Issue #490, opened 2020-03-30
Cursor disappears after indent/outdent


raineorshine:
### Indent

- a
- b [cursor]

Indent (tab key) immediately after typing `b` (without moving the cursor or causing a `blur` in any way). The cursor will disappear instead of staying on `a.b`.

### Outdent

- a
  - b [cursor]

Outdent (shift + tab) immediately after typing `b` (without moving the cursor or causing a `blur` in any way). The cursor will disappear instead of staying on `b`.

PR #508, opened 2020-04-04
Fixed cursor disappear issue after indent and outdent.

umair-khanzada:
Ref #490

### DIFF
--- a/src/action-creators/indent.js
+++ b/src/action-creators/indent.js
@@ -58,5 +58,7 @@ export const indent = () => dispatch => {
       newPath: cursorNew,
       offset
     })
+
+    dispatch({ type: 'setCursor', thoughtsRanked: cursorNew, offset })
   }
 }
--- a/src/action-creators/outdent.js
+++ b/src/action-creators/outdent.js
@@ -50,5 +50,7 @@ export const outdent = () => dispatch => {
       newPath: cursorNew,
       offset
     })
+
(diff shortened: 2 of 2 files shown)


raineorshine:
Hi! I think the fix actually belongs in `existingThoughtMove`. The question is, since the new cursor depends on `editing`, [why is `editing` false in this particular case](https://github.com/umair-khanzada/em/blob/14ed500535600ada0370604768da01f6a70c439d/src/reducers/existingThoughtMove.js#L186)?

raineorshine:
I thought about it more, and I think it makes sense to dispatch a separate action. The default behavior of `existingThoughtMove` should not change the cursor unless the cursor is already on it. Drag and drop is a special case, where the cursor should follow the drop.

umair-khanzada:
But what I am thinking it is not a good practice to dispatch two events from action-creator, we should move this in exec? what do you think?

umair-khanzada:
> ~I thought about it more, and I think it makes sense to dispatch a separate action. The default behavior of `existingThoughtMove` should not change the cursor unless the cursor is already on it. Drag and drop is a special case, where the cursor should follow the drop.~
> 
> Never mind, this isn’t drag and drop. I would expect `editing` to be true.

So the issue is `equalPath` function returning `false` because of `cursorBeforeEdit.value` is an empty string while the `oldPath.value` is different in our case it is `'b'` that's why the editing flag is false

raineorshine:
Replaced by #512

Issue #456, opened 2020-03-24
Autohide does not hide distant relatives 


raineorshine:
`distance-from-cursor` currently only measures depth. With tables and pinned open contexts, it is possible to have both visible and hidden thoughts rendered at the same depth. `distance-from-cursor` should be replaced by inline styles that hide the correct thoughts. 

1. `b` should be dimmed, like `a`:
- a:
  - b
- c (cursor)
  - d
    - e

2. `b` should be hidden:
- a:
  - b
- c
  - d (cursor)
    - e

PR #510, opened 2020-04-04
Autohide does not hide distant relatives fix

shresthabijay:
refers to #456


shresthabijay:
@raineorshine This issue was really tricky because `distance-from-cursor` implementation doesn't work quite well with table view and pinned context. They introduce edge cases . 

> b should be dimmed, like a:
>- a:
>   - b
>- c (cursor)
>   - d
>     - e
>

The reason for this issue is that the current implantation doesn't consider the condition where a node which is neither ancestor or descendant of cursor can have zero `distance-from-cursor`.  As you can see in the example wrapper of thoughts at level of **b** will have zero `distance-from-cursor`. So it won't be dim and will have full opacity. 

>b should be hidden:
>- a:
>   - b
>- c
>   - d (cursor)
>     - e
>

Unlike normal view where there is only one expanded thought in a context at a time, table view node has all its children expand and render their respective `subthoughts` all at once. So if  we select any grandchildren of the main table view node, all it's children will disappear but the grandchildren will still show up. It is because the children will have  2 cursor distance (hide) from the main table view node while have 1 cursor distance (dim). 

**Solution:** With this PR I have written logic to find these specific edge cases and used it to set distance appropriately. Please review the changes and let me know what you think. Thank you!

Issue #504, opened 2020-04-03
"s" character added multiple times to context


raineorshine:
## Steps to Reproduce

Type "s" and hit Enter.

## Current Behavior
"s" has superscript "2" indicating that it belongs to two contexts. 

- Every time you hit Enter the superscript of "s" will increment. 
- Only an issue for "s" (???)

You can call `getThought('s')` in the console to see that is has been added multiple times to the same context.

```js
{
  value: "s",
  contexts: [
    { context: ["__ROOT__"], rank: 0 },
    { context: ["__ROOT__"], rank: 1 }
  ],
  created: "2020-04-03T15:07:06.949Z",
  lastUpdated: "2020-04-03T15:07:06.956Z",
}
```

## Expected Behavior
"s" should only be in a single context.

## Notes
Not a recent regression; goes back to at least v95.0.0.

PR #513, opened 2020-04-04
Multiple context for 's' fix

shresthabijay:
refers to #504

### DIFF
--- a/src/util/hashThought.js
+++ b/src/util/hashThought.js
@@ -14,6 +14,10 @@ const strip = s => {
   return stripped.length > 0 ? stripped : s
 }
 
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine  `HashThought`  made every value singular before hashing.  The issue was when we `singularize` character **s** it would just be an empty string . So it caused  "s" character to have same no of context as empty thoughts in the entire tree. Hitting **Enter** will create new thought every time and context for **s** increased with it. 

I made an exception to `singularize` function to skip character **s** . Please review it!

raineorshine:
Ha! How amusing. Thanks :)

PR #515, opened 2020-04-05
Remove .js imports

raineorshine:


raineorshine:
Turns out 9675c1bbb22cdf6dd14f19dca921f6bea48a324b works just fine. The cause was #516.

Issue #342, opened 2020-02-17
Divider: Should set cursor after creation


raineorshine:
When the user types `---` (three dashes), the thought is converted into a horizontal [`Divider`](https://github.com/cybersemics/em/blob/dev/src/components/Divider.js). Doing a [search for "isDivider"](https://github.com/cybersemics/em/search?q=isDivider&unscoped_q=isDivider) provides a good survey of all of the divider business logic.

## Steps to Reproduce
1. Hit Enter and type "A" to create a new thought
2. Hit Enter to create a new empty thought
3. Type `---` to create a divider

## Current Behavior

Divider is not highlighted (cursor).

## Expected Behavior

The cursor should be on the divider thought and the divider should be highlighted.

parthpnp:
Can not reproduce the issue. Can you please verify.

raineorshine:
@parthpnp Thanks! Turns out it is only an issue when the divider is not the first child. I updated the steps to reproduce accordingly.

raineorshine:
This was supposed to be fixed as described in https://github.com/cybersemics/em/pull/522#discussion_r404212502, but it is happening again.

PR #502, opened 2020-04-03
Prevent moving thought into divider

incora-dev:
#390

PR #522, opened 2020-04-06
Fix cursor after divider creation

incora-dev:
#342

### DIFF
--- a/src/action-creators/newThoughtAtCursor.js
+++ b/src/action-creators/newThoughtAtCursor.js
@@ -19,14 +19,18 @@ import {
   pathToContext,
   perma,
   splitChain,
-} from '../util'
+  isDivider
+} from '../util.js'
 
 export const newThoughtAtCursor = () => dispatch => {
 
   const { cursor, contextViews } = store.getState()
 
   let value = '' // eslint-disable-line fp/no-let
-  const offset = window.getSelection().focusOffset
+  const offset = isDivider(headValue(cursor))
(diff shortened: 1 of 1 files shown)


raineorshine:
> Make sure to specify the issue number in the PR description, as specified in the [Contributing Guidelines](https://github.com/cybersemics/em/blob/dev/CONTRIBUTING.md).

https://github.com/cybersemics/em/pull/502#issuecomment-608439241

@incora-dev Please follow the PR naming protocol. You have been informed of this requirement twice already. The next time will involve an escalation.

src/action-creators/newThoughtAtCursor.js
```diff
@@ -19,14 +19,18 @@ import {
...
 
 export const newThoughtAtCursor = () => dispatch => {
 
   const { cursor, contextViews } = store.getState()
 
   let value = '' // eslint-disable-line fp/no-let
-  const offset = window.getSelection().focusOffset
+  const offset = isDivider(headValue(cursor))
+    ? headValue(cursor).length
+    : window.getSelection().focusOffset
```
raineorshine, line 32: While this does prevent the issue from occurring, it doesn't go to the root of the problem. You have to ask, why is the offset wrong to begin with? It looks like the browser selection is still active when the divider is created. If you clear the browser selection when the divider is created, the problem is solved without any custom logic in `newThoughtAtCursor`. Fixed in e8a71d953df437646615257984b2988ce3612d33.

Issue #483, opened 2020-03-29
Convert util functions to selectors


raineorshine:
Pure functions that slice or derive data from state should take an initial argument of `state` and be stored in `/selectors`.

raineorshine:
Should direct slices be separated from computed data?

raineorshine:
@umair-khanzada Would you like to take on this larger refactoring task? It involves converting any `util` functions that compute derived data from `state` with no side effects into selector functions that take `state` as the first variable.

The basic steps for each selector are:
1. `git mv src/util/something.js src/selectors/something.js`.
1. `export default`.
1. Change signature to take `state` as first argument.
1. Update `import` statements and update callers to match new signature.

It is pretty rote, just requires some care in step 4. Each selector refactor should go in a separate commit so we can more easily trace regressions. See any of the commits in #507 for examples.

Here is a list of the util functions that need to be converted to selectors:

- [decodeThoughtsUrl](https://github.com/cybersemics/em/blob/dev/src/util/decodeThoughtsUrl.js)
- [exists](https://github.com/cybersemics/em/blob/dev/src/util/exists.js)
- [expandThoughts](https://github.com/cybersemics/em/blob/dev/src/util/expandThoughts.js)
- [exportContext](https://github.com/cybersemics/em/blob/dev/src/util/exportContext.js)
- [getChildPath](https://github.com/cybersemics/em/blob/dev/src/util/getChildPath.js)
- [getContexts](https://github.com/cybersemics/em/blob/dev/src/util/getContexts.js)
- [getContextsSortedAndRanked](https://github.com/cybersemics/em/blob/dev/src/util/getContextsSortedAndRanked.js)
- [getDescendants](https://github.com/cybersemics/em/blob/dev/src/util/getDescendants.js)
- [getNextRank](https://github.com/cybersemics/em/blob/dev/src/util/getNextRank.js)
- [getNgrams](https://github.com/cybersemics/em/blob/dev/src/util/getNgrams.js)
- [getPrevRank](https://github.com/cybersemics/em/blob/dev/src/util/getPrevRank.js)
- [getRankAfter](https://github.com/cybersemics/em/blob/dev/src/util/getRankAfter.js)
- [getRankBefore](https://github.com/cybersemics/em/blob/dev/src/util/getRankBefore.js)
- [getSetting](https://github.com/cybersemics/em/blob/dev/src/util/getSetting.js)
[...]

PR #524, opened 2020-04-07
Convert util functions to selectors

umair-khanzada:
#483


src/util/exists.js
```diff
@@ -1,8 +0,0 @@
-import { store } from '../store'
-
-// util
-import { getThought } from './getThought'
-
-/** Returns true if the head of the given context exists in the thoughtIndex */
-export const exists = (value, thoughtIndex = store.getState().thoughtIndex) =>
-  value != null && !!getThought(value, thoughtIndex)
```
raineorshine, line 8: Use `git mv` so that git can track the move.

src/util/dataIntegrityCheck.js
```diff
@@ -53,7 +55,7 @@ export const dataIntegrityCheck = path => {
 
   // recreate thoughts missing in thoughtIndex
   for (const child of (contextIndex[encoded] || [])) { // eslint-disable-line fp/no-loops,fp/no-let
-    const childExists = exists(child.value, thoughtIndex)
+    const childExists = exists(thoughtIndex, child.value)
```
raineorshine, line 58: Should selectors take the full state, or a slice?

@anmolarora1 We had talked about this and I think decided on full state, but do you still think that's the right thing? It seems odd to pass the full state when it only needs a slice.

src/components/SearchSubthoughts.js
```diff
@@ -77,7 +79,7 @@ const SearchSubthoughts = ({ search, searchLimit = DEFAULT_SEARCH_LIMIT, dispatc
     // must go into DOM to modify the parent li classname since we do not want the li to re-render
     ref={onRef}
   >
-    {!exists(search) ? <NewThought path={[]} label={`Create "${search}"`} value={search} type='button' /> : null}
+    {!exists(store.getState().thoughtIndex, search) ? <NewThought path={[]} label={`Create "${search}"`} value={search} type='button' /> : null}
```
raineorshine, line 82: Since `thoughtIndex` is passed from `mapStateToProps`, you can use that instead of accessing state directly.

src/util/dataIntegrityCheck.js
raineorshine: @umair-khanzada I discussed with @anmolarora1, and we would like to pass the whole `state` object to each selector (or least in the same shape as the full state, e.g. `{ thoughtIndex }` rather than `thoughtIndex`).

src/components/Superscript.js
```diff
@@ -15,10 +14,13 @@ import {
...
 } from '../util'
 
+// selectors
+import { exists } from '../selectors'
+
 const mapStateToProps = ({ contextViews, cursor, cursorBeforeEdit, modalData, showModal }, props) => {
 
   // track the transcendental identifier if editing
-  const editing = equalArrays(pathToContext(cursorBeforeEdit || []), pathToContext(props.thoughtsRanked || [])) && exists(headValue(cursor || []))
+  const editing = equalArrays(pathToContext(cursorBeforeEdit || []), pathToContext(props.thoughtsRanked || [])) && exists(store.getState().thoughtIndex, headValue(cursor || []))
```
anmolarora1, line 23: store.getState() is similar to the first argument of mapStateToProps, which is being de-structured. So you can use that here and pass state to `exists` function

src/components/Superscript.js
```diff
@@ -42,7 +44,7 @@ const mapStateToProps = ({ contextViews, cursor, cursorBeforeEdit, modalData, sh
     // thoughtRaw is the head that is removed when showContexts is true
     thoughtRaw: props.showContexts ? head(props.thoughtsRanked) : head(thoughtsRankedLive),
     empty: thoughtsLive.length > 0 ? head(thoughtsLive).length === 0 : true, // ensure re-render when thought becomes empty
-    numContexts: exists(head(thoughtsLive)) && getContexts(head(thoughtsLive)).length,
+    numContexts: exists(store.getState().thoughtIndex, head(thoughtsLive)) && getContexts(head(thoughtsLive)).length,
```
anmolarora1, line 47: Replace with `exists(state, head(thoughtsLive))`

src/reducers/existingThoughtDelete.js
```diff
@@ -23,7 +25,7 @@ import getThoughtsRanked from '../selectors/getThoughtsRanked'
 export default (state, { context, thoughtRanked, showContexts }) => {
 
   const { value, rank } = thoughtRanked
-  if (!exists(value, state.thoughtIndex)) return
+  if (!exists(state.thoughtIndex, value)) return
```
anmolarora1, line 28: Replace with `exists(state, value)`

src/selectors/index.js
```diff
@@ -0,0 +1 @@
+export { default as exists } from './exists'
```
anmolarora1, line 1: I'm not sure if there was a discussion to follow this pattern of having a map file for selectors. 
@raineorshine, what's your opinion?

raineorshine: I'm agnostic. An index creates cleaner bulk imports, but involves manually maintaining a separate file and it makes refactoring individual imports less automatable. 

@umair-khanzada originally suggested the index file, so perhaps he has more reasons he can offer if he has a strong opinion.

src/util/dataIntegrityCheck.js
umair-khanzada: Got it

src/selectors/index.js
umair-khanzada: Yes, I had suggested this, there is a couple of benefits of adopting this technic
1. React and other most popular lib using the same structure ie:(`{Fragment, Component, PureComponent, etc}`) 
2. Create a public interface for accessing a file
3. Import more than one selector in a single statement ie:(`{S1, S2} from 'selectors'`)
4. We are doing the same thing for util and sometimes it is confusing because both have the same name

src/components/ThoughtAnnotation.js
```diff
@@ -20,12 +19,16 @@ import {
   pathToContext,
   unroot,
 } from '../util'
+import { store } from '../store'
```
anmolarora1, line 22: We don't need to import store here. The first argument of mapStateToProps is `state`, which is same as store.getState()

src/components/ThoughtAnnotation.js
```diff
@@ -20,12 +19,16 @@ import {
...
 
 // components
 import HomeLink from './HomeLink'
 import StaticSuperscript from './StaticSuperscript'
 import ContextBreadcrumbs from './ContextBreadcrumbs'
 
+// selectors
+import { decodeThoughtsUrl } from '../selectors'
+
 const mapStateToProps = ({ cursor, cursorBeforeEdit, focusOffset, invalidState, editingValue }, props) => {
```
anmolarora1, line 32: This can be replaced with
```
const mapStateToProps = (state, props) => {
    const { cursor, cursorBeforeEdit, focusOffset, invalidState, editingValue } = state
```

src/components/ThoughtAnnotation.js
```diff
@@ -94,7 +97,7 @@ const ThoughtAnnotation = ({ dark, thoughtsRanked, showContexts, showContextBrea
             // eslint-disable-next-line no-undef
               ? <a href={(!url.startsWith('http:') && !url.startsWith('https:') && !url.startsWith('localhost:') ? 'https://' : '') + url} rel="noopener noreferrer" target='_blank' className='external-link' onClick={e => {
                 if (url.startsWith(window.location.origin)) {
-                  const { thoughtsRanked, contextViews } = decodeThoughtsUrl(url.slice(window.location.origin.length))
+                  const { thoughtsRanked, contextViews } = decodeThoughtsUrl(store.getState(), url.slice(window.location.origin.length))
```
anmolarora1, line 100: `const { thoughtsRanked, contextViews } = decodeThoughtsUrl(state, url.slice(window.location.origin.length))`

src/util.js
```diff
@@ -15,7 +15,7 @@ import { contextChainToPath } from './util/contextChainToPath'
 import { contextOf } from './util/contextOf'
 import { dataIntegrityCheck } from './util/dataIntegrityCheck'
 import { decodeCharacterEntities } from './util/decodeCharacterEntities'
-import { decodeThoughtsUrl } from './util/decodeThoughtsUrl'
+// import { decodeThoughtsUrl } from './util/decodeThoughtsUrl'
```
anmolarora1, line 18: I think we should delete these now

src/util/updateUrlHistory.js
```diff
@@ -23,7 +25,7 @@ export const updateUrlHistory = (thoughtsRanked = RANKED_ROOT, { replace, though
     return
   }
 
-  const decoded = decodeThoughtsUrl(window.location.pathname, thoughtIndex, contextIndex)
+  const decoded = decodeThoughtsUrl({ thoughtIndex, contextIndex }, window.location.pathname)
```
anmolarora1, line 28: Maybe we should pass the state here too, just like we're doing in every other call to this selector. 
Because this might break or need to be updated in case you update your selector to use more items from the state in the future.

raineorshine: Yes, although I have instructed Umair to convert one function per commit, so this should show up in a separate commit.

src/components/ThoughtAnnotation.js
umair-khanzada: @anmolarora1 Actually we need store here, I have replaced store call in `mapStateToProps` but we need a store in `ThoughtAnnotation`.

src/util.js
umair-khanzada: Deleted.

src/components/Superscript.js
```diff
@@ -20,7 +20,7 @@ import { exists } from '../selectors'
 const mapStateToProps = ({ contextViews, cursor, cursorBeforeEdit, modalData, showModal }, props) => {
 
   // track the transcendental identifier if editing
-  const editing = equalArrays(pathToContext(cursorBeforeEdit || []), pathToContext(props.thoughtsRanked || [])) && exists(store.getState().thoughtIndex, headValue(cursor || []))
+  const editing = equalArrays(pathToContext(cursorBeforeEdit || []), pathToContext(props.thoughtsRanked || [])) && exists(store.getState(), headValue(cursor || []))
```
anmolarora1, line 21: Please replace `exists(store.getState())` with `exists(state)`. Use state from mapStateToProps' arguments

src/components/Superscript.js
```diff
@@ -44,7 +44,7 @@ const mapStateToProps = ({ contextViews, cursor, cursorBeforeEdit, modalData, sh
     // thoughtRaw is the head that is removed when showContexts is true
     thoughtRaw: props.showContexts ? head(props.thoughtsRanked) : head(thoughtsRankedLive),
     empty: thoughtsLive.length > 0 ? head(thoughtsLive).length === 0 : true, // ensure re-render when thought becomes empty
-    numContexts: exists(store.getState().thoughtIndex, head(thoughtsLive)) && getContexts(head(thoughtsLive)).length,
+    numContexts: exists(store.getState(), head(thoughtsLive)) && getContexts(head(thoughtsLive)).length,
```
anmolarora1: Please replace `exists(store.getState())` with `exists(state)`. Use state from mapStateToProps' arguments

src/components/Superscript.js
```diff
@@ -44,7 +43,7 @@ const mapStateToProps = ({ contextViews, cursor, cursorBeforeEdit, modalData, sh
     // thoughtRaw is the head that is removed when showContexts is true
     thoughtRaw: props.showContexts ? head(props.thoughtsRanked) : head(thoughtsRankedLive),
     empty: thoughtsLive.length > 0 ? head(thoughtsLive).length === 0 : true, // ensure re-render when thought becomes empty
-    numContexts: exists(store.getState(), head(thoughtsLive)) && getContexts(head(thoughtsLive)).length,
+    numContexts: exists(store.getState(), head(thoughtsLive)) && getContexts(store.getState(), head(thoughtsLive)).length,
```
anmolarora1, line 45: Replace with `getContexts(state)`. Use state from mapStateToProps' arguments

src/util/deleteThought.js
```diff
@@ -65,18 +69,19 @@ export const deleteThought = () => {
...
+    !meta(state, context.concat(thoughtRanked.value)).hidden
   )
 
+  // must call store.getState() to use the new state after existingThoughtDelete
   const next = perma(() => showContexts
-    ? unroot(getContextsSortedAndRanked(headValue(contextOf(path))))[0]
+    ? unroot(getContextsSortedAndRanked(state, headValue(contextOf(path))))[0]
     // get first visible thought
-    : (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(context)
+    : (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(store.getState(), context)
```
anmolarora1, line 84: Is there any specific reason for using `store.getState()` here instead of `state`?

src/util/updateUrlHistory.js
```diff
@@ -23,7 +24,7 @@ export const updateUrlHistory = (thoughtsRanked = RANKED_ROOT, { replace, though
     return
   }
 
-  const decoded = decodeThoughtsUrl(window.location.pathname, thoughtIndex, contextIndex)
+  const decoded = decodeThoughtsUrl(store.getState(), window.location.pathname)
```
anmolarora1, line 27: I think we can use the same `state` instance on lines 27 and 37 instead of calling `store.getState()` twice. Please correct me if I'm wrong.

src/components/Editable.js
```diff
@@ -244,10 +243,10 @@ const Editable = ({ isEditing, thoughtsRanked, contextChain, cursorOffset, showC
       return
     }
 
-    const newNumContext = getContexts(newValue).length
+    const newNumContext = getContexts(store.getState(), newValue).length
     const isNewValueURL = isURL(newValue)
 
-    const contextLengthChange = newNumContext > 0 || newNumContext !== getContexts(oldValueRef.current).length - 1
+    const contextLengthChange = newNumContext > 0 || newNumContext !== getContexts(store.getState(), oldValueRef.current).length - 1
```
anmolarora1, line 249: There are multiple places in this function where store.getState() is being used. We can retrieve it once and use the same instance all over.

src/shortcuts/pinOpen.js
```diff
@@ -22,9 +24,11 @@ export default {
   keyboard: { key: 'p', alt: true },
   svg: Icon,
   exec: () => {
-    const { cursor } = store.getState()
+    const state = store.getState()
+    const cursor = state.cursor
```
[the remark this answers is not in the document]
raineorshine: No point destructuring if `state` is needed to pass to `attribute`.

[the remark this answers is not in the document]
anmolarora1: Just to be clear, this is what I mean here.
`const {cursor} = state`

raineorshine:
> It'd really help if you can use `git mv` and push the changes for all such instances.

Yes, this is a requirement.

src/shortcuts/pinOpen.js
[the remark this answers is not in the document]
raineorshine: Ah, I see. Yes I agree.

src/selectors/expandThoughts.js
```diff
@@ -30,7 +31,7 @@ import getThoughtsRanked from '../selectors/getThoughtsRanked'
     ...
   }
 */
-export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews = {}, contextChain = [], { depth = 0 } = {}) => {
+export default ({ thoughtIndex, contextIndex, contextViews }, path, contextChain = [], { depth = 0 } = {}) => {
```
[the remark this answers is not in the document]
umair-khanzada: I don't think we need set default, because of `initialState` contextViews is set empty object in initialState

src/selectors/decodeThoughtsUrl.js
```diff
@@ -19,7 +24,7 @@ export default ({ thoughtIndex, contextIndex }, pathname) => {
       [hashContext(pathUnranked.slice(0, i + 1))]: true
     }) : accum,
   {})
-  const thoughtsRanked = rankThoughtsFirstMatch(pathUnranked, { state: { thoughtIndex, contextIndex, contextViews } })
+  const thoughtsRanked = rankThoughtsFirstMatch(state, pathUnranked, { state: { thoughtIndex, contextIndex, contextViews } })
```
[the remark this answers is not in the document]
umair-khanzada: I have made this more cool :)

src/selectors/getSetting.js
```diff
@@ -11,6 +11,7 @@ import {
 import getThoughtsRanked from '../selectors/getThoughtsRanked'
 
 /** Returns subthoughts of /em/Settings/...context, not including meta subthoughts */
-export const getSetting = (context, state = store.getState(), depth = 0) =>
+// TODO: I have passes state everywhere from getSetting is calling but still getting undefined in some cases so the temp hack is store.getState()
+export default (state = store.getState(), context, depth = 0) =>
```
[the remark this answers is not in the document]
umair-khanzada: Sure, I'll check and let you update.

src/components/__tests__/ThoughtDeleteEmpty.js
```diff
@@ -19,7 +23,7 @@ it('delete empty thought', async () => {
   jest.runAllTimers()
 
   // state
-  const subthoughts = getThoughts([ROOT_TOKEN])
+  const subthoughts = getThoughts(store.getState(), [ROOT_TOKEN])
```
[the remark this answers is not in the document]
umair-khanzada: Yes, you are right we need to mock the store

src/shortcuts/proseView.js
```diff
@@ -26,16 +28,16 @@ export default {
...
-    const { cursor } = store.getState()
-    if (cursor) {
+    const state = store.getState()
+    const { cursor } = state
 
-      // if the cursor is on a leaf, activate prose view for the parent
-      const path = cursor.length > 1 && getThoughtsRanked(cursor).length === 0
-        ? contextOf(cursor)
-        : cursor
+    // if the cursor is on a leaf, activate prose view for the parent
```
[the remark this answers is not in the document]
umair-khanzada: Didn't get you! Did you mean `cursor  && cursor.length` ?

umair-khanzada:
> > It'd really help if you can use `git mv` and push the changes for all such instances.
> 
> Yes, this is a requirement.

I have ran `mv` command for each selector

umair-khanzada:
I have updated all the changes that suggested, I also have commented on some please check.

raineorshine:
> > > It'd really help if you can use `git mv` and push the changes for all such instances.
> > 
> > 
> > Yes, this is a requirement.
> 
> I have ran `mv` command for each selector

There are several files that have not been `git mv` properly, e.g. `isTutorial.js`.

Note also that you have a lot of merge conflicts to resolve, so I suggest doing a rebase on `dev` as soon as possible.

src/selectors/expandThoughts.js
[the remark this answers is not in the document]
anmolarora1: Not every selector is being called with the complete state. Though it might work in this case, I think unless the functions are type-safe, it's always good to have a fallback.

src/components/Subthoughts.js
```diff
@@ -213,7 +208,9 @@ const evalCode = ({ thoughtsRanked }) => {
 
   const state = store.getState()
   const { thoughtIndex } = state
-  const thought = getThought(headValue(thoughtsRanked), 1)
+  // TODO: This getThought call looking bit ambitious to me I am commenting the previous statement please check this.
+  // const thought = getThought(headValue(thoughtsRanked), 1)
+  const thought = getThought(state, headValue(thoughtsRanked))
```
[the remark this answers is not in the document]
raineorshine: Correct

raineorshine:
Resumed in #640.

Issue #527, opened 2020-04-08
Tree nodes re-render when cursor hovers over toolbar shortcut icons ( setToolbarOverlay )


shresthabijay:
**Issue:**
When hovered over any one of the shortcut icons from the toolbar, an overlay describing the shortcut is rendered. But the event  is causing whole tree nodes to re-render. This side effect causes unnecessary re-renders and hinders application performance.

**Expected behaviour**
Just rendering an overlay for shortcuts in a toolbar , should not cause whole tree to re-render. I strongly believe that this event should only re-render toolbar. It doesn't cause any state changes that is consumed by `Thoughts`.

_Hovering over shortcuts icons_

![captured (3)](https://user-images.githubusercontent.com/25879100/78765050-270ec480-79a7-11ea-8709-dc15d5dc5140.gif)

raineorshine:
I agree! Good catch.

(How do you get the render boxes to appear???)

shresthabijay:
> (How do you get the render boxes to appear???)

Go to the react dev tool component. Click at the top right setting icon and check highlight update.
It will highlight re-rendered components. Colors vary with frequency of re-renders.
  
![higlight-update](https://user-images.githubusercontent.com/25879100/78801838-e0d45800-79dc-11ea-9d94-762adfeda41b.png)

anmolarora1:
@raineorshine Adding to what @shresthabijay pointed out, I also noticed this issue sometime back, and many such issues can be caught using dev tools and profiler. 
Afaik, this one is happening because of `rootThoughts` is being re-calculated on every render and even if its value doesn't change, its reference does. 
![Screen Shot 2020-04-08 at 6 16 34 PM](https://user-images.githubusercontent.com/7765764/78808528-d3f83a00-79c5-11ea-997f-39e50d956d1d.png)

shresthabijay:
@anmolarora1 good catch. @raineorshine This is one of the few places in codebase where referential inequality in `mapStateToProps` is causing unnecessary re-renders. I believe I suggested you about this when I first reviewed the em codebase.

It is better to use `mapStateToProps`  to return only state from the store that visual component depends on and do all the necessary calculations inside render function. We only use such calculations inside `mapStateToProps` if these calculation result are entirely independent to other state and affects UI. If they are dependent on any one of the state, we return those required states from the `mapStateToProps` and then do calculation on render function. This will eliminate referential inequality and keeps `mapStateToProps` simple and easy to follow. 

[...]

PR #530, opened 2020-04-08
Content: Move getRootThoughts to render function

shresthabijay:
refers to #527


shresthabijay:
@raineorshine I moved calculation of `getRootThoughts` from `mapStateToProps` to render function to avoid referential inequality. With this change tree stopped re-rendering from the root. But cursor and all its descendants are being re-rendered. I am certain we can refactor it.

raineorshine:
@shresthabijay This causes a regression in which the Content is not re-rendered when root thoughts change. Reverted in d419f7d06c23301ab65a16b6341dedfdf9219cf6.

## Steps to Reproduce
1. Create a thought in ROOT
2. Delete the thought in ROOT

Issue #276, opened 2020-01-06
Cursor disappears after moving parent


raineorshine:
`existingThoughtMove` causes the cursor to disappear in some circumstances.

## Steps to Reproduce

Given the following structure and cursor position:

- A
  - C
    - y
  - B
    - x (cursor)

Drag "B" above "C".

## Current Behavior

The thought cursor disappears.

## Expected Behavior

The thought cursor should stay on the same element even after it is re-ordered or moved.

- A
  - B
    - x (cursor)
  - C
    - y

## Notes

This appears to be caused by the new rank of "B" not being updated in the `cursor` state variable.

## Requirements

The cursor should similarly be preserved when dragging to a parent or uncle.

sparklingW:
I figured out that problem occurs not only because of the cursor(state variable) doesn't update, also it may occur because newPath contain elements except last one, so if we moved our element above and our cursor was on a child element of C(took from your example above), newPath will contain only 'C' element.

incora-dev:
@raineorshine  Estimate: 2.5h

incora-dev:
@raineorshine 
Spent a lot of time I have found a possible solution. The possible problem is the wrong calculation for `isEditing` variable in Thought.js(line 94). When I try to calculate this variable without rank (just using value), it works partially, but I don't like this solution. I don't have any idea about other solutions for now.

My temporary solution(without good code style):
```
  // const isEditing = equalPath(cursorBeforeEdit, thoughtsResolved)
  const isEditing = thoughtsResolved && cursorBeforeEdit.every((e, i) => {
    return e.value === (thoughtsResolved[i] && thoughtsResolved[i].value)
  })
```

raineorshine:
Great! Thanks for identifying the cause of the problem.

If you ignore `rank` then `isEditing` will incorrectly return `true` when editing a duplicate thought in the same context.

Why is `cursorBeforeEdit` still referring to the old cursor? I would think that by the time the thoughts are re-rendered, the `existingThoughtMove` will have already updated it.

I think the problem is here:
https://github.com/cybersemics/em/blob/dev/src/reducers/existingThoughtMove.js#L186-L187

It updates the cursor when the editing thought itself is moved, but not when its ancestor is moved. We may need to check [`isEditingPath`](https://github.com/cybersemics/em/blob/dev/src/components/Subthoughts.js#L90) like the Subthoughts component.

PR #532, opened 2020-04-09
Fix: cursor disappears after moving parent #276

incora-dev:
It works, but I am not exactly sure about the rest of the functionality. Cursor always keep old state because `editing` always was false. We tried to check is path equal for cursor and oldPath (oldPath has 2 child and cursor has 3 when we move parent)

### DIFF
--- a/src/reducers/existingThoughtMove.js
+++ b/src/reducers/existingThoughtMove.js
@@ -180,11 +180,18 @@ export default (state, { oldPath, newPath, offset }) => {
     }
   })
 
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks! Unfortunately, this breaks in other cases.

e.g.

- a
  - b [cursor]

Move `b` above `a`, and now the cursor disappears.

As [mentioned](https://github.com/cybersemics/em/issues/276), I think you want to broaden `editing` to `editingPath`.

Issue #429, opened 2020-03-21
Alert drag and drop


raineorshine:
Dispatch an `alert` when the user taps and holds a thought and/or drags a thought with text: “Drag and drop to move thought”.

Hide the `alert` on release.

Issue #428, opened 2020-03-21
Style drag bullet on tap and hold


raineorshine:
## Steps to Reproduce
Tap and hold a thought

## Current Behavior
Thought is selected but bullet does not turn lightblue until touchmove. 

## Expected Behavior
Bullet should turn blue after tap and hold delay even without any movement

raineorshine:
Updated PR: #1388

PR #533, opened 2020-04-10
Drag and drop issues

pushkov-fedor:
Issues #428 and #429. Looks like ReactDnD event functions beginDrag() and isDragging() fire up after click and move. Used mouse events up and down to fixed this. Inside them wrapped logic in setTimeout() to prevent turning on lightblue on click


src/components/Thought.js
```diff
@@ -435,7 +446,20 @@ const ThoughtContainer = ({
       dragPreview(getEmptyImage())
     }
   }}>
-    <div className='thought-container'>
+    <div className='thought-container'
+      onMouseDown={e => {
```
raineorshine, line 450: You'll need to handle `onTouchStart` and `onTouchEnd` so that this also works on mobile.

src/components/Thought.js
```diff
@@ -435,7 +446,20 @@ const ThoughtContainer = ({
...
+        // Used timeout to prevent turning lightblue on click
+        const timeoutId = setTimeout(() => {
+          store.dispatch({ type: 'dragInProgress', value: true, draggedThoughtsRanked: thoughtsRankedLive })
+          alert('Drag and drop to move thought')
+        }, TIMEOUT_BEFORE_DRAG)
+        store.dispatch({ type: 'dragTimeoutId', timeoutId })
```
raineorshine, line 456: I think we can store `dragTimeoutId` locally rather than add an additional global reducer. I'm not sure whether closure, `useState` or `useEffect` is the right solution. 

@shresthabijay What do you think?

src/components/Thought.js
pushkov-fedor: I've done research today about drag-and-drop after click/tap and move on text. It looks like we need in dragging process **disable** ContentEditable component and apply style with **user-select: none** to prevent text selection while moving. But after that, event dragEnd fires up after cursor left ContentEditable border, and dragging stops. Need some time to figure out what's happening.

We cant start dragging from click on text because of text selection so we have to disable it. But by disabling it, we prevent text selection, when user wants to select part of the text by mouse. Need to come up with idea how to deal with it. What do you think? Or correct me, if i'm wrong somewhere. Also want to ask, what potential issues or pitfalls i could come across by  disabling ContentEditable?

raineorshine: Thanks, Fedor, you helped me realize I missed an important point. Thoughts can be moved by dragging the text of the thought **on mobile only**. On desktop, clicking and dragging text should do the default browser behavior of selecting text within the editable.

So the dragging behavior is great as-is, and you just need a find a way to enable the blue bullet on mobile when the thought is tapped. I think the key will be to re-render the cursor overlay bullet without re-rendering the ContentEditable. You may have to move the cursor overlay bullet to its own component and `connect` it to redux.

Hopefully that is much easier for you! Thanks for being willing to investigate and let me know I'm missing something or you have any other questions.

src/components/Thought.js
shresthabijay: @raineorshine Yes it's better to store `dragTimeoutId` locally. I recommend creating a custom hook that manages long press event.

```javascript
// useLongPress.js custom hook to manage long press
import { useState, useEffect, useCallback, useRef } from 'react'

export default function useLongPress(onLongPressStart = () => {}, onLongPressEnd = () => {}, ms = 250) {
  const [startLongPress, setStartLongPress] = useState(false)
  const timerIdRef = useRef()

  useEffect(() => {
    if (startLongPress) timerIdRef.current = setTimeout(onLongPressStart, ms)
    else clearTimeout(timerIdRef.current)

    return () => clearTimeout(timerIdRef.current)
  }, [startLongPress])

  const start = useCallback(() => {
    setStartLongPress(true)
  }, [])

  const stop = useCallback(() => {
    setStartLongPress(false)
    onLongPressEnd()
  }, [])

  return {
    onMouseDown: start,
    onMouseUp: stop,
    onTouchStart: start,
    onTouchEnd: stop,
  }
}

```

Now we use this custom hook wherever we need long press event.

```javascript

// inside ThoughtContainer component 
 const onLongPressStart = () => {
    store.dispatch({ type: 'dragInProgress', value: true, draggedThoughtsRanked: thoughtsRankedLive })
    alert('Drag and drop to move thought')
  }

  const onLongPressEnd = () => {
    store.dispatch({ type: 'dragInProgress', value: false })
    alert(null)
  }

  const longPressHandlerProps = useLongPress(onLongPressStart, onLongPressEnd, TIMEOUT_BEFORE_DRAG)

  // pass long press handler props to the target div
  return (
    <div className='thought-container' {...longPressHandlerProps}></div>
   )
```
This way we can abstract long press logic properly in separate file without needing to handle timer id inside `ThoughtContainer`.

pushkov-fedor: Yeah, this looks great

pushkov-fedor:
Did some work. Please point me out, if i've missed something. Also want to discuss instant focus and blur events on text after dragging it causing keyboard appearance and disappearance.

raineorshine:
> Did some work. Please point me out, if i've missed something.

Excellent work! Looking really good.

1. On desktop, I notice a difference between the `onLongPress` region and the actual drag region. I think we want to unify them. I can click and hold on the bullet, to the left of the bullet, or to the right of the text to drag a thought. When I click and drag on the text itself, it selects the text and does not drag. This behavior is perfect.

`onLongPress` only gets triggered when the bullet itself is long pressed, not the space to the left of the bullet or to the right of the text. I think the blue bullet should be shown when any valid drag area is clicked and held.

Forgive me for not noticing this earlier, since it pertains to our last conversation about the active drag area. 

This would correspond to the [`li.child` dropTarget in Subthoughts.js](https://github.com/cybersemics/em/blob/dev/src/components/Subthoughts.js#L458-L464) rather than Thought.js. Do you think we should create a container component for the `<li>` and move the logic there? Why does the bullet appear on mobile when the other drag area is long pressed, but not on desktop?

2. On mobile, if I initiate a scroll and my finger happens to be on a thought, it will enable the long press. It should immediately be disabled if there is scrolling. The drag and drop library seems to handle this, so we need to have similar behavior for the blue bullet.

> Also want to discuss instant focus and blur events on text after dragging it causing keyboard appearance and disappearance.

Sure, can you provide steps and explain a little more so I can reproduce?

src/components/Editable.js
```diff
@@ -208,6 +209,13 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
     }
   }, [isEditing, cursorOffset])
 
+  // I'm not sure what's happening, but on mobiles sometimes after dragging keyboard apears and disappears instantly
+  useEffect(() => () => {
+    if (contentRef.current !== undefined) {
```
raineorshine, line 214: Simplify to `if (contentRef.current)`.

pushkov-fedor:
I moved action listeners from thought container to parent (<li>), had to deal with event bubbling. Easiest solution was to stopPropogation, but that causes on mobile  to stop working of all drag listeners at all. So i've changed  dragInProgress reducer. Now it prevents setting draggedThoughtsRanked, if previous value wasn't undefined, that also seems logical. Early, because of bubbling, onMouseDown events fire up for child and all parent, and in the end top-level parent bullet was highlighted.

Also, react-dnd deals perfect with scroll, but we start to highlight bullet from onMouseDown event, so react-dnd doesn't help. I added additional eventListener in useLongPress to fix it.

pushkov-fedor:
> Sure, can you provide steps and explain a little more so I can reproduce?
 I think, now it works okay, useEffect on line 213 in Editable.js fixes the problem, but if you comment it out, you will see

raineorshine:
Nice job! This is working great now. That ended up being quite tricky. 

Only one issue I can find. On mobile, we are losing some functionality that we don't want to. You should be able to tap and hold on a word to select it (default browser behavior). You can see this on `dev`. There is a bug (not related to your work) that causes the tap-and-hold browser selection to only work when the cursor is already on the thought, but you can set the cursor first and then see the behavior.

We still want the blue bullet, since the thought can still be dragged from this state, but we need to retain the default browser selection behavior. I might consider putting the overlay bullet into its own component so it can be re-rendered without re-rendering the thought.

I'll add a couple more hours to the budget so you have time to work on this!

raineorshine:
Make sure to `rebase` on `dev` when you get the chance. There are conflicts. Make sure **not** to `merge` or `pull`, as that will result in a messier diff. 

I recommend rebasing every time you start work so that it doesn't get so far behind. You'll get fewer conflicts that way. Your code is on v107 right now but `dev` is up to `v113`.

src/reducers/dragInProgress.js
```diff
@@ -1,3 +1,4 @@
-export default (state, { value }) => ({
-  dragInProgress: value
+export default (state, { value, draggedThoughtsRanked }) => ({
+  dragInProgress: value,
+  draggedThoughtsRanked: state.draggedThoughtsRanked ? (!draggedThoughtsRanked ? draggedThoughtsRanked : state.draggedThoughtsRanked) : draggedThoughtsRanked
```
raineorshine, line 3: Can you write a comment that helps explain the logic here?

I tried multiple times to simplify a statement of this form but I'm not sure its reducible!

```js
a
  ? (!b ? null : a)
  : b
```

pushkov-fedor:
@raineorshine Raine, i sincerely apology for not responding you yesterday. I've checked out fresh dev branch, looked on your bug. On **dev** without selecting a thought, after tap and hold browser offers me _"select al"_ and selects **all** thoughts on screen, including bullets. After selecting a thought and hold, edit cursor appears and browser automatically selects last word in the thought. I want to clarify what hold behavior is expected. 

How do you want users to work with word selection and drag-and-drop? In which moment we decide that it's drag-and-drop or word selection? I wish to know all use-cases in order to to provide more accurate time estimatation and implement all functionality and test it at once.

raineorshine:
> @raineorshine Raine, i sincerely apology for not responding you yesterday. 

No problem!

> I've checked out fresh dev branch, looked on your bug. On **dev** without selecting a thought, after tap and hold browser offers me _"select al"_ and selects **all** thoughts on screen, including bullets. 

In this case, the word under your finger should be selected, and the `cursor` should not change. 

By "selecting a thought", I assume you mean set the `cursor` on the thought. The `cursor` is the gray circle on the bullet that indicates which thought is being focused on. I am considering renaming this to `focus` to avoid confusing it with browser selection.

> After selecting a thought and hold, edit cursor appears and browser automatically selects last word in the thought. I want to clarify what hold behavior is expected.

This isn't what's happening for me on mobile safari. On `dev`, when the cursor is on a thought and I tap and hold, the word under my finger is selected and the "Cut, Copy, Replace" context menu pops up. This is the correct behavior.

> How do you want users to work with word selection and drag-and-drop? In which moment we decide that it's drag-and-drop or word selection? I wish to know all use-cases in order to to provide more accurate time estimatation and implement all functionality and test it at once.

Thanks for your thoroughness. Hopefully this covers all the cases, but let me know if I'm missing something!

1. Tapping and holding a word on a `cursor` thought should perform the default browser behavior. On Mobile Safari it selects the word under the thought. On Android it might do something different. It should enable the blue overlay bullet.
2. Tapping and holding a word on a non-`cursor` thought should do the same as above. (Not working in `dev`; may be addressed in separate PR.)
3. Tapping and holding a thought and dragging should should move the thought. The browser selection, if any, should be removed as soon as the drag movement starts.
4. Tapping and holding to the bullet, the space to the left of the bullet, or the space to the right of the text should not select any text, and should enable the blue overlay bullet.
5. Scrolling should disable the blue overlay bullet.

pushkov-fedor:
As i see it, the key to fix this bug somewhere in Editable component. I think it takes a hour to fix, but lets add one more for force majeure circumstances. Also, what about other tasks?

raineorshine:
If you rebase correctly on `dev`, all of your commits should appear on top, so that it it can be merged with `--ff-only`. I'm not able to review your code currently because there isn't a clean diff. Instead it is showing changes from commits that are already in `dev`.

pushkov-fedor:
@raineorshine Raine, what about this pull request? I synced code yesterday, did you check it? Tell me please, if there's something wrong with code

raineorshine:
Thanks! I must have forgotten to circle back to this. I'll review now.

raineorshine:
Hi! Did you test it after you resolved the merge conflicts? On c40535378b58c3b6752d07cc36a83bc413423f99, I'm getting compile errors of variables not defined.

src/components/Editable.js
```diff
@@ -74,7 +74,7 @@ const stopPropagation = e => e.stopPropagation()
   @contexts indicates that the thought is a context rendered as a child, and thus needs to be displayed as the context while maintaining the correct thoughts path
 */
 // use rank instead of headRank(thoughtsRanked) as it will be different for context view
-const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOffset, showContexts, rank, style, dispatch }) => {
```
raineorshine, line 77: This looks like an incorrect attempt to resolve a merge conflict.

raineorshine:
Merge conflicts are such a pain! :)

raineorshine:
It feels like we're so close. Unfortunately none of the tap-and-hold behavior is working for me on mobile Safari. It always cancels the default browser behavior of selecting the word under the touch. Can I ask what mobile platform you are testing on?

pushkov-fedor:
> Merge conflicts are such a pain! :)

Agree with you, sorry that you had to spend your time on fix merge aftermath

pushkov-fedor:
> It feels like we're so close. Unfortunately none of the tap-and-hold behavior is working for me on mobile Safari. It always cancels the default browser behavior of selecting the word under the touch. Can I ask what mobile platform you are testing on?

Oh, Safari is always pain in the neck. I tested it on Android with Google Chrome

raineorshine:
I started working on the fix for mobile Safari and found another issue.

`onLongPressStart` bubbles up and fires on all ancestors.

- a
  - b
    - c [tap and hold]

You'll notice that `onLongPressStart` fires three times. I tried `e.stopPropagation` but unfortunately that broke dragging on mobile. Maybe you will have better luck than me.

pushkov-fedor:
I've checked your issue. It partially works, but something strange happens with drag-and-drop in conjunction with scroll. Do you have the same on Safari, or drag-and-drop doesn't work at all?

raineorshine:
> I've checked your issue. It partially works, but something strange happens with drag-and-drop in conjunction with scroll. Do you have the same on Safari, or drag-and-drop doesn't work at all?

The drag and drop does not work at all on mobile Safari.

pushkov-fedor:
@raineorshine Raine, i think we need somehow resolve this PR. Maybe we should try to decompose all issues here. Because of COVID-19, I work from home now so i don't have access to IOS device. Without text selection, drag-and-drop works pretty well, but preventing bubling of `onLongPress` function by `stopPropagation` breaks part of drag-and-drop logic. Maybe problem in part of code which is responsible for default browser behavior. Did drag-and-drop work on your Safari before enabling text selection?

raineorshine:
I understand. It's unfortunate that the regression happened to be on a device that you are unable to test currently.

Yes, drag-and-drop and text selection were working correctly on mobile Safari prior to this PR.

I will take another stab at it myself.

pushkov-fedor:
Yes, agree with you. I continue to work on fix of `onLongPressStart` bubbling

raineorshine:
Thanks! I know this issue has been extremely tricky. It'll feel good when we get to a high quality solution though.

pushkov-fedor:
Looks like stopPropagation in onTouchStart event breaks drag-and-drop on mobiles at all. I tried to invoke stopPropagation on top level in Subthoughts component and it also breaks drag-and-drop. So i decided to use dragInProgress value from store to prevent multiple calling.
If it's necessery, I am ready to investigate why stopPropagation breaks drag-and-drop, but it could take a couple of hours.

raineorshine:
> If it's necessery, I am ready to investigate why stopPropagation breaks drag-and-drop, but it could take a couple of hours.

I'm quite nervous about this. Even if we understand why `stopPropagation` breaks drag-and-drop, it's possible that there is nothing we can do to fix it.

I wish there was a way to add the long press functionality independently of drag-and-drop. It's not good that they are interfering with each other.

I'm not really sure. What do you think the best path forward is? The tap-and-hold to select functionality on mobile cannot be sacrificed. But I would like to find a way to salvage this feature.

pushkov-fedor:
It's so sad that react-dnd doesn't have events for detecting touch/mouseDown and hold without movement. Actually any of our custom logic above this library don't look like ideal solution. I don't have wide experience with drag-and-drop, maybe we should try different libraries (e.g. https://use-gesture.netlify.app/docs/hooks, which has cool hooks for different events). But such investigation could take a lot of time.

Maybe we need to finish current drag-and-drop and text selection in order to have working prototype and then think about optimization.

raineorshine:
> I don't have wide experience with drag-and-drop, maybe we should try different libraries (e.g. https://use-gesture.netlify.app/docs/hooks, which has cool hooks for different events). But such investigation could take a lot of time.

Yeah, that would be a large time investment. `react-dnd` is working great in every other respect, and uses `react-native`, so I don't think it's worth it at the moment.

> Maybe we need to finish current drag-and-drop and text selection in order to have working prototype and then think about optimization.

Sure, what do you have in mind? It seems like I might have to work on this on my end to get a working solution for Mobile Safari.

pushkov-fedor:
Looks like i stay at home until the end of May. Give me a day, i will try to get IOS device and reply you tomorrow about it.

pushkov-fedor:
@raineorshine Sorry, that i didn't respond you yesterday. I will take iphone s6 tomorrow and test it. If he's okay, i will be ready to finish text selection.

raineorshine:
Sure. I'd like to know your plan and time estimate after you've checked it out. This task has grown in complexity and I want to make sure it doesn't go too much over budget.

pushkov-fedor:
It looks like that on Safari you can't tap and hold to select the word, if it's text input. You can do double tap for this. On plain text elements tap and hold works. We can try to disable EditableContent for some moment so he will become plain text element but i'm not sure about it.

Let's re-clarify it again please, i've tested tap and hold behavior on text inputs on a few sites, and text selection doesn't work there, but double tap works. I use iphone 6s. Do you have other behavior?

raineorshine:
Try tapping and holding on the editable text when there is more than one word. It doesn't seem to work with a single word element.

pushkov-fedor:
Oh, sorry, i don't have wide experience with Safari :) I think the problem with disabled attribute and its current value. I would like to spend 1-1.5 hour on this tomorrow, is it okay?

pushkov-fedor:
When cursor is on Thought and Thought is editing, tap-and-hold fires "Select and Select All" modal.

raineorshine:
The "Select and Select All" popup appears, but the word under the tap is still not selected for me.

pushkov-fedor:
I found the problem. Dispatching dragInProgress with true flag prevents text selection. Looks like there's somewhere logic with dragInProgress value that blocks text selection behavior. I added another reducer for bullet highlighting and it helped. 
I spent 1.5 hours on this, so i was out of time budget and couldn't find what caused our issue. What do you think about it? I think we shouldn't leave it unclear and should find out what happens

raineorshine:
Great! I will test it out.

My guess is that one of the components is subscribed to `dragInProgress`, which causes a re-render, which blocks the text selection behavior. Let me see if there's any remaining functionality needed, in which case we can investigate further. Otherwise it might be better to move on.

raineorshine:
It's working!!! I feel like our (your) persistence has paid off. Thank you!

src/components/Editable.js
```diff
@@ -220,6 +220,12 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
     }
   }, [isEditing, cursorOffset])
 
+  useEffect(() => () => {
+    if (contentRef.current) {
+      contentRef.current.blur()
+    }
+  })
```
raineorshine, line 227: Can you verify that this is still needed, and if so, add a comment explaining why? I couldn't tell a difference with or without when I tested it, but of course I may have been missing a specific edge case.

src/components/Thought.js
```diff
@@ -387,9 +401,27 @@ const ThoughtContainer = ({
   thoughtsRankedLive,
   url,
   view,
+  viewContext
 }) => {
 
   const state = store.getState()
+  console.log(highlightBullet)
```
raineorshine, line 408: Remove `console.log`

src/components/Thought.js
```diff
@@ -437,7 +469,7 @@ const ThoughtContainer = ({
     'cursor-parent': isCursorParent,
     'cursor-grandparent': isCursorGrandparent,
     'code-view': isCodeView,
-    dragging: isDragging,
+    dragging: (highlightBullet && equalThoughtsRanked(draggedThoughtsRanked, thoughtsRanked)) || isDragging,
```
raineorshine, line 472: I recommend putting `isDragging` first. Putting pre-calculated values first in a boolean expression allows it to short-circuit and avoid calculating the full expression when it doesn't have to.

src/reducers/dragInProgress.js
```diff
@@ -1,3 +1,5 @@
-export default (state, { value }) => ({
-  dragInProgress: value
-})
+export default (state, { value }) => {
+  return {
+    dragInProgress: value,
+  }
+}
```
raineorshine, line 5: Revert to compact representation

src/reducers/highlightBullet.js
```diff
@@ -0,0 +1,7 @@
+export default (state, { value = false, draggedThoughtsRanked }) => {
+  return {
```
raineorshine, line 2: Use compact representation with paren-wrapped object rather than explicit return.

src/reducers/highlightBullet.js
```diff
@@ -0,0 +1,7 @@
...
+    highlightBullet: value,
```
raineorshine, line 3: Let's call this (and the file) `dragHold`. That shifts the semantics from the presentation to the behavior.

src/util/equalThoughtsRanked.js
```diff
@@ -0,0 +1,5 @@
+import { equalThoughtRanked } from './equalThoughtRanked'
+
+/** Compares two arrays of ThoughtRanked based on equalThoughtRanked */
+export const equalThoughtsRanked = (a, b) =>
+  a && b && Array.isArray(a) && Array.isArray(b) && a.length === b.length && a.every((_, i) => equalThoughtRanked(a[i], b[i]))
```
raineorshine, line 5: You can use the existing `equalPath` util function.

src/components/Thought.js
```diff
@@ -372,6 +382,9 @@ const ThoughtContainer = ({
   distance,
   dragPreview,
   dragSource,
+  dragInProgress = false,
+  draggedThoughtsRanked,
+  dragHold,
```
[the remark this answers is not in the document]
pushkov-fedor: Yeah, you're right, thanks. I moved highlighting logic to `BulletCursorOverlay` component. `isDragging` from react-dnd monitor is still passed to `ThoughtComponent`, so it will cause re-render on dragStart. Other part of logic are in new component now. It works, but i'm a little confused in css rules in App.css. I tried to add the same rules as to  `.dragging > .... > .bullet-cursor-overlay` selectors, but it didn't work. Could you point me out please, what rules use for it?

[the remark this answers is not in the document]
raineorshine: Cool, that does look better.

I don't quite understand your question. Are you asking where something is, or how to solve something?

src/hooks/useLongPress.js
```diff
@@ -0,0 +1,42 @@
...
+    setStartLongPress(true)
+  }, [])
+  const stop = useCallback(() => {
+    setStartLongPress(false)
+    onLongPressEnd()
+    setStartCallbackDispatched(false)
+  }, [])
+  const scroll = useCallback(() => {
+    if (!startCallbackDispatched) setStartLongPress(false)
+    else setStartLongPress(true)
```
raineorshine, line 31: Simplify to `setStartLongPress(startCallbackDispatched)`.

raineorshine: Fixed in aa9456c.

raineorshine:
Wow. I think we did it!!! Thank you so much for your perseverance. I had no idea that one of your first issues would be so involved. On the plus side, I really got to see what you are capable of.

Note the comments I added in aa9456c.

Thanks again!

PR #534, opened 2020-04-10
Fix: Cursor disappears after moving parent

incora-dev:
#276

### DIFF
--- a/src/reducers/existingThoughtMove.js
+++ b/src/reducers/existingThoughtMove.js
@@ -5,7 +5,6 @@ import {
   hashContext,
   equalArrays,
   equalThoughtRanked,
-  equalPath,
   getThought,
   hashThought,
   moveThought,
@@ -15,6 +14,7 @@ import {
   rootedContextOf,
   head,
   headRank,
+  subsetThoughts,
   sync,
   timestamp,
   pathToContext,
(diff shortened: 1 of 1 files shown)


src/reducers/existingThoughtMove.js
```diff
@@ -40,7 +40,7 @@ export default (state, { oldPath, newPath, offset }) => {
   const sameContext = equalArrays(oldContext, newContext)
   const oldThought = getThought(value, thoughtIndex)
   const newThought = removeDuplicatedContext(moveThought(oldThought, oldContext, newContext, oldRank, newRank), newContext)
-  const editing = equalPath(state.cursor, oldPath)
+  const editing = subsetThoughts(state.cursor, oldPath)
```
raineorshine, line 43: Let's rename this to `isEditingPath` or `isPathInCursor`. The variable name should change to match the new semantics.

src/reducers/existingThoughtMove.js
```diff
@@ -180,11 +180,24 @@ export default (state, { oldPath, newPath, offset }) => {
     }
   })
 
+  /**
+   * Keep cursor position on child element after moving parent
+   */
+  const generateCursorNewPath = () => {
```
raineorshine, line 186: I don't think we need a function for this. You can just use local variables and expressions.

src/reducers/existingThoughtMove.js
```diff
@@ -180,11 +180,24 @@ export default (state, { oldPath, newPath, offset }) => {
...
+    if (editing) {
+      const children = (state.cursor || []).slice(oldPath.length)
+      return [...newPath, ...children]
```
raineorshine, line 189: Prefer `newPath.concat(cursorDescendantPath)`.

src/reducers/existingThoughtMove.js
```diff
@@ -180,11 +180,24 @@ export default (state, { oldPath, newPath, offset }) => {
     }
   })
 
+  /**
+   * Keep cursor position on child element after moving parent
+   */
+  const generateCursorNewPath = () => {
+    if (editing) {
+      const children = (state.cursor || []).slice(oldPath.length)
```
raineorshine, line 188: Let's call this `cursorDescendantPath`. Children refer to the subthoughts beneath a parent thought. Here rather you are getting the slice of the path that descends from the cursor.

src/reducers/existingThoughtMove.js
incora-dev: changed to isPathInCursor

raineorshine:
Looks good, thanks!

Issue #526, opened 2020-04-07
Convert selectNextEditable/selectPrevEditable to state-based nextUncle


raineorshine:
`selectNextEditable` and `selectPrevEditable` use the DOM to determine the next thought (across contexts). We should replace this with a state-based function that retrieves the next uncle, or the **next sibling of the nearest ancestor that has a next sibling**. 

*Note 1: `nextUncle` should be kept independent from `firstChild` and `nextSibling`. `nextThought` will use all three: `firstChild`, `nextSibling`, and `nextUncle`, trying each one in that order.*

e.g.

- a
  - b
    - c [cursor]
- d

If `nextUncle` is called on the cursor, it should return `d`. 

The steps are:

1. Walk up the path (`c` to `b` to `a`) until you find a context that has a next sibling. 
2. `b` does not have a next sibling so it is skipped
3. `a` has a next sibling: `d` is returned

*Note 2: This gets more complicated when moving across context views.*

Issue #525, opened 2020-04-07
Fix context view cursor navigation


raineorshine:
The cursor navigation actions do not work in and across context views.

- `cursorDown`
- `cursorUp`
- `cursorNext`
- `cursorPrev`
- `cursorBack`
- `cursorForward`

raineorshine:
I will break these out into separate issues and write tests.

PR #535, opened 2020-04-11
adds cursorDown handling for contextView and replaces `selectNextEditable`

anmolarora1:
Refers #525 & #526 
@raineorshine This is an initial draft for the updates that I've made enable `cursorUp` action to handle `contextView`. I've also replaced the `selectNextEditable` which relied on DOM to find out the next editable element.

### DIFF
--- a/src/util/getNextThoughtsWithContextChain.js
+++ b/src/util/getNextThoughtsWithContextChain.js
@@ -1,0 +1,113 @@
+import { store } from '../store.js'
+
+import {
+  ROOT_TOKEN,
+  RANKED_ROOT
+} from '../constants.js'
+// util
+import {
+  head,
+  contextOf,
+  unroot,
+  getThoughtsRanked,
+  getThoughtsSorted,
+  getSortPreference,
+  isFunction,
+  meta,
+  pathToContext,
+  getContextsSortedAndRanked,
+  hashThought,
+  splitChain,
+  nextSibling as thoughtNextSibling
+} from '../util.js'
+import { isContextViewActive } from './isContextViewActive.js'
+
+const contextToThought = (context, thoughtIndex) => context.map(contextValue => {
+  const thought = thoughtIndex[hashThought(contextValue)]
+  return { value: thought.value, rank: thought.contexts[0] ? thought.contexts[0].rank : 0 }
+})
+
+const mapContextsToThoughts = (contexts, thoughtIndex) => contexts.map(({ context }) => contextToThought(context, thoughtIndex))
+
+const getContextSiblings = (context, thoughtIndex) => {
+  const contexts = getContextsSortedAndRanked(head(context))
+  return mapContextsToThoughts(contexts, thoughtIndex)
+}
+
+const nextSiblingContext = (value, context, thoughtIndex) => {
+  const contextSiblings = getContextSiblings(context, thoughtIndex)
+  const currentIndex = contextSiblings.findIndex(thoughts => thoughts.find(thought => thought.value === value))
+  return contextSiblings[currentIndex + 1]
+}
+
+const firstChildOfContext = (value, cursor, thoughtIndex) => {
+  const context = pathToContext(cursor)
+  return nextSiblingContext(value, context, thoughtIndex)
+}
+
+const firstChildOfThought = (thoughtsRanked, showHiddenThoughts) => {
+  const contextMeta = meta(thoughtsRanked)
+  const sortPreference = getSortPreference(contextMeta)
+  const children = (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(thoughtsRanked)
+  const notHidden = child => !isFunction(child.value) && !meta(pathToContext(thoughtsRanked).concat(child.value)).hidden
+  const childrenFiltered = showHiddenThoughts ? children : children.filter(notHidden)
+  return childrenFiltered[0]
+}
+
+/** Gets nextThoughts and their respective contextChain */
+export const getNextThoughtsWithContextChain = (value, context, rank, cursor, contextRanked, contextChain) => {
+  const { showHiddenThoughts, thoughtIndex, contextViews } = store.getState()
+  if (!contextChain) {
+    contextChain = splitChain(cursor, contextViews)
+  }
+
+  if (context.length === 0 || cursor.length === 0) {
+    return null
+  }
+
+  // if the cursor is at a thought with context view open, move it into context view - jump in
+  if (isContextViewActive(pathToContext(cursor))) {
+    const withinContextView = isContextViewActive(contextRanked)
+    const thoughtsRanked = withinContextView ? getContextSiblings(context, thoughtIndex) : getThoughtsRanked(context)
+
+    const currentThought = withinContextView ? thoughtsRanked.find(thoughts => thoughts.find(thought => thought.value === value))[0] : thoughtsRanked.find(thought => thought.value === value && thought.rank === rank)
+    const firstChild = firstChildOfContext(value, cursor, thoughtIndex)
+    if (firstChild) {
+      return { nextThoughts: [...firstChild, currentThought], contextChain }
+    }
+    // jump out if there are no context children
+    return getNextThoughtsWithContextChain(value, context, rank, cursor, contextOf(contextRanked), contextOf(contextChain))
+  }
+  // if cursor is within a context view, return the next context
+  else if (isContextViewActive(contextRanked)) {
+    const nextSibling = nextSiblingContext(value, context, thoughtIndex, showHiddenThoughts)
+    const contextRankedHead = head(contextRanked)
+
+    if (nextSibling) {
+      return { nextThoughts: [...nextSibling, contextRankedHead], contextChain: contextOf(contextChain) }
+    }
+    // jump out
+    return getNextThoughtsWithContextChain(contextRankedHead.value, contextOf(context), contextRankedHead.rank, cursor, contextOf(contextRanked), contextOf(contextChain))
+  }
+
+  const thoughtsRanked = cursor || RANKED_ROOT
+  const firstChild = firstChildOfThought(thoughtsRanked, showHiddenThoughts)
+  if (firstChild) {
+    return { nextThoughts: unroot(thoughtsRanked.concat(firstChild)), contextChain: [] }
+  }
+
+  const nextSibling = thoughtNextSibling(value, context, rank, showHiddenThoughts)
+  const sortPreference = getSortPreference(meta(pathToContext(context)))
+  const siblings = sortPreference === 'Alphabetical' ? getThoughtsSorted(context) : getThoughtsRanked(context)
+
+  const contextHead = head(context)
+  if (siblings.length === 0) {
+    return getNextThoughtsWithContextChain(contextHead, contextOf(context), rank, cursor, contextRanked, contextChain)
+  }
+
+  if (!nextSibling) {
+    const parentContext = context.length === 1 ? [ROOT_TOKEN] : contextOf(context)
+    return getNextThoughtsWithContextChain(contextHead, parentContext, head(contextRanked).rank, cursor, contextOf(contextRanked), contextChain)
+  }
+  return { nextThoughts: unroot(contextRanked.concat(nextSibling)), contextChain: [] }
+}
--- a/src/util/getNextThoughtsWithContextChain.js
+++ b/src/util/getNextThoughtsWithContextChain.js
@@ -24,6 +24,10 @@ import {
 } from '../util.js'
 import { isContextViewActive } from './isContextViewActive.js'
 
+/**
+ * Used to check whether a context view is valid and there are more than one child
+ @returns {boolean}
+ */
 const isValidContextView = context => {
   if (!context.length) {
     return false
@@ -31,6 +35,10 @@ const isValidContextView = context => {
   return isContextViewActive(context) && getContextsSortedAndRanked(head(pathToContext(context))).length > 1
 }
 
+/**
+ * Transforms a context of a thought to rankedContext
+ @returns {array | null} rankedContext
+ */
 const getThoughtContextsWithRank = (contextInfo, thoughtIndex) => {
   return contextInfo && contextInfo.context.map(c => {
     const thought = getThought(c, thoughtIndex)
@@ -38,6 +46,10 @@ const getThoughtContextsWithRank = (contextInfo, thoughtIndex) => {
   })
 }
 
+/**
+ * Returns the next sibling of the focused context
+ @returns {array | null} rankedContext
+ */
 const nextSiblingContext = (rank, context, thoughtIndex) => {
   const contextSiblings = getContextsSortedAndRanked(head(context))
   const currentContextIndex = contextSiblings.findIndex(context => context.rank === rank)
@@ -46,6 +58,10 @@ const nextSiblingContext = (rank, context, thoughtIndex) => {
   return getThoughtContextsWithRank(nextSibling, thoughtIndex)
 }
 
+/**
+ * Returns the first child of context at the given path
+ @returns {array | undefined} rankedContext
+ */
 const firstChildOfContext = (path, thoughtIndex) => {
   const context = pathToContext(path)
   const contextChildren = getContextsSortedAndRanked(head(context))
@@ -53,6 +69,10 @@ const firstChildOfContext = (path, thoughtIndex) => {
   return getThoughtContextsWithRank(firstChild, thoughtIndex)
 }
 
+/**
+ * Returns the first subthought of thought at the given path
+ @returns {object | undefined} thought
+ */
 const getSubThought = (path, showHiddenThoughts) => {
   const contextMeta = meta(path)
   const sortPreference = getSortPreference(contextMeta)
@@ -62,6 +82,10 @@ const getSubThought = (path, showHiddenThoughts) => {
   return childrenFiltered[0]
 }
 
+/**
+ * Returns the path to the current thought by stripping out any context views
+ @returns {array} path
+ */
 const getPathFromContextChain = contextChain => {
   // last of second last item in context chain gives us the current context
   const context = head(head(contextOf(contextChain)))
@@ -73,6 +97,19 @@ const getPathFromContextChain = contextChain => {
   return [...matchedContext, context, ...head(contextChain).slice(1)]
 }
 
+/**
+ * Computes the value of next thoughts  and context chain if-
+ * 1) the focused thought has context view open
+ * 2) the focus is on a context
+ * Delegates control to getNextThoughtsAndContextChainFromThoughtview if none of the above conditions meet
+ * @param {string} value - The value of focused thought
+ * @param {array} context - Context of focused thought
+ * @param {string} rank - Rank of focused thought
+ * @param {array} path - Path to focsued thought
+ * @param {array} rankedContext - Context with rank
+ * @param {array} contextChain - ContextChain for the focused thought
+ * @param {(boolean | undefined)} ignoreChildren - Used to ignore the children context if they've been traversed already
+ */
 const getNextThoughtsAndContextChainFromContextview = (value, context, rank, path, rankedContext, contextChain, ignoreChildren) => {
   const { showHiddenThoughts, thoughtIndex } = store.getState()
 
@@ -112,6 +149,19 @@ const getNextThoughtsAndContextChainFromContextview = (value, context, rank, pat
   return getNextThoughtsAndContextChainFromThoughtview(value, context, rank, path, contextChain, contextWithNoChildren)
 }
 
+/**
+ * Computes the value of next thoughts  and context chain if-
+ * 1) the focused thought is not within a context view
+ * 2) the focused thought is at level 2 or further down in a context tree
+ * Delegates control to getNextThoughtsAndContextChainFromContextview if none of the above conditions meet
+ * @param {string} value - The value of focused thought
+ * @param {array} context - Context of focused thought
+ * @param {string} rank - Rank of focused thought
+ * @param {array} path - Path to focsued thought
+ * @param {array} rankedContext - Context with rank
+ * @param {array} contextChain - ContextChain for the focused thought
+ * @param {(boolean | undefined)} ignoreChildren - Used to ignore the subthoughts if they've been traversed already
+ */
 const getNextThoughtsAndContextChainFromThoughtview = (value, context, rank, path, contextChain, ignoreChildren) => {
   const { showHiddenThoughts } = store.getState()
 
--- a/src/util/getNextThoughtsWithContextChain.js
+++ b/src/util/getNextThoughtsWithContextChain.js
@@ -67,7 +67,7 @@ const getSubThought = (thoughtsRanked, showHiddenThoughts) => {
 
 const getPathFromContextChain = contextChain => {
   // last of second last item in context chain gives us the current context
-  const context = head(contextChain.slice(-2, -1)[0])
+  const context = head(head(contextOf(contextChain)))
   const contexts = getContextsSortedAndRanked(context.value)
   const currentContextTop = head(contextChain)[0].value
   const matchedContextChain = contexts.find(c => c.context.includes(currentContextTop))
--- a/src/util/getNextThoughtsWithContextChain.js
+++ b/src/util/getNextThoughtsWithContextChain.js
@@ -24,43 +24,40 @@ import {
 } from '../util.js'
 import { isContextViewActive } from './isContextViewActive.js'
 
-const contextToThought = (context, thoughtIndex) => context.map(contextValue => {
-  const thought = getThought(contextValue, thoughtIndex)
-  return { value: thought.value, rank: thought.contexts[0] && thought.contexts[0].rank }
-})
-
-const mapContextsToThoughts = (contexts, thoughtIndex) => contexts.map(({ context }) => contextToThought(context, thoughtIndex))
-
 const isValidContextView = context => {
-  // const allowSingleContext = false
   if (!context.length) {
     return false
   }
   return isContextViewActive(context) && getContextsSortedAndRanked(head(pathToContext(context))).length > 1
 }
 
-const getContextSiblings = (context, thoughtIndex) => {
-  const contexts = getContextsSortedAndRanked(head(context))
-  return mapContextsToThoughts(contexts, thoughtIndex)
+const getThoughtContextsWithRank = (contextInfo, thoughtIndex) => {
+  return contextInfo && contextInfo.context.map(c => {
+    const thought = getThought(c, thoughtIndex)
+    return { ...(thought || { value: c }), rank: contextInfo.rank }
+  })
 }
 
-const nextSiblingContext = (value, rank, context, thoughtIndex) => {
-  const contextSiblings = getContextSiblings(context, thoughtIndex)
-  const currentIndex = contextSiblings.findIndex(thoughts => thoughts.find(thought => thought.value === value))
-  return contextSiblings[currentIndex + 1]
+const nextSiblingContext = (rank, context, thoughtIndex) => {
+  const contextSiblings = getContextsSortedAndRanked(head(context))
+  const currentContextIndex = contextSiblings.findIndex(context => context.rank === rank)
+  // const currentIndex = contextSiblings.findIndex(thoughts => thoughts.find(thought => thought.value === value))
+  const nextSibling = contextSiblings[currentContextIndex + 1] ? contextSiblings[currentContextIndex + 1] : null
+  return getThoughtContextsWithRank(nextSibling, thoughtIndex)
 }
 
-const firstChildOfContext = (value, cursor, thoughtIndex) => {
-  const context = pathToContext(cursor)
-  const contextSiblings = getContextSiblings(context, thoughtIndex)
-  return contextSiblings[0]
+const firstChildOfContext = (path, thoughtIndex) => {
+  const context = pathToContext(path)
+  const contextChildren = getContextsSortedAndRanked(head(context))
+  const firstChild = contextChildren[0]
+  return getThoughtContextsWithRank(firstChild, thoughtIndex)
 }
 
-const getSubThought = (thoughtsRanked, showHiddenThoughts) => {
-  const contextMeta = meta(thoughtsRanked)
+const getSubThought = (path, showHiddenThoughts) => {
+  const contextMeta = meta(path)
   const sortPreference = getSortPreference(contextMeta)
-  const children = (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(thoughtsRanked)
-  const notHidden = child => !isFunction(child.value) && !meta(pathToContext(thoughtsRanked).concat(child.value)).hidden
+  const children = (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(path)
+  const notHidden = child => !isFunction(child.value) && !meta(pathToContext(path).concat(child.value)).hidden
   const childrenFiltered = showHiddenThoughts ? children : children.filter(notHidden)
   return childrenFiltered[0]
 }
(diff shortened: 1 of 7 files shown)


raineorshine:
Did you run into the bug in `dev` described by #551, which I just observed and documented? I assume that would get in the way of implementing `cursorDown` in the context view.

raineorshine:
Here are the cases I tested. The ones that are not working yet are unchecked:

## Normal View

- [ ] `cursorDown` null cursor
e.g. Hit escape until the cursor is `null`, then activate `cursorDown`. It should move to the first thought in the root.
- [x] `cursorDown` next sibling
- [x] `cursorDown` first child
- [x] `cursorDown` next uncle

## Context View

- [ ] `cursorDown` context view → next sibling
e.g. Activate `cursorDown` from `a~`. Cursor should move to `b`:
  - a [context view]
  - b

- [x] `cursorDown` context view → first context
- [ ] `cursorDown` context subthought → next sibling
e.g. Activate `cursorDown` from `a.m~.a.x`. Cursor should move to `a.m~.a.y`
  - a
    - m
      - x
      - y
  - b
    - m
- [ ] `cursorDown` context → first subthought
e.g. Activate `cursorDown` from `a.m~.a`. Cursor should move to `a.m~.a.x`
  - a
    - m
      - x
  - b
    - m
- [x] `cursorDown` context → next uncle
- [ ] `cursorDown` context subthought → next uncle
e.g. Activate `cursorDown` from `a.m~.a.x`. Cursor should move to `a.m~.b`
  - a
    - m
      - x
  - b
    - m

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,113 @@
...
+  pathToContext,
+  getContextsSortedAndRanked,
+  hashThought,
+  splitChain,
+  nextSibling as thoughtNextSibling
+} from '../util.js'
+import { isContextViewActive } from './isContextViewActive.js'
+
+const contextToThought = (context, thoughtIndex) => context.map(contextValue => {
+  const thought = thoughtIndex[hashThought(contextValue)]
```
raineorshine, line 26: I recommend using `getThought` in cases like this. You should not be accessing `thoughtIndex` directly, but instead using the selectors and accessor methods.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,113 @@
...
+  return { value: thought.value, rank: thought.contexts[0] ? thought.contexts[0].rank : 0 }
```
raineorshine, line 27: Defaulting the rank to `0` is a red flag... seems very likely to lead to problems.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,113 @@
...
+const mapContextsToThoughts = (contexts, thoughtIndex) => contexts.map(({ context }) => contextToThought(context, thoughtIndex))
+
+const getContextSiblings = (context, thoughtIndex) => {
+  const contexts = getContextsSortedAndRanked(head(context))
+  return mapContextsToThoughts(contexts, thoughtIndex)
+}
+
+const nextSiblingContext = (value, context, thoughtIndex) => {
+  const contextSiblings = getContextSiblings(context, thoughtIndex)
+  const currentIndex = contextSiblings.findIndex(thoughts => thoughts.find(thought => thought.value === value))
```
raineorshine, line 39: If you ignore `rank`, this might not work with duplicate thoughts (thoughts with the same value and different ranks within the same context).

raineorshine: `equalThoughtRanked` does the full comparison of both `value` and `rank`.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,113 @@
...
+      return { nextThoughts: [...nextSibling, contextRankedHead], contextChain: contextOf(contextChain) }
+    }
+    // jump out
+    return getNextThoughtsWithContextChain(contextRankedHead.value, contextOf(context), contextRankedHead.rank, cursor, contextOf(contextRanked), contextOf(contextChain))
+  }
+
+  const thoughtsRanked = cursor || RANKED_ROOT
+  const firstChild = firstChildOfThought(thoughtsRanked, showHiddenThoughts)
+  if (firstChild) {
+    return { nextThoughts: unroot(thoughtsRanked.concat(firstChild)), contextChain: [] }
```
raineorshine, line 96: Having `return` statements interspersed throughout the function makes it harder to see the control flow. Instead, build up the correct result and return it at the end.

src/util/getNextThoughtsWithContextChain.js
anmolarora1: I've noticed some ambiguities in the rank information. I will look further into that and update the code accordingly.

raineorshine:
Context View looks great! Nice work. Just two small edge cases in the normal view.

## Normal View

- [ ] `cursorDown` null cursor
e.g. Hit escape until the cursor is `null`, then activate `cursorDown`. It should move to the first thought in the root.
- [x] `cursorDown` next sibling
- [x] `cursorDown` first child
- [x] `cursorDown` next uncle
- [ ] `cursorDown` last thought
e.g. Select the last thought and activate `cursorDown`. The cursor should not move.

## Context View

- [x] `cursorDown` context view → next sibling
e.g. Activate `cursorDown` from `a~`. Cursor should move to `b`:
  - a [context view]
  - b

- [x] `cursorDown` context view → first context
- [x] `cursorDown` context → next sibling
e.g. Activate `cursorDown` from `a.m~.a`. Cursor should move to `a.m~.b`
  - a
    - m
  - b
    - m

- [x] `cursorDown` context subthought → next sibling
e.g. Activate `cursorDown` from `a.m~.a.x`. Cursor should move to `a.m~.a.y`
  - a
    - m
      - x
      - y
  - b
    - m
- [x] `cursorDown` context → first subthought
e.g. Activate `cursorDown` from `a.m~.a`. Cursor should move to `a.m~.a.x`
  - a
    - m
      - x
  - b
    - m
- [x] `cursorDown` context → next uncle
- [x] `cursorDown` context subthought → next uncle
e.g. Activate `cursorDown` from `a.m~.a.x`. Cursor should move to `a.m~.b`
  - a
    - m
      - x
  - b
    - m

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+  const context = head(contextChain.slice(-2, -1)[0])
+  const contexts = getContextsSortedAndRanked(context.value)
+  const currentContextTop = head(contextChain)[0].value
+  const matchedContextChain = contexts.find(c => c.context.includes(currentContextTop))
+  const matchedContext = matchedContextChain.context.map(i => ({ value: i, rank: matchedContextChain.rank }))
+
+  return [...matchedContext, context, ...head(contextChain).slice(1)]
+}
+
+const contextViewLookup = (value, context, rank, cursor, contextRanked, contextChain, ignoreChildren) => {
```
raineorshine, line 77: What is a "Lookup"? This name does not give a clear picture of what this function does. What does the function do, and what does it return? This needs to be clearly documented.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+      return { nextThoughts: [...nextSibling, contextRankedHead], contextChain: contextOf(contextChain) }
+    }
+    // jump out
+    return thoughtViewLookUp(contextRankedHead.value, contextOf(derivedContext), contextRankedHead.rank, contextOf(cursor), contextOf(contextChain), true)
+  }
+  const contextWithNoChildren = contextChain.length === 1 && isContextViewActive(contextChain[0]) && getContextsSortedAndRanked(head(pathToContext(contextChain[0]))).length <= 1
+  return thoughtViewLookUp(value, context, rank, cursor, contextChain, contextWithNoChildren)
+}
+
+const thoughtViewLookUp = (value, context, rank, cursor, contextChain, ignoreChildren) => {
```
raineorshine, line 120: What's the relationship between `contextViewLookUp` and `thoughtViewLookup`? This deserves a thorough explanation (think 1-2 paragraphs). All I can tell is that you have the recursion split across two functions, but I have no idea what the purpose of this implementation is, how it works, and why you chose it.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+    return getNextThoughtsWithContextChain(contextHead, contextOf(derivedContext), rank, contextOf(derivedCursor), derivedContextRanked, contextChain.length > 1 ? [contextChain.slice(-1)] : contextChain)
+  }
+
+  const nextSibling = thoughtNextSibling(value, derivedContext, rank, showHiddenThoughts)
+  if (nextSibling) {
+    return { nextThoughts: unroot(derivedContextRanked.concat(nextSibling)), contextChain: contextOf(contextChain) }
+  }
+  // look for uncle/ancestor sibling
+  else {
+    const parentThought = derivedCursor.slice(-2, -1)[0]
```
raineorshine, line 149: Do not use array methods on paths or thoughts. That is breaking the separation between interface and implementation. Use the interface methods, `head` and `contextOf`.

This is a really important rule that you should apply everywhere. Directly operating on the array implementation is going to make it impossible to change the implementation in the future. We should reply on the API that is exposed through the available methods.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+    const rankedContextOfCurrentContext = contextOf(contextChain)[0]
+    const contextChainTillFirstChildOfContext = [...contextOf(contextChain), [head(contextChain)[0]]]
+    const firstThoughtInContext = head(contextChain)[0]
+
+    // reached last child of last thought
+    if (!parentThought) {
+      return { nextThoughts: [siblings[0]] || [ROOT_TOKEN], contextChain: [] }
+    }
+    const thoughtsWithContextChain = thoughtViewLookUp(parentThought.value, parentContext, parentThought.rank, parentCursor, contextChainForParentThought, true)
+    return thoughtsWithContextChain || contextViewLookup(firstThoughtInContext.value, rankedContextOfCurrentContext.map(c => c.value), firstThoughtInContext.rank, cursorToFirstThoughtInContext, rankedContextOfCurrentContext, contextChainTillFirstChildOfContext, true)
```
raineorshine, line 165: Use `pathToContext` instead of `.map(c => c.value)`.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+  return contextSiblings[currentIndex + 1]
+}
+
+const firstChildOfContext = (value, cursor, thoughtIndex) => {
+  const context = pathToContext(cursor)
+  const contextSiblings = getContextSiblings(context, thoughtIndex)
+  return contextSiblings[0]
+}
+
+const firstChildOfThought = (thoughtsRanked, showHiddenThoughts) => {
```
raineorshine, line 57: I don't know what it means to have a child of a thought. Only contexts have children. Can you explain?

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+  return mapContextsToThoughts(contexts, thoughtIndex)
+}
+
+const nextSiblingContext = (value, rank, context, thoughtIndex) => {
+  const contextSiblings = getContextSiblings(context, thoughtIndex)
+  const currentIndex = contextSiblings.findIndex(thoughts => thoughts.find(thought => thought.value === value))
+  return contextSiblings[currentIndex + 1]
+}
+
+const firstChildOfContext = (value, cursor, thoughtIndex) => {
```
raineorshine, line 51: All of the references to `cursor` should simply be to `path`. It doesn't have to be the `cursor`.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+    return getNextThoughtsWithContextChain(value, context, rank, cursor, contextOf(contextRanked), contextOf(contextChain))
+  }
+  // if cursor is within a context view
+  else if (isValidContextView(contextRanked)) {
+    const headContext = contextOf(cursor).map(c => c.value)
+    if (!ignoreChildren && isValidContextView(headContext)) {
+      const derivedCursor = getDerivedCursor(contextChain)
+      const firstChild = firstChildOfThought(derivedCursor || RANKED_ROOT, showHiddenThoughts)
+      if (firstChild) {
+        return { nextThoughts: unroot(cursor.concat(firstChild)), contextChain: [] }
```
raineorshine, line 101: As mentioned before, I do not approve of `return` statements in the middle of function bodies. It is imperative and unnecessarily breaks the invariant that the shape of the code should reflect the shape of the data and control flow. You will need to refactor this to return a single value at the end.

The only exception is for short-circuit return statements at the very beginning of a function, e.g. as done at the top of this function: `if (context.length === 0 || cursor.length === 0) return null`.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+    const firstChild = firstChildOfContext(value, cursor, thoughtIndex)
+    if (!ignoreChildren && firstChild) {
+      return { nextThoughts: [...firstChild, currentThought], contextChain }
+    }
+    // jump out if there are no context children
+    return getNextThoughtsWithContextChain(value, context, rank, cursor, contextOf(contextRanked), contextOf(contextChain))
+  }
+  // if cursor is within a context view
+  else if (isValidContextView(contextRanked)) {
+    const headContext = contextOf(cursor).map(c => c.value)
```
raineorshine, line 96: Use `pathToContext` instead of `.map(c => c.value)`.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+const firstChildOfThought = (thoughtsRanked, showHiddenThoughts) => {
+  const contextMeta = meta(thoughtsRanked)
+  const sortPreference = getSortPreference(contextMeta)
+  const children = (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(thoughtsRanked)
+  const notHidden = child => !isFunction(child.value) && !meta(pathToContext(thoughtsRanked).concat(child.value)).hidden
+  const childrenFiltered = showHiddenThoughts ? children : children.filter(notHidden)
+  return childrenFiltered[0]
+}
+
+const getDerivedCursor = contextChain => {
```
raineorshine, line 66: What is a "derived cursor"? When introducing a new term not used anywhere before in the codebase, please include a description.

anmolarora1:
The code is written in accordance with the current tree structure that we have for representing thoughts and contexts.  On a root level, there are 2 functions -
1. **`getNextThoughtsAndContextChainFromContextview`**
2. **`getNextThoughtsAndContextChainFromThoughtview`**

Why the approach of recursion with switching between functions was chosen - 
At any point, the `cursor` or `focus` can be either in the normal or the context view.  The recursive pattern is mostly used to find the `uncle` thought/context when there are no further siblings. Interchanging between these two functions is needed when the cursor is at an Inflexion point (where it needs to transition from context to thought view or vice-versa).  Eg. If the cursor is at `a/m~/b/o`, we need to search for the next thoughts using the `context view` logic Whereas, if it’s at `a/m~/c`, the next thought needs to be determined using the `thought view` logic

We can use this example for a better understanding -  
```

▼ a 
     ᐁ m
         ∙ x 		
         ▼ b
             ∙ n
             ∙ o
	 ∙ c
▼ b
     ▼ m
	∙ n
	∙ o
▼ c
   ∙ m  
```
  Here’s the basic pseudocode -
1. If the current cursor is at a thought with its context view open, try searching for its children (a/m~) 
2. If the cursor within a context view, see if  1. The thought in focus has got sub-thoughts within it. If `a/m~/b`, next is - `a/m~b/n`  2. No sub thoughts? Look for the sibling. If `a/m~/x`, next is - `a/m~/b`  3. No sibling? Shift cursor one level up and try navigating thought view (This is where we transition from one view to another). (`a/m~/c`) 
3. 1. The thought in focus has got sub-thoughts within it. If `b/m`, next is - `b/m/n` 2. No sub thoughts? Look for the sibling. If `a/m~/x`, next is - `a/m~/b`  3. No sibling? Shift cursor one level up and try navigating thought view `b/m/o` (recursion wishing thought view) 4. No result (reached context view)? Shift the cursor to the first thought in that context view and try navigating context view.  If `a/m~/b/o`, try searching for the siblin in the context view `a/m~` with `b` as the current value.

src/util/getNextThoughtsWithContextChain.js
anmolarora1: I've updated the terminologies to make them sound more explicit.

src/util/getNextThoughtsWithContextChain.js
anmolarora1: Updated the names and documented in the description

src/util/getNextThoughtsWithContextChain.js
anmolarora1: Thanks for sharing. It's updated now.

src/util/getNextThoughtsWithContextChain.js
anmolarora1: I've tried to make it more declarative by using `perma` in some places. We can discuss how we can improve it further.

src/util/getNextThoughtsWithContextChain.js
anmolarora1: Added description with example

src/util/getNextThoughtsWithContextChain.js
anmolarora1: I agree with you on that. And I've updated such instances. 
We'll need a util function to return the first item in the list of thoughts/contexts to ger rid of all array operations.

src/util/getNextThoughtsWithContextChain.js
```diff
@@ -0,0 +1,178 @@
...
+      return { nextThoughts: [siblings[0]] || [ROOT_TOKEN], contextChain: [] }
+    }
+    const thoughtsWithContextChain = thoughtViewLookUp(parentThought.value, parentContext, parentThought.rank, parentCursor, contextChainForParentThought, true)
+    return thoughtsWithContextChain || contextViewLookup(firstThoughtInContext.value, rankedContextOfCurrentContext.map(c => c.value), firstThoughtInContext.rank, cursorToFirstThoughtInContext, rankedContextOfCurrentContext, contextChainTillFirstChildOfContext, true)
+
+  }
+}
+
+/** Gets nextThoughts and their respective contextChain */
+export const getNextThoughtsWithContextChain = (value, context, rank, cursor, contextRanked, contextChain) => {
```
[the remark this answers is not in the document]
anmolarora1: Updated

src/util/getNextThoughtsWithContextChain.js
anmolarora1: Updated

src/util/nextThought.js
```diff
@@ -31,56 +31,58 @@ import {
...
  * @returns {boolean}
  */
-const isValidContextView = context =>
-  context.length &&
-  isContextViewActive(context) &&
-  getContexts(head(context)).length > 1
+const isValidContextView = path =>
+  path.length &&
+  isContextViewActive(pathToContext(path)) &&
+  getContexts(head(path).value).length > 1
```
raineorshine, line 37: Technically `head` should only be called on `Path` types, `Contexts` are meant to be unordered. Since the current implementation of `Context` is an array, it ends up working on both. I haven't been consistently using `head` in this way myself so this is not that important right now.

src/util/nextThought.js
```diff
@@ -116,19 +118,20 @@ const nextInContextView = (value, rank, path, rankedContext, contextChain, ignor
...
     const currentThought = head(path)
-    const firstChild = firstChildOfContext(path, thoughtIndex)
     // jump out if there are no context children
-    return !ignoreChildren && firstChild
-      ? { nextThoughts: [...firstChild, currentThought], contextChain }
-      : nextInContextView(value, rank, path, contextOf(rankedContext), contextOf(contextChain))
+    return {
+      nextThoughts: [...firstChild(), currentThought],
+      contextChain
+    }
```
raineorshine: When there are no context children, we can allow it to move to the default case. This simplifies this case.

src/util/nextThought.js
```diff
@@ -176,19 +179,25 @@ const nextInThoughtView = (value, context, rank, path, contextChain, ignoreChild
...
+      return (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(thoughtViewContext)
+    }
+
+    // only calculate uncle if not at root
+    const nextUncle = () => {
+      const parentContext = context.length === 1 ? [ROOT_TOKEN] : contextOf(thoughtViewContext)
+      const contextChainForParentThought = [...contextOf(contextChain), contextOf(head(contextChain))]
+      const parentPath = contextOf(thoughtViewPath)
+      return nextInThoughtView(parentThought.value, parentContext, parentThought.rank, parentPath, contextChainForParentThought, true)
+    }
```
raineorshine, line 197: No need to calculate the uncle when we are at the root thought.

src/util/nextThought.js
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * Adds the rank of the child thought to every thought in a context.
+ * @param {ContextInfo} contextInfo.   A { context, rank } object returned from getContexts.
+ * @returns {array | null} rankedContext
+ */
+const contextWithThoughtRank = contextInfo => {
+  return contextInfo && contextInfo.context.map(value => {
+    const thought = getThought(value)
+    return { ...(thought || { value }), rank: contextInfo.rank }
```
raineorshine, line 47: `contextInfo.rank` is just the rank of the head thought. Here it is incorrectly applied to every thought in the path that is returned. This will result in an incorrect `thoughtsRanked` passed to `setCursor` in the end.

(Technically I'm not sure if these steps apply to this instance of the problem or the one in `getPathFromContextChain` mentioned below. Either way it demonstrates the issue.)

## Steps to Reproduce

Create the following thoughts from top to bottom:

  - a (rank: 0)
    - b
      - c
        - x
        - y
  - d (rank: 1)
  - e (rank: 2)
    - y

Set the cursor on `a.b.c.y~.c` and hit `cursorDown` to move the cursor to `a.b.c.y~.e`.

## Current Behavior

`nextThought` returns: 

```js
[
  { value: 'e', rank: 1 },
  { value: 'y', rank: 1 }
]
```

## Expected Behavior

```js
[
  { value: 'e', rank: 2 },
  { value: 'y', rank: 1 }
]
```

src/util/nextThought.js
```diff
@@ -0,0 +1,240 @@
...
+ * Returns the path to the current thought by stripping out any context views
+ * @returns {array} path
+ */
+const getPathFromContextChain = contextChain => {
+  // last of second last item in context chain gives us the current context
+  const context = head(head(contextOf(contextChain)))
+  const contexts = getContextsSortedAndRanked(context.value)
+  const currentContextTop = head(contextChain)[0].value
+  const matchedContextChain = contexts.find(c => c.context.includes(currentContextTop))
+  const matchedContext = matchedContextChain.context.map(i => ({ value: i, rank: matchedContextChain.rank }))
```
raineorshine, line 98: Setting `rank` here causes the same issue described above. It should only be applied to the relevant thought, not all thoughts.

src/util/nextThought.js
```diff
@@ -0,0 +1,240 @@
...
+      : nextInThoughtView(rankedContextHead.value, contextOf(context), rankedContextHead.rank, contextOf(path), contextOf(contextChain), true)
+  }
+  // not a valid context view. Try navigating thoughts
+  else {
+    const contextWithNoChildren = contextChain.length === 1 &&
+      isContextViewActive(contextChain[0]) &&
+      getContexts(head(pathToContext(contextChain[0]))).length <= 1
+
+    return nextInThoughtView(value, context, rank, path, contextChain, contextWithNoChildren)
+  }
```
raineorshine, line 156: In which cases does this `else` get executed? Is it possible to not call `nextInContextView` at all if this is the case, and instead call `nextInThoughtView` at the caller? I'm wondering if that might be cleaner.

If this is only hit when `nextThought` is called the first time, then we can definitely move some handling logic there and keep `nextInContextView` strictly for within a context view. If not, it might be more complicated, but I would like to understand the effort involved.

src/util/nextThought.js
```diff
@@ -101,59 +102,62 @@ const getPathFromContextChain = contextChain => {
...
+ * Delegates control to nextInThoughtView if none of the above conditions meet
  * @param {string} value - The value of focused thought
- * @param {array} context - Context of focused thought
  * @param {string} rank - Rank of focused thought
  * @param {array} path - Path to focsued thought
  * @param {array} rankedContext - Context with rank
  * @param {array} contextChain - ContextChain for the focused thought
  * @param {(boolean | undefined)} ignoreChildren - Used to ignore the children context if they've been traversed already
- */
-const getNextThoughtsAndContextChainFromContextview = (value, context, rank, path, rankedContext, contextChain, ignoreChildren) => {
```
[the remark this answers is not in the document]
anmolarora1: I see. Thanks for pointing out. Will keep it mind

src/util/nextThought.js
```diff
@@ -101,59 +102,62 @@ const getPathFromContextChain = contextChain => {
...
     const rankedContextHead = head(rankedContext)
 
-    if (!ignoreChildren) {
-      const firstChild = getFirstChild()
-      if (firstChild) {
-        return { nextThoughts: unroot(path.concat(firstChild)), contextChain: [] }
-      }
+    return !ignoreChildren && firstChild() ? {
+      nextThoughts: unroot([...path, firstChild()]),
+      contextChain: []
```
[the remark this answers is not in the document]
anmolarora1: If there's no additional overhead in calling it again, the new approach makes sense

src/util/nextThought.js
```diff
@@ -162,61 +166,66 @@ const getNextThoughtsAndContextChainFromContextview = (value, context, rank, pat
...
-
-  })
+    return nextInContextView(firstThoughtInContext.value, firstThoughtInContext.rank, pathToFirstThoughtInContext, rankedContextOfCurrentContext, contextChainTillFirstChildOfContext, true)
+  }
 
   const firstChild = getSubThought(thoughtViewPath || RANKED_ROOT, showHiddenThoughts)
-  if (!ignoreChildren && firstChild) {
-    return { nextThoughts: unroot(path.concat(firstChild)), contextChain: [] }
-  }
+  const nextSibling = () => thoughtNextSibling(value, thoughtViewContext, rank, showHiddenThoughts)
```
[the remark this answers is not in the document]
anmolarora1: That's right. Thanks for correcting.

src/util/nextThought.js
anmolarora1: Do you think we can create a util function that returns the last item from a context?

src/util/nextThought.js
```diff
@@ -42,9 +43,15 @@ const isValidContextView = path =>
  * @returns {array | null} rankedContext
  */
 const contextWithThoughtRank = contextInfo => {
```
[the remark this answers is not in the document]
raineorshine: `cursor` contains the rank information, while `contextChain` only contains context. Unfortunately, as long as those data structures are separate I think recreating a `Path` from `contextChain` might be inherently inefficient.

src/util/nextThought.js
```diff
@@ -34,8 +34,7 @@ import { equalArrays } from './equalArrays.js'
  */
 const isValidContextView = path =>
   path.length &&
-  isContextViewActive(pathToContext(path)) &&
-  getContexts(head(path).value).length > 1
+  isContextViewActive(pathToContext(path))
```
raineorshine: Since this has been simplified, perhaps we can use `isContextViewActive` directly and avoid the extra function definition.

src/util/nextThought.js
```diff
@@ -126,6 +125,14 @@ const nextInContextView = (value, rank, path, rankedContext, contextChain, ignor
   const context = pathToContext(rankedContext)
   const firstChild = perma(() => firstChildOfContextView(path, thoughtIndex))
 
+  const contextWithoutChildren = contextChain.length === 1 &&
+  isContextViewActive(contextChain[0]) &&
+  getContexts(head(pathToContext(contextChain[0]))).length <= 1
```
raineorshine, line 130: Splitting a long logical statement onto separate lines is great, just make sure to indent the second and third lines.

src/util/nextThought.js
```diff
@@ -241,6 +240,15 @@ export const nextThought = path => {
...
   const rankedContext = rootedContextOf(path)
   const contextChain = splitChain(path, contextViews)
+  const context = pathToContext(rankedContext)
+
+  if (isValidContextView(rankedContext) || isValidContextView(path)) {
+    return nextInContextView(value, rank, path, rankedContext, contextChain)
+  }
+  else {
+  // not a valid context view. Try navigating thoughts
+    return nextInThoughtView(value, context, rank, path, contextChain)
```
raineorshine, line 247: Please use the ternary operator to express both cases as a single return statement.

src/util/nextThought.js
```diff
@@ -241,6 +240,15 @@ export const nextThought = path => {
   const { value, rank } = head(path)
   const rankedContext = rootedContextOf(path)
   const contextChain = splitChain(path, contextViews)
+  const context = pathToContext(rankedContext)
+
+  if (isValidContextView(rankedContext) || isValidContextView(path)) {
+    return nextInContextView(value, rank, path, rankedContext, contextChain)
+  }
```
raineorshine, line 244: Great, this looks better.

src/util/nextThought.js
```diff
@@ -237,6 +237,9 @@ const nextInThoughtView = (value, context, rank, path, contextChain, ignoreChild
 /** Gets the next thought whether it is a child, sibling, or uncle, and its respective contextChain */
 export const nextThought = path => {
   const { contextViews } = store.getState()
+  if (!path) {
+    path = RANKED_ROOT
+  }
```
raineorshine, line 236: This could be more cleanly specified as a default argument.

src/util/nextThought.js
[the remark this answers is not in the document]
anmolarora1: > `contextWithThoughtRank` is getting very computationally expensive with `equalArrays` nested in `map`. Are you sure this is necessary? Does the rank need to be recreated at every level, or is only the rank of the head thought used? I think we need to do some deeper analysis of the requirements and performance of this function. Let me know how you'd like to approach that.

We need to have the rank created at every level for the duplicate thoughts. The `setCursor` action might work with incorrect ranks sometimes, as it did in [this case](https://github.com/cybersemics/em/pull/535#discussion_r418649603), but it's important to have the correct rank information to avoid unexpected issues. 
We can improve the performance by 
1) By memorizing the `contextWithThoughtRank` function
2) Albeit not the best way, but using array.join() to compare arrays if it works for our use case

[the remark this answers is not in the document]
anmolarora1: > `cursor` contains the rank information, while `contextChain` only contains context. Unfortunately, as long as those data structures are separate I think recreating a `Path` from `contextChain` might be inherently inefficient.

I agree with that. Do you think there's an alternative approach to solve this with the given data structures?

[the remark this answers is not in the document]
raineorshine: > We need to have the rank created at every level for the duplicate thoughts. The setCursor action might work with incorrect ranks sometimes, as it did in this case, but it's important to have the correct rank information to avoid unexpected issues.

Yes, that's true. I was asking because I wasn't sure how the result of `contextWithThoughtRank` was used. If it was used to generate the new `cursor`, then yes it would need the rank at each level. If it was used for some other check, it's possible that it does not need it. I can do a more thorough analysis but I was hoping you could articulate exactly how the result of `contextWithThought` is used so that we can make an informed decision.

> By memorizing [sic] the contextWithThoughtRank function

Memoization is a good choice when you have necessary computation that is called multiple times with the same inputs. In this case I'm arguing that the computation is not necessary. Rather than implement an inefficient algorithm and then memoize it, I suggest optimizing the algorithm first.

> Albeit not the best way, but using array.join() to compare arrays if it works for our use case.

Not a bad idea, although unfortunately it looks like `Array.prototype.join` is considerably slower:
https://jsperf.com/array-string-equality-with-join/

> I agree with that. Do you think there's an alternative approach to solve this with the given data structures?

That is my intuition, but I would like you to investigate this. I have some other priorities I would like to focus on. If you are unable to find a solution, at that point I can allocate some time to investigate more deeply.

I would start by identifying exactly how the result of `contextWithThoughtRank` is used, then look for ways that we already have some of the rank information  or more efficient ways to get it.

src/util/nextThought.js
raineorshine: No, the concept of "first" and "last" assumes that the list is ordered, and a `Context` should be considered unordered. It is implemented as array for simplicity. A more advanced implementation would probably represent it as a `Set` and have some serialization logic for persistence.

As mentioned, the codebase unfortunately does not respect the unordered property of `Contexts` in many cases.

`Paths` and `Children` are ordered.

src/util/nextThought.js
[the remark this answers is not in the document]
anmolarora1: The result of `contextWithThoughtRank` in `context views` when the cursor down action should traverse the thoughts as if it's a `thought view`. An example - 

```
▼ a 
     ᐁ m
         ∙ x 		
         ▼ b
             ∙ n
             ∙ o
```

When the cursor is at `a.m~.b.n` or further deep down in the context view, the `cursorDown` action needs to work the same way as it does in the `thought view`.If there are any subsequent children (`firstChild`) of the thought in focus, the rank information isn't needed (added in [this commit](https://github.com/cybersemics/em/pull/535/commits/bbd219fb54273a0dd15a65b4f247f854733d29f9)). However, if we need to determine the `nextSibling` or `nextUncle` of the thought, we need the absolute path with rank information, and hence, `contextWithThoughtRank` is called.

src/util/nextThought.js
```diff
@@ -0,0 +1,252 @@
...
+
+    const nextSibling = nextSiblingContext(rank, context, thoughtIndex, showHiddenThoughts)
+    const rankedContextHead = head(rankedContext)
+
+    return !ignoreChildren && firstChild() ? {
+      nextThoughts: unroot([...path, firstChild()]),
+      contextChain: []
+    }
+      : nextSibling ? {
+        nextThoughts: [...nextSibling, rankedContextHead],
```
raineorshine, line 167: I'm going to start here, where `nextSibling` is used, and work my way backwards.

`nextSibling` is the result of calling `nextSiblingContext`, and it is the only place `nextSiblingContext` is used, so we don't have the added complexity of multiple cases.

[`nextSiblingContext` calls `contextWiththoughtRank`](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L61), which is the function that laboriously converts a `Context` to a `Path`. (`contextWithThoughtRank` is used in other places, but for now I am only considering it as it is used in `nextSiblingContext`.)

The problem is that `contextWithThoughtRank` goes through all the trouble of recreating ranks, when we already have all of the ranks, not including the next sibling, in `rankedContext` [that is provided to `nextInContextView`](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L131). Recreating the rank at every level is unnecessary. We just need to concatenate the correct `Child` onto `rankedContext`.

This is the kind of optimization I'm taking about, and may apply to the other uses of `contextWithThoughtRank`. It's usually just the current level where we may need to get a thought or rank that we don't have. All of the ancestors are being passed around in `path` and `contextRanked`.

anmolarora1: In this particular case that you've pointed out, we're not computing the ranks of the entire path. As you mentioned, [`nextSiblingContext calls contextWithThoughRank`](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L61), the ranks are only computed for the **context** that's being passed, which is the `nextSibling` and its rank information is not included in the `rankedContext`.
The only case where we're computing the `ranks` for the full path is when we need to compute the [`nextUncleInThoughtView` within a context view](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L194). Reason for the same is mentioned [here](https://github.com/cybersemics/em/pull/535#discussion_r421526834)

raineorshine: > In this particular case that you've pointed out, we're not computing the ranks of the entire path. As you mentioned, [`nextSiblingContext calls contextWithThoughRank`](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L61), the ranks are only computed for the **context** that's being passed, which is the `nextSibling` and its rank information is not included in the `rankedContext`.

The rank of the `head` of the next sibling does need to be calculated. I agree with that. But the ranks of the rest of the sibling context should already be known. After all, it is the sibling, so it has the same ancestors as `path`. It seems like we only need to calculate one rank, the sibling, not the rank for every thought in the context. We just need to copy them over from `path`. If this is the case, It seems unnecessary to [iterate through `contextInfo.context`](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L38).

Here is the case I am considering. Set the cursor on `a.b.c.d.x~` and activate `cursorDown`. You'll see that [`matchedContext`](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L40-L44) gets called four times (for each ancestor of `x`).

```
- a
  - b
    - c
      - d
        - x
      - e
        - x
```

> The only case where we're computing the `ranks` for the full path is when we need to compute the [`nextUncleInThoughtView` within a context view](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L194). Reason for the same is mentioned [here](https://github.com/cybersemics/em/pull/535#discussion_r421526834)

I'd like to confirm the exact case you're talking about, but if [`getPathFromContextChain` is called](https://github.com/cybersemics/em/blob/bbd219fb54273a0dd15a65b4f247f854733d29f9/src/util/nextThought.js#L194), then `contextWithThoughtRank` will be called and the same redundant calls to `matchedContext` will occur.

Perhaps I am missing something, but more than happy to discuss either way!

Issue #518, opened 2020-04-05
Prevent line break at superscript 


raineorshine:
![2A67D5C2-DA8A-4CDF-94EB-A0CC2D323C40](https://user-images.githubusercontent.com/750276/78498743-5dc7af00-7709-11ea-8e95-a7fd374899c9.jpeg)

## Steps to Reproduce

- a word 
- b [cursor]
  - a word

Convert “b” into a table (click the table icon in the toolbar)

## Current Behavior

The “2” superscript is on a separate line from “word”. 

## Expected Behavior

The “2” superscript should be positioned immediately to the right of “word” on the same line.

incora-dev:
Spent: 1h

PR #549, opened 2020-04-13
fix linebreak at superscript for table view

incora-dev:
#518

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -2395,9 +2395,6 @@ code, .code {
 .table-view > .children > .child > .thought-container > .thought-annotation > .subthought {
   word-break: normal;
 }
-.table-view > .children > .child > .thought-container > .thought {
-  margin-right: 2em;
-}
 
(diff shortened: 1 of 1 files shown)


src/App.css
```diff
@@ -2396,6 +2396,13 @@ code, .code {
   word-break: normal;
 }
 
+@media screen and (min-width: 600px) {
+  .table-view > .children > .child > .thought-container > .thought,
+  .table-view > .children > .child > .thought-container > .thought-annotation {
+    margin-right: 1rem;
+  }
+}
```
raineorshine, line 2404: @incora-dev The root issue was that a style was being applied to `.thought` but not `.thought-annotation`. `.thought-annotation` is the thought overlay that contains the superscript, so it should use the same layout styles as `.thought`.

In this case I decided I liked some margin on larger screens.

raineorshine:
Thanks!

Issue #550, opened 2020-04-13
Fix ArrowUp/ArrowDown in multi-line thought


raineorshine:
## Current Behavior
When the browser selection is on the second or greater line of a multi-line thought, hitting `ArrowUp` moves the selection to the previous thought.

## Expected Behavior
Hitting `ArrowUp` should move the selection up one line (default browser behavior).

e.g.

```
- This is a thought that
  is no|w wrapping.
```

Hitting `ArrowUp` should result in:

```
- This |is a thought that
  is now wrapping.
```

The inverse behavior should occur for `ArrowDown`.

incora-dev:
@raineorshine 2h

Issue #333, opened 2020-02-15
ArrowUp moves selection to beginning of previous sibling


raineorshine:
## Current Behavior

Hitting `ArrowUp` moves the selection to the beginning of the previous thought.

## Expected Behavior
Hitting ArrowUp should move the selection to the last line of the previous thought, directly up. This mimics the default browser behavior when hitting `ArrowUp` within a multi-line thought (although this is currently broken. See: #550).

e.g.

```
- This is a thought that
  is now wrapping.
- Test|
```

Hitting `ArrowUp` should result in:

```
- This is a thought that
  is n|ow wrapping.
- Test
```

raineorshine:
When a shortcut keycode or gesture is detected, the cursor and application state may or may not allow for that shortcut to be executed (for example trying to indent a thought that is already indented all the way). If `canExecute` returns `true`, then `e.preventDefault()` will be called on the native event and the `exec` function of the shortcut will be called.

This statement returns `false` so that in prose view the default browser behavior for the up arrow is executed. This allows the browser selection to move up one line (browser default) instead of moving the cursor up to the previous paragraph (the cursorUp action-creator). It has the intended effect describe above. We need to change the behavior when moving the cursor between thoughts while preserving the correct up arrow behavior when moving the browser selection within a thought.

Athuln:
Presently an unexpected behaviour i came across was in a multi line thought ,when cursor is presently at second line, and up arrow is pressed cursor should go to the first line ,but the cursor jumps to the thought present above. 
Same is the case for down arrow,when down arrow is pressed cursor goes to start of next thought,but cursor should go to the line below.Are they intended behaviours.

raineorshine:
Good catch! No, that is not intended. I think we need the requested behavior implemented not just in prose view, but for any multi-line thought.

raineorshine:
I added that as a separate issue: #550.

PR #558, opened 2020-04-15
Fix arrow up arrow down for multiline thought

incora-dev:
#550


raineorshine:
Thanks! We're close. just one case that is incorrect.

- [x] ArrowDown from line 1
- [ ] ArrowDown from line 2
- [x] ArrowUp from line 2

## Steps to Reproduce

ArrowDown on the following:

```
- This is a thought that 
  is |now wrapping.
- Test
```

## Expected Behavior

```
- This is a thought that 
  is now wrapping.
- Tes|t
```

## Current Behavior

```
- This is a thought that 
  is now wrapping.|
- Test
```

## Notes

In this case, the cursor should go to the next thought, like the current behavior. Getting the exact offset can be done either in this issue or in #333. Either way it should not be moving the selection to the end of the element.

incora-dev:
@raineorshine The main problem of this task is detecting a caret position in contenteditable block. Ideally, getting line number and offset from the left side (but I didn't found any good solution for that). Maybe you have some ideas have we can solve it

I have written some function that checks if cursor on the first or last line of contenteditable block, but it doesn't work pretty good 
```
export const isCursorOnEdgeMultiline = () => {
  const range = getSelection().getRangeAt(0)
  const span = document.createElement('span')
  range.insertNode(span)
  const { offsetTop, parentNode, parentElement, offsetHeight } = span
  const isTop = ((offsetTop - parentNode.offsetTop) === 0)
  const isBottom = ((offsetTop - parentElement.offsetTop + offsetHeight + 20) >= parentElement.offsetHeight)
  span.parentNode.removeChild(span)
  return { isTop, isBottom }
}
```

raineorshine:
Yes, that sounds right. I haven't worked with calculating selection position before, so I don't have any ideas right now. I would assume that people have posted similar problems on StackOverflow.

incora-dev:
What exactly is wrong? Could you confirm that case `ArrowDown from line 2` still has the same behavior?

raineorshine:
Sorry, I should have explained more. Here's what I am seeing:

- [ ] ArrowUp from line 1 should move to previous sibling
- [x] ArrowUp from line 2 should move up a line (default browser behavior)
- [x] ArrowDown from line 1 should move down a line (default browser behavior)
- [x] ArrowDown from line 2 should move to next sibling

So the issue was fixed, but there is a regression which causes the selection to go to the beginning of line 1 instead of to the previous sibling when the selection is on line 1 and you hit ArrowUp.

incora-dev:
@raineorshine To handle first case I need 3 hours more as it's a bit different behavior as arrow down. For arrow down I just needed to move cursor to the next thought but here I also need to set it in the specific position.

raineorshine:
@incora-dev Let's go with a simpler solution where ArrowUp from line 1 moves the selection to the beginning of the previous sibling (current `dev` behavior). Then we can avoid having to do the specific position calculations.

This is what WorkFlowy does, so it's not a bad solution. I updated the [list of requirements](https://github.com/cybersemics/em/pull/558#issuecomment-625900691).

raineorshine:
Looks good, thank you!

PR #559, opened 2020-04-16
Edited subthought of edited thought disappearing fix

### LINKED ISSUE #555: Edited subthought of edited thought disappears
## Steps to Reproduce

1. Create `a`
2. Create `b` as a subthought of `a`
3. Change `a` → `aa`
4. Change `b` → `bb`
5. `cursorUp`

## Current Behavior
`bb` disappears

## Expected Behavior
`bb` should not disappear

## Notes
If `bb` is added again and you switch to the context view, you can see that the original `bb` got added to `a` instead of `aa`.

Screen Shot 2020-04-14 at 10 19 12 AM

shresthabijay:
fixes #555

### DIFF
--- a/src/components/Editable.js
+++ b/src/components/Editable.js
@@ -134,8 +134,8 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
 
   // dispatches existingThoughtChange and has tutorial logic
   // debounced from onChangeHandler
-  const thoughtChangeHandler = newValue => {
-
+  // since variables inside this function won't get updated between re-render so passing latest context, rank etc as params
+  const thoughtChangeHandler = (newValue, { context, showContexts, rank, thoughtsRanked, contextChain }) => {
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine This issue was introduced by throttled `thoughtChangeHandler` function. Since the throttled function had to be stored in ref , the variables inside the `thoughtChangeHandler` couldn't get updated between re-renders. It resulted in passing old context values to `exisitingThoughtChange` dispatch which resulted this regression.

So to fix this I passed required updated values as function parameters. So every time throttled function is called it is sure to receive updated values of context, rank etc. Please review it ! Thanks

raineorshine:
Interesting. I'm still new to refs... I wouldn't have guessed that! Thanks so much.

shresthabijay:
> Interesting. I'm still new to refs... I wouldn't have guessed that! Thanks so much.

Refs can be tricky sometimes, Always happy to help.

Issue #541, opened 2020-04-12
Hamburger menu covers long Alert


raineorshine:
![475D2FCB-B1B2-498F-9A23-E879E6383BBE](https://user-images.githubusercontent.com/750276/79068419-f323ef80-7c83-11ea-9e5e-b1bb37abcb3f.jpeg)

PR #560, opened 2020-04-16
Hide Hamburger on showModal

shresthabijay:
refers #541 
@raineorshine This is fix for hiding hamburger when modal appears. Previous fix worked only for desktop. It was because wrapper of `MultiGesture` was causing problems with z-index. Now it works properly for both mobile and desktop. Please review the changes!


raineorshine:
Looks good, thanks!

Issue #498, opened 2020-04-02
Ancestor url changes along with descendant on edit


shresthabijay:
Create a `thought`  and edit it's value to become same as one of its direct `ancestor`. Then as you keep editing that descendant thought the parts of url representing both edited thought as well as that specific ancestor keeps on changing the same. Using **New Thought** shortcut immediately won't work and using **New Subthought** will cause **cursor to loose focus**.

![captured](https://user-images.githubusercontent.com/25879100/78269636-94df6a00-7529-11ea-875f-18fed7ad3cdd.gif)

In above case url becomes **"/helloasdasd/helloasdasd/"**

raineorshine:
Related: #23

raineorshine:
Fixed with unique ids

PR #567, opened 2020-04-18
Duplicate URL components fix

shresthabijay:
refers #498

### DIFF
--- a/src/reducers/existingThoughtChange.js
+++ b/src/reducers/existingThoughtChange.js
@@ -51,10 +51,10 @@ export default (state, { oldValue, newValue, context, showContexts, thoughtsRank
     ? contextOf(contextOf(thoughtsRanked)).concat({ value: oldValue, rank: headRank(contextOf(thoughtsRanked)) }).concat(head(thoughtsRanked))
     : contextOf(thoughtsRanked).concat({ value: oldValue, rank })
 
-  const cursorNew = state.cursor && state.cursor.map(thought => thought.value === oldValue && thought.rank === rankInContext
-    ? { value: newValue, rank: thought.rank }
-    : thought
-  )
+  const currentFocusedThought = state.cursor ? state.cursor[state.cursor.length - 1] : null
+
+  const cursorNew = (currentFocusedThought && currentFocusedThought.value === oldValue && currentFocusedThought.rank === rankInContext) ?
+    state.cursor.slice(0, -1).concat({ value: newValue, rank: currentFocusedThought.rank }) : state.cursor
 
   const oldPath = rankThoughtsFirstMatch(thoughtsOld, { state })
   const newPath = oldPath.slice(0, oldPath.length - 1).concat({ value: newValue, rank: oldPath.slice(oldPath.length - 1)[0].rank })


shresthabijay:
@raineorshine In previous implementation `cursorNew` was calculated updating all the thoughts whose value and rank is equal to `oldValue` and `oldRank`. So in cases like :

- a
   - m
      - apple

Let's break down what happens when we type **apple** . At first **a** is registered as the `oldValue'. And when we write **ap**, state.cursor will look something like this  ['a', 'm', 'a']. Then `cursorNew` will be ['ap', 'm', 'ap'] because all the thoughts that has value equal to `oldValue` **a** is replaced by `newValue` **ap**. So URL also becomes **ap/m/ap**. But this is not what we want.

To solve this instead of iterating and replacing all the thoughts inside `state.cursor`, we just replace the last thought. It is because that is the only thought we are actually editing. With this change it works properly. I don't know if the previous implementation is needed for something else. Please review the changes. Thanks!

raineorshine:
> Let's break down what happens when we type **apple** . At first **a** is registered as the `oldValue`. And when we write **ap**, state.cursor will look something like this ['a', 'm', 'a']. Then `cursorNew` will be ['ap', 'm', 'ap'] because all the thoughts that has value equal to `oldValue` **a** is replaced by `newValue` **ap**. So URL also becomes **ap/m/ap**. But this is not what we want.

Yes, that makes perfect sense.

> To solve this instead of iterating and replacing all the thoughts inside `state.cursor`, we just replace the last thought. It is because that is the only thought we are actually editing. With this change it works properly. I don't know if the previous implementation is needed for something else.

There are two other important cases to consider here.

1. `existingThoughtChange` can edit any thought, regardless of where the `cursor` is. So the assumption that only the last thought in the `cursor` is being edited is incorrect.

2. There may be multiple thoughts that need to be replaced in the path due to cycles and context views. **em** uses a directed graph for data and a tree for rendering.

e.g. In the example below, the context view is open on `b`, so `a` is rendered as both a parent and a child. If either are edited, they should both change.

- a
  - b~
    - a

Given that this is a non-trivial task marked as low-priority, let's table this for now.

src/reducers/existingThoughtChange.js
```diff
@@ -51,10 +51,10 @@ export default (state, { oldValue, newValue, context, showContexts, thoughtsRank
     ? contextOf(contextOf(thoughtsRanked)).concat({ value: oldValue, rank: headRank(contextOf(thoughtsRanked)) }).concat(head(thoughtsRanked))
     : contextOf(thoughtsRanked).concat({ value: oldValue, rank })
 
-  const cursorNew = state.cursor && state.cursor.map(thought => thought.value === oldValue && thought.rank === rankInContext
-    ? { value: newValue, rank: thought.rank }
-    : thought
-  )
+  const currentFocusedThought = state.cursor ? state.cursor[state.cursor.length - 1] : null
```
raineorshine, line 54: @shresthabijay For future reference, we should not use array indexing or slicing with `Path` types. That creates a tight coupling between the implementation of the `Path` and the consumer. Instead, use the accessor methods, `head` and `contextOf`.

Issue #546, opened 2020-04-12
Table: Do not expand only-child in second column


raineorshine:
- a [table view] [cursor]
  - b
    - c
      - d

When `a` is in normal view, `c` is correctly expanded since it is the only child of `b`.

When `a` is in table view, `c` should not be expanded automatically when it is an only child, because the table already shows a lot of thoughts due to column 1 being expanded.

PR #569, opened 2020-04-18
Prevent expanding only-child in second column for Table View

shresthabijay:
refers #546

### DIFF
--- a/src/util/expandThoughts.js
+++ b/src/util/expandThoughts.js
@@ -32,13 +32,15 @@ const publish = new URLSearchParams(window.location.search).get('publish') != nu
     ...
   }
 */
-export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews = {}, contextChain = [], { depth = 0 } = {}) => {
+export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews = {}, contextChain = [], { depth = 0, distanceFromTableAncestor = -1 } = {}) => {
 
   if (
     // arbitrarily limit depth to prevent infinite context view expansion (i.e. cycles)
     depth > MAX_EXPAND_DEPTH ||
     // suppress expansion if left command is held down after cursorDown
-    globals.suppressExpansion
+    globals.suppressExpansion ||
+    // if distance from the laest ancestor with table view is more than 1 then it should stop expanding
+    distanceFromTableAncestor > 1
   ) return {}
 
   const thoughtsRanked = !path || path.length === 0 ? RANKED_ROOT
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine I have added logic to prevent only-child in second column to expand. It took me a while to fully understand all the logic inside `expandThoughts`. It was tricky due to recursion. I used a variable  called `distanceFromTableAncestor` to keep track of the distance of the latest ancestor with table view. With my tests it works properly. Please review the code and also test if it introduces any regression related to expansion.  Thanks!

shresthabijay:
@raineorshine I noticed something unusual with `expandThoughts`. Since it also does recursion to the parent path, this resulting in another recursion path from that parent node as well. That is resulting in two recursion path, one starting from cursor (starting path) itself and another from direct parent of cursor. It means everything is executed twice. 

_line 95ish (expandThoughts.js)_
```javascript
    {
      // expand current thought
      [hashContext(path || [])]: true,

      // expand context
      // this allows expansion of column 1 when the cursor is on column 2 in the table view, and uncles of the cursor that end in ":"
      // RECURSION
      ...(path && path.length >= 1 && depth <= 1
        ? expandThoughts(contextOf(path), thoughtIndex, contextIndex, contextViews, contextChain, { depth: depth + 1 })
        : {})
    }
```
chrome-capture
    
I suggest that we first calculate the actual starting path from the path and then start recursion.

src/util/expandThoughts.js
```diff
@@ -32,13 +32,15 @@ const publish = new URLSearchParams(window.location.search).get('publish') != nu
     ...
   }
 */
-export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews = {}, contextChain = [], { depth = 0 } = {}) => {
+export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews = {}, contextChain = [], { depth = 0, distanceFromTableAncestor = -1 } = {}) => {
```
raineorshine, line 35: Rather than adding data to the recursion and increasing the complexity of the function, just check directly if the grandparent is a table.

```js
const isTableColumn2 = () => attribute(
  contextOf(contextOf(thoughtsRanked)), 
  '=view', 
  { state: { thoughtIndex, contextIndex } }
) === 'Table'
```

src/util/expandThoughts.js
```diff
@@ -32,13 +32,15 @@ const publish = new URLSearchParams(window.location.search).get('publish') != nu
...
 
   if (
     // arbitrarily limit depth to prevent infinite context view expansion (i.e. cycles)
     depth > MAX_EXPAND_DEPTH ||
     // suppress expansion if left command is held down after cursorDown
-    globals.suppressExpansion
+    globals.suppressExpansion ||
+    // if distance from the laest ancestor with table view is more than 1 then it should stop expanding
+    distanceFromTableAncestor > 1
```
raineorshine, line 43: This will break the `=pin` functionality. Instead of short circuiting, use `isTableColumn2` to affect the main conditional (L71). In particular, you will want to modify `isOnlyChildNoUrl` with `&& !isTableColumn2`.

raineorshine:
> With current implementation recursion only goes up to direct parent of the cursor(starting path) resulting in two recursion path. With increased number of visible thoughts, this can be troublesome. I suggest that we first calculate the actual starting path from the path and then start recursion.

Agree! Created #586.

src/util/expandThoughts.js
shresthabijay: Sure! I was trying to avoid iteration using recursion logic.

shresthabijay:
@raineorshine  I have made the requested changes. Please review it. Thanks!

src/util/expandThoughts.js
raineorshine: `attribute` involves iteration due to `.includes`, but this can be optimized when testing equality. We could create a new function to do it in O(1) like this:

```js
const attributeEquals = (context, attr, value) => {
  const firstChild = getThoughts(context.concat(attr))[0]
  return firstChild && firstChild.value === value
}
```

raineorshine: 7791ec6b383baa25861aebdccfb2cd5ccf1f620d

shresthabijay:
@raineorshine  This fix was not successful. I had to make small change. I have sent a PR https://github.com/cybersemics/em/pull/588

PR #574, opened 2020-04-19
Prevent text selection from calling cursorBack

### LINKED ISSUE #557: Text selection incorrectly calls clickOnEmptySpace
## Steps to Reproduce

- a
  - b [cursor]

1. Begin selecting the `b` text by clicking and dragging with the mouse.
2. Before releasing the mouse, move the cursor down below the thoughts.
3. Release on the empty space below the thoughts.

## Current Behavior
`cursorBack` gets triggered from the [`onClick` handler in Content.js](https://github.com/cybersemics/em/blob/dev/src/components/Content.js#L71).

## Expected Behavior
When text is being selected and the mouse is released, `onClick` should no be called.

shresthabijay:
fixes #557

### DIFF
--- a/src/components/Content.js
+++ b/src/components/Content.js
@@ -63,6 +63,11 @@ const Content = props => {
   // remove the cursor if the click goes all the way through to the content
   // extends cursorBack with logic for closing modals
   const clickOnEmptySpace = () => {
+
+    // click event occured during text selection has focus node of type text unlike normal event which has node of type element
+    // prevent text selection from calling cursorBack incorrectly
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine It took a bit of research to find this solution. The `focusNode` we get from `window.getSelection` is of type text during active text selection. So we can check the type of `focusNode` and prevent calling `cursorBack` if there is an active text selection. Please review the changes. Thanks!

raineorshine:
Beautiful! Thank you!

PR #584, opened 2020-04-21
Fix bottom clipping after font resize

### LINKED ISSUE #536: Scaled text gets cut off at bottom
When font size is increased (`scaleSize > 1`) and there are many (50-100) thoughts in the root, then they get cut off at the bottom of the window. The `<Scale>` element gets scaled but `<body>` doesn’t scale to fit. If we know the `scaleSize`, can we add the requisite padding-bottom without having to read the element height from the DOM?

@edsonayllon Could you take a look at this? I didn’t catch this bug in the original PR. 

See: https://stackoverflow.com/questions/34474826/how-to-fit-html-parent-elements-bounding-box-to-a-css-transformed-child

Related: #223

edsonayllon:
Fixes #536


src/components/AppComponent.tsx
```diff
@@ -144,7 +144,7 @@ const AppComponent: FC<Props> = (props) => {
               <Toolbar />
               {tutorial && !isLoading ? <Tutorial /> : null}
               <SplitPane
-                style={{ position: 'relative' }}
+                style={{ position: 'relative', paddingBottom: Math.pow(10, scale * 2.02) }}
```
raineorshine, line 147: Hi! I think your calculation might be off still. This increases the padding-bottom exponentially, when it should increase proportionally. When I increase the font just a few times, I get thousands of pixels of padding-bottom.

Also, we should try to decouple this from the `SplitPane`.

edsonayllon: Screen Shot 2020-04-21 at 9 02 04 PM
Screen Shot 2020-04-21 at 9 02 16 PM

It doesn't seem to be creating thousands of pixels for me.

edsonayllon: I'll add some changes

edsonayllon: `SplitPane` is what's causing the clipping.

raineorshine: > `SplitPane` is what's causing the clipping.

Oh, I see.

raineorshine: > It doesn't seem to be creating thousands of pixels for me.

Try a font size 10 steps higher. That's still within an order of magnitude so it should support that.

Unless I'm wrong, I believe the height will grow proportionally to the font size, not exponentially.

edsonayllon: Yeah, I'm going to remove that and do something else.

edsonayllon: Bottom margin should scale proportionally now. 474d4cd

edsonayllon:
Unfortunately, `flex` is required for this method. Flex fixes a lot of other issues that pop up during resize. The annotation margin was fixed in 5dca597.

raineorshine:
Sure! Happy to keep `flex` as long as we can fix the regressions.

raineorshine:
Closing due to inactivity.

edsonayllon:
Hi. Sorry, I became busy with my full-time job. 

Was this resolved? The only thing left was centering the active bullet point iirc.

raineorshine:
That's okay. I will assign this to one of my developers. Perhaps they can reuse what you've done so far.

Issue #586, opened 2020-04-21
expandThoughts: Eliminate redundant recursion


raineorshine:
#569

> @raineorshine I noticed something unusual with `expandThoughts`. Since it also does recursion to the ancestor paths at specific condition, so this is resulting in new recursion paths from these nodes as well. That is resulting in mutiple recursion paths, one starting from cursor (starting path) itself and other from ancestors. It means everything is executed multiple times.
> 
> _line 95ish (expandThoughts.js)_
> 
> ```js
>     {
>       // expand current thought
>       [hashContext(path || [])]: true,
> 
>       // expand context
>       // this allows expansion of column 1 when the cursor is on column 2 in the table view, and uncles of the cursor that end in ":"
>       // RECURSION
>       ...(path && path.length >= 1 && depth <= 1
>         ? expandThoughts(contextOf(path), thoughtIndex, contextIndex, contextViews, contextChain, { depth: depth + 1 })
>         : {})
>     }
> ```
> 
> ![chrome-capture (1)](https://user-images.githubusercontent.com/25879100/79642296-741b4380-81bc-11ea-9c7b-d6a620c7102b.jpg)
> 
> With current implementation recursion only goes up to direct parent of the cursor(starting path) resulting in two recursion path. With increased number of visible thoughts, this can be troublesome. I suggest that we first calculate the actual starting path from the path and then start recursion.

PR #588, opened 2020-04-23
Changed isTableColumn2 to isTableColumn1 (Table expand fix)

shresthabijay:
@raineorshine Previous PR #569 used `isTableColumn2` . But it didn't actually solve the problem. It should be `isTableColumn1` because `expandThoughts` function expands the current path by default (except if it has reached maximum depth) and the filters its children.

PR #589, opened 2020-04-23
Redundant recursion and Table expand fix

shresthabijay:
refers #586

### DIFF
--- a/src/util/expandThoughts.js
+++ b/src/util/expandThoughts.js
@@ -11,10 +11,12 @@ import {
 import {
   contextChainToPath,
   contextOf,
+  equalPath,
   excludeMetaThoughts,
   getChildPath,
   getThoughtsRanked,
   hashContext,
+  isDescendant,
   isURL,
   pathToContext,
   publishMode,
@@ -33,7 +35,7 @@ import attributeEquals from '../selectors/attributeEquals'
     ...
   }
 */
-export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews = {}, contextChain = [], { depth = 0 } = {}) => {
+const expandThoughtsRecursion = (path, thoughtIndex, contextIndex, contextViews = {}, contextChain = [], cursor, { depth } = {}) => {
 
   if (
     // arbitrarily limit depth to prevent infinite context view expansion (i.e. cycles)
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine I have solved the issue by starting recursion from the grandparent of the given path(cursor). I had to add some checks to make sure cursor and its direct ancestor always expands. Please review the changes. Thanks !

Also I happened to branch out of another fix branch #588  . So merging this PR will solve both the issues.

src/util/expandThoughts.js
```diff
@@ -80,31 +82,36 @@ export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews =
   return (isOnlyChildNoUrl || isTable() || pinChildren() || publishPinChildren
     ? children
     : children.filter(child => {
+      const childPath = (path || []).concat(child)
```
raineorshine, line 85: Have you tested this in the context view? I'm skeptical that this works in all cases, but I could be wrong. The `child` is not necessarily the last item in a path. It seems like we might be mixing up `Path` types with `Context` types here.

src/util/expandThoughts.js
```diff
@@ -80,31 +82,36 @@ export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews =
...
+      const isCursor = equalPath(childPath, cursor)
```
raineorshine, line 86: Since only `child` changes on each iteration, you only need to compare `child` with `head(cursor)`.

src/util/expandThoughts.js
```diff
@@ -80,31 +82,36 @@ export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews =
...
+      const isDirectAncestorOfCursor = isDescendant(childPath, cursor)
```
raineorshine, line 87: This is an expensive operation within a loop. It seems like we've traded extra recursion for extra computation. I'm not really satisfied with this solution.

src/util/expandThoughts.js
```diff
@@ -80,31 +82,36 @@ export const expandThoughts = (path, thoughtIndex, contextIndex, contextViews =
...
   )
 }
+
+export const expandThoughts = (cursor, thoughtIndex, contextIndex, contextViews = {}, contextChain = [], { print = false } = {}) => {
+
+  const startingPath = contextOf(contextOf(cursor || []))
+  const startingDepth = startingPath.length - (cursor || []).length
+
+  return expandThoughtsRecursion(startingPath, thoughtIndex, contextIndex, contextViews, contextChain, cursor || [], { print, depth: startingDepth })
+}
```
raineorshine, line 117: Since there is minimal logic here, I would prefer default options on the main recursive function rather than a separate function to jump start the recursion.

shresthabijay:
> Given that this function will probably be superseded by `getFlatArray`, I suggest we fix the table issue in a separate PR and put this on hold rather than investing more time in it.
> 

@raineorshine Yes you are right! It will get superseded by `getFlatArray` some point in the future. Let's keep this in hold for now.

raineorshine:
Closing; we can re-open if needed.

Issue #579, opened 2020-04-20
Change sort toolbar icon


raineorshine:
The solid arrow deviates from the line drawing style of the rest of the icons. Try to find a similar icon at thenounproject that is a better match.

Issue #578, opened 2020-04-20
Table toolbar icon should be slightly smaller


raineorshine:
To match other icons. 

Time Estimate: 30 min

PR #607, opened 2020-04-29
Tweak toolbar icon styles

pushkov-fedor:
#578, #579


pushkov-fedor:
Thank you for remarks. I changed sort icon. 
I'm not sure that i understand your issue about vertical alignment. On my laptop, as i see, it's centered, also i only changed width and height of Icon and didn't edit styles, which are responsible for alignment. I compared 2 icons and didn't find difference. Could you please point me out, what's wrong with them?
Screenshot from 2020-04-30 20-01-42
Screenshot from 2020-04-30 20-01-52

raineorshine:
The table was just a couple pixels too high. I decided to thicken up the lines so that they don't get too thin after scaling. I also lined up the arrow better on the sort icon while I was at it.

Thank you!

raineorshine:
Fix table and sort icons.

Issue #600, opened 2020-04-27
Note: Cursor doesn't change on select


pushkov-fedor:
## Steps to Reproduce
Toggle note icon in the toolbar and switch between different context

- a
[note]
  - b
  - c

## Current Behavior

When the cursor is on `b`, `cursor` is in `[root, a]` context. If you click on note, `cursor` will stay the same: `[root, a]`. The note icon is not highlighted. The `cursor` will be changed only after clicking on `a`.

## Expected Behavior

Click on note should cause changing of current context change to [root] and highlight note icon

PR #610, opened 2020-04-29
Fix Note:Cursor doesn't change on select

pushkov-fedor:
Issue #600
Tried change context by dispatching setCursor in onFocus event of Note component. It solves the issue, but this action sets focus on parent Thought of the Note. As i see, dispatch action triggers render of Thought component, where isEditing is computed by`isEditing = equalPath(cursorBeforeEdit, thoughtsResolved)`, i think it is the reason why Thought, parent of Note, is getting focused. We need somehow prevent this focus event, do you have any ideas?

Timeout with focus was added just for test


pushkov-fedor:
@raineorshine  I added additional check before `setSelection`, now Thought isn't getting focus after click on Note component. I found one bug. It works pretty well, when we have Note with text. But if we click/tap on empty note, cursor will be set and reset immediately. I'm trying to find the reason of this behavior. What do you think about this?

raineorshine:
> I found one bug. It works pretty well, when we have Note with text. But if we click/tap on empty note, cursor will be set and reset immediately. I'm trying to find the reason of this behavior. What do you think about this?

Do you have Redux DevTools? I'm seeing two calls to `setCursor` in this case. The first has `noteFocus: true` and the next does not. I don't know what is calling the `setCursor` but if you do a stack trace I'm sure you can find out. Perhaps it's the blur event.

I'm also noticing a bug in `dev` where the cursor disappears when a note is first added. I think it has to do with `cursorBeforeEdit` not being updated when the focus changes between the thought and its note. We can create a separate issue for that, but I want to make sure it's not interfering with your work on this issue.

pushkov-fedor:
@raineorshine Okay, i'll fix them, thanks for review. Can i log extra 0.5 hour for this? Is it okay?

raineorshine:
Yes, an extra 1-2 hours is fine since it was a little more complicated than expected.

pushkov-fedor:
Fixed, #600 is also here

pushkov-fedor:
@raineorshine It's so strange. I've tested it again on local and it works.

- thought
[note] **(cursor is here)**

After click on Note button:

- thought **(cursor is here)**
[note]

And reverse for the next situation. Please tell me, if i don't understand something about desired behavior. I've tested in on PC and Android mobile device.

raineorshine:
Okay, thanks for confirming! It must be a platform issue. I'll see if I can fix up the issues on Desktop and Mobile Safari. Then you can test them out again on your platforms. What desktop browser are you using btw?

raineorshine:
Okay, working now on my end. Will you test 325e640 at your convenience? Thanks, and have a good holiday!

pushkov-fedor:
It's Chrome on both devices. Yeah, i will test it tomorrow on mobile and PC.

pushkov-fedor:
I've tested it. It works on both mobile and desktop Chrome

raineorshine:
Thanks! I'll merge soon.

PR #611, opened 2020-04-30
Prevent split if focus is not on Editable

### LINKED ISSUE #609: TUTORIAL_STEP_SECONDTHOUGHT: First thought split
## Steps to Reproduce

1. Clear data and refresh
2. Click "Start Tutorial"
3. When prompted to create your first thought, type "Hello World"
4. Hit Next
5. When prompted to create your second thought, hit Enter.

## Current Behavior

- Hell
- o World [cursor]

## Expected Behavior

- Hello World
- [cursor]

shresthabijay:
fixes #609

### DIFF
--- a/src/shortcuts/newThought.js
+++ b/src/shortcuts/newThought.js
@@ -52,12 +52,15 @@ const exec = (e, { type }) => {
   }
 
   const offset = window.getSelection().focusOffset
+  // making sure the current focus in on the editable component to prevent splitting
+  const isFocusOnEditable = document.activeElement.classList.contains('editable')
+
   const showContexts = cursor && isContextViewActive(contextOf(cursor), { state: store.getState() })
 
   // split the thought at the selection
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine This issue #609  took quite a while to solve. It was very confusing at the beginning. The behavior of thought splitting even when the focus is away from the input was happening only during the tutorial. It was not happening when the tutorial was off. I noticed that  `cursorOffset` was changing to 4 when we reach second tutorial. At first I tried looking into tutorial logic to see if anything mutates `cursorOffset`. But I couldn't find anything. 

Later I found that when clicking on "**Next**" button, the focus was set to that specific text node and the `focusOffset` was set to 4. Also `newThought` had to check to prevent splitting if focus was not on `Editable` component. The fix is fairly easy. We just check if the current `activeElement` is an `Editable` component. If not we prevent splitting.

raineorshine:
Excellent work, Bijay. Appreciate the thoroughness and clear solution. Thanks!

Issue #437, opened 2020-03-23
Remove autoProse


raineorshine:
The `autoProse` feature will automatically convert a context to prose view when its children have enough characters. This was an experimental feature that ended up being too confusing for users. 

Remove all use of `autoProse` and let the the user activate prose view from the toolbar.

raineorshine:
Seemingly caused by embedded HTML. Still need steps to reproduce.

PR #615, opened 2020-05-01
remove auto prose

anmolarora1:
Refers to issue #437 
Removes `autoProse` util and all its usages


raineorshine:
Thanks! I ended up implementing this and forgot to push my changes. Oops.

Issue #597, opened 2020-04-26
Superscript too far right on <i>


raineorshine:
The superscript appears too far to the right when the thought is wrapped in `<i>`, `<b>`, or `<u>`. It should appear immediately after the text of the thought, just like without the formatting a tag. 

Must work for both single and multi-line thoughts.

PR #616, opened 2020-05-01
Fixes superscript positioning for thoughts with embedded HTML tags

anmolarora1:
Refers to issue #597  

Problem - We're using `thoughtAnnotation` to display the number of duplicates of a thought. The text inside `thoughtAnnotation` didn't interpret the markup and rendered it with character codes, hence adding extra space and pushing the superscript(no. of duplicate thoughts) further.

Solution - To position the number(superscript) correctly, the text within `thoughtAnnotation` should exactly match it's corresponding thought, with all formatting preserved. So we've used `dangerouslySetInnerHTML` function to correctly format the text. 

**The assumption with this solution is that the text which is being passed to the function has been pre-sanitized and there's no risk of XSS**


src/components/ThoughtAnnotation.js
```diff
@@ -31,6 +31,14 @@ import StaticSuperscript from './StaticSuperscript'
 import ContextBreadcrumbs from './ContextBreadcrumbs'
 import UrlIcon from './icons/UrlIcon'
 
+const getSubThoughtTextMarkup = (isEditing, subthought, thoughtMeta) => ({
```
raineorshine, line 34: I reeeeeeeally would like you to get in the habit of adding JSDOC comments to EEEEEEEEEVERY function declaration 😊😊😊

Issue #539, opened 2020-04-12
Sort empty thought to top 


raineorshine:
Empty thought should always be sorted to top in sorted context. Currently it is sorted after thoughts that start with certain punctuation e.g. `*`

e.g. 

- My Sorted Context
  - `*`
  - [New Thought]
  - A
  - B

incora-dev:
@raineorshine Estimate: 1.5h

raineorshine:
@incora-dev This is already assigned to another developer.

PR #618, opened 2020-05-04
Sort empty thought to top

aqkhan:
References to #539

### DIFF
--- a/src/util/compareThought.js
+++ b/src/util/compareThought.js
@@ -19,10 +19,6 @@ export const compare = (a, b) => a > b ? 1 : a < b ? -1 : 0
 //       )
 // }
 
-export const compareEmptyThoughts = (a, b) => {
-  return a === '' ? -1 : 1
-}
-
 /* A comparator that sorts numbers ahead of non-numbers */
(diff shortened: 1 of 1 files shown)


src/util/compareThought.js
```diff
@@ -43,7 +43,7 @@ export const compareLowercase = (a, b) => compare(lower(a), lower(b))
 export const comparePunctuationAndOther = (a, b) => {
   const aIsPunctuation = regexPunctuation.test(a)
   const bIsPunctuation = regexPunctuation.test(b)
-  return aIsPunctuation && !bIsPunctuation ? -1
+  return (aIsPunctuation && !bIsPunctuation) || (a === '') ? -1
```
raineorshine, line 46: I would prefer this in its own comparator function to keep things clean. I'm very proud of my comparator composition 😂. Please define `compareEmpty` and add it to the `makeOrderedComparator` argument list.

aqkhan: Sure.

aqkhan:
Closing due to failed updates.

Issue #25, opened 2019-09-26
Rerank subthoughts when sibling collision


raineorshine:
When moving (dragging-and-dropping) a thought into a context, that context may have a thought with the same value, which may have different subthoughts (children). When this occurs, the subthoughts from the duplicate thoughts are merged.

## Current Behavior
Subthoughts are combined into the destination (dropped) thought without modifying their rank, potentially creating duplicate ranks and/or resulting in an unexpected order.

## Expected Behavior
The source subthoughts (drag) should be re-ranked starting with the rank of the last destination subthought (drop) + 1.

e.g. Given the following:

``` 
- A
  - B
    - x (rank 7)
    - y (rank 8)
- M
  - B
    - z (rank 4.5)
``` 

Moving `/M/B` into `/A` should result in the following:

``` 
- A
  - B
    - x (rank 7)
    - y (rank 8)
    - z (rank 9)
- M
```

PR #631, opened 2020-05-06
Prevent rank change if thought moves within same context

### LINKED ISSUE #614: existingThoughtMove: Old context not deleted
Sometimes when a thought is moved, its descendants are not properly moved from the old context. 

Here `Week` and all its descendants were moved to `/Today/Week`. `A` incorrectly still exists in the `/Week` context. 

6C55B891-DB4E-47A4-B34C-BBD9DBB71029

## Possible causative factors;

- [x] ~Root?~ No
- [x] ~Different thoughts with same hash?~ No
- [x] ~Race condition?~ No
- [x] ~Moving to/from ancestor vs sibling?~
- [x] ~Drag only or indent?~ Both
- [x] ~Rank?~ Yes!

## Steps to Reproduce
- a
  - b
    - c

1. Move `b` before `a`:

- b
  - c
- a

2. Move `b` after `a`:

- a
- b
  - c

3. Indent or move `b` within `a`:

- a
  - b
    - c

## Expected Behavior

The thoughts should return to their original state:

- a
  - b
    - c

## Current Behavior

`c` exists in two contexts, the old context and the new context of the last move:

- a
  - b
    - c~
      - a • b
      - a

shresthabijay:
Fixes #614.

@raineorshine The reason  for  `contextIndex` not getting updated as `thoughtIndex` was because of the logic that prevented `contextIndexDescendantUpdates` from getting new updates when `sameContext` is true. That was in place to prevent unnecessary computation and that is the right thing to do. So instead I prevented updating ranks for the first depth child  (i.e updating `thoughtIndex`)  if thought moves within same context. Please review it. Thanks!

### DIFF
--- a/src/reducers/existingThoughtMove.js
+++ b/src/reducers/existingThoughtMove.js
@@ -84,8 +84,8 @@ export default (state, { oldPath, newPath, offset }) => {
       // remove and add the new context of the child
       const contextNew = newThoughts.concat(contextRecursive)
 
-      // update rank of first depth of childs
-      const movedRank = newLastRank ? newLastRank + i : child.rank
+      // update rank of first depth of childs except when a thought has been moved within the same context
+      const movedRank = !sameContext && newLastRank ? newLastRank + i : child.rank
       const childNewThought = removeDuplicatedContext(addContext(removeContext(childThought, pathToContext(oldThoughtsRanked), child.rank), contextNew, movedRank), contextNew)
 
       // update local thoughtIndex so that we do not have to wait for firebase


shresthabijay:
> Unfortunately this causes duplicate ranks when merging thoughts of (valid) duplicate child values. Follow the steps in #25 to reproduce.

@raineorshine This won't be a problem. When a thought is moved within same context then we don't have to worry about merging. Also inconsistency was introduced when a thought was moved with in its own context and it's descendant thoughts rank was increased in `thoughtIndex` while there was a logic preventing update of `contextIndex` when `sameContext` was true.

For example:

```
- A
  - B
     - D
     - E
  - F
```

If we move A/B after A/F  within same context (i.e `sameContext` is true) we shouldn't update rank of A/B.

``` 
- A
  - B
    - x (rank 7)
    - y (rank 8)
- M
  - B
    - z (rank 4.5)
``` 
While in this case when M/B is moved to A/B  to different context (i.e `sameContext` is false) , it updates ranks and merges `subthoughts`.

I tested it and it's working for me. Please let me know if I am missing something or is not working for you. Thanks!

raineorshine:
You're absolutely right. My apologies. I must have been testing on the wrong commit! 🙏

Issue #540, opened 2020-04-12
Sort thought during editing


raineorshine:
When editing a thought in a sorted context, it should be dynamically sorted to the correct place in real-time. 

This must be done without interrupting editing, i.e. losing the browser selection.

pushkov-fedor:
@raineorshine I work on solution, but there's a moment i would like to clarify. Loos like we need to rerender Subthoughts component while editing. In Editable component dispatching render action or calling setCursorOnThought function sort thoughts, but cause some unnecessary side effects. What do you think about it? Is it correct way to start solving this issue? I'm just a little confusing about thoughtsRanked, contexts, thoughtsResolved entities and their impact on render :)

raineorshine:
> Loos like we need to rerender Subthoughts component while editing. 

This is a tough decision, which I'm quite ambivalent about. We can't re-render the Subthoughts component of a thought we are editing because we would lose the browser selection. I don't trust that we can restore the browser selection without interrupting the user's typing, which is not acceptable.

I'm almost tempted to do direct DOM manipulation. As bad as that is, it might be the least disruptive, and it will get re-rendered in sorted order anyway.

I'm not sure which approach will have the most negative side effects.

> In Editable component dispatching render action or calling setCursorOnThought function sort thoughts, but cause some unnecessary side effects.

Yes

> What do you think about it? Is it correct way to start solving this issue? 

It's very hard for me to tell. I'm no React expert. Maybe @shresthabijay can offer an opinion.

> I'm just a little confusing about thoughtsRanked, contexts, thoughtsResolved entities and their impact on render :)

That's understandable. There is a steep learning curve. Let me know know if there's anything specific about the architecture I can explain for you. Although the documentation is quite sparse, it might be worth going over [Understanding the Data Model](https://github.com/cybersemics/em/wiki/Technical-Overview#2-understanding-the-data-model).

shresthabijay:
> It's very hard for me to tell. I'm no React expert. Maybe @shresthabijay can offer an opinion.

It's quite tricky. I tried re-rendering `Subthoughts` and  it is not causing loss of input focus but it definitely is causing weird `cursorOffset` behviour. But we can possibly solve that by storing `cursorOffset` within component, so after a re-render it can restore the offset properly. But still current architecture is made in a way that doesn't favor re-rendering top level sub thoughts on edit.

shresthabijay:
[...]

PR #637, opened 2020-05-07
add setCursor on change in Editable

pushkov-fedor:
Issue #540
@raineorshine Raine, these days i was thinking a lot about this issue and trying different options. These 2 lines of code make sorting working in someway, but break backspace logic and cause some bugs. I think i should have published it earlier in order to get different feedbacks and opinions, sorry for that. :( The task is exciting and i'm ready to continue working on it, also it gave better understanding of current codebase, but takes a lot of time. 

As i see, we just need to re-execute `getThoughtsSorted` in Subthought compoentnt and set the cursor back, but the devil in the details
I also tried dispatching `setCursor` in `onChange` method of Editable component, it sorted thoughts but Editable component lose focus  after throttled `thoughtChangeHandler` function invocation. Maybe the problem is that value of Thought changed.

What do you think about it?

### DIFF
--- a/src/components/Editable.js
+++ b/src/components/Editable.js
@@ -187,6 +187,7 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
       )) {
         tutorialNext()
       }
+      dispatch({ type: 'setCursor', thoughtsRanked })
     }
   }
 
@@ -266,6 +267,7 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
       thoughtChangeHandler(newValue, { context, showContexts, rank, thoughtsRanked, contextChain })
     }
     else throttledChangeRef.current(newValue, { context, showContexts, rank, thoughtsRanked, contextChain })
+    dispatch({ type: 'setCursor', thoughtsRanked })
   }
 
   const onPaste = e => {


raineorshine:
> The task is exciting and i'm ready to continue working on it, also it gave better understanding of current codebase, but takes a lot of time.

Good, I'm glad to hear that. This is challenging, but getting more familiar with the data model and codebase is great.

> These 2 lines of code make sorting working in someway, but break backspace logic and cause some bugs.

It's kind of hard to tell with just that small change. Is it the wrong approach, or does it just need a lot of fixes? It breaks a lot of behavior from what I can tell. I'm concerned that it breaks fundamental invariants of the throttled editing logic. 

> As i see, we just need to re-execute getThoughtsSorted in Subthought compoentnt and set the cursor back

The `cursor` should still be valid, it's the browser selection that needs to be moved back. 

I [raised concerns about being able to restore the browser selection](https://github.com/cybersemics/em/issues/540#issuecomment-620773084) without it breaking editing. It may be worth trying though, since direct DOM manipulation is problematic [as Bijay pointed out](https://github.com/cybersemics/em/issues/540#issuecomment-620966712).

> I suggest we invest only 3-4 hours to this issue. If this feature has higher priority then we can invest more . But if it takes lot of effort and major architectural changes , then we should put it in hold. We can reopen if we feel #503 will take much longer to address this issue.

I agree with Bijay on this. How much time have you put in so far, and what else you try within that window?

pushkov-fedor:
I spent 4 hours for this, figuring out what is going on by react/redux-dev-tools. I'd spent a lot of time trying to figure out, why after re-setting cursor we loose browser selection of Thought. Also I'd tried re-render Subthoughts component through callback from Editable component, but it needs more complex approach. I tried to use Sort button logic for triggering sorting function. All of these don't look like an acceptable solution, but i tried to find point to start.

raineorshine:
Thank you for digging in and trying out some possibilities. I think given the time investment it would be a good idea to put this on hold. We can look into this again after #503.

PR #638, opened 2020-05-07
Duplicate thought in a context after merge (existingThoughtMove) fix

### LINKED ISSUE #632: Duplicate thoughts within context.
When moving (dragging-and-dropping) a thought into a context, that context may have a thought with the same value but different rank the `subthoughts` are not merged.

### Steps To Reproduce

- A
  - B
    - C
- B
  - D

1. Create above structure.
2. Move root contexts **B** inside **A**.

### Current Behavior

- A
  - B
    - C
    - D
  - B
    - C
    - D

### Expected Behavior

- A
  - B
    - C
    - D

shresthabijay:
Fixes #632 
@raineorshine In the current implementation duplicate thoughts in the `subthoughtsNew`  is done by checking if both rank and value are equal. 

https://github.com/cybersemics/em/blob/f5949fb217cfa1c09ba8182e277ac1070e531f33/src/reducers/existingThoughtMove.js#L68-L69

But the rank of duplicate thoughts won't always be the same. 

``` 
- A
  - B (rank 0)
    - x (rank 7)
    - y (rank 8)
- M
  - B (rank 0)
    - z (rank 4.5)
``` 
In this example when M/B is moved to A/B, it filters out the duplicate thought because both thought B have same rank 0.

``` 
- A
  - B (rank 0)
    - x (rank 7)
    - y (rank 8)
- M
  - N (rank 0)
  - B (rank 1)
    - z (rank 4.5)
``` 
If it was something like this then it won't be filtered out causing duplicate thoughts in the context after moving M/B to A/B.

**Solution:** We only check if the value of the thoughts are equal.

```javascript
  .filter(child => child.value !== value)
```

Please review the changes! Thanks.

### DIFF
--- a/src/reducers/existingThoughtMove.js
+++ b/src/reducers/existingThoughtMove.js
@@ -1,3 +1,8 @@
+// constants
+import {
+  ID,
+} from '../constants'
+
 // util
 import {
   addContext,
@@ -72,7 +77,7 @@ export default (state, { oldPath, newPath, offset }) => {
     .filter(child => child.value !== value)
     .concat({
       value,
-      rank: (isDuplicateMerge) ? duplicateSubthought.rank : newRank,
+      rank: isDuplicateMerge ? duplicateSubthought.rank : newRank,
       lastUpdated: timestamp()
     })
 
@@ -183,6 +188,7 @@ export default (state, { oldPath, newPath, offset }) => {
     })
   }
 
+  /** Updates the ranks within the given path to match those in descendantUpdatesResult */
   const updateMergedThoughtsRank = path => path.map(
     child => {
       const updatedThought = descendantUpdatesResult[hashThought(child.value)]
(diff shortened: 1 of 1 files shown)


shresthabijay:
> I'm seeing one regression: the cursor disappears when there is a duplicate merge. It is discarding the source rank and using the destination rank, but `cursorDescendantPath` is being generated from `oldPath` which has the source rank. We may have to add another case to set `newCursorPath` correctly when the the old rank has been filtered out.

@raineorshine  I have updated the `newPath` to use the duplicate thought's rank in the new context instead of using the new rank. `Also whenever there is a duplicate merge and old cursor is in old path then I have updated rank of thoughts of `cursorDescendantPath` according to `descendantUpdatesResult`. It's working properly for me. Please review the changes. Thanks!

raineorshine:
Works great for me!

Small stylistic changes in 38af4853.

Issue #626, opened 2020-05-05
Archive: Do not count archived thoughts in superscript


raineorshine:
## Steps to Reproduce

- a
  - =archive
    - x
  - x
- b
  - x

## Current Behavior

When `showHiddenThoughts` is false, `b` has superscript 3.

## Expected Behavior

When `showHiddenThoughts` is false, `b` should have superscript 2.

PR #639, opened 2020-05-08
Archive: Do not count archived thoughts in superscript

aqkhan:
References to #626


src/components/Superscript.js
```diff
@@ -34,6 +34,10 @@ const mapStateToProps = ({ contextViews, cursor, cursorBeforeEdit, modalData, sh
     ? (props.showContexts ? contextOf(cursor || []) : cursor || [])
     : thoughtsRanked
 
+  const notInArchive = context => (
+    showHiddenThoughts || context.context.indexOf('=archive') === -1
```
raineorshine: Since `showHiddenThoughts` does not change over the course of the loop, it should be checked outside `filter` for better performance.

Issue #396, opened 2020-03-02
Archive: Alert undo


raineorshine:
- [x] Depends on #394 

When a thought is archived, dispatch an alert that says ‘Deleted “THOUGHT_TEXT. Undo”’.

“Undo” should be a link that when clicked will move the archived thought back into the context it was deleted from. It should have the same rank it had before, unless that would conflict with an existing thought, in which case it should be inserted after that thought. 

Automatically close the alert after 10 seconds.

PR #641, opened 2020-05-10
Archive: Alert undo

aqkhan:
References to #396


src/action-creators/undoArchive.js
```diff
@@ -1,12 +1,25 @@
...
+
 // action-creators
 import alert from '../action-creators/alert'
 
 export const undoArchive = ({ originalPath, currPath, offset }) => dispatch => {
 
   const state = store.getState()
 
+  const archiveContext = contextOf(pathToContext(currPath))
+  const contextOfArchive = splice(archiveContext, archiveContext.length - 1, 1)
```
raineorshine, line 25: Any reason you are using `splice` instead of `contextOf` here?

raineorshine: The names `archiveContext` and `contextOfArchive` seem to signify the same thing. Can we clarify these names so that they are semantically different? I would suggest:

Rename `archiveContext` → `context`
Rename `contextOfArchive` → `archiveContext`

src/action-creators/undoArchive.js
```diff
@@ -22,6 +35,14 @@ export const undoArchive = ({ originalPath, currPath, offset }) => dispatch => {
     offset
   })
 
+  if (getThoughtsRanked(archiveContext).length === 0) {
+    dispatch({
+      type: 'existingThoughtDelete',
+      context: contextOfArchive,
+      thoughtRanked: head(rankThoughtsFirstMatch(archiveContext))
```
raineorshine: Rather than recreating the ranks from the context, which is expensive and does not work if there are duplicate thoughts, build the correct thought using `currPath`, `contextOf`, and `getThoughtsRanked`. 

`rankThoughtsFirstMatch` should be avoided if at all possible. It's mainly used for guessing the cursor from the url, although it has admittedly leaked out into other code.

src/action-creators/undoArchive.js
aqkhan: `splice` was used to build the context of archive, but it can also be retrieved from `originalPath`, which is being used now.

src/action-creators/undoArchive.js
aqkhan: Removed `rankThoughtsFirstMatch`.

src/action-creators/undoArchive.js
```diff
@@ -21,8 +18,8 @@ export const undoArchive = ({ originalPath, currPath, offset }) => dispatch => {
 
   const { getState } = store
```
raineorshine, line 19: Use the second argument to the thunk rather than importing the store. See other action-creators for this pattern.

src/action-creators/undoArchive.js
```diff
@@ -39,11 +36,12 @@ export const undoArchive = ({ originalPath, currPath, offset }) => dispatch => {
     offset
   })
 
-  if (getThoughtsRanked(getState(), archiveContext).length === 0) {
+  // Check if =archive is empty
+  if (getThoughtsRanked(getState(), context).length === 0) {
```
raineorshine, line 40: When the order doesn't matter, use `getThoughts`, which is `O(1)`, instead of `getThoughtsRanked`.

src/action-creators/undoArchive.js
```diff
@@ -21,8 +18,8 @@ export const undoArchive = ({ originalPath, currPath, offset }) => dispatch => {
 
   const { getState } = store
 
-  const archiveContext = contextOf(pathToContext(currPath))
-  const contextOfArchive = splice(archiveContext, archiveContext.length - 1, 1)
+  const context = contextOf(pathToContext(currPath))
+  const archiveContext = contextOf(pathToContext(originalPath))
```
raineorshine, line 22: You'll want to use `rootedContextOf` instead of `contextOf` here so that `archiveContext` gets set to `[ROOT]` instead of `[]`. `[]` is not a valid context.

You can test this out by archiving a thought in the root and then clicking "Undo".

Issue #627, opened 2020-05-05
Archived thoughts should be given a new rank


raineorshine:
When a thought is archived, it should use `getPrevRank` to be added to the top of the `=archive` context. The following is a simple example where the archived thought gets `rank: 0`, but the rank will vary and depend on previously archived thoughts.

## Steps to Reproduce

- a
  - b [rank: 0]
  - c [rank: 1]

Archive `c`.

## Current Behavior

- a
  - `=archive`
    - c [rank: 1]
  - b [rank: 0]

## Expected Behavior

- a
  - `=archive`
    - c [rank: 0]
  - b [rank: 0]

raineorshine:
Another case:

## Steps to Reproduce

- a
  - `=archive`
    - c [rank: 0]
  - b [rank: 0]

Archive `b`.

## Current Behavior

- a
  - `=archive`
    - c [rank: 0]
    - b [rank: 0]

## Expected Behavior

- a
  - `=archive`
    - b [rank: -1]
    - c [rank: 0]

PR #642, opened 2020-05-10
Archived thoughts should be given a new rank

aqkhan:
References to #627


aqkhan:
Closing to re-open in order to clean diff.

PR #654, opened 2020-05-13
Rank check for thoughts with context length less than one (existingThoughtDelete)

### LINKED ISSUE #521: Table view and pinned context toggle stops working
**Steps to reproduce:** 
1. Take a node and set it to table view from the toolbar.
2. Now select context view and then click table view in the toolbar again.

Now the table view shortcut stops working and it doesn't toggle back off. 
If we follow the same process with pinned context, it stops working too.

shresthabijay:
fixes #521 
@raineorshine  I have added rank check for thoughts with context less than one. Please review the changes. Thanks!

### DIFF
--- a/src/reducers/existingThoughtDelete.js
+++ b/src/reducers/existingThoughtDelete.js
@@ -44,12 +44,13 @@ export default (state, { context, thoughtRanked, showContexts }) => {
   }
 
   // the old thought less the context
-  const newOldThought = thought.contexts && thought.contexts.length > 1
+  const newOldThought = thought.contexts
     ? removeContext(thought, context, showContexts ? null : rank)
     : null
 
   // update state so that we do not have to wait for firebase
-  if (newOldThought) {
+  // delete thought directly if new context is empty
+  if (newOldThought && newOldThought.contexts.length > 1) {
     thoughtIndexNew[key] = newOldThought
   }
   else {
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine  So I just iterated over the contexts and check if there is a valid thought (i.e has exact rank and context). If there are none then I just prevented further logic and returned old state from the reducer.Please review the changes!

src/reducers/existingThoughtDelete.js
```diff
@@ -33,6 +34,11 @@ export default (state, { context, thoughtRanked, showContexts }) => {
   const thoughtIndexNew = { ...state.thoughtIndex }
   const oldRankedThoughts = rankThoughtsFirstMatch(state, thoughts)
 
+  const isValidThought = thought.contexts.find(parent => equalArrays(context, parent.context) && rank === parent.rank)
+
+  // if thought is not valid then just return current state
+  if (!isValidThought) return { ...state }
```
raineorshine, line 40: You can return nothing, as the main reducer handles it when [merging the updated fields into state](https://github.com/cybersemics/em/blob/dev/src/reducers/index.js#L98).

Can you add a `console.error` here? This *shouldn't* happen normally, so I'd like to be aware of it if it does.

raineorshine:
> Quick question do we even need to worry about context view when deleting a thought ? Because when context view is on `rankThoughtsFirstMatch` seems to return first matching thought ranked according to context view. But we need to find actual thought according to `contextIndex` with original rank. If so then is there specific reason for using `rankThoughtsFirstMatch` inside `exisitingThoughDelete` to find thought that needs to deleted ?

I think there may be a couple issues at play here.

`existingThoughtDelete` has a `showContexts` argument which is only used by [removeContext](https://github.com/cybersemics/em/blob/dev/src/reducers/existingThoughtDelete.js#L48). When `showContexts` is true, it ignores the rank. Presumably this is because the Context View uses autogenerated ranks, so the rank of the thought to be deleted will not match the rank within its context. On first impression, this does not seem like an ideal solution. Instead of matching it with the correct rank, it just ignores it. That might lead to problems deleting a context that contains a duplicate thought.

I'm not sure why `rankThoughtsFirstMatch` [was added to `existingThoughtDelete`](https://github.com/cybersemics/em/commit/915f01dcf85f536684ad773d575d5a2597142368#diff-aec2cdfb0f36273ed52f010d3ff441bfR33). It seems like `thoughtsRanked` would have worked, but it was changed intentionally.

`rankThoughtsFirstMatch` is [also used in `existingThoughtChange`](https://github.com/cybersemics/em/blob/dev/src/reducers/existingThoughtChange.js#L66), prompting the same question. 

A deeper investigation should be done not only to answer the above questions, but to determine what exactly requires the full `Path`. I would think that all you would need would be the `value`, `rank`, and `context` to delete a thought.

src/reducers/existingThoughtDelete.js
shresthabijay: Sure!

Issue #350, opened 2020-02-17
=unextendable should disallow `New Subthought` from parent


raineorshine:

pushkov-fedor:
I added =unextendable as child of Thought and it works in a strange way.

- parent
    thought1
    =unextendable

But it works only after copy and paste **=unextendable**. If i try to type it by hand and press enter, em cuts last characters of **=unextendable** and saves part of it (e.g. **=unextenda** or just **=**). After that submiting new Thoughts causes to the same cut behavior (e.g. 12345 => 123 saves). Looks like throttled function with timeout in Editable.js does not have time to fires again and save new value. That's why after pressing enter we have a cropped thoughts,
This happens only after an attempt to save meta thought (started with `=`), which gets cutted. Do you have the same experience?

raineorshine:
Yes, it appears to be a regression on `dev`. I will investigate and file a separate issue.

raineorshine:
Editing issue fixed in 2e2b0dfa. You should be able to test `=unextendable` now.

pushkov-fedor:
It works! newThoughts.js in shortcuts folder has check if parent is unextendable. I think we need add kind of the same check in newSubthoughts.js for thoughts itself. I think it takes 1-1.5 hours

PR #655, opened 2020-05-14
Disallow 'New Subthoughts' from parent with =unextendable

pushkov-fedor:
Issue #350


raineorshine:
Beautiful! Now we get the check on `newUncle` for free.

Issue #519, opened 2020-04-05
Convert to typescript


raineorshine:
- One commit per file
- Start with small, frequently used util functions

Issue #273, opened 2020-01-05
Staged local loading


raineorshine:
Large amounts of data (5,000+ thoughts) causes the app to load very slowly from IndexedDB. Instead of loading the data all at once (`loadLocalState`), load ~the data and render in stages~ only the thoughts that are within a few steps from the cursor. The `thoughtIndex` and `contextIndex` will then serve as a cache of thoughts rather than a complete index.

Thoughts should be loaded in two stages. The first stage is the most critical as it is the "time-till-first-load" that most affects the user's perception of the loading time.

1. Load `expanded` thoughts and their children.
2. Load children, grandchildren, and great grandchildren.

After the initial load, thoughts should be swapped out as follows:

- [ ] When the `cursor` moves, recalculate `expanded` and load thoughts as described above for the new `cursor`.
- [ ] Show “Loading...” in place of children that are still loading (also on startup). 
- [ ] Delete least recently accessed thoughts when they exceed a configurable cache size.

raineorshine:
For reference, this is some code I was using to load just the root thoughts:

```js
  if (!schemaVersion || schemaVersion === SCHEMA_LATEST) {
    localForage.getItem('contextIndex-' + hashContext([ROOT_TOKEN])).then(rootThoughts => {
      store.dispatch({
        type: 'thoughtIndex',
        contextIndexUpdates: {
          [hashContext([ROOT_TOKEN])]: rootThoughts
        },
        thoughtIndexUpdates: (rootThoughts || []).reduce((accum, current) => {
          return {
            ...accum,
            [hashThought(current.value)]: {
              contexts: [[ROOT_TOKEN]],
              value: current.value,
            }
          }
        }, {}),
        forceRender: true
      })
    })
  }
```

PR #656, opened 2020-05-14
Conversion to Typescript

anmolarora1:
Addresses #519 

As a part of this PR, we'll introduce TS to util functions, add frequently used types from business-context to `src/types.ts` and other types shared by util functions to `src/utilTypes.ts`


src/utilTypes.ts
```diff
@@ -3,3 +3,6 @@ export type ComparatorValue = 1 | -1 | 0
 
 /* A standard comparator function used within sort */
 export type ComparatorFunction<T> = (a: NonNullable<T>, b: NonNullable<T>) => ComparatorValue
+
+/* Set of file types supported for exporting thoughts */
+export type FileTypes = 'text/plain' | 'text/html'
```
raineorshine, line 8: `MimeType` would be a more appropriate name.

A type should generally be in the singular. The plural would likely be its own array or collection type.

src/util/equalArrays.ts
```diff
@@ -1,7 +1,5 @@
-//@ts-nocheck
-
 /** Equality for lists of lists. */
-export const equalArrays = (a, b) =>
+export const equalArrays = (a: unknown[], b: unknown[]): boolean =>
```
raineorshine, line 2: Does `a: []` work, or is `unknown` required?

src/util/equalPath.ts
```diff
@@ -1,7 +1,6 @@
-//@ts-nocheck
-
 import { equalThoughtRanked } from './equalThoughtRanked'
+import { Thought, Child } from '../types'
 
 /** Compares two thoughtsRanked arrays using { value, rank } as identity and ignoring other properties. */
-export const equalPath = (a, b) =>
+export const equalPath = (a: (Thought | Child)[], b: (Thought | Child)[]): boolean =>
```
raineorshine, line 5: `a` and `b` should be type `Path`.

src/util/equalThoughtRanked.ts
```diff
@@ -1,5 +1,5 @@
-//@ts-nocheck
+import { Thought, Child } from "../types";
 
 /** Compares two thought objects using { value, rank } as identity and ignoring other properties. */
-export const equalThoughtRanked = (a, b) =>
+export const equalThoughtRanked = (a: (Thought | Child), b: (Thought | Child)): boolean =>
```
raineorshine, line 4: `Child` only

src/util/equalThoughtSorted.ts
```diff
@@ -1,5 +1,5 @@
-//@ts-nocheck
+import { Thought, Child } from "../types";
 
 /** Compares two thought objects using { value } as identity and ignoring other properties. */
-export const equalThoughtSorted = (a, b) =>
+export const equalThoughtSorted = (a: (Thought | Child), b: (Thought | Child)): boolean =>
```
raineorshine, line 4: `Child` only

src/util/equalArrays.ts
anmolarora1: Afaik, that doesn't work. We need to specify the type of items.

src/utilTypes.ts
anmolarora1: Agreed

src/util/formatNumber.ts
```diff
@@ -1,11 +1,9 @@
-//@ts-nocheck
-
 /** Adds commas to a number */
 // TODO: Localize
-export const formatNumber = n => {
+export const formatNumber = (n: number) => {
   let s = '' // eslint-disable-line fp/no-let
   const digits = n.toString().split('')
-  digits.forEach((value, i) => {
+  digits.forEach((_, i) => {
```
anmolarora1, line 6: @raineorshine Have you considered using `toLocaleString` instead of this?

src/util/equalArrays.ts
raineorshine: Did you try it?

src/util/formatNumber.ts
raineorshine: Sure, I'm not super familiar with `toLocaleString`.

anmolarora1: It does the same job that this function is doing. You can check it out [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)

src/util/equalArrays.ts
anmolarora1: Yes. Please check [this](https://www.typescriptlang.org/play/#code/MYewdgzgLgBA5gUygGQWOUAWMC8MAUAhgFwwDaAugJS4B8MhAdADZoaYDcAUF6JCKxYg4+RCjZZ8ZAIwAaAEzUqQA)

raineorshine: Got it, thanks!

src/util/formatNumber.ts
raineorshine: Cool! I created #682 to replace it.

package-lock.json
```diff
@@ -8623,25 +8623,25 @@
           "dependencies": {
             "abbrev": {
               "version": "1.1.1",
-              "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+              "resolved": false,
```
raineorshine, line 8626: Revert `package-lock.json`

src/util/contextOf.ts
```diff
@@ -1,4 +1,5 @@
-import { Context } from "../types";
+import { Thought, Child } from "../types";
 
+// type C = 
 /** Gets the context of a context */
-export const contextOf = (context: Context): Context => context.slice(0, -1)
+export const contextOf = <T extends Array<Thought | Child>>(context: T) => context.slice(0, -1)
```
raineorshine: I think this might need to be `Context | Path`.

src/util/deleteThought.ts
```diff
@@ -37,17 +35,18 @@ import {
...
   splitChain,
 } from '../selectors'
+import { Child, Thought } from '../types'
 
 /** Deletes a thought. */
 export const deleteThought = () => {
 
   const state = store.getState()
-  const path = state.cursor
+  const path:(Thought[] | Child[]) = state.cursor
```
raineorshine, line 44: Why use `Child[]` when we have a `Path` type? Also, why `Thought[]` at all? A `Thought` is an entry in the `thoughtIndex` which is not related to a `Path`.

src/util/deleteThought.ts
```diff
@@ -37,17 +35,18 @@ import {
...
 
   // same as in newThought
   const contextChain = splitChain(state, path)
   const showContexts = isContextViewActive(state, contextOf(path))
-  const thoughtsRanked = contextChain.length > 1
+  const thoughtsRanked:Thought[] = contextChain.length > 1
```
raineorshine, line 49: Doesn't Typescript infer types of variables?

src/util/deleteThought.ts
```diff
@@ -73,12 +72,12 @@ export const deleteThought = () => {
...
 
   // prev must be calculated before dispatching existingThoughtDelete
-  const prev = showContexts
+  const prev: Child | undefined = showContexts
     ? prevContext()
     : prevSibling(state, value, context, rank)
 
   /** Returns true when thought is not hidden due to being a function or having a =hidden attribute. */
-  const isVisible = thoughtRanked => state.showHiddenThoughts || (
+  const isVisible = (thoughtRanked: Thought) => state.showHiddenThoughts || (
```
raineorshine, line 80: I think this should be `Child`.

src/util/deleteThought.ts
```diff
@@ -103,7 +102,7 @@ export const deleteThought = () => {
   }
 
   /** Sets the cursor or moves it back if it doesn't exist. */
-  const setCursorOrBack = (thoughtsRanked, { offset } = {}) => {
+  const setCursorOrBack = (thoughtsRanked: Array<Thought | Child | string> = [], { offset } = {offset: 0}) => {
```
raineorshine, line 105: I think there are times when `offset` is undefined. If you set it to `0` it will change the behavior.

raineorshine, line 105: Wouldn't this type just be a nullable `Path`?

src/util/deleteThought.ts
```diff
@@ -112,14 +111,16 @@ export const deleteThought = () => {
...
     // Case III: delete last thought in context; set cursor on context
-    thoughts.length > 1 ? [rootedContextOf(path), { offset: head(context).length }]
-    // Case IV: delete very last thought; remove cursor
-    : [null]
-  )
+    : setCursorOrBack(contextOf(path).concat(next()), { offset: 0 })
+  }
+  // Case IV: delete very last thought; remove cursor
+  return thoughts.length>1 ?setCursorOrBack(rootedContextOf(path), { offset: head(context).length }) : setCursorOrBack([])
 }
```
raineorshine, line 126: Do not use multiple return statements, as it breaks the invariant that the shape of the code should reflect the control flow. A single return statement with multiple cases captures the logic with less redundancy.

src/selectors/canShowModal.ts
```diff
@@ -1,6 +1,6 @@
 import { store } from '../store'
 
-const canShowModal = (state: any, id: number) => {
+const canShowModal = (state: any, id: number | string) => {
```
raineorshine, line 3: Would this work with just `string`?

src/util/ancestors.ts
```diff
@@ -0,0 +1,6 @@
+import { equalThoughtRanked } from './equalThoughtRanked'
+import { Path,Child } from '../types'
+
+
```
raineorshine, line 4: Extra blank line. Can you make sure the typescript linter is picking this up?

src/util/ancestors.ts
```diff
@@ -0,0 +1,6 @@
+import { equalThoughtRanked } from './equalThoughtRanked'
+import { Path,Child } from '../types'
```
raineorshine, line 2: Space after comma

src/util/compareThought.ts
```diff
@@ -104,4 +106,4 @@ const compareReasonable = makeOrderedComparator([
 ])
 
 /** Compare the value of two thoughts. */
-export const compareThought = (a, b) => compareReasonable(removeEmojis(a.value), removeEmojis(b.value))
+export const compareThought = (a: Thought, b: Thought) => compareReasonable(removeEmojis(a.value), removeEmojis(b.value))
```
raineorshine, line 109: `Child`

src/util/decodeCharacterEntities.ts
```diff
@@ -1,5 +1,5 @@
 /** Decode &, >, and < character entities. */
-export const decodeCharacterEntities = s => s
+export const decodeCharacterEntities = (s:string): string => s
```
raineorshine, line 2: Space after `:`. Can the linter catch this? (Can be in another PR)

src/util/contextOf.ts
```diff
@@ -0,0 +1,10 @@
+import { Context, Path } from "../types";
+
+type ReturnType = {
+  (param:Context): Context;
+  (param:Path): Path;
+  (param: Path | Context) : Path | Context
+}
```
raineorshine, line 7: Can you explain this? I don't quite understand.

src/util/ellipsize.ts
```diff
@@ -1,6 +1,6 @@
 import { THOUGHT_ELLIPSIZED_CHARS } from '../constants'
 
 /** Returns a string truncated with an ellipsis at a given limit n. */
-export const ellipsize = (s, n = THOUGHT_ELLIPSIZED_CHARS) =>
+export const ellipsize = (s: string, n:number = THOUGHT_ELLIPSIZED_CHARS): string =>
```
raineorshine, line 4: Space after `:`

src/util/ellipsizeUrl.ts
```diff
@@ -2,7 +2,7 @@ import { URL_MAX_CHARS } from '../constants'
 import { isURL } from '../util/isURL'
 
 /** Ellipsize a value if it is a url. */
-export const ellipsizeUrl = value =>
+export const ellipsizeUrl = (value:string): string =>
```
raineorshine, line 5: Space after `:`

src/util/escapeRegExp.ts
```diff
@@ -1,4 +1,4 @@
 /** Make a string safe to instantiate a RegExp.
  * NOTE: The escape-string-regexp npm package crashes the react production build.
  */
-export const escapeRegExp = s => s.replace(/[-[\]{}()*+?.\\^$|#\s]/g, '\\$&')
+export const escapeRegExp = (s:string): string => s.replace(/[-[\]{}()*+?.\\^$|#\s]/g, '\\$&')
```
raineorshine, line 4: Space after `:`

src/util/escapeSelector.ts
```diff
@@ -3,4 +3,4 @@ import { escapeRegExp } from './escapeRegExp'
 export const regExpEscapeSelector = new RegExp('[' + escapeRegExp(' !"#$%&\'()*+,./:;<=>?@[]^`{|}~') + ']', 'g')
 
 /** Replace characters that are invalid in document.querySelector with their respective character codes. Prepend _ to escape leading digits. */
-export const escapeSelector = s => '_' + s.replace(regExpEscapeSelector, s => '_' + s.charCodeAt())
+export const escapeSelector = (s:string): string => '_' + s.replace(regExpEscapeSelector, s => `_${s.charCodeAt(0)}`)
```
raineorshine, line 6: Space after `:`

src/util/formatNumber.ts
```diff
@@ -1,3 +1,3 @@
 /** Adds commas to a number. */
 // TODO: Localize
-export const formatNumber = (n: number, locale='en-US') => n.toLocaleString(locale)
+export const formatNumber = (n: number, locale='en-US') => n.toLocaleString(locale)
```
raineorshine, line 3: Restore newline at end of file.

Do you have `EditorConfig` configured in your editor? Usually that will automatically enforce this.

src/util/makeCompareByProp.ts
```diff
@@ -1,16 +1,18 @@
...
 /**
  * Creates a function that takes two values and compares the given key.
  * Does case insensitive comparison with strings.
  */
-export const makeCompareByProp = key => (a, b) =>
+export const makeCompareByProp = (key:any):ComparatorFunction<Object> => (a: Object, b: Object) =>
   isGreater(a, b, key) ? 1
   : isSmaller(a, b, key) ? -1
-  : 0
+  : 0
```
raineorshine, line 18: Newline

src/util/makeCompareByProp.ts
```diff
@@ -1,16 +1,18 @@
...
 
+import { ComparatorFunction } from '../utilTypes'
+
 /** Returns true if the first object's key is greater than the second object's key. */
-export const isGreater = (a, b, key) => lower(a[key]) > lower(b[key])
+export const isGreater = (a:Object, b:Object, key: keyof typeof a & keyof typeof b )  => lower(a[key]) > lower(b[key])
 
 /** Returns true if the first object's key is smaller than the second object's key. */
-export const isSmaller = (a, b, key) => lower(a[key]) < lower(b[key])
+export const isSmaller = (a:Object, b:Object, key: keyof typeof a & keyof typeof b )  => lower(a[key]) < lower(b[key])
```
raineorshine, line 9: Space after `:`

src/util/makeCompareByProp.ts
```diff
@@ -1,16 +1,18 @@
 import { lower } from './lower'
 
+import { ComparatorFunction } from '../utilTypes'
+
 /** Returns true if the first object's key is greater than the second object's key. */
-export const isGreater = (a, b, key) => lower(a[key]) > lower(b[key])
+export const isGreater = (a:Object, b:Object, key: keyof typeof a & keyof typeof b )  => lower(a[key]) > lower(b[key])
```
raineorshine, line 6: Can you explain the expression `keyof typeof a & keyof typeof b`?

src/utilTypes.ts
```diff
@@ -0,0 +1,13 @@
...
+/* A standard comparator function used within sort */
+export type ComparatorFunction<T> = (a: NonNullable<T>, b: NonNullable<T>) => ComparatorValue
+
+/* Set of file types supported for exporting thoughts */
+export type MimeType = 'text/plain' | 'text/html'
+
+/* Generic type to allow null */
+export type Nullable<T> = T | null;
+
+export type GenericObject<T = any> = {[key: string]: T}
```
raineorshine, line 13: What's the advantage of using this over any? Since everything is a generic object in Javascript.

src/util/parseJsonSafe.ts
```diff
@@ -1,8 +1,10 @@
+import { Nullable } from "../utilTypes"
+
 /** Safely JSON parse an unknown value, and default to a given fallback value. */
-export const parseJsonSafe = (value, defaultValue) => {
+export const parseJsonSafe = (value: Nullable<string>, defaultValue: number) => {
```
raineorshine, line 4: `defaultValue` should be `any`.

src/util/pathToContext.ts
```diff
@@ -0,0 +1,8 @@
+import { Context, Path } from "../types"
+
+/** Converts paths [{ value, rank }, ...] to contexts [key, ...]. */
+// if already converted, return a shallow copy
+// if falsey, return as-is
+
+const isPath = (param: any): param is Path =>  !!(param as Path).length && (param as Path)[0].hasOwnProperty('value')
+export const pathToContext = (thoughts: Path | Context): Context  => isPath(thoughts) ? thoughts.map(child => child.value) : thoughts.slice()
```
raineorshine, line 8: Newline

src/util/rankThoughtsSequential.ts
```diff
@@ -1,3 +1,5 @@
+import { Context, Child } from "../types";
+
 /** Ranks the thoughts from 0 to n in the given array order. */
-export const rankThoughtsSequential = thoughts =>
+export const rankThoughtsSequential = (thoughts: Context) =>
```
raineorshine, line 4: Technically this should be `string[]` since this operates on thoughts within a context rather than along a path.

src/util/setSelection.ts
```diff
@@ -5,20 +5,23 @@
...
-  const sel = window.getSelection()
+  const sel = window.getSelection() || new Selection()
   // automatically constrain offset to text length
   // this may still throw an error if the text node does no exist any longer
-  try {
-    range.setStart(textNode, offset ? Math.min(offset, textNode.textContent.length) : end ? textNode.textContent.length : 0)
-  }
-  catch (e) {
-    console.warn(e)
+  if(textNode!==null) {
```
raineorshine, line 18: I believe the `textNode` is guaranteed to be valid since it is appended at the beginning of the function.

src/util/setSelection.ts
```diff
@@ -5,20 +5,23 @@
...
-export const setSelection = (el, { offset, end } = {}) => {
+type SelectionOptionsType = {offset: number, end: number}
+export const setSelection = (el: HTMLElement, { offset, end }: SelectionOptionsType = {offset: 0, end: 0}) => {
   if (el.childNodes.length === 0) {
     el.appendChild(document.createTextNode(''))
   }
   const textNode = el.firstChild
   const range = document.createRange()
-  const sel = window.getSelection()
+  const sel = window.getSelection() || new Selection()
```
raineorshine, line 15: Can you explain this?

raineorshine:
> - isRoot & isEm are being used interchangeably by contexts/Path/Thought object’s context field. What’s the correct expectation?

`isRoot` tests if a thought is a descendant of the root thought. `isEm` tests if a thought is a descendant of the **em** thought. Currently the root thought contains user thoughts and **em** contains Settings.

Does that answer your question, or are you seeking something else?

> - mergeUpdates - why are we using for in and not Object.keys?

For [performance reasons](https://jsperf.com/object-keys-vs-for-in-loop-test), but this will become less important when we switch to a thought cache (#273) rather than loading all thoughts into memory.

> - notNull function doesn’t seem to be filtering undefined whereas its JSDoc description states it filters both null and undefined.

`a != null` is shorthand for `a !== null && a !== undefined`

> - importHTML - line 92 . Is null a valid value in thoughtindexupdates?

The line number has changed, but I think I can still answer your question. Yes, `thoughtIndexUpdates` may have `null` values. This is what tells Firebase to delete the key. They are removed locally into `thoughtIndex` via `mergeUpdates`. Same for `contextIndexUpdates`.

> - What's the type of items in sumSubthoughtsLength?

That's a strange one. `Child | ThoughtContext`.

src/selectors/canShowModal.ts
anmolarora1: Yes, I'll update it

src/util/ancestors.ts
anmolarora1: We can add a config for picking up such issues. If it's a long process, we can tackle it in a different issue.

src/util/contextOf.ts
anmolarora1: An explicit return type is added here to allow the compiler to infer the type of results in the calling functions. If we don't do it, the return type would be `Path | Context`.  
An example from `importHtml.ts` - 
[importCursor](https://github.com/cybersemics/em/blob/adab4290f2a0c59d6fc53f52a3dd8040b3f21dba/src/util/importHTML.js#L98) can be `Path|Context`. But [here](https://github.com/cybersemics/em/blob/adab4290f2a0c59d6fc53f52a3dd8040b3f21dba/src/util/importHTML.js#L182), we're assuming its type to be `Path`, which results in a compile-time error.

src/utilTypes.ts
anmolarora1: That's true, everything is a generic object in JS and we can allow them to be like that. But this helps to narrow down the allowed value types for an object. 
Example - We can enforce the thoughtIndex to have only `thoughtObject` values - [`thoughtIndex: GenericObject<Thought>`](https://github.com/cybersemics/em/blob/5226dcfec237353436ae84973c23cacfa2b76764/src/util/initialState.ts#L33).

src/util/setSelection.ts
anmolarora1: That's right, but the compiler doesn't quite understand that. 
Maybe we can refactor our code to make it more explicit. Something like this helps to get rid of the compiler error - 
```
const textNode = el.firstChild || el.appendChild(document.createTextNode(''))
```

src/util/setSelection.ts
anmolarora1: When used in Firefox, `window.getSelection()` [returns null](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection) for some values, and hence, TS compilation errors on [these lines](https://github.com/cybersemics/em/blob/094828d323f63e1d19d865947e619322c9dc2519/src/util/setSelection.ts#L27)

src/util/makeCompareByProp.ts
anmolarora1: The intention behind adding this was to ensure that the key passed to this function belongs to both the objects, so it doesn't end up doing `lower(undefined) >lower(undefined)`
It didn't work as expected but it's fixed in [this](https://github.com/cybersemics/em/pull/656/commits/fb102a1254eee7ce8d99aa6157aa3606b16ce29b) commit

PR #662, opened 2020-05-15
Fix for toolbar scaling issue

### LINKED ISSUE #608: Toolbar does not expand fully on desktop
Screen Shot 2020-04-29 at 7 32 33 AM

## Steps to Reproduce

Open **em** in a window about 700px wide or more.

## Expected Behavior

Toolbar should expand to its full width and not have any scroll bars.

## Current Behavior

Toolbar container width is smaller than content, causing scrollbars to appear.

## Notes

Important to test that the fix does not introduce regressions at any device size.

anmolarora1:
Fixes #608 

Issue - Hard-coding the `transformOrigin` property for the `Scale` component didn't allow it to use the `transforms` appropriately. Its value should ideally be based on where the scaled elements are positioned in the DOM.

Fix- Allow the value of `transformOrigin` to be passed as props based on the positioning of the component, and set its value to `top right` to represent the toolbar's position.


anmolarora1:
> I don't believe we should use `isMobile` here. What matters is the width of the window, not whether it is mobile or not. We need to maintain the following invariants, regardless of the platform:
> 
We need to use `isMobile` here in order to change the `transform-origin` property of the toolbar. The default value for the same is `50% 50%` or `center`, which cannot be used in our case because -
1) In the desktop view, the toolbar needs to be right-aligned and scaled towards left
2) In the mobile view, the toolbar needs to be left-aligned and scaled towards left
3) On hovering, the `toolbar-overlay` will cause the toolbar to bump up because it's statically positioned.

An alternative approach involves adding fixed margins on both sides in order to address the aforementioned issues. And, making some changes to the `toolbar-overlay` (like absolute positioning).

> 1. The Toolbar expands fully when there is room
> 2. The Toolbar does not overlap the Recently Edited icon
> 
> The second point is not quite working yet:
> 
> Screen Shot 2020-05-15 at 12 19 25 PM

raineorshine:
Yes, absolutely! I’m in full agreement about the need to use different alignment and transform origin. 

Regarding my point about `isMobile`. I think you may be  conflating the platform (mobile/desktop) with screen width. Those are correlated but semantically different things. I still think the condition should be based on screen width rather than platform. 

i.e. 

``` 
if (toolbarWidth < containerWidth) alignLeft
else alignRight
``` 

This seems more relevant and flexible than `isMobile`, which is parsed from the UserAgent and has nothing to do with screen width. Using `isMobile`, for example, may return `true` for tablets or large phones in landscape mode, and incorrectly left align the toolbar when there is plenty of space to right align it.

raineorshine:
I'm not familiar with `useContext`, so I'm not in a position to assess the viability of that approach.

Another approach would be to always align the toolbar right, set the origin to "top right", and then make sure the scrollbar appears correctly when the screen is too small.

anmolarora1:
> Yes, absolutely! I’m in full agreement about the need to use different alignment and transform origin.
> 
> Regarding my point about `isMobile`, I think you may be conflating the platform (mobile/desktop) with screen width. Those are correlated but semantically different things. I still think the condition should be based on screen width rather than platform.
> 
> i.e.
> 
> ```
> if (toolbarWidth < containerWidth) alignLeft
> else alignRight
> ```
> 
> This seems more relevant and flexible than `isMobile`, which is parsed from the UserAgent and has nothing to do with screen width. Using `isMobile`, for example, may return `true` for tablets or large phones in landscape mode, and incorrectly left align the toolbar when there is plenty of space to right align it.

That approach seems sensible. Since we do not have an existing solution in place to comply with responsive behavior and make the component re-render when the screen size changes, here are some options that I'd like to suggest- 
1. Use context API to store the screen dimensions with an attached event listener to handle screen-size changes. Have the app wrapped up inside the provider and consume the values wherever we need them (In this case, the `Toolbar` component)
2. Store those variables locally within the `Toolbar` component and have an event listener created in the `useEffect` hook to handle screen size changes.

@shresthabijay Would you like to share your thoughts on this?

shresthabijay:
> > Yes, absolutely! I’m in full agreement about the need to use different alignment and transform origin.
> > Regarding my point about `isMobile`, I think you may be conflating the platform (mobile/desktop) with screen width. Those are correlated but semantically different things. I still think the condition should be based on screen width rather than platform.
> > i.e.
> > ```
> > if (toolbarWidth < containerWidth) alignLeft
> > else alignRight
> > ```
> > 
> > 
> > This seems more relevant and flexible than `isMobile`, which is parsed from the UserAgent and has nothing to do with screen width. Using `isMobile`, for example, may return `true` for tablets or large phones in landscape mode, and incorrectly left align the toolbar when there is plenty of space to right align it.
> 
> That approach seems sensible. Since we do not have an existing solution in place to comply with responsive behavior and make the component re-render when the screen size changes, here are some options that I'd like to suggest-
> 
> 1. Use context API to store the screen dimensions with an attached event listener to handle screen-size changes. Have the app wrapped up inside the provider and consume the values wherever we need them (In this case, the `Toolbar` component)
> 2. Store those variables locally within the `Toolbar` component and have an event listener created in the `useEffect` hook to handle screen size changes.
> 
> @shresthabijay Would you like to share your thoughts on this?

@anmolarora1 Yes I agree. Context API will make it easier to channel screen size data to multiple components with ease. But if it's gonna be used only for `Toolbar` component then I suggest using custom hook like this https://usehooks.com/useWindowSize/  inside the component.

anmolarora1:
> > > Yes, absolutely! I’m in full agreement about the need to use different alignment and transform origin.
> > > Regarding my point about `isMobile`, I think you may be conflating the platform (mobile/desktop) with screen width. Those are correlated but semantically different things. I still think the condition should be based on screen width rather than platform.
> > > i.e.
> > > ```
> > > if (toolbarWidth < containerWidth) alignLeft
> > > else alignRight
> > > ```
> > > 
> > > 
> > > This seems more relevant and flexible than `isMobile`, which is parsed from the UserAgent and has nothing to do with screen width. Using `isMobile`, for example, may return `true` for tablets or large phones in landscape mode, and incorrectly left align the toolbar when there is plenty of space to right align it.
> > 
> > 
> > That approach seems sensible. Since we do not have an existing solution in place to comply with responsive behavior and make the component re-render when the screen size changes, here are some options that I'd like to suggest-
> > 
> > 1. Use context API to store the screen dimensions with an attached event listener to handle screen-size changes. Have the app wrapped up inside the provider and consume the values wherever we need them (In this case, the `Toolbar` component)
> > 2. Store those variables locally within the `Toolbar` component and have an event listener created in the `useEffect` hook to handle screen size changes.
> > 
> > @shresthabijay Would you like to share your thoughts on this?
> 
> @anmolarora1 If we want to provide screen size data to multiple components then we should go with context API. But if it's gonna be used only for `Toolbar` component then I suggest using custom hook like this https://usehooks.com/useWindowSize/ inside the component.

@shresthabijay I agree with that. We can go for this hook for now. Thanks for suggesting :)

raineorshine:
That sounds like a could way to encapsulate the `resize` event. Note that we should adapt the hook to use the width of the `.toolbar-container` which determines if the `.toolbar` gets a scroll bar, not the window size itself. The window size is analogous but overgeneral for the problem at hand.

shresthabijay:
> That sounds like a could way to encapsulate the `resize` event. Note that we should adapt the hook to use the width of the `.toolbar-container` which determines if the `.toolbar` gets a scroll bar, not the window size itself. The window size is analogous but overgeneral for the problem at hand.

@raineorshine @anmolarora1  If we wanna respond to change in the width of `.toolbar-container` then we can use a custom hook like this. It uses `resize-observer-polyfill` and I am using this same hook for flat-render branch.

```javascript
import { useLayoutEffect, useRef, useState } from 'react'
import ResizeObserver from 'resize-observer-polyfill'

/** hook to observe elements size and position */
const useMeasure = () => {
  const ref = useRef()
  const [bounds, set] = useState({ left: 0, top: 0, width: 0, height: 0 })
  const [ro] = useState(() => new ResizeObserver(([entry]) => set(entry.contentRect)))

  useLayoutEffect(() => {
    if (ref.current) ro.observe(ref.current)
    return () => ro.disconnect()
  }, [])

  return [{ ref }, bounds]
}

export default useMeasure
``` 

Now we can simply bind the div and respond to width or height changes. 

```javascript
const [bind, { width }] = useMeasure()

return <div {...bind} > </div> // targeted div
```

anmolarora1:
> That sounds like a could way to encapsulate the `resize` event. Note that we should adapt the hook to use the width of the `.toolbar-container` which determines if the `.toolbar` gets a scroll bar, not the window size itself. The window size is analogous but overgeneral for the problem at hand.

@raineorshine Correct me if I'm wrong here, but to have the below comparison, we'll need the `width of the window (containerWidth)` as well as the `.toolbar-container (toolbarWidth)`
```
if (toolbarWidth < containerWidth) alignLeft
else alignRight
```

The hook which @shresthabijay suggested works well for getting the size of any element. But we'll need to have a `resize` event-listener or the same hook added at the top-level of our app to get the `containerWidth`. Which approach do you think, would be more appropriate? 
Or would you like to suggest a third alternative?

raineorshine:
> @raineorshine Correct me if I'm wrong here, but to have the below comparison, we'll need the `width of the window (containerWidth)` as well as the `.toolbar-container (toolbarWidth)`

Based on my tests, the `clientWidth` of the toolbar changes, while the `scrollWidth` is fixed:

// overflow
toolbar.clientWidth 289
toolbar.scrollWidth 576

// no overflow
toolbar.clientWidth 512
toolbar.scrollWidth 576

In this case, it looks like `toolbar.clientWidth + 64 < toolbar.scrollWidth` should evaluate to `true` if there is any overflow. That constant should be checked out though to make sure it doesn't change based on zoom or font size.

If this works, then you'll just need to calculate this in response to the `ResizeObserver`.

anmolarora1:
> > @raineorshine Correct me if I'm wrong here, but to have the below comparison, we'll need the `width of the window (containerWidth)` as well as the `.toolbar-container (toolbarWidth)`
> 
> Based on my tests, the `clientWidth` of the toolbar changes, while the `scrollWidth` is fixed:
> 
> ```js
> // overflow
> toolbar.clientWidth 289
> toolbar.scrollWidth 576
> 
> // no overflow
> toolbar.clientWidth 512
> toolbar.scrollWidth 576
> ```
> 
> In this case, it looks like `toolbar.clientWidth + 64 < toolbar.scrollWidth` should evaluate to `true` if there is any overflow. That constant should be checked out though to make sure it doesn't change based on zoom or font size.
> 
> If this works, then you'll just need to calculate this in response to the `ResizeObserver`.

I think there are some caveats with this approach -

1. The difference between the `scrollWidth` and the `clientWidth` or the `delta` differs based on the `screen size` and the `scale` values.
These are the respective measures, that I got for clientWidth, scrollWidth, device-type, scale, difference respectively.
```
720  576  Desktop 1.25 - 144
413  330  Mobile 1.25 - 83
866  576  Desktop 1.5 - 110
495  330  Mobile 1.5 - 165
```

2. We'll have to use the DOM element ref to calculate the `clientHeight` and `scrollWidth` because these values aren't provided by the `ResizeObserver`.

src/hooks/useResizeObserver.ts
```diff
@@ -0,0 +1,18 @@
...
+    if (ref.current === null) {
+      return;
+    }
+
+    const ro = new ResizeObserver(onResize);
+    ro.observe(ref.current!);
+
+    return () => ro.disconnect()
+  }, []);
+}
```
shresthabijay, line 17: Looks good! But we can make it tighter with these changes.

```javascript
const useResizeObserver = (ref: RefObject<HTMLElement>, onResize: (target: Element) => void) => {
  const [ro] = useState(() => new ResizeObserver(([entry]) => onResize(entry.target)))
  useLayoutEffect(() => {
    if (ref.current) ro.observe(ref.current)
    return () => ro.disconnect()
  }, []);
}
```
So we don't have do access ref again in the callback.

```javascript
useResizeObserver(toolbarRef, target => {
    const { scrollWidth, clientWidth } = target
    updateToolbarOrigin(scrollWidth - clientWidth > TOOLBAR_WRAPPER_DELTA ? toolbarOrigins.topRight : toolbarOrigins.topLeft)
})
```

src/components/Toolbar.js
```diff
@@ -85,9 +91,15 @@ const Toolbar = ({ cursorOnTableView, cursorOnAlphabeticalSort, cursorPinOpen, c
...
   const fg = dark ? 'white' : 'black'
   // const bg = dark ? 'black' : 'white'
+  const toolbarRef = useRef()
+  useResizeObserver(toolbarRef, () => {
+    const { scrollWidth, clientWidth } = toolbarRef.current
+    updateToolbarOrigin(scrollWidth - clientWidth > TOOLBAR_WRAPPER_DELTA ? toolbarOrigins.topRight : toolbarOrigins.topLeft)
+  })
 
   useEffect(() => {
     if (toolbarOverlay) {
```
shresthabijay, line 105: This side effect is not really needed. We can simply do this. This will eliminate two state variables.

```javascript
  const shortcut = toolbarOverlay ? shortcutById(toolbarOverlay) : null
  const overlayName = shortcut ? shortcut.name : ''
  const overlayDescription = shortcut ? shortcut.description : ''
```

shresthabijay:
@anmolarora1 There is a regression introduced by the recent changes.

chrome-capture (1)

Just hover over a toolbar icon to trigger overlay. As you can see in the gif that it is weirdly shifting to the right when overlay is visible. Somehow it is triggering toolbar origin update.

On troubleshooting I found the cause. When overlay is rendered inside toolbar, it doesn't change the width but changes it's height. Since resize observer observes `contentRect` of the target div, so it will trigger resize on height change too. Somehow `scrollWidth` is also changing.

One quick reliable solution will be to make sure our hook observer only triggers resize event on width change. I have included changes I suggested above in this code.

```javascript
const useResizeObserver = (ref: RefObject<HTMLElement>, onResize: (target: Element) => void) => {

  const boundsRef = useRef({ width: 0, height: 0 })
  const [ro] = useState(() => new ResizeObserver(([entry]) => {
    console.log(entry.contentRect)
    // resize callback gets only in width change
    if (boundsRef.current.width !== entry.contentRect.width) onResize(entry.target)
    boundsRef.current = entry.contentRect
  }))

  useLayoutEffect(() => {
    if (ref.current) ro.observe(ref.current)
    return () => ro.disconnect()
  }, []);
}
```

As you can see using a ref we can easily achieve this. If we do this we should rename the hook to `useWidthObserver`. But that will dumb down the hook to just observing width. Also there is another possible solution where we can prevent change of height of toolbar due to overlay div. What do you think we should do ? I think first one is easier and reliable.

anmolarora1:
> @anmolarora1 There is a regression introduced by the recent changes.
> 
> chrome-capture (1)
> 
> Just hover over a toolbar icon to trigger overlay. As you can see in the gif that it is weirdly shifting to the right when overlay is visible. Somehow it is triggering toolbar origin update.
> 
> On troubleshooting I found the cause. When overlay is rendered inside toolbar, it doesn't change the width but changes it's height. Since resize observer observes `contentRect` of the target div, so it will trigger resize on height change too. Somehow `scrollWidth` is also changing.
> 
> One quick reliable solution will be to make sure our hook observer only triggers resize event on width change. I have included changes I suggested above in this code.
> 
> ```js
> const useResizeObserver = (ref: RefObject<HTMLElement>, onResize: (target: Element) => void) => {
> 
>   const boundsRef = useRef({ width: 0, height: 0 })
>   const [ro] = useState(() => new ResizeObserver(([entry]) => {
>     // resize callback gets only in width change
>     if (boundsRef.current.width !== entry.contentRect.width) onResize(entry.target)
>     boundsRef.current = entry.contentRect
>   }))
> 
>   useLayoutEffect(() => {
>     if (ref.current) ro.observe(ref.current)
>     return () => ro.disconnect()
>   }, []);
> }
> ```
> 
> As you can see using a ref we can easily achieve this. If we do this we should rename the hook to `useWidthObserver`. But that will dumb down the hook to just observing width. Also there is another possible solution where we can prevent change of height of toolbar due to overlay div. What do you think we should do ? I think first one is easier and reliable.

@shresthabijay Thanks for pointing out. The solution looks good but I'd prefer something simpler while it's possible. Updating the position of the  `toolbar-container` to `absolute` does the job because absolutely positioned elements don't add up to the parent height. This is what I did  -
```
  position: absolute;
  right: 0;
```

Let me know your thoughts.

src/components/Toolbar.js
anmolarora1: That's right. Though there's more work to be done with this component, I think we should do that refactoring in a separate issue.

src/hooks/useResizeObserver.ts
anmolarora1: > Looks good! But we can make it tighter with these changes.
> 
> ```js
> const useResizeObserver = (ref: RefObject<HTMLElement>, onResize: (target: Element) => void) => {
>   const [ro] = useState(() => new ResizeObserver(([entry]) => onResize(entry.target)))
>   useLayoutEffect(() => {
>     if (ref.current) ro.observe(ref.current)
>     return () => ro.disconnect()
>   }, []);
> }
> ```
> 
> So we don't have do access ref again in the callback.

With regard to splitting the logic between `useState` and `useLayoutEffect`, I'm not sure quite sure about that because we're creating `ResizeObserver` instance without even checking if the current reference has a valid value, thus making it a dead store in some cases.
`if (ref.current), then initialize and observer`
Ideally, we should get rid of the null-check altogether using TS but that'd be possible when `Toolbar.js` gets migrated to ts (hopefully soon). But for now, we'll have to live with it.
What do you think?

> 
> ```js
> useResizeObserver(toolbarRef, target => {
>     const { scrollWidth, clientWidth } = target
>     updateToolbarOrigin(scrollWidth - clientWidth > TOOLBAR_WRAPPER_DELTA ? toolbarOrigins.topRight : toolbarOrigins.topLeft)
> })
> ```

Passing the target element in `onResize` callback definitely helps with some logic abstraction and I'll update that.

raineorshine:
> @shresthabijay Thanks for pointing out. The solution looks good but I'd prefer something simpler while it's possible. Updating the position of the `toolbar-container` to `absolute` does the job because absolutely positioned elements don't add up to the parent height. This is what I did -
> 
> ```
>   position: absolute;
>   right: 0;
> ```
> 
> Let me know your thoughts.

`position: absolute` locks the element to the top of the document (or container) rather than the top of the window. We need `position: fixed` to ensure that the toolbar is visible when the user scrolls down.

anmolarora1:
> > @shresthabijay Thanks for pointing out. The solution looks good but I'd prefer something simpler while it's possible. Updating the position of the `toolbar-container` to `absolute` does the job because absolutely positioned elements don't add up to the parent height. This is what I did -
> > ```
> >   position: absolute;
> >   right: 0;
> > ```
> > 
> > 
> > Let me know your thoughts.
> 
> `position: absolute` locks the element to the top of the document (or container) rather than the top of the window. We need `position: fixed` to ensure that the toolbar is visible when the user scrolls down.

Sorry, it was a typo from my end. What I meant was `toolbar-overlay` because we need to update its position, not the position of `toolbar-container`. That remains unchanged.

shresthabijay:
> Updating the position of the `toolbar-container` to `absolute` does the job because absolutely positioned elements don't add up to the parent height. This is what I did -

You are correct. Overlay was adding height to the parent div. This works and is much simpler.

shresthabijay:
> Sorry, it was a typo from my end. What I meant was `toolbar-overlay` because we need to update its position, not the position of `toolbar-container`. That remains unchanged.

@raineorshine He is correct. It will only affect the overlay not the `toolbar-container`.

raineorshine:
Amazing, such a simple solution!

I will create a separate issue for fixing the scroll arrows.

I reverted your package-lock which removed hard dependencies. You have submitted broken package-locks before. Please find a better system to avoid this in the future, whether it is finding a fix for the npm issue or manually checking the package-lock before committing.

anmolarora1:
> Amazing, such a simple solution!
> 
> I will create a separate issue for fixing the scroll arrows.
> 
> I reverted your package-lock which removed hard dependencies. You have submitted broken package-locks before. Please find a better system to avoid this in the future, whether it is finding a fix for the npm issue or manually checking the package-lock before committing.

Okay, I'll ensure that. Perhaps we can match our versions of `npm` and `node`.

raineorshine:
I made it easier by adding a CI check in #759!

```
node v12.13.0
npm v6.14.5
```

Issue #378, opened 2020-02-26
Re-architect store/local/remote storage


raineorshine:
The redux store should serve as a single source of truth for the app. Currently `localStorage`, `localForage`, and `firebase` are used in too ad hoc a fashion. Consider how best to re-architect this aspect of the app to consolidate state persistence.

- Is there an existing library that syncs the redux store to other data stores?
- Could this be accomplished with custom middleware?

raineorshine:
- `sync` is my early attempt to consolidate storage logic
- Redux supports `persist`, which may be able to plug in to `localForage` and `firebase`
- Updates are done slightly differently between `localForage` (key-value store) and `firebase` (nested object)
- We can use the deep key path that is passed to `firebase` (e.g. `thoughtIndex/f3b88c`) as the key in `localForage`.
- Must support merging individual thoughts from different states based on timestamps (different devices).
- `migrate` must be run on `localForage` and `firebase` data due to offline mode and multiple devices.

raineorshine:
Some thoughts:

There are plenty of persistor libraries like [react-persist](https://github.com/rt2zz/redux-persist/), but it’s not clear how much effort it would take to configure the specific combination of `localStorage`, `indexedDB`, and `Firebase`. 

The Redux middleware only has access to the dispatched action, but the root reducer has access to the before and after states. If we have to write our own storage engines, it might be better to simply wrap one or more reducers.

Consider wrapping the `thoughtIndex` reducer since it is the last place we still have access to `thoughtIndexUpdates` before they become `thoughtIndex`. Other reducers will have to be wrapped for other persisted values (cursor, lastUpdated, tutorial, etc). Luckily they don’t require access to intermediate values. 

We should keep the persistence layer as decoupled as possible. Instead of calling `db` and `sync` functions directly, they should be called automatically on the appropriate state slice.

raineorshine:
Settled for isolating `sync` calls to `updateThought` reducer in order to avoid fat action-creators or extra diffing on state to determine updates.

raineorshine:
**Another idea:** 

1. Add updates to state. Merge with previous updates. 
1. Subscribe to the store 
1. If `updateQueue` is not empty, call `flushQueue` action-creator
1. `flushQueue` syncs to local db and remote, and clears the queue

Thus the reducers remain pure.

raineorshine:
[...]

PR #666, opened 2020-05-17
Move sync side effect to flushQueue action-creator

raineorshine:
Uses a middleware, action-creator, and reducer to coordinate detecting updates, syncing them, and clearing the queue. Well-defined, but tightly coupled across three different modules.

Contender PR vs #667 for #378.


anmolarora1:
This approach seems in sync with the existing flow of the app. The only difference from #667 is that we're using middleware to delegate `syncQueue` processing to the `flushQueue` action-creator. Apart from [this](https://github.com/cybersemics/em/pull/666/commits/6041e95a3bff5dd028ba181d556c33015fc0aae7#diff-e94e99d1ba66d976dc82b084b18c3f90R17) extra check, everything else is pretty easy to understand. As per my understanding, adding a `debounce` here would involve the same set of challenges as #667. 
Unless that's incorrect, I'd vote for this one due to its simplistic nature

raineorshine:
> This approach seems in sync with the existing flow of the app. The only difference from #667 is that we're using middleware to delegate `syncQueue` processing to the `flushQueue` action-creator. Apart from [this](https://github.com/cybersemics/em/pull/666/commits/6041e95a3bff5dd028ba181d556c33015fc0aae7#diff-e94e99d1ba66d976dc82b084b18c3f90R17) extra check, everything else is pretty easy to understand. 

Yup, that's right.

> As per my understanding, adding a `debounce` here would involve the same set of challenges as #667.

Since `flushQueue` is an action-creator, it can easily return a debounced thunk. The same is not true in #667 since the modified state has to be returned synchronously.

shresthabijay:
@raineorshine This looks good. I prefer this over #667 because store enhancer needs to be synchronous and it will be problem when we have to asynchronously clear the queue after sync has completed.

src/action-creators/flushQueue.js
```diff
@@ -0,0 +1,24 @@
...
+const syncQueue = () => (dispatch, getState) => {
+  const state = getState()
+
+  if (!hasQueue(state)) return
+
+  sync(state.syncQueue.thoughtIndexUpdates, state.syncQueue.contextIndexUpdates, { recentlyEdited: state.recentlyEdited })
+
+  dispatch({
+    type: 'clearQueue'
+  })
```
shresthabijay, line 21: As I understand we should clear the queue only after sync is complete. Since `sync` returns promise we can chain it to dispatch `clearQueue` only after sync is a success.

```javascript
sync()
.then(() => dispatch({ type: 'clearQueue' }))
```

src/reducers/updateThoughts.js
```diff
@@ -11,22 +10,24 @@ import {
...
   const recentlyEditedNew = recentlyEdited || state.recentlyEdited
 
-  setTimeout(() => {
-    sync(thoughtIndexUpdates, contextIndexUpdates, { recentlyEdited: recentlyEditedNew })
-  })
+  // updates are queued, detected by the syncQueue middleware, and sync'd with the local and remote stores
+  const syncQueueNew = {
+    thoughtIndexUpdates: { ...syncQueueOld.thoughtIndexUpdates, ...thoughtIndexUpdates },
+    contextIndexUpdates: { ...syncQueueOld.contextIndexUpdates, ...contextIndexUpdates },
+  }
```
shresthabijay, line 24: Here `syncQueueOld` is not defined. Maybe you forgot to get old queue from the state.

src/reducers/updateThoughts.js
```diff
@@ -11,22 +10,24 @@ import {
...
 
   return {
     contextIndex,
     expanded: expandThoughts(state, state.cursor, contextChain),
     recentlyEdited: recentlyEditedNew,
+    syncQueue,
```
shresthabijay, line 30: It should be 
```javascript
{ syncQueue : syncQueueNew }
```

src/action-creators/flushQueue.js
raineorshine: Not quite. If we don't clear the queue immediately, then `flushQueue` could get called again and the `syncQueue` could try to sync the same data. We have to clear the queue synchronously to prevent this loop.

If anything, we should store the pending queue separately so that it can be either cleared if the sync succeeds or restored if the sync fails. `Firebase` has its own internal retry mechanism, so there is no need to preserve `syncQueue` for `Firebase`. We could implement a retry/recovery mechanism for `IndexedDB`, but since that does not rely on a network connection it seems much less likely to fail, so I would consider it a lower priority.

src/reducers/updateThoughts.js
raineorshine: Thanks, you're right

PR #667, opened 2020-05-17
Move sync side effect to syncQueue store enhancer

raineorshine:
Use a store enhancer to detect, sync, and clear updates in one place. 

Might not be as amenable to debouncing since store enhancers must be synchronous (?).

Contender PR vs #666 for #378.

### DIFF
--- a/src/store-enhancers/syncQueueEnhancer.js
+++ b/src/store-enhancers/syncQueueEnhancer.js
@@ -1,0 +1,38 @@
+import {
+  sync,
+} from '../util'
+
+const initialSyncQueue = {
+  thoughtIndexUpdates: {},
+  contextIndexUpdates: {},
+}
+
+/* Returns true if the state has updates in the syncQueue */
+const hasUpdates = state => state && state.syncQueue && (
+  Object.keys(state.syncQueue.thoughtIndexUpdates).length > 0 ||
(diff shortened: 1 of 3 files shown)


anmolarora1:
This approach looks pretty solid and in congruence with our goal of separating the concerns. `Enhancers` require some in-depth understanding though so comprehending it might take some time. 
As you mentioned, implementing `debouncing` is going to be tricky and we'll have to explore some options around batching the updates and calling a `debounced sync` function. So the enhancer might get more complex, but we'll still have it separated from the rest of our code.

src/store-enhancers/syncQueueEnhancer.js
```diff
@@ -0,0 +1,38 @@
...
+    const newState = reducer(state, action)
+
+    // check if the new state has updates in the queue
+    if (hasUpdates(newState)) {
+
+      // sync updates to remote
+      sync(newState.syncQueue.thoughtIndexUpdates, newState.syncQueue.contextIndexUpdates, { recentlyEdited: newState.recentlyEdited })
+
+      // clear sync queue
+      newState.syncQueue = initialSyncQueue
```
shresthabijay, line 31: Store enhancer should be synchronous. If we want to clear the state only after sync is a success or call other dispatch calls when sync fails then it will be a problem.

PR #671, opened 2020-05-19
Prevent setting alert to null for valid gesture

### LINKED ISSUE #665: Mobile: Undo Archive Alert flashes
When a thought is archived on mobile, the undo alert flashes briefly then disappears. It should not immediately disappear. 

Tested on iOS Safari.

shresthabijay:
fixes #665

### DIFF
--- a/src/shortcuts.js
+++ b/src/shortcuts.js
@@ -113,6 +113,7 @@ export const handleGestureSegment = (g, sequence, e) => {
 }
 
 export const handleGestureEnd = (gesture, e) => {
+
   const state = store.getState()
   const { scrollPrioritized } = state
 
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine The issue was specific to mobile because alert was set to null at `handleGestureEnd` without checking if the gesture was valid. 

https://github.com/cybersemics/em/blob/06ad37bc7d2fe679d8c8776ca634cd5b2ea82f9a/src/shortcuts.js#L135

So whenever `delete` gesture was complete, `archiveThought` showed undo alert and immediately after that `handleGestureEnd`  set it to null. 

I made changes so that alert is set to null inside `handleGestureEnd` only if the gesture is not valid. Please review the changes. Thank you!

raineorshine:
That's definitely the cause, but unfortunately we need to clear the gesture hint when the gesture is valid.

In the process of investigating, I stumbled upon a solution. Will you review? #677

PR #677, opened 2020-05-19
Add alert.alertType to fix shortcut alert

raineorshine:


shresthabijay:
@raineorshine This solution is working. It's good to go.

PR #680, opened 2020-05-22
Sort: Cursor disappears on newly sorted thought

### LINKED ISSUE #629: Sort: Cursor disappears on newly sorted thought
## Steps to Reproduce

1. Create a new thought and give it a value
1. Activate “Sort” from the toolbar

## Current Behavior

Cursor disappears. 

## Expected Behavior

Cursor should stay on the thought. 

## Additional Information

Only an issue on a newly created thought.

aqkhan:
Fixes #629

### DIFF
--- a/src/shortcuts/toggleSort.js
+++ b/src/shortcuts/toggleSort.js
@@ -33,6 +33,7 @@ export default {
     const sortPreference = globalSort === 'Alphabetical' ? 'None' : 'Alphabetical'
     if (cursor) {
       dispatch(toggleAttribute(pathToContext(cursor), '=sort', sortPreference))
+      dispatch({ type: 'setCursor', thoughtsRanked: state.cursor })
     }
   }
 }


raineorshine:
> The issue I'd found was the cursorBeforeEdit not being set properly when the thought is blurred. I created a reducer just to update cursorBeforeEdit onBlur but it conflicted with cursor and the thought being edited started disappearing on the render. I ended up setting cursor on toggleSort to preserve it's presence there.> 
> 
> PS: The cursor is not actually disappearing, the isEditing prop is not being determined properly as the cursorBeforeEdit does not update correctly on blur.

raineorshine:
`cursor` and `cursorBeforeEdit` can be quite confusing. `cursorBeforeEdit` is not updated until the cursor moves. This was used to prevent the cursor thought from being re-rendered during editing. There may be a better way to do that.

Your solution works perfectly in this case. Thanks!

PR #685, opened 2020-05-23
Added space after starting emoji in thought value.

### LINKED ISSUE #663: Enforce space after emoji
During editing, when a thought begins with an emoji followed by a non-whitespace, non-emoji character, insert a space between the emoji and the character. 

e.g. “🧠Cybersemics” → “🧠 Cybersemics”

shresthabijay:
fixes #663


shresthabijay:
@raineorshine I used regex to check if value starts with consecutive emojis, followed by non whitespace and non emoji character. If so then we add space after those emojis.

🐼🐶animals    -->   🐼🐶 animals 

Please review the changes. Thank you!

src/util/addEmojiSpace.js
```diff
@@ -0,0 +1,15 @@
...
+// regex pattern that checks if string starts with number of emoji followed by non-whitespace, non emoji character
+const emojiWithoutSpaceRegex = new RegExp(`^${emojiRegexString}+(?!(\\s|${emojiRegexString}))`, '')
+
+/**
+ * If a string starts with an emoji followed by a non-whitespace, non-emoji character,
+ * then insert a space between the emoji and the character.
+ */
+export const addEmojiSpace = text => {
+  const match = text.match(emojiWithoutSpaceRegex)
+  return (match ? match[0] + ' ' + text.substring(match[0].length) : text).trim()
```
raineorshine, line 14: Let's remove the `trim` as it overreaches the purpose of `addEmojiSpace`.

If you are using it to remove a trailing space in the case where there is a match, instead detect that case and avoid adding the space to begin with, either via the regexp or checking the length of the rest of the text after the emoji.

PR #696, opened 2020-05-25
Convert newThought action-creator to reducer

raineorshine:
- [x] Use a selector to get rank of new thought. Adding `newRank` to the state just to pass it to the next reducer is subpar.


src/reducers/newThought.js
```diff
@@ -102,46 +109,46 @@ export default ({ at, insertNewSubthought, insertBefore, value = '', offset, pre
...
+        : tutorialStep === TUTORIAL2_STEP_CONTEXT2_PARENT_HINT ?
+          state => tutorialStepReducer(state, { value: TUTORIAL2_STEP_CONTEXT2_PARENT })
+          : tutorialStep === TUTORIAL2_STEP_CONTEXT2_HINT ?
+            state => tutorialStepReducer(state, { value: TUTORIAL2_STEP_CONTEXT2 })
+            : null,
+
+    // return new rank in case composed reducers need it
+    state => ({
+      newRank
+    })
```
shresthabijay, line 150: Won't this be adding `newRank` to the store state too ?

raineorshine: Yes, it is a subpar implementation. Hence the subtask:

>  Use a selector to get rank of new thought. Adding `newRank` to the state just to pass it to the next reducer is subpar.

Issue #602, opened 2020-04-28
Action-Creators and Reducers: Business Logic


raineorshine:
Should we have more business logic in action-creators or reducers?

This is mostly a response to “Recommendations for best practices regarding action-creators, reducers, and selectors”:
https://github.com/reduxjs/redux/issues/1171. 

This thread is 5 years old, so a newer source should probably be consulted before making a final conclusion. Nevertheless I read the whole thread and the consensus seemed to be that action-reducers should be small, simply representing the user or browser action, and the logic should go into the reducers. The original post that recommended fat action-creators was speculation from someone who had been using Redux for only a couple months. 

Reducers are easier to test and compose since they are pure functions. Plus you get the benefit of things like hot reloading when the logic is in reducers. There was a fair amount of pragmatism represented in true Redux fashion, but the recommend approach was in favor of logic in reducers. 

Here’s my evidence:

Dan Abramov:
https://github.com/reduxjs/redux/issues/1171#issuecomment-167704896

Imperative vs Reactive: https://github.com/reduxjs/redux/issues/1171#issuecomment-167714850

“Action creators are just a pattern for organizing the code. It’s convenient to have factories for actions because you want to make sure that actions of the same type have consistent structure, have the same side effect before they are dispatched, etc. But from Redux point of view, action creators don’t exist—Redux only sees actions.”

Hot reloading benefit: https://github.com/reduxjs/redux/issues/1171#issuecomment-205746859

A comment Dan made also suggested that composing reducers is generally a good thing: https://github.com/reduxjs/redux/issues/1171#issuecomment-196819727
“For reducers, we also don’t recommend using classes *because it will be harder to use reducer composition, that is, reducers that call other reducers*.” [emphasis mine]

More:
https://github.com/reduxjs/redux/issues/1171#issuecomment-205756731
https://github.com/reduxjs/redux/issues/1171#issuecomment-205865840

Other considerations:
- By auto calling the reducer with the same name as the action, it appears I’m following the [ducks](https://github.com/erikras/ducks-modular-redux) approach of co-locating actions and reducers. This is not something that Dan Abramov supports. I like the simplicity of it, but I will have to pay attention to how it couples actions to state and keep an eye out for potential cases where it would be more elegant to have multiple reducers respond to a single action. 
- There was some support expressed for the idea of handling side effects in middleware. I’d really like to do that to get `sync` out of the reducers, although I’m also consider wrapping a reducer so that it can be called just with the state delta (change in state).
[...]

PR #697, opened 2020-05-25
Convert subCategorizeAll action-creator to reducer

raineorshine:
Here is a simple example of converting a synchronous action-creator with no side effects into a reducer. This is the intended direction of the "fat reducers" approach discussed in #602.

Instead of dispatching multiple actions procedurally, the entry reducer composes multiple reducers using [reducerFlow](https://github.com/cybersemics/em/blob/newThought-reducer/src/util/reducerFlow.ts).

To reiterate, this is only possible with *synchronous*, *pure* action-creators. Action-creators with side effects will still need to be action-creators.

General advantages:
- Amenable to time travel, hot reloading, and undo/redo
- Pure functions are easier to test
- Eliminates need for `setTimeout` and other timing hacks (e.g. [L69-R83](https://github.com/cybersemics/em/pull/697/files#diff-3c28b615310e78bff4f6fecd88fe1c1bL69-R83))


shresthabijay:
@raineorshine This looks pretty good to me. Use of `reducerFlow` is a great improvement. Now we just need to convert synchronous action-creators to reducers.

PR #699, opened 2020-05-26
Adding UUID to thoughts

shresthabijay:
## Description

This PR adds `uuid` to thought on creation and persists it throughout it's lifecycle i.e (exisitingThoughtChange, existingThoughtMove, exisitingThoughtDelete etc). We also need to set `uuid` properly along with `rank` and `value` for `thoughtsRanked` and pass it to all the functions that directly or indirectly calls these methods. 

## Related PR
This PR originated from #503 . Flat render architecture needs unique identifier for each individual to properly identify a thought throughout the component's life cycle.


shresthabijay:
### Tasks

- [ ] Add `uuid` to thoughts on creation.
- [ ] Persist `uuid` on thoughts on `exisitingThoughtChange` .
- [ ] Persist `uuid` on thoughts on `exisitingThoughtMove`.

shresthabijay:
### Update

@raineorshine  I used a library called `uuid`. So I created and passed `uuid` in places where dispatch `newThoughtSubmit` is called. Instead of creating identifier inside `newThoughtSubmit` reducer itself , I created it in the function that dispatches it because at some cases `setCursor` needs to be called immediately after creating new thought and we need pass that `uuid` to the `cursor`. 

Also I changed `existingThoughtChange` to persist the `uuid`. So I had to change files that called `existingThoughtChange` and add `uuid` to calculated `thoughtsRanked` where ever needed. I also made similar changes to files that calls `setCursor` so that `uuid` is passed properly to `cursor`.

I tested most of the changes that I made to make sure that I didn't create any regressions. Please review my approach. Thanks!

raineorshine:
> So I created and passed `uuid` in places where dispatch `newThoughtSubmit` is called. Instead of creating identifier inside `newThoughtSubmit` reducer itself , I created it in the function that dispatches it because at some cases `setCursor` needs to be called immediately after creating new thought and we need pass that `uuid` to the `cursor`.

I think `newThoughtSubmit` is the proper place to create the uuid. Pay attention, as this is the kind of judgment you can improve! We want the architecture to be driven by good semantics and encapsulation, not by implementation details. The caller should simply specify the new thought’s value, where the new thought belongs, and what rank it has in that context. The uuid is an artifact of the creation process. Your solution skirts this problem by weakening the separation of concerns and creating unnecessary redundancy at the caller level. 

If we don’t want to change A to fit B, we have to change B to fit A. 

We should use a selector after the thought has been created to get its uuid. This is necessary since `newThoughtSubmit` is a reducer and returns the whole state.

I ran into the same issue in #696 to get `newRank` which I still need to refactor. 

Let’s make this change and see how the PR looks then.

src/selectors/rankThoughtsFirstMatch.js
```diff
@@ -74,7 +74,8 @@ export default (state, pathUnranked) => {
       value,
       // NOTE: we cannot throw an error if there is no parent, as it may be a floating context
       // unfortunately this that there is no protection against a (incorrectly) missing parent
-      rank: parent ? parent.rank : 0
+      rank: parent ? parent.rank : 0,
+      uuid: parent ? parent.uuid : ''
```
raineorshine, line 78: Let’s set it to `null` if there is no uuid. Thats more explicit than an empty string.

src/components/Editable.js
```diff
@@ -76,7 +76,7 @@ const stopPropagation = e => e.stopPropagation()
  * An editable thought with throttled editing.
  * Use rank instead of headRank(thoughtsRanked) as it will be different for context view.
  */
-const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOffset, showContexts, rank, style, dispatch }) => {
+const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOffset, showContexts, rank, uuid, style, dispatch }) => {
```
raineorshine, line 79: Can the uuid be derived from `thoughtsRanked`?

src/shortcuts/bumpThoughtDown.js
```diff
@@ -59,13 +60,15 @@ export default {
...
+          uuid: createUuid()
         })
 
         dispatch({
           type: 'setCursor',
           thoughtsRanked: unroot(contextOf(thoughtsRanked).concat({
             value: '',
-            rank
+            rank,
+            uuid: cursor.uuid
```
raineorshine, line 71: I believe uuid is a property of the child rather than the path as a whole.

shresthabijay:
> We want the architecture to be driven by good semantics and encapsulation, not by implementation details.

Yes agreed! I appreciate that you are helping me better understand about these important aspects.

shresthabijay:
> We should use a selector after the thought has been created to get its uuid. This is necessary since `newThoughtSubmit` is a reducer and returns the whole state.

How will we create a selector that exactly knows the latest newly added thought to the tree ? I believe the state doesn't provide with something like `latestAddedThought`.

src/shortcuts/bumpThoughtDown.js
shresthabijay: Thanks for pointing this our. We need to use `headUuuid` here.

src/components/Editable.js
shresthabijay: Yes. `uuid` is just redundant here. I will remove this.

src/selectors/rankThoughtsFirstMatch.js
shresthabijay: Sure!

raineorshine:
> How will we create a selector that exactly knows the latest newly added thought to the tree ? I believe the state doesn't provide with something like `latestAddedThought`.

That's right, but whoever called `newThoughtSubmit` specified the context and value of the new thought. That should be enough to select it. I'll implement it with rank to show you what I mean.

raineorshine:
See: 4b4b4f9

You can think about it as shifting an architecture problem to a performance problem. We can almost always optimize a specific performance problem without changing the interface. Getting the architecture right and reducing redundancy improves the codebase as a whole.

shresthabijay:
> You can think about it as shifting an architecture problem to a performance problem. We can almost always optimize a specific performance problem without changing the interface. Getting the architecture right and reducing redundancy improves the codebase as a whole.

Got it. I will make necessary changes and push it. Thanks!

shresthabijay:
> Looks good!
> 
> Could you actually change the property name from `uuid` to `id`, and `headUuid` to `headId`? I think that is cleaner and abstracts the concept of an id from the particular implementation of a uuid.

Sure !

shresthabijay:
@raineorshine I have made necessary changes and rebased to latest `dev`. Since thoughts now have id associated with them, I changed `cursorBack` test to check equality of cursor with just rank and value instead of comparing all the fields inside including `id`. Please review the changes. Thanks!

raineorshine:
Thanks! It looks like you may have recreated the action-creators that were converted to reducers. Please use the new reducers and ensure that the deleted action-creators stay deleted.

- `indent`
- `outdent`
- `setAttribute`
- `toggleAttribute`

`createAttribute` was removed completely since it was not being used.

shresthabijay:
@raineorshine I have made necessary changes. Please review it. Thanks!

raineorshine:
I'm having some doubts about the breadth of these changes. There are many functions that now have to manage ids that should not have to know about ids at all. For example, when moving a thought, does the caller really need to know the id? It seems like value and rank should be enough.

Can you store the id in the `contextIndex` entry and look it up when needed? This would avoid all of the id coupling that is going on.

shresthabijay:
> Can you store the id in the `contextIndex` entry and look it up when needed? This would avoid all of the id coupling that is going on.

Initially I thought `thoughtsRanked` should have all the three value, rank and id of thoughts. I agree it seems like it would be unnecessary. Do you suggest that instead of passing `id` inside `thoughtsRanked` we pass only `rank` and `value` like before and pass it to `exisitingThoughtChange` and `existingThoughtMove` ? Then we take the `rank` and `value` of thoughts and find their `id` inside these reducers and then persist them.

raineorshine:
Yes, exactly. We want to minimize the surface area of change, as each place that has to interact with id adds another coupling.

shresthabijay:
@raineorshine  I have changed `exisitingThoughtChange` and `exisitingThoughtMove` to find `id` and persist them. I have reverted all the unnecessary changes and added a test to check if `id` persists on `exisitingThoughtMove`. Please review it. Thanks!

src/reducers/existingThoughtChange.js
```diff
@@ -51,12 +52,14 @@ export default (state, { oldValue, newValue, context, showContexts, thoughtsRank
     ? contextOf(contextOf(thoughtsRanked)).concat({ value: oldValue, rank: headRank(contextOf(thoughtsRanked)) }).concat(head(thoughtsRanked))
     : contextOf(thoughtsRanked).concat({ value: oldValue, rank })
 
+  const oldPath = rankThoughtsFirstMatch(state, thoughtsOld)
```
raineorshine, line 55: `rankThoughtsFirstMatch` is off limits. It merely guesses at the ranks, and should only be used for restoring the cursor from a lossy url.

shresthabijay: Oh! I understand now. 

> @shresthabijay Can you clarify: Where are you storing the id? I assume you want to store it in one place and then look it up as needed.
> 
> * thoughtIndex
> * contextIndex
> * cursor

I have stored it in `thoughtIndex` and `contextIndex` and I persist them only when thought is moved or changed.

shresthabijay: > `rankThoughtsFirstMatch` is off limits. It merely guesses at the ranks, and should only be used for restoring the cursor from a lossy url.

Oh! I thought it was used to get the exact thought from the given context. So should I use `getThoughts` and then use `find` to match the exact thought ?

shresthabijay:
> @shresthabijay Can you clarify: Where are you storing the id? I assume you want to store it in one place and then look it up as needed.

I am storing it inside both `contextIndex` and `thoughtIndex`. Since `id` is directly associated with `thought`, it makes sense to store it inside `thoughtIndex`  and `contextIndex`. I am getting old thought's id inside `existingThoughtChange` and `existingThoughtMove` and then persist them.

src/reducers/existingThoughtChange.js
raineorshine: Hmmm I see now that `rankThoughtsFirstMatch` already snuck into this function. This isn't really what we want here, but since it's already in `dev` we can refactor it in a separate PR.

https://github.com/cybersemics/em/commit/1c74cc5f5097e42a7ca35901c10fe92aab0677b2#diff-f3709f42ab08e75fdd90c210ec93a4e3R62-R63

raineorshine: Can you get id from `thoughtOld`? You shouldn't have to do any iteration.

raineorshine:
> > @shresthabijay Can you clarify: Where are you storing the id? I assume you want to store it in one place and then look it up as needed.
> 
> I am storing it inside both `contextIndex` and `thoughtIndex`. Since `id` is directly associated with `thought`, it makes sense to store it inside `thoughtIndex` and `contextIndex`. I am getting old thought's id inside `existingThoughtChange` and `existingThoughtMove` and then persist them.

Okay, that sounds good. Can we avoid storing it also in `path` or `cursor` then?

src/reducers/existingThoughtChange.js
shresthabijay: Sure I will do that!

shresthabijay: @raineorshine Since id is associated with individual thought we need to iterate over `thoughtOld`'s `contexts` to find the exact thought.

shresthabijay:
> Okay, that sounds good. Can we avoid storing it also in `path` or `cursor` then?

I think I have made that change. I have avoided passing `id` to `thoughtsRanked` that is passed into `exisitingThoughtChange` and `existingThoughtMove` . And also avoided storing it inside `cursor`.

src/reducers/existingThoughtChange.js
raineorshine: That's right. Okay. At least see if there is any way to utilize existing computations in the function to avoid an extra iteration.

shresthabijay: @raineorshine Sure! Is there other way we can get exact ranked path instead of using `rankedThoughtsFirstMatch` ? We can simply get the head of the ranked path to access the exact thought. We won't have to do extra iteration over.

raineorshine: The only good way would be to actually store the `id` in the cursor. I was hoping to avoid that since there is so much logic that works with the cursor. Preserving the id there is non-trivial.

`existingThoughtChange` etc intentionally takes a `Child` and a `Context` rather than a `Path`. i.e. it shouldn't matter how the user got there, just which thought to change.

shresthabijay: Yes I agree about `existingThoughtChange` taking only `Child` and `Context`. But with addition of recently edited thought we still need a `thoughtsRanked`  inside `existingThoughtChange`. Also we need to get `id` of `head` to persist.

In current code base we have `thoughtsRanked` being passed to `existingThoughtChange`. `thoughtsRanked` has `RankedThought` with `rank`, `value` and `lastUpdated`. Now it will have `id` too.  This `thoughtsRanked` is used to get `headRank` and `headValue` easily inside `existingThoughtChange`. We could use same `rankedThoughts` to get `headId` or use it for recently edited thoughts. But there is a problem with this. Normal editing from `Editable` passes `thoughtsRanked` with all the available properties `rank`, `value`, `lastUpdated` and `id`. But functions like `splitFunctions` that also calls `existingThoughtChange`, pass `thoughtsRanked` with the `head` RankedThought with just `rank` and `value`. 

https://github.com/cybersemics/em/blob/cc1b235094d7685410d128058bd4793b9b5c017c/src/reducers/splitThought.js#L50

So we don't have access to `lastUpdated` and `id` of `head` thought. This inconsistency is forcing us to find ranked path again inside `existingThoughtChange` even though we already have `rankedThoughts` passed as function paramater.

But the question is should properties like `lastUpdated` and `id` be part of `RankedThought` inside `thoughtsRanked` ? If yes then we should pass it everywhere without exception otherwise it will cause confusion. If it's a `thoughtsRanked` then should we expect it to have only `rank` and `value` or should it have all properties associated with the thought ?

raineorshine: Thank you, that makes sense.

I guess when it comes down to it, I am willing to add `id` everywhere it's justified. However I want to keep the interfaces clean and there may be places where it's better to derive `id` then force a caller to know about it. It's a bit fuzzy where to draw the line. I agree we want to reduce confusion, and it should be made very clear what is required.

- In one case, `{ value, rank }` serves as the minimal information to uniquely identify a thought within a context.
- In the other case, `{ value, rank, id, created, lastUpdated }` represents the full `Child` with all of its metadata.

So:

- If we *have* the metadata (id, etc), we should be passing it through. 
- If the metadata is required by the callee, we should be passing it through, even if we need to get or calculate it.
- If the metadata is not required by the callee, and it would take additional work to get the metadata, perhaps it represents a minimal interface rather than a full `Child` node.

Thanks for the discussion. At this point, after taking the above into account, I am open to your recommendation.

shresthabijay: Yes. It's very fuzzy where to draw the line. Let's list out what changes we need to make when we want caller (ex: `importText.js`) to pass `thoughtsRanked` with metadata that is required by callee that needs to persists id (ex: `exisitingThoughtChange`).

- At most places caller has easy access to metadata and can pass `id`, `lastUpdated` along with `rank` and `value` to `thoughtsRanked`
- We need to make changes to some utility functions that is used to calculate `thoughtsRanked` to pass necessary metadata like `id`. ex: `addContext.js`.
- Since many places uses `cursor` to derive `thoughtsRanked` to pass to callee , so we need to make sure `cursor` is always populated with `metadata`. That means we need to make necessary changes where `setCursor` is called.
- Sometimes `setCursor` is called right after `newThoughtSubmit`. We need to find the newly added thought and then pass required metadata to make sure cursor is consistently populated with `id`.

Now on the other hand if we want to callee to handle finding metadata of thought instead of making caller to pass metadata on `thoughtsRanked` itself.

- I am assuming that except `head` of the `thoughtsRanked`  passed to `exisitingThoughtChange` all other thoughts have the meta data associated with them. In this case we need to make extra iteration of `thoughtOld`'s `context`  property to find the exact thought and it's metadata. We can use it to persist `id` and create a path that has all metadata of its thought to pass it to recently edited thought.

shresthabijay: > * If the metadata is not required by the callee, and it would take additional work to get the metadata, perhaps it represents a minimal interface rather than a full `Child` node.

Calee like `exisitingThoughtChange` and `existingThoughtMove` are responsible for persisting metadata like `id`. So every caller that calls these functions need to make sure that they pass `thoughtsRanked` with all metadata. We just need to make changes to these callers that call these specific functions.

raineorshine: Thank you, this has given me the chance to think about this more, although I am led to make the counter recommendation.
> 
> * At most places caller has easy access to metadata and can pass `id`, `lastUpdated` along with `rank` and `value` to `thoughtsRanked`
> * We need to make changes to some utility functions that is used to calculate `thoughtsRanked` to pass necessary metadata like `id`. ex: `addContext.js`.
> * Since many places uses `cursor` to derive `thoughtsRanked` to pass to callee , so we need to make sure `cursor` is always populated with `metadata`. That means we need to make necessary changes where `setCursor` is called.
> * Sometimes `setCursor` is called right after `newThoughtSubmit`. We need to find the newly added thought and then pass required metadata to make sure cursor is consistently populated with `id`.

These changes are extensive and add tight coupling to the id across multiple modules. I still think we should avoid this. The more minimal solution of deriving the id as needed gives us a minimal implementation and defers the additional risk. I would be more comfortable considering the more extensive implementation in a refactor down the road after we have been using ids with the new render pipeline for some time.

Another possibility is that we *only* pass id, and not pass value or rank at all, looking those up as needed. That's a huge change, but it should be considered. Deferring the decisions gives us more time to evaluate the merits of that approach without holding up the flat render task.

> Now on the other hand if we want callee to handle finding metadata of thought instead of making caller to pass metadata on `thoughtsRanked` itself.
> 
> * I am assuming that except `head` of the `thoughtsRanked`  passed to `exisitingThoughtChange` all other thoughts have the meta data associated with them. In this case we need to make extra iteration of `thoughtOld`'s `context`  property to find the exact thought and it's metadata. We can use it to persist `id` and create a path that has all metadata of its thought to pass it to recently edited thought.

Yes, and it will get better since I expect a refactor where `thoughtObject.context` becomes an object rather than an array so we can do O(1) lookup by value + rank.

Does recently edited need the id? I thought it was the flat render that needed it.

> * We can use a function that gives proper ranked path when given a context. But it will just be too much to do that and it doesn't make sense when we already have `thoughtsRanked` as parameter.

A Path cannot, even in theory, be deterministically derived from a Context, since there are many Paths that may lead to a Context.

> The later solution is easy to implement and requires less changes but it still doesn't solve inconsistency on the `thoughtsRanked` we are receiving from the caller. It will create confusion on why are we finding the metadata specifically for the `head` thought. So I would suggest we take the first approach. Let me know what you think.

I would rather a little confusion in one or two places than an extensive interface change across the app that we, or at least I, am yet convinced is the optimal architecture.

shresthabijay: > Another possibility is that we _only_ pass id, and not pass value or rank at all, looking those up as needed. That's a huge change, but it should be considered. Deferring the decisions gives us more time to evaluate the merits of that approach without holding up the flat render task.

Yes I agree with you. Since id can uniquely identify thoughts, so it makes sense to go for this approach.

shresthabijay: > A Path cannot, even in theory, be deterministically derived from a Context, since there are many Paths that may lead to a Context.

Ahh yes. So the problem still remains that we don't have all metadata inside `thoughtsRanked` passed inside `existingThoughtChange` and we cannot get full path just from `context` to derive exact id of head thought. So calculating full path inside `exisitingThoughtChange` can be ruled out. 

Either we pass `id` from all the callers or we go with the plan you suggested to pass only id and instead use it to derive exact thought.

raineorshine: > So calculating full path inside exisitingThoughtChange can be ruled out.

But `existingThoughtChange` is already passed the full path: `thoughtsRanked`. It may have been shoehorned in, but it's there. I suggest we use it for now. Unless this isn't what you're needing?

> Either we pass id from all the callers or we go with the plan you suggested to pass only id and instead use it to derive exact thought.

Interesting, those were the two options I ruled out 😂. I was suggesting that the least risky approach right now would be to pass the id only where needed (not all callers) or derive it when possible.

shresthabijay: @raineorshine Let's start from where we left last time. I am still not sure about which implementation we are going for :joy: . I think for now we should just derive `id` from `contextIndex` inside `exisitingThoughtsChange` and `exisitingThoughtsMove`. It would only require us making changes to these files.

raineorshine: Yes, let's go with that for now! Apologies for the confusion.

shresthabijay:
@raineorshine I am little confused by this test.
https://github.com/cybersemics/em/blob/131128333d754b4686e7e2cbc479c7d3425cbde3/src/reducers/__tests__/existingThoughtMove.js#L199-L237

As you can see in line 211 that this specific `oldPath` is not actually available in the tree given by 
```
- a 
   - m 
    - x 
  - m 
    - y
```

https://github.com/cybersemics/em/blob/131128333d754b4686e7e2cbc479c7d3425cbde3/src/reducers/__tests__/existingThoughtMove.js#L211

Since a thought with this path is not available inside `thoughtIndex` , proper `id` cannot be derived in side`exisitingThoughtMove`. This is causing two similar tests to fail.

shresthabijay:
@raineorshine I have fixed indentation errors for move merge tests. Please review the changes.

src/reducers/existingThoughtChange.js
```diff
@@ -43,18 +45,21 @@ export default (state, { oldValue, newValue, context, showContexts, thoughtsRank
     ? contextOf(contextOf(thoughtsRanked)).concat({ value: oldValue, rank: headRank(contextOf(thoughtsRanked)) }).concat(head(thoughtsRanked))
     : contextOf(thoughtsRanked).concat({ value: oldValue, rank })
 
+  /** Find exact thought from thoughtIndex. */
+  const exactThought = () => thoughtOld.contexts.find(thought => equalArrays(thought.context, context) && thought.rank === rank)
+  const id = headId(thoughtsRanked) || exactThought().id
```
raineorshine, line 50: Since `exactThought` is always called, you can use a normal variable rather than a function:

```js
const exactThought = thoughtOld.contexts.find(thought => equalArrays(thought.context, context) && thought.rank === rank)
```

src/util/addContext.ts
```diff
@@ -1,18 +1,18 @@
...
   ...notNull({
     contexts: (thought.contexts || [])
-      .filter(parent =>
+      .filter((parent: ThoughtContext) =>
         !(equalArrays(parent.context, context) && parent.rank === rank)
       )
-      .concat({ context, rank }),
-    created: thought.created || timestamp(),
+      .concat({ context, rank, id }),
+    created: thought.created,
```
raineorshine, line 15: Let's  keep `|| timestamp()` to be safe.

PR #703, opened 2020-05-27
Thought Cache & Iterative Loading

### LINKED ISSUE #273: Staged local loading

raineorshine:
Fixes #273.


raineorshine:
## UPDATE:

On `dev` I changed `Subthoughts` to re-render based on changed children without a forced `render` action. This made it possible to solve some issues without the rendering issues I was running into before. I fixed issues related to buffered thoughts not loading and thoughts loading more than once, creating duplicates.

Iterative loading is a highly fragile because of all the concurrent dependencies:

- It depends on `cursor` and `expanded` to determine which thoughts need to be loaded
- It depends on the `syncQueue` because it has to make sure not to load thoughts that are queued to be deleted or modified. 
- It depends on `dexie` and `firebase` because it keeps them in sync and updates state.

We'll see how it holds up to continued testing. The issues I fixed today were all created with very simple structures in offline mode. Not sure how it is behaving when connected to the remote. Hopefully there are few bugs left in simple cases and I can move on to test with real user data soon.

iterative-loading

This shows thoughts loading in iteratively from dexie, with a short buffer depth (2) so that it is more visible. It will do the same thing from firebase, seamlessly sync with state and dexie. In production I will increase the buffer depth so it will pre-load thoughts that are not visible. This should make load times very low for any number of thoughts. The initial load and searching (which has to be done server-side now) should be the only things with noticeable delay. At least that's the idea :).

shresthabijay:
@raineorshine I tested iterative loading. It's working well. There are few things I noticed.
- Thoughts are not being synced from firebase. I logged in and created thoughts. I tried logging out and logging in again. It doesn't load data from firebase. I checked out to latest dev and it is working there.
-  When cursor is in home context its grand children do not show if they are expandable until its parent is focused.
Screen Capture_select-area_20200820202842

- In this gif I am reloading the app. At initial load you can see that tree expands at the beginning and then contracts back. Then expands again.
Screen Capture_select-area_20200820220746

raineorshine:
Thanks for your help!

> * Thoughts are not being synced from firebase. I logged in and created thoughts. I tried logging out and logging in again. It doesn't load data from firebase. I checked out to latest dev and it is working there.

Were you logging into an account with existing data? I think the data schema has changed, but I will need to confirm. If that's the case, it probably explains why it's ignoring the data completely. I was able to delete my remote data manually in the Firebase console and confirm that the basics are working with a clean start. Need to test more complex reconciliation scenarios.

To test, you can login with a new account, set up your own Firebase Realtime Database instance, or give me permission to delete your existing Firebase data.

> * When cursor is in home context its grand children do not show if they are expandable until its parent is focused.

Fixed in 580af2413.

> * In this gif I am reloading the app. At initial load you can see that tree expands at the beginning and then contracts back. Then expands again.

Possible fix in 413271e2d. Could you try again? If it's still broken, please provide any additional information that might help me reproduce (login status, tutorial status, etc)

## Next Steps

- Logout is throwing an error
- Remote loading and reconciliation needs to be tested more rigorously
- Buffer size needs to be expanded and further tested (currently loads 1 level beyond visible thoughts). I purposefully set it low since that will result in the most frequent loading and possible concurrency issues.

shresthabijay:
@raineorshine  I tested the latest PR. Thought is being synced from firebase. I used fresh account for the test. However I still see some issues.
- Create following structure in the tree without logging in.
`
- a
  - b
    - c
  - d
`
Screen Capture_select-area_20200821103102

Screen Capture_select-area_20200821103206

raineorshine:
> > Also I noticed that `firebaseProvider.getManyDescendants` takes significantly long time to return data for fairly small amount of thoughts. So for the first sync of local db from firebase it takes significant time to navigate.
> 
> Okay, thanks. I suspected that might be slow. I will look into optimizing this once the functionality is stable. Currently it uses a naive implementation that calls `getContext` separately for each thought. This can be combined into a single fetch for all children without changing the API. That should be a significant improvement, even if it still requires multiple fetches for deeper descendants.

Added `getContextsByIds` to fetch child context in a single db call (f272ab4).

The function still blocks until all descendants are fetched (up to the given `maxDepth`). Instead, we should probably return a stream-like object so that the first level of thoughts can be rendered immediately.

PR #704, opened 2020-05-28
Set oldCursor of Editable on importText

### LINKED ISSUE #686: Editing stops working after paste
## Steps to reproduce

1. Create a new thought.
2. Copy and paste any text into the thought.
3. Try editing the thought again.

## Current Behavior

Editing stops working completely.

## Expected Behavior

Editing should work normally.

shresthabijay:
fixes #686 

### Description
`onPaste` function on `Editable` calls `importText` but doesn't update `oldCursor`. So `oldCursor` gets out of sync and on further edit , it cannot find thought associated with out of sync `oldCursor` , preventing `onExistingThoughtChange` dispatch.

https://github.com/cybersemics/em/blob/b173be7a9160b9c51c3ba4ae996508e71e1dba08/src/components/Editable.js#L162-L165

### Fix

`importText` updates the original thought only when `numLines === 1`. That means we need to update the `oldCursor` when this happens. So I returned `newValue` from the `importText` action creator when original thought is updated. Then I updated `oldCursor` inside `Editable` if `newValue` is returned by `importText` dispatch.

### DIFF
--- a/src/action-creators/importText.js
+++ b/src/action-creators/importText.js
@@ -148,6 +148,10 @@ export default (thoughtsRanked, inputText, { preventSetCursor, preventSync, rawD
         offset: startOffset + newText.length
       })
     }
+
+    return Promise.resolve({
+      newValue
+    })
   }
   else {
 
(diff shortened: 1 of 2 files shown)


raineorshine:
Well done! Thank you!

PR #705, opened 2020-05-28
Prevent selecting meta thoughts on cursor forward

### LINKED ISSUE #692: cursorForward: Ignore hidden thoughts
When `cursorForward` is activated with an empty cursor history, the cursor is moved to its first child if it exists. This should not include hidden thoughts or meta attributes when `showHiddenThoughts` is false. 

Mobile only. 

## Steps to Reproduce

- a
  - =test
  - b

0. Make sure `showHiddenThoughts` is off. 
1. Tap outside the thoughts until the cursor is `null`. 
2. Swipe ← to activate `cursorForward` which places the cursor on `a`. 
3. Swipe ← again to activate `cursorForward`. 

## Current Behavior

Cursor is on `a.=test`

## Expected Behavior

Cursor should be on `a.b`

shresthabijay:
fixes #692 

### Description
When `cursorHistory` length is zero `cursor` is moved to first child without checking if the child is a `meta` function and showHiddenThoughts is `false`.

### Fix
Instead of selecting the first child directly, we check if `showHiddenThoughts` is false and find the first child that is not a `meta` thought.

### DIFF
--- a/src/reducers/cursorForward.js
+++ b/src/reducers/cursorForward.js
@@ -3,6 +3,11 @@ import {
   getThoughtsRanked,
 } from '../selectors'
 
+// utils
+import {
+  isFunction,
+} from '../util'
+
 // reducers
 import setCursor from './setCursor'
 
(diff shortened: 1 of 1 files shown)


shresthabijay:
> You'll also want to check the `hidden` attribute. I've used this function before (which we really ought to factor out):
> 
> ```
> const notHidden = child => 
>   !isFunction(child.value) && !meta(state, [...context, child.value]).hidden
> ```

Ahh, I almost forgot about thought with `hidden` attribute. I will make this change. Thank you !

shresthabijay:
@raineorshine I have made the requested change and rebased to latest `dev` too. Please review the changes!

raineorshine:
Thank you!

Issue #719, opened 2020-06-02
Wrong splitThought behavior: valueLeft is always empty


pushkov-fedor:
Mobile devices

# Steps to Reproduce
Create a new thought with two or more words.
Tap between two words
Hit Enter

# Current Behavior
An empty thought is created before the dictated thought

# Expected Behavior
The dictated thought should be split into two thoughts at the browser selection

PR #720, opened 2020-06-02
Fix wrong splitThought behavior

pushkov-fedor:
Issue #719

@raineorshine The problem is in `asyncFocus` function within newThought shortcut. On desktop this function does nothing, on mobiles it creates input element and set focus on it. I think, that's why expression `offset || window.getSelection().focusOffset` in splitThought.js evaluates to 0. Can you explain me, what is purpose of `asyncFocus` function? I haven't removed it yet.
Another solution - we can leave `asyncFocus` function and just pass offset to `splitThought` reducer

### DIFF
--- a/src/shortcuts/newThought.js
+++ b/src/shortcuts/newThought.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { isMobile } from '../browser'
+import { isMobile, isSafari } from '../browser'
 
 // constants
 import {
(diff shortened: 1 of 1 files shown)


pushkov-fedor:
Actually passing an offset to reducer is a good way, because we calculate offset in `newThought` shortcut and then calculate it again in `splitThought` reducer

raineorshine:
> @raineorshine The problem is in `asyncFocus` function within newThought shortcut. On desktop this function does nothing, on mobiles it creates input element and set focus on it. I think, that's why expression `offset || window.getSelection().focusOffset` in splitThought.js evaluates to 0. Can you explain me, what is purpose of `asyncFocus` function? I haven't removed it yet.

Basically, if there is no active selection, Mobile Safari will only allow programmatic selection within a click or touch event handler. Otherwise trying to focus or set the selection does nothing. To be able to set the selection in an asynchronous callback, you have to first set the selection to an arbitrary element in the initial click or touch handler. Then setting the selection will work.

See https://stackoverflow.com/a/45703019/480608 for more information. (This link is given in `asyncFocus.js`.)

raineorshine:
> Actually passing an offset to reducer is a good way, because we calculate offset in `newThought` shortcut and then calculate it again in `splitThought` reducer

I'm good with that.

If `asyncFocus` is really only needed for Mobile Safari, we could also tighten that condition so it doesn't run on Android.

PR #721, opened 2020-06-02
Updated regex and added remove spaces regex

### LINKED ISSUE #691: Some emojis not sorted correctly
Some emojis are not sorted correctly with text. All emoji should be sorted by the text with the other emojis. 

## Steps to Reproduce

```
- x
  - ❤️ e
  - 🧡 d
  - 💛 c
  - ⭐️ b
  - 🌋 a
```

1. Paste the above text
2. Sort `x`

## Current Behavior

```
- x
  - 🌋 a
  - 💛 c
  - 🧡 d
  - ⭐️ b
  - ❤️ e
```

## Expected Behavior

```
- x
  - 🌋 a
  - ⭐️ b
  - 💛 c
  - 🧡 d
  - ❤️ e
```

aqkhan:
Fixes #691


src/util/compareThought.js
```diff
@@ -1,10 +1,11 @@
...
+const regexEmojis = /([#0-9]\u20E3)|[\xA9\xAE\u203C\u2047-\u2049\u2122\u2139\u3030\u303D\u3297\u3299][\uFE00-\uFEFF]?|[\u2190-\u21FF][\uFE00-\uFEFF]?|[\u2300-\u23FF][\uFE00-\uFEFF]?|[\u2460-\u24FF][\uFE00-\uFEFF]?|[\u25A0-\u25FF][\uFE00-\uFEFF]?|[\u2600-\u27BF][\uFE00-\uFEFF]?|[\u2900-\u297F][\uFE00-\uFEFF]?|[\u2B00-\u2BF0][\uFE00-\uFEFF]?|(?:\uD83C[\uDC00-\uDFFF]|\uD83D[\uDC00-\uDEFF])[\uFE00-\uFEFF]?|[\u20E3]|[\u26A0-\u3000]|\uD83E[\udd00-\uddff]|[\u00A0-\u269F]/g
+const regexSpaces = /^\s+|\s+$/g
 
-/** Remove emojis from camparator inputs using regex. */
-const removeEmojis = str => str.replace(regexEmojis, '')
+/** Remove emojis and trailing/leading spaces from camparator inputs using regex. */
+const removeEmojisAndSpaces = str => str.replace(regexEmojis, '').replace(regexSpaces, '')
```
raineorshine, line 8: I think you can use `.trim()` instead of `regexSpaces`.

raineorshine:
Confirmed fixes #691.

raineorshine:
But there is a regression:

## Steps to Reproduce

- x
  - ❤️ e
  - 🧡 d
  - c
  - ⭐️ b
  - a

## Current Behavior

- x
  - a
  - ⭐️ b
  - c
  - 🧡 d
  - ❤️ e

## Expected Behavior

- x
  - ⭐️ b
  - 🧡 d
  - ❤️ e
  - a
  - c

raineorshine:
This is also a great case for a test! Would you write a test for this modeled after `Subthoughts`? You can dispatch `toggleSort` at the appropriate time and then check the order of the rendered subthoughts.

https://github.com/cybersemics/em/blob/dev/src/components/__tests__/Subthoughts.js

aqkhan:
@raineorshine Just to confirm; the requested regression resolution depicts thoughts starting with emojis should always be placed above the ones without emojis.

aqkhan:
> This is also a great case for a test! Would you write a test for this modeled after `Subthoughts`? You can dispatch `toggleSort` at the appropriate time and then check the order of the rendered subthoughts.
> 
> https://github.com/cybersemics/em/blob/dev/src/components/__tests__/Subthoughts.js

Sure.

raineorshine:
> @raineorshine Just to confirm; the requested regression resolution depicts thoughts starting with emojis should always be placed above the ones without emojis.

Correct

src/util/compareThought.js
```diff
@@ -23,6 +23,18 @@ export const compare = (a, b) => a > b ? 1 : a < b ? -1 : 0
...
 
+/** A comparator that sorts string starting with emojis ahead of other. */
+export const compareStringsWithEmojis = (a, b) => {
+  const aStartsWithEmoji = regexEmojis.test(a)
+  const bStartsWithEmoji = regexEmojis.test(b)
+  const aStripped = removeEmojisAndSpaces(a)
+  const bStripped = removeEmojisAndSpaces(b)
+  return aStartsWithEmoji && !bStartsWithEmoji ? -1
+    : bStartsWithEmoji && !aStartsWithEmoji ? 1
+    : aStartsWithEmoji && bStartsWithEmoji ? compareReasonable(aStripped, bStripped)
```
raineorshine, line 34: You'll want to simply return `0` if they both start with an emoji. Each comparator functions independently and should only compare its relevant criterion. `makeOrderedComparator` handles composing the comparators together in the correct order.

src/util/compareThought.js
```diff
@@ -100,8 +110,9 @@ const compareReasonable = makeOrderedComparator([
   compareNumbers,
   compareDateAndOther,
   compareDateStrings,
+  compareStringsWithEmojis,
```
raineorshine, line 113: I would give this higher precedence: after `comparePunctuationAndOther`.

raineorshine:
Will you be writing the test in this PR or another? Thanks!

aqkhan:
> I would suggest looking for a way to compose `removeEmojisAndSpaces` using `makeOrderedComparator`. Perhaps you can make a nested comparator. It's preferable not to couple `removeEmojisAndSpaces` with all the other comparator functions since they are meant to function independently.

Acknowledged. I am working on that already.

src/util/compareThought.ts
```diff
@@ -84,10 +94,14 @@ export const makeOrderedComparator = (comparators: ComparatorFunction<any>[]): C
...
       ? 0
       // if the values are non equal by the initial comparator, return the result of the comparator
-      : comparators[0](a, b) ||
+      : execComparator(comparators[0], a, b) ||
         // if they are equal, move on to the next comparator
         makeOrderedComparator(comparators.slice(1))(a, b) // RECURSION
 
+/** Executes and comparator with respective string modifications. */
+const execComparator = (comparator: ComparatorFunction<any>, a:any, b: any) =>
+  comparator.name === 'compareStringsWithEmojis' ? comparator(a, b) : comparator(removeEmojisAndSpaces(a), removeEmojisAndSpaces(b))
```
raineorshine, line 103: This is very creative, but unfortunately couples the wrong things together. `execComparator` sounds like a general function, but is tightly coupled to `compareStringsWithEmojis` and `removeEmojisAndSpaces`. Why would a general function have specialized knowledge about emojis? Further, `makeOrderedComparator` depends on `execComparator`, inheriting all of those same couplings, defeating the purpose of a generalized `makeComparator` function.

I will see if I can come up with a preferable solution so you can see what I mean.

src/components/__tests__/Subthoughts.js
```diff
@@ -140,3 +140,40 @@ it('context view', async () => {
       thoughtsRanked: [{ value: 'b' }, { value: 'ones' }, { value: 'y', rank: 7 }],
     })
 })
+
+// Sort context
+it('normal view', async () => {
```
raineorshine, line 145: Rather than test the detailed comparison functionality within the Subthoughts component, we should test it lower down, in `util/__tests__/compareThought.js`.

aqkhan: Makes sense.

raineorshine: Wait till I push my change since I'll need to create at least one test myself.

aqkhan: Sure.

raineorshine:
677a2f1 demonstrates the general idea of nesting `makeOrderedComparator` while stripping emojis from the inner comparator. I *think* this should work, or something similar, but unfortunately I wasn't able to test properly because the emoji regex is not working with other text.

```
> regexEmojis.test('a')
false
> regexEmojis.test('😂')
true
> regexEmojis.test('😂 a')
false
> regexEmojis.test('😂 b')
true
> regexEmojis.test('😂 c')
false
> regexEmojis.test('😂 d')
true
```

Once you get the regex working, I bet you'll be able to come up with a working comparator. Thanks!

src/components/__tests__/Subthoughts.js
raineorshine: I acknowledge that I steered you towards the Subthoughts test to begin with, so that was my error.

aqkhan:
Acknowledged. Thanks!

aqkhan:
@raineorshine Just pushed the updates. One of the tests regarding "context view" is failing though. I've ran tests before pushing the code and it worked fine on my machine. Looking into it now.

raineorshine:
> @raineorshine Just pushed the updates. One of the tests regarding "context view" is failing though. I've ran tests before pushing the code and it worked fine on my machine. Looking into it now.

Tests were failing in `dev`. That's my bad. Fixed now.

Issue #538, opened 2020-04-12
Correctly position sorted drop hover


raineorshine:
When a thought is dragged over any of the subthoughts of a sorted context, the drop hover (blue line) should be shown at the sorted drop location. 

Currently the drop hover position is misleading. It should instead be positioned at the point where the dragged thought will be sorted to.

incora-dev:
@raineorshine Estimate: 2.5h

raineorshine:
@incora-dev This is a challenging one. If you want to take this on, I ask for a written description of your proposed solution before you begin implementing it. 2-3 hours for the proposal and 4-6 hours total.

incora-dev:
@raineorshine I've investigated a bit current functionality and looks like we need adjust current condition `globals.simulateDropHover || isHovering` based on sort is enabled or not.

raineorshine:
Thanks! I will need a bit more of a plan before proceeding to implementation.

- The current `isHovering` boolean is set to `monitor.isOver({ shallow: true }) && monitor.canDrop()`. How will the new condition be determined? i.e. How will you determine when to show the `drop-hover`, and how will you determine which `drop-hover` to show?
- There are three different `drop-hover` elements. Which ones will be used? Will they have the same conditions?

incora-dev:
- If sorting is disabled current `isHovering` variable will be used. Else `isHovering` should check if mouse is over parent (or root) thought and then check if draggable thought is alphabetically next after current thought.
- All three will be used. I think they should have the same conditions (at least very similar).

raineorshine:
Great, what is your time estimate?

> * If sorting is disabled current `isHovering` variable will be used. Else `isHovering` should check if mouse is over parent (or root) thought and then check if draggable thought is alphabetically next after current thought.

That sounds good. It will need to specifically exclude (return `false`) older ancestors, as they will technically be under the mouse cursor.

> * All three will be used. I think they should have the same conditions (at least very similar).

Okay, makes sense.

incora-dev:
@raineorshine I think it will take 3-4 hours.

PR #723, opened 2020-06-03
Corrected drop hover position for sorted lists

incora-dev:
#538

### DIFF
--- a/src/components/Thought.js
+++ b/src/components/Thought.js
@@ -59,6 +59,7 @@ import {
 // selectors
 import {
   attribute,
+  attributeEquals,
   chain,
   getNextRank,
   getRankBefore,
@@ -185,15 +186,19 @@ const canDrag = props => {
 }
 
 // eslint-disable-next-line jsdoc/require-jsdoc
-const beginDrag = props => {
+const beginDrag = ({ thoughtsRankedLive }) => {
   // disable hold-and-select on mobile
   if (isMobile) {
     setTimeout(() => {
       document.getSelection().removeAllRanges()
     })
   }
-  store.dispatch({ type: 'dragInProgress', value: true })
-  return { thoughtsRanked: props.thoughtsRankedLive }
+  store.dispatch({
+    type: 'dragInProgress',
+    value: true,
+    draggingThought: thoughtsRankedLive,
+  })
+  return { thoughtsRanked: thoughtsRankedLive }
 }
 
 // eslint-disable-next-line jsdoc/require-jsdoc
@@ -302,7 +307,8 @@ const drop = (props, monitor, component) => {
 // eslint-disable-next-line jsdoc/require-jsdoc
 const dropCollect = (connect, monitor) => ({
   dropTarget: connect.dropTarget(),
-  isHovering: monitor.isOver({ shallow: true }) && monitor.canDrop()
+  isHovering: monitor.isOver({ shallow: true }) && monitor.canDrop(),
+  isAnyChildHovering: monitor.isOver()
 })
 
 /**********************************************************************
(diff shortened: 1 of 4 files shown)


src/components/Thought.js
```diff
@@ -459,6 +468,17 @@ const ThoughtContainer = ({
...
 
+  const cursorOnAlphabeticalSort = cursor && attributeEquals(state, context, '=sort', 'Alphabetical')
+
+  const draggingThoughtPath = pathToContext(state.draggingThought)
+  const draggingThoughtValue = draggingThoughtPath && draggingThoughtPath[draggingThoughtPath.length - 1]
+
+  const shouldDisplayHover = cursorOnAlphabeticalSort
+    ? state.dragInProgress && isParentHovering
+      && draggingThoughtValue <= value
+      && (prevChild ? draggingThoughtValue > prevChild.value : true)
```
raineorshine, line 479: Simplify to: `&& (!prevChild || draggingThoughtValue > prevChild.value)`

src/components/Thought.js
```diff
@@ -459,6 +468,17 @@ const ThoughtContainer = ({
   const styleContainer = getStyle(state, thoughts, { container: true })
   const styleContainerZoom = isEditingPath ? getStyle(state, thoughts.concat('=focus', 'Zoom'), { container: true }) : null
 
+  const cursorOnAlphabeticalSort = cursor && attributeEquals(state, context, '=sort', 'Alphabetical')
+
+  const draggingThoughtPath = pathToContext(state.draggingThought)
+  const draggingThoughtValue = draggingThoughtPath && draggingThoughtPath[draggingThoughtPath.length - 1]
```
raineorshine, line 474: Use `contextOf` instead of indexing into `draggingThoughtPath` manually.

src/components/Thought.js
```diff
@@ -459,6 +468,17 @@ const ThoughtContainer = ({
   const styleContainer = getStyle(state, thoughts, { container: true })
   const styleContainerZoom = isEditingPath ? getStyle(state, thoughts.concat('=focus', 'Zoom'), { container: true }) : null
 
+  const cursorOnAlphabeticalSort = cursor && attributeEquals(state, context, '=sort', 'Alphabetical')
+
+  const draggingThoughtPath = pathToContext(state.draggingThought)
```
raineorshine, line 473: This should be called `draggingThoughtContext` since `pathToContext` returns type `Context`, not type `Path`.

src/components/Thought.js
```diff
@@ -459,6 +468,17 @@ const ThoughtContainer = ({
...
+  const draggingThoughtValue = draggingThoughtPath && draggingThoughtPath[draggingThoughtPath.length - 1]
+
+  const shouldDisplayHover = cursorOnAlphabeticalSort
```
raineorshine, line 476: A comment here explaining what this does would be great.

src/components/Thought.js
```diff
@@ -390,6 +397,8 @@ const ThoughtContainer = ({
   isEditing,
   isEditingPath,
   isHovering,
+  isParentHovering,
+  prevChild,
```
raineorshine, line 401: Adding `prevChild` is not ideal, but maybe there is no better way.

src/components/Thought.js
incora-dev: @raineorshine How can I use `contextOf` in this case? It returns all elements of array except the last but I need only the last.

raineorshine: Sorry, I meant `head`!

raineorshine:
One more note: Careful with your `package-lock.json`. For some reason your editor or `npm` changed a lot of the `resolved` fields to `false` which is not correct.

PR #734, opened 2020-06-08
Drag and pin to top/bottom of sorted

### LINKED ISSUE #537: Drag and pin to top/bottom of sorted
When a list is sorted, it would be useful to be able to drag thoughts to the top or bottom, removing them from the sort order. 

e.g. 

- My Sorted List
  - *To Do* [pinned top]
  - a
  - b
  - c
  - *Archive* [pinned bottom]

## Requirements

- [ ] Drag a sorted thought to the top or bottom of the context 
- [ ] Show the drop target correctly at the top or bottom 
- [ ] Drag a pinned thought back to the sorted sections
- [ ] Add some space in between the pinned and sorted sections. 
- [ ] All cursor navigation should take into account the hybrid order of thoughts. **This must be decoupled so that the cursor navigation logic does not know about the sorting algorithm.**

aqkhan:
Fixes #537


aqkhan:
Note: Cursor navigation needs to be fixed entirely for sorted context. Will push in the next commit.

aqkhan:
Resolving Drag/Drop issue that emerged after merging into dev.

raineorshine:
Please use reducers for pure, synchronous state changes rather than action creators. [reducerFlow](https://github.com/cybersemics/em/blob/dev/src/util/__tests__/reducerFlow.js) provides an easy way to compose reducers to create higher order functionality.

src/action-creators/index.js
```diff
@@ -18,6 +18,8 @@ export { default as preloadSources } from './preloadSources'
 export { default as setEditingValue } from './setEditingValue'
 export { default as setInvalidState } from './setInvalidState'
 export { default as updateSplitPosition } from './updateSplitPosition'
+export { default as pinToTop } from './pinToTop'
+export { default as pinToBottom } from './pinToBottom'
```
raineorshine, line 22: Let's combine these into a single, parameterized reducer. Possible input values may be `'top'`, `'bottom'`, or `null`. `null` can be used to remove pins.

src/action-creators/removePins.js
```diff
@@ -0,0 +1,24 @@
...
+  const state = getState()
+  const context = pathToContext(path)
+  const pinnedTop = hasChild(state, path, '=pinnedTop')
+  const pinnedBottom = hasChild(state, path, '=pinnedBottom')
+  if (pinnedTop) {
+    dispatch({ type: 'deleteAttribute', context, key: '=pinnedTop' })
+  }
+  if (pinnedBottom) {
+    dispatch({ type: 'deleteAttribute', context, key: '=pinnedBottom' })
+  }
```
raineorshine, line 23: Notice the similarity between these two blocks. Try to move conditionals as deep as possible to reduce redundancy.

src/components/Subthoughts.js
```diff
@@ -395,15 +409,33 @@ export const SubthoughtsComponent = ({
...
+    const value = showContexts ? head(child.context) : child.value
+    const childContext = pathToContext(unroot(thoughtsRanked)).concat(value)
+    return hasChild(state, childContext, '=pinnedTop')
+  })
+
+  const bottomPinned = children.filter(child => {
+    const value = showContexts ? head(child.context) : child.value
+    const childContext = pathToContext(unroot(thoughtsRanked)).concat(value)
+    return hasChild(state, childContext, '=pinnedBottom')
+  })
```
raineorshine, line 434: Notice how similar these blocks of code are. Try to abstract the similarities and parameterize the differences to reduce redundancy.

src/reducers/existingThoughtMove.js
```diff
@@ -41,6 +42,9 @@ export default (state, { oldPath, newPath, offset }) => {
   const newThought = removeDuplicatedContext(moveThought(oldThought, oldContext, newContext, oldRank, newRank), newContext)
   const isPathInCursor = subsetThoughts(state.cursor, oldPath)
 
+  console.log('oldThought', oldThought)
+  console.log('newThought', newThought)
```
raineorshine, line 46: Remove console.logs

aqkhan:
@raineorshine I am facing an issue while creating pin attributes for the thoughts dropped at top and bottom of the sorted context. The thoughts that are to be pinned are duplicated on the drop, can be seen if hidden thought are revealed (see attached screenshot) for reference. Also, I am using the following to create pin attributes. 

`dispatch({ type: 'newThought', at: path, insertNewSubthought: true, insertBefore: true, value: '=pinnedBottom', preventSetCursor: true })`

Need your input on this. Thanks!

Screenshot 2020-07-28 at 6 22 33 PM

raineorshine:
> @raineorshine I am facing an issue while creating pin attributes for the thoughts dropped at top and bottom of the sorted context. The thoughts that are to be pinned are duplicated on the drop, can be seen if hidden thought are revealed (see attached screenshot) for reference. Also, I am using the following to create pin attributes.
> 
> `dispatch({ type: 'newThought', at: path, insertNewSubthought: true, insertBefore: true, value: '=pinnedBottom', preventSetCursor: true })`
> 
> Need your input on this. Thanks!

Are they duplicated in `state` or just in `render`? i.e., when you refresh the page are they still duplicated?

You may want to use [`setAttribute`](https://github.com/cybersemics/em/blob/dev/src/reducers/setAttribute.ts) or [`toggleAttribute`](https://github.com/cybersemics/em/blob/dev/src/reducers/toggleAttribute.ts).

aqkhan:
> Are they duplicated in state or just in render? i.e., when you refresh the page are they still duplicated?

The duplicates seems to appear only in the render and are not persisted on page refresh. However, the pin attributes also disappear after page refresh. 

> You may want to use setAttribute or toggleAttribute.

Thanks. I'll take a look.

raineorshine:
> > Are they duplicated in state or just in render? i.e., when you refresh the page are they still duplicated?
> 
> The duplicates seems to appear only in the render and are not persisted on page refresh. 

Seems like you may be getting duplicate `ranks` somehow. `rank` must be unique as it is used to [generate the React key](https://github.com/cybersemics/em/blob/dev/src/components/Subthoughts.tsx#L554)

> However, the pin attributes also disappear after page refresh.

That's strange. Creating a new thought should trigger `updateThoughts` → `sync` and persist to `dexie`.

raineorshine:
Alternative: #900

PR #735, opened 2020-06-09
Fixes regressions caused in cursorDown due to util functions' refactoring

### LINKED ISSUE #658: cursorDown regression
#640 may have broken the new `cursorDown` from #535.

anmolarora1:
Fixes #658 and adds unit tests for `cursorDown` reducer

### DIFF
--- a/src/reducers/__tests__/cursorDown.js
+++ b/src/reducers/__tests__/cursorDown.js
@@ -1,6 +1,7 @@
-import { initialState, reducerFlow } from '../../util'
+import { initialState, pathToContext, reducerFlow } from '../../util'
 import { NOOP, RANKED_ROOT } from '../../constants'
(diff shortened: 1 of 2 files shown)


raineorshine:
Since the input text is defined using a template literal string, the indentation is preserved. When the `describe` block was added, everything was indented, including the template literals. I fixed them, but it did reveal that one of the tests is broken, likely due to an error in `importText`.

Also, I recommend defining the text explicitly in each test block since it varies in some of them. `beforeEach` should be reserved for setup that is not test specific. Defining a default and overriding it in certain cases is less explicit and more error prone than defining the variable explicitly in each test.

raineorshine:
Fixed the tests! I used `rankThoughtsFirstMatch` to make the tests more readable and decouple rank, which can be tested separately.

Issue #668, opened 2020-05-18
JSDOC documentation


raineorshine:
Try these and compare:

- https://github.com/documentationjs/documentation
- https://github.com/clenemt/docdash
- https://github.com/nijikokun/minami

pushkov-fedor:
I've tried these today. docdash and minami - just theme templates for jsdoc generator output (https://github.com/jsdoc/jsdoc).
I didn't find configuration, but looks like documentation.js, out-of-the-box, doesn't work well with arrow function. Here's screenshots for isLeftSpaceClick function defined as arrow function, and as standard function
![Screenshot from 2020-05-19 20-37-37](https://user-images.githubusercontent.com/32064251/82336730-212fe880-9a15-11ea-90a0-66d47e752d81.png)
![Screenshot from 2020-05-19 20-38-03](https://user-images.githubusercontent.com/32064251/82336745-255c0600-9a15-11ea-9ecb-d8394c548e05.png)
Bellow screenshots for standard jsdoc theme, minami and docdash
![Screenshot from 2020-05-19 20-41-27](https://user-images.githubusercontent.com/32064251/82336906-51778700-9a15-11ea-93fc-0f0f5817734f.png)
![Screenshot from 2020-05-19 20-42-13](https://user-images.githubusercontent.com/32064251/82336911-52101d80-9a15-11ea-99a3-938769bbee0e.png)
![Screenshot from 2020-05-19 20-42-31](https://user-images.githubusercontent.com/32064251/82336913-52a8b400-9a15-11ea-8918-635ec52363ab.png)

As you see, jsdoc also pastes source links and generates files for them
minami/docdash
![Screenshot from 2020-05-19 21-16-57](https://user-images.githubusercontent.com/32064251/82337642-35281a00-9a16-11ea-9f3f-fec9ddcd9231.png)
![Screenshot from 2020-05-19 21-17-11](https://user-images.githubusercontent.com/32064251/82337644-36594700-9a16-11ea-8c48-72383bc774ff.png)

What do you think? Maybe you have certain vision of tools and functions that doc generators should have?

raineorshine:
> I've tried these today. docdash and minami - just theme templates for jsdoc generator output ([jsdoc/jsdoc](https://github.com/jsdoc/jsdoc)).

Great!

> I didn't find configuration, but looks like documentation.js, out-of-the-box, doesn't work well with arrow function. Here's screenshots for isLeftSpaceClick function defined as arrow function, and as standard function.

[...]

PR #739, opened 2020-06-11
TypeDoc documentation

pushkov-fedor:
Issue #668 

I think here's a starting point for this epic :)
We need **typedoc** as dev dependency for generating documentation and **typedoc-plugin-external-module-name** for grouping files in any way we want. I added npm script "docs" so don't forget to update node_modules before running it.

TypeDoc generates 2 top level groups: internal and external. Here's explanation of what it means:

> Typedoc categorizes your code into "Internal" and "External". Essentially:
> 
> Internal: Your own code
> External: Everything else
> When using the default theme, typedoc provides a checkbox to show/hide the generated "External" documentation.
> 
> Typedoc uses the files: [] array (in tsconfig.json) to determine if code is "Internal". If a file being parsed is in the files: [] array, then the code in that file is "Internal".

So i added js files to tsconfig. Without it TypeDoc defines all TypeScript files as Internal and all JS files as External. We can handle this issue with another plugin - typedoc-plugin-internal-external. It uses `@internal` and `@external` tags.

I grouped all files by directories, but we can group them in any way we want. **typedoc-plugin-external-module-name** uses **@packageDocumentation** and **@module** tags.
Here's explanation of what each tag means:
`You can explicitly specify a Module name using the @module annotation. Add a comment block at the top of a Typescript file with @module modulename. Mark the comment block as @packageDocumentation to let typedoc know that this is documentation for the file (Module) itself`
For example, i grouped all `tutorial*` files in reducers directory in a single module.

We need to allow @packageDocumentation tag for eslint checking, because now there's a lot of errors in project

I replaced all `export default () => {}` with 
`const fun = () => {}; export default fun; `

We can't document destructured params like this: 
```
/**
 * Dispatches an alert action.
 *
 * @param value The string or React Component that will be rendered in the alert.
 * @param showCloseLink Show a small 'x' in the upper right corner that allows the user to close the alert. Default: true.
 * @param type An arbitrary alert type that can be added to the alert. This is useful if specific alerts needs to be detected later on, for example, to determine if the alert should be closed, or if it has been superceded by a different alert type.
 */
export default (value, { showCloseLink, alertType } = {}) => {
```
In last commit i added example of how to do this. But i think i can google for different solutions in order to place all comment above the function.

Please generate docs, test it and tell me what you think about it :)


raineorshine:
Fantastic! This is awesome.

> We need **typedoc** as dev dependency for generating documentation and **typedoc-plugin-external-module-name** for grouping files in any way we want. I added npm script "docs" so don't forget to update node_modules before running it.

Great. I removed "npx" since npm scripts automatically use the binaries in `node_modules`.

> So i added js files to tsconfig. Without it TypeDoc defines all TypeScript files as Internal and all JS files as External. We can handle this issue with another plugin - typedoc-plugin-internal-external. It uses `@internal` and `@external` tags.

Explicitly specifying `@internal` and `@external` is extra work without any real benefit. The distinction between internal and external is meaningless in our project. So whatever we can do to automate this and strip out the distinction would be great.

> I grouped all files by directories, but we can group them in any way we want. **typedoc-plugin-external-module-name** uses **@packageDocumentation** and **@module** tags.

Okay, great. I will think about how to organize them.

> We need to allow @packageDocumentation tag for eslint checking, because now there's a lot of errors in project

Yes, this is high priority since the linter is failing. Can you work on this if you are not already?

> I replaced all `export default () => {}` with
> `const fun = () => {}; export default fun; `

Perfect

> 
> We can't document destructured params
> In last commit i added example of how to do this. But i think i can google for different solutions in order to place all comment above the function.

I like what you did in 96e926b7d21ff70bca09e68b5fefb318e54909d4. That works for me. Hopefully we can change the output from "__namedParameters" to "options".

> Please generate docs, test it and tell me what you think about it :)

Looks great! Here are some desired changes:

- [ ] Render options output as  "options" instead of "__namedParameters" for options objects (96e926b7d21ff70bca09e68b5fefb318e54909d4)
- [ ] Does `/docs` need to be in version control for GitHub pages to work, or is it possible to run it in a build process?
- [ ] Is there a way to specify the docs home page? I think we need a different message than `README.md`. We should provide a summary and instructions that are appropriate for someone looking for API docs.
- [ ] Remove `<em>Module</em>` from sidebar.
- [ ] Remove "Globals" namespace. action-creators, components, hooks, etc should be in the root.
- [ ] How do we extend the default theme? I can see there will be some things we want to change.

src/action-creators/ModuleInfoDocs.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Here's documentation for all action-creators.
+ *
+ * @packageDocumentation
+ * @module action-creators
+ * @preferred
+ */
```
raineorshine, line 7: What's the purpose of this file, and why is it a js file? I don't like having dummy files for the docs.

src/action-creators/alert.js
```diff
@@ -1,13 +1,21 @@
+/**
+ * @packageDocumentation
```
raineorshine, line 2: Why is `@packageDocumention` needed here?

src/action-creators/alert.js
```diff
@@ -1,13 +1,21 @@
...
+ * @module action-creators.alert
```
raineorshine, line 3: Is there a way to infer the `@module` from the directory structure? I don't like the burden of having to explicitly mark up every file when it can be generated automatically.

typedoc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "inputFiles": ["src/*"],
+    "exclude": ["**/index.js", "src/*/*/*", "src/globals.js", "src/registerServiceWorker.js"],
+    "mode": "modules",
+    "out": "docs"
+}
```
raineorshine, line 6: Do you make sure you have the [EditorConfig](https://editorconfig.org/) plugin set up correctly in your editor? The [`.editorconfig`](https://github.com/cybersemics/em/blob/dev/.editorconfig) should ensure that all files end with a newline character.

raineorshine:
I [forked typedoc-default-themes](https://github.com/cybersemics/em-typedoc-theme) and added it as a dependency in 29acef9.

pushkov-fedor:
> Is there a way to specify the docs home page? I think we need a different message than README.md. We should provide a summary and instructions that are appropriate for someone looking for API docs.

See 0b12c01

pushkov-fedor:
> Does /docs need to be in version control for GitHub pages to work, or is it possible to run it in a build process?

As i understand, GitHub pages use static files from your repository to host them so we need to have **_docs_** folder in dev branch.

Screenshot from 2020-06-15 16-10-10

pushkov-fedor:
> Render options output as "options" instead of "__namedParameters" for options objects

Yes, i don't like current "__namedParameters" too. I created an issue for this case in typedoc repo, let's wait a little.

pushkov-fedor:
>  Remove <em>Module</em> from sidebar.

I'm not sure what's happened but it's removed. Just re-installed node_modules and ran typedoc with em theme. Do you have the same?
Screenshot from 2020-06-15 17-04-45

raineorshine:
> > Remove _Module_ from sidebar.
> 
> I'm not sure what's happened but it's removed. Just re-installed node_modules and ran typedoc with em theme. Do you have the same?

Yeah, same for me!

raineorshine:
> > Render options output as "options" instead of "__namedParameters" for options objects
> 
> Yes, i don't like current "__namedParameters" too. I created an issue for this case in typedoc repo, let's wait a little.

Sounds good. Actually, this issue will probably be eliminated once we convert more code to Typescript, because the `Options` type is usually specified:

https://github.com/cybersemics/em/blob/93b2e3bf1a471c338b68fd96320da22ca6b1e2d9/src/action-creators/suppressExpansion.ts#L6-L13

And when there are multiple Options types, they can be named:

https://github.com/cybersemics/em/blob/4a41328b99cdb2e05ac81fa2b6efa5466e71e5e1/src/util/importHTML.ts#L36-L49

This should result in a clean output, and the code seems cleaner, too. What do you think?

pushkov-fedor:
> > > Render options output as "options" instead of "__namedParameters" for options objects
> > 
> > 
> > Yes, i don't like current "__namedParameters" too. I created an issue for this case in typedoc repo, let's wait a little.
> 
> Sounds good. Actually, this issue will probably be eliminated once we convert more code to Typescript, because the `Options` type is usually specified:
> 
> https://github.com/cybersemics/em/blob/93b2e3bf1a471c338b68fd96320da22ca6b1e2d9/src/action-creators/suppressExpansion.ts#L6-L13
> 
> And when there are multiple Options types, they can be named:
> 
> https://github.com/cybersemics/em/blob/4a41328b99cdb2e05ac81fa2b6efa5466e71e5e1/src/util/importHTML.ts#L36-L49
> 
> This should result in a clean output, and the code seems cleaner, too. What do you think?

Actually right now it doesn't work in a way we want 
Screenshot from 2020-06-17 20-47-33
Screenshot from 2020-06-17 20-46-11

src/action-creators/ModuleInfoDocs.js
pushkov-fedor: I've added these files as illustration that we are able to add special description for each module.
Text from these files is displayed above the list of functions inside a module.
Screenshot from 2020-06-17 21-04-13


We can remove it, if we doesn't need it


Screenshot from 2020-06-17 21-03-49

Or we can move such files to another directory. See a5bcdc3

Also typedoc supports markdown, so we can write something like this:


Screenshot from 2020-06-17 21-13-26

src/action-creators/alert.js
pushkov-fedor: > Mark the comment block as @packageDocumentation to let typedoc know that this is documentation for the file (Module) itself 

I removed all `@packageDocumentation` tags from action-creators for testing and typedoc mixed almost all files together.
Screenshot from 2020-06-17 21-55-17
It defines expandContextThought, setEditingValue, setInvalidState ... as modules because these files don't import anything.

raineorshine:
> Actually right now it doesn't work in a way we want

I see. Yeah, that's subpar.

> > Render options output as "options" instead of "__namedParameters" for options objects
> 
> Yes, i don't like current "__namedParameters" too. I created an issue for this case in typedoc repo, let's wait a little.

Will you post a link to the issue here?

Do you think you could add the functionality on their end and issue a PR?

pushkov-fedor:
> > Actually right now it doesn't work in a way we want
> 
> I see. Yeah, that's subpar.
> 
> > > Render options output as "options" instead of "__namedParameters" for options objects
> > 
> > 
> > Yes, i don't like current "__namedParameters" too. I created an issue for this case in typedoc repo, let's wait a little.
> 
> Will you post a link to the issue here?
> 
> Do you think you could add the functionality on their end and issue a PR?

Sure! https://github.com/TypeStrong/typedoc/issues/1321
But i'm not sure about this issue now. I asked about replacing "__namedParameters" because this default name looks a bit ugly. For solving this small issue we can add replacement script and substitute all default `__namedParameters` strings with default `options` string or with something else. 

If we use destructuring for 2 params, output will look like this:
Screenshot from 2020-06-18 13-27-14
After script execution:
Screenshot from 2020-06-18 13-30-00
Output looks prettier, but the issue still isn't solved.

And i don't like such approach for TypeScript files. I would like to see, for such function `const foo = ({ cancel, duration }: Options): void => {}`, generated output like this:

**Parameters**:
- **Options**
  - **cancel**: boolean
  - **duration**: number

What do you think? 
I would like to work on it and add such functionality, but i'm not sure about how much time it could take. Let's investigate a little bit more, maybe solution for this already exists.

pushkov-fedor:
Check f918e20

src/action-creators/alert.js
pushkov-fedor: Yes, there's another way to do this, we can write mapping function. See c806a04

pushkov-fedor:
I've added `packageDocumentation` and `preferred` tags in "definedTags" for eslint (f3b4a07, b30eaae), but compiler still fails on `preferred`

raineorshine:
> Sure! [TypeStrong/typedoc#1321](https://github.com/TypeStrong/typedoc/issues/1321)
> But i'm not sure about this issue now. I asked about replacing "__namedParameters" because this default name looks a bit ugly. For solving this small issue we can add replacement script and substitute all default `__namedParameters` strings with default `options` string or with something else.

Thanks! Eager to hear what they say.

Yes, the replacement script would work. There are two advantages to making the change in TypeDoc though:

1. Reduces complexity in **em**
2. Contributes back to Open Source

Arguably it may be worth forking Typedoc anyway to achieve (1). I think it would be a fun task, but maybe that's just me :). Let me know what you're up for.

> Output looks prettier, but the issue still isn't solved.

Sorry, which issue exactly?


>
> And i don't like such approach for TypeScript files. I would like to see, for such function `const foo = ({ cancel, duration }: Options): void => {}`, generated output like this:
> 
> **Parameters**:
> 
> * **Options** (with link)
>   
>   * **cancel**: boolean
>   * **duration**: number
> 
> What do you think?
> I would like to work on it and add such functionality, but i'm not sure about how much time it could take. Let's investigate a little bit more, maybe solution for this already exists.

Nice! I like the looks of that. It's probably too low priority to allocate time to right now, but let's keep it in mind. You are welcome to add an issue for it so it's tracked. 

Modifying TypeDoc to rename `__namedParameters` is sufficient for now.

src/action-creators/ModuleInfoDocs.js
raineorshine: > I've added these files as illustration that we are able to add special description for each module.

Yes, I think that would be useful to have a description for each module. It's just that storing it in an arbitrary js file that is only used for documentation seems less than ideal.

> Or we can move such files to another directory.

Good to know, although I think this is worse, as they should definitely be colocated.

raineorshine: The logical place for the module description to go would be the `index.js` file in each module folder, however TypeDoc didn't seem to pick up the package documentation there when I added it. Can you look into this to see if it's possible?

src/action-creators/alert.js
raineorshine: Well, we obviously don't want that. Do you see any other solutions? The point of inline documentation is to automate as much as possible from JSDOC comments and folder structure. It's not great to have to specify something explicitly that could be determined automatically.

src/action-creators/alert.js
raineorshine: Great! So we can remove `@module`, but `@packageDocumentation` is still required?

src/action-creators/alert.js
raineorshine: For inner modules with single functions (e.g. `alert`) I kind of like it showing up on the main module page, but including all the internal variables and functions is unacceptable. We probably need the dedicated module page for each inner module to accommodate internal variables.

typedoc.json
raineorshine: @pushkov-fedor

src/action-creators/alert.js
pushkov-fedor: I would like to clarify what you meant.
Here's how submodules are displayed on the main module page. 

Screenshot from 2020-06-19 15-26-03

Internal variables and functions are placed inside submodule

Screenshot from 2020-06-19 15-27-33

Is it correct or do you want to show only functions having JSDoc comment without any additional functions/variables?

src/action-creators/ModuleInfoDocs.js
pushkov-fedor: Yes, you are right! TypeDoc didn't pick index.js files because i excluded them in config file :)
I did it because TypeDoc creates References for re-exported files.
Screenshot from 2020-06-19 16-36-41
I added `typedoc-plugin-exclude-references` to fix this, now any description could be placed in index.js files

Commit: 395942c

typedoc.json
pushkov-fedor: Sorry for not responding, this page is getting massive and messy. 
I forgot to enable plugin in vscode, now it works :)

pushkov-fedor:
> I'm getting `sed: 1: "./docs/index.html": invalid command code .` on the sed command. This won't work cross-platform anyway. We still want Windows developers to be able to run the project locally, so we should have a node-based cross platform solution.
> 
> I vote for a TypeDoc fork, which eschews the substitution script.

Okay, it would be interesting to work on fix for TypeDoc :)

pushkov-fedor:
> Sorry, which issue exactly?

I meant that we have 2 issues about typedoc output:

1. hardcoded __namedParameters
2. desired output with destructured params and interface

I was talking about the second one

pushkov-fedor:
I didn't find eslint-plugin or configuration that forces `@packageDocumentation` to be on top of the file. Maybe we should write custom plugin for it?

pushkov-fedor:
No compile errors now

src/action-creators/ModuleInfoDocs.js
raineorshine: Excellent!

src/action-creators/alert.js
raineorshine: That's correct, thanks!

raineorshine:
Please rebase or merge `dev`.

raineorshine:
> I didn't find eslint-plugin or configuration that forces `@packageDocumentation` to be on top of the file. Maybe we should write custom plugin for it?

Another idea: What if added a flag to TypeDoc or write a plugin to act "as if" `@packageDocumentation` was set at the top of every file. Would that work, or are there files we *don't* want to annotate with that?

If that was doable, we wouldn't have to explicitly add the annotation to every file and we wouldn't have to write a new linter rule.

.gitignore
```diff
@@ -1,13 +1,9 @@
...
 
-# dependencies
 /node_modules
-
-# testing
 /coverage
-
-# production
 /build
+/docs
```
raineorshine, line 6: If GitHub pages requires `/docs`, then we can't ignore it, right?

typedoc.json
```diff
@@ -0,0 +1,10 @@
+{
+  "inputFiles": ["src/*"],
+  "exclude": ["src/*/*/*", "src/globals.js", "src/registerServiceWorker.js"],
```
raineorshine, line 3: What is `src/*/*/*`?

Can we include `globals.js`? My previous comment about removing `Globals` was the namespaces that it added to everything, not the globals file and its exports.

.gitignore
pushkov-fedor: Yes!

typedoc.json
pushkov-fedor: I used `src/*/*/*` to exclude all files in second level subdirectories of src (e.g. components/Tutorial, components/__tests__).  Are they necessary for documentation?

pushkov-fedor:
> > I didn't find eslint-plugin or configuration that forces `@packageDocumentation` to be on top of the file. Maybe we should write custom plugin for it?
> 
> Another idea: What if added a flag to TypeDoc or write a plugin to act "as if" `@packageDocumentation` was set at the top of every file. Would that work, or are there files we _don't_ want to annotate with that?
> 
> If that was doable, we wouldn't have to explicitly add the annotation to every file and we wouldn't have to write a new linter rule.

It's a great idea! I think i need to take a deeper look at TypeDoc' plugin development. If we solve both tasks (automatic file annotation with @packageDocumentation and renaming __namedParams) with plugin, it will be amazing.

typedoc.json
raineorshine: Let's exclude `tests`, but document the rest.

raineorshine:
Note: https://github.com/cybersemics/em/pull/739#discussion_r443670663

> It's a great idea! I think i need to take a deeper look at TypeDoc' plugin development. 

Wonderful! You're doing great work. Let's continue this direction. I think this might be the final thing to do.

> If we solve both tasks (automatic file annotation with @packageDocumentation and renaming __namedParams) with plugin, it will be amazing.

Agreed. Two independent plugins, to be clear. If you'd like, you can create repos for them and then transfer them to [cybersemics](https://github.com/cybersemics).

pushkov-fedor:
I temporary removed /docs from .gitignore. Such a massive amount of files was a little noisy :)

I added plugin for renaming __namedParameters (https://github.com/pushkov-fedor/typedoc-plugin-rename-named-parameters)

pushkov-fedor:
Also i noticed, that many files defines by TypeDoc as modules. I think the reason of such behavior could be transition from js to ts. I will test this assumption and reply you about it in coming days.

raineorshine:
> I temporary removed /docs from .gitignore. Such a massive amount of files was a little noisy :)

Yeah, let me know if you have any better ideas about where to store them that will still work with GitHub pages.

> I added plugin for renaming __namedParameters ([pushkov-fedor/typedoc-plugin-rename-named-parameters](https://github.com/pushkov-fedor/typedoc-plugin-rename-named-parameters))

Awesome! It looks great!

Will you transfer to [cybersemics](https://github.com/cybersemics) when you get a chance?

pushkov-fedor:
> > I temporary removed /docs from .gitignore. Such a massive amount of files was a little noisy :)
> 
> Yeah, let me know if you have any better ideas about where to store them that will still work with GitHub pages.
> 
> > I added plugin for renaming __namedParameters ([pushkov-fedor/typedoc-plugin-rename-named-parameters](https://github.com/pushkov-fedor/typedoc-plugin-rename-named-parameters))
> 
> Awesome! It looks great!
> 
> Will you transfer to [cybersemics](https://github.com/cybersemics) when you get a chance?

Yes! Could you explain please what needs from me to make the transfer?

raineorshine:
Please follow the instructions here: https://help.github.com/en/github/administering-a-repository/transferring-a-repository

raineorshine:
Hey! I have a couple good solutions for deploying the docs without adding the output to version control. 

1. Deploy to GitHub Pages Action 
https://github.com/marketplace/actions/deploy-to-github-pages

2. Fleek can publish static sites using IPFS and automatically deploy using GitHub webhook. https://fleek.co

raineorshine:
Hey Fedor! Just noticed that even though `/docs` has been removed it's still taking up a lot of space in the commit history: f0dd8a031bca19b1dac2ec7333bf414f1e5325f1. 

Please rebase and modify/omit that commit (and any other commits that add `/docs`) before we merge into `dev`.

pushkov-fedor:
> Hey Fedor! Just noticed that even though `/docs` has been removed it's still taking up a lot of space in the commit history: [f0dd8a0](https://github.com/cybersemics/em/commit/f0dd8a031bca19b1dac2ec7333bf414f1e5325f1).
> 
> Please rebase and modify/omit that commit (and any other commits that add `/docs`) before we merge into `dev`.

Okay, i will fix it!

pushkov-fedor:
> Hey! I have a couple good solutions for deploying the docs without adding the output to version control.
> 
> 1. Deploy to GitHub Pages Action
>    https://github.com/marketplace/actions/deploy-to-github-pages
> 2. Fleek can publish static sites using IPFS and automatically deploy using GitHub webhook.
>    https://fleek.co

I will take a look and reply you about them on the next week as soon as possible

pushkov-fedor:
> Also i noticed, that many files defines by TypeDoc as modules. I think the reason of such behavior could be transition from js to ts. I will test this assumption and reply you about it in coming days.

The reason of such behavior is our custom mapping function. I checked source code of TypeDoc, and it looks like they use @packageDocumentation tag for displaying module-wide comments (e.g. comments which we store in index files to describe the module as a whole). After removing @packageDocumentation tag we lose such description functionality. So i wrote plugin, which allows without @packageDocumentation tag to display module wide comments.

Here's plugin: https://github.com/pushkov-fedor/typedoc-plugin-module-wide-comments

I'm ready to transfer it and previous one to em.

The thing, that i'm worried about, is displaying <  em>Module<  /em> near other modules.
Screenshot from 2020-07-03 21-32-13

I think that we have some syntax construction in em, that are parsed incorrectly. I offer to write another special plugin for em and fix this issue there.

raineorshine:
> > Also i noticed, that many files defines by TypeDoc as modules. I think the reason of such behavior could be transition from js to ts. I will test this assumption and reply you about it in coming days.
> 
> The reason of such behavior is our custom mapping function. I checked source code of TypeDoc, and it looks like they use @packageDocumentation tag for displaying module-wide comments (e.g. comments which we store in index files to describe the module as a whole). After removing @packageDocumentation tag we lose such description functionality. So i wrote plugin, which allows without @packageDocumentation tag to display module wide comments.

Great work! I don't completely understand, but I'm sure it will make more sense to me when I look at the code. It seems like you have done a good job tracking down the solution and writing a plugin to solve it.

> 
> Here's plugin: [pushkov-fedor/typedoc-plugin-module-wide-comments](https://github.com/pushkov-fedor/typedoc-plugin-module-wide-comments)

Could you add descriptive comments to `index.ts` and a `README.md`?

> The thing, that i'm worried about, is displaying < em>Module< /em> near other modules.
> Screenshot from 2020-07-03 21-32-13
> 
> I think that we have some syntax construction in em, that are parsed incorrectly. I offer to write another special plugin for em and fix this issue there.

What does "syntax construction" mean? What exactly would the plugin do? I want to make sure the plugin has a clearly definable purpose and is well-justified. Otherwise we should just look for a workaround. Our other plugins were clearly defined and potentially reusable.

We could also modify the theme to not render `<em>Module< /em>`, but maybe that is too much of a hack. Open to your thoughts.

pushkov-fedor:
> Could you add descriptive comments to index.ts and a README.md?

Yes

> What does "syntax construction" mean? What exactly would the plugin do?
It seems to me that somewhere in em is some part of code that is parsed by TypeDoc incorrectly and defined like module (near components, reducers etc). I plan to find out where this code belongs to and why it renders this way. Then we can fix this part of code in em or write plugin that exclude this module. Plugin wouldn't be reusable, i agree, but it allows us not to touch em codebase. 
I don't think that we can fix out issue by editing theme, because < em> Module< /em> isn't hardcoded but generated automatically after resolving the project.

Let's find out what part of code is causing this and then decide our next step.

.typedoc-plugin-external-module-name.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Custom mapping function for documentation modules.
```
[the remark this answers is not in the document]
pushkov-fedor: No problem!

pushkov-fedor:
I've tried "Deploy to GitHub Pages Action" and it works!

Here's the configuration file: https://github.com/pushkov-fedor/documentation-test/blob/master/.github/workflows/blank.yml

The difference between my config file and config from action documentation is NoJekyll step. GitHub Pages could be integrated with Jekyll - static site generator. Jekyll ignores files which are started with _ (e.g. https://github.com/pushkov-fedor/documentation-test/blob/gh-pages/modules/_index_.html) TypeDoc generates such files so we need to add .nojekyll file in order to explicitly tell GitHub that we don't use Jekyll.

Also we can hack our issue about` < em> Module < /em>`. We can run bash script after NoJekyll step and before deployment in order to remove this link from generated output.

pushkov-fedor:
So, the problem was in src/index.js file, i just excluded it from typedoc.json.

pushkov-fedor:
I added README.md for https://github.com/pushkov-fedor/typedoc-plugin-module-wide-comments

raineorshine:
> I've tried "Deploy to GitHub Pages Action" and it works!

Great!

> 
> Here's the configuration file: [pushkov-fedor/documentation-test:.github/workflows/blank.yml@`master`](https://github.com/pushkov-fedor/documentation-test/blob/master/.github/workflows/blank.yml?rgh-link-date=2020-07-06T09%3A33%3A39Z)

Makes sense

> 
> The difference between my config file and config from action documentation is NoJekyll step. GitHub Pages could be integrated with Jekyll - static site generator. Jekyll ignores files which are started with _ (e.g. [pushkov-fedor/documentation-test:modules/_index_.html@`gh-pages`](https://github.com/pushkov-fedor/documentation-test/blob/gh-pages/modules/_index_.html?rgh-link-date=2020-07-06T09%3A33%3A39Z)) TypeDoc generates such files so we need to add .nojekyll file in order to explicitly tell GitHub that we don't use Jekyll.

Understood. Thanks for the explanation!

> 
> Also we can hack our issue about` < em> Module < /em>`. We can run bash script after NoJekyll step and before deployment in order to remove this link from generated output.

We need to maintain support for Windows, so a bash script won't work.

raineorshine:
> I added README.md for [pushkov-fedor/typedoc-plugin-module-wide-comments](https://github.com/pushkov-fedor/typedoc-plugin-module-wide-comments)

Thank you. The README file looks good. Would you also comment the `index` file? We want to provide clear comments so that future developers (including ourselves) can quickly understand why it exists and what it does.

pushkov-fedor:
> > I added README.md for [pushkov-fedor/typedoc-plugin-module-wide-comments](https://github.com/pushkov-fedor/typedoc-plugin-module-wide-comments)
> 
> Thank you. The README file looks good. Would you also comment the `index` file? We want to provide clear comments so that future developers (including ourselves) can quickly understand why it exists and what it does.
> 
> Once that's done feel free to transfer it to cybersemics.

Yes, will do it!

pushkov-fedor:
> > Also we can hack our issue about < em> Module < /em>. We can run bash script after NoJekyll step and before deployment in order to remove this link from generated output
> 
> We need to maintain support for Windows, so a bash script won't work.

We don't need it now, because this issue is fixed.
But i meant that if we do some manipulations via github actions (e.g. building documentation), as i understand, we can specify os of runner in yml file. 
For example:
```
build-and-deploy:
    runs-on: ubuntu-latest
```
And then we can use scripts. It was just for your information, because we don't need it right now

raineorshine:
I see. Thank you.

pushkov-fedor:
I've fixed lint errors and added comments to https://github.com/pushkov-fedor/typedoc-plugin-module-wide-comments

raineorshine:
Thanks! Just to confirm, did you remove all traces of `/docs` from the commit history?

raineorshine:
Will you be removing `<em>Module</em>`?

.eslintrc.json
```diff
@@ -168,7 +174,12 @@
     "jsdoc/check-alignment": 2,
     "jsdoc/check-indentation": 2,
     "jsdoc/check-syntax": 2,
-    "jsdoc/check-tag-names": 2,
+    "jsdoc/check-tag-names": [
+      2,
+      {
+        "definedTags": ["packageDocumentation"]
+      }
+    ],
```
raineorshine, line 182: You should be able to remove this since we don't use `packageDocumentation` any more.

package.json
```diff
@@ -91,13 +93,19 @@
     "jest-environment-jsdom-sixteen": "^1.0.3",
     "jest-localstorage-mock": "^2.4.0",
     "md5": "^2.2.1",
-    "raw.macro": "^0.3.1",
+    "raw.macro": "^0.3.0",
```
raineorshine, line 96: `raw.macro` got reverted

src/browser.ts
```diff
@@ -1,4 +1,7 @@
-/** Defines client/browser-specific variables that do not change over the course of a session. */
+/**
+ * Defines client/browser-specific variables that do not change over the course of a session.
+ *
+ */
```
raineorshine, line 4: Out of curiosity, is the expanded comment format required for the docs to render correctly, or is it just preference?

pushkov-fedor:
> Thanks! Just to confirm, did you remove all traces of `/docs` from the commit history?

Not yet

> Will you be removing `<em>Module</em>`?

It's removed because i excluded src/index.js from input

src/browser.ts
pushkov-fedor: I haven't yet cleaned all files after removing @packageDocumentation. We can write comments in one line.

pushkov-fedor:
@raineorshine did you update em-typedoc-theme on npm after approving my PR?

pushkov-fedor:
> Thanks!  Just to confirm, did you remove all traces of /docs from the commit history?

I squashed commits where i added and removed `/docs` folder, history should be clean now.

raineorshine:
> @raineorshine did you update **em-typedoc-theme** on npm after approving my PR?

Thanks for the reminder! Will do so now.

> > Thanks!  Just to confirm, did you remove all traces of /docs from the commit history?
> 
> I squashed commits where i added and removed `/docs` folder, history should be clean now.

Great, thank you.

pushkov-fedor:
Thanks for updating the theme! I've tested it today, everything works. "Globals" doesn't displayed, `<em>Module</em>` is removed, we don't need @packageDocumentation tag, **__namedParameters** is replaced, git history is cleaned. You can add GitHub Action for auto-deployment the documentation to GithubPages. Is there anything else that we have to do?

raineorshine:
> Thanks for updating the theme! I've tested it today, everything works. "Globals" doesn't displayed, `<em>Module</em>` is removed, we don't need @packageDocumentation tag, **__namedParameters** is replaced, git history is cleaned. You can add GitHub Action for auto-deployment the documentation to GithubPages. Is there anything else that we have to do?

Beautiful! No, I think we're good to go! Thanks so much for your work!

pushkov-fedor:
Thank you for opportunity to work on such wide and interesting task!

PR #741, opened 2020-06-12
Adds check for context view in cursorUp

### LINKED ISSUE #736: cursorUp: move cursor from context's first child to parent
Test: 900d4e91f0938dd3b8e831b35e65d9b690fa5bc0
Branch: `736`

anmolarora1:
Fixes #736

### DIFF
--- a/src/reducers/__tests__/cursorUp.js
+++ b/src/reducers/__tests__/cursorUp.js
@@ -1,9 +1,8 @@
-import { initialState, reducerFlow } from '../../util'
-
-// reducers
-import newThought from '../newThought'
-import cursorUp from '../cursorUp'
-import setCursor from '../setCursor'
+import { initialState, pathToContext, reducerFlow } from '../../util'
+import { NOOP, RANKED_ROOT } from '../../constants'
+import { importText } from '../../action-creators'
+import { cursorUp, newThought, setCursor, updateThoughts, toggleContextView } from '../../reducers'
+import { rankThoughtsFirstMatch } from '../../selectors'
 
 it('move cursor to previous sibling', () => {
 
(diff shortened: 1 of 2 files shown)


src/selectors/prevSibling.js
```diff
@@ -1,34 +1,50 @@
...
  *
  * @param context   Can be a context or path.
  */
 export default (state, value, context, rank) => {
+
   const { showHiddenThoughts } = state
   const sortPreference = getSortPreference(state, context)
-  const siblings = (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(state, context)
+
+  const contextViewActive = perma(() => isContextViewActive(state, context))
```
raineorshine, line 14: Since `contextViewActive` will always be used, you can use a normal variable:

```js
const contextViewActive = isContextViewActive(state, context)
```

`perma` is only needed if there is at least one case where it does not get used it all, in which case the function wrapper emulates a lazy-evaluated variable.

src/selectors/prevSibling.js
```diff
@@ -1,34 +1,50 @@
...
+
+  /** Gets siblings of a context. */
+  const getContextSiblings = () => getContextsSortedAndRanked(state, head(context))
+
+  /** Gets siblings of thought. */
+  const getThoughtSiblings = () => (sortPreference === 'Alphabetical' ? getThoughtsSorted : getThoughtsRanked)(state, context)
```
raineorshine, line 20: Great, this is clear.

src/selectors/prevSibling.js
```diff
@@ -1,34 +1,50 @@
...
+
+  const siblings = contextViewActive() ? getContextSiblings() : getThoughtSiblings()
   let prev// eslint-disable-line fp/no-let
 
+  /** Returns true when context is not hidden due to being a function or having a =hidden attribute. */
+  const isVisibleContext = contextRanked => showHiddenThoughts || (
```
raineorshine, line 26: I believe what is passed is an unranked `context`.

src/selectors/prevSibling.js
```diff
@@ -1,34 +1,50 @@
...
+    else if (!(contextViewActive() ? isVisibleContext(child.context) : isVisible(child))) {
       return false
     }
     else {
       prev = child
       return false
     }
   })
-  return prev
+  return prev && { ...prev, value: contextViewActive() ? head(prev.context) : prev.value }
```
raineorshine, line 49: `prevSibling` should return a `Child`, not a `string`.

anmolarora1: The return value is a `child` already

src/selectors/prevSibling.js
anmolarora1: That's right, I'll update the name

src/selectors/prevSibling.js
anmolarora1: That makes sense

src/selectors/prevSibling.js
raineorshine: Sorry, misread that! Thanks!

raineorshine:
@shresthabijay Can you test? Thanks!

raineorshine:
Actually, I was mistaken, no manual testing is needed!

This issue uses the approach of test-driven development. I wrote a failing test in a branch. The task is to make the test pass. The test provides a more accurate description of the requirements than a written description could provide.  Since the PR includes the test and it passes the GitHub Action, no other testing is needed!

PR #747, opened 2020-06-18
Formatting split fix

### LINKED ISSUE #742: Split thought breaks on formatting tags

shresthabijay:
fixes #742 
@raineorshine I found very easy solution of the problem we discussed yesterday. We can insert node at the end of the currently focused range. So we can add a text node with non printable character at the point and then take the inner html of the content editable and find the index of that non printable character. It works. 

Also I used `xhtml-purifier` to purify the values after split. `xhtml-purifier` adds some `<p>` tags and changes `<i>` to `<em>`, `<b>` to `<strong>`. So I stripped the values and updated the formatting tag regex to include **em** and **strong** tag. Please review the changes. Thanks!


src/reducers/splitThought.js
```diff
@@ -45,8 +48,10 @@ export default (state, { path, offset } = {}) => {
 
   // split the value into left and right parts
   const value = headValue(path)
-  const valueLeft = value.slice(0, offset)
-  const valueRight = value.slice(offset)
+
+  // Note: xhtmlPurifier adds unwanted <p> tags. So using strip to remove such tags
```
raineorshine, line 52: Would be helpful to add a comment here *why* `xhtmlPurifier` is being used. I would recommend including the url to the issue as well. This will make it easy for future developers to trace the reasoning for this code, which would otherwise be quite mysterious!

src/shortcuts/newThought.js
```diff
@@ -30,7 +31,8 @@ const exec = (dispatch, getState, e, { type }) => {
   // cancel if tutorial has just started
   if (tutorial && tutorialStep === TUTORIAL_STEP_START) return
 
-  const offset = window.getSelection().focusOffset
+  // Note: e.traget should be a content editable node
```
raineorshine, line 34: typo

raineorshine:
> @raineorshine I found very easy solution of the problem we discussed yesterday. We can insert node at the end of the currently focused range. So we can add a text node with non printable character at the point and then take the inner html of the content editable and find the index of that non printable character. It works.

Brilliant!!!

> 
> Also I used `xhtml-purifier` to purify the values after split. `xhtml-purifier` adds some `<p>` tags and changes `<i>` to `<em>`, `<b>` to `<strong>`. So I stripped the values and updated the formatting tag regex to include **em** and **strong** tag.

Sounds good

src/reducers/splitThought.js
shresthabijay: Sure will do that

shresthabijay:
@raineorshine I have made some changes . All tests are now passing. Please review it. Thanks!

raineorshine:
Thanks!

PR #748, opened 2020-06-18
Merges lint_typescript.json into eslintrc

### LINKED ISSUE #725: Combine lint_typescript.json and .eslintrc.json
Can we run the same linting process on both Javascript and Typescript?

If not, let's have lint_typescript.json extend .eslintrc.json so that the rules are not duplicated.

anmolarora1:
Fixes #725


.eslintrc.json
```diff
@@ -7,21 +7,40 @@
...
+      "parserOptions": {
+        "ecmaFeatures": { "jsx": true },
+        "ecmaVersion": 2018,
+        "sourceType": "module",
+        "project": "./tsconfig.eslint.json"
+      },
+      "plugins": ["@typescript-eslint"],
+      "rules": {
+        "import/prefer-default-export": 0,
+        // temporary fix from 'typescript-elint' docs  https://bit.ly/2XhUE8h
```
raineorshine, line 34: You can use the actual url here. `bit.ly` links give you no indication about the destination, forcing you to click on it to learn anything about the source.

lint_typescript.json
```diff
@@ -1,153 +0,0 @@
...
-    ],
-
-    // import
-    "import/order": ["error", {
-      "groups": [
-        ["external", "builtin"],
-        ["internal", "index", "sibling", "parent"]
-      ],
-      "newlines-between": "ignore"
-    }]
```
raineorshine, line 151: It appears that these rules have been removed and were not added in `.eslintrc.json`. Can you explain?

package.json
```diff
@@ -91,11 +91,8 @@
...
     "clean": "rm -rf ./node_modules/.cache/eslint-loader",
     "deploy": "npm version major && npm run build && firebase deploy",
-    "lint": "npm run lintts && npm run lintjs",
-    "lint:fix": "npm run lintts:fix && npm run lintjs:fix",
-    "lintjs": "eslint src",
-    "lintjs:fix": "npm run lintjs -- --fix",
-    "lintts": "eslint -c lint_typescript.json \"src/**/*.{ts,tsx}\" --no-ignore",
+    "lint": "eslint \"src/**/*.{js,ts,tsx}\"",
+    "lint:fix": "npm run lint -- --fix",
     "lintts:fix": "npm run lintts -- --fix",
```
raineorshine, line 96: Remove `lintts:fix`

.eslintrc.json
```diff
@@ -7,21 +7,40 @@
...
+        "plugin:@typescript-eslint/eslint-recommended",
+        "plugin:@typescript-eslint/recommended"
+      ],
+      "globals": { "Atomics": "readonly", "SharedArrayBuffer": "readonly" },
+      "parser": "@typescript-eslint/parser",
+      "parserOptions": {
+        "ecmaFeatures": { "jsx": true },
+        "ecmaVersion": 2018,
+        "sourceType": "module",
+        "project": "./tsconfig.eslint.json"
```
raineorshine, line 29: Any reason to have a separate `tsconfig.json` and `tsconfig.eslint.json`?

.eslintrc.json
```diff
@@ -7,21 +7,40 @@
...
+      },
+      "plugins": ["@typescript-eslint"],
+      "rules": {
+        "import/prefer-default-export": 0,
+        // temporary fix from 'typescript-elint' docs  https://bit.ly/2XhUE8h
+        "no-unused-vars": 0,
+        "@typescript-eslint/no-unused-vars": [2],
+        "@typescript-eslint/no-explicit-any": 0,
+        "@typescript-eslint/explicit-module-boundary-types": 0,
+        "@typescript-eslint/ban-ts-comment": 1
```
raineorshine, line 39: Do these only work as `overrides` rather than normal `rules`?

.eslintrc.json
```diff
@@ -40,7 +59,8 @@
     "fp",
     "jsdoc",
     "react",
-    "@typescript-eslint"
+    "@typescript-eslint",
+    "disable"
```
raineorshine, line 63: What does "disable" do?

.eslintrc.json
anmolarora1: Okay

.eslintrc.json
anmolarora1: They do work as overrides.

.eslintrc.json
anmolarora1: Sorry, this needs to be removed. It was added in order to disable all rules from `ts plugins` for `non-ts files`. But the suggested way was to use `overrides` for that.

lint_typescript.json
anmolarora1: I'll have to add `"@typescript-eslint/array-type"` and `"jsx-quotes"` rules back again. It was a manual error while I was trying to reduce the number of explicit rules. 

`semi` and `import/order ("sort-imports")` are a part of pre-existing eslint config already and they apply to ts files. And rule `"@typescript-eslint/explicit-function-return-type"` doesn't seem to throw an error anymore

.eslintrc.json
raineorshine: I wasn’t asking if they work as overrides. I was asking if they *only* work as overrides. If so, why? If not, why? Please provide a more thorough response.

anmolarora1: The rules under the overrides section are supposed to work `only` for the `matched files`. You can refer [this](https://eslint.org/docs/user-guide/configuring#disabling-rules-only-for-a-group-of-files
) more details.

.eslintrc.json
anmolarora1: Merged them in the latest commit

.eslintrc.json
raineorshine: Yes, I understand what `overrides` does. I am asking why it is needed here. You still have explained the reason. What happens when they are added as regular rules?

package.json
raineorshine: @anmolarora1 I think you forgot to do this.

PR #753, opened 2020-06-20
Make children container 100% width

### LINKED ISSUE #752: Thoughts get narrower with greater depth
A2364D36-D758-4D31-9F78-842AC31814C3

shresthabijay:
fixes #752

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -476,6 +476,7 @@ a.text-link {
   font-size: 15px;
   margin-top: 0;
   margin-bottom: 0;
+  width: 100%;
 }
 
 .children .child:first-child {


shresthabijay:
@raineorshine The child container width was not set to cover it's parent container width. Changing it to 100% worked. Please review the changes. Thanks!

shresthabijay:
@raineorshine I took some time to look into this issue but css is really confusing me. What I found is the width of `children` div is shrinking by some certain factor and is not retaining width of its parent div `child`. So at every depth width of the descendants keep getting smaller. I tried making changes but I couldn't do it without breaking other things.

raineorshine:
Okay, thanks. I'll take a stab at it.

PR #754, opened 2020-06-21
Store archived timestamp and persist it

### LINKED ISSUE #733: Store archived timestamp
Add an optional `archived` property to each Child. It should be set to `timestamp()` when a thought is archived and persisted across all modifications like `created`.

shresthabijay:
fixes #733

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "ts-key-enum": "^2.0.2",
     "typescript": "^3.7.5",
     "uuid": "^3.4.0",
-    "xhtml-purifier": "git+https://github.com/cybersemics/node-xhtml-purifier#types"
+    "xhtml-purifier": "git+https://github.com/cybersemics/node-xhtml-purifier.git#types"
   },
   "devDependencies": {
     "@sindresorhus/slugify": "^0.3.0",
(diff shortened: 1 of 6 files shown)


shresthabijay:
@raineorshine I have added `archived` field to individual thought inside `thoughtIndex`. It will be created only when thoughts are being archived and it persists on thought change and thought move. Please review the changes. Thanks!

src/reducers/existingThoughtMove.js
```diff
@@ -40,13 +40,17 @@ export default (state, { oldPath, newPath, offset }) => {
   const sameContext = equalArrays(oldContext, newContext)
   const oldThought = getThought(state, value)
 
-  /** Find exact thought from thoughtIndex. */
-  const exactThought = () => oldThought.contexts.find(thought => equalArrays(thought.context, oldContext) && thought.rank === oldRank)
+  const isArchived = newThoughts[newThoughts.length - 2] === '=archive'
```
raineorshine, line 43: Recommend to use the existing `util/isArchived`.

shresthabijay: `isArchived` takes `Lexeme` as parameter but we only have access to `newPath` in `exisitingThoughtMove`.

raineorshine: Of course, my bad.

Regardless, a thought is archived if any of its ancestors is `=archive`, not just the penultimate components of its path.

```js
newThoughts.indexOf('=archive') !== -1
```

PR #756, opened 2020-06-22
Add position:absolute to toolbar when keyboard is up

### LINKED ISSUE #545: Mobile Safari: Toolbar disappears when keyboard is up
Toolbar should be position: fixed and always visible.

pushkov-fedor:
Fixes #545

I found out plenty issues about position:fixed when keyboard is up on stackoverflow. Common solution - make element to position absolute. 
It works now, but i don't like small toolbar lagging during scrolling. Also i discovered that left and right arrow on toolbar are positioned a little tricky, we need to fix them separately from` toolbar-container` and `hamburger-menu` or change their layout in html and css. What do you think? 
Also we need to hide caret when toolbar is above it.
I stopped with current progress because 1.5hrs are over.


src/components/Editable.js
```diff
@@ -62,6 +62,33 @@ const EMPTY_THOUGHT_TIMEOUT = 5 * 1000
...
+}
+/** Update position of toolbar elements while scrolling in order to show them always on top. */
+const updateToolbarPositionOnScroll = () => {
+  const hamburgerMenu = document.getElementsByClassName('hamburger-menu')[0]
+  const toolbar = document.getElementsByClassName('toolbar-container')[0]
+  Array.from([hamburgerMenu, toolbar]).forEach(el => {
+    el.style.top = `${window.scrollY}px`
+  })
+}
+
```
raineorshine, line 91: Could this be done in the HamburgerMenu and Toolbar components via `state` and the appropriate reducers? I would prefer a more idiomatic react approach than direct DOM manipulation.

src/components/Editable.js
```diff
@@ -348,6 +378,8 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
    * Prevented by mousedown event above for hidden thoughts.
    */
   const onFocus = e => {
+    setTimeout(makeToolbarPositionFixed, 100)
```
raineorshine, line 381: What is the purpose of the delay, and does the length of time matter?

pushkov-fedor:
I fixed arrows, tried to add smooth animation via `transition`, but it doesn't work. I would like to search for solution.
Also we need to hide caret when it's below the toolbar, it could be tricky. What about 1-1.5 hour for researching and testing different solutions?

src/components/Editable.js
pushkov-fedor: Yes, i wrote this for testing, i can reorganize this code in any way.

src/components/Editable.js
pushkov-fedor: Author of this solution said that delay is required, but it works for 6s even without delay.  I don't think that length matters. I removed timeout. Does it still work on your device?

raineorshine: Removing the `setTimeout` works on iPhone 11, but only when applied to f4e3dfe54dc31f2368584c78966a8d8b0016a5d2. Something broke in 326233aa.

src/components/Editable.js
raineorshine: Sounds good. Once we get the functionality perfected than we can turn our attention to architecture.

raineorshine:
Sure, that deserves so me extra time. See what you can do in 2 hours of work. Thanks!

pushkov-fedor:
I decided to try hiding caret while it's below toolbar. `caret-color: transparent` works statically (e.g. if i add this style to `editable` class), but doesn't work when i add class `caret-transparent` with this style dynamically based on position of element. Have you ever encountered such behavior?

pushkov-fedor:
Also we can blur Editable, while it's below Toolbar, and caret will disappear

raineorshine:
> I decided to try hiding caret while it's below toolbar. `caret-color: transparent` works statically (e.g. if i add this style to `editable` class), but doesn't work when i add class `caret-transparent` with this style dynamically based on position of element. Have you ever encountered such behavior?

No, that’s curious. If it works statically I would expect it also work when attached dynamically. 

> Also we can blur Editable, while it's below Toolbar, and caret will disappear

Good to know, but we don’t want to remove the caret without the user choosing to do so themselves.

pushkov-fedor:
Hi, Raine! I was thinking about this issue today and come up with idea.
What i don't like now: 

1. Lagy toolbar behavior during scrolling especially if we scroll all page.

1. Strange caret behavior: dynamically changing caret color doesn't work, caret is above toolbar while thought is below (iphone 6s). Also i tested on mac mojave, we don't have such caret problem here, but we still have lags during scrolling.

Fixing these problems will take a lot of time. I think we should try another approach: we can remove position fixed of toolbar at all and make it position: block. Then we add max-height: 100vh to body and add fixed height to content and let content to be scrollable. We won't need to do some hacks with html&css and toolbar will be always on top. What do you think?

raineorshine:
> 1. Lagy toolbar behavior during scrolling especially if we scroll all page.

This is not ideal, but it is still better than having the toolbar disappear completely. Unfortunately I have encountered a lot of apps that have the same problem due to the underlying browser issue. I don't think the lagginess is a dealbreaker though.

> 2. Strange caret behavior: dynamically changing caret color doesn't work, caret is above toolbar while thought is below (iphone 6s). 

Okay, will you create a separate issue for the caret issue? It doesn't have to be a blocker on this, and maybe we can look for other solutions in the future.

> Fixing these problems might take some time. I think we should try another approach: we can remove position fixed of toolbar at all and make it position: block. Then we add max-height: 100vh to body and add fixed height to content and let content to be scrollable. We won't need to do some hacks with html&css and toolbar will be always on top. What do you think?

Sounds possible. I am very hesitant to head down the CSS rabbit hole on this though. It feels like it could be a big time sink. We really don't know what we're getting into or how long it will take. Let's go with the current solution for now so that we can stay focused on our Objectives.

pushkov-fedor:
I was thinking about how to organize it in the best way and came to this, but open to any ideas. Reorganization and tests on Chrome Desktop/Mobile, Safari Desktop/Mobile took some time.

pushkov-fedor:
Ready to create caret issue after approving PR

raineorshine:
Thanks Fedor. Unfortunately the refactor made the lagginess worse. Tested on iPhone 11.

Here is a video from before the refactor (135de69d750b8d9806a68ee4949f6796f1d4a9f3):

https://youtu.be/oypNPxwwXFA

Here is after (0d9a7c0f937d913a3f71a6f208dfcce795ce05ff):

https://youtu.be/oypNPxwwXFA

I'd like to figure how the best way to allocate our time and wrap up this task. Do you think the new performance issue is fixable? Or should we go back to the previous commit and do some simpler re-organization that does not change the behavior? Please let m know estimated time frames on any suggested lines of action.

pushkov-fedor:
Looks like direct DOM manupilations works faster then via state, i've tested both of them. I throttled `updateToolbarPositionOnScroll` and added transition. Could you check it please? If it's okay, i will return `updateToolbarPositionOnScroll` logic to reducer as it was in 0038d14.

But i believe that the best way to solve this problem is to fix height of toolbar and content to 100vh, remove overflow of body and then add overflow to content. But i'm afraid that this could cause some side problems with gestures and any scroll logic, i didn't dive in current implementation of these.

raineorshine:
Unfortunately this one is also worse than 135de69. Please halt all development on this task until we decide an appropriate path forward.

raineorshine:
> But i believe that the best way to solve this problem is to fix height of toolbar and content to 100vh, remove overflow of body and then add overflow to content. But i'm afraid that this could cause some side problems with gestures and any scroll logic, i didn't dive in current implementation of these.

@pushkov-fedor Do you think a couple hours is enough to get this prototyped so we can see if it is viable?

Otherwise I will revert to 135de69 so we can get this merged.

Thanks!

pushkov-fedor:
Sorry for such long delay! I think it's worth trying. I would like to spend extra 1.5-2 hours on reorganizing components and adding custom styles. I believe in 2 hours we definitely would understand whether its correct way or not.

raineorshine:
Okay, go for it!

pushkov-fedor:
I reorganized HTML and added CSS styles. It is far for ideal, still has some scroll problems, home icon and breadcrumbs hide below. Fixing these would take some time. Maybe it would be faster and cheaper in terms of time to merge 135de69? What do you think about current solution? Test it and let me know please

raineorshine:
> I reorganized HTML and added CSS styles. It is far for ideal, still has some scroll problems, home icon and breadcrumbs hide below. Fixing these would take some time. 

How much time?

> Maybe it would be faster and cheaper in terms of time to merge [135de69](https://github.com/cybersemics/em/commit/135de69d750b8d9806a68ee4949f6796f1d4a9f3)? 

Not sure yet. It would be really nice to solve both the jittery problem and the caret problem on mobile.

> What do you think about current solution? Test it and let me know please

The toolbar disappears when the keyboard is up (Mobile Safari), so the main problem remains. Are you not having this issue?

pushkov-fedor:
> The toolbar disappears when the keyboard is up (Mobile Safari), so the main problem remains. Are you not having this issue?

RPReplay-Final1596523251

pushkov-fedor:
> > I reorganized HTML and added CSS styles. It is far for ideal, still has some scroll problems, home icon and breadcrumbs hide below. Fixing these would take some time.
> 
> How much time?

Actually it's hard to predict right now.
1. We have 2 scrolls: inner scroll inside of content block, outer scroll of the whole page. Because of 2 scrolls, sometimes you can trigger outer (page) scroll instead of inner (content). This leads to hiding toolbar above viewport. I set page height limits up to 100vh, but my assumption that actual height exceeds these limits for some reason.

2. Home icon and breadcrumbs are related to each other and would be fixed together. Sticky position is broken for them.

3. Footer is hidden at the bottom of screen.

I think that these issues are related to each other, I would like to go by small steps and spend 45 minutes on third, i have a feeling that fixing it will simplify the remaining.

But what worries me the most is that toolbar still disappears for you. Is it not working at all?

raineorshine:
Yes, the toolbar disappears sometimes, in particular when entering editing mode on a thought further down the page (therefore activating the browser's autoscroll): https://youtu.be/ycAAw6Xj2j0

Are you able to reproduce this? You may have not encountered it if you were always initiating editing mode when already scrolled to the top.

pushkov-fedor:
Yes, i saw it, thanks! I will figure out whats the reason of such behavior.

pushkov-fedor:
@raineorshine Raine, hi! I've spent 1.5 hours yesterday on our main toolbar issue, but couldn't make any progress. My hypothesis is that on focus event, safari automatically scrolls to focused input, and it causes such behavior. Safari is always a pain in the neck, and i would like to make important statement. I think we are seriously mired in this problem. It works on Android, but I am not a Safari engineer :), so I can't predict any further time budgets that this issue could take, because any future work is researching and testing. I think that this tasks requires a special and uncommon solution.
We can continue to search other solutions for this problem and try to apply it to **em**. We can try to find websites, where navbar/header is fixed, when input is focused, and apply their styles to **em**. HTM and CSS of such solutions also could differ from em, so we must take this into account to. Another option is to try to use one of popular css frameworks for this task (e.g. Bootstrap), and see, if they've solved this problem. Or we can revert to previous not ideal but working solution with position: absolute style for toolbar.
I think that eventually we would come to solution, but i don't know how much time it could take
Please think about what i've written, and let me know about your decision.

raineorshine:
Thanks for your analysis and estimation of effort.

> Or we can revert to previous not ideal but working solution with position: absolute style for toolbar.

Let's use the suboptimal solution since we have other priorities. We can can come back to this in the future.

pushkov-fedor:
Okay, then i revert changes to 135de69, yes?

raineorshine:
> Okay, then i revert changes to [135de69](https://github.com/cybersemics/em/commit/135de69d750b8d9806a68ee4949f6796f1d4a9f3), yes?

Yes, thanks!

pushkov-fedor:
> > Okay, then i revert changes to [135de69](https://github.com/cybersemics/em/commit/135de69d750b8d9806a68ee4949f6796f1d4a9f3), yes?
> 
> Yes, thanks!

Done

raineorshine:
I'm getting an issue where it scrolls horizontally when the keyboard is up:

https://youtu.be/Zo1pnH6PdNA

pushkov-fedor:
> I'm getting an issue where it scrolls horizontally when the keyboard is up:
> 
> https://youtu.be/Zo1pnH6PdNA

Fixed

raineorshine:
Thank you!

PR #757, opened 2020-06-22
archiveThought: Should archive empty thought with descendants

### LINKED ISSUE #744: archiveThought: Should archive empty thought with descendants
When `archiveThought` is activated on an empty thought (`value: ""`), it is deleted permanently. However, it should not do this if it has any descendants.

Please write the corresponding unit test as well.

aqkhan:
Fixes #744

### DIFF
--- a/src/reducers/archiveThought.js
+++ b/src/reducers/archiveThought.js
@@ -23,6 +23,7 @@ import {
 // selectors
 import {
   getContextsSortedAndRanked,
+  getDescendants,
   hasChild,
   isContextViewActive,
   lastThoughtsFromContextChain,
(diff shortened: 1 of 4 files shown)


src/reducers/archiveThought.js
```diff
@@ -67,7 +68,8 @@ export default (state, { path } = {}) => {
   const isEmpty = value === ''
   const isArchive = value === '=archive'
   const isArchived = isThoughtArchived(path)
-  const isDeletable = isEmpty || isArchive || isArchived
+  const hasDescendants = getDescendants(state, path).length !== 0
+  const isDeletable = (isEmpty || isArchive || isArchived) && !hasDescendants
```
raineorshine, line 72: I would use `getThoughts` here. It is much faster than `getDescendants`, and has the same effect in this case.

src/selectors/getDescendants.ts
```diff
@@ -7,7 +6,7 @@ import { Child, Path } from '../types'
 
 /** Generates a flat list of all descendants. */
 const getDescendants = (state: State, thoughtsRanked: Path, recur?: boolean/* INTERNAL */): Child[] => {
-  const children = getThoughtsRanked(store.getState(), thoughtsRanked)
+  const children = getThoughtsRanked(state, thoughtsRanked)
```
raineorshine, line 9: Good catch

PR #760, opened 2020-06-25
Fix search trim

### LINKED ISSUE #751: Search gets trimmed during editing
Seems to be caused by the search editable being re-rendered during editing.

shresthabijay:
fixes #751

### DIFF
--- a/src/components/Search.js
+++ b/src/components/Search.js
@@ -29,9 +29,6 @@ const onKeyDown = e => {
   }
 }
 
-// eslint-disable-next-line jsdoc/require-jsdoc
-// const mapStateToProps = ({ archived }) => ({ search, archived })
-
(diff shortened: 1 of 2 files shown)


shresthabijay:
@raineorshine `Search` component was subscribed to both `search` and `archived` from redux store. So whenever new search value is dispatched from the component , it -re-rendered because of the subscription. Due to this `Editable` html property was set to new trimmed search value from the store.

I removed all the subscription from the `Search` component. We also don't need to pass  props `search` and `archived` to the `SearchSubthoughts` from `Search` component. `SearchSubthoughts` It can subscribe those values on its own. Please review the changes. Thanks!

raineorshine:
Excellent, thanks!

PR #761, opened 2020-06-25
fixes scaling of toolbars' scroll arrows

### LINKED ISSUE #758: Toolbar: Scroll arrows do not scale with toolbar
When the font size is increased, the scroll arrows should scale with the toolbar. When they are not scaled properly:

1. The scroll arrows are not vertically centered
2. The black background does not cover the toolbar.

Screen Shot 2020-06-24 at 7 49 12 AM

anmolarora1:
fixes #758

### DIFF
--- a/src/components/Toolbar.js
+++ b/src/components/Toolbar.js
@@ -85,7 +85,7 @@ const Toolbar = ({ cursorOnTableView, cursorOnAlphabeticalSort, cursorPinOpen, c
   const [overlayDescription, setOverlayDescription] = useState()
 
   const fg = dark ? 'white' : 'black'
-  const arrowHeight = fontSize / 1.5
+  const arrowWidth = fontSize / 3
 
   useEffect(() => {
     if (toolbarOverlay) {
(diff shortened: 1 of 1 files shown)


raineorshine:
Screen Shot 2020-06-25 at 8 02 59 AM

The arrows are much larger than they should be at font size 18.

raineorshine:
Also, the arrows are not square, so you can't set the width and height to the same value.

anmolarora1:
> Screen Shot 2020-06-25 at 8 02 59 AM
> 
> The arrows are much larger than they should be at font size 18.



> Screen Shot 2020-06-25 at 8 02 59 AM
> 
> The arrows are much larger than they should be at font size 18.

Updated in the latest commit to match the former dimensions

raineorshine:
@shresthabijay Can you test this? Thanks!

src/components/Toolbar.js
```diff
@@ -247,13 +248,13 @@ const Toolbar = ({ cursorOnTableView, cursorOnAlphabeticalSort, cursorPinOpen, c
                     : id === 'undo' ? fg
                     : id === 'redo' ? fg
                     : 'gray',
-                    width: fontSize,
-                    height: fontSize
+                    width: fontSize + 4,
+                    height: fontSize + 4,
```
shresthabijay, line 252: @anmolarora1 Why are you adding 4 to the `fontSize` ?

shresthabijay:
@raineorshine Arrows are of proper size, vertically aligned and black background covers toolbar in all font sizes. Looks pretty good to me!

src/components/Toolbar.js
raineorshine: The output size of text is slightly different than the output size of the svg icons. We need to add a constant amount to the font size so that they are more evenly matched.

Issue #605, opened 2020-04-28
Fade out toolbar and breadcrumbs


raineorshine:
On desktop when the user starts typing, fade out the toolbar and breadcrumbs (800ms, ease-in-out).

When the user moves their mouse, fade them back in.

This mimics the behavior in Notion.

pushkov-fedor:
Tried it on Notion, this behavior looks really cool. I would like to take this issue. I think it's 1-1.5 hour with development and testing.

raineorshine:
@anmolarora1 claimed this just before you, I just hadn't updated the issue yet! Thank you, though.

I agree it's a slick feature.

PR #763, opened 2020-06-26
Fade out toolbar and breadcrumbs on action/typing

anmolarora1:
#605 

Added two state variables to toggle the visibility of the `toolbar` and `breadcrumbs`. The action-creator that controls them is dispatched on `mouseMove` and `keyUp (within the thought)` events with toggle values `true` and `false` respectively.


src/action-creators/toggleToolbarAndBreadCrumbs.ts
```diff
@@ -0,0 +1,19 @@
+import { store } from '../store'
+import { ActionCreator } from '../types'
+
+/**
+ * Dispatches an alert action.
+ *
+ * @param value A boolean to represent the visibility state of toolbar & breadcrumbs.
+ */
+const toggleToolbarAndBreadCrumbs = (value: boolean): ActionCreator => dispatch => {
```
raineorshine, line 9: The second argument after `dispatch` is `getState`. Use that to avoid accessing the `store` directly.

src/action-creators/toggleToolbarAndBreadCrumbs.ts
```diff
@@ -0,0 +1,19 @@
+import { store } from '../store'
+import { ActionCreator } from '../types'
+
+/**
+ * Dispatches an alert action.
```
raineorshine, line 5: Comment is out of sync!

src/reducers/toggleToolbarAndBreadCrumbs.ts
```diff
@@ -0,0 +1,10 @@
+import { State } from '../util/initialState'
+
+/** Toggles the Toolbar Visibility. */
+const toggleToolbarAndBreadCrumbs = (state: State, { value }: { value?: boolean }) => ({
+  ...state,
+  showToolbar: value === null ? !state.showToolbar : value,
```
raineorshine, line 6: This is a bit more terse: `value ?? !state.showToolbar`

src/util/initEvents.ts
```diff
@@ -20,6 +22,10 @@ export const initEvents = () => {
     store.dispatch({ type: 'setCursor', thoughtsRanked, replaceContextViews: contextViews })
   })
 
+  window.addEventListener('mousemove', _.throttle(() => {
+    store.dispatch(toggleToolbarAndBreadCrumbs(true))
+  }, 1000, { leading: true }))
```
raineorshine, line 27: Only dispatch if toolbar is hidden to prevent spamming the action history.

src/util/initEvents.ts
```diff
@@ -3,9 +3,11 @@
 import { store } from '../store'
 import { keyDown, keyUp } from '../shortcuts'
 import * as db from '../db'
+import _ from 'lodash'
 
 // util
 import { decodeThoughtsUrl } from '../selectors'
+import toggleToolbarAndBreadCrumbs from '../action-creators/toggleToolbarAndBreadCrumbs'
```
raineorshine, line 10: Import from `action-creators` index instead of directly.

raineorshine:
Let's hide the toolbar when a shortcut is activated by the keyboard as well:

https://github.com/cybersemics/em/blob/dev/src/shortcuts.js#L170-L178

src/action-creators/toggleToolbarAndBreadCrumbs.ts
anmolarora1: I should've noticed. Will update

src/reducers/toggleToolbarAndBreadCrumbs.ts
anmolarora1: Nullish coalescing, cool 😃

raineorshine: I know, I just learned about it!

src/util/initEvents.ts
anmolarora1: It's being handled in the [action-creator](https://github.com/cybersemics/em/blob/9c863f46910b550f679c250b65bb81a416330da8/src/action-creators/toggleToolbarAndBreadCrumbs.ts#L11)

src/reducers/toggleToolbarAndBreadCrumbs.ts
raineorshine: You can use this technique for `showToolbar` as well.

raineorshine:
Code looks good. Thanks!

@shresthabijay Will you test? It needs to be tested on mobile and desktop. Layout and positioning of toolbar and breadcrumbs should be identical to `dev`.

shresthabijay:
@anmolarora1 Have you noticed this ? Sometimes toolbar flickers when typing. I could not reproduce it consistently but it's happening on both mobile and desktop.

Screen Capture_select-area_20200630093223

shresthabijay:
@raineorshine  @anmolarora1 I think we should trigger fade on key down.

Screen Capture_select-area_20200630094410

shresthabijay:
@raineorshine I think a little slower animation similar to Notion would look better.

raineorshine:
> @raineorshine @anmolarora1 I think we should trigger fade on key down.
> 
> Screen Capture_select-area_20200630094410

The difference between keyup and keydown is negligible given the delay. What you’re seeing in the screenshot is an issue with throttle. 

> @raineorshine I think a little slower animation similar to Notion would look better.

Yes, let’s match their delay.

src/reducers/toggleToolbarAndBreadCrumbs.ts
```diff
@@ -0,0 +1,10 @@
...
+  showBreadcrumbs: value ?? !state.showToolbar,
```
raineorshine, line 7: This should be `state.showBreadcrumbs`.

anmolarora1:
> > @raineorshine @anmolarora1 I think we should trigger fade on key down.
> > Screen Capture_select-area_20200630094410
> 
> The difference between keyup and keydown is negligible given the delay. What you’re seeing in the screenshot is an issue with throttle.
@raineorshine I think Bijay is right. It's because we're using `keyUp`.  The `throttle` is being used on `mousemove` action. 
We should be using `keyDown` instead

> 
> > @raineorshine I think a little slower animation similar to Notion would look better.
> 
> Yes, let’s match their delay.

The current delay is `200ms`. Should we go for `800`?

anmolarora1:
> @raineorshine @anmolarora1 I think we should trigger fade on key down.
> 
> Screen Capture_select-area_20200630094410

I tried a couple of times but haven't been able to replicate this one. Any specific steps that you're following, or maybe, any pattern that you might've noticed?

anmolarora1:
> @anmolarora1 Have you noticed this ? Sometimes toolbar flickers when typing. I could not reproduce it consistently but it's happening on both mobile and desktop.
> 
> Screen Capture_select-area_20200630093223

@shresthabijay I tried a couple of times but haven't been able to replicate this one. Any specific steps that you're following, or maybe, any pattern that you might've noticed?

shresthabijay:
> I tried a couple of times but haven't been able to replicate this one. Any specific steps that you're following, or maybe, any pattern that you might've noticed?

@anmolarora1 I don't have specific steps to reproduce this. It doesn't happen consistently. I think we should ignore this for now.

shresthabijay:
> The difference between keyup and keydown is negligible given the delay. What you’re seeing in the screenshot is an issue with throttle.

Actually I am continuously pressing down on a key and as soon I release the key toolbar fade is triggered. I don't think this is an issue with throttle.

raineorshine:
> > The difference between keyup and keydown is negligible given the delay. What you’re seeing in the screenshot is an issue with throttle.
> 
> Actually I am continuously pressing down on a key and as soon I release the key toolbar fade is triggered. I don't think this is an issue with throttle.

Ah, that makes sense! So neither keyup nor keydown is the correct event to attach to. We need to attach to the change event.

anmolarora1:
> > > The difference between keyup and keydown is negligible given the delay. What you’re seeing in the screenshot is an issue with throttle.
> > 
> > 
> > Actually I am continuously pressing down on a key and as soon I release the key toolbar fade is triggered. I don't think this is an issue with throttle.
> 
> Ah, that makes sense! So neither keyup nor keydown is the correct event to attach to. They do not account for a key held down (although it should be pointed out that this would rarely occur in normal use). We need to attach to the change event.

@raineorshine The `keydown` event would work for a key held down. If we attach it to the change event though, the toolbar won't fade while the user is moving his cursor within a thought, which is different from `notion`.

raineorshine:
> @raineorshine The `keydown` event would work for a key held down.

You’re right! Thank you. That will work. 

> If we attach it to the change event though, the toolbar won't fade while the user is moving his cursor within a thought, which is different from `notion`.

Great point as well. Thanks for the attention to detail!

raineorshine:
@shresthabijay Ready to test again

shresthabijay:
@raineorshine Slower animation is looking great. Layout and positioning of toolbar and breadcrumbs are identical to `dev`. There are no regressions.

shresthabijay:
@raineorshine @anmolarora1 I was analyzing Notion's implementation of this feature and I noticed that when we start typing and move mouse too at the same time, it doesn't hide the toolbar. However our current implantation causes toolbar to fade in and out. I don't know how often people use keyboard and mouse that way but still I thought it was worth pointing out.

raineorshine:
> @raineorshine @anmolarora1 I was analyzing Notion's implementation of this feature and I noticed that when we start typing and move mouse too at the same time, it doesn't hide the toolbar. However our current implementation causes toolbar to fade in and out. I don't know how often people use keyboard and mouse that way but still I thought it was worth pointing out.

Good observation. While users may not type and move the mouse at the same time, they could activate keyboard shortcuts and move the mouse at the same time. It would be nice if the toolbar didn't fade in and out.

@anmolarora1 If you can think of a relatively quick solution, go for it, otherwise we can backlog it since it is lower priority.

anmolarora1:
> > @raineorshine @anmolarora1 I was analyzing Notion's implementation of this feature and I noticed that when we start typing and move mouse too at the same time, it doesn't hide the toolbar. However our current implementation causes toolbar to fade in and out. I don't know how often people use keyboard and mouse that way but still I thought it was worth pointing out.
> 
> Good observation. While users may not type and move the mouse at the same time, they could activate keyboard shortcuts and move the mouse at the same time. It would be nice if the toolbar didn't fade in and out.
> 
> @anmolarora1 If you can think of a relatively quick solution, go for it, otherwise we can backlog it since it is lower priority.

@raineorshine I've added a fix and pushed the [code](https://github.com/cybersemics/em/pull/763/commits/a682c17b1f75219949c04efab7d0b897544a278c). Please review

raineorshine:
Thanks, will do!

You missed this comment: https://github.com/cybersemics/em/pull/763/files#r447179605

raineorshine:
- [x] Fade animation looks good
- [ ] Toolbar should fade back in as soon as I start moving my mouse, not when I stop moving my mouse.
- [ ] Toolbar should not fade out when only a modifier key is pressed: Ctrl, Shift, Alt.

src/components/Breadcrumbs.js
```diff
@@ -53,7 +53,7 @@ export const Breadcrumbs = ({ path, thoughtsLimit, charLimit, className }) => {
       <TransitionGroup>
         {overflowArray.map((thoughtRanked, i) => {
           const subthoughts = !thoughtRanked.isOverflow ? ancestors(path, thoughtRanked) : null
-          return <CSSTransition key={i} timeout={200} classNames='fade'>
+          return <CSSTransition key={i} timeout={500} classNames='fade-700'>
```
raineorshine, line 56: Should the timeout and CSS duration be the same amount?

src/util/initEvents.ts
```diff
@@ -20,6 +21,8 @@ export const initEvents = () => {
     store.dispatch({ type: 'setCursor', thoughtsRanked, replaceContextViews: contextViews })
   })
 
+  window.addEventListener('mousemove', () => store.dispatch(toggleToolbarAndBreadCrumbs(true)))
```
raineorshine, line 24: It would be more efficient to throttle outside the dispatch. Any reason why it was moved inside?

anmolarora1: The throttle has been moved to the [action-creator](https://github.com/cybersemics/em/blob/c954efc5607f85cac8e251d0ccc60d16b09cf3f8/src/action-creators/toggleToolbarAndBreadCrumbs.ts) to handle this [issue](https://github.com/cybersemics/em/pull/763#issuecomment-653035806), where the key-press and `mousemove` events are fired together. It's to ensure that only the last action gets called in an event where multiple actions are dispatched within a certain timespan (100ms in this case)

src/components/Breadcrumbs.js
anmolarora1: Yes, it should be the same ideally. I've updated it to `600ms`. The [throttled action-creator](https://github.com/cybersemics/em/blob/c954efc5607f85cac8e251d0ccc60d16b09cf3f8/src/action-creators/toggleToolbarAndBreadCrumbs.ts#L7) adds a delay of `100ms`. 
(Notion uses `700`)

src/util/initEvents.ts
raineorshine: Okay. Can you look for another way to do it? I'm really not comfortable dispatching an action on every mousemove trigger.

anmolarora1: This is a thunk being dispatched, and [this check](https://github.com/cybersemics/em/blob/c954efc5607f85cac8e251d0ccc60d16b09cf3f8/src/action-creators/toggleToolbarAndBreadCrumbs.ts#L9) within the action-creator ensures that we don't fire too many actions. Let me know if that addresses your concern. 
The other way of handling it could be wrapping [this](https://github.com/cybersemics/em/blob/4b2284367269839d48f752c438aa4e7e3c94014f/src/util/initEvents.ts#L24) function with a `debounce`.

raineorshine: > This is a thunk being dispatched, and this check within the action-creator ensures that we don't fire too many actions. Let me know if that addresses your concern.

No, I don't care for this approach. Too many thunks being dispatched unnecessarily.

> The other way of handling it could be wrapping this function with a debounce and set fire it on the leading edge, which would ensure that only a single dispatch gets called right when the mouse starts to move, instead of getting called too frequently while the mouse is being moved.

Yes, I would appreciate this other approach.

raineorshine:
> * [x]  Fade animation looks good
> * [ ]  Toolbar should fade back in as soon as I start moving my mouse, not when I stop moving my mouse.
> * [ ]  Toolbar should not fade out when only a modifier key is pressed: Ctrl, Shift, Alt.

@anmolarora1 I think you may have missed these...?

src/components/Editable.tsx
```diff
@@ -443,6 +444,14 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
     }
   }
 
+  /**
+   * Prevents onKeyDownAction call for shift, alt or ctrl keys.
+   */
+  const onKeyDown = (e: React.KeyboardEvent) => {
+    if (e.shiftKey || e.altKey || e.ctrlKey) return
```
raineorshine, line 451: This incorrectly returns when pressing a modifier key in combination with another key, e.g. `Ctrl + RightArrow`. We only want to cancel when *only* a modifier key is pressed.

Also, you will need to add `e.metaKey` for Mac.

src/components/Editable.tsx
```diff
@@ -443,6 +445,16 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
     }
   }
 
+  /**
+   * Prevents onKeyDownAction call for shift, alt or ctrl keys.
+   */
+  const onKeyDown = (e: React.KeyboardEvent) => {
+    if ((e.shiftKey || e.altKey || e.ctrlKey || e.metaKey) && MODIFIER_KEYS.find(keyName => keyName === e.key)) {
+      return
+    }
```
raineorshine, line 454: Glad you created a data structure for `MODIFIER_KEYS`!

Rather than store `MODIFIER_KEYS` as an array, store it as an object to allow O(1) lookup by key. This will also mitigate the explicit check for `e.shiftKey || e.altKey || e.ctrlKey || e.metaKey`.

raineorshine:
Please write "Fixes #605" instead of just "#605" in the PR description. Then GitHub will automatically close the associated issue. Thanks!

PR #767, opened 2020-06-29
Add subthought when empty second column is tapped

### LINKED ISSUE #509: Table: Tap to create column 2 thought
In a table, allow the user to tap in an empty column 2 to create a new subthought. 

e.g. 

- a *[table]*
  - b      *[tap here]*

Tapping to the right of `b` should result in:

- a *[table]*
  - b      *[empty thought]*

Which in list view would be:

- a
  - b
    - *[empty thought]*

shresthabijay:
fixes #509

### DIFF
--- a/src/components/Editable.js
+++ b/src/components/Editable.js
@@ -195,12 +195,7 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
   const setSelectionToCursorOffset = () => setSelection(contentRef.current, { offset: cursorOffset })
 
   useEffect(() => {
-
-    console.log(thoughtsRanked, value, 'yehawww')
-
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine I have added functionality to add a subthought when empty second column with no visible children is clicked or tapped. But it introduced strange cursor selection regression in mobile. At certain cases cursor get focused on different nodes back and forth in some kind of loop. I am still looking into it.

raineorshine:
I think there may be a misunderstanding. The task is to create a new subthought when no subthought yet exists. The click event can’t go in the Editable component, because there is no Editable yet in column 2. It is an empty column, not an empty thought that needs to be clickable. 

ACE2CE8F-EF4B-426E-92AF-3FFAA1FC599A

PR #770, opened 2020-06-30
Fix note disappearing on thought edit

### LINKED ISSUE #766: Note disappears when editing thought
## Steps to Reproduce

- A [cursor]
  - `=note`
    - B

Create the above structure and then edit `A`. 

## Current Behavior

Note `B` disappears after edit throttle delay. 

## Expected Behavior

Note should not disappear. 

## Notes
This is likely caused by using `cursorBeforeEdit` instead of `cursor`, or vice versa, in the logic for either determine if the note should be rendered or the logic for fetching the note itself. 

Make sure the solution does not cause a re-render and does not cause a `cursor` regression elsewhere.

shresthabijay:
fixes #766

### DIFF
--- a/src/components/Thought.js
+++ b/src/components/Thought.js
@@ -449,6 +449,7 @@ const ThoughtContainer = ({
     thoughtsRanked.length > 2
 
   const thoughts = pathToContext(thoughtsRanked)
+  const thoughtsLive = pathToContext(thoughtsRankedLive)
   const context = contextOf(thoughts)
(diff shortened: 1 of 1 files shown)


shresthabijay:
@raineorshine We needed to pass live `context` and `thoughtsRanked` to the `Note` component. However it took me some to realize this solution.

raineorshine:
> @raineorshine We needed to pass live `context` and `thoughtsRanked` to the `Note` component. However it took me some to realize this solution.

Hopefully it was a learning experience at least!

Issue #727, opened 2020-06-05
Ancestor drop target incorrectly highlighted


raineorshine:
## Steps to Reproduce

- a
  - b
    - c

1. Sort `a`
2. Drag `b` over `c`

## Current Behavior

Drop target above `b` is incorrectly highlighted.

## Expected Behavior

No drop target should be highlighted since `b` cannot be dragged to its descendants. It should be the same behavior as an unsorted context.

incora-dev:
@raineorshine Here I've found an issue when we drag some item to descendants of sorted thought. Then some child of sorted thought is highlighted. Looks like I've found solution already but still working on it.

shresthabijay:
> 2\. Drag `b` over `c`

@raineorshine I think it should Drag `a` over `b` since `a` is the sorted thought.

raineorshine:
> > 2. Drag `b` over `c`
> 
> @raineorshine I think it should be Drag `a` over `b`.

No, the steps listed are correct. That may be a separate issue.

PR #773, opened 2020-07-01
Ancestor drop target incorrectly highlighted

incora-dev:
#727


src/components/Thought.js
```diff
@@ -224,7 +224,7 @@ const canDrop = (props, monitor) => {
   const thoughtsTo = props.thoughtsRankedLive
   const thoughts = pathToContext(props.thoughtsRankedLive)
   const context = contextOf(thoughts)
-  const isSorted = getSortPreference(state, context) === 'Alphabetical'
+  const isSorted = getSortPreference(state, context).includes('Alphabetical')
```
raineorshine, line 227: Can you explain this? The short preference should exactly match `Alphabetical`.

raineorshine:
@shresthabijay Can you test? Thanks!

src/components/Thought.js
incora-dev: @raineorshine It returns array, so, it couldn't match and this check never was true.

raineorshine: I would expect it to return a `string` value, since the sort preference is indicated by a string like “Alphabetical”. Wouldn’t you agree? If it is returning an array than it has a bug.

incora-dev: @raineorshine Yes, I agree. But I'm not sure if it affect other places where it's used. So, I just used `includes` instead of changing `getSortPreference `.

raineorshine: What is the faulty condition that is causing `getSortPreference` to return an array? That should be investigated. We want to get to the root of the problem rather than covering it up.

raineorshine: Typescript would throw an error if any of the callers were expecting an array. There are only seven places in the codebase where `getSortPreference` is called, so it is easy to confirm by hand.

shresthabijay:
> @shresthabijay Can you test? Thanks!

Sure. On it!

shresthabijay:
@raineorshine  This PR solves the mentioned issue. However I noticed few other things.

Screen Capture_select-area_20200702164054

Here `a` has sort toggled on. As you can see in the gif we can drag the grandchildren into the sorted context but the drop target is not highlighted.

Unnamed


Somehow two thoughts get highlighted at once. I don't have exact steps to reproduce it but it happens quite frequently. 

However looking at the degree of changes made to the codebase by this PR, I don't think these are regressions introduced by it.

raineorshine:
@shresthabijay Thank you. Please create one or more separate issues for the aforementioned observations.

src/components/Thought.js
incora-dev: @raineorshine I was written on JS at first. I got TS version only during merge. In JS version there are no types. So, I thought that there can be a list of sort preferences as it also make sense if we will be able to sort not only alphabetically.

raineorshine: Okay, that makes sense. We don't have that now, but maybe in the future.

PR #774, opened 2020-07-01
Outdent thought on enter if thought is empty

### LINKED ISSUE #769: Hit Enter on empty thought to outdent
- [ ] Unit Test

Condition: Only if the thought is the last visible child in its parent context. 

## Why?

The default behavior of hitting Enter on an empty thought is to create a new thought after. This is unused and unneeded. Instead, we can coopt this specific case to provide a convenient way to outdent a newly created thought, to mitigate the mistake of creating a new thought instead of a new uncle, or to quickly create a “distant uncle” at higher levels. 

## Precedent

This feature exists on WorkFlowy and Roam.

shresthabijay:
fixes #769


shresthabijay:
@raineorshine I have added logic to dispatch `outdent` event inside `newThought` shortcut's exec function when thought is empty. Also I have written a test for it. Please review the changes. Thanks!

src/shortcuts/__tests__/newThought.js
```diff
@@ -0,0 +1,59 @@
...
+    { value: 'a', rank: '0' },
+    { value: 'b', rank: '1' },
+    { value: 'c', rank: '2' },
+    { value: 'd', rank: '3' },
+    { value: 'e', rank: '4' },
+    { value: 'f', rank: '5' },
+  ] })
+
+  // mount wrapper component to simulate global key down events
+  const wrapper = await mount(<TestWrapper/>)
```
raineorshine, line 37: What about calling `keyDown` directly to avoid `mount`?

raineorshine:
Nice work! The test case is great. 

There’s a question I’d like to pose to you. Notice how `outdent` gets dispatched from `newThought`. This is a bit odd. Why would the `newThought` shortcut dispatch `outdent`? Ah, because `newThought` is bound to the Enter key. This would be confusing to a new developer. They probably would go through the same thought process and trace the code to understand this. How could we refactor this to avoid the misunderstanding?

src/shortcuts/__tests__/newThought.js
shresthabijay: `keyDown` uses synthesized event `e`. So we have to pass an object with properties such as `preventDefault`, `metaKey`. We can pass a noop function as `preventDefault` to avoid error. Should we do that ?

raineorshine: Yes, I think that would be cleaner. We are then just mocking the event rather than the whole mounted node.

shresthabijay:
> There’s a question I’d like to pose to you. Notice how `outdent` gets dispatched from `newThought`. This is a bit odd. Why would the `newThought` shortcut dispatch `outdent`? Ah, because `newThought` is bound to the Enter key. This would be confusing to a new developer. They probably would go through the same thought process and trace the code to understand this. How could we refactor this to avoid the misunderstanding?

Yes indeed, it would be confusing to have `outdent` event inside `newThought`. Now a shortcut can perform different events depending on the some condition. But only a single event handler is bound to a shortcut. We should separate `gesture` and `keyboard` from the `shortcut` object and then create a global shortcut array that binds `gesture`, `keyboard` and array of shortcuts associated with them. So number of shortcuts can run on single `keyboard` or `gesture`.

```js
[
 {
   keyboard: { key: 'Enter' },
   gesture: 'rd',
   shortcuts: [newThought, outdent]
 }
 ,
 ...
]
```
We also may need to pass `keyboard` and `gesture` to `canExecute` to run different logic depending on which type of shortcut key triggered the event. This may not be the best solution. Let's discuss.

raineorshine:
One downside to this is that the condition would need to be duplicated within each shortcut. In the above case, `outdent` would have to check if the condition for its execution were true, and `newThought` would have to check if the same condition is false. Duplicating and inverting the condition is more error prone as it introduces the risk of them becoming out of sync.

There is a simpler solution. Do you see it?

shresthabijay:
> One downside to this is that the condition would need to be duplicated within each shortcut. In the above case, `outdent` would have to check if the condition for its execution were true, and `newThought` would have to check if the same condition is false. Duplicating and inverting the condition is more error prone as it introduces the risk of them becoming out of sync.
> 

Totally agreed on this.

shresthabijay:
> There is a simpler solution. Do you see it?

Not yet. Can you explain ?

raineorshine:
I suggest moving most of the logic of the `newThought` shortcut to an action-creator. Then, instead of binding the Enter key directly to `newThought`, bind it to a generic shortcut `enter`. Then `enter` can check the condition and dispatch either `newThought` or `outdent` as needed. 

While it might seem odd to create a generic `enter` shortcut, it is a better reflection of the polymorphic behavior of that shortcut. We don’t know yet what effect Enter should have when we bind the shortcut. When the condition is checked, the appropriate action can be dispatched. 

You can think of this as an alternative solution that keeps the state-dependent condition out of the shortcut logic. We utilize the shortcut architecture as-is, and push the complexity downstream into the shortcut exec function and subsequent action-creators. Thus the shortcut architecture does not need to “know” that there are multiple branches that could be called from a given shortcut key. It is just responsible for binding the shortcut key. 

Most importantly, the business logic for each branch is separated and handled in an appropriate place that matches the developer’s expectations of where that code should be found.

shresthabijay:
> While it might seem odd to create a generic `enter` shortcut, it is a better reflection of the polymorphic behavior of that shortcut. We don’t know yet what effect Enter should have when we bind the shortcut. When the condition is checked, the appropriate action can be dispatched.

I was thinking of this yesterday but it seemed too odd. We have been naming all the shortcuts based on type of action they perform and now we have a odd shortcut file that has a generic name based on its key. But this is a good idea to reflect the polymorphic behavior of the shortcut as you mentioned. May be we should write down this convention for developers.

shresthabijay:
> You can think of this as an alternative solution that keeps the state-dependent condition out of the shortcut logic. We utilize the shortcut architecture as-is, and push the complexity downstream into the shortcut exec function and subsequent action-creators. Thus the shortcut architecture does not need to “know” that there are multiple branches that could be called from a given shortcut key. It is just responsible for binding the shortcut key.

Agreed!

shresthabijay:
@raineorshine I have migrated `newThought` exec function's logic to a new `action-creator` and made necessary changes. I also made changes to the test file. Please review it. Thanks!

src/shortcuts/enter.js
```diff
@@ -0,0 +1,63 @@
...
+  id: 'enterAliases',
+  name: 'Enter',
+  hideFromInstructions: true,
+  gesture: ['rdld', 'rdldl', 'rdldld', 'rldl', 'rldld', 'rldldl'],
+  // on mobile, the shift key should cause a normal newThought, not newThoughtAbove
+  // smuggle it in with the aliases
+  ...isMobile ? { keyboard: { key: 'Enter', shift: true } } : null,
+  canExecute: () => isDocumentEditable(),
+  exec
+}
```
raineorshine, line 63: I'd invite you to think more about the semantics of this. We want to think about the architectural semantics at all levels, just like the reasoning we took to separate `newThought` and `enter`.

**Why would gestures be bound to a shortcut called `enter`?**

This catches one's attention because it is semantically incongruous. Once investigated it also reveals an unintended effect: Doing the gesture for `newThought` on an empty thought will cause an outdent, though the intention is that this will only happen when the Enter key is pressed.

I'd like to hear a proposal to resolve the incongruity and prevent the outdent on gesture.

shresthabijay: > Why would gestures be bound to a shortcut called `enter`?

We shouldn't name it based on `keyboard` or `gesture`. We should go with something like `newThoughtOrOutdent` like you mentioned earlier.

shresthabijay: > Doing the gesture for `newThought` on an empty thought will cause an outdent, though the intention is that this will only happen when the Enter key is pressed.

We should encapsulate individual actions into their own separate `exec` functions. Also we should create separate functions to check if the certain actions can be executed. This way we can write main `exec` function that can decide which action to trigger based on certain conditions or `type` of event (i.e `keyboard` or `gesture`). Even if the aliases behave differently we can create a exec function that only runs certain actions with very little code.

```js
{
  exec : (dispatch, getState, e, { type }) => {
    if (shouldExecOutdent(getState) && type === 'keyboard') dispatch({ type: 'outdent' })
    else newThoughtExec(dispatch, getState, e, { type })
  }
}

// for alias that doesn't need to trigger outdent
{
  exec: newThoughtExec
}
```

src/selectors/isLastVisibleChild.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const filteredChildren = children.filter(child => {
+    return state.showHiddenThoughts ||
+      // exclude meta thoughts when showHiddenThoughts is off
+      (!isFunction(child.value) && !hasChild(state, unroot(parentContext.concat(child.value)), '=hidden'))
+  })
+
+  const index = filteredChildren.findIndex(child => equalThoughtRanked(head(path), child))
+
+  return index === filteredChildren.length - 1
```
raineorshine, line 32: This can be simplified and optimized by:

```js
return equalThoughtRanked(head(path), filteredChildren[filteredChildren.length - 1])
``` 

Since you know the exact object to compare, there is no need to iterate through all children.

raineorshine:
Note the changes I added. In general you want to only split up functionality when needed, or if there is a convention that allows small functions to be easily understood (such as selectors) as opposed to ad hoc functions that require the developer to asses their purpose and use. 

In this case, having separate functions for different checks was harder to read then inlining them, as the `exec` function is only about 10 lines of code with two simple branches.

PR #775, opened 2020-07-02
Indent thought on adding space at the beginning

### LINKED ISSUE #771: Hit space at beginning of thought to indent
- [x] Unit Test

This provides a convenient way to indent thoughts on mobile where there is no tab key. (For consistency, however, this functionality should not be constrained to mobile.)

shresthabijay:
fixes #771


shresthabijay:
@raineorshine  I have used `onChangeHandler` inside `Editable` to check if space has been added at the beginning of the thought and then dispatch `indent`. I also prevented `indent` if thought has  `=readonly` or `=immovable` meta children. I have written two tests for this one. One to check if the indent happens on adding space at the beginning and another to check if `indent` doesn't happen when thought is `=immovable`. 

I may have mentioned you earlier that I ran into some trouble while I was writing test. Turns out `await` was missing before `createTestApp` inside `beforeAll` function. It caused tests to run before app was initialized. I added `await`.

```js
beforeAll(async () => {
  await createTestApp()
})
```

src/components/Editable.js
```diff
@@ -248,6 +250,12 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
     dispatch(setEditingValue(newValue))
 
     if (newValue === oldValue) {
+
+      // if new value starts with space then trigger indent
+      if (isDocumentEditable() && !readonly && !immovable && state.cursor && e.target.value.startsWith(' ')) {
```
raineorshine, line 255: `isDocumentEditable()` is unnecessary since the `Editable` is [disabled in Thought.js](https://github.com/cybersemics/em/blob/dev/src/components/Thought.js#L351).

src/components/__tests__/Subthoughts.js
```diff
@@ -20,7 +20,7 @@ const whereContext = context => node => equalArrays(pathToContext(node.props().t
 
 // cannot figure out how to unmount and reset after each test so that we can use beforeEach
 beforeAll(async () => {
-  createTestApp()
+  await createTestApp()
```
raineorshine, line 23: Good catch. Question is, why is `beforeAll` in this test file affecting tests in other files? Will you create an issue for this please? Thanks!

src/components/__tests__/Editable.js
```diff
@@ -0,0 +1,99 @@
...
+  await store.dispatch(importText(RANKED_ROOT, `
+    - a
+      - b
+        - c
+        - d`))
+
+  await store.dispatch({
+    type: 'setCursor',
+    thoughtsRanked: [{ value: 'a', rank: 0 }, { value: 'b', rank: 1 }, { value: 'd', rank: 3 }],
+    flag: true })
```
raineorshine, line 33: Was `flag` used for debugging?

raineorshine:
> @raineorshine I have used `onChangeHandler` inside `Editable` to check if space has been added at the beginning of the thought and then dispatch `indent`. I also prevented `indent` if thought has `=readonly` or `=immovable` meta children. I have written two tests for this one. One to check if the indent happens on adding space at the beginning and another to check if `indent` doesn't happen when thought is `=immovable`.

Great! Thanks for your thoroughness.

Can I ask why you chose to handle this in the `onChangeHandler` instead of higher up? I was thinking that it would be handled in a shortcut. `Editable` is already so complex.

src/components/__tests__/Editable.js
shresthabijay: Yes. I forgot to remove it. I will do that.

src/components/Editable.js
shresthabijay: Sure. I will remove it.

src/components/__tests__/Subthoughts.js
shresthabijay: I don't think it is affecting tests in other files. It was surprisingly working in this test without `await`.

shresthabijay:
> Can I ask why you chose to handle this in the `onChangeHandler` instead of higher up? I was thinking that it would be handled in a shortcut. `Editable` is already so complex.

Getting access to value from event handler was easy and simple. I don't have other specific reason for this. I think you are right. I will convert this to a shortcut.

raineorshine:
Okay, thanks. We also get the benefit of being able to test it by invoking the shortcut directly and avoiding `mount`. This will speed up tests by a couple seconds.

I would suggest a shortcut named `indentWithSpace` since it will always either indent or not fire at all. If it had branching functionality like `enter`, we may be forced to simply call it `space`, as anything more specific creates a semantic incongruity. But in this case it will always indent.

PR #778, opened 2020-07-04
Mock dexie

### LINKED ISSUE #664: Mock dexie
I was not able to successfully mock dexie in `setupTests.js`. I tried conditionally mocking within `db.js` itself, which fixes `initDB`, but the tests still fail. All db calls have to be ignored until this is solved. 

```js
if (process.env.NODE_ENV === 'test') {
  Dexie.dependencies.indexedDB = require('fake-indexeddb')
  Dexie.dependencies.IDBKeyRange = require('fake-indexeddb/lib/FDBKeyRange')
}
```

Based on https://github.com/dfahlander/Dexie.js/issues/495#issuecomment-288803216

Also tried this to no avail:

```js
require('fake-indexeddb/auto')
```

shresthabijay:
fixes #664

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -23,8 +23,8 @@ import {
   preloadSources,
 } from './action-creators'
 
-// export the promise for testing
-export const initialized = (async () => {
+/** Initilaize local db , firebase and window events. */
+export const initialize = async () => {
 
   // load local state unless loading a public context or source url
   await initDB()
@@ -52,8 +52,10 @@ export const initialized = (async () => {
   initEvents()
 
   return localStateLoaded
+}
 
-})()
+// Note: Manually initialize during test environment
+if (process.env.NODE_ENV !== 'test') initialize()
 
 /** Partially apply state to a function. */
 const withState = f => (...args) => f(store.getState(), ...args)
--- a/src/components/__tests__/Mount.js
+++ b/src/components/__tests__/Mount.js
@@ -30,7 +30,7 @@ it('create, navigate, and edit thoughts', async () => {
   windowEvent('keydown', { key: 'Enter', shiftKey: true, ctrlKey: true })
 
   act(() => {
-    jest.runAllTimers()
+    jest.runOnlyPendingTimers()
   })
 
   // state
--- a/src/components/__tests__/Mount.js
+++ b/src/components/__tests__/Mount.js
@@ -1,9 +1,11 @@
 import { store } from '../../store'
 import { getThoughtsRanked } from '../../selectors'
-import { createTestApp, windowEvent } from '../../setupTests'
+import windowEvent from '../../test-helpers/windowEvent'
+import createTestApp from '../../test-helpers/createTestApp'
+import { act } from 'react-dom/test-utils'
 
 beforeEach(async () => {
-  createTestApp()
+  await createTestApp()
 })
 
 // test basic thought operations using fully mounted app
@@ -27,7 +29,9 @@ it('create, navigate, and edit thoughts', async () => {
   // create top subthought
   windowEvent('keydown', { key: 'Enter', shiftKey: true, ctrlKey: true })
 
-  jest.runAllTimers()
+  act(() => {
+    jest.runAllTimers()
+  })
 
   // state
   const subthoughts = getThoughtsRanked(store.getState(), ['a'])
--- a/src/components/__tests__/db-test.js
+++ b/src/components/__tests__/db-test.js
@@ -1,11 +1,6 @@
-import { initialize } from '../../App'
 import { hashThought } from '../../util'
 import { store } from '../../store'
 import * as db from '../../db'
-
-beforeAll(async () => {
-  await initialize()
-})
 
 it('load settings into indexedDB on initialization', async () => {
   const hash = hashThought('Settings')
--- a/src/components/__tests__/db-test.js
+++ b/src/components/__tests__/db-test.js
@@ -1,0 +1,42 @@
+import { initialize } from '../../App'
+import { hashThought } from '../../util'
+import { store } from '../../store'
+import * as db from '../../db'
+
+beforeAll(async () => {
+  await initialize()
+})
+
+it('load settings into indexedDB on initialization', async () => {
+  const hash = hashThought('Settings')
+
+  const thoughtState = store.getState().thoughts.thoughtIndex[hash]
+
+  expect(thoughtState).not.toBeUndefined()
+  expect(thoughtState.contexts.length).toEqual(1)
+
+  const thoughtDB = await db.getThought(hash)
+
+  expect(thoughtDB).not.toBeUndefined()
+  expect(thoughtDB.contexts.length).toEqual(1)
+
+  expect(thoughtState.contexts[0].id).toEqual(thoughtDB.contexts[0].id)
+})
+
+it('presist data on indexedDB after adding new thought', async () => {
+
+  await store.dispatch({ type: 'newThought', value: 'db-test' })
+  const hash = hashThought('db-test')
+
+  const thoughtState = store.getState().thoughts.thoughtIndex[hash]
+
+  expect(thoughtState).not.toBeUndefined()
+  expect(thoughtState.contexts.length).toEqual(1)
+
+  const thoughtDB = await db.getThought(hash)
+
+  expect(thoughtDB).not.toBeUndefined()
+  expect(thoughtDB.contexts.length).toEqual(1)
+
+  expect(thoughtState.contexts[0].id).toEqual(thoughtDB.contexts[0].id)
+})
--- a/src/db.js
+++ b/src/db.js
@@ -9,6 +9,7 @@ import { timestamp } from './util'
 
 /** Returns true if the app is running as a test. */
 const isTest = () => process.env.NODE_ENV === 'test'
+
 if (isTest()) {
   Dexie.dependencies.indexedDB = require('fake-indexeddb')
   Dexie.dependencies.IDBKeyRange = require('fake-indexeddb/lib/FDBKeyRange')
@@ -26,9 +27,6 @@ const initHelpers = async () => {
 
 /** Initializes the database tables. */
 const initDB = async () => {
-
-  if (isTest()) return
-
   await db.version(1).stores({
     thoughtIndex: 'id, value, *contexts, created, lastUpdated',
     contextIndex: 'id, *context, lastUpdated',
@@ -82,22 +80,22 @@ export const getContextIndex = async () => {
 }
 
 /** Updates the recentlyEdited helper. */
-export const updateRecentlyEdited = async recentlyEdited => isTest() || db.helpers.update('EM', { recentlyEdited })
+export const updateRecentlyEdited = async recentlyEdited => db.helpers.update('EM', { recentlyEdited })
 
 /** Updates the schema version helper. */
-export const updateSchemaVersion = async schemaVersion => isTest() || db.helpers.update('EM', { schemaVersion })
+export const updateSchemaVersion = async schemaVersion => db.helpers.update('EM', { schemaVersion })
 
 /** Updates the lastUpdates helper. */
-export const updateLastUpdated = async lastUpdated => isTest() || db.helpers.update('EM', { lastUpdated })
+export const updateLastUpdated = async lastUpdated => db.helpers.update('EM', { lastUpdated })
 
 /** Gets all the helper values. */
 export const getHelpers = async () => db.helpers.get({ id: 'EM' })
 
 /** Updates the cursor helper. */
-export const updateCursor = async cursor => isTest() || db.helpers.update('EM', { cursor })
+export const updateCursor = async cursor => db.helpers.update('EM', { cursor })
 
 /** Deletes the cursor helper. */
-export const deleteCursor = async () => isTest() || db.helpers.update('EM', { cursor: null })
+export const deleteCursor = async () => db.helpers.update('EM', { cursor: null })
 
 /** Gets the full logs. */
 export const getLogs = async () => db.logs.toArray()
--- a/src/db.test.js
+++ b/src/db.test.js
@@ -2,6 +2,7 @@ import { store } from './store'
 import * as db from './db'
 import { initialize } from './initialize'
 import { hashThought } from './util'
+import { getThought } from './selectors'
 
 beforeAll(async () => {
   await initialize()
@@ -10,33 +11,15 @@ beforeAll(async () => {
 it('load settings into indexedDB on initialization', async () => {
   const hash = hashThought('Settings')
 
-  const thoughtState = store.getState().thoughts.thoughtIndex[hash]
+  const thoughtState = getThought(store.getState(), 'Settings')
 
   expect(thoughtState).not.toBeUndefined()
-  expect(thoughtState.contexts.length).toEqual(1)
+  expect(thoughtState.contexts).toHaveLength(1)
 
   const thoughtDB = await db.getThought(hash)
 
   expect(thoughtDB).not.toBeUndefined()
-  expect(thoughtDB.contexts.length).toEqual(1)
-
-  expect(thoughtState.contexts[0].id).toEqual(thoughtDB.contexts[0].id)
-})
-
-it('presist data on indexedDB after adding new thought', async () => {
-
-  await store.dispatch({ type: 'newThought', value: 'db-test' })
-  const hash = hashThought('db-test')
-
-  const thoughtState = store.getState().thoughts.thoughtIndex[hash]
-
-  expect(thoughtState).not.toBeUndefined()
-  expect(thoughtState.contexts.length).toEqual(1)
-
-  const thoughtDB = await db.getThought(hash)
-
-  expect(thoughtDB).not.toBeUndefined()
-  expect(thoughtDB.contexts.length).toEqual(1)
+  expect(thoughtDB.contexts).toHaveLength(1)
 
   expect(thoughtState.contexts[0].id).toEqual(thoughtDB.contexts[0].id)
 })
--- a/src/db.test.js
+++ b/src/db.test.js
@@ -1,6 +1,11 @@
-import { hashThought } from '../../util'
-import { store } from '../../store'
-import * as db from '../../db'
+import { store } from './store'
+import * as db from './db'
+import { initialize } from './initialize'
+import { hashThought } from './util'
+
+beforeAll(async () => {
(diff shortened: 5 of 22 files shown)


shresthabijay:
@raineorshine I made changes to the `setupFiles` and separated test utilities like `createTestApp` to separate file. I added a test called `db-test` to check db persistence. `Dexie` mock is working and all the tests are passing in my local system. But for some reason window object from `jsdom`  is not working properly for some files. It gives this error.
```js
TypeError: Cannot read property '_location' of null
```

However all the tests are passing in my local system. But tests is failing in the github check. Can you try it in your local system ?

raineorshine:
Unfortunately it is failing for me with the same error.

Do you have the same `jsdom` versions in your node_modules?

```
$ npm ls jsdom
em@143.0.0 /Users/raine/projects/em
├─┬ jest-environment-jsdom-sixteen@1.0.3
│ └── jsdom@16.2.2 <------------------------------- v16.2.2
└─┬ react-scripts@3.4.1
  ├─┬ jest@24.9.0
  │ └─┬ jest-cli@24.9.0
  │   └─┬ jest-config@24.9.0
  │     └─┬ jest-environment-jsdom@24.9.0
  │       └── jsdom@11.12.0 <---------------------- v11.12.0
  └─┬ jest-environment-jsdom-fourteen@1.0.1
    └── jsdom@14.1.0 <----------------------------- v14.1.0
```

It is confusing, but it should end up using v16 due to `jest-environment-jsdom-sixteen`.

shresthabijay:
@raineorshine Can you run just `db-test` and see if that passes ?

raineorshine:
Hmmm they are all passing now. `db-test` on its own, as well as all tests run together.

raineorshine:
You should still post the result you get from running `npm ls jsdom` please.

raineorshine:
I notice that locally `db-test` runs before everything else, but the failed GitHub Action runs it after the mount tests. Could that be causing the issue?

shresthabijay:
> You should still post the result you get from running `npm ls jsdom` please.

```js
─┬ jest-environment-jsdom-sixteen@1.0.3
│ └── jsdom@16.2.2 
└─┬ react-scripts@3.4.1
  ├─┬ jest@24.9.0
  │ └─┬ jest-cli@24.9.0
  │   └─┬ jest-config@24.9.0
  │     └─┬ jest-environment-jsdom@24.9.0
  │       └── jsdom@11.12.0 
  └─┬ jest-environment-jsdom-fourteen@1.0.1
    └── jsdom@14.1.0 
```

shresthabijay:

src/util/sync.ts
```diff
@@ -225,7 +223,7 @@ export const sync = (thoughtIndexUpdates = {}, contextIndexUpdates = {}, { local
     logWithTime('sync: localPromises complete')
 
     // firebase
-    if (isDocumentEditable() && remote) {
+    if (process.env.NODE_ENV !== 'test' && isDocumentEditable() && remote) {
```
raineorshine, line 226: If `remote` is true, `process.env.NODE_ENV !== 'test'` will always be true, so you can remove it.

src/db.js
```diff
@@ -9,6 +9,7 @@ import { timestamp } from './util'
 
 /** Returns true if the app is running as a test. */
 const isTest = () => process.env.NODE_ENV === 'test'
+
 if (isTest()) {
   Dexie.dependencies.indexedDB = require('fake-indexeddb')
   Dexie.dependencies.IDBKeyRange = require('fake-indexeddb/lib/FDBKeyRange')
```
raineorshine, line 15: Is it possible to move these into the test files? We should try to reduce test detection in the regular codebase if possible. I had put them here only because I could not get them working in the tests.

src/setupTests.js
```diff
@@ -1,33 +1,23 @@
...
+import { configure } from 'enzyme'
 import 'jest-localstorage-mock'
+import './store'
+import { act } from 'react-dom/test-utils'
+import { initialize } from './App'
 
-import App, { initialized } from './App'
+/*******************************************************
+ *    Global setup that runs before each test file     *
+ * ****************************************************/
```
raineorshine, line 10: When a former developer created `setupTests`, all of the tests used `mount`. Now that we only have a couple mount tests, I'm thinking all the logic in this file should actually belong in `createTestApp`. We don't want to `initialize` the app for non-mount tests. Does that seem right?

src/App.js
```diff
@@ -52,8 +52,10 @@ export const initialized = (async () => {
   initEvents()
 
   return localStateLoaded
+}
 
-})()
+// Note: Manually initialize during test environment
+if (process.env.NODE_ENV !== 'test') initialize()
```
raineorshine, line 58: What if we called `initialize` in onComponentMount of the `App` component? Then it would only load when the `App` itself loaded, and we could avoid test detection.

src/components/__tests__/Mount.js
```diff
@@ -27,7 +29,9 @@ it('create, navigate, and edit thoughts', async () => {
   // create top subthought
   windowEvent('keydown', { key: 'Enter', shiftKey: true, ctrlKey: true })
 
-  jest.runAllTimers()
+  act(() => {
+    jest.runOnlyPendingTimers()
+  })
```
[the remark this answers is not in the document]
shresthabijay: > Could you explain what `act` does here? I've never quite understood it.

`act` makes sure all state updates have been processed and applied to the DOM before assertion is made. Generally all the `api` provided by enzyme is already wrapped in `act`. But timeout functions happens outside the React callstack. So we need to wrap it with `act` our self when using jest fake timers. So basically `act` is being used here to make sure the state updates happening inside those timeout functions are processed before assertion is made.

You check out the docs for more details.
https://reactjs.org/docs/testing-recipes.html#act

Here is a great article on when to use `act`
https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning#1-when-using-jestusefaketimers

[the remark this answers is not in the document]
shresthabijay: > Also, what's the significance of using `runOnlyPendingTimers` vs `runAllTimers`?

It is used to prevent endless loop caused by recursive timer functions. I was getting these issues. 
https://jestjs.io/docs/en/timer-mocks#run-pending-timers

src/setupTests.js
```diff
@@ -1,33 +1,23 @@
-import React from 'react'
-import { act } from 'react-dom/test-utils'
 import Adapter from 'enzyme-adapter-react-16'
-import { configure, mount } from 'enzyme'
+import { configure } from 'enzyme'
 import 'jest-localstorage-mock'
+import './store'
```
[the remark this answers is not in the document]
shresthabijay: In some reducers test store wasn't getting initialized before store is subscribed or action is dispatched causing issues. So I imported store here to make sure it is initialized before every test file. I will make sure if it needed or not after the changes I have made. If not I will remove this.

src/components/__tests__/db-test.js
```diff
@@ -0,0 +1,37 @@
...
+import { store } from '../../store'
+import * as db from '../../db'
+
+it('load settings into indexedDB on initialization', async () => {
+  const hash = hashThought('Settings')
+
+  const thoughtState = store.getState().thoughts.thoughtIndex[hash]
+
+  expect(thoughtState).not.toBeUndefined()
+  expect(thoughtState.contexts.length).toEqual(1)
```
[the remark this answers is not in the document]
shresthabijay: Sure

src/components/__tests__/db-test.js
```diff
@@ -0,0 +1,37 @@
+import { hashThought } from '../../util'
+import { store } from '../../store'
+import * as db from '../../db'
+
+it('load settings into indexedDB on initialization', async () => {
+  const hash = hashThought('Settings')
+
+  const thoughtState = store.getState().thoughts.thoughtIndex[hash]
```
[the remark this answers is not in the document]
shresthabijay: > A better name for this variable would be `lexeme` or `settingsLexeme`.

Got it!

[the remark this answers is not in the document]
shresthabijay: > Always use the `getThought` selector for this. Never access `thoughtIndex` directly.

My bad. I will make the necessary changes.

src/db.js
shresthabijay: I was thinking the same.  Instead of directly creating `dexie` object inside `db`. Let's create a class and abstract all the db functions inside. Then we can inject `dexie` from the constructor. So we can easily pass `dexie` object with fake index db for the test. What do you think ?

src/components/__tests__/Mount.js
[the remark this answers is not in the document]
raineorshine: > > Could you explain what `act` does here? I've never quite understood it.
> 
> `act` makes sure all state updates have been processed and applied to the DOM before assertion is made. Generally all the `api` provided by enzyme is already wrapped in `act`. But timeout functions happens outside the React callstack. So we need to wrap it with `act` our self when using jest fake timers. So basically `act` is being used here to make sure the state updates happening inside those timeout functions are processed before assertion is made.
> 
> You check out the docs for more details.
> [reactjs.org/docs/testing-recipes.html#act](https://reactjs.org/docs/testing-recipes.html#act)
> 
> Here is a great article on when to use `act`
> [kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning#1-when-using-jestusefaketimers](https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning#1-when-using-jestusefaketimers)

Awesome, thanks!

> > Also, what's the significance of using `runOnlyPendingTimers` vs `runAllTimers`?
> 
> It is used to prevent endless loop caused by recursive timer functions. I was getting these issues.
> [jestjs.io/docs/en/timer-mocks#run-pending-timers](https://jestjs.io/docs/en/timer-mocks#run-pending-timers)

Interesting! That makes sense. I hadn't run into that yet.

src/setupTests.js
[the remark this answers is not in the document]
raineorshine: Most tests do not need the global `store` since they initialize state themselves and call reducers directly. Only some mount tests and some shortcut tests should require the global store. Even those should likely be refactored to accept a mock store via dependency injection.

src/db.js
raineorshine: I like the idea of not automatically creating the `dexie` object in `db.js`. I prefer not to use classes, but presumably we could do the same thing with a simple factory function.

Are you imagining `db.js` would still contain a singleton instance? It seems like this would be still be required wherever `db.js` is imported, since the db functions operate on the `db` instance.

src/setupTests.js
shresthabijay: Yes that makes sense. But we may need to `initialize` db for tests that use `store.dispatch` because it calls local db sync. So we should probably stop local db sync unless we need it. Is there any available way we can start application with local sync behavior set to desired state ?

raineorshine: > But we may need to `initialize` db for tests that use `store.dispatch` because it calls local db sync. 

I'm not sure I follow. What calls local db sync? And what is the connection between `store.dispatch` and initialize?

I do recognize that the db needs to be initialized for any tests that use the db though.

> So we should probably stop local db sync unless we need it. Is there any available way we can start application with local sync behavior set to desired state ?

By "stop" do you mean cancel something that has been started, or avoid starting it to begin with? Which part of the local db sync are you trying to cancel exactly? Is it the dispatch of [`loadLocalState` in App.js](https://github.com/cybersemics/em/blob/dev/src/App.js#L36)? If not, maybe you can post a link to the line number.

src/App.js
shresthabijay: `onComponentMount` will run after we do `wrapper.update`. It will cause `initialize` to run twice. The problem is that initialization logic runs as soon as we import component from `src/App.js`. We should separate all the initialization logic to a separate file and not export `App` component from the same file. We can call initialization logic in `index.js` and import and render `App` directly from the components folder. This way we can separately call and wait for initialization logic to complete in test and then mount `App` component.

raineorshine: Yes! I like that plan. Would you be willing to do that?

If I understand correctly, `initialize` would go into something like `/src/util/initialize.js`, `/src/index.js` would invoke `initialize`, and `/src/App.js` would be removed completely.

src/setupTests.js
shresthabijay: > By "stop" do you mean cancel something that has been started, or avoid starting it to begin with? Which part of the local db sync are you trying to cancel exactly? Is it the dispatch of [`loadLocalState` in App.js](https://github.com/cybersemics/em/blob/dev/src/App.js#L36)? If not, maybe you can post a link to the line number

I meant after `store.dispatch` is called with `thoughtIndex` and `contextIndex` updates then `syncQueue` will be flushed which will update data in local indexed db. Some test like shortcuts use global `store` unlike reducers tests. So if db is not initialized it will throw errors. So for such tests we must prevent sync. Does this make sense ?

src/setupTests.js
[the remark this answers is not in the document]
shresthabijay: > Most tests do not need the global `store` since they initialize state themselves and call reducers directly. Only some mount tests and some shortcut tests should require the global store. Even those should likely be refactored to accept a mock store via dependency injection.

That makes sense. We should use dependency injection to mock store too.

src/App.js
shresthabijay: > If I understand correctly, `initialize` would go into something like `/src/util/initialize.js`, `/src/index.js` would invoke `initialize`, and `/src/App.js` would be removed completely.

Yes.

shresthabijay: > Yes! I like that plan. Would you be willing to do that?

Sure. I can do that.

src/setupTests.js
raineorshine: Yes, that makes sense. I will create a separate issue for this refactor. Use `store` as-is/as-needed in this PR. Thanks!

raineorshine: #781

src/setupTests.js
```diff
@@ -1,33 +1,23 @@
...
-    jest.useFakeTimers()
-
-    // wait for app to be initialized
-    // specifically for initialSettings to be loaded via loadLocalState
-    await initialized
+beforeAll(async () => {
+  jest.useFakeTimers()
+  // initialize db and window events before every test
+  await initialize()
+})
```
raineorshine, line 18: Unless I'm missing something, we still need to move `initialize` out into individual tests that need it, as we don't want it to run for every test. This is independent from mocking `store` per our previous discussion (#781).

shresthabijay:
@raineorshine I have moved initialize to tests file that need it instead of calling it before every test file. Separated db-test into a separate `db.test.js`. The new changes from the `dev` was causing some warnings. Turns out `window.scrollTo` was not implemented , so I added a `noop` function as `window.scrollTo`.

shresthabijay:
@raineorshine I ran into some sort of circular dependency issue. It goes away on importing `store` in `setupTests.js`. We don't this in `dev` because we import `App` inside `setupTests.js` which loads `store` before any `utils` have been imported directly from `./util/index.js`.

Import any `util` function before `App` is imported in `setupTest.js` in `dev`. Then run some tests. You will get this error.

```js
TypeError: Cannot read property 'parseJsonSafe' of undefined

      at Object.get [as parseJsonSafe] (src/util/index.ts:387:27)
      at initialState (src/util/initialState.ts:107:20)
      at appReducer (src/reducers/app.ts:8:36)
      at dispatch (node_modules/redux/lib/redux.js:219:22)
      at createStore (node_modules/redux/lib/redux.js:305:3)
      at node_modules/redux/lib/redux.js:635:31
      at createStore (node_modules/redux/lib/redux.js:85:33)
      at Object.<anonymous> (src/store.js:14:22)
      at Object.<anonymous> (src/selectors/canShowModal.ts:1:1)
      at Object.<anonymous> (src/selectors/index.ts:4:1)
      at Object.<anonymous> (src/util/addThought.ts:3:1)
      at Object.<anonymous> (src/util/index.ts:3:1)
      at Object.<anonymous> (src/util/__tests__/splice.js:1:1)
```

I think this a circular dependency error because some `utils` like `addThought` import some `selectors`. It causes all the `selectors` to load. Since `selectors` use `utils` too it causes this error.  Since `store` somewhere in it's dependency tree imports a util `initialState` directly from it's file `./util/initialState` instead from `./util/index.js`, it resolves this issue. It pretty confusing. Have you encountered this before ?

raineorshine:
The reason is fails on `parseJsonSafe` is because it is the [first util function that is called in `initialState`](https://github.com/cybersemics/em/blob/dev/src/util/initialState.ts#L107). Importing and using `initialState` in the main [appReducer](https://github.com/cybersemics/em/blob/dev/src/reducers/app.ts#L8) is where the entry point seems to be. If you remove `initialState` there, the error goes away.

Importing `store` in `canShowModal` was also causing problems, which I fixed.

Importing the util functions individually, and thus not importing the util index file, seems to fix the issue. It's preferable to importing `store` in `setupTests`, but it still feels like a hack. I wonder if extricating the global `store` from other util functions might get closer to the root of the issue.

raineorshine:
Created #784 to track the circular dependency issue separately.

src/db.test.js
```diff
@@ -0,0 +1,42 @@
...
+  expect(thoughtState).not.toBeUndefined()
+  expect(thoughtState.contexts.length).toEqual(1)
+
+  const thoughtDB = await db.getThought(hash)
+
+  expect(thoughtDB).not.toBeUndefined()
+  expect(thoughtDB.contexts.length).toEqual(1)
+
+  expect(thoughtState.contexts[0].id).toEqual(thoughtDB.contexts[0].id)
+})
```
raineorshine, line 42: This test appears to be identical to the first test.

src/db.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+  const thoughtState = store.getState().thoughts.thoughtIndex[hash]
+
+  expect(thoughtState).not.toBeUndefined()
+  expect(thoughtState.contexts.length).toEqual(1)
+
+  const thoughtDB = await db.getThought(hash)
+
+  expect(thoughtDB).not.toBeUndefined()
+  expect(thoughtDB.contexts.length).toEqual(1)
```
raineorshine, line 21: Use `toHaveLength`

src/db.test.js
shresthabijay: The first one checks if the Settings gets synced in the db on load. The second one checks if a new thought added gets synced to the db. They are kinda identical. I will remove the later one.

PR #779, opened 2020-07-05
Outdent on delete

### LINKED ISSUE #772: Hit delete at beginning of only child to outdent
- [ ] Unit Test

When the caret is located at `offset: 0` of a thought which is the only child of a context, then hitting delete should outdent the thought. 

This provides a way to “undo” indent via space in at least some circumstances, without interfering with other commands.

shresthabijay:
fixes #772


shresthabijay:

src/shortcuts/__tests__/deleteEmptyThoughtOrOutdent.js
```diff
@@ -0,0 +1,74 @@
...
+  await store.dispatch({ type: 'modalComplete', id: 'welcome' })
+  await store.dispatch({ type: 'tutorial', value: false })
+
+  // import thoughts
+  await store.dispatch(importText(RANKED_ROOT, `
+  - a
+    - b
+      - c`))
+
+  await store.dispatch({ type: 'setCursor', thoughtsRanked: [
```
raineorshine, line 24: This dispatch is also synchronous.

src/shortcuts/deleteEmptyThoughtOrOutdent.js
```diff
@@ -60,6 +63,33 @@ const exec = (dispatch, getState) => {
   dispatch({ type: 'deleteEmptyThought' })
 }
 
+// eslint-disable-next-line jsdoc/require-jsdoc
+const canExecuteOutdent = getState => {
```
raineorshine, line 67: JSDOC should only be disabled in rare cases, such as for extremely trivial functions or canonical functions like `mapStateToProps`.

src/shortcuts/deleteEmptyThoughtOrOutdent.js
```diff
@@ -21,7 +24,7 @@ import {
 } from '../selectors'
 
 // eslint-disable-next-line jsdoc/require-jsdoc
-const canExecute = getState => {
+const canExecuteDeleteEmptyThought = getState => {
```
raineorshine, line 27: Subtle point, but I would recommend passing `state` directly here, as then the function better conforms to a selector signature.

src/shortcuts/deleteEmptyThoughtOrOutdent.js
```diff
@@ -60,6 +63,33 @@ const exec = (dispatch, getState) => {
...
+  return isDocumentEditable() && headValue(cursor).length !== 0 && offset === 0 && filteredChildren().length === 1
+}
+
+// eslint-disable-next-line jsdoc/require-jsdoc
+const canExecute = getState => canExecuteOutdent(getState) || canExecuteDeleteEmptyThought(getState)
+
+// eslint-disable-next-line jsdoc/require-jsdoc
+const exec = (dispatch, getState) => {
+  if (canExecuteOutdent(getState)) dispatch({ type: 'outdent' })
+  else deleteEmptyThought(dispatch, getState)
```
raineorshine, line 90: This deserves a comment indicating that `canExecuteDeletEmptyThought` has already been checked in `canExecute`.

raineorshine: It can also be called in the conventional thunk action-creator form:

```js
dispatch(deleteEmptyThought)
```

src/shortcuts/deleteEmptyThoughtOrOutdent.js
raineorshine: Now that I think about it, I think `shortcuts` should pass `state` directly so that `canExecute` also matches the selector signature. That doesn't have to be done in this PR.

raineorshine:
I addressed the comments in b60eed2. Feel free to take a look.

Thanks!

PR #783, opened 2020-07-08
Drag Error Divider Fix

### LINKED ISSUE #777: Drag error: "Cannot move the em context to another context"
## Steps to Reproduce

- a
  - b
  - c
  - --- (divider)

Drag `b` in between `c` and `---`.

## Current Behavior

The highlighted drop target appears too far to the left. Then when dropped, there is the message: "Cannot move the em context to another context."

## Expected Behavior

The drop target should appear in between `c` and `---` and allow `b` to be dropped there.

shresthabijay:
fixes #777

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -2455,6 +2455,10 @@ code, .code {
   margin-left: -125px;
 }
 
+.child-divider .drop-hover {
+  margin-left: 0.4em;
+}
+
 .child-divider .bullet,
 .child-divider .bullet-cursor-overlay
 {
--- a/src/App.css
+++ b/src/App.css
@@ -2456,7 +2456,7 @@ code, .code {
 }
 
 .child-divider .drop-hover {
-  margin-left: 0.4em;
+  margin-left: 6px;
 }
 
 .child-divider .bullet,
--- a/src/App.css
+++ b/src/App.css
@@ -2095,6 +2095,10 @@ code, .code {
   padding-bottom: 1.1em;
 }
 
+.empty-children .drop-end.inside-divider {
+  margin-left: 60px;
+}
+
 .mobile .empty-children .drop-end {
   margin-left: 25%;
 }
@@ -2456,6 +2460,10 @@ code, .code {
 }
 
 .child-divider .drop-hover {
+  margin-left: 6px;
+}
+
+.inside-divider .drop-hover {
   margin-left: 6px;
 }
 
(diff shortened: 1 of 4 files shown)


shresthabijay:
@raineorshine The issue was that the margin left of the `Divider` component didn't match that `Editable` component. So I added required margin left to the `drop-hover` if thought is a divider. The cannot move error was because there was a condition to prevent drop if the thought is a divider. Please review the changes. Thanks!

shresthabijay:
> I would advise more caution around removing branches of logic. It's usually there for a reason. This one was added in [28b004c](https://github.com/cybersemics/em/commit/28b004c9ecd16fdc6b59ee0f1b4830fac27fe1b7), although it appears that perhaps the condition was only needed in `Subthoughts`. Until you understand the reason it was put there, you don't know for sure that you're not removing something that is needed. I recommend the use of git blame in the GitHub interface to find the commit in which a line of code was added. That can reveal the purpose and help you better evaluate its necessity.
> 
> If you are aware of this, my apologies. I read, perhaps wrongly, from your comment that you had not identified when the code was added and for what purpose.

I was aware of this. This condition was added to `Thoughts` to prevent dragging thoughts into divider. But that condition is only necessary for drop target inside  `Subthought` component.

raineorshine:
> I was aware of this. This condition was added to `Thoughts` to prevent dragging thoughts into divider. But that condition is only necessary for drop target inside `Subthought` component.

Great! When you describe the cause ("there was a condition to prevent drop if the thought is a divider") just go one step further and describe the reason why it is not needed. Otherwise I have to do the work of tracking it down that perhaps you have already done. Thanks!

shresthabijay:
> Great! When you describe the cause ("there was a condition to prevent drop if the thought is a divider") just go one step further and describe the reason why it is not needed. Otherwise I have to do the work of tracking it down that perhaps you have already done. Thanks!

I will keep that in mind.

shresthabijay:
> It looks like there is a slightly different case that still causes the error. In the following, drag `c` under the divider and slightly to the right:

I will look into it.

shresthabijay:
@raineorshine There are three drop targets responsible for showing `drop-hover`. With the `Divider` component the `drop-hover` associated with `EmptyChildrenDropTarget` doesn't have enough margin left to appear as drop target to add subthought. So it was causing confusion. I made changes to add the margin. But with `Divider` component we do not want to activate `drop-target` associated with `EmptyChildrenDropTarget` because we don't allow to drag into `Divider` thoughts anyways. So I made necessary changes to the `canDrop` logic to prevent to prevent drag if thought is a divider. Please review the changes. Thanks!

raineorshine:
Great! That was tricky. Thank you for working through the details and getting to a solution. Excellent work.

PR #787, opened 2020-07-11
Separate test store for shortcuts test

### LINKED ISSUE #781: Mock store in tests
This issue is about mocking `store` to make the tests cleaner.

## Previous Dialog

**raineorshine**
When a former developer created setupTests, all of the tests used mount. Now that we only have a couple mount tests, I'm thinking all the logic in `setupTests` should actually belong in `createTestApp`. We don't want to initialize the app for non-mount tests. Does that seem right?

**shresthabijay**
Yes that makes sense. But we may need to initialize db for tests that use `store.dispatch` because it calls local db sync. So we should probably stop local db sync unless we need it. Is there any available way we can start application with local sync behavior set to desired state ?

**raineorshine**
> But we may need to initialize db for tests that use `store.dispatch` because it calls local db sync.

I'm not sure I follow. What calls local db sync? And what is the connection between `store.dispatch` and initialize?

I do recognize that the db needs to be initialized for any tests that use the db though.

> So we should probably stop local db sync unless we need it. Is there any available way we can start application with local sync behavior set to desired state ?

By "stop" do you mean cancel something that has been started, or avoid starting it to begin with? Which part of the local db sync are you trying to cancel exactly? Is it the dispatch of loadLocalState in App.js? If not, maybe you can post a link to the line number.

**shresthabijay**
> By "stop" do you mean cancel something that has been started, or avoid starting it to begin with? Which part of the local db sync are you trying to cancel exactly? Is it the dispatch of loadLocalState in App.js? If not, maybe you can post a link to the line number.

I meant after `store.dispatch` is called with thoughtIndex and contextIndex updates then `syncQueue` will be flushed which will update data in local indexed db. Some test like shortcuts use global store unlike reducers tests. So if db is not initialized it will throw errors. So for such tests we must prevent sync.

shresthabijay:
fixes #781

### DIFF
--- a/src/shortcuts.js
+++ b/src/shortcuts.js
@@ -86,7 +86,7 @@ let handleGestureSegmentTimeout // eslint-disable-line fp/no-let
 /**
  * Keyboard handlers factory function.
  */
-export const keyboardHandlers = store => ({
+export const inputHandlers = store => ({
 
   /** Handles gesture hints when a valid segment is entered. */
   handleGestureSegment: (g, sequence, e) => {
--- a/src/shortcuts.js
+++ b/src/shortcuts.js
@@ -84,7 +84,7 @@ const isGestureHint = state => state.alert && state.alert.alertType === 'gesture
 let handleGestureSegmentTimeout // eslint-disable-line fp/no-let
 
 /**
- *
+ * Keyboard handlers factory function.
  */
 export const keyboardHandlers = store => ({
 
--- a/src/shortcuts.js
+++ b/src/shortcuts.js
@@ -2,7 +2,6 @@
 
 import Emitter from 'emitter20'
 import { isMac } from './browser'
-import { store } from './store'
 import globals from './globals'
 import { alert, suppressExpansion } from './action-creators'
 import { GESTURE_SEGMENT_HINT_TIMEOUT } from './constants'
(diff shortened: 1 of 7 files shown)


shresthabijay:
@raineorshine I have exported a factory function that takes `store` as param and encapsulates shortcut handlers. I have made necessary changes to the places where these handlers were used. I also created a `createTestStore` function that helps create new store for every shortcut tests. Please review the changes. Thanks!

src/shortcuts.js
```diff
@@ -84,100 +83,106 @@ const isGestureHint = state => state.alert && state.alert.alertType === 'gesture
...
-    if (isGestureHint(store.getState())) {
-      alert(null)
+/**
+ * Keyboard handlers factory function.
+ */
+export const keyboardHandlers = store => ({
+
+  /** Handles gesture hints when a valid segment is entered. */
+  handleGestureSegment: (g, sequence, e) => {
+
```
raineorshine, line 93: Since `handleGestureSegment` and `handleGestureEnd` are not keyboard inputs, a better name would be `inputHandlers`.

src/shortcuts/__tests__/newThoughtOrOutdent.js
```diff
@@ -1,10 +1,12 @@
...
-import { keyDown } from '../../shortcuts'
+import { keyboardHandlers } from '../../shortcuts'
 import { importText } from '../../action-creators'
 import { RANKED_ROOT, ROOT_TOKEN } from '../../constants'
 import { exportContext } from '../../selectors'
 import { noop } from 'lodash'
 import { initialize } from '../../initialize'
+import { createTestStore } from '../../test-helpers/createTestStore'
+
+const store = createTestStore()
```
raineorshine, line 9: This should probably be assigned in `beforeEach`, otherwise data could carry over between tests.

src/shortcuts.js
shresthabijay: Sure!

shresthabijay:
@raineorshine I have removed `initialize` from the shortcut tests and create new store within the scope of each test. I renamed `keyboardHandlers` to `inputHandlers`. Please review the changes. Thanks!

PR #788, opened 2020-07-12
Archive thought from context

### LINKED ISSUE #737: deleteThought: Delete thought from context.
Test: ed27494a
Branch: `737`

pushkov-fedor:
Fixes #737 

@raineorshine, hi! I'm not really familiar with archive logic, but this issue needs to fix `deleteThought` and `archiveThought` reducers. I decided convert wrong path in **Context View** to correct path in **Normal View** and let remaining logic to do its job. As i get it, reducers works correctly for **Normal View**. Is it correct approach? 

This issue belongs to `deleteThought` reducer, but deleting logic is replaced with archiving now. I have feeling that you forgot about it or maybe i don't understand something? Correct me please! Deleting/Archiving logic from context view works and tests are passed, but after all steps in test code we are having archived thought in **Context View**. It doesn't look like something i expect to see in **Context View.** Is it okay?
Screenshot from 2020-07-12 17-33-19

### DIFF
--- a/src/reducers/archiveThought.js
+++ b/src/reducers/archiveThought.js
@@ -31,7 +31,6 @@ import {
   lastThoughtsFromContextChain,
   nextSibling,
   prevSibling,
-  rankThoughtsFirstMatch,
   splitChain,
   thoughtsEditingFromChain,
 } from '../selectors'
@@ -57,21 +56,21 @@ const archiveThought = (state, { path }) => {
 
   // same as in newThought
   const showContexts = isContextViewActive(state, contextOf(path))
-  if (showContexts) {
-    // Get thought in ContextView
-    const thoughtInContextView = head(contextOf(path))
-    // Get context from which we are going to delete thought
-    const context = pathToContext(_.last(splitChain(state, path)))
-    if (context) {
-      // Convert to path
-      path = rankThoughtsFirstMatch(state, [...context, thoughtInContextView.value])
-    }
-  }
+  // if (showContexts) {
+  //   // Get thought in ContextView
+  //   const thoughtInContextView = head(contextOf(path))
+  //   // Get context from which we are going to delete thought
+  //   const context = pathToContext(_.last(splitChain(state, path)))
+  //   if (context) {
+  //     // Convert to path
+  //     path = rankThoughtsFirstMatch(state, [...context, thoughtInContextView.value])
+  //   }
+  // }
   const contextChain = splitChain(state, path)
   const thoughtsRanked = contextChain.length > 1
     ? lastThoughtsFromContextChain(state, contextChain)
     : path
-  const context = showContexts ? pathToContext(contextOf(path)) : pathToContext(showContexts && contextChain.length > 1 ? contextChain[contextChain.length - 2]
+  const context = pathToContext(showContexts && contextChain.length > 1 ? contextChain[contextChain.length - 1]
     : !showContexts && thoughtsRanked.length > 1 ? contextOf(thoughtsRanked) :
     RANKED_ROOT)
 
@@ -96,16 +95,24 @@ const archiveThought = (state, { path }) => {
     }
   }
 
+  /** Gets the next sibling context in the context view. */
+  const nextContext = () => {
+    const contextsSortedAndRanked = getContextsSortedAndRanked(state, headValue(contextOf(path)))
+    const contextsFiltered = contextsSortedAndRanked
+      .filter(({ context }) => head(context) !== '=archive')
+      .map((context, i) => Object.assign({}, context, { rank: i }))
+    return contextsFiltered.find(c => head(c.context) !== head(context))
+  }
+
   // prev must be calculated before dispatching existingThoughtDelete
   const prev = showContexts
     ? prevContext()
     : prevSibling(state, value, context, rank)
 
   const next = !prev && showContexts
-    ? getContextsSortedAndRanked(state, headValue(contextOf(path)))[0]
+    ? nextContext()
     // get first visible thought
     : nextSibling(state, value, context, rank)
-
   const [cursorNew, offset] =
     // Case I: set cursor on prev thought
     prev ? [contextOf(path).concat(prev), prev.value.length] :
@@ -132,7 +139,7 @@ const archiveThought = (state, { path }) => {
 
     isDeletable
       ? existingThoughtDelete({
-        context: contextOf(pathToContext(thoughtsRanked)),
+        context: showContexts ? context : contextOf(pathToContext(thoughtsRanked)),
         showContexts,
         thoughtRanked: head(thoughtsRanked),
       })
@@ -151,7 +158,7 @@ const archiveThought = (state, { path }) => {
 
         // undo alert
         alert({
-          value: <div>Deleted "{ellipsize(headValue(path))}"&nbsp;
+          value: <div>Deleted "{ellipsize(headValue(showContexts ? thoughtsRanked : path))}"&nbsp;
             <a onClick={() => {
               store.dispatch({
                 type: 'undoArchive',
@@ -168,8 +175,8 @@ const archiveThought = (state, { path }) => {
 
         // execute existingThoughtMove after newThought has updated the state
         state => existingThoughtMove(state, {
-          oldPath: path,
-          newPath: pathToArchive(state, path, context),
+          oldPath: showContexts ? thoughtsRanked : path,
+          newPath: pathToArchive(state, showContexts ? thoughtsRanked : path, context),
           offset
         })
       ])
--- a/src/reducers/archiveThought.js
+++ b/src/reducers/archiveThought.js
@@ -63,7 +63,6 @@ const archiveThought = (state, { path }) => {
   const context = pathToContext(showContexts && contextChain.length > 1 ? contextChain[contextChain.length - 1]
     : !showContexts && thoughtsRanked.length > 1 ? contextOf(thoughtsRanked) :
     RANKED_ROOT)
-  console.log('thoughtsRanked: ', thoughtsRanked)
   const { value, rank } = head(thoughtsRanked)
   const thoughts = pathToContext(thoughtsRanked)
 
--- a/src/reducers/archiveThought.js
+++ b/src/reducers/archiveThought.js
@@ -24,6 +24,7 @@ import {
 
 // selectors
 import {
+  getContexts,
   getContextsSortedAndRanked,
   getThoughts,
   hasChild,
@@ -31,6 +32,7 @@ import {
   lastThoughtsFromContextChain,
   nextSibling,
   prevSibling,
+  rankThoughtsFirstMatch,
   splitChain,
   thoughtsEditingFromChain,
 } from '../selectors'
@@ -55,8 +57,18 @@ const archiveThought = (state, { path }) => {
   if (!path) return state
 
   // same as in newThought
-  const contextChain = splitChain(state, path)
   const showContexts = isContextViewActive(state, contextOf(path))
+  if (showContexts) {
+    // Get thought in ContextView
+    const thoughtInContextView = head(contextOf(path))
+    // Get context from which we are going to delete thought
+    const context = getContexts(state, thoughtInContextView.value).map(({ context }) => context).find(context => head(context) === headValue(path))
+    if (context) {
+      // Convert to path
+      path = rankThoughtsFirstMatch(state, [...context, thoughtInContextView.value])
+    }
+  }
+  const contextChain = splitChain(state, path)
   const thoughtsRanked = contextChain.length > 1
     ? lastThoughtsFromContextChain(state, contextChain)
     : path
--- a/src/reducers/archiveThought.js
+++ b/src/reducers/archiveThought.js
@@ -24,7 +24,6 @@ import {
 
 // selectors
 import {
-  getContexts,
   getContextsSortedAndRanked,
   getThoughts,
   hasChild,
@@ -45,7 +44,6 @@ import {
   newThought,
   setCursor,
 } from '../reducers'
-import { last } from 'lodash'
 
 /** Moves the thought to =archive. If the thought is already in =archive, permanently deletes it.
  *
@@ -63,7 +61,7 @@ const archiveThought = (state, { path }) => {
     // Get thought in ContextView
     const thoughtInContextView = head(contextOf(path))
     // Get context from which we are going to delete thought
-    const context = pathToContext(last(splitChain(state, path)))
+    const context = pathToContext(_.last(splitChain(state, path)))
     if (context) {
       // Convert to path
       path = rankThoughtsFirstMatch(state, [...context, thoughtInContextView.value])
--- a/src/reducers/archiveThought.js
+++ b/src/reducers/archiveThought.js
@@ -56,16 +56,6 @@ const archiveThought = (state, { path }) => {
 
   // same as in newThought
   const showContexts = isContextViewActive(state, contextOf(path))
-  // if (showContexts) {
-  //   // Get thought in ContextView
-  //   const thoughtInContextView = head(contextOf(path))
-  //   // Get context from which we are going to delete thought
-  //   const context = pathToContext(_.last(splitChain(state, path)))
-  //   if (context) {
-  //     // Convert to path
-  //     path = rankThoughtsFirstMatch(state, [...context, thoughtInContextView.value])
-  //   }
-  // }
   const contextChain = splitChain(state, path)
   const thoughtsRanked = contextChain.length > 1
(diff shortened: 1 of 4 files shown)


src/reducers/archiveThought.js
```diff
@@ -54,8 +56,18 @@ const archiveThought = (state, { path } = {}) => {
...
-  const contextChain = splitChain(state, path)
   const showContexts = isContextViewActive(state, contextOf(path))
+  if (showContexts) {
+    // Get thought in ContextView
+    const thoughtInContextView = head(contextOf(path))
+    // Get context from which we are going to delete thought
+    const context = getContexts(state, thoughtInContextView.value).map(({ context }) => context).find(context => head(context) === headValue(path))
+    if (context) {
+      // Convert to path
+      path = rankThoughtsFirstMatch(state, [...context, thoughtInContextView.value])
```
raineorshine, line 67: In general, we can't convert a Context to a Path since the rank information has been lost. `rankThoughtsFirstMatch` infers the rank, choosing the first in a context if there are duplicates, which is not suitable for an exact path. `rankThoughtsFirstMatch` is used to infer the `cursor` from the URL and should not be used for any other purpose.

raineorshine:
> This issue belongs to deleteThought reducer, but deleting logic is replaced with archiving now.

Ah, you are correct! I did not take into account archive view when I wrote this test. Would you be interested in re-writing the test and then we can go from there? The thought should be archived rather than deleted from the context.

pushkov-fedor:
> > This issue belongs to deleteThought reducer, but deleting logic is replaced with archiving now.
> 
> Ah, you are correct! I did not take into account archive view when I wrote this test. Would you be interested in re-writing the test and then we can go from there? The thought should be archived rather than deleted from the context.

Sure, will do it!

pushkov-fedor:
@raineorshine, i found another issue regarding archiving.

After deleting **m** thought from test case, i expect to see `=archive` in` [__ROOT__, b]` context, but empty archive creates in `[__ROOT__]`. Also Context View of **m** shows that **m** are inside `[b, =archive]`, but =archive in [__ROOT__, b] doesn't renders.
Screenshot from 2020-07-12 17-33-19
If we add another thought in `[__ROOT__, b]` and delete it, archive will appear in` [__ROOT__,b]` containing **m** and a thought that we created and deleted.

# Shortly:
- a
  - m
    - x
- b
  - m

Set cursor on [a, m]. Toggle ContextView. Delete m in context of b.

# Expected:
- a
  - m
    - x
- b
  -  =archive
     - m

# Current:
- =archive(empty)
- a
  - m
    - x
- b
***  
Add a thought to [b]
- =archive(empty)
- a
  - m
    - x
- b
  - thought
***
Delete **thought** in normal view
# What we get
- =archive(empty)
- a
  - m
    - x
- b
  - =archive
    - thought
    - m

raineorshine:
> After deleting **m** thought from test case, i expect to see `=archive` in` [__ROOT__, b]` context

ROOT is a special case, and thus is always excluded from Contexts. Thus `=archive` should be added to `['b']`, technically. `[ROOT, b]` is not a valid Context. You can use [`unroot`](https://github.com/cybersemics/em/blob/dev/src/util/unroot.ts) to remove ROOT from the beginning of a Context or Path.

It seems like that problem may account for the later issues, but correct me if I'm wrong.

Also your test looks correct so I think you have the right starting point.

pushkov-fedor:
Raine, i would like to clarify some cases that could happen. Imagine we have tree from example and toggle Context View on **m**. If we delete **b** context of **m**, we will delete **m** from b context. If we delete child **x** from Context View ([a, x]), we will simply delete a child :) If we delete context **a** of **m**, we will delete **m** from context **a**, but what will happen with child **x**?

src/reducers/archiveThought.js
pushkov-fedor: My approach was to re-create Path from Context View to Path in Normal View. As i understand, we can't do it, because we lose rank information, so do i need to fix this issue in another way?

pushkov-fedor: I've tried this case:
- parentA
  - a
    - m
      - x
- parentB
  - b
    - m

In Context View we have [parentA, a] and [parentB, b].
We get path from cursor and in this case we don't have **parentB** there (the same problen as in https://github.com/cybersemics/em/pull/800). 
Screenshot from 2020-07-15 18-03-25
We need either to restore path in Normal View from what we get or pass some additional data to cursor/reoragnize cursor for Context View in order to restore path from it. What do you think about it? If i'm kind of lost in path/cursor logic or forgot about some util functions that could help, please tell me :) Thanks!

pushkov-fedor:
Archiving works in some way even with `rankThoughtsFirstMatch` for tree from issue, but tests are failed, because `=archive` thought  isn't created (as i described above). I will find out what the reason of such behavior.

raineorshine:
> Raine, i would like to clarify some cases that could happen. Imagine we have tree from example and we toggle Context View on **m**. 

The Context View applies to a Context, not a Child or ThoughtContext, so you should specify whether you mean `[a, m]` or `[b, m]`.

> If we delete **b** context of **m**, we will delete **m** from b context. 

Correct. To be more precise: If we activate `archiveThought` on `a/m~/b`, then it will archive `m` from `[b]` context.

> If we delete child **x** from Context View ([a, x]), we will simply delete a child :) 

Correct, it will delete Child `x` from Context `[a, m]`.

> If we delete context **a** of **m**, we will delete **m** from context **a**, but what will happen with child **x**?

This should be the same as the behavior of archiving `m` when in Normal View of `[a]`. Thus, it should archive all descendants (`a/m/x`).

**Before:**

- a
  - m
    - x
- b
  - m

**After:**

- a
  - m
    - x
- b
  - m

Note that you should be specific as to whether you mean delete `a` from Context `a/m` or from Context `b/m`, which may result in different `cursor` behavior due to the `Path` structure.

src/reducers/archiveThought.js
raineorshine: Gosh, this is difficult to talk about!

> We get path from cursor

The `cursor` *is* a `Path`. It is a transhierarchical `Path` in that it may cross multiple Context View boundaries.

> We get path from cursor and in this case we don't have **parentB** there (the same problen as in #800).

If I understand correctly, you're saying that when the `cursor` is `parentA/a/m~/b` then **parentB** is not explicitly stored in the `Path`, even though it is implicitly represented in the last simple (monohierarchical) `Path` of the `cursor`: `parentB/b/m`. Because it is not explicitly stored, we don't have all of the rank information.

> We need either to restore path in Normal View from what we get or pass some additional data to cursor/reoragnize cursor for Context View in order to restore path from it.

I'm a little lost at this point, but maybe this is because I lost the original context of this discussion.

The `contextChain` contains the full `Path` information for each segment in a transhierarchical `Path`. Thus, by deriving `contextChain` from `cursor`, you can recover the `Child` at any point in the `Path`.

Uh oh... `lastThoughtsFromContextChain` itself uses `rankThoughtsFirstMatch`. I wrote that function so I have nobody to blame but myself!

raineorshine:
I think the bug must be related to the `showContexts` logic in `archiveThought`. I suggest the following approach:

1. Set the cursor to `b/m` (Normal View) and activate `archiveThought`.
2. Observe the exact arguments sent in the [call to `newThought`](https://github.com/cybersemics/em/blob/dev/src/reducers/archiveThought.js#L130-L139) and the [call to `existingThoughtMove`](https://github.com/cybersemics/em/blob/dev/src/reducers/archiveThought.js#L158-L163) within the `reducerFlow` composition.

Now repeat these steps archiving the same thought from within the Context View (`a/m~/b`) and see how the arguments to `newThought` and `existingThoughtMove` are different. They should be the same.

src/reducers/archiveThought.js
raineorshine: @pushkov-fedor I can address the refactoring of `lastThoughtsFromContextChain` in a separate issue. We should still avoid all explicit uses of `rankThoughtsFirstMatch` here.

src/reducers/deleteThought.ts
```diff
@@ -40,8 +42,18 @@ const deleteThought = (state: State, { path }: { path?: Path }) => {
...
-  const contextChain = splitChain(state, path)
   const showContexts = isContextViewActive(state, pathToContext(contextOf(path)))
+  if (showContexts) {
+    // Get thought in ContextView
+    const thoughtInContextView = head(contextOf(path))
+    // Get context from which we are going to delete thought
+    const context = getContexts(state, thoughtInContextView.value).map(({ context }) => context).find(context => head(context) === headValue(path!))
+    if (context) {
+      // Convert to path
+      path = rankThoughtsFirstMatch(state, [...context, thoughtInContextView.value])
```
raineorshine, line 53: Let's avoid the use of `rankThoughtsFirstMatch`.

raineorshine:
The cursor after archiving should behave the same as in Normal View. I added failing tests.

src/reducers/archiveThought.js
```diff
@@ -55,12 +55,22 @@ const archiveThought = (state, { path }) => {
...
+  // if (showContexts) {
+  //   // Get thought in ContextView
+  //   const thoughtInContextView = head(contextOf(path))
+  //   // Get context from which we are going to delete thought
+  //   const context = pathToContext(_.last(splitChain(state, path)))
+  //   if (context) {
+  //     // Convert to path
+  //     path = rankThoughtsFirstMatch(state, [...context, thoughtInContextView.value])
+  //   }
+  // }
```
raineorshine, line 68: Delete unused comments

src/reducers/archiveThought.js
```diff
@@ -85,16 +95,24 @@ const archiveThought = (state, { path }) => {
     }
   }
 
+  /** Gets the next sibling context in the context view. */
+  const nextContext = () => {
+    const contextsSortedAndRanked = getContextsSortedAndRanked(state, headValue(contextOf(path)))
+    const contextsFiltered = contextsSortedAndRanked
+      .filter(({ context }) => head(context) !== '=archive')
+      .map((context, i) => Object.assign({}, context, { rank: i }))
+    return contextsFiltered.find(c => head(c.context) !== head(context))
```
raineorshine, line 104: Rather than doing `filter` and `map` first, combine the logic into `find`. This saves unnecessary iterations when the matching context is at the beginning of the list.

src/reducers/deleteThought.ts
```diff
@@ -35,8 +37,18 @@ const deleteThought = (state: State, { path }: { path?: Path }) => {
   if (!path) return state
 
   // same as in newThought
-  const contextChain = splitChain(state, path)
   const showContexts = isContextViewActive(state, pathToContext(contextOf(path)))
+  if (showContexts) {
+    // Get thought in ContextView
+    const thoughtInContextView = head(contextOf(path))
+    // Get context from which we are going to delete thought
+    const context = getContexts(state, thoughtInContextView.value).map(({ context }) => context).find(context => head(context) === headValue(path!))
```
raineorshine, line 45: Similar: It is inefficient to `map` across the entire array when `find` naturally short circuits. I recommend combining into `find`.

raineorshine:
Excellent work. The Context View is extremely tricky, and you found the places that it needed to be detected and the logic changed to support it.

Just a few requests to clean up the code.

pushkov-fedor:
> Excellent work. The Context View is extremely tricky, and you found the places that it needed to be detected and the logic changed to support it..

Thank you so much !for your detailed explanation and time you spent. Eventually i figured out :)

> Just a few requests to clean up the code

Will do!

pushkov-fedor:
@raineorshine, hi! I came across a small problem with `nextContext` and `prevContext` functions. When Thought is in Context View, it has already been archived in any context, we can't use just `removedContextIndex - 1` or `removedContextIndex + 1` for getting prev or next context. Here are contexts which are returned from `getContextsSortedAndRanked`.
Screenshot from 2020-07-22 11-02-38
If we delete thought in **a** context, we can't use the next one. But also we can't use **b** context, its rank incorrect, because it takes into `=archive` contexts. The correct head value for cursor should be `{ value: 'b', rank: 1 }`
That's why i use filter to remove '=archive' contexts and recalculates rank. I can't just use `find` or `findIndex`, because we need to know how much `=archive` contexts are before. Please point me out, if it could be done with another way.

Also i haven't tested anything with deleteReducer. Am i right, that there we have the same behavior as archiving, but deleting should take place in `=archive` context?

raineorshine:
Can you provide steps to reproduce? Since you did not describe the thought structure that led to your test case, I am not able to know what you are referring to. If you export your thought structure as plain text and paste it there, I can import it exactly. 

- When Thought is in Context View, it has already been archived in any context

Only `Contexts` can be in `Context View`, not thoughts. It might be helpful to refer to specific types when getting into detailed discussions like this. It's not clear what kind of thought you're talking about, or where it is. Maybe the exported thoughts will clear this up.

> Also i haven't tested anything with deleteReducer. Am i right, that there we have the same behavior as archiving, but deleting should take place in =archive context?

`deleteThought` takes place in all contexts, not just an archived context. 

Deleting is different than archiving, but the cursor movement at the end is the same.

pushkov-fedor:
> deleteThought takes place in all contexts, not just an archived context.
> 
> Deleting is different than archiving, but the cursor movement at the end is the same.

Then i'm a little bit confused right now. What do you mean by deleting? Before archiving logic, deleting was pretty understandable. Now deleting thought via shortcut fires archive logic. Even invoking delete in `[...., =acrhive]` context triggers `archiveThought` reducer. In which case do we actually **delete** thought?

pushkov-fedor:
> Can you provide steps to reproduce? Since you did not describe the thought structure that led to your test case, I am not able to know what you are referring to. If you export your thought structure as plain text and paste it there, I can import it exactly.

Sure, sorry for that. 
Use tree structure from test cases for context view:
- a
  - m
    - x
- b
  - m

Use my last commit and please log to console `contexts` variable in `prevContext()` function inside archiveThought reducer. Toggle context view on `[ b, m ]`, toggle hidden thoughts. Set cursor on [ b, m, a ] and archive thought. You will see 2 contexts. Then add again
- m
  - x
in [ a ] context. Again set cursor on [ b, m, a ] and archive thought. You will see 3 contexts, including [a, =archive]

raineorshine:
> @raineorshine, hi! I came across a small problem with `nextContext` and `prevContext` functions. When Thought is in Context View, it has already been archived in any context, we can't use just `removedContextIndex - 1` or `removedContextIndex + 1` for getting prev or next context. 

In this situation, when `b/m~/a` is deleted while `a/=archive/m` already exists, the Context View will effectively replace `b/m~/a` with `b/m~/=archive`, where `=archive` represents `a/=archive`. Thus there are two `ThoughtContexts` in the Context View before and two after, but one of them has changed (i.e. `a/m` has been archived, or `m` has been moved from `['a']` to `['a', '=archive']`). This is as you have described.

In the current implementation this means the cursor will need to move from `b/m~/a` to `b/m~/=archive`. However, as I believe you are pointing out, if `b/=archive/m` exists, then `b/m~/=archive` could refer to `a/=archive/m` or `b/=archive/m`. This is the url problem we have identified in https://github.com/cybersemics/em/pull/800#issuecomment-658331702.

If that's the case, we can't go any further on this without coming up with a solution to the url, i.e. resolving path ambiguities. I intended to open up the conversation in a group chat, but I haven't had the time to gather all of our comments on this issue and summarize them for others to respond to. I will put that on my list, as it seems to be interfering with the Context View in several areas.

> > deleteThought takes place in all contexts, not just an archived context.
> > Deleting is different than archiving, but the cursor movement at the end is the same.
> 
> Then i'm a little bit confused right now. What do you mean by deleting? Before archiving logic, deleting was pretty understandable. Now deleting thought via shortcut fires archive logic. Even invoking delete in `[...., =acrhive]` context triggers `archiveThought` reducer. In which case do we actually **delete** thought?

Delete means to remove completely. Archive means to move to `=archive`, or remove completely if it is already in archive.

`deleteThought` is not currently used since it was replaced with archiving. However I still want to support it, as future features like the API will need to be able to remove a thought completely.

I changed the title to reflect that we're working primarily with archiving.

raineorshine:
@pushkov-fedor Do you think we should get this merged and handle the other case in a separate issue? This does partially solve the original issue.

pushkov-fedor:
I think so. I'm a bit lost in out discussion regarding this issue and not sure what next step should be. My solution solves test case, but we came across some difficulties and ambiguities with Context View. Let's close this and move out discussion to the next issue.

raineorshine:
Sure. Could you rebase so I can merge? Thanks!

pushkov-fedor:
> Sure. Could you rebase so I can merge? Thanks!

Done

PR #800, opened 2020-07-14
Cursor duplication in context view

### LINKED ISSUE #590: Cursor Duplication in context view
For thoughts with a similar tree structure, the `setCursor` action in the context view doesn't seem to work for the contexts of children at second or subsequent levels of the tree. Clicking on one context focuses on all the contexts with a similar structure. 
Here's an example -

cursor-duplication_contextView

pushkov-fedor:
Fixes #590

@raineorshine, hi! I found what causes this issue. Here's cursor value from issue test case.
Screenshot from 2020-07-14 13-24-33
Looks like we use only head Thought in context and don't pay attention for remaining part. Because of this, we can't distinguish selected Thought and set cursor on it. I think cursor should be like that: `a, x, y, a, x` or `a, x, y, b, x`. I tried to fix it and was totally confused in contexts and thoughts manipulations, because i'm not familiar with cursor code. Code in PR - attempt to fix it and is pretty messy.

What do you think about it?  If it's okay, could you please navigate me to what part of logic should be refactored? Thanks!
>  I think cursor should be like that: `a, x, y, a, x` or `a, x, y, b, x`


raineorshine:
> Looks like we use only head Thought in context and don't pay attention for remaining part. Because of this, we can't distinguish selected Thought and set cursor on it. 

Yes, that is a problem. Usually the rank distinguishes duplicates. In this case however, since they are `ThoughtContext` objects rather than `Child` objects, they may have the same rank, albeit from different contexts. There is a [temporary rank that is generated in getContextsSortedAndRanked](https://github.com/cybersemics/em/blob/dev/src/selectors/getContextsSortedAndRanked.ts) which could potentially be used to distinguish them. It could also use `id`, but that is a newer addition to the code that I am not as familiar with.

> I tried to fix it and was totally confused in contexts and thoughts manipulations, because i'm not familiar with cursor code. 

Yes, it is incredibly confusing.

> Code in PR - attempt to fix it and is pretty messy.

It's not working for me. There is no overlay bullet shown on the cursor at all on the selected ThoughtContext. I think we need to think through a proper solution anyway given the issue you found with uniquely identifying a ThoughtContext.

> 
> What do you think about it? If it's okay, could you please navigate me to what part of logic should be refactored?

I wish I could, but I don't know the solution any more than you :).

> 
> > I think cursor should be like that: `a, x, y, a, x` or `a, x, y, b, x`

Unfortunately that won't work since the appended `a, x` is ambiguous: it could refer to `x` in a specific context, as in this case, or it could refer to `a` in a different context with its own `y`. I would have to think about it more to come up with a specific tree to reproduce, but it seems clear that appending/embedding contexts (`a, x`) within contexts (`a, x, y`) will present problems since the boundaries are not clear between them.

A potential option is to store the full Context of each Path component rather than just the `head` in the `cursor`. Thus the cursor would have enough information to disambiguate ThoughtContexts. However we are faced with the issue of the URL scheme. Currently the URL is a shallow Path without rank information. The `cursor` must be recreated with a best guess via `rankThoughtsFirstMatch`. The URL would not be able to include the additional information of the full Context for each Path component.

I can think of two options to 
1. Encode the rank into the URL. Use the dynamically constructed rank of the Context View to disambiguate ThoughtContexts.
2. Do not allow transhierarchical Paths (Paths that include at least one Context View), converting them to the context of the last Thought, i.e. `/~/a/x/y~/x` → `/~/a/x`

The problem with (1) is that it makes the URLs quite ugly. We also don't have the option of only using ranks if we need to, as it is possible a ThoughtContext added in the future may suddenly make a rankless url ambiguous.

The problem with (2) is that we lose functionality, namely the ability to save/bookmark/share not just Contexts but Paths. When there are no duplicate ThoughtContext values, this is an unnecessary loss, i.e. the current implementation *does* support transhierarchical Paths in most cases.

Let me know if anything needs explaining. We are getting into very detailed aspects of the way that Contexts, URLs, and the Context View work together. It may help to refer to [types.ts](https://github.com/cybersemics/em/blob/dev/src/types.ts), as the capitalized types I mention are specifically referencing the types defined there.

pushkov-fedor:
Thank you for your answer and opinion!

Because of massive amount of code and logic, i don't understand difference between `thoughtsRanked`, `thoughtsRankedLive`, `thoughtsResolved` and `thoughtsResolvedLive`, i took a look in their structure, but i'm not sure that i understand purposes of each of them.

I decided to try this approach: as we use head Thought in context and can't distinguish Thoughts by its value in our case, i set rank value for cursor head, which equals rank of appropriate context from `getContextsSortedAndRanked`. 
For our case:
If we select  `a x` context, rank of x in cursor will be 0. If we select `b x` context, rank will be 1. What do you think about it?

It seems to me that it would be ideal to keep Context View value in cursor, so cursor 
{ value: 'a', ... }
{ value: 'x', ... }
{ value: 'y', contextView: true, ... }
{ value: 'a', ... }
{ value: 'x', ... }
will be different from
{ value: 'a', ... }
{ value: 'x', ... }
{ value: 'y', contextView: false, ... }
{ value: 'a', ... }
{ value: 'x', ... }
But i don't know what consequences such change could cause.

raineorshine:
> Because of massive amount of code and logic, i don't understand difference between `thoughtsRanked`, `thoughtsRankedLive`, `thoughtsResolved` and `thoughtsResolvedLive`, i took a look in their structure, but i'm not sure that i understand purposes of each of them.

This confusion is quite understandable. I will try to explain.

- **`thoughtsRanked`** - A simple `Path`, i.e. a `Path` with no Context Views. e.g. `a/x/y`
- **`thoughtsRankedLive`** - When the cursor thought is edited, its value changes. Due to the rendering logic, we cannot change `thoughtsRanked` during editing, as it would re-render the Thought component, messing up typing. `thoughtsRankedLive` reflects the changed value, while `thoughtsRanked` reflects the value when the Thought was first rendered. Bijay pointed out that the editable element should not re-render even if the Thought component re-renders, so long as the React key has not changed. Therefore we should be able to refactor this and consolidate `thoughtsRanked` and `thoughtsRankedLive`. Until then, we have to live with both, and if one doesn't work you may have to try the other.
- **`thoughtsResolved`** - A `Path` that may cross active Context Views. e.g. `a/x/y~/x`. This has the same type as `cursor`. It is usually built up from the `contextChain` which keeps track of multiple `Path` segments that cross Context Views. `thoughtsResolved` is usually needed for `cursor` type calculations, while `thoughtsRanked` is needed for calculations that require a Context, such as `getThoughtsRanked`.
- **`thoughtsResolvedLive`** - The live version of `thoughtsResolved`, as described above.

> I decided to try this approach: as we use head Thought in context and can't distinguish Thoughts by its value in our case, i set rank value for cursor head, which equals rank of appropriate context from `getContextsSortedAndRanked`.
> For our case:
> If we select `a x` context, rank of x in cursor will be 0. If we select `b x` context, rank will be 1. What do you think about it?

This is the way the `cursor` works currently. The cursor reflects the visible rank, so in the Context View the cursor will use the dynamic rank from `getContextsSortedAndRanked` rather than the source rank from the ThoughtContext. In other data structures this may not be the case, but I just confirmed that this is true for `cursor`.

> It seems to me that it would be ideal to keep Context View value in cursor, so cursor
> { value: 'a', ... }
> { value: 'x', ... }
> { value: 'y', contextView: true, ... }
> { value: 'a', ... }
> { value: 'x', ... }
> will be different from
> { value: 'a', ... }
> { value: 'x', ... }
> { value: 'y', contextView: false, ... }
> { value: 'a', ... }
> { value: 'x', ... }
> But i don't know what consequences such change could cause.

This is a refactor I have considered. It would be fairly challenging to make this change without breaking things. Currently which contexts are in Context View is captured in `state.contextViews`, while the Path segments of the `cursor` are captured in the active `contextChain`. `contextChain` is not saved in state, but rather built up through the rendering process or recreated from the `cursor`. I think it may be necessary to combine `contextChain` and `cursor`, although I dread the work involved.

src/selectors/chain.ts
```diff
@@ -1,19 +1,26 @@
...
+  const superAsString = superContext.join('/')
+  const subAsString = subContext.join('/')
+  return superAsString.includes(subAsString)
+}
+
 /** Merges thoughts into a context chain, removing the overlapping head. */
 // use autogenerated rank of context
 // if there is no/empty context chain, return thoughtsRanked as-is
-const chain = (state: State, contextChain: Path[], path: Path) => {
+const chain = (state: State, contextChain: Path[], path: Path, showContexts?: boolean) => {
```
raineorshine, line 16: The head of each `Path` in the `contextChain` represents a Context View, so we shouldn't need an extra variable `showContext`. We should discuss our approach more however before attempting an implementation.

pushkov-fedor:
> This confusion is quite understandable. I will try to explain.
> 
> thoughtsRanked - A simple Path, i.e. a Path with no Context Views. e.g. a/x/y
> thoughtsRankedLive - When the cursor thought is edited, its value changes. Due to the rendering logic, we cannot change thoughtsRanked during editing, as it would re-render the Thought component, messing up typing. thoughtsRankedLive reflects the changed value, while thoughtsRanked reflects the value when the Thought was first rendered. Bijay pointed out that the editable element should not re-render even if the Thought component re-renders, so long as the React key has not changed. Therefore we should be able to refactor this and consolidate thoughtsRanked and thoughtsRankedLive. Until then, we have to live with both, and if one doesn't work you may have to try the other.
> thoughtsResolved - A Path that may cross active Context Views. e.g. a/x/y~/x. This has the same type as cursor. It is usually built up from the contextChain which keeps track of multiple Path segments that cross Context Views. thoughtsResolved is usually needed for cursor type calculations, while thoughtsRanked is needed for calculations that require a Context, such as getThoughtsRanked.
> thoughtsResolvedLive - The live version of thoughtsResolved, as described above.

Okay, difference between them is more clear for me now. Thank you! Could you tell me more about `contextChain` please, e.g. give some examples? It is array of `Paths`, but in issue cases it usually contains only single `Path`. 

> thoughtsResolved is usually needed for cursor type calculations,

What do you mean by cursor type calculations?

pushkov-fedor:
> This is the way the cursor works currently. The cursor reflects the visible rank, so in the Context View the cursor will use the dynamic rank from getContextsSortedAndRanked rather than the source rank from the ThoughtContext. In other data structures this may not be the case, but I just confirmed that this is true for cursor.

Actually, it didn't work like this. In `chain` selector we calculates **appendedThoughtInContext**. In previous code for tree from issue:
```
const pivot = head(contextChain[contextChain.length - 1])
const i = path.findIndex(child => child.value === pivot.value)
const append = path.slice(i - 1)
const contexts = getContextsSortedAndRanked(state, pivot.value)
const appendedThoughtInContext = contexts.find(child => head(child.context) === append[0].value)
```
For path `a/x/y` or `b/x/y` in **append** calculations we were losing top level Child (i.e. **a** or **b**) and path always was `x/y`. In **appendedThoughtInContext** we used head of each context (`["a","x']` and `["b","x"]`), which is always **x** for both of them. So this is a reason why we got selection for both Thoughts.

pushkov-fedor:
Wow, i haven't known about contextViews. As i see it uses hash of thought as property, and true/false as value. What is hash calculated from? Only thought value or also context? What do you think about adding to state another cursor for thought in Context View. Something like this:
{ value: 'a', ... }
{ value: 'x', ... }
{ value: 'y', contextView: true, ... }
{ value: 'a', ... }
{ value: 'x', ... }
It will differ from common cursor by having contextView flag and full path (includes selected thought and its all rendered parent thoughts). Having full path allows us to simplify archiving thoughts from another PR, where we have to determine whether a thought is a child of another Thought in Context View or a parent. Such approach is kind of workaround - it saves our time and doesn't break current cursor logic, but i'm worried that later we will come across with another data-structure-problem, which solving will take much more time. I'm still getting familiar with some part of codebase, my view could be incomplete and opinion could be wrong, pardon me please in this case :)

raineorshine:
> Could you tell me more about `contextChain` please, e.g. give some examples? It is array of `Paths`, but in issue cases it usually contains only single `Path`.

Sure. The `contextChain` is calculated by `splitChain(state, cursor)`. Consider the following:

- a
  - m
    - x
- b
  - m
    - y

When `cursor` is `a/m~/b/y`, then `contextChain` is (ranks omitted for readability):

```js
[
  ['a', 'm'],
  ['b', 'y']
]
```

That is, the `cursor` consists of the initial segment `a/m`, then we enter the Context View of `m`, then `b/y`.

This allows the `cursor` to move across multiple Context Views. A more complicated example (copy and paste into **em** to test):

```
  - Books
    - Read
      - C. S. Peirce
        - Philosophical Writings
          - Three Categories
          - Philosophy of Math
          - Philosophy Logic
          - Semiotics
  - Personal
    - Influences
      - Gregory Bateson
      - Michael Polanyi
      - C. S. Peirce
  - Philosophy
    - Philosophy of Math
      - Statistical Inference
      - Probability as Potential
    - Philosophy of Science
    - Metaphysics
      - Sri Aurobindo
      - Forrest Landry
        - Potentiality
          - Probability as Potential
            - Potentiality vs Actuality
```

The `cursor` is `/Books/Read/C.S. Peirce/Philosophical Writings/Philosophy of Math~/Philosophy/Probability as Potential~/Potentiality/Potentiality vs Actuality` has two Context Views (`Philosophy of Math` and `Probability as Potential`) that are crossed, thus there are three segments in the `contextChain`:

```js
[
  ['Books', 'Read', 'C. S. Peirce', 'Philosophical Writings', 'Philosophy of Math'],
  ['Philosophy', 'Probability as Potential'],
  ['Potentiality', 'Potentiality vs Actuality'],
]
```

> 
> > thoughtsResolved is usually needed for cursor type calculations,
> 
> What do you mean by cursor type calculations?

`thoughtsRanked` represents a simple `Path` which does not tell you which Context Views you may have used to get there. If you want to the full `cursor` including traversals across Context Views, we have to use `thoughtsResolved` which *resolves* the `contextChain` into a full `Path`. We can look at a specific example if you would like more of an explanation.

pushkov-fedor:
Okay, i understand now, thanks a lot!

pushkov-fedor:
By the way, moving along these thoughts `/Books/Read/C.S. Peirce/Philosophical Writings/Philosophy of Math~/Philosophy/Probability as Potential~/Potentiality/Potentiality vs Actuality` on `/Books/Read/C.S. Peirce/Philosophical Writings/Philosophy of Math~/Philosophy/Probability as Potential~/Potentiality/` i got an error.
Screenshot from 2020-07-17 17-16-57

raineorshine:
> Okay, i understand now, thanks a lot!

Great. I copied my explanation to the Wiki, and added some links to functions that work with the `contextChain`:

https://github.com/cybersemics/em/wiki/Technical-Overview#2222-contextchain

> By the way, moving along these thoughts `/Books/Read/C.S. Peirce/Philosophical Writings/Philosophy of Math~/Philosophy/Probability as Potential~/Potentiality/Potentiality vs Actuality` on `/Books/Read/C.S. Peirce/Philosophical Writings/Philosophy of Math~/Philosophy/Probability as Potential~/Potentiality/` i got an error.

Excellent, please post an issue with steps to reproduce!

raineorshine:
Too far behind now

PR #801, opened 2020-07-14
Strip tags on clearThought

### LINKED ISSUE #743: Strip formatting tags from placeholder text
When using the [`clearThought`](https://github.com/cybersemics/em/blob/682a796c0b68af31d9868204c93ab2757ab41685/src/shortcuts/clearThought.js#L22) shortcut, the thought value becomes the placeholder text. If the value contains a formatting tag like `<i>`, it will be incorrectly encoded in the placeholder. It should either be rendered as html or stripped in the placeholder.

shresthabijay:
fixes #743

### DIFF
--- a/src/shortcuts/clearThought.js
+++ b/src/shortcuts/clearThought.js
@@ -21,7 +21,7 @@ const clearThoughtShortcut = {
       // need to delay DOM changes on mobile for some reason so that this works when edit mode is false
       setTimeout(() => {
         editable.innerHTML = ''
-        editable.setAttribute('W', text)
+        editable.setAttribute('placeholder', text)
       })
     }
   }
--- a/src/shortcuts/clearThought.js
+++ b/src/shortcuts/clearThought.js
@@ -3,8 +3,7 @@ import {
   isDocumentEditable,
   setSelection,
 } from '../util'
-
-const tagsRegex = /<(?:.|\n)*?>/gmi
+import { REGEXP_TAGS } from '../constants'
 
 const clearThoughtShortcut = {
   id: 'clearThought',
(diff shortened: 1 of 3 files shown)


shresthabijay:
@raineorshine We cannot render html within placeholder so it would make sense to just strip away all the tags. What do you think ?

raineorshine:
Yes, that's correct. It won't have any non-formatting tags anyway since they have already been stripped out during editing.

src/shortcuts/clearThought.js
```diff
@@ -4,6 +4,8 @@ import {
   setSelection,
 } from '../util'
 
+const tagsRegex = /<(?:.|\n)*?>/gmi
```
raineorshine, line 7: We currently have two other regex definitions that strip tags/html. Could you assess whether they can be combined into a single regex, and if so, have them all (including `clearThought`) reference that regex from `constants`? Thanks!

- https://github.com/cybersemics/em/blob/dev/src/constants.ts#L194
- https://github.com/cybersemics/em/blob/dev/src/util/hashThought.ts#L8

raineorshine:
@shresthabijay Any updates on this?

src/shortcuts/clearThought.js
shresthabijay: @raineorshine The `REGEXP_HTML` in constants `/<\/?[a-z][\s\S]*>/i` doesn't do strict tag checks. `<p>test>` as a whole will pass the regex test. Is this intentional ?

raineorshine: `REGEXP_HTML` is currently used to detect HTML, not replace tags. I'm not sure what you mean by "strict tag checks."

You should be able to see exactly where each of the HTML-related regular expressions are being used, and determine if any of them can be combined based on their use.

shresthabijay:
@raineorshine The `REGEXP_HTML` is quite different from the regex used for tags in `hashThought`. But the tags regex used in clear thought is same as used in `hashThought`, so I exported it as a separate constant.

raineorshine:
Perfect, thanks!

PR #802, opened 2020-07-17
Curry reducers

raineorshine:
Curry reducers to clean up `reducerFlow`.

This...

```js
  reducersFlow([
    state => newThought(state, { value: 'a' }),
    state => newThought(state, { value: 'b' }),
    state => existingThoughtMove(state, {
      oldPath: [{ value: 'b', rank: 1 }],
      newPath: [{ value: 'b', rank: -1 }],
    }),
  ])(state)
```

... can now change to:

```js
  reducerFlow([
    newThought({ value: 'a' }),
    newThought({ value: 'b' }),
    existingThoughtMove({
      oldPath: [{ value: 'b', rank: 1 }],
      newPath: [{ value: 'b', rank: -1 }],
    }),
  ])(state)
```

### Notes:

- Reducers with no payload are not curried.
- Payload objects are no longer optional, e.g. ~`archiveThought()`~ `archiveThought({})`. Omitting the payload will return a curried function. This might be corrected with a custom `curry` function, but for now the standard `_.curryRight` is being used.

### DIFF
--- a/src/reducers/__tests__/archiveThought.js
+++ b/src/reducers/__tests__/archiveThought.js
@@ -9,9 +9,9 @@ import { archiveThought, cursorUp, newThought, setCursor } from '../../reducers'
 it('archive a thought', () => {
 
   const steps = [
-    state => newThought(state, { value: 'a' }),
-    state => newThought(state, { value: 'b' }),
-    archiveThought,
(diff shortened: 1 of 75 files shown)


anmolarora1:
Looks much cleaner. Good job 👍

PR #808, opened 2020-07-21
Pluralize values inside chain

### LINKED ISSUE #426: Context View: Hash matched contexts do not show children
## Steps to Reproduce
- a
  - cats [cursor, context view]
- b
  - cat
    - x

## Current Behavior
When the cursor view is activated on `a.cats`, it correctly shows both contexts, but does not ~show the child `x` under `b`~ allow the cursor to be set on `b/cat~/x`. 

## Expected Behavior
`b/cat~/x` should be selectable. It work exactly the same as when the values are an exact match:

- a
  - cat
- b
  - cat
    - x

## Notes
- It has to do with the fact that `cat` and `cats` correctly hash match (i.e. `hashThought('cat') === hashThought('cats')`), but the cursor logic is not using the correct `value` in this case.
- I am not sure if the problem is in the [generation of the `thoughtsRanked` value for `setCursor`](https://github.com/cybersemics/em/blob/dev/src/components/Editable.tsx#L128-L149), or in `setCursor` itself, so I am not sure yet where the proper unit test belongs.

shresthabijay:
fixes #426


shresthabijay:
@raineorshine @raineorshine  The issue was due to incorrect `thoughtResolved` calculated by `chain`. `chain` compares values to find index of `pivot` inside `path`. Since values are not pluralized, the resulting index was incorrect. I pluralized values before comparison which fixed the issue. But I think we should should compare hashes instead. What do you think?

shresthabijay:
> Think about the encapsulation boundary. It's really important to think about which modules should know about other modules. `chain` should not know about `pluralize`, since `pluralize` is part of the hashing functionality. The hash function _could_ use any kind of hash. In order to maintain this decoupling, we have to abstract away from the implementation details of the hash. By breaking this encapsulation boundary, you are unnecessarily adding more surface area to the internal api.

Agreed. I think we should compare hashes instead.

raineorshine:
Seems like we're on the same page :).

Does the encapsulation boundary make sense?

shresthabijay:
> Does the encapsulation boundary make sense?

Yes. If we change the implementation of hash function then it will not break the functionality of chain.

raineorshine:
Thank you, looks good! Could you add the appropriate unit test?

https://github.com/cybersemics/em/blob/dev/src/util/__tests__/chain.js

shresthabijay:
> Thank you, looks good! Could you add the appropriate unit test?
> 
> https://github.com/cybersemics/em/blob/dev/src/util/__tests__/chain.js

Sure will do that.

shresthabijay:
> So it's working when I click on `a/cats~/b/x` but not when I activate `cursorDown` on `a/cats~/b`. We may have two bugs: one in `chain` and one in `nextThought`.

I am looking into it.

shresthabijay:
> So it's working when I click on `a/cats~/b/x` but not when I activate `cursorDown` on `a/cats~/b`. We may have two bugs: one in `chain` and one in `nextThought`.

@raineorshine The bug in `nextThought` was caused because the `hashContext` doesn't pluralize before creating hash. So hash of `['a', 'cats']` and `['a', 'cat']` are different. We will get a context `['b', 'cats']` from `contextChain` `a/cats~/b` using `getContextFromContextChain(state, contextChain)`. Down the line `firstVisibleChild` use `getThoughtsRanked` which uses `contextIndex` to find children of specific context. Since `['b', 'cats']` doesn't have same hash as `['b', 'cat']` it returns zero children and `firstChild` becomes undefined. This is causing the issue.

https://github.com/cybersemics/em/blob/ca07a005cb9f195c0707ecb2d8ba60b6ef7183bf/src/util/nextThought.ts#L158

I tried quick solution by pluralizing values inside `hashContext` and it works. The problem is that it causes weird children duplication when a thought with plural value is changed to singular value.

raineorshine:
> > So it's working when I click on `a/cats~/b/x` but not when I activate `cursorDown` on `a/cats~/b`. We may have two bugs: one in `chain` and one in `nextThought`.
> 
> @raineorshine The bug in `nextThought` was caused because the `hashContext` doesn't pluralize before creating hash. 

Good catch! I had assumed `hashContext` was using `hashThought` on each thought value within the `Context`. Maybe I had omitted that for performance reasons not realizing it would cause this issue.

Note that if we change `hashContext` it will break all of the existing `Context` hashes in production (i.e. my data!). I will have to coordinate this change with a script that re-calculates all of the hashes. This will be a good time to sort the thoughts that go into `hashContext` so that `['a', 'b']` will hash match `['b', 'a']` which was part of the original design.

> I tried quick solution by pluralizing values inside `hashContext` and it works. 

Okay, good. As you know pluralize is just this specific case. We need to generalize to `hashThought` as there are other things like case sensitivity, emojis, etc that are also compressed.

> The problem is that it causes weird children duplication when a thought with plural value is changed to singular value.

Got it. I assume you are troubleshooting this now.

shresthabijay:
@raineorshine I have made the necessary changes. Please review it. Thanks!

shresthabijay:
> This will be a good time to sort the thoughts that go into `hashContext` so that `['a', 'b']` will hash match `['b', 'a']` which was part of the original design.

I am little confused about this. As I understand `['a', 'b']` and `['b', 'a']` can be two unique context in the tree. If we sort thoughts in context then two different contexts can have same children right ?

raineorshine:
> > This will be a good time to sort the thoughts that go into `hashContext` so that `['a', 'b']` will hash match `['b', 'a']` which was part of the original design.
> 
> I am little confused about this. As I understand `['a', 'b']` and `['b', 'a']` can be two unique context in the tree. If we sort thoughts in context then two different contexts can have same children right ?

The intention is that `Contexts` are like sets and therefore unordered. So `['a', 'b']` and `['b', 'a']` are actually the same `Context`. This was omitted from the original implementation due to the simplicity/serializability of arrays.

If there is anywhere in the code that treats `Contexts` as ordered (such as calling `contextOf`) they would need to change to operate on a `Path`, which is ordered, and then convert it to a `Context`.

src/util/hashContext.ts
```diff
@@ -1,16 +1,31 @@
...
 /** Encode the thoughts (and optionally rank) as a string. */
 export const hashContext = (thoughts: Context | Path, rank?: number): string => (globals.disableThoughtHashing ? ID : murmurHash3.x64.hash128)(pathToContext(thoughts)
   .map(thought => thought ? escapeSelector(thought) : '')
+  .map(thought => _.flow([
+    stripTags,
+    lower,
+    trim,
+    stripEmojiWithText,
+    singularize,
+  ])(thought))
```
raineorshine, line 29: The composition of string modifiers that are fed into  murmurHash should be abstracted into a function that both `hashContext` and `hashThought` use. Otherwise it will be too easy for one of them to get out of sync.

```js
const compress = _.flow([
    stripTags,
    lower,
    trim,
    stripEmojiWithText,
    singularize,
  ])
```

shresthabijay: Sure.

shresthabijay:
> The intention is that `Contexts` are like sets and therefore unordered. So `['a', 'b']` and `['b', 'a']` are actually the same `Context`. This was omitted from the original implementation due to the simplicity/serializability of arrays.
> 
> If there is anywhere in the code that treats `Contexts` as ordered (such as calling `contextOf`) they would need to change to operate on a `Path`, which is ordered, and then convert it to a `Context`.

That makes sense. Let's create a separate issue for this.

raineorshine:
> > The intention is that `Contexts` are like sets and therefore unordered. So `['a', 'b']` and `['b', 'a']` are actually the same `Context`. This was omitted from the original implementation due to the simplicity/serializability of arrays.
> > If there is anywhere in the code that treats `Contexts` as ordered (such as calling `contextOf`) they would need to change to operate on a `Path`, which is ordered, and then convert it to a `Context`.
> 
> That makes sense. Let's create a separate issue for this.

Sure, but we'll need to branch off `shresthabijay:426` and merge together, there will be incompatible data models floating around.

raineorshine:
I have a method of migrating existing hashes now, so I’ve taken this off hold. We can sort the contexts before hashing without having to make other adjustments to the codebase, so since that will only affect the hash. We can deal with actual unordered contexts separately since that’s a bigger change. 

Will you rebase on `dev`?

PR #809, opened 2020-07-22
Add =focus/Zoom for =children

### LINKED ISSUE #651: =children/=focus/Zoom
Allow `=focus/Zoom` to be added to the `=children` attribute. This should create the same effect as adding `=focus/Zoom` to each child.

pushkov-fedor:
Fixes #651

### DIFF
--- a/src/components/Subthoughts.tsx
+++ b/src/components/Subthoughts.tsx
@@ -464,12 +464,14 @@ export const SubthoughtsComponent = ({
     1. Force actualDistance to 2 to hide thoughts.
     2. Set zoomCursor and zoomParent CSS classes to handle siblings.
   */
-  console.log('thoughtsRanked: ', thoughtsRanked)
-  console.log('parentChildrenZoom: ', parentChildrenZoom)
-  const zoomCursor = cursor && (attribute(state, pathToContext(cursor), '=focus') === 'Zoom')
-  const zoomParent = cursor && (attribute(state, pathToContext(contextOf(cursor)), '=focus') === 'Zoom')
+  const zoomCursor = cursor && (attribute(state, pathToContext(cursor), '=focus') === 'Zoom' || parentChildrenZoom)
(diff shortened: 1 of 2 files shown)


src/components/Subthoughts.tsx
```diff
@@ -462,8 +462,10 @@ export const SubthoughtsComponent = ({
     1. Force actualDistance to 2 to hide thoughts.
     2. Set zoomCursor and zoomParent CSS classes to handle siblings.
   */
-  const zoomCursor = cursor && attribute(state, pathToContext(cursor), '=focus') === 'Zoom'
-  const zoomParent = cursor && attribute(state, pathToContext(contextOf(cursor)), '=focus') === 'Zoom'
+  const zoomCursor = cursor && (attribute(state, pathToContext(cursor), '=focus') === 'Zoom'
+    || attribute(state, pathToContext(contextOf(cursor)).concat('=children'), '=focus') === 'Zoom')
+  const zoomParent = cursor && (attribute(state, pathToContext(contextOf(cursor)), '=focus') === 'Zoom'
+    || attribute(state, pathToContext(contextOf(contextOf(cursor))).concat('=children'), '=focus') === 'Zoom')
```
raineorshine, line 468: This works, but if you do it this way, then each child will have to individually check its parent. It would make more sense if the parent checked itself and then passed the result to its children. This is the way that `=children/=style` currently works.

pushkov-fedor:
Can't figure out yet, why it doesn't workt. Instead of checking attribute in component i pass calculated value to child and use it there. I think it maybe related to isEditingAncestor check, because its value differs on fbeb996 and 54a8c65 for same tree. What do you think?

raineorshine:
Yes, the zoom does involve a tricky combo between the attributes and which ancestor is being edited. I don't know the exact issue but you seem to be on the right track.

Unfortunately we have reached the budget on this task and I don't wish to add additional hours. I will use your first solution. Perhaps we can re-address it when we attempt to generalize `=children` to all attributes.

raineorshine:
Fixed in d12d416e

PR #810, opened 2020-07-22
Test: existingThoughtMove

### LINKED ISSUE #804: Test: existingThoughtMove
`getContexts` is only used in a couple tests in [`existingThoughtMove`](https://github.com/cybersemics/em/blob/dev/src/reducers/__tests__/existingThoughtMove.js). This leaves a gap in coverage over the changes made to `thoughtIndex`.

Add one or more `expect` statements that check that `getContexts` is correct in the following tests. You may need more than one, for example to check the contexts of a thought and its descendants.

- move within root
- move within context
- move across contexts
- move descendants
- move descendants with siblings

aqkhan:
Fixes #804

### DIFF
--- a/src/reducers/__tests__/existingThoughtMove.js
+++ b/src/reducers/__tests__/existingThoughtMove.js
@@ -154,7 +154,7 @@ it('move descendants', () => {
       rank: -1,
     }])
 
-  // contexts of both the descendants of b should remain unchanged
+  // contexts of both the descendants of b should change
   expect(getContexts(stateNew, 'b1'))
     .toMatchObject([{
       context: ['b'],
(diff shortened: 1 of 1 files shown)


src/reducers/__tests__/existingThoughtMove.js
```diff
@@ -213,13 +253,31 @@ it('move descendants with siblings', async () => {
...
     - d`)
 
+  // b should exist in the ROOT context
+  expect(getContexts(stateNew, 'b'))
+    .toMatchObject([{
+      context: [ROOT_TOKEN],
+      rank: 1,
+    }])
+
+  // context for both the descendants of b should remain unchanged
```
raineorshine, line 269: changed

PR #811, opened 2020-07-22
Adds a basic undo/redo implementation using 'fast-json-patch'

### LINKED ISSUE #105: Undo/Redo
Add Undo and Redo functionality.

Requirements:

- [x] Review 2-3 undo/redo libraries for Redux and provide reasoning for or against.
- [ ] Add undo and redo buttons to toolbar with appropriate svg icons.
- [ ] Undo history should persist in state, local, and remote stores (See: [Data Storage](https://github.com/cybersemics/em/wiki#data-storagepersistence)).
- [ ] Work with iterative loading (#703).
- [ ] Undo groups of actions at once.

## Groups

Each `undo` should be associated with one or more actions.

Consider the following sequence of actions:

1. newThought
2. existingThoughtChange1
3. existingThoughtChange2
4. existingThoughtChange3

These should be undoable in a single `undo`.

Another example, based on the following state:

- a
  - b [cursor]
    - c

1. existingThoughtChange1
1. existingThoughtChange2
1. existingThoughtChange3
1. cursorDown
1. existingThoughtChange4
1. existingThoughtChange5
1. existingThoughtChange6
1. cursorUp

Activating `undo` should undo 5-8, i.e. navigation and contiguous edits.

### Undoable

The following action types should be undoable:

```
archiveThought
bumpThoughtDown
cursorBack (navigation)
cursorBeforeSearch (navigation)
cursorDown (navigation)
cursorForward (navigation)
cursorHistory (navigation)
cursorUp (navigation)
deleteAttribute
deleteData
deleteEmptyThought
deleteThought
existingThoughtChange
existingThoughtDelete
existingThoughtMove
expandContextThought
indent
moveThoughtDown
moveThoughtUp
newThought
newThoughtSubmit
outdent
searchLimit
setAttribute
setCursor (navigation)
setFirstSubthought
settings
splitThought
subCategorizeAll
subCategorizeOne
toggleAttribute
toggleCodeView
toggleContextView
toggleHiddenThoughts
toggleSplitView
toolbarOverlay
tutorial
tutorialChoice
tutorialNext
tutorialPrev
tutorialStep
```

### Non-undoable

Patches should not be generated for the following actions and they should be ignored completely by the undo/redo history:

```
alert
authenticate
clear
clearQueue
dragHold
dragInProgress
editing
editingValue
error
invalidState
loadLocalState
loadLocalThoughts
modalComplete
modalRemindMeLater
prependRevision
render
search
selectionChange
setResourceCache
showModal
status
toggleSidebar
undoArchive (will be replaced with new undo/redo logic)
updateSplitPosition
unknownAction
updateThoughts (complicated)
```

anmolarora1:
Fixes #105 
  
This adds a `store-enhancer` to allow creating `patches/inverse-patches (diffs)` between the states associated with all `undoable` actions, and also two reducers, for `undo` and `redo` operations respectively. 
The purpose of this draft is to get the approach verified and identify any bottlenecks that we see at this point.


src/util/initialState.ts
```diff
@@ -76,6 +76,8 @@ export interface State {
   toolbarOverlay: string | null,
   tutorialStep?: number,
   user?: any,
+  patches?: any[],
+  inversePatches?: any[],
```
raineorshine, line 80: TODO: types

src/shortcuts/undo.js
```diff
@@ -5,8 +5,12 @@ const undoShortcut = {
   name: 'Undo',
   description: 'Undo.',
   svg: UndoIcon,
-  canExecute: () => false,
-  exec: (dispatch, getState) => { }
+  canExecute: () => true,
```
raineorshine, line 8: You can omit `canExecute` for the same effect.

src/redux-enhancers/undoRedoReducerEnhancer.js
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Checks if the action type is existingThoughtChange.
+ */
+const isExistingThoughtChange = actionType => actionType === 'existingThoughtChange'
+
+/**
+ * Gets the last item of an array.
+ */
+const getLastItem = arr => arr[arr.length - 1]
```
raineorshine, line 12: prefer `last`

src/redux-enhancers/undoRedoReducerEnhancer.js
```diff
@@ -0,0 +1,85 @@
...
+  let lastActionType
+
+  /**
+   * Reducer to handle undo/redo actions and add/merge inverse-patches for other actions.
+   */
+  const undoAndRedoReducer = (state = initialState, action) => {
+    if (!state) return reducer(initialState, action)
+    const { patches, inversePatches } = state
+    if (action.type === 'undoAction') {
+      return inversePatches.length ? getLastItem(inversePatches) && NAVIGATION_ACTIONS[getLastItem(inversePatches)[0].actions[0]] ? undoReducer(undoReducer(state)) : undoReducer(state) : state
```
raineorshine, line 54: Split long lines onto multiple lines for readability. `?` and `:` would be a good place to split this line.

src/redux-enhancers/undoRedoReducerEnhancer.js
```diff
@@ -0,0 +1,85 @@
...
+
+    // combine navigation and thoughtChange actions
+    if ((NAVIGATION_ACTIONS[action.type] && NAVIGATION_ACTIONS[lastActionType]) || (isExistingThoughtChange(lastActionType) && isExistingThoughtChange(action.type))) {
+      lastActionType = action.type
+      const lastInversePatch = getLastItem(state.inversePatches)
+      const lastState = lastInversePatch ? applyPatch(deepClone(state), lastInversePatch).newDocument : state
+      const combinedInversePatch = compare(newState, lastState)
+      return {
+        ...newState,
+        inversePatches: [...newState.inversePatches.slice(0, -1), combinedInversePatch.map(operation => ({ ...operation, actions: [...lastInversePatch ? lastInversePatch[0].actions : [], action.type] }))]
```
raineorshine, line 72: Split onto multiple lines for readability.

src/redux-enhancers/undoRedoReducerEnhancer.js
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+
+    lastActionType = action.type
+    const inversePatch = compare(newState, state)
+
+    return { ...newState, inversePatches: [...newState.inversePatches, inversePatch.map(operation => ({ ...operation, actions: [action.type] }))] }
```
raineorshine, line 79: Split onto multiple lines for readability.

anmolarora1:
@raineorshine  I've made some changes to the code and the basic `undo/redo` functionality seems to work correctly for now. It's based on two assumptions -
1. The caret positioning doesn't need to be maintained for now and would be added once the `ContentEditable` changes are merged
2. The  [setCursor issue](https://github.com/cybersemics/em/issues/814) is resolved, that is, the `setCursor` is not being called after every `newThought` action. If we don't want that, the existing code can be updated to call the `undo` and `redo` actions thrice in case of an `existingThoughtChange` action followed by `newThought` and `setCursor` actions (We'd have to revert three actions [instead of two](https://github.com/cybersemics/em/blob/7ae2fedfd7080dd8ddcd623c793cfa774827e556/src/redux-enhancers/undoRedoReducerEnhancer.js#L54) in a single undo/redo op)

raineorshine:
Here's what I found:

- The undo/redo toolbar buttons are visually larger than the other buttons. They should be the same size.
- The undo/redo button should be disabled when there is nothing left to undo/redo, respectively.
- You should not be allowed to undo back to the tutorial.
- `toggleSidebar`, `search`, and `modal*` actions should be ignored.
- The `cursor` disappears after one or two undo/redo actions.
- Undo `toggleAttribute` (e.g. `=sort`) does not work

raineorshine:
`TypeError: Cannot read property 'value' of null`

## Steps to Reproduce

1. Create `a`
2. Create `b`
3. Activate `deleteThought` on `b`
4. Activate `deleteThought` on `a`
5. Close alert
6. Undo

anmolarora1:
> * The undo/redo toolbar buttons are visually larger than the other buttons. They should be the same size.
Fixed it in the latest commit

> * The undo/redo button should be disabled when there is nothing left to undo/redo, respectively.
Fixed now.

> * You should not be allowed to undo back to the tutorial.
Updated

> * `toggleSidebar`, `search`, and `modal*` actions should be ignored.
Sure. I've made the changes in the latest commit.
Besides, I'd suggest we should remove these actions from the [issue description](https://github.com/cybersemics/em/issues/105#issue-507586134) that we created earlier

> * The `cursor` disappears after one or two undo/redo actions.
@raineorshine Do you mean the `caret` or the `cursor`? I couldn't replicate this one.

> * Undo `toggleAttribute` (e.g. `=sort`) does not work
I wasn't able to replicate this. I tried with some other `toggleAttributes` like the `table view` & `pin open` and they seemed to work fine. Would you mind sharing the steps to replicate this?

anmolarora1:
> `TypeError: Cannot read property 'value' of null`
> 
> ## Steps to Reproduce
> 1. Create `a`
> 2. Create `b`
> 3. Activate `deleteThought` on `b`
> 4. Activate `deleteThought` on `a`
> 5. Close alert
> 6. Undo

Couldn't replicate this as well. I'll try to share a screencast for the same.
Meanwhile, can you please check it at your end with the latest changes that I've made?

raineorshine:
> Besides, I'd suggest we should remove these actions from the [issue description](https://github.com/cybersemics/em/issues/105#issue-507586134) that we created earlier

Done

raineorshine:
> > `TypeError: Cannot read property 'value' of null`
> > ## Steps to Reproduce
> > 
> > 1. Create `a`
> > 2. Create `b`
> > 3. Activate `deleteThought` on `b`
> > 4. Activate `deleteThought` on `a`
> > 5. Close alert
> > 6. Undo
> 
> Couldn't replicate this as well. I'll try to share a screencast for the same.
> Meanwhile, can you please check it at your end with the latest changes that I've made?

Yes, still occurring for me on 61835d7a in Chrome.

em - 811

raineorshine:
> > * The `cursor` disappears after one or two undo/redo actions.
> 
> @raineorshine Do you mean the `caret` or the `cursor`? I couldn't replicate this one.

The `cursor`:

em - 811 cursor mov

raineorshine:
> > * Undo `toggleAttribute` (e.g. `=sort`) does not work
> 
> I wasn't able to replicate this. I tried with some other `toggleAttributes` like the `table view` & `pin open` and they seemed to work fine. Would you mind sharing the steps to replicate this?

`toggleAttribute` seems to be working for all of them now. The only thing a little off with `toggleSort` is that sometimes there is a dead `undo`.

## Steps to Reproduce

1. Create `a`
2. Create three subthoughts: `b`, `z`, and `c`
3. Select `a` and activate `toggleSort`
4. Add another subthought `m` to `a`
5. Activate `undo` twice

## Current Behavior

The first `undo` reverses the creation of `m`, and the second `undo` does nothing. 

If `undo` is activated a third time, `toggleSort` will be reversed.

## Expected Behavior

The second `undo` should reverse the activation of `toggleSort`.

anmolarora1:
> > > * Undo `toggleAttribute` (e.g. `=sort`) does not work
> > 
> > 
> > I wasn't able to replicate this. I tried with some other `toggleAttributes` like the `table view` & `pin open` and they seemed to work fine. Would you mind sharing the steps to replicate this?
> 
> `toggleAttribute` seems to be working for all of them now. The only thing a little off with `toggleSort` is that sometimes there is a dead `undo`.
> 
> ## Steps to Reproduce
> 1. Create `a`
> 2. Create three subthoughts: `b`, `z`, and `c`
> 3. Select `a` and activate `toggleSort`
> 4. Add another subthought `m` to `a`
> 5. Activate `undo` twice
> 
> ## Current Behavior
> The first `undo` reverses the creation of `m`, and the second `undo` does nothing.
> 
> If `undo` is activated a third time, `toggleSort` will be reversed.
> 
> ## Expected Behavior
> The second `undo` should reverse the activation of `toggleSort`.

That dead undo represents the `setCursor` action that's being triggered after `toggleAttribute`. It updates the `datanonce`. An inverse of that operation doesn't produce any noticeable results, hence appearing dead

anmolarora1:
> > > `TypeError: Cannot read property 'value' of null`
> > > ## Steps to Reproduce
> > > 
> > > 1. Create `a`
> > > 2. Create `b`
> > > 3. Activate `deleteThought` on `b`
> > > 4. Activate `deleteThought` on `a`
> > > 5. Close alert
> > > 6. Undo
> > 
> > 
> > Couldn't replicate this as well. I'll try to share a screencast for the same.
> > Meanwhile, can you please check it at your end with the latest changes that I've made?
> 
> Yes, still occurring for me on [61835d7](https://github.com/cybersemics/em/commit/61835d7a467b41af1ecf955de55cc88b13113dad) in Chrome.
> 
> em - 811

Yeah, I could replicate it now. I was using the toolbar icons to trigger the same set of operations, and it worked well. But with shortcuts, it didn't.
Nevertheless, after some digging here's what I found -

The reason for this error that we're getting - 
On two consecutive deletes, this is how the patches look like
Screenshot 2020-08-18 at 1 40 12 PM


After a delete operation, when we manually close the alert box, the value of `state.alert` is changed from a valid object to null. The `undo` op tries to restore the alert value at this path - `state/alert/value`. But since that path is null, we get the error - `TypeError: Cannot read property 'value' of null`
One way to deal with this is to create an undo op for the `close alert` action. That would help in preserving the correct state value but won't be enough to give us the intended result due to the following issue (Non-serializable state)

From my understanding of our current approach for `archiveThough`, we're storing a react element in the redux state, hence making it non-serializable. So an undo/redo doesn't work as expected. 
This approach doesn't seem ideal unless there's a strong reason for backing it.

https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state
https://github.com/reduxjs/redux/issues/1248

raineorshine:
> That dead undo represents the `setCursor` action that's being triggered after `toggleAttribute`. It updates the `datanonce`. An inverse of that operation doesn't produce any noticeable results, hence appearing dead

Okay, great. What solution do you propose?

> From my understanding of our current approach for `archiveThough`, we're storing a react element in the redux state, hence making it non-serializable. So an undo/redo doesn't work as expected.
> This approach doesn't seem ideal unless there's a strong reason for backing it.
> 
> [redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state](https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state)
> [reduxjs/redux#1248](https://github.com/reduxjs/redux/issues/1248)

Yes, storing a component in the `state` seems like a bad idea! Will you propose a solution? Thanks!

anmolarora1:
> > From my understanding of our current approach for `archiveThough`, we're storing a react element in the redux state, hence making it non-serializable. So an undo/redo doesn't work as expected.
> > This approach doesn't seem ideal unless there's a strong reason for backing it.
> > [redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state](https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state)
> > [reduxjs/redux#1248](https://github.com/reduxjs/redux/issues/1248)
> 
> Yes, storing a component in the `state` seems like a bad idea! Will you propose a solution? Thanks!

I inspected all the occurrences of the `alert` reducer and it seems that we're storing the component only for the `archiveThought` action. These are the 2 solutions that I'd suggest -
1. The component stored in the `value` represents - 1) the deleted thought, which is just a piece of text. 2) The `undo button` which is used to undo the `archive` op. Since we're working on the generic undo approach which will cover `undoing an archived thought`, we can get rid of this redundant button, unless it serves some definitive purpose
2. Assuming that we need to have both the `undo` buttons in place, we can add an undo button to the alert box based on the meta-data - `alertType` - And trigger the generic `undo action` on its click.
This approach would save us from the state conflicts that we're likely face when the user hits `undo` from within the alert modal, and then tries to do the same from the toolbar.

anmolarora1:
> > That dead undo represents the `setCursor` action that's being triggered after `toggleAttribute`. It updates the `datanonce`. An inverse of that operation doesn't produce any noticeable results, hence appearing dead
> 
> Okay, great. What solution do you propose?

So far, we've found out two fatuous occurrences of `setCursor` where it's just being used to update the `datanonce` so far - `newThought` and `toggleAttribute`. And there might be other such instances so handling them individually doesn't seem like the best idea. 
The proposed solution - While monitoring the actions in the `undoRedoReducerEnhancer`, we can easily detect if a `setCursor` action is just updating the `datanonce`, and merge it with the preceding `inverse-patch` in the state. That'd help us retain all the state changes in the `patches/inverse-patches` while getting rid of the `dead` undo/redo ops at the same time. 
What are your thoughts?

anmolarora1:
> > I inspected all the occurrences of the `alert` reducer and it seems that we're storing the component only for the `archiveThought` action. These are the 2 solutions that I'd suggest -
> > 
> > 1. The component stored in the `value` represents - 1) the deleted thought, which is just a piece of text. 2) The `undo button` which is used to undo the `archive` op. Since we're working on the generic undo approach which will cover `undoing an archived thought`, we can get rid of this redundant button, unless it serves some definitive purpose
> > 2. Assuming that we need to have both the `undo` buttons in place, we can add an undo button to the alert box based on the meta-data - `alertType` - And trigger the generic `undo action` on its click.
> >    This approach would save us from the state conflicts that we're likely face when the user hits `undo` from within the alert modal, and then tries to do the same from the toolbar.
> 
> That's a great explanation, thanks! Since that's the only place it's used, and we are adding real Undo functionality, I'd say we can remove the Undo button completely and use a regular alert.

Makes sense. Let's create a new issue for this one and fix it quickly

> 
> > So far, we've found out two fatuous occurrences of `setCursor` where it's just being used to update the `datanonce` so far - `newThought` and `toggleAttribute`. And there might be other such instances so handling them individually doesn't seem like the best idea.
> 
> Agreed
> 
> > The proposed solution - While monitoring the actions in the `undoRedoReducerEnhancer`, we can easily detect if a `setCursor` action is just updating the `datanonce`, and merge it with the preceding `inverse-patch` in the state. That'd help us retain all the state changes in the `patches/inverse-patches` while getting rid of the `dead` undo/redo ops at the same time.
> > What are your thoughts?
> 
> This sounds right to me. My only question is whether it needs to be more general, or if `setCursor` with only the `dataNonce` is the only situation that needs to be detected. I wonder if there are other undoable actions that might cause a dead undo state as well.
> 
> I guess we can start with your proposed solution and then expand it if we discover there are more dead undo states. Could you take 15 minutes or so to try a variety of different actions and see if there are any other dead undo states? If not, then proceed with the proposed solution. If there are, incorporate that into your analysis and see if you can find a solution that handles both cases in as generic way as possible. Thanks!

Thanks! I'll do that

raineorshine:
> Makes sense. Let's create a new issue for this one and fix it quickly
> 

Created #837

raineorshine:
> After a bit of digging, I wasn't able to find out any such dead actions. I didn't spend much time though and will try to come up with an approach that's flexible enough to accommodate any other dead actions that we might discover in the future.

Okay, thanks for looking. We can always extend the solution if needed in the future.

> With regard to the approach that we finalized, we'll have to do an extra state comparison to find out if the changes (dead actions) need to merged or not. We can do that using `fast-json-patch`, that we're using already, or another library like [this](https://github.com/mattphillips/deep-object-diff#readme). Besides, we don't need a patch to find out the difference, so we should go for something which is less performance-heavy.
> @raineorshine What do you suggest?

Can you explain this more on our call? I don't yet understand the need for the additional state comparison. I thought we already have a patch that can be examined.

raineorshine:
> If `cursor` and `cursorBeforeEdit` were synchronized on undo, would that result in the correct `isEditing` value and fix the issue?

Thoughts?

anmolarora1:
> > If `cursor` and `cursorBeforeEdit` were synchronized on undo, would that result in the correct `isEditing` value and fix the issue?
> 
> Any thoughts on this?

That works as a good `hack`. But it messes the caret positioning and places the caret at the beginning of each thought.
Besides, it is a hack that involves changing parts of the state (cursorBeforeEdit) which isn't directly related to the `undo/redo` functionality, so I'm not very confident about this approach.

anmolarora1:
> A couple things with redo:
> 
> * The redo toolbar button should be greyed out when `patches` is empty.
That makes sense

> * Executing any of `UNDOABLE_ACTIONS` normally (and only `UNDOABLE_ACTIONS`) should clear `patches`. (It doesn't make sense to apply a redo patch after the state has changed.)
> 
I'm not sure what you mean by `clear patches`. Do you mean that an action that's redone shouldn't be undoable again, and we should clear the `inversePatches`? Can you please elaborate?

raineorshine:
> > > If `cursor` and `cursorBeforeEdit` were synchronized on undo, would that result in the correct `isEditing` value and fix the issue?
> > 
> > 
> > Any thoughts on this?
> 
> That works as a good `hack`. But it messes the caret positioning and places the caret at the beginning of each thought.
> Besides, it is a hack that involves changing parts of the state (cursorBeforeEdit) which isn't directly related to the `undo/redo` functionality, so I'm not very confident about this approach.

Fair enough. I want a solid solution as well. However I don't want to delay this PR by making it depend on a high risk refactor like `cursor`/`cursorBeforeEdit`. Having a working cursor with a breaking caret is functional and seems like a reasonable means of getting this PR merged.

> > * Executing any of `UNDOABLE_ACTIONS` normally (and only `UNDOABLE_ACTIONS`) should clear `patches`. (It doesn't make sense to apply a redo patch after the state has changed.)
> 
> I'm not sure what you mean by `clear patches`. Do you mean that an action that's redone shouldn't be undoable again, and we should clear the `inversePatches`? Can you please elaborate?

Sure. This is normal behavior for linear undo/redo functionality. Consider the following:

1. Perform actions A, B, and C.
2. Undo action C.
3. Perform action D.
4. Redo... what should it do?

We can't redo action C, because D has already changed the state. The standard behavior is to wipe all redos (clear `patches`) as soon as a different action (from `UNDOABLE_ACTIONS`) is executed.

Nonlinear undo/redo is possible with a full version control system like git, but that's outside the scope for the initial version of undo/redo.

raineorshine:
This is looking really good!

Editing notes should be grouped together into a single undoable state, rather than character by character. Not sure why this isn't working automatically, since notes are just thoughts, but apparently there is a difference in the action sequence.

1. Create thought `a`
2. Add note `test`
3. Create thought `test`
4. Undo back to Step 1

You'll see that the note `test` will undo one character at a time instead of the whole thing.

However before you work on that, let's get some tests written so we have safeguards against regressions.

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  // import thoughts
+  store.dispatch(importText(RANKED_ROOT, `
+  - a
+  - b`))
+
+  store.dispatch({ type: 'setCursor', thoughtsRanked: [
+    { value: 'a', rank: '0' },
+    { value: 'b', rank: '1' }
+  ] })
```
raineorshine, line 22: You are setting the cursor to `a/b`, but `b` is a child of the root, not `a`. Do you mean `[{ value: 'b', rank: 1 }]`?

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  store.dispatch({ type: 'existingThoughtChange',
+    newValue: 'aa',
+    oldValue: 'a',
+    context: [ROOT_TOKEN],
+    thoughtsRanked: [{ value: 'a', rank: 0 }]
+  })
+
+  // undo thought change
+  executeShortcut(undoShortcut, { store, type: 'keyboard', event })
```
raineorshine, line 32: It would be better to dispatch `{ type: 'undoAction' }` so that you isolate the reducer. We can test that the shortcut is hooked up to `undoAction` correctly in a separate test.

src/shortcuts/__tests__/redo.ts
```diff
@@ -0,0 +1,56 @@
...
+    oldValue: 'a',
+    context: [ROOT_TOKEN],
+    thoughtsRanked: [{ value: 'a', rank: 0 }]
+  })
+
+  // undo thought change
+  executeShortcut(undoShortcut, { store, type: 'keyboard', event })
+
+  // eslint-disable-next-line fp/no-let
+  let exported = exportContext(store.getState(), [ROOT_TOKEN], 'text/plain')
```
raineorshine, line 36: Please do not use `let`. You can call the first variable `exportedAfterUndo` and the second variable `exportedAfterRedo`.

src/shortcuts/__tests__/redo.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  // eslint-disable-next-line fp/no-let
+  let expectedOutput = `- ${ROOT_TOKEN}
+  - a
+  - b`
+
+  expect(exported).toEqual(expectedOutput)
+
+  // undo thought change
```
raineorshine, line 45: Comment does not correspond to code

src/shortcuts/__tests__/redo.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  store.dispatch({ type: 'existingThoughtChange',
+    newValue: 'aa',
+    oldValue: 'a',
+    context: [ROOT_TOKEN],
+    thoughtsRanked: [{ value: 'a', rank: 0 }]
+  })
+
+  // undo thought change
+  executeShortcut(undoShortcut, { store, type: 'keyboard', event })
```
raineorshine, line 33: It would be better to dispatch `{ type: 'undoAction' }` so that you isolate the reducer. We can test that the shortcut is hooked up to `undoAction` correctly in a separate test.

src/shortcuts/__tests__/redo.ts
```diff
@@ -0,0 +1,56 @@
...
+    { value: 'a', rank: '0' },
+    { value: 'b', rank: '1' }
+  ] })
+
+  store.dispatch({ type: 'existingThoughtChange',
+    newValue: 'aa',
+    oldValue: 'a',
+    context: [ROOT_TOKEN],
+    thoughtsRanked: [{ value: 'a', rank: 0 }]
+  })
```
raineorshine, line 30: `store.dispatch` accepts an array of actions.

raineorshine (changes requested):
Overall, looks good! Minor changes suggested.

I assume you'll be testing the `cursor` restoration separately.

src/shortcuts/__tests__/undo.ts
anmolarora1: Agreed. I think we can use [redux-mock-store ](https://redux.js.org/recipes/writing-tests#async-action-creators)to test if correct actions are hooked to the `undo shortcut`. 
Let me know if you've got any other approach in mind.

src/shortcuts/__tests__/redo.ts
anmolarora1: Updated

src/shortcuts/__tests__/redo.ts
anmolarora1: Thanks! That makes sense

src/shortcuts/__tests__/redo.ts
anmolarora1: https://github.com/cybersemics/em/pull/811#discussion_r490990907

src/shortcuts/__tests__/redo.ts
raineorshine: I should mention that redux doesn't support arrays out of the box; it comes from a tiny middleware called [redux-multi](https://github.com/ashaffer/redux-multi/blob/master/src/index.js).

src/shortcuts/__tests__/undo.ts
raineorshine: Yes, I like that approach.

src/redux-enhancers/undoRedoReducerEnhancer.js
```diff
@@ -0,0 +1,188 @@
...
+const nthLast = (arr, n) => arr[arr.length - n]
+
+/**
+ * Applies the last inverse-patch to get the next state and adds a corresponding reverse-patch for the same.
+ */
+const undoReducer = state => {
+  const { patches, inversePatches } = state
+  const lastInversePatch = nthLast(inversePatches, 1)
+  if (!lastInversePatch) return state
+  const newState = applyPatch(deepClone(state), lastInversePatch).newDocument
```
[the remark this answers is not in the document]
raineorshine: @anmolarora1

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,126 @@
...
+    oldValue: 'a',
+    context: [ROOT_TOKEN],
+    thoughtsRanked: [{ value: 'a', rank: 0 }]
+  }, {
+    type: 'cursorDown'
+  }, {
+    type: 'cursorDown'
+  }, {
+    type: 'setCursor'
+  }, {
```
raineorshine, line 60: Use `thoughtsRanked: null` explicitly with `setCursor`. `undefined` is not a valid value for `cursor`.

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,126 @@
...
+    thoughtsRanked: [
+      {
+        value: 'c',
+        rank: 2,
+      }
+    ]
+  }])
+
+  // undo thought change and preceding navigation actions
+  store.dispatch({ type: 'undoAction' })
```
raineorshine, line 76: Include this in the previous `store.dispatch` call.

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+    ]
+  },
+  // dispensible set cursor (which only updates datanonce)
+  {
+    type: 'setCursor'
+  }])
+
+  // undo setCursor and thoughtChange in a sinle action
+  store.dispatch({ type: 'undoAction' })
```
raineorshine, line 115: Include this in the previous `store.dispatch` call.

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,126 @@
...
+      {
+        value: 'b',
+        rank: 1,
+      }
+    ]
+  },
+  // dispensible set cursor (which only updates datanonce)
+  {
+    type: 'setCursor'
+  }])
```
raineorshine, line 112: Use `thoughtsRanked: null` explicitly with `setCursor`. `undefined` is not a valid value for `cursor`. You can also dispatch `render` if you want to just update `dataNonce`.

raineorshine (changes requested):
Looks good!

You're missing some tests. Think through all the functionality of undo/redo we have so far. What tests would you need to guarantee that there were no regressions if it was refactored?

src/redux-enhancers/undoRedoReducerEnhancer.js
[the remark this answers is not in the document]
anmolarora1: @raineorshine The state gets mutated if we don't use `deepClone`.

src/redux-enhancers/undoRedoReducerEnhancer.js
```diff
@@ -0,0 +1,188 @@
...
+
+    // add a new inverse patch
+    const inversePatch = compareWithOmit(newState, state)
+
+    // if the patch is dispensable, combine it with the last patch
+    // Note: we can't simply ignore a dispensable patch because that would result in
+    // inconsistencies between the original state, and the one built using patches
+    if (isDispensable(inversePatch)) {
+      const lastPatch = nthLast(inversePatches, 1)
+      console.log(lastPatch, inversePatch)
```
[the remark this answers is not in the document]
anmolarora1: Removed

src/redux-enhancers/undoRedoReducerEnhancer.js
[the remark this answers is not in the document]
raineorshine: Yes, but does it break anything? `deepClone` is very expensive here.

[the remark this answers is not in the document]
raineorshine: Yes, it's just a very expensive operation, especially when most of the `state` can be shallow copied. We can leave it as-is and measure performance later.

raineorshine (changes requested):
Here are the other test cases that I can think of:

1. Undo/redo contiguous changes
2. Update cursor after undo/redo
3. Group contiguous navigation actions preceding a thought change on redo
4. Ignore non-undoable actions
5. `state.alert` is omitted from the applied patch
6. Clear `patches` when any `UNDOABLE_ACTIONS` is dispatched
  (as described in https://github.com/cybersemics/em/pull/811#issuecomment-684022102)

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,294 @@
...
+      context: [ROOT_TOKEN],
+      thoughtsRanked: [{ value: 'A', rank: 0 }]
+    },
+    {
+      type: 'existingThoughtChange',
+      newValue: 'Atlantic ',
+      oldValue: 'Atlantic',
+      context: [ROOT_TOKEN],
+      thoughtsRanked: [{ value: 'Atlantic', rank: 0 }]
+    },
```
raineorshine, line 217: This isn't a realistic action since values are always trimmed.

src/shortcuts/__tests__/redo.ts
```diff
@@ -0,0 +1,140 @@
...
+  { type: 'cursorDown' },
+  {
+    type: 'existingThoughtChange',
+    newValue: 'ar',
+    oldValue: 'a',
+    context: [ROOT_TOKEN],
+    thoughtsRanked: [{ value: 'a', rank: 0 }]
+  },
+  { type: 'cursorUp' },
+  { type: 'setCursor', thoughtsRanked: [{ value: 'ar', rank: 0 }] },
```
raineorshine, line 64: Isn't the cursor already on `ar` here?

raineorshine (changes requested):

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,294 @@
...
+    { type: 'setCursor', thoughtsRanked: null },
+    { type: 'cursorBack' },
+    { type: 'undoAction' }
+  ])
+
+  // restore the cursor to it's state before navigation actions
+  const { cursor } = store.getState()
+
+  expect(cursor).toMatchObject([
+    { value: 'aa' }])
```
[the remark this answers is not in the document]
anmolarora1: No. The navigation actions preceding the `thought change` action are undone in the same operation as the thought change. Although the navigation actions following `existingThoughtChange` are grouped together and undone as a single op, the thought change isn't a part of that operation. Let me know if that deviates from our requirements.

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,294 @@
...
+    const store = mockStore(initialState())
+
+    executeShortcut(undoShortcut, { store, type: 'keyboard', event })
+
+    expect(store.getActions()).toEqual([])
+  })
+
+})
+
+it('undo thought change', async () => {
```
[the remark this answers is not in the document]
anmolarora1: Got it.

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,294 @@
...
+      newValue: 'd1',
+      rankInContext: 3,
+      thoughtsRanked: [
+        {
+          value: 'd',
+          rank: 3,
+        }
+      ]
+    },
+    // undo thought change and preceding navigation actions
```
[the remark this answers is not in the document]
anmolarora1: That makes sense. I'll be careful

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,294 @@
...
+        - d`
+    ),
+    { type: 'setCursor', thoughtsRanked: null },
+    {
+      type: 'existingThoughtChange',
+      context: ['a'],
+      oldValue: 'b',
+      newValue: 'bd',
+      rankInContext: 1,
+      thoughtsRanked: [{ value: 'b', rank: 1 }]
```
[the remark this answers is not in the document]
anmolarora1: Can you please explain why it's invalid?

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,294 @@
...
+    importText(RANKED_ROOT, `
+      - A
+      - B`
+    ),
+    { type: 'alert', value: 'test' }
+  ])
+
+  expect(store.getState().inversePatches.length).toEqual(0)
+
+})
```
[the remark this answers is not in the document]
anmolarora1: I'm not sure if that's the correct approach for writing this unit test. Because we're assuming that `archiveThought` would always include `alert`. If the alert is excluded from the `archiveThought`, the test would still pass even if the `undo reducer` logic is incorrect.

src/shortcuts/__tests__/redo.ts
anmolarora1: Good catch. I've removed this action.

src/shortcuts/__tests__/redo.ts
```diff
@@ -0,0 +1,140 @@
...
+  { type: 'redoAction' },
+  ])
+
+  const exportedAfterRedo = exportContext(store.getState(), [ROOT_TOKEN], 'text/plain')
+
+  const expectedOutputAfterRedo = `- ${ROOT_TOKEN}
+  - arizona
+  - b`
+
+  expect(exportedAfterRedo).toEqual(expectedOutputAfterRedo)
```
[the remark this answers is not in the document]
anmolarora1: Makes sense.

src/shortcuts/__tests__/redo.ts
```diff
@@ -49,20 +49,23 @@ it('group contiguous navigation actions preceding a thought change on redo', ()
   const store = createTestStore()
 
   store.dispatch([importText(RANKED_ROOT, `
-      - a
-      - b`
+    - a
+    - b`
   ),
   { type: 'cursorDown' },
+  { type: 'setCursor', thoughtsRanked: [{ value: 'b', rank: 1 }] },
```
raineorshine, line 56: I think the `cursor` is already on `b` after the `cursorDown`.

src/shortcuts/__tests__/redo.ts
```diff
@@ -49,20 +49,23 @@ it('group contiguous navigation actions preceding a thought change on redo', ()
...
     newValue: 'ar',
     oldValue: 'a',
     context: [ROOT_TOKEN],
     thoughtsRanked: [{ value: 'a', rank: 0 }]
   },
-  { type: 'cursorUp' },
-  { type: 'setCursor', thoughtsRanked: [{ value: 'ar', rank: 0 }] },
   { type: 'cursorBack' },
+  { type: 'cursorUp' },
+  { type: 'setCursor', thoughtsRanked: [{ value: 'b', rank: 0 }] },
```
raineorshine, line 67: `[{ value: 'b', rank: 0 }]` is not a valid `Path` since `b` was indented to become `[{ value: 'a', rank: 0 }, { value: 'b', rank: 0 }]`.

src/shortcuts/__tests__/undo.ts
```diff
@@ -208,13 +206,6 @@ it('undo contiguous changes', () => {
       context: [ROOT_TOKEN],
       thoughtsRanked: [{ value: 'A', rank: 0 }]
     },
-    {
-      type: 'existingThoughtChange',
-      newValue: 'Atlantic ',
-      oldValue: 'Atlantic',
-      context: [ROOT_TOKEN],
-      thoughtsRanked: [{ value: 'Atlantic', rank: 0 }]
-    },
```
raineorshine: You need to change `oldValue` in the next `existingThoughtChange` to `Atlantic` instead of `Atlantic `.

raineorshine (changes requested):

src/shortcuts/__tests__/undo.ts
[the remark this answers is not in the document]
raineorshine: @anmolarora1 The test should assert that `state.alert` is not restored after an undo when it has been dismissed. This would fail if it weren't for `stateSectionsToOmit`. Effectively we need a separate test for `stateSectionsToOmit` since it is different than just skipping a non-undoable action.

src/shortcuts/__tests__/undo.ts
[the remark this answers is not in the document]
raineorshine: @anmolarora1 Sure. It's because `b` is a child of `a`. There is no thought at `[{ value: 'b', rank: 1 }]`. It should be `[{ value: 'a', rank: 0 }, { value: 'b', rank: 1 }]`.

src/shortcuts/__tests__/undo.ts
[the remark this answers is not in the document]
raineorshine: @anmolarora1 It's actually the other way around. The navigation actions following an edit are undone with the edit. You can see this functionality in Google Docs or other word processors.

Redo is the inverse: Navigation actions preceding an edit (temporally) are redone before the edit is redone.

Also see: https://github.com/cybersemics/em/pull/811#discussion_r492329929

src/shortcuts/__tests__/undo.ts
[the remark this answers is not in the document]
anmolarora1: That's correct. 
And I see two approaches for handling this - 1) We write our test using an existing reducer which uses `state.alert`, e.g. `archiveThought`. But as I mentioned [here](https://github.com/cybersemics/em/pull/811#discussion_r498657964), it relies on the assumption that the reducer wouldn't change.
2) We use a `test/mock reducer` and a `mock state`. Afaik, it's not possible with the existing `testStore`, and so we'll need to update it to allow passing a mock reducer as an argument instead of using the [appReducer](https://github.com/cybersemics/em/blob/5be62eb42d613aade4fecdecc70974beb16414d8/src/test-helpers/createTestStore.js#L13)
Besides, we'd need to inspect the patches that are generated to verify that they don't contain `alert` related information. 
What are your thoughts?

src/shortcuts/__tests__/undo.ts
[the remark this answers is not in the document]
anmolarora1: That's right. 
But do we need to specify the full path while we're explicitly passing the context as `['a']` at [line 117](https://github.com/cybersemics/em/blob/d2eb0b239f3e979b6838331f3656c2dfe3d72c83/src/shortcuts/__tests__/undo.ts#L117)?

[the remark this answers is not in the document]
raineorshine: Yes, there is some redundancy there, but `existingThoughtChange` would have to be refactored to eliminate it.

src/shortcuts/__tests__/undo.ts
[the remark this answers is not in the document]
raineorshine: The first approach seems much more straightforward. I don't mind having a dependency on `archiveThought` this time.

src/shortcuts/__tests__/redo.ts
anmolarora1: No, the cursor is on `a`

raineorshine: Ah, you're right. Thanks!

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,295 @@
...
+    { type: 'cursorBack' },
+    { type: 'moveThoughtDown' },
+    { type: 'cursorDown' },
+    // undo 'moveThoughtDown' and 'cursorDown'
+    { type: 'undoAction' }
+  ])
+
+  const cursorAfterFirstUndo = store.getState().cursor
+  expect(cursorAfterFirstUndo).toMatchObject([
+    { value: 'a' }])
```
raineorshine, line 100: You'll need to assert the changed thought as well. You've only demonstrated that the `cursorDown` has been undone, not that `moveThoughtDown` has been undone in the same `undoAction`.

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,295 @@
...
+
+  // undo 'cursorBack' and 'existingThoughtChange'
+  store.dispatch({ type: 'undoAction' })
+
+  const cursorAfterSecondUndo = store.getState().cursor
+  expect(cursorAfterSecondUndo).toMatchObject([
+    { value: 'a' }, { value: 'b' }])
```
raineorshine, line 107: Also assert changed thought here.

raineorshine (changes requested):
Please convert `undoRedoReducerEnhancer.js` to typescript. Thanks! I recommend writing typescript from the beginning.

src/shortcuts/__tests__/redo.ts
```diff
@@ -0,0 +1,144 @@
...
+  },
+  { type: 'cursorBack' },
+  { type: 'cursorUp' },
+  {
+    type: 'existingThoughtChange',
+    newValue: 'arizona',
+    oldValue: 'ar',
+    context: [ROOT_TOKEN],
+    thoughtsRanked: [{ value: 'ar', rank: 0 }]
+  },
```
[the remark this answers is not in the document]
anmolarora1: That's not the existing behavior of `redo`, and I'm not able to recall that being listed as a requirement either. So let's discuss that first.

src/shortcuts/__tests__/redo.ts
```diff
@@ -0,0 +1,144 @@
...
+    type: 'existingThoughtChange',
+    newValue: 'ar',
+    oldValue: 'a',
+    context: [ROOT_TOKEN],
+    thoughtsRanked: [{ value: 'a', rank: 0 }]
+  },
+  { type: 'cursorBack' },
+  { type: 'cursorUp' },
+  {
+    type: 'existingThoughtChange',
```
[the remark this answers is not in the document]
anmolarora1: Agreed

src/shortcuts/__tests__/undo.ts
```diff
@@ -0,0 +1,295 @@
...
+      - A
+      - B`
+    ),
+    { type: 'alert', value: 'test' },
+    { type: 'archiveThought' },
+  ])
+
+  expect(store.getState().inversePatches.length).toEqual(0)
+
+})
```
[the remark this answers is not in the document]
anmolarora1: That's a valid point. We should be looking into the `inversePatch` to verify `alert` is actually omitted. 
On a side note, the `inversePatches` is empty because the cursor value is null, and hence, the `archiveThought` doesn't update the state. Since there's no change in the state, no patch is generated.
I've updated the test nevertheless, to work as intended.

src/shortcuts/__tests__/redo.ts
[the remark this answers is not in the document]
raineorshine: It is already working as expected. Just add a `cursorDown` action and we'll at least have the test that demonstrates that navigation actions following the edit are not grouped with the edit on redo.

raineorshine:
Beautiful! Thanks!

raineorshine (approved):

PR #815, opened 2020-07-27
Test: existingThoughtChange

### LINKED ISSUE #803: Test: existingThoughtChange
Add unit tests for [`existingThoughtChange`](https://github.com/cybersemics/em/blob/dev/src/reducers/existingThoughtChange.ts). You can ignore Context View for now. 

Follow the basic pattern used by existing reducer tests like [`existingThoughtDelete`](https://github.com/cybersemics/em/blob/dev/src/reducers/__tests__/existingThoughtDelete.js) or [`archiveThought`](https://github.com/cybersemics/em/blob/dev/src/reducers/__tests__/archiveThought.js).

## Test Cases
- edit a thought
  - Test both thoughtIndex (using `getContexts`) and contextIndex (using `getThoughts`)
- edit a thought with descendants
  - Test `getContexts` and `getThoughts` of all descendants
- edit a thought that exists in another context
- update the cursor with the new thought value

aqkhan:
Fixes #803

### DIFF
--- a/src/reducers/__tests__/existingThoughtChange.js
+++ b/src/reducers/__tests__/existingThoughtChange.js
@@ -1,0 +1,36 @@
+import { ROOT_TOKEN } from '../../constants'
+import { initialState, reducerFlow } from '../../util'
+import { exportContext, getContexts, getThoughtsRanked } from '../../selectors'
+import { existingThoughtChange, newThought, setCursor } from '../../reducers'
+
+it('edit a thought', () => {
+
+  const steps = [
+    newThought({ value: 'a' }),
+    newThought({ value: 'b' }),
+    setCursor({ thoughtsRanked: [{ value: 'a', rank: 0 }] }),
+    existingThoughtChange({
+      newValue: 'c',
+      oldValue: 'a',
+      context: [ROOT_TOKEN],
+      thoughtsRanked: [ { value: 'a', rank: 0 }, { value: 'b', rank: 1 } ]
+    })
+  ]
+
+  // run steps through reducer flow and export as plaintext for readable test
+  const stateNew = reducerFlow(steps)(initialState())
+  console.log('stateNew', stateNew)
+  const exported = exportContext(stateNew, [ROOT_TOKEN], 'text/plaintext')
+
+  expect(exported).toBe(`- ${ROOT_TOKEN}
+  - a
+  - b`)
+
+  // b should exist in the ROOT context
+  // expect(getContexts(stateNew, 'b'))
+  //   .toMatchObject([{
+  //     context: [ROOT_TOKEN],
+  //     rank: -1,
+  //   }])
+
+})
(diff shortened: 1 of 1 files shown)


src/reducers/__tests__/existingThoughtChange.js
```diff
@@ -10,27 +10,176 @@ it('edit a thought', () => {
...
+
+  // aa should exist in ROOT context
+  expect(getContexts(stateNew, 'aa'))
+    .toMatchObject([{
+      context: [ROOT_TOKEN]
+    }])
+  expect(getThoughts(stateNew, [ROOT_TOKEN]))
+    .toMatchObject([{ value: 'b', rank: 1 }, { value: 'aa', rank: 0 }])
+
+  expect(stateNew.cursor)
```
aqkhan, line 35: Cursor does not update upon editing the active thought.

src/reducers/__tests__/existingThoughtChange.js
```diff
@@ -10,27 +10,176 @@ it('edit a thought', () => {
...
+  const exported = exportContext(stateNew, [ROOT_TOKEN], 'text/plaintext')
+
+  expect(exported).toBe(`- ${ROOT_TOKEN}
+  - a
+    - ab
+  - b
+    - ab`)
+
+  // ab should exist in both contexts a and b
+  expect(getContexts(stateNew, 'ab'))
```
aqkhan, line 167: `getContexts` return the context b with `rank: 0`.

raineorshine:
It looks like the tests are failing. Can you provide an update on this? Is this still in progress?

I couldn't tell whether your comments were notes to yourself, or questions, or something else. Thanks!

src/reducers/__tests__/existingThoughtChange.js
aqkhan: @raineorshine The state.cursor in this case returns the original path, before edit, i.e `[ { value: 'a', rank: 0 } ]`

src/reducers/__tests__/existingThoughtChange.js
aqkhan: @raineorshine 
`getContexts` in this case returns the following:
`[
      {
        context: [ 'a' ],
        id: '266e7ff0-aa5c-4927-a8a2-bcc2591b8e85',
        rank: 0
      },
      {
        context: [ 'b' ],
        rank: 0,
        id: '0783bcd8-c2b8-4e48-9487-a530b3b44be2'
      }
    ]`
Notice that the `b` should have `rank: 1`, but instead it returns `rank: 0`, which is not desirable.

aqkhan:
@raineorshine The comments are notes/reasons for why the tests are failing. Just added more details. The methods does not seem to return the desired values. Let me know how should I proceed. Thanks!

src/reducers/__tests__/existingThoughtChange.js
raineorshine: Good catch. It appears that `existingThoughtChange` does not update the `cursor` at all. This will be fixed separately.

src/reducers/__tests__/existingThoughtChange.js
raineorshine: This is correct actually. Since `getContexts` returns `ThoughtContexts` not `Child` nodes, `rank` represents the rank of the value within each context. i.e. in `Context` `['a']` the thought `ab` has rank `0`, and in `Context` `['b']` the thought `ab` also has rank `0`. Ranks are only unique within the same `Context`.

PR #818, opened 2020-07-28
adds check to stop setCursor dispatch when thought value is same

### LINKED ISSUE #814: Remove the setCursor action on new thought creation
## Current Behavior

When a new thought is created by pressing the enter key, the actions `newThought` and `setCursor` are being dispatched. The `setCursor` is a gratuitous action that updates the `dataNonce`, which is already being updated by the former action.

## Expected Behavior

The `setCursor` action shouldn't be dispatched on the creation of a new thought.

anmolarora1:
fixes #814 

**Issue:**
The setCursor action was being dispatched on every `onFocus` event of the `Editable(Thought)` component. On every `enter` keypress, a new thought is created, and brought to focus, hence dispatching the `setCursor` action fatuously.

**Proposed solution:**
Dispatching the `setCursor` action which ensures that it's fired only when the `thoughtsResolved` and `cursor` differ

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -387,7 +387,7 @@ const Editable = ({ disabled, isEditing, thoughtsRanked, contextChain, cursorOff
         // clicking a different thought (when not editing)
         (!state.editing && !equalPath(thoughtsResolved, state.cursorBeforeEdit))
 
-      const thoughtChanged = !state.cursor || thoughtsResolved.some((thought, index) => state.cursor![index].id !== undefined && state.cursor![index].id !== thought.id)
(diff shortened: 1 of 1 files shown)


anmolarora1:
> Unfortunately this breaks the caret on mobile (Safari, iPhone 11).
> 
> When I enter [editing mode](https://github.com/cybersemics/em/wiki/Technical-Overview#9-mobile-editing-mode) and tap on a different thought, the caret does not move with the cursor.
> Also, when I enter editing mode and delete a thought, the caret is removed.

I've added a check that seems to rectify the issue. Please review whenever you can

raineorshine:
I suggest moving the check up into `setCursorOnThought`: https://github.com/cybersemics/em/pull/828

That said, if the condition is disabled on mobile and an extra `setCursor` is dispatched, wouldn't that result in the same problem for undo/redo?

raineorshine (commented):
Alternative: #828

PR #820, opened 2020-07-30
Refactor import html

### LINKED ISSUE #714: Refactor importHTML
`importHTML` currently converts HTML directly into thought updates. Instead, break it up into two distinct functions that can be used independently as well as in a pipeline. 

1. Convert HTML to JSON (lossy) - Responsible for parsing the HTML and converting it to a strict hierarchy. 
1. Convert JSON to thought updates (lossless)

pushkov-fedor:
Fixes #714 

One test still fails and code needs to be cleaned.

I'm a little confused with this test and can't figure out what's the problem. Test, i'm talking about, is `merge duplicate with new rank` of existingThoughtMove.js. Tree for this test should look like this:
  - a
    - m
      - x
  - m
    - y

I've logged import html structure and it looks like this: `<li>a<ul><li>m<ul><li>x</li></ul></li></ul></li><li>m<ul><li>y</li></ul></li>`. If you paste it to em, the tree will be exactly as expected. I would like to clarify one moment: `/m` thought after importing has rank of 4. I thought that if we don't have any other thoughts, after pasting `/a` thought should have rank **0** and `/m` should have rank **1**. Currently after importing `/a` has rank 1 and `/m` has rank 4. Especially interesting that `/a/m` has rank 2 and `/a/m/x` has rank 3. I left the rankStart logic exactly the same as in dev branch https://github.com/cybersemics/em/blob/07b115563b9c81bb8a671b4749af929213683f8d/src/util/importHTML.ts#L66 so rank 1 of `/a` is calculated by it. I modified code in order to instead of this
  - a (1)
    - m (2)
      - x (3)
  - m (4)
    - y (5)

get tree like this (we don't have any other thoughts)
  - a (1)
    - m (0)
      - x (0)
  - m (2)
    - y (0)

But these changes lead to failing a couple of other tests (which were passed before). In addition, in test we have this line
https://github.com/cybersemics/em/blob/07b115563b9c81bb8a671b4749af929213683f8d/src/reducers/__tests__/existingThoughtMove.js#L297
As i understand, `/m` is expected to have rank **5**. At this point i'm totally confused, and it seems to me that i don't understand something about rank logic. Could you please clarify this moment?


raineorshine:
> I would like to clarify one moment: `/m` thought after importing has rank of 4. I thought that if we don't have any other thoughts, after pasting `/a` thought should have rank **0** and `/m` should have rank **1**. Currently after importing `/a` has rank 1 and `/m` has rank 4. Especially interesting that `/a/m` has rank 2 and `/a/m/x` has rank 3. I left the rankStart logic exactly the same as in dev branch.

The `rank` is evaluated relative to its siblings, so the absolute value does not matter. Creating a new thought in an empty context (with `newThought`) happens to give the thought rank `0`, but it could give it any value. 

When there are siblings, the new thought will receive a rank that is half-way in between its surrounding siblings (or a simple integer decrement or increment for adding to the beginning or end of the context, respectively). This is done for efficiency; it allows new thoughts to be added at any position without having to modify the existing ranks.

`importHTML` simply autoincrements the ranks across contexts for efficiency. This results in thoughts within a context having ascending ranks. The start rank does not matter.

From https://github.com/cybersemics/em/wiki/Technical-Overview:

> The first (chronological) thought in a context has `rank: 0`, but the absolute value does not matter. It is only used to compare with other ranks to determine sort order. Ranks can be negative. For example, creating a new thought before `{ value: 'A', rank: 0 }` will yield `{ value: 'MyThought', rank: -1 }`. Creating a new thought in between two thoughts will result in a rank that is numerically halfway between them (e.g. `-0.5`). The user may change the rank by dragging and dropping the thought. 

Hope that helps!

> But these changes lead to failing a couple of other tests (which were passed before). 

Yes, I do not know the implications of changing the ranking logic in this way. I would recommend preserving it as-is unless you have a good reason to change it.

> In addition, in test we have this line
> https://github.com/cybersemics/em/blob/07b115563b9c81bb8a671b4749af929213683f8d/src/reducers/__tests__/existingThoughtMove.js#L297
> 
> 
> As i understand, `/m` is expected to have rank **5**. At this point i'm totally confused, and it seems to me that i don't understand something about rank logic. Could you please clarify this moment?

In this case, the thoughts are imported in this order (shown with ranks):

0: `/a`
1: `/a/m`
2: `/a/m/x`
3: —
4: —
5: `/m`
6: `/m/y`

It appears that the rank is unintentionally incremented after popping out of each context. However as described above, this has no effect on the relative ranks within each context.

Hopefully this helps, but if you get stuck and want me to evaluate why `merge duplicate with new rank` is failing, just let me know!

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+interface InsertThoughtOptions {
+  indent?: boolean,
+  outdent?: boolean,
+  insertEmpty?: boolean,
+}
+interface ThoughtJSON {
+  scope: string,
+  children: ThoughtJSON[],
+}
```
raineorshine, line 26: Let's use the `Block` type that is already declared in `importText.ts`. I'm not sure if it ends up being used in more than one file, but if so you can add it to the general `types.ts`. Otherwise define it only in the file that uses it.

pushkov-fedor:
Thanks you for detailed explanation! I've read Technical Overview and understand that absolute value of rank doesn't matter in general.
But i thought that for **empty database** creating this tree iterative by hand or by importing html should lead to the same rank result.
- a (0)
   - m (0)
    - x (0)
- m (1)
  - y (0)

> It appears that the rank is unintentionally incremented after popping out of each context. However as described above, this has no effect on the relative ranks within each context.

This explains my problem! Let me know please what's the reason for this decision? Because in Technical Overview says that 

> The first (chronological) thought in a context has rank: 0, but the absolute value does not matter 

Considering this, the first thought in a context (if context doesn't have any thoughts) should have rank 0. But after importing we get this: 

0: /a
1: /a/m
2: /a/m/x
3: —
4: —
5: /m
6: /m/y

Should we set ranks this way after importing? Correct me please, if i'm wrong somewhere.

raineorshine:
> But i thought that for **empty database** creating this tree iterative by hand or by importing html should lead to the same rank result.

No, that's not the case.

> This explains my problem! Let me know please what's the reason for this decision? 

It seems to be just an outcome of the current `importHTML` implementation. It is efficient to simply autoincrement across multiple contexts.

> Technical Overview says that
> > The first (chronological) thought in a context has rank: 0, but the absolute value does not matter

I will update the Technical Overview! That is misleading since it does not apply to `importHTML`.

> Should we set ranks this way after importing? Correct me please, if i'm wrong somewhere.

The first thought in a context does not need to have rank 0. I'm not going to introduce a constraint that is not needed. Relative ranks are perfectly functional and actually more efficient.

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,161 @@
...
+      } as PreBlock
+    }
+    // WorkFlowy import with notes
+    if (nodes.some(node => node && !Array.isArray(node) && node.scope === '=note')) {
+      const parent = _.first(nodes)
+      const children = _.tail(nodes)
+      return Object.assign({}, parent, { children: children.flat() }) as PreBlock
+    }
+    if (nodes.every(node => Array.isArray(node))) return nodes.flat()
+    if (nodes.some(node => Array.isArray(node))) {
```
raineorshine, line 65: By the 4th `if` statement, the nodes have been iterated over up to 4 times. See if you can use a single loop to determine the appropriate return value, i.e. Refactor `joinChildren` to be `O(n)` instead of `O(4n)`.

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,161 @@
...
+      }
+      if (node.type === 'comment') return false
+      // remove text nodes containing only empty space and new line characters, such nodes could be created by himalaya in case of multi-line html input
+      if (node.type === 'text' && isStringIncludesOnly(node.content, ['\n', ' '])) return false
+      return node.content.length
+    }) as (Element | Text)[]
+  }
+
+  /** Append children to parent as children property if it's necessary. */
+  const joinChildren = (nodes: (PreBlock[] | PreBlock)[]): PreBlock[] | PreBlock => {
```
raineorshine, line 45: Why does `joinChildren` take `PreBlock[] | PreBlock` instead of `PreBlock[]`? Same with the return value.

In general we want to avoid polymorphic behavior in function interfaces, i.e. accepting different types of inputs or returning different types of outputs. There are some exceptions in the codebase such as when functions handle either `Path` or `context`, but usually it is undesirable.

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,161 @@
...
+        return true
+      }
+      if (node.type === 'comment') return false
+      // remove text nodes containing only empty space and new line characters, such nodes could be created by himalaya in case of multi-line html input
+      if (node.type === 'text' && isStringIncludesOnly(node.content, ['\n', ' '])) return false
+      return node.content.length
+    }) as (Element | Text)[]
+  }
+
+  /** Append children to parent as children property if it's necessary. */
```
raineorshine, line 44: Could you provide more of an explanation of the purpose of `joinChildren`? I don't totally understand.

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  /** Convert PreBlock array to Block array. */
+  const convertToBlock = (nodes: PreBlock[]): Block[] => nodes.map(node => {
+    if (node.children.length > 0) return Object.assign({}, node, { children: convertToBlock(node.children) }) as Block
+    else return node as Block
+  })
+
+  /** Converts each Array of HimalayaNodes to PreBlock. */
+  const convert = (nodes: (Element | Text)[]): PreBlock[] | PreBlock => {
```
raineorshine, line 118: Similar question about polymorphic return type

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,161 @@
+import _ from 'lodash'
+import { Element, HimalayaNode, Text, parse } from 'himalaya'
+import { Block } from '../action-creators/importText'
+
+interface PreBlock extends Block {
+  wasSpan?: boolean,
+  hasSpaces?: boolean,
```
raineorshine, line 7: Could you explain what `PreBlock` is for me, and the use of `wasSpan` and `hasSpaces`? They seem like very low-level, parsing details, so I'm curious why an intermediate structure must be created for them. Thanks!

src/util/convertHTMLtoJSON.ts
pushkov-fedor: I explain. E.g. such HTML:
```
<li>a<ul>
  <li>x</li>
  <li>y</li>
</ul></li>
```
converts by himalaya to such structure:
```
[
  {
    "type": "element",
    "tagName": "li",
    "attributes": [],
    "children": [
      {
        "type": "text",
        "content": "a"
      },
      {
        "type": "element",
        "tagName": "ul",
        "attributes": [],
        "children": [
          {
            "type": "element",
            "tagName": "li",
            "attributes": [],
            "children": [
              {
                "type": "text",
                "content": "x"
              }
            ]
          },
          {
            "type": "element",
            "tagName": "li",
            "attributes": [],
            "children": [
              {
                "type": "text",
                "content": "y"
              }
            ]
          }
        ]
      }
    ]
  }
]
```

In this JSON, thoughts **a** and **ul** are siblings. In `convert` function i iterate through tree and transform raw himalaya output to PreBlock.
For above output, after mapping and before `joinChildren` call, we have such data:
```
[
  {
    scope: 'a',
    children: null,
  },
  [
    {
      scope: 'x',
      children: null,
    },
    {
      scope: 'y',
      children: null
    }
  ]
]
```
So we have PreBlock (parent) and array of PreBlock (children). Array of PreBlock, which is placed next to PreBlock object in array, is determined as children of this object. 
Another example: 
```
[
  PreBlock (a), 
  [
    PreBlock (b), 
    [
      PreBlock, PreBlock, PreBlock
    ] (c), 
    PreBlock (d)
  ]
], 
```
results in
- a
  - b
    - PreBlock, PreBlock, PreBlock (c)
  - d


Two or more PreBlocks (**not** array) together are determined as siblings. 
E.g.
`[PreBlock, PreBlock, PreBlock, [Preblock, PreBlock]]` ->
```
- PreBlock
- PreBlock
- PreBlock
  - [PreBlock, PreBlock]
```

src/util/convertHTMLtoJSON.ts
pushkov-fedor: Answered above.

src/util/convertHTMLtoJSON.ts
pushkov-fedor: I have simplified it, now it uses only `partOfThought`. It needs for joining formatting tags.
Example: 
For such HTML:
`<b>one</b> and <i>two</i>`
Himalaya output is
```
[
  {
    "type": "element",
    "tagName": "b",
    "attributes": [],
    "children": [
      {
        "type": "text",
        "content": "one"
      }
    ]
  },
  {
    "type": "text",
    "content": " and "
  },
  {
    "type": "element",
    "tagName": "i",
    "attributes": [],
    "children": [
      {
        "type": "text",
        "content": "two"
      }
    ]
  }
]
```
Result of conversion to PreBlock:
```
[
  {
    scope: "<b>one<b/>",
    children: null,
    partOfThought: true,
  },
  {
    scope: " and ",
    children: null,
  },
  {
    scope: "<i>two<i>",
    children: null,
    partOfThought: true,
  }
]
```
As i mentioned above, PreBlock elements (**not** array of PreBlock) are determined as siblings, but in current case we need to merge such elements together, that's why i use `partOfThought`

src/util/convertHTMLtoJSON.ts
raineorshine: Thank you for the good explanation. That helps me understand the himalaya output and the conversion to the nested PreBlock structure.

I have some reservations about the PreBlock structure. In the himalaya output, the node type makes the difference between siblings and children explicit. In the PreBlock structure, this becomes *implicit* and is determined indirectly by `Array.isArray`. The notion of "children" is defined in two ways now: either explicitly in the `children` property or in the array representation. This seems like a step backwards to me in terms of clarity. I will consider next steps and which one of us should undertake them, but feel free to let me if you have any thoughts.

src/util/convertHTMLtoJSON.ts
raineorshine: I see. Thank you for the explanation. Maybe we can combine these in an earlier step to avoid the additional complexity of an intermediate step.

src/util/importJSON.ts
```diff
@@ -0,0 +1,160 @@
...
+  const lastThoughtFirstLevelIndex = thoughtsJSON.length - 1
+  const lastThoughtFirstLevel = thoughtsJSON[lastThoughtFirstLevelIndex]
+  const rankOffset = lastThoughtFirstLevelIndex === 0 ? 0 : getRankOffset(thoughtsJSON, lastThoughtFirstLevelIndex)
+  return { value: lastThoughtFirstLevel.scope, rank: rankStart + rankOffset * rankIncrement }
+}
+
+/** Recursively iterate through thoughtsJSON and call insertThought for each thought individually to save it. */
+const saveThoughts = (context: Context, rank: number, rankIncrement: number, thoughtsJSON: Block[], insertThought: (value: string, context: Context, rank: number) => void) => {
+  thoughtsJSON.forEach((thought, index) => {
+    const rankOffset = index === 0 ? 0 : getRankOffset(thoughtsJSON, index)
```
raineorshine, line 57: It looks like `getRankOffset` is called for each thought in `O(n)`, and `getRankOffset` itself loops through each thought in `O(n)` recursively, resulting in computational complexity somewhere around `O(n^2*m)`.

Please confirm and suggest a solution with more acceptable performance characteristics.

src/util/importJSON.ts
```diff
@@ -0,0 +1,160 @@
...
+    insertThought(thought.scope, context, rank + rankOffset * rankIncrement)
+    if (thought.children.length > 0) {
+      saveThoughts([...context, thought.scope], rank + (rankOffset + 1) * rankIncrement, rankIncrement, thought.children, insertThought)
+    }
+  })
+}
+
+/** Return number of contexts in ThoughtJSON array. */
+const getContextsNum = (thoughts: Block[]): number => {
+  return thoughts.map(thought => thought.children.length > 0 ? 1 + getContextsNum(thought.children) : 1).reduce((acc, val) => acc + val, 0)
```
raineorshine, line 67: For long lines like this, I would suggest breaking them up for better readability:

```js
  return thoughts
    .map(thought => thought.children.length > 0 
      ? 1 + getContextsNum(thought.children) 
      : 1
    )
    .reduce((acc, val) => acc + val, 0)

```

src/util/importJSON.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/** Calculate rankIncrement value based on rank of next sibling or its absence. */
+const getRankIncrement = (thoughtsJSON: Block[], state: State, context: Context, destThought: Child, rankStart: number) => {
```
raineorshine, line 71: It's recommended that `state` is always the first argument so it follows the convention of other selectors.

src/util/importJSON.ts
```diff
@@ -0,0 +1,160 @@
...
+  const numContexts = getContextsNum(thoughtsJSON)
+  const destValue = destThought.value
+  const destRank = destThought.rank
+  const next = nextSibling(state, destValue, context, destRank) // paste after last child of current thought
+  const rankIncrement = next ? (next.rank - rankStart) / (numContexts || 1) : 1 // prevent divide by zero
```
raineorshine, line 76: Notice that `numContexts` (`O(n)`) is always calculated even though it is only used in one branch of the ternary. How can you avoid this unnecessary calculation?

You should develop a habit of noticing where expensive functions (generally >= `O(n)`) are being used, and take steps to prevent unnecessary calls.

src/util/importJSON.ts
```diff
@@ -0,0 +1,160 @@
...
+        }
+      },
+      value,
+      rank,
+      id,
+      rootContext
+    )
+
+    thoughtIndex[hashThought(value)] = thoughtNew
+    thoughtIndexUpdates[hashThought(value)] = thoughtNew
```
raineorshine, line 135: Let's call `hashThought(value)` once and reuse it (this was in the original code).

src/util/importJSON.ts
pushkov-fedor: Yes, it is. I think we can calculate offsets directly in `saveThoughts` or before it, and save them to Map

src/util/importJSON.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/** Recursively calculate rank offset for a given position based on thoughts before and their children. Increment offset each time we pop out of context. */
+const getRankOffset = (thoughts: Block[], position?: number): number => {
+  const before = !position ? [...thoughts] : thoughts.slice(0, position)
+  const result = before.map(thought => thought.children.length > 0 ? 1 + getRankOffset(thought.children) : 1).reduce((acc, val) => acc + val, 0)
+  return position ? result : result + 1
+}
+
+/** Prepare thoughts for saving, skip root if necessary. */
```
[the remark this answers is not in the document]
pushkov-fedor: Actually, it used for skipping root thought, if `skipRoot` flag is true. My bad, I have renamed it.

src/util/importJSON.ts
pushkov-fedor: I've replaced it with one calculation for all thoughts. Check 5643990

src/util/importJSON.ts
pushkov-fedor: Yes, you are right. I replaced `numContext` with direct `getContextsNum` call.

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,156 @@
...
+  const removeEmptyNodesAndComments = (nodes: HimalayaNode[]): (Element | Text)[] => {
+    const filteredNodes = nodes.filter(node => {
+      if (node.type === 'element') {
+        if (node.tagName === 'br') return false
+        return true
+      }
+      if (node.type === 'comment') return false
+      // remove text nodes containing only empty space and new line characters, such nodes could be created by himalaya in case of multi-line html input
+      if (node.type === 'text' && isStringIncludesOnly(node.content, ['\n', ' '])) return false
+      return node.content.length
```
raineorshine, line 35: Please re-write the filter predicate as a single (nested) boolean expression. You can break the boolean onto separate lines to make it easier to read. The current use of `if` statements and explicit boolean returns does not fit the existing style of the project.

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,156 @@
...
+        }
+        else if (childNode.type === 'element' && childNode.tagName === 'ul') {
+          return {
+            scope: '',
+            children: convert(childNode.children as (Element | Text)[])
+          } as PreBlock
+        }
+      }
+      else return convert(node.children as (Element | Text)[])
+    }).filter(node => node !== undefined) as (PreBlock | PreBlock[])[]
```
raineorshine, line 145: Let's use an explicit `return null` in the appropriate place within the mapping function instead of the implicit `undefined`. Functions should never return `undefined` in my opinion unless they are a void function (i.e. always return `undefined`).

src/util/importJSON.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+interface RankInfo {
+  rank: number,
+  deepLevel: number,
+}
+
+/** If skipRoot is true, replace head block with its children, or drop it, if head has no children. */
+const skipRootThought = (skipRoot: boolean, blocks: Block[]) => {
+  if (!skipRoot) return blocks
```
raineorshine, line 37: A function should generally not include a dedicated boolean to bail. That logic should be located at the caller.

src/util/importJSON.ts
```diff
@@ -0,0 +1,209 @@
...
+/** Calculate last thought of the first level, as this is where the selection will be restored to. */
+const calculateLastThoughtFirstLevel = (rankMap: Map<Block, RankInfo>, blocks: Block[]) => {
+  const lastThoughtFirstLevelIndex = blocks.length - 1
+  const lastThoughtFirstLevel = blocks[lastThoughtFirstLevelIndex]
+  const { rank } = rankMap.get(lastThoughtFirstLevel)!
+  return { value: lastThoughtFirstLevel.scope, rank }
+}
+
+/** Return map of thought ranks. */
+const createRankMap = (blocks: Block[], rankStart: number, rankIncrement: number) => {
```
raineorshine, line 53: No, this is not the right way to do this. You have separated the ranking algorithm from the insertion algorithm and created a lot of additional computation in the process. Thoughts are already traversed in order during the insertion process, and the rank increases at regular intervals. This should not be a separate/pre-process. (Not to mention we already do a separate full traversal for `getRankIncrement`).

Does this make sense? Can you describe a better solution? I want to make sure you understand before proceeding.

raineorshine:
Regarding the [use of `PreBlock`](https://github.com/cybersemics/em/pull/820#discussion_r477684885), I would like to try to only use `Block` and remove `partOfThought` by joining the children earlier in the pipeline. This will reduce the complexity of the intermediary structure. I could be wrong, but I think this is doable. Is this something you can think you can do?

src/util/importJSON.ts
pushkov-fedor: Okay, i can we can merge `createRankMap` and `saveThoughts` together. Iterate through tree once, save ranks information in `Map`, and then retrieve info in next iterations. Result function will be a bit complex and large, but we will iterate only once.

pushkov-fedor:
> Regarding the [use of `PreBlock`](https://github.com/cybersemics/em/pull/820#discussion_r477684885), I would like to try to only use `Block` and remove `partOfThought` by joining the children earlier in the pipeline. This will reduce the complexity of the intermediary structure. I could be wrong, but I think this is doable. Is this something you can think you can do?

Okay, give me some time for coming up with updated algorithm for this and i will tell you my thoughts. If it's okay, i will implement it.

src/util/importJSON.ts
raineorshine: No, we don't want to iterate a separate time, and we don't want a separate `Map`. That is effectively what you have already done, it would just be wrapped in a single function. It doesn't change the performance characteristics. You need to do it all in one iteration. Additional iterations are not necessary. Can the ranks be generated within `insertThought`? That's what we want.

pushkov-fedor: Actually, i don't think that it's a good idea. Purpose of `insertThought` is saving thoughts, and i don't think that it needs additional computation inside it. I like that it takes minimum required arguments: `value`, `context` and `rank` and saves thought. It's simple function and its purpose is clear. 
Maybe you don't understand me correctly, but `saveThoughts` and `calculateRanks` are the same in terms of iteration. They both recursively iterate through tree and do some operations on each iteration. I meant that we can substitute these 2 functions with single one, because they merge easily. On first level of recursion we can create `Map` for storing previous ranks and use this `Map` only inside this single function.
We need to know the rank of previous saved thought and how deep this thought is inside the tree.
E.g. If we recursively iterate such tree
- a
  - b
    - c
- d

on `d` thought iteration we need to know `RankInfo` of `c` thought. It could be retrieved from global mutating object (this approach was used in previous version, where `let` variable was used for it), or Map.

raineorshine: > Actually, i don't think that it's a good idea. Purpose of `insertThought` is saving thoughts, and i don't think that it needs additional computation inside it. I like that it takes minimum required arguments: `value`, `context` and `rank` and saves thought. It's simple function and its purpose is clear.

Ah, yes, you're right, `insertThought` is not the place for it.

> Maybe you don't understand me correctly, but `saveThoughts` and `calculateRanks` are the same in terms of iteration. They both recursively iterate through tree and do some operations on each iteration. I meant that we can substitute these 2 functions with single one, because they merge easily.

Yes! Exactly.

> We need to know the rank of previous saved thought and how deep this thought is inside the tree.

Ranks are only relative to siblings, as I have documented and we have discussed in depth elsewhere. So they don't need to know how deep the thought is inside the tree. They just need to know the index within their parent. The previous implementation autoincremented the whole tree for efficiency. However, it's not the only possible implementation that meets the rank requirements. If it's more efficient now to increment siblings only, do that. Or if there is an efficient way to do a global autoincrement, that would be fine, too. Either way, the rank should be generated within the main traversal function, as you pointed out.

> On first level of recursion we can create `Map` for storing previous ranks and use this `Map` only inside this single function. We will iterate tree only **once**.

(I quoted this out of order since it may be invalidated by the above paragraph. Feel free to only respond to these questions if they are still relevant.)

By "first level of recursion" do you mean the first set of children? i.e. just "when the iteration starts"? I thought you meant "first iteration", my bad.

If you're only iterating the tree once, what is the purpose of `Map`? If the ranks are generated on the fly, I don't see why you would need to store them in a global structure. The starting rank can be threaded through the traversal and incremented as needed, unless I'm mis-assessing the situation.

pushkov-fedor: Okay, i was a bit confused with previous ranking implementation and that fact that some tests is based on this implementation. If it's not necessary to increment rank each time we save thought/pop out of context, then i would like to implement ranking based on siblings
Instead of
- a (0)
  - b (1)
    - c (2)
- d (5)
- d (6)

it will be (of course ranking could be a little different in "real life", `startRank` and `rankIncrement` could differ from this example, but i hope you get the idea)
- a (0)
  - b (0)
    - c (0)
- d (1)
- e (2)

Such implementation is much more easier and doesn't require any additional variables for storing previous ranks.

raineorshine: 💯

raineorshine: > some tests is based on this implementation.

Yes, this will be a good opportunity to remove those misleading dependencies. Tests that do not rely on specific ranks should generally not be testing specific ranks. We can use `matchObject` instead of `toEqual` in some cases. I also created [`setCursorFirstMatch`](https://github.com/cybersemics/em/blob/dev/src/test-helpers/setCursorFirstMatch.ts) so that we could `setCursor` without depending on rank.

pushkov-fedor:
@raineorshine Raine, hi! I have changed ranking algorithm, but tests, based on previous ranking, fail. I changed some of them, but i'm not sure how to implement it in correct way.

raineorshine:
You'll want to fix the linter errors. That is preventing the tests from running in the GitHub Action, so I am unable to see which tests are failing right here on the PR.

> I have changed ranking algorithm

So much better!

> I changed some of them, but i'm not sure how to implement it in correct way.

You can do this.The tests are not changing in any fundamental way. Go through the failing tests one at a time. Where is it failing? What is causing it? Is it asserting a rank based on the previous algorithm? If so, change the assertion so it asserts the expected new rank. The tests were all passing before the ranking algorithm change, so we can assume that if the test is failing, it must be solely due to the rank change. 

The easiest approach will be to test exact ranks. A more flexible approach would only test the minimum requirement: that ranks of siblings are unique and in the correct order. I'm okay with either approach at this point.

If you get stuck, link to the test that you are unsure about, explain the problem, and I will offer assistance.

src/action-creators/__tests__/cursorPrev.js
```diff
@@ -7,21 +7,21 @@ import {
...
+    store.dispatch(importText(RANKED_ROOT, `
       - a
         - a1
-      - b`),
-      { type: 'setCursor', thoughtsRanked: [{ value: 'b', rank: 1 }] },
-      cursorPrev()
-    ])
+      - b`))
+
+    setCursorFirstMatch(['b'])(store.getState())
```
raineorshine, line 23: This line won't do anything, since you're not using the return value. `setCursorFirstMatch` is a reducer, so it returns the new state.

raineorshine (changes requested):
> Raine, hi! I've fixed all failing tests in [ff0e2b6](https://github.com/cybersemics/em/commit/ff0e2b68730edcb124725cd80ed6ff1a23d6ce35), but i believe it's not the best way to do it. I don't like that ranks is hardcoded.

Good work!

For the purpose of testing, I actually think the hardcoded ranks are good. Most of those functions end up changing the rank (e.g. `existingThoughtMove`) and we want to make sure it changes exactly as expected. Having it tightly coupled to `importText` in this small way seems like a reasonable tradeoff to me. The alternative would be to determine the existing rank, then calculate the expected relative rank for the assertion, which I think would make the tests harder to read and more error prone without much benefit.

> In [5c09bba](https://github.com/cybersemics/em/commit/5c09bba9c8b232566537874f3746fa103370db17) i tried to replace hardcoded ranks in `setCursor` with `setCursorFirstMatch`. It works for `cursorPrev`, but doesn't work for others.

See comment below.

pushkov-fedor:
Okay, thanks, in this case i can revert 5c09bba.

raineorshine:
> Okay, thanks, in this case i can revert [5c09bba](https://github.com/cybersemics/em/commit/5c09bba9c8b232566537874f3746fa103370db17).

I'm pretty sure you will still need to set the cursor in some cases. You just need to do it so it actually changes the state.

raineorshine (changes requested):

src/util/convertHTMLtoJSON.ts
raineorshine: @pushkov-fedor

src/action-creators/__tests__/cursorPrev.js
raineorshine: @pushkov-fedor

raineorshine:
Hi Fedor. The new ranking calculations look good. 

There were some comments that were unaddressed. I tagged you on them. I would appreciate more thoroughness in responding to and addressing all requests for changes. You should generally address all comments, and if one has not been addressed, let me know explicitly that you're still working on it. If you're requesting a partial review, let me know that, otherwise I assume it's supposed to be a full review and waste time tagging you on unaddressed comments which I'm not sure if you're aware of or not.

Comments that GitHub hides automatically when the file changes are the easiest to lose track of.

> > Regarding the [use of `PreBlock`](https://github.com/cybersemics/em/pull/820#discussion_r477684885), I would like to try to only use `Block` and remove `partOfThought` by joining the children earlier in the pipeline. This will reduce the complexity of the intermediary structure. I could be wrong, but I think this is doable. Is this something you can think you can do?
> 
> Okay, give me some time for coming up with updated algorithm for this and i will tell you my thoughts. If it's okay, i will implement it.

What's your status on factoring out `PreBlock`? I think it's the last big piece.

Thanks :)

src/util/convertHTMLtoJSON.ts
```diff
@@ -0,0 +1,161 @@
...
+interface PreBlock extends Block {
+  wasSpan?: boolean,
+  hasSpaces?: boolean,
+}
+
+/** Parses input HTML and saves in JSON array using Himalaya. */
+export const convertHTMLtoJSON = (html: string) => {
+
+  /** Check if string contains only characters from given 'includes' array. */
+  const isStringIncludesOnly = (str: string, includes: string[]) => {
```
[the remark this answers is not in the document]
pushkov-fedor: Sure, replaced.

src/util/convertHTMLtoJSON.ts
pushkov-fedor: We can use `for` loop for this. It doesn't look elegant, but performance is better.

pushkov-fedor:
> Hi Fedor. The new ranking calculations look good.
> 
> There were some comments that were unaddressed. I tagged you on them. I would appreciate more thoroughness in responding to and addressing all requests for changes. You should generally address all comments, and if one has not been addressed, let me know explicitly that you're still working on it. If you're requesting a partial review, let me know that, otherwise I assume it's supposed to be a full review and waste time tagging you on unaddressed comments which I'm not sure if you're aware of or not.
> 
> Comments that GitHub hides automatically when the file changes are the easiest to lose track of.
> 
> > > Regarding the [use of `PreBlock`](https://github.com/cybersemics/em/pull/820#discussion_r477684885), I would like to try to only use `Block` and remove `partOfThought` by joining the children earlier in the pipeline. This will reduce the complexity of the intermediary structure. I could be wrong, but I think this is doable. Is this something you can think you can do?
> > 
> > 
> > Okay, give me some time for coming up with updated algorithm for this and i will tell you my thoughts. If it's okay, i will implement it.
> 
> What's your status on factoring out `PreBlock`? I think it's the last big piece.
> 
> Thanks :)

Yes, i understand. Sorry for such unclearness in review asking. Next time i will directly tell you what part of PR is waiting for review.

About PreBlock:
We have such line:
`const preBlocks = convert(removeEmptyNodesAndComments(nodes))`. Here i clear himalaya output from empty nodes, and then pass it to `convert` function. I think we can write additional function, which would be called before `convert` execution. In this function i will iterate through clear nodes tree and convert formatting tags or group of formatting tags into single 'text' node. Then formatted tree would be passed to `convert` function, where we will simply convert it to JSON. What do you think?
How i see it would be called:
`const preBlocks = convert(handleFormattingTags(removeEmptyNodesAndComments(nodes)))`

raineorshine:
> About PreBlock:
> We have such line:
> `const preBlocks = convert(removeEmptyNodesAndComments(nodes))`. Here i clear himalaya output from empty nodes, and then pass it to `convert` function. I think we can write additional function, which would be called before `convert` execution. In this function i will iterate through clear nodes tree and convert formatting tags or group of formatting tags into single 'text' node. Then formatted tree would be passed to `convert` function, where we will simply convert it to JSON. What do you think?
> It would look something like this:
> `const preBlocks = convert(handleFormattingTags(removeEmptyNodesAndComments(nodes)))`

We want to avoid anything that adds additional iterations. Use the iterations that are already there. More importantly, we want to isolate the formatting tag complexity to a single place and eliminate the `PreBlock` intermediary structure.

The main issue is within `convert`. It relies on the intermediary `PreBlock` structure which adds additional complexity that is not needed. Think about it: You are doing work to generate the `PreBlocks`, then doing work to turn them into `Blocks`. Why not make `Blocks` to begin with? The idea is to convert formatting tags or groups of formatting tags into a proper `Block` the first time they are encountered—at `convertFormattingTags`. Rather than generating `PreBlocks`, consolidate the nodes into `Blocks` right there. 

With the new implementation, the logic for converting formatting tags into regular `Blocks` is still there, but it is limited to a function call at a single point within the himalaya conversion. When you convert it to `PreBlock` early and convert back to `Block` later, then everything in between has to "know" about `PreBlock` when it really doesn't have to. `PreBlock` spreads that one piece of logic around throughout many more functions. It would be better to isolate it to as small a space as possible. Does that make sense?

This is an important principle for good architecture, so I hope I have explained it well! When I learned this concept my programming skills really leveled up. Your job as a programmer is not to make things work; that's only the most basic step. Your job as a programming is to manage complexity.

pushkov-fedor:
Thank you for explanation! I really appreciate it. 

In my opinion, `convertFormattingTags` in current implementation isn't correct place to do it. It works with single Element (`i`,`b`,`span`). In some cases we need to merge formatting tags in a single element. e.g. `a<span>fter</span>word`. So we need information not only about current element, but about its siblings too. Then after merging the whole group of elements should be replaced with single element. In current implementation in `convert` function we retrieve information from himalaya nodes, converts nodes to JSON, based on their type and tags, and then reorganize `Block` tree in `joinChildren`. Original himalaya tree structure differs from desired Block structure.

Maybe it would be better if we handle and reorganize himalaya tree to our desired structure (it's what `joinChildren` function does now) before `convert` call, and on this step we handle formatting tags and merge them if it's necessary. And after that we simply iterate through reorganized tree and replace himalaya nodes with Block (what `convert` function does now). It sounds more logical.

raineorshine:
> In my opinion, `convertFormattingTags` in current implementation isn't correct place to do it. It works with single Element (`i`,`b`,`span`). In some cases we need to merge formatting tags in a single element. e.g. `a<span>fter</span>word`. So we need information not only about current element, but about its siblings too. Then after merging the whole group of elements should be replaced with single element.

Yes, it would need to be done one step up, on the parent. A function operating on the parent node can read the children and recognize if there are formatting tags that need to be merged.

> Maybe it would be better if we handle and reorganize himalaya tree to our desired structure (it's what `joinChildren` function does now) before `convert` call, and on this step we handle formatting tags and merge them if it's necessary. And after that we simply iterate through reorganized tree and replace himalaya nodes with Block (what `convert` function does now). It sounds more logical.

That approach does an additional traversal. There is no reason to split the functionality into two different traversals.

What you're suggesting is analogous to:

```js
blocks.map(foo).map(bar)
```

What I'm suggesting is:

```js
blocks.map(compose(bar, foo))
```

It is the same functionality, but fewer iterations.

You need to think through how to do it in a single iteration. You are breaking it into multiple iterations to make it easier on your brain, but you are sacrificing performance. The end result of a single iteration approach will be just as readable, if not more, because it will be in one place.

pushkov-fedor:
Raine, hi! I have encountered some problems with Workflowy and its html structure.
Screenshot from 2020-09-30 18-13-25

It slightly differs from previous html structure in test, because it has `br` tag, which separate `a` parent tag from its children (all tags after `br`). Also we have `ul` tag after `Note span`.
Usually `ul` tag determines children of its previous sibling.
For example:
Screenshot from 2020-09-30 18-13-55

But it doesn't work the same way with Workflowy, so i should also handle this situation. Now everything works.
Also i simplified `joinChildren` logic and removed additional loops.
But we need 2 loops at the beginning of `convert` function in order to determine whether nodes include formatting tags (which should be joined) or `br` tag.

raineorshine:
> i simplified `joinChildren` logic and removed additional loops.

Great, thank you.

> But we need 2 loops at the beginning of `convert` function in order to determine whether nodes include formatting tags (which should be joined) or `br` tag.

This is acceptable since it is contained within the `convert` function and only applies to the current level, as opposed to a full traversal.

> But it doesn't work the same way with Workflowy, so i should also handle this situation. Now everything works.

That's a good catch, but we need to make sure it works for other uses of `<br>`. `<br>` tags will not always be accompanied by a `<ul>`. We should insert multiple thoughts for text separated by `<br>` tags. I added the missing tests for this case. It would be great if you could update the code to get the tests passing.

I did some widespread refactoring in 4fe2c4b2a. I recommend looking closely at this commit for areas of improvement. Some of the main issues:

- Overuse of explicit type casts
- Simplify logic in multiple places
- Move helper functions outside `convertHTMLtoJSON` function scope, since they are encapsulated by the module scope already.
- Rename functions that start with `convert` since it is doesn't add any meaning (i.e. all functions "convert" something to something else; that's essentially the definition of a function).
- Consolidate return statements (especially in the `convert` function)

pushkov-fedor:
@raineorshine Thanks for review and explanation! I really appreciate it. I have fixed new tests.
I would like to discuss my solution. Now we have 2 different conversion for `br` tag.
The first one is Workflowy:
Screenshot from 2020-10-05 15-26-22
`br` here is used for saving all tags after `br` as children.
The second one is tests that you've added.
Screenshot from 2020-10-05 15-27-51
`br` here is used as simply new line indicator.

Syntactically, they are the same. I decided to use small workaround for Workflowy. If the next element after `br` is workflowy `span`, i convert all nodes after `br` as children. In the another case, i filter all `br`, because they're unnecessary. What do you think about it?

raineorshine:
> Syntactically, they are the same. I decided to use small workaround for Workflowy. If the next element after `br` is workflowy `span`, i convert all nodes after `br` as children. In the another case, i filter all `br`, because they're unnecessary. What do you think about it?

Excellent! That totally solves it.

Thanks so much for all your hard work. This was a challenging task and I appreciate you sticking it out. Hopefully you learned some useful things in the process.

PR #823, opened 2020-08-05
Content Editable

shresthabijay:
`ContentEditable` component that only sets `innerHTML` blur. It prevents setting `innerHTML` when it is currently focused. This will allow `Editable` component to re-render without causing reset of `focusOffset` on `ContentEditable`.

### DIFF
--- a/src/components/ContentEditable.tsx
+++ b/src/components/ContentEditable.tsx
@@ -1,0 +1,69 @@
+import React, { useRef } from 'react'
+import { GenericObject } from '../utilTypes'
+
+interface ContentEditableProps extends React.HTMLProps<HTMLDivElement>{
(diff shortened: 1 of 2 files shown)


shresthabijay:
> Looks good! Just two regressions that I can find:
> 
> 1. The `Mount` test is failing.
> 2. The thought temporarily reverts back to the previous `innerHTML` on blur. This occurs on desktop when switching windows, or on mobile when closing the keyboard:
> 
> RPReplay_Final1596755579-save mov

@raineorshine I have made changes to make sure mount test passes. Also I have fixed issue of temporarily reverting back to the previous innerHTML on blur. Please test it. Thanks!

raineorshine:
Works great!

PR #831, opened 2020-08-17
Tests: cursorNext/cursorPrev

### LINKED ISSUE #806: Tests: cursorNext/cursorPrev
Write unit tests for [`cursorNext`](https://github.com/cybersemics/em/blob/dev/src/action-creators/cursorNext.ts) and [`cursorPrev`](https://github.com/cybersemics/em/blob/dev/src/action-creators/cursorPrev.ts). They function like `cursorUp` and `cursorDown` but skip descendants and do not move the cursor outside of the cursor context. They can be activated with `Control + ↓` and `Control + ↑`, respectively.

Instantiate a `store` using [`createTestStore`](https://github.com/cybersemics/em/tree/dev/src/test-helpers/createTestStore.js). Do not import the global `store` and do not do a full component mount. You can dispatch the action-creator thunk directly to the test `store`. 

Copy and adapt the following tests from [`reducers/__tests__/cursorDown`](https://github.com/cybersemics/em/blob/dev/src/reducers/__tests__/cursorDown.js) for both `cursorNext` and `cursorPrev`:
 
- move cursor to next sibling (or prev sibling when testing `cursorPrev`)
- move to first root child when there is no cursor
- do nothing when the cursor is on the last sibling (or first sibling when testing `cursorPrev`)
- do nothing when there are no thoughts
- work for sorted thoughts

Add one additional test:
- skip descendants

aqkhan:
Fixes #806

### DIFF
--- a/src/action-creators/__tests__/cursorNext.js
+++ b/src/action-creators/__tests__/cursorNext.js
@@ -2,7 +2,7 @@ import { RANKED_ROOT } from '../../constants'
 
 // action-creators
 import {
-  cursorPrev,
+  cursorNext,
   importText,
 } from '../../action-creators'
 
@@ -10,7 +10,7 @@ import { createTestStore } from '../../test-helpers/createTestStore'
 
 describe('normal view', () => {
 
-  it('move cursor to previous sibling', async() => {
+  it('move cursor to next sibling', async() => {
 
     const store = createTestStore()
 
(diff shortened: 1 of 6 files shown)


aqkhan:
@raineorshine I've found the following issues with the action-creators cursorNext and cursorPrev:

- Both the action-creators do not work with the sorted context as expected. 
- Both the action-creators do not move the cursor to first child of root when the cursor is not set. 

I've fixed the second issue, i.e move the cursor to first child of root when the it is not set but need more time to fix the first issue. I will move forward after your approval.

raineorshine:
Nicely done!

I did some minor refactors, but otherwise looks great:

- Moved tests to `/action-creators/__tests__/` directory.
- Removed unnecessary async/await (it's not obvious that this works since `importText` appears to be asynchronous and needs to be updated).
- Dispatch multiple actions at once using [redux-multi](https://github.com/ashaffer/redux-multi).
- Remove `rank` from `toMatchObject` since that creates an unnecessary dependency on importText's somewhat arbitrary ranks.

raineorshine:
I will merge and you can open another PR for the sort issue! Thanks!

PR #834, opened 2020-08-18
Merge sibling fix

### LINKED ISSUE #816: Edit after merge siblings broken
## Steps to Reproduce

- a
  - b
  - |c

1. Hit Backspace to merge `b` and `c`
2. Type `d` to edit
3. Hit Escape or Up

## Current Behavior

`a/bdc` is never saved and reverts to `a/bc` as soon as the cursor moves.

## Expected Behavior

`a/bdc` should be saved and editing should work as normal after merging siblings.

## Note

`editingValue` is dispatched but not `existingThoughtChange`.

shresthabijay:
fixes #816

### DIFF
--- a/src/components/ContentEditable.tsx
+++ b/src/components/ContentEditable.tsx
@@ -6,15 +6,17 @@ interface ContentEditableProps extends React.HTMLProps<HTMLDivElement>{
     html: string,
     disabled?: boolean,
     innerRef?: React.RefObject<HTMLDivElement>,
+    isEditing?: boolean,
     onChange: (originalEvt: ContentEditableEvent) => void,
 }
 
 /**
  * Content Editable Component.
  */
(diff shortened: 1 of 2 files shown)


shresthabijay:
@raineorshine The issue was the `oldValue` was not being updated after the `thoughtsRanked` being accessed from props was updated. Since `thoughtChangeHandler` checks if the `oldValue` has any associated thought using `getThought`, it prevented dispatching `exisitingThoughtChange`. I made the necessary changes to fix it. Apart from that I have made some changes to the `ContentEditable` to use `isEditing` to track focus state instead of using `document.activeElement` which was not reliable between re-renders. Please test the changes. Thanks!

raineorshine:
Thanks for the quick turnaround! Works great.

PR #838, opened 2020-08-20
Import text render fix

### LINKED ISSUE #836: Imported text not rendered
## Steps to Reproduce

1. Create an empty thought
2. Paste "a"

## Current Behavior

The thought is still empty. 

Refreshing the page or otherwise forcing the thought to re-render will show "a".

## Expected Behavior

Thought should be re-rendered as "a" as soon as the text is pasted.

## Note

The `state` is being updated, so it seems like a rendering problem.

## Test

Here is a rough draft of a unit test, but I cannot guarantee it's functionality so please review. You'll need `npm install --save-dev @types/enzyme`.

`/src/action-creators/__tests__/importText.ts`

```ts
import { ROOT_TOKEN } from '../../constants'
import { store } from '../../store'
import * as db from '../../db'
import { getThoughtsRanked } from '../../selectors'
import windowEvent from '../../test-helpers/windowEvent'
import createTestApp from '../../test-helpers/createTestApp'
import { act } from 'react-dom/test-utils'
import { ReactWrapper } from 'enzyme'
import { importText } from '../'

type TestDocument = Document & {
  // See: enzyme types
  // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/enzyme/index.d.ts
  wrapper: ReactWrapper,
}

beforeEach(async () => {
  await createTestApp()
})

afterEach(async () => {
  store.dispatch({ type: 'clear' })
  await db.clearAll()
})

it('re-render thought after importing text', async () => {

  const document = global.document as TestDocument

  // create empty thought
  windowEvent('keydown', { key: 'Enter' })
  document.wrapper.update()

  // import text
  store.dispatch(importText([{ value: '', rank: 0 }], 'a'))

  act(() => {
    jest.runOnlyPendingTimers()
  })

  // state
  const subthoughts = getThoughtsRanked(store.getState(), [ROOT_TOKEN])
  expect(subthoughts).toHaveLength(1)
  expect(subthoughts[0]).toMatchObject({ value: 'a' })

  // DOM
  document.wrapper.update()
  const editable = document.wrapper.find('div.editable')
  expect(editable.text()).toBe('a')

})
```

shresthabijay:
fixes #836

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1873,6 +1873,15 @@
         "@babel/types": "^7.3.0"
       }
     },
+    "@types/cheerio": {
+      "version": "0.22.21",
+      "resolved": "https://registry.npmjs.org/@types/cheerio/-/cheerio-0.22.21.tgz",
+      "integrity": "sha512-aGI3DfswwqgKPiEOTaiHV2ZPC9KEhprpgEbJnv0fZl3SGX0cGgEva1126dGrMC6AJM6v/aihlUgJn9M5DbDZ/Q==",
+      "dev": true,
+      "requires": {
+        "@types/node": "*"
+      }
+    },
     "@types/classnames": {
       "version": "2.2.10",
       "resolved": "https://registry.npmjs.org/@types/classnames/-/classnames-2.2.10.tgz",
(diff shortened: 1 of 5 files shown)


shresthabijay:
@raineorshine The logic to prevent `innerHTML` updates while editing thought was causing this issue. I found a better way around to do this. I also added the test you provided. Please test the changes. Thanks!

raineorshine:
Beautiful, thanks!

The only problem I encountered was that the caret gets set to `offset: 0` after pasting. The caret should go to the end of the pasted text. It was doing this before the innerHTML regression, but I'm not sure if there was special logic for it or if it was happening by default.

I am going to merge this PR since I am in need of this change, but can you create a separate PR to address the caret issue? Thanks so much!

shresthabijay:
> Beautiful, thanks!
> 
> The only problem I encountered was that the caret gets set to `offset: 0` after pasting. The caret should go to the end of the pasted text. It was doing this before the innerHTML regression, but I'm not sure if there was special logic for it or if it was happening by default.
> 
> I am going to merge this PR since I am in need of this change, but can you create a separate PR to address the caret issue? Thanks so much!

The caret offset always gets set to zero after `innerHTML` is updated. We can write a function to put the caret at the end after every `innerHTML` updates. I will create a separate PR for it.

PR #844, opened 2020-08-23
Set selection after paste

shresthabijay:
@raineorshine I tried to place cursor position at the end of the pasted text after `innerHTML` is updated but it turns out quite tricky to achieve. With this PR I have set the selection at the end of the thought after paste. I will be working on a proper solution in the sidelines.


raineorshine:
Thanks! Actually, this messes up pasting at the beginning of a thought, which is not much of an improvement, so let's hold off until a proper solution.

PR #845, opened 2020-08-24
Command: Collapse Context

### LINKED ISSUE #387: Command: Collapse Context
- [x] Unit Test
 
Create a pure, functional reducer called `collapseContext` that archives a context after outdenting all its children. Use `reducerFlow` in combination with `existingThoughtMove`, `archiveThought`, and any other reducers or custom state manipulation needed.

Create a shortcut that dispatches `{ type: 'collapseContext' }`

Keyboard: <kbd>Alt</kbd> + <kbd>p</kbd>
Gesture: ←↓↑ 

e.g. 

## Before

- a
  - b [cursor]
    - c

## After

- a
  - c [cursor]

aqkhan:
Fixes #387


aqkhan:
@raineorshine I've created a reducers and added a shortcut for collapse context. There is an issue that I am facing that need to have your input:

- Somehow the app is not re-rendered when setCursor is called. The cursor is set properly with the URL updated but the existing thought value shows on the cursor, instead of the updated value. 

Let me know if you need more elaboration.

raineorshine:
That stumped me as well. I figured out it's because of the way the Thought's [React `key` is defined](https://github.com/cybersemics/em/blob/dev/src/components/Subthoughts.tsx#L556). The old implementation only re-rendered if the `context` and `rank` changed (`value` causes issues). When a thought with a single child is collapsed, it must be producing an identical key and not getting re-rendered. 

We recently added a unique `id` to each thought. That should be a much more reliable `key`.

src/reducers/collapseContext.ts
```diff
@@ -0,0 +1,41 @@
...
+        (state: State) => existingThoughtMove(state, {
+          oldPath: cursor.concat(child),
+          newPath: contextOf(cursor).concat(child),
+        })
+      ),
+      archiveThought({ path: cursor }),
+      setCursor({
+        thoughtsRanked: contextOf(cursor).concat(children[0]),
+        editing: state.editing,
+        offset: children.length
```
raineorshine: `offset` refers to the caret offset, which should be `0` here.

raineorshine:
Looks good, thanks!

PR #847, opened 2020-08-27
Fix: Child with same value as parent breaks cursor

### LINKED ISSUE #824: Child with same value as parent breaks cursor
## Steps to Reproduce

1. Create thought `a`
2. Create subthought `ab` (by typing `a` then `b`)

## Current Behavior

`cursor` becomes `[{ value: 'ab', rank: 0 }, { value: 'ab', rank: 0 }]`

## Expected Behavior

`cursor` should be `[{ value: 'a', rank: 0 }, { value: 'ab', rank: 0 }]`

## Notes

When the ranks differ, the problem does not occur:

- a
  - x
  - a

1. Create thought `a`
2. Create subthought of `a`: `x`
3. Create subthought of `a`: `a`
4. Edit `/a/a` to `/a/ab`

This will correctly result in `cursor: [{ value: 'a', rank: 0 }, { value: 'ab', rank: 1 }]`

pushkov-fedor:
Fixes #824

### DIFF
--- a/src/reducers/existingThoughtChange.ts
+++ b/src/reducers/existingThoughtChange.ts
@@ -37,9 +37,9 @@ interface Payload {
 
 /** Changes the text of an existing thought. */
 const existingThoughtChange = (state: State, { oldValue, newValue, context, showContexts, thoughtsRanked, rankInContext, contextChain }: Payload) => {
-
   if (oldValue === newValue || isDivider(oldValue)) return state
 
+  const { cursor } = state
   // thoughts may exist for both the old value and the new value
   const thoughtIndex = { ...state.thoughts.thoughtIndex }
   const value = headValue(thoughtsRanked)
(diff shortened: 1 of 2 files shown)


src/reducers/existingThoughtChange.ts
```diff
@@ -56,18 +56,12 @@ const existingThoughtChange = (state: State, { oldValue, newValue, context, show
...
   const archived = exactThought ? exactThought.archived : null
-
-  const cursorNew = state.cursor && state.cursor.map(thought => thought.value === oldValue && thought.rank === (rankInContext || rank)
-    ? { value: newValue, rank: thought.rank }
-    : thought
-  )
-
-  const newPath = thoughtsRanked.slice(0, thoughtsRanked.length - 1).concat({ value: newValue, rank: thoughtsRanked.slice(thoughtsRanked.length - 1)[0].rank })
+  const cursorNew = state.cursor && [...state.cursor.slice(0, -1), { ...head(state.cursor), value: newValue }]
+  const newPath = thoughtsRanked.slice(0, thoughtsRanked.length - 1).concat({ value: newValue, rank: rankInContext || rank })
```
raineorshine, line 64: Please do not use array methods on `Path` types. They are meant to be ADT's with clear methods of interaction. The underlying array is an implementation detail that consumers should not be aware of.

Use `head` and `contextOf`.

raineorshine:
Thanks!

PR #849, opened 2020-08-28
Show NewThoughtInstruction if there are no visible thoughts & Test

### LINKED ISSUE #843: Hide NewThoughtInstructions only when there are visible thoughts
Hide the `NewThoughtInstructions` component only when there are visible thoughts in the ROOT. Currently the message is hidden when there are any thoughts, regardless of visibility. 

## Steps to Reproduce

1. Create thought `a`
2. Archive `a`

## Current Behavior

Screen is empty. 

## Expected Behavior

Should show “Hit Enter to add a new thought”. 

## Tests

Write a unit test for the `Content` component.

shresthabijay:
fixes #843 

@raineorshine Please test the changes. Thanks.

### DIFF
--- a/src/components/Content.tsx
+++ b/src/components/Content.tsx
@@ -4,7 +4,7 @@ import classNames from 'classnames'
 import { isMobile } from '../browser'
 import expandContextThought from '../action-creators/expandContextThought'
 import { EM_TOKEN, MODAL_CLOSE_DURATION, RANKED_ROOT, ROOT_TOKEN, TUTORIAL2_STEP_SUCCESS } from '../constants'
-import { getSetting, getThoughts, hasChild } from '../selectors'
+import { getSetting, getThoughts, hasChild, isChildVisible } from '../selectors'
 import { publishMode } from '../util'
 import { Child } from '../types'
 import { State } from '../util/initialState'
@@ -34,7 +34,7 @@ const tutorialStepLocal = +(localStorage['Settings/Tutorial Step'] || 1)
 
 // eslint-disable-next-line jsdoc/require-jsdoc
 const mapStateToProps = (state: State) => {
-  const { focus, isLoading, noteFocus, search, showModal } = state
+  const { focus, isLoading, noteFocus, search, showModal, showHiddenThoughts } = state
   const isTutorial = isLoading ? tutorialLocal : hasChild(state, [EM_TOKEN, 'Settings', 'Tutorial'], 'On')
 
   // @typescript-eslint/eslint-plugin does not yet support no-extra-parens with nullish coallescing operator
@@ -43,7 +43,7 @@ const mapStateToProps = (state: State) => {
   const tutorialStep = isLoading ? tutorialStepLocal : +(getSetting(state, 'Tutorial Step') ?? 1)
 
   // do no sort here as the new object reference would cause a re-render even when the children have not changed
-  const rootThoughts = getThoughts(state, [ROOT_TOKEN])
+  const rootThoughts = getThoughts(state, [ROOT_TOKEN]).filter(({ value, rank }) => showHiddenThoughts || isChildVisible(state, [value], { value, rank }))
 
   return {
     focus,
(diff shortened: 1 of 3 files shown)


src/components/Content.tsx
```diff
@@ -43,7 +43,7 @@ const mapStateToProps = (state: State) => {
   const tutorialStep = isLoading ? tutorialStepLocal : +(getSetting(state, 'Tutorial Step') ?? 1)
 
   // do no sort here as the new object reference would cause a re-render even when the children have not changed
-  const rootThoughts = getThoughts(state, [ROOT_TOKEN])
+  const rootThoughts = getThoughts(state, [ROOT_TOKEN]).filter(({ value, rank }) => showHiddenThoughts || isChildVisible(state, [value], { value, rank }))
```
raineorshine, line 46: I think you will have to `filter` within the `render` function. Otherwise `filter` creates a new object references and `mapStateToProps` will force a re-render even if the thoughts have not changed.

raineorshine: It's also more efficient to check `showHiddenThought` outside the `filter` predicate since it is the same for all children.

src/components/__tests__/Content.js
```diff
@@ -0,0 +1,52 @@
...
+beforeEach(async () => {
+  wrapper = await createTestApp()
+})
+
+afterEach(async () => {
+  await cleanupTestApp()
+  wrapper = null
+})
+
+it('show NewThoughtInstructions when there are no visible thoughts in the root context', async () => {
```
raineorshine, line 18: Test looks good!

raineorshine:
If you have a trivial change like a typo and it has not been merged into `dev` yet, I recommend using `git commit --amend`.

src/components/Content.tsx
shresthabijay: Agreed!

src/components/Content.tsx
```diff
@@ -43,15 +42,15 @@ const mapStateToProps = (state: State) => {
...
+  const rootThoughtsLength = (showHiddenThoughts ? getThoughts(state, [ROOT_TOKEN]) : getThoughts(state, [ROOT_TOKEN]).filter(({ value, rank }) => isChildVisible(state, [value], { value, rank }))).length
 
   return {
     focus,
     search,
     showModal,
     isTutorial,
     tutorialStep,
-    rootThoughts,
+    rootThoughtsLength,
```
raineorshine, line 53: Won't this prevent a re-render when a thought has been edited?

shresthabijay: It will re-render if the edit causes change of total number of visible children of root context. I don't think we need to re-render `Content` if edit doesn't cause visible children to change.

shresthabijay:
@raineorshine  `Thought` component test is failing in the check. But it is passing in my local setup. Can you please check if it failing for you ?

src/components/Content.tsx
raineorshine: If other children change, such as from another data source, then we will need to re-render.

raineorshine:
> @raineorshine `Thought` component test is failing in the check. But it is passing in my local setup. Can you please check if it failing for you ?

I re-ran the job and it passes. Still a timing issue in the `Thought` test that I am not able to reproduce.

shresthabijay:
> > @raineorshine `Thought` component test is failing in the check. But it is passing in my local setup. Can you please check if it failing for you ?
> 
> I re-ran the job and it passes. Still a timing issue in the `Thought` test that I am not able to reproduce.

I tried to reproduce in my local system. But I was not able to. It's quite strange.

src/components/Content.tsx
raineorshine: @shresthabijay

shresthabijay: > If other children change, such as from another data source, then we will need to re-render.

I don't understand. Can you explain briefly ?

My understanding is any changes in the store that cause number of visible child to change should re-render `Content`. Changes in other context should simply be handled by `Subthoughts` component.

raineorshine: Oh, I guess you're right. I forgot that Subthoughts would take care of it. Or maybe I was thinking of before Subthoughts' `mapStateToProps` was changed.

raineorshine (approved):

PR #850, opened 2020-08-28
Toggle parent attribute for some shortcuts and write tests

### LINKED ISSUE #842: Attribute togglers should target parent of cursor
In practical usage, when there is a need to activate an attribute toggler—`pinSubthoughts`, `toggleSort`, `toggleTableView`, `proseView`—it is more likely that the `cursor` will be on a child of the intended context rather than the context itself. This often requires `cursorUp`, toggle the attribute, then navigating back to the original `cursor`.

Change each of attribute toggler shortcuts to toggle the attribute on the parent of the cursor rather than the cursor thought itself.

- `pinSubthoughts`
- `proseView`
- `toggleSort`
- `toggleTableView`

e.g.

- a
  - c
  - b [cursor]

## Current Behavior

`toggleSort` sorts `a/b`.

## Expected Behavior

`toggleSort` should sort `a`.

## Testing

Create unit tests for each toggler. You can create a unit test in `/src/shortcuts/__tests__/` or you can convert the `exec` function to an action-creator and create a unit test in `/src/action-creators/__tests__/`.

shresthabijay:
fixes #842 
@raineorshine  I made the required changes to target cursor's parent for attribute togglers. I also wrote tests for them. Please review the changes. Thanks!

### DIFF
--- a/src/shortcuts/__tests__/pinSubthoughts.ts
+++ b/src/shortcuts/__tests__/pinSubthoughts.ts
@@ -1,0 +1,41 @@
+import { inputHandlers } from '../../shortcuts'
+import { importText } from '../../action-creators'
+import { RANKED_ROOT } from '../../constants'
+import { noop } from 'lodash'
+import { createTestStore } from '../../test-helpers/createTestStore'
+import { attributeEquals } from '../../selectors'
+
+it('toggle =pinChildren attribute of parent of cursor', async () => {
+
+  const store = createTestStore()
+
+  const { keyDown } = inputHandlers(store)
+
+  // import thoughts
+  await store.dispatch(importText(RANKED_ROOT, `
+  - a
+    - b
+      - c
+      - d
+    - e
+      - f
+      - g
+`))
+
+  await store.dispatch({ type: 'setCursor', thoughtsRanked: [
+    { value: 'a', rank: '0' },
+    { value: 'b', rank: '1' },
+  ] })
+
+  // toggle =pinChildren attribute
+  keyDown({ preventDefault: noop, key: 'p', altKey: true, shiftKey: true } as KeyboardEvent)
+
+  // parent of cursor should have =pinChildren set to true
+  expect(attributeEquals(store.getState(), ['a'], '=pinChildren', 'true')).toBeTruthy()
+
+  // toggle =pinSubthoughts attribute
+  keyDown({ preventDefault: noop, key: 'p', altKey: true, shiftKey: true } as KeyboardEvent)
+
+  // parent of cursor should not have =pinChildren attribute
+  expect(attributeEquals(store.getState(), ['a'], '=pinChildren', 'false')).toBeFalsy()
+})
--- a/src/shortcuts/__tests__/pinSubthoughts.ts
+++ b/src/shortcuts/__tests__/pinSubthoughts.ts
@@ -1,18 +1,18 @@
-import { inputHandlers } from '../../shortcuts'
 import { importText } from '../../action-creators'
-import { RANKED_ROOT } from '../../constants'
-import { noop } from 'lodash'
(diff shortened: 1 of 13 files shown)


src/shortcuts/__tests__/pinSubthoughts.ts
```diff
@@ -0,0 +1,41 @@
...
+  - a
+    - b
+      - c
+      - d
+    - e
+      - f
+      - g
+`))
+
+  await store.dispatch({ type: 'setCursor', thoughtsRanked: [
```
raineorshine, line 25: You can omit `await` since `store.dispatch` is synchronous.

src/shortcuts/__tests__/pinSubthoughts.ts
```diff
@@ -0,0 +1,41 @@
...
+  keyDown({ preventDefault: noop, key: 'p', altKey: true, shiftKey: true } as KeyboardEvent)
+
+  // parent of cursor should have =pinChildren set to true
+  expect(attributeEquals(store.getState(), ['a'], '=pinChildren', 'true')).toBeTruthy()
+
+  // toggle =pinSubthoughts attribute
+  keyDown({ preventDefault: noop, key: 'p', altKey: true, shiftKey: true } as KeyboardEvent)
+
+  // parent of cursor should not have =pinChildren attribute
+  expect(attributeEquals(store.getState(), ['a'], '=pinChildren', 'false')).toBeFalsy()
```
raineorshine, line 40: In general I recommend breaking up assertions for different behaviors into separate tests. In this test, if toggle on fails, then it will never execute the assertion for toggle off. It would be more informative to have a separate test for each. Technically we should also have a test for toggle on from `=pinChildren/false`, not just from the initial state (no `=pinChildren`).

Another area of improvement in this test is the implicit test of the keyboard shortcut. It would be better to import the shortcut directly and call `exec`. We do not need to test the keyboard shortcut itself as we don't want this test to fail if there is something wrong with `keyDown`. That should be tested separately in a shortcut test if at all.

src/shortcuts/__tests__/pinSubthoughts.ts
```diff
@@ -0,0 +1,41 @@
...
+import { attributeEquals } from '../../selectors'
+
+it('toggle =pinChildren attribute of parent of cursor', async () => {
+
+  const store = createTestStore()
+
+  const { keyDown } = inputHandlers(store)
+
+  // import thoughts
+  await store.dispatch(importText(RANKED_ROOT, `
```
raineorshine, line 15: You can actually omit `await` here. `importText` technically returns a `Promise`, but it is already resolved so `store.dispatch` returns immediately. I will change `importText` to not return a `Promise`, but Fedor is in the middle of a large refactor of that function at the moment.

src/shortcuts/__tests__/toggleTableView.ts
```diff
@@ -0,0 +1,39 @@
...
+  // set =view attribute to Table
+  keyDown({ preventDefault: noop, key: 't', altKey: true } as KeyboardEvent)
+
+  // parent of cursor should have =view set to Table
+  expect(attributeEquals(store.getState(), ['a'], '=view', 'Table')).toBeTruthy()
+
+  // remove value of Table from =view attribute
+  keyDown({ preventDefault: noop, key: 't', altKey: true } as KeyboardEvent)
+
+  // parent of cursor should not have =view attribute set to Prose
```
raineorshine, line 37: <strike>Prose</strike> Table

src/shortcuts/__tests__/pinSubthoughts.ts
shresthabijay: Sure. It makes sense.

raineorshine:
Looks great, thanks!

PR #859, opened 2020-09-03
Command: splitSentences

### LINKED ISSUE #853: Command: splitSentences
- Add a reducer `splitSentences` that splits multiple sentences within a thought into separate thoughts. 
- Add a corresponding shortcut and toolbar button. 
- Reducer unit tests

### Before

- One. Two. Three.

### After

- One.
- Two.
- Three.

### Before

- a
- One. Two. Three.
- b

### After

- a
- One.
- Two.
- Three.
- b

pushkov-fedor:
Fixes #853 
Raine, hi! It's starting point for this PR, could you please take a look?


src/reducers/__tests__/splitSentences.js
```diff
@@ -0,0 +1,42 @@
...
+import { initialState, reducerFlow } from '../../util'
+import { exportContext } from '../../selectors'
+
+import { newThought, splitSentences } from '../../reducers'
+import setCursorFirstMatch from '../../test-helpers/setCursorFirstMatch'
+
+it('split sentences with only one thought', () => {
+  const steps = [
+    newThought('One. Two. Three.'),
+    setCursorFirstMatch(['One. Two. Three.']),
```
raineorshine, line 11: `newThought` automatically sets the cursor to the newly created thought as long as `preventSetCursor` is not set to `true`, so you can omit `setCursorFirstMatch` here.

src/reducers/__tests__/splitSentences.js
```diff
@@ -0,0 +1,42 @@
...
+  - Two.
+  - Three.`)
+})
+
+it('split sentences with other thoughts nearby', () => {
+  const steps = [
+    newThought('a'),
+    newThought('One. Two. Three.'),
+    newThought('b'),
+    setCursorFirstMatch(['One. Two. Three.']),
```
raineorshine, line 29: This one is obviously needed though since the `cursor` is on `b`.

src/shortcuts/splitSentences.tsx
```diff
@@ -0,0 +1,45 @@
...
+  id: 'splitSentences',
+  name: 'Split Sentences',
+  keyboard: { shift: true, key: 's' },
+  svg: Icon,
+  canExecute: getState => getState().cursor !== null,
+  exec: (dispatch: Dispatch<SplitSentencesAction | ActionCreator>, getState: () => State) => {
+    const { cursor } = getState()
+    const sentences = headValue(cursor!)
+    const splitSentencesRegExp = /[^.!?]+[.!?]+/g
+    if (!splitSentencesRegExp.test(sentences)) {
```
raineorshine, line 29: Would `String.prototype.split` work?

src/reducers/splitSentences.ts
```diff
@@ -0,0 +1,43 @@
...
+  const rank = headRank(cursor)
+  const value = headValue(cursor)
+  const regExp = /[^.!?]+[.!?]+/g
+
+  if (!regExp.test(value)) {
+    console.log('This thought has no sentences')
+    return state
+  }
+
+  const splittedSentences = value.match(regExp)!.map(s => s.trim())
```
raineorshine, line 23: Would `String.prototype.split` work?

src/reducers/splitSentences.ts
```diff
@@ -0,0 +1,43 @@
...
+import { contextOf, head, headRank, headValue, pathToContext, reducerFlow } from '../util'
+import { existingThoughtChange, newThought, setCursor } from '../reducers'
+import { ROOT_TOKEN } from '../constants'
+
+/** .... */
+const splitSentences = (state: State) => {
+  const { cursor } = state
+  if (!cursor) return state
+  const thoughts = pathToContext(cursor)
+  const cursorContext = thoughts.length > 1 ? contextOf(pathToContext(cursor)) : [ROOT_TOKEN]
```
raineorshine, line 12: Save the result of the first `pathToContext(cursor)` to avoid calling it twice.

src/reducers/splitSentences.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const firstSentence = _.head(splittedSentences)!
+  const otherSentences = _.tail(splittedSentences)!
+  const newCursor = contextOf(cursor).concat({ ...head(cursor), value: firstSentence })
+
+  const reducers = [
+    existingThoughtChange({ oldValue: value, newValue: firstSentence, context: cursorContext, thoughtsRanked: cursor, rankInContext: rank }),
+    ...otherSentences.map(sentence => newThought({ value: sentence })),
+    setCursor({ thoughtsRanked: newCursor })
+  ]
```
raineorshine, line 38: This looks good.

I might suggest breaking `existingThoughtChange` onto multiple lines. We usually don't want lines that long.

src/reducers/splitSentences.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const splittedSentences = value.match(regExp)!.map(s => s.trim())
+
+  if (splittedSentences.length === 1) {
+    console.log('This thought has only one sentence')
+    return state
+  }
+
+  const firstSentence = _.head(splittedSentences)!
+  const otherSentences = _.tail(splittedSentences)!
```
raineorshine, line 31: Cleaner with es6 syntax:

```js
const [firstSentence, ...otherSentences] = splittedSentences
```

src/reducers/splitSentences.ts
```diff
@@ -0,0 +1,43 @@
...
+  const cursorContext = thoughts.length > 1 ? contextOf(pathToContext(cursor)) : [ROOT_TOKEN]
+  cursorContext.toString()
+  const rank = headRank(cursor)
+  const value = headValue(cursor)
+  const regExp = /[^.!?]+[.!?]+/g
+
+  if (!regExp.test(value)) {
+    console.log('This thought has no sentences')
+    return state
+  }
```
raineorshine, line 21: No logging to the console in **em**.

src/shortcuts/splitSentences.tsx
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    const splittedSentences = sentences.match(splitSentencesRegExp)!.map(s => s.trim())
+
+    if (splittedSentences.length === 1) {
+      dispatch(alert('Cannot split sentences: thought has only one sentence.', { alertType: 'splitSentencesErr2', clearTimeout: 3000 }))
+      return
+    }
+
+    dispatch({ type: 'splitSentences', splittedSentences })
```
raineorshine, line 41: `splittedSentences` is not used. Shouldn't be passed here anyway; should be in the reducer.

src/shortcuts/splitSentences.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const splitSentencesShortcut: Shortcut = {
+  id: 'splitSentences',
+  name: 'Split Sentences',
+  keyboard: { shift: true, key: 's' },
+  svg: Icon,
+  canExecute: getState => getState().cursor !== null,
+  exec: (dispatch: Dispatch<SplitSentencesAction | ActionCreator>, getState: () => State) => {
+    const { cursor } = getState()
+    const sentences = headValue(cursor!)
```
raineorshine, line 27: Call this `value`, which is the conventional variable to use for the thought's value. We don't know yet if it has sentences.

src/shortcuts/splitSentences.tsx
```diff
@@ -0,0 +1,45 @@
...
+    const splitSentencesRegExp = /[^.!?]+[.!?]+/g
+    if (!splitSentencesRegExp.test(sentences)) {
+      dispatch(alert('Cannot split sentences: thought has no sentences.', { alertType: 'splitSentencesErr1', clearTimeout: 3000 }))
+      return
+    }
+
+    const splittedSentences = sentences.match(splitSentencesRegExp)!.map(s => s.trim())
```
raineorshine, line 34: Call this `sentences` after you have renamed the existing `sentences`.

raineorshine (changes requested):
You're on the right track! Just made some small suggestions.

src/shortcuts/splitSentences.tsx
pushkov-fedor: We can use such expression: 
`const sentences = value.split(/[.!?]+/g).filter(s => s !== '').map(s => '${s.trim()}.')`

raineorshine: Would a simpler statement work?

```
value.split(' ').map(s => s.trim())
```

src/reducers/__tests__/splitSentences.js
```diff
@@ -0,0 +1,41 @@
...
+  const stateNew = reducerFlow(steps)(initialState())
+  const exported = exportContext(stateNew, [ROOT_TOKEN], 'text/plaintext')
+
+  expect(exported).toBe(`- ${ROOT_TOKEN}
+  - One.
+  - Two.
+  - Three.`)
+})
+
+it('split sentences with other thoughts nearby', () => {
```
raineorshine, line 23: "nearby" is too vague; you can say "siblings" to be precise.

src/reducers/splitSentences.ts
```diff
@@ -0,0 +1,39 @@
...
+import { contextOf, head, headRank, headValue, pathToContext, reducerFlow } from '../util'
+import { existingThoughtChange, newThought, setCursor } from '../reducers'
+import { ROOT_TOKEN } from '../constants'
+
+/** Split thought by sentences. Create new thought for each sentence. Thought value, on which cursor is on, replace with first sentence. */
+const splitSentences = (state: State) => {
+  const { cursor } = state
+  if (!cursor) return state
+  const thoughts = pathToContext(cursor)
+  const cursorContext = thoughts.length > 1 ? contextOf(thoughts) : [ROOT_TOKEN]
```
raineorshine, line 12: You can also use [rootedContextOf](https://github.com/cybersemics/em/blob/dev/src/util/rootedContextOf.ts).

raineorshine (changes requested):

src/shortcuts/splitSentences.tsx
pushkov-fedor: It wouldn't, because in this case we split by whitespace. Sentences with more then 1 word would be processed incorrectly.
E.g. `First sentence. Second sentence.` => `["First", "sentence.", "Second", "sentence."]`

pushkov-fedor:
@shresthabijay Bijay, hi! Could you help me please with one moment. It's about new `ContentEditable` component.
I use line of reducers in order to split thought by sentences and create new thought for each sentence except the first. The first sentence is used as new value for thought on which is cursor on. I use `existingThoughtChange` reducer for first sentence and then `newThought` for other sentences. But this leads to the cursor thought rendered value stays the same.

# Steps to Reproduce
1. Create thought "a. b. c.".
2. Wait for throttled `thoughtChangeHandler` exectuion and thought will be saved completely.
3. Invoke `splitSentences` shortcut (`meta + shift + s`).

# Current behavior
- a. b. c. 
- b.
- c.

# Expected behavior
- a.
- b.
- c.

As i change thought value by `existingThoughtChange`, `thoughtsRanked` in `Editable` component should be changed too. I've noticed that in `ContentEditable` component `allowInnerHTMLChange.current` value is false. As i understand, it prevents rendering new value. Could you give me advice on how to fix it?

src/shortcuts/splitSentences.tsx
raineorshine: Sorry, obvious typo. I meant:

```js
value.split('.').map(s => s.trim())
```

src/shortcuts/splitSentences.tsx
```diff
@@ -0,0 +1,34 @@
...
+const splitSentencesShortcut: Shortcut = {
+  id: 'splitSentences',
+  name: 'Split Sentences',
+  keyboard: { meta: true, shift: true, key: 's' },
+  svg: Icon,
+  canExecute: getState => getState().cursor !== null,
+  exec: (dispatch: Dispatch<Action | ActionCreator>, getState: () => State) => {
+    const { cursor } = getState()
+    const value = headValue(cursor!)
+    const sentences = value.split(/[.!?]+/g).filter(s => s !== '').map(s => `${s.trim()}.`)
```
raineorshine, line 24: You can use `match` here and count the sentences, which will be more efficient. Sorry if my previous comment gave you the wrong message.

src/reducers/splitSentences.ts
```diff
@@ -0,0 +1,38 @@
...
+/** Split thought by sentences. Create new thought for each sentence. Thought value, on which cursor is on, replace with first sentence. */
+const splitSentences = (state: State) => {
+  const { cursor } = state
+  if (!cursor) return state
+  const thoughts = pathToContext(cursor)
+  const cursorContext = rootedContextOf(thoughts)
+  const rank = headRank(cursor)
+  const value = headValue(cursor)
+
+  const splittedSentences = value.split(/[.!?]+/g).filter(s => s !== '').map(s => `${s.trim()}.`)
```
raineorshine, line 15: Please rename to `sentences`

src/shortcuts/splitSentences.tsx
shresthabijay: > @shresthabijay Bijay, hi! Could you help me please with one moment. It's about new `ContentEditable` component.
> I use line of reducers in order to split thought by sentences and create new thought for each sentence except the first. The first sentence is used as new value for thought on which is cursor on. I use `existingThoughtChange` reducer for first sentence and then `newThought` for other sentences. But this leads to the cursor thought rendered value stays the same.

@pushkov-fedor Hi! I am sorry for late reply. This seems to be a limitation of `ContentEditable`. It prevents `innerHTML` changes when it re-renders and only allows changes on when a node is `blurred` or in events of `paste` and thought split i.e on hitting `Enter`. So even though the first sentence `ContentEditable` is receiving new value, it is not allowing `innerHTML` change because of the constraints I explained above. We can use `shortcutEmitter` inside `ContentEditable` to allow `innerHTML` change. But it is not a elegant solution. @raineorshine I believe the original intention to create custom `ContentEditable` was to allow `Editable` to re-render without loosing focus offset. I don't see `Editable` re-rendering on edit. Can you confirm if we need this behavior ?

shresthabijay:
> @shresthabijay Bijay, hi! Could you help me please with one moment. It's about new `ContentEditable` component.
> I use line of reducers in order to split thought by sentences and create new thought for each sentence except the first. The first sentence is used as new value for thought on which is cursor on. I use `existingThoughtChange` reducer for first sentence and then `newThought` for other sentences. But this leads to the cursor thought rendered value stays the same.

@pushkov-fedor Hi! I am sorry for late reply. This seems to be a limitation of `ContentEditable`. It prevents `innerHTML` changes when it re-renders and only allows changes on when a node is `blurred` or in events of `paste` and thought split i.e on hitting `Enter`. So even though the first sentence `ContentEditable` is receiving new value, it is not allowing `innerHTML` change because of the constraints I explained above. We can use `shortcutEmitter` inside `ContentEditable` to allow `innerHTML` change. But it is not a elegant solution. @raineorshine I believe the original intention to create custom `ContentEditable` was to allow `Editable` to re-render without loosing focus offset. I don't see `Editable` re-rendering on edit. Can you confirm if we need this behavior ?

raineorshine:
> We can use `shortcutEmitter` inside `ContentEditable` to allow `innerHTML` change. But it is not a elegant solution.

Yeah, that's not a dependency we want to create.

> @raineorshine I believe the original intention to create custom `ContentEditable` was to allow `Editable` to re-render without loosing focus offset. I don't see `Editable` re-rendering on edit. Can you confirm if we need this behavior ?

We need whatever behavior is needed to fix the issue. Take another stab at figuring this out together before getting me involved... I believe in you! 😊

pushkov-fedor:
> > We can use `shortcutEmitter` inside `ContentEditable` to allow `innerHTML` change. But it is not a elegant solution.
> 
> Yeah, that's not a dependency we want to create.
> 
> > @raineorshine I believe the original intention to create custom `ContentEditable` was to allow `Editable` to re-render without loosing focus offset. I don't see `Editable` re-rendering on edit. Can you confirm if we need this behavior ?
> 
> We need whatever behavior is needed to fix the issue. Take another stab at figuring this out together before getting me involved... I believe in you!

@shresthabijay Bijay, hi! I am not expert of your part of code. What in your opinion is the best of approach to fix this issue? I think we need come up to some solution together.

shresthabijay:
> @shresthabijay Bijay, hi! I am not expert of your part of code. What in your opinion is the best of approach to fix this issue? I think we need come up to some solution together.

@pushkov-fedor Thanks for your patience. looked into this. I tried few different things but it always included creating unwanted dependencies with shortcuts. @raineorshine The problem I am facing with this is that we are not allowing content editable to update `innerHTML` on editing but we still want it to update on certain shortcut events. Either we have to find a way to save and restore caret position or introduce dependency with `shortcut`. I don't like the later one. Since with the current implementation the `Editable` doesn't re-render on edit. I think the quick and safe solution will be to fallback on using previous `ContentEditable` for now.

raineorshine:
The new `ContentEditable` is working everywhere else. If this is really the only place it fails, maybe you can look for a solution just to this problem (such as manually restoring the caret, or forcing a re-render) so that we can continue using the new `ContentEditable`.

shresthabijay:
> The new `ContentEditable` is working everywhere else. If this is really the only place it fails, maybe you can look for a solution just to this problem (such as manually restoring the caret, or forcing a re-render) so that we can continue using the new `ContentEditable`.

I think we could force re-render for this specific situation for now. However I would keep looking for better solution. I was really hoping to find some way to restoring caret that could work for wide range of scenarios. I wonder how Notion implements it.

shresthabijay:
@raineorshine I have added `editableNonce` that will be used to force update inner html of `ContentEditable`. This will take away need of adding shortcuts dependency to `ContentEditable`. Please review the changes. Thanks!

raineorshine:
> @raineorshine I have added `editableNonce` that will be used to force update inner html of `ContentEditable`. This will take away need of adding shortcuts dependency to `ContentEditable`. Please review the changes. Thanks!

Yes, this looks good. Hopefully we can remove it in the future, but at least this touches less parts of the codebase. Thanks @shresthabijay!

raineorshine:
@pushkov-fedor I think there are just a couple requested changes left.

pushkov-fedor:
@raineorshine  Looks like everything else is fixed. Could you check it please?

src/components/ThoughtAnnotation.tsx
```diff
@@ -49,7 +49,7 @@ const getSubThoughtTextMarkup = (state: State, isEditing: boolean, subthought: {
   const { editingValue } = state
   return {
     __html: isEditing
-      ? editingValue && subthought.text !== editingValue ? editingValue : subthought.text
+      ? editingValue && subthought.text === editingValue ? editingValue : subthought.text
```
raineorshine, line 52: `editingValue && subthought.text === editingValue ? editingValue : subthought.text ` is equivalent to `subthought.text`.

This code is intended to use the live value in the `ThoughtAnnotation` when it is being edited. What's purpose of this change? It seems like removing `editingValue` would eliminate existing functionality.

pushkov-fedor: Hmmm
Sometimes we have such render after `splitSentences`. In this case `editingValue` is `1.2.` and `subthought.text` is `1.` Maybe we should just set correct `editingValue` in the last step of `splitSentences` reducer.
Screenshot from 2020-10-07 12-42-37

pushkov-fedor:
@raineorshine Raine, this command could create duplicates. I think we should handle this in `canExecute` of `splitSentences` shortcut. What do you think?

src/components/ThoughtAnnotation.tsx
raineorshine: > Maybe we should just set correct `editingValue` in the last step of `splitSentences` reducer.

Yes, try that and see if it works.

`editingValue` should be [set on focus](https://github.com/cybersemics/em/blob/dev/src/components/Editable.tsx#L511) but maybe that's too late.

raineorshine:
> @raineorshine Raine, this command could create duplicates. I think we should handle this in `canExecute` of `splitSentences` shortcut. What do you think?

Good catch! Yes, that sounds right. Thanks for looking into that.

pushkov-fedor:
> > @raineorshine Raine, this command could create duplicates. I think we should handle this in `canExecute` of `splitSentences` shortcut. What do you think?
> 
> Good catch! Yes, that sounds right. Thanks for looking into that.

Done!

pushkov-fedor:
Actually we still have this `ContentEditable` behavior sometimes. :(
Screenshot from 2020-10-09 18-52-12

As seen on screenshot, superscript is positioned correctly, but first thought is rendered as multi-sentence thought.

pushkov-fedor:
I could catch the bug!

# Steps To Reproduce
1. Create thought `1.2.`.
2. Execute `splitSentences`
3. You have a tree like this:
- 1 .
- 2 .
4. Set cursor on first thought (`1.`).
5. Change the value from `1.` -> `1.3.`.
6. Execute `splitSentences` shortcut.

# Expected Behavior
- 1 .
- 3 .
- 2 .

# Current Behavior
- 1 . 3.
- 3 .
- 2 .
First thought is saved correctly. After refreshing the page it becomes `1.`.

raineorshine:
Okay, thanks! Will you have time to fix this, or should I give this task to someone else to finish?

pushkov-fedor:
> Okay, thanks! Will you have time to fix this, or should I give this task to someone else to finish?

Unfortunately, no, I don't have time. Sorry that I didn't finish this task completely.

shresthabijay:
@raineorshine  I made some changes to `Editable` and `ContentEditable`. It's working for me. Please test the changes.

raineorshine:
Works great! Thanks!

PR #860, opened 2020-09-04
Mobile incorrectly enters editing mode after drag

### LINKED ISSUE #821: Mobile incorrectly enters editing mode after drag
Tested in Safari, iPhone 11

## Steps to Reproduce

- a
  - b
  - c

1. Enter editing mode on `c`
2. Drag `c` above `b`
3. Close the keyboard 
4. Activate `cursorBack` (swipe right)

## Current Behavior

After Step 2, the caret disappears. The keyboard stays up on iOS and is closed on Android.

After Step 4, the keyboard comes up and [editing mode](https://github.com/cybersemics/em/wiki/Technical-Overview#9-mobile-editing-mode) is activated. 

## Expected Behavior

The caret should not disappear and the keyboard should not be closed automatically on drag.

Editing mode should stay disabled after closing the keyboard manually.

pushkov-fedor:
Fixes #821 

Raine, hi! Could you test please if it works on IOS now?


raineorshine (changes requested):
`editing` mode should not be disabled on drag. It should be disabled when the keyboard is closed.

pushkov-fedor:
Okay, let's clarify one moment. 
On IOS:
- a
  - b
  - c
1. Enter editing mode on c
2. Drag c above b
After dragging caret disappears, but keyboard is open. Do you have the same behavior?
On Android after dragging caret disappears and keyboard closes. For me on Android side it looks like after dragging editing is canceled. That's why i dispatch `editing` with value `false`. It leads to keyboard stays closed after `cursorBack`.

raineorshine:
> After dragging caret disappears, but keyboard is open. Do you have the same behavior?

Yes, I have the same behavior, and I didn't notice that the caret disappears at this step. The caret should be restored if the keyboard is still up.

> After dragging caret disappears and keyboard closes. For me on Android side it looks like after dragging editing is canceled. That's why i dispatch `editing` with value `false`. It leads to keyboard stays closed after `cursorBack`.

The keyboard should not be closed automatically. So it seems like there are two issues here:

1. The caret disappears on drag (which result in the keyboard being closed on Android)
2. After manually closing the keyboard, `editing` is still `true`. `editing` should always match the state of the keyboard.

pushkov-fedor:
Raine, hi! I'm a bit stuck on this issue.

What we have on Android:
Closing keyboard isn't invoke blur, caret stays. Tapping on left space near Thought lets us quit editing mode. But then after `cursorForward/cursorBack` shortcuts invocation, keyboard always pops up and editing mode turns on. Additional check in `Editable` component helps to solve this problem. Now after `cursorForward/cursorBack` execution, keyboard stays closed/open (based on editing mode).

Removing `clearSelection` call allows caret to stay on its place after drag-and-drop. On IOS this hack does nothing. It seems to me, that it's not the best approach to fix caret behavior.

For me nothing changes on IOS, but could you test it please on your device?

raineorshine:
> What we have on Android:
> Closing keyboard isn't invoke blur, caret stays.

Manually closing the keyboard should remove the caret. Usually the browser handles this.

> Tapping on left space near Thought lets us quit editing mode. But then after `cursorForward/cursorBack` shortcuts invocation, keyboard always pops up and editing mode turns on.

Correct (original issue)

> Additional check in `Editable` component helps to solve this problem. Now after `cursorForward/cursorBack` execution, keyboard stays closed/open (based on editing mode).

Yes, keyboard is staying closed and editing mode is staying disabled for me!

> Removing `clearSelection` call allows caret to stay on its place after drag-and-drop. On IOS this hack does nothing. It seems to me, that it's not the best approach to fix caret behavior.

Yeah, on my iOS the caret is disappearing after drag and drop, even though the keyboard is still up.  

You may want to look into the [logic in Editable.tsx that determines if `setSelectionToCursorOffset`](https://github.com/cybersemics/em/blob/dev/src/components/Editable.tsx#L300) is called. It's currently set to prevent the selection if `dragHold` is true, which may be part of the problem.

pushkov-fedor:
> > What we have on Android:
> > Closing keyboard isn't invoke blur, caret stays.
> 
> Manually closing the keyboard should remove the caret. Usually the browser handles this.

Keyboard closing on Android:
ezgif-6-01ac71ecb294

pushkov-fedor:
> You may want to look into the logic in Editable.tsx that determines if setSelectionToCursorOffset is called. It's currently set to prevent the selection if dragHold is true, which may be part of the problem.

Okay, thanks, i will take a look!

raineorshine:
> Keyboard closing on Android:

Is it normal on Android for the caret to remain when the keyboard is closed? It doesn't make much sense since the user is unable to type.

pushkov-fedor:
> > Keyboard closing on Android:
> 
> Is it normal on Android for the caret to remain when the keyboard is closed? It doesn't make much sense since the user is unable to type.

Yes. On my device editing works this way on any websites.

pushkov-fedor:
Raine, hi! The problem was in effect inside Editable component, which should call `setSelectionToCursorOffset`. This effect didn't run and nothing happened. 
I added another reducer to fire this effect after `endDrag` and also save `cursorOffset` in order to restore selection with correct offset after drag-and-drop. We can discuss this solution. Could you please check on you device that caret appears after drag-and-drop?
Also could you please re-confirm what keyboard behavior you have now?

raineorshine:
> The problem was in effect inside Editable component, which should call `setSelectionToCursorOffset`. This effect didn't run and nothing happened.

Thanks! Good catch.

> `endDrag` and also save `cursorOffset` in order to restore selection with correct offset after drag-and-drop. We can discuss this solution.

Sure, I'll take a look once the behavior is correct.

> Could you please check on you device that caret appears after drag-and-drop?
> Also could you please re-confirm what keyboard behavior you have now?

Yes, the caret appears after drag-and-drop. However, it also appears after drag-and-drop even when `editing` is disabled. `editing` should always match the state of the keyboard.

Unfortunately the original issue remains. It re-renters editing mode and the keyboard opens when it shouldn't.

pushkov-fedor:
> Yes, the caret appears after drag-and-drop. However, it also appears after drag-and-drop even when editing is disabled. editing should always match the state of the keyboard.

My bad! I added check on editing mode, now it should work correctly.

> Unfortunately the original issue remains. It re-renters editing mode and the keyboard opens when it shouldn't.

I'm afraid, in this case i cannot do anything further about this issue. It works on 6S, it works on Android (we can't blur by closing keyboard there, but can tap on the left space of thought. After tapping and invoking `cursor shortcut` keyboard doesn't pop up. It did before.)

raineorshine:
Great, I'll see if I can get it to work on iPhone 11. Thanks for all your work.

raineorshine:
I think I found a way to make it work without the additional `restoreSelection` reducer. It solves all the observed issues on iPhone 11. Will you test on Android?

There was code that called `clearSelection` on drag that was no longer needed. Then it was just a matter of setting `cursorOffset` in the `dragInProgress' reducer and adding `state.dragInProgress` to the `onEffect` dependencies. 

I did way more trial-and-error than I'd care to admit. That selection logic is nasty.

raineorshine (approved):

pushkov-fedor:
> I think I found a way to make it work without the additional restoreSelection reducer. It solves all the observed issues on iPhone 11. Will you test on Android?

Yes! It works! If editing mode is on, caret restores after drag-and-drop. 
After closing keyboard and `cursor` shortcuts invocation, keyboard doesn't pop up.

raineorshine:
> Yes! It works! If editing mode is on, caret restores after drag-and-drop.
> After closing keyboard and `cursor` shortcuts invocation, keyboard doesn't pop up.

Thank god!

PR #861, opened 2020-09-04
Mobile: Moving cursor from note to thought incorrectly enters edit mode

### LINKED ISSUE #858: Mobile: Moving cursor from note to thought incorrectly enters edit mode
While this looks like a regression, it was actually an underlying issue that didn't become visible until #856 was fixed.

## Steps to Reproduce

- a
[note]
  - b
  - c

1. Tap on note (enters edit mode)
2. Close keyboard
3. Tap on `b`

## Current Behavior

Keyboard opens. It think edit mode is still set from the Note's `onFocus`. Somehow edit mode needs to be disabled when the keyboard closes. However if you disable edit mode in `onBlur`, it might break cursor movement when the keyboard is kept up.

Worst case, we set a timer and only disable edit mode if the selection is really removed (i.e. keyboard is down).

## Expected Behavior

Keyboard should stay closed and edit mode should be false.

pushkov-fedor:
Fixes #858

### DIFF
--- a/src/components/Note.tsx
+++ b/src/components/Note.tsx
@@ -81,6 +81,13 @@ const Note = ({ context, thoughtsRanked, contextChain }: { context: Context, tho
     dispatch({ type: 'setCursor', thoughtsRanked, contextChain, cursorHistoryClear: true, editing: true, noteFocus: true })
   }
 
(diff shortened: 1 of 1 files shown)


raineorshine:
Works! Thanks

PR #876, opened 2020-09-30
Sorting should work in the root

### LINKED ISSUE #863: Sorting should work in the root
When `cursor` is `null`, toggling sort should sort the root. Currently it does nothing.

aqkhan:
Fixes #863

### DIFF
--- a/src/components/Subthoughts.tsx
+++ b/src/components/Subthoughts.tsx
@@ -364,8 +364,7 @@ export const SubthoughtsComponent = ({
   const state = store.getState()
   const [page, setPage] = useState(1)
   const globalSort = getSetting(state, ['Global Sort']) || 'None'
-  const rootSort = isRoot(thoughtsRanked) ? attribute(store.getState(), thoughtsRanked!, '=sort') : null
-  const sortPreference = (contextSort || rootSort) || globalSort
+  const sortPreference = contextSort || globalSort
   const { cursor } = state
   // TODO: This getThought call looking bit ambitious to me I am commenting the previous statement please check this.
   const thought = getThought(state, headValue(thoughtsRanked))
--- a/src/components/Subthoughts.tsx
+++ b/src/components/Subthoughts.tsx
@@ -364,7 +364,8 @@ export const SubthoughtsComponent = ({
   const state = store.getState()
   const [page, setPage] = useState(1)
   const globalSort = getSetting(state, ['Global Sort']) || 'None'
-  const sortPreference = contextSort || globalSort
+  const rootSort = isRoot(thoughtsRanked) ? attribute(store.getState(), thoughtsRanked!, '=sort') : null
+  const sortPreference = (contextSort || rootSort) || globalSort
   const { cursor } = state
   // TODO: This getThought call looking bit ambitious to me I am commenting the previous statement please check this.
   const thought = getThought(state, headValue(thoughtsRanked))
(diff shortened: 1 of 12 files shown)


aqkhan:
@raineorshine The sorting has been enabled in the ROOT context but the application does not render automatically when the sorted is enabled. Your input is needed on this.

src/reducers/toggleAttribute.ts
```diff
@@ -32,7 +32,7 @@ const toggleAttribute = (state: State, { context, key, value }: { context: Conte
 
       // set attribute value
       setFirstSubthought({
-        context: context.concat(key),
+        context: isRoot(context) ? [key] : context.concat(key),
```
raineorshine, line 35: Let's use `[...unroot(context), key]`

src/selectors/attribute.ts
```diff
@@ -1,13 +1,13 @@
...
-import { pathToContext } from '../util'
+import { isRoot, pathToContext } from '../util'
 import { getThoughts, isChildVisible } from '../selectors'
 import { State } from '../util/initialState'
 
 /** Returns the value of an attribute of the given context. O(children). Use attributeEquals for O(1) attribute equality check. Use a combination of hasChild and attribute to check if an attribute exists but has no children. */
 const attribute = (state: State, pathOrContext: Path|Context, attributeName: string) => {
   const context = pathToContext(pathOrContext)
-  const children = getThoughts(state, [...context, attributeName])
+  const children = getThoughts(state, isRoot(context) ? [attributeName] : [...context, attributeName])
```
raineorshine, line 10: Prefer `[...unroot(context), attributeName]`

src/selectors/attributeEquals.ts
```diff
@@ -12,7 +13,7 @@ import {
 
 /** Returns true if the given attribute equals the given value. O(1). Use over attribute when possible for performance. */
 const attributeEquals = (state: any, context: Context, attr: string, value: string) => {
-  const children = getThoughts(state, context.concat(attr))
+  const children = getThoughts(state, isRoot(context) ? [attr] : context.concat(attr))
```
raineorshine, line 16: Prefer `[...unroot(context), attr]`

src/components/Subthoughts.tsx
```diff
@@ -364,7 +364,8 @@ export const SubthoughtsComponent = ({
   const state = store.getState()
   const [page, setPage] = useState(1)
   const globalSort = getSetting(state, ['Global Sort']) || 'None'
-  const sortPreference = contextSort || globalSort
+  const rootSort = isRoot(thoughtsRanked) ? attribute(store.getState(), thoughtsRanked!, '=sort') : null
+  const sortPreference = (contextSort || rootSort) || globalSort
```
shresthabijay: Since `Subthoughts` component gets `sort` as a prop, it won't re-render if we put the `rootSort` logic into `Subthoughts` render function. `Content` is the actually the parent component of the root `Subthoughts`, so we need to place that logic inside `Content` to make it work.

src/shortcuts/toggleSort.tsx
```diff
@@ -42,8 +43,7 @@ const toggleSortShortcut: Shortcut = {
     const sortPreference = globalSort === 'Alphabetical' ? 'None' : 'Alphabetical'
 
     if (cursor) {
```
raineorshine, line 45: Find a way to do this with a single `dispatch` call (i.e. with a ternary statement inside the dispatch argument). The logic will be clearer if it's lower down since the two paths have so much in common.

raineorshine (changes requested):

shresthabijay:
@aqkhan @raineorshine. I have pushed some changes and the PR is now working. I have included following changes as well.

- Attribute togglers that targets parent of cursor to use  `[ROOT_TOKEN]` as context if the cursor is `null`.

raineorshine:
@aqkhan @shresthabijay Great! It works for me. Can one of you correct the small changes requested in the review?

PR #878, opened 2020-10-06
Circular path fix for chain

shresthabijay:
# Changes

- Pass pivot index to the `chain` to create `thoughtsResolved` for circular paths.
- Add generic return type to `ActionCreator` interface.
- Tests for circular paths context view.


src/components/Editable.tsx
```diff
@@ -110,6 +110,7 @@ interface EditableProps {
   rank: number,
   showContexts?: boolean,
   style?: GenericObject<string>,
+  pivotIndex: Nullable<number>,
```
raineorshine, line 113: Why not `number | null`? It's shorter and doesn't require an import.

src/selectors/__tests__/expandThoughts.ts
```diff
@@ -62,7 +62,7 @@ describe('normal view', () => {
   - c
     - d`
 
-    const thoughts = await importText(RANKED_ROOT, text)(NOOP, initialState)
+    const thoughts = await importText(RANKED_ROOT, text, {})(NOOP, initialState)
```
raineorshine, line 65: The options object should be optional.

src/action-creators/importText.ts
```diff
@@ -3,6 +3,7 @@ import { parse } from 'jex-block-parser'
 import he from 'he'
 import { contextOf, convertHTMLtoJSON, head, importJSON, pathToContext, rootedContextOf, strip } from '../util'
 import { ActionCreator, Path } from '../types'
+import { Payload } from '../reducers/updateThoughts'
```
raineorshine, line 6: This would be clearer as `Payload as UpdateThoughtsPayload` since `Payload` is so generic.

src/components/__tests__/Subthoughts.js
```diff
@@ -168,3 +169,48 @@ describe('context view', () => {
...
+  store.dispatch({ type: 'setCursor', thoughtsRanked: [
+    { value: 'a', rank: 0 },
+    { value: 'b', rank: 0 },
+    { value: 'c', rank: 0 },
+    { value: 'd', rank: 0 },
+    { value: 'c', rank: 0 },
+  ]
+  })
+
+  store.dispatch({ type: 'toggleContextView' })
```
raineorshine, line 192: We have middleware that supports passing an array of actions to `store.dispatch`.

raineorshine (changes requested):

raineorshine:
Thanks, this is a great start. The tests look good, although I think we probably should have a separate test for circular paths (`a/b/c~/b/c`) vs paths with duplicate lexemes ('a/b/c/d/c`) vs circular paths with duplicate lexemes ('a/b/c/d/c~/d').

Notice how `pivotIndex` is now a dependency for all these components:

- Content.tsx
- Editable.tsx
- Note.tsx
- SearchSubthoughts.tsx
- Subthoughts.tsx
- Thought.tsx
- ThoughtAnnotation.tsx

Think of what has happened here: you have an implementation detail that is now a dependency for seven different components that should not care about how the `contextChain` is generated. I really can't emphasize this point enough; it's a critical architectural decision and an important skill that I've been trying to teach you. The programmer's main job is to manage complexity. You need to think more about encapsulation and dependencies.

Here are the questions I would ask as we consider how to re-architect this:

- Can `pivotIndex` be generated internally so that components do not have to be aware of it?
- If not, what is the minimum surface area required? Is there any way to isolate this behavior to avoid cutting across multiple interfaces?

shresthabijay:
> Think of what has happened here: you have an implementation detail that is now a dependency for seven different components that should not care about how the `contextChain` is generated. I really can't emphasize this point enough; it's a critical architectural decision and an important skill that I've been trying to teach you. The programmer's main job is to manage complexity. You need to think more about encapsulation and dependencies.
> 
> Here are the questions I would ask as we consider how to re-architect this:
> 
> * Can `pivotIndex` be generated internally so that components do not have to be aware of it?
> * If not, what is the minimum surface area required? Is there any way to isolate this behavior to avoid cutting across multiple interfaces?

I was about to discuss this. Thanks for pointing it out. The `pivotIndex` can be calculated from `Child` of the `Subthoughts` with `contextView` toggled on. So we cannot calculate it internally in the places where `chain` is used to generate `thoughtsResolved`. Since `pivotIndex` is actually the part of the `contextChain`, we may need to change the structure of the `contextChain` to something like this.

```javascript
   const contextChain = [
    {
      chain: ['a', 'b', 'c', 'd', 'c'],
      pivotIndex: 4
    }
   ]
```

raineorshine:
> I was about to discuss this. Thanks for pointing it out. The `pivotIndex` can be calculated from `Child` of the `Subthoughts` with `contextView` toggled on. So we cannot calculate it internally in the places where `chain` is used to generate `thoughtsResolved`.

Great, I'm really glad to know it can be done internally.

> Since `pivotIndex` is actually the part of the `contextChain`, we may need to change the structure of the `contextChain` to something like this.
> 
> ```js
>    const contextChain = [
>     {
>       chain: ['a', 'b', 'c', 'd', 'c'],
>       pivotIndex: 4
>     }
>    ]
> ```

I think `pivotIndex` is a property of a `Path` not a `ContextChain`. Since the `contextChain` consists of unbroken `Paths` in between context views, there is no need for a pivot to split them further. It's when we have a `Path` which consists of undifferentiated `Path` segments that we need to calculate the `pivotIndex`.

That said, I'd like to ask about another possibility. I notice that the signature of `chain` is:

```js
const chain = (state: State, contextChain: Path[], path: Path) => {...}
```

What if we could pass a `thoughtsRanked` instead of `path`? Then there would be no need to calculate a pivot. `chain` is only used in five places, all of which have `thoughtsRanked` in scope:

- Editable
- Subthoughts
- Thought
- ThoughtAnnotation
- setCursor

Could this work?

shresthabijay:
> That said, I'd like to ask about another possibility. I notice that the signature of `chain` is:
> 
> ```js
> const chain = (state: State, contextChain: Path[], path: Path) => {...}
> ```
> 
> What if we could pass a `thoughtsRanked` instead of `path`? Then there would be no need to calculate a pivot. `chain` is only used in five places, all of which have `thoughtsRanked` in scope:
> 
> * Editable
> * Subthoughts
> * Thought
> * ThoughtAnnotation
> * setCursor
> 
> Could this work?

I don't think it will work. In fact we are passing `thoughtsRanked` as path to `chain` in all these places already. The `pivotIndex` is actually the length of the direct `child` of a `context` view. 

```
- a
  - b ~
     - a.b (pivotIndex is 1)
        - k 
     - c.d.b (pivotIndex is 2)
        - e 
          - f ~
            - c.d.b.e.f (pivotIndex is 4)
            - j.m.i.f (pivotIndex is 3)
     - a.b.c.d.b (pivotIndex is 4)
```

As you can see `pivotIndex` needs to be passed down from direct `child` of the contextView to its descendants. It will reset again after a nested contextView is encoutered. With just `thoughtsRanked` we can't know the index of the pivot value for a circular path.

raineorshine:
Let's use the term `SimplePath` to refer to a contiguous `Path` with no cycles or context view jumps. `thoughtsRanked` is a `SimplePath` and `cursor` is a `Path`.

> With just `thoughtsRanked` we can't know the index of the pivot value for a circular path.

I believe you're talking about a `SimplePath` with duplicate lexemes here rather than a `Path` with cycles, e.g. `a/b/c/d/c`, but correct me if I'm wrong. We should take care to distinguish these two types of paths (and their combination), as described here:

> [Raine] circular paths (`a/b/c~/b/c`) vs paths with duplicate lexemes ('a/b/c/d/c`) vs circular paths with duplicate lexemes ('a/b/c/d/c~/d').

Next...

> The `pivotIndex` is actually the length of the direct `child` of a `context` view.

Yes, because we only ever call `chain` to append a `SimplePath` to the end of a `Path`. e.g. `a/b/c/d/c` will always have `pivotIndex` of 4 (the last `c`) not 2 (the first `c`). Is this true? If so, is the `pivotIndex` needed at all?

I probably had a reason for doing the `find` rather than just taking from the end, but I can't recall what it was. This seems like a critical point though to be able to justify the use of a `pivotIndex`.

> `pivotIndex` will reset again after a nested contextView is encoutered. With just `thoughtsRanked` we can't know the index of the pivot value for a circular path.

This is merely an addendum, but it might add clarity to our conversation. `pivotIndex` doesn't get "reset", as its only used for appending a single `SimplePath` to the current `contextChain`. Your language suggests that `pivotIndex` persists across levels, but I don't think that's appropriate since it is only in scope at a single level. Ancestor pivots are not needed to calculate descendant pivots. Once the `contextChain` is built up, it is no longer needed. It is an ephemeral value used in a single `chain` call. `contextChain` is where the information is persisted. Of course, I'm questioning the existence of a `pivotIndex` that is `find`-based (current implementation) or parameterized (passed down).

Lastly...

I think the `chain` test for a longer `contextChain` is incorrect; it should pass a `SimplePath` rather than a `Path`. Here's the current test I think is wrong:

```js
  expect(pathToContext(chain(
    initialState(),
    [
      rankThoughtsSequential(['2', 'A']),
      rankThoughtsSequential(['1', 'A', 'Nope']),
    ],
    rankThoughtsSequential(['START', 'B', 'Nope', 'Butter', 'Bread']) // WRONG
  ))).toEqual(['2', 'A', '1', 'Nope', 'B', 'Butter', 'Bread'])
```

The following tests the construction of a `Path` with no duplicate lexemes: `a/b/m~/y~/z`. Is this correct?

```js
  expect(pathToContext(chain(
    initialState(),
    [
      rankThoughtsSequential(['a', 'b', 'm']),
      rankThoughtsSequential(['c', 'd', 'm', 'y']),
    ],
    rankThoughtsSequential(['c', 'd', 'm', 'y', 'z'])
  ))).toEqual(['a', 'b', 'm', 'y', 'z'])
```

shresthabijay:
> Let's use the term `SimplePath` to refer to a contiguous `Path` with no cycles or context view jumps. `thoughtsRanked` is a `SimplePath` and `cursor` is a `Path`.

That's much clear to understand.

shresthabijay:
> I believe you're talking about a `SimplePath` with duplicate lexemes here rather than a `Path` with cycles, e.g. `a/b/c/d/c`, but correct me if I'm wrong. We should take care to distinguish these two types of paths (and their combination), as described here:

Yes exactly. `SimplePath` with duplicate lexemes.

shresthabijay:
> Yes, because we only ever call `chain` to append a `SimplePath` to the end of a `Path`. e.g. `a/b/c/d/c` will always have `pivotIndex` of 4 (the last `c`) not 2 (the first `c`). Is this true? If so, is the `pivotIndex` needed at all?

Use of `find` in the `chain` made me think that pivot value couldn't be simply be taken as the last element. If there is no any specific reason for using find, then it could be so much simpler to just take the last value.

shresthabijay:
> This is merely an addendum, but it might add clarity to our conversation. `pivotIndex` doesn't get "reset", as its only used for appending a single `SimplePath` to the current `contextChain`. Your language suggests that `pivotIndex` persists across levels, but I don't think that's appropriate since it is only in scope at a single level. Ancestor pivots are not needed to calculate descendant pivots. Once the `contextChain` is built up, it is no longer needed. It is an ephemeral value used in a single `chain` call. `contextChain` is where the information is persisted. Of course, I'm questioning the existence of a `pivotIndex` that is `find`-based (current implementation) or parameterized (passed down).

You are correct. I got it wrong. It doesn't persist across levels.

shresthabijay:
> Yes, because we only ever call `chain` to append a `SimplePath` to the end of a `Path`. e.g. `a/b/c/d/c` will always have `pivotIndex` of 4 (the last `c`) not 2 (the first `c`). Is this true? If so, is the `pivotIndex` needed at all?

`chain` is called to generate `thoughtsResolved` for every `Subthoughts` component where `contextChain.length > 1`

https://github.com/cybersemics/em/blob/2678137e03364e88b846d80096903279be79bc11/src/components/Subthoughts.tsx#L376-L378

`
- a
  - b
    - c ~ 
      - a.b.c
      - a.b.c.d.c
`

shresthabijay:
> The following tests the construction of a `Path` with no duplicate lexemes: `a/b/m~/y~/z`. Is this correct?
> 
> ```js
>   expect(pathToContext(chain(
>     initialState(),
>     [
>       rankThoughtsSequential(['a', 'b', 'm']),
>       rankThoughtsSequential(['c', 'd', 'm', 'y']),
>     ],
>     rankThoughtsSequential(['c', 'd', 'm', 'y', 'z'])
>   ))).toEqual(['a', 'b', 'm', 'y', 'z'])
> ```

This will fail. This will generate `[ 'a', 'b', 'm', 'c', 'm', 'y', 'm', 'z']`. It's another issue with `chain`. The current implementation simply removes value from the first index of the `simplePath` in the `contextChain`.

shresthabijay:
> Your language suggests that `pivotIndex` persists across levels, but I don't think that's appropriate since it is only in scope at a single level. Ancestor pivots are not needed to calculate descendant pivots. Once the `contextChain` is built up, it is no longer needed.

I hope the example above shows why there is a dependency between ancestor and the descendant for creating pivots.

raineorshine:
Thank you for the thorough explanation!

> The inputs for `chain` to calculate `thoughtsResolved` for both **first** and **second** are identical. We cannot simply use `2` as the pivot index for both of them. This will cause `thoughtsResolved` to be same for both **first** and **second** `f`.

This really clarifies it for me. Indeed `chain` does not have enough information with just `contextChain` and `simplePath`, because they are missing the `Path` segment since the last context view.

> `chain` needs extra dependency when there are duplicate lexemes in provided `simplePath`. The dependency is the length of the `simplePath` of the nearest ancestor that is a direct child of a context view. I tried solving it passing `pivotIndex` to the children but introduces dependencies to many components.

Yes, this is spreading highly implementation-specific dependencies into many components. We need to find another way.

> This will fail. This will generate `[ 'a', 'b', 'm', 'c', 'm', 'y', 'm', 'z']`. It's another issue with `chain`. The current implementation simply removes value from the first index of the `simplePath` in the `contextChain`.

We both know the current implementation will fail. I was asking if the test was correct.

> > Your language suggests that `pivotIndex` persists across levels, but I don't think that's appropriate since it is only in scope at a single level. Ancestor pivots are not needed to calculate descendant pivots. Once the `contextChain` is built up, it is no longer needed.
> 
> I hope the example above shows why there is a dependency between ancestor and the descendant for creating pivots.

You're right, thank you!

raineorshine:
If we passed `Path` instead of `SimplePath` down the component hierarchy, then could that be directly appended to instead of having to use `contextChain`?

shresthabijay:
> If we passed `Path` instead of `SimplePath` down the component hierarchy, then could that be directly appended to instead of having to use `contextChain`?

`thoughtsResolved` is a unique path to a thought in a tree. We can build up the `thoughtsResolved` for a child by appending to the parent's `thoughtsResolved`. We may not need to depend on `contextChain`. It makes sense for a thought to know about it's parent's path because it's own path can be derived from it.

raineorshine:
> > If we passed `Path` instead of `SimplePath` down the component hierarchy, then could that be directly appended to instead of having to use `contextChain`?
> 
> I understand `thoughtsResolved` as a unique path to a thought in the tree. We can build up the `thoughtsResolved` for a child by appending to it's parent's `thoughtsResolved`. We may not need to depend on `contextChain`. It makes sense for a thought to know about it's parent's path because it's own path can be derived from it. This will eliminate the need of `chain` and problem of finding correct `pivot` when there are duplicate lexemes. Each thought will directly depend only on its immediate parent for calculating `thoughtsResolved`.

Makes sense to me! I wish I could remember why I implemented it as a `contextChain` to begin with! Could you review the uses of `contextChain` just to make sure that the n-1 `SimplePath` segments are not being used? If so, I think this would be a great avenue to pursue.

shresthabijay:
> Makes sense to me! I wish I could remember why I implemented it as a `contextChain` to begin with! Could you review the uses of `contextChain` just to make sure that the n-1 `SimplePath` segments are not being used? If so, I think this would be a great avenue to pursue.

Sure. I will do that.

shresthabijay:
https://github.com/cybersemics/em/blob/f5a7bdaf5a154d6aa2e14294b8c6e652c233338c/src/reducers/setCursor.ts#L78-L84

I couldn't understand use of `contextChain` here to calculate `expanded`. Do you think we can calculate expanded here with just `thoughtsResolved` ?

shresthabijay:
I have reviewed uses of `contextChain`. It is used extensively in many places. Most places generate `contextChain` from given `path` using `splitChain`. However we can remove use of `contextChain` in places like `Subthoughts`, `Thoughts`, `Editable` etc where it is used just to generate `thoughtsResolved`. We can calculate the `thoughtsResolved` easily as we discussed above.

We also don't need to pass `contextChain` to the `setCursor` reducer, as it only uses it to generate `thoughtsResolved` at the end. The only place `contextChain` is passed to the `setCursor` is at `cursorDown`.

https://github.com/cybersemics/em/blob/c49c64baa619f00be647a69e2b6acd45e271c8d9/src/reducers/cursorDown.ts#L13-L23

shresthabijay:
@raineorshine Can't we always replace head with cursor here ? https://github.com/cybersemics/em/blob/c49c64baa619f00be647a69e2b6acd45e271c8d9/src/components/Subthoughts.tsx#L130-L132

raineorshine:
> I have reviewed uses of `contextChain`. It is used extensively in many places. Most places generate `contextChain` from given `path` using `splitChain`. However we can remove use of `contextChain` in places like `Subthoughts`, `Thoughts`, `Editable` etc where it is used just to generate `thoughtsResolved`. We can calculate the `thoughtsResolved` easily as we discussed above.

That's great. I would suggest removing `contextChain` one-by-one from these components and running the tests (or at least the tests for that component) each time.

> @raineorshine Can you explain the need of checking `contextChain` length here ?
> 
> https://github.com/cybersemics/em/blob/c49c64baa619f00be647a69e2b6acd45e271c8d9/src/components/Subthoughts.tsx#L130-L132

It's used to generate a `SimplePath` that reflects that most recent edit when the `Subthoughts` component is part of a context view. `props.simplePath` is the `SimplePath` that the containing `Thought` component passed as a prop, which is not updated in real-time in order to prevent excessive re-rendering. Perhaps the rendering logic has changed such that we don't need that separate any longer (i.e. your work on ContentEditable), but that would require a refactor and good testing to confirm. It was needed at one point.

I believe we can just check `showContextsParent` rather than use `contextChain` here. I made the change in 876f74f8a8c58bd84a7205b6ac2b5e55fa4d57d5 and the tests are passing.

PR #879, opened 2020-10-06
Could not find a declaration file for module 'himalaya'

### LINKED ISSUE #877: docs: Could not find a declaration file for module 'himalaya'.
For some reason `himalaya` is causing an error in `typedocs`. It is not seeing the types declared in the `himalaya` module. There is no `@types/himalaya` package. I'm not sure why it's failing in `docs` but not in `dev`. 

I tried to find a solution but no luck so far. Sorry to assign another task to you, but could you take a look?

```
$ npm run docs

> em@162.0.0 docs /Users/raine/projects/em
> typedoc --options typedoc.json

typedoc-plugin-external-module-name: Using custom module name mapping function from /Users/raine/projects/em/.typedoc-plugin-external-module-name.js
Loaded plugin typedoc-plugin-external-module-name
Loaded plugin typedoc-plugin-exclude-references
New __namedParameters name: options
Loaded plugin typedoc-plugin-rename-named-parameters
Loaded plugin typedoc-plugin-module-wide-comments

Using TypeScript 3.9.7 from /Users/raine/projects/em/node_modules/typescript/lib
Error: /Users/raine/projects/em/src/util/convertHTMLtoJSON.ts(1)
 Could not find a declaration file for module 'himalaya'. '/Users/raine/projects/em/node_modules/himalaya/lib/index.js' implicitly has an 'any' type.
  Try `npm install @types/himalaya` if it exists or add a new declaration (.d.ts) file containing `declare module 'himalaya';`
```

pushkov-fedor:
Fixes #877

### DIFF
--- a/typedoc.json
+++ b/typedoc.json
@@ -1,6 +1,6 @@
 {
   "inputFiles": ["src/*"],
-  "exclude": ["src/*/*/*", "src/registerServiceWorker.js", "src/index.js"],
+  "exclude": ["src/!(@types)/*/*", "src/registerServiceWorker.js", "src/index.js"],
   "excludeExternals": true,
   "mode": "modules",
   "out": "docs",


raineorshine:
Interesting! I forgot we had a module declaration for `himalaya`. Glad I asked!

Could you explain the purpose of excluding `"src/!(@types)/*/*"` (and the original `"src/*/*/*"` for that matter)? I'd like to confirm it's the appropriate level of specificity since it seems like it was too general before.

pushkov-fedor:
> Interesting! I forgot we had a module declaration for `himalaya`. Glad I asked!
> 
> Could you explain the purpose of excluding `"src/!(@types)/*/*"` (and the original `"src/*/*/*"` for that matter)? I'd like to confirm it's the appropriate level of specificity since it seems like it was too general before.

It is primarily used to exclude `__test__` folders. It also excludes `Tutorial` and `icons` folders inside `components`

raineorshine:
I see. That *might* be too general, but I'm okay with it for now. Now I know what the problem is if it's not picking up a type. Thanks!

PR #882, opened 2020-10-11
cursorNext/cursorPrev: Fix movement in sorted context

### LINKED ISSUE #880: cursorNext/cursorPrev: Fix movement in sorted context
`cursorNext` and `cursorPrev` should work on sorted contexts the same way they works on unsorted contexts. Currently they are skipping around the list because it using the order provided by [`getThoughtsRanked`](https://github.com/cybersemics/em/blob/dev/src/selectors/getThoughtsRanked.ts) rather than [`getChildrenSorted`](https://github.com/cybersemics/em/blob/dev/src/selectors/getChildren.ts#L41-L46) which is sorted.

Tests are written and just need to be uncommented:
- https://github.com/cybersemics/em/blob/dev/src/action-creators/__tests__/cursorNext.js#L75-L102
- https://github.com/cybersemics/em/blob/dev/src/action-creators/__tests__/cursorPrev.js#L75-L102

Followup from https://github.com/cybersemics/em/pull/831#issuecomment-675176412

aqkhan:
Fixes #880


raineorshine:
Looks good! Please amend the commit to remove the yarn.lock, and then we're good to go!

PR #885, opened 2020-10-14
Fix multiple thoughts import overwrites siblings.

### LINKED ISSUE #832: Importing multiple thoughts overwrites siblings
## Steps to Reproduce

- a
  - b [cursor]

Paste two or more thoughts at once:

```
- x
- y
```

## Current Behavior

- a
  - x
  - y

(`b` is still in `thoughtIndex`)

## Expected Behavior

- a
  - b
  - x
  - y

## Tests

Write a failing unit test (recommend doing this first for TDD). Test that both `contextIndex` and `thoughtIndex` are correct after pasting.

**This is a rough draft of a unit test for `importHtml.js` but it is failing differently than I expected. You will have to fix this or write your own.**

```js
it('paste multiple thoughts after the cursor', () => {

  /** Import HTML and merge into state. */
  const importHtmlReducer = (state, insertionPath, html) => {
    const { contextIndexUpdates, thoughtIndexUpdates } = importHtml(state, insertionPath, html)
    return {
      ...state,
      thoughts: {
        contextIndex: contextIndexUpdates,
        thoughtIndex: thoughtIndexUpdates,
      }
    }
  }

  const initialHtml = `
<li>a<ul>
  <li>b</li>
</ul></li>
`

  const importedHtml = `
<li>x</li>
<li>y</li>
`

  const state1 = importHtmlReducer(initialState, RANKED_ROOT, initialHtml)
  const state2 = importHtmlReducer(state1, [{ value: 'a', rank: 0 }, { value: 'b', rank: 1 }], importedHtml)

  const exported = exportContext(state2, [ROOT_TOKEN], 'text/plaintext')

  expect(exported).toBe(`- __ROOT__
  - a
    - b
    - x
    - y`)
})

```

shresthabijay:
fixes #832

### DIFF
--- a/src/util/importJSON.ts
+++ b/src/util/importJSON.ts
@@ -132,9 +132,9 @@ export const importJSON = (state: State, thoughtsRanked: Path, blocks: Block[], 
     // update contextIndexUpdates
     const contextEncoded = hashContext(rootContext)
 
-    const childrenUpdates = contextIndexUpdates[contextEncoded]
-      ? contextIndexUpdates[contextEncoded].children
-      : [...contextIndex[contextEncoded] ? contextIndex[contextEncoded].children : []] // concat children from already available contextIndex
+    const childrenUpdates =
+      contextIndexUpdates[contextEncoded]?.children ||
+      contextIndex[contextEncoded]?.children || []
 
     contextIndexUpdates[contextEncoded] = {
       ...contextIndexUpdates[contextEncoded],
--- a/src/util/importJSON.ts
+++ b/src/util/importJSON.ts
@@ -88,6 +88,7 @@ export const importJSON = (state: State, thoughtsRanked: Path, blocks: Block[], 
   const destThought = head(thoughtsRanked)
   const destEmpty = destThought.value === '' && getThoughts(state, pathToContext(thoughtsRanked)).length === 0
   const thoughtIndex = { ...state.thoughts.thoughtIndex }
+  const contextIndex = { ...state.thoughts.contextIndex }
   const rankStart = getRankAfter(state, thoughtsRanked)
   const rankIncrement = getRankIncrement(state, blocks, context, destThought, rankStart)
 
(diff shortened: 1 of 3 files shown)


shresthabijay:
@raineorshine There was issue with the test you provided earlier. Thoughts updates needed to be merged with the previous thoughts. And the reason for overwriting sibling was that `importJSON` calculated `contextIndexUpdates`  without adding `children` from already available `contextIndex`  from the state. I have made necessary changes and it's working. Please review it. Thanks!

 However I noticed that the tree is not re-rendering after import even in `dev`. Let me know if that is intended.

raineorshine:
> @raineorshine There was a issue with the test you provided earlier.

Yes, did you see the comment where pointed it out in the original issue? Anyway, thanks for fixing it!

> And the reason for overwriting siblings was that `importJSON` calculated `contextIndexUpdates` without adding `children` from already available `contextIndex` from the state. I have made necessary changes and it's working. Please review it. Thanks!

Excellent, thanks!

> However I noticed that the tree is not re-rendering after import even in `dev`. Let me know if that is intended.

No, we definitely want it to re-render after import.

src/util/importJSON.ts
```diff
@@ -130,7 +131,11 @@ export const importJSON = (state: State, thoughtsRanked: Path, blocks: Block[],
 
     // update contextIndexUpdates
     const contextEncoded = hashContext(rootContext)
-    const childrenUpdates = contextIndexUpdates[contextEncoded] ? contextIndexUpdates[contextEncoded].children : []
+
+    const childrenUpdates =
+      contextIndexUpdates[contextEncoded]?.children ||
+      contextIndex[contextEncoded]?.children || []
```
raineorshine, line 137: @shresthabijay We can clean this up a bit with the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

raineorshine (approved):

PR #888, opened 2020-10-21
Basic importRoam

### LINKED ISSUE #865: Import from Roam Research
## Part I
- [x] Import thoughts from Roam's JSON format (see below)
- [x] Convert to `Blocks` JSON format then use `importJSON`

## Part II
- [x] Save `create-time` as `created` property
- [x] Save `edit-time` as `lastUpdated` property
- [x] Save `create-email` as `=create-email` subthought
- [x] Save `edit-email` as `=edit-email` subthought

## Part III
- [ ] Autodetect Roam JSON in `importText`

### JSON:

```js
[
{
    "title": "September 4th, 2020",
    "children": [
    {
        "string": "A",
        "create-email": "raine@cybersemics.org",
        "create-time": 1600111381583,
        "children": [
        {
            "string": "B",
            "create-email": "raine@cybersemics.org",
            "create-time": 1600111383054,
            "children": [
            {
                "string": "C",
                "create-email": "raine@cybersemics.org",
                "create-time": 1600111383911,
                "uid": "HMN_YQtZZ",
                "edit-time": 1600111385719,
                "edit-email": "raine@cybersemics.org"
            }],
            "uid": "JBXKlMcxh",
            "edit-time": 1600111383918,
            "edit-email": "raine@cybersemics.org"
        }],
        "uid": "0VQBPmUiy",
        "edit-time": 1600111383064,
        "edit-email": "raine@cybersemics.org"
    }],
    "edit-time": 1599270696877,
    "edit-email": "raine@cybersemics.org"
},
{
    "title": "September 5th, 2020",
    "children": [
    {
        "string": "X",
        "create-email": "raine@cybersemics.org",
        "create-time": 1600111456859,
        "children": [
        {
            "string": "Y",
            "create-email": "raine@cybersemics.org",
            "create-time": 1600111457621,
            "children": [
            {
                "string": "[[September 4th, 2020]]",
                "create-email": "raine@cybersemics.org",
                "create-time": 1600111458385,
                "uid": "Wt5NR3b56",
                "edit-time": 1600111478479,
                "edit-email": "raine@cybersemics.org"
            }],
            "uid": "obXRTMWqJ",
            "edit-time": 1600111458396,
            "edit-email": "raine@cybersemics.org"
        }],
        "uid": "Pu444IoIi",
        "edit-time": 1600111457625,
        "edit-email": "raine@cybersemics.org"
    }],
    "edit-time": 1599290307051,
    "edit-email": "raine@cybersemics.org"
}
```

### Thought structure:

- September 4th, 2020
  - A
    - B
      - C
- September 5th, 2020
  - X
    - Y
      - [[September 4th, 2020]]

anmolarora1:
fixes #865 

Adds some basic functions to parse ROAM JSON into { contextIndexUpdates, thoughtIndexUpdates } that can be sync'd with the state


src/@types/ROAM/index.d.ts
```diff
@@ -0,0 +1,19 @@
+declare module 'roam' {
+  import { GenericObject } from '../../utilTypes'
+
+  interface ROAMChild {
```
raineorshine, line 4: Let's call this `RoamBlock` since that's what they are called within Roam.

src/@types/ROAM/index.d.ts
```diff
@@ -0,0 +1,19 @@
+declare module 'roam' {
+  import { GenericObject } from '../../utilTypes'
```
raineorshine, line 2: `utilTypes` was removed and `GenericObject` was renamed to `Index`. Not sure how this is compiling.

src/test-helpers/exportWithoutRoot.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Remove root, de-indent (trim), and append newline to make tests cleaner.
+ */
```
raineorshine, line 3: Unless you are using this function across multiple files, it should be defined in the module it is being used. There's no point adding to the surface area of the internal API if it is only used in one module.

src/test-helpers/exportWithoutRoot.js
```diff
@@ -0,0 +1,8 @@
...
+export const exportedWithoutRoot = exported => exported.slice(exported.indexOf('\n'))
```
raineorshine, line 4: `removeRoot` would be a better name for this function.

src/util/importROAM.ts
```diff
@@ -0,0 +1,34 @@
...
+      scope: item.title,
+      children: convertROAMChildrenToBlocks(item.children)
+    }
+  })
+}
+
+/**
+ * Parses ROAM JSON and generates { contextIndexUpdates, thoughtIndexUpdates } that can be sync'd to state.
+ */
+export const importROAM = (state: State, thoughtsRanked: SimplePath, ROAM: RoamNode[]) => {
```
raineorshine, line 31: We're calling `thoughtsRanked` by the name `simplePath` now for clarity.

src/@types/ROAM/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+    uid: string,
+    string: string,
+    'create-email': string,
+    'create-time': number,
+    children?: ROAMChild[],
+    'edit-time'?: number,
+    'edit-email'?: string,
+  }
+
+  interface RoamNode {
```
raineorshine, line 14: We can call this `RoamPage`.

src/util/importROAM.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+const convertROAMChildrenToBlocks = (children: ROAMChild[]): Block[] => children.map(child => ({
+  scope: child.string,
+  children: child.children && child.children.length ? convertROAMChildrenToBlocks(child.children) : []
+}))
+
+/**
+ * Converts the ROAM JSON to an array of blocks.
+ */
+const convertROAMJSONToBlocks = (ROAM: RoamNode[]) => {
```
raineorshine, line 18: Let's call this `roamJsonToBlocks`. The word `convert` is extraneous. All pure functions "convert` one thing to another.

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,180 @@
...
+import {
+  exportContext,
+} from '../../selectors'
+import { importROAM } from '../importROAM'
+import { State, initialState } from '../initialState'
+import { RoamNode } from 'roam'
+import { exportedWithoutRoot } from '../../test-helpers/exportWithoutRoot'
+import { SimplePath, Timestamp } from '../../types'
+
+const RANKED_ROOT = [{ value: ROOT_TOKEN, rank: 0 }]
```
raineorshine, line 22: This can be imported from `constants`.

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,180 @@
...
+const testState: State = {
+  ...initialState(),
+  thoughts: {
+    contextIndex: {
+      [hashContext([ROOT_TOKEN])]: {
+        context: [ROOT_TOKEN],
+        children: [],
+        lastUpdated: '11-10-2020' as Timestamp
```
raineorshine, line 30: `11-10-2020` is not a valid `Timestamp`. A `Timestamp` is an ISO string, i.e. `new Date().toISOString()`. I recommend importing the `timestamp` function to create new `Timestamps`.

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,180 @@
...
+    .toBe(`
+- September 4th, 2020
+  - A
+    - B
+      - C
+- September 5th, 2020
+  - X
+    - Y
+      - [[September 4th, 2020]]
+`)
```
raineorshine, line 179: Nicely done!

src/@types/ROAM/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+  import { GenericObject } from '../../utilTypes'
+
+  interface ROAMChild {
+    uid: string,
+    string: string,
+    'create-email': string,
+    'create-time': number,
+    children?: ROAMChild[],
+    'edit-time'?: number,
+    'edit-email'?: string,
```
raineorshine, line 11: Great

raineorshine (changes requested):
Functionality looks great! I made some recommendations regarding naming and structure.

In addition, please rename all instances of `ROAM` to `Roam` (or `roam`, as the case may be). Do the same thing for `JSON` and `Json`.

raineorshine:
Please name PR titles succinctly. Full descriptions go in the description field.

src/test-helpers/exportWithoutRoot.js
anmolarora1: Yes, this is being used by [some tests](https://github.com/cybersemics/em/blob/2678137e03364e88b846d80096903279be79bc11/src/util/__tests__/importHtml.js#L53). Moreover, we can use it in other tests where we're using `exportContext`, to get rid of the `ROOT_TOKEN` from our tests' assertions, which is quite redundant.

src/@types/ROAM/index.d.ts
anmolarora1: You're right. I found out that it's being ignored by `eslint` based on its current configuration.

src/test-helpers/exportWithoutRoot.js
anmolarora1: That makes sense

src/util/importROAM.ts
anmolarora1: Good catch, thanks 👍

src/@types/ROAM/index.d.ts
raineorshine: Yes, thanks!

src/test-helpers/exportWithoutRoot.js
raineorshine: That sounds good. Please replace the duplicate code in other files with a reference to `exportWithoutRoot`.

src/util/__tests__/importRoam.ts
anmolarora1: Okay, I'll do that

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,246 @@
...
+  ]
+
+  const roamBlocks = [...testData[0].children, ...testData[1].children]
+
+  const {
+    thoughtIndexUpdates: thoughtIndex,
+  } = importROAM(testState, RANKED_ROOT as SimplePath, testData)
+
+  Object.keys(thoughtIndex)
+    // ignore root blocks of ROAM page since they won't have create/edit time
```
raineorshine, line 239: Let's test that the root blocks have `created` and `lastUpdated` set to a new `Timestamp`.

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,246 @@
...
+    .filter((_, index) => index !== 0 && index !== 4)
```
raineorshine, line 240: Matching by index here is quite fragile. If the test changes at all it will break, or worse create a false positive. In general we want to avoid creating dependencies between things that are not semantically related, or only incidentally related. 

Instead, I would suggest filtering out based on the key.

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,246 @@
...
+    .forEach((thoughtHash, index) => {
+      expect(thoughtIndex[thoughtHash].created).toEqual(new Date(roamBlocks[index]['create-time']).toISOString())
+      expect(thoughtIndex[thoughtHash].lastUpdated).toEqual(new Date(roamBlocks[index]['edit-time']).toISOString())
```
raineorshine, line 243: Nicely done!

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+  ]
+
+  const roamBlocks = [...testData[0].children, ...testData[1].children]
+
+  const {
+    thoughtIndexUpdates: thoughtIndex,
+  } = importROAM(testState, RANKED_ROOT as SimplePath, testData)
+
+  Object.keys(thoughtIndex)
```
raineorshine, line 238: We should also assert that `contextIndex` entries have a matching `lastUpdated`.

src/util/addThought.ts
```diff
@@ -18,9 +18,7 @@ export const addThought = (state: State, value: string, rank: number, id: string
       rank,
       id
     }),
-    created: thoughtOld && thoughtOld.created
-      ? thoughtOld.created
-      : timestamp(),
-    lastUpdated: timestamp()
+    created: thoughtOld?.created || created,
+    lastUpdated
```
raineorshine, line 22: Excellent

src/util/importJSON.ts
```diff
@@ -25,6 +25,10 @@ import {
 interface ImportHtmlOptions {
   skipRoot? : boolean,
 }
+
+/** Insert the given value at the context. Modifies contextIndex and thoughtIndex. */
+type insertThought = (value: string, context: Context, rank: number, created?:Timestamp, lastUpdated?: Timestamp) => void
```
raineorshine, line 30: Thanks for eliminating the duplicate code! 

Another way to do this is to type `insertThought` inline and then use `typeof insertThought` to type the `saveThoughts` parameter.

src/util/importROAM.ts
```diff
@@ -0,0 +1,37 @@
...
+  scope: string,
+  created: new Date(createTime).toISOString() as Timestamp,
+  lastUpdated: editTime ? new Date(editTime).toISOString() as Timestamp : timestamp(),
+  children: children && children.length ? convertRoamBlocksToBlocks(children) : []
+}))
+
+/**
+ * Converts the ROAM JSON to an array of blocks.
+ */
+const roamJsontoBlocks = (ROAM: RoamPage[]) => {
```
raineorshine, line 21: Please change all instances of `ROAM` to `roam` as appropriate. This was requested in my previous review. You should also rename `ROAM` to `Roam` in the comments.

src/util/importROAM.ts
```diff
@@ -0,0 +1,37 @@
...
+import { State } from './initialState'
+import { importJSON } from './importJSON'
+import { Block } from '../action-creators/importText'
+import { RoamBlock, RoamPage } from 'roam'
+import { timestamp } from './timestamp'
+
+/**
+ * Recursively converts the roam children to blocks.
+ */
+const convertRoamBlocksToBlocks = (children: RoamBlock[]): Block[] => children.map(({ string, children, 'edit-time': editTime, 'create-time': createTime }) => ({
```
raineorshine, line 11: Please rename as requested in previous review.

src/util/importROAM.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Converts the ROAM JSON to an array of blocks.
+ */
+const roamJsontoBlocks = (ROAM: RoamPage[]) => {
+
+  return ROAM.map((item: RoamPage) => {
+    return {
+      scope: item.title,
+      children: convertRoamBlocksToBlocks(item.children)
+    }
+  })
```
raineorshine, line 28: I recommend the more succinct inline functions when the function body consists only of a return statement:

```js
  return ROAM.map((item: RoamPage) => ({
    scope: item.title,     
    children: convertRoamBlocksToBlocks(item.children)
  }))
```

raineorshine (changes requested):

src/util/importROAM.ts
raineorshine: @anmolarora1

src/test-helpers/exportWithoutRoot.js
raineorshine: @anmolarora1

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,246 @@
...
+          'create-email': 'testemail@gmail.com',
+          'create-time': 1600111389054,
+          'edit-time': 1600111389050,
+          uid: 'BK11233',
+        }
+      ],
+    }
+  ]
+
+  const roamBlocks = [...testData[0].children, ...testData[1].children]
```
raineorshine, line 232: Let's generalize this to work for any amount of testData:

```js
const roamBlocks = testData.map(roamBlock => roamBlock.children).flat()
```

src/util/importROAM.ts
anmolarora1: Sorry about this miss, @raineorshine . Would refactor it

src/test-helpers/exportWithoutRoot.js
anmolarora1: Sure. Just to be clear, are you referring to [this file](https://github.com/cybersemics/em/blob/2678137e03364e88b846d80096903279be79bc11/src/util/__tests__/importHtml.js#L53), or all the possible use-cases of this function that are currently using `exportContext`?

raineorshine: Just duplicate functions that can be removed, like in importHtml, that look like this:

```js
  const exportedWithoutRoot = exported.slice(exported.indexOf('\n'))
    .split('\n')
    .map(line => line.slice(2))
    .join('\n')
    + '\n'
```

src/@types/ROAM/index.d.ts
anmolarora1: @raineorshine  Do you think we can do this in a different PR or maybe towards the end of this PR because there are other `.d.ts` files as well that might need to be handled?

src/util/__tests__/importRoam.ts
anmolarora1: I'm not sure what this means. Can you please explain?

src/util/importROAM.ts
anmolarora1: Yes, I observed this as well. Updated now 👍

src/@types/ROAM/index.d.ts
raineorshine: Yes, let's do a separate PR. I'd like to know what your overall plan is before starting on it.

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,271 @@
...
+    })
+
+})
+
+test('it should set the created and lastUpdated properties as current timestamp for root blocks', () => {
+  const {
+    thoughtIndexUpdates: thoughtIndex,
+  } = importRoam(testState, RANKED_ROOT as SimplePath, testData)
+
+  Object.keys(thoughtIndex)
```
raineorshine, line 260: Also assert number of keys here

src/util/importROAM.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+const roamJsonToBlocks = (roam: RoamPage[]) => roam.map((item: RoamPage) => ({
+  scope: item.title,
+  children: roamBlocksToBlocks(item.children)
+}))
+
+/**
+ * Parses Roam and generates { contextIndexUpdates, thoughtIndexUpdates } that can be sync'd to state.
+ */
+export const importRoam = (state: State, simplePath: SimplePath, roam: RoamPage[]) => {
```
raineorshine, line 50: `importRoam` should take the same options as `importJSON` so they can be passed through.

raineorshine (changes requested):
Looks good! Just a couple small points.

src/util/importROAM.ts
anmolarora1: `importRoam` is a part of `importText` now so I don't think we need this.

src/util/validateRoam.ts
```diff
@@ -0,0 +1,32 @@
...
+  'create-time': createTime,
+  'create-email': createEmail
+}: RoamBlock) =>
+  Array.isArray(children)
+  && isRoamBlock(children)
+  && typeof uid === 'string'
+  && typeof string === 'string'
+  && typeof createTime === 'number'
+  && typeof createEmail === 'string'
+)
```
raineorshine, line 19: I think it's generally cleaner to extract the mapping logic outside of functions. So `isRoamBlock` would just take a single `RoamBlock`. The aggregate can be calculated with `children.every(isRoamBlock)` in `validateRoam`. This decouples the validation from the aggregate structure.

src/util/__tests__/validateRoam.ts
```diff
@@ -0,0 +1,99 @@
...
+
+test('it returns false for an invalid string value', () => {
+  const invalidRoamString = JSON.stringify([{ ...testData[0], children: testData[0].children.map(child => ({ ...child, string: 1011 })) }])
+  expect(validateRoam(invalidRoamString)).toBe(false)
+})
+
+test('it returns false for an invalid children value', () => {
+  const invalidRoamString = JSON.stringify([{ ...testData[0], children: 1234 }])
+  expect(validateRoam(invalidRoamString)).toBe(false)
+})
```
raineorshine, line 99: Very thorough! Thanks!

src/util/__tests__/importText.js
```diff
@@ -128,3 +123,97 @@ it('two root thoughts', async () => {
...
+      ],
+    }
+  ])
+  //   /**
+  //  * Parses Roam and generates { contextIndexUpdates, thoughtIndexUpdates } that can be sync'd to state.
+  //  */
+  // export const importRoam = (state: State, simplePath: SimplePath, roam: RoamPage[]) => {
+  //   const thoughtsJSON = roamJsonToBlocks(roam)
+  //   return importJSON(state, simplePath, thoughtsJSON, { skipRoot: false })
+  // }
```
raineorshine, line 186: Old code should be removed

raineorshine (approved):
I really like how this finished up. Thorough testing!

src/util/__tests__/importRoam.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  const {
+    contextIndexUpdates: contextIndex,
+    thoughtIndexUpdates: thoughtIndex,
+  } = importRoam(testState, RANKED_ROOT as SimplePath, testData)
+
+  Object.keys(contextIndex)
+    .forEach(contextHash => {
+      expect(contextIndex[contextHash].lastUpdated).toEqual('2020-11-02T01:11:58.869Z')
+    })
```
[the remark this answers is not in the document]
raineorshine: @anmolarora1 I found an even better way to do this. Using `toMatchObject` we can assert arbitrary keys in a single assertion. This obviates the need for asserting the number of keys and has the advantage of displaying all values that fail instead of just the first one that fails in the `forEach` loop. See: https://github.com/cybersemics/em/blob/staged-loading/src/util/__tests__/roamJsonToBlocks.ts#L226-L237

The object can be built up abstractly (i.e. `reduce`), but here I prefer the hard-coded values as they are easier to read and make it easier to assert exceptions, such as `Fruits` and `Veggies` using the edit time of their child.

PR #890, opened 2020-10-23
Replaces typedoc-plugin-external-module-name with a forked version

### LINKED ISSUE #889: docs script fails
The `docs` script is failing due to a recent update, possibly #886.

Please confirm which commit broke the script, and remediate.

https://github.com/cybersemics/em/runs/1277664876

anmolarora1:
fixes #889

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,29 +5,17 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.10.3",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.3.tgz",
-      "integrity": "sha512-fDx9eNW0qz0WkUeqL6tXEXzVlPh6Y5aCDEZesl0xBGA8ndRukX91Uk44ZqnkECp01NAZUdCAl+aiQNGi0k88Eg==",
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
+      "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
       "requires": {
-        "@babel/highlight": "^7.10.3"
+        "@babel/highlight": "^7.10.4"
       }
     },
     "@babel/compat-data": {
-      "version": "7.11.0",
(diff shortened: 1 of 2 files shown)


anmolarora1:
@raineorshine Though the `docs` script error has been fixed in the forked version of the plugin, the docs aren't being generated correctly. It's probably with the configuration mismatch between the plugin and the `typedoc` itself. 
I'm looking into their documentation to figure out if we need to make any changes to our configuration.

raineorshine:
I'll hold off until merging until you either find a resolution or we decide to revert.

anmolarora1:
> I'll hold off until merging until you either find a resolution or we decide to revert.

@raineorshine  Though the error is fixed with our recent changes, I've found some compatibility issues with the plugin and `typedoc v019.2` which prevents the docs from being generated as expected.
[The function](https://github.com/christopherthielen/typedoc-plugin-external-module-name/blob/5af43c4189123b63dee3065392a06a7ea5d28126/typedoc-plugin-external-module-name.ts#L70) that is supposed to trigger the [custom name mapper](https://github.com/christopherthielen/typedoc-plugin-external-module-name/blob/5af43c4189123b63dee3065392a06a7ea5d28126/typedoc-plugin-external-module-name.ts#L136) function isn't being called at all. 
We can dig deep into this to find out if it's an issue with the `typedoc` or the `plugin`, but unless there's a strong reason for updating `typedoc`, I'd suggest we should probably revert the typedoc version for now, until we get a response from the plugin's creator on [this issue](https://github.com/christopherthielen/typedoc-plugin-external-module-name/issues/548)

raineorshine:
> [The function](https://github.com/christopherthielen/typedoc-plugin-external-module-name/blob/5af43c4189123b63dee3065392a06a7ea5d28126/typedoc-plugin-external-module-name.ts#L70) that is supposed to trigger the [custom name mapper](https://github.com/christopherthielen/typedoc-plugin-external-module-name/blob/5af43c4189123b63dee3065392a06a7ea5d28126/typedoc-plugin-external-module-name.ts#L136) function isn't being called at all.
> We can dig deep into this to find out if it's an issue with the `typedoc` or the `plugin`, but unless there's a strong reason for updating `typedoc`, I'd suggest we should probably revert the typedoc version for now, until we get a response from the plugin's creator on [this issue](https://github.com/christopherthielen/typedoc-plugin-external-module-name/issues/548)

Great, thank you for looking into this! Since it's not a trivial change, I agree we should wait until the developer responds before doing additional work.

In the mean time, let's revert typedoc and move on. Thanks!

PR #891, opened 2020-10-27
Downgrades typedoc version

anmolarora1:


package.json
```diff
@@ -101,7 +101,7 @@
     "md5": "^2.3.0",
     "raw.macro": "^0.4.2",
     "redux-mock-store": "^1.5.4",
-    "typedoc": "^0.19.2",
+    "typedoc": "^0.17.8",
```
raineorshine, line 104: Can you add `typedoc` to our [.ncurc](https://github.com/cybersemics/em/blob/dev/.ncurc.json) so that we don't accidentally auto upgrade? Thanks!

raineorshine (changes requested):

package.json
anmolarora1: Done!
And I came to know about [this package](https://www.npmjs.com/package/npm-check-updates) as well. Great job!

raineorshine (approved):

PR #899, opened 2020-11-08
Refactor filter children logic 

### LINKED ISSUE #896: Root thought is hidden when `toggleHiddenThoughts` is turned off
## Steps to Reproduce

1. Activate `toggleHiddenThoughts`
1. Create `=one`
2. Create `=two`
3. Activate `toggleHiddenThoughts`

## Current Behavior

`=one` and `=two` disappear

## Expected Behavior

The thought under `cursor` (`=two`) should not disappear. All other metaprogramming attributes should disappear.

This is a problem with the root context rendering. It works as expected in non-root contexts:

e.g.

1. Create thought `a`
1. Activate `toggleHiddenThoughts`
1. Create subthought `a/=one`
2. Create `a/=two`
3. Activate `toggleHiddenThoughts`

i.e.:

- a
  - =one
  - =two [cursor]

shresthabijay:
fixes #896 

### Changes
- Refactor filter visible child logic.
- Add separate selector for this logic.
- Add tests for hidden thoughts functionality.


shresthabijay:

src/selectors/getChildren.ts
```diff
@@ -74,3 +74,31 @@ export const getChildrenRanked = (state: State, context: Context): Child[] =>
...
+ * Returns visible children.
+ *
+ * @param state - Redux state.
+ * @param path - Parent's path.
+ * @param simplePath - Parent's simple path.
+ * @param children - Child or ThoughtContext array.
+ * @param showContexts - Parent's context view status.
+ */
+export const filterChildren = (state: State, path: Path, simplePath: SimplePath, children: (Child | ThoughtContext)[], showContexts?: boolean) => {
+  return children.filter(child => {
```
raineorshine, line 88: In general you want to avoid starting functions with `filter` (or `map`) since it conflates the filtering with the predicate. A standalone predicate function is cleaner and makes it easier to compose filters.

src/selectors/getChildren.ts
```diff
@@ -74,3 +74,31 @@ export const getChildrenRanked = (state: State, context: Context): Child[] =>
...
+/**
+ * Returns visible children.
+ *
+ * @param state - Redux state.
+ * @param path - Parent's path.
+ * @param simplePath - Parent's simple path.
+ * @param children - Child or ThoughtContext array.
+ * @param showContexts - Parent's context view status.
+ */
+export const filterChildren = (state: State, path: Path, simplePath: SimplePath, children: (Child | ThoughtContext)[], showContexts?: boolean) => {
```
raineorshine, line 87: The comment and the name of the function need some work. `filterChildren` is too general: what kind of filter? Also see below comment about the appropriate function boundary for predicates.

src/selectors/getChildren.ts
```diff
@@ -74,3 +74,31 @@ export const getChildrenRanked = (state: State, context: Context): Child[] =>
...
+    const childPath = appendChildPath(state, childSimplePath, path)
+
+    const value = showContexts
+      ? head((child as ThoughtContext).context)
+      : (child as Child).value
+
+    return state.showHiddenThoughts ||
+      // exclude meta thoughts when showHiddenThoughts is off
+      // NOTE: child.rank is not used by isChildVisible
+      isChildVisible(state, pathToContext(simplePath), { value, rank: child.rank }) ||
```
raineorshine, line 100: Both `state.showHiddenThoughts` and `isChildVisible` checks are provided used in this module, e.g. `getVisibleThoughts`. You'll want to reuse code rather than duplicating it. Notice how much higher order functions are leveraged in this module in order to make everything composable without duplication. If you need to split up functions so that they are more modular you can do that.

src/components/Subthoughts.tsx
```diff
@@ -405,17 +403,7 @@ export const SubthoughtsComponent = ({
...
-      ? head((child as ThoughtContext).context)
-      : (child as Child).value
-    return showHiddenThoughts ||
-      // exclude meta thoughts when showHiddenThoughts is off
-      // NOTE: child.rank is not used by isChildVisible
-      isChildVisible(state, pathToContext(simplePath), { value, rank: child.rank }) ||
-      // always include thoughts in cursor
-      (cursor && equalThoughtRanked(cursor[simplePath.length], child as Child))
-  })
+  const filteredChildren = filterChildren(state, resolvedPath, simplePath, children, showContexts)
```
raineorshine, line 406: `children` has already filtered out visible thoughts, so there is some redundancy between the two. This was present before your addition, but it's worth pointing out. Ideally we could find a way to remove any redundancy between the original filtering where `children` is defined and the filtering here.

raineorshine (changes requested):
This is a great start! It's a good opportunity to push your refactoring skills further and develop stronger intuition for good function boundaries that enable composability.

shresthabijay:
@raineorshine To check if a `Child` is within the cursor then we would need it's resolved `path`.  To calculate resolved 
 `path` for  all `children` of a `context` we need the parent's `path` and `simplePath`. Currently `getVisibleThoughts` doesn't include predicate to check if the child is within the cursor. I am having problem deciding if that logic needs to be handled by `getVisibleThoughts`. Since it would introduce parent's `path` and `simplePath` as dependency. `getVisibleThoughts` is being used by selectors like `getChildrenSorted` in many places. And most of those use cases doesn't seem to care if the child is within the cursor. So do you think we should create two separate functions one that includes cursor check `getVisibleThoughtsWithCursorCheck` (not necessarily with this name) and another one that doesn't `getVisibleThoughts` ?

raineorshine:
> To check if a `Child` is within the cursor then we would need it's resolved `path`.

It appears that the existing logic only uses the `Child` and the length of its parent `SimplePath` (or equivalently, `Context`), not the full `Path`. Hopefully this means you don't have to add the additional dependencies, which I agree would be too much.

```js
      // always include thoughts in cursor
      (cursor && equalThoughtRanked(cursor[simplePath.length], child as Child))
```

https://github.com/cybersemics/em/pull/899/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-ce31b1adb16bc6b8a583afa05fee00f72248145486e1098d46cb580b8267a1a9L417

> Currently `getVisibleThoughts` doesn't include predicate to check if the child is within the cursor. I am having problem deciding if that logic needs to be handled by `getVisibleThoughts`.

It seems like `getChildren` and `getChildrenSorted`, which are composed of `getVisibleThoughts`, would want to include thoughts in the cursor. The code that calls them assumes that all of the thoughts they return are actually visible to the user, so that should include what's in the cursor.

> It would introduce parent's `path` and `simplePath` as dependency.

Hopefully not, but let me know if I'm missing something.

I do think that `isChildVisible` needs to accept `Child | ThoughtContext` rather than `Child` so that it can replace this logic:

```js
      isChildVisible(state, pathToContext(simplePath), { value, rank: child.rank }) ||
```

https://github.com/cybersemics/em/pull/899/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-ce31b1adb16bc6b8a583afa05fee00f72248145486e1098d46cb580b8267a1a9L415

shresthabijay:
> Hopefully not, but let me know if I'm missing something.

I have explained the reason in earlier discussion.
https://github.com/cybersemics/em/issues/896#issuecomment-721601717

> I would also like to point out another flaw in this implementation. As you can see in line `424` that `cursor[simplePath.length]` is being compared with `child`. The current logic expects a thought that lies in a `cursor` to have `simplePath` to be proper ancestor of `cursor`. But it is not always true. Specially when `cursor` has one or more active context view.
> 
> * a
>   
>   * b
>     
>     * c ~
>       
>       * a.b
>         
>         * k (cursor)
>       * i
> * i
> * c
> 
> Here, let's analyse `a.b.c.b.k`. `simplePath` and `path` is being represented as `context` for clarity.
> 
> `['a', 'b', 'c']` is the `simplePath` of its parent which will be used here and `['a'. 'b', 'c', 'b', 'k']` is the cursor. As you can see `cursor[simplePath.length]` will return `b` instead of `k`. So `cursor[simplePath.length]` won't always return the desired `child` from the cursor. But we want this to work across context views too.

As you can see we cannot simply take `Child` at `simplePath.length` index from the `cursor`. It would not work with context views. Thoughts with metaprogramming attributes inside context view will disappear even when it lies within the cursor.

raineorshine:
> As you can see we cannot simply take `Child` at `simplePath.length` index from the `cursor`. It would not work with context views. Thoughts with metaprogramming attributes inside context view will disappear even when it lies within the cursor.

Ah, that's right.

> Do you think we should create two separate functions one that includes cursor check `getVisibleThoughtsWithCursorCheck` (not necessarily with this name) and another one that doesn't `getVisibleThoughts` ?

If you believe that will solve the issue, maybe start there and then we can refactor after? I think I got confused refactoring before fixing the issue. Thanks!

shresthabijay:
@raineorshine I made some changes. We discussed earlier that `path` and `simplePath` of a parent is required to check if it's children lie within the `cursor` path. But all the places don't need that predicate to check is a thought is visible. So I changed current `getChildrenSorted` to `getVisibleChildrenSorted`. It returns all the visible children of a `context` without `cursor` check. `getChildrenSorted` will just return correct `getThoughts` function. This way we can use `getChildrenSorted` in `Subthoughts` and use `getVisibleChildrenWithoutCursorCheck` to get filtered children.

src/selectors/getChildren.ts
```diff
@@ -46,13 +47,20 @@ export const hasChildren = (state: State, context: Context) => {
...
-  return getVisibleThoughts(getThoughtsFunction, state, context)
+  return getThoughtsFunction(state, context)
+}
+
+/** Gets all visible children within a context sorted by rank or sort preference.
+ * Note: It doesn't check if thought lies within the cursor path.
+ */
+export const getVisibleChildrenSorted = (state: State, context: Context) => {
+  return getVisibleThoughts(getChildrenSorted, state, context)
 }
```
raineorshine, line 64: Since `getVisibleThoughts` is curried, you can simplify as follows:

```js
export const getVisibleChildrenSorted = getVisibleThoughts(getChildrenSorted)
```

src/selectors/getChildren.ts
raineorshine: @shresthabijay Do you have questions about this?

src/selectors/getChildren.ts
```diff
@@ -46,13 +47,20 @@ export const hasChildren = (state: State, context: Context) => {
 /** Gets all visible children within a context. */
 export const getChildren = getVisibleThoughts(getAllChildren)
 
-/** Gets all visible children within a context sorted by rank or sort preference. */
+/** Gets all children within a context sorted by rank or sort preference. */
 export const getChildrenSorted = (state: State, context: Context) => {
```
raineorshine, line 51: Great! I like the way you factored out the visible predicate so that we can use it both ways.

There was an inversion of semantics that occurred that I would like to correct though. Currently, `getChildren*` returns only visible children by default, and `getAllChildren*` includes hidden children. This was intentional, since the most common case is to get only visible children. In your refactor, you have inverted this: the unprefixed functions include hidden children, and functions prefixed with `getVisible` do not.

Let's change this back to the original semantics, specifically:
- Rename `getChildrenSorted` to `getAllChildrenSorted`
- Rename `getVisibleChildrenSorted` to `getChildrenSorted`
- Rename `getVisibleChildrenWithCursorCheck` to `getChildrenWithCursorCheck`

src/selectors/getChildren.ts
```diff
@@ -73,4 +81,32 @@ export const getChildrenRanked = (state: State, context: Context): Child[] =>
...
+  const showContexts = isContextViewActive(state, pathToContext(path))
+
+  return children.filter(child => {
+
+    const childSimplePath = getChildPath(state, child, simplePath, showContexts)
+    const childPath = appendChildPath(state, childSimplePath, path)
+
+    const value = showContexts
+      ? head((child as ThoughtContext).context)
+      : (child as Child).value
```
raineorshine, line 105: If you'd like a little challenge, you can try to refactor the filter predicate as a composition ;)

```json
or(showHiddenThoughts, or(isChildVisible, isChildInCursor))
```

raineorshine (changes requested):

src/selectors/getChildren.ts
shresthabijay: Sure. This makes sense. Thanks!

src/selectors/getChildren.ts
shresthabijay: @raineorshine Thanks for pointing it out. I understand it now. Having a separate predicate function makes it easier to compose.

shresthabijay:
@raineorshine I have made few changes. Please review it.

src/selectors/getChildren.ts
```diff
@@ -74,3 +80,38 @@ export const getChildrenRanked = (state: State, context: Context): Child[] =>
...
+/** Checks if a child lies within the cursor path. */
+const isChildInCursor = (state: State, path: Path, simplePath: SimplePath, child: Child | ThoughtContext) => {
+  const showContexts = isContextViewActive(state, pathToContext(path))
+  const childSimplePath = getChildPath(state, child, simplePath, showContexts)
+  const childPath = appendChildPath(state, childSimplePath, path)
+  return state.cursor && equalThoughtRanked(state.cursor[childPath.length - 1], head(childPath))
+}
+
+/** Returns head of context if parent has active context view. */
+const getResolvedValue = (state: State, path: Path, child: Child | ThoughtContext) => {
```
raineorshine, line 93: Since this wants a `Context` rather than a `Path`, it would be better to accept a `Context` directly rather than accept a `Path` and then convert it to a `Context`.

raineorshine, line 93: We've removed the "resolved" terminology FYI. This is fine for now.

src/selectors/getChildren.ts
```diff
@@ -74,3 +80,38 @@ export const getChildrenRanked = (state: State, context: Context): Child[] =>
...
+  return showContexts
+    ? head((child as ThoughtContext).context)
+    : (child as Child).value
+}
+
+/** Checks if the child is visible and also checks if the child lies within the cursor. */
+const isChildVisibleWithCursorCheck = _.curry((state: State, path: Path, simplePath: SimplePath, child: Child | ThoughtContext) => {
+  return state.showHiddenThoughts ||
+  isChildVisible(state, pathToContext(simplePath), { value: getResolvedValue(state, path, child), rank: child.rank }) ||
+  isChildInCursor(state, path, simplePath, child)
```
raineorshine, line 104: This is fine, though it would be great to see you push the limits a bit more with FP. Here's a library I made with some simple `and`/`or` predicates that can be used for composition: https://github.com/raineorshine/fp-and-or

src/selectors/getChildren.ts
```diff
@@ -74,3 +80,38 @@ export const getChildrenRanked = (state: State, context: Context): Child[] =>
...
+ * Returns only visible thoughts from the children array. It checks if a thought lies within the cursor path.
+ *
+ * @param state - Redux state.
+ * @param path - Parent's path.
+ * @param simplePath - Parent's simple path.
+ * @param children - Children of the given simplePath's context.
+ */
+export const getChildrenWithCursorCheck = (state: State, path: Path, simplePath: SimplePath, children: (Child | ThoughtContext)[]) => {
+  return children.filter(isChildVisibleWithCursorCheck(state, path, simplePath))
+}
```
raineorshine, line 117: If it has to be passed `children`, it's not really `getting` the children, so this isn't the best name. This function is just wrapping filter and hiding `isChildVisibleWithCursorCheck`, so I don't think it adds much value. It may be unnecessary. Unless it is able to query the children internally, we should just export `isChildVisibleWithCursorCheck` and use that.

raineorshine (approved):
Please note the minor points of feedback, but good to merge. Thanks!

src/selectors/getChildren.ts
shresthabijay: What do you think would be the appropriate name for this ?

src/selectors/getChildren.ts
shresthabijay: Agreed! I think we should just export `isChildVisibleWithCursorCheck` and use it where necessary.

src/selectors/getChildren.ts
shresthabijay: Agreed!

src/selectors/getChildren.ts
shresthabijay: Thanks! I will take some notes from this.

src/selectors/getChildren.ts
raineorshine: `getPathValue` or `pathHead` would be slightly better, but it's fine the way it is. Don't worry about changing it.

src/selectors/getChildren.ts
raineorshine: Cool, I'll do this when I merge `dev`.

PR #907, opened 2020-11-13
Makes splitSentences undoable

anmolarora1:
Makes `splitSentences` action undoable, and adds related tests


raineorshine:
Thanks! 

Technically we don't need a test since there's no special handling of `splitSentences`. Writing a test in order to make sure we don't forget `splitSentences` in `UNDOABLE_ACTIONS` is tautological; if we remember to write the test, then by definition we have remembered `splitSentences` :).

raineorshine:
Fixed in 8b2f259a2

PR #909, opened 2020-11-14
Removes userRef from state

### LINKED ISSUE #904: Remove `userRef` from State
`state.userRef` is not serializable and should be removed from State.

We already store the user's `uid` in `state.user.uid`. Let's reconstruct the `userRef` from `uid` where needed:

```js
firebase.database().ref('users/' + state.user.uid)
```

anmolarora1:
fixes #904 

This PR is intended to remove `userRef` from the state and replace all its instances with the function ['getUserRef'](https://github.com/anmolarora1/em/blob/b5816caebf1f9f05ab9516109c40a79f7829b3f5/src/util/getUserRef.ts).
It also adds a global mock for the same since it's being used at multiple places.


src/util/__mocks__/getUserRef.ts
```diff
@@ -0,0 +1,36 @@
...
+      _.set(firebaseStore, `${path}/${key}`, value)
+    })
+    cb(null)
+  },
+  on: noop
+})
+
+/**
+ *
+ */
```
raineorshine, line 35: Comment

src/redux-middleware/__tests__/pullQueue.ts
```diff
@@ -6,6 +6,8 @@ import * as dexie from '../../data-providers/dexie'
 import getContext from '../../data-providers/data-helpers/getContext'
 import { DataProvider } from '../../data-providers/DataProvider'
 
+// mock getUserRef (firebase's database.ref)
+jest.mock('../../util/getUserRef')
```
raineorshine, line 10: Should this be mocked as logged in or logged out? Does it vary per test?

src/shortcuts/__tests__/undo.ts
```diff
@@ -341,3 +341,37 @@ it('non-undoable actions are ignored', () => {
...
+    {
+      type: 'importText',
+      path: RANKED_ROOT,
+      text: `
+        - First Sentence. Second Sentence
+        - t`
+    },
+    { type: 'setCursor', path: [{ value: 'First Sentence. Second Sentence', rank: 0 }] },
+    { type: 'splitSentences' }
+  ])
```
raineorshine, line 358: I think you may have accidentally branched off your other PR. You can remove this test as described in https://github.com/cybersemics/em/pull/907/#issuecomment-727037674. Thanks!

raineorshine (changes requested):
Looks great! Good solution, solves the problem nicely.

Just some small requests, and a question about a different mock needed in `dexie` and `pullQueue` tests.

src/data-providers/__tests__/dexie.ts
```diff
@@ -25,6 +24,9 @@ jest.mock('lodash', () => {
   )
 })
 
+// mock getUserRef (firebase's database.ref)
+jest.mock('../../util/getUserRef')
+
```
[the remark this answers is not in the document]
anmolarora1: It was mocked because the function `getUserRef` was always expected to return a valid value (an invalid assumption). Now that it returns `Ref | null`, we don't need to mock it

src/redux-middleware/__tests__/pullQueue.ts
anmolarora1: No, it doesn't vary per test. Besides, it works even without the mock (with the newly added support for null values) so we can safely remove it now.
Whether it should be mocked as `logged in` or `logged out` is something that I'd let you answer since you're more familiar with this function

src/shortcuts/__tests__/undo.ts
anmolarora1: That's correct. Removed it now

src/redux-middleware/__tests__/pullQueue.ts
raineorshine: I think it should be mocked as logged in since it's only mocked in the firebase tests now.

raineorshine:
Thanks! You can push your final changes any time.

src/redux-middleware/__tests__/pullQueue.ts
anmolarora1: Okay, I've updated the test

src/util/getUserRef.ts
```diff
@@ -0,0 +1,7 @@
+import { State } from './initialState'
+import { Ref } from '../types'
+
+/**
+ * Get the user ref from an authenticated user's details stored in the state.
+ */
+export const getUserRef = (state: State): Ref | null => window.firebase?.database().ref('users/' + state.user!.uid)
```
raineorshine, line 7: You changed the return type but I think you forgot to change the actual function so it returns `null` if the user is not logged in.

raineorshine (approved):

PR #914, opened 2020-11-17
Adds multiline check to cursorDown shortcut

### LINKED ISSUE #913: cursorDown does not work on empty thought
## Steps to Reproduce

1. Create empty thought
2. Create a thought after the empty thought
3. cursorDown

## Current Behavior

Nothing happens

## Expected Behavior

Cursor should move to the next though.

anmolarora1:
fixes #913 

Problem - The function [`isSelectionOnLastLine`](https://github.com/cybersemics/em/blob/394cbff80b687031db4a0d76cc6b0618ff72f52c/src/shortcuts/cursorDown.tsx#L15) was being [called](https://github.com/cybersemics/em/blob/394cbff80b687031db4a0d76cc6b0618ff72f52c/src/shortcuts/cursorDown.tsx#L57) for every `cursorDown` action, and is currently based on an assumption that every thought would be either a multiline thought or have a valid `selection`
An empty thought, however, doesn't have any valid selection, and the function always [evaluates to false](https://github.com/cybersemics/em/blob/394cbff80b687031db4a0d76cc6b0618ff72f52c/src/shortcuts/cursorDown.tsx#L24), hence, skipping the shortcut execution.

Solution - Check if a thought has a valid selection at all, before checking the cursor position within a multi-line editable.

### DIFF
--- a/src/shortcuts/cursorDown.tsx
+++ b/src/shortcuts/cursorDown.tsx
@@ -43,7 +43,6 @@ const getSelectionAttributes = (): SelectionAttributes | null => {
   const { y: baseNodeY, height: baseNodeHeight } = baseNodeParentEl.getClientRects()[0]
   const [paddingTop, , paddingBottom] = getElementPaddings(baseNodeParentEl)
 
-  // allow error of 5px
   const isMultiline = Math.abs(rangeY - baseNodeY - paddingTop) > 0
   return isMultiline ? { rangeY, rangeHeight, baseNodeY, baseNodeHeight, paddingTop, paddingBottom } : null
 }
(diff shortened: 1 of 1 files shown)


src/shortcuts/cursorDown.tsx
```diff
@@ -53,11 +71,9 @@ const cursorDownShortcut: Shortcut = {
...
     if (isProseMode) return false
 
-    // use default browser if selection is not on the last line of a multi-line editable
-    const isOnLastLine = isSelectionOnLastLine()
-    if (!isOnLastLine) return false
-
-    return true
+    const selectionAttributes = getSelectionAttributes()
+    // use default browser if there is a valid selection and it's not on the last line of a multi-line editable
+    return selectionAttributes ? isSelectionOnLastLine(selectionAttributes) : true
```
raineorshine: This can be simplified to:

```js
return !selectionAttributes || isSelectionOnLastLine(selectionAttributes)
```

Boolean literals in conditionals are usually sign that the expression can be simplified. Though your use of the ternary is an improvement on what was there before!

raineorshine (approved):
Thank you for the great writeup!

Issue #894, opened 2020-11-02
Superscript positioning breaks on edit


raineorshine:
Reproduced on mobile. Not sure if it's also a problem on desktop. It may be reproducible on desktop by simulating mobile via setting isMobile to true.

This is a fairly recent regression. The width of the `ThoughtAnnotation` is what determines the superscript positioning, so the bug must be there. I think it's updating the `ThoughtAnnotation` text based on `editingValue` even when a different thought is edited. 

For thoroughness, please identify the commit that broke this and make sure that fixing it doesn't revert whatever that commit fixed. 

## Steps to Reproduce

- a
  - m
- b
  - m

1. Edit `a` to `alabama`
2. Close keyboard/blur selection
2. Click on `m`

## Current Behavior

Superscript is rendered too far to the right. 

## Expected Behavior

Superscript should be rendered in the correct place on the upper right corner of the `m`.

shresthabijay:
@raineorshine I was not able to reproduce this issue.

raineorshine:
I missed a step: blur the selection before clicking on `m`. Tested on Safari desktop.

shresthabijay:
Hi Raine! I was able to reproduce this issue. On closer inspection I found that the cause for the issue is very similar to the #895. `editingValue` was out of sync. 

https://github.com/cybersemics/em/blob/0e95af78232b43f5b52af8e054616cf5e5e753c9/src/components/ThoughtAnnotation.tsx#L63

As you can see here that `isEditing` is `true`.

https://github.com/cybersemics/em/blob/0e95af78232b43f5b52af8e054616cf5e5e753c9/src/components/ThoughtAnnotation.tsx#L49-L56

Since `editingValue` is not in sync. `getSubThoughtTextMarkup` returns wrong text. 

I confirmed that the PR for #895  also solves this issue. We can additionally add this to make sure that `editingValue` is not used when `editing` is `false`.

```javascript
const isEditing = editing !== null && editing && equalPath(cursor, props.path)
```

raineorshine:
> I confirmed that the PR #915 solves this issue too. We can additionally add this to make sure that `editingValue` is not used when `editing` is `false`.

Perfect!

I'll hold off on the other change—`isEditing` is used in many places and always refers to whether the cursor is on the thought or not, regardless of edit mode. Changing it here would create an inconsistency, and changing it everywhere would break other code. Making sure `editingValue` stays in sync should be sufficient.

raineorshine:
[...]

PR #915, opened 2020-11-17
Editable onBlur: set editingValue to null

raineorshine:


src/components/Editable.tsx
```diff
@@ -473,15 +473,17 @@ const Editable = ({ disabled, isEditing, simplePath, path, cursorOffset, showCon
...
+
     if (isMobile && isSafari()) {
       resetToolbarPosition()
       document.removeEventListener('scroll', updateToolbarPositionOnScroll)
     }
 
     const { invalidState } = state
     throttledChangeRef.current.flush()
-    // set editingValue in order to position superscript correctly if edited thought is duplicate
-    oldValueRef.current && dispatch(setEditingValue(oldValueRef.current))
```
raineorshine, line 484: Superscript positioning is broken anyway and needs to be solved a different way, so we can remove this. See #894.

PR #918, opened 2020-11-19
Defers setCursor until `=archive` context is added for archive action

### LINKED ISSUE #911: Archiving the last thought moves cursor to archived thought
Archiving the last thought in the root moves the cursor to the archived thought even when `showHiddenThoughts` is `false`.

## Steps to Reproduce

1. Create thought `a`
2. Archive `a`

## Current Behavior

Cursor set to `=archive/a`

## Expected Behavior

Cursor should be set to `null`

anmolarora1:
fixes #911 

**Problem**
The cursor is correctly being set to null by [the setCursor call](https://github.com/cybersemics/em/blob/547b2c9049c2cc049897e5bc2e07155f56866f40/src/reducers/archiveThought.ts#L128) within `archiveThought` if there's only one thought. For the very first thought, we don't have `=archive` context in place, so we're [creating a new thought](https://github.com/cybersemics/em/blob/547b2c9049c2cc049897e5bc2e07155f56866f40/src/reducers/archiveThought.ts#L144) for that. However, this eventually triggers the `updateThoughts` reducer, which tries to [decode the cursor from null](https://github.com/cybersemics/em/blob/547b2c9049c2cc049897e5bc2e07155f56866f40/src/reducers/updateThoughts.ts#L137) (based on the assumption that it's due to page load). So the cursor gets reset to the original thought: `{value: "a", rank: 0, id: ""}`. The [`existingThoughtMove`](https://github.com/cybersemics/em/blob/547b2c9049c2cc049897e5bc2e07155f56866f40/src/reducers/archiveThought.ts#L162) [updates the cursor](https://github.com/cybersemics/em/blob/547b2c9049c2cc049897e5bc2e07155f56866f40/src/reducers/existingThoughtMove.ts#L288) yet again, with an underline assumption that the cursor would be null for and won't be updated (which isn't true for the very first thought).

**Solution**
By calling [setCursor](https://github.com/anmolarora1/em/blob/7b454ea4ec28e8ea7ed56a4f07ea639fdeb09e8e/src/reducers/archiveThought.ts#L150) after the [new thought ](https://github.com/anmolarora1/em/blob/7b454ea4ec28e8ea7ed56a4f07ea639fdeb09e8e/src/reducers/archiveThought.ts#L142)(`=archive`) has been created, so that the cursor is correctly set to `null` we can ensure that `existingThoughtMove` doesn't move the cursor to an archive thought (or put simply, update the cursor)

### DIFF
--- a/src/reducers/archiveThought.ts
+++ b/src/reducers/archiveThought.ts
@@ -123,19 +123,16 @@ const archiveThought = (state: State, { path }: { path: Path }): State => {
   }
 
   return reducerFlow([
+
     ...isDeletable
+
       ? [
-        // set the cursor away from the current cursor before archiving so that existingThoughtMove does not move it
-        setCursor({
-          path: cursorNew,
-          editing: state.editing,
-          offset,
-        }),
         existingThoughtDelete({
           context: showContexts ? context : parentOf(pathToContext(simplePath)),
           showContexts,
           thoughtRanked: head(simplePath),
         })]
+
       : [
         // create =archive if it does not exist
         (state: State) => !hasChild(state, context, '=archive')
@@ -147,11 +144,6 @@ const archiveThought = (state: State, { path }: { path: Path }): State => {
             preventSetCursor: true
           })
           : null,
-        setCursor({
-          path: cursorNew,
-          editing: state.editing,
-          offset,
-        }),
 
         // undo alert
         alert({
@@ -162,15 +154,21 @@ const archiveThought = (state: State, { path }: { path: Path }): State => {
         }),
 
         // execute existingThoughtMove after newThought has updated the state
-        (state: State) => {
-          return existingThoughtMove(state, {
-            oldPath: showContexts ? simplePath : path,
-            // TODO: Are we sure pathToArchive cannot return null?
-            newPath: pathToArchive(state, showContexts ? simplePath : path, context)!,
-            offset
-          })
-        }
-      ]
+        (state: State) => existingThoughtMove(state, {
+          oldPath: path,
+          // TODO: Are we sure pathToArchive cannot return null?
+          newPath: pathToArchive(state, showContexts ? simplePath : path, context)!,
+          offset
+        })
+
+      ],
+
+    setCursor({
+      path: cursorNew,
+      editing: state.editing,
+      offset,
+    }),
+
   ])(state)
 }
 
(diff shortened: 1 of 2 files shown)


src/reducers/archiveThought.ts
```diff
@@ -123,24 +123,22 @@ const archiveThought = (state: State, { path }: { path: Path }): State => {
...
-      : reducerFlow([
-
+    ...isDeletable
+      ? [
+        // set the cursor away from the current cursor before archiving so that existingThoughtMove does not move it
+        setCursor({
+          path: cursorNew,
+          editing: state.editing,
+          offset,
+        }),
```
raineorshine, line 133: Why not call `setCursor` once at the tend, since it's the same regardless of `isDeletable`? It seems strange to duplicate this code in both cases, i.e.

```js
if (a) {
  b()
  c()
}
else {
  d()
  b()
}
```

raineorshine (commented):
Brilliant?

src/reducers/archiveThought.ts
anmolarora1: @raineorshine  I noticed that duplication earlier and tried moving the `setCursor` to the end. But some of the `archiveThoughts` tests start to break if we do that, so I believe that the order of execution is important here (as implied by [this comment](https://github.com/cybersemics/em/blob/3d3ed667401c08b2b95158037988eca51a9102c4/src/reducers/archiveThought.ts#L127)). 
```
// set the cursor away from the current cursor before archiving so that existingThoughtMove does not move it
```

raineorshine: I did some more investigating and found the cause of the ordering issue. `existingThoughtMove` was not just setting the cursor incorrectly but crashing, so I suspected there was more to it. 

`archiveThought` was passing `SimplePaths` to `existingThoughtMove` instead of `Paths`, thus preventing the `cursor` updating logic from working. `existingThoughtMove` still needed `SimplePath` for some things (e.g. context, rank, and id), so I had to update it to do an internal `simplifyPath` conversion for just those particular cases, without changing the `Path` that was used for cursor calculations.

raineorshine:
I tried to make a test for this but couldn't make it fail! I even did a fully mounted test with a delay to give time for `updateThoughts` to be called. This is a hole in our test coverage that I'd like to figure out some day, but I will let it be for now.

Issue #884, opened 2020-10-12
Back/Forward after deleting results in invalid cursor


raineorshine:
History back/forward should never result in an invalid cursor. History states with invalid `Paths` should be skipped. Thus if there are three invalid history states in a row (e.g. due to a thought being archived and `showHiddenThoughts` being `false`), then they should all be skipped and the next valid `Path` in the cursor history should be restored.

Start with a failing unit test.

## Steps to Reproduce

- a
- b
- c [cursor]

1. Create `a`, `b`, `c`, in order as shown above.
2. Select `a`.
2. Archive `a`.
3. Hit Back in the browser

## Current Behavior

Cursor disappears and url changes to `/a` even though `a` has been archived.

## Expected Behavior

Cursor should move to `c`.

anmolarora1:
@raineorshine I wasn't able to replicate the issue. Can you please confirm if it still exists?

raineorshine:
The current behavior has changed, but the expected behavior is still not happening.

anmolarora1:
After some further digging, I think the best place to handle this would be [the updateUrlHistory middleware](https://github.com/cybersemics/em/blob/dev/src/redux-middleware/updateUrlHistory.ts). 
IMO, we should simply skip the archived thoughts' states from the window's history object, and the browser's back button would work as we're expecting.
Albeit unused, the logic to handle this scenario (or something similar) is in place already. [This is](https://github.com/cybersemics/em/blob/5baf45d1bc0e188dd2dd744dd806f3c885ef3100/src/redux-middleware/updateUrlHistory.ts#L55) where we're deciding whether to replace the push to the state or replace it.

Also, [the throttled function](https://github.com/cybersemics/em/blob/5baf45d1bc0e188dd2dd744dd806f3c885ef3100/src/redux-middleware/updateUrlHistory.ts#L68) doesn't work correctly for varying arguments.
[Here's the issue](https://github.com/lodash/lodash/issues/2403) which captures this problem. So we might have to combine the `throttle` function with `memoize`.
@raineorshine What are your thoughts?

raineorshine:
> Albeit unused, the logic to handle this scenario (or something similar) is in place already. [This is](https://github.com/cybersemics/em/blob/5baf45d1bc0e188dd2dd744dd806f3c885ef3100/src/redux-middleware/updateUrlHistory.ts#L55) where we're deciding whether to replace the push to the state or replace it.

[...]

PR #920, opened 2020-11-27
On history back, skip invalid Paths

raineorshine:
#884

- Doesn't handle `forward`
- Doesn't skip duplicate states


raineorshine:
Calling `history.back()` multiple times in succession is too glitchy. I think it would be better to just move the cursor to the deepest valid ancestor.

PR #922, opened 2020-11-30
884

### LINKED ISSUE #884: Back/Forward after deleting results in invalid cursor

anmolarora1:
fixes #884 

@raineorshine As discussed, this implementation is based on [our discussion](https://github.com/cybersemics/em/issues/884#issuecomment-734508411) over the issue. The back button works as expected, whereas the forward button doesn't. 
Please share your thoughts, and we'll proceed accordingly.


src/util/initEvents.ts
```diff
@@ -16,16 +16,25 @@ declare global {
...
+  let lastState: number
+
   /** Popstate event listener; setCursor on browser history forward/backward. */
-  const onPopstate = () => {
+  const onPopstate = (e: PopStateEvent) => {
 
     const state = store.getState()
 
-    const { path, contextViews } = decodeThoughtsUrl(state, window.location.pathname)
+    const { path, contextViews } = decodeThoughtsUrl(state, window.location.pathname, { exists: true })
```
raineorshine, line 27: If you add `exists: true`, then `path` will be `null` in the case of a nonexistent pathname and you won't be able to differentiate it from the root context, as described here:

> It seems that `decodeThoughtsUrl` is a little inconsistent; when `exists: true` is given it returns `null` for a valid root context. This might be a bug.

https://github.com/cybersemics/em/issues/884#issuecomment-734508411

Not sure if that was intended, but I think with the current implementation of `decodeThoughtsUrl` you'll have to omit `exists` and use `pathExists` directly.

src/redux-middleware/updateUrlHistory.ts
```diff
@@ -53,7 +52,8 @@ const updateUrlHistory = (state: State, path = RANKED_ROOT, { replace, contextVi
   // update browser history
   try {
     window.history[replace ? 'replaceState' : 'pushState'](
```
raineorshine, line 54: There is something causing `[ROOT]` to get pushed onto the stack when skipping an invalid history state. This seems to happen on `dev` as well. I would suggest identifying the cause of that, as it might be interfering with the skipping back/forward functionality.

raineorshine (commented):
Let me know how else I can help, or if you get stuck!

src/util/initEvents.ts
anmolarora1: @raineorshine Removing `exists:true` solves the problem. You can test the functionality now.
As discussed, I'll look into the `decodeThoughtsUrl` bug and it's relation to the state in a different issue.

raineorshine (changes requested):
Great, back and forward are working for the basic case now. One thing:

Duplicate states should also be skipped. i.e. if the user navigates `a`, `b`, `c`, `b`, and `c` gets deleted, then hitting back on `b` should go to `a`, not the other `b`.

raineorshine:
Looks good, thanks!

PR #929, opened 2020-12-09
Adds store enhancer to detect cursor changes and handle clear selection

### LINKED ISSUE #829: Remove side effect from setCursor
https://github.com/cybersemics/em/blob/dev/src/reducers/setCursor.ts#L44-L48

It might have to go in every cursor action creator that ends up calling `setCursor`, but if you can think of a solution with less surface area we should do that.

anmolarora1:
fixes #829

### DIFF
--- a/src/redux-enhancers/cursorChanged.ts
+++ b/src/redux-enhancers/cursorChanged.ts
@@ -1,0 +1,50 @@
+import { Action, Store, StoreEnhancer, StoreEnhancerStoreCreator } from 'redux'
+import { RANKED_ROOT } from '../constants'
+import { chain, simplifyPath } from '../selectors'
+import { Index, Path, SimplePath } from '../types'
+import { clearSelection, equalPath, headValue, isDivider } from '../util'
+import { State } from '../util/initialState'
+
+interface SetCursorAction extends Action<any> {
+    contextChain?: SimplePath[],
+    cursorHistoryClear?: boolean,
+    cursorHistoryPop?: boolean,
+    editing?: boolean | null,
+    noteFocus?: boolean,
+    offset?: number | null,
+    replaceContextViews?: Index<boolean>,
+    path: Path | null,
+}
+
+/**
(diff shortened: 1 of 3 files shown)


src/redux-enhancers/cursorChanged.ts
```diff
@@ -0,0 +1,50 @@
...
+    cursorHistoryPop?: boolean,
+    editing?: boolean | null,
+    noteFocus?: boolean,
+    offset?: number | null,
+    replaceContextViews?: Index<boolean>,
+    path: Path | null,
+}
+
+/**
+ * Store enhancer to append the ability to undo/redo for all undoable actions.
```
raineorshine, line 20: Comment is incorrect

src/redux-enhancers/cursorChanged.ts
```diff
@@ -0,0 +1,50 @@
...
+      const { path, contextChain } = action
+      const simplePath = path ? simplifyPath(state, path) : RANKED_ROOT
+      const thoughtsResolved = path && contextChain && contextChain.length > 0
+        ? chain(state, contextChain, simplePath!)
+        : path || updatedState.cursor
+
+      // clear browser selection if the cursor is on a divider
+      if (thoughtsResolved && isDivider(headValue(thoughtsResolved))) {
+        clearSelection()
+      }
```
raineorshine, line 41: I'm not sure I understand why all this logic is necessary. Can you just check `updatedState.cursor && isDivider(headValue(updatedState.cursor))`?

raineorshine (changes requested):

src/redux-enhancers/cursorChanged.ts
anmolarora1: You're right, checking that alone would suffice because we don't need to rely on the action payload anymore since we're using the computed state.
I'll update this

raineorshine (approved):
Looks good, thanks!

PR #933, opened 2020-12-17
Search remote

### LINKED ISSUE #931: Server-Side Search
With the introduction of the `pullQueue`, not all thoughts are loaded into `state`. This breaks the current implementation of search. We need to change it to search Firebase, if logged in, or IndexedDB otherwise.

- The behavior should be close to the previous state search, although we may need to add a `LoadingEllipsis`.
- Firebase search should be done server-side. You might need to create your own `Firebase` database to test this.
- Need to search IndexedDB without loading the entire `thoughtIndex` into memory somehow.

shresthabijay:
fixes #931
Basic search functionality with cloud functions and algolia.


raineorshine:
This is great! I tested it and it works well.

shresthabijay:
@raineorshine I have pushed some changes to the branch. The changes are as follows:

- I have added `searchContexts` that is used by `pullQueue` to dispatch `pull.`
-  `searchContexts` is populated by `SearchSubthought` after retrieving search `thoughtHash` results from algolia and using `firebaseProvider` to get the contexts for these `thoughtHash`. 
- Rest is handled by `pullQueue`

However there are few challenges I am facing. 

- Since search results can contain thoughts that are deeply nested. Currently it only retrieves it's context and few children as specified by  `BUFFER_DEPTH`. However we may want to pull its all direct ancestors too. It seems odd to have just a thought in the memory that doesn't have it's ancestors loaded yet. What do you think ?

- Also some thoughts are not being loaded after expanding thoughts in search mode.

raineorshine:
> * Since search results can contain thoughts that are deeply nested. Currently it only retrieves it's context and few children as specified by  `BUFFER_DEPTH`. However we may want to pull its all direct ancestors too. It seems odd to have just a thought in the memory that doesn't have it's ancestors loaded yet. What do you think ?

That's a good point; `pending` only works for descendants, not ancestors. I'm not sure what the implications are for having a thought loaded but not its ancestors. 

The problem with loading all ancestors is that we're working in a graph structure, not a tree. Every thought can have multiple parents, and every parent can have multiple parents. Loading all the ancestors could be a large amount of data (even infinite since cycles are allowed). We would have to implement the `pending` mechanic in reverse; `pending` ancestors so that we only loaded up to `BUFFER_DEPTH`. I'm not sure this is necessary though. The context view only shows one level of parents. If a thought is loaded into state, then its `Lexeme` is loaded and we will have access to those parents. I have not thoroughly tested iteratively loading on the context view, so there are bound to be some issues there. Unless you can anticipate a specific issue, we'll know when we encounter it, or when we write more comprehensive tests for the context view.

> * Also some thoughts are not being loaded after expanding thoughts in search mode.

I believe they will need `pending` to be set properly, and then the `pullQueue` will load in the next descendants when they are expanded. This use case hasn't been tested though and the `pullQueue` will likely need some modifications.

> * I have added `searchContexts` that is used by `pullQueue` to dispatch `pull.`
> * `searchContexts` is populated by `SearchSubthought` after retrieving search `thoughtHash` results from algolia and using `firebaseProvider` to get the contexts for these `thoughtHash`.
> * Rest is handled by `pullQueue`

I think that makes sense, since the search results would not show up in `visibleContexts`. I can think more about it when I'm back at work later this week.

raineorshine:
I would recommend fixing the linter ASAP since it's blocking the tests from running. Better to catch a regression early rather than at the end.

shresthabijay:
@raineorshine There is a problem with `pullQueue` not being able to correctly pull for thoughts inside context view. The reason for the issue is that we are using context that includes context view. Instead we need to use simple context that doesn't include context view.

For example:
- a
  - b 
   - c  ~ (cursor)
      - a.k
        - i 
Here a.b.c.k is the context of the cursor. Since this context includes context view, we cannot use it to pull from the sever or indexed db. We need to convert this context to context chain and use last simple context to pull the data.

One possible solution would be to use `splitChain`  for each context in `pullQueue` index. But `splitChain` uses `Path` instead of `Context`. It also use `rankThoughtsFirstMatch` which may not work properly because all the required thoughts may not have loaded yet.

raineorshine:
> Instead we need to use simple context that doesn't include context view.

Makes sense

> The reason for the issue is that we are using context that includes context view.

Could you point out where the unranked `Path` is being passed?

> * Here a.b.c.k is the context of the cursor. Since this context includes context view, we cannot use it to pull from the sever or indexed db.

I'm wondering how this ends up getting passed to the `pullQueue`. I would think that it would get simplified before being passed.

> One possible solution would be to use `splitChain` for each context in `pullQueue` index. But `splitChain` uses `Path` instead of `Context`. It also use `rankThoughtsFirstMatch` which may not work properly because all the required thoughts may not have loaded yet.

Hopefully we can simplify it before it gets to that point, but let me know if I'm missing something.

shresthabijay:
> I'm wondering how this ends up getting passed to the `pullQueue`. I would think that it would get simplified before being passed.
> 
> >

We are using `expanded` in `pullQueue`. `expanded` includes context view with context view. 

image

Here `a.b.c.k` is also in the `expanded`. This way it is getting passed to the `pullQueue`.

raineorshine:
Is it possible to simplify the unranked `Path` in `pullQueue`? Or is this affected by what you stated (below)?

> It also use rankThoughtsFirstMatch which may not work properly because all the required thoughts may not have loaded yet.

On another note, the context view consists of the `Lexeme`'s `ThoughtContexts` which contain actual `Contexts` (not unranked `Paths`). Theoretically that should give us what we need to pull the contexts shown in the context view. I haven't thought through the relationship between `expanded` and context view though. 

> One possible solution would be to use `splitChain` for each context in `pullQueue` index. But `splitChain` uses `Path` instead of `Context`. It also use `rankThoughtsFirstMatch` which may not work properly because all the required thoughts may not have loaded yet.

Maybe you've thought through everything I mentioned above and this is your conclusion. I'm not quite there yet in my own understanding and am hoping to avoid low-level functions like `splitChain`, but maybe that's not possible. Thanks!

shresthabijay:
#1055 supersedes this PR.

Issue #942, opened 2020-12-25
=pin not working in ROOT


raineorshine:
## Steps to Reproduce

- a
  - =pin
    - true
  - b
- c

Remove the cursor and refresh the page. 

## Current Behavior

`a` is collapsed (although its bullet indicates expanded).

## Expected Behavior

`a` should be expanded.

PR #947, opened 2020-12-26
Rerender editable on thoughts bump

### LINKED ISSUE #935: Bump Thought Down does not re-render root context
## Steps to Reproduce

- a
  - b

Activate "Bump Thought Down" on `a`.

## Current Behavior

The cursor thought does not re-render, making it appear like:

- a
  - a
  - b

Moving the cursor causes a re-render, showing that the thoughts are actually correct:

- [empty]
  - a
  - b

## Expected Behavior

The cursor thought should be re-rendered so the correct thoughts are visible. This may be related to the special `mapStateToProps` logic in the `Content` component.

shresthabijay:
fixes #935

Turns out it was not the problem with `mapStateToProps` logic in `Content`. Actually we had to just call `editableRender` reducer on thoughts bump. This will let `ContentEditable` know that it is safe to set `innerRef` and updates the value.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ yarn-error.log*
 .idea
 .history
 .eslintcache
+
+.vscode
--- a/src/components/__tests__/Editable.ts
+++ b/src/components/__tests__/Editable.ts
@@ -1,0 +1,46 @@
+import { ReactWrapper } from 'enzyme'
+import { store } from '../../store'
+import { RANKED_ROOT } from '../../constants'
+import createTestApp, { cleanupTestApp } from '../../test-helpers/createTestApp'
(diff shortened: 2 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}
```
raineorshine, line 3: Let's add `.vscode` to our `gitignore`

raineorshine:
Thanks! Also check out #942; seems similar.

raineorshine (changes requested):
Can you add tests? Thanks!

raineorshine (approved):

PR #951, opened 2020-12-27
Rerender Sub thought on expand

### LINKED ISSUE #942: =pin not working in ROOT

shresthabijay:
fixes #942 

## Changes
- Added `isExpanded` to `mapStateToProps` of `Subthought` to re-render it's children if thought is either expanded or collapsed.

### DIFF
--- a/src/components/__tests__/Subthoughts.ts
+++ b/src/components/__tests__/Subthoughts.ts
@@ -3,12 +3,13 @@ import { store } from '../../store'
 import createTestApp, { cleanupTestApp } from '../../test-helpers/createTestApp'
 import { RANKED_ROOT } from '../../constants'
 import { equalArrays, pathToContext, timestamp } from '../../util'
-import { importText, setCursor } from '../../action-creators'
+import { importText, setCursor, toggleAttribute } from '../../action-creators'
 import Editable from '../Editable'
 import Thought from '../Thought'
 import Subthoughts from '../Subthoughts'
 import { Context, Path, SimplePath } from '../../types'
 import { setCursorFirstMatchActionCreator } from '../../test-helpers/setCursorFirstMatch'
+import { getAllChildren } from '../../selectors'
 
 // type for Thoughts or Subthoughts component that has a simplePath prop
 interface ThoughtOrSubthoughtsComponent {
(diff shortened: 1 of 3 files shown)


raineorshine (changes requested):
Can you add tests? Thanks!

shresthabijay:
@raineorshine I have added a test for this particular scenario where `Subthought` needs to re-render when `isExpanded` changes.

src/components/__tests__/Subthoughts.ts
```diff
@@ -396,3 +397,61 @@ describe('hidden thoughts', () => {
     expect(nestedMetaThought2Child()).toHaveLength(1)
   })
 })
+
+describe('expand thoughts', () => {
+
+  it('re-render Subthought if it is expanded', () => {
```
raineorshine, line 403: This is a good test, although it does not test the issue in #942. This tests whether toggling the `=pin` value will be re-rendered properly, which was always working. #942 is a problem with the initial render (no toggle).

Is there a way to test that case? If it's too complicated that's okay, but I think it's worth a discussion.

shresthabijay: @raineorshine I have considered this. Creating that exact scenario of initial render requires populating db with `thoughts` and then mounting the app. However the core problem was that  a `Subthought` wouldn't re render if it `store.getState().expanded[]hashContext(context)])` changes.  This is now calculated and returned by `mapStateToProps`, so we can be sure that it will re-render if it changes. This test exactly recreates this situation .  I have set cursor to `null` to not affect `expanded` and manually toggled pin open attribute in a way that it updates `expanded` property but not `children` of the context. I ran this same test in `dev` which failed exactly like the issue we encountered.

I admit this test is kinda hacky. It is dependent to the implementation of `toggleAttribute`. If we want to test this in a clean way we either have to populate db before mounting app or we would need a way to update `expanded` to desired value maybe with a test reducer.

raineorshine: I stand corrected! Thank you for the explanation. The test was different looking enough from the issue that I didn't realize you had found a way to test the same mechanism . As you pointed out, it's better to test the user interaction rather than an implementation detail. Let's leave this for now, and we can come back to it in the future when we wish to refactor. 

Here is an example of simulating a page refresh for future reference: https://github.com/cybersemics/em/blob/dev/src/redux-middleware/__tests__/pullQueue.ts#L76-L80.

raineorshine (approved):

PR #957, opened 2020-12-31
Fix caret update on empty thought

### LINKED ISSUE #936: Caret not updated when moved from empty thought
## Steps to Reproduce

- [empty]
  - a

1. Set the cursor on the empty thought
2. Activate `cursorDown`

## Current Behavior

The cursor moves to `a` but the caret stays on the empty thought.

## Expected Behavior

The caret should move with the cursor.

## Notes

This occurs in all contexts (including root) and when the cursor is moved through any means except a click.

shresthabijay:
fixes #936 

## Cause of the issue

Empty strings are falsey in js. This check returns false even for empty string, not just for null value. So on empty thought, `cursorOffset` was being set to `null`

https://github.com/cybersemics/em/blob/4885039066a2315426b0728e2094b6d1f0b92699/src/reducers/setCursor.ts#L134

That prevented `setSelection` from being called in `Editable`.

https://github.com/cybersemics/em/blob/4885039066a2315426b0728e2094b6d1f0b92699/src/components/Editable.tsx#L314-L322

## Changes

Used strict `null` check to prevent falsey empty string.

### DIFF
--- a/src/reducers/setCursor.ts
+++ b/src/reducers/setCursor.ts
@@ -131,7 +131,7 @@ const setCursor = (state: State, {
       : state.cursorHistory,
       // set cursorOffset to null if editingValue is null
       // (prevents Editable from calling setSelection on click since we want the default cursor placement in that case)
-      cursorOffset: offset ?? (state.editingValue ? 0 : null),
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks so much!

PR #958, opened 2020-12-31
Add isActive to Shortcut

### LINKED ISSUE #939: Refactor Toolbar icon highlighting logic
Each icon in the toolbar is highlighted under certain conditions, e.g. when the cursor is on a context with `=view/Table`, then the table icon is highlighted. Currently these conditions are defined in the Toolbar component which tightly couples the Toolbar with specific shortcuts.

https://github.com/cybersemics/em/blob/611b6215eeb0afb8c59bfb51eb0a641a47be5216/src/components/Toolbar.tsx#L235-L258

Instead, add an `isActive?: (state: State) => boolean` property to the `Shortcut` type and perform the shortcut-specific logic in each shortcut. Then Toolbar will only be coupled to the `isActive` signature itself. Refactor the Toolbar component to set the SVG fill to `fg` if `isActive` returns `true` or does not exist, and `gray` otherwise. (This is the same functionality that exists now, just a re-architecture.)

shresthabijay:
fixes #939 

### Changes

- Added optional `isActive` function to the `Shortcut`.
- Refactored `Toolbar` to use `isActive` to properly fill shortcut's svg.

### DIFF
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -16,9 +16,8 @@ import { shortcutById } from '../shortcuts'
 import { isTouch } from '../browser'
 import { store } from '../store'
 import { overlayHide, overlayReveal, scrollPrioritize } from '../action-creators/toolbar'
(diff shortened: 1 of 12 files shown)


raineorshine (approved):
Perfectly executed! This is exactly what I had in mind. Thank you so much!!!

PR #961, opened 2020-12-31
Hide archived contexts from context view

### LINKED ISSUE #941: Do not show archived contexts in context view
## Steps to Reproduce

- a
  - m
- b
  - m

1. Archive `b`
2. Activate context view on `a/m`

## Current Behavior

The context view of `m` shows two contexts: `=archive/b` and `a` when `showHiddenThoughts` is false.

## Expected Behavior

The context view of `m` should only show `a` when `showHiddenThoughts` is false. It should show `=archive/b` when `showHiddenThoughts` is true.

shresthabijay:
fixes #941 

## Changes

- Added `getUnarchivedContextsSortedAndRanked` to filter out archived thoughts.
- Use `showHiddenThoughts` to select `getUnarchivedContextsSortedAndRanked`  and `getContextsSortedAndRanked` on context view.


raineorshine (changes requested):
This looks great, and solves the problem as presented! Unfortunately I believe I didn't generalize the requirements enough, which is my mistake. This should not be specific to archived thoughts; all descendants of hidden thoughts should be hidden (i.e. metaprogramming attributes for now and perhaps `=hidden` in the future). This would make more sense then only hiding archived thoughts.

After looking at the call sites of `getContextsSortedAndRanked`, I believe we should modify it to filter out hidden thoughts by default when `showHiddenThoughts` is false, similar to `getChildren`. `getContextsSortedAndRanked` is only used in places that are user-facing (hence the sorting), so it should be safe to assume that showing only visible thoughts is appropriate and won't break anything. A quick review of the call sites suggests this will work everywhere the function is currently being used, but let me know if you see anything that indicates otherwise.

If `getContextSortedAndRanked` handles hidden thoughts, then you shouldn't have to change the `Subthoughts` component which would be ideal.

Thanks!

shresthabijay:
> getContextsSortedAndRanked

@raineorshine  Agreed! We can simply filter out in `getContextSortedAndRanked`. I have made the changes

shresthabijay:
@raineorshine Ther  is a function called `isContextVisible` exported from `getChildren`. It doesn't exactly work for this use case, so I created a new selector `isContextVisible` and exported in a different name `isContextVisibleUser`. The new one first checks if `showHiddenThoughts`, it checks is it's direct ancestors are meta attributes, and in last it checks if the ancestors with `=hidden` attribute as one of their children.  Can you check if the new function supersedes the old one ?

src/selectors/isContextVisible.ts
```diff
@@ -0,0 +1,21 @@
+import { Context } from '../types'
+import { isFunction } from '../util'
+import { State } from '../util/initialState'
+import _ from 'lodash'
+import hasChild from './hasChild'
+
+/**
+ * Checks if the context is visible.
+ */
+const isContextVisible = _.curry((state: State, context: Context) => {
```
raineorshine, line 10: Let's call this `isAncestorsVisible` and rename the file to match.

src/selectors/getContextsSortedAndRanked.ts
```diff
@@ -1,12 +1,12 @@
...
-import { getContexts } from '../selectors'
+import { getContexts, isContextVisibleForUser } from '../selectors'
 import { State } from '../util/initialState'
 import { ThoughtContext } from '../types'
 
 /** Gets all contexts that the given thought is in, sorted and ranked. */
 const getContextsSortedAndRanked = (state: State, value: string): ThoughtContext[] =>
   sort(
-    getContexts(state, value),
+    getContexts(state, value).filter(({ context }) => isContextVisibleForUser(state, context.concat(value))),
```
raineorshine, line 9: When concatenating `Contexts` or `Paths` you always have to `unroot` the result since the ROOT token is only used in the ROOT context itself, not any of its descendants.

raineorshine (changes requested):

raineorshine:
> The new one first checks `showHiddenThoughts` then it checks is it's direct ancestors are meta attributes, and in last it checks if any ancestors have `=hidden` attribute. Can you check if the new function supersedes the old one ?

Yup, let's replace the existing `isContextVisible` with the new one. `isContextVisible` is only used in `prevSibling` which is used to determine the `cursor` after `cursorUp` and delete/move operations. The `cursor` should only be automatically moved to visible thoughts.

Issue #953, opened 2020-12-27
Cursor not exported


raineorshine:
## Steps to reproduce

- a
  - b
    - c

1. Set the cursor on `a`
1. Export `a`

## Current Behavior

`b` and `c` are exported correctly but the cursor thought itself is excluded. 

- b
  - c

## Expected Behavior

Cursor thought should be exported. 

- a
  - b
    - c

shresthabijay:
@raineorshine This behavior is introduced by this logic.

https://github.com/cybersemics/em/blob/edec0655310835a1fbcbcb375fd90c5e6b33d278/src/components/ModalExport.tsx#L73
Here `removeRoot` removes first thought from the `exported` if `titleChild` is not available. Shouldn't it check if the thought'a value is `__ROOT__` before removing ?

raineorshine:
Indeed! Let's add that condition in `removeRoot` itself, and if the first line is not the ROOT then return `exported` as-is. Thanks!

PR #963, opened 2021-01-03
Fix loading ellipsis on export modal

### LINKED ISSUE #948: Loading hangs when exporting leaf
## Steps to Reproduce

1. Create a thought
2. Export

## Current Behavior

Loading messages in export model hangs. 

## Expected Behavior

Loading message should go away after all descendants are loaded (which in this case are none).

shresthabijay:
fixes #948

## Changes
- Prevent falsey empty strings and use strict null check for `exportContent`.


raineorshine:
Thanks! That's better. It was hiding the underlying #953.

PR #968, opened 2021-01-04
On Subcategorize all: Child deleted fix

### LINKED ISSUE #962: Subcategorize All: Children deleted
## Steps to Reproduce

``` 
- a
  - b
    - =archive
      - b1
    - c
      - =archive
        - c1
      - d
``` 

1. Select `c`
2. Activate Subcategorize All

## Current Behavior

`c` and `d` disappear

## Expected Behavior

`a/b/=archive` and `a/b/c` should be moved to a new empty Thought in `a/b`.

shresthabijay:
fix #962

# Cause of Issue

`subCategrorizeAll` first adds a new thought and then moves all the children in the context to the new thought. However `getThoughtNew` logic may not always return the new thought. 

https://github.com/cybersemics/em/blob/edec0655310835a1fbcbcb375fd90c5e6b33d278/src/reducers/subCategorizeAll.ts#L47-L51

`getPrevRank` gets a rank that comes before all visible thoughts. So in a context with meta attributes with `showHIddenThoughts` set to false, the new thought added in the context would not be at the start of the sorted children array. In the above mentioned issue, `a.b.=archieved` is returned by `getThoughtNew` instead of newly added empty thought. So all the new thoughts were being moved to the `a.b.=archieved` context.

# Changes
`getPrevRank` has `aboveMeta` param that we can pass to allow rank to be calculated above hidden meta thoughts. Using that solved the issue.


raineorshine:
Tested and reviewed. Looks good!

Could you add a unit test before we merge? Thanks!

shresthabijay:
@raineorshine I have added the test. Thanks!

raineorshine:
Thanks! Is it possible to test the reducer directly? That would be preferable over running it through the store.

shresthabijay:
> Thanks! Is it possible to test the reducer directly? That would be preferable over running it through the store.

Thanks for pointing it out! I migrated it to existing reducers test.

raineorshine (changes requested):
Linter is failing. Otherwise looks good.

PR #974, opened 2021-01-07
Fix note focus

### LINKED ISSUE #969: Toggling new note multiple times does not move caret
Tested in Chrome

## Steps to Reproduce

1. Create thought `a`
2. Activate note (caret moves to note)
3. Activate note (caret moves back to thought)
4. Activate note (caret **should** move to note)

shresthabijay:
fixes #969

# Cause of Issue

`noteFocus` is not set to false after focus is set to the `ContentEditable`. This prevented setting focus back to note as `noteFocus` is already true.

# Changes

- Create `setNoteFocus` reducer and set focus back to `false` when needed on note shortcut.


raineorshine:
I'm only seeing the caret move after a second activation, i.e.:

1. Create thought `a`
2. Activate note (caret moves to note)
3. Activate note (caret moves back to thought)
4. Activate note (caret **should** move to note)
5. Activate note (caret moves back to note - one step too late)

shresthabijay:
@raineorshine I couldn't reproduce the issue you mentioned above. It is working fine for me.
Screen Capture_select-area_20210109101241

Which device are you using for the test ?

raineorshine:
Just figured out it works when the note is activated from the toolbar, but not when it is activated via the keyboard shortcut.

> Which device are you using for the test ?

Unless marked otherwise, all issues are reproduced from Chrome Desktop.

shresthabijay:
> Just figured out it works when the note is activated from the toolbar, but not when it is activated via the keyboard shortcut.
> 
> > Which device are you using for the test ?
> 
> Unless marked otherwise, all issues are reproduced from Chrome Desktop.

@raineorshine I can't reproduce it. It's working for me even via the keyboard shortcut.

raineorshine:
Okay, thanks for looking. Must be on my end. I'll look into it.

raineorshine:
I needed to add `setNoteFocus({ value: false })` to the Note component shortcut to get it to work for me for some reason.

PR #980, opened 2021-01-09
Fix getPrevRank

### LINKED ISSUE #970: Unhandled Rejection when activating new note after deleting subthought
## Steps to Reproduce

1. Create thought `a`
2. Create subthought `b`
3. Archive `b`
3. Activate new note

## Current Behavior

`Unhandled Rejection (TypeError): Cannot read property 'rank' of undefined`

> src/selectors/getPrevRank.ts:25

## Expected Behavior

New note is created without error.

shresthabijay:
fixes #970

# Cause of Issue

The logic of `getPrevRank` didn't handled case where all the children in the context are hidden.

# Changes

- Add specific logic to handle above mentioned case in `getPrevRank`.
- Add appropriate selector test for `getPrevRank`.

### DIFF
--- a/src/selectors/__tests__/getPrevRank.ts
+++ b/src/selectors/__tests__/getPrevRank.ts
@@ -1,20 +1,70 @@
-import { initialState, reducerFlow } from '../../util'
-import { getAllChildren } from '../../selectors'
-import { newSubthought, newThought, cursorBack } from '../../reducers'
+import { initialState, isFunction, reducerFlow } from '../../util'
+import { RANKED_ROOT } from '../../constants'
+import { getAllChildrenSorted } from '../../selectors'
+import { importText, newSubthought, newThought } from '../../reducers'
 import getPrevRank from '../getPrevRank'
 
-it('getPrevRank when all the thoughts in the context are hidden', () => {
+it('get rank above all children', () => {
 
   const steps = [
     newThought('a'),
-    newSubthought('=archive'),
-    cursorBack
+    newSubthought('b'),
+    newThought('c'),
   ]
 
   const stateNew = reducerFlow(steps)(initialState())
 
-  const children = getAllChildren(stateNew, ['a'])
+  const children = getAllChildrenSorted(stateNew, ['a'])
 
   expect(getPrevRank(stateNew, ['a'])).toBeLessThan(children[0].rank)
+
+})
+
+it('get rank less than visible children but greater than hidden children', () => {
+
+  const text = `
+    - a
+      - =archive
+      - b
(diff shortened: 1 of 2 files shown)


src/selectors/__tests__/getPrevRank.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  const steps = [
+    newThought('a'),
+    newSubthought('=archive'),
+    cursorBack
+  ]
+
+  const stateNew = reducerFlow(steps)(initialState())
+
+  const children = getAllChildren(stateNew, ['a'])
```
raineorshine, line 16: Use `getAllChildrenSorted`, since we want the first child to have the lowest rank.

src/selectors/getPrevRank.ts
```diff
@@ -16,8 +16,11 @@ const getPrevRank = (state: State, context: Context, { aboveMeta }: { aboveMeta?
     ? 0
     : children.findIndex(child => !isFunction(child.value))
 
+  // there could be no visible thoughts in the context
+  const noVisibleChildren = i === -1
+
   // between last metaprogramming attribute and first visible child
-  if (i === 0) return children[0].rank - 1
+  if (i === 0 || noVisibleChildren) return children[0].rank - 1
```
raineorshine, line 23: This is the correct behavior when `aboveMeta` is true. However if `aboveMeta` is false, the new rank should not be above hidden thought. I will add tests to clarify!

raineorshine (changes requested):

src/selectors/getPrevRank.ts
shresthabijay: Got it!

src/selectors/__tests__/getPrevRank.ts
```diff
@@ -1,20 +1,70 @@
...
-import { getAllChildren } from '../../selectors'
-import { newSubthought, newThought, cursorBack } from '../../reducers'
+import { initialState, isFunction, reducerFlow } from '../../util'
+import { RANKED_ROOT } from '../../constants'
+import { getAllChildrenSorted } from '../../selectors'
+import { importText, newSubthought, newThought } from '../../reducers'
 import getPrevRank from '../getPrevRank'
 
-it('getPrevRank when all the thoughts in the context are hidden', () => {
+it('get rank above all children', () => {
```
raineorshine: @shresthabijay FYI You can exclude the module name from the individual tests since it is automatically included in the Jest output.

raineorshine:
Looks good, thanks!

Issue #851, opened 2020-08-28
Command: join


raineorshine:
- [ ] Unit Test
 
Create a pure, functional reducer called `join` that joins all siblings into a single thought. Joined thoughts should be separated by a single space. Children should be merged.

Create a shortcut that dispatches `{ type: 'join' }`

Keyboard: <kbd>Meta</kbd> + <kbd>Option</kbd> + <kbd>j</kbd>

# Case 1

## Before

- a
  - b [cursor]
  - c
  - d

## After

- a
  - b c d [cursor]

# Case 2 (children)

## Before

- a
  - b [cursor]
    - c
  - d
    - e

## After

- a
  - b d [cursor]
    - c
    - e

pushkov-fedor:
I liked to work on splitSentences, i would like to pick this too! 7-9hrs

raineorshine:
I'm afraid I can't budget that much time for this. Given how similar it is to `splitSentences`, it should be very straightforward. 3-4 hours at most.

pushkov-fedor:
Okay, i still want to take it and try to implement it in your estimation. But if you want to assign it to someone, who is more confident in his estimation, it's okay :)

raineorshine:
I'd be happy to give it to you if you log no more than 4 hours, even if it takes you a bit more. I'll leave that up to you. I understand if you want to work strictly hourly.

anmolarora1:
@raineorshine Can we close this one?

PR #983, opened 2021-01-09
adds join command and related tests

anmolarora1:
#851


src/reducers/join.ts
```diff
@@ -0,0 +1,54 @@
...
+import existingThoughtDelete from './existingThoughtDelete'
+
+/** Join two or more thoughts split by spaces. */
+const join = (state: State) => {
+
+  const { cursor } = state
+
+  if (!cursor) return state
+
+  const contextRanked = pathToContext(parentOf(cursor))
```
raineorshine, line 17: Since `pathToContext` converts the `Path` to a `Context`, it's actually unranked. `context` would be a better name.

src/reducers/join.ts
```diff
@@ -0,0 +1,54 @@
...
+      const oldPath = getChildPath(state, child, pathToSibling as SimplePath)
+      const newPath = [...cursor, { ...child, rank: index === 1 ? getNextRank(state, pathToContext(pathToSibling)) : getPrevRank(state, pathToContext(pathToSibling)) }]
+
+      return existingThoughtMove({ oldPath, newPath })
+
+    })]
+
+  }, [] as _.RightCurriedFunction1<State, State>[])
+
+  const newValue = contextChildren.reduce((acc, { value }) => `${acc} ${value}`, '').trim()
```
raineorshine, line 37: `.map(child => child.value).join(' ')` would be a little more idiomatic but this is okay.

src/reducers/join.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const updateThoughtAction = existingThoughtChange({
+    oldValue: value,
+    newValue,
+    context: contextRanked,
+    path: cursor as SimplePath,
```
raineorshine, line 43: `cursor` has to be converted to a `SimplePath` here, as above.

src/reducers/join.ts
```diff
@@ -0,0 +1,54 @@
...
+
+      return existingThoughtMove({ oldPath, newPath })
+
+    })]
+
+  }, [] as _.RightCurriedFunction1<State, State>[])
+
+  const newValue = contextChildren.reduce((acc, { value }) => `${acc} ${value}`, '').trim()
+
+  const updateThoughtAction = existingThoughtChange({
```
raineorshine, line 39: These are actually reducers, not actions.

src/reducers/join.ts
```diff
@@ -0,0 +1,54 @@
...
+    oldValue: value,
+    newValue,
+    context: contextRanked,
+    path: cursor as SimplePath,
+  })
+
+  const removalActions = siblings.map(sibling => existingThoughtDelete({
+    context: contextRanked,
+    thoughtRanked: sibling
+  }))
```
raineorshine, line 49: Looks good!

src/reducers/join.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { value, rank } = head(cursor)
+  const siblings = contextChildren.filter(child => child.value !== value && child.rank !== rank)
+
+  const actions = siblings.reduce((acc, sibling, index) => {
+
+    const pathToSibling = [...parentOf(cursor), sibling]
+    const children = getAllChildren(state, pathToContext(pathToSibling))
+
+    return [...acc, ...children.map(child => {
+      const oldPath = getChildPath(state, child, pathToSibling as SimplePath)
```
raineorshine, line 28: `pathToSibling` has to be properly converted to a `SimplePath` here. Casting is only appropriate if the `Path` is guaranteed to be a `SimplePath`, such as concatenating a thought to a `SimplePath`. Since `pathToSibling` concatenates a thought to a `Path`, that guarantee doesn't exist here.

You can use `util/simplifyPath` for this purpose.

src/reducers/join.ts
```diff
@@ -0,0 +1,54 @@
...
+      const newPath = [...cursor, { ...child, rank: index === 1 ? getNextRank(state, pathToContext(pathToSibling)) : getPrevRank(state, pathToContext(pathToSibling)) }]
```
raineorshine, line 29: Since all the merged children are added to the end of the context, we should be able to call `getNextRank` once outside the loop and then increment by 1 for each subsequent sibling.

e.g.

- a
  - m
    - w (rank 0)
    - x (rank 1)
  - n
    - y (rank 0)
  - o
    - z (rank 0)

`getNextRank(state, [a', 'm'])` would return 1, and we could increment the rank of each merged child:

- a
  - m
    - w (rank 0)
    - x (rank 1)
    - y (rank 2)
    - z (rank 3)

src/reducers/__tests__/join.ts
```diff
@@ -0,0 +1,55 @@
...
+  const exported = exportContext(newState, [ROOT_TOKEN], 'text/plain')
+  const expectedOutput = `
+- a
+  - m n
+- b
+`
+  expect(removeRoot(exported)).toEqual(expectedOutput)
+})
+
+it('joins two thoughts and merges their children', () => {
```
raineorshine, line 29: Test looks great

src/reducers/__tests__/join.ts
```diff
@@ -0,0 +1,55 @@
...
+  ]
+
+  const newState = reducerFlow(steps)(initialState())
+  const exported = exportContext(newState, [ROOT_TOKEN], 'text/plain')
+  const expectedOutput = `
+- a
+  - m n o
+    - x
+    - y
+    - z
```
raineorshine, line 51: Let's test that the merged siblings do not have conflicting ranks.

src/reducers/join.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { cursor } = state
+
+  if (!cursor) return state
+
+  const contextRanked = pathToContext(parentOf(cursor))
+  const contextChildren = getAllChildren(state, contextRanked)
+  const { value, rank } = head(cursor)
+  const siblings = contextChildren.filter(child => child.value !== value && child.rank !== rank)
+
+  const actions = siblings.reduce((acc, sibling, index) => {
```
raineorshine, line 22: Good organization and use of `reducerFlow`!

raineorshine (changes requested):

raineorshine:
Looks good! Minor recommendations.

Let's add a shortcut as well. <kbd>Meta</kbd> + <kbd>Option</kbd> + <kbd>j</kbd>

src/reducers/join.ts
anmolarora1: I agree. Apparently, string concatenation is slightly faster than the join. So I opted for the former approach.

src/reducers/join.ts
anmolarora1: I couldn't understand why this conversion won't work. Can you please explain?

src/reducers/join.ts
anmolarora1: > Since all the merged children are added to the end of the context, we should be able to call `getNextRank` once outside the loop and then increment by 1 for each subsequent sibling.

I didn't know that the merged children were supposed to be added to the end of the context because it's not stated anywhere in the [issue description](https://github.com/cybersemics/em/issues/851#issue-688219266).
So my assumption was that we'd keep the order of children. For instance, running a join command on this structure

> * a
>   
>   * m
>     
>     * w (rank 0)
>     * x (rank 1)
>   * n (cursor)
>     
>     * y (rank 0)
>   * o
>     
>     * z (rank 0)

should result in

> * a
>   
>   * m n o (cursor)
>     * w
>     * x
>     * y
>     * z

If we append all the children to the end, however, we'd get something like this

> * a
>   
>   * m n o (cursor)
>     * y
>     * w
>     * x
>     * z


@raineorshine Please confirm which one is correct. And let's add the same to the issue description.

raineorshine: Sorry for the confusion! Your original interpretation is correct. By "end of context" I didn't mean append without changing their ranks. I was trying to refer to what you were already doing.

The suggestion is related to efficiency rather than functionality. Do you think that an autoincrement would work and allow us to avoid calling `getNextRank` multiple times?

src/reducers/__tests__/join.ts
anmolarora1: okay

src/reducers/join.ts
anmolarora1: I agree. In fact, the previous approach wasn't fully correct as it resulted in duplicate ranks. 
Though the `incrementing option` does the job, the code doesn't look very efficient now. So I'd like to hear your thoughts on that.

src/reducers/join.ts
anmolarora1: Thanks! I've updated it

src/reducers/__tests__/join.ts
```diff
@@ -0,0 +1,88 @@
...
+    importText({ path: RANKED_ROOT, text }),
+    setCursorFirstMatch(['a', 'n']),
+    join()
+  ]
+
+  const newState = reducerFlow(steps)(initialState())
+
+  const children = getChildrenRanked(newState, ['a', 'm n o'])
+
+  expect(new Set(children.map(child => child.rank)).size).toBe(4)
```
raineorshine, line 86: Nice! I wasn't aware of this trick.

src/reducers/join.ts
raineorshine: > Though the `incrementing option` does the job, the code doesn't look very efficient now. So I'd like to hear your thoughts on that.

It is actually more efficient than before since it doesn't call `getNextRank` every iteration. It can simplified though.

I think you can just use `nextRankMin + i`...?

I'm not understanding the purpose of `index < cursorIndex`. If you think it's still needed maybe you can help me understand.

raineorshine:
Thanks for the close attention to the recommend changes, all good. Let me know if you have any questions about the ranking logic!

src/reducers/join.ts
anmolarora1: I'm not sure about that. With the current approach, we're updating (incrementing or decrementing) the rank of every child **except the cursor head** (`y` in the below example). 
> * a
>   * m
>     * w (rank 0)
>     * x (rank 1)
>   * n (cursor)
>     * y (rank 0)
>   * o
>     * z (rank 0)

The result looks like this 
> * a
>   * m n o
>     * w (rank -2)
>     * x (rank -1)
>     * y (rank 0)
>     * z (rank 1)


The notion behind this approach was that we don't update the ranks of any thought in the current cursor as that might result in some side-effects. 

However, if you think otherwise, we can update the ranks of all merged children.
The result of that approach might look something like this -

> * a
>   * m n o
>     * w (rank 1)
>     * x (rank 2)
>     * y (rank 3)
>     * z (rank 4)

raineorshine: > The notion behind this approach was that we don't update the ranks of any thought in the current cursor as that might result in some side-effects.

I don't believe it's necessary. We can re-rank from the new first child, starting at `0`, with no side effects.

raineorshine:
I tightened up the code a bit in b97331d. Feel free to take a look! Everything looks good though.

raineorshine (approved):

PR #997, opened 2021-01-12
Confirm page reload if push is active

### LINKED ISSUE #965: No confirmation when closing during push
## Steps to Reproduce

1. Paste a large number of thoughts (2000+). [em-sample-large.txt](https://github.com/cybersemics/em/files/5762228/em-sample-large.txt)
2. Soon after the UI recovers and re-renders, hit refresh. (If you wait too long all the thoughts will have been saved and you won't reproduce the issue.)

## Current Behavior

Only some of the thoughts are persisted to IndexedDB.

## Expected Behavior

If a push is still in progress when the user tries to close or refresh the app, they should be presented with a browser confirmation dialog (via `onbeforeunload`).

shresthabijay:
fixes #965 

# Changes

- Add `beforeUnload` to confirm reload if `state.isPushing` is `true`. However custom messages is not supported by modern browsers. 
- Remove handler on `cleanup`.


raineorshine:
Perfect

PR #998, opened 2021-01-12
Get simplified context for attribute check in toggle sort

### LINKED ISSUE #977: Toolbar buttons not highlighted in nested view
Toolbar buttons for context attributes (e.g. pinSubthoughts, sort, table view) are not highlighted when the cursor is on a thought nested within a context view. 

## Steps to Reproduce

``` 
- a
  - m
    - =sort
      - Alphabetical
    - z
    - t
- b
  - m
``` 

1. Select `b/m`
1. Activate context view 
1. Select `b/m~/a`

## Current Behavior

Sort is not highlighted in the toolbar. 

## Expected Behavior

Sort should be highlighted in the toolbar.

shresthabijay:
fixes #977 

# Changes
- Use `simplifyPath` to get simplified context for attribute check in `toggleSort`.

### DIFF
--- a/src/shortcuts/note.ts
+++ b/src/shortcuts/note.ts
@@ -1,10 +1,10 @@
 import { isTouch } from '../browser'
-import { attribute, hasChild } from '../selectors'
+import { attribute, hasChild, simplifyPath } from '../selectors'
 import PencilIcon from '../components/icons/PencilIcon'
 import { asyncFocus, editableNode, isDocumentEditable, pathToContext, setSelection } from '../util'
 import { setAttribute } from '../action-creators'
 import { Shortcut } from '../types'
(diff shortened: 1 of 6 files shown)


shresthabijay:
@raineorshine I think we should make this changes to other shortcuts too.

raineorshine:
> @raineorshine I think we should make this change to other shortcuts too.

Yes! Could you do that in this PR for all other shortcuts with `isActive` defined?

raineorshine (changes requested):
> @raineorshine I think we should make this change to other shortcuts too.

Yes! Could you do that in this PR for all other shortcuts with `isActive` defined?

raineorshine (approved):

PR #1001, opened 2021-01-15
Make thought Annotation take space instead of Thought

### LINKED ISSUE #986: Url icon overlaps next thought when wrapped
## Steps to Reproduce

Adjust length of `aaa...` so that it wraps immediately before the url icon.

```
- a
  - aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    - https://stackoverflow.com/
  - b
```

## Current Behavior

The url icon overlaps the next thought when it is wrapped onto two lines.

image

## Expected Behavior

The next thought `b` should be rendered after the url icon. 

The height of `a` should be based on the ThoughtAnnotation, which contains the url icon, rather than the Thought.

shresthabijay:
fixes #986

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -433,13 +433,18 @@ a.text-link {
   white-space: pre-wrap;
 }
 
+.thought-container {
+  position: relative;
+}
+
+.thought {
+  position: absolute;
+}
+
 .thought-annotation {
-  position: absolute; /* cannot use position:absolute as it changes what the width is relative to */
   pointer-events: none;
-  /* offset parent container (.child) padding-left since position:absolute changes the width calculation; */
   box-sizing: border-box;
   width: auto;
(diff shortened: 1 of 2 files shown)


raineorshine:
The url icon looks good and is working as desired.

This change brings up bigger questions about the purpose of `ThoughtAnnotation`, which was intended to be a position absolute overlay for displaying non-interactive visual elements like underlines and superscripts. Yet the superscripts seem to work fine after this change. At the very least there is some careful renaming and refactoring work needed in order to bring the semantics in alignment with the functionality.

Given the additional work necessary, plus the additional testing needed of the new cursor overlay positioning across devices, browsers, and font sizes, I am going to put this on hold for now.

Issue #924, opened 2020-12-02
Command: Extract


raineorshine:
Create a new command that creates a new child with the selected text. 

e.g. 

- This is a thought

If the word "thought" is selected and Extract is activated, it should result in:

- This is a 
  - thought

## Requirements
- Extracted thought added to end of context
- Cursor doesn't change
- Activating with no selection results in an alert: "No text selected to extract"
- Keyboard: `Meta + Control + e`

PR #1002, opened 2021-01-16
adds extract thought command and tests (using RTL)

anmolarora1:
#924


package.json
```diff
@@ -30,6 +30,9 @@
   },
   "dependencies": {
     "@bit/mui-org.material-ui.swipeable-drawer": "^4.9.10",
+    "@testing-library/jest-dom": "^5.11.6",
+    "@testing-library/react": "^11.2.2",
+    "@testing-library/user-event": "^12.1.7",
```
raineorshine, line 35: I believe these should be `devDependencies`

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  it('the selected part of a thought is extracted as a child thought', async () => {
+
+    store.dispatch([
+      newThought({ value: 'this is a thought' }),
+      setCursorFirstMatchActionCreator(['this is a thought'])
+    ])
+
+    const thought = screen.queryAllByText('this is a thought')
+    expect(thought).toHaveLength(2)
```
raineorshine, line 37: I wonder why this returns two instances. Is it the `Thought` and the `ThoughtAnnotation`? If so, should probably just select what is visible (`Thought`), though I am open to other ideas. Selecting a visible and a hidden instance and assuming the first one is the visible one feels like too much implementation detail leaking into the tests.

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,69 @@
...
+    const thoughtWrapper = thought[0]
+
+    const selectedText = setSelection(thoughtWrapper, 10, 17)
+    store.dispatch([extractThought()])
+
+    const updatedThought = await screen.findAllByText(thoughtWrapper.textContent!.slice(0, 9), { exact: true })
+    expect(updatedThought).toHaveLength(2)
+
+    const createdThought = await screen.findAllByText(selectedText, { exact: true })
+    expect(createdThought).toHaveLength(2)
```
raineorshine, line 47: This tests that the thoughts have been updated and are present on the screen, but does not test the relationship between them. We need to test that the extracted text became the child, and that it is added to the end of the context if there are other children.

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,69 @@
...
+    const selectedText = setSelection(thoughtEl, 10, 22)
+    store.dispatch([extractThought()])
+
+    const createdThought = await screen.findAllByText(selectedText, { exact: true })
+    expect(createdThought).toHaveLength(2)
+
+    expect(store.getState().cursor).toMatchObject([{ value: thoughtEl.textContent!.slice(0, 9) }])
+
+  })
+})
```
raineorshine, line 69: One more requirement to test:

- Activating with no selection results in an alert: "No text selected to extract"

src/reducers/extractThought.ts
```diff
@@ -0,0 +1,42 @@
+import _ from 'lodash'
+import { State } from '../util/initialState'
+import { headRank, headValue, pathToContext, reducerFlow, rootedParentOf } from '../util'
+import existingThoughtChange from './existingThoughtChange'
+import newThought from './newThought'
+import { simplifyPath } from '../selectors'
+
+/** Extract the selection as child thought. */
+const extractThought = (state: State) => {
```
raineorshine, line 9: Nicely done! Good implementation.

raineorshine (changes requested):
Looks great! I see this is still a draft, so you may already be aware of some of the points I brought up, but I reviewed anyway so that you had some feedback. Implementation looks great.

src/shortcuts/__tests__/extractThought.ts
anmolarora1: I couldn't find a query that ignores hidden elements, so I've used a matching element selector.

src/shortcuts/__tests__/extractThought.ts
anmolarora1: We'll need to use DOM selectors for this. I've added a test. Please go through.

src/reducers/extractThought.ts
anmolarora1: Thanks!

src/test-helpers/createRtlTestApp.tsx
```diff
@@ -0,0 +1,106 @@
...
+    ])
+
+    jest.runOnlyPendingTimers()
+
+    // make DND ref available for drag and drop tests.
+    document.DND = dndRef.current
+  })
+
+  // since act cannot return anything, we have to wait for wrapper to be set and then convince Typescript that it is not being used before it is assigned
+  return wrapper as unknown as ReactWrapper<unknown, unknown>
```
raineorshine, line 75: ~`ReactWrapper` is an Enzyme type. This should be typed as the return type of RTL's `render` function.~

Since RTL's queries are global, we may not need to return `wrapper` at all.

src/test-helpers/createRtlTestApp.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { store } from '../store'
+import * as db from '../data-providers/dexie'
+import { clear } from '../action-creators'
+import { Await } from '../types'
+
+// components
+import AppComponent from '../components/AppComponent'
+import ErrorBoundaryContainer from '../components/ErrorBoundaryContainer'
+import TouchMonitor from '../components/TouchMonitor'
+import { render } from '@testing-library/react'
```
raineorshine, line 17: Let's place the `render` import in the top section of imports. This section is just for components.

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,96 @@
...
+    expect(thought).toBeTruthy()
+
+    const selectedText = setSelection(thought, 10, 17)
+    store.dispatch([extractThought()])
+
+    const updatedThought = await findAllByText(thought, thought.textContent!.slice(0, 9), { exact: true })
+    expect(updatedThought).toHaveLength(1)
+
+    const createdThought = await findAllByText(content, selectedText, { exact: true })
+    expect(createdThought).toHaveLength(2)
```
raineorshine, line 66: We need a query that returns one thought rather than two here in order to more closely match what the user sees. Whatever it is returning two for is an implementation detail that should not be tested against.

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    // created thought gets appended to the end
+    const thoughtChildrenWrapper = thought.closest('li')?.lastElementChild as HTMLElement
+    const thoughtChildren = await findAllByPlaceholderText(thoughtChildrenWrapper, 'Add a thought')
```
raineorshine, line 70: Nice! I was wondering how to best query thoughts, and I like `findAllByPlaceholderText` in this situation.

src/components/Content.tsx
```diff
@@ -100,7 +100,7 @@ const Content: ContentComponent = props => {
     publish: publishMode(),
   }), [tutorialStep, isTutorialLocal])
 
-  return <div id='content-wrapper' onClick={e => {
+  return <div id='content-wrapper' data-testid='content-wrapper-test' onClick={e => {
```
raineorshine, line 103: Could you help me understand the purpose of this a bit more? I see how it's being used but I don't totally understand the problem that it solves. What happens when only `screen` is used, and why?

raineorshine (changes requested):
Looking good!

I'd like to fix the queries that return two thoughts instead of the expected one. I understand that you've already worked on that problem, so just let me know if you feel like you've done everything you can and I can take a stab at it myself. Perhaps there is no way to avoid it. Thanks!

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,91 @@
...
+      setCursorFirstMatchActionCreator(['this is a thought'])
+    ])
+
+    store.dispatch([extractThought()])
+
+    const alert = await screen.findByText('No text selected to extract')
+    expect(alert).toBeTruthy()
+
+    const [thought] = await screen.findAllByText('this is a thought', { selector: 'div' })
+    expect(thought).toBeTruthy()
```
raineorshine, line 42: We're getting closer. Extracting the first item in the array is not totally clear to me. Does `findAllByText` return more than one item? Why? What does `selector: `div` do in this case? 

The reason I am spending more time on this is that it still feels like there are implementation details mixed in with the test. If there are no alternatives then we can factor out the specifics into a custom `findThought` query. That will decouple the individual tests from the implementation details.

raineorshine (changes requested):

src/shortcuts/__tests__/extractThought.ts
anmolarora1: Agreed. We need to abstract all the implementation details that we can, and I guess one way to do that is to create custom query utilities, as you mentioned.

Regarding your questions,

`findAllByText` returns an array by default as the name implies. We can use `findByText` to get rid of that.

The  `selector: div` is needed because there's no direct query to find only the visible elements, and since each thought comprises two elements - `subthought-text` (a hidden span) and an `editable` (the visible div), we need to be specific with our queries to get the correct(visible) element and avoid any array logic.

raineorshine: > `findAllByText` returns an array by default as the name implies. We can use `findByText` to get rid of that.

Great! That's better.

> The `selector: div` is needed because there's no direct query to find only the visible elements, and since each thought comprises two elements - `subthought-text` (a hidden span) and an `editable` (the visible div), we need to be specific with our queries to get the correct(visible) element and avoid any array logic.

Ah, I see. Thank you for the explanation. I think selecting on something more semantic than the element type may be more transparent and understandable for other devs. It seems that the key difference is that the editable thought has a `content-editable` attribute. We should be able to create a custom query that is a variant of `findByText`. I would suggest calling `findByText` and then filtering its results by attribute rather than the other way around for efficiency. Let me know what you think of this approach.

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  it('an alert should be shown if there is no selection', async () => {
+    const thoughtValue = 'this is a thought'
+    store.dispatch([
+      newThought({ value: thoughtValue }),
+      newThought({ value: 'sub-thought', insertNewSubthought: true }),
+      setCursorFirstMatchActionCreator([thoughtValue])
+    ])
+
+    store.dispatch([extractThought()])
```
raineorshine, line 45: This should be included in the previous `store.dispatch`

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    const alert = await screen.findByText('No text selected to extract')
+    expect(alert).toBeTruthy()
+
+    const thought = await findThoughtByText(thoughtValue)
+    expect(thought).toBeTruthy()
```
raineorshine, line 51: This is testing `newThought` and is not relevant for this test.

src/shortcuts/__tests__/extractThought.ts
```diff
@@ -0,0 +1,101 @@
...
+  range.setStart(element.childNodes[0], selectionStart)
+  range.setEnd(element.childNodes[0], selectionEnd)
+
+  sel?.removeAllRanges()
+  sel?.addRange(range)
+
+  return range.toString()
+}
+
+describe('Extract thought', () => {
```
raineorshine, line 31: The `describe` block is mainly useful to group tests within a test module. Since there is only one `describe` block here, you can omit it. The filename will show up in the test output and describes the same information.

raineorshine (approved):

PR #1010, opened 2021-01-20
Recently Edited: Formatting and filtering

### LINKED ISSUE #959: Recently Edited: formatting and filtering
There are some small improvements we can make before the next major upgrade:

- Exclude EM_TOKEN and metaprogrammimg attributes
- Format breadcrumbs like context view

shresthabijay:
fixes #959 

# Changes

- Hide recently edited meta thought and ROOT thoughts when `showHiddenThoughts` is `true`.
- Add test for recenty edited `findTreeDescendants` for above mentioned functionaity.
- Move optional overflow collapsing and character limiting to `ContextBreadcrumbs` and use it in `NavBar` and `Sidebar` too.
- Create `RecentlyEditeBreadcrumbs`  a variant of `ContextBreadcrumbs` for recently edited thoughts sidebar.


shresthabijay:
@raineorshine  Please review the changes. Thank you!

src/App.css
```diff
@@ -765,7 +770,7 @@ body.dark {
   vertical-align: bottom;
 }
 
-.nav-breadcrumbs.nonempty {
+.nav-breadcrumbs. pty {
```
raineorshine, line 773: This appears to be a typo

src/components/ContextBreadcrumbs.tsx
```diff
@@ -43,16 +83,43 @@ const ContextBreadcrumbs = ({ homeContext, simplePath }: ContextBreadcrumbProps
...
+
+  const parentSimplePath = props.simplePath.slice(0, -1) as SimplePath
+  const simplePath = props.simplePath
+
+  return <div className='recently-edited-breadcrumbs'>
+    <ContextBreadcrumbs {...props} simplePath={parentSimplePath}/>
+    <Link simplePath={simplePath} label={headValue(simplePath)} />
+    <Superscript simplePath={simplePath} />
+  </div>
+})
```
raineorshine, line 123: Let's stick to our convention of one exported component per file.

src/components/ContextBreadcrumbs.tsx
```diff
@@ -43,16 +83,43 @@ const ContextBreadcrumbs = ({ homeContext, simplePath }: ContextBreadcrumbProps
...
     }
   </div>
 }
 
+/**
+ * Varaint of ContextBreadcrumbs for recently edited with collapsing overflow.
+ */
+export const RecentlyEditedBreadcrumbs = connect(mapStateToProps)((props: ContextBreadcrumbProps & ReturnType<typeof mapStateToProps>) => {
+
+  const parentSimplePath = props.simplePath.slice(0, -1) as SimplePath
```
raineorshine, line 115: Use `parentOf`

src/components/Sidebar.tsx
```diff
@@ -3,35 +3,28 @@ import SwipeableDrawer, { SwipeableDrawerProps } from '@bit/mui-org.material-ui.
...
+  const showHiddenThoughts = useSelector((state: State) => state.showHiddenThoughts)
+  const recentlyEdited = _.reverse(_.sortBy(findTreeDescendants(recentlyEditedTree, { startingPath: [], showHiddenThoughts }), 'lastUpdated')) // eslint-disable-line fp/no-mutating-methods
 
   return (
     <div className='recently-edited-sidebar'>
       <div className='header'>Recently Edited Thoughts</div>
       <div style={{ padding: '0 2em' }}>
         {
-          recentlyEdited.map((recentlyEditedThought, i) => <ThoughtsTab path={recentlyEditedThought.path} key={i} />)
+          recentlyEdited.map((recentlyEditedThought, i) => <RecentlyEditedBreadcrumbs key={i} path={recentlyEditedThought.path as Path} charLimit={32} thoughtsLimit={10} />)
```
raineorshine, line 27: Why the cast?

raineorshine (changes requested):
This looks great! The new design feels much cleaner for deeply nested thoughts. New organization looks great and thanks for adding a test. Minor changes requested.

Could you also exclude thoughts that are within the `[EM_TOKEN]` context when `showHiddenThoughts` is false? Thanks!

src/components/ContextBreadcrumbs.tsx
shresthabijay: Since `RecentlyEditedBreadcrumbs` uses `mapStateToProps` to calculate `simplePath`, `ContextBreadcrumbs` doesn't need to calculate simplified path again. Defining `RecentlyEditedBreadcrumbs` in this file, we can use `ContextBreadcrumbs` that is not connected to redux.

I will create a separate file for `RecentlyEditedBreadcrumbs`. However same `mapStateToProps` will be exeuted twice. Once for both connected components.

src/components/ContextBreadcrumbs.tsx
shresthabijay: Sure

shresthabijay:

src/components/ContextBreadcrumbs.tsx
raineorshine: That's a fair point. Another way to achieve that is to export the unconnected `ContextBreadcrumbs` so that it can be imported "laterally" by `RecentlyEditedBreadcrumbs`. That allows you to avoid the extra `mapStateToProps` while allowing `RecentlyEditedBreadcrumbs` to be a default import.

raineorshine:
Looks good! Unless I am mistaken, I believe there's just one more thing:

> Could you also exclude thoughts that are within the `[EM_TOKEN]` context when `showHiddenThoughts` is false? Thanks!

shresthabijay:
> Could you also exclude thoughts that are within the `[EM_TOKEN]` context when `showHiddenThoughts` is false? Thanks!

I missed that part. Thank you for reminding. I just pushed making the changes.

src/util/isEM.ts
```diff
@@ -1,8 +1,11 @@
 import { EM_TOKEN } from '../constants'
-import { Path } from '../types'
+import { Child } from '../types'
 
 /** Returns true if the Path is the EM_TOKEN. */
-export const isEM = (path: Path): boolean =>
-  path.length === 1 &&
-  !!path[0] &&
-  path[0].value === EM_TOKEN
+export const isEM = (thoughts: (string | Child)[]): boolean =>
```
raineorshine, line 5: Slightly cleaner type: `thoughts: Context | Path`

src/util/isEM.ts
```diff
@@ -1,8 +1,11 @@
...
+  thoughts.length === 1 &&
+  !!thoughts[0] &&
```
raineorshine, line 7: Since we already know `thought.length === 1`, then we don't need to check `!!thoughts[0]`. An empty thought will be handled by the subsequent case.

src/util/isEM.ts
```diff
@@ -1,8 +1,11 @@
...
+  (
+    (thoughts[0] as Child).value === EM_TOKEN ||
+    (thoughts[0] as string) === EM_TOKEN
```
raineorshine, line 10: This works on strings since the `value` property will return `undefined`, but it would be better to explicitly check `typeof thoughts[0]`. You could also use a function like this:

```js
/** Checks if an object is of type Path. */
const isPath = (o: Context | Path): o is Path =>
  o.length > 0 && Object.prototype.hasOwnProperty.call(o[0], 'value')
```

raineorshine (changes requested):

src/util/isEM.ts
shresthabijay: Thanks!  I didn't know about user defined type guards.

raineorshine (approved):

PR #1012, opened 2021-01-24
converts subsetThoughts to descendantThoughts

### LINKED ISSUE #231: Do not expand matching contexts
When a thought is expanded within a context view, it incorrectly expands the matching thought outside the context view.

Expected behavior: Only the cursor path should be expanded.

See Video Capture: https://youtu.be/3Xt_MAGmNBQ

anmolarora1:
Fixes #231


src/util/descendantThoughts.ts
```diff
@@ -2,11 +2,11 @@ import { equalPath } from './equalPath'
 import { Path } from '../types'
 
 /** Returns true if thoughts subset is contained within superset (inclusive). */
-export const subsetThoughts = (superset: Path | null, subset: Path | null) => {
+export const descendantThoughts = (superset: Path | null, subset: Path | null) => {
```
raineorshine, line 5: Let's call this `isDescendantPath`. The `is` prefix will help communicate that it is a predicate.

src/util/__tests__/descendantThoughts.ts
```diff
@@ -0,0 +1,32 @@
...
+  expect(descendantThoughts([{ value: 'a', rank: 0 }], [{ value: 'a', rank: 0 }])).toBe(true)
+})
+
+it('descendant returns true', () => {
+  expect(descendantThoughts([{ value: 'a', rank: 0 }, { value: 'b', rank: 0 }], [{ value: 'a', rank: 0 }])).toBe(true)
+  expect(descendantThoughts([{ value: 'a', rank: 0 }, { value: 'b', rank: 0 }, { value: 'c', rank: 0 }], [{ value: 'a', rank: 0 }, { value: 'b', rank: 0 }])).toBe(true)
+})
+
+it('subset returns false', () => {
+  expect(descendantThoughts([{ value: 'a', rank: 0 }, { value: 'b', rank: 0 }], [{ value: 'a', rank: 0 }])).toBe(true)
```
raineorshine, line 13: You can remove the first `expect` from this test since it is part of the "descendant returns true" test.

raineorshine (changes requested):
Looks good!

I'd like to have an integration test as well to test the expansion behavior directly. We should be able to mount the test case and then use RTL to test that the correct contexts are expanded.

src/components/__tests__/ContextView.ts
```diff
@@ -0,0 +1,29 @@
+import { cursorDown, newThought } from '../../action-creators'
+import { store } from '../../store'
+import createTestApp, { cleanupTestApp } from '../../test-helpers/createRtlTestApp'
+import { findThoughtsByText } from '../../test-helpers/queries'
+import { setCursorFirstMatchActionCreator } from '../../test-helpers/setCursorFirstMatch'
+
+describe('Extract thought', () => {
+  beforeEach(async () => {
+    await createTestApp()
+  })
```
raineorshine, line 10: This can be simplified to `beforeEach(createTestApp)` since the return value is not being used.

src/test-helpers/queries.ts
```diff
@@ -0,0 +1,17 @@
+import { findAllByText, screen } from '@testing-library/dom'
+
+/**
+ * Gets a thought by value.
+ */
+export const findThoughtByText = async (value: string, container?: HTMLElement) => {
+  const thoughtNodes = await (container ? findAllByText(container!, value, { exact: true }) : screen.findAllByText(value, { exact: true }))
+  return thoughtNodes.find(t => t.hasAttribute('contenteditable'))
+}
```
raineorshine, line 9: I prefer separate modules for each exported function. Having the module name be the same as the function name makes it easier to know where functions live, and it also make for very fast access using your editor's quick file find feature.

src/test-helpers/queries.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Gets a thought by value.
```
raineorshine, line 12: Make sure comments are in alignment with the code, as it can represent a point of confusion for other developers if it doesn't make sense.

src/test-helpers/queries.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+export const findThoughtsByText = async (value: string, container?: HTMLElement) => {
```
raineorshine, line 14: Let's call this `findAllThoughtsByText` to follow the RTL naming convention.

raineorshine (changes requested):

src/test-helpers/queries/findThoughtByText.ts
```diff
@@ -0,0 +1,9 @@
+import { findAllByText, screen } from '@testing-library/dom'
+
+/**
+ * Gets a thought that matches the given value.
+ */
+export const findThoughtByText = async (value: string, container?: HTMLElement) => {
+  const thoughtNodes = await (container ? findAllByText(container!, value, { exact: true }) : screen.findAllByText(value, { exact: true }))
```
raineorshine, line 7: I recommend splitting long ternaries into multiple lines for better readability.

raineorshine (approved):
Looks good! Thanks!

PR #1014, opened 2021-01-31
Fix hash match descendant disappearing

### LINKED ISSUE #862: existingThoughtMove: Hash matched descendants disappear
## Steps to Reproduce

- a [cursor]
  - =note
    - note

Activate `subcategorizeOne` or `subcategorizeAll`.

## Current Behavior

- [empty]
  - a
    - Note

## Expected Behavior

- [empty]
  - a [cursor]
    - =note
      - note

shresthabijay:
fixes #862


src/reducers/existingThoughtMove.ts
```diff
@@ -32,6 +32,7 @@ type RecursiveMoveResult = {
   archived?: Timestamp,
   contextsNew: Context[],
   contextsOld: Context[],
+  hashedKey: string,
```
raineorshine, line 35: Let's call this `thoughtEncoded` to be consistent with similar variables used in other part of the codebase.

src/reducers/existingThoughtMove.ts
```diff
@@ -169,6 +170,7 @@ const existingThoughtMove = (state: State, { oldPath, newPath, offset }: {
         // Order matters: accum must have precendence over accumRecursive so that contextNew is correct
         // merge current thought update
         value: child.value,
+        hashedKey: hashedKey,
```
raineorshine, line 173: Use shorter syntax: `hashedKey,` (or after the rename, `thoughtEncoded`)

src/reducers/existingThoughtMove.ts
```diff
@@ -210,9 +212,9 @@ const existingThoughtMove = (state: State, { oldPath, newPath, offset }: {
         const accumInnerChildrenOld = accumInner.contextIndex[contextEncodedOld]?.children
         const accumInnerChildrenNew = accumInner.contextIndex[contextEncodedNew]?.children
         const childrenOld = (accumInnerChildrenOld || getAllChildren(state, contextOld))
-          .filter((child: Child) => child.value !== result.value)
+          .filter((child: Child) => hashThought(child.value) !== result.hashedKey)
```
raineorshine, line 215: I believe it would be more efficient to use `thoughtTransform` rather than `hashThought`, and result in the same comparison.

src/util/hashThought.ts
```diff
@@ -44,12 +19,7 @@ const singularize = (s: string) => s !== 's' ? pluralize.singular(s) : s
  */
 export const hashThought: (s: string) => ThoughtHash = _.memoize(
   (value: string) => _.flow([
-    // placed before stripEmojiWithText because stripEmojiWithText partially removes angle brackets
-    stripTags,
-    lower,
-    trim,
-    stripEmojiWithText,
-    singularize,
+    thoughtTransform,
```
raineorshine, line 22: Great. I've been wanting this!

src/util/thoughtTransform.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Tranformations for thought value:
+ *
+ * - case-insensitive
+ * - ignore punctuation & whitespace (when there is other text)
+ * - ignore emojis (when there is other text)
+ * - singularize.
+ */
+export const thoughtTransform = _.flow([
```
raineorshine, line 37: Let's call this `normalizeThought`

raineorshine (changes requested):

shresthabijay:
# Description

This PR introduces some major changes to `exisitingThoughtMove`.

### LexemeUpdate
It keeps tracks of updated `newThought` i.e lexeme. It also stores contextNew and contextOld associated with each lexeme.

### ChildUpdate
It keeps tracks of descendant child that have been updated. It has pathNew and pathOld and most importantly it has actual value, id, rank to the thought.

### RecurisveMoveResult
It is object with two attributes `lexemeUpdates` and `childUpdates`. It is returned by `recursiveUpdates`.

### lexemeUpdates
It is an index with key (hashed normalized thought) and value(it's associated LexemeUpdate). Since there can be multiple descendants with same normalized value, so recursively LexemeUpdate for each hashKey is updated with latest newThought.

### childUpdates
It is an index with key (hashed context of oldPath)and value (ChildUpdate). It stores information for all updated child recurively. This is used to get newPath of the child from the oldContext. For example we can easily get newPath from the context of the stale cursor and use it to create updated cursor.


# Change Log

- `oldSimplePath`  and `updatedNewSimplePath` is used instead of ``oldPath` and `newPath` to initiate recursiveUpdates function call. 
- Use `childUpdates` to find updated child exact value, id and ranks. Use it to create new children for `contextIndex`.
- Use `childUpdates` to find updated cursor.
- Change `transfromThought` to `normalizeThought`.
- Fix all failing test introduced by the side effect of new changes.

src/reducers/__tests__/existingThoughtMove.ts
```diff
@@ -458,23 +457,13 @@ it('move with duplicate descendant', () => {
...
-      - x
-      - y
-        - x`)
-
-  // x should have contexts a/b and a/b/y
-  expect(getContexts(stateNew, 'x'))
-    .toMatchObject([
-      { context: ['a', 'b'], rank: 0 },
-      { context: ['a', 'b', 'y'], rank: 0 }
-    ])
```
raineorshine, line 478: You'll want to test `getContexts` as well, since `exportContext` is only touching `contextIndex` while `getThought` or `getContexts` will touch `thoughtIndex`.

src/reducers/existingThoughtMove.ts
```diff
@@ -28,19 +28,28 @@ import {
   timestamp,
 } from '../util'
 
-type RecursiveMoveResult = {
-  archived?: Timestamp,
+type LexemeUpdate = {
+  value: string,
+  newThought: Lexeme,
```
raineorshine, line 33: Can we eliminate `value` and just use `newThought.value`?

src/reducers/existingThoughtMove.ts
```diff
@@ -202,24 +258,27 @@ const existingThoughtMove = (state: State, { oldPath, newPath, offset }: {
...
         const contextEncodedOld = hashContext(contextOld)
         const contextEncodedNew = hashContext(contextNew)
+
+        const childUpdate = childUpdates[hashContext(contextOld.concat(result.value))]
+
         const accumInnerChildrenOld = accumInner.contextIndex[contextEncodedOld]?.children
         const accumInnerChildrenNew = accumInner.contextIndex[contextEncodedNew]?.children
         const childrenOld = (accumInnerChildrenOld || getAllChildren(state, contextOld))
-          .filter((child: Child) => child.value !== result.value)
+          .filter((child: Child) => normalizeThought(child.value) !== normalizeThought(result.value))
```
raineorshine, line 272: You can normalize `result.value` as soon as `result` is available (L261) for better performance. Otherwise it's called unnecessarily in both `result.contextsOld.reduce` and `.filter`.

src/reducers/existingThoughtMove.ts
```diff
@@ -292,29 +351,18 @@ const existingThoughtMove = (state: State, { oldPath, newPath, offset }: {
...
-      } as Child
-    }
-  )
-
-  // if duplicate subthoughts are merged then update rank of thoughts of cursor descendants
-  const cursorDescendantPath = (isPathInCursor && isDuplicateMerge ? updateMergedThoughtsRank : ID)(state.cursor || []).slice(oldPath.length)
+  /** Returns new path for the given old context.
+   * Note: This uses childUpdates so it works only for updated descendants. For main ancestor oldPath that has been moved (ancestor) use updatedNewPath instead.
+   */
+  const getNewPathFromOldContext = (path: Path) => childUpdates[hashContext(pathToContext(path))].pathNew
```
raineorshine, line 357: I think this should take a `SimplePath` rather than a `Path`. `Contexts` are contiguous parent-child sequences and are not allowed to cross context views like `Paths` can. Unfortunately we lose our branded type safety when we call `pathToContext` (which should probably only take `SimplePath` also). You can call `simplifyPath` on `state.cursor` before passing it in.

raineorshine (changes requested):
Wow, Bijay, excellent work on a very complicated part of the codebase! Really impressed. Thanks for digging in and going the extra kilometer! Tests are passing so we should be good to go after a couple minor changes.

src/reducers/__tests__/existingThoughtMove.ts
shresthabijay: Sure. That makes sense.

raineorshine (approved):

raineorshine:
Yay! I'm so glad to have this fixed. I can actually move around my thoughts without worrying about them disappearing now.

Issue #1021, opened 2021-02-07
Children multiply inside a context after thought is edited.


shresthabijay:
# Steps to Reproduce

- Create the following structure.novalue

```
- a
 - b
```

- Change `a` to `as`
- Change `as` back to `a`

# Current Behavior
```
 - a
  - b
  - b
```

# Expected Behaviour

```
- a
  - b
```

PR #1025, opened 2021-02-08
Merge import descendants and fix duplicate children

### LINKED ISSUE #993: Import: Merge descendants
## Steps to Reproduce

1. Create the following thoughts:
    - a
      - b
        - c
    - [empty]
2. Set the cursor on the empty thought
3. Paste the following:
    ```
    - a
      - b
        - q
      - x
        - y
    - j
    ```

## Current Behavior

- a
  - b
    - c
- a
  - b
    - q
  - x
    - y
- j

## Expected Behavior

Imported thought and its descendants should be merged with existing thoughts so that there are no duplicates.

- a
  - b
    - c
    - q
  - x
    - y
- j

## Proposed Solution

The desired behavior is achieved with the following:

1. Insert dummy thought `__IMPORT__`
2. Paste
3. Set cursor back on `__IMPORT__`
4. Activate `collapseContext`

I recommend reproducing this in a reducer flow in `importText` to avoid having to modify the import logic itself. Rename the existing function to `importTextIntoLeaf` and throw an error if the destination `path` has any children. Create a new `importText` function to be exported with the same signature as the old one and perform the reducer flow described in the above steps, calling `importTextIntoLeaf` for Step 2.

Let me know if you have any concerns about this approach!

shresthabijay:
fixes #993 and #1021 

# Changes

- Prevent children duplication on `exisitingThoughtChange`.
- Use dummy thought and `collapseContext` to merge imported thoughts.
- Add test for import and merge descendants.
- Add options to delete cursor instead of archive for `collapseContext` reducer.
- Add id for `contextIndexUpdates` in `exisitingThoughtMove` and `exisitingThoughtDelete`.
- Fix faulty `deleteEmptyThought` "merging siblings, caret should be in between" test.
- Use `getNextRank` instead of `getRankAfter` for `rankStart` inside `importJson` because if destination path is not empty then we need to import thoughts inside it.
- Fix all necessary tests and side effects.


src/reducers/importText.ts
```diff
@@ -157,22 +160,55 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
...
-          offset: headValue(imported.lastImported).length
-        })
-        : null,
+    const destEmpty = destThought.value === '' && getAllChildren(state, pathToContext(simplePath)).length === 0
 
-    ])(state)
+    // New dummy thoughts for collapsing
+    const newDummyThought = getAllChildren(updatedState, pathToContext(simplePath)).find(child => child.value === uuid)
+
+    const newDestinationPath = unroot(newDummyThought ? simplePath.concat(newDummyThought) : simplePath)
```
raineorshine, line 180: It's preferable to cast `newDestinationPath` to `SimplePath` as soon as it is assigned, as it is always a `SimplePath`.

raineorshine: I suggest `[...simplePath, newDummyThought]` instead of `simplePath.concat(newDummyThought)`. It's clean and slightly faster. While I'm sure the performance is not relevant here, I think it's a good habit to get into since `Array.prototype.concat` doesn't offer any advantages based on my research.

src/reducers/collapseContext.ts
```diff
@@ -1,15 +1,23 @@
...
+import { head, parentOf, pathToContext, reducerFlow, unroot } from '../util'
 import { State } from '../util/initialState'
-import { getChildren } from '../selectors'
+import { getAllChildren, simplifyPath } from '../selectors'
 import { archiveThought, existingThoughtMove, setCursor } from '../reducers'
+import _ from 'lodash'
+import existingThoughtDelete from './existingThoughtDelete'
+
+interface Options {
+  preventCursorArchive?: boolean,
```
raineorshine, line 9: Let's call this option `deleteCursor`. "prevent" suggests that it's prevented altogether.

src/reducers/collapseContext.ts
```diff
@@ -19,7 +27,13 @@ const collapseContext = (state: State) => {
           newPath: unroot([...parentOf(cursor), child]),
         })
       ),
-      archiveThought({ path: cursor }),
+      !preventCursorArchive
+        ? archiveThought({ path: cursor }) :
+        existingThoughtDelete({
+          context: parentOf(pathToContext(cursor)),
```
raineorshine, line 33: Also here: `pathToContext(simpleCursor)`.

It's not important, but it would be a bit cleaner to store it in a `context` variable to avoid calling `pathToContext` multiple times.

src/reducers/collapseContext.ts
```diff
@@ -19,7 +27,13 @@ const collapseContext = (state: State) => {
...
+          showContexts: false,
```
raineorshine, line 34: You can omit `showContexts` instead of explicitly setting it to false since it is an optional variable.

src/reducers/importText.ts
```diff
@@ -157,22 +160,55 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
...
-        ? setCursor({
-          path: imported.lastImported,
-          offset: headValue(imported.lastImported).length
-        })
-        : null,
+    const destEmpty = destThought.value === '' && getAllChildren(state, pathToContext(simplePath)).length === 0
 
-    ])(state)
+    // New dummy thoughts for collapsing
+    const newDummyThought = getAllChildren(updatedState, pathToContext(simplePath)).find(child => child.value === uuid)
```
raineorshine, line 178: It's not super important, but you should probably store `pathToContext(simplePath)` in a variable since you call it multiple times.

src/reducers/importText.ts
```diff
@@ -157,22 +160,55 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
...
+      updateThoughts(imported),
+      // set cursor to the dummy thought
+      setCursor({
+        path: newDestinationPath,
+      }),
+      collapseContext({ preventCursorArchive: true }),
+      // if original destination has empty then collapse once more
+      ...destEmpty ? [setCursor({
+        path: parentOf(newDestinationPath)
+      }), collapseContext({ preventCursorArchive: true })] : [],
```
raineorshine, line 206: I'm concerned about the use of `setCursor` when `preventSetCursor` is true. What about adding an `at` option to `collapseContext` that defaults to `state.cursor` but can be set to any `Path`? Then we can pass `parentOf(newDestinationPath)` directly. We use this pattern elsewhere so there is some precedent.

raineorshine (changes requested):
Great, overall approach looks good! Just some code quality suggestions.

src/reducers/importText.ts
shresthabijay: > I suggest `[...simplePath, newDummyThought]` instead of `simplePath.concat(newDummyThought)`. It's clean and slightly faster. While I'm sure the performance is not relevant here, I think it's a good habit to get into since `Array.prototype.concat` doesn't offer any advantages based on my research.

Thanks. I was not aware of the slight performance benefit.

src/reducers/collapseContext.ts
shresthabijay: Sure.

src/reducers/existingThoughtChange.ts
```diff
@@ -277,12 +277,15 @@ const existingThoughtChange = (state: State, { oldValue, newValue, context, show
...
+      const isSameContext = hashContext(contextOld) === hashContext(contextNew)
+
       return {
         [contextOldEncoded]: null,
         [contextNewEncoded]: {
           ...(state.thoughts.contextIndex || {})[contextOldEncoded],
           context: contextNew,
-          children: [...thoughtsOld, ...thoughtsNew],
+          // if previous and new context is the same then do not duplicate children
+          children: [...isSameContext ? [] : thoughtsOld, ...thoughtsNew],
```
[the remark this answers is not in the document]
shresthabijay: Sure.

shresthabijay:

src/reducers/__tests__/existingThoughtChange.ts
```diff
@@ -221,3 +221,35 @@ it('edit a child with the same value as its parent', () => {
...
+  // run steps through reducer flow and export as plaintext for readable test
+  const stateNew = reducerFlow(steps)(initialState())
+  const exported = exportContext(stateNew, [ROOT_TOKEN], 'text/plain')
+
+  expect(exported).toBe(`- ${ROOT_TOKEN}
+  - a
+    - b`)
+
+  expect(getAllChildren(stateNew, ['a']))
+    .toMatchObject([{ value: 'b', rank: 0 }])
```
raineorshine, line 253: Technically the `getAllChildren` assertion is unnecessary since `exportContext` walks the `contextIndex` and returns the children.

src/reducers/collapseContext.ts
```diff
@@ -1,15 +1,26 @@
...
 
 /** Collapses the active thought. */
-const collapseContext = (state: State) => {
+const collapseContext = (state: State, { deleteCursor, at }: Options) => {
   const { cursor } = state
 
   if (!cursor) return state
 
-  const children = getChildren(state, pathToContext(cursor))
+  const simpleCursor = simplifyPath(state, cursor)
```
raineorshine, line 20: `at` is supposed to completely replace the `cursor`, e.g.: https://github.com/cybersemics/em/blob/cd80687634aff5c34b7f9e0b03fa0a16a77f2e75/src/reducers/newThought.ts#L94 It allows any `Path` to be collapsed. An interface that operates partly on the cursor and partly on an arbitrary `Path` would be confusing. 

At the risk of saying things you already know, let me share a little more of my thought process. I understand it may not apply completely if the dual `Path` approach is needed here, but hopefully you can find something useful in it anyway :).

I ask myself, does the interface accurately reflect the purpose of this function? Am I adding an option  only as an implementation detail, where it doesn't relate directly to the purpose? There are some cases where this is necessary, in which case comments are essential. Otherwise I want to be very careful about how I am designing the internal API.

In this case we have a straightforward reducer that archives a context and moves its children to its parent. It makes sense that you may want to delete instead of archive, but not always, so we make that an optional argument. It also makes sense that we want to be able to collapse any `Path`, not just the `cursor`, so we have a way to set the `Path` and default to the `cursor`. `at` is a variable name that reflects where the operation will be executed. We could also opt for a more explicit and less "natural language" approach and call it `path`, although the `at` convention is used elsewhere so I opted for that. If two paths are needed (the site of the operation, and where to move the cursor to), then we should rename `at` to something that reflects that its more narrow purpose. Note that we should also consider the possibility of removing the `setCursor` completely from the reducer, and setting the cursor in the action creator instead. However, the `cursor` movement is considered to be atomic in most cases; if the operation is performed on the `cursor`, the `cursor` should be updated. Otherwise it's possible to leave the `cursor` in an invalid state (such as a `Path` that has been moved or deleted). So this leads us to automatic `cursor` movement as the default, with some options to override if necessary (like in `importText`).

src/reducers/collapseContext.ts
```diff
@@ -1,15 +1,26 @@
...
+  const simpleContext = pathToContext(simpleCursor)
```
raineorshine, line 21: All `Contexts` are simple, so you can just call this variable `context`.

src/reducers/importText.ts
```diff
@@ -114,6 +117,7 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
   const isRoam = validateRoam(text)
 
   const simplePath = simplifyPath(state, path)
+  const simpleContext = pathToContext(simplePath)
```
raineorshine, line 120: All `Contexts` are simple, so you can just call this variable `context`.

src/reducers/importText.ts
```diff
@@ -157,22 +161,58 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
...
-        : null,
+    const destEmpty = destThought.value === '' && getAllChildren(state, simpleContext).length === 0
 
-    ])(state)
+    // New dummy thoughts for collapsing
+    const newDummyThought = getAllChildren(updatedState, simpleContext).find(child => child.value === uuid)
+
+    const newDestinationPath = unroot(newDummyThought ? [...simplePath, newDummyThought] : simplePath) as SimplePath
+
+    const imported = importJSON(updatedState, newDestinationPath as SimplePath, json, { lastUpdated, skipRoot })
```
raineorshine, line 183: `newDestinationPath` is already a `SimplePath` since you cast it on assignment, so you don't need the extra cast here.

raineorshine (changes requested):

src/reducers/collapseContext.ts
shresthabijay: @raineorshine Thanks for the detailed thought process. I now properly understand the use of `at`. I was trying to achieve different thing using `at`. I was using it to override default `cursor` setting logic of `collapseContext` and directly set given `path` as the cursor passed using `at`. Back of the mind I was questioning myself that this doesn't reflect the purpose of the function. I was very wrong.

shresthabijay:
@raineorshine I have made the changes. Please review the changes. Thanks again for guiding through things that I am working on :).

src/reducers/collapseContext.ts
raineorshine: No worries! I see now that `at` was clever but properly not the best variable name to use :).

src/reducers/collapseContext.ts
```diff
@@ -1,32 +1,50 @@
...
-const collapseContext = (state: State) => {
+const collapseContext = (state: State, { deleteCursor, at }: Options) => {
   const { cursor } = state
 
-  if (!cursor) return state
+  const path = at || cursor
+
+  if (!path) return state
+
+  const simpleCursor = simplifyPath(state, path)
```
raineorshine, line 22: Since `path` may not be the `cursor` any more, we should call this `simplePath`.

src/reducers/importText.ts
```diff
@@ -157,22 +161,56 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
...
+      collapseContext({ deleteCursor: true, at: newDestinationPath }),
+      // if original destination has empty then collapse once more.
+      destEmpty ?
+        collapseContext({ deleteCursor: true, at: parentOfDestination }) : null,
+      // restore the selection to the last imported thought on the first level
+      // Note: Since collapseContext behavior sets cursor to the first children, we need to set cursor back to the old cursor if preventSetCursor is true.
+      !preventSetCursor ? setLastImportedCursor : setCursor({
+        path: state.cursor
+      })
+    ])(updatedState)
```
raineorshine, line 213: Great, this is cleaner.

raineorshine (approved):

PR #1034, opened 2021-02-13
compare hash in chain

### LINKED ISSUE #426: Context View: Hash matched contexts do not show children

shresthabijay:
fix #426


shresthabijay:
@raineorshine The previous PR #808 was too bar behind from the dev. It was faster to migrate changes using another PR. Please review. Thanks!

src/util/normalizeThought.ts
```diff
@@ -1,31 +1,12 @@
...
-}
-
-/** Strips all html tags. */
-const stripTags = (s: string) => s.replace(REGEXP_TAGS, '')
-
-/**
- * Making character 's' will just become an empty value ''.
- * Skip it else it will cause "s" character to have same no of context as empty thoughts in the entire tree. */
-const singularize = (s: string) => s !== 's' ? pluralize.singular(s) : s
-
```
raineorshine, line 28: I would prefer to keep these functions in this module. It keeps the definitions close to their use and minimizes the surface area of the project's internal API. Then if you need to use one of these functions in a separate module then you can factor it out.

src/selectors/chain.ts
```diff
@@ -1,18 +1,18 @@
...
+import { Path } from '../types'
 import { State } from '../util/initialState'
-import { Path, SimplePath } from '../types'
 
 /** Merges thoughts into a context chain, removing the overlapping head. */
 // use autogenerated rank of context
-// if there is no/empty context chain, return path as-is
-const chain = (state: State, contextChain: SimplePath[], path: SimplePath): Path => {
+// if there is no/empty context chain, return thoughtsRanked as-is
+const chain = (state: State, contextChain: Path[], path: Path) => {
```
raineorshine, line 9: A `contextChain` should only contain `SimplePaths`, so `Path` is incorrect here. Was there a reason you needed to change the type?

src/selectors/chain.ts
```diff
@@ -1,18 +1,18 @@
...
 
   if (!contextChain || contextChain.length === 0) return path
 
   // get the head thought in the last segment of the contextChain
   const pivot = head(contextChain[contextChain.length - 1])
-  const i = path.findIndex(child => child.value === pivot.value)
+  const i = path.findIndex(child => hashThought(child.value) === hashThought(pivot.value))
```
raineorshine, line 15: Let's use `normalizeThought` for comparisons since it's equivalent and faster.

src/selectors/chain.ts
```diff
@@ -22,16 +22,16 @@ const chain = (state: State, contextChain: SimplePath[], path: SimplePath): Path
   //   console.warn('pivot', pivot)
   // }
 
-  const append = path.slice(i - 1) as SimplePath
+  const append = path.slice(i - 1)
   const contexts = getContextsSortedAndRanked(state, pivot.value)
-  const appendedThoughtInContext = contexts.find(child => head(child.context) === append[0].value)
+  const appendedThoughtInContext = contexts.find(child => hashThought(head(child.context)) === hashThought(append[0].value))
```
raineorshine, line 27: Let's use `normalizeThought` for comparisons since it's equivalent and faster.

raineorshine (changes requested):

src/util/normalizeThought.ts
shresthabijay: Understood.

src/selectors/chain.ts
shresthabijay: Ahh yes! I made a mistake here. Thanks for pointing it out.

src/util/normalizeThought.ts
```diff
@@ -1,30 +1,30 @@
...
+import * as pluralize from 'pluralize'
 import { REGEXP_TAGS } from '../constants'
-
-/** Converts a string to lowecase. */
-const lower = (s: string) => s.toLowerCase()
+import emojiStrip from 'emoji-strip'
 
 /** Trims a string. */
-const trim = (s: string) => s.replace(
+export const trim = (s: string) => s.replace(
```
raineorshine, line 7: These are internal API and should not be exported, but not a big issue.

raineorshine (approved):

Issue #1035, opened 2021-02-13
Use of setTimeout outside a function


anmolarora1:
In our current code, we're using a hack to load whitelisted thoughts on the app startup - using [setTimeout](https://github.com/cybersemics/em/blob/b2bfc9e110185c4a823ff81e716a64cf25b19169/src/reducers/updateThoughts.ts#L31) outside a function

The main issue is that the util imports are undefined at load time for some reason. However, the whitelisted thoughts need to be loaded on app startup. The setTimeout ensures that the util functions are imported before importHTML is called.

raineorshine:
Suspect `raw.macro`

PR #1038, opened 2021-02-15
Removes setTimeout used for whitelisted thoughts

anmolarora1:
#1035 

This moves the initialization of whitelisted thoughts to a singleton function, and also removes the cyclic dependencies from the [test store](https://github.com/cybersemics/em/blob/0c634375e991c1a4ac5253273ddaf5b4c8dfe4cf/src/test-helpers/createTestStore.ts#L19)


raineorshine (approved):

raineorshine:
@anmolarora1 Just a small reminder to write "Closes #1035" or "Fixes #1035" so that we can take advantage of GitHub's automatic issue closing on merge.

Issue #987, opened 2021-01-10
Undo/Redo: Push changes


raineorshine:
Currently an undo or redo action manually updates the state. If `state.thoughts` are updated, it needs to also trigger the `pushQueue` to save the changes to the data providers (IndexedDB and Firebase). 

This can be done by calling [updateThoughts](https://github.com/cybersemics/em/blob/dev/src/reducers/updateThoughts.ts). You will need to extract the `thoughtIndex` and `contextIndex` updates from the patch to pass to `updatesThoughts`. (None of the other arguments are needed.)

Here is how to [simulate a page refresh](https://github.com/cybersemics/em/blob/cbc247e10522ec9488e918a50c46c2f77fbf3e83/src/redux-middleware/__tests__/pullQueue.ts#L76) for the test.

raineorshine:
As discussed, call the `updateThoughts` reducer in the enhancer and merge it into the new state.

If all goes well, the updates will get propagated to IndexedDB before the `pullQueue` tries to sync. That should fix the undo/redo being negated by `pull`.

anmolarora1:
@raineorshine Can we close this issue now?

Issue #919, opened 2020-11-21
Upgrade to react-scripts v4


raineorshine:
The [pullQueue tests](https://github.com/cybersemics/em/blob/dev/src/redux-middleware/__tests__/pullQueue.ts) fail on `react-scripts` v4. It is related to the fake timer and mock debounce logic, but I can't figure out how to fix it. We should be able to use `useFakeTimers('modern')` now and not have to mock lodash's debounce, but that causes an error.

 Here are the CHANGELOGS although I didn't find anything relevant:
- create-react-app CHANGELOG: https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md
- jest CHANGELOG: https://github.com/facebook/jest/blob/master/CHANGELOG.md

shresthabijay:
@raineorshine I finally figured why `useFakeTimers('modern')` is not working. Even though we upgraded the `react-scripts` which has implementation for `modern` implementation of fake timer, we are still explicitly using `jest-environment-jsdom-sixteen` as the testing environment.

https://github.com/cybersemics/em/blob/5baf45d1bc0e188dd2dd744dd806f3c885ef3100/package.json#L120

It still does not pass `modern` implementation of fake timer to its environment.  `jest` 26 ships with `jsdom` 16 by default. I tested the lodash's `debounce` with upgraded `react-scripts` and `jest` and it's working with `useFakeTimers('modern')`.

However tests are failing for `pullQueue` and `dexie`. I am working on it.

shresthabijay:
@raineorshine  `dexie` tests are failing after upgrade. Somehow Table.get is timing out when using fake timer. I isolated and tested get functions of db ex db.helpers.get in both versions of react-scripts. In the latest version it is timing out while using jest's fake timer (for both legacy and modern). I looked into Dexie codebase for any time dependent logic but I couldn't find any leads. But it's working without using fake timers on both versions. So it cannot be problem with the mock indexed db.

shresthabijay:
@raineorshine I found the reason for dexie not working with modern implementation of fake timer. Turns out the reason is `fake-indexeddb`. `modern` fake timer freezes `setImmediate` calls too. `fake-indexeddb` uses `setImmediate` everywhere to execute call backs and handlers in next iteration of event loop. But `dexie` resolves its promises using these handlers.

```js
          get: function (_a) {
                var trans = _a.trans, key = _a.key;
                return new Promise(function (resolve, reject) {
                    resolve = wrap(resolve);
[...]

PR #1039, opened 2021-02-15
adds persistence for undo/redo actions

anmolarora1:
#987 

As per the discussion [here](https://github.com/cybersemics/em/issues/987#issuecomment-771038879), implemented the persistence for undo/redo actions by calling `updateThoughts` from the undoRedoReducerEnhancer - 
We're extracting thought and context index updates from the undo/redo patches, passing them to the `updateThoughts` reducer, and merging the resulting state with the state returned by the undo/redo reducer.


src/shortcuts/__tests__/undo.ts
```diff
@@ -69,6 +74,45 @@ it('undo thought change', () => {
   expect(exported).toEqual(expectedOutput)
 })
 
+it('persists undo thought change', async () => {
+
+  initialize()
```
raineorshine, line 79: `initialize` is async and needs to be awaited

src/shortcuts/__tests__/undo.ts
```diff
@@ -69,6 +74,45 @@ it('undo thought change', () => {
...
+    importText({
+      path: RANKED_ROOT,
+      text: `
+        - a
+        - b`
+    }),
+    setCursorFirstMatchActionCreator(['a']),
+  ]
+  )
+  await fakeTimer.runAllAsync()
```
raineorshine, line 93: Maybe this can be removed and the actions can be combined into one dispatch once `initialize` is awaited?

src/shortcuts/__tests__/undo.ts
```diff
@@ -69,6 +74,45 @@ it('undo thought change', () => {
...
+  appStore.dispatch([newThought({ value: 'alpha', insertNewSubthought: true }), { type: 'undoAction' }])
+  await fakeTimer.runAllAsync()
+
+  fakeTimer.useRealTimer()
+
+  // clear and call initialize again to reload from local db (simulating page refresh)
+  appStore.dispatch(clear())
+  fakeTimer.useFakeTimer()
+  initialize()
+  await fakeTimer.runAllAsync()
```
raineorshine, line 104: Try `await initialize()` here as well

src/redux-enhancers/undoRedoReducerEnhancer.ts
```diff
@@ -4,9 +4,30 @@ import { Action, Store, StoreEnhancer, StoreEnhancerStoreCreator } from 'redux'
...
 import { State } from '../util/initialState'
 import { Index, Patch } from '../types'
+import { updateThoughts } from '../reducers'
 
 const stateSectionsToOmit = ['alert', 'pushQueue', 'user']
 
+/**
+ * Manually extract thought and context index updates along with pushQueue.
+ */
+const extractUpdates = (patch: Patch, newState: State) => {
```
raineorshine, line 14: Think of any function that takes a `State` and returns a new `State` as a reducer. `state: State` should be the first argument by convention.

src/redux-enhancers/undoRedoReducerEnhancer.ts
```diff
@@ -4,9 +4,30 @@ import { Action, Store, StoreEnhancer, StoreEnhancerStoreCreator } from 'redux'
...
+
+  const thoughtIndexUpdates = thoughtIndexChanges.reduce((acc, { path }) => {
+    const [thoughtId] = path.slice(thoughtIndexPath.length).split('/')
+    return { ...acc, [thoughtId]: newState.thoughts.thoughtIndex[thoughtId] ?? null }
+  }, {})
+  const contextIndexUpdates = contextIndexChanges.reduce((acc, { path }) => {
+    const [contextId] = path.slice(contextIndexPath.length).split('/')
+    return { ...acc, [contextId]: newState.thoughts.contextIndex[contextId] ?? null }
+  }, {})
+  return updateThoughts({ thoughtIndexUpdates, contextIndexUpdates })(newState)
```
raineorshine, line 28: Nice, this looks great

src/redux-enhancers/undoRedoReducerEnhancer.ts
```diff
@@ -89,11 +110,17 @@ const undoHandler = (state: State, inversePatches: Patch[]) => {
...
+  if (!inversePatches.length) return state
+
+  const undoTwice = penultimateInversePatch && !!((lastAction && NAVIGATION_ACTIONS[lastAction]) ||
+  (penultimateAction && (NAVIGATION_ACTIONS[penultimateAction] || penultimateAction === 'newThought')))
+
+  const newState = undoTwice ? undoReducer(undoReducer(state)) : undoReducer(state)
+
+  const poppedInversePatches = undoTwice ? [penultimateInversePatch, lastInversePatch] : [lastInversePatch]
+  const asyncUpdates = extractUpdates(poppedInversePatches.flat(), newState)
+  return { ...newState, ...asyncUpdates }
```
raineorshine, line 123: Note that you could also use `reducerFlow` here to generate the new state.

raineorshine (changes requested):

anmolarora1:
@raineorshine You pointed out correctly that the `initialize` is an async function and needs to be called using await. But waiting for the initialize function twice in a test exceeds the jest's default timeout threshold of `5000ms`. We should be able to fix that with some updates to our jest config, but I don't think a test should take more than 5s. 
Furthermore, I noticed that we're not awaiting for `initialize` call in [the tests for pullQueue](https://github.com/cybersemics/em/blob/5f6c27167372dd834cf188c9204fdff530e069c6/src/redux-middleware/__tests__/pullQueue.ts#L24), and adding `await` to them made them fail as well with the same timeout reason.
However, every call to initialize is followed by ```await fakeTimer.runAllAsync()```, and I suppose that's been added to handle async initialize call without breaking jest's timeout. [Here's](https://github.com/cybersemics/em/issues/919#issuecomment-739135971) an excellent explanation written by @shresthabijay , where he's mentioned that it's the Sinon's fake timers (precisely, the `runAllAsync` function in this case) that help us avoid the timeout failures related to `dexie`. 
So as per my understanding of it, we probably don't need to await those calls. Let me know what you think.

src/shortcuts/__tests__/undo.ts
raineorshine: Let's call `await fakeTimer.runAllAsync` immediately after.

raineorshine:
> @raineorshine You pointed out correctly that the `initialize` is an async function and needs to be called using await. But waiting for the initialize function in a test exceeds the jest's default timeout threshold of `5000ms`. We should be able to fix that with some updates to our jest config, but I don't think a test should take more than 5s.
> Furthermore, I noticed that we're not awaiting for `initialize` call in [the tests for pullQueue](https://github.com/cybersemics/em/blob/5f6c27167372dd834cf188c9204fdff530e069c6/src/redux-middleware/__tests__/pullQueue.ts#L24), and adding `await` to them made them fail as well with the same timeout reason.

Ah, thanks for the reminder! I had all sorts of trouble with timers before Bijay found a fix.

> [Here's](https://github.com/cybersemics/em/issues/919#issuecomment-739135971) an excellent explanation written by @shresthabijay , where he's mentioned that it's the Sinon's fake timers (precisely, the `runAllAsync` function in this case) that help us avoid the timeout failures related to `dexie`.

Beautiful. Let's add a comment that points to that comment so that others (i.e. me) do not forget the reasoning behind `runAllAsync`.

> So as per my understanding of it, we probably don't need to await those calls. Let me know what you think.

Yup, I understand now. Are you able to move the `runAllAsync` to immediately after `initialize`? It would be cleaner to have all the actions batched together in one dispatch since they are contiguous.

raineorshine (approved):

PR #1042, opened 2021-02-18
Quick add

### LINKED ISSUE #749: Quick Add
This task adds a "Quick Add" feature to quickly add a thought to **em**. It renders and adds new thoughts to a unique `[ABSOLUTE]` context which represents no context.

- Add a plus icon with `position: fixed` in the bottom right corner. Find a suitable svg icon from nounproject.com.
- Define `const ABSOLUTE = '__ABSOLUTE__'` in [constants.ts](https://github.com/cybersemics/em/blob/dev/src/constants.ts). This represents a context whose children do not appear in any other context. (We need to represent an absolute "no context" as an actual context in order to allow O(1) lookup of children.)
- Add `state.startingContext: [ROOT_TOKEN]` to track the starting context of the Content component. Replace the [hardcoded `[ROOT_TOKEN]` value in the Content component](https://github.com/cybersemics/em/blob/dev/src/components/Content.tsx#L40) with `state.startingContext`. 
- **When the icon is clicked, show thoughts that have no context and were created after the icon is clicked**. Click the icon to set `startingContext` to `[ABSOLUTE]`. Add a condition to the calculation of `rootThoughts` in the Content component: if `startingContext` is `[ABSOLUTE]`, then filter the `rootChildren` by `child.created > t` where `t` is the time when the icon is clicked. Thus it will initially be empty until the user creates a new thought.
- Editing any child at the top level should save it to the `[ABSOLUTE]` context.
- Editing and navigating below the top level should function as normal.
- Add a `quickAdd` shortcut that has the same effect as clicking the icon. Its keyboard shortcut should be `Alt + n`.

- **Faster way back than Recently Edited:** When Quick Add is activated, save the `cursor` in `cursorBeforeQuickAdd` which functions just like `cursorBeforeSearch`. i.e. When `startingContext` is `[ABSOLUTE]`, there is no `cursor`, and the user activates `cursorBack`, reset `startingContext` to `[ROOT]` and restore `cursor` to `cursorBeforeQuickAdd`  This should go in the same place as the `cursorBeforeSearchLogic`. 
- **Faster new thought:** Render an empty thought and set the browser selection when Quick Add is activated. Do not save it to state until the user starts typing.

shresthabijay:
fixes #749


shresthabijay:
@raineorshine I have implemented the feature. I have not written any tests yet. Please review it. Thanks!

raineorshine:
I tested the functionality and it's exactly what I wanted! Thanks! 

There are some bugs/edge cases I didn't anticipate. Would you prefer dealing with them in this PR or separate issues?

Before I review, I'm sorry to have to ask for this, but can you rebase and do the renames in a separate commit? I know you'll have to rename them back to the originals, then rename them in a separate commit, but that will still be faster than me manually reviewing 139 files 😳.

raineorshine (changes requested):
Before I review, I'm sorry to have to ask for this, but can you rebase and do the renames in a separate commit? I know you'll have to rename them back to the originals, then rename them in a separate commit, but that will still be faster than me manually reviewing 139 files 😳.

shresthabijay:
> Before I review, I'm sorry to have to ask for this, but can you rebase and do the renames in a separate commit? I know you'll have to rename them back to the originals, then rename them in a separate commit, but that will still be faster than me manually reviewing 139 files .

I still have to to change `isRoot` to something like `isStartingContext` and convert `rootedParentOf` to a selector. This will still affect many files. How about I separate  and list the files that needs to reviewed from the files that only had rename changes ? Will it help ?

raineorshine:
> I still have to to change `isRoot` to something like `isStartingContext` and convert `rootedParentOf` to a selector. This will still affect many files.

Sure. This could be done first, so it's separated from the other changes, right?

> How about I separate and list the files that needs to reviewed from the files that only had rename changes ? Will it help ?

Yes, that would work. Seems like that involves the process suggested of creating a separate commit for the renames? Or are you thinking of doing it manually?

shresthabijay:
> Yes, that would work. Seems like that involves the process suggested of creating a separate commit for the renames? Or are you thinking of doing it manually?

I think I will go with the suggested process.

shresthabijay:
@raineorshine I renamed everything back and squashed everything to a single commit. The number of files changed is down to 59. I also then renamed everything for consistent semantics as we discussed. 

You can just review the diff of the first commit. I hope this helps. [https://github.com/cybersemics/em/pull/1042/commits/09eaf919f7c06df31b89611a978a7ad2266d1c20](renamed commit)

raineorshine:
@shresthabijay 
> There are some bugs/edge cases I didn't anticipate. Would you prefer dealing with them in this PR or separate issues?

src/selectors/getChildren.ts
```diff
@@ -95,3 +94,18 @@ export const isChildVisibleWithCursorCheck = _.curry((state: State, path: Path,
...
+
+/**
+ * Children filter predicate used for rendering.
+ *
+ * 1. Checks if the child is visible.
+ * 2. Checks if child is within cursor.
+ * 3. Checks if child is created after latest absolute context toggle if starting context is absolute.
+ */
+export const childrenFilterPredicate = _.curry((state: State, path: Path, simplePath: SimplePath, child: Child | ThoughtContext) => {
+  return isChildVisibleWithCursorCheck(state, path, simplePath, child) &&
```
raineorshine, line 109: Is `isChildVisibleWithCursorCheck` still used externally to this module? If not, we can remove `export` and make it internal.

src/util/isAbsolute.ts
```diff
@@ -0,0 +1,13 @@
+import { ABSOLUTE_TOKEN } from '../constants'
+import { Child, ThoughtContext } from '../types'
+
+/** Returns true if the thoughts or path is the absolute root. */
```
raineorshine, line 4: Change to "Returns true if the thoughts or path is the absolute context"

The absolute context is not *inherently* a root, so "the absolute root" is a bit misleading; the current view has a root context which theoretically can point to any context (though I'm aware the current implementation isn't quite flexible enough for that).

src/util/isHome.ts
```diff
@@ -0,0 +1,13 @@
+import { ROOT_TOKEN } from '../constants'
+import { Child, ThoughtContext } from '../types'
+
+/** Returns true if the thoughts or path is the home root. */
```
raineorshine, line 4: Change to "Returns true if the thoughts or path is the home context"

raineorshine (changes requested):
Looks good! Thanks for dealing with the weird edge cases for this (the whole task is kind of a weird edge case!)

The change of `rootedParentOf` from a util to a selector should have been in its own commit. That would have made reviewing easier for me.

One subtle difference that I would have made, which I think is purely a judgment call, is to hardcode the values of `TRANSIENT_ABSOLUTE_CHILD_PATH` and `ROOT_PATH_MAP` rather than define constants. They're intuitive/motivated rather than arbitrary like `ROOT_TOKEN` or `ABSOLUTE_TOKEN` so they would make sense in context, they are trivially small, only used in two places, and hardcoding them helps minimize the surface area of the internal API. The downside is that if the transient path implementation were to change, the constant would make that a bit easier.

shresthabijay:
@raineorshine I have made the changes. The tests are passing in my system. But the tests are failing in the github. It seems to be the problem with latest dev.

raineorshine:
> The tests are passing in my system. But the tests are failing in the github. It seems to be the problem with latest dev.

This happened when I added the dependency `https://github.com/magic-akari/page-lifecycle#feat/add-types` using npm `v7`. There is a bug in npm `v7` that changes `https://` to `git+ssh://` in the package-lock.json, which then fails the CI install. It's being tracked here: https://github.com/npm/cli/issues/2610.

We'll have to either stick to npm `v6` or manually change `git+ssh` back to `https` in package-lock.json whenever installing dependencies.

src/components/Content.tsx
```diff
@@ -20,15 +20,20 @@ import { SimplePath } from '../types'
 const tutorialLocal = localStorage['Settings/Tutorial'] === 'On'
 const tutorialStepLocal = +(localStorage['Settings/Tutorial Step'] || 1)
 
+const TransientChildPath = [{
```
raineorshine, line 23: camelCase

src/selectors/getRootPath.ts
```diff
@@ -0,0 +1,18 @@
+import { ABSOLUTE_PATH, ABSOLUTE_TOKEN, HOME_PATH, HOME_TOKEN } from '../constants'
+import { SimplePath } from '../types'
+import { State } from '../util/initialState'
+
+const RootPathMap: Record<string, SimplePath> = {
```
raineorshine, line 5: camelCase

raineorshine (approved):

raineorshine:
@shresthabijay Third time's the charm? :)

> @shresthabijay
> 
> > There are some bugs/edge cases I didn't anticipate. Would you prefer dealing with them in this PR or separate issues?

shresthabijay:
> @shresthabijay Third time's the charm? :)
> 
> > @shresthabijay
> > 
> > > There are some bugs/edge cases I didn't anticipate. Would you prefer dealing with them in this PR or separate issues?
> 
> 

@raineorshine Sure :) Please describe the bugs/edge cases in new issues and  assign them to me.

PR #1050, opened 2021-03-02
Adds pendingPulls and fixes existingThoughtMove for pending thoughts

### LINKED ISSUE #1026: existingThoughtMove: Pending merged thoughts deleted
## Steps to Reproduce

``` 
- a
  - b
    - c
- d
  - b
    - e
``` 

1. Paste the above thoughts
2. Set the cursor to null
3. Refresh the page
4. Set the cursor on `a` to expand it
5. Move `a/b` into `d`

## Current Behavior

- a
- d
  - b
    - c

## Expected Behavior

- a
- d
  - b
    - e
    - c

## Notes

[flushMoves](https://github.com/cybersemics/em/blob/b2bfc9e110185c4a823ff81e716a64cf25b19169/src/redux-middleware/pushQueue.ts#L110) in the `pushQueue` only uses `pathOld`, but should also flush `pathNew`.

Analyze what occurs in between the first 'existingThoughtMove` (in-memory thoughts) and the second (pending thoughts) to confirm the existing approach will still work if merging pending.

anmolarora1:
fixes #1026


raineorshine:
Hi Anmol. I try not to review code with no tests. Having tests in place assures me that the functionality is working and allows me to focus on architecture and performance. The tests won't change even if our approach changes, so there's no downside to writing them up front. Also, if I can nudge you towards TDD, that would be great too!

Once you write a test I can do a general review, but if you have specific questions please let me know any time. Thanks!

anmolarora1:
> Hi Anmol! Just to let you know, I try not to review code with no tests. Having tests in place assures me that the functionality is working and allows me to focus on architecture and performance. I understand that it might be nice to get my feedback early on, but it's not the best use of my time. The tests won't change even if our approach changes, so there's no downside to writing them up front. Also, if I can nudge you towards TDD, that would be great too!

I agree that having TDD is useful in some scenarios, this being one of them. I've added the tests :)

src/reducers/__tests__/existingThoughtMove.ts
```diff
@@ -560,3 +567,62 @@ it('move with nested duplicate thoughts and merge their children', () => {
...
+
+  await timer.runAllAsync()
+
+  timer.useFakeTimer()
+
+  appStore.dispatch([setCursorFirstMatchActionCreator(['a'])])
+  await timer.runAllAsync()
+
+  appStore.dispatch([
+    setCursorFirstMatchActionCreator(['a']),
```
raineorshine, line 612: Looks like `setCursorFirstMatchActionCreator(['a'])` is dispatched twice. Is that intentional?

src/reducers/__tests__/existingThoughtMove.ts
```diff
@@ -560,3 +567,62 @@ it('move with nested duplicate thoughts and merge their children', () => {
...
+    existingThoughtMoveAction({
+      oldPath: [{ value: 'a', rank: 0 }, { value: 'b', rank: 0 }],
+      newPath: [{ value: 'd', rank: 1 }, { value: 'b', rank: 1 }],
+    }),
+    setCursorFirstMatchActionCreator(['d', 'b'])
```
raineorshine, line 617: I believe setting the cursor at the end is unnecessary for this test.

src/reducers/__tests__/existingThoughtMove.ts
```diff
@@ -560,3 +567,62 @@ it('move with nested duplicate thoughts and merge their children', () => {
...
+  ]
+  )
+  await timer.runAllAsync()
+
+  timer.useRealTimer()
+
+  const mergedChildren = getAllChildren(appStore.getState(), ['d', 'b', 'c']).map(child => child.value)
+
+  expect(new Set(mergedChildren)).toMatchObject(new Set(['1', '2', '3', '4']))
```
raineorshine, line 626: You should test the ranks too to make sure they are not conflicting.

raineorshine (commented):
Functionality and code looks good! Just a couple changes requested on the tests.

raineorshine (changes requested):
Oops, meant to request changes.

raineorshine:
@anmolarora1 Is this blocked or did it just get sidelined? Thanks!

src/redux-middleware/pushQueue.ts
```diff
@@ -108,27 +108,42 @@ const flushEdits = (pushQueue: PushBatch[]): Thunk<Promise<void>> => async (disp
 /** Pull all descendants of pending moves and dispatch existingThoughtMove to fully move. */
 const flushMoves = (pushQueue: PushBatch[]): Thunk => async (dispatch, getState) => {
 
-  // if there are pending thoughts that need to be moved, dispatch an action to be picked up by the pullQueue middleware which can load pending thoughts before dispatching another existingThoughtMove
-  const pendingMoves = pushQueue.map(batch => batch.pendingMoves || []).flat()
-  if (pendingMoves?.length) {
+  // if there are pending thoughts that need to be deleted, dispatch an action to be picked up by the pullQueue middleware which can load pending thoughts before dispatching another existingThoughtDelete
```
raineorshine, line 111: This comment should refer to moved thoughts rather than deleted thoughts

src/redux-middleware/pushQueue.ts
```diff
@@ -108,27 +108,42 @@ const flushEdits = (pushQueue: PushBatch[]): Thunk<Promise<void>> => async (disp
...
-      return {
-        [hashContext(context)]: context
-      }
+      // skip the pull for loaded descendants
+      return { [hashContext(context)]: context }
     })
 
     await dispatch(pull(pending, { maxDepth: Infinity }))
+    maxDepth = getDepth(getState(), pathToContext(descendantMoves[0].pathOld))
+  }
```
raineorshine, line 126: Let's add a comment that clarifies that this pulls only source thoughts since that's important.

src/redux-middleware/pushQueue.ts
```diff
@@ -108,27 +108,42 @@ const flushEdits = (pushQueue: PushBatch[]): Thunk<Promise<void>> => async (disp
...
       const context = pathToContext(pathOld)
-      return {
-        [hashContext(context)]: context
-      }
+      // skip the pull for loaded descendants
+      return { [hashContext(context)]: context }
     })
 
     await dispatch(pull(pending, { maxDepth: Infinity }))
+    maxDepth = getDepth(getState(), pathToContext(descendantMoves[0].pathOld))
```
raineorshine, line 125: `descendantMoves` combines multiple batches (L112), so `descendantMoves[0]` is not necessarily the root of all `descendantMoves`. For example, if there was a batch that involved moving thoughts with depth 3 and a batch that involved moving thoughts with depth 10, this would mistakenly load only 3 levels.

It would be more correct to use a different pull depth for each batch, but we don't want the overhead of separate pulls. What do you think? We could call `getDepth` on the root of each batch and take the max. That would cover all the cases, and the the inefficiency of pulling a few of the batches too deep is better than the inefficiency of multiple pulls or the data integrity issues of not pulling a batch deeply enough.

src/util/getDepth.ts
```diff
@@ -0,0 +1,18 @@
+import { hashContext } from '.'
+import { Context } from '../types'
+import { State } from './initialState'
+import { unroot } from './unroot'
+
+/** Get max depth of a visible context. */
+export const getDepth = (state: State, context: Context): number => {
```
raineorshine, line 7: We can just use `getDescendants(...).length` to avoid adding an additional function, and it has the benefit of having a single traversal implementation in case the thought structure changes.

raineorshine (changes requested):
Thanks! I think this will work and meet all our requirements, a few small requests notwithstanding.

Please make sure to address the older comments in the test, too.

src/util/getDepth.ts
anmolarora1: As per my understanding, `getDescendants` serves a very different purpose from `getDepth` despite some similarities in the traversing logic. The former is a flat map of all the children whereas we need the depth of the longest branch.

raineorshine: Oops! You're right. Thanks.

src/util/getDepth.ts
```diff
@@ -0,0 +1,18 @@
...
+  const contextIndex = state.thoughts.contextIndex
+  const contextEncoded = hashContext(context)
+  const parentEntry = contextIndex[contextEncoded]
+  const children = parentEntry?.children ?? []
```
raineorshine, line 11: I believe you can can use `getAllChildren` here to decouple from the implementation details of the `contextIndex`.

anmolarora1: This isn't the latest code and I'm using [getChildrenRanked](https://github.com/cybersemics/em/blob/5c44ef66c5a536c0002ea367aacb893da74937de/src/reducers/existingThoughtMove.ts#L70) there.
But `getAllChildren` makes more sense since we don't need ranks.

src/reducers/__tests__/existingThoughtMove.ts
```diff
@@ -620,3 +627,125 @@ it('consitent rank between thoughtIndex and contextIndex on duplicate merge', ()
...
+  expect(mergedChildren).toMatchObject([
+    { value: 'three', rank: 0 },
+    { value: 'four', rank: 1 },
+    { value: 'one', rank: 2 },
+    { value: 'two', rank: 3 },
+  ])
+
+})
+
+it('only fetch the descendants upto the possible conflicting path', async () => {
```
raineorshine, line 693: Great test!

raineorshine (approved):
🎉

PR #1052, opened 2021-03-06
Handle archiving of thoughts with hidden children

### LINKED ISSUE #864: `deleteEmptyThought` with hidden child is cancelled
Normally an empty thought cannot be archived using Backspace if it has any children. However, this is confusing when all of its children are hidden. In this case, instead of doing nothing, we should archive all of the cursor's children in the parent thought then delete the empty thought as normal.

See: https://github.com/cybersemics/em/blob/dev/src/shortcuts/deleteEmptyThoughtOrOutdent.tsx

## Steps to Reproduce

- a
  - [empty] [cursor]
    - =test

Put the caret on the empty thought and hit backspace.

## Current Behavior

Nothing happens.

## Expected Behavior

- a [cursor]
  - =archive
    - =test

In the special case where all of the cursor's children are hidden, instead of doing nothing, archive all of the children in the cursor's parent and then delete the cursor (empty thought).

anmolarora1:
fixes #864

### DIFF
--- a/src/reducers/__tests__/deleteEmptyThought.ts
+++ b/src/reducers/__tests__/deleteEmptyThought.ts
@@ -77,7 +77,6 @@ it(`archive thought with hidden children - arvhive all children in cursor's pare
           - =a
           - =b`
     }),
-
     setCursorFirstMatch(['']),
     deleteEmptyThought
   ]
@@ -88,8 +87,9 @@ it(`archive thought with hidden children - arvhive all children in cursor's pare
   const exported = exportContext(stateNew, [HOME_TOKEN], 'text/plain')
 
   expect(exported).toBe(`- ${HOME_TOKEN}
-  - =a
-  - =b`)
+  - =archive
+    - =a
+    - =b`)
 
 })
 
--- a/src/reducers/__tests__/deleteEmptyThought.ts
+++ b/src/reducers/__tests__/deleteEmptyThought.ts
@@ -3,7 +3,7 @@ import { exportContext } from '../../selectors'
 import { importText } from '../../action-creators'
 import { store } from '../../store'
 import createTestApp, { cleanupTestApp } from '../../test-helpers/createTestApp'
-import { setCursorFirstMatchActionCreator } from '../../test-helpers/setCursorFirstMatch'
+import setCursorFirstMatch, { setCursorFirstMatchActionCreator } from '../../test-helpers/setCursorFirstMatch'
 import { HOME_PATH, HOME_TOKEN } from '../../constants'
 
 // reducers
@@ -13,7 +13,7 @@ import deleteEmptyThought from '../deleteEmptyThought'
 import newSubthought from '../newSubthought'
 import newThought from '../newThought'
 import setCursor from '../setCursor'
-import archiveThought from '../archiveThought'
+import importTextReducer from '../importText'
 
 it('delete empty thought', () => {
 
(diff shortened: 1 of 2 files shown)


src/reducers/__tests__/deleteEmptyThought.ts
```diff
@@ -66,23 +67,23 @@ it('do not delete thought with children', () => {
...
 
   const steps = [
     newThought(''),
-    newSubthought('=one'),
-    cursorBack,
-    deleteEmptyThought,
+    newSubthought('1'),
+    archiveThought({}),
+    deleteEmptyThought
   ]
```
raineorshine, line 77: I think this might be clearer if we did it with a single `importText` and `setCursor` before calling `deleteEmptyThought`. I see how `archiveThought` sets up the condition of having a hidden child, but it took me a minute to figure out that it was being used to set up the test case.

raineorshine: Also, let's use an inert hidden thought like `=one` or `=test` rather than `=archive`. `=archive` is a special case and usually is deleted permanently when it is archived.

src/reducers/__tests__/deleteEmptyThought.ts
```diff
@@ -66,23 +67,23 @@ it('do not delete thought with children', () => {
...
   // run steps through reducer flow and export as plaintext for readable test
   const stateNew = reducerFlow(steps)(initialState())
+
   const exported = exportContext(stateNew, [HOME_TOKEN], 'text/plain')
 
   expect(exported).toBe(`- ${HOME_TOKEN}
   - =archive
-    - ${''/* prevent trim_trailing_whitespace */}
-      - =one`)
+    - 1`)
```
raineorshine, line 86: We'll want to test multiple children, otherwise it's possible that the test passes for a single child but fails when there are multiple children.

src/reducers/deleteEmptyThought.ts
```diff
@@ -28,7 +28,14 @@ const deleteEmptyThought = (state: State): State => {
   }
   // archive an empty thought with hidden children
   else if (isEmpty && visibleChildren.length === 0) {
-    return archiveThought(state, {})
+    const archivedChild = allChildren[0]
+    return reducerFlow([
+      existingThoughtMove({ oldPath: [...simplePath, archivedChild], newPath: [...parentOf(cursor), archivedChild] }),
```
raineorshine, line 33: You're using `simplePath` to create `oldPath`, but `parentOf(cursor)` to create `newPath`. They need to either both be based on `SimplePath` or `Path`. I think `Path` is probably the correct choice here so that the cursor is correctly restored within nested context views.

raineorshine (changes requested):

src/reducers/__tests__/deleteEmptyThought.ts
```diff
@@ -66,23 +67,29 @@ it('do not delete thought with children', () => {
...
   const stateNew = reducerFlow(steps)(initialState())
+
   const exported = exportContext(stateNew, [HOME_TOKEN], 'text/plain')
 
   expect(exported).toBe(`- ${HOME_TOKEN}
-  - =archive
-    - ${''/* prevent trim_trailing_whitespace */}
-      - =one`)
+  - =a
+  - =b`)
```
raineorshine, line 92: This doesn't look quite right. Don't we expect it to archive the children, per the original issue?

raineorshine (changes requested):

src/reducers/__tests__/deleteEmptyThought.ts
anmolarora1: I misunderstood the description and assumed that only the archived thoughts need to be archived in the parent, and the hidden thoughts needed to be moved to the parent. I'm sorry for getting this wrong.
It's fixed now, so please review.

PR #1053, opened 2021-03-08
Add plain text lossless export option

### LINKED ISSUE #991: Export: Plain Text (lossless)
Add an export type of "Plain Text (lossless)" with all metaprogramming attributes included. 

Change the default export type of "Plain Text" to exclude all metaprogramming attributes.

shresthabijay:
fixes #991 

# Changes
- Add plain text lossless and without loss export option.
- Refactor dropdown ui.
- Add on click outside to close dropdown.


shresthabijay:

src/util/filter.ts
```diff
@@ -0,0 +1,10 @@
+interface Predicate<P> {
+  (item: P, index: number): boolean,
+}
+
+/**
+ * Filter with multiple predicate functions.
+ */
+export const filter = <T>(array: T[], predicateArray: (Predicate<T>)[]) => {
+  return predicateArray.length > 0 ? array.filter((item: T, index) => predicateArray.every(predicate => predicate(item, index))) : array
+}
```
raineorshine, line 10: This functionality can be achieved more simply with function composition:

```
items.filter(and(predicate1, predicate2, etc))
```

`and` is a basic higher order function. In order to get access to it you can import [fp-and-or](https://github.com/raineorshine/fp-and-or).

raineorshine (changes requested):
Great! Works as expected. Just one change requested.

src/util/filter.ts
shresthabijay: That makes sense. Thank you!

shresthabijay:
@raineorshine  I have made the change.

src/util/filter.ts
raineorshine: Yeah, super handy. I'm surprised they're not in lodash. I finally made a package just for them since I use them so often :).

raineorshine (approved):

PR #1060, opened 2021-03-22
Split sentence on comma if period is not present

### LINKED ISSUE #1041: Split sentences should split on comma when there are no periods
## Steps to Reproduce

1. Create thought "one, two, three"
2. Activate "Split Sentences"

## Current Behavior

Cannot split sentences.

## Expected Behavior

- one
- two
- three

## Notes

Thoughts with a combination of periods and commas should be split on periods only. 

e.g. Activating Split Sentences on "one. two, three." should yield:

- one.
- two, three.

shresthabijay:
fixes #1041

# Changes
- Seperate splitting value logic to a separate `splitSentence` logic. 
- Add test for the changes.


shresthabijay:

raineorshine (changes requested):
Looks great! I thought of a couple small tweaks I wanted to make to the requirements:

- If a thought with a mix of commas and periods has only one period and it's the last character, remove it and split on comma
- Split on semicolon just like period

shresthabijay:
> * If a thought with a mix of commas and periods has only one period and it's the last character, remove it and split on comma

What if there are other split characters too in the string ? For example `One!, Two, Three.`

shresthabijay:
> * Split on semicolon just like period

Do you mean the same behavior with mix of commas and semilcon ?

raineorshine:
> > * If a thought with a mix of commas and periods has only one period and it's the last character, remove it and split on comma
> 
> What if there are other split characters too in the string ? For example `One!, Two, Three.`

Normally if there are a mix of split characters, it will split on period. However if the period is at the end, you should remove the period and split on comma. Does that answer?

shresthabijay:
> Normally if there are a mix of split characters, it will split on period.

Current behavior is that on mix of split characters, it will split on every split character. For example: `One?Two!Three.Four` will split into `['One.'. 'Two.', 'Three.', 'Four.']` . So if there is a string such as `One,Seven?Two!Three.`, then it should split as `['One', 'Seven?Two!Three']`. Is this correct ?

raineorshine:
> > Normally if there are a mix of split characters, it will split on period.
> 
> Current behavior is that on mix of split characters, it will split on every split character.

That is correct for thoughts without commas.

> So if there is a string such as `One,Seven?Two!Three.`, then it should split as `['One', 'Seven?Two!Three']`. Is this correct ?

No, hard split characters should always take precedence over commas. So it would split as `One,Seven?', 'Two!', 'Three.'`.

raineorshine (approved):

PR #1061, opened 2021-03-22
Use nodeValue of element to access decoded value for calculating offset

### LINKED ISSUE #1040: Thought with "&" split in wrong place
It is off by 4 (perhaps the length of "amp;", which could be added when the amperand is HTML encoded).

## Steps to Reproduce

1. Create a thought with text "fire & brimstone"
2. Set the caret between "fire & brim" and "stone" and hit Enter

## Current Behavior

- fire & brimston
- e

## Expected Behavior

- fire & brim
- stone

shresthabijay:
fixes #1040

### DIFF
--- a/src/util/getOffsetWithinContent.ts
+++ b/src/util/getOffsetWithinContent.ts
@@ -11,8 +11,11 @@ export const getOffsetWithinContent = (el: HTMLElement) => {
   // insert dummy text at the end of the range
   sel!.getRangeAt(0).insertNode(target)
 
+  const nodeBeforeCaret = el.childNodes[0]
+
   // find the index of the dummy text within the inner html of the contenteditable
-  const offset = el.innerHTML.indexOf('\u0001')
+  const offset = nodeBeforeCaret?.nodeValue?.length
+
   target.parentNode?.removeChild(target)
-  return offset !== -1 ? offset : 0
+  return offset || 0
 }


shresthabijay:

src/util/getOffsetWithinContent.ts
```diff
@@ -11,8 +11,11 @@ export const getOffsetWithinContent = (el: HTMLElement) => {
   // insert dummy text at the end of the range
   sel!.getRangeAt(0).insertNode(target)
 
+  const nodeBeforeCaret = el.childNodes[0]
+
   // find the index of the dummy text within the inner html of the contenteditable
-  const offset = el.innerHTML.indexOf('\u0001')
+  const offset = nodeBeforeCaret?.nodeValue?.length
```
raineorshine, line 17: Could you help me understand how this works now? It looks like you are still inserting `\u0001` but not checking it any more?

shresthabijay: The editable target node has only one child text node before inserting the dummy node.

The child nodes of the target element before inserting dummy node is  `["fire & brimstone"]`.
After inserting the dummy node we get `["fire & brim", "", "stone"]`. 

At the first index we have the text before the caret. At the second index we have the dummy itself. At the third index we get the text after the caret. Also `node.nodeValue` will return text without any encoding. So we can take the length of the first text node as the offset.

After we calculate the offest we just remove the dummy node from the target element.

raineorshine: Got it! Thanks!

raineorshine (approved):

PR #1064, opened 2021-03-26
Assume the imported text is html when it starts with a closed html tag

### LINKED ISSUE #1008: Import breaks on <li>
## Steps to Reproduce

Paste the following into an empty thoughtspace:

```
- a
  - b
  - <li>c</li>
```

Nested `<p>` causes a similar problem.

## Current Behavior

All text is imported as a single bullet.

## Expected Behavior

Text should be imported in the expected hierarchy:

- a
  - b
  - c

shresthabijay:
fixes #1008

# Cause of Issue

On importing the following structure
```
- a
  - b
  - <li>c</li>
```

The current logic detected it as a html text  with one li tag instead of indented text. 

# Solution

We cannot simply assume the imported text is html because there is at least one `li` or `p` tag. It is because intended text can have these tags too. In most html import, the first value starts with a tag. So the better approach would be to assume that the imported text is an html when the text starts with a properly closed html tag. 

# Changes

- Changes assume imported text is html when it starts with a closed html tag instead of checking for at least on li tag.
- Add test for the specific case.

### DIFF
--- a/src/reducers/importText.ts
+++ b/src/reducers/importText.ts
@@ -16,8 +16,9 @@ const regexpListItem = /<li(?:\s|>)/gmi
 // '*'' must be followed by a whitespace character to avoid matching *footnotes or *markdown italic*
 const regexpPlaintextBullet = /^\s*(?:[-—▪◦•]|\*\s)/m
 
-// has at least one list item or paragraph
-const regexpHasListItems = /<li|p(?:\s|>).*?>.*<\/li|p>/mi
+// regex that checks if the value starts with closed html tag
+// Note: This regex cannot check properly for a tag nested within itself. However for general cases it works properly.
+const regexStartsWithClosedTag = /^<([A-Z][A-Z0-9]*)\b[^>]*>(.*?)<\/\1>/ism
 
 /** Converts data output from jex-block-parser into HTML.
  *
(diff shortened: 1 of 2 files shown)


shresthabijay:

src/reducers/importText.ts
```diff
@@ -12,12 +12,13 @@ import collapseContext from './collapseContext'
 
 // a list item tag
 const regexpListItem = /<li(?:\s|>)/gmi
-// starts with '-', '—' (emdash), ▪, ◦, •, or '*'' (excluding whitespace)
+// starts with '-', 'â€”' (emdash), â–ª, â—¦, â€¢, or '*'' (excluding whitespace)
 // '*'' must be followed by a whitespace character to avoid matching *footnotes or *markdown italic*
-const regexpPlaintextBullet = /^\s*(?:[-—▪◦•]|\*\s)/m
+const regexpPlaintextBullet = /^\s*(?:[-â€”â–ªâ—¦â€¢]|\*\s)/m
```
raineorshine, line 17: These characters appear to have been corrupted.

src/util/__tests__/importText.ts
```diff
@@ -572,3 +572,17 @@ it('decode HTML entities', () => {
...
 })
+
+// related: https://github.com/cybersemics/em/issues/1008
+it('do not parse as html when value has tags inside indented text', () => {
+
+  expect(importExport(`
+  - a
+    - b
+    - <li>c</li>
+  `).trim())
```
raineorshine, line 583: Let's remove the `trim`; I think in all our tests we do an exact match on the `importExport` result.

raineorshine (changes requested):

raineorshine:
Good solution!

src/util/__tests__/importText.ts
shresthabijay: Sure.

shresthabijay:
@raineorshine I have made the changes. Please review it!

raineorshine (approved):

Issue #495, opened 2020-04-01
Independent Editing


raineorshine:
Decouple thought identity from value so that thoughts can be edited without having to update all descendants.

- a `← should be able to edit without changing all descendants`
  - b

## Migration strategy:

1. **Make `Child.id` required** - This should still be done iteratively, as mentioned, since it has a large surface area. You can detect `Child` nodes that are missing an `id` and add them one by one.
2. **Convert `Child.id` from uuid to contextIndex key** - Ultimately we need to unify `Child.id` and `Parent.id`. If we change `Parent.id` to a uuid to match `Child.id`, it will break `hashContext`. Thus we change `Child.id` to match `Parent.id`. They will be treated as separate universes at first, but this will make it easier to unify them later.
3. **Disable pull/push** - We need to somehow limit the domain of the initial structural change since there are 20+ low-level functions. Disabling pull/push is the best seam I can see to split it along.
*(3b)* **Disable Context View** - Here is another domain that can be disabled without affecting the initial conversion.
5. **Change `Parent.id` to an arbitrary uuid** - This will change the thought structure and require changes in all low-level functions that are still enabled. Preserve their API to avoid any higher-level changes at this point. e.g. `hashContext` will traverse from the HOME context to the desired `Parent` and return its id instead of hashing the `Context`. (This will be slow. We can optimize it later by also accepting a `Path`, which will allow O(1) `Parent` lookup with `head(path).id`.)
7. **Remove `Parent.context`** - We begin removing `Context` from `State`. We are anticipating that once `Parent` entries become stable (i.e. unchanged with edits and ancestors), `Contexts` for a given thought can change. Thus `Context` shifts from a unique key to an ad hoc structure that is eventually replaced with simpler and more efficient access by id. Doing this now will make it easier to refactor recursive functionality since `Parents` will no longer have any reference to their ancestors.
8. **Delete descendant updates** - At this point, we can pretty much delete all the descendant update code in `moveThought`, `editThought`, and `deleteThought`, since they no longer depend on ancestors.
6. **Re-enable pull/push** - All low-level push/pull functions will need to be updated to use the new `state.thoughts` structure. 
9. **Convert `Child` to `Parent` or `Parent.id`** - That is, wherever `Child` is used, replace it with a `Parent.id` (if persisted, e.g. `Parent.children`) or a direct reference to a `Parent` object (in memory business logic). Move `Child.rank` to `Parent.rank`.
[...]

PR #1065, opened 2021-03-27
adds middleware to fetch pending descendants before existingThoughtMove

### LINKED ISSUE #1026: existingThoughtMove: Pending merged thoughts deleted

anmolarora1:
fixes #1026 

Uses a middleware to fetch any pending descendants of the destination context before calling `existingThoughtMove` reducer, as explained [here](https://github.com/cybersemics/em/issues/1026#issuecomment-800888062)

### DIFF
--- a/src/redux-middleware/pullBeforeMove.ts
+++ b/src/redux-middleware/pullBeforeMove.ts
@@ -7,7 +7,7 @@ import { State } from '../util/initialState'
 /**
  * Middleware that fetches any pending descendants before a thought-move operation.
  */
-const fetchBeforeMove: ThunkMiddleware<State> = ({ getState, dispatch }) => {
+const pullBeforeMove: ThunkMiddleware<State> = ({ getState, dispatch }) => {
 
   return next => async action => {
 
@@ -32,4 +32,4 @@ const fetchBeforeMove: ThunkMiddleware<State> = ({ getState, dispatch }) => {
   }
 }
 
-export default fetchBeforeMove
+export default pullBeforeMove
--- a/src/redux-middleware/pullBeforeMove.ts
+++ b/src/redux-middleware/pullBeforeMove.ts
@@ -24,7 +24,7 @@ const fetchBeforeMove: ThunkMiddleware<State> = ({ getState, dispatch }) => {
       const isNewContextPending = conflictedPath && isPending(state, newContext)
       if (isNewContextPending) {
         const toPull = { [hashContext(newContext)]: newContext }
-        await pull(toPull, { maxDepth: Infinity })(dispatch, getState)
+        await dispatch(pull(toPull, { maxDepth: Infinity }))
       }
     }
     next(action)
--- a/src/redux-middleware/pullBeforeMove.ts
+++ b/src/redux-middleware/pullBeforeMove.ts
@@ -1,0 +1,35 @@
+import { ThunkMiddleware } from 'redux-thunk'
+import pull from '../action-creators/pull'
+import { isPending, pathExists, simplifyPath } from '../selectors'
+import { hashContext, head, parentOf, pathToContext } from '../util'
+import { State } from '../util/initialState'
+
+/**
+ * Middleware that fetches any pending descendants before a thought-move operation.
+ */
+const fetchBeforeMove: ThunkMiddleware<State> = ({ getState, dispatch }) => {
+
+  return next => async action => {
+
+    if (action.type === 'existingThoughtMove') {
+      const state = getState()
+
+      const oldContext = pathToContext(simplifyPath(state, action.oldPath))
+
+      const newContext = pathToContext(simplifyPath(state, action.newPath))
+
+      const updatedNewContext = [...parentOf(newContext), head(oldContext)]
+      const conflictedPath = pathExists(state, updatedNewContext)
+
+      const isNewContextPending = conflictedPath && isPending(state, newContext)
+      if (isNewContextPending) {
+        const toPull = { [hashContext(newContext)]: newContext }
+        await pull(toPull, { maxDepth: Infinity })(dispatch, getState)
+      }
+    }
+    next(action)
+
+  }
+}
+
+export default fetchBeforeMove
--- a/src/reducers/__tests__/existingThoughtMove.ts
+++ b/src/reducers/__tests__/existingThoughtMove.ts
@@ -594,7 +594,7 @@ it('data integrity test', () => {
   expect(noOfUpdates).toBe(0)
 })
 
-it('pending thoughts should be merged correctly(fetch pending before move)', async () => {
+it('pending destination should be merged correctly (fetch pending before move)', async () => {
   initialize()
 
   const text = `
@@ -632,13 +632,74 @@ it('pending thoughts should be merged correctly(fetch pending before move)', asy
   timer.useFakeTimer()
 
   appStore.dispatch([setCursorFirstMatchActionCreator(['a'])])
+
+  // wait for pullBeforeMove middleware to execute
   await timer.runAllAsync()
 
   appStore.dispatch([
-    setCursorFirstMatchActionCreator(['a']),
     existingThoughtMoveAction({
       oldPath: [{ value: 'a', rank: 0 }, { value: 'b', rank: 0 }],
       newPath: [{ value: 'd', rank: 1 }, { value: 'b', rank: 1 }],
+    })
+  ])
+  await timer.runAllAsync()
+
+  timer.useRealTimer()
+
+  const mergedChildren = getAllChildren(appStore.getState(), ['d', 'b', 'c'])
+  expect(mergedChildren).toMatchObject([
+    { value: 'three', rank: 0 },
+    { value: 'four', rank: 1 },
+    { value: 'one', rank: 2 },
+    { value: 'two', rank: 3 },
+  ])
+
+})
+
+it('pending destination descendants should be merged correctly (fetch pending before move)', async () => {
+  initialize()
+
+  const text = `
+  - a
+    - b
+      -c
+        - one
+        - two
+  - b
+    - c
+      - three
+      - four`
+
+  timer.useFakeTimer()
+
+  appStore.dispatch([
+    importTextAction({
+      path: HOME_PATH,
+      text
+    }),
+  ])
+  await timer.runAllAsync()
+
+  timer.useFakeTimer()
+  // clear and call initialize again to reload from local db (simulating page refresh)
+  appStore.dispatch(clear())
+  await timer.runAllAsync()
+
+  initialize()
+
+  await timer.runAllAsync()
+
+  timer.useFakeTimer()
+
+  appStore.dispatch([setCursorFirstMatchActionCreator(['a'])])
+
+  // wait for pullBeforeMove middleware to execute
+  await timer.runAllAsync()
+
+  appStore.dispatch([
+    existingThoughtMoveAction({
+      oldPath: [{ value: 'a', rank: 0 }, { value: 'b', rank: 0 }],
+      newPath: [{ value: 'b', rank: 0.5 }],
     })
   ]
   )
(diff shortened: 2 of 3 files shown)


raineorshine:
The CI is failing because the package-lock was rebuilt using npm 7, which has a bug that changes https to ssh: https://github.com/npm/cli/issues/2610

Until the bug is fixed, you'll have to use npm 6 to install dependencies. In this PR, the package file wasn't changed so you can just reset the package lock file. Thanks!

src/action-creators/pull.ts
```diff
@@ -113,4 +114,13 @@ const pull = (contextMap: Index<Context>, { maxDepth }: PullOptions = {}): Thunk
...
 }
 
+/**
+ * Calls pullDescendants to fetch and update descendants.
+ * WARNING: Unknown behavior if thoughtsPending takes longer than throttleFlushPending.
+ */
+const pull = (contextMap: Index<Context>, { maxDepth }: PullOptions = {}): Thunk<Promise<boolean>> => async (dispatch, getState) => {
+  const result = await pullDescendants(contextMap, { maxDepth }, dispatch, getState)
+  return result
+}
```
raineorshine, line 124: Could you explain the purpose of separating `pull` from `pullDescendants`? It looks like you're using `pullDescendants` to remove the thunk, but I'm not sure why.

src/redux-middleware/fetchBeforeMove.ts
```diff
@@ -0,0 +1,36 @@
+import { ThunkMiddleware } from 'redux-thunk'
+import { pullDescendants } from '../action-creators/pull'
+import { isPending, pathExists } from '../selectors'
+import { hashContext, head, parentOf, pathToContext } from '../util'
+import { State } from '../util/initialState'
+
+/**
+ * Middleware that fetches any pending descendants before a thought-move operation.
+ */
+const fetchBeforeMove: ThunkMiddleware<State> = ({ getState, dispatch }) => {
```
raineorshine, line 10: Let's call this `pullBeforeMove`

src/redux-middleware/fetchBeforeMove.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  return next => async action => {
+
+    // reset internal state variables when clear action is dispatched
```
raineorshine, line 14: Comment doesn't make sense...?

src/redux-middleware/fetchBeforeMove.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (action.type === 'existingThoughtMove') {
+      const state = getState()
+
+      const oldContext = pathToContext(action.oldPath)
+
+      const newContext = pathToContext(action.newPath)
```
raineorshine, line 20: You should simplify the Paths before converting to Contexts. Otherwise it will break context views. All Contexts must be simple, we just don't have that built into the `pathToContext` type yet.

raineorshine (changes requested):

src/action-creators/pull.ts
anmolarora1: We need to have the ability to load the descendant thoughts `synchronously` before calling `existingThoughtMove` reducer. I noticed that we're using `await dispatch` for the `pull` action in some places but it doesn't seem to wait for the thoughts to be loaded (unless that's the intended behavior, we'd need to update those occurrences as well). 
So I separated that function as async and exported the same.

raineorshine: > We need to have the ability to load the descendant thoughts `synchronously` before calling `existingThoughtMove` reducer.

It's impossible to load thoughts synchronously; accessing Firebase or IndexedDB is inherently asynchronous.

> I noticed that we're using `await dispatch` for the `pull` action in some places but it doesn't seem to wait for the thoughts to be loaded (unless that's the intended behavior, we'd need to update those occurrences as well).

The `pull` action creator awaits in several places, so it seems like `await dispatch(pull(...))` would wait for the thoughts to be loaded. If not, that could be a bug.

src/redux-middleware/pullBeforeMove.ts
```diff
@@ -0,0 +1,35 @@
...
+import { isPending, pathExists, simplifyPath } from '../selectors'
+import { hashContext, head, parentOf, pathToContext } from '../util'
+import { State } from '../util/initialState'
+
+/**
+ * Middleware that fetches any pending descendants before a thought-move operation.
+ */
+const pullBeforeMove: ThunkMiddleware<State> = ({ getState, dispatch }) => {
+
+  return next => async action => {
```
raineorshine, line 12: Could you explain how asynchronous middleware works? What happens to other actions that have been dispatched while this is fetching?

src/redux-middleware/pullBeforeMove.ts
```diff
@@ -0,0 +1,35 @@
...
+
+      const newContext = pathToContext(simplifyPath(state, action.newPath))
+
+      const updatedNewContext = [...parentOf(newContext), head(oldContext)]
+      const conflictedPath = pathExists(state, updatedNewContext)
+
+      const isNewContextPending = conflictedPath && isPending(state, newContext)
+      if (isNewContextPending) {
+        const toPull = { [hashContext(newContext)]: newContext }
+        await dispatch(pull(toPull, { maxDepth: Infinity }))
```
raineorshine, line 27: We'll want to limit the depth of the pull, otherwise there could be a large number of thoughts pulled unnecessarily.

src/redux-middleware/pullBeforeMove.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    if (action.type === 'existingThoughtMove') {
+      const state = getState()
+
+      const oldContext = pathToContext(simplifyPath(state, action.oldPath))
+
+      const newContext = pathToContext(simplifyPath(state, action.newPath))
+
+      const updatedNewContext = [...parentOf(newContext), head(oldContext)]
+      const conflictedPath = pathExists(state, updatedNewContext)
```
raineorshine, line 22: This check is a bit too narrow. The destination could be pending, but also a descendant of the destination could be pending which could also be a conflict. I have added a failing unit test in e40d11f90 for clarity.

src/redux-middleware/pullBeforeMove.ts
```diff
@@ -0,0 +1,35 @@
+import { ThunkMiddleware } from 'redux-thunk'
+import pull from '../action-creators/pull'
+import { isPending, pathExists, simplifyPath } from '../selectors'
+import { hashContext, head, parentOf, pathToContext } from '../util'
+import { State } from '../util/initialState'
+
+/**
+ * Middleware that fetches any pending descendants before a thought-move operation.
+ */
+const pullBeforeMove: ThunkMiddleware<State> = ({ getState, dispatch }) => {
```
raineorshine, line 10: Does this replace `flushMoves`? Or is that still needed?

https://github.com/cybersemics/em/blob/18a3dc9b98386e4225b7e30dd5a8577ceeb0721e/src/redux-middleware/pushQueue.ts#L107-L133

raineorshine (changes requested):

src/redux-middleware/pullBeforeMove.ts
anmolarora1: It doesn't block the other actions that are dispatched while it's fetching

raineorshine:
> It doesn't block the other actions that are dispatched while it's fetching

Okay, thanks. I have a couple more questions.

What happens if a subsequent action modifies the thought structure before the pull completes?

Second, what is the user experience during this pull? Is there going to be a delay between the time they activate the move and the time they see changes to the thought structure? For example, with the `pendingMoves` logic the loaded thoughts are immediately moved and the pending source thoughts, which are generally not visible to the user, are deferred to `flushMoves`.

src/redux-middleware/pullBeforeMove.ts
anmolarora1: In order to optimize this and pull only until a certain depth, we need to know the full conflicting path, which cannot be determined without loading the source context's tree, which is [loaded after the reducer](https://github.com/cybersemics/em/blob/a44a411697efb8112440386c3f4cb7f343bca746/src/reducers/existingThoughtMove.ts#L359) `existingThoughtMove` gets triggered.
For instance, 
`
 - a
    - b
      - c
        - d1
        - d2
          - e1 (pending)
          - e2 (pending)
 - p
    - b
      - c
        - d2
          - e3 (pending)
          - e4 (pending 
`

At the time of running the middleware, we'd only know about the conflict `['p', 'b', 'c']`, and any conflicts beyond that level cannot be predicted, unless we load the source path in the middleware. 
Let me know what you think.

src/redux-middleware/pullBeforeMove.ts
anmolarora1: Thanks, I'll address this

anmolarora1:
> > It doesn't block the other actions that are dispatched while it's fetching
> 
> Okay, thanks. I have a couple more questions.
> 
> What happens if a subsequent action modifies the thought structure before the pull completes?
>
The thought won't get moved

 
> Second, what is the user experience during this pull? Is there going to be a delay between the time they activate the move and the time they see changes to the thought structure? For example, with the `pendingMoves` logic the loaded thoughts are immediately moved and the pending source thoughts, which are generally not visible to the user, are deferred to `flushMoves`.

Yes, there's a delay until the thoughts get fetched. I haven't yet looked into the `flushMoves` so I'll check that to see if we can use that to pull hidden descendant thoughts

src/redux-middleware/pullBeforeMove.ts
anmolarora1: `pendingMoves` is being used for pulling the [source thoughts](https://github.com/cybersemics/em/blob/a44a411697efb8112440386c3f4cb7f343bca746/src/reducers/existingThoughtMove.ts#L359) that are not yet loaded, so we can't replace that yet

src/redux-middleware/pullBeforeMove.ts
raineorshine: Okay, that makes sense. If there is a deep conflict we need to pull source and destination.

There is a case where the source is very shallow (e.g. 2 levels) and the destination is very deep (e.g. 10 levels). In that case, it would be best to set `maxDepth: 2` since we know there can be no conflicts beyond that.

This may sound like an edge case, but it's actually a pretty common use case in my experience. Let's say I'm moving a few thoughts to a different part of my thoughtspace. I will probably have to move them up a few levels and even into the root before moving them back down the hierarchy to where I wanted them (although Bijay just finished up expand-on-drag which makes it possible to do in a single move). Pulling the root with `maxDepth: Infinity` will crash the app.

We may have to pull the source, count the depth, then pull the destination up to that depth.

src/redux-middleware/pullBeforeMove.ts
raineorshine: It would take some refactoring, but is it possible to handle source and destination pulling using the same mechanism? That would be much cleaner than having two completely separate mechanisms.

raineorshine:
> > > It doesn't block the other actions that are dispatched while it's fetching
> > 
> > 
> > Okay, thanks. I have a couple more questions.
> > What happens if a subsequent action modifies the thought structure before the pull completes?
> 
> The thought won't get moved

Could you explain more what you mean here? Do you mean that something will cancel the move?

> > Second, what is the user experience during this pull? Is there going to be a delay between the time they activate the move and the time they see changes to the thought structure? For example, with the `pendingMoves` logic the loaded thoughts are immediately moved and the pending source thoughts, which are generally not visible to the user, are deferred to `flushMoves`.
> 
> Yes, there's a delay until the thoughts get fetched. Currently, `flushMoves` is being used to pull pending thoughts from the source and I'd need to see if we can use it to do the same for descendants. From what I remember, we started with [a similar approach](https://github.com/cybersemics/em/pull/1050) and ended up with rank conflicts, that's why we switched to the middleware solution.

I've given this some thought and unfortunately I think this would be too big an impact on the user experience. We need to find a way to immediately move non-pending thoughts and pull only pending thoughts.

raineorshine:
The ideal solution would be #495, because then there would be no pending descendants to conflict with, but that is such a massive refactor I’m not sure if it’s realistic right now. It’s a tricky situation. Open to ideas!

raineorshine:
Fixed by #1050 instead

PR #1067, opened 2021-04-03
adds remoteSync middleware for real-time firebase pull

### LINKED ISSUE #932: Real-time Firebase sync
When the previous syncing mechanism was replaced with the `pullQueue`, subscribing to changes from Firebase was removed. This needs to be added back in so that if the user has **em** open on two or more devices, their changes are sync'd in real-time. We should be able to use `reconcile` to handle merging, so it's just the subscription logic that needs to be added.

I can think of two approaches which have different performance characteristics. Please assess and provide a recommendation.

## 1.

- Subscribe to changes from all [visible thoughts](https://github.com/cybersemics/em/blob/dev/src/redux-middleware/pullQueue.ts#L22-L41)
- When the visible thoughts change, subscribe to newly visible thoughts and unsubscribe from thoughts that are no longer visible.
- **Advantage:** Only subscribe to visible thoughts; less bandwidth usage. Optimized for multi-user editing.
- **Disadvantage:** Managing a large number of subscriptions is complex and uses additional memory.

## 2.

- Subscribe to all `thoughtIndex` and `contextIndex` changes
- Ignore changes to thoughts that are not visible.
- **Advantage:** Simple, low memory subscription, less complex code. Optimized for single-user editing.
- **Disadvantage:** Increased bandwidth and processing for thoughts that are never rendered.

anmolarora1:
fixes #932

### DIFF
--- a/src/redux-middleware/remoteSync.ts
+++ b/src/redux-middleware/remoteSync.ts
@@ -1,0 +1,121 @@
+import _ from 'lodash'
+import { ThunkMiddleware } from 'redux-thunk'
+import { decodeContextUrl, expandThoughts } from '../selectors'
+import { equalArrays, hashContext, hashThought, head, keyValueBy, pathToContext } from '../util'
+import { State } from '../util/initialState'
+import { Index, Context, Ref, Parent, Lexeme } from '../types'
+import { reconcile } from '../action-creators'
+import getFirebaseProvider from '../data-providers/firebase'
+
+/** Generates a map of all visible contexts, including the cursor, all its ancestors, and the expanded contexts. */
+const getVisibleContexts = (state: State, expandedContexts: Index<Context>): Index<Context> => {
+
+  const { cursor } = state
+
+  // if there is no cursor, decode the url so the cursor can be loaded
+  // after loading the ranks will be inferred to update the cursor
+  const contextUrl = decodeContextUrl(state, window.location.pathname)
+  const contextCursor = cursor ? pathToContext(cursor) : contextUrl
+
+  return {
+    ...expandedContexts,
+    // generate the cursor and all its ancestors
+    // i.e. ['a', b', 'c'], ['a', 'b'], ['a']
+    ...keyValueBy(contextCursor, (value, i) => {
+      const subcontext = contextCursor.slice(0, contextCursor.length - i)
+      return subcontext.length > 0 ? { [hashContext(subcontext)]: subcontext } : null
+    }),
+  }
+}
+
+/**
+ * Middleware to auto-sync thoughts from firebase. 2 major functions that this serves
+ * 1) Maintains the list(thoughtListeners) of listeners for visible contexts, while deactivating the ones that are hidden.
+ * 2) Builds the thoughtIndex by fetching all lexemes and triggers reconcile with thought and context index updates.
+ */
+const remoteSyncMiddleware: ThunkMiddleware<State> = ({ getState, dispatch }) => {
+
+  let thoughtListeners: Index<Ref> = {}
+  let lastExpanded: Index<Context> = {} // eslint-disable-line fp/no-let
+  let thoughts: (Lexeme | undefined)[] = []// eslint-disable-line fp/no-let
+
+  return next => async action => {
+
+    next(action)
+    if (action.type !== 'existingThoughtChange') return
+
+    const state = getState()
+
+    if (!state.authenticated || !state?.user?.uid) return
+
+    const expandedContexts = expandThoughts(state, state.cursor, {
+      returnContexts: true,
+    })
+
+    const visibleContexts = getVisibleContexts(state, expandedContexts)
+    const visibleContextIds = Object.keys(visibleContexts)
+
+    if (Object.keys(state.expanded).length > 0
+    && equalArrays(Object.keys(expandedContexts), Object.keys(lastExpanded))) return
+
+    lastExpanded = expandedContexts
+
+    const listenersToAdd = _.difference(visibleContextIds, Object.keys(thoughtListeners))
+    const listenersToRemove = _.difference(Object.keys(thoughtListeners), visibleContextIds)
+
+    const refsToAdd: Index<Ref> = listenersToAdd.reduce((acc, id) => {
+      const url = `users/${state.user!.uid}/contextIndex/${id}`
+      return { ...acc, [id]: window.firebase?.database().ref(url) }
+    }, {} as Index<Ref>)
+
+    const refsToRemove = listenersToRemove.reduce((acc, id) => {
+      return { ...acc, [id]: thoughtListeners[id] }
+    }, {} as Index<Ref>)
+    const provider = getFirebaseProvider(getState(), dispatch)
+
+    Object.keys(refsToRemove).forEach(id => {
+      refsToRemove[id]?.off('value')
+    })
+
+    Object.keys(refsToAdd).forEach(id => {
+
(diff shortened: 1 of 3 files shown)


src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+import { ThunkMiddleware } from 'redux-thunk'
+import { decodeContextUrl, expandThoughts } from '../selectors'
+import { equalArrays, hashContext, hashThought, head, keyValueBy, pathToContext } from '../util'
+import { State } from '../util/initialState'
+import { Index, Context, Ref, Parent, Lexeme } from '../types'
+import { reconcile } from '../action-creators'
+import getFirebaseProvider from '../data-providers/firebase'
+
+/** Generates a map of all visible contexts, including the cursor, all its ancestors, and the expanded contexts. */
+const getVisibleContexts = (state: State, expandedContexts: Index<Context>): Index<Context> => {
```
raineorshine, line 11: Is this identical to `getVisibleContexts` in `pullQueue`? If so, it should be factored out to avoid duplication.

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+    }),
+  }
+}
+
+/**
+ * Middleware to auto-sync thoughts from firebase. 2 major functions that this serves
+ * 1) Maintains the list(thoughtListeners) of listeners for visible contexts, while deactivating the ones that are hidden.
+ * 2) Builds the thoughtIndex by fetching all lexemes and triggers reconcile with thought and context index updates.
+ */
+const remoteSyncMiddleware: ThunkMiddleware<State> = ({ getState, dispatch }) => {
```
raineorshine, line 36: Let's use the term `remoteSubscribe` rather than `remoteSync` since it handles subscribing to remote changes and sync could also refer to pull or push.

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  let thoughtListeners: Index<Ref> = {}
+  let lastExpanded: Index<Context> = {} // eslint-disable-line fp/no-let
+  let thoughts: (Lexeme | undefined)[] = []// eslint-disable-line fp/no-let
```
raineorshine, line 40: These could use some descriptive comments to help developers understand the local state of this middleware.

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+    const visibleContexts = getVisibleContexts(state, expandedContexts)
+    const visibleContextIds = Object.keys(visibleContexts)
+
+    if (Object.keys(state.expanded).length > 0
+    && equalArrays(Object.keys(expandedContexts), Object.keys(lastExpanded))) return
+
+    lastExpanded = expandedContexts
+
+    const listenersToAdd = _.difference(visibleContextIds, Object.keys(thoughtListeners))
+    const listenersToRemove = _.difference(Object.keys(thoughtListeners), visibleContextIds)
```
raineorshine, line 64: This is nice and clear

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const refsToAdd: Index<Ref> = listenersToAdd.reduce((acc, id) => {
+      const url = `users/${state.user!.uid}/contextIndex/${id}`
+      return { ...acc, [id]: window.firebase?.database().ref(url) }
+    }, {} as Index<Ref>)
+
+    const refsToRemove = listenersToRemove.reduce((acc, id) => {
+      return { ...acc, [id]: thoughtListeners[id] }
+    }, {} as Index<Ref>)
+    const provider = getFirebaseProvider(getState(), dispatch)
```
raineorshine, line 74: Let's move the `provider` declaration up next to `listenersToRemove` so that variable declarations are grouped together.

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+      refsToRemove[id]?.off('value')
+    })
+
+    Object.keys(refsToAdd).forEach(id => {
+
+      refsToAdd[id]?.on('value', _.debounce(async snapshot => {
+
+        const state = getState()
+
+        const remoteState: Parent = snapshot.val()
```
raineorshine, line 86: Let's call this variable `parent`

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        if (!remoteState || refsToRemove[id]) return
+
+        const contexts: Context[] = remoteState.children.map(child => [...remoteState.context, child.value])
+        const thoughtIds = contexts.map(context => hashThought(head(context)))
```
raineorshine, line 91: Why generate `contexts` if all you need are the child values? You might be able to simplify this.

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        const thoughtPromises = thoughtIds.map(id => provider.getThoughtById(id))
```
raineorshine, line 93: Use `getThoughtsByIds` for batching efficiency.

src/redux-middleware/remoteSync.ts
raineorshine, line 40: Let's rename `thoughts` to `lexemes` for clarity.

In general we want to use the type-specific names. I just use "thoughts" when speaking informally or if it is clear from the context.

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        thoughts = await Promise.all(thoughtPromises)
+        const thoughtindex: Index<Lexeme> = thoughts.reduce((acc, thought, i) => {
+          return { ...acc, ...thought ? { [thoughtIds[i]]: { ...thought, id: thoughtIds[i] } } : false }
```
raineorshine, line 97: Prefer `null` instead of `false` for this use

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+        const remoteState: Parent = snapshot.val()
+
+        if (!remoteState || refsToRemove[id]) return
+
+        const contexts: Context[] = remoteState.children.map(child => [...remoteState.context, child.value])
+        const thoughtIds = contexts.map(context => hashThought(head(context)))
+
+        const thoughtPromises = thoughtIds.map(id => provider.getThoughtById(id))
+
+        thoughts = await Promise.all(thoughtPromises)
```
raineorshine, line 95: Mutation is pretty ugly here. You may want to consider changing the `forEach` into a `map`.

src/redux-middleware/remoteSync.ts
```diff
@@ -0,0 +1,121 @@
...
+          contextIndex: { [id]: remoteState },
+          thoughtCache: [],
+          contextCache: []
+        }] }))
+
+      }, 500))
+
+    })
+
+    thoughtListeners = _.omit({ ...thoughtListeners, ...refsToAdd }, listenersToRemove)
```
raineorshine, line 116: I'm on the fence about whether to allow this mutation. It's something we're really trying to avoid on this project.

raineorshine (changes requested):
Comments mostly refer to name changes for clarity. Logic is clear and succinct, though I haven't tested it out so I look forward to hearing from you how well it's working! Thanks!!

PR #1068, opened 2021-04-03
Add puppeteer tests

### LINKED ISSUE #937: Puppeteer caret tests
The caret behavior is critical in **em** given the precise editing controls. Since JSDOM does not fully support browser selection (and may never model it perfectly accurately), we should test the caret behavior on a headless browser.

- In a test, open the app in [Puppeteer](https://github.com/puppeteer/puppeteer) and test the caret position. A basic test will consist of roughly the following (from an Enzyme test that does not work consistently):

```js
  windowEvent('keydown', { key: 'Enter' })
  jest.runOnlyPendingTimers()
  wrapper.update()
  const { focusNode, focusOffset } = window.getSelection() || {}
  expect(focusNode?.textContent).toEqual('')
  expect(focusOffset).toEqual(0)
```

- You may have to use [puppeteer-direct](https://github.com/wix-playground/puppeteer-direct) to tunnel the selection information from the browser back to the test.

shresthabijay:
fixes #937 

# Changes
- Add puppeter environment for jest.
- Add basic caret tests.
- Add helper functions to handle clicking editable and bullets.


shresthabijay:
@raineorshine I have added puppeteer and caret tests. It completely uses mouse click to simulate setting caret in the desired position in `Editable`. I will add further tests including caret behavior of `Notes` and simulating mobile after you review the changes. Thanks!

src/e2e/puppeteer-environment.js
```diff
@@ -0,0 +1,73 @@
...
+const delay = ms => new Promise(resolve => setTimeout(() => resolve(true), ms))
+
+/** */
+class PuppeteerEnvironment extends JsDomEnvironment {
+  constructor(config) {
+    super(config)
+  }
+
+  async setup() {
+    console.log(chalk.yellow('Setup Test Environment.'))
```
raineorshine, line 23: Use `console.info` for informational logging and preserve `console.log` for debugging.

src/e2e/puppeteer-environment.js
```diff
@@ -0,0 +1,73 @@
...
+  async teardown() {
+    console.log(chalk.yellow('Teardown Test Environment.'))
+    await this.global.browser.close()
+    await teardownDevServer()
+    await super.teardown()
+  }
+
+  runScript(script) {
+    return super.runScript(script)
+  }
```
raineorshine, line 70: Unused

src/initialize.ts
```diff
@@ -43,10 +44,23 @@ export const initialize = async () => {
 const withState = <T, R>(f: (state: State, ...args: T[]) => R) =>
   (...args: T[]) => f(store.getState(), ...args)
 
+/** Set cursor function that is exposed to window for puppeteer testing. */
+const setCursorFirstMatch = (unrankedPath: string[]) => store.dispatch(setCursorFirstMatchActionCreator(unrankedPath))
+
+/** Imports text on given unranked path. Used in puppeteer testing. */
+const importTextFirstMatch = (unrankedPath: string[], text: string) => store.dispatch(importOnFirstMatchPathActionCreator(unrankedPath, text))
```
raineorshine, line 51: Maybe rename `importTextFirstMatch` to `paste` to make the tests more readable and less implementation-specific?

src/test-helpers/delay.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Delay for tests.
+ */
+export const delay = (ms: number) => new Promise<boolean>(resolve => setTimeout(() => resolve(true), ms))
```
raineorshine, line 4: You can pass `resolve` directly to `setTimeout` for a slightly more compact implementation, i.e. `setTimeout(resolve)`

src/test-helpers/e2e-helpers/getEditableContextFromSelection.ts
```diff
@@ -0,0 +1,23 @@
+import { Page } from 'puppeteer'
+
+/**
+ * Get editable context hash from the selection.
+ */
+const getEditableContextHashFromSelection = (page: Page) => page.evaluate(() => {
```
raineorshine, line 6: I'm going to make a note here to come back to this after the other changes are done. Maybe we leave this as-is, or maybe we find a way to do this without context hashes.

raineorshine (changes requested):
Thanks @shresthabijay! This looks great. I like your approach and am happy to see these tests working!

Most of my comments are related to renaming, following RTL guidelines, or attempting to tighten up the API by removing implementation details.

.github/workflows/test.yml
```diff
@@ -32,5 +32,8 @@ jobs:
       - name: Lint
         run: npm run lint
 
+      - name: Build
+        run: npm run build
+
```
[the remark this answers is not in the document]
shresthabijay: Sure. We can seprate it into e2e workflow. It will be fater to run tests locally.

src/e2e/__tests__/caret.ts
```diff
@@ -0,0 +1,190 @@
...
+    await testHelpers.clearAll()
+    await localStorage.clear()
+  })
+
+  await page.reload({
+    waitUntil: 'load'
+  })
+
+  // wait for alerts to get dismissed
+  await delay(50)
```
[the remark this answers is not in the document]
shresthabijay: I added a callback to dismiss alerts on page. However it didn't seem to work properly unless small delay is added after page reload.

src/components/Thought.tsx
```diff
@@ -536,6 +536,7 @@ const ThoughtContainer = ({
     'child-divider': isDivider(thought!.value ?? ''),
     'cursor-parent': isCursorParent,
     'cursor-grandparent': isCursorGrandparent,
+    [`context-${hashContext(pathToContext(path), rank)}`]: true,
```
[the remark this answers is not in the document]
shresthabijay: Yes indeed. We can implement searching ediatble node by text content and then find the associated bullet node. However we won't be able to differentiate between two nodes with same text content.

src/initialize.ts
shresthabijay: Sure.

src/test-helpers/delay.ts
shresthabijay: Will do that. Thanks!

src/test-helpers/e2e-helpers/setCursorAndSelection.ts
```diff
@@ -0,0 +1,33 @@
...
+  offset?: number,
+  end?: boolean,
+}
+
+/**
+ * Set cursor to the first matched path and set it's selection in puppeteer environment.
+ */
+const setCursorAndSelection = async (page: Page, {
+  unrankedPath, offset, end
+}: Options) => {
```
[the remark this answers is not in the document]
shresthabijay: Sure.

src/e2e/__tests__/caret.ts
```diff
@@ -0,0 +1,190 @@
...
+    })
+
+    await page.keyboard.press('ArrowDown')
+
+    const focusedEditableContextHash = await getEditableContextHashFromSelection(page)
+
+    const contextHash = hashContext(['Dogs', 'Labrador'], 1)
+
+    // the focus must be in Dogs/Labrador after cursor down
+    expect(contextHash).toBe(focusedEditableContextHash)
```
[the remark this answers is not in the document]
shresthabijay: Ahh. I too think we can simply remove `getEditableContextHashFromSelection`.

src/test-helpers/e2e-helpers/clickEditable.ts
```diff
@@ -0,0 +1,48 @@
...
+    range.setStart(textNode, offset)
+    const { right, top, height } = range.getBoundingClientRect()
+    return {
+      x: right,
+      y: top + (height / 2)
+    }
+  }, editableNode, offset || 0)
+
+  const coordinate = !offset ? {
+    x: boundingBox.x + (horizontalClickLine === 'left' ? 0 : horizontalClickLine === 'right' ? boundingBox.width : boundingBox.width / 2),
```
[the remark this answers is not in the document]
shresthabijay: The first one. It click on the left edge of the `Editable` component.

src/components/Thought.tsx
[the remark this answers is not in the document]
raineorshine: That will work for normal view 99% of the time. For context view we will have to have a different approach since the same thought can appear on the screen more than once. 

I’m flexible here, just wanted you to give it some thought and see if there was an easy way to improve the testing interface.

src/test-helpers/e2e-helpers/clickEditable.ts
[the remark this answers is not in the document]
raineorshine: Ah, must have been a miscommunication. The test case is the other. It should click to the left of the thought, outside the editable.

.github/workflows/test.yml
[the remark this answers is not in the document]
raineorshine: We still need to run e2e tests locally.

src/e2e/__tests__/caret.ts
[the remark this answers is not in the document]
raineorshine: So just to be clear, you found that alerts persist across page reloads?

shresthabijay:
@raineorshine I have made the changes. I have simplified some testing interface by removing dependency with `context` hash class to the `Thought`. Instead we can now search `Editable` node from text. However it won't work for view with multiple thoughts wirth same value. Since most of the test are simple we don't have to worry about it rn.

src/test-helpers/e2e-helpers/clickWithOffset.ts
```diff
@@ -0,0 +1,42 @@
...
+    range.setStart(textNode, offset)
+    const { right, top, height } = range.getBoundingClientRect()
+    return {
+      x: right,
+      y: top + (height / 2)
+    }
+  }, nodeHandle, offset || 0)
+
+  const coordinate = !offset ? {
+    x: boundingBox.x + (horizontalClickLine === 'left' ? 0 : horizontalClickLine === 'right' ? boundingBox.width : boundingBox.width / 2),
```
raineorshine, line 33: It looks like "left" still clicks on the left edge of the element instead of to the left of the element (outside). Likewise for "right".

raineorshine (commented):
Tests read really nice now! I'm happy with the structure.

src/test-helpers/e2e-helpers/clickWithOffset.ts
raineorshine: I added a `HORIZONTAL_OFFSET` in 2135e4ced but the "right" test now fails. Can you investigate?

shresthabijay: @raineorshine I am a little confused here. By clicking right of the thought. Do you mean to the right(outside) of the `Editable`? If that is the case it activates `cursorUp` instead of caret being set at the end of the editable ?

raineorshine: I don’t think it should activate cursorUp when clicking to the right of a thought. Does the Editable itself go all the way to the right edge? I thought it might have been a container element or ThoughtAnnotation. I’m not at my computer so I can’t test. 

On mobile the caret correctly goes to the end of the thought if the cursor was previously on a sibling. If the cursor is not on a sibling, it incorrectly puts the cursor at the beginning.

shresthabijay: It's working on mobile. However, I still cannot get consistent results. It's putting caret at the beginning sometimes. Should the caret always be positioned at the end of the thought if clicked anywhere right (outside) of the editable? Or does it depend on where the previous cursor was?

shresthabijay: https://user-images.githubusercontent.com/25879100/114599822-55f30100-9cb3-11eb-8b5d-81ba045c4039.mov

shresthabijay: I am using a simulator here. I have highlighted the editable. I am getting inconsistent caret placement on clicking on the right side of the editable. Am I missing something here ?

raineorshine: > It's working on mobile.

On iOS, the caret goes to the beginning of a thought instead of the end when the cursor is on a non-sibling before the tap. I will create a separate issue for this, though I remember there being some complication with it when I tried to solve it before.

> Should the caret always be positioned at the end of the thought if clicked anywhere right (outside) of the editable? 

If user clicks slightly to the right of the text, it should definitely place the caret at the end of the editable. For the dead space way to the right, I'm not sure, but I've never clicked there and expected something different to happen, so I say leave it for now. I added `.skip` to the test for clicking to the right.

> Or does it depend on where the previous cursor was?

The caret placement should never depend on where the cursor was previously (except for determining edit mode).

> I am getting inconsistent caret placement on clicking on the right side of the editable. Am I missing something here ?

I can't explain the inconsistent behavior, but I think it's okay to ignore it since that has not caused any UX issues that I have noticed.

Issue #1080, opened 2021-04-08
Sort does not work in HOME context


raineorshine:
## Steps to Reproduce

- b
- a

1. Click on an empty space to set the `cursor` to `null`. 
2. Activate `sortContext`

## Current Behavior

Nothing happens

## Expected Behavior

- a
- b

PR #1081, opened 2021-04-08
remove 'the' preposition from comparator inputs

### LINKED ISSUE #1079: Ignore “the “ when sorting
*Case insensitive

## Steps to Reproduce

1. Create a thought
2. Activate `sortContext`
3. Import the following thoughts as children:

- a
- theatre
- d
- the c
- b

## Current Behavior

- a
- b
- d
- the c
- theatre

## Expected Behavior

- a
- b
- the c
- d
- theatre

onurpolattimur:
fixes #1079 

Notes: This doesn't work for home context. There is already opened issue related to this #1080

### DIFF
--- a/src/util/compareThought.ts
+++ b/src/util/compareThought.ts
@@ -5,9 +5,22 @@ import { EMOJI_REGEX } from '../constants'
 const regexPunctuation = /^[!@#$%^&*()\-_=+[\]{};:'"<>.,?\\/].*/
 const regexShortDateWithDash = /\d{1,2}-\d{1,2}/
 const regexShortDateWithSlash = /\d{1,2}\/\d{1,2}/
+const ignoredPrefixes = ['the ']
 
 /** Remove emojis and trailing/leading spaces from camparator inputs using regex. */
 const removeEmojisAndSpaces = (str: string) => str.replace(EMOJI_REGEX, '').trim()
+
+/** Remove ignored prefixes from comparator inputs. */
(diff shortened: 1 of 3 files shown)


src/util/compareThought.ts
```diff
@@ -5,10 +5,23 @@ import { EMOJI_REGEX } from '../constants'
...
 const regexShortDateWithSlash = /\d{1,2}\/\d{1,2}/
+const ignoredPrefixes = ['the ']
 
 /** Remove emojis and trailing/leading spaces from camparator inputs using regex. */
 const removeEmojisAndSpaces = (str: string) => str.replace(EMOJI_REGEX, '').trim()
 
+/** Remove ignored prefixes from comparator inputs. */
+const removeIgnoredPrefixes = (str: string) => {
+  const ignoredKeywords = ignoredPrefixes.join('|')
+  const regex = new RegExp(`^(${ignoredKeywords})(.*)`, 'gm')
```
raineorshine, line 16: You can move the regex to the module scope since it does not change between invocations. Regex instantiation is not trivial like object literals.

raineorshine (changes requested):
Thanks for the contribution! I like your approach.

raineorshine:

src/util/compareThought.ts
onurpolattimur: Yes, you are right. Also I have moved the ignoredPrefixes variable to `constants.ts`. Because one day you may need to access these ignored prefixes.

PR #1082, opened 2021-04-09
Trigger local pull only when multiple sessions(browser tabs) are active

### LINKED ISSUE #1071: Thoughts duplicated when typing fast
The thoughts get duplicated in the state when typing too fast. For some reason, the duplicate thought doesn't show up in the DOM but can be seen in the redux state. 
[Here's the video](https://www.awesomescreenshot.com/video/3274829?key=9dc6c6ad6cd5dfeda1da928875229116) capturing this issue.

anmolarora1:
fixes #1071

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { Provider } from 'react-redux'
 import { store } from '../store'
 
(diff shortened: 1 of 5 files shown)


raineorshine:
Thanks for putting this together! Unless I am mistaken, this fixes the issue when there is just a single tab open, but doesn't address the issue if there are multiple tabs open. I don't think this is the direction I want to go. I'd rather come up with a complete solution that works for multiple tabs.

PR #1083, opened 2021-04-09
Unroot context before getting value of sort attribute

onurpolattimur:
Fixes: #1080 

We need to unroot the context before getting value of sorting attribute for the context because root token is removed from the context when adding. 

https://github.com/cybersemics/em/blob/e2c66a7c0b46127f43c2978145b490aed0e0458b/src/reducers/toggleAttribute.ts#L35

### DIFF
--- a/src/selectors/getSortPreference.ts
+++ b/src/selectors/getSortPreference.ts
@@ -1,10 +1,11 @@
 import { getSetting, getAllChildren } from '../selectors'
 import { State } from '../util/initialState'
 import { Context } from '../types'
+import { unroot } from '../util'
 
 /** Get the sort setting from the given context meta or, if not provided, the global sort. */
 const getSortPreference = (state: State, context: Context) => {
-  const childrenSort = getAllChildren(state, [...context, '=sort'])
+  const childrenSort = getAllChildren(state, [...unroot(context), '=sort'])
   return childrenSort.length > 0
     ? childrenSort[0].value
     : getSetting(state, ['Global Sort']) || 'None'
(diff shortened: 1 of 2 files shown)


raineorshine (changes requested):
Thank you! Could you add the test to https://github.com/cybersemics/em/blob/dev/src/shortcuts/__tests__/toggleSort.ts?

raineorshine:
Oops, I just realized that a pure state test won't suffice (it passes even if the change is reverted). We need to test the DOM since that is where the bug is.

I would suggest an RTL test like in https://github.com/cybersemics/em/blob/dev/src/shortcuts/__tests__/extractThought.ts which will allow you to test the order of the subthoughts. Is this something you'd like to try, or should I add it?

onurpolattimur:
> Oops, I just realized that a pure state test won't suffice (it passes even if the change is reverted). We need to test the DOM since that is where the bug is.
> 
> I would suggest an RTL test like in https://github.com/cybersemics/em/blob/dev/src/shortcuts/__tests__/extractThought.ts which will allow you to test the order of the subthoughts. Is this something you'd like to try, or should I add it?

Yes, you are right. I wrote the RTL tests but  I didn't find any RTL tests regarding sorting. Where should we put the tests?

raineorshine:
You can put them in `/src/shortcuts/__tests__/toggleSort.ts` with the non-RTL tests. Thanks!

onurpolattimur:
I have added the tests. Can you review it please?

raineorshine:
Thank you! The tests are still giving me a false positive when I revert them. Any ideas why? They look good to me, and they're clearly testing the sort order in the DOM.

onurpolattimur:
Are you sure? Which file did you revert? When I revert the changes in `getSortPreference.ts`, the home context RTL test ('thoughts are sorted alphabetically') is failed.

raineorshine:
Ah, I'm glad I asked! I must have run the wrong test or it didn't build properly. I was able to fail it this time! Tests work perfectly. (Always amusing when the fix is 1 line of code and the tests are 100 😆.)

Thanks again for your contribution!

PR #1084, opened 2021-04-12
Fix subCategorizeOne in sorted context

### LINKED ISSUE #1076: subcategorizeOne breaks in sorted context
When activating `subcategorizeOne` on a sorted context, the thought is added to a sibling rather than the new empty thought. 

## Steps to Reproduce

``` 
- A
  - =sort
    - Alphabetical
  - D
  - B
  - C
  - E
``` 

1. Set the cursor in `E`
2. Activate `subcategorizeOne`

## Current Behavior

``` 
- A
  - =sort
    - Alphabetical
  - D
    - E
  - B
  - C
``` 

## Expected Behavior

``` 
- A
  - =sort
    - Alphabetical
  - D
  - B
  - C
  - 
    - E
  ```

### PR DESCRIPTION
**Pull request by shresthabijay**
fixes #1076

### CODE REVIEW

**Comment by shresthabijay** [comment]

File: src/selectors/getThoughtBefore.ts [hunk 1]
```diff
@@ -1,16 +1,16 @@
...
 
 /** Gets a new rank before the given thought in a list but after the previous thought. */
-const getThoughtBefore = (state: State, simplePath: SimplePath) => {
+const getThoughtBefore = (state: State, simplePath: SimplePath, sort = true) => {
 
   const value = headValue(simplePath)
   const rank = headRank(simplePath)
   const parentPath = rootedParentOf(state, simplePath)
-  const children = getChildrenSorted(state, pathToContext(parentPath))
+  const children = (sort ? getChildrenSorted : getVisibleChildrenRanked)(state, pathToContext(parentPath))
```
raineorshine, line 13: `getChildrenSorted` already checks the sort preference and calls `getChildrenSortedAlphabetical` if needed. Why isn't that working in this case? If you could investigate and provide an explanation for that it will help me better assess the proposed solution. Thanks!

shresthabijay: We cannot use sorted children to find the thought before the cursor. Since the empty thought will always be at the beginning of the sorted children array, we will get the wrong thought at the thought before the cursor. On `subCategorizeOne` we first add the empty thought before the cursor. Since the sort preference of the context is alphabetical, `getThoughtBefore` doesn't give the empty child as the result. So in short we always need to use ranked children instead of sorted children for `subCategorizeOne`.

raineorshine: I see, thank you, that makes sense. So I'm thinking of rather than parameterizing `getThoughtBefore` (which increases the surface area of the internal API), I wonder if there is a better way to determine the path to the newly created thought in `subcategorizeOne`. `getThoughtBefore` is kind of a hacky way to figure out the thought that was just created since `newThought` is a reducer and doesn't return the newly created thought itself.

Is there a way to pass information from one reducer to another in a reducer flow? Would adding something like `_newThought` to the state just for this purpose be even more hacky? 

Aside: This reminds me of database functions that create new records. They usually return the id of the new record for similar purposes.

shresthabijay: > I see, thank you, that makes sense. So I'm thinking of rather than parameterizing `getThoughtBefore` (which increases the surface area of the internal API), I wonder if there is a better way to determine the path to the newly created thought in `subcategorizeOne`. `getThoughtBefore` is kind of a hacky way to figure out the thought that was just created since `newThought` is a reducer and doesn't return the newly created thought itself.
> 
> Is there a way to pass information from one reducer to another in a reducer flow? Would adding something like `_newThought` to the state just for this purpose be even more hacky?
> 
> Aside: This reminds me of database functions that create new records. They usually return the id of the new record for similar purposes.

I think adding transient fields like `_newThought` to the state may not be a good idea. As we can end up with too many fields in the state. It would just pollute the `State`.

We can change the interface of the reducer to return an index instead of directly returning the updated state. Something like this.

```javascript

interface ReducerReturn<R> {
 value: R | null,
 updatedState: State
}

function Reducer<T>(state: State, options: Options): ReducerReturn<T>
```

This way reducers will still be pure functions that are responsible to return an updated state. We can easily pass information from one reducer to another in a reducer flow through `value`. All we have to do is to change the interface of all the reducers except the `appReducer`. It should return an updated state by accessing `updatedState` from the return value of the called reducer. What do you think ?

raineorshine: > I think adding transient fields like `_newThought` to the state may not be a good idea. As we can end up with too many fields in the state. It would just pollute the `State`.

Yeah, that's what I suspected.

> We can change the interface of the reducer to return an index instead of directly returning the updated state. Something like this.

Okay, that's an interesting idea.

> All we have to do is to change the interface of all the reducers except the `appReducer`. It should return an updated state by accessing `updatedState` from the return value of the called reducer. What do you think ?

Hmmm not excited about changing the signature on all the reducers. Especially since it's so clean right now.

I think I'll hold off on any architectural changes for now, but if we get many more instances of this need I will consider it. Thanks for discussing!

raineorshine: @shresthabijay I thought of another solution that would only affect this case. Instead of calling `newThought` which determines the rank internally, we can calculate the new rank manually and use the lower level `newThoughtSubmit` reducer. We don't need all the rank cases and tutorial logic anyway. That allows us to know the rank from within `subcategorizeOne` without relying on `getThoughtBefore`. What do you think?

shresthabijay: > @shresthabijay I thought of another solution that would only affect this case. Instead of calling `newThought` which determines the rank internally, we can calculate the new rank manually and use the lower level `newThoughtSubmit` reducer. We don't need all the rank cases and tutorial logic anyway. That allows us to know the rank from within `subcategorizeOne` without relying on `getThoughtBefore`. What do you think?

That's a better approach. We just want need to calculate rank. Should I implement this and push the changes?

raineorshine: Sure, that would be great! Thanks for talking through it!

shresthabijay:

src/reducers/subCategorizeOne.ts
```diff
@@ -14,6 +15,8 @@ import {
   reducerFlow,
   isRoot,
 } from '../util'
+import newThoughtSubmit from './newThoughtSubmit'
+import setCursor from './setCursor'
```
raineorshine, line 19: You can add these to the single import from the reducers index above.

PR #1085, opened 2021-04-12
Some rendering optimization

raineorshine:


raineorshine:
@shresthabijay Can you do a quick review of these render performance optimizations I made? They were easy optimizations, but nothing systematic. I'm not very experienced with React optimizations, so I'm just wondering if there is a better way to do anything. Thanks!

shresthabijay:
> @shresthabijay Can you do a quick review of these render performance optimizations I made? They were easy optimizations, but nothing systematic. I'm not very experienced with React optimizations, so I'm just wondering if there is a better way to do anything. Thanks!

These optimizations look good. I don't see any React specific optimizations that are necessary right away. It would be better if we first profile the app and find unnecessary re-renders. It would be easy to make optimizations that way.

raineorshine:
I totally agree. These were some ad hoc optimizations that I only did because I had a personalized session with a performance expert.

Anmol is going to be spearheading performance after he finishes up his current issues.

Thanks!

Issue #1072, opened 2021-04-07
HTML not rendered in alert


raineorshine:
## Steps to Reproduce

- a
- b

1. Make `b` bold
2. Drag `b` into `a`

## Current Behavior

Alert: `”<b>b</b>“ moved to “a”`

## Expected Behavior

HTML in thoughts should be decoded and rendered as HTML in alerts. 

Alert: ”**b**“ moved to “a”

onurpolattimur:
What is the expected behaviour if length of the bold text is greater than THOUGHT_ELLIPSIZED_CHARS which is currently 16?

For example; 
- **Lorem Ipsum is simply dummy text of the printing and typesetting industry. **

The thought's value is wrapped up `<b> ` tag, but we ellipsize the value before showing alert. In this situation there will be opening tag at the head of the ellipsized message and no closing tag at the end of the ellipsized message. So the full text of the alert will be bold.

https://github.com/cybersemics/em/blob/dcc557e8646ae484c70accce7384a8e40c3df8eb/src/components/Subthoughts.tsx#L223

raineorshine:
Good catch! We would want to close all the open tags after ellipsizing so that they don't leak into the rest of the alert. I think `ellipsize` itself is the right place for that.

PR #1089, opened 2021-04-16
Add truncate library to ellipsize html thoughts properly

onurpolattimur:
Fixes: #1072 

I added `truncate-html` library because otherwise it is really too risky to handle all cases and too complex.

- If we assume the character limit is 16, then user should see exactly 16 character. So html tags shouldn't be counted in this character limit.
Example: 
If the thought's value is `<b>L</b><i>o</i><b>r</b><i>e</i><b>m</b>` then user shouldn't see `<b>L</b><i>o</i>`. Because there is only two character in this value. So user should see `<b>L</b><i>o</i><b>r</b><i>e</i><b>m</b>` 
Rendered value should be: <b>L</b><i>o</i><b>r</b><i>e</i><b>m</b>

- All openings tag should be closed to not affect rest of alert message.
Example: 
If the thought's value is `<b>Lorem Ipsum is simply dummy text.</b>` then user shouldn't see `<b>Lorem Ipsum...` User should see `<b>Lorem Ipsum is...</b>` 
Rendered value should be: <b>Lorem Ipsum is...</b>

- If thought's value contains `<` `>` characters for other purposes than html,  user should see as is.
Example: 
If the thought's value is `Zero < Three > One` then user should see `Zero < Three >...` 
Rendered value should be: Zero < Three >...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2403,7 +2403,6 @@
       "version": "0.22.22",
       "resolved": "https://registry.npmjs.org/@types/cheerio/-/cheerio-0.22.22.tgz",
       "integrity": "sha512-05DYX4zU96IBfZFY+t3Mh88nlwSMtmmzSYaQkKN48T495VV1dkHSah6qYyDTN5ngaS0i0VonH37m+RuzSM0YiA==",
-      "dev": true,
       "requires": {
         "@types/node": "*"
       }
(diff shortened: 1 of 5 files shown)


raineorshine (approved):
Looks great! Yes, I think `truncate-html` is the way to go.

I assume it works when bundled, despite the repo saying it does not work in the browser.

Issue #1086, opened 2021-04-13
Cursor lost on refresh


raineorshine:
Labeled `important` as refreshes occur frequently on the PWA.

## Steps to Reproduce

``` 
- a
- b
``` 

With the cursor on `b', refresh the page.

## Current Behavior

Cursor is not on `b`.

## Expected Behavior

Cursor should be on `b` after refresh. (Cursor should be restored from the URL on refresh.)

## Notes

This is a regression. The logic that supports restoring the cursor is not trivial due to timing issues with the initial load. When the page first loads, the thoughts have not yet been loaded so the cursor cannot be restored. Furthermore, we can't even determine the correct `Path` for the `cursor` until the thoughts are loaded, because the url (intentionally) lacks ranks. This is the original and proper use of `rankThoughtsFirstMatch`. 

We should set the cursor as soon as its respective thoughts are loaded into state. Then `state.cursorInitialized` is set to true to ensure it only occurs once.

https://github.com/cybersemics/em/blob/3a0367c4f28703b1259567f6441f0a202bb0acbf/src/reducers/updateThoughts.ts#L106-L119

PR #1094, opened 2021-04-18
Set cursor after thoughts are loaded properly

onurpolattimur:
Fixes: #1086 

-------

Notes: To show thought with the highlighted bullet, isEditing prop is used in Bullet component, and we assign equalPath result to isEditing in Though component "equalPath(cursor, path)". The equalPath method takes ranks into consideration too but in this time thoughts are not loaded properly and the result is false. If thoughts are loaded properly the decodeThoughtsUrl result contains ranks and id. I didn't check according to **rank** because we set **zero** as default in rankThoughtsFirstMatch method so I used id for condition.

### DIFF
--- a/src/reducers/updateThoughts.ts
+++ b/src/reducers/updateThoughts.ts
@@ -2,7 +2,7 @@ import _ from 'lodash'
 import { State, PushBatch } from '../util/initialState'
 import { decodeThoughtsUrl, expandThoughts } from '../selectors'
 import { ExistingThoughtChangePayload } from '../reducers/existingThoughtChange'
-import { hashContext, logWithTime, mergeUpdates, reducerFlow, getWhitelistedThoughts, isRoot } from '../util'
+import { hashContext, logWithTime, mergeUpdates, reducerFlow, getWhitelistedThoughts, isRoot, hashThought } from '../util'
 import { CONTEXT_CACHE_SIZE, EM_TOKEN, HOME_TOKEN, THOUGHT_CACHE_SIZE } from '../constants'
 import { Child, Context, ContextHash, Index, Lexeme, Parent, Path, SimplePath, ThoughtHash, ThoughtsInterface } from '../types'
 
@@ -104,15 +104,20 @@ const updateThoughts = (state: State, { thoughtIndexUpdates, contextIndexUpdates
     }),
 
     // Reset cursor on first load. The pullQueue can determine which contexts to load from the url, but cannot determine the full cursor (with ranks) until the thoughts have been loaded. To make it source agnostic, we decode the url here.
-    // state.cursor should always be null here, but check to make sure we're not overriding it
-    !state.cursor && thoughts
+    !state.cursorInitialized
       ? state => {
         const { contextViews, path } = decodeThoughtsUrl(state, window.location.pathname)
         const cursorNew = !path || isRoot(path) ? null : path
-        return {
+        const thoughtsLoaded = cursorNew?.every(pathItem => thoughts.thoughtIndex[hashThought(pathItem.value)])
+
+        return thoughtsLoaded || !cursorNew ? {
           ...state,
           contextViews,
           cursor: cursorNew,
+          cursorInitialized: true,
+        } : {
+          ...state,
+          cursor: cursorNew
         }
       }
       : null,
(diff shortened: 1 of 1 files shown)


src/reducers/updateThoughts.ts
```diff
@@ -105,15 +105,14 @@ const updateThoughts = (state: State, { thoughtIndexUpdates, contextIndexUpdates
 
     // Reset cursor on first load. The pullQueue can determine which contexts to load from the url, but cannot determine the full cursor (with ranks) until the thoughts have been loaded. To make it source agnostic, we decode the url here.
     // state.cursor should always be null here, but check to make sure we're not overriding it
-    !state.cursorInitialized && !state.cursor
+    !state.cursor && thoughts
```
raineorshine, line 108: There is a problem with checking `state.cursor` instead of `cursorInitialized`. null is the value expected when the cursor is not on any thought, e.g. when you hit escape enough times. Thus `cursor` can be null throughout regular use of the app, not only when it has not yet been initialized. This may solve the issue, but it results in this block of code being called redundantly instead of once on load. We need to initialize the cursor only once.

raineorshine: Note that `thoughts` is always truthy, so `&& thoughts` is a noop.

src/reducers/updateThoughts.ts
```diff
@@ -122,8 +121,20 @@ const updateThoughts = (state: State, { thoughtIndexUpdates, contextIndexUpdates
     state => ({
       ...state,
       expanded: expandThoughts(state, state.cursor),
-    })
+    }),
+
+    !state.cursorInitialized ? state => {
+      const { path, contextViews } = decodeThoughtsUrl(state, window.location.pathname)
+      return !path || path.some(pathItem => pathItem.id === '') ?
```
raineorshine, line 128: The empty id seems like a tenuous thing to code against. That is, I'm skeptical about treating that as a formal API. I'd like to consider other ways to detect this.

raineorshine (changes requested):
Thanks! I left comments below for requested changes. This issue may have a few too many confounding variables to make it a good first task, but feel free to take another stab at it. If you get stuck just let me know.

Note: There is a regression to be aware of also. When the cursor is `null` and you refresh the page, it incorrectly sets the cursor to `[{ value: HOME_TOKEN, rank: 0 }]` when it should be `null`. Technically the home token is not a valid `Path` since does not have a visible bullet, so we use `null` to represent no cursor.

src/reducers/updateThoughts.ts
onurpolattimur: You are right. I didn't like it.  I updated the code, now we are checking whether thought is loaded using thoughtIndex.

src/reducers/updateThoughts.ts
```diff
@@ -104,16 +104,20 @@ const updateThoughts = (state: State, { thoughtIndexUpdates, contextIndexUpdates
...
 
     // Reset cursor on first load. The pullQueue can determine which contexts to load from the url, but cannot determine the full cursor (with ranks) until the thoughts have been loaded. To make it source agnostic, we decode the url here.
-    // state.cursor should always be null here, but check to make sure we're not overriding it
-    !state.cursorInitialized && !state.cursor
+    !state.cursorInitialized
       ? state => {
         const { contextViews, path } = decodeThoughtsUrl(state, window.location.pathname)
         const cursorNew = !path || isRoot(path) ? null : path
-        return {
+        const isCursorLoaded = cursorNew?.every(child => getThought(state, child.value))
```
raineorshine, line 111: Note the use of the `getThought` selector (which should really be called `getLexeme`). You'll get to know the different selectors and util functions over time. `thoughtIndex` and `contextIndex` are low level so you will usually never access them directly.

raineorshine (approved):
Works great, and meets all the requirements! Thanks!

src/reducers/updateThoughts.ts
raineorshine: Also regarding terminology: a [`Path`](https://github.com/cybersemics/em/wiki/Technical-Overview#314-path) consists of `Child` nodes from different `Parents`. So I renamed `pathItem` to `child`.

onurpolattimur: 👍🏻

Issue #1096, opened 2021-04-19
Data Integrity Issue: exisitingThoughtMove


shresthabijay:
# Steps To Reproduce

- Create a new empty thought and paste the following structure.
   ```
     - k
       - a
         - b
           - a
     - m
   ```
- Move `k.a` under `m`
- Use `fixContextIndex` to check the number of missing lexemes in the `contextIndex`

# Current Behaviour

`fixContextIndex` shows that one lexeme is not in sync with `contextIndex`.

# Expected Behavior

There should be no data integrity issue.

Issue #1095, opened 2021-04-19
Data Integrity Issue: existingThoughtChange


shresthabijay:
# Steps to reproduce

- Create new empty thoughts and paste the following.
   ```
    - a
      - b
        - d
      - d  
   ```
- Set cursor to `a` and change it to `azkaban.

# Current Behaviour

Thought `d` under the context `a.b` disappears. Also, you can use `fixContextIndex` to check the number of missing lexemes in `contextIndex`.

# Expected Behaviour

There should be no data integrity issue.

PR #1097, opened 2021-04-19
existingThoughtChange/Move data integrity fix

### LINKED ISSUE #1069: Missing thoughts in contextIndex
I have noticed a large number of thoughts go missing from my personal thoughtspace. 

- Mostly affects grandchildren of the HOME context. (This could be due to a specific way that the HOME context is loaded, or it could be inspecific and simply the result of the high number of times the HOME context is loaded.)
- Affects thoughts that are not being edited; related to the loading/reconciling logic 
- Thoughts appear to be orphaned rather than fully deleted, i.e. removed from a context but all descendants intact.

For example, I haven’t touched the “💬 Language” thought in the HOME context in a long time. When I open it, its children are intact but they are incorrectly shown as leaves. They should have large subtrees. 

A96D4C5E-0772-491F-A81C-65B4AC35D436

The issue could possibly be mitigated by converting the `Parent`‘s `children` property from an array to an object (#230), which would eliminate the possibility of overwriting the children with a single child update. It still doesn’t address the source of the conflict.

shresthabijay:
fixes #1069 #1095 #1096

# Changes

- Update `existingThoughtChange` and `exisitingThoughtMove` to fix data integrity issue.
- Add tests specific to these issues.


shresthabijay:
@raineorshine I have made the changes. Please review it. Thanks!

I have added a selector called `checkDataIntegrity` which is used in data integrity tests. It is basically `fixContextIndex`  but returns `contextIndex` updates if there are missing lexemes. Since the majority code is redundant between `fixContextIndex` and `checkDateIntegrity` should I expose the `fixContextIndex` to the `window.em`?

raineorshine:
> I have added a selector called `checkDataIntegrity` which is used in data integrity tests. It is basically `fixContextIndex` but returns `contextIndex` updates if there are missing lexemes. Since the majority code is redundant between `fixContextIndex` and `checkDateIntegrity` should I expose the `fixContextIndex` to the `window.em`?

You can remove `fixContextIndex`. It's no longer useful now that not all thoughts are loaded into memory.

src/selectors/checkDataIntegrity.ts
```diff
@@ -0,0 +1,71 @@
+import { hashContext, hashThought } from '../util'
+import { State } from '../util/initialState'
+import { Index, Parent, Timestamp } from '../types'
+
+/**
+ * This script runs client-side (in the browser console) and operate directly on in-memory data using window.em and then dispatching updateThoughts to sync.
+ */
+
+/** Iterates through each lexeme in thoughtIndex, identifies lexeme.contexts that are not replicated in contextIndex, and generates contextIndexUpdates that are dispatched to restore them. */
+const checkDataIntegrity = (state: State, max = 100000) => {
```
raineorshine, line 10: Since `checkDataIntegrity` is only used in tests, we should put it in `/test-helpers` in order to keep `/selectors` just for runtime functions.

src/selectors/checkDataIntegrity.ts
```diff
@@ -0,0 +1,71 @@
+import { hashContext, hashThought } from '../util'
+import { State } from '../util/initialState'
+import { Index, Parent, Timestamp } from '../types'
+
+/**
+ * This script runs client-side (in the browser console) and operate directly on in-memory data using window.em and then dispatching updateThoughts to sync.
+ */
```
raineorshine, line 7: This comment is no longer relevant and can be deleted.

raineorshine (changes requested):
Looks good! Much needed. Thanks again!

PR #1100, opened 2021-04-21
Add e2e test for cursor and add some helper functions for e2e tests

onurpolattimur:
Hi, I wrote e2e tests for the issue #1086. It was challenging and there were some limitations, timing issues etc. I have added some util methods which can be useful for future e2e tests too. I will try to explain one by one.

In e2e tests, every operation that act as user is really fast and some elements may not be visible when we want to click the element. We can solve this issues by adding delay functions but we can not be sure how long we have to wait. Some tests may fail in other developer's environment or if one day we change the github's action runner image(ubuntu-latest). But if we really need to use delay, there is nothing to do. We will use it.

But I always avoid using delay as much as possible. So, I wrote some `wait` functions.

- `waitForEditable`
Wait for editable div that contains given thought value to be visible.
Why? If you don't wait, tests may fail randomly when you want to access the element.
- `waitForState`
Wait until state of the given property equals to given value.
Why? This makes sure that state is in desired condition. For example, I need to refresh the page in the test but also I need to wait for the isPushing state to become false before page reload. Or you may want to wait cursor state become what you desire when you click a thought.
- `waitForThoughtToExistInDb`
Wait until given thought value is written to db. 
Why? The waitForState function is not enough for this situation. waitForState function waits for isPushing state to become false but in this time transaction may not be committed yet and if we don't wait and refresh the page, upps! some thoughts do not exist in db anymore.

Also I wrote a small function `getEditingText` to get currently editing thought value. 

I would like to point on another thing that was really time consuming for me, I don't want you to waste time on this too. 
When I run the tests one by one, everything is green. But when I run all tests some of them randomly are red and some of them are green. Let me explain with an example; we reload the page after each test and click the skip tutorial before each test. When I run all tests, the problem was that the first test passes but second test fails. Then I ran test in non-headless mode which open a chromium browser and I saw that the second test stuck on popup screen and I suspected localStorage but we have already cleared the localStorage between tests.  But it is not atomic, so I added this line to make sure localStorage is empty before reload page. 
`await page.waitForFunction(() => localStorage.length === 0)`

Also, I think we should not directly update the state from e2e tests. We should write test act as user. In `caret` test, I realised some codes that change the state. So, I also edited some part of the test. Now we have only one function to dispatch an event for paste operation. For now we can keep this.


src/test-helpers/getThoughtFromDB.ts
```diff
@@ -0,0 +1,10 @@
+import * as db from '../data-providers/dexie'
+import dbGetThought from '../data-providers/data-helpers/getThought'
```
raineorshine, line 2: I believe this file should go in `/src/test-helpers/e2e-helpers` since it is embedded in the puppeteer environment vs being called directly in test code.

src/test-helpers/getThoughtFromDB.ts
```diff
@@ -0,0 +1,10 @@
...
+
+/**
+ * Get thought from db for given thought value.
+ */
+const getThoughtFromDB = async (thought: string) => {
```
raineorshine, line 7: Let's call this `getLexemeFromDB`, otherwise it could be incorrectly assumed to get a `Child`.

src/e2e/__tests__/caret.ts
```diff
@@ -39,7 +36,14 @@ describe('caret testing', () => {
 
     await page.keyboard.press('Enter')
     await paste(page, [''], importText)
-    await setCursor(page, ['puppeteer', 'web scrapping'], { offset: 3 })
+
+    await clickBullet(page, 'puppeteer')
```
raineorshine, line 40: Thanks, this is an improvement over setting the `cursor` directly.

Clicking the bullet usually has the same effect as clicking on the editable, but it's more of an edge case, so I'd prefer not to use it for our primary means of moving the cursor around. (In many outliners you click the bullet to expand/collapse an item, but that's not the case in **em**. In **em**, you move the cursor and the bullets auto-expand/collapse for you.)

You could create a `clickThought` helper or just use `(await getEditable(page, value)).click()`.

src/e2e/__tests__/caret.ts
```diff
@@ -39,7 +36,14 @@ describe('caret testing', () => {
...
+    await clickBullet(page, 'web scrapping')
```
raineorshine, line 41: Note that `setCursor(page, ['puppeteer', 'web scrapping'])` sets the cursor once (on "web scrapping" [sic]). It just takes the full `Context` as an argument. So you should replace this with a single click.

src/e2e/__tests__/caret.ts
```diff
@@ -39,7 +36,14 @@ describe('caret testing', () => {
...
+
+    await waitForEditable(page, 'web scrapping')
+    const editableNodeHandle = await getEditable(page, 'web scrapping')
+    await clickWithOffset(page, editableNodeHandle, { horizontalClickLine: 'left', offset: 3 })
```
raineorshine, line 45: `offset` overrides `horizontalClickLine`, so in this case you can omit `horizontalClickLine`.

src/e2e/__tests__/caret.ts
```diff
@@ -54,9 +58,12 @@ describe('caret testing', () => {
...
     await page.keyboard.press('Enter')
     await paste(page, [''], importText)
-    await setCursor(page, ['Don\'t stay awake for too long', 'I don\'t wanna fall asleep'], { offset: 10 })
-    await clickBullet(page, 'Don\'t stay awake for too long')
 
+    await waitForEditable(page, 'I don\'t wanna fall asleep')
+    const editableNodeHandle = await getEditable(page, 'I don\'t wanna fall asleep')
+    await clickWithOffset(page, editableNodeHandle, { horizontalClickLine: 'left', offset: 10 })
+
+    await clickBullet(page, 'Don\'t stay awake for too long')
```
raineorshine, line 66: This instance of `clickBullet` is still valid since the test is actually about clicking on a bullet.

src/e2e/__tests__/caret.ts
```diff
@@ -83,8 +94,12 @@ describe('caret testing', () => {
 
     await page.keyboard.press('Enter')
     await paste(page, [''], importText)
-    await setCursor(page, ['Purple Rain'], { offset: 5 })
+
+    await waitForEditable(page, 'Purple Rain')
     const editableNodeHandle = await getEditable(page, 'Purple Rain')
```
raineorshine, line 99: Could we combine these two lines (or make it possible to do so)?

```ts
const editableNodeHandle = await waitForEditable(page, 'Purple Rain')
```

raineorshine:
> In e2e tests, every operation that act as user is really fast and some elements may not be visible when we want to click the element. We can solve this issues by adding delay functions but we can not be sure how long we have to wait. Some tests may fail in other developer's environment or if one day we change the github's action runner image(ubuntu-latest). But if we really need to use delay, there is nothing to do. We will use it.
> 
> But I always avoid using delay as much as possible. So, I wrote some `wait` functions.

Great, I concur! We couldn't figure out a way to get rid of the `delay`, so it was definitely a compromise. Glad you figured it out!

raineorshine (changes requested):

src/test-helpers/getThoughtFromDB.ts
onurpolattimur: done

onurpolattimur: 👍🏻

src/e2e/__tests__/caret.ts
onurpolattimur: I couldn't understand what you mean. How can I click the collapsed subthought with a single click?

raineorshine: Oops! You’re right :).

src/test-helpers/e2e-helpers/getEditingText.ts
```diff
@@ -0,0 +1,10 @@
+import { Page } from 'puppeteer'
+
+/**
+ * Get the thought value that cursor on.
+ */
+const getEditingText = (page: Page) => page.evaluateHandle(() => {
+  return document.querySelector('.editing .editable')?.innerHTML
+})
```
shresthabijay, line 8: You can use `page.evaluate` instead. That way it will simply return a string instead of `JSHandle` and you don't have to access `_remoteObject.value`

src/e2e/__tests__/caret.ts
```diff
@@ -83,8 +91,11 @@ describe('caret testing', () => {
 
     await page.keyboard.press('Enter')
     await paste(page, [''], importText)
-    await setCursor(page, ['Purple Rain'], { offset: 5 })
-    const editableNodeHandle = await getEditable(page, 'Purple Rain')
+
+    const editableNodeHandle = await waitForEditable(page, 'Purple Rain')
+
+    await clickWithOffset(page, editableNodeHandle, { horizontalClickLine: 'left', offset: 5 })
```
shresthabijay, line 97: `offset` overrides `horizontalClickLine`. So you can remove `horizontalClickLine` here.

shresthabijay (changes requested):
@onurpolattimur  Great work! Just a small change.

shresthabijay:
> Now we have only one function to dispatch an event for the paste operation. For now, we can keep this.

I tried using many workarounds for the paste operation. But nothing worked with headless mode. Hope we can figure it out soon.

src/test-helpers/e2e-helpers/getEditingText.ts
onurpolattimur: You're right! Thanks.

src/test-helpers/e2e-helpers/waitForEditable.ts
```diff
@@ -5,8 +5,11 @@ import { Page } from 'puppeteer'
  */
 const waitForEditable = async (page: Page, value: string) =>
   await page.waitForFunction((value: string) => {
-    return Array.from(document.getElementsByClassName('editable'))
-      .some(element => element.innerHTML === value)
+    const editableElement = Array.from(document.getElementsByClassName('editable'))
+      .find(element => element.innerHTML === value)
+    if (editableElement) {
+      return editableElement
+    }
```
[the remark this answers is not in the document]
onurpolattimur: Sorry, I missed it. waitForFunction waits until pageFunction returns truthy value. We can return editableElement.

raineorshine (approved):

raineorshine:
Excellent, thank you!

PR #1102, opened 2021-04-22
Remove unwanted text selection after clipboard copy

### LINKED ISSUE #1077: Gestures stop working after export
Tested on mobile Safari. 

## Steps to Reproduce

- a
  - b

1. Set the cursor on `b`
2. Open the export modal
3. Copy “b” to clipboard
4. Swipe right to activate `cursorBack`

## Current Behavior

Nothing happens. No gestures work.

## Expected Behavior

`cursorBack` should be activated. All gestures should work as normal. 

## Notes

- Problem does not occur if export is cancelled
- Problem goes away if edit mode is entered

shresthabijay:
fixes #1077 

# Source of issue

After the clipboard copy in `ModalExport` there was an unwanted text selection, probably caused by `clipboard`. This was preventing the gesture handler to work as it only works when there is no text selection in the document.

# Solution

Removing the text selection after a successful clipboard copy operation solves the problem.

### DIFF
--- a/src/components/ModalExport.tsx
+++ b/src/components/ModalExport.tsx
@@ -110,6 +110,9 @@ const ModalExport = () => {
 
     clipboard.on('success', () => {
 
+      // Note: clipboard leaves unwanted text selection after copy operation. so removing it to prevent issue with gesture handler
+      if (document.getSelection()?.toString()) document.getSelection()?.removeAllRanges()
+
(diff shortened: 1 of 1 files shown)


raineorshine (approved):

raineorshine:
Yay! Works great

Issue #1098, opened 2021-04-20
Second part of tutorial crashes


KICKdesigns:
Steps to reproduce:

On the second screen of the second part of the tutorial (where you get the choice of To-Do List, Journal Theme, and Book/Podcast Notes), click on any of the options.

**Result:** `TypeError: Cannot read property 'length' of undefined`

PR #1106, opened 2021-04-23
Fix rootSubthoughts conflicts to fix tutorial crashes

KICKdesigns:
Issue #1098 

@raineorshine

FYI the function checks for rootSubthoughts.length > 0, which of course crashes because rootSubthoughts is null, and doesn't have a length field or any field.

I added a test for rootSubthoughts (truthiness = false if null) which fixed this particular crash. But then it crashes on the next tutorial screen, not immediately, but after I follow its instructions, which are to type something.

That crash is taking place in Tutorial2StepContext1.js, where it is also using rootSubthoughts, without considering that it might be null.

We need to decide if the problem is
a) the fact that rootSubthoughts is null, and it shouldn't be
b) it is allowed to be null, but there has to be tests and handling for when it is.

Question: Where is the code that calls Tutorial2StepContext1Parent? This would be the code that could potentially pass it a rootSubthoughts that is not null. I can't find any code that is calling it directly. It gets there through React via the dispatch function, but nowhere else based on the repo files.

One thing that makes this time consuming is that every time I run the project to test it, or run it with some added console.log statements, I have to manually go through all the tutorial steps from the beginning of the first tutorial, after first manually clearing the IndexedDB.

### DIFF
--- a/src/components/Tutorial/Tutorial2StepContext1.js
+++ b/src/components/Tutorial/Tutorial2StepContext1.js
@@ -14,29 +14,21 @@ import {
 } from '../../constants'
 
 // eslint-disable-next-line jsdoc/require-jsdoc
-const Tutorial2StepContext1 = ({ cursor, tutorialChoice, rootSubthoughts }) => <Fragment>
+const Tutorial2StepContext1 = ({ cursor, tutorialChoice, rootChildren }) => <Fragment>
   <p>Let's say that {
     tutorialChoice === TUTORIAL_VERSION_TODO ? 'you want to make a list of things you have to do at home.' :
     tutorialChoice === TUTORIAL_VERSION_JOURNAL ? 'one of the themes in your journal is "Relationships".' :
     tutorialChoice === TUTORIAL_VERSION_BOOK ? `you hear a podcast on ${TUTORIAL_CONTEXT[tutorialChoice]}.` :
     null
   } Add a thought with the text "{TUTORIAL_CONTEXT[tutorialChoice]}" <i>within</i> “{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}”.</p>
-  <p>Do you remember how to do it?
-    <TutorialHint>
-      <br /><br />{!cursor || headValue(cursor).toLowerCase() !== TUTORIAL_CONTEXT1_PARENT[tutorialChoice].toLowerCase() ? `Select "${TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". ` : null}{isTouch ? 'Trace the line below with your finger' : `Hold ${isMac ? 'Command' : 'Ctrl'} and hit Enter`} to create a new thought <i>within</i> "{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". Then type "{TUTORIAL_CONTEXT[tutorialChoice]}".
-    </TutorialHint>
-  </p>
+  {rootChildren.find(child => child.value.toLowerCase() === TUTORIAL_CONTEXT1_PARENT[tutorialChoice].toLowerCase())
+    ? <p>Do you remember how to do it?
+      <TutorialHint>
+        <br /><br />{!cursor || headValue(cursor).toLowerCase() !== TUTORIAL_CONTEXT1_PARENT[tutorialChoice].toLowerCase() ? `Select "${TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". ` : null}{isTouch ? 'Trace the line below with your finger' : `Hold ${isMac ? 'Command' : 'Ctrl'} and hit Enter`} to create a new thought <i>within</i> "{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". Then type "{TUTORIAL_CONTEXT[tutorialChoice]}".
+      </TutorialHint>
+    </p>
+    : <p>Oops, somehow “{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}” was changed or deleted. Click the Prev button to go back.</p>
+  }
 </Fragment>
 
 export default Tutorial2StepContext1
-// eslint-disable
-/*
-{rootSubthoughts.find(child => child.value.toLowerCase() === TUTORIAL_CONTEXT1_PARENT[tutorialChoice].toLowerCase())
-  ? <p>Do you remember how to do it?
-    <TutorialHint>
-      <br /><br />{!cursor || headValue(cursor).toLowerCase() !== TUTORIAL_CONTEXT1_PARENT[tutorialChoice].toLowerCase() ? `Select "${TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". ` : null}{isTouch ? 'Trace the line below with your finger' : `Hold ${isMac ? 'Command' : 'Ctrl'} and hit Enter`} to create a new thought <i>within</i> "{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". Then type "{TUTORIAL_CONTEXT[tutorialChoice]}".
-    </TutorialHint>
-  </p>
-  : <p>Oops, somehow “{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}” was changed or deleted. Click the Prev button to go back.</p>
-}
-*/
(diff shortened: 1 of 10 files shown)


raineorshine:
> I added a test for rootSubthoughts (truthiness = false if null) which fixed this particular crash. But then it crashes on the next tutorial screen, not immediately, but after I follow its instructions, which are to type something.

Maybe related? Hard to tell if they are the same bug or not.

> We need to decide if the problem is
> a) the fact that rootSubthoughts is null, and it shouldn't be
> b) it is allowed to be null, but there has to be tests and handling for when it is.

Exactly. My guess is that `rootSubthoughts` should never be null, but it's hard to tell immediately since it's Javascript instead of Typescript (one of the only places in the codebase!).

> Question: Where is the code that calls Tutorial2StepContext1Parent? This would be the code that could potentially pass it a rootSubthoughts that is not null. I can't find any code that is calling it directly. It gets there through React via the dispatch function, but nowhere else based on the repo files.

`Tutorial2StepContext1Parent` is added to the [tutorialStepComponentMap](https://github.com/cybersemics/em/blob/dev/src/components/Tutorial/TutorialStepComponentMap.js#L81), a somewhat indirect way to render Tutorial components based on the value of `state.tutorialStep`. It is rendered from [Tutorial/index.js](https://github.com/cybersemics/em/blob/dev/src/components/Tutorial/index.js#L66).

KICKdesigns:
@raineorshine Thanks! PS, yes I have project-wide search on the IDE. I just wanted to pose some questions that may be relevant. I'll push through the tutorial to see if I can get it working. It may be layers of bugs (IE one bug fix may reveal a bug that appears later in the tutorial). Thanks.

KICKdesigns:
Issue #1098

src/components/Tutorial/Tutorial2StepContext1.js
```diff
@@ -21,14 +21,22 @@ const Tutorial2StepContext1 = ({ cursor, tutorialChoice, rootSubthoughts }) => <
...
-        <br /><br />{!cursor || headValue(cursor).toLowerCase() !== TUTORIAL_CONTEXT1_PARENT[tutorialChoice].toLowerCase() ? `Select "${TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". ` : null}{isTouch ? 'Trace the line below with your finger' : `Hold ${isMac ? 'Command' : 'Ctrl'} and hit Enter`} to create a new thought <i>within</i> "{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". Then type "{TUTORIAL_CONTEXT[tutorialChoice]}".
-      </TutorialHint>
-    </p>
-    : <p>Oops, somehow “{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}” was changed or deleted. Click the Prev button to go back.</p>
-  }
+  <p>Do you remember how to do it?
+    <TutorialHint>
+      <br /><br />{!cursor || headValue(cursor).toLowerCase() !== TUTORIAL_CONTEXT1_PARENT[tutorialChoice].toLowerCase() ? `Select "${TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". ` : null}{isTouch ? 'Trace the line below with your finger' : `Hold ${isMac ? 'Command' : 'Ctrl'} and hit Enter`} to create a new thought <i>within</i> "{TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}". Then type "{TUTORIAL_CONTEXT[tutorialChoice]}".
+    </TutorialHint>
+  </p>
```
raineorshine, line 28: Look like you have removed some functionality. If the parent is changed or deleted, the tutorial needs to instruct them to go back.

src/components/Tutorial/Tutorial2StepContext1SubThought.js
```diff
@@ -28,9 +28,8 @@ const Tutorial2StepContext1SubThought = ({ cursor, tutorialChoice, rootSubthough
     }</p>
     {
       // e.g. Home
-      rootSubthoughts.find(child => child.value.toLowerCase() === TUTORIAL_CONTEXT1_PARENT[tutorialChoice].toLowerCase()) &&
-        // e.g. Home/To Do
-        getChildrenRanked(store.getState(), [TUTORIAL_CONTEXT1_PARENT[tutorialChoice]]).find(child => child.value.toLowerCase() === TUTORIAL_CONTEXT[tutorialChoice].toLowerCase())
+      // e.g. Home/To Do
+      getChildrenRanked(store.getState(), [TUTORIAL_CONTEXT1_PARENT[tutorialChoice]]).find(child => child.value.toLowerCase() === TUTORIAL_CONTEXT[tutorialChoice].toLowerCase())
```
raineorshine, line 32: If you remove the `rootSubthoughts` check, then it's possible that the parent no longer exists, so it wouldn't be appropriate to ask them to go to the next step.

raineorshine (changes requested):
Hey, I'm a bit confused here. It looks like you have simply removed functionality in order to make the error go away rather than fixing the issue. Did you understand the purpose of the `rootSubthoughts` checks? Let me know if I can help.

KICKdesigns:
@raineorshine Not sure if you saw the notes in the latest commit, but let me respond to some of your notes:

- Did you understand the purpose of the `rootSubthoughts` checks?
  - I've been working through understanding it's purpose through some posed questions and code exploration. When I do a quick search of `rootSubthoughts` it's being used in 2 main TS files (not including the references in the tutorial JS files) Divider.ts and Thought.ts. Based on the code references to it, it looks like it's used as a ranking / ordering var for the list of to-dos. Is that correct?

- If you remove the `rootSubthoughts` check, then it's possible that the parent no longer exists, so it wouldn't be appropriate to ask them to go to the next step.
  - Currently, `rootSubthoughts` doesn't contain any information about whether the parent still exists. So how do we put that information into `rootSubthoughts`, or otherwise test for if the parent still exists? In the TutorialNavigationNext.js file around line 32, it seems to get set to `undefined`, which is what is causing a lot of the errors within the tutorial.

raineorshine:
> * I've been working through understanding it's purpose through some posed questions and code exploration. When I do a quick search of `rootSubthoughts` it's being used in 2 main TS files (not including the references in the tutorial JS files) Divider.ts and Thought.ts. Based on the code references to it, it looks like it's used as a ranking / ordering var for the list of to-dos. Is that correct?

Yes, sort of. The `rank` property is indeed used for ordering and is part of the `Child` type, although that's not relevant here. `rootSubthoughts` contains the list of subthoughts in the root context, i.e. the thoughts at the top level in the hierarchy. Part II of the Tutorial walks them through the creation of a thought that is added to two different contexts. For variety, they can choose one of three themes: A To-Do List, Journal, or Book/Podcast. So the `rootSubthoughts` logic exists to make sure that they create the top level thought before proceeding to the next step in the tutorial.

Since `rootSubthoughts` is expected to be an array but it is `undefined`, we need to trace the props to see why it isn't being set. In the `Tutorial` component which dynamically renders the correct component for the current tutorial step, you can see that `rootSubthoughts` is not passed at all.

https://github.com/cybersemics/em/blob/5474052ae8e9d5f306f9f8afc5a8282348a9fe8e/src/components/Tutorial/index.js#L42-L51

https://github.com/cybersemics/em/blob/5474052ae8e9d5f306f9f8afc5a8282348a9fe8e/src/components/Tutorial/index.js#L58

https://github.com/cybersemics/em/blob/5474052ae8e9d5f306f9f8afc5a8282348a9fe8e/src/components/Tutorial/index.js#L66

Instead, something called `rootChildren` is passed. It seems like that was accidentally renamed without updating the tutorial components and that is causing it to be `undefined`.

KICKdesigns:
@raineorshine The latest push has a bunch of fixes for the var conflicts and it enables the user to run through the tutorial all the way through now.

src/components/Tutorial/TutorialUtils.js
```diff
@@ -2,26 +2,49 @@ import { store } from '../../store'
...
+    else {
+      // console.log('returning neither (FAIL)')
+      return undefined
+    }
+  }
+}
+
 /** Returns true if the first context thought has been created, e.g. /Home/To Do/x. */
-export const context1SubthoughtCreated = ({ rootSubthoughts, tutorialChoice }) => {
+export const context1SubthoughtCreated = ({ rootSubthoughts, rootChildren, tutorialChoice }) => {
```
raineorshine, line 27: Why take both `rootSubthoughts` and `rootChildren`? They're the same thing, just accidentally split up due to a renaming mixup.

raineorshine (approved):

PR #1107, opened 2021-04-24
Check if more than one child exist for empty thought sorting

onurpolattimur:
In resortEmptyInPlace function If there is one empty child, the app will be crashed when Global Sort is Alphabetical. Because in this function we get nearestSibling which will be undefined when there is one empty child.

When you toggle sort a context you can not reproduce this issue because there will be two child in the context.

### DIFF
--- a/src/selectors/getChildren.ts
+++ b/src/selectors/getChildren.ts
@@ -84,7 +84,7 @@ const rankDiff = (a: Child, b: Child) => Math.abs(a?.rank - b?.rank)
 const getChildrenSortedAlphabetical = (state: State, context: Context): Child[] => {
   const sorted = getChildrenSortedBy(state, context, compareThought)
   const emptyIndex = sorted.findIndex(child => !child.value)
-  return emptyIndex === -1 ? sorted : resortEmptyInPlace(sorted)
+  return emptyIndex === -1 || sorted.length === 1 ? sorted : resortEmptyInPlace(sorted)
 }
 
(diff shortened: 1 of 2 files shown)


src/selectors/getChildren.ts
```diff
@@ -84,7 +84,7 @@ const rankDiff = (a: Child, b: Child) => Math.abs(a?.rank - b?.rank)
 const getChildrenSortedAlphabetical = (state: State, context: Context): Child[] => {
   const sorted = getChildrenSortedBy(state, context, compareThought)
   const emptyIndex = sorted.findIndex(child => !child.value)
-  return emptyIndex === -1 ? sorted : resortEmptyInPlace(sorted)
+  return emptyIndex === -1 || sorted.length === 1 ? sorted : resortEmptyInPlace(sorted)
```
raineorshine, line 87: If you place the guard here, then `resortEmptyInPlace` still has a possible input that is valid to the type system but will cause the app to crash. It would be better to place the guard in `resortEmptyInPlace`.

Another consideration is module responsibility: `getChildrenSortedAlphabetical` shouldn't have to know that much about `resortEmptyInPlace`. It should only have to know that it takes `Child[]` and returns `Child[]`.

raineorshine (changes requested):

src/selectors/getChildren.ts
onurpolattimur: Thanks!

raineorshine (approved):

PR #1108, opened 2021-04-24
Use global sort value as default sort instead of None

onurpolattimur:
This pr fixes the global sorting issue. When global sorting is Alphabetical, sorting is not working.

## Steps to Reproduce
Set global sorting to `Alphabetical`.

- b
- a
 
## Current Behavior
- b
- a
 

## Expected Behavior
- a
- b

### DIFF
--- a/src/shortcuts/__tests__/toggleSort.ts
+++ b/src/shortcuts/__tests__/toggleSort.ts
@@ -195,6 +195,91 @@ describe('DOM', () => {
     expect(thoughtChildren.map((child: HTMLElement) => child.textContent)).toMatchObject(['1', '2', '3'])
   })
 
+  describe('sort with Global Sort settings', () => {
+    it('thoughts are sorted alphabetically when "Global Sort" settings is Alphabetical', async () => {
+      const thoughtValue = 'c'
+      store.dispatch([
+        newThought({ value: thoughtValue }),
+        newThought({ value: 'b' }),
+        newThought({ value: 'a' }),
+        setCursor({ path: null }),
+
+        ((dispatch, getState) => dispatch(existingThoughtChange({
+          context: [EM_TOKEN, 'Settings', 'Global Sort'],
+          oldValue: 'None',
+          newValue: 'Alphabetical',
+          path: rankThoughtsFirstMatch(getState(), [EM_TOKEN, 'Settings', 'Global Sort', 'None']) as SimplePath
+        }))) as Thunk,
+      ])
+
+      const thought = await findThoughtByText(thoughtValue)
+      expect(thought).toBeTruthy()
+
+      const thoughtsWrapper = thought!.closest('ul') as HTMLElement
+      const thoughts = await findAllByPlaceholderText(thoughtsWrapper, 'Add a thought')
+
+      expect(thoughts.map((child: HTMLElement) => child.textContent)).toMatchObject(['a', 'b', 'c'])
+    })
+
+    it('subthoughts are sorted alphabetically when "Global Sort" settings is Alphabetical', async () => {
+
+      const thoughtValue = 'a'
+      store.dispatch([
+        newThought({ value: thoughtValue }),
+        newThought({ value: '3', insertNewSubthought: true }),
+        newThought({ value: '1' }),
+        newThought({ value: '2' }),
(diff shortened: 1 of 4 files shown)


src/shortcuts/__tests__/toggleSort.ts
```diff
@@ -195,6 +195,91 @@ describe('DOM', () => {
     expect(thoughtChildren.map((child: HTMLElement) => child.textContent)).toMatchObject(['1', '2', '3'])
   })
 
+  describe('sort with Global Sort settings', () => {
+    it('thoughts are sorted alphabetically when "Global Sort" settings is Alphabetical', async () => {
+      const thoughtValue = 'c'
+      store.dispatch([
+        newThought({ value: thoughtValue }),
```
raineorshine, line 202: It would be preferable to hardcode `c` rather than using `thoughtValue`.

Unlike in normal programming, in tests you want to hardcode duplicate literal values in order to maximize concreteness. This reduces the chance of a false positive.

This doesn't extend to complex values, e.g. `await findThoughtByText('c')`, as duplicating the full expression decreases readability and is more likely to introduce a bug than using a variable.

src/shortcuts/__tests__/toggleSort.ts
```diff
@@ -195,6 +195,91 @@ describe('DOM', () => {
...
+      const thought = await findThoughtByText(thoughtValue)
+      expect(thought).toBeTruthy()
+
+      const thoughtChildrenWrapper = thought!.closest('li')?.lastElementChild as HTMLElement
+      const thoughtChildren = await findAllByPlaceholderText(thoughtChildrenWrapper, 'Add a thought')
+
+      expect(thoughtChildren.map((child: HTMLElement) => child.textContent)).toMatchObject(['1', '2', '3'])
+    })
+
+    it('subthoughts are not sorted alphabetically when toggling sort and "Global Sort" settings is Alphabetical', async () => {
```
raineorshine, line 250: Make this test description a little more specific: "when toggling sort" → "when context sort is None"

src/shortcuts/__tests__/toggleSort.ts
```diff
@@ -195,6 +195,91 @@ describe('DOM', () => {
...
+          newValue: 'Alphabetical',
+          path: rankThoughtsFirstMatch(getState(), [EM_TOKEN, 'Settings', 'Global Sort', 'None']) as SimplePath
+        }))) as Thunk,
+      ])
+
+      store.dispatch(toggleAttribute({
+        context: ['a'],
+        key: '=sort',
+        value: 'None',
+      }))
```
raineorshine, line 271: I think this can be placed in the previous dispatch array?

raineorshine (changes requested):
Thanks! A few small requests.

src/shortcuts/__tests__/toggleSort.ts
onurpolattimur: Yes, but my intention was to show the second dispatch as action. 
I mean, the first dispatch is for `initial` and the second one is for the action. I think it is better for readability but if you want we can move the second one to the previous dispatch.

src/shortcuts/__tests__/toggleSort.ts
onurpolattimur: > This doesn't extend to complex values, e.g. await findThoughtByText('c'), as duplicating the full expression decreases readability and is more likely to introduce a bug than using a variable.

Sorry, I didn't understand this part.

raineorshine: I just mean that hardcoding should only be done for literal values (strings, numbers) and not longer expressions.

src/shortcuts/__tests__/toggleSort.ts
raineorshine: Okay, I can see that. Thanks for explanation your intention!

I still prefer a single dispatch, as it makes it clear that there is nothing that happens in between. When I see multiple dispatches instead of one it makes me look for an intervening side effect or functional reason to separate them out. You can add a blank line above `toggleAttribute` or add comments like `// given` and `// when` if you'd like to make that distinction clear.

raineorshine (approved):

Issue #1103, opened 2021-04-23
Caret not set after subcategorizeOne


raineorshine:
Tested on iOS. 

Note that it is only an issue with `subcategorizeOne`, not `subcategorizeAll` for some reason. 

This regression has occurred a couple times, so it’ll be good to get test coverage now that we have caret testing with Puppeteer. 

## Steps to Reproduce

- a
  - b

1. Set the cursor on `b`
2. Close the keyboard to disable edit mode
3. Activate `subcategorizeOne` via the toolbar or gesture

## Current Behavior

New parent thought is created and caret is not set. 

## Expected Behavior

Caret should be set and edit mode enabled on the new thought.

onurpolattimur:
I think the problem is that we don't set editing true in subCategorizeOne.ts. When I add `editing: true`, the keyboard is opened. I tested on iOS/Safari.

https://github.com/cybersemics/em/blob/9768232839b32511828bc22a2fab8ed1f29d5bcf/src/reducers/subCategorizeOne.ts#L61-L64

onurpolattimur:
Also, this problem occurs only on touchable platforms because we set editing to `false` only for the touchable platforms.

https://github.com/cybersemics/em/blob/9768232839b32511828bc22a2fab8ed1f29d5bcf/src/components/Editable.tsx#L574-L579

raineorshine:
> I think the problem is that we don't set editing true in subCategorizeOne.ts. When I add `editing: true`, the keyboard is opened. I tested on iOS/Safari.

Yup, that should work. `subCategorizeOne` should always enter edit mode. I guess `subCategorizeAll` doesn't suffer from the same problem because `newThought` handles the caret.

PR #1112, opened 2021-04-26
Add mobile tests and fix sub categorize one caret issue

onurpolattimur:
I fixed the caret issue mentioned in #1103. We didn't have any mobile tests so I decided to add for our all e2e tests.

Now, all e2e tests run for both mobile and desktop. Here is the output from IDE.

image

At this time we don't have any platform-specific test. So I wrote two noop describe block for caret testing to use it as a template. 
 
1. Only for mobile platform
2. Only for desktop platfrom

If we need to write mobile-specific tests then we should write into mobile describe block, and we need to write desktop specific tests then we should write into desktop describe block. 

After writing mobile tests, there was another timing issue. Some tests randomly failed. So, I added a new function in addition to functions in #1100 

- `waitForContextHasChildWithValue`:
Wait until the given context and its child that has given value are loaded properly.
Why? There was a problem with the welcome modal, even we click the skip tutorial link it was stuck on this screen because related states were not loaded at this time. If we don't use this, we can get `'Lexeme not found'` or `'Missing oldThoughtRanked in Settings update'` error.


raineorshine:
Oops! I totally pointed you in the wrong direction with my comment "The bulk of the work here is the e2e test". I forgot this was reported on Mobile Safari!! I was imagining desktop caret testing, but that's not relevant here. I'm very sorry about that.

iPhone behavior is specific to mobile Safari. Unfortunately spoofing the user agent and setting the device size does not recreate any Safari-specific behavior. It would just test how Chrome renders the app on a small screen size, which is not necessary.

I would love the `subCategorizeOne` fix, but I don't think we want the other additions unfortunately!

onurpolattimur:
I don't think this issue occurs only on safari. I reverted the changes and tried on chrome with iPhone 11 and the same issue existed. Because it is not browser-specific, it is about whether touchable or not.

https://github.com/cybersemics/em/blob/f55a4b2d9f011b3a98a6446b12a41a31238f8b3b/src/browser.ts#L6

In puppeteer tests, it does not set only device size it also set additional properties to emulate mobile.

You can see the details at https://github.com/puppeteer/puppeteer/blob/ad6b736039436fcc5c0a262e5b575aa041427be3/src/common/DeviceDescriptors.ts#L554

If you revert the changes in `src/reducers/subCategorizeOne.ts`,  the mobile test related to the subCategorizeOne will fail. Before running the test, make sure to run `npm build `command after reverting the changes.

raineorshine:
Ah, I see. So it does test something more than just screen size. The media is different so `isTouch` is true. That will allow us to test some of the `isTouch` related functionality without an actual mobile device which I do think would be valuable. Thanks for clarifying!

For identical behavior on desktop and mobile, we don't need emulated mobile tests, since they're not actually testing anything different. Let's reduce the mobile tests down to just the one test that fails without the fix.

> At this time we don't have any platform-specific test. So I wrote two noop describe block for caret testing to use it as a template.

You can use `.skip` so they don't get added to the final test count.

> * `waitForContextHasChildWithValue`:
>   Wait until the given context and its child that has given value are loaded properly.
>   Why? There was a problem with the welcome modal, even we click the skip tutorial link it was stuck on this screen because related states were not loaded at this time. If we don't use this, we can get `'Lexeme not found'` or `'Missing oldThoughtRanked in Settings update'` error.

I'd like to understand the cause of this better.

1. Does the error occur when clicking on the skip tutorial link, or after clicking?
2. Does this happen consistently or intermittently?
3. What do you mean by "related states"?

Thanks!

src/e2e/emulated-devices.js
```diff
@@ -0,0 +1,16 @@
+import { devices } from 'puppeteer'
+
+export const iPhone = devices['iPhone 11']
+
+export const desktop = {
```
raineorshine, line 5: Why emulate desktop? Can we use puppeteer without emulation in this case?

onurpolattimur: If we set the emulator once for any test, we need to set the emulator again for another platform. Puppeteer does not have any default value or function like a reset for desktop.

onurpolattimur:
> Ah, I see. So it does test something more than just screen size. The media is different so isTouch is true. That will allow us to test some of the isTouch related functionality without an actual mobile device which I do think would be valuable. Thanks for clarifying!

Yes, exactly!

> You can use .skip so they don't get added to the final test count.

Sorry, I had removed .skip to show in the screenshot. 

> Does the error occur when clicking on the skip tutorial link, or after clicking?

It occurs after clicking on the skip tutorial link.

> Does this happen consistently or intermittently?

As I see, this happen consistently.

> What do you mean by "related states"?

When we click the link, `[EM_TOKEN, 'Settings', 'Tutorial]` context and its children are not loaded at this time properly. So it can not get old value in `settings.ts `https://github.com/cybersemics/em/blob/f55a4b2d9f011b3a98a6446b12a41a31238f8b3b/src/reducers/settings.ts#L15

And UI becomes this:

image

Maybe you cannot reproduce this issue in your local environment, to reproduce this issue you need to change 
await `page.click('#skip-tutorial') ` line to  
`await page.evaluate(() => document.getElementById('skip-tutorial')?.click())`

Because, page.click() is slightly different from document.getElementById(..).click(). You can see details at https://stackoverflow.com/questions/49979069/puppeteer-element-click-not-working-and-not-throwing-an-error. But if you don't change this line, it will be stuck on Welcome Modal, not screen I added above.

raineorshine:
> When we click the link, `[EM_TOKEN, 'Settings', 'Tutorial]` context and its children are not loaded at this time properly. So it can not get old value in `settings.ts `

Thanks. I created #1113 so we can track that issue separately. 

I'm hesitant to add `waitForContextHasChildWithValue` as a work-around for a bug that should be fixed. I guess we can add it now and then remove it when #1113 is fixed.

src/e2e/emulated-devices.js
raineorshine: Is the page not destroyed after each test? Seems like we would like full isolation between tests so I’m not sure how it was set up. 

If not, we could reset to defaults in `afterEach`. `page.emulate` is an alias for `page.setUserAgent` which has defaults, and `page.setViewport`, which may be able to be set to an empty string or a basic Chrome string. 

https://pptr.dev/#?product=Puppeteer&version=v9.0.0&show=api-pageemulateoptions

onurpolattimur: No, the page is not destroyed after each test. We only reset DB and local storage as you can see here:
https://github.com/cybersemics/em/blob/5474052ae8e9d5f306f9f8afc5a8282348a9fe8e/src/e2e/__tests__/caret.ts#L16-L21 

I will look at how we can achieve isolation between tests. The way it is doesn't give me much confidence too.

Maybe, we can pass an empty string to `page.setUserAgent` but I think [page.setViewPort](https://pptr.dev/#?product=Puppeteer&version=v9.0.0&show=api-pagesetviewportviewport)
 doesn't get string. We need to pass a Viewport object.

raineorshine: > I will look at how we can achieve isolation between tests. The way it is doesn't give me much confidence too.

Great. I think that will be better.

> Maybe, we can pass an empty string to `page.setUserAgent` but I think [page.setViewPort](https://pptr.dev/#?product=Puppeteer&version=v9.0.0&show=api-pagesetviewportviewport)
> doesn't get string. We need to pass a Viewport object.

Sorry! I meant the other way around. `page.emulate` is an alias for `page.setViewport` which has defaults (except dimensions), and `page.setUserAgent`, which may be able to be set to an empty string or a basic Chrome string.

onurpolattimur:
Hi @raineorshine, I have refactored the e2e tests. Now, every test has its own context. We don't need to reset indexedDB or localStorage etc. anymore. So, It gives me much confidence now.

raineorshine (changes requested):
Thanks for removing the emulated devices and isolating the tests! That's a big improvement.

After reviewing the PR, I'm unsatisfied with the use of `this.global`. `setup` was already  using `this.global` as a hacky way to pass information to `teardown`. However adding `this.global.emulatedDevice` and exposing `openNewPage` and `closePage` as globals is a whole other layer of hackiness. We need to modularize this functionality and not rely on globals to pass information between phases of the test.

If you can assess this problem and propose a solution that would be great. Otherwise I can spend some time with the code myself.

src/test-helpers/e2e-helpers/initPage.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+  page.on('dialog', async dialog => {
+    await dialog.accept()
+  })
+
+  await page.goto(url)
+
+  skipTutorial && await skipTutorialScreen(page)
+
+  return { page, context }
```
raineorshine, line 29: Let's just return `page`. It's a cleaner api and we can get the context by calling `page.browserContext()`.

src/test-helpers/e2e-helpers/initPage.ts
```diff
@@ -0,0 +1,32 @@
...
+  puppeteerBrowser?: Browser,
+  url?: string,
+  skipTutorial?: boolean,
+  emulatedDevice?: Device,
+}
+
+/**
+ * Create a new incognito context and page.
+ */
+const initPage = async ({ puppeteerBrowser = browser, url = 'http://localhost:3000', emulatedDevice, skipTutorial = true }: InitPageOptions) => {
```
raineorshine, line 14: Options argument should also be optional so that we can call `initPage()` instead of `initPage({})`.

raineorshine (changes requested):

raineorshine (approved):

PR #1114, opened 2021-04-30
adds email & password based authentication

### LINKED ISSUE #1111: Email/Password Login
Users should not be required to log in with their Google account. They should be able to create an account with an email and password. 

- Clicking login should go to a Login Modal rather than directly to Google login. 
  - Add email/password fields and “Sign Up” button in Login Modal. 
  - Add “I already have an account” link.
  - Add “Forgot your password?” link.
  - Add “Sign in with Google” link (current "Log In" behavior)

anmolarora1:
fixes #1111


raineorshine:
Review in progress

src/components/ModalAuth.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+  useEffect(() => {
+    updateInputValue(value)
+  }, [value])
+
+  /** On input change handler. */
+  const onChange = (e: ChangeEvent<HTMLInputElement>) => updateInputValue(e.target.value)
+
+  return <input type={type} placeholder={placeholder} value={inputValue} onChange={onChange} onBlur={onBlur}/>
+}
```
raineorshine, line 41: What's the purpose of this `Input` component? It looks like a simple wrapper for `<input>`.

src/components/ModalAuth.tsx
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * Enable Password recovery assistance.
+   */
+  const handleForgotPassword = () => {
+    updateModalKey(modalKeys.resetPassword)
+    updateTitle(modalTitles.resetPassword)
+    updateForgotPassword(true)
+    resetFormFields()
+    submitAction.current = resetPassword
+  }
```
raineorshine, line 178: This section can be improved a bit. The problem is that the variables are treated as independent when they aren't: they should all flow from the active mode. With separate variables it is possible to update some and not others and thus is more error-prone. 

Instead, store a single local state `active` or `activeMode` which is set to `login`, `signup`, or `resetPassword`. Then for example the title is just `modalTitles[active]` and does not require additional state.

I like the `modalKeys`, `modalTitles`, etc structure. It will enable clean computed values with minimal state.

raineorshine (changes requested):

raineorshine:
> @raineorshine I've added an initial level of implementation covering email/password signup, login, and account recovery using reset-password. We might need to make some adjustments to the UX so it'd be great if you can review and share your thoughts.

Looks good! I tried sign up and log in and they works great.

One thing I would like is for the "Forget password" link to always be visible on the "Log In" page. It took me a while to figure out how to access it. Thanks!

src/components/ModalAuth.tsx
anmolarora1: It's to prevent the re-renders of the entire modal component while the user is typing.

src/components/ModalAuth.tsx
anmolarora1: Agreed

src/components/ModalAuth.tsx
raineorshine: Okay, thanks

src/components/ModalAuth.tsx
```diff
@@ -0,0 +1,203 @@
...
+   * Login using google account.
+   */
+  const signInWithGoogle = () => {
+    updateIsSubmitting(true)
+    dispatch(login())
+  }
+
+  useEffect(() => {
+    resetFormFields()
+    submitAction.current = isModeActive(modes.login) ? loginWithEmailAndPassword : isModeActive(modes.signup) ? signUp : resetPassword
```
raineorshine: The state management is much better now with `modes`!

`submitAction` should also be refactored in a similar way. With the current architecture, it is treated as an independent variable when it is actually a dependent variable: it can always be derived from the active mode.

You can define `submitAction` as a small function that calls the appropriate function based on the active mode. Or you can perform the conditional logic directly in the render function rather than in `useEffect`.

raineorshine (approved):

raineorshine:
I did some minor refactoring and tweaked the UX a bit to my liking.

You may want to review the changes in 970f245 (email was exactly the same in all three cases) and e750385 (`submitAction` can be done without state).

Issue #1078, opened 2021-04-07
Sort descending


raineorshine:
## Steps to Reproduce

Toggle `sortContext`

## Current Behavior

Toggles between `Alphabetical` and none. 

## New Behavior

3-way toggle between `Alphabetical`, `Alphabetical/desc`, and none. 

Add the appropriate logic for sorting in descending order with `Alphabetical/desc`.

Duplicate the sort icon and reverse the direction of the arrow. Or parameterize the current component. 

## Tests

- [ ] [None] → Alphabetical
- [ ] Alphabetical → Alphabetical/desc
- [ ] Alphabetical/desc → [None]
- [ ] Alphabetical/asc (should work the same as Alphabetical)

onurpolattimur:
Should we reverse the direction of the arrow or bars? I think it seems better when we change the order of bars. 

For ascending order:
![image](https://user-images.githubusercontent.com/23136437/115916858-0e703000-a47e-11eb-869e-14a1633fcc32.png)

For descending order:
![image](https://user-images.githubusercontent.com/23136437/115916754-e4b70900-a47d-11eb-84dd-bef8191ac287.png)

raineorshine:
Sure, I’d be happy either way!

PR #1115, opened 2021-05-01
Add descending sort feature

onurpolattimur:
Adds feature #1078 

----

I didn't use the array reverse function because I think it is more efficient to reverse the comparator. So, I decided to negate the comparator value if the sort type is descending.

Also, there was a problem that took much time for me. I had changed the `initialSettings.html` file to test manually some functionality. After completing my test, I reverted the changes. But after that, somehow unit tests were failed and I started to investigate the problem and I realize that the `raw(initialSettings.html)`  was reading a changed version of initialSettings.html but actually there was no change in this file. 

I was sure that there was a problem with the cache. And here is the solution, you can use `jest --clearCache` command to clear the cache for tests.


raineorshine:
> Also, there was a problem that took much time for me. I had changed the `initialSettings.html` file to test manually some functionality. After completing my test, I reverted the changes. But after that, somehow unit tests were failed and I started to investigate the problem and I realize that the `raw(initialSettings.html)` was reading a changed version of initialSettings.html but actually there was no change in this file.
> 
> I was sure that there was a problem with the cache. And here is the solution, you can use `jest --clearCache` command to clear the cache for tests.

Wow, I didn't know jest cached that.

I've had some major issues with `raw.macro`, such as introducing circular imports in some cases. I think it messes with webpack in a way that is very fragile.

There was a time when we could only import HTML, but now we can import JSON directly. That should allow us to eliminate `raw.macro` completely.

src/components/Toolbar.tsx
```diff
@@ -205,11 +205,13 @@ const Toolbar = ({ dark, fontSize, toolbarOverlay, scrollPrioritized, showTopCon
                     exec(store.dispatch, store.getState, e, { type: 'toolbar' })
                   }}
                 >
-                  <SVG style={{
-                    fill: !isActive || isActive(store.getState) ? fg : 'gray',
-                    width: fontSize + 4,
-                    height: fontSize + 4,
-                  }} />
+                  <SVG
+                    getState={store.getState}
```
raineorshine, line 209: Is there an alternative to passing 'getState` as a prop? We want to constrain the SVG api.

src/util/compareThought.ts
```diff
@@ -131,5 +131,16 @@ export const compareReasonable = makeOrderedComparator<string>([
   (a, b) => compareReasonableText(makeReasonable(a), makeReasonable(b)),
 ])
 
+/** A comparator that sorts in descending order. */
+export const compareReasonableDescending = (a: string, b: string): ComparatorValue => {
+  const comparatorValue = compareReasonable(a, b)
+  return comparatorValue !== 0 ?
+    comparatorValue === -1 ? 1 : -1
+    : 0
```
raineorshine, line 139: Even better, create a higher order function that allows you to reverse a comparator:

```js
const reverse = comp => (a, b) => comp(b, a)
const compareThoughtDescending = reverse(compareThought)
```

src/shortcuts/toggleSort.tsx
```diff
@@ -17,6 +20,44 @@ const Icon = ({ size = 20, style }: IconType) => <svg version='1.1' className='i
...
+    else {
+      return 'Alphabetical'
+    }
+  }
+
+  return getSortType() === 'Alphabetical/desc' ? <IconDesc size={size} style={style}/> : <IconAsc size={size} style={style}/>
+
+}
+
+// eslint-disable-next-line jsdoc/require-jsdoc
```
raineorshine, line 55: This is not a canonical function like `mapStateToProps` so it deserves a JSDOC comment.

raineorshine (changes requested):
Ah, you're not familiar yet with our shorthand for descendants! My apologies for being unclear. "Alphabetical/desc" means:

- Alphabetical
  - desc

Otherwise looks very good!

src/util/compareThought.ts
onurpolattimur: Brilliant!

src/components/Toolbar.tsx
onurpolattimur: I am not sure but we can add `additionalSVGProps?: (getState: () => State) => any` to Shortcut inferface. Besides this, we need to add additionalProps to `Icon`.
In`toggleSort.tsx` we can decide which Icon we want to render and return a json object like {icon: IconDesc} and we can pass this to Icon (SVG) in `Toolbar.tsx`

onurpolattimur:
> Ah, you're not familiar yet with our shorthand for descendants! My apologies for being unclear. "Alphabetical/desc" means:
> 
> * Alphabetical
>   
>   * desc
> 
> Otherwise looks very good!

There are some things that I can not understand well. 
First of all, if I add desc as descandant of 'Alphabetical', I still need to return something like 'AlphabeticalDesc' from getSortPreference. 
And I need to write some conditions into toggleSort.tsx. For example: If the next sort preference is 'AlphabeticalDesc' then toggleAttribute with key '=sort' and value 'Alphabetical' and then toggleAttribute with key 'Alphabetical' and value 'desc'. If the next sort preference is equal to AlphabeticalDesc and also the global sort is AlphabeticalDesc then toggleAttribute with key 'Alphabetical' and value 'desc' to toggle off. I am not sure but I think there will be more than 3 if blocks.

And also in getSortPreference, we are getting global preference with getSetting(state, ['Global Sort']) that returns a string. But in this case, I think we can not use this function to get Global Sort preference. We need to call again if it is Alphabetical.

I think that the current version is much more understandable. When I look at the initialSettings.html there will be 3 options for Global Sort. 

<li>Global Sort
    <ul>
      <li>=readonly</li>
      <li>=options
        <ul>
          <li>None</li>
          <li>Alphabetical</li>
          <li>AlphabeticalDesc</li>
        </ul>
      </li>
      <li>None</li>
    </ul>
</li>

But if we make these changes, it will look complicated.

<li>Global Sort
    <ul>
      <li>=readonly</li>
      <li>=options
        <ul>
          <li>None</li>
          <li>Alphabetical
              <ul>
                  <li>=options
                      <ul>
                          <li>asc</li>
                          <li>desc</li>
                      </ul>
                  </li>
                  <li>desc</li>
          </li>
        </ul>
      </li>
      <li>None</li>
    </ul>
</li>

raineorshine:
I hear what you're saying and you have a good point. From a simplicity perspective, a flat structure would be easier to implement and easier to import.

The thing about **em** is that deep hierarchies are a key part of its design. We intentionally want people to de-structure sentences and render their thoughts deeper into the hierarchy. This increases the connectivity between thoughts and provides more surface area for adding, editing, and re-organizing thoughts. We could make an exception this time for convenience, but I think it's important to stay consistent if we're trying to teach users about this new way of structuring their thoughts.

In addition to the principle described above, there is a practical reason in that parsing the sort direction as its own thought is more general and will be easier to apply when we have other sort criteria, or user-defined sort criteria, rather than adding two sort keys for every field, e.g. `Created`, `CreatedDesc`, `Updated`, `UpdatedDesc`, etc. Another way to think about it is that if you were programming a sort configuration you would probably have two separate variables `sortField` and `sortDirection` rather than combining them into a single field that needed to be parsed. The latter is more a normalized data structure. **em** encourages normalized thought structures.

raineorshine:
> First of all, if I add desc as descandant of 'Alphabetical', I still need to return something like 'AlphabeticalDesc' from `getSortPreference`.
> And I need to write some conditions into `toggleSort.tsx`. For example: If the next sort preference is 'AlphabeticalDesc' then toggleAttribute with key '=sort' and value 'Alphabetical' and then `toggleAttribute` with key 'Alphabetical' and value 'desc'. If the next sort preference is equal to AlphabeticalDesc and also the global sort is AlphabeticalDesc then `toggleAttribute` with key 'Alphabetical' and value 'desc' to toggle off. I am not sure but I think there will be more than 3 if blocks.
> 
> And also in `getSortPreference`, we are getting global preference with`getSetting(state, ['Global Sort'])`that returns a string. But in this case, I think we can not use this function to get Global Sort preference. We need to call again if it is Alphabetical.

Yes, all of this logic gets more complicated because the sort preference is no longer represented as a single string.

onurpolattimur:
Hi @raineorshine, I have made the changes. Can you review it, please? Now our sort preference is represented with two fields that are type and direction. There are some points that do not satisfy me, especially in `toggleSort.tsx`. If you find a better way, we can implement that.

src/selectors/prevSibling.ts
```diff
@@ -10,14 +11,13 @@ import { Child, Context, Index, ThoughtContext } from '../types'
...
   const { showHiddenThoughts } = state
-  const sortPreference = getSortPreference(state, context)
   const contextViewActive = isContextViewActive(state, context)
 
   /** Gets siblings of a context. */
   const getContextSiblings = () => getContextsSortedAndRanked(state, head(context))
 
   /** Gets siblings of thought. */
-  const getThoughtSiblings = () => (sortPreference === 'Alphabetical' ? getChildrenSorted : getChildrenRanked)(state, context)
+  const getThoughtSiblings = () => (isSortPreferenceAlphabetical(state, context) ? getChildrenSorted : getChildrenRanked)(state, context)
```
onurpolattimur, line 20: I think we can use new isContextSorted function here.

raineorshine (changes requested):
Thanks for your thoroughness, I don't see anything that was missed. The tests are great and the logic is flawless! Just a few architectural recommendations.

src/util/compareThought.ts
```diff
@@ -131,5 +131,11 @@ export const compareReasonable = makeOrderedComparator<string>([
   (a, b) => compareReasonableText(makeReasonable(a), makeReasonable(b)),
 ])
 
+/** Get reverse of the given comparator. */
+const reverse = (comparator: (a: Child, b:Child) => ComparatorValue) => (a: Child, b: Child) => comparator(b, a)
```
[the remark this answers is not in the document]
onurpolattimur: done

src/util/__tests__/compareThought.ts
```diff
@@ -108,5 +110,33 @@ describe('compareReasonable', () => {
...
   })
+})
+
+describe('compareReasonableDescending', () => {
+
+  /**
+   * Build Child object for tests.
+   */
+  const buildChild = (value: string): Child =>
+    ({ archived: undefined, id: '0', lastUpdated: undefined, rank: 0, value: value })
```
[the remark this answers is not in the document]
onurpolattimur: Sorry, I missed these fields are optional.

src/selectors/isSortPreferenceAlphabetical.ts
```diff
@@ -0,0 +1,11 @@
+import { getSortPreference } from '../selectors'
+import { State } from '../util/initialState'
+import { Context } from '../types'
+
+/** Get whether the given context sorted alphabetically. */
+const isSortPreferenceAlphabetical = (state: State, context: Context) => {
```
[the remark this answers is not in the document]
onurpolattimur: I removed them.

src/selectors/getSortPreference.ts
```diff
@@ -1,14 +1,55 @@
 import { getSetting, getAllChildren } from '../selectors'
 import { State } from '../util/initialState'
-import { Context } from '../types'
+import { Context, SortPreference } from '../types'
 import { unroot } from '../util'
 
+/**
+ * Todo:: I don't like this.
+ */
+const getSortDirectionFromString = (sortDirection: string | null) => {
```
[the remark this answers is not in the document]
onurpolattimur: 🥳  perfect!

onurpolattimur:
Hi @raineorshine, I have pushed the changes. Can you review it, please? 
Thanks!

src/selectors/getCursorSortDirection.ts
```diff
@@ -0,0 +1,15 @@
+/** A selector that returns the sort direction of the cursor context. */
+import { State } from '../util/initialState'
+import { getSortPreference, simplifyPath } from './index'
+import { HOME_PATH } from '../constants'
+import { pathToContext } from '../util'
+
+/** Get sort direction of cursor. */
+const getCursorSortDirection = (state: State) => {
```
raineorshine, line 8: It would be better to define this inline in the `toggleSort` module since it's not used anywhere else. This helps reduce the surface area of the internal API.

src/types.ts
```diff
@@ -232,6 +240,7 @@ export interface Shortcut {
   svg?: (icon: Icon) => ReactNode,
   canExecute?: (getState: () => State, e: Event) => boolean,
   exec: (dispatch: Dispatch, getState: () => State, e: Event | GestureResponderEvent | React.MouseEvent, { type }: { type: string }) => void,
+  additionalIconProps?: (getState: () => State) => any,
```
raineorshine, line 243: `additionalIconProps` can be removed.

raineorshine (changes requested):

raineorshine:
In the future I recommend making smaller commits and providing a descriptive commit message for each one. It's a good habit to get into and even if it doesn't matter a lot of the time it really pays off sometimes, for example allowing me to review code one commit at a time, or allowing more granular identification of regressions.

raineorshine (approved):

raineorshine:
Thanks!

Issue #1110, opened 2021-04-25
Toolbar button click animation


raineorshine:
This will make user actions more visible during mobile screen captures. 

- On touchstart/mousedown: animate toolbar button down a few pixels
- On touchend/mouseup: animate toolbar button back up

onurpolattimur:
Adding background may be better especially for mobile. Even we move down 10px the button when I touch down the toolbar item I can not see still this item because my finger will be on the button.

![image](https://user-images.githubusercontent.com/23136437/116757083-0bdd7f80-aa16-11eb-824a-cd2dde088168.png)

raineorshine:
> If the purpose of this feature is to make the user action more visible when we capture the screen, it is ok to move the button down.

Yes, this is the main purpose.

> Even we move down 10px the button when I touch down the toolbar item, I can not see still this item because my finger will be on the button.

Yeah, it will be pretty hidden for the mobile user.

> So we can add background.

What do you mean?

onurpolattimur:
Okay good. I have pushed the changes with #1116. Now, we move down the button 10px, it is really useful while capturing the screen.

I mean, if we also want to give feedback to users while pressing the button, we can add background to the toolbar item. Because the user's finger will be on the button, the user will not see that the button is going down.

raineorshine:
Do you mean set a background color on the toolbar item as they press it?
Wouldn't that be covered by their finger too?

onurpolattimur:
> Do you mean set a background color on the toolbar item as they press it?
> Wouldn't that be covered by their finger too?

Yes, but the user can see the place the top of the button as you can see in my screenshot that I have posted before.

raineorshine:
True. I don't think a solid background color is the aesthetic I want though. I'll keep this in mind but for now it's low priority so I think we should come back to it later.

PR #1116, opened 2021-05-01
Move down toolbar item when pressing the button

onurpolattimur:
Adds feature #1110

### DIFF
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -55,7 +55,7 @@ const Toolbar = ({ dark, fontSize, toolbarOverlay, scrollPrioritized, showTopCon
   const [lastScrollLeft, setLastScrollLeft] = useState<number | undefined>()
   const [leftArrowElementClassName = 'hidden', setLeftArrowElementClassName] = useState<string | undefined>()
   const [rightArrowElementClassName = 'hidden', setRightArrowElementClassName] = useState<string | undefined>()
-  const [pressingToolbarId, setPressingToolbarId] = useState('')
+  const [pressingToolbarId, setPressingToolbarId] = useState<string | null>(null)
   const fg = dark ? 'white' : 'black'
   const arrowWidth = fontSize / 3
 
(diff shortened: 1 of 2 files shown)


src/components/Toolbar.tsx
```diff
@@ -55,7 +55,7 @@ const Toolbar = ({ dark, fontSize, toolbarOverlay, scrollPrioritized, showTopCon
   const [lastScrollLeft, setLastScrollLeft] = useState<number | undefined>()
   const [leftArrowElementClassName = 'hidden', setLeftArrowElementClassName] = useState<string | undefined>()
   const [rightArrowElementClassName = 'hidden', setRightArrowElementClassName] = useState<string | undefined>()
-
+  const [pressingToolbarId, setPressingToolbarId] = useState('')
```
raineorshine, line 58: Default state should be `null` which makes it more explicit. Empty string is more error prone.

```ts
useState<null | string>(null)
```

src/components/Toolbar.tsx
```diff
@@ -195,12 +196,17 @@ const Toolbar = ({ dark, fontSize, toolbarOverlay, scrollPrioritized, showTopCon
                 <div
                   key={name}
                   id={id}
+                  style={pressingToolbarId === id ? { paddingTop: '10px' } : {}}
```
raineorshine, line 199: Let's add a `transition` so it's a bit more smooth.

raineorshine (changes requested):

src/components/Toolbar.tsx
```diff
@@ -195,12 +196,17 @@ const Toolbar = ({ dark, fontSize, toolbarOverlay, scrollPrioritized, showTopCon
                 <div
                   key={name}
                   id={id}
+                  style={pressingToolbarId === id ? { paddingTop: '10px', transition: 'padding 0.2s ease 0s' } : {}}
```
raineorshine, line 199: If `transition` is removed when `pressingToolbarId` is no longer equal to `id`, won't that remove the return animation? I think you want `transition` to always be set, and only remove `paddingTop`. Unless I am mistaken and it already transitions both directions even when `transition` is removed.

onurpolattimur: Yes, you are right! It will not apply animation for returning. I fixed that now.

raineorshine (approved):

PR #1120, opened 2021-05-02
Fix childrenFilterPredicate to work with forcedChildren

### LINKED ISSUE #1092: Allow empty Algolia config
Currently the app cannot load any thoughts from the remote when the Algolia config is left empty. This makes it harder for new developers to get started on the project or to stand up staging sites.

Instead, if the Algolia config is missing, everything should still work except the search. If the user attempts to search, an alert should be shown that says that search functionality is not enabled.

shresthabijay:
fix #1092

# Changes
- Fix `childrenFilterPredicate` to use pass `showContexts` instead of `isForcedChildren`.
- Do not initialize remote search if required configs are not available.

### DIFF
--- a/src/components/Subthoughts.tsx
+++ b/src/components/Subthoughts.tsx
@@ -375,7 +375,7 @@ export const SubthoughtsComponent = ({
     return cursor[depth] && cursor[depth].rank === child.rank
   }) : 0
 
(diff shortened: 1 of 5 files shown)


shresthabijay:
@raineorshine Previously `childrenFilterPredicate` failed with forced children. I fixed it. If the remote search is not available it will fall back to local search. Please review the changes. Thanks!

raineorshine (approved):
Thanks!

PR #1122, opened 2021-05-03
Priortize plainText over hmtlText that is passed to importText

### LINKED ISSUE #1049: Safari: Import duplicate uncle doesn't work
Safari only

## Steps to Reproduce

Paste the following:

```
- a
  - b
- b
```

## Current Behavior

The pasted thoughts are either invisible or failed to paste.

## Expected Behavior

The pasted thoughts should be added correctly and visible.

shresthabijay:
fixes #1049

# Cause of issue

In the `Editable`component, `plainText` was passed to `importText` function only if it is an HTML. Basically prioritizing `htmlText` over `plainText` made it more complicated. Even the indented text was passed as HTML. In chrome the `htmlText` has  a `meta` tag at the beginning. Since `importText` checks if the text is an HTML by the presence of a proper closing tag at the beginning, so in the Chrome it was not detected as an HTML and was parsed as blocks. That's why it was working properly. However, it is not the case for the Safari. So it was causing the inconsistent imports.

# Solution

We simply prioritize `plainText`. Rest is handled by already available logic in `importText`

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -40,7 +40,6 @@ import {
   headValue,
   isDivider,
   isElementHiddenByAutoFocus,
-  isHTML,
   isURL,
   pathToContext,
   setSelection,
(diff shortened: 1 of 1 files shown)


shresthabijay:

raineorshine:
Unfortunately this breaks HTML paste. That is, copying and pasting from a web page with formatting, rather than copying and pasting HTML in plaintext. We can't prioritize `plainText` over `htmlText` because `plainText` is lossy.

Try to copy the below thoughts and paste directly into **em**:

- Hello
  - Good**bye**

If we prioritize `plainText` we lose the formatting and structure.

Note: We don't have test coverage of `onPaste` unfortunately. We would have to find a way to mock `e.clipboardData.getData`.

Issue #981, opened 2021-01-09
Mobile: Caret moves to beginning on cursor change


raineorshine:
## Steps to Reproduce

1. Create thought `a`
1. Create subthought `foo bar`
1. Set the cursor on `a`
1. Tap in between `foo` and `bar`

## Current behavior

The caret goes to the beginning of `foo bar`. (This does not occur when moving between siblings.)

If you look closely, you can see that the caret moves to the correct position for a split second then gets reset to the beginning. 

## Expected Behavior

The caret should move to in between `foo` and `bar`, where the user tapped. 

When the user taps on a thought and the cursor changes from parent to child, or child to parent, the caret should move to the point that the user tapped.

onurpolattimur:
Here is what I have found so far. 

The Editable component is rendered twice after we tap in between foo and bar. After the second render, the caret position is set the zero. Because ` { offset: cursorOffset || state.cursorOffset || 0 }` is always zero. That is why caret at the beginning.

It is after the second render because `isEditing` is false and so, it doesn't enter the if block after the first render.
https://github.com/cybersemics/em/blob/fdd917169a202b537bd987ac858ccbbb8747c2fc/src/components/Editable.tsx#L367-L389

If you block the second render, you will see it works. I blocked the second render by adding the code below(just for debug) at the end of the `Editable.tsx`.

```
export default connect()(React.memo(Editable, (prev: any, next: any) => {
  console.log(prev, next)
  return !prev.isEditing && next.isEditing
}))

```

Same issue occurs on desktop if we add some delay before calling `setSelectionToCursorOffset`

I think we only need ` setSelectionToCursorOffset()` for newly created thoughts. I am not sure but maybe we can add a local state like 'isFirstRender' and we can set true this in useEffect. I tried and I can not see any problem.

raineorshine:
@onurpolattimur Thank you! That is super helpful information.

Before considering blocking the second render, why is it rendering twice to begin with? If we can remove the extra render we could fix the bug and improve performance.

onurpolattimur:
I am so sorry, I think I was wrong. It is not rendered twice, the first render log was for editable element contains 'a'. 

But I think, I found the root cause. As I said before, we always set caret position to zero. We need to continue to set zero for newly created thoughts to set caret.

```jsx
{ offset: cursorOffset || state.cursorOffset || 0 }
```

[...]

PR #1125, opened 2021-05-06
Don't set caret when user click an Editable

onurpolattimur:
fixes: #981 
_____
> #### Why do we set the caret?
> 
> On events that don't contain user clicks, we should set the caret for editable elements. Because the user doesn't click anywhere but we want the caret to be on a thought. Here are some examples of this situation;
> 
> - When we refresh the page, the caret should be on the thought in the path.
> - When we delete a thought, the caret should be on the previous thought.
> - When we add a thought, the caret should be on the newly created thought.
> ....
> #### What should we do?
> 
> We should not set the caret if we want to see that the caret on the correct position where the user clicks. Because the browser already set the caret where the user clicks. The only thing we should know before setting caret is whether the user clicked or not.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash'
-import React, { useEffect, useRef } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import { Dispatch } from 'redux'
 import { connect } from 'react-redux'
 import { unescape } from 'html-escaper'
@@ -181,6 +181,7 @@ const Editable = ({ disabled, isEditing, simplePath, path, cursorOffset, showCon
   // store the old value so that we have a transcendental head when it is changed
   const oldValueRef = useRef(value)
   const editableNonceRef = useRef(state.editableNonce)
+  const [isTapped, setIsTapped] = useState(false)
 
   useEffect(() => {
     editableNonceRef.current = state.editableNonce
@@ -362,7 +363,6 @@ const Editable = ({ disabled, isEditing, simplePath, path, cursorOffset, showCon
     //     focusNode: !!focusNode,
     //   })
     // }
-
     // allow transient editable to have focus on render
     if (transient || (
       isEditing &&
@@ -370,7 +370,8 @@ const Editable = ({ disabled, isEditing, simplePath, path, cursorOffset, showCon
       !noteFocus &&
       contentRef.current &&
       (cursorWithoutSelection || isAtBeginning) &&
-      !dragHold
+      !dragHold &&
+      !isTapped
     )) {
       /*
         When a new thought is created, the Shift key should be on when Auto-Capitalization is enabled.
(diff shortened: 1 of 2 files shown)


src/components/Editable.tsx
```diff
@@ -388,6 +389,8 @@ const Editable = ({ disabled, isEditing, simplePath, path, cursorOffset, showCon
       }
     }
 
+    isTapped && setIsTapped(false)
```
raineorshine, line 392: Let's stick with explicit `if` statements on expressions with side effects. I'm not strictly opposed to the shorthand, but it deviates from the convention we have been using so I'd like to stay consistent for now.

src/components/Editable.tsx
```diff
@@ -611,6 +614,7 @@ const Editable = ({ disabled, isEditing, simplePath, path, cursorOffset, showCon
 
     const isHiddenByAutofocus = isElementHiddenByAutoFocus(e.target as HTMLElement)
     const editingOrOnCursor = state.editing || equalPath(path, state.cursor)
+    setIsTapped(true)
```
raineorshine, line 617: When adding behavior like this that may not be obvious to other developers, it might be good to leave a descriptive comment and link to the issue.

raineorshine (changes requested):
Tested on mobile, works great! Very low footprint, well done.

I just found one regression on desktop.

## Steps to Reproduce

1. Create a thought
2. Refresh the page
3. Click on the thought

## Current Behavior

`5bd9895a8`

Cursor moves to thought but caret is not set.

## Expected Behavior

`dev`

Caret should always be set on the cursor on desktop.

onurpolattimur:
Hi @raineorshine, I found the root cause of the issue you wrote. We are using e.preventDefault() in onTap. Because `state.editing` is false after refreshing the page `editingOrOnCursor` is also false, so it is preventing onFocus event and caret can not be set by the browser. I fixed, can you review it again?

raineorshine (approved):
Works great! Thanks!

Do you think it would be possible to produce a failing e2e test case?

https://github.com/cybersemics/em/blob/dev/src/browser.ts#L5-L6

onurpolattimur:
> Works great! Thanks!
> 
> Do you think it would be possible to produce a failing e2e test case?
> 
> https://github.com/cybersemics/em/blob/dev/src/browser.ts#L5-L6

Hi @raineorshine, I added the e2e test for the issue that you wrote above. But unfortunately, we can not write an e2e  test for the issue #981, because it occurs only on the mobile browsers.  It doesn't matter whether isTouch true or false. We can not produce this issue on our desktop browser with forcing isTouch true .

raineorshine:
> > Works great! Thanks!
> > Do you think it would be possible to produce a failing e2e test case?
> > [https://github.com/cybersemics/em/blob/dev/src/browser.ts#L5-L6](https://github.com/cybersemics/em/blob/dev/src/browser.ts?rgh-link-date=2021-05-08T18%3A47%3A41Z#L5-L6)
> 
> Hi @raineorshine, I added the e2e test for the issue that you wrote above.

I'm not able to get the test to fail on `dev`. Do you experience otherwise?

> But unfortunately, we can not write an e2e test for the issue #981, because it occurs only on the mobile browsers. It doesn't matter whether isTouch true or false. We can not reproduce this issue on our desktop browser with forcing isTouch true .

Got it, makes sense.

Issue #1022, opened 2021-02-07
Expand on drag hover


raineorshine:
When dragging a thought over a context for 1 second, the context should expand. 

e.g. 

- a [cursor]
  - b
- c
  - d
  
If the cursor in on `a`, then `c` will be collapsed. Dragging `b` over `c` so that the blue drop indicator is in place to move `b` to `c/b` then, without releasing for 1 second, `c` should be expanded. 

When the drag ends, expansion should go back to normal.

PR #1127, opened 2021-05-07
Expand on hover

### LINKED ISSUE #1023: Expand parent on drag
Related to #1022

When dragging a thought above the first visible thought (closest to the root) for 1 second, the hidden context and its children should become visible. 

e.g. 

- a
  - b
    - c
      - d [cursor]
  - e
  
If the cursor in on `d`, then `a` and `e` will be invisible. Dragging `d` above `b` so that the blue drop indicator is in place to move `d` to `a/d` then, without releasing for 1 second, `a` and `e` should become visible.

shresthabijay:
fixes #1023 #1022

# Changes

- Expand context when hovered on the empty drop. Works with nested expansions.
- Make parent and its children visible when visible thought nearest to the root hovers. Thoughts can be dropped into the parent and siblings after the expansion is active.

https://user-images.githubusercontent.com/25879100/117482149-b0fed780-af83-11eb-9554-27e52dc3ba8a.mov

@raineorshine  There are few issues. On expanding parent, I have translated the hidden parent and the children to the left for now. It works for desktop but may not work for mobile as there is less space. Also, parent expansion doesn't work at the root. Need to figure it out. I will do that after you review the current changes.


shresthabijay:
@raineorshine I have added the feature. Please review it. Thanks!

raineorshine (changes requested):
Thanks Bijay! It looks really great so far. It's a challenging task. I'm not as experienced with hooks as you, but I looked over the code and it seems like a reasonable approach.

Here are the issues I am seeing with the behavior:

- Expand parent not usable on mobile (as you pointed out, but just confirming since it's critical).
- When expanding on drag hover, it only expands a single level even if there are single children to expand. It should expand the hovered thought the same way that the `expandThoughts` algorithm already works, i.e. with special cases for single children, pinOpen, etc. It should use the existing algorithm though so that if the algorithm changes this hover expand behavior also changes. It doesn't look like `expandThoughts` is used in this implementation so I'm not sure how best to integrate it. It's my fault for not specifying it in the initial requirements. It wasn't obvious to me until I tried it out. When I tested it, it felt weird that the thoughts were not expanding as they normally do, so I think it's important to keep that consistent for the user experience.
- When expanding on drag hover, it collapses too quickly which makes it a bit hard to use. Would it be possible to only collapse the hover expansion if the drag ends, or if another thought is hover expanded? If the user drags to another thought but does not hover over it for the requisite 1s, it shouldn't collapse the current hover expansion. I think this will make it feel less choppy.

shresthabijay:
> * When expanding on drag hover, it only expands a single level even if there are single children to expand. It should expand the hovered thought the same way that the `expandThoughts` algorithm already works, i.e. with special cases for single children, pinOpen, etc. It should use the existing algorithm though so that if the algorithm changes this hover expand behavior also changes. It doesn't look like `expandThoughts` is used in this implementation so I'm not sure how best to integrate it. It's my fault for not specifying it in the initial requirements. It wasn't obvious to me until I tried it out. When I tested it, it felt weird that the thoughts were not expanding as they normally do, so I think it's important to keep that consistent for the user experience.

Understood! I should have thought of it too.

shresthabijay:
- On hover on the top. It will expand the parent just. It works on mobile too.
- Used `expandThoughts` with the hover expansion logic. It worked like a charm.
- Expanded context will only collapse after drag ends.

# Hover Expansion (IOS Simulator)
https://user-images.githubusercontent.com/25879100/117546328-a3ac2080-b049-11eb-91c3-792021bd077b.mov

shresthabijay:
> @raineorshine I have made the changes. Please review them. Thanks!
> 
> * On hover on the top. It will expand the parent just. It works on mobile too.
> * Used `expandThoughts` with the hover expansion logic. It worked like a charm.
> * Expanded context will only collapse after drag ends.
> 
> # Hover Expansion (IOS Simulator)
>  Screen.Recording.2021-05-08.at.21.50.03.mov

raineorshine:
The video on that screen capture is not playing here, but it has a great soundtrack 😂.

It worked when I downloaded it. Let me know if it doesn't work for you.

shresthabijay:
> The problem I’m having is that it only expands up once. This makes it impossible to drop something further up, or in a different part of the thoughtspace that is farther away.

Sure. I will make it expand up to farther contexts too.

shresthabijay:
@raineorshine I have made the changes. I realized that the hooks were overkill for this task. Also, they needed to be used with each `Thoughts`. So I refactored one of the hooks to the subscriber pattern. If you approve of this pattern I will refactor other hook as well. Please review the changes. Thanks!

# DEMO

https://user-images.githubusercontent.com/25879100/117840439-e1a08300-b29b-11eb-8603-0870da3f0ddb.mov

src/hooks/useExpandOnHover.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }, [trigger])
+
+  return value
+}
+
+/**
+ * Handles expansion of the context due to hover on the thought's empty drop.
+ */
+export const useExpandBottomOnHover = ({ path }: Options, expandDelay: number) => {
```
raineorshine, line 46: The functionality looks good, just a couple questions:

1. `useExpandBottomOnHover` is tightly coupled to the `dragInProgress` reducer (via `state.hoveringPath` and `state.dragInProgress`) but this fact is quite hidden when you look at the hook and its arguments. I guess all hooks that use `useSelector` have that risk. Any ideas how to make these dependencies more explicit? We could pass them as arguments, but that just pushes `useSelector` to the component. Is it just a matter of clearer documentation that `useExpandBottomOnHover` only works with `dragInProgress`?

2. Could you clarify why this must be implemented as a hook instead of an action-creator? It looks like it just dispatches `expandBottom` based on `state.hoveringPath` and `state.dragInProgress`. Could this logic be called from the `dragInProgress` action-creator?

src/initialize.ts
```diff
@@ -48,6 +50,8 @@ const withState = <T, R>(f: (state: State, ...args: T[]) => R) =>
 /** Imports text on given unranked path. Used in puppeteer testing. */
 const importTextFirstMatch = (unrankedPath: string[], text: string) => store.dispatch(importOnFirstMatchPathActionCreator(unrankedPath, text))
 
+store.subscribe(expandOnHoverTop)
```
raineorshine, line 53: Could you explain a little more the reason for using the subscription api? Why not use a middleware or storeEnhancer? 

Or even simpler, make the appropriate updates in `dragInProgress` (basically the same question as asked above, though I may be missing an obvious reason why that is not viable)?

> If you approve of this pattern I will refactor other hook as well.

If I am understanding you correctly, you suggest the use of the subscription api for both expanding top and bottom, is that right?

raineorshine (commented):
Hey Bijay! Bear with me, I have a lot of questions 😄.

Tested on desktop and iOS and it works great! I think I just need to understand the architectural decisions a little better. Thanks!

src/components/ExpandOnHover.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useExpandBottomOnHover } from '../hooks/useExpandOnHover'
+import { Path } from '../types'
+
+interface Props {
+  path?: Path,
+  depth: number,
+}
+
+/**
+ * It is separated into this component to isolate the affects of re-renders which can be caused by hook.
```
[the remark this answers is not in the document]
shresthabijay: When one of the hooks rerenders, it rerenders the component. It doesn't depend if the final result returned by the component is the same. Also, I need to refactor this. We don't need a hook for this now. Everything has been moved to the redux.

src/util/initialState.ts
```diff
@@ -56,9 +56,11 @@ export interface State {
   editingValue: (string | null),
   error?: string | null,
   expanded: Index<Path>,
+  expandedBottom: Index<Path>,
+  expandHoverBottomPaths: Index<Path>,
```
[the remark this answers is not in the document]
shresthabijay: `expandHoverBottomPaths` are the paths that have active bottom hover i.e their context is expanded. `expandedBottom` is the paths are the same as `expanded`. But it results from `expandHoverBottomPaths`.

src/hooks/useExpandOnHover.ts
shresthabijay: > 2\. Could you clarify why this must be implemented as a hook instead of an action-creator? It looks like it just dispatches `expandBottom` based on `state.hoveringPath` and `state.dragInProgress`. Could this logic be called from the `dragInProgress` action-creator?

As you can see, `useExpandBottomOnHover` takes `path` as an argument.  The hook needs to maintain a timer for each thought. It would have been hard to maintain a timer for all thoughts with the action creator. That was the reason I implemented it as the hook. But I think it is possible to refactor this to use only a single timer just like with expand hover top. So I will change this to an action creator.

src/initialize.ts
shresthabijay: > Could you explain a little more the reason for using the subscription api? Why not use a middleware or storeEnhancer?

You are correct. It can be easily converted to an action creator.

shresthabijay:
@raineorshine I have made the changes. I converted them to action creators. Please review the changes. Thanks!

raineorshine (changes requested):
Excellent! I like that. Now it's more similar to our existing architecture.

All set to remove extra code and add JSDOCs. Thanks!

src/action-creators/expandOnHoverBottom.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const state = getState()
+
+  const { hoveringPath, hoverId, expandHoverBottomPaths, dragInProgress } = state
+
+  const hoveringContext = hoveringPath && pathToContext(hoveringPath)
+
+  const shouldExpand = hoverId === DROP_TARGET.EmptyDrop && hoveringPath && getChildren(state, pathToContext(hoveringPath)).length > 0
+
+  // eslint-disable-next-line jsdoc/require-jsdoc
```
raineorshine, line 24: JSDOC

src/action-creators/expandOnHoverBottom.ts
```diff
@@ -0,0 +1,60 @@
...
+  const shouldCancel = !dragInProgress
+
+  if (shouldCancel) dispatch(clearExpandBottom())
+
+  if (!shouldExpand) {
+    clearTimeout()
+    return
+  }
+
+  // eslint-disable-next-line jsdoc/require-jsdoc
```
raineorshine, line 41: JSDOC

src/action-creators/expandOnHoverTop.ts
```diff
@@ -0,0 +1,91 @@
...
+  const shouldCancel = !dragInProgress && expandHoverTopPath
+
+  if (shouldCancel) dispatch(expandHoverTop({ path: null }))
+
+  if (!shouldExpand) {
+    clearTimeout()
+    return
+  }
+
+  // eslint-disable-next-line jsdoc/require-jsdoc
```
raineorshine, line 71: JSDOC

src/components/Thought.tsx
```diff
@@ -45,6 +45,7 @@ import {
   isDescendantPath,
   isRoot,
   unroot,
+  // isDescendant,
```
raineorshine, line 48: ✗

src/components/Thought.tsx
```diff
@@ -250,23 +257,35 @@ const dragCollect = (connect: DragSourceConnector, monitor: DragSourceMonitor) =
...
+
+  const distance = cursor ? cursor?.length - thoughtsTo.length : 0
+
+  /** If the epxand hover top is active then all the descenendants of the current active expand hover top path should be droppable. */
+  const isExpandedTop = () => expandHoverTopPath && path.length > expandHoverTopPath.length && isDescendantPath(path, expandHoverTopPath)
+
+  const isHidden = distance >= visibleDistanceAboveCursor(state) && !isExpandedTop()
+
+  // check if thought is the visible ancestor of cursor nearest to the root
+  // const isNearestVisibleThought = distance === distanceAboveCursor && isCursorAncestor()
```
raineorshine, line 276: ✗

src/action-creators/expandOnHoverTop.ts
```diff
@@ -0,0 +1,91 @@
...
+ * On expand the context and its hidden children should be visible.
+ */
+const expandOnHoverTop = (): Thunk => (dispatch, getState) => {
+
+  const state = getState()
+
+  const { hoveringPath, expandHoverTopPath, dragInProgress } = state
+  const shouldExpand = shouldAllowActiveHoverTop(state)
+
+  // eslint-disable-next-line jsdoc/require-jsdoc
```
raineorshine, line 53: JSDOC

raineorshine (changes requested):
Let's use JSDOC on all functions. I would disable them only for canonically named functions like `mapStateToProps`, but even though it is arguable that they should always be set so we are not disabling the linter.

Everything else looks good!

shresthabijay:

raineorshine (approved):

Issue #1029, opened 2021-02-10
Cursor moves to thought hidden by autofocus


raineorshine:
This is a regression. It would be helpful to identify the originating commit. 

## Steps to Reproduce

```
- a
  - b
    - c
- d
``` 

1. Set the cursor on `c` so `d` disappears
2. Tap 1 line height below `c`, to the right of where `d` would be. 

## Current Behavior

Cursor moves to `d`. 

## Expected Behavior

Cursor should move to `b`. 

Tapping on empty space with no visible thoughts to the left or right should always activate `cursorUp`.

raineorshine:
Only occurs on mobile, but you can reproduce it on desktop by forcing `isTouch = true`.

raineorshine:
@shresthabijay Thank you for the solution! We should still do some forensic work to understand the bigger picture of this. When there is a discrete regression, it's important to identify the commit that broken to inform our fix. I have often fixed a regression only to realize I compromised the original behavior in some way. Usually it's better to identify why something stopped working rather than adding additional code to make it work again.

I put it softly in the description, but I probably should have made it clear that it was a requirement :). Could you find out which commit broke this and why? Thanks!

shresthabijay:
@raineorshine Understood. I opted to go with the fix as it was fairly easy. I will get back to you after I identify the commit that created the regression.

shresthabijay:
@raineorshine The regression seems to have been introduced from this commit 
https://github.com/cybersemics/em/commit/e12699e3524da5ee9bdcd9714e3429a59683fba7. 

`onTap` replaced `onMouseDown` for device with touch screens. Previously `cursorBack` was being called even on touch screen devices because `onMouseDown` was being called after `onTapEnd`. This is because touch events also emit artificial mouse events on succession unless touch events are prevented.

`onTap` doesn't have `cursorBack` logic. With this commit `onTap` is being called on `onmousedown` event for touch screens, thus this regression was introduced.

raineorshine:
> @raineorshine The regression was introduced from this commit [e12699e](https://github.com/cybersemics/em/commit/e12699e3524da5ee9bdcd9714e3429a59683fba7).

Great, thanks! 

Reference: #946

[...]

PR #1128, opened 2021-05-07
Activate cursorBack when clicking the hidden thought

onurpolattimur:
fixes: #1029 

____

This issue occurs because we are clicking the editable area of hidden thoughts. I added border for .editable class to show you the boundaries of editable elements.

image

There is already a function written for handle this situation `isElementHiddenByAutoFocus` . But the problem was in tap handler, it was not entering the if block because `editingOrOnCursor` is true when we click the area of `d`

https://github.com/cybersemics/em/blob/e30375659d5d5ad98c5df5bbe88f3e9a7d92ff1d/src/components/Editable.tsx#L612-L630

I was thinking solve this issue with css `visible: hidden` property but this will cause other issues. For example, if we add  `.distance-from-cursor-3 > .child { visibility: hidden }`then we can not see the thoughts in `.distance-from-cursor-3` We were going to need to add other css properties for visible thughts in `.distance-from-cursor-3`

### DIFF
--- a/src/components/Content.tsx
+++ b/src/components/Content.tsx
@@ -1,4 +1,4 @@
-import React, { FC, MouseEvent, useMemo, useRef } from 'react'
+import React, { FC, MouseEvent, useMemo, useRef, useState } from 'react'
 import { Dispatch } from 'redux'
 import { connect } from 'react-redux'
 import classNames from 'classnames'
@@ -100,13 +100,15 @@ const isLeftSpaceClick = (e: MouseEvent, content?: HTMLElement) => {
 const Content: ContentComponent = props => {
   const { search, isTutorialLocal, tutorialStep, showModal, showRemindMeLaterModal, cursorBack: moveCursorBack, toggleSidebar, rootThoughtsLength, noteFocus, rootSort, isAbsoluteContext } = props
   const contentRef = useRef<HTMLDivElement>(null)
+  const [isPressed, setIsPressed] = useState<boolean | null>(null)
 
   /** Removes the cursor if the click goes all the way through to the content. Extends cursorBack with logic for closing modals. */
-  const clickOnEmptySpace = (e: any) => {
-    // Stop event bubbling. We need to handle this event if we click really on content element not other elements.
-    if (e.target.id !== 'content') {
+  const clickOnEmptySpace = () => {
+    // We need to make sure the user starts the action by pressing the Content element.
+    if (!isPressed) {
       return
     }
+    setIsPressed(false)
 
     // click event occured during text selection has focus node of type text unlike normal event which has node of type element
     // prevent text selection from calling cursorBack incorrectly
@@ -140,8 +142,8 @@ const Content: ContentComponent = props => {
       id='content'
       ref={contentRef}
       className={contentClassNames}
-      onTouchEnd={clickOnEmptySpace}
-      onMouseDown={clickOnEmptySpace}
+      onClick={clickOnEmptySpace}
+      onMouseDown={() => setIsPressed(true)}
     >
       {search != null
         ? <Search />
--- a/src/components/Content.tsx
+++ b/src/components/Content.tsx
@@ -102,7 +102,12 @@ const Content: ContentComponent = props => {
   const contentRef = useRef<HTMLDivElement>(null)
 
   /** Removes the cursor if the click goes all the way through to the content. Extends cursorBack with logic for closing modals. */
-  const clickOnEmptySpace = () => {
+  const clickOnEmptySpace = (e: any) => {
+    // Stop event bubbling. We need to handle this event if we click really on content element not other elements.
+    if (e.target.id !== 'content') {
+      return
+    }
+
     // click event occured during text selection has focus node of type text unlike normal event which has node of type element
     // prevent text selection from calling cursorBack incorrectly
     const selection = window.getSelection()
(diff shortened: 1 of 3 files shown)


raineorshine (changes requested):
I appreciate the succinct description of the cause of the problem, existing infrastructure, and choice of solution. I tested and it fixes the issue, but it unfortunately breaks a test case when tapping on an uncle.

## Steps to Reproduce

```
- a
  - b
    - c
  - d
``` 

(`d` is indented one level compared to the original test case in #1029.)

1. Set the cursor on `c`
2. Tap one line height below `c`, just to the right of where `d` would be. 

## Current Behavior

Cursor moves to `a`. 

## Expected Behavior

Cursor and caret should move to `d`.

onurpolattimur:
Hi @raineorshine, I reproduced your issue and I saw that onClick event of the Content div triggered.  First, I suspected event bubbling, but it was not. In onTap of Editable we have already prevented the default events. 

The problem was that if we click the thought '**d**', the '**b**' thought will be collapsed and the place where we touch up is on `Content` element and this will trigger the onClick event of `Content`.

src/components/Content.tsx
```diff
@@ -102,7 +102,12 @@ const Content: ContentComponent = props => {
   const contentRef = useRef<HTMLDivElement>(null)
 
   /** Removes the cursor if the click goes all the way through to the content. Extends cursorBack with logic for closing modals. */
-  const clickOnEmptySpace = () => {
+  const clickOnEmptySpace = (e: any) => {
+    // Stop event bubbling. We need to handle this event if we click really on content element not other elements.
+    if (e.target.id !== 'content') {
```
onurpolattimur, line 107: I couldn't get id from `React.MouseEvent` so I used `any` for here.

raineorshine (changes requested):
Fix one regression and two more pop up 😂! Tested on iOS.

f54684140 results in incorrect behavior when clicking on empty space.

Can you see if this is reproducible on desktop with `isTouch = true`? That will tell us if we can write an e2e case for this as well.

## Steps to Reproduce

```
- a
  - b
    - c
  - d
``` 

1. Set the cursor on `c`
2. Tap on empty space below `d`

## Current Behavior

If edit mode is enabled, edit mode is disabled and the cursor doesn't move.
If edit mode is disabled, the cursor moves to `a`. 

## Expected Behavior

Cursor and caret should move to `d`.

raineorshine:
> Hi @raineorshine, I reproduced your issue and I saw that onClick event of the Content div triggered. First, I suspected event bubbling, but it was not. In onTap of Editable we have already prevented the default events.

I apologize if I am misunderstanding, but `preventDefault` doesn't have any effect on event bubbling. `e.preventDefault()` prevents the default browser behavior (such as navigating to the `href` of an `<a>`) and `e.stopPropagation()` prevents the same event from firing on ancestors (event bubbling). Here's a Code Sandbox that demonstrates this more clearly: https://codesandbox.io/s/divine-morning-0cw49?file=/src/App.js.

> The problem was that if we click the thought '**d**', the '**b**' thought will be collapsed and the place where we touch up is on `Content` element and this will trigger the onClick event of `Content`.

I'm not sure I understand yet. Can you explain this more? What happens before the `Content` onClick and why does it get that far?

onurpolattimur:
Yes, but e.stopPropogation() doesn't work for touch events. But we can prevent event bubbling using e.preventDefault() for touch events. Here is an example for you, you can test in https://qf8e8.csb.app/ with forcing mobile view by developer tools.
Source codes: https://codesandbox.io/s/withered-water-qf8e8?file=/src/App.js

But our problem was not about event bubbling. I will try to explain more clearly.
First of all, onClick events are triggered after we release the mouse button or release the touch. 

```
- a
  - b
    - c
  - d
```
Think about the case above, when the cursor is on `c` and we touch the thought 'd', context 'b' will be collapsed, and We won't have released the touch yet, the current touch position will be on Content div, and after we release the touch, the onClick event of Content div will be triggered.

I will also look into the case you wrote and will try to make some explanation.

raineorshine:
> Yes, but e.stopPropogation() doesn't work for touch events. But we can prevent event bubbling using e.preventDefault() for touch events. Here is an example for you, you can test in [qf8e8.csb.app](https://qf8e8.csb.app/) with forcing mobile view by developer tools.
> Source codes: [codesandbox.io/s/withered-water-qf8e8?file=/src/App.js](https://codesandbox.io/s/withered-water-qf8e8?file=/src/App.js)

Okay, thanks! I didn't know the touch and click events were competing like that.

> But our problem was not about event bubbling. I will try to explain more clearly.
> First of all, onClick events are triggered after we release the mouse button or release the touch.
> 
> ```
> - a
>   - b
>     - c
>   - d
> ```
> 
> Think about the case above, when the cursor is on `c` and we touch the thought 'd', context 'b' will be collapsed, and We won't have released the touch yet, the current touch position will be on Content div, and after we release the touch, the onClick event of Content div will be triggered.

Got it, that makes sense now!

onurpolattimur:
I made an in-depth review. I think the only option is that using the onMouseDown event instead of onClick. I explained the reason why we should not use onClick above. 

But when we use onMouseDown, another issue will occur. Because click/touch operation is not completed yet selection will be on Editable, so it will not pass the condition below.

So I added a guard, to make sure onClick is triggered only if the user starts the action by pressing the Content element.

https://github.com/cybersemics/em/blob/a08761ac79377927e39630451df47644a6104f96/src/components/Content.tsx#L108-L110

PR #1130, opened 2021-05-08
Add theme before react loads

### LINKED ISSUE #269: White flash on initial load
The toolbar and hamburger menu need to be hidden when the welcome modal is active. Otherwise they produce a white flash on load and appear during momentum scroll. 

Reproduce by refreshing the page; you should see the hamburger menu and toolbar flash on the screen.

shresthabijay:
fixes #269 

# Cause of issue

Theme was only set up after the first load of the `AppComponent`. It momentarily causes `body` to have no theme class. This was causing initial white flash.

# Solution

Adding required style and js logic to set initial theme before react loads solves the problem.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -28,6 +28,14 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>em</title>
+
+    <style>
+      body.dark {
+        color: white;
+        background-color: black;
+      }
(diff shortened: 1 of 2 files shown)


shresthabijay:
@raineorshine Please review the changes!

raineorshine:
Thanks!

Issue #230, opened 2019-12-24
contextIndex: Array → Object 


raineorshine:
Convert `Parent`‘s `children` property to an object to allow concurrent writes.

PR #1133, opened 2021-05-11
adds session manager(to identify tabs) and dexie subscriptions 

anmolarora1:


anmolarora1:
@raineorshine I'm facing some issues with mocking `dexie-observable` and noticed that you'd created a [similar issue](https://github.com/cybersemics/em/issues/664#issuecomment-629691193) sometime back.  I went through the issue description and PR #778  but there have been a lot of changes since then and I couldn't find the solution that you added. 
Can you please share that?

raineorshine:
> > > The solution involves pulling thoughts for selected actions instead of every action. I'm hoping that this would be completed soon.

> > The solution needs to be generalized so that it works for any actions. This was an intentional design choice so that the system is extensible as new actions are added to the app. Feel free to blacklist certain actions, but don’t use a whitelist approach that requires new actions to be manually added.

> Here's what I think -
> Now that we have subscription-based updates to sync visible thoughts in real-time, we should consider only those actions that impact the non-visible thoughts. 
> While blacklisting is an easier option, I think an effective solution is to whitelist actions that would need a pull because that saves us from the overhead of calculating visible and pending contexts for every action - something that we're currently doing.
> Do you think that grouping actions into different categories would be helpful? Something of this sort -
> 1. UX critical actions - The actions that impact visible thoughts (editThought, moveThought/Up/Down, indent, outdent, extractThought, splitThought, etc). With real-time updates in place, we shouldn't need a manual pull in response to any of these actions.
> 2. Nav actions - actions like setCursor, cursorUp, cursorDown, cursorBack, to name a few. We need to trigger a pull for all these actions since they change visible thoughts/contexts that might've been updated in the background.
> 3. Meta actions - such as tutorial, search, alert, showModal, toggleTopControlsAndBreadcrumbs, error, dragInProgress, etc. These do not directly impact the thoughts so we probably don't need to worry about them.
> We can add more categories, or add further details to each one. This would likely help us figure out an appropriate solution for this issue and also be useful for subsequent issues like undo/redo. 
> I might be missing something here so please feel free to add.

In order to determine if the visible thoughts have changed, we have `getVisibleContexts` in the `pullQueue` which uses `expandThoughts`. `expandThoughts` is decoupled from any specific actions. There is no way to strictly know which actions will change visible thoughts or not, since the expansion logic has a variety of conditions, including meta attributes (`=pin`, `=view/Table`) and content (thoughts ending in `:`). It is better to check which thoughts are visible regardless of which actions have been called.

For actions that do not affect the cursor or visible thoughts at all, then `expandedContexts` will not change, and `updatePullQueue` will short circuit, so this path should have negligible overhead. If certain actions are a performance bottleneck, we can blacklist them from the `pullQueue` without losing the generality of the solution.

src/data-providers/firebase.ts
```diff
@@ -94,4 +112,53 @@ const getFirebaseProvider = (state: State, dispatch: Dispatch<any>) => ({
   },
 })
 
+/** Creates a subscription handler that converts a Parent snapshot to a ThoughtUpdate and invokes a callback. */
+const createParentSubscriptionHandler =
+  (onUpdate: (updates: ThoughtUpdates) => void, { value }: { value?: Parent | null } = {}) =>
```
shresthabijay, line 117: We can replace `value`  with something like `delete` which would be a boolean. If `delete` is true, then simply use `null` as value. I think it would be more clear to understand.

src/action-creators/updateThoughtsFromSubscription.ts
```diff
@@ -0,0 +1,46 @@
...
+  updatedBy?: string
+}
+
+/** Returns true if the updateable entity is coming from the given SessionType and is not self. */
+export const isValidSource = (updateable: Updateable, updateType: SessionType) => {
+  // on a change event, the firebase snapshot only contains updated fields, which may not include updatedBy
+  // in this case, return true so it gets added locally
+  return (
+    !updateable.updatedBy ||
+    (updateable.updatedBy !== getSessionId() && getSessionType(updateable.updatedBy) === updateType)
```
shresthabijay, line 16: Currently, a tab session id is added to the local session list only after visibility has changed. In the case where we open two tabs with a window split or in a dual-screen setup, the visibility of the two tabs doesn't change. So updates from local DB are rejected because they are mistaken for remote updates. Maybe we should call `keepAlive` once at the initialization.

shresthabijay (changes requested):

src/data-providers/firebase.ts
raineorshine: I tried it both ways, and I find `{ value: null }` to read better than an unmarked Boolean. I'll add a comment though. The Boolean would require a comment anyway to understand.

src/action-creators/updateThoughtsFromSubscription.ts
raineorshine: Good idea. I may use a dedicated init function, which seems like a better place to initialize the in-memory `sessionId` too.

raineorshine: The dedicated `init` won't work without some larger refactoring, because `getSessionId` is used in `initialState`, which gets called before `initialize`. I will just call `keepalive` in `initialize` and otherwise keep the structure as-is.

raineorshine: Should we call `keepalive` also on edit to prevent the session from expiring if the tab stays visible?

raineorshine: I added a simple middleware that calls `keepalive` on any action.

raineorshine:
Error editing with multiple tabs.

## Steps to Reproduce

```
- a
- b
```

Open em in two tabs with the above thoughts, and execute the following in quick succession:

1. Create a new thought after `b`
2. Edit the empty thought to `c`
3. Hit backspace to edit `c` so it is empty again.
4. Hit backspace again to delete the empty thought.

## Current Behavior

`Cannot read property 'replace' of undefined

image

raineorshine:
@shresthabijay Here's what's left:

**Remove/refactor pulls:** With subscriptions, we can now assume that thoughts in the Redux state are up-to-date with the remote. This means that we can skip some pulls such as pull-before-push and pull on navigate. EXCEPT we still need to pull pending descendants.

My rough plan is to make any `pull` that is initiated on a non-pending thought instead traverse the descendants and only pull pending thoughts. This seems more robust than putting the logic outside `pull` which would add complexity to some of the call sites. 

We may also want to filter out non-pending thoughts in `pullQueue` to prevent them from being passed to `pull` at all when the `pullQueue` is flushed.

We need to analyze each call to `pull` to make sure this solution will work and not have any unexpected consequences. Luckily there are only four files that call `pull`:

```ts
/Users/raine/projects/em/src/shortcuts/copyCursor.ts:
    4:  import { alert, pull } from '../action-creators'
    42:        await dispatch(pull({ [hashContext(context)]: context }, { maxDepth: Infinity }))

/Users/raine/projects/em/src/redux-middleware/pushQueue.ts:
    9:    pull,
    70:        await dispatch(pull(pending, { maxDepth: Infinity }))
    95:        await dispatch(pull(pending, { maxDepth: Infinity }))
    121:        await dispatch(pull(pending, { maxDepth: Infinity }))
    134:        await dispatch(pull(pathToLoad, { maxDepth }))

/Users/raine/projects/em/src/redux-middleware/pullQueue.ts:
    13:  import { pull } from '../action-creators'
    119:      const hasMorePending = await dispatch(pull(extendedPullQueue))

/Users/raine/projects/em/src/components/ModalExport.tsx:
    23:  import { alert, error, closeModal, pull } from '../action-creators'
    68:        pull(
```

Lastly, we'll need some rigorous manual testing to see if there are any concurrency issues.

---

- [ ] Error quickly editing multiple tabs (above)
- [ ] Remove/refactor pulls
- [ ] Test

shresthabijay:
> we'll need some rigorous manual testing to see if there are any concurrency issues.

@raineorshine Thanks for the update! I will continue and share the updates here.

raineorshine:
- [x] ~Error quickly editing multiple tabs (above)~
  - Moved to #1406; not blocking
- [x] Pull only pending thoughts
- [x] Test multiple tabs (Dexie)
- [ ] Test multiple devices (Firebase)

raineorshine:
All done and ready to merge. I have tested a variety of scenarios for online and offline editing across 2 devices when logged in. The below table indicates the different combinations tested. Both devices began synced, then edits were made on one or both devices online or offline.

e.g.

- (3) Online Edit — Offline: Device B was taken offline, then Device A made an edit, then Device B was reconnected.
- (4) Online — Offline Edit: Device B was taken offline, then Device B  made an edit, then Device B was reconnected.
- (7) Offline Edit — Offline Edit: both devices were taken offline and edited independently before reconnecting.

All scenarios resulted in correctly synced data once each device was brought back online and time was given for them to sync with Firebase.

image

One exception is that editing the same `Parent` on both devices causes `Parent.children` to be overwritten. Converting `Parent.children` to an object (#230) so that individual `Child` entries can be updated without updating the whole `Parent` will solve simultaneous child editing. Independent Editing (#495) will solve simultaneous editing of ancestor and descendant .

Given the complexity of this task and the large amount of concurrency involved, I imagine that there will be more bugs and timing issues. At this point, it works well for the simple test cases given. I think it's ready to merge and continue monitoring during real-world usage.

@shresthabijay I'd love to get a final review and sanity check from you! Thanks!

shresthabijay:
> All done and ready to merge. I have tested a variety of scenarios for online and offline editing across 2 devices when logged in. The below table indicates the different combinations tested. Both devices began synced, then edits were made on one or both devices online or offline.
> 
> e.g.
> 
> * (3) Online Edit — Offline: Device B was taken offline, then Device A made an edit, then Device B was reconnected.
> * (4) Online — Offline Edit: Device B was taken offline, then Device B  made an edit, then Device B was reconnected.
> * (7) Offline Edit — Offline Edit: both devices were taken offline and edited independently before reconnecting.
> 
> All scenarios resulted in correctly synced data once each device was brought back online and time was given for them to sync with Firebase.
> 
> image
> 
> One exception is that editing the same `Parent` on both devices causes `Parent.children` to be overwritten. Converting `Parent.children` to an object (#230) so that individual `Child` entries can be updated without updating the whole `Parent` will solve simultaneous child editing. Independent Editing (#495) will solve simultaneous editing of ancestor and descendant .
> 
> Given the complexity of this task and the large amount of concurrency involved, I imagine that there will be more bugs and timing issues. At this point, it works well for the simple test cases given. I think it's ready to merge and continue monitoring during real-world usage.
> 
> @shresthabijay I'd love to get a final review and sanity check from you! This is also a good time to point out anywhere you think I can provide better comments/documentation. Thanks!

Sure, I will do the review.

shresthabijay (commented):
@raineorshine Overall approach looks good. I have left few comments for things I wanted to understand.

src/action-creators/updateThoughtsFromSubscription.ts
```diff
@@ -0,0 +1,75 @@
...
+        : null,
+    )
+
+    if (Object.keys(contextIndexUpdates).length === 0 && Object.keys(thoughtIndexUpdates).length === 0) return
+
+    dispatch(
+      updateThoughts({
+        contextIndexUpdates,
+        thoughtIndexUpdates,
+        local: false,
```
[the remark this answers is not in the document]
raineorshine: You're right! I wasn't noticing this because the remote thoughts are force pulled on refresh.

Fixed in d053e61bcb.

src/action-creators/pull.ts
```diff
@@ -25,22 +28,57 @@ async function itForEach<T>(it: AsyncIterable<T>, callback: (value: T) => void)
...
+    { force, maxDepth, onLocalThoughts, onRemoteThoughts }: PullOptions = {},
   ): Thunk<Promise<boolean>> =>
   async (dispatch, getState) => {
     if (Object.keys(contextMap).length === 0) return false
 
+    let contextMapFiltered = contextMap
+
+    // pull only pending contexts parents unless forced
+    if (!force) {
+      // Missing parents only occur in 2-part deletes (see flushDeletes and associated logic).
```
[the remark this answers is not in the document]
raineorshine: Sure! As you know, in a 2-part delete, all of the descendants that are in the Redux store are deleted in Part I, and pending descendants are pulled and then deleted in Part II. 

With the old style `pull`, Part II pulled the pending descendants as expected. With the new style `pull` that *only* pulls pending thoughts, `pull` will short circuit in Part II since there is no `Parent` which is marked as pending (it was deleted in Part I). Thus we need to check both `pending` and check that the `Parent` is there at all. The only way it could be missing is if it was removed in Part I of a 2-part delete.

Does that make sense?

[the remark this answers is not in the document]
shresthabijay: Yes it makes sense.

shresthabijay (approved):

PR #1138, opened 2021-05-13
Normalize thought for duplication check in editable

### LINKED ISSUE #1132: Disallow siblings with same lexeme
## Steps to Reproduce

1. Create thought `cat`
2. Create thought `cats`

## Current Behavior

- cat
- cats

## Expected Behavior

“Duplicate thoughts” alert should occur for siblings that would have the same `Lexeme` i.e. hash to the same value. They should be treated just like exact duplicates (e.g. “cat”) in style, behavior, alert, etc.

shresthabijay:
fixes #1132

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -45,6 +45,7 @@ import {
   pathToContext,
   setSelection,
   strip,
+  normalizeThought,
 } from '../util'
 
 // selectors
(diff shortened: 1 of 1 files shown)


src/components/Editable.tsx
```diff
@@ -439,7 +440,7 @@ const Editable = ({ disabled, isEditing, simplePath, path, cursorOffset, showCon
     const oldValueClean = oldValue === EM_TOKEN ? 'em' : ellipsize(oldValue)
 
     const thoughtsInContext = getAllChildren(state, context)
-    const hasDuplicate = newValue !== '' && thoughtsInContext.some(thought => thought.value === newValue)
+    const hasDuplicate = newValue !== '' && thoughtsInContext.some(thought => normalizeThought(thought.value) === normalizeThought(newValue))
```
raineorshine, line 443: You’ll want to normalize `newValue` outside the loop for efficiency.

raineorshine (changes requested):

src/components/Editable.tsx
shresthabijay: Sure

raineorshine (approved):

raineorshine:
Thanks!

PR #1141, opened 2021-05-13
Scale `Content` using fontSize and em unit

onurpolattimur:
fixes: #536

### DIFF
--- a/src/test-helpers/e2e-helpers/clickWithOffset.ts
+++ b/src/test-helpers/e2e-helpers/clickWithOffset.ts
@@ -36,7 +36,7 @@ const clickEditable = async (page: Page, nodeHandle: JSHandle, { horizontalClick
   const coordinate = !offset ? {
     x: boundingBox.x + (
       horizontalClickLine === 'left' ? 0
-      : horizontalClickLine === 'right' ? boundingBox.width
+      : horizontalClickLine === 'right' ? boundingBox.width - 1
       : boundingBox.width / 2
     ),
     y: boundingBox.y + (boundingBox.height / 2)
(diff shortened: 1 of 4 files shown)


src/test-helpers/e2e-helpers/clickWithOffset.ts
```diff
@@ -44,7 +44,7 @@ const clickEditable = async (page: Page, nodeHandle: JSHandle, { horizontalClick
 
   if (!coordinate) throw new Error('Coordinate not found.')
 
-  await page.mouse.click(coordinate.x + x, coordinate.y + y)
+  await page.mouse.click(coordinate.x + x - (horizontalClickLine === 'right' ? 1 : 0), coordinate.y + y)
```
raineorshine, line 47: I suggest changing `boundingBox.width` to `boundingBox.width - 1` on L39 instead of here so that the `horizontalClickLine` logic is all in one place.

raineorshine (changes requested):
Looks good! Scaling the font size seems to work much better than transforming the whole `Content` component.

There are a few non-text elements that need to be scaled explicitly now that they are not transformed by the ancestor `Content` component. Nothing you could have known ahead of time. I will post screenshots below. Thanks!

raineorshine:
The url icon needs to be scaled.

## Steps to Reproduce

``` 
- https://emtheapp.com
``` 

Increase font size +4

## Current Behavior

image

## Expected Behavior

As in `dev`:

image

raineorshine:
Context breadcrumbs need to be scaled.

## Steps to Reproduce

``` 
- a
  - b
    - c
      - x
    - d
      - x
``` 

- Increase font size +4
- Activate Context View on `a/b/c/x`

## Current Behavior

image

## Expected Behavior

As in `dev`:

image

raineorshine (approved):

raineorshine:
Looks great! Thanks!

Issue #1140, opened 2021-05-13
Export: Advanced options


raineorshine:
Add an "Advanced" toggle to the Export screen that displays more advanced export options. The Export screen can be activated with the last icon on the right of the toolbar.

- Move "Copy to clipboard" above the download message.
- Add an "Advanced" link below the download message.
- Clicking the Advanced link should open up a div below it where advanced options can go.

![image](https://user-images.githubusercontent.com/750276/118134002-b1053a00-b3be-11eb-9416-3344114c376f.png)

raineorshine:
## Part II

- Add `ArrowDownWhite` (or `ArrowDownWhite` if `dark` is true) to the right of the "Advanced" link to show the user that it expands content.
- When expanded, the arrow should point up.
- Currently there are two plaintext options in the dropdown menu: "Plain Text (lossless)" and "Plain Text". Actually, lossless should be an advanced option that applies to any content type.
  - Remove "Plain Text (lossless)" from the dropdown menu
  - Add a "Lossless" checkbox to the Advanced section as shown below. Default to checked.
  - Rewire the "Plain Text (lossless)" functionality to use the "Lossless" checkbox.

<img width="1068" alt="Screen Shot 2021-05-14 at 3 12 51 PM" src="https://user-images.githubusercontent.com/750276/118319441-03765180-b4b8-11eb-8e92-b4d1c4b3f48e.png">

ruchijn:
Based on current functionality, user can either select `Plain Text(Lossless)`or `Plain Text`. Moving forward when `Plain Text (Lossless)` is moved to advanced with auto-selected, should I go ahead and make sure nothing is auto-checked in the existing dropdown?

raineorshine:
The new options in the dropdown menu will be `Plain Text` and `HTML`, with `Plain Text` selected by default.

ruchijn:
My main question is, can we have user download as `Plain Text` and `Lossless` both selected at same time? If so, then will they download 2 files?

raineorshine:
`Plain Text` is the file format and `Lossless` corresponds to the `excludeMeta` option that is passed to `getDescendants` and `exportContext`, e.g.:

https://github.com/cybersemics/em/blob/7ff12d46358d2942fafb8a5604d06994aaa5a8b0/src/components/ModalExport.tsx#L53-L55

[...]

PR #1142, opened 2021-05-14
Adding advanced setting toggle

ruchijn:
For issue: https://github.com/cybersemics/em/issues/1140

em_advanced

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -3135,10 +3135,23 @@ code, .code {
   margin-bottom: 20px;
 }
 
+p.advance-setting-label {
+  margin-bottom: 5px;
+  line-height: 1.2;
+}
+
+p.advance-setting-description {
+  font-size: 90%;
+}
+
 .advance-setting-link {
   display: flex;
   position: relative;
+  transition: opacity 100ms ease-in-out;
 }
+
+a.advance-setting-link { color: #000; }
+.dark a.advance-setting-link { color: #fff; }
 
 .advance-setting-chevron {
   display: flex;
@@ -3150,8 +3163,8 @@ code, .code {
 .advance-setting-section {
   display: flex;
   justify-content: center;
-  margin: 0 auto 20px;
-  width: 90%;
+  margin: 0 auto;
+  max-width: 34em;
 }
 
 .chevron-up {
@@ -3165,9 +3178,6 @@ code, .code {
   padding-left: 35px;
   margin-bottom: 12px;
   cursor: pointer;
-  -webkit-user-select: none;
-  -moz-user-select: none;
-  -ms-user-select: none;
   user-select: none;
 }
 
--- a/src/App.css
+++ b/src/App.css
@@ -3127,3 +3127,12 @@ code, .code {
   transform-origin: 0 0; /* see note above */
 }
 .ripple_loader div { box-sizing: content-box; }
+
+.advance-setting-wrapper {
+  margin-bottom: 15px;
+  text-align: center;
+}
+
+.advanced-setting-section {
+  margin: 15px auto;
+}
--- a/src/App.css
+++ b/src/App.css
@@ -2847,6 +2847,7 @@ code, .code {
   position: absolute;
   right: 3px;
   padding: 0px 5px;
+  z-index: 1;
 }
 
 .drop-down-wrapper .drop-down-option-wrapper {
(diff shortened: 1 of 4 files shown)


raineorshine:
Perfect, thank you!

I'm going to add some requirements to the issue before we merge.

src/components/ModalExport.tsx
```diff
@@ -210,9 +210,21 @@ const ModalExport = () => {
     setPublishing(false)
   }
 
+  const [advancedSettings, setadvancedSettings] = useState(false)
```
raineorshine, line 213: Prefer full camelCase: `setAdvancedSettings`

ruchijn: Changed

src/components/ModalExport.tsx
```diff
@@ -210,9 +209,46 @@ const ModalExport = () => {
...
+
+  const [shouldIncludeMetaAttributes, setShouldIncludeMetaAttributes] = useState(true)
+
+  /** Updates checkbox value when clicked and set the appropriate value in the selected option. */
+  const onChangeLosslessCheckbox = () => {
+    const updatedValue = {
+      ...selected,
+      excludeMeta: !shouldIncludeMetaAttributes
+    }
+    setSelected(updatedValue)
```
raineorshine, line 225: The `ExportOption` type needs to be refactored since we are handling `excludeMeta` as an independent variable now. `ExportOption` should no longer contain `excludeMeta`. We're halfway in between right now which is resulting in some extraneous structure for what should be a single, local component state (`shouldIncludeMetaAttributes`).

The fact that you have to call `setSelected` and `setShouldIncludeMetaAttributes` to update a single value is a sign that they should be combined.

src/components/ModalExport.tsx
```diff
@@ -210,9 +209,46 @@ const ModalExport = () => {
...
+  /** Created an array of objects so that we can just add object here to get multiple checkbox options created. */
+  const checkboxArray = [
+    {
+      id: 'lossless-checkbox',
+      onChangeFunc: onChangeLosslessCheckbox,
+      seq: 1,
+      title: 'Lossless',
+      description: 'When checked, include all metaprogramming attributes such as archived thoughts, pins, table view, etc. Check this option for a backup-quality export that can be re-imported with no data loss. Uncheck this option for social sharing or exporting to platforms that do not support em metaprogramming attributes. Which is, uh, all of them.'
+    }
+  ]
```
raineorshine, line 238: Great, I like this structure and it is nicely extensible!

src/App.css
```diff
@@ -3091,3 +3092,105 @@ code, .code {
...
+}
+
+/* Custom checkbox style starts */
+.checkbox-container {
+  display: block;
+  position: relative;
+  padding-left: 35px;
+  margin-bottom: 12px;
+  cursor: pointer;
+  user-select: none;
```
raineorshine, line 3145: The build process will automatically add "-moz-", "-webkit-" prefixes based on our `browserslist`.

src/App.css
```diff
@@ -3091,3 +3092,105 @@ code, .code {
...
+}
+
+.advance-setting-section {
+  display: flex;
+  justify-content: center;
+  margin: 0 auto;
+  max-width: 34em;
+}
+
+.chevron-up {
```
raineorshine, line 3134: It would be better to give this a more general name like `.rotate180` since it is not limited to just chevrons. Or add it inline since we're moving towards inline styles eventually.

src/components/ModalExport.tsx
```diff
@@ -210,9 +209,38 @@ const ModalExport = () => {
...
+    setSelected(updatedValue)
+    setShouldIncludeMetaAttributes(!shouldIncludeMetaAttributes)
+  }
+
+  /** Created an array of objects so that we can just add object here to get multiple checkbox options created. */
+  const checkboxArray = [
+    {
+      id: 'lossless-checkbox',
+      onChangeFunc: onChangeLosslessCheckbox,
+      seq: 1,
```
raineorshine, line 234: Let's call this `defaultChecked`. I'm not sure what `seq` means.

raineorshine (changes requested):
Looks great! Just a few request changes below.

src/components/ModalExport.tsx
ruchijn: Done

src/App.css
ruchijn: Done

src/App.css
ruchijn: Removed

src/components/ModalExport.tsx
ruchijn: Changed. Also I created `ChevronImg` as it was mostly same code duplicated.

src/components/ModalExport.tsx
```diff
@@ -210,25 +217,39 @@ const ModalExport = () => {
...
-            style={{ cursor: 'pointer' }}
-            onClick={() => setIsOpen(!isOpen)}
-          />
+          <ChevronImg dark={dark} onClickHandle={() => setIsOpen(!isOpen)} />
           <div ref={setWrapper}>
             <DropDownMenu
               isOpen={isOpen}
               selected={selected}
               onSelect={(option: ExportOption) => {
+                setShouldIncludeMetaAttributes(!shouldIncludeMetaAttributes)
```
raineorshine, line 252: I'm not sure I understand this. Selecting the export format should not affect `shouldIncludeMetaAttributes`. Can you explain what your intention was?

src/components/ModalExport.tsx
```diff
@@ -11,13 +9,21 @@ import { exportContext, getDescendants, getAllChildren, simplifyPath, theme } fr
...
+  id: string,
+  onChangeFunc: () => void,
+  defaultChecked: boolean,
+  title: string,
+  description: string,
+}
+
 const exportOptions: ExportOption[] = [
-  { type: 'text/plain', label: 'Plain Text (lossless)', extension: 'txt' },
-  { type: 'text/plain', label: 'Plain Text', extension: 'txt', excludeMeta: true },
```
raineorshine, line 20: You can now remove `excludeMeta` from the `ExportOption` interface itself since it is no longer being used:

https://github.com/cybersemics/em/blob/84ec54dec3133bfa2e8e2d0827efc4b2ad40ab5f/src/types.ts#L93-L99

raineorshine (changes requested):

src/components/ModalExport.tsx
ruchijn: Oh.. I left it by mistake. I was doing some testing. It is not needed here. Sorry about that

src/components/ModalExport.tsx
ruchijn: Done

raineorshine (approved):

raineorshine:
Thanks!

PR #1143, opened 2021-05-14
Conistent rank on duplicate merge

### LINKED ISSUE #1136: existingThoughtMove: Duplicate context added to lexeme
## Steps to Reproduce

```
- a
  - b
- b
```

1. Outdent `a/b` *or* move `a/b` above `b`.
2. Indent `b` *or* move `b` into `a`.

## Current Behavior

`b` has superscript "2" even though it only exists in one context.

## Expected Behavior

`b` should not have a superscript.

## Notes

Note that this does not occur if moving `a/b` *below* `b`. 

It may be related to the updated rank logic I added in 276690282a516d2e3af129b112bd31b3b2241983. Or it could be related to recent work done on the merge logic in `existingThoughtMove`.

shresthabijay:
fix #1136 

# Cause of issue

```
- a
  - b
- b
```

On moving `b` to `a.b` there was inconsistent rank saved in `contextIndex` and `thoughtIndex`. The `newRank` generated is `0.5` and the rank of the duplicate `b` is `1`.  `newRank` is used by `movedThought` . But to update the child it uses previous rank because it checks for the duplicate merge. This causes inconsistent rank between the two indexes. Now moving `b` to `a.b`, `moveThoughts` cannot remove the old context as it checks for the rank.

# Changes

- Use previous duplicate rank for both `contextIndex` and `thoughtIndex` updates if there is a duplicate merge. 
- Add tests for this specific case.

### DIFF
--- a/src/reducers/__tests__/existingThoughtMove.ts
+++ b/src/reducers/__tests__/existingThoughtMove.ts
@@ -1,8 +1,10 @@
 import { HOME_PATH, HOME_TOKEN } from '../../constants'
 import { equalArrays, initialState, reducerFlow } from '../../util'
-import { exportContext, getContexts, getThought, getAllChildren, getChildrenRanked } from '../../selectors'
+import { exportContext, getContexts, getThought, getAllChildren, getChildrenRanked, getRankAfter, getChildren } from '../../selectors'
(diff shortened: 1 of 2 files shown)


shresthabijay:

raineorshine:
Thank you!!!

raineorshine (approved):

PR #1145, opened 2021-05-14
Gesture diagram overlay for mobile

### LINKED ISSUE #1109: Gesture diagram overlay
Draw an on-screen gesture diagram overlay for mobile screen captures. 

- Render the gesture diagram immediately when a shortcut is entered. 
- Fade out. Try delay 300ms and duration 500ms. 
- Render with name of shortcut. 
- Location
  - In alert? Has text anyway. 
  - Right top?
  - Center top?
  - Can’t go bottom since the keyboard pops up with some shortcuts.

shresthabijay:
fixes #1109 

# Changes
- Add gesture diagram overlay at the bottom of the screen. Always appears relative to the actual viewport ( above keyboard).
- Shortcuts diagram can show up to 3 gestures if the consecutive gestures are activated before the gesture diagram duration.
- Add development setting in help modal to enable showing gesture diagrams on mobile. (I believe this will be used only for screen captures so)

# Steps to enable gesture diagram
- Go to help modal.
- Click enable gesture diagrams.


shresthabijay:
# Demo

https://user-images.githubusercontent.com/25879100/118298922-d77ebe80-b4ff-11eb-8355-3df323cc142c.mp4

shresthabijay:

raineorshine:
Wow, exciting! I will review soon.

raineorshine:
- Positioning is great
- Fade and duration is great
- Diagram needs to be centered in the gray container

> Always appears relative to the actual viewport ( above keyboard).

I’m not seeing diagrams while the keyboard is up on Mobile Safari. This is the platform from which I’ll be doing all my screen captures. 

https://user-images.githubusercontent.com/750276/118527109-f30be400-b70e-11eb-8e72-b7dd7a46e675.mov

src/components/LatestShortcutsDiagram.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <CSSTransition in={latestShortcuts.length > 0} classNames={'latest-shortcuts-transition'} timeout={400} unmountOnExit={true}>
+        <div className={classNames({
+          'latest-shortcuts': true,
+          [position]: true
+        })}>
+          {shortcutsList.map((shortcut, index) => {
+            return (
+              <div key={index} className='shortcut-tab-wrapper'>
+                <div className='shortcut-tab'>
+                  <GestureDiagram path={shortcut.gesture as GesturePath} size={30} color='white' strokeWidth={2}/>
```
raineorshine, line 40: Color should use the `theme` selector and switch between `black` and `white`.

raineorshine (changes requested):
Good structure, thanks!

src/components/LatestShortcutsDiagram.tsx
shresthabijay: Sure.

shresthabijay:
> * Diagram needs to be centered in the gray container

It is already centered. It seems like for some `GestureDiagram` view box is not properly calculated.

shresthabijay:
@raineorshine I have made the following changes.
- Latest gestures diagram appears above the keyboard for ios safari.
- `NavBar` also now appears above the keyboard for ios safari.
- Add the theme to `LatestShortcutsDiagram`.

I have used `visualViewport API for keeping things above the keyboard on ios. But this API is supported only from ios safari 13. We may need to add some polyfills to support previous ios safari.

raineorshine:
https://user-images.githubusercontent.com/750276/118653514-9dd6dd80-b7b5-11eb-8dd3-ab13a21d6056.mov

shresthabijay:
> `position: fixed/sticky` is broken on Mobile Safari, so proceed with caution. You may have to use a similar approach as used for the Toolbar (#756). The Toolbar still quite choppy scrolling with the keyboard up. Restructuring the layout and setting overflow scroll on a div rather than the body might be a better solution, but that's a larger effort and it may affect secondary scroll behaviors such as autoscroll, scroll bounce, etc.
> 
> It may be preferable to find a working solution just for the latest shortcut diagrams to avoid getting too sucked into this.
> 
>  FullSizeRender.mov

I would like to have a go at it. The latest shortcuts diagram will get affected by the scroll as shown in the video. Because it's actually inside the scrolling container. We need to place absolutely placed views like toolbar, nav, and gesture diagram outside the scrolling container. We should change the layout. What do you think?

raineorshine:
> I would like to have a go at it. The latest shortcuts diagram will get affected by the scroll as shown in the video. Because it's actually inside the scrolling container. We need to place absolutely placed views like toolbar, nav, and gesture diagram outside the scrolling container. We should change the layout. What do you think?

It's worth a try! It would make the Toolbar look a lot better, too. If we can get the `Content` to scroll while preserving the secondary scrolling behavior that we have now, that would be great.

I'm going to revert the last commit (`da51069`) and merge what we have then you can work off `dev`.

raineorshine (approved):

PR #1146, opened 2021-05-14
Fade Out Alert 

### LINKED ISSUE #1046: Fade out alert
Alert should fade out over ~300ms~ 800ms rather than disappearing instantly when its timer expires.

Use the `CSSTransition` component.

linghong:
Fixes #1046

I didn't find any code has fade time 300 ms, but I noticed that the Alert.tsx had fade out time 200 ms, and disappeared immediately after the timer expired. Actually,  it  turned out the code has no error,  just because the time was too short, the fade out was hardly recognized by human eyes. I thus increased the time to 800 ms.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -13,18 +13,14 @@
   color: lightblue;
 }
 
-html,
-body,
-#root,
-#app {
+html, body, #root, #app {
   height: 100%;
   font-size: 16px;
 }
 
-body,
-textarea {
+body, textarea {
   font-weight: 300;
-  font-family: "Helvetica";
+  font-family: 'Helvetica';
   line-height: 1.25;
 }
 
@@ -97,6 +93,7 @@ input:focus {
   outline: 0 none;
 }
 
+
 .dark input:focus {
   border: solid 1px #999;
 }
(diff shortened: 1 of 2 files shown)


raineorshine (changes requested):
Great! That's good that the CSSTransition was already in place. It feels good for the fade-out, but now the fade-in is too long. The fade-in should still be 200ms. You may have to look up how to have separate fade-out and fade-in durations in the `react-transition-group` documentation.

Please revert the formatting changes that were made to `App.css`. This changes the style and lengthens many expressions that would be better in the current more compact format.

linghong:
Sorry. I didn't notice that the FormateOnSave on my editor was set on true. I have moved the formate back to original.  

I think how it animates is determined by the CSS code, and the JavaScript  code timeout controls the maximum time limit on the animation. Hence, if you want to increase the Alert ease-out  animation time from 800 ms to 2 s, you have to change  the css code fade-enter-active to 2s, and also increase the js code timeout time; while if you want to reduce ease-in animation time, you only need to change the CSS code.

raineorshine:
> Sorry. I didn't notice that the FormateOnSave on my editor was set on true. I have moved the formate back to original.

Great! The diff is much easier to read too :).

> I think how it animates is determined by the CSS code, and the JavaScript code timeout controls the maximum time limit on the animation. Hence, if you want to increase the Alert ease-out animation time from 800 ms to 2 s, you have to change the css code fade-exit-active to 2s, and also increase the js code timeout time; while if you want to reduce ease-in animation time, you only need to change the CSS fade-enter-active time.

Thanks for the explanation. Looks good to me!

raineorshine (approved):

Issue #1147, opened 2021-05-15
Mobile: cursorUp incorrectly activated on gesture


raineorshine:
Hi @onurpolattimur. It looks like we have just a small regression from ~#1135~ #1128. You can reproduce in emulation mode on desktop. Could you look into this when you have a chance? Thanks!

## Steps to Reproduce

``` 
- a
  - x
    - y
- b
- c
- d
- e
- f
- g
- h
- i
``` 

1. Set the cursor on `a/x/y`
2. Activate the swipe →↓  gesture (newThought) in the space below `y`

## Current Behavior

`cursorUp` is activated and then it creates a new thought after `x`.

## Expected Behavior

A new thought should be created after 'y`.

raineorshine:
I have confirmed that this is introduced in 4a59734fd. I didn't catch it during my testing.

PR #1148, opened 2021-05-15
Don't set cursor if any gesture is activated when pressing the hidden thought

onurpolattimur:
fixes: #1147


raineorshine (approved):
Tested all the [caret test cases](https://github.com/cybersemics/em/wiki/Test-Cases#caret) and everything works! 🙏

Issue #1033, opened 2021-02-12
Import: Incorrect thought structure from iOS Notes app


raineorshine:
## Steps to Reproduce

Copy and paste the following from iOS Notes app to **em**:

```
A
  - B
  - C
``` 

## Current Behavior

- [empty]
  - A
- [empty]
  - A
- [empty]
  - A

## Expected Behavior

It should create the same thoughts that the plain text version creates:

- A
  - B
  - C

onurpolattimur:
The current behaviour is different on my iPhone. Only one thought is created whose value is `A-B-C`.

![IMG_1646](https://user-images.githubusercontent.com/23136437/118253499-aaf08580-b4b2-11eb-92de-d956bd3e7793.PNG)
![IMG_1645](https://user-images.githubusercontent.com/23136437/118253508-ad52df80-b4b2-11eb-82dd-102056bb3cb0.PNG)

raineorshine:
Thanks @onurpolattimur. I get this behavior now, too. I updated the original description.

onurpolattimur:
Similar behavior exists when we copy text from different text editors/ide.

Here is the clipboard data when we copy text from macos notes app:

```
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
<meta http-equiv="Content-Style-Type" content="text/css">
<title></title>
<meta name="Generator" content="Cocoa HTML Writer">
<meta name="CocoaVersion" content="1894.6">
<style type="text/css">
p.p1 {margin: 0.0px 0.0px 0.0px 0.0px; font: 12.0px 'Helvetica Neue'}
</style>
</head>
<body>
<p class="p1">A</p>
<p class="p1"><span class="Apple-converted-space">  </span>- B</p>
<p class="p1"><span class="Apple-converted-space">  </span>- C</p>
</body>
</html>
```

PR #1154, opened 2021-05-16
Fix incorrect thought structure when copy from text editors

onurpolattimur:
fixes: #1033

---

This issue is not about only ios notes app, similar behaviour exists when we copy text from other applications such as webstorm, macos notes app etc.

First of all, there was an error in the code that gets the body content of html string. 
https://github.com/cybersemics/em/blob/cdeabad628dda0219ff34a751d5c0735cf55a1c7/src/reducers/importText.ts#L64-L70

Here is the clipboard data when we copy text from the macOS notes app.

```jsx
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
<meta http-equiv="Content-Style-Type" content="text/css">
<title></title>
<meta name="Generator" content="Cocoa HTML Writer">
<meta name="CocoaVersion" content="1894.6">
<style type="text/css">
p.p1 {margin: 0.0px 0.0px 0.0px 0.0px; font: 12.0px 'Helvetica Neue'}
</style>
</head>
<body>
<p class="p1">A</p>
<p class="p1"><span class="Apple-converted-space">  </span>- B</p>
<p class="p1"><span class="Apple-converted-space">  </span>- C</p>
</body>
</html>
```

For this input, bodyContent was returning below.

```
</p>
```

So I decided to use regex to get body content. Now the first problem is solved. Here is the result that bodyContent returns.

```html
<p class="p1">A</p>
<p class="p1"><span class="Apple-converted-space">  </span>- B</p>

```

Also here is the clipboardData, when we copy the text from WebStorm.

```html
<html><head><meta http-equiv="content-type" content="text/html; charset=UTF-8"></head><body><pre style="background-color:#2b2b2b;color:#a9b7c6;font-family:'JetBrains Mono',monospace;font-size:9.8pt;">A<br>&#32;-B<br>&#32;-C<br></pre></body></html>
```

And here is the clipboardData, when we copy the text from IOS notes app.

```html
<meta charset="UTF-8"><p class="p1" style="margin: 0px; font-style: normal; font-variant-caps: normal; font-weight: normal; font-stretch: normal; font-size: 14px; line-height: normal; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-tap-highlight-color: rgba(26, 26, 26, 0.3); -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;"><span class="s1" style="font-weight: normal; font-style: normal; font-size: 14px;">A</span></p><p class="p1" style="margin: 0px; font-style: normal; font-variant-caps: normal; font-weight: normal; font-stretch: normal; font-size: 14px; line-height: normal; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-tap-highlight-color: rgba(26, 26, 26, 0.3); -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;"><span class="s1" style="font-weight: normal; font-style: normal; font-size: 14px;"><span class="Apple-converted-space"> </span>- B</span></p><p class="p1" style="margin: 0px; font-style: normal; font-variant-caps: normal; font-weight: normal; font-stretch: normal; font-size: 14px; line-height: normal; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-tap-highlight-color: rgba(26, 26, 26, 0.3); -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;"><span class="s1" style="font-weight: normal; font-style: normal; font-size: 14px;"><span class="Apple-converted-space"> </span>- C</span></p>
```

---

Okay, the second problem is the `numLines`  is 0 for these inputs. Because there is no `<li>` tag in this input. So, It goes the if block but it shouldn't.
https://github.com/cybersemics/em/blob/cdeabad628dda0219ff34a751d5c0735cf55a1c7/src/reducers/importText.ts#L123

First, I thought I can handle this situation in `importText`. After getting body content, we can strip the content and after that, we can call blocksToHtml to get the result contains `<li> `tag. But it can break existing functionalities and we can not get exact text value as same as the value we would get with `e.clipboardData.getData('text/plain')` by stripping the text. 

And as a result, I decided to handle this issue on paste operation. The only similarity between the examples above I can see is the meta tag. So if the clipboard data contains a meta tag, we can pass the plainText to importText instead of htmlText.

For all examples above, we will pass `A\n - B\n - C` to import text.

---
Note:
If we copy the below text from the notes app or any other ide/text editor, we will pass it as is. There shouldn't be any issue with this example.

```html
<li><i><b>A</b></i></li>
<li><i><b>B</b></i></li>
<li><i><b>C</b></i></li>
```


src/components/Editable.tsx
```diff
@@ -535,7 +541,7 @@ const Editable = ({ disabled, isEditing, simplePath, path, cursorOffset, showCon
         path,
         text: isHTML(plainText)
           ? plainText
-          : htmlText || plainText,
+          : isCopiedFromApp(htmlText) ? plainText : htmlText || plainText,
```
raineorshine, line 544: We need to preserve formatting such as `<b>` and `<i>`, so we can't just take the `plainText`.

src/util/__tests__/importHtml.ts
```diff
@@ -394,3 +399,145 @@ it('set cursor on last thought after importing multiple thoughts in non-empty cu
...
 })
+
+describe('DOM', () => {
+  beforeEach(async () => {
+    await createTestApp()
+  })
+  afterEach(cleanupTestApp)
+
+  it('should paste text properly that is copied from MacOS Notes.app', async () => {
+    /* eslint-disable*/
```
raineorshine, line 410: Explanation for disabling the linter?

raineorshine (changes requested):
Thanks for the detailed explanation!

> And as a result, I decided to handle this issue on paste operation. The only similarity between the examples above I can see is the meta tag. So if the clipboard data contains a meta tag, we can pass the plainText to importText instead of htmlText.

Unfortunately this won't work from both a functional and architectural perspective.

Functionally, we can't use `plainText` because we lose formatting such as `<b>` and `<i>`. We have to maintain the formatting.

Architecturally, `importText` should be able to detect any kind of HTML and handle it appropriately before passing it to `convertHTMLtoJSON`. (I don't think `importHTML` is used anymore.) Processing the input in the paste handler leaves `importText` exposed if it is used elsewhere, such as if we add file upload functionality.

src/util/__tests__/importHtml.ts
onurpolattimur: There are irregular whitespaces in span tags.

`        <p class="p1"><span class="Apple-converted-space">  </span>- B</p>
`

raineorshine: Got it. I recommend just disabling the specific rule rather than all rules.

onurpolattimur:
There were so much different cases than I thought. The good news, we handle more different cases and we import text keeping formatting now! I tried to write a test for each case. You can see these test in importText.ts

Also, we properly import text that is copied from the IOS notes app. But there is one issue about the text that contains formatting copied from IOS notes. IOS notes app doesn't add `<b>, <i>` tags for text formatting. IOS Notes app uses the style attribute to style texts. Here is an example that contains bold text that is copied from the app.

```html
<meta charset="UTF-8"><p class="p1"
                         style="margin: 0px; font-style: normal; font-variant-caps: normal; font-weight: normal; font-stretch: normal; font-size: 14px; line-height: normal; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-tap-highlight-color: rgba(26, 26, 26, 0.3); -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;">
    <span class="s1" style="font-weight: bold; font-style: normal; font-size: 14px;">A</span></p><p class="p1"
                                                                                                    style="margin: 0px; font-style: normal; font-variant-caps: normal; font-weight: normal; font-stretch: normal; font-size: 14px; line-height: normal; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-tap-highlight-color: rgba(26, 26, 26, 0.3); -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;">
    <span class="s1" style="font-weight: bold; font-style: normal; font-size: 14px;"><span class="Apple-converted-space"> </span>B</span></p>
```

We import the text properly for the above input but we are losing formatting. I will look into this problem on another issue because before fixing this issue we need to discuss about this.

raineorshine:
> There were so much different cases than I thought. The good news, we handle more different cases and we import text keeping formatting now! I tried to write a test for each case. You can see these test in importText.ts

Awesome! Thank you!

> We import the text properly for the above input but we are losing formatting. I will look into this problem on another issue because before fixing this issue we need to discuss about this.

Added #1158

raineorshine (approved):
Beautiful! 😍

PR #1155, opened 2021-05-18
Fix same thought duplicate alert regression

shresthabijay:
fixes regression introduced by #1132 

# Changes
- Add rank check to find a duplicate. Prevent checking the value of thought with itself.


raineorshine (approved):

raineorshine:
Thank you!

Issue #1157, opened 2021-05-19
Mobile: Can’t set caret on =focus thought in root


raineorshine:
`=focus/Zoom` is a metaprogramming attribute that zooms in on a specific context by hiding its ancestors and siblings when it is selected. There is an edge case that interacts with the new caret logic. 

Tested on Mobile Safari though it may be reproducible in emulation mode on desktop. 

## Steps to Reproduce

``` 
- A
- B
  - =focus
    - Zoom
``` 

2. Exit edit mode
3. Set the cursor on `B`
4. Tap `B`

## Current Behavior

Cursor becomes `null`

## Expected Behavior

Enter edit mode on `B`

PR #1159, opened 2021-05-19
Update logic that controls whether 'Editable' is visible

onurpolattimur:
fixes: #1157 

___

I fixed the issue but actually, I am not satisfied with `isElementHiddenByAutoFocus` function. 
I thought about this function and I think we can only check the alpha color value of `Editable`

We are hiding the Editable elements by changing their color.  So we can only check the computed color value of Editable in `isElementHiddenByAutoFocus` function. 

Here is my alternative solution:

```js
/** Get alpha value of given color. */
const getAlphaValue = (color: string) => {
  if (color.startsWith('rgba')) {
    const alpha = color.replace(/^.*,(.+)\)/, '$1')
    return parseFloat(alpha)
  }
  return 1
}

/** Returns true if the element has been hidden by the distance-from-cursor autofocus. */
export const isElementHiddenByAutoFocus = (el: HTMLElement) => {
  const color = window.getComputedStyle(el, null).color
  return getAlphaValue(color) === 0
}

```


raineorshine:
The current implementation couples hidden element detection to `editing` and `distance-from-cursor-x` while your alternative solutions couple is to color alpha, which is more straightforward and solves any issues that may arise due to different autofocus behavior.

Feel free to implement it, I think it's a great idea.

raineorshine (changes requested):
Looks good!

Since we're changing the implementation of the function but not the intention or purpose, we shouldn't change its name. We want the call sites to be decoupled from the implementation details. All they should know is that the function will return `true` if the element is currently hidden by the autofocus. Thus I think changing the name back to `isElementHiddenByAutoFocus.ts` would be perfect.

onurpolattimur:
Hi @raineorshine, I have made changes and also I added an e2e util function that is named `waitForElementBecomeHidden`
We need this util function because as you know every operation is really fast in puppeteer and before clicking the hidden thought we need to wait until its transition end.

I looked into whether there is a good way to check a transition is in progress on the element. Unfortunately, there is no good way to handle this without libraries. `transitionend ` event couldn't be enough for this because before adding the event handler, the transition may have already ended.

raineorshine (approved):

raineorshine:
Thanks!

PR #1164, opened 2021-05-22
Pull lexeme if not avaible before exisitingThoughtChange

### LINKED ISSUE #1074: New thoughts not merged into pending lexemes
## Steps to Reproduce

```
- a
- b
  - c
    - d
      - e
        - f
```

1. Set the cursor on `a`. 
1. Refresh the page.
2. Hit Enter to create a new thought and edit it to `f`.

## Current Behavior

`f` has no superscript.

~~If the page is refreshed, the superscript appears, indicating that the Lexeme was actually updated correctly in the db.~~ (No longer true as of the initial merge of the treecrdt.)

## Expected Behavior

`f` should have superscript 2.

## Testing

Unskip https://github.com/cybersemics/em/blob/5e38e2028a34c620d06bf52b530ff3cd12c7e069/src/redux-middleware/__tests__/pushQueue.ts#L16.

shresthabijay:
fixes #1074 

# Cause of the issue

Most of the data is lazily loaded depending on the `expanded`. So when thought is edited to a new value, it is not guaranteed that its related `lexeme` is already loaded in the `state`. So if the `lexeme` is not available `exisitingThoughtChange` simply overrides that `lexeme` in both the `local` and `remote` data sources. So even if the thoughts are loaded later on, the `contextIndex` being pulled from the data sources is already incorrect.

# Solution

Adding an action creator `thoughtChange` that wraps `existingThoughtChange`. It checks if the `lexeme` related to the new value is available in the `state`. If `lexeme` is available then it simply dispatches `exisitingThoughtChange`. If not available, it pulls the `lexeme` from both `local` and `remote`, picks the latest updated `lexeme` among two of them, then updates the state with `updateThoughts` with `remote` and `local` sync turned off and then dispatches the `exisitingThoughtChange`.

@raineorshine  I have changed the `exisitingThoughtChange` in `Editable` to `thoughtChange` and it works. However there are many places `exisitingThoughtChange` is being called besides `Editable`. These may be potential places for similar issues to resurface. We may need to change them as well.

I will write tests and make other required changes after you approve of this solution. Thanks!


raineorshine (changes requested):
Thanks for investigating and explaining the problem!

Edits have to be synchronous for both UX reasons and architectural reasons. Syncing and reconciliation can be asynchronous.

From a UX perspective, it's not acceptable for the user to initiate a change in the editor and then have to wait for IndexedDB and Firebase to be queried before it shows up on the screen. While this may not occur in the `Editable` case due to the throttled edits, there are many other commands that execute all their changes immediately and must be reflected in the `state` immediately.

From an architectural perspective, there are many composite commands that edit thoughts. Making them all asynchronous with side effects would increase complexity dramatically. We made a big effort to convert all action-creators to reducers last year for this reason.

We need a solution that allows the `state` change to be committed immediately, and triggers a `Lexeme` pull that is merged properly so that the new context that the thought appears in is added to the existing `Lexeme`. `existingThoughtChange` should not have to know the details of this mechanism, although it is possible that it may need to expose more information to the `syncQueue` if the middleware does not have what it needs to do the reconciliation. `reconcile` may need to change to be able to handle this case.

shresthabijay:
> Edits have to be synchronous for both UX reasons and architectural reasons. Syncing and reconciliation can be asynchronous.
> 
> From a UX perspective, it's not acceptable for the user to initiate a change in the editor and then have to wait for IndexedDB and Firebase to be queried before it shows up on the screen. While this may not occur in the `Editable` case due to the throttled edits, there are many other commands that execute all their changes immediately and must be reflected in the `state` immediately.
> 
> From an architectural perspective, there are many composite commands that edit thoughts. Making them all asynchronous with side effects would increase complexity dramatically. We made a big effort to convert all action-creators to reducers last year for this reason.

Totally agreed!

shresthabijay:
@raineorshine I have a solution. What do you think ?

- Inside `updateThoughts` check for `thoughtIndex` updates that don't have an entry in the current `thoughtIndex`.

- Create a  `pendingLexemeBatches` in the state where such updates are pushed. If such updates are found then push them to `pendingLexemeBatches` instead of `pushQueue`.
- A separate middleware will handle the `pendingLexemeBatches` that will pull the lexemes for each entry that don't have `thoughtIndex` in the state. Create synced lexeme by getting the latest updated lexeme from the data sources, check for duplicates, and then concat the incoming lexeme update contexts to it. Then add the batch to the `syncedLexemeBatches` (keeps batches that have been sycned) and remove that batch from `pendingLexemeBatches`. (I am suggesting union to reconcile the lexeme contexts because I am assuming that for a delete operation, this scenario never occurs. As for a loaded thought, the corresponding lexeme is always pulled)

- `pushQueue` middlware should prioritize pushing from `syncedLexemeBatches` until `pendingLexemeBatches` and `syncedLexemeBatches` are empty. So all the push batches on `pushQueue` will be on hold till then. Else it may cause updates in wrong order, causing incosistency.

raineorshine:
> * Create a  `pendingLexemeBatches` in the state where such updates are pushed.

Yup, that sounds good.

> * A separate middleware will handle the `pendingLexemeBatches`

Since the `pushQueue` already handles `pendingDeletes`, `pendingEdits`, and `pendingMoves`, calling `pull` as needed, I think it should also handle `pendingLexemeBatches` instead of another middleware.

> * Create synced lexeme by getting the latest updated lexeme from the data sources, check for duplicates, and then concat the incoming lexeme update contexts to it.

We'll want to use the existing `reconcile` reducer to merge the lexeme, as the reconciliation logic, state update, and caching mechanism is too complex to duplicate. It was designed to handle both `contextIndexUpdates` and `thoughtIndexUpdates` but I think it will work with an empty `contextIndexUpdates`.

> * Then add the batch to the `syncedLexemeBatches` (keeps batches that have been sycned) and remove that batch from `pendingLexemeBatches`

If we do everything in `pushQueue`, can we avoid an additional batch array? There seems to be a lot of state to maintain between different pull/push steps.

> * I am suggesting union to reconcile the lexeme contexts because I am assuming that for a delete operation, this scenario never occurs. As for a loaded thought, the corresponding lexeme is always pulled

That is correct.

> * `pushQueue` middlware should prioritize pushing from `syncedLexemeBatches` until `pendingLexemeBatches` and `syncedLexemeBatches` are empty. So all the push batches on `pushQueue` will be on hold till then. Else it may cause updates in wrong order, causing incosistency.

I'm not quite able to wrap my head around this part. Is there any way to do this with a single `pushQueue`? Batches are already quite complex with `contextIndexUpdates`, `thoughtIndexUpdates`, `pendingEdits`, etc. 

Multi-step syncs are really difficult to pull off elegantly, and I think the current pending/flush system may already be vulnerable to some data inconsistencies (even more of a reason we need #495).

raineorshine:
Resume discussion in #1074

Issue #949, opened 2020-12-26
splitSentence should work on `;?!`


raineorshine:
Currently the `splitSentences` shortcut just splits on period. It should split on `/[.;?!]/`. 

See #853 and #859 for how the `splitSentences` shortcut works.

linghong:
It seems that the splitSentences also works for /[.;?!]/, but it always replace those symbols  with "."  Do you mean you want to replace "."  with the original symbol?

Currently in some special cases, for example 333.12345, it will splits into 333. and 12345. Do you want to keep it like that way, or  if it is a decimal  number, just don't split them.  

There are also other special cases, like the below:
sentense A?! sentense B
"sentense A." sentenseB.
(sentense A!) sentenseB.

Do you prefer our code to address those issues, or just split them whenever any of the /[.;?!]/ symbols occurs  and let the users themselves to adjust the result?

raineorshine:
> It seems that the splitSentences also works for /[.;?!]/, but it always replace those symbols with "." Do you mean you want to replace "." with the original symbol?

Yes, I didn't realize that! We want to preserve the original symbol.

> Currently in some special cases, for example 333.12345, it will splits into 333. and 12345. Do you want to keep it like that way, or if it is a decimal number, just don't split them.

It is not expected that the user would activate `splitSentences` on such thoughts, so I think we can leave it as-is. Thanks for noticing though.

> There are also other special cases, like the below:
> sentense A?! sentense B
> "sentense A." sentenseB.
> (sentense A!) sentenseB.

- `sentense A?! sentense B` → `sentense A?!` and `sentense B`
- `"sentense A." sentenseB.` → `"sentense A."` and `sentenseB.`
- `(sentense A!) sentenseB.` → `(sentense A!)` and `sentenseB.`

Good catch on those. It is more logic than I originally expected when I designed the feature, but they are all cases that I would like to cover.

> Do you prefer our code to address those issues, or just split them whenever any of the /[.;?!]/ symbols occurs and let the users themselves to adjust the result?

Thanks for your thoroughness, I really appreciate that. I'd love for you to address all the cases you raised.

PR #1168, opened 2021-05-22
Split sentences (#949)

linghong:
Split sentences when one ore more of symbols  .;!? exist.

It only split sentences when the result will look like real sentences. For example, words like Mr. Jack Smith or $3.2M won't be split. 

An example of sentences that can be split nicely:
call Dr. Smith to schedule an office visit on Sept. 23; remember at least free one full day this week to clean my house!!! call Verizon to ask why it charges me  an extra $35.5 on top of regular bill; buy some fruits tomorrow, e.g. banana, apple, orange; do more exercises (increase to 2 hours per day?)

### DIFF
--- a/src/util/isAbbreviation.ts
+++ b/src/util/isAbbreviation.ts
@@ -1,73 +1,78 @@
 /**
- * Function: isAbbrEndSplitter.
+ * Function: isAbbrev.
  *
- * @param str The sentence just added into the newSentences array, it contains the sentence and the first splitter.
+ * @param str1 The sentence just added into the result sentence, it contains the sentence and the splitter.
  * @param s The current sentence which is right behind the spliter.
  * @param spliter2 The second spliter.
  * @returns A boolean value that tells whether the dot comes from an Abbrev word, and shouldn't be split
- * Examples: Mr. Dr. Apt. Prof. Ph.D.
+ * Examples: Mr. Dr.q Apt. Prof. Ph.D.
  */
-export function isAbbrev(prevSentence : string, s: string, spliter2 : string) {
-  return isAbbrEndSplitter(prevSentence, s) || isAbbrMidSplitter(prevSentence, s) || isAbbrDoubleSplitter(prevSentence, s, spliter2)
+export function isAbbrev(str1 : string, s: string, spliter2 : string) {
+
+  return isAbbrEndSplitter(str1, s) || isAbbrMidSplitter(str1, s) || isAbbrDoubleSplitter(str1, s, spliter2)
 }
 
 /**
  * Function: isAbbrEndSplitter.
  *
- * @param str The sentence just added into the newSentences array, it contains the sentence and the first splitter.
- * @param s The current sentence which is right behind the spliter.
- * @returns A boolean value that tells whether the dot comes from an Abbrev word, and shouldn't be split
+ * @param str1 The sentence just added into the result sentence, it contains the sentence and the splitter.
+ * @param s The current sentence which sits right behind the spliter.
+ * @returns A boolean value that tells whether the dot comes from an Abbrev word that has an dot on the word end, and shouldn't be split
  * Examples: Mr. Dr. Apt. Feb. Prof.
  */
-function isAbbrEndSplitter (str : string, s : string) {
+function isAbbrEndSplitter (str1 : string, s : string) {
+  if (str1[str1.length - 1] !== '.') return false
+
   const twoLetters = ['Mr.', 'Ms.', 'Dr.', 'Rm.', 'No.', 'no.', 'vs.', 'Rd.', 'St.', 'Co.', 'Jr.', 'pl.', 'Sr.']
   const threeLetters = ['Mrs.', 'Jan.', 'Feb.', 'Jun.', 'Aug.', 'Oct.', 'Nov.', 'Dec.', 'Apt.', 'est.', 'Ave.', 'Tel.', 'alt.', 'Col.', 'inc.', 'Ltd.', 'vol.']
   const fourLetters = ['Sept.', 'addr.', 'Prof.', 'Dept.', 'temp.', 'Blvd.', 'abbr.', 'Assn.', 'Corp.', 'Univ.']
   const fiveLetters = ['et al.']
   const sixLetters = ['Abbrev.', 'approx.']
 
-  const len = str.length
-  const isTwoLetters = twoLetters.includes(str.slice(len - 3, len))
-  const isThreeLetters = threeLetters.includes(str.slice(len - 4, len))
-  const isFourLetters = fourLetters.includes(str.slice(len - 5, len))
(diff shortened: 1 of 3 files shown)


raineorshine (changes requested):
Before I do a detailed code review, please make the following changes:

1. Fix failing test
2. Add tests for your new test cases
3. Find a way to implement the desired functionality without mutations. (Mutations are considered a bad practice. The requirement for a more functional programming style is described in the Contributor Guidelines and job posting.)

Thanks and let me know if you have any questions!

linghong:
I checked your test file. I got really confused. You file is named as splitSentence.ts. But the example thoughts used on your test code is not sentences.  The reason it raises my concern is because in my code if it is Abbrev. word,  such as Dr. or Mr.,  the sentence won't get split. I clarify  an Abbrev. word as the first letter is capital letter, followed by a few small letters and ends with a dot and an empty space.  In your test case the sample sentences is "One. Two. Three.", which match that description.

My question is what is a real use case of your app? Are the thoughts only for listing phrases or words, not for sentences?

raineorshine:
> I checked your test file. I got really confused. You file is named as splitSentence.ts. But the example thoughts used on your test code is not sentences.

The test cases are short for simplicity but they do demonstrate the expected behavior of `splitSentence`. The length of the sentence should not affect the behavior. For example, the thought "One. Two. Three." should split the same way as "Number One. Number Two. Number Three." or "I have one. I have two. I have three." All of those examples have three sentences ending in periods from the perspective of **em**.

> I clarify an Abbrev. word as the first letter is capital letter, followed by a few small letters and ends with a dot and an empty space.

This is overgeneral and matches many cases that are not abbreviations, as you discovered. We have to whitelist abbreviations, as there is no general way to determine the difference between a sentence ending and an abbreviation without parsing grammar and doing other natural language processing (which **em** does not do).

> My question is what is a real use case of your app? Are the thoughts only for listing phrases or words, not for sentences?

Common use cases of **em** include the user inputting words, phrases, sentences, or paragraphs, either through typing or pasting other sources. `splitSentence` is most useful for pasting a paragraph from a source and then splitting it into sibling thoughts for a more idiomatic structure. 

Entering small words or phrases will be more common than entering abbreviations, so the former should be the base case and the latter should be a special case handled via a whitelist. Let's add a few common abbreviations like "Mr.", "Mrs." and "Dr." and leave it at that for now. We can always add to the list or add more intelligent heuristics later.

linghong:
Thank you for the detailed explanation. I'll change the abbreviations to a white list.

linghong:
New modifications:
1.  added test cases;
2. replaced the forEach function with array reduce function to pass through a newly formatted sentence string;
3. added code for dealing with  email and url address;
4. fixed some edge cases.

src/util/isAbbreviation.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @param str1 The sentence just added into the result sentence, it contains the sentence and the splitter.
+ * @param s The current sentence which is right behind the spliter.
+ * @param spliter2 The second spliter.
+ * @returns A boolean value that tells whether the dot comes from an Abbrev word, and shouldn't be split
+ * Examples: Mr. Dr.q Apt. Prof. Ph.D.
+ */
+export function isAbbrev(str1 : string, s: string, spliter2 : string) {
+
+  return isAbbrEndSplitter(str1, s) || isAbbrMidSplitter(str1, s) || isAbbrDoubleSplitter(str1, s, spliter2)
+}
```
raineorshine, line 13: Use a single default export:

```js
export default isAbbrev
```

src/util/splitSentence.ts
```diff
@@ -1,16 +1,160 @@
...
+  /**
+   * When the setences can be split, it has multiple situations.
+   */
+  const sentences = value.split(mainSplitRegex)
+
+  /**
+   * The reduce function return a string, which is a combination of all the sentences, we then can split it during the next step.
+   * Google searching keyword '^^' or '^#' return 0 result for both of them. Hence, it is safe in using the below seperator to link each sentence into one string.
+   */
+  const seperator = value.indexOf('^^') > 0 ? '^#' : '^^'
```
raineorshine, line 34: This is a good idea, but let's use a dummy value like `__SEP__` instead, which is clearer and even more conflict-proof. Define it as a const at the top of the file, like this:

https://github.com/cybersemics/em/blob/66855eafb13b103520aaee9d0a7a4a60f00346fd/src/util/hashContext.ts#L10

src/util/splitSentence.ts
```diff
@@ -1,16 +1,160 @@
...
+  const initialValue = sentences[0]
+
+  const resultSentences = sentences.reduce((newSentence : string, s : string, i : number) => {
+
+    if (i === 0) return newSentence + spliters[0]
```
raineorshine, line 39: There is a typo: `spliters` should be `splitters`. Please do a search and replace to fix all instances of `spliter`.

raineorshine (changes requested):
Thank you! This turned into a much more complex task than imagined, but you have been extremely thorough! Thanks for writing all the test cases and documenting it so well. It is very clean.

What do you think about ignoring splitter characters that are not followed by whitespace? Sentences are always followed by one or two spaces, so anything that is not followed by whitespace should not be split. I think that may simplify the logic, for example abbreviations, email addresses, etc.

linghong:
> What do you think about ignoring splitter characters that are not followed by whitespace? Sentences are always followed by one or two spaces, so anything that is not followed by whitespace should not be split. I think that may simplify the logic, for example abbreviations, email addresses, etc.

I think it is better not split if there is no white space followed by a splitter. By doing that, we still need to have white list for abbreviations, but we can simplify the logic for email and url address.  Or we might do the below way:  we  only split the thought if the first letter followed by the splitter is a Capital letter when there is no empty space immediately after the splitter.  Because, what I am more concern is beside the special cases I have written, I now find other specials ones, such as file names, js library names, IP address. Maybe there are even more we will find later. However, it seems that the complex issue most comes from the "." symbol.  Hence, I think another way we can do is that we make the default condition as below: if the splitter is a dot, we only split the thought when it has empty spaces after the dot, but for other splitters we will still split them.

linghong:
> Yes, I totally agree!
> 
> Thanks for sticking with it 🙂

I would like to be clear, for the three default rules I mentioned before, which way do you prefer?
1. Do not split the thought if there is no white space followed by a splitter. 
2. Only split the thought if the first letter followed by the splitter is a capital letter when there is no empty space immediately after the splitter.
3. Only split the thought when it has empty spaces after the dot, but for other splitters we will still split them no matter it has white spaces or not.

raineorshine:
Let's go with #<b/>3!

src/util/isAbbreviation.ts
raineorshine: Oops! I actually got this wrong. You were right the first time :). No problem though.

raineorshine (approved):
Thank you!!!

PR #1171, opened 2021-05-23
add proper rank to moved thoughts in collapse context

### LINKED ISSUE #1165: collapseContext: Wrong rank
## Steps to Reproduce

``` 
- a
- b
- c
  - d
``` 

Activate `collapseContext` in `c`

## Current Behavior

``` 
- a
- d
- b
``` 

## Expected Behavior

``` 
- a
- b
- d
``` 

## Notes

The existing `collapseContext` test must not be testing the order of children after merge. If that is the case, you should be able to add an additional assertion instead of a new unit test.

shresthabijay:
fixes #1165 

# Cause of the issue

`newRank` was not provided to the moved thoughts in `collapseContext`. Moved thoughts were using the rank they were assigned in the previous context.

# Solution

Calculate `newRank` and pass it to the moved thoughts.


shresthabijay:
@raineorshine The changes I made seems to break many tests. It is because the `importText` also uses the `collapseContext`. `importText` first creates a dummy thought and imports the thought inside it and then collapses it to merge. These causes rank to be different than expected.

Let's take an example here. 

```
- a
- b
- c
```

If you import this text we will get the following ranks.
```
- a (-1)
- b (-0.5)
- c (-0.25)
```
It is because of the intermediary steps that is involved in `importText`

First a dummy thought is created.

```
- dummy (0)
 - a (0)
 - b (1)
 - c (2)
```

Then we collapse dummy thought.

The first step of collapsing is to move the children to the parent context. Doing that we need to use `getRankBefore` the current cursor as the cursor has not been deleted yet. Since the rank of cursor (dummy) is 0, it will generate -1, -0.5 and -0.25 for the moved thoughts. Only after that the cursor is deleted or archived.

The generated ranks are correct but looking from the viewpoint of just import in text, it doesn't make sense. That is why most of the test are failing.

One solution is to find the child right before `cursor` and then make a copy of `state `with the updated `contextIndex`entry of the parent context with cursor removed. Pass that state to the `getRankAfter` to get the correct ranks as we expect. What do you think we should do?

raineorshine (commented):
> The generated ranks are correct but looking from the viewpoint of just import in text, it doesn't make sense.

Maybe the tests are wrong to assume specific ranks? All that matters should be the relative order.

> One solution is to find the child right before `cursor` and then make a copy of `state `with the updated `contextIndex`entry of the parent context with cursor removed. Pass that state to the `getRankAfter` to get the correct ranks as we expect. Just because we know the cursor will eventually be deleted or achieved after the childrens are moved.

That seems possible, although it seems like a lot of effort to get specific absolute ranks when ranks are inherently not absolute. If we are doing that in order to make ranks in tests more readable, it may be better to simply fix the tests so they do not depend on absolute ranks.

Another approach would be to avoid the collapsing dummy thought completely when the destination context is empty, which would apply to all tests since they always import into `HOME_CONTEXT`.

> That is why most of the tests are failing.

I could be wrong but it seems that some of the tests are failing due to some other issue. 

For example here it appears the ordering is wrong:

```
 ● merge children

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 2
    + Received  + 2

      - __ROOT__
        - a
          - =archive
            - b
          - c
    -     - x
    +     - d
    -     - d
    +     - x

      92 |   const exported = exportContext(stateNew, [HOME_TOKEN], 'text/plain')
      93 | 
    > 94 |   expect(exported).toBe(`- ${HOME_TOKEN}
         |                    ^
      95 |   - a
      96 |     - =archive
      97 |       - b

      at Object.<anonymous> (src/reducers/__tests__/collapseContext.ts:94:20)
```

And here the move is not completed correctly:

```
FAIL src/reducers/__tests__/existingThoughtMove.ts
  ● move root thought into another root thought

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 0
    + Received  + 1

      - __ROOT__
        - x
          - a
            - b
              - c
    +   - a

      255 |   const stateNew = reducerFlow(steps)(initialState())
      256 |   const exported = exportContext(stateNew, [HOME_TOKEN], 'text/plain')
    > 257 |   expect(exported).toBe(`- ${HOME_TOKEN}
          |                    ^
      258 |   - x
      259 |     - a
      260 |       - b

      at Object.<anonymous> (src/reducers/__tests__/existingThoughtMove.ts:257:20)
```

shresthabijay:
> For example here it appears the ordering is wrong:

The test itself seems to assume the wrong order here. I have changed it.

shresthabijay:
> Maybe the tests are wrong to assume specific ranks? All that matters should be the relative order.

I agree. Only the relative order matters. Should we slowly remove rank checks from test where not needed ?

src/reducers/importText.ts
```diff
@@ -199,30 +199,49 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
...
     const uuid = createId()
 
+    const isDestContextEmpty = getAllChildren(state, context).length === 0
+
+    /** Check if destination's parent context has more than one children. */
+    const isDestParentContextEmpty = () => getAllChildren(state, rootedParentOf(state, context)).length <= 1
+
+    const destEmpty = destThought.value === '' && isDestContextEmpty
+
+    const noNeedOfCollapse = (!destEmpty && isDestContextEmpty) || (destEmpty && isDestParentContextEmpty())
```
raineorshine, line 209: I think this is the same as  `destEmpty ? isDestParentContextEmpty() : isDestContextEmpty`? If so, that might be easier to read.

src/reducers/importText.ts
```diff
@@ -234,9 +253,9 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
     return reducerFlow([
       updateThoughts(imported),
       // set cusor to destination path's parent after collapse unless it's em or cusor set is prevented.
-      collapseContext({ deleteCursor: true, at: newDestinationPath }),
+      !noNeedOfCollapse ? collapseContext({ deleteCursor: true, at: newDestinationPath }) : null,
```
raineorshine, line 256: Double negatives are harder to read, so it would be better to negate the value and call it something like `useDummy` or `importIntoDummy`.

src/reducers/importText.ts
```diff
@@ -199,30 +199,49 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
...
     const setLastImportedCursor = (state: State) => {
 
       const lastImportedContext = pathToContext(imported.lastImported)
-      const cursorContextHead = lastImportedContext.slice(0, newDestinationPath.length - (destEmpty ? 2 : 1))
-      const cursorContextTail = lastImportedContext.slice(newDestinationPath.length)
-      const newCursorContext = cursorContextHead.concat(cursorContextTail)
-      const newCursor = rankThoughtsFirstMatch(state, newCursorContext)
+
+      /** Get last iumported cursor after using collapse. */
+      const getLastImportedAfterCollpase = () => {
```
raineorshine, line 237: Typo in function name

src/reducers/importText.ts
```diff
@@ -199,30 +199,49 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
...
+        const cursorContextHead = lastImportedContext.slice(0, newDestinationPath.length - (destEmpty ? 2 : 1))
+        const cursorContextTail = lastImportedContext.slice(newDestinationPath.length)
+        const newCursorContext = cursorContextHead.concat(cursorContextTail)
```
raineorshine, line 240: I recommend getting in the habit of `[...cursorContextHead, ...cursorContextTail]` as its cleaner and faster.

raineorshine (changes requested):
Logic looks good to me! Just a few small recommendations for code readability. Thanks!

raineorshine:
> I have fixed some `collapseContext` texts that assume the wrong order and ranks.

Good catch! Thanks!

> > Maybe the tests are wrong to assume specific ranks? All that matters should be the relative order.
> 
> I agree. Only the relative order matters. Should we slowly remove rank checks from test where not needed ?

Yes, we can write all new tests without absolute ranks and slowly migrate the old tests.

In general it will be preferable to call `children.map(c => c.value)` and perform the assertion on the array of strings since `{ value }` is not necessary (unless other properties like `id` are needed). It will create more readable test output.

raineorshine (approved):
Beautiful, thanks!

PR #1179, opened 2021-05-23
Add iOS safari testing using webdriverio & browserstack

onurpolattimur:
We will now be able to perform mobile tests on ios via webdriverio and browserstack.🥳 Because we are using jest for puppeteer, I set up a new jest environment for webdriverio.

The device on browserstack can not access localhost, so we need browserstacklocal. You can see details at [here](https://www.browserstack.com/local-testing/automate).

I wrote some util functions for webdriveio. They can be useful for the next tests. I also added one test as an example that is like the caret test. 

If you want to run these tests on your local machine, first of all, you need to have macOS and you need to setup some required applications such as Appium, Xcode, etc.. But, I think we can register the team in browserstack and the team can run tests by setting some environment variables on their local development environment without setup an application.

After the test run, you can see all details about the tests including a video that is captured while the test is running on browserstack.com

Here is an example:

image

### How to run?

To run tests on BrowserStack, you need to have an access key. @raineorshine will send you an invite. If you are not registered yet, you should create an account on BrowserStack.

After logging, you will see your username and access key on the dashboard page. Copy these values.

```
BROWSERSTACK_USERNAME={your_username} BROWSERSTACK_ACCESS_KEY={your_access_key} npm run test:e2e:ios
```

After running the above command, you can see your test result on BrowserStack.


src/test-helpers/e2e-helpers/mobile/clickThought.ts
```diff
@@ -0,0 +1,14 @@
+import getEditable from './getEditable'
+import { Browser } from 'webdriverio'
+
+/** Click the thought for the given thought value. */
+const clickThought = async (browser: Browser<'async'>, value: string) => {
+
+  const editableNode = await getEditable(browser, value)
+
+  if (!editableNode) throw new Error('editable node for the given value not found.')
```
raineorshine, line 9: You could add `value` to the error message to make it a bit more useful in the test output.

src/test-helpers/e2e-helpers/mobile/execute.ts
```diff
@@ -0,0 +1,11 @@
+import { Browser } from 'webdriverio'
+
+/** Execute a function in browser context. */
+const execute = (browser: Browser<any>, func: (...args: any) => any, ...args: any[]) => {
+  // https://github.com/webdriverio/webdriverio/issues/6206
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  return browser.execute(func, ...args)
+}
```
raineorshine, line 9: This function just wraps `browser.execute`. Can you call `browser.execute` directly instead?

src/test-helpers/e2e-helpers/mobile/gestures/newThoughtGesture.ts
```diff
@@ -0,0 +1,18 @@
+import { Browser } from 'webdriverio'
+
+/** New thought gesture. */
+const newThoughtGesture = (browser: Browser<'async'>) => {
+  return browser.touchPerform(
```
raineorshine, line 5: That's so awesome that we can execute touch gestures!!!

src/test-helpers/e2e-helpers/mobile/paste.ts
```diff
@@ -0,0 +1,14 @@
+import { Browser } from 'webdriverio'
+import execute from './execute'
+
+/** Import text on given unranked path using exposed testHelpers. */
+const paste = async (browser: Browser<'async'>, unrankedPath: string[], text: string) => {
```
raineorshine, line 5: Does `webdriverio` have a way to trigger a native `paste` event so that we don't have to use the hack that we needed for puppeteer?

src/test-helpers/e2e-helpers/mobile/tapReturn.ts
```diff
@@ -0,0 +1,12 @@
+import { Browser } from 'webdriverio'
+
+/** Tap 'return' on the keyboard. */
+const tapReturn = async (browser: Browser<'async'>) => {
```
raineorshine, line 4: Let's call this `tapReturnKey` which is a bit less ambiguous. It took me a second to realize what it was referring to.

raineorshine (changes requested):
Beautiful! I'm very excited about this!

raineorshine:
Can you provide some more guidance about how to run the tests on browserstack? I created a trial account and added `BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY` to the repo secrets.

src/test-helpers/e2e-helpers/mobile/execute.ts
onurpolattimur: I did this because `browser.execute` has some type problems and I didn't want to add lines 5-6-7 to everywhere that call `browser.execute`.

src/test-helpers/e2e-helpers/mobile/gestures/newThoughtGesture.ts
onurpolattimur: Yesss! 🥳

onurpolattimur:
To BrowserStack can access em app, we need to download and run the app named BrowserStackLocal. To prevent download/run it from CLI, I added `browserstacklocal` library. Now, it is so easy to run tests!

To run tests on BrowserStack, you need to have an access key. @raineorshine will send you an invite. If you are not registered yet, you should create an account on BrowserStack.

After logging, you will see your username and access key on the dashboard page. Copy these values.

```
BROWSERSTACK_USERNAME={your_username} BROWSERSTACK_ACCESS_KEY={your_access_key} npm run test:e2e:ios
```

After running the above command, you can see your test result on BrowserStack.

src/test-helpers/e2e-helpers/mobile/paste.ts
raineorshine: No problem. If webdriverio doesn't support it then we probably need to inject it like we are doing. What do you mean by separate util function?

src/test-helpers/e2e-helpers/mobile/execute.ts
raineorshine: I think I fixed it in 556c93b28

package.json
```diff
@@ -33,12 +33,14 @@
...
     "test": "react-scripts test --testPathIgnorePatterns=./src/e2e",
-    "test:e2e": "react-scripts test ./src/e2e"
+    "test:e2e": "react-scripts test ./src/e2e --testPathIgnorePatterns=./src/e2e/__tests__/iOS",
+    "test:e2e:ios": "react-scripts test ./src/e2e/__tests__/iOS"
   },
   "dependencies": {
     "@bit/mui-org.material-ui.swipeable-drawer": "^4.9.10",
     "algoliasearch": "^4.9.1",
     "axios": "^0.21.1",
+    "browserstack-local": "^1.4.8",
```
raineorshine, line 43: This should be a `devDependency`

src/e2e/__tests__/iOS/sample.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Browser } from 'webdriverio'
+import getEditingText from '../../../test-helpers/e2e-helpers/mobile/getEditingText'
+import waitForElement from '../../../test-helpers/e2e-helpers/mobile/waitForElement'
+import newThoughtGesture from '../../../test-helpers/e2e-helpers/mobile/gestures/newThoughtGesture'
+import paste from '../../../test-helpers/e2e-helpers/mobile/paste'
+import waitForEditable from '../../../test-helpers/e2e-helpers/mobile/waitForEditable'
+import clickThought from '../../../test-helpers/e2e-helpers/mobile/clickThought'
+import clickWithOffset from '../../../test-helpers/e2e-helpers/mobile/clickWithOffset'
+// import execute from '../../../test-helpers/e2e-helpers/mobile/execute'
+import tapReturnKey from '../../../test-helpers/e2e-helpers/mobile/tapReturnKey'
```
raineorshine, line 14: What do you think about housing the e2e test helpers in `/src/e2e/helpers`? They are currently a bit far away.

src/e2e/webdriverio-environment.js
```diff
@@ -0,0 +1,85 @@
...
+        port: 3000
+      })
+    }
+    try {
+      this.startBrowserStackLocal()
+      this.global.browser = await wdio.remote(config)
+    }
+    catch (e) {
+      console.error(e)
+    }
```
raineorshine, line 55: Why catch the error? I think we would want it to fail.

src/test-helpers/e2e-helpers/mobile/waitForEditable.ts
```diff
@@ -0,0 +1,14 @@
+import { Browser } from 'webdriverio'
+import getEditable from './getEditable'
+
+/** Wait for editable element that contains given value. */
+const waitForEditable = async(browser: Browser<'async'>, value: string) => {
+  await browser.waitUntil(async () => await browser.execute(value => {
+    return Array.from(document.getElementsByClassName('editable'))
+      .filter(element => element.innerHTML === value).length > 0
+  }, value)
+  , { timeout: 11000 })
```
raineorshine, line 10: Can we defer to the containing test's timeout? Ideally helper functions should not be specifying their own timeouts.

Same for `waitForElement` and `waitForElementNotExist`

raineorshine (changes requested):

raineorshine:
I'm seeing two tests run on browserstack even though we only have one mobile test. One of them is empty and exits with success as soon as the session starts.

image

Any ideas what might be causing that?

My only guess is that maybe it is the call to `reloadSession` at the end of the test:

https://github.com/cybersemics/em/blob/556c93b28bcfd80442871a482b475ab7183c330b/src/e2e/__tests__/iOS/sample.ts#L27-L29

src/test-helpers/e2e-helpers/mobile/execute.ts
onurpolattimur: 😄 Thank you!

src/test-helpers/e2e-helpers/mobile/paste.ts
onurpolattimur: Webdriver io doesn't support paste events but I think we can combine some events to achieve paste events.
For example, Long press the editable and then find a native element that contains 'paste' text and then touch this native element, etc.  I am not sure if it will work but I will give it a shot at a more convenient time.

src/e2e/__tests__/iOS/sample.ts
onurpolattimur: Yes, good idea.

src/e2e/webdriverio-environment.js
onurpolattimur: Otherwise, I couldn't not see the error message when error occurred on wdio.remote call. It will also fail but if you want we can throw an exception.

onurpolattimur:
> I'm seeing two tests run on browserstack even though we only have one mobile test. One of them is empty and exits with success as soon as the session starts.
> 
> image
> 
> Any ideas what might be causing that?
> 
> My only guess is that maybe it is the call to `reloadSession` at the end of the test:
> 
> https://github.com/cybersemics/em/blob/556c93b28bcfd80442871a482b475ab7183c330b/src/e2e/__tests__/iOS/sample.ts#L27-L29

Yes, that is the reason. I don't know how we can prevent this. Do you have any idea?

src/e2e/webdriverio-environment.js
raineorshine: Yes, you can both `console.error` and throw if you need to.

raineorshine:
> Yes, that is the reason. I don't know how we can prevent this. Do you have any idea?

I suggest moving it to the beginning of the `beforeEach` in a conditional that ensures is only runs if a browser session is already active. You can add a short comment clarifying the use.

onurpolattimur:
> > Yes, that is the reason. I don't know how we can prevent this. Do you have any idea?
> 
> I suggest moving it to the beginning of the `beforeEach` in a conditional that ensures is only runs if a browser session is already active. You can add a short comment clarifying the use.

I think it won't change anything. Because the first session is created on `await wdio.remote(config)` call. So when we move the reloadSession statement from `afterEach`  to `beforeEach` block, it will create still two sessions on BrowserStack. The only change possibly will be the order of the sessions on BrowserStack.

raineorshine:
> > > Yes, that is the reason. I don't know how we can prevent this. Do you have any idea?
> > 
> > 
> > I suggest moving it to the beginning of the `beforeEach` in a conditional that ensures is only runs if a browser session is already active. You can add a short comment clarifying the use.
> 
> I think it won't change anything. Because the first session is created on `await wdio.remote(config)` call. So when we move the reloadSession statement from `afterEach` to `beforeEach` block, it will create still two sessions on BrowserStack. The only change possibly will be the order of the sessions on BrowserStack.

Wouldn't the conditional statement prevent that? If `reloadSession` is not called, it won't start an additional session. In this case it's not about the ordering: I just figured it would be easier to detect if it was the first test rather than the last test (which would rely on knowing how many tests there are).

The blank test is unnecessarily increasing the browserstack minutes that are used. The documentation even gives a warning:

> Be careful though, this command affects your test time tremendously since spawning new Selenium sessions is very time consuming especially when using cloud services.

https://webdriver.io/docs/api/browser/reloadSession/

If you could take a stab at it that would be great! If you're not able to get anywhere with it I am happy to give it a try myself. Thanks!

onurpolattimur:
Hi @raineorshine, I made changes. Now, we see one session on BrowserStack.

raineorshine (approved):
Beautiful!

Issue #1174, opened 2021-05-23
Empty table context has extra space 


raineorshine:
## Steps to Reproduce

``` 
- a
  - =view
    - Table
- b
``` 

Set the cursor on `a`

## Current Behavior

Space is created beneath `a`

![1D0E63CF-16C9-4836-AF43-9BDB83FDCDC1](https://user-images.githubusercontent.com/750276/119262310-c4778880-bba8-11eb-8248-466890ea1439.jpeg)

## Expected Behavior

No space would be created since `a` has no children.

PR #1180, opened 2021-05-24
Fix empty table context having extra space

ruchijn:
Fix : #1174 

Before:
1174_bug

After:
1174_fix

### DIFF
--- a/src/components/__tests__/Subthoughts.ts
+++ b/src/components/__tests__/Subthoughts.ts
@@ -248,6 +248,70 @@ describe('context view', () => {
 
   })
 
+  it('renders EmptyChildrenDrop when no subthought ', () => {
+
+    const now = timestamp()
+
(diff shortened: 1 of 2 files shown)


ruchijn:
Please do not review. Need some changes.

src/components/__tests__/Subthoughts.ts
```diff
@@ -248,6 +248,70 @@ describe('context view', () => {
...
+
+    // import thoughts
+    store.dispatch([
+      importText({
+        path: HOME_PATH,
+        text: `
+          - a
+          - b
+        `,
+        lastUpdated: now,
```
raineorshine, line 263: Since `now` is the default value of `lastUpdated`, I think you can omit it.

raineorshine (approved):

Issue #1177, opened 2021-05-23
Input styling broken on login page


raineorshine:
![image](https://user-images.githubusercontent.com/750276/119267352-7a43d680-bbab-11eb-9b43-1b120b83ad69.png)

PR #1183, opened 2021-05-24
Fix login screen styling issue

ruchijn:
Fix: #1177

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -77,7 +77,9 @@ li {
   list-style: none;
 }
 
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
Let's use a whitelist approach rather than a blacklist approach since [there are many input types](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types).

raineorshine (approved):

raineorshine:
Thank you!

PR #1190, opened 2021-05-27
Use pullLexemes to pull pending lexemes for each batch

### LINKED ISSUE #1074: New thoughts not merged into pending lexemes

shresthabijay:
fixes #1074 

# Changes

- Add `pendingLexemes` to `PushBatch`. Calculating `pendingLexemes`  in `pushQueue` won't work as the application state is already updated in `updatedThoughts`. So it needs to be done in `updatedThoughts` and added to the batch.
- `pushQueue` now first checks for `pendingLexemes` in the batches and calls `pullLexemes`. `pullLexemes` gets the lexeme from local and remote source and reconciles them.
- `reconcile` is updated to call `updateThoughts` if only `thoughtIndexOnly` is available.


src/reducers/__tests__/deleteEmptyThought.ts
```diff
@@ -248,28 +251,32 @@ describe('mount', () => {
...
 
-  it('after deleteEmptyThought, caret should move to end of previous thought', () => {
-    store.dispatch([
-      { type: 'newThought', value: 'apple' },
-      { type: 'newThought' },
-      { type: 'deleteEmptyThought' }
-    ])
-    jest.runOnlyPendingTimers()
+  it('after deleteEmptyThought, caret should move to end of previous thought', async () => {
+    await fakeTimer.runAsyncCode(() => {
```
raineorshine, line 255: Could you explain the reason for adding `runAsyncCode` here? I'd like to understand this more.

src/reducers/updateThoughts.ts
```diff
@@ -62,6 +62,15 @@ const updateThoughts = (state: State, { thoughtIndexUpdates, contextIndexUpdates
 
   const recentlyEditedNew = recentlyEdited || state.recentlyEdited
 
+  //  lexemes from the updates that are not available in the state yet.
+  const pendingLexemes = Object.keys(thoughtIndexUpdates).reduce<Index<string>>((acc, thoughtId) => {
+    const lexemeInState = state.thoughts.thoughtIndex[thoughtId]
+    return {
+      ...acc,
+      ...lexemeInState ? {} : { [thoughtId]: thoughtId }
```
raineorshine, line 70: Why use `{ [thoughtId]: thoughtId }` instead of an array of `thoughtId`? Do you need O(1) lookup? 

If so, I'd suggest `{ [thoughtId]: true }` anyway which is a little cleaner and forces downstream code to use `Object.keys` instead of either `Object.keys` and `Object.values`, which could result in mixed use.

src/redux-middleware/pushQueue.ts
```diff
@@ -139,13 +133,54 @@ const flushPushQueue = (): Thunk<Promise<void>> => async (dispatch, getState) =>
...
+      ...acc,
+      ...lexemeInUpdatedState ? { [thoughtId]: { ...lexemeInBatch, contexts: [...lexemeInBatch.contexts, ...updatedLexemeContexts()] } } : {}
+    }
+  }, {})
+
+  // Pull all the pending lexemes that are not available in state yet, reconcile and update the thoughtIndexUpdates. Prevents local and remote lexemes getting overriden by incomplete application state (due to lazy loading).
+  // Related issue: https://github.com/cybersemics/em/issues/1074
+  const lexemeSyncedPushQueue = await Promise.all(pushQueue.map(async batch => {
+    if (!batch.pendingLexemes || Object.keys(batch.pendingLexemes).length === 0) return batch
+    await dispatch(pullLexemes(Object.values(batch.pendingLexemes!)))
```
raineorshine, line 169: I think we need to collate `pendingLexemes` and then make a single `pullLexemes` call, as one `pullLexemes` for every batch is inefficient and defeats the purpose of the `pushQueue`, which is to throttle database access. The logic is already there with `mergeBatch`. You just need to add `pendingLexemes` to `mergeBatch` and then use that instead of pulling every batch separately.

If you need to have access to `lexemeSyncedPushQueue` before `mergeBatch` is called for some reason, then you may need to merge the `pendingLexemes` manually, but I still think we want to find a way to use a single `pullLexemes`.

Let me know if this doesn't make sense or if I'm missing something!

src/redux-middleware/pushQueue.ts
```diff
@@ -139,13 +133,54 @@ const flushPushQueue = (): Thunk<Promise<void>> => async (dispatch, getState) =>
...
-  dispatch(flushMoves())
+  /**
+   * Returns updated thoughtIndexUpdates using the latest state after pulling pending lexemes.
+   */
+  const getSyncedThoughtsUpdate = (batch: PushBatch) => Object.keys(batch.thoughtIndexUpdates).reduce<Index<Lexeme | null>>((acc, thoughtId) => {
+
+    const updatedState = getState()
+
+    // Return early if this particular lexeme was not pending
+    const hadPendingLexeme = batch.pendingLexemes?.[thoughtId]
```
raineorshine, line 144: Idea: Is it possible to calculate `batch.pendingLexemes` here instead of in `updateThoughts`? It isn't possible with `pendingMoves` for example which needs information from `existingThoughtMoves`, but `pendingLexemes` is a simple comparison with the state. One less dependency and one batch variable.

src/redux-middleware/pushQueue.ts
```diff
@@ -139,13 +133,54 @@ const flushPushQueue = (): Thunk<Promise<void>> => async (dispatch, getState) =>
 
   if (pushQueue.length === 0) return Promise.resolve()
 
-  dispatch(flushDeletes())
-  dispatch(flushEdits())
-  dispatch(flushMoves())
+  /**
+   * Returns updated thoughtIndexUpdates using the latest state after pulling pending lexemes.
+   */
+  const getSyncedThoughtsUpdate = (batch: PushBatch) => Object.keys(batch.thoughtIndexUpdates).reduce<Index<Lexeme | null>>((acc, thoughtId) => {
```
raineorshine, line 139: Instead of taking the whole `batch` you can just take `pendingLexemes` and `thoughtIndexUpdates`, which is a narrower dependency.

src/redux-middleware/pushQueue.ts
```diff
@@ -139,13 +133,54 @@ const flushPushQueue = (): Thunk<Promise<void>> => async (dispatch, getState) =>
...
+
+    const updatedState = getState()
+
+    // Return early if this particular lexeme was not pending
+    const hadPendingLexeme = batch.pendingLexemes?.[thoughtId]
+    if (!hadPendingLexeme) return acc
+
+    // Note: Assuming thoughtIndexUpdates entry with null value won't occur, as they will have associated lexeme already pulled.
+    const lexemeInBatch = batch.thoughtIndexUpdates[thoughtId] as Lexeme
```
raineorshine, line 148: Use `batch.thoughtIndexUpdates[thoughtId]!`

src/redux-middleware/pushQueue.ts
```diff
@@ -139,13 +133,54 @@ const flushPushQueue = (): Thunk<Promise<void>> => async (dispatch, getState) =>
...
+   */
+  const getSyncedThoughtsUpdate = (batch: PushBatch) => Object.keys(batch.thoughtIndexUpdates).reduce<Index<Lexeme | null>>((acc, thoughtId) => {
+
+    const updatedState = getState()
+
+    // Return early if this particular lexeme was not pending
+    const hadPendingLexeme = batch.pendingLexemes?.[thoughtId]
+    if (!hadPendingLexeme) return acc
+
+    // Note: Assuming thoughtIndexUpdates entry with null value won't occur, as they will have associated lexeme already pulled.
```
raineorshine, line 147: Could you clarify "they will have associated lexeme already pulled"? Are null updates somehow removed before this point?

raineorshine (changes requested):

src/reducers/updateThoughts.ts
shresthabijay: > Why use `{ [thoughtId]: thoughtId }` instead of an array of `thoughtId`? Do you need O(1) lookup?

Yes. We need O(1) lookup here in `getSyncedThoughtsUpdate`. Else we will have to iterate the array every time to check if a particular lexeme was pending or not.

https://github.com/shresthabijay/em/blob/5ec0459fe264dd8ab5f8683fac5a2976cd12e2db/src/redux-middleware/pushQueue.ts#L144

src/redux-middleware/pushQueue.ts
shresthabijay: I don't think so. `updateThoughts` updates the `thoughts` itself and we need to compare `batch` `thoughtIndexUpdates` against the current `thoughts`, not the updated one. In `pushQueue` we will already have the updated state so comparison will show no differences.

src/redux-middleware/pushQueue.ts
shresthabijay: Sure.

src/redux-middleware/pushQueue.ts
shresthabijay: null updates are caused by thought deletion. For thought deletion, we already need to pull the associated lexemes and parent (from the pullQueue). So after null updates won't run into a state where the associated lexeme is not available. Please correct me if I am wrong to assume this.

src/redux-middleware/pushQueue.ts
shresthabijay: Sure

src/redux-middleware/pushQueue.ts
shresthabijay: Thanks. It makes sense.

src/redux-middleware/pushQueue.ts
raineorshine: An edit consists of a null update + a new or modified update. There will be no conflicting Lexemes from a null update, but that doesn't mean there are no null updates. 

Or do you mean we know there will be no null updates because of the check on L168 of pushQueue?

shresthabijay: `pendingLexemes` are the `thoughtIndexUpdates` that do not have an entry in the `thoughtIndex`. For all null `thoughtIndexUpdates`, there would always be an entry in the `thoughtIndex` as they are pulled by the `pullQueue` before the delete operation.

raineorshine: Is it the conditional above, `if (!hadPendingLexeme) return acc`, that ensures this? If not, I give up 😅

shresthabijay: > Is it the conditional above, `if (!hadPendingLexeme) return acc`, that ensures this? If not, I give up 😅

Yes. It checks if the update was pending or not.

shresthabijay:
- Updated `createTestApp` to flush all the promises before the test starts. Asynchronous calls from the previous `createTestApp` were actually affecting the state of the new tests. It was causing the actual problem in tests.
- Merged `pendingLexemes` and called a single `pullLexeme` call.
- Add tests for this specific issue.

src/reducers/__tests__/deleteEmptyThought.ts
```diff
@@ -248,29 +248,26 @@ describe('mount', () => {
...
-  it('after deleteEmptyThought, caret should move to end of previous thought', () => {
+  it('after deleteEmptyThought, caret should move to end of previous thought', async () => {
     store.dispatch([
       { type: 'newThought', value: 'apple' },
       { type: 'newThought' },
       { type: 'deleteEmptyThought' }
     ])
-    jest.runOnlyPendingTimers()
-    expect(window.getSelection()?.focusOffset).toBe('apple'.length)
   })
```
raineorshine, line 257: There is no longer an assertion in this test

raineorshine (changes requested):
Great! I think it's really coming together now.

src/redux-middleware/__tests__/pushQueue.ts
```diff
@@ -0,0 +1,96 @@
...
+const fakeTimer = testTimer()
+
+const db = dexie as DataProvider
+
+beforeEach(async () => {
+  fakeTimer.useFakeTimer()
+  initialize()
+  await fakeTimer.runAllAsync()
+  fakeTimer.useRealTimer()
+})
```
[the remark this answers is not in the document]
shresthabijay: It doesn't need to test components. But it would be alright to use `createTestApp` instead of using `initialize` and  fake timers here. I will change it.

src/redux-middleware/__tests__/pushQueue.ts
```diff
@@ -0,0 +1,96 @@
...
+
+afterEach(async () => {
+  fakeTimer.useFakeTimer()
+  store.dispatch(clear())
+  await fakeTimer.runAllAsync()
+  fakeTimer.useRealTimer()
+  await db.clearAll()
+})
+
+it('if pending lexemes are found in push batches then pull lexemes, reconcile and update the thoughtIndexUpdates before they are pushed', async () => {
```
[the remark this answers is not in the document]
shresthabijay: Sure

src/redux-middleware/__tests__/pushQueue.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  initialize()
+  await fakeTimer.runAllAsync()
+
+  fakeTimer.useRealTimer()
+
+  // lexeme for 'f; should not be loaded into the state yet.
+  expect(getThoughtSelector(store.getState(), 'f')).toBeFalsy()
+
+  fakeTimer.useFakeTimer()
```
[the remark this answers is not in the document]
shresthabijay: Ahh! I will remove it.

src/redux-middleware/__tests__/pushQueue.ts
```diff
@@ -0,0 +1,96 @@
...
+    oldValue: 'h',
+    newValue: 'f',
+    context: ['g'],
+    path: [{ value: 'g', rank: 0 }, { value: 'h', rank: 0 }] as SimplePath,
+  }))
+  await fakeTimer.runAllAsync()
+
+  fakeTimer.useRealTimer()
+
+  // after existing thought change the pushQueue should pull the pending lexemes and update the thoughtIndexUpdates accordingly.
```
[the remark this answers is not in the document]
shresthabijay: Sure.

src/redux-middleware/__tests__/pushQueue.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  // clear state
+  // call initialize again to reload from db (simulating page refresh)
+
+  fakeTimer.useFakeTimer()
+  store.dispatch(clear())
+  await fakeTimer.runAllAsync()
+
+  initialize()
+  await fakeTimer.runAllAsync()
```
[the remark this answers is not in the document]
shresthabijay: Sure. We can use that test helper in other places like `pullQueue` tests too.

raineorshine (approved):

PR #1193, opened 2021-05-29
Remove static server spin up inside puppeteer env configuration

### LINKED ISSUE #1191: Set up static server for e2e tests outside the puppeteer environment config.
# Issue
`puppeteer-environment` spins up static server at port 3000 for setup and then closes the server on teardown. This will work properly for a single test. But the jest runs tests in parallel. So every tests tries to spin up static server and close it. It causes inconsistent test failures.

# Possible solution

Run static server at port 3000 before e2e tests and then run the tests.

shresthabijay:
fixes #1191

# Changes

- Remove static server spin up from `puppeteer-environment.js`
- Spin up and kill static server  from bash command  in `test:e2e` script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,8 @@
     "lint:package-lock": "node ./scripts/validate-package-lock.js",
     "servebuild": "serve -s build -l 3000",
     "test": "react-scripts test --testPathIgnorePatterns=./src/e2e",
-    "test:e2e": "react-scripts test ./src/e2e"
+    "test:e2e": "react-scripts test ./src/e2e",
+    "test:e2e:prod": "npm run servebuild & react-scripts test ./src/e2e; kill $!"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "lint:package-lock": "node ./scripts/validate-package-lock.js",
     "servebuild": "serve -s build -l 3000",
     "test": "react-scripts test --testPathIgnorePatterns=./src/e2e",
-    "test:e2e": "react-scripts test ./src/e2e"
+    "test:e2e": "npm run servebuild & react-scripts test ./src/e2e; kill $!"
```
raineorshine, line 36: Let's keep `test:e2e` as-is and create a new script `test:e2e:prod`. I'd like to still have an npm script for easily running the e2e tests against a running dev server.

raineorshine (changes requested):

raineorshine (approved):

Issue #1170, opened 2021-05-22
Desktop safari shows Share instead of Download


ruchijn:
### Steps to Reproduce:

Open app in safari. 
Add thoughts and then toggle Export content

<img width="1436" alt="Screen Shot 2021-05-22 at 2 31 28 PM" src="https://user-images.githubusercontent.com/18288628/119237462-eae6e680-bb0a-11eb-9343-e341d9a521ba.png">

### Current Behavior
User is seeing `Share` on desktop safari

### Expected Behavior
User should see `Download` on desktop browsers

raineorshine:
Also we can remove "Copy to clipboard" from mobile since the OS provides that option on the Share popup.

PR #1194, opened 2021-05-29
Fix issue with detecting mobile to render appropriately on export popup.

ruchijn:
Fix: #1170 

Chrome: 

https://user-images.githubusercontent.com/18288628/120073547-ca68e000-c066-11eb-925e-2bdd0e981793.mp4

Safari:

https://user-images.githubusercontent.com/18288628/120074061-0f8e1180-c069-11eb-92dc-3a5a5eec942b.mp4


src/components/ModalExport.tsx
```diff
@@ -61,7 +62,8 @@ const ModalExport = () => {
     filterFunction: !shouldIncludeMetaAttributes ? child => !isFunction(child.value) : undefined
   }).length
 
-  const exportWord = 'share' in navigator ? 'Share' : 'Download'
+  const { isMobile } = DetectUserDevice()
+  const exportWord = isMobile ? 'Share' : 'Download'
```
raineorshine, line 66: I would prefer a simpler touch detection mechanism. You can use [isTouch](https://github.com/cybersemics/em/blob/dev/src/browser.ts#L6) which is already defined.

raineorshine (changes requested):

src/hooks/detectUserDevice.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from 'react'
+
+/** Custom hook detectUserDevice.ts to detect user device. */
+const DetectUserDevice = () => {
```
raineorshine, line 4: For future reference, hooks should always start with `use`, e.g. `useDetectUserDevice`.

src/components/ModalExport.tsx
ruchijn: I saw existing `isTouch`. But won't it also be true for touch screen laptops? 🤔

raineorshine: Yes, but I don't want to risk a false negative on mobile. It's a reasonable approach though.

Issue #1158, opened 2021-05-19
Preserve (limited) <style> attributes


raineorshine:
> we properly import text that is copied from the IOS notes app. But there is one issue about the text that contains formatting copied from IOS notes. IOS notes app doesn't add `<b>, <i>` tags for text formatting. IOS Notes app uses the style attribute to style texts. Here is an example that contains bold text that is copied from the app.
> 
> ```
> <meta charset="UTF-8"><p class="p1"
>                          style="margin: 0px; font-style: normal; font-variant-caps: normal; font-weight: normal; font-stretch: normal; font-size: 14px; line-height: normal; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-tap-highlight-color: rgba(26, 26, 26, 0.3); -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;">
>     <span class="s1" style="font-weight: bold; font-style: normal; font-size: 14px;">A</span></p><p class="p1"
>                                                                                                     style="margin: 0px; font-style: normal; font-variant-caps: normal; font-weight: normal; font-stretch: normal; font-size: 14px; line-height: normal; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-tap-highlight-color: rgba(26, 26, 26, 0.3); -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;">
>     <span class="s1" style="font-weight: bold; font-style: normal; font-size: 14px;"><span class="Apple-converted-space"> </span>B</span></p>
> ```
> 
> We import the text properly for the above input but we are losing formatting. I will look into this problem on another issue because before fixing this issue we need to discuss about this.

## Possible Solutions

1. Convert `<span>` style attributes to `<b>` and `<i>`
2. Render inline `<span>` in Editable

## Discussion

I prefer solution #<b></b>2 because a) it does not require any special conversions, just making the HTML stripping a bit smarter, and b) it allows any style to be applied inline not just `<b>` and `<i>`.

onurpolattimur:
I prefer solution 2 too.

[...]

PR #1195, opened 2021-05-29
Preserve style attributes of span while importing text

onurpolattimur:
fixes: #1158

### DIFF
--- a/src/util/__tests__/stripStyleAttribute.ts
+++ b/src/util/__tests__/stripStyleAttribute.ts
@@ -1,0 +1,29 @@
+import stripStyleAttribute from '../stripStyleAttribute'
+
+it('only allowed properties', () => {
+  expect(stripStyleAttribute('color: red; font-weight: bold')).toBe('color: red;font-weight: bold;')
+})
+
+it('allowed properties and font-size', () => {
+  expect(stripStyleAttribute('color: red; font-weight: bold; font-size: 14px')).toBe('color: red;font-weight: bold;')
+})
+
+it('allowed properties and font-weight: normal', () => {
+  expect(stripStyleAttribute('color: red; font-weight: bold; font-style: normal')).toBe('color: red;font-weight: bold;')
+})
+
+it('with black color and no background, should not include color property', () => {
+  expect(stripStyleAttribute('color: black;')).toBe('')
+})
+
+it('with white color and no background, should not include color property', () => {
+  expect(stripStyleAttribute('color: white;')).toBe('')
+})
+
+it('with black color and yellow background', () => {
+  expect(stripStyleAttribute('color: black; background: yellow')).toBe('color: black;background: yellow;')
+})
+
+it('with white background and no color, should not include background property', () => {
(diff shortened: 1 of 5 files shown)


src/util/__tests__/stripStyleAttribute.ts
```diff
@@ -0,0 +1,29 @@
+import stripStyleAttribute from '../stripStyleAttribute'
+
+it('only allowed properties', () => {
+  expect(stripStyleAttribute('color: red; font-weight: bold')).toBe('color: red;font-weight: bold;')
+})
```
raineorshine, line 5: This test demonstrates that `stripStyleAttribute` does not strip `color` and `font-weight`, but it does not test that it strips other properties. Include a non-whitelisted property that is stripped for a complete test.

src/util/stripStyleAttribute.ts
```diff
@@ -0,0 +1,70 @@
+interface StyleProperty {
+  name: string,
+  value: string,
+}
+
+const blackColors = ['#000000', '#000', 'rgb(0,0,0)', 'rgba(0,0,0,1)', 'black']
```
raineorshine, line 6: Also `rgb(0, 0, 0) `

src/util/stripStyleAttribute.ts
```diff
@@ -0,0 +1,70 @@
...
+    })
+}
+
+/** Parse and strip style attribute, returns stripped style attribute string. */
+const stripStyleAttribute = (style: string) => {
+  const styles = parseStyleString(style)
+  return styles.reduce((acc, property) => {
+    const styleProperty = allowedStyleProperties.find(allowedStyleProperty => property.name.startsWith(allowedStyleProperty.property))
+    if (styleProperty && (!styleProperty.test || styleProperty.test(property, styles))) {
+      acc += `${property.name}: ${property.value};`
```
raineorshine, line 64: Please avoid mutations when possible

src/util/stripStyleAttribute.ts
```diff
@@ -0,0 +1,70 @@
...
+      }
+    })
+}
+
+/** Parse and strip style attribute, returns stripped style attribute string. */
+const stripStyleAttribute = (style: string) => {
+  const styles = parseStyleString(style)
+  return styles.reduce((acc, property) => {
+    const styleProperty = allowedStyleProperties.find(allowedStyleProperty => property.name.startsWith(allowedStyleProperty.property))
+    if (styleProperty && (!styleProperty.test || styleProperty.test(property, styles))) {
```
raineorshine, line 63: You'll want to `toLowerCase` the property.

raineorshine (changes requested):
Thank you! Good structure, just a few small suggestions.

src/util/__tests__/stripStyleAttribute.ts
onurpolattimur: There is a test like you said. The purpose of this test is to check the output for the allowed properties only.

raineorshine (approved):

Issue #1131, opened 2021-05-09
Import Error: Cannot read property 'trim' of undefined


raineorshine:
## Steps to Reproduce

Paste the following:

```
- a
- b
- <span>c
- d
```
or:

```
- a
 - b
 - b
 - <em>c</em>
```

or:

```
- a
- b
- c<br>
```

## Current Behavior

![image](https://user-images.githubusercontent.com/750276/117586059-0e556e80-b0d3-11eb-9f57-35ccafaa837c.png)

## Expected Behavior

- a
- b
- c
- d

shresthabijay:
@raineorshine Is this issue fixed ?

raineorshine:
No, the referenced commits are workarounds for a script that diffs database snapshots.

PR #1197, opened 2021-05-29
Fix Import Error - 'Cannot read property 'trim' of undefined'  

onurpolattimur:
fixes: #1131 

___

There was more than one issue.

- When text contains nonclosed html tags, it was not converting nodes to blocks properly.
- `<em>` tag was not being considered a formatting tag.
- It was returning an array of blocks even it has only one child while handling br.


raineorshine:
> * `<em>` tag was not being considered a formatting tag.

Ha... how ironic 😂

src/util/closeNonClosedHtmlTags.ts
```diff
@@ -0,0 +1,8 @@
+import truncate from 'truncate-html'
+
+/** Close non closed html tags. */
+const closeNonClosedHtmlTags = (html :string) => {
+  return truncate(html, html.length)
+}
+
+export default closeNonClosedHtmlTags
```
raineorshine, line 8: Since `closeNonClosedHtmlTags` is only used in one file, it would be better to define it internal to that module in order to avoid increasing the surface area of the util api.

The separate file does decouple `convertHMTLtoJSON` from the `truncate-html` dependency, but it's not a major interface with large implementation details to hide, so I think the simpler architectural is a better fit.

raineorshine (changes requested):
Looks great! Thanks for tightening up our import feature!

raineorshine (approved):

Issue #1202, opened 2021-05-31
setSelection fails to focus on proper offset.


shresthabijay:
# Steps to reproduce

```
- guardians <strong>of the </strong><strong>galaxy </strong>
- volume 1
```

- Paste the following
- Set cursor at `volume 1` at offset 0.
- Hit backspace.

# Current behavior

Two thoughts are merged but the cursor position is not after `galaxy` but is after `guardians`.

# Expected behavior

Two thoughts should merge and the cursor position should be at the point of merge i.e after `galaxy`.

# Cause of issue

The current `setSelection` assumes that selection should always be in the  `firstChild` of the editable node. But this assumption is wrong. It won't work properly as soon as there are other nodes introduced from format tags, just like in the example.

To understand the issue, we need to understand how the focus selection and range work.  At first, the editable node can contain more than one text node as `childNodes`. Let's break down the example.

For `guardians <strong>of the </strong><strong>galaxy </strong>`, the node structure would be something like this.

- editable node
  - text (guardians)
  - strong
    - text (of the)
  - strong
    - text (galaxy) 
 
Currently, the setSelection is trying to set focus on text (guardians) at offset which is the length of the full text. But it doesn't only get focused at the end of the `guardians`.

shresthabijay:
# Possible solution.

The solution is to introduce a path to the actual node which needs to be focused. Let's say `focusNodePath`

- editable node
  - text (guardians)
  - strong
    - text (of the)
  - strong
    - text (galaxy) 
    
The path is an array of numbers, each number representing the index of the node respective to the parent node starting from root editable node.
  
If we want to set selection at the beginning of the `galaxy`, the `focusNodePath` must be `[3, 0]` which selects the node text(galaxy). And the offset must be 0.
  
Similarly, if we need to set selection at the end of the `of the` text, the `focusNodePath` can be `[2, 0]` with offset of `7`. It can also be done with `focusNodePath` set to `null` (means it will select `editable node`) and with an offset of 3. This will keep caret at the beginning second `strong` tag which is also the end of the `of the` text.

raineorshine:
@shresthabijay The application is crashing as soon as I paste. Is that not happening for you? On Chrome?

shresthabijay:
> @shresthabijay The application is crashing as soon as I paste. Is that not happening for you? On Chrome?
[...]

PR #1203, opened 2021-05-31
Split thought using browser range api

### LINKED ISSUE #1184: Thought with formatting split false positive
## Steps to Reproduce

``` 
a <strong>b</strong> c
``` 

Place the caret at the end of the thought and hit enter. 

## Current Behavior

- a
- <strong>b</strong> c

## Expected Behavior

- a <strong>b</strong> c
- 

Should perform normal `newThought` and not split the sentence since the caret is at the end of the thought.

shresthabijay:
fixes #1184 #1202

# Cause of issue for #1184

Since the actual value of the thought is an HTML value, using the selection offset to split the thought doesn't work as intended.

For example:

`a <strong>b</strong>c`

Keeping the cursor at the end of the `c` will have a selection offset as 3. But splitting the thought value at  3 doesn't make any sense.

Note: Please refer to #1202 for the detailed information of its cause.
# Solution

For #1184:

I managed to find a solution that works. It works as follows:
- Using the current selection range, `firstChild` and `lastChild` of the editable node we can create two different `Range` that exactly splits thought at the cursor selection.
- Add two dummy divs and append the cloned contents of those range. 
- Extract the `innerHTML` of the dummy divs.

This way we can use browser to split the thoughts exactly at the half.

For #1202

~~I have added two fields to the `state` called `focusNodePath` and `cusorFocusPosition` that will help set the offset at any position of the editable we want. I have explained it in the issues itself https://github.com/cybersemics/em/issues/1202#issuecomment-851175690~~ 

~~`focusNodePath` will be the path to the focus which we want to set selection. Using `focusNodePath`  and offset together we can keep caret on exact places we want. `cusorFocusPosition` will be used to override the `focusNodePath` and `offset` to set cursor at the very beginning or the very end of the editable node. Previously `end` didn't work well when there were multiple `childNodes` inside editable node.~~

The previous implementation was working but it added complexity with `focusNodePath` and `cursorFocusPosition`. The idea was to get correct path to the focus node and add it to `focusNodePath`, so that we can retrieve the actual focus node on `Editable` to set focus on.

The new approach simplifies things by removing both  `focusNodePath` and `cursorFocusPosition` and using current `cursorOffset` as a relative offset.

So `cursorOffset` represents the offset relative to the plain text content of the thought value without any HTML tags.

For example: 
Thought value: `This is a <b>basic</b> example`
Text content: `This is a basic example`

Let's say we want the caret at the end of the word `basic`. Previously we included the HTML tags in the offset. In that case the offset would have been `18`. But the relative offset would be with respect to the plain text content. So the relative offset would now be `15`.

I have created a function called `getTextContentFromHTML` that will take the HTML thought value and return the plain text content stripping all the tags. It accurately represents how the user sees the text content in the browser as it uses browser API to get the text conent.

A utility function called `getCaretPositionDetails` will take the relative offset and the editable node and return the exact focus node and offset relative to that focus node.

So to set the caret position at the desired relative offset of an editable we need to do following.
- Get text content from the thought value using `getTextContentFromHTML`.
- Get `focusNode` and `offset` using `getCaretPositionDetails` and pass it to `setSelection`.

Here's an example how we could merge the thoughts and keep the caret at the merged point.

```javascript

const left = 'This is a <b>basic</b>'
const right = 'example'

const newValue = left + right

// Assuming an editable is rendered with merged thought value.

const textContent = getTextContentFromHTML(left)
const { focusNode, offset } = getCaretPositionDetails(editableNode, left.length)

setSelection(focusNode, { offset })

/*
{
  focusNode: #text ('basic'),
  offset: 5
}
*/
```


shresthabijay:
@raineorshine I have made significant changes to the split logic and cursor selection logic. Please review my approach for solving #1202 and #1184. I am planning to solve these two issues in this single PR as they are closely related to each other.

I have made changes to `setCursor` and only changed few places that uses `setCursor` to demonstrate how they work. So tests may fails for now. I will make necessary changes after you approve the approach. Thanks!

src/shortcuts/newThoughtOrOutdent.tsx
```diff
@@ -35,12 +35,12 @@ const exec: Shortcut['exec'] = (dispatch, getState, e, { type }: { type: string
     const target = e.target as HTMLElement
 
     // Note: e.target should be a HTMLElement and a content editable node
-    const offset = cursor && isTargetHTMLElement && target.hasAttribute('contenteditable')
-      ? getOffsetWithinContent(target)
-      : 0
+    const isTargetAnEditable = isTargetHTMLElement && target.hasAttribute('contenteditable')
+
+    const splitResult = cursor && isTargetAnEditable ? splitAtSelection(target) : null
```
raineorshine, line 40: Just for my understanding, does `splitResult.previous + splitResult.next` always equal `target.innerHTML`?

raineorshine (changes requested):
Thanks for getting to the bottom of this. It makes perfect sense that a single offset would not handle nested formatting nodes. I think the overall approach should work. I have a few naming suggestions and questions for clarification.

src/util/initialState.ts
```diff
@@ -60,6 +61,7 @@ export interface State {
   expandHoverBottomPaths: Index<Path>,
   expandHoverTopPath?: Path | null,
   expandedContextThought?: Path,
+  focusNodePath?: number[] | null,
```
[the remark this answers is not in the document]
shresthabijay: > * Do these both represent the current caret position? How are they kept up-to-date when the caret moves?

Sometimes we want to set the selection at the specific place in the editable. So we need `focusNodePath` and `offset` to do that.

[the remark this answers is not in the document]
shresthabijay: > * Can `cursorFocusPosition` be derived from `focusNodePath`, or is it independent?

In many cases we just want to set the focus at the end or the start of the editable. That doesn't require `focusNodePath` and `offset`. So `cursorFocusPosition` is basically an enum with values `start` and `end`. If it is provided it will just override the  `focusNodePath` and `offset` and set the focus at the start or the end of the editable. Similar to `end` parameter in the previous implementation of `setSelection`.

src/shortcuts/newThoughtOrOutdent.tsx
shresthabijay: Not always. So if we have following value at the target.innerHTML `<b> bold and <i> italic <i> </b>`.

So splitting with cursor selection right after the word `and`, we will get following result.

`splitResult.previous` will be `<b> bold and </b>`
`splitResult.next` will be `<b><i> italic </i></b>`

We are basically using browser to do the splitting 😄

src/reducers/splitThought.ts
```diff
@@ -7,15 +7,19 @@ import { editableRender, existingThoughtChange, existingThoughtMove, newThought,
 import { State } from '../util/initialState'
 import { Path } from '../types'
 
+interface SplitResult {
+  previous: string,
+  next: string,
```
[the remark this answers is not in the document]
shresthabijay: Sure. I will change this.

src/reducers/deleteEmptyThought.ts
```diff
@@ -1,10 +1,20 @@
...
-import { head, headRank, headValue, isDivider, parentOf, pathToContext, reducerFlow } from '../util'
+import { hashContext, head, headRank, headValue, isDivider, parentOf, pathToContext, reducerFlow } from '../util'
 import { getNextRank, getChildren, getChildrenRanked, isContextViewActive, prevSibling, simplifyPath, rootedParentOf } from '../selectors'
 import { deleteThought, existingThoughtChange, existingThoughtDelete, existingThoughtMove, setCursor } from '../reducers'
 import { State } from '../util/initialState'
 import { SimplePath } from '../types'
 import archiveThought from './archiveThought'
+import { getNodePathToParentEditable } from '../util/getNodePathFromEditable'
+
+/** */
```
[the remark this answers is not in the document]
shresthabijay: Will do. Thanks!

shresthabijay:
Just for the information, I am keeping this here. I think it is also possible to use relative offset (taking editable value as a textContent, not as HTML value) to set the selection exactly at the right position within the nested nodes.

We take the offset and iterate the editable node children using depth-first search and keep adding the text nodes content length and return the node and its path where the relative offset is equal or greater to the text nodes content sum.

This may be useful when there is no way to calculate `focusNodePath` directly from another editable reference.

src/util/initialState.ts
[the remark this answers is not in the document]
raineorshine: Okay, but why is this part of the global state? The way you described them they seem more like arguments to a reducer, i.e. a specific action instead of a persistent value.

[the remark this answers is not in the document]
shresthabijay: > Okay, but why is this part of the global state? The way you described them they seem more like arguments to a reducer, i.e. a specific action instead of a persistent value. 

You are correct we can just use ‘focusNodePath‘ and ‘offset‘ to keep selection at end or start too. I will remove this from.the global state. Thanks!

raineorshine:
Okay, I think we have a way to move forward! Let me know if I missed anything before your next revision.

src/util/splitAtSelection.ts
```diff
@@ -0,0 +1,44 @@
...
+export const splitAtSelection = (el: HTMLElement) => {
+  const sel = document.getSelection()
+  if (!sel || sel.rangeCount === 0) return null
+
+  const splitNodesResult = splitNode(sel, el)
+
+  if (!splitNodesResult) return null
+
+  const previousDiv = document.createElement('div')
+  const nextDiv = document.createElement('div')
```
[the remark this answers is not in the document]
shresthabijay: I don't think that would be necessary here. Creating just an element without appending it to any existing node in the DOM would not cause any reflow. Also we need to create `div` to access `innerHTML`  which is not provided by `DocumentFragment`.

shresthabijay:
@raineorshine I have made the changes. I have updated the approach and added required tests for it. Please review it!

src/util/splitAtSelection.ts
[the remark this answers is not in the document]
raineorshine: Ah, I see! I forgot you weren't adding into to the DOM. Thanks.

src/reducers/deleteThought.ts
```diff
@@ -107,7 +110,7 @@ const deleteThought = (state: State, payload: { path?: Path }) => {
           : parentOf(path).concat(next() as Child)
         ), { offset: 0 }] :
         // Case III: delete last thought in context; set cursor on context
-        thoughts.length > 1 ? [rootedParentOf(state, path), { offset: head(context).length }]
+        thoughts.length > 1 ? [rootedParentOf(state, path), { offset: (getTextContentFromHTML(contextHeadValue) || contextHeadValue).length }]
```
raineorshine, line 113: I'm seeing this guard expression used a lot: `(getTextContentFromHTML(contextHeadValue) || contextHeadValue)`

However, is there ever a case where `getTextContextFromHTML` could return `null`? We are setting the `innerHTML` of the dummy div to a string, so it has to have `textContext`. Maybe you can have it return `return dummyDiv.textContent!` and avoid the guards upstream.

src/util/getCaretPositionDetails.ts
```diff
@@ -0,0 +1,60 @@
...
+  // case where the caret should be positioned at the end of the editable.
+  if (editableNode.textContent && relativeOffset >= editableNode.textContent.length) {
+    return {
+      focusNode: editableNode,
+      offset: editableNode.childNodes.length
+    }
+  }
+
+  return recursiveFocusNodeFinder(editableNode, relativeOffset)
+}
```
raineorshine, line 60: I think the `editableNode` argument can be named `node`, as it is not specific to Editables, unless I am mistaken.

src/util/getCaretPositionDetails.ts
```diff
@@ -0,0 +1,60 @@
+interface CaretPositionDetails {
+  focusNode: Node | null,
+  offset: number,
+}
+
+/**
+ * Recursively iterates the nodes children and returns focusNode and offset where the relative offset ends.
+ */
+export const recursiveFocusNodeFinder = (node: Node, relativeOffset: number): CaretPositionDetails | null => {
```
raineorshine, line 9: `recursiveFocusNodeFinder` shouldn't be exported.

src/util/splitAtSelection.ts
```diff
@@ -0,0 +1,44 @@
...
+  previousRange.setEnd(range.startContainer, range.startOffset)
+
+  const nextRange = document.createRange()
+  nextRange.setStart(range.endContainer, range.endOffset)
+  nextRange.setEndAfter(lastChild)
+
+  return {
+    previous: previousRange,
+    next: nextRange,
+  }
```
raineorshine, line 21: Let's use left and right instead of previous and next for consistency.

raineorshine (changes requested):
I like the new approach!

raineorshine:
@onurpolattimur Could you look into this? It should be getting the username from the repo secret.

```
Run browserstack/github-actions/setup-env@master
Error: Action Failed: Error: Action input failed for reason: Input required and not supplied: username
```

src/util/getCaretPositionDetails.ts
shresthabijay: Yes!

src/util/getCaretPositionDetails.ts
shresthabijay: Will fix this. Thanks!

src/util/splitAtSelection.ts
shresthabijay: I missed it here. I will change it.

src/reducers/deleteThought.ts
shresthabijay: Sure. That makes sense.

shresthabijay:

raineorshine (approved):

Issue #1186, opened 2021-05-25
Share icon


raineorshine:
Use a standard share icon instead of a download icon for `exportContext` when on mobile. 

I recommend finding one on the Noun Project.

PR #1206, opened 2021-05-31
Replace download icon with share when on mobile

ruchijn:
Fixes: #1186 

share_download


raineorshine (changes requested):
Thanks! Let's actually use this icon: https://thenounproject.com/term/share/1058861. I should have looked that up when I created the issue, so sorry for the extra work.

Also, can you add the icon credits to `ModalHelp`? You should see how the other icon credits are listed.

raineorshine (approved):

PR #1215, opened 2021-06-04
Remove unallowed tags from html before converting html to json

### LINKED ISSUE #1205: importText crashes on paste
# Steps to reproduce

```
- guardians <strong>of the </strong><strong>galaxy </strong>
- volume 1
```

- Paste the following.

# Current behavior

Crashes the application.

# Expected behavior

Should work without crashing.

shresthabijay:
fixes #1205 

@raineorshine  This solution fixes the problem. But `convertHTMLtoJSON` may still return `(Block | Block[])[]` depending on the input, which is incompatible with `importJSON`. I suggest that if the return of the `convertHTMLtoJSON` is an array, we always flat it at one level.


shresthabijay:

raineorshine:
> @raineorshine This solution fixes the problem. But `convertHTMLtoJSON` may still return `(Block | Block[])[]` depending on the input, which is incompatible with `importJSON`. I suggest that if the return of the `convertHTMLtoJSON` is an array, we always flat it at one level.

Okay, that sounds reasonable. Do you want to do that in this PR or separately?

shresthabijay:
@raineorshine Can you explain what this part of the code does ?https://github.com/cybersemics/em/blob/8b928779270be3f6eb662912b58aa32a58f505d1/src/util/convertHTMLtoJSON.ts#L201-L208

raineorshine:
I'm actually not sure. It's not obvious what would produce that scenario, and I didn't originally write that code.

I would recommend removing it and seeing which tests fail and how they fail to determine its purpose.

shresthabijay:
> I'm actually not sure. It's not obvious what would produce that scenario, and I didn't originally write that code.
> 
> I would recommend removing it and seeing which tests fail to determine its purpose.

I was just making sure before I make the changes. I will look into it.

shresthabijay:
> Looks good!
> 
> Make sure to use npm v7 for installs now that we are using the newer lockfile.

Sure

Issue #1150, opened 2021-05-16
Mobile: Do not outdent during clearThought


raineorshine:
Mobile only

https://user-images.githubusercontent.com/750276/118383704-60fcc200-b5ce-11eb-8ec0-c380f2661b35.mov

## Steps to Reproduce

1. Create thought `A`
2. Create subthought `B`
2. Activate `clearThought` by swiping right-left. 
3. Hit Backspace

## Current Behavior

`B` is outdented

## Expected Behavior

`B` should be deleted. 

## Notes

In `newThoughtOrOutdent`, activate `outdent` only if the placeholder attribute of the Editable does not equal the innerHTML stripped of tags. 

https://github.com/cybersemics/em/blob/cdeabad628dda0219ff34a751d5c0735cf55a1c7/src/shortcuts/newThoughtOrOutdent.tsx#L27-L30

Factor out the code that strips the innerHTML of tags into a helper function `toPlaceholder: string → string` since `newThoughtOrOutdent` must check for the exact value set at the placeholder in `clearThought`:

https://github.com/cybersemics/em/blob/cdeabad628dda0219ff34a751d5c0735cf55a1c7/src/shortcuts/clearThought.ts#L15

ruchijn:
I was looking into this. I am not sure if I am doing it wrong, but it doesn't seem to be going to https://github.com/cybersemics/em/blob/cdeabad628dda0219ff34a751d5c0735cf55a1c7/src/shortcuts/newThoughtOrOutdent.tsx#L27-L30. It seems to be dispatching `outdent` from https://github.com/cybersemics/em/blob/cdeabad628dda0219ff34a751d5c0735cf55a1c7/src/shortcuts/deleteEmptyThoughtOrOutdent.tsx#L111. 

Just wanted to confirm if I am right track before I go in deeper.

raineorshine:
Good catch! You're right, it's in `deleteEmptyThoughtOrOutdent`, not `newThoughtOrOutdent`.

PR #1216, opened 2021-06-04
Prevent outdent when clearThought. Instead delete the thought

ruchijn:
Fix: #1150 

clear_thought_fix

### DIFF
--- a/src/shortcuts/__tests__/deleteEmptyThoughtOrOutdent.ts
+++ b/src/shortcuts/__tests__/deleteEmptyThoughtOrOutdent.ts
@@ -26,7 +26,6 @@ it('do nothing when there is no cursor', () => {
 })
 
 it('outdent on pressing backspace at the beginning of the thought', () => {
-
   const store = createTestStore()
 
   // import thoughts
@@ -74,6 +73,36 @@ it('do not outdent thought with siblings', () => {
   - a
     - b
       - cd`
+
+  expect(exported).toEqual(expectedOutput)
+})
+
+it('delete the thought when user triggered clearThought and then hit back', () => {
+  const container = document.createElement('div')
+  container.setAttribute('class', 'editing')
+  container.innerHTML = `<div class="preventAutoscroll editable" placeholder="c" contenteditable="true"></div>`
+  document.body.appendChild(container)
+
+  const store = createTestStore()
+
+  // import thoughts
+  store.dispatch(importText({
+    path: HOME_PATH,
+    text: `
+      - a
+        - b
+          - c
+  ` }))
+
+  store.dispatch(setCursor({ path: rankThoughtsFirstMatch(store.getState(), ['a', 'b', 'c']) }))
+
+  executeShortcut(deleteEmptyThoughtOrOutdent, { store })
+
+  const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
+
+  const expectedOutput = `- ${HOME_TOKEN}
+  - a
+    - b`
 
   expect(exported).toEqual(expectedOutput)
 })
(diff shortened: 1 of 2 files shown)


src/shortcuts/deleteEmptyThoughtOrOutdent.tsx
```diff
@@ -71,11 +71,17 @@ const canExecuteOutdent = (state: State) => {
   if (!cursor || !selection) return false
 
   const offset = selection.focusOffset
+  const focusedEditingValue = state.editingValue || ''
+  const contextOfCursor = pathToContext(cursor)
+  const parentOfCursor = parentOf(contextOfCursor)
+  const filteredList = contextOfCursor.filter(el => !parentOfCursor.includes(el))
```
raineorshine, line 77: This `filteredList` doesn't make sense so me. Paths and contexts are contiguous, e.g. a → b → c. Filtering it breaks the invariance of contiguity. Maybe you can explain you intention here and I can help point you in the right direction.

raineorshine (changes requested):
This fixes the issue but creates a false negative, i.e. a case that should be outdented but now is not.

## Steps to Reproduce

- a
  - b

Set the caret at the beginning of `a/b` and hit delete.

## Current Behavior

967ec0fa0

Nothing happens.

## Expected Behavior

`a/b` is outdented.

src/shortcuts/__tests__/deleteEmptyThoughtOrOutdent.ts
```diff
@@ -77,3 +78,34 @@ it('do not outdent thought with siblings', () => {
...
+    const initialExportedData = exportContext(store.getState(), ['a'], 'text/plain')
+    expect(initialExportedData).toBe(bothThoughtsExported)
+
+    executeShortcut(clearThoughtShortcut, { store })
+
+    executeShortcut(deleteEmptyThoughtOrOutdent, { store })
+
+    /** This ensures that the thought b doesn't exist now. */
+    const exported = exportContext(store.getState(), ['a'], 'text/plain')
+    expect(exported).toBe(`a`)
```
raineorshine, line 109: Actually, this ensures that thought `b` is not within the context `a`. If thought `b` is deleted, it is not in the context `a`. Unfortunately, if thought `b` is outdented (the fail case), thought `b` is *also* not in context `a`. It has just been moved to the root. So this doesn't actually fail when the bug is present.

If you export the entire thoughtspace with `exportContext(store.getState(), [HOME_TOKEN], 'text/plain')`, then it will export all thoughts, and you can be sure that `b` has been deleted.

raineorshine (changes requested):

src/shortcuts/__tests__/deleteEmptyThoughtOrOutdent.ts
```diff
@@ -77,3 +78,34 @@ it('do not outdent thought with siblings', () => {
...
+    store.dispatch([
+      newThought({ value: 'a' }),
+      newThought({ value: 'b', insertNewSubthought: true }),
+      setCursor({ path: rankThoughtsFirstMatch(store.getState(), ['a', 'b']) })
+    ])
+
+    const bothThoughtsExported = `- a
+  - b`
+
+    /** This ensures that the thought b exists so we can confirm later that it is deleted. */
```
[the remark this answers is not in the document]
ruchijn: oh ok. Changed

src/shortcuts/__tests__/deleteEmptyThoughtOrOutdent.ts
ruchijn: I changed the context to `HOME_TOKEN` for both the tests before/after.

raineorshine (approved):

PR #1220, opened 2021-06-05
Refactor expandThoughts and allow supressExpansion to afftect only non pinned expansion path.

### LINKED ISSUE #1212: cursorPrev to pinned thought should not collapse and expand
## Steps to Reproduce

```js
- a
  - b
    - =pin
      - true
    - c
  - d
```

1. Set the cursor on `a/d`
2. Activate `cursorPrev` (<kbd>cmd</kbd> + <kbd>up</kbd>)

## Current Behavior

`a/b` collapses for a moment and then re-expands

## Expected Behavior

When `cursorPrev` moves to a pinned thought, it should not collapse and expand since the thought is already expanded.

The same should be done for `cursorNext` in the other direction.

Screen Recording 2021-06-02 at 6 25 13 AM mov

shresthabijay:
fixes #1212 

# Changes
- Make `suppressExpansion` prevent the expansion of the given `path` if it is a non-pinned thought.
- Refactored `expandThoughts` to simplify it and prevent unnecessary recursions.
- Add tests for `expandThoughts` for the cases of `suppressExpansion`.

### DIFF
--- a/src/reducers/setCursor.ts
+++ b/src/reducers/setCursor.ts
@@ -5,6 +5,7 @@ import { equalPath, equalThoughtRanked, hashContext, headValue, isDescendant, pa
 import { render, settings } from '../reducers'
 import { State } from '../util/initialState'
 import { Index, Path, SimplePath, TutorialChoice } from '../types'
+import globals from '../globals'
 
 /**
  * Sets the cursor on a thought.
@@ -71,7 +72,7 @@ const setCursor = (state: State, {
   //   }
   // })
 
-  const expanded = expandThoughts({ ...state, contextViews: newContextViews }, thoughtsResolved)
+  const expanded = globals.suppressExpansion ? {} : expandThoughts({ ...state, contextViews: newContextViews }, thoughtsResolved)
 
   const tutorialChoice = +(getSetting(state, 'Tutorial Choice') || 0) as TutorialChoice
   const tutorialStep = +(getSetting(state, 'Tutorial Step') || 1)
--- a/src/selectors/expandThoughts.ts
+++ b/src/selectors/expandThoughts.ts
@@ -1,9 +1,8 @@
-import globals from '../globals'
-import { EXPAND_THOUGHT_CHAR, HOME_PATH, HOME_TOKEN, MAX_EXPAND_DEPTH } from '../constants'
-import { attribute, attributeEquals, getAllChildren, getChildPath, getContexts, isContextViewActive, rootedParentOf, simplifyPath } from '../selectors'
+import { EXPAND_THOUGHT_CHAR, HOME_PATH, HOME_TOKEN, MAX_DISTANCE_FROM_CURSOR, MAX_EXPAND_DEPTH } from '../constants'
+import { attribute, attributeEquals, getAllChildren, getChildPath, getContexts, isContextViewActive, simplifyPath } from '../selectors'
 import { Child, Context, Index, Path, ThoughtContext } from '../types'
 import { State } from '../util/initialState'
-import { equalThoughtRanked, hashContext, head, headValue, isFunction, isURL, keyValueBy, parentOf, pathToContext, publishMode, strip, unroot } from '../util'
+import { equalArrays, equalThoughtRanked, hashContext, head, headValue, isDescendant, isFunction, isURL, keyValueBy, parentOf, pathToContext, publishMode, strip, unroot } from '../util'
 
 /** Get the value of the Child | Th oughtContext. */
 const childValue = (child: Child | ThoughtContext, showContexts: boolean) => showContexts
@@ -37,11 +36,12 @@ const publishPinChildren = (state: State, context: Context) => publishMode() && 
   'true'
 )
 
-function expandThoughts(state: State, path: Path | null, options?: { depth?: number }): Index<Path>
-function expandThoughts<B extends boolean>(state: State, path: Path | null, options?: { depth?: number, returnContexts?: B }): Index<B extends true ? Context : Path>
+function expandThoughts(state: State, path: Path | null): Index<Path>
+function expandThoughts<B extends boolean>(state: State, path: Path | null, options?: { returnContexts?: B }): Index<B extends true ? Context : Path>
 
-/** Returns an expansion map marking all contexts that should be expanded.
+/** Returns an expansion map marking all contexts that should be expanded when for the given path.
  *
+ * @param suppressExpansion - Prevents expansion of non pinned expansion path.
  * @example {
  *   [hashContext(context)]: true,
  *   [hashContext([...context, childA])]: true,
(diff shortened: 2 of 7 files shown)


shresthabijay:

src/selectors/expandThoughts.ts
```diff
@@ -50,23 +50,41 @@ function expandThoughts<B extends boolean>(state: State, path: Path | null, opti
...
+    return ancestorPath && ancestorPath.length !== 0 ? ancestorPath : HOME_PATH
+  })()
+
+  return expandThoughtsRecursive(state, path || HOME_PATH, startingPath, options)
+}
+
+/**
+ * Recursively generate expansion map based by checking if the children of the current path should expand based on the given expansion path.
+ *
+ * @param expansionPath - The path based on which expansion happens.
```
raineorshine, line 67: Let's call this `startingPath` or `expansionBasePath` to be a bit more specific.

src/selectors/expandThoughts.ts
```diff
@@ -85,13 +103,30 @@ function expandThoughts(state: State, path: Path | null, { depth = 0, returnCont
...
+
+      /** Check of the path is the ancestor of the expansion path. */
+      const isAncestor = () => isDescendant(childContext, expansionPathContext)
+
+      /** Check if the path is equal to the expansion path. */
+      const isExpansionPath = () => equalArrays(childContext, expansionPathContext)
+
+      const shouldHide = isFunction(value) && !isAncestor() && !isExpansionPath()
+      return !shouldHide
+    })
```
raineorshine, line 125: This is a lot of work within a filter predicate. For every child in the expansion tree it is calling multiple functions of `O(path.length)`: `pathToContext`, `isDescendant`, `equalArrays`.

Is there a way to reuse work from previous recursive calls and improve the running time complexity of this? For example, if a thought A was not a descendant of the expansion base path in its iteration, then we know that the child of thought A is not a descendant of the expansion base path without having to call `isDescendant`.

src/selectors/expandThoughts.ts
```diff
@@ -100,22 +135,31 @@ function expandThoughts(state: State, path: Path | null, { depth = 0, returnCont
...
+    : visibleChildren.filter(child => {
       const value = strip(childValue(child, showContexts))
-      return value[value.length - 1] === EXPAND_THOUGHT_CHAR || isPinned(child) === 'true'
+
+      const childPath = unroot([...path || [], { ...child, value }])
+
+      const childContext = pathToContext(childPath)
+
+      /** Check of the path is the ancestor of the expansion path. */
+      const isAncestor = () => isDescendant(childContext, expansionPathContext)
```
raineorshine, line 147: I think the same `isAncestor` is called in the previous `childrenUnfiltered.filter`. Is it doing the same work twice?

raineorshine (changes requested):
Thanks, Bijay.

It looks like the overall approach for the `suppressExpansion` fix is to have `expandThoughts` check for pinned thoughts, and ignore `suppressExpansion` in that case. So it's kind of like `suppressSuppressExpansion`... a lot of negatives. 

My question is, why go to the trouble of suppressing expansion at all if it's going to be cancelled downstream? Why not change the `suppressExpansion` condition in `cursorNext` and `cursorPrev`? One reason would be if it is hard to access the pin state there. Another would be if it creates an undesirable dependency. I'm not sure if either of those are relevant, or if there is another reason to avoid that approach. Let me know what you think. Thanks!

src/reducers/setCursor.ts
```diff
@@ -71,7 +72,9 @@ const setCursor = (state: State, {
   //   }
   // })
 
-  const expanded = expandThoughts({ ...state, contextViews: newContextViews }, thoughtsResolved)
+  const expanded = expandThoughts({ ...state, contextViews: newContextViews }, thoughtsResolved, {
+    suppressExpansion: globals.suppressExpansion
+  })
```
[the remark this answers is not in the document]
shresthabijay: In`setCursor`,  `expandThoughts`is used to calculate the expanded state, which will actually be reflected in the view. In this case, global suppression should suppress the current cursor expansion.

However, in places like `expandBottom` , `expandThoughts` is used to know what will be the expanded state if the `expansionBasePath` is other than the current `cursor`.  Since this use case doesn't affect the view, global preference of suppression should not affect it.

So it made sense to pass the option to suppress the given `expansionBasePath` instead of directly using global preference.  It depends on what use case the `expandThoughts` is being called to pass the option. `setCursor` is using it with the current `cursor` as `expansionBasePath` which will directly be reflected in the view. That is why it needs to know about the global preference if the current cursor in the view should suppress or not, and pass it as the option to `expandThoughts`.

src/selectors/expandThoughts.ts
shresthabijay: Sure.

src/selectors/expandThoughts.ts
shresthabijay: Yes. I will try to come up with a solution.

src/reducers/setCursor.ts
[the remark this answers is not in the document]
shresthabijay: In`setCursor`,  `expandThoughts`is used to calculate the expanded state, which will actually be reflected in the view. In this case, global suppression should suppress the current cursor expansion.

However, in places like `expandBottom` , `expandThoughts` is used to know what will be the expanded state if the `expansionBasePath` is other than the current `cursor`.  global preference of suppression should not affect it.

So it made sense to pass the option to suppress the given `expansionBasePath` instead of directly using global preference.  It depends on what use case the `expandThoughts` is being called to pass the option. `setCursor` is using it with the current `cursor` as `expansionBasePath` which will directly be reflected in the view. That is why it needs to know about the global preference if the current cursor in the view should suppress or not, and pass it as the option to `expandThoughts`.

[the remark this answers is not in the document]
raineorshine: Thank you for the explanation! I figured there was some other reason.

What do you think a global vs a reducer here? Is my concern over performance unfounded?

raineorshine:
> It looks like the overall approach for the `suppressExpansion` fix is to have `expandThoughts` check for pinned thoughts, and ignore `suppressExpansion` in that case. So it's kind of like `suppressSuppressExpansion`... a lot of negatives.
> 
> My question is, why go to the trouble of suppressing expansion at all if it's going to be cancelled downstream? Why not change the `suppressExpansion` condition in `cursorNext` and `cursorPrev`? One reason would be if it is hard to access the pin state there. Another would be if it creates an undesirable dependency. I'm not sure if either of those are relevant, or if there is another reason to avoid that approach. Let me know what you think. Thanks!

@shresthabijay Thoughts?

shresthabijay:
> My question is, why go to the trouble of suppressing expansion at all if it's going to be cancelled downstream? Why not change the `suppressExpansion` condition in `cursorNext` and `cursorPrev`? One reason would be if it is hard to access the pin state there. Another would be if it creates an undesirable dependency. I'm not sure if either of those are relevant, or if there is another reason to avoid that approach. Let me know what you think. Thanks!

I don't think it would be hard to access the pin state here. Also, we have to create a dependency somewhere. I think it's okay for `cursorNext` and `cursorPrev` to change the `supressExpansion`.

src/selectors/expandThoughts.ts
shresthabijay: Yes. I need to fix this.

src/reducers/setCursor.ts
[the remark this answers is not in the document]
shresthabijay: > What do you think a global vs a reducer here?

It can be both in my opinion. The only reason I see it needs to be global is the concern of re-render but some related reducers like `cursorPrev` always would be called along side with setting the `suppressExpansion`.

raineorshine:
> I don't think it would be hard to access the pin state here. Also, we have to create a dependency somewhere. I think it's okay for `cursorNext` and `cursorPrev` to change the `supressExpansion`.

It sounds like you agree, if I am reading you correctly!

One more thought:

You can think of the `suppressExpansion` global as an API. Any code can check that variable to see whether to suppress expansion or not, and that code should not know the implementation details of `suppressExpansion`. If there are additional conditions for the suppression that are known at the time of activation, then it is ideal for the action-creator that initiates the suppression to handle that logic.

shresthabijay:
@raineorshine  I have made the changes.

- I have completely removed the `suppressExpansion` from `the expandThoughts`. It is now the responsibility of related logic to pass empty object as `expanded` when there is an active `suppressExpansion`. I think this makes sense. As `expandThoughts` simply returns an empty object anyways when there is an active `suppressExpansion`.

- `cursorPrev` and `cursorNext` prevents activating `suppressExpansion` if cursor is pinned.
- Add tests for both `cursorPrev` and `cursorNext`  to test whether they activate `suppressExpansion` on certain conditions.

raineorshine (approved):
> * I have completely removed the `suppressExpansion` from `the expandThoughts`. It is now the responsibility of related logic to pass empty object as `expanded` when there is an active `suppressExpansion`. I think this makes sense. As `expandThoughts` simply returns an empty object anyways when there is an active `suppressExpansion`.

I still think that just because a module calls `expandThoughts` does not mean it should have to know about `suppressExpansion`. It seems like it is an implementation detail that should be encapsulated in `expandThoughts` and in commands like `cursorNext` and `cursorPrev` that initiate the suppression. I'll defer to you though.

Thank you!

Issue #1224, opened 2021-06-07
Export Modal: Does not export descendants correctly in lossy mode


raineorshine:
## Steps to Reproduce

```js
- a
  - =pin
    - true
  - =archive
    - c
  - b
```

1. Open Export Modal
2. Open Advanced Options
3. Uncheck "Lossless"
4. Copy to clipboard

## Current Behavior

Only copies a single thought `a` to the clipboard.

## Expected Behavior

Should copy to the clipboard:

```js
- a
  - b
```

raineorshine:
Related: It doesn't copy all thoughts correctly when "Lossless" is re-enabled after the above.

## Steps to Reproduce

```js
- a
  - =pin
    - true
  - =archive
    - c
  - b
```

1. Open Export Modal
2. Open Advanced Options
3. Uncheck "Lossless"
4. Check "Lossless"
4. Copy to clipboard

## Current Behavior

Only copies a single thought `a` to the clipboard.

## Expected Behavior

Should copy all thoughts to the clipboard.

PR #1226, opened 2021-06-08
Fix issue with export in lossy mode.

ruchijn:
Fix: #1224 

https://user-images.githubusercontent.com/18288628/121106562-420bdd00-c7d4-11eb-8bfd-8afe4c545cc7.mp4

### DIFF
--- a/src/selectors/__tests__/exportContext.ts
+++ b/src/selectors/__tests__/exportContext.ts
@@ -1,9 +1,8 @@
 import { initialState, reducerFlow } from '../../util'
 import { setCursor, importText } from '../../reducers'
-import { HOME_PATH } from '../../constants'
 import exportContext from '../exportContext'
 
-it('test when meta and archived thoughts are included', () => {
+it('meta and archived thoughts are included', () => {
   const text = `- a
   - =archive
     - c
@@ -12,7 +11,7 @@ it('test when meta and archived thoughts are included', () => {
   - b`
 
   const steps = [
-    importText({ path: HOME_PATH, text }),
+    importText({ text }),
     setCursor({ path: [{ value: 'a', rank: 0 }] })
   ]
 
@@ -29,7 +28,7 @@ it('test when meta and archived thoughts are included', () => {
   - b`)
 })
 
-it('test when meta is included but archived thoughts are excluded', () => {
+it('meta is included but archived thoughts are excluded', () => {
   const text = `- a
   - =archive
     - c
@@ -38,7 +37,7 @@ it('test when meta is included but archived thoughts are excluded', () => {
   - b`
 
   const steps = [
-    importText({ path: HOME_PATH, text }),
+    importText({ text }),
     setCursor({ path: [{ value: 'a', rank: 0 }] })
   ]
 
@@ -53,7 +52,7 @@ it('test when meta is included but archived thoughts are excluded', () => {
   - b`)
 })
 
-it('test when meta is excluded', () => {
+it('meta is excluded', () => {
   const text = `- a
   - =archive
     - c
@@ -62,7 +61,7 @@ it('test when meta is excluded', () => {
   - b`
 
   const steps = [
-    importText({ path: HOME_PATH, text }),
+    importText({ text }),
     setCursor({ path: [{ value: 'a', rank: 0 }] })
   ]
 
@@ -77,7 +76,7 @@ it('test when meta is excluded', () => {
   - b`)
 })
 
-it('test when meta is excluded but archived is included', () => {
+it('meta is excluded but archived is included', () => {
   const text = `- a
   - =archive
     - c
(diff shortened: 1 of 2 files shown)


raineorshine:
Note quite there yet:

## Steps to Reproduce

```js
- a
  - =pin
    - true
  - =archive
    - c
  - b
```

Export with "Lossless" checked and "Archived" unchecked.

## Current Behavior

0a50b9a43

```js
- a
  - =pin
    - true
  - =archive
    - c
```

## Expected Behavior

```js
- a
  - =pin
    - true
  - b
```

raineorshine:
I suggest writing some tests for `exportContext`, since we don't have any yet.

- No options
- Exclude meta
- Exclude archived (but include other meta)

This should make it easier to ensure that you cover all the cases!

raineorshine (changes requested):
I suggest writing some tests for `exportContext`, since we don't have any yet.

- No options
- Exclude meta
- Exclude archived (but include other meta)

This should make it easier to ensure that you cover all the cases!

src/selectors/__tests__/exportContext.ts
```diff
@@ -0,0 +1,78 @@
+import { initialState, reducerFlow } from '../../util'
+import { newThought, setCursor, importText } from '../../reducers'
+import { HOME_PATH } from '../../constants'
+import exportContext from '../exportContext'
+
+describe('exportContext', () => {
```
raineorshine, line 6: You can omit the outer `describe` since Jest groups the tests by filename.

src/selectors/__tests__/exportContext.ts
```diff
@@ -0,0 +1,78 @@
...
+  it('test when meta and archived thoughts are included', () => {
```
raineorshine, line 7: This test doesn't do what it says: it can't test when meta and archived thoughts are included if there are no meta or archived thoughts. Either you mean to test the base case of `exportContext` with no arguments and no meta thoughts, or you mean to test the case where meta/archived thoughts are included.

src/selectors/__tests__/exportContext.ts
```diff
@@ -0,0 +1,78 @@
...
+      newThought({ value: 'c' }),
+      setCursor({ path: [{ value: 'a', rank: 0 }] })
+    ]
+
+    // run steps through reducer flow and export as plaintext for readable test
+    const stateNew = reducerFlow(steps)(initialState())
+
+    const exported = exportContext(stateNew, ['a'], 'text/plain', {
+      excludeMeta: false, excludeArchived: false
+    })
```
raineorshine, line 20: By convention, boolean options can be omitted if they are false. I realize that `false` is more explicit, but we are already omitting them everywhere on the project so I think it's important to be consistent.

src/selectors/__tests__/exportContext.ts
```diff
@@ -0,0 +1,78 @@
...
+    const stateNew = reducerFlow(steps)(initialState())
+
+    const exported = exportContext(stateNew, ['a'], 'text/plain', {
+      excludeMeta: false, excludeArchived: true
+    })
+
+    expect(exported).toBe(`- a
+  - b
+    - =pin
+      - true
```
raineorshine, line 75: If `=pin` is in the output, then it is not excluding meta thoughts as described.

raineorshine (changes requested):

src/selectors/__tests__/exportContext.ts
ruchijn: Removed

src/selectors/__tests__/exportContext.ts
ruchijn: Done

src/selectors/__tests__/exportContext.ts
ruchijn: Unit test was right. The test title was wrong. :(

src/selectors/__tests__/exportContext.ts
```diff
@@ -0,0 +1,78 @@
...
+      excludeMeta: false, excludeArchived: true
+    })
+
+    expect(exported).toBe(`- a
+  - =pin
+    - true
+  - b`)
+  })
+
+  it('test when meta is excluded', () => {
```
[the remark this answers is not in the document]
ruchijn: Unit test was right. The test title was wrong. :(  Fixed now

src/selectors/__tests__/exportContext.ts
ruchijn: Updated.

ruchijn:
I also added a new test to test the scenario when `meta` is excluded but `archived` was included. This would ignore the value of `archive` and export everything.

raineorshine:
Looks good, thanks!

raineorshine (approved):

Issue #1227, opened 2021-06-08
Site crashes during the last step of the Tutorial


linghong:
During the last step on the Start Tutorial, i.e. hitting the 'a'  text or the circle before a as shown on the below screenshot, the site got crashed with the error message: Type Error: s.replace is not a function
<img width="1149" alt="Screen Shot 2021-06-07 at 10 12 01 PM" src="https://user-images.githubusercontent.com/5419448/121112384-5228ba00-c7de-11eb-8cc8-13b6fbdee6d8.png">

This error doesn't exist on the main page when doing the same step of creating thoughts and hiding the thought. 
<img width="1237" alt="Screen Shot 2021-06-07 at 10 12 25 PM" src="https://user-images.githubusercontent.com/5419448/121113664-6cfc2e00-c7e0-11eb-80ad-ab3bff40076f.png">

PR #1228, opened 2021-06-08
Fix the site crash issue during the step of "hiding the thought" on the Start Tutorial 

linghong:
Fix Ticket #1227

Cause: the incoming data is not always a string, thus the String.replace  function won't work correctly.

### DIFF
--- a/src/util/normalizeThought.ts
+++ b/src/util/normalizeThought.ts
@@ -16,7 +16,10 @@ export const stripEmojiFromText = (s: string) => {
 }
 
 /** Strips all html tags. */
-export const stripTags = (s: string) => s.replace(REGEXP_TAGS, '')
+export const stripTags = (s: string | {value: string, rank: number, id: string, lastUpdated: string}) => {
+  const str = typeof s === 'string' ? s : s.value
+  return str.replace(REGEXP_TAGS, '')
+}
 
 /**
  * Making character 's' will just become an empty value ''.


raineorshine (changes requested):
Thanks for identifying the cause. We want to take a slightly different approach on the solution. `stripTags` should only have to know about strings. It should not know anything about the `Child` type, i.e. `{ value: string, rank: number, ... }`. Notice the rest of the file and think about its purpose: all of the functions in the file operate on strings, and are not dependent on types like `Child` or `Context` that are specific to **em**'s data structure.

Also, polymorphism is *generally* a bad idea. We do use it in a couple places, e.g. `isRoot`, but we want to avoid it in favor of strict API's as much as possible.

Instead, identify the problem upstream that is passing a `Child` to `stripTags` instead of the expected string.

Notice that `stripTags` only accepts a string, yet is failing at runtime due to a bad type. This tells us that it must be being called by untyped code (i.e. `js` file). The tutorial contains some of the few `js` files left in the project. A good option here is to convert the caller to `ts` and let the compiler assist in showing you where the bad argument is being generated.

linghong:
New Changes:
1. The problem comes from two original functions on the TutorialStepAutoExpandExpand: The rootSubthoughtNotCursorWithSubthoughts and the rootGrandchildNotCursor
2. Removed multiple calls of those two functions, so that old smart phones could have better performance.
3. There are some spaces missing on the text, fixed it.

src/components/Tutorial/TutorialStepSuccess.jsx
```diff
@@ -2,7 +2,7 @@ import React, { Fragment } from 'react'
 
 // eslint-disable-next-line jsdoc/require-jsdoc
 const TutorialStepSuccess = () => <Fragment>
-  <p>Huzzah... You have completed the tutorial. Creating thoughts and subthought is all you need to create a personalized thoughtspace. There's a lot more to <b>em</b> however. Additional commands have been designed to facilitate a truly fluid sensemaking experience.</p>
+  <p>Huzzah... You have completed the tutorial. Creating thoughts and subthought is all you need to create a personalized thoughtspace. There's a lot more to <b>em</b>; however, additional commands have been designed to facilitate a truly fluid sensemaking experience.</p>
```
raineorshine: The original punctuation was correct.

src/components/Tutorial/TutorialStepAutoExpandExpand.jsx
```diff
@@ -1,29 +1,34 @@
...
+    <p>{isTouch ? 'Tap' : 'Click'} {uncle ? `"${ellipsize(uncle.value)}"` : 'a thought'} to reveal its subthought "{hiddenChild}".</p>
+  </>
+}
+
+/**
+ * @param cursor The array that display the thought string which has a cursor.
+ * @param rootChildren The object array that show all the root thoughts.
+ * @returns The array that holds all the thoughts that that don't have a cursor, but have children.
+ */
+const ThoughtsNoCursorWithChild = (cursor, rootChildren) => {
```
raineorshine: Pascal case is only used on components and classes.

raineorshine (approved):
I completed the tutorial with no bugs. Thanks!

Issue #971, opened 2021-01-06
Mobile: Breadcrumbs, home, and footer tap do not work


raineorshine:

raineorshine:
Fixed in `react-native-web` `v0.15.6`

raineorshine:
Fixed (again) in #1259.

PR #1237, opened 2021-06-09
Expo setup

brunokiafuka:
In this PR we are setting React Native using Expo, which will be used to convert the web components to mobile components.


package.json
```diff
@@ -39,6 +41,10 @@
     "test:e2e:prod": "npm run servebuild & npm run test:e2e; kill $!"
   },
   "dependencies": {
+    "expo": "~41.0.1",
+    "expo-status-bar": "~1.0.4",
+    "react-native": "https://github.com/expo/react-native/archive/sdk-41.0.0.tar.gz",
+    "react-native-web": "~0.13.12",
```
raineorshine, line 47: Why downgrade?

package.json
```diff
@@ -39,6 +41,10 @@
     "test:e2e:prod": "npm run servebuild & npm run test:e2e; kill $!"
   },
   "dependencies": {
+    "expo": "~41.0.1",
+    "expo-status-bar": "~1.0.4",
+    "react-native": "https://github.com/expo/react-native/archive/sdk-41.0.0.tar.gz",
```
raineorshine, line 46: Use the `git+https://git@github.com` protocol, otherwise it will fail the CI due to a bug in npm v7.

raineorshine (changes requested):
Looks good!

Please re-install the node modules with npm v7.

The app does not require npm v7 (hence the engine value of >= v6) but it is recommended and consistent with what my team is using. Thanks!

package.json
brunokiafuka: We ensure that the version of the Expo SDK and React Native community packages that you install are compatible with the app by running `expo install`.

[Ref](https://docs.expo.io/introduction/walkthrough/)

package.json
brunokiafuka: Setting it using the `git+https://git@github.com` doesn't work it seems to behave differently since we are getting a `tar` rather than a repository. Can you test it out to see if the CI will work with syntax

package.json
raineorshine: Okay, thanks!

package.json
brunokiafuka: @raineorshine this worked locally.
`"react-native": "git+https://github.com/expo/react-native/archive/sdk-41.0.0.tar.gz"`

raineorshine: You can see the checks are failing in this PR. https://github.com/cybersemics/em/pull/1237/checks

brunokiafuka: Fixed

package.json
```diff
@@ -47,6 +49,8 @@
     "dexie": "^3.0.3",
     "emitter20": "^2.0.0",
     "emoji-strip": "^1.0.1",
+    "expo": "~41.0.1",
+    "expo-status-bar": "~1.0.4",
```
raineorshine, line 53: Please use `^` instead of `~`

raineorshine (approved):
Thanks! I can merge as soon as the conflicts are resolved.

raineorshine:
I upgraded `react` and `react-native-web` to latest in `dev`. Note that we need `react-native-web` >= 0.15.6 otherwise some click handlers stop working (#971).

PR #1240, opened 2021-06-10
Do not dim leaf cursor siblings

### LINKED ISSUE #1239: Leaf siblings should not be dimmed
Recent regression (assuming #1220). Missing test coverage. 

## Steps to Reproduce

```js
- a
  - b
  - c
    - d
``` 

Set the cursor on `a/b`  

## Current Behavior

`a/c` is dimmed. 

## Expected Behavior

`a/c` should not be dimmed. 

Lead siblings should not be dimmed.

shresthabijay:
fixes #1239

### DIFF
--- a/src/components/Subthoughts.tsx
+++ b/src/components/Subthoughts.tsx
@@ -435,6 +435,7 @@ export const SubthoughtsComponent = ({
   const isDescendantOfCursor = cursor && cursor.length === (cursorSubcontextIndex + 1) && resolvedPath.length > cursor?.length
 
   const isCursor = cursor && resolvedPath.length === (cursorSubcontextIndex + 1) && resolvedPath.length === cursor?.length
+  const isCursorParent = cursor && isAncestorOfCursor && (cursor.length - resolvedPath.length) === 1
 
   /*
     The thoughts that are not the ancestor of cursor or the descendants of first visible thought should be shifted left and hidden.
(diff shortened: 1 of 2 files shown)


shresthabijay:
@raineorshine I have made the changes. This was certainly a regression from #1220. The visibility tests should now prevent such regressions. However, It would be much better if we could add docs related to when thought should dim, shift to left, etc. This will help anyone who will need to work on this part of the codebase again.

For example:

# Thoughts Dimming

### Thoughts that should not be dimmed
  - Cursor and its descendants.
  - Thoughts that are both descendant of the first visible thought and ancestor of the cursor.
  - Siblings of the cursor if the cursor is a leaf thought.

### Thoughts that should be dimmed
  - First visible thought should be dimmed if it is not direct parent of the cursor.
  - Besides the above mentioned thoughts in the above "Should not dim section", all the other thoughts that are descendants of the first visible thought should be dimmed.
  
 I will write docs related to hiding thoughts and related things. You can verify it and add them to the project wiki. Please let me know what you think. Thanks!

raineorshine (approved):

raineorshine:
Thanks Bijay!

I acknowledge that these rules are subtle and impossible to know ahead of time when there is no test coverage. It will be much better now that we have proper test coverage. Even if we move away from the `distance-from-cursor` implementation, the test cases are the same and will be easy to convert.

> However, It would be much better if we could add docs related to when thought should dim, shift to left, etc. This will help anyone who will need to work on this part of the codebase again.

Unit tests are the best documentation. I think the block comment you added is perfect. It complements the unit tests by providing a quick, high-level overview of the behavior, and is close to the code where it is most relevant. Duplicating this in the Wiki is unnecessary. The Wiki is better for high level concepts that cannot be demonstrated easily with some sample code or unit tests.

Issue #1208, opened 2021-06-01
Mobile: Enter edit mode


raineorshine:
1. Create a thought `a`
2. Close the keyboard
   - Expect no browser selection
3. Tap on `a`
   - Expect the browser selection to be on `a`

(Working as expected. Issue created for documentation purposes.)

PR #1245, opened 2021-06-10
Add ios e2e test for 'Mobile: Enter edit mode' case

onurpolattimur:
- Automate test case for Enter edit mode  (#1208)
- Refactor gesture helper function.

### DIFF
--- a/src/e2e/helpers/mobile/gestures/gesture.ts
+++ b/src/e2e/helpers/mobile/gestures/gesture.ts
@@ -29,6 +29,8 @@ const gesture = async (browser: Browser<'async'>, path: Direction[], { xStart = 
     { action: 'release' }
   ] as TouchAction[]
 
+  // webdriverio has some problem about type for TouchAction[] that is why we add @ts-ignore
+  // We didn't use touchPerform here because webdriverio calls touchPerform in the background. https://github.com/webdriverio/webdriverio/blob/aea3d797ab1970309a60c43629f74154453597e9/packages/webdriverio/src/commands/constant.ts#L100
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
   await browser.touchAction(actions)
--- a/src/e2e/helpers/mobile/gestures/gesture.ts
+++ b/src/e2e/helpers/mobile/gestures/gesture.ts
@@ -1,5 +1,5 @@
 import { Direction } from '../../../../types'
-import { Browser } from 'webdriverio'
+import { Browser, TouchAction } from 'webdriverio'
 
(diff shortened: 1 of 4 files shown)


src/e2e/helpers/mobile/gestures/gesture.ts
```diff
@@ -1,37 +1,37 @@
...
+    { action: 'press', x: xStart, y: yStart },
     ...moveActions,
     { action: 'release' }
-  ]
+  ] as TouchAction[]
 
-  await browser.touchPerform(actions)
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  await browser.touchAction(actions)
```
onurpolattimur, line 34: As I can see, webdriverio has a type issue in touchAction function. Can you confirm that, please? @raineorshine

raineorshine: `browser.touchAction` performs a single action. Do you mean `touchPerform`?

> 'Touch Perform' works similarly to the other singular touch interactions, except that this allows you to chain together more than one touch action as one command. This is useful because Appium commands are sent over the network and there's latency between commands. This latency can make certain touch interactions impossible because some interactions need to be performed in one sequence. Vertical, for example, requires pressing down, moving to a different y coordinate, and then releasing. For it to work, there can't be a delay between the interactions.

https://appium.io/docs/en/commands/interactions/touch/touch-perform/

src/e2e/helpers/mobile/gestures/gesture.ts
```diff
@@ -1,37 +1,37 @@
...
 export interface GestureOptions {
   xStart?: number,
   yStart?: number,
   segmentLength?: number,
 }
 
-interface Action<T = any> {
-  action: string,
-  options?: T,
-}
```
raineorshine, line 13: Thanks, I didn't look hard enough for webdriverio's `TouchAction` 🙂

raineorshine (approved):

src/e2e/helpers/mobile/gestures/gesture.ts
onurpolattimur: 🤔
Here is webdriverio documentation: 
https://webdriver.io/docs/api/browser/touchAction/

Here is an example from this document: 
```    
// multi action on an element
// drag&drop from position 200x200 down 100px on the screen
  browser.touchAction([
      { action: 'press', x: 200, y: 200 },
      { action: 'moveTo', x: 200, y: 300 },
      'release'
  ])```

I will look into the details of the touchPerform.

PR #1252, opened 2021-06-11
Render modal on the top

### LINKED ISSUE #1235: Render modal on top
## Steps to Reproduce

Activate a modal such as the Export Modal. 

## Current Behavior

The modal is rendered and all other components are not rendered. 

## Expected Behavior

Fade in and render the modal on top of other components such as the Content,
Toolbar, and Navigation. Fade out on close such that you can see the underlying components. 

Make sure the body scroll position and toolbar scroll position are maintained after the modal is closed.

shresthabijay:
fixes #1235


shresthabijay:

raineorshine (changes requested):
Thanks! Here are the issues I am seeing:

- The modal should not scroll unless the modal content exceeds the height of the screen. Currently there is a lot of empty space at the bottom (possibly from the Content underneath) that causes it to scroll. See `dev` for the correct behavior.
- The fade-in animation seems to get cut off at the end. It is smooth up until about 20% from the end and then it abruptly cuts to the end. It should be a smooth `ease-in-out` that is a bit shorter overall.
- We don't want the scroll to be noticeable to the user. The modal is just covering what the user was looking at. The modal should fade out and the Content should be in exactly the same place that it was when they activated the modal.
- The toolbar and hamburger menu appear above the Help Modal after scrolling down.
- The bottom margin on the Help Modal is gone.

shresthabijay:
@raineorshine  With a scrollable body, we had two options. Either give modal top offset equal to the body scroll (not so reliable when the keyboard is up) or first scroll the body up and then show the modal (user notices the scroll). So I made the body overflow behavior hidden and made a scrollable container instead. 

Now modal always shows at the top even if the content is scrolled. On ios safari when the keyboard is up, the toolbar won't get janky too. But it has problems with ios safari. They are as follows:

- Body scroll cannot be fully prevented. So sometimes on scrolling certain sections, the body gets scrolled instead of a nested scrollable container. This may cause weird visual problems.

- Since ios safari does body scroll to keep the element at the viewport on focus, the nested scroll may not be able to show some content at the top of the scroll container. However, I have solved showing content not showing at the bottom. 

The problems I described above are only for ios safari when the keyboard is up. On all the other platforms it should work properly. I took some time to check if some other note-taking apps had solved this problem, but sadly apps like **notion** have the same exact issue. So this may be as far as we can go with a non-scrolling body with a nested scrollable container in ios safari.

Some tests are failing. I will fix them if you approve this approach. Thanks!

raineorshine:
Thanks for the explanation. The problems on iOS Safari are major so unfortunately we can't go with this approach. I understand that the browser itself has bugs so there is only so much we can do. I would be interested in considering a different approach that does not rely on a scrollable container, but I think this issue is not high enough priority to devote more time to it at this time. The work you did will be useful when we return to the problem in the future. Thanks!

Issue #1209, opened 2021-06-01
Mobile: Preserve editing true


raineorshine:
1. Create a thought `a`
1. Create a subthought `b`
1. Tap on `a`
   - Expect the browser selection to be on `a`

(Working as expected. Issue created for documentation purposes.)

PR #1260, opened 2021-06-11
Add ios e2e test for 'Preserve editing: true' case

onurpolattimur:
Automate test for "Preserve editing: true" case (#1209)


src/e2e/helpers/constants.ts
```diff
@@ -1,3 +1,5 @@
 import { Direction } from '../../types'
 
 export const NEW_THOUGHT_GESTURE: Direction[] = ['r', 'd']
+
+export const NEW_SUB_THOUGHT_GESTURE: Direction[] = ['r', 'd', 'r']
```
raineorshine, line 5: This list is going to grow a lot and have to be maintained in parallel with the shortcuts themselves. I have two ideas of how to make it more maintainable:

1. Remove these and import the shortcuts directly in the tests.
2. Define these in a single object with keys that match the shortcut names.

(1) eliminates all redundancy, or (2) tidies it up.

Which one do you like better?

src/e2e/helpers/mobile/sendKeysForThought.ts
```diff
@@ -0,0 +1,10 @@
+import { Browser } from 'webdriverio'
+import waitForEditable from './waitForEditable'
+
+/** Send keys and wait for editable exists for the given values. */
+const sendKeysForThought = async (browser: Browser<'async'>, keys: string[]) => {
+  await browser.sendKeys(keys)
+  await Promise.all(keys.map(key => waitForEditable(browser, key)))
+}
```
raineorshine, line 8: Can you help me understand the need for this a bit better?

I have two questions:

1. When would you want to send more than one string?
2. Why do we need to wait for the editable? Isn't the editable's innerHTML changing right away once the keys are sent?

raineorshine (changes requested):

src/e2e/helpers/mobile/sendKeysForThought.ts
onurpolattimur: Yes of course I will do my best to make you understand. 

1. For now, we don’t have any case that includes more than one string but we may need to have. For example; we can create 2 thoughts bypassing [“foo”, "\n", “bar”].  

2. As you know some of the operations in tests happen truly quickly. Before you even complete writing something, another action may happen. We need to be sure that the current action affects our app before running the next action. In this case, we want to be sure that there is a thought with the given value. Here is a result when we don't do that; 

https://automate.browserstack.com/dashboard/v2/builds/b6036aca2ddad6636c22c83fd2e9e90dc7af0dba/sessions/5d1d1bec14bc4ee7bdcb9cf36dbb858d9a1d4561

src/e2e/helpers/constants.ts
onurpolattimur: I am not sure but I prefer the second option. I don't have strict reasons but I can say that I don't want our e2e test to be domain-dependent.

Think about this scenario, when somebody changes the gesture of the newThought by mistake, I want to catch this issue in our tests. 
I always think of the e2e tests from an end-user perspective. If you change the shortcut of any action, you need to inform me. Maybe you can mention this on the changelog or release description.  So, I think you need to inform our e2e test also.

src/e2e/helpers/mobile/sendKeysForThought.ts
raineorshine: Thanks, that's helpful!

> 1. we can create 2 thoughts bypassing [“foo”, "\n", “bar”]

I hadn't thought of `\n`. We do have `tapReturnKey`. I suspect that explicit `tapReturnKey` will be more clear for taking steps one at a time like the user, and `paste` will be used for setting up a test case by importing lots of thoughts.

I recommend making the argument polymorphic: `keys: string | string[]`. Otherwise the common case of typing a single thought will require the string to be wrapped in an array which is awkward.

> 2\. As you know some of the operations in tests happen truly quickly. Before you even complete writing something, another action may happen. We need to be sure that the current action affects our app before running the next action. In this case, we want to be sure that there is a thought with the given value. Here is a result when we don't do that;
> 
> [automate.browserstack.com/dashboard/v2/builds/b6036aca2ddad6636c22c83fd2e9e90dc7af0dba/sessions/5d1d1bec14bc4ee7bdcb9cf36dbb858d9a1d4561](https://automate.browserstack.com/dashboard/v2/builds/b6036aca2ddad6636c22c83fd2e9e90dc7af0dba/sessions/5d1d1bec14bc4ee7bdcb9cf36dbb858d9a1d4561)

Let's add this as a comment, otherwise other developers will have the same question I did. It helps to know specifically what will go wrong if this is omitted. The last thing we want is newer developers adding delays or waits without understanding why they are doing so.

Also: How about we rename it `editThought` for better readability. It captures the intention that we are typing in an editable but do not want to continue before it is updated. We can use `browser.sendKeys` directly for non-editing keyboard interactions.

onurpolattimur: Yes, I thought it can be useful but I realized an issue after I saw your comment. It will not work if we pass  `[“foo”, "\n", “bar”]`. Because it will try to find an editable that contains only  `\n`. 

We have 2 options left
1. Ignore the issue above and make the type of parameter keys: `string | string[]`
2. Make the `keys` parameter of type `string`

raineorshine: Ah, good point. Okay let's make it just edit a single thought for now. We can always add a higher level function if we encounter the need to abstract editing larger numbers of thoughts, and then we will be informed by concrete cases.

src/e2e/helpers/constants.ts
```diff
@@ -1,3 +1,6 @@
 import { Direction } from '../../types'
 
-export const NEW_THOUGHT_GESTURE: Direction[] = ['r', 'd']
+export const gestures = {
+  newThought: ['r', 'd'] as Direction[],
+  newSubThought: ['r', 'd', 'r'] as Direction[]
+}
```
raineorshine: You can type the whole object as `Index<Direction[]>` instead of each property.

raineorshine (approved):

PR #1264, opened 2021-06-12
iOS test cleanup

raineorshine:


onurpolattimur (approved):
I was planning to write something like initSession but you acted before me and it really looks great!  Thank you, Raine!

Issue #612, opened 2020-04-30
=let


raineorshine:
Define variables that are accessible to all descendants. 

## Example

Apply a defined header style:

- a
  - `=let`
    - heading
      - font-size
        - 24px
  - b
    - c
      - `=style`
        - `=heading`

## Requirements

- Lexically scoped
  - `a` should pass its `=let` variables to each of its children. Children should pass them to their children. 
- Consider including the context in which the variable is defined for better tracing the origin of variables in scope. 
- Parse **em** functions using the in-scope env. 
- Merge children

PR #1266, opened 2021-06-12
=let style variables

raineorshine:
#612

Add support for defining scoped `=style` variables with `=let`.

```js
- =let
  - =h1
    - =style
      - fontSize
        - 30px
      - fontWeight
        - bold
- Header
  - =h1
```

will apply the `=h1` styles to the `Header`  as if it was:

```js
- Header
  - =style
    - fontSize
      - 30px
    - fontWeight
      - bold
```


onurpolattimur:
It is really a perfect feature Raine! 

Just a few question;
- What do we expect when users enter our metaprogramming keys as a variable name? For example, when the user enters =sort instead of =h1 while defining a variable, the app will crash.

- Also, the style is applied to thought '=let' itself. Do we want this?
Here is a screenshot: 
image

raineorshine:
> It is really a perfect feature Raine!

Thank you! **em** has so much potential once we start adding more programmability.

> * What do we expect when users enter our metaprogramming keys as a variable name? For example, when the user enters =sort instead of =h1 while defining a variable, the app will crash.

User-defined variables should probably be namespaced so they do not interfere with native metaprogramming attributes.

> * Also, the style is applied to thought '=let' itself. Do we want this?

It should only be applied to `=let/=h1` and `onur` (and arguably only `onur`, since `=let/=h1` is just the definition. It is happening because the `=let` expression is being parsed and applied to its containing context, with of course includes `=let` itself. I didn't know it was possible to have a function definition and application at the same time 😂.

onurpolattimur:
> User-defined variables should probably be namespaced so they do not interfere with native metaprogramming attributes.

Yes, we should prevent that users enter our metaprogramming attibute as a variable name.

> It should only be applied to =let/=h1 and onur (and arguably only onur, since =let/=h1 is just the definition. It is happening because the =let expression is being parsed and applied to its containing context, with of course includes =let itself. I didn't know it was possible to have a function definition and application at the same time 😂.

This can be useful as a preview. I mean user can see their variable effect without adding this variable to a thought 😅

raineorshine:
> > User-defined variables should probably be namespaced so they do not interfere with native metaprogramming attributes.
> 
> Yes, we should prevent that users enter our metaprogramming attibute as a variable name.

Ideally the user can mask most native attributes using lexical scoping. They can still use the same names, but they will just hide native attributes. However since the all operate in the same space, that could definitely cause problems when **em** toggles a specific attribute like `=sort` that it expects to behave in a certain way. Not sure how to handle that yet.

PR #1272, opened 2021-06-13
Test Helper Cleanup

raineorshine:


onurpolattimur:
I don't why it looks success but puppeteer tests are failing because of timeout. https://github.com/cybersemics/em/pull/1272/checks?check_run_id=2814354163#step:6:24

After running the tests in non-headless mode I saw the error below and they stuck in here.  So it gets timeout.
image

The reason why test results look successful is probably it checks the exit code of the last command to decide the test is successful or failure. 

The command to run test is: `npm run servebuild & npm run test:e2e; kill $!`
So the last command is `kill$!` and this command runs even any error occurs while running ` npm run test:e2e` Therefore, GitHub will get the result of this command, not the `npm run test:e2e` command.

src/e2e/helpers/mobile/paste.ts
```diff
@@ -1,13 +1,23 @@
...
+import { HOME_TOKEN } from '../../../constants'
+
+const em = window.em as WindowEm
+
+async function paste(browser: Browser<'async'>, text: string): Promise<void>
+async function paste(browser: Browser<'async'>, pathUnranked: string[], text: string): Promise<void>
 
 /** Import text on given unranked path using exposed testHelpers. */
-const paste = async (browser: Browser<'async'>, unrankedPath: string[], text: string) => {
+async function paste(browser: Browser<'async'>, pathUnranked: string | string[], text?: string): Promise<void> {
```
onurpolattimur, line 11: What do you think about below? Isn't it enough?

```ts
async function paste(page: Page, text: string, pathUnranked = [HOME_TOKEN]): Promise<void> {
  // Note: This helper is exposed because copy paste seemed impossible in headless mode. With headless false copy paste with ctrl + v seems to work. ??
  await page.evaluate((_pathUnranked, text) => {
    const testHelpers = em.testHelpers
    testHelpers.importToContext(_pathUnranked, text)
  }, pathUnranked, text)
}
```

raineorshine: It would work. I think it’s more intuitive to pass the path first though, since it is short while the text can take up multiple lines. I’m willing to have a slightly more complex definition of it makes the tests flow better.

raineorshine:
The puppeteer test should be fixed in #1273. 

@shresthabijay @onurpolattimur Do either of you have a better solution for serving the `build`? I didn’t catch the problem with kill when it was changed. I think we did it originally so that we could have one server running for all the puppeteer tests instead of restarting it each time.

onurpolattimur:
> The puppeteer test should be fixed in #1273.
> 
> @shresthabijay @onurpolattimur Do either of you have a better solution for serving the `build`? I didn’t catch the problem with kill when it was changed. I think we did it originally so that we could have one server running for all the puppeteer tests instead of restarting it each time.

I think we can add serve step into our pipeline after the build step. So we can run only `npm run test:e2e` after serve step. It should be enough for CI runs.
https://github.com/cybersemics/em/blob/4cef26ad1d2b15c3982da66faba5eb94b35b02b3/.github/workflows/puppeteer.yml#L32

onurpolattimur (approved):

PR #1273, opened 2021-06-13
Test: Import thoughts into empty home context

raineorshine:


src/reducers/importText.ts
```diff
@@ -159,7 +159,7 @@ const importText = (state: State, { path, text, lastUpdated, preventSetCursor, r
   const destValue = rawDestValue || destThought.value
 
   // if we are only importing a single line of html, then simply modify the current thought
-  if (numLines <= 1 && !isRoam) {
+  if (numLines <= 1 && !isRoam && !isRoot(path)) {
```
onurpolattimur, line 162: I didn't quite get why you added `!isRoot(path)` Can you help me understand the need for this?

raineorshine: There was a bug that was crashing the app when pasting a single thought into an empty root. The first new test covers this now.

onurpolattimur (approved):

PR #1275, opened 2021-06-13
Rename thought reducers

raineorshine:
The original reducers for basic thought editing are unnecessarily long. There are also two different versions of some reducers which operate on different levels but are confusingly named (`deleteThought` vs `existingThoughtDelete`, `newThought` vs `newThoughtSubmit`).

3027f5aec Rename newThoughtSubmit -> createThought
6fe31f992 Rename existingThoughtDelete -> deleteThought
b3917e563 Rename deleteThought -> deleteThoughtWithCursor
d8b5155bc Rename existingThoughtMove -> moveThought
fa117e131 Rename moveThought -> moveLexemeThought
d08675eab Rename existingThoughtChange -> editThought

Here is a summary of their functionality:

```ts
/** Removes a thought from a context. If it was the last thought in that context, removes it 
completely from the thoughtIndex. Does not update the cursor. Use deleteThoughtWithCursor or 
archiveThought for high-level functions. */
const deleteThought = (state: State, { context, thoughtRanked, showContexts }: Payload) => ...

/** Deletes a thought and moves the cursor to a nearby valid thought. */
const deleteThoughtWithCursor = (state: State, payload: { path?: Path }) => ...

/** Adds a new thought to the cursor. Calculates the rank to add the new thought above, below, 
or within a thought. */
const newThought = (state: State, payload: {
  at?: Path,
  insertNewSubthought?: boolean,
  insertBefore?: boolean,
  value?: string,
  offset?: number,
  aboveMeta?: boolean,
  preventSetCursor?: boolean,
} | string) => ...

/** Creates a new thought with a known context and rank. Does not update the cursor. Use the 
newThought reducer for a higher level function. */
const createThought = (state: State, { 
  context, 
  value, 
  rank, 
  id, 
  addAsContext 
}: Payload) => ...

/** Moves a thought from one context to another, or within the same context. */
const moveThought = (state: State, { oldPath, newPath, offset }) => ...

```


shresthabijay (approved):
Looks good to me!

PR #1278, opened 2021-06-14
Unify e2e helpers

raineorshine:
## Requirements

- Create a single api for device interaction that can be used across jsdom, puppeteer, and webdriver. 
- Make tests as readable as possible
- Avoid globals


src/test-helpers/partialWithRef.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * const { click, press, type } = partialWithRef(ref, { click, press, type })
+ *
+ * it('test', () => {
+ * await press('Enter')
+ * await type('Hello')
+ * })
+ *
+ **/
+const partialWithRef = <HelperObject extends ObjectWithInstanceMethods>(ref: Ref<Parameters<HelperObject[keyof HelperObject]>[0]>, helpers: { [Key in keyof HelperObject]: HelperObject[Key] }) => {
```
onurpolattimur, line 29: A poem written by Raine, from 21st century.

src/e2e/iOS/helpers/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  const init = initSession()
+  const browserRef = {} as { current?: Browser<'async'> }
+  const index = partialWithRef(browserRef, helpers)
+
+  beforeEach(async () => {
+    browserRef.current = await init()
+  })
+
+  return index as typeof index & {
+    paste: typeof pasteOverload,
```
onurpolattimur, line 65: Can we also add browserRef?  To avoid accessing webdriverio browser methods with browserRef.current we can add it like below.

```
 get mobileBrowser() {
    return browserRef.current!
  }
```

onurpolattimur, line 65: Can we also add browserRef? 
To avoid accessing webdriverio browser methods with browserRef.current we can add it as a function. So, I can access the current browserRef by `mobileBrowser()`

src/e2e/iOS/helpers/getSelection.ts
```diff
@@ -0,0 +1,36 @@
...
+
+/** Returns a proxy getSelection object with async getters for selection properties. WebdriverIO cannot retrieve the entire selection object so we delegate to asynchronous getters. Top-level selection properties (e.g. .focusNode) are compatible with nested properties (e.g. .focusNode.textContent) are compatible with parent properties by defining ad hoc getters on promises.
+ *
+ * @example
+ * await getSelection().focusOffset
+ * await getSelection().focusNode
+ * await getSelection().focusNode?.textContent
+ *
+ **/
+const getSelection = (browser: BrowserEnvironment) => {
```
onurpolattimur, line 12: Should this file be in iOS/helpers? It is also used by puppeteer.

src/e2e/iOS/__tests__/caret.ts
```diff
@@ -3,66 +3,67 @@
...
-  expect(selection?.focusOffset).toBe(0)
-  expect(selection?.focusNode).toBe(null)
+  await gesture(gestures.newThought)
+  await type('foo')
+  await hideKeyboardByTappingDone()
 
-  const editableNodeHandle = await waitForEditable(mobileBrowser, 'foo')
-  await tapWithOffset(mobileBrowser, editableNodeHandle, { offset: 0 })
+  const editableNodeHandle = await waitForEditable('foo')
+  await tapWithOffset(editableNodeHandle, { offset: 0 })
```
onurpolattimur, line 35: We can remove offset:0

src/e2e/puppeteer/helpers/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  beforeEach(async () => {
+    pageRef.current = await setup(...setupArgs)
+  })
+
+  afterEach(async () => {
+    await pageRef.current!.browserContext().close()
+  })
+
+  return index as typeof index & {
+    paste: typeof pasteOverload,
```
onurpolattimur, line 70: Can we add pageRef here?

onurpolattimur (approved):
It looks really amazing. I have better knowledge of typescript now, thanks to you.

src/e2e/iOS/helpers/getSelection.ts
raineorshine: It can be used as-is with webdriverio, since `Browser<'async'>` is a `BrowserEnvironment`. However, the puppeteer helper should not have to know about `iOS` at all. I agree it would be better in a more generic directory.

PR #1280, opened 2021-06-15
Add ios e2e test for 'Tab hidden uncle' case

onurpolattimur:
Automate test for "Tap hidden uncle" case (https://github.com/cybersemics/em/pull/1128#pullrequestreview-654800218)


src/e2e/iOS/__tests__/caret.ts
```diff
@@ -104,3 +105,30 @@ it('Tap hidden root thought', async () => {
...
+
+  const editableNodeHandle = await waitForEditable(mobileBrowser, 'c')
+  const elementRect = await getElementRectByScreen(mobileBrowser, editableNodeHandle)
+
+  // click near uncle. see: https://github.com/cybersemics/em/pull/1128#issuecomment-837427219
+  await mobileBrowser.touchAction({
+    action: 'tap',
+    x: elementRect.x + (elementRect.width / 2),
+    y: elementRect.y + elementRect.height + 10,
+  })
```
raineorshine, line 129: Could you do this with `tapWithOffset`? `getElementRectByScreen` would be handy in other situations, but maybe here we can use what we already have. It would be nice to abstract away the pixel calculations anyway.

raineorshine (approved):

PR #1281, opened 2021-06-15
Add ios e2e test for 'Tap empty content' case

onurpolattimur:
Automate test for "Tap empty content" case (https://github.com/cybersemics/em/pull/1128#pullrequestreview-656073834)
________
Rebase after merging #1280


raineorshine (changes requested):
Let's do two tests: one with the keyboard up and one with the keyboard down. It is possible that only one is affected, as having the keyboard up can change the caret behavior. The behavior should be the same in both cases, while preserving the keyboard up or down, respectively.

src/e2e/iOS/__tests__/caret.ts
```diff
@@ -124,3 +124,46 @@ it('Tap hidden uncle', async () => {
...
+  await clickThought(mobileBrowser, 'c')
+  await hideKeyboardByTappingDone(mobileBrowser)
+
+  const editableNodeHandleD = await waitForEditable(mobileBrowser, 'd')
+  await tapWithOffset(mobileBrowser, editableNodeHandleD, { x: 20, y: 200 })
+
+  // Wait until cursor change
+  await mobileBrowser.waitUntil(async () => await getEditingText(mobileBrowser) === 'b')
+  const editingText = await getEditingText(mobileBrowser)
+  expect(editingText).toBe('b')
```
raineorshine, line 168: We should also test that the keyboard has stayed down, i.e. there is no focusNode.

src/e2e/iOS/__tests__/caret.ts
```diff
@@ -124,3 +124,46 @@ it('Tap hidden uncle', async () => {
...
+  await clickThought(mobileBrowser, 'b')
+  await clickThought(mobileBrowser, 'c')
+
+  const editableNodeHandleD = await waitForEditable(mobileBrowser, 'd')
+  await tapWithOffset(mobileBrowser, editableNodeHandleD, { x: 20, y: 200 })
+
+  // Wait until cursor change
+  await mobileBrowser.waitUntil(async () => await getEditingText(mobileBrowser) === 'b')
+  const editingText = await getEditingText(mobileBrowser)
+  expect(editingText).toBe('b')
```
raineorshine, line 146: We should also test that the keyboard has stayed up, i.e. there is a focusNode.

raineorshine (changes requested):

src/e2e/iOS/__tests__/caret.ts
onurpolattimur: @raineorshine, I made changes.
After tapping the empty area while the keyboard is down, the focusNode is `<span role="presentation" class="Resizer vertical "></span>`  element and it isn't null. That is why I used isKeyboardShown function.

raineorshine (approved):

Issue #1253, opened 2021-06-11
toggleHiddenThoughts button doesn’t re-render 


raineorshine:
## Steps to Reproduce

Click the `toggleHiddenThoughts` button in the toolbar. 

## Current Behavior

The button stays dimmed. 

## Expected Behavior

The button should be white after activation. 

The button should re-render whenever `showHiddenThoughts` is changed.

PR #1288, opened 2021-06-17
Fix issue with toggleHiddenThoughts not being set to active when clicked

ruchijn:
Fix: #1253 

Before:
toggleHiddenThoughts_before

After:
toggleHiddenThoughts_after

### DIFF
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -31,9 +31,20 @@ const ARROW_SCROLL_BUFFER = 20
 // eslint-disable-next-line jsdoc/require-jsdoc
 const mapStateToProps = (state: State) => {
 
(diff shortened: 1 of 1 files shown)


src/components/Toolbar.tsx
```diff
@@ -31,18 +31,9 @@ const ARROW_SCROLL_BUFFER = 20
...
     dark: theme(state) !== 'Light',
-    isLoading,
-    fontSize,
-    scrollPrioritized,
-    toolbarOverlay,
-    // We cannot know if any one the shortcut's active status,has changed, so we re-render everytime thoughts or cursor is changed
-    thoughts,
-    cursor,
-    showTopControls
+    ...state
```
raineorshine, line 36: Re-rendering the Toolbar when anything in the state changes is overkill and likely to introduce performance issues. It should only re-render when it needs to.

raineorshine (changes requested):

raineorshine (approved):

Issue #1242, opened 2021-06-10
Gesture hint is not cleared after gesture is abandoned


raineorshine:
Tested on Mobile Safari

## Steps to Reproduce

Tap and swipe just a few pixels down, just enough to activate a scroll, then, in the same gesture, swipe right until the Back gesture hint is shown, then swipe down. 

![2FAEA860-7F41-4A3C-894D-79A28AD3064A](https://user-images.githubusercontent.com/750276/121501413-39223380-c9ad-11eb-9c2a-5f955cbdc2ee.jpeg)

## Current Behavior

The gesture hint is shown until touchend. 

## Expected Behavior

The gesture hint should be dismissed as soon as the gesture is abandoned,
i.e. internally cancelled once the scroll threshold is reached.

PR #1292, opened 2021-06-19
Dismiss gesture hint when gesture is abandoned on scroll

onurpolattimur:
fixes: #1242


raineorshine (changes requested):
Thanks! It is an improvement, though it still happens sometimes unfortunately. It's even harder to reproduce now. I can reproduce it most consistently when the scrollbar is completely still at the start, and the final down swipe is just a few pixels long. If you are unable to reproduce then I will have to get remote debugging working and try to track the touchmove events.

Architecturally, we need to take a slightly different approach. The `MultiGesture` component should not know about the global store or gesture hints. I suggest adding a new event handler `onCancel` that allows the upstream component to add the appropriate logic.

onurpolattimur:
I could reproduce only once but I made some changes. Can you check it again, please?

src/components/MultiGesture.tsx
```diff
@@ -87,6 +88,9 @@ class MultiGesture extends React.Component<MultiGestureProps> {
 
       onPanResponderMove: (e: GestureResponderEvent, gestureState: GestureState) => {
         if (this.abandon) {
+          if (this.props.onCancel) {
+            this.props.onCancel()
+          }
```
raineorshine, line 93: You can also use the fancy new [Optional Chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator: `this.props.onCancel?.()`

raineorshine (commented):
It still occurs consistently for me unfortunately. Would you like to keep working on it or shall I give it a try?

onurpolattimur:
> It still occurs consistently for me unfortunately. Would you like to keep working on it or shall I give it a try?

I can't re-produce still after 8384ab9. So you can give it a try if you want.

raineorshine:
@onurpolattimur The main issue was the scrolling was causing the `MultiGesture` to terminate without `onPanResponderRelease` (and thus `onEnd`) being called.

In order to make the gesture system more forgiving, we currently allow a tiny bit of scroll without abandoning the gesture. Unfortunately, in the case of this issue, `onPanResponderRelease` is not called, even after `touchend`. Neither is `onPanResponderReject` or `onPanResponderEnd`.

`onPanResponderTerminate` is called consistently, but it is also called for any scroll event. I am not aware of a way to differentiate when `onPanResponderTerminate` is called from a scroll event vs a final termination where release it never called.

So instead of eliminating the scroll lenience, we can handle `touchend` manually and ensure `onEnd` is called if there is a valid sequence.

```ts
    document.body.addEventListener('touchend', e => {
      if (this.sequence) {
        setTimeout(() => {
          if (!this.abandon && this.sequence) {
            this.props.onEnd?.(this.sequence, e as unknown as GestureResponderEvent)
            this.reset()
          }
        })
      }
    })
```

I also cleaned up some state access that was happening outside of a thunk, which could lead to timing issues, but I don't think that was causing the problem.

raineorshine:
Apparently I can't request a review from you since you created the PR, but feel free to take a look.

onurpolattimur:
Hi @raineorshine. Glad you found it.
I reviewed and it looks good to me.

raineorshine (approved):

Issue #1291, opened 2021-06-18
Toolbar disabled briefly after clearing alert


raineorshine:
## Steps to Reproduce

```js
- a
  - b
  - c
```

1. Set the cursor on `c`
2. Tap delete in the toolbar
3. Tap the ‘x’ in the upper right corner to close the alert
4. Quickly tap the outdent button on the toolbar

You have to execute steps 2-4 rather quickly, otherwise it works fine. 

## Current Behavior

Nothing happens

## Expected Behavior

b is outdented

PR #1293, opened 2021-06-19
Remove alert components after dismiss alert by clicking the 'x'

onurpolattimur:
fixes: #1291  
____
CSSTransition apply *-enter, *-enter-active, *-exit and *-exit-active respectively. When we dismiss the alert, it will stick on *-exit-active class until the timeout ends. And the toolbar stays behind the alert component. So, that's why we can't click the items in the toolbar.
Now, we set the timeout to 0 after clicking the dismiss button to remove alert components from DOM.

### DIFF
--- a/src/components/Alert.tsx
+++ b/src/components/Alert.tsx
@@ -29,7 +29,7 @@ const AlertWithTransition = ({ alert, close }: AlertComponentProps) => {
     close()
   }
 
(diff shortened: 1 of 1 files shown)


raineorshine (approved):

raineorshine:
Thanks!

I added a bit more explanation to the comment. It helps to explain both what the code does and why it is necessary.

Issue #1284, opened 2021-06-16
Strip html when exporting as Plain Text with excludeMeta true


raineorshine:
## Steps to Reproduce

```
- Hello <b>world</b>
```

Export the thought in Plain Text with Lossless unchecked (which will call exportContext with `excludeMeta: true`)

## Current Behavior

```
- Hello <b>world</b>
```

## Expected Behavior

```
- Hello world
```

ruchijn:
Shouldn't this the HTML tag be controlled by `Plain text` vs `HTML` option and not by `Lossless`? 🤔

raineorshine:
Exactly

PR #1296, opened 2021-06-20
Fix for stripping HTML tag when exporting as plain text

ruchijn:
Fix: #1284

### DIFF
--- a/src/selectors/__tests__/exportContext.ts
+++ b/src/selectors/__tests__/exportContext.ts
@@ -1,5 +1,6 @@
 import { initialState, reducerFlow } from '../../util'
 import { setCursor, importText } from '../../reducers'
+import { EMPTY_SPACE } from '../../constants'
 import exportContext from '../exportContext'
 
 it('meta and archived thoughts are included', () => {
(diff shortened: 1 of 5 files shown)


src/selectors/__tests__/exportContext.ts
```diff
@@ -97,3 +98,44 @@ it('meta is excluded but archived is included', () => {
...
+
+  const steps = [
+    importText({ text }),
+    setCursor({ path: [{ value: 'a', rank: 0 }] })
+  ]
+
+  // run steps through reducer flow and export as plaintext for readable test
+  const stateNew = reducerFlow(steps)(initialState())
+
+  const exported = exportContext(stateNew, ['a'], 'text/plain', { excludeMeta: true })
```
raineorshine, line 114: Why `excludeMeta: true`? There are no meta tags in this test case.

src/selectors/exportContext.ts
```diff
@@ -69,8 +76,11 @@ export const exportContext = (state: State, context: Context, format: MimeType =
     ? `${childrenPrefix}\n${childrenFiltered.map(exportChild).join('\n')}${childrenPostfix}${format === 'text/html' ? indent === 0 ? tab0 : tab1 : ''}`
     : ''
 
-  const text = `${tab0}${linePrefix}${head(context)}${exportedChildren && format === 'text/html' ? tab1 : ''}${exportedChildren}${linePostfix}`
+  let text = `${tab0}${linePrefix}${head(context)}${exportedChildren && format === 'text/html' ? tab1 : ''}${exportedChildren}${linePostfix}`
 
+  if (format === 'text/plain') {
+    text = stripHTMLTag(text)
+  }
```
raineorshine, line 83: It would be better to use two separate variables instead of overriding `text`. We try to avoid mutations as much as possible.

src/util/__tests__/importText.ts
```diff
@@ -1,25 +1,18 @@
...
-  // remove root, de-indent (trim), and append newline to make tests cleaner
-  const exportedWithoutRoot = exported.slice(exported.indexOf('\n'))
-    .split('\n')
-    .map(line => line.slice(2))
-    .join('\n')
-    + '\n'
+  const exported = exportContext(stateNew, [HOME_TOKEN], isHTML ? 'text/html' : 'text/plain')
 
-  return exportedWithoutRoot
+  return removeHome(exported)
```
raineorshine, line 15: Thanks for catching this.

raineorshine (approved):
Thanks for fixing all the tests! I didn't realize the tests were tightly coupled to that functionality. The `isHTML` argument is a bit awkward but I don't see a different way to do it and I agree it should probably be explicit.

src/selectors/exportContext.ts
ruchijn: I see you approved and merged. I will update this in my next PR.

raineorshine: Yeah, if they are minor points sometimes I'll approve and merge and just mention it for future reference. You don't have to fix it this round.

PR #1299, opened 2021-06-21
Converting Components to React Native 

brunokiafuka:
Components: 
- Toolbar
- FeedbackModal
- Modal wrapper


raineorshine:
I only had time to look briefly, but I am getting the gist of it! The native versions of components make a lot of sense.

Could you explain the plan with Redux state? It looks like `useSelector` and `mapStateToProps` has been removed temporarily. Are those going to be added back in? It seems like there maybe a lot of duplication between native and web in shortcuts and action-creators.

Second question is why not shim `localStorage` instead of creating two versions of every module that uses localStorage?

Thanks!

brunokiafuka:
> I only had time to look briefly, but I am getting the gist of it! The native versions of components make a lot of sense.
> 
> Could you explain the plan with Redux state? It looks like `useSelector` and `mapStateToProps` has been removed temporarily. Are those going to be added back in? It seems like there maybe a lot of duplication between native and web in shortcuts and action-creators.
> 
> Second question is why not shim `localStorage` instead of creating two versions of every module that uses localStorage?
> 
> Thanks!

1. `useSelector` and `mapStateToProps` have been removed temporarily so that we could build the interfaces, for now, they will be added as we start activating each one of them (our strategy is to build the components first and then activate the states according to the feature we are working on). 

2. Shortcuts duplications we are using `react-native-svg `to build the SVG icons used on the shortcuts and this required us to recreate them. 

3.Action creators, just due to web dependency but we have done in two files thus far one was because of the localStorage which we can solve and the other one due to `window.scrollTo(0, 0)` which we should be able to shim.

4. for localStorage we can shim it but it's not a great idea to just be setting a global variable in React native run time the way that it is accessed in the browser.

Another issue is that in React Native we use AsyncStorage which is asynchronous whereas localStorage is synchronous, so we propose creating a gateway where we can abstract the local storage implementations and then export that localStorage implementation everywhere needed so don't have to recreate a specific file abstracting the implementation.
This will involve some changes to the existing files to import the storage gateway and make the calls async.

src/util/publishMode.native.ts
```diff
@@ -0,0 +1,2 @@
+/** Returns the publish mode query string. */
+export const publishMode = () => true
```
raineorshine, line 2: You'll want to hardcode this to `false` on native.

src/util/initEvents.native.ts
```diff
@@ -0,0 +1,134 @@
+import _ from 'lodash'
+import { Store } from 'redux'
+import { inputHandlers /* isGestureHint */ } from '../shortcuts'
+import * as db from '../data-providers/dexie'
+import { clearSelection, isRoot, pathToContext } from '.'
+import { State } from './initialState'
+import { decodeThoughtsUrl, pathExists } from '../selectors'
+import { /* alert */ error, scrollCursorIntoView, setCursor, toggleTopControlsAndBreadcrumbs } from '../action-creators'
```
raineorshine, line 8: FYI the `alert` action-creator is not a browser/native alert but an in-app alert system, as you're probably aware. The only web dependency it has is `window.setTimeout`.

raineorshine (commented):
Great, I love it! I did a more thorough review. With the help of your answers, I am completely on board.

> 1. `useSelector` and `mapStateToProps` have been removed temporarily so that we could build the interfaces, for now, they will be added as we start activating each one of them (our strategy is to build the components first and then activate the states according to the feature we are working on).

Got it! That sounds good. Better to focus on the components first than try to deal with state at the same time. The large majority of the web components and shortcuts will not change, so it shouldn't be hard to bring any update dlogic back into native.

> 2\. Shortcuts duplications we are using `react-native-svg `to build the SVG icons used on the shortcuts and this required us to recreate them.

I see, thanks. We could also factor out the shortcut icons into separate components. But since you are making native versions of shortcuts anyway to disable Redux state, it doesn't really matter. Is that your reasoning, too?

> 3.Action creators, just due to web dependency but we have done in two files thus far one was because of the localStorage which we can solve and the other one due to `window.scrollTo(0, 0)` which we should be able to shim.

👍

> 1. for localStorage we can shim it but it's not a great idea to just be setting a global variable in React native run time the way that it is accessed in the browser.
> 
> Another issue is that in React Native we use AsyncStorage which is asynchronous whereas localStorage is synchronous, so we propose creating a gateway where we can abstract the local storage implementations and then export that localStorage implementation everywhere needed so don't have to recreate a specific file abstracting the implementation.
> This will involve some changes to the existing files to import the storage gateway and make the calls async.

I totally agree. We should move to an asynchronous version that works across web and mobile.

src/util/publishMode.native.ts
brunokiafuka: Done

src/util/initEvents.native.ts
brunokiafuka: Thanks @raineorshine.

src/browser.ts
```diff
@@ -1,5 +1,5 @@
 /** Defines client/browser-specific variables that do not change over the course of a session. */
-
+import 'react-native-get-random-values'
```
Foosballfan, line 2: we import this in quite a few places, should probably only be once

brunokiafuka: I removed the one on the app entry, the others are pretty much needed.

src/components/NavBar.native.tsx
```diff
@@ -0,0 +1,52 @@
...
+    showBreadcrumbs,
+  }
+}
+
+/** A navigation bar that contains a link to home and breadcrumbs. */
+const NavBar = ({
+  cursor,
+  position,
+  showBreadcrumbs,
+}: {
```
Foosballfan, line 27: nicer as interface

src/components/Toolbar.native.tsx
```diff
@@ -0,0 +1,60 @@
...
+const Toolbar = ({ dark = true, fontSize = 12, toolbarOverlay = null, scrollPrioritized = true, showTopControls = true }/* : ReturnType<typeof mapStateToProps> */) => {
+
+  const fillColor = dark ? 'white' : 'black'
+
+  const shortcutIds = TOOLBAR_DEFAULT_SHORTCUTS
+
+  const keyExtractor = useCallback((item: string) => item, [])
+  const renderItem: ListRenderItem<string> = useCallback(({ item: id, index }) => {
+
+    if (id === 'delete') return null
```
Foosballfan, line 30: a few ids hanging around as strings that would be safer as enums

src/components/Toolbar.native.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <SVG key={name} fill={fillColor} size={35} />
+    </TouchableOpacity>
+
+  }, [])
+
+  /**********************************************************************
+   * Render
+   **********************************************************************/
+  return (
+    <View style={{ flexDirection: 'row', alignContent: 'center' }}>
```
Foosballfan, line 46: do we have a plan for styling?

brunokiafuka: We could add our inline style linting rule, but there are a quite good number of components that have inline style. A possible solution would be to create a rule that only checks on files that finish with `*.native.ts`.

src/data-providers/dexie.native.ts
```diff
@@ -0,0 +1,227 @@
...
+export interface Log {
+  created: Timestamp
+  message: string
+  stack?: any
+}
+
+const db = new Dexie('EM') as EM
+
+/** Initializes the EM record where helpers are stored. */
+const initHelpers = async () => {
```
Foosballfan, line 72: there is a lot of duplicated code between this and non native, can we abstract just the code which is unique per platform?

src/util/initialState.native.ts
```diff
@@ -0,0 +1,259 @@
...
+  recentlyEdited: RecentlyEditedTree
+  pendingDeletes?: { context: Context; child: Child }[]
+  pendingEdits?: editThoughtPayload[]
+  descendantMoves?: { pathOld: Path; pathNew: Path; pathToPull?: Path }[]
+  pendingPulls?: { path: Path }[]
+  updates?: Index<string>
+  pendingLexemes?: Index<boolean>
+}
+
+export interface State {
```
Foosballfan, line 53: all this state in React 😢

brunokiafuka:
> I see, thanks. We could also factor out the shortcut icons into separate components. But since you are making native versions of shortcuts anyway to disable Redux state, it doesn't really matter. Is that your reasoning, too?

Do you mean to separate out the SVG components and have a single file where we implement the logic? - If so, would like me to do it or will your team to it?

raineorshine:
> > I see, thanks. We could also factor out the shortcut icons into separate components. But since you are making native versions of shortcuts anyway to disable Redux state, it doesn't really matter. Is that your reasoning, too?
> 
> @raineorshine - Do you mean to separate out the SVG components and have a single file where we implement the logic? - If so, would like me to do it or will your team to it?

Yes. So instead of just `shortcuts/archive` with an internal `Icon` component`, we would have `shortcuts/archive` import `components/icons/archive`. Then we could add `components/icons/archive.native`.

But if you are disabling the Redux state and thus forking `shortcuts/archive` anyway, then it doesn't seem necessary at this time to create a separate component file. If the only change in `shortcuts/archive` were the icon, then it would make sense to factor it out and thus fork only the component rather than the shortcut. So I think you can continue as you are doing, just wanted to make sure we were thinking about it the same way.

brunokiafuka:
> > > I see, thanks. We could also factor out the shortcut icons into separate components. But since you are making native versions of shortcuts anyway to disable Redux state, it doesn't really matter. Is that your reasoning, too?
> > 
> > 
> > @raineorshine - Do you mean to separate out the SVG components and have a single file where we implement the logic? - If so, would like me to do it or will your team to it?
> 
> Yes. So instead of just `shortcuts/archive` with an internal `Icon` component`, we would have `shortcuts/archive`import`components/icons/archive`. Then we could add `components/icons/archive.native`.
> 
> But if you are disabling the Redux state and thus forking `shortcuts/archive` anyway, then it doesn't seem necessary at this time to create a separate component file. If the only change in `shortcuts/archive` were the icon, then it would make sense to factor it out and thus fork only the component rather than the shortcut. So I think you can continue as you are doing, just wanted to make sure we were thinking about it the same way.

It sounds like a great idea, should we plan the refactor once we are good with Redux?

raineorshine:
> It sounds like a great idea, should we plan the refactor once we are good with Redux?

Yeah, it's an easy job to factor them out.

Foosballfan (approved):

src/constants.ts
```diff
@@ -330,7 +331,30 @@ export const ALLOWED_FORMATTING_TAGS = ['b', 'i', 'u', 'em', 'strong', 'span']
...
+  DELETE = 'delete',
+  TOGGLE_CONTEXT_VIEW = 'toggleContextView',
+  PROSE_VIEW = 'proseView',
+  // TOGGLESPLITVIEW = // 'toggleSplitView',
+  SPLIT_SENTENCES = 'splitSentences',
+  SUBCATEGORIZE_ONE = 'subcategorizeOne',
+  SUBCATEGORIZE_ALL = 'subcategorizeAll',
+  TOGGLE_HIDDEN_THOUGHTS = 'toggleHiddenThoughts',
+  EXPORT_CONTEXT = 'exportContext',
+}
```
raineorshine, line 360: This would need to be manually maintained in parallel with the shortcuts index. If you want type safe shortcut ids without the overhead of a parallel enum, you can derive the type from the keys of https://github.com/cybersemics/em/blob/dev/src/shortcuts/index.ts.

Case in point, this list includes just the default shortcuts, not the full list of shortcuts.

(I'm aware that there are other instances in this file that do not follow this advice.)

src/data-providers/win.native.ts
```diff
@@ -0,0 +1,12 @@
+import SQLite from 'react-native-sqlite-2'
+import setGlobalVars from 'indexeddbshim/dist/indexeddbshim-noninvasive'
+
+const win: { indexedDB?: any; IDBKeyRange?: any } = {}
```
Foosballfan, line 4: winning

src/constants.ts
brunokiafuka: @raineorshine, the derived keys don't match the ids. ie.: the delete is named `deleteEmptyThoughtOrOutdent` and the id is `delete`.

raineorshine: They are two separate shortcuts. The id `delete` should be called `archive`.

https://github.com/cybersemics/em/blob/db627f09a57d5def2a053098978b12bdd4b9c696/src/shortcuts/archive.tsx#L69-L78

brunokiafuka: Fixed, by creating the native archive component.

PR #1302, opened 2021-06-25
Add updated emoji regex

### LINKED ISSUE #1289: Emoji incorrectly altered
Testes on iOS Safari

## Steps to Reproduce

1. Create a thought with the value “👩🏻‍🏫”
2. Close the keyboard

## Current Behavior

- 👩🏻 ‍🏫

## Expected Behavior

Should render and preserve emoji as entered

shresthabijay:
fixes #1289

# Problem with the previous emoji regex.

- Detected components of emojis that are not actually emojis.
- Couldn't properly detect emojis with variant selector `\ufeof`.
- Could not detect ZWJ sequenced emojis like 👩‍👩‍👦‍👦 👨‍👦‍👦 as a single emojis.

# Solution

 New emoji regex is based on `emoji-regex/text.js` https://github.com/mathiasbynens/emoji-regex. The original regex detects 1234567890*&^’%$#@!-+\ characters as emoji too. So I had to modify it to fit our use case.

 Text-based `emoji-regex/text.js` is used instead of more standardized `emoji-regex/RGI_Emoji.js` because of the following problem.

 Some IOS and macOS emojis don't have variant selector at the end but it is required in comparison to the standardized RGI emojis. So RGI based regex doesn't detect them as emojis.

However, there is a downside to this. This regex also matches emoji that are not `Emoji_Presentation` symbols and that aren’t forced to render as emoji by a variation selector. i.e It matches symbols like '❤', '❣', '©', '☯', '®'. However it doesn't detect all of these symbols. Only some portion of it.

But I think it won't be much of an issue compared to the issues of the prev implementation.

# Changes

- Added new emoji regex.
- Added more tests related to emoji regex and its related utils.
- Updated `addEmojiSpace`.


raineorshine (approved):
Wonderfully documented and perfectly executed. Thank you for getting to the bottom of it!

Issue #1230, opened 2021-06-09
Dim disabled toolbar buttons 


raineorshine:
Toolbar buttons should be dimmed if `canExecute` is false. 

For example: Archive should be dimmed when the `cursor` is null, since the command cannot be executed.

ruchijn:
I was looking at this. But it seems even when `cursor` is `null`, `canExecute` is still set to true. It is not checking for `cursor` value. I think I might need to connect with you on this.

raineorshine:
You're right. In the case of the `archive` shortcut, `canExecute` does not consider the `cursor` at all, though it should:

https://github.com/cybersemics/em/blob/f23343267e25271d900b8e741fb4ced14acaaac5/src/shortcuts/archive.tsx#L76

raineorshine:
I went through all the shortcuts and made sure `canExecute` was correct. You should be good to know now.

PR #1304, opened 2021-06-26
Dim disabled toolbar buttons

ruchijn:
Fix: #1230


src/components/ModalHelp.tsx
```diff
@@ -116,7 +116,7 @@ const ModalHelp = ({ tutorialStep, showQueue, dispatch, enableLatestShorcutsDiag
 
     <code>=focus</code>
     <p>Options: Normal, Zoom<br/>
-    When the cursor is on this thought, hide its parent and sibliings for additional focus. Excellent for study time or when you have had too much coffee.</p>
+    When the cursor is on this thought, hide its parent and siblings for additional focus. Excellent for study time or when you have had too much coffee.</p>
```
ruchijn, line 119: Fixing typo

src/selectors/exportContext.ts
```diff
@@ -76,14 +76,15 @@ export const exportContext = (state: State, context: Context, format: MimeType =
     ? `${childrenPrefix}\n${childrenFiltered.map(exportChild).join('\n')}${childrenPostfix}${format === 'text/html' ? indent === 0 ? tab0 : tab1 : ''}`
     : ''
 
-  let text = `${tab0}${linePrefix}${head(context)}${exportedChildren && format === 'text/html' ? tab1 : ''}${exportedChildren}${linePostfix}`
+  const text = `${tab0}${linePrefix}${head(context)}${exportedChildren && format === 'text/html' ? tab1 : ''}${exportedChildren}${linePostfix}`
+
+  const strippedHTMLText = stripHTMLTag(text)
+
+  const finalizedText = format === 'text/plain' ? strippedHTMLText : text
```
ruchijn, line 83: Changes made as per review feedback https://github.com/cybersemics/em/pull/1296#discussion_r654996121

src/util/isShortcutExecutable.ts
```diff
@@ -0,0 +1,5 @@
+import { State } from '../util/initialState'
```
ruchijn, line 1: Created a common util that can be reused rather than having the same logic in multiple files. This way if needed to modify in future, it will just one file change.

src/shortcuts/archive.tsx
```diff
@@ -73,7 +73,8 @@ const archiveShortcut: Shortcut = {
   gesture: 'ldl',
   svg: Icon,
   keyboard: { key: Key.Backspace, shift: true, meta: true },
-  canExecute: getState => isDocumentEditable() && !!getState().cursor,
+  isActive: getState => isShortcutExecutable(getState),
+  canExecute: getState => isShortcutExecutable(getState),
```
raineorshine, line 77: Instead of copying `canExecute` to `isActive`, change the Toolbar logic to fallback to `canExecute` when there is no `isActive`.

https://github.com/cybersemics/em/blob/862aa3b7e97f5904999b1abb9fee3640880d5d74/src/components/Toolbar.tsx#L215

That way there is less duplication and fewer required fields for new shortcuts.

src/shortcuts/note.ts
```diff
@@ -69,7 +69,7 @@ const noteShortcut: Shortcut = {
     const state = getState()
     const { cursor } = state
     const context = pathToContext(cursor ? simplifyPath(state, cursor) : HOME_PATH)
-    return attribute(state, context, '=note') != null
+    return attribute(state, context, '=note') != null || isShortcutExecutable(getState)
```
raineorshine, line 72: This would destroy the meaning of the existing `isActive`, as `isShortcutExecutable` will always return true when the cursor is valid, regardless of `attribute`.

We should either use the existing behavior (dim when there is no note) or the new functionality (dim when it is not possible to add a note). I would keep the existing behavior.

Some shortcuts will dim when they are not executable, and some shortcuts will dim when their respective attribute is disabled on the cursor. This is currently ambiguous to the user, which is a tad confusing. Yet it seems wrong to not dim a shortcut when it is not executable.

If that doesn't make sense that's okay. Just know that we should leave existing `isActive` methods as-is.

raineorshine (changes requested):

src/util/isShortcutExecutable.ts
ruchijn: Hmmm.... The places that I have modified does the same check. Hence I moved it to common file. If any shortcut does a different validation based on `isDocumentEditable` and `cursor`, I haven't touched those. Do you still think I should revert the changes and have the validation in respective shortcut file?

src/shortcuts/note.ts
ruchijn: Currently non-executable icon is also clickable. Should we look into making them non-clickable if they can not be executed?

src/util/isShortcutExecutable.ts
raineorshine: Yes, reverting would be best.

To offer a little more explanation: Your approach is functionally correct, but ill-advised. I don't think colocation of code is a sufficient criterion for abstraction. In this case, grouping them into a function would suggest that they belong together, rather than that they are together for ad hoc reasons and fully independent. Hope that helps!

src/shortcuts/note.ts
raineorshine: Could you explain what you mean by clickable? If `canExecute` returns false, clicking the toolbar button will do nothing. It will produce an animation though which I think should be disabled.

ruchijn: What I mean is if action related to icon is not executable, user can still click on it and expect something to happen. Based on current set up, when user clicks on it, we will show them the animation, but nothing will happen which might be confusing for them. So, if toolbar icon is non-executable, we should disable it and stop the animation. Also on desktop, we should not have cursor as pointer. When cursor is changed to pointer, it means it is a link and user can click on it and expect something to happen.

src/util/isShortcutExecutable.ts
ruchijn: Done.

src/components/Toolbar.tsx
```diff
@@ -190,7 +190,8 @@ const Toolbar = ({ dark, fontSize, toolbarOverlay, scrollPrioritized, showTopCon
           >
             <span id='left-arrow' className={leftArrowElementClassName}><TriangleLeft width={arrowWidth} height={fontSize} fill='gray' /></span>
             {shortcutIds.map(id => {
-              const { name, svg, exec, isActive } = shortcutById(id)!
+              const { name, svg, exec, isActive, canExecute } = shortcutById(id)!
+              const isToolbarExecutable = isActive ? !isActive || isActive(store.getState) : !canExecute || canExecute(store.getState)
```
raineorshine, line 194: `!isActive` is always false in the case that `isActive` is true, so it should be omitted.

raineorshine (changes requested):
> What I mean is if action related to icon is not executable, user can still click on it and expect something to happen. Based on current set up, when user clicks on it, we will show them the animation, but nothing will happen which might be confusing for them. So, if toolbar icon is non-executable, we should disable it and stop the animation. Also on desktop, we should not have cursor as pointer. When cursor is changed to pointer, it means it is a link and user can click on it and expect something to happen.

Yes, I totally agree! That would be great if you can do that, too.

src/components/Toolbar.tsx
raineorshine: Done in 2a2a12201e

raineorshine: Done in 2a2a12201e

src/components/Toolbar.tsx
```diff
@@ -192,13 +192,18 @@ const Toolbar = ({ dark, fontSize, toolbarOverlay, scrollPrioritized, showTopCon
...
               const SVG = svg as React.FC<Icon>
               return (
                 <div
                   key={name}
                   id={id}
-                  style={pressingToolbarId === id ? { paddingTop: '10px' } : {}}
+                  style={{
+                    paddingTop: pressingToolbarId === id ? '10px' : '',
+                    cursor: isButtonActive ? 'pointer' : 'default',
+                    pointerEvents: isButtonActive ? 'auto' : 'none',
```
raineorshine, line 205: This will incorrectly disable the button when the shortcut is executable but inactive, such as when the Table View is not active on a thought, but can be activated. 

Instead, we want to disable the button only if it cannot be executed.

raineorshine (changes requested):

src/components/Toolbar.tsx
ruchijn: But if the button is executable, shouldn't it be active too?

raineorshine: No, there are some commands that show active if a meta attribute is on the thought. For example:

```js
- a
  - =note
    - hello
- b
```

If the cursor is on `a`, then the Note button will be *executable* and *active*. If the cursor is on `b`, the Note button will be *executable* and *inactive*.

It is confusing because it is ambiguous. I am overloading the `isButtonActive` state to handle two different things. I tried to explain it here, but it is understandably confusing: https://github.com/cybersemics/em/pull/1304#discussion_r659219440

It's like `isButtonActive` reflects if `isActive` returns true for the given shortcut, or if `isActive` is not defined at all, with the added exception that if the shortcut is not executable, it should not show as active.

ruchijn: I re-read your comment and tried to follow it in app. It makes sense. I will make the change. Thanks

src/components/Toolbar.tsx
```diff
@@ -190,20 +190,26 @@ const Toolbar = ({ dark, fontSize, toolbarOverlay, scrollPrioritized, showTopCon
           >
             <span id='left-arrow' className={leftArrowElementClassName}><TriangleLeft width={arrowWidth} height={fontSize} fill='gray' /></span>
             {shortcutIds.map(id => {
-              const { svg, exec, isActive } = shortcutById(id)!
+              const { svg, exec, isActive, canExecute } = shortcutById(id)!
+              const isButtonActive = !isActive || isActive(store.getState)
+              const isButtonExecutable = !canExecute || canExecute(store.getState)
```
raineorshine, line 195: This is what I was going for: `isButtonActive && isButtonExecutable`

raineorshine (approved):

PR #1306, opened 2021-06-29
Prevent moving some meta thoughts on subCategorizeAll

### LINKED ISSUE #1244: subcategorizeAll should not move some meta attributes
## Steps to Reproduce

```js
- a
  - =note
    - My Note
  - c
```

1. Set the cursor on `a/c`. 
2. Activate `subcategorizeAll`. 

## Current Behavior

`=note` is moved to the new empty thought. 

## Expected Behavior

The following meta attributes should not move with `subcategorizeAll`:

- =archive
- =bullet
- =focus
- =label
- =note
- =pin
- =publish

shresthabijay:
fixes #1244

# Changes

- Filter out meta exceptions on `subCategorizeAll`.
- Update tests.

### DIFF
--- a/src/reducers/__tests__/subCategorizeAll.ts
+++ b/src/reducers/__tests__/subCategorizeAll.ts
@@ -1,4 +1,4 @@
-import { HOME_TOKEN } from '../../constants'
+import { HOME_PATH, HOME_TOKEN } from '../../constants'
 import { initialState, reducerFlow } from '../../util'
 import { exportContext } from '../../selectors'
 
@@ -7,7 +7,8 @@ import newSubthought from '../newSubthought'
 import newThought from '../newThought'
 import subCategorizeAll from '../subCategorizeAll'
 import setCursor from '../setCursor'
-import cursorBack from '../cursorBack'
+import importText from '../importText'
+import setCursorFirstMatch from '../../test-helpers/setCursorFirstMatch'
 
 it('subcategorize multiple thoughts', () => {
 
@@ -89,28 +90,50 @@ it('set cursor on new empty thought', () => {
 
 })
 
-it('move all visible and hidden thoughts into a new empty thought after subcategorizeAll', () => {
+it('move all non meta thoughts and only allowed meta thoughts into new empty thought after subCategorizeAll', () => {
+
+  const text = `
+  - a
+    - =archive
+    - =bullet
+    - =focus
+    - =label
+    - =note
+    - =pin
+    - =publish
+    - =style
+    - =view
+    - c
+    - d
+    - e`
 
   const steps = [
-    newThought('b'),
-    newSubthought('=archive'),
-    newThought('c'),
-    newSubthought('d'),
-    cursorBack,
-    subCategorizeAll
+    importText({
+      text,
+      path: HOME_PATH
+    }),
+    setCursorFirstMatch(['a', 'c']),
+    subCategorizeAll,
   ]
 
-  // run steps through reducer flow
+  // run steps through reducer flow and export as plaintext for readable test
   const stateNew = reducerFlow(steps)(initialState())
-
   const exported = exportContext(stateNew, [HOME_TOKEN], 'text/plain')
 
-  const expectedOutput = `- ${HOME_TOKEN}
-  - b
-    -${' '}
-      - =archive
-      - c
-        - d`
 
-  expect(exported).toEqual(expectedOutput)
+  expect(exported).toBe(`- ${HOME_TOKEN}
+  - a
+    - ${''/* prevent trim_trailing_whitespace */}
+      - =style
+      - =view
+      - c
+      - d
+      - e
+    - =archive
+    - =bullet
+    - =focus
+    - =label
+    - =note
+    - =pin
+    - =publish`)
 })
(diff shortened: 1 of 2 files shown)


shresthabijay:

src/reducers/__tests__/subCategorizeAll.ts
```diff
@@ -89,28 +90,42 @@ it('set cursor on new empty thought', () => {
 
 })
 
-it('move all visible and hidden thoughts into a new empty thought after subcategorizeAll', () => {
+it('move all non meta thoughts and only allowed meta thoughts into new empty thought after subCategorizeAll', () => {
+
+  const metaExceptions = SUBCATEGORIZE_ALL_META_EXCEPTIONS.reduce(
+    (acc, meta, index) => (acc + `    - ${meta}${index === (SUBCATEGORIZE_ALL_META_EXCEPTIONS.length - 1) ? '' : '\n'}`)
+  , '')
```
raineorshine, line 97: I would hardcode the exceptions into the test. We want tests to be as concrete as possible. Principles of abstraction that we use in application architecture do not apply in the test environment. It forces each case to be considered individually, which is what we want to do if there are new meta attributes.

Of course, we do generalize tests at times but we have to be careful to avoid complexity and abstraction when possible.

src/reducers/subCategorizeAll.ts
```diff
@@ -42,6 +42,9 @@ const subCategorizeAll = (state: State) => {
   const children = getChildrenRanked(state, pathToContext(simplifyPath(state, path)))
   const pathParent = cursor.length > 1 ? cursorParent : HOME_PATH
 
+  // filter out meta children that should not be moved
+  const filteredChildren = children.filter(child => !SUBCATEGORIZE_ALL_META_EXCEPTIONS.includes(child.value))
```
raineorshine, line 46: Once the exceptions are hardcoded in the tests, we can move them out of constants into this module.

src/reducers/__tests__/subCategorizeAll.ts
```diff
@@ -89,28 +90,42 @@ it('set cursor on new empty thought', () => {
...
+  expect(exported).toBe(`- ${HOME_TOKEN}
+  - a
+    - ${''/* prevent trim_trailing_whitespace */}
+      - =style
+      - =view
+      - c
+      - d
+      - e
+${metaExceptions}`)
+})
```
raineorshine, line 131: Do you have your editor set up to use `.editorconfig`? It should not be allowing you to save files without final newlines. Unless that's a Windows thing.

raineorshine (changes requested):

src/reducers/__tests__/subCategorizeAll.ts
shresthabijay: No, I am using Macbook. I think the `eslint-config-prettier` is turning off some important eslint rules from `eslint-config-raine` and possibly other plugins too. All formatting related rules should be handled by the `prettier` now.

src/reducers/subCategorizeAll.ts
shresthabijay: Sure will change this.

src/reducers/__tests__/subCategorizeAll.ts
shresthabijay: Understood.

src/reducers/__tests__/subCategorizeAll.ts
raineorshine: I don't think prettier has the equivalent of `insert_final_newline` from [EditorConfig](https://editorconfig.org/). You should install EditorConfig in your editor if you don't have it already.

This isn't related to eslint rules, but I am working with App Sapiens to make sure https://github.com/prettier/eslint-config-prettier is not a problem. It is disabling rules but only ones that Prettier can fix automatically.

raineorshine: This issue seems to suggest that Prettier does this automatically: https://github.com/prettier/prettier/issues/2575

However it may depend on your editor settings.

src/reducers/subCategorizeAll.ts
```diff
@@ -4,6 +4,16 @@ import { State } from '../util/initialState'
...
 
+const SUBCATEGORIZE_ALL_META_EXCEPTIONS = [
+  '=archive',
+  '=bullet',
+  '=focus',
+  '=label',
+  '=note',
+  '=pin',
+  '=publish',
+]
```
raineorshine, line 15: Can you use an object for this? The array is O(n) lookup while an object is O(1).

```ts
const SUBCATEGORIZE_ALL_META_EXCEPTIONS = {
  '=archive': true,
  '=bullet': true,
  ...
}
```

```ts
const filteredChildren =  childen.filter(child => child.value in SUBCATEGORIZE_ALL_META_EXCEPTIONS)
```

raineorshine (changes requested):

src/reducers/__tests__/subCategorizeAll.ts
shresthabijay: > I don't think prettier has the equivalent of `insert_final_newline` from [EditorConfig](https://editorconfig.org/). You ~should~ could install EditorConfig in your editor ~if you don't have it already.~ if it doesn't have a setting to automatically add a newline at the end of a file.

Sure.

PR #1307, opened 2021-06-29
Ref: custom local storage implementation.

brunokiafuka:
In this PR we re-implemented the local storage functionalities so we can support them on mobile (React Native) and web, avoiding having to duplicate all the files/scripts for both mobile and web. 

> The mobile implementation is asynchronous. [Read More](https://react-native-async-storage.github.io/async-storage/docs/usage/)


src/util/localStorage.native.ts
```diff
@@ -0,0 +1,26 @@
...
+  async clear(): Promise<void> {
+    await AsyncStorage.clear()
+  },
+
+  async getItem(key: string): Promise<string | null> {
+    return AsyncStorage.getItem(key)
+  },
+
+  async removeItem(key: string): Promise<void> {
+    storage = _.omit(storage, key)
```
raineorshine, line 16: `_.omit` is O(n). Can you do `delete storage.key` or do you need a new object reference?

src/util/localStorage.native.ts
```diff
@@ -0,0 +1,26 @@
...
+    await AsyncStorage.removeItem(key)
+  },
+
+  async setItem(key: string, value: string): Promise<void> {
+    storage[`${key}`] = value
+    await AsyncStorage.setItem(key, value)
```
raineorshine, line 22: Can you explain the reason for having two different assignments, one to a plain object and one to `AsyncStorage`?

src/util/localStorage.ts
```diff
@@ -0,0 +1,24 @@
+import _ from 'lodash'
+import { IStorage } from '../types'
+
+let storage: IStorage = {
+  clear(): void {
+    localStorage.clear()
+  },
+
+  getItem(key: string): string | null {
```
raineorshine, line 9: This `getItem` returns `string | null` while the native version returns `Promise<string | null>`. I think they should both return `Promise<string | null>`. If web and native have a different interface, it defeats the purpose of abstracting away the `localStorage` implementation. Same for the other functions.

src/util/localStorage.ts
```diff
@@ -0,0 +1,24 @@
+import _ from 'lodash'
+import { IStorage } from '../types'
+
+let storage: IStorage = {
```
raineorshine, line 4: Let's name this file `storage.ts`.

We have a convention of naming the file after the default export when possible.

src/action-creators/fontSize.ts
```diff
@@ -1,11 +1,14 @@
...
-  setTimeout(() => {
-    localStorage.fontSize = value
-  })
-}
+const fontSize =
+  (value: number): Thunk =>
+  (dispatch, getState) => {
+    dispatch({ type: 'fontSize', value })
+    setTimeout(() => {
+      storage.fontSize = value
```
raineorshine, line 10: This doesn't work on web. It saves to `storage.fontSize`, not the browser's `localStorage`. I think you need `storage.setItem('fontSize', value)`.

You can reproduce this by increasing the font size using the links in the footer and then refreshing the page.

raineorshine (changes requested):

src/util/localStorage.native.ts
brunokiafuka: Refactored using `delete`. didn't use it earlier because eslint was complaining 😅

src/util/localStorage.ts
brunokiafuka: 💯, will refactor it.

src/util/localStorage.native.ts
brunokiafuka: There were some places where you were accessing the localStorage value using the storage object. Like: ` storage['Settings/' + value]`. I think I will just refactor it to access using the `getItem` method so we avoid double assignment.

raineorshine: Yes. `localStorage` is special in that assignment is equivalent to `setItem`. We should use `getItem` and `setItem` instead.

src/util/storage.native.ts
```diff
@@ -0,0 +1,21 @@
+import AsyncStorage from '@react-native-async-storage/async-storage'
+
+const storage = {
```
raineorshine, line 3: Can you use `export const storage` instead of exporting `{ storage }`? That would be a bit more standard.

src/util/localStorage.ts
brunokiafuka: we just made async store sync to prevent changing to async call the whole app.

src/util/storage.native.ts
brunokiafuka: Fixed

src/util/localStorage.native.ts
brunokiafuka: Fixed.

Foosballfan (approved):

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable fp/no-class */
+import AsyncStorage from '@react-native-async-storage/async-storage'
+
+/**
+ *
+ */
```
raineorshine, line 6: Missing JSDOC

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,78 @@
...
+function handleError(func: string, param?: string): Promise<string> {
+  let message
```
raineorshine, line 8: I'm not sure why the linter is not catching this, but we try to avoid `let` and mutations. This expression can easily be rewritten with a ternary statement.

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (!param) {
+    message = func
+  } else {
+    message = `${func}() requires at least ${param} as its first parameter.`
+  }
+  console.warn(message) // eslint-disable-line no-console
+  return Promise.reject(message)
```
raineorshine, line 15: You can throw an Error instead of returning a rejected Promise.

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ *
+ */
```
raineorshine, line 20: Missing JSDOC

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,78 @@
...
+class SyncStorage {
+  data: Map<string, string> = new Map()
+
+  loading = true
+
+  async init(): Promise<void> {
+    await AsyncStorage.getAllKeys().then((keys: string[]) =>
```
raineorshine, line 27: This would be cleaner with `async/await` (and `try/catch` as needed).

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,78 @@
...
+      AsyncStorage.multiGet(keys).then((data: any): any[] => {
+        data.forEach(this.saveItem.bind(this))
+
+        return [...this.data]
```
raineorshine, line 31: I don't think the return value goes anywhere. Did you mean to return it from `init`, or is not used?

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable fp/no-class */
```
raineorshine, line 1: We would prefer to avoid classes on this project when possible. It's okay this time.

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,78 @@
...
+ *
+ */
+function handleError(func: string, param?: string): Promise<string> {
+  let message
+  if (!param) {
+    message = func
+  } else {
+    message = `${func}() requires at least ${param} as its first parameter.`
+  }
+  console.warn(message) // eslint-disable-line no-console
```
raineorshine, line 14: `warn` is [set to allow]() so I am not sure why the linter is erroring here.

https://github.com/cybersemics/em/blob/22ca9a103482a38ebb8e0e6cfc205b35ce89c14d/package.json#L33

raineorshine (changes requested):

src/util/nativeStorageHelper.ts
```diff
@@ -1,52 +1,39 @@
...
-
-        return [...this.data]
-      }),
-    )
+    const keys = await AsyncStorage.getAllKeys();
+    const data = await AsyncStorage.multiGet(keys) as string[][];
+    data.forEach(this.saveItem.bind(this));
 
+    // eslint-disable-next-line no-unused-expressions
+    [...this.data];
```
raineorshine, line 21: Sorry, I'm still not understanding this line. Could you explain?

brunokiafuka: I just removed this line, it wasn't actually doing anything 😅.

raineorshine: Phew... my Javascript understanding was about to break 😂

src/util/nativeStorageHelper.ts
```diff
@@ -0,0 +1,61 @@
...
+      [, value] = item
+    }
+
+    this.data.set(item[0], value)
+    this.loading = false
+  }
+}
+
+// eslint-disable-next-line export-default-identifier/export-default-identifier
+export default new SyncStorage()
```
raineorshine, line 61: This can be fixed without having to disable the linter:

```
const syncStorage = new syncStorage
export default syncStorage
```

raineorshine (changes requested):

raineorshine (approved):

PR #1308, opened 2021-07-01
Restore cursor selection aftter toolbar click

### LINKED ISSUE #1233: Edit mode lost after executing command
Tested on mobile Safari. Unknown if reproducible in emulation mode on desktop. 

## Steps to Reproduce

1. Create thought `a` and stay in edit mode
2. Tap the Pin button 

## Current Behavior

Edit mode is disabled

## Expected Behavior

Edit mode should stay enabled after any command. 

You can use [asyncFocus](https://github.com/cybersemics/em/blob/dev/src/util/asyncFocus.ts) on Safari before activating the command to allow the selection to be set outside of a handler.

shresthabijay:
fixes #1233

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -1,5 +1,5 @@
 import _ from 'lodash'
-import React, { useEffect, useRef, useState } from 'react'
+import React, { useEffect, useRef, useState , FocusEventHandler } from 'react'
 import { Dispatch } from 'redux'
 import { connect } from 'react-redux'
 import { unescape } from 'html-escaper'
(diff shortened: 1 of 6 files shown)


shresthabijay:

raineorshine:
This causes a regression unfortunately:

```
- a
  - [empty thought]
```

Backspace to delete the empty thought. The keyboard is still up by the caret has disappeared and typing has no effect.

Can we discuss the cause of the problem (in #1233) and the nature of the proposed solution (here)? It looks complex, and it tightly couples the Toolbar to the thought structure.

raineorshine (changes requested):
This causes a regression unfortunately:

```
- a
  - [empty thought]
```

Hit backspace to delete the empty thought. The keyboard is still up but the caret has disappeared and typing has no effect.

Can we discuss the cause of the problem (in #1233) and the nature of the proposed solution (here)? I am hesitant to tightly couple the Toolbar to the thought structure and I would like to understand the problem better.

shresthabijay:
@raineorshine I found a quite easy and simple solution. I had no idea preventing the default `mouseDown` event also prevented blur. I also have made some changes to `splitResult` because of the previous implementation. But we can keep it. Please review the changes. Thanks!

raineorshine (approved):
Excellent work!

PR #1309, opened 2021-07-01
Improve browserstack tests

onurpolattimur:
As you know, we have some trouble with our iOS e2e tests on BrowserStack. I investigated all of the issues and made some changes. There wasn't just one issue. There were different kinds of problems. Some of them were related to us and some of them were related to BrowserStack.

- One of the big problems was that our application is loaded like zoomed. Everything looks huge. Here is a captured screen video:

https://user-images.githubusercontent.com/23136437/124398007-166d0b80-dd1c-11eb-84ad-8c5ba3a15028.mp4

I looked at the all details, logs, etc and I couldn't find anything that is wrong with our codes. As a result, I contacted the BrowserStack team and they said that this can be related to the ios version and they asked me to try an iOS version that is different from iOS 14. It looks like changing the iOS version solved this problem but I am not sure because I encountered this issue only once after changing the iOS version. Also, I am waiting for a response from BrowserStack team, they are investigating this issue now. 
For now, some of our tests may fail due to this zoom issue. There is not much that we can do for now, we have to wait for BrowserStack to solve this issue because it's directly related to them. At the very beginning, it loads as shown in the picture, thus it cannot run the tests properly. 
 
image

- Another major problem was that gestures were causing the scroll. Looks like fixed for now. Here are captured screen videos that show the issue:

https://user-images.githubusercontent.com/23136437/124398047-47e5d700-dd1c-11eb-927c-53495a800952.mp4

https://user-images.githubusercontent.com/23136437/124398046-46b4aa00-dd1c-11eb-8d3b-4d10f762ca84.mp4

- ERROR webdriver: `Request failed with status 404 due to Error: Session not started or terminated`
I think, it was related to BrowserStack but I believe we overcame the problem with some changes in configs. Also, I notified the BrowserStack team about this issue.

Those were the main problems. Other than these I noticed some other issues and fixed them. I hope our iOS tests will run accurately from now on. Also here is a good documentation on flaky tests [https://www.browserstack.com/guide/how-to-avoid-flaky-tests](https://www.browserstack.com/guide/how-to-avoid-flaky-tests).

Also from now on, we can see the test names on the BrowserStack.

i1

We can't say for sure that we won't have any problems because we are dependant on a 3rd party (BrowserStack). At least, from now on we can make sure that our codes about iOS e2e tests don't cause any problem.


onurpolattimur:
There are two opened tickets on the BrowserStack side and I will write updates here.

### Issue 1 - One of the big problems was that our application is loaded like zoomed. Everything looks huge. Here is a captured screen

BrowserStack is currently investigating the issue and there is no update for now.

### Issue 2 - Request failed with status 404 due to Error: Session not started or terminated

BrowserStack: 
> The error "Session not started or terminated" was encountered because your test script tried to send commands after the test execution was terminated. 
​
I would request to you add more additional logs in your test script and check where the driver is quit abruptly. 

I replied to this, I told that we encounter this issue every once in a while. And now I am waiting for response.

src/e2e/config/webdriverio.config.js
```diff
@@ -9,25 +9,27 @@ const user = process.env.BROWSERSTACK_USERNAME
...
   user,
   key: process.env.BROWSERSTACK_ACCESS_KEY,
-
   capabilities: {
     platformName: 'iOS',
-    platformVersion: '14.4',
-    deviceName: 'iPhone 12',
     browserName: 'Safari',
+    device: 'iPhone 11',
+    osVersion: '13.3',
```
raineorshine, line 16: Can you add a comment here that describes why the device and osVersion is important for the tests? Thanks!

src/e2e/config/webdriverio.local.config.js
```diff
@@ -5,9 +5,11 @@ const config = {
...
   capabilities: {
     platformName: 'iOS',
-    platformVersion: '14.5',
-    deviceName: 'iPhone 12 Pro Max',
-    browserName: 'Safari',
+    platformVersion: '13.3',
+    deviceName: 'iPhone 11',
+    browserName: 'safari',
+    maxInstances: 1,
+    automationName: 'XCUITest',
```
raineorshine, line 12: Any comments you can add here would be helpful too

src/e2e/iOS/helpers/gesture.ts
```diff
@@ -25,18 +33,17 @@ const gesture = async (
...
 
   // webdriverio has some problem about type for TouchAction[] that is why we add @ts-ignore
   // We didn't use touchPerform here because webdriverio calls touchPerform in the background. https://github.com/webdriverio/webdriverio/blob/aea3d797ab1970309a60c43629f74154453597e9/packages/webdriverio/src/commands/constant.ts#L100
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
   await browser.touchAction(actions)
+
+  if (waitForEditableAfterGesture !== undefined) {
+    await waitForEditable(browser, waitForEditableAfterGesture)
+  }
```
raineorshine, line 46: I would avoid doing this here. It's not really related to the gesture itself. The consuming code can call `waitForEditable` afterwards. It's about the same amount of code as specifying the option.

If you'd like to provide an alias for the new thought gesture + wait + edit, you could create a `newThought` helper which executes all three, like we do for Puppeteer:

https://github.com/cybersemics/em/blob/3dcad97b2e7f1b1150218f60095492a6433ced18/src/e2e/puppeteer/helpers/newThought.ts#L4-L12

src/e2e/iOS/helpers/initSession.ts
```diff
@@ -15,9 +16,11 @@ const initSession = (): (() => Promise<Browser<'async'>>) => {
     }
 
     await mobileBrowser.url('http://bs-local.com:3000')
-    const skipElement = await waitForElement(mobileBrowser, '#skip-tutorial')
-    await skipElement.click()
-
+    const skipElement = await waitForElement(mobileBrowser, '#skip-tutorial', { timeout: 90000 })
+    await mobileBrowser.waitUntil(async () => await skipElement.isClickable())
```
raineorshine, line 20: Good catch, thank you

raineorshine (changes requested):

src/e2e/iOS/__tests__/caret.ts
```diff
@@ -40,10 +39,8 @@ it('Enter edit mode', async () => {
 })
 
 it('Preserve Editing: true', async () => {
-  await gesture(gestures.newThought)
-  await editThought('foo')
-  await gesture(gestures.newSubThought)
-  await editThought('bar')
+  await newThought('foo')
+  await newThought('bar', true)
```
raineorshine, line 43: Let's use an Options object `{ insertNewSubthought?: boolean }`. The option nicely mirrors the option to the `newThought` reducers.

In my experience I have found that Options objects are almost always preferable to optional arguments. They are self-documenting at the call-site since the option name is explicitly provided, and they provide an easy way to add additional options without order mattering.

src/e2e/webdriverio-environment.js
```diff
@@ -22,38 +19,29 @@ class WebdriverIOEnvironment extends JsDomEnvironment {
...
       this.global.browser.executeScript(
         'browserstack_executor: {"action": "setSessionStatus", "arguments": {"status":"failed","reason": "Failed"}}',
       )
     }
+
+    if (event.name === 'test_fn_start') {
+      this.global.browser.executeScript(
+        `browserstack_executor: {"action": "setSessionName", "arguments": {"name": "${event.test.name}"}}`,
+      )
+    } */
```
raineorshine, line 32: Can this be removed, or are you keeping it around for future debugging?

raineorshine (approved):

src/e2e/iOS/__tests__/caret.ts
raineorshine: Done in fda6386f77

raineorshine: Done in `fda6386f77`

raineorshine:
Yay!!! 🎉

PR #1313, opened 2021-07-02
fix: site crashes when no cursor detected but the splitSentence icon clicked

linghong:
Immediately after the site is  loaded, click the split sentence icon. At that time the cursor doesn't point to any thought. The site will crash with the below error:
Screen Shot 2021-07-01 at 2 43 31 PM

Fixed the above bug.


raineorshine:
A more general solution is to prevent the shortcut from being executed by checking `canExecute` when the Toolbar button is clicked. This check already exists when entering a keyboard shortcut, but it was missing from the Toolbar logic.

This was identified a few days ago and will be fixed in https://github.com/cybersemics/em/pull/1304#pullrequestreview-693945536

PR #1314, opened 2021-07-02
Cleaning up hideUntil property as remind me later functionality is no longer being used.

ruchijn:
Cleaning up `hideUntil` property


src/components/Content.tsx
```diff
@@ -120,9 +119,7 @@ const Content: ContentComponent = props => {
     if (focusNode && focusNode.nodeType === Node.TEXT_NODE) return
 
     // if disableOnFocus is true, the click came from an Editable onFocus event and we should not reset the cursor
-    if (showModal) {
-      dispatch(modalRemindMeLater({ duration: MODAL_CLOSE_DURATION }))
-    } else if (!noteFocus) {
+    if (!showModal && !noteFocus) {
```
raineorshine, line 122: I assume this should be replaced with `closeModal` rather than removing?

src/reducers/closeModal.ts
```diff
@@ -17,11 +15,10 @@ const modalRemindMeLater = (state: State, { id, duration = 0 }: { id?: string; d
     modals: {
       ...state.modals,
       [id ?? state.showModal!]: {
-        ...state.modals[id ?? state.showModal!],
-        hideuntil: time
+        ...state.modals[id ?? state.showModal!]
       }
```
raineorshine, line 19: This is a NOOP

src/reducers/closeModal.ts
```diff
@@ -5,9 +5,7 @@ import { State } from '../util/initialState'
 /**
  * Closes a modal temporarily.
  */
-const modalRemindMeLater = (state: State, { id, duration = 0 }: { id?: string; duration?: number }) => {
-
-  const time = Date.now() + duration
+const closeModal = (state: State, { id }: { id?: string }) => {
```
raineorshine, line 8: You should not need to pass the `id` to close a modal.

raineorshine (changes requested):
Looks good! Just a few small suggestions.

src/components/Content.tsx
ruchijn: But this is just passing the duration which was used to set `hideUntil`. And since we aren't using `hideUntil` anymore, I thought we should be ok to delete this. I will take a look again at this.

raineorshine: `modalRemindMeLater` closed the current Modal and set `hideuntil`.  We still need to close the Modal.

src/reducers/closeModal.ts
ruchijn: This was existing code but seems we do not need this. Hence deleted.

raineorshine (approved):
Looks good!

PR #1317, opened 2021-07-05
Prevent archived child from getting deleted on deleteEmptyThought

### LINKED ISSUE #1282: Delete empty thought with =archive subthought crashes
## Steps to Reproduce

```js
- =archive
- 
  - =archive
    - a
```

1. Turn `showHiddenThoughts` off.
2. Set the cursor on the empty thought
3. Hit Backspace

## Current Behavior

Error: Cannot read property 'value' of undefined

## Expected Behavior

Descendants of empty thought should be moved to the sibling `=archive`, as is done normally when deleting an empty thought with hidden children.

```js
- =archive
  - =archive
    - a
```

## Notes

- Error is caused by an undefined archivedChild that is passed to `existingThoughtMove`:

https://github.com/cybersemics/em/blob/830478be0b794757a8574de3fe22fda36df3f3d4/src/reducers/deleteEmptyThought.ts#L35-L36

- Only occurs if `showHiddenThoughts` is false
- Does not occur with other hidden thoughts, only `=archive`
- Occurs in both the root and nested contexts

shresthabijay:
fixes #1282 

# Cause of issue.

On deleting an empty thought all its hidden thoughts are archived which includes the `=archive` meta child too. Archiving the `=archive` thought will get deleted. So we cannot move already deleted thought to the new archived context.

# Solution
 
Just filtering out the `=archive` meta solves the problem. 

However, we don't get the following expected output as you described in the issue. 

```
- =archive
  - =archive
    - a
```

Since the logic is to copy the archived child to the context above it. Two duplicate `=archive` context will merge together to become this instead

```
=archive
  - a
```

Inorder to get the expected output as you mentioned we need to copy the nested `['', '=archive']` to `['=archive', '=archive']` instead of  `['=archive']` (current implementation).

I have not made this change. Please let me know if this is necessary.

### DIFF
--- a/src/reducers/deleteEmptyThought.ts
+++ b/src/reducers/deleteEmptyThought.ts
@@ -47,6 +47,9 @@ const deleteEmptyThought = (state: State): State => {
   // archive an empty thought with hidden children
   else if (isEmpty && visibleChildren.length === 0) {
     return reducerFlow([
+      // trying to archive child =archive causes an error when it is no longer available to be moved
+      // filter out child =archive to avoid
+      // https://github.com/cybersemics/em/issues/1282
       ...allChildren.map(child => (child.value === '=archive' ? null : archiveThought({ path: [...cursor, child] }))),
       state => {
         const archivedChild = getChildrenRanked(state, context)[0]
--- a/src/reducers/deleteEmptyThought.ts
+++ b/src/reducers/deleteEmptyThought.ts
@@ -47,7 +47,7 @@ const deleteEmptyThought = (state: State): State => {
   // archive an empty thought with hidden children
   else if (isEmpty && visibleChildren.length === 0) {
     return reducerFlow([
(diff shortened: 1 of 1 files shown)


raineorshine (approved):

raineorshine:
> #
> On deleting an empty thought all its hidden thoughts are archived which includes the `=archive` meta child too. Archiving the `=archive` thought will get deleted. So we cannot move already deleted thought to the new archived context.

Ah, good point. I didn't anticipate the nested archive being deleted.

> Just filtering out the `=archive` meta solves the problem.
> 
> However, we don't get the following expected output as you described in the issue.
> 
> ```
> - =archive
>   - =archive
>     - a
> ```
> 
> Since the logic is to copy the archived child to the context above it. Two duplicate `=archive` context will merge together to become this instead
> 
> ```
> =archive
>   - a
> ```

I like the merged solution!

PR #1318, opened 2021-07-05
Native side bar component

brunokiafuka:
This PR contains the sidebar menu component conversion to RN (React Native), using `DrawerLayout` wrapper from the [RN gesture handler library](https://docs.swmansion.com/react-native-gesture-handler/docs/api/components/drawer-layout).


src/components/Sidebar.native.tsx
```diff
@@ -0,0 +1,46 @@
...
+    _.sortBy(findTreeDescendants(recentlyEditedTree, { startingPath: [], showHiddenThoughts }), 'lastUpdated'),
+  )
+
+  return (
+    <View style={styles.container}>
+      <Text style={styles.headerText}>Recently Edited Thoughts</Text>
+
+      <View style={styles.padding}>
+        <Text style={styles.text}>RecentlyEditedBreadcrumbs will go here</Text>
+        {/* {recentlyEdited.map((recentlyEditedThought, i) => (
```
Foosballfan, line 26: is this still pending conversion?

src/components/HamburgerMenu.native.tsx
```diff
@@ -1,11 +1,14 @@
 import React from 'react'
 import { TouchableOpacity } from 'react-native'
 import Svg, { Path } from 'react-native-svg'
+import { useDrawerLayoutContext, IDrawerLayoutContext } from '../context/DrawerLayoutContext.native'
 
 //  const tutorialLocal = localStorage['Settings/Tutorial'] !== 'Off'
 
 /** An options menu with three little bars that looks like a hamburger. */
 const HamburgerMenu = ({ fill = '#fff', size = 45 }) => {
+  const { openDrawer } = useDrawerLayoutContext() as IDrawerLayoutContext
```
Foosballfan, line 10: is this state currently in redux? Is Raine ok for us to use context instead?

src/components/Sidebar.native.tsx
brunokiafuka: yip

brunokiafuka: will do it separated the `RecentlyEditedBreadcrumbs` component has some other links that require a bit of thinking so I can refactor it correctly.

src/components/HamburgerMenu.native.tsx
raineorshine: Since we don't currently use context in the app, I would need to understand the advantages and disadvantages vs Redux state in order to make a decision.

There is some precedent for such a transition. We are gradually switching to `useSelector` from `mapStateToProps`.

Foosballfan (approved):

src/components/HamburgerMenu.native.tsx
brunokiafuka: using redux now.

raineorshine (approved):

PR #1320, opened 2021-07-05
Prevent showing note if the parent is pending or not loaded

### LINKED ISSUE #1256: Do not render pending empty note
## Steps to Reproduce

```js
- a
  - b
    - c
      - =note
        - Note 
```

1. Add an artificial delay to dexie so the pending thoughts load more slowly. 
2. Refresh the page. 

## Current Behavior

Shows “Enter a note” while the note is loading

## Expected Behavior

Should not render the note if it is empty and pending.

shresthabijay:
fixes #1256 

# Changes
 - Add check to not show `Note` if the related `Parent` has not been loaded or if it is pending.

@raineorshine Should we write a test for this? We may need to mock `store.getState` to mimick the pending parent.

### DIFF
--- a/src/components/Note.tsx
+++ b/src/components/Note.tsx
@@ -4,7 +4,7 @@ import { isTouch } from '../browser'
 import { store } from '../store'
 import { attribute, hasChild, isContextViewActive } from '../selectors'
 import { deleteAttribute, editing, setAttribute, setNoteFocus } from '../action-creators'
-import { asyncFocus, selectNextEditable, setSelection, strip } from '../util'
+import { asyncFocus, hashContext, selectNextEditable, setSelection, strip } from '../util'
 import ContentEditable, { ContentEditableEvent } from 'react-contenteditable'
 import { Context } from '../types'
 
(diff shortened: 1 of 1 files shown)


src/components/Note.tsx
```diff
@@ -27,7 +27,15 @@ const Note = ({ context, onFocus }: NoteProps) => {
   const [justPasted, setJustPasted] = useState(false)
 
   const hasNote = hasChild(state, context, '=note')
-  if (!hasNote || isContextViewActive(state, context)) return null
+
+  /** Check if the note thought is pending or not. */
+  const isNotePending = () => {
+    const hashedContext = hashContext([...context, '=note'])
+    const noteThought = state.thoughts.contextIndex[hashedContext]
```
raineorshine, line 34: Avoid accessing `state.thoughts` directly. That is a low-level API that should only be accessed in low-level thought edit operations.

You can use the isPending selector: 

https://github.com/cybersemics/em/blob/b03a97a8b1a7cc0c53ce253b0ef9a16f33429510/src/selectors/isPending.ts#L5-L6

raineorshine (changes requested):

src/components/Note.tsx
```diff
@@ -27,7 +27,15 @@ const Note = ({ context, onFocus }: NoteProps) => {
...
+    return !noteThought || noteThought.pending
```
raineorshine, line 35: I think `!noteThought` is already covered by `hasNote`.

raineorshine (approved):

raineorshine:
> @raineorshine Should we write a test for this? We may need to mock `store.getState` to mimick the pending parent.

Seems like too much work given the priority. Thanks though!

PR #1322, opened 2021-07-07
convert modal components into react native.

brunokiafuka:
This PR contains the initial conversion of the modal components to react-native.


src/components/ModalComponent.native.tsx
```diff
@@ -31,26 +30,30 @@ const ModalComponent: React.FC<ModalProps> = props => {
   // const remindMeLater = () => dispatch(closeModal({ id: props.id }))
 
   /** Dispatches a modalComplete action for the modal. */
-  const complete = () => dispatch(modalComplete(props.id))
+  const close = () => dispatch(modalComplete(props.id))
```
brunokiafuka, line 33: Temporarily using close to hide the modals, @raineorshine is there any difference between close and complete modal?

raineorshine: `modalComplete` will permanently close the modal, while `closeModal` closes it allows it to be opened again. Generally you want `closeModal`. I'm not sure what will happen if you call `modalComplete` instead. The complete logic is older code that I think is only used for the Welcome modal now and should probably be factored out.

raineorshine: HomeLink apparently references it but I don't even see a Home modal anymore 🤷‍♀️

brunokiafuka: I am using `modalComplete` and I am being able to re-open the modals without any problems. They seem to perform very similar actions. Yet, the `closeModal` action calls an extra util function to clean up the modals (`modalCleanup`) and set the `showModal` to `null`.

raineorshine: What a mess 😕

I didn't do the modal cleanup code. You can git blame and find who on the team did it if you want to find out. Should have been better commented probably.

raineorshine: You can use modalComplete if it's working. I'll have one of my developers refactor it later.

brunokiafuka: Thanks, @raineorshine 🚀.

raineorshine (approved):

PR #1323, opened 2021-07-07
Prevent second onTap when gesture is below threshold

### LINKED ISSUE #1268: Short swipe right causes cursorBack and touch
Tested on iOS Safari. 

## Steps to Reproduce

```js
- a
  - b
```

1. Set the cursor on `a/b` with the keyboard down. 
2. Perform a very short swipe directly on top of `b` or to the right. If it is too short it will just enter edit mode. If it is too long it will just activate `cursorBack`. 

## Current Behavior

The cursor rapidly moves from `b` to `a` and back again. 

In this video I reproduce it twice in a row. The only touch I perform is two short swipe rights. 

https://user-images.githubusercontent.com/750276/121792996-15d3d000-cbc9-11eb-93fb-033ac6b3fcd2.mov

## Expected Behavior

The cursor should move back on swipe right.

raineorshine:
Fixes #1268

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -651,9 +651,15 @@ const Editable = ({
 
   /** Sets the cursor on the thought on mousedown or tap. Handles hidden elements, drags, and editing mode. */
   const onTap = (e: React.MouseEvent | React.TouchEvent) => {
-    // Stop propagation to not trigger onMouseDown event of Content.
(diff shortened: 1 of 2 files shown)


onurpolattimur (approved):

raineorshine:
I'll wait till #1309 before merging! 😅

PR #1329, opened 2021-07-09
Fixes deleteEmptyThought for thoughts with archived children

### LINKED ISSUE #864: `deleteEmptyThought` with hidden child is cancelled

anmolarora1:
fixes #864 

As described in [this comment](https://github.com/cybersemics/em/issues/864#issuecomment-813031897), the deleteEmptyThought didn't work as expected for archived children. 

Problem - we were trying to [archive all hidden children](https://github.com/cybersemics/em/blob/139130d3595fac8358a2cc6746d8b0e89fe58f76/src/reducers/deleteEmptyThought.ts#L50) of the empty thought, but the `archiveThoughts` handles one type of hidden thoughts differently - the thoughts that are already archived. Such thoughts get deleted based on [this condition](https://github.com/cybersemics/em/blob/139130d3595fac8358a2cc6746d8b0e89fe58f76/src/reducers/archiveThought.ts#L135), which isn't what we intend to do here. 

Solution - Do not try to archive the already archived thoughts, and simply move them to the parent instead.

### DIFF
--- a/src/reducers/__tests__/deleteEmptyThought.ts
+++ b/src/reducers/__tests__/deleteEmptyThought.ts
@@ -14,6 +14,7 @@ import newSubthought from '../newSubthought'
 import newThought from '../newThought'
 import setCursor from '../setCursor'
 import importTextReducer from '../importText'
+import archiveThought from '../archiveThought'
 
 it('delete empty thought', () => {
   const steps = [newThought('a'), newThought(''), deleteEmptyThought]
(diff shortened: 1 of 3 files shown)


raineorshine (approved):

raineorshine:
Thanks, works well!

> Problem - we were trying to [archive all hidden children](https://github.com/cybersemics/em/blob/139130d3595fac8358a2cc6746d8b0e89fe58f76/src/reducers/deleteEmptyThought.ts#L50) of the empty thought, but the `archiveThoughts` handles one type of hidden thoughts differently - the thoughts that are already archived. Such thoughts get deleted based on [this condition](https://github.com/cybersemics/em/blob/139130d3595fac8358a2cc6746d8b0e89fe58f76/src/reducers/archiveThought.ts#L135), which isn't what we intend to do here.
> 
> Solution - Do not try to archive the already archived thoughts, and simply move them to the parent instead.

Perfect explanation, thank you!

PR #1333, opened 2021-07-09
React Native footer component

brunokiafuka:
This PR contains the conversion of the Footer component. 

We also have created a Text wrapper component that will allow us to dynamically change the font size across the app.


src/components/Footer.native.tsx
```diff
@@ -0,0 +1,118 @@
...
+// import { TUTORIAL2_STEP_SUCCESS } from '../constants'
+import { logout, showModal } from '../action-creators'
+import { getSetting, isTutorial } from '../selectors'
+import { scaleFontDown, scaleFontUp } from '../action-creators/scaleSize'
+import { State } from '../util/initialState'
+import { View, StyleSheet, TouchableOpacity } from 'react-native'
+import { Text } from './Text.native'
+
+// eslint-disable-next-line jsdoc/require-jsdoc
+const mapStateToProps = (state: State) => {
```
Foosballfan, line 13: I think we are moving away from this and should use useselector

src/components/Footer.native.tsx
```diff
@@ -0,0 +1,118 @@
...
+      <View style={styles.flexEnd}>
+        <Text style={styles.textOpacityWhite}>{`Version: ${pkg.version}`}</Text>
+      </View>
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
+  topContentMargin: { marginRight: 20 },
+  flexEnd: { alignItems: 'flex-end' },
```
Foosballfan, line 99: are we implementing common styles?

brunokiafuka: I think we should.

src/components/Footer.native.tsx
brunokiafuka: Converting to use useSelector 🏃🏾‍♂️

src/components/Footer.native.tsx
brunokiafuka: this will be added on a different PR

package.json
```diff
@@ -49,9 +50,10 @@
     "emitter20": "^2.0.0",
     "emoji-strip": "^1.0.1",
     "expo": "^41.0.1",
-    "expo-clipboard": "^1.0.2",
+    "expo-clipboard": "1.0.2",
```
raineorshine, line 53: If this dependency needs to be locked, please also add an entry and comment to the `reject` list in https://github.com/cybersemics/em/blob/dev/.ncurc.js. We use [ncu](https://github.com/raineorshine/npm-check-updates) to keep dependencies up-to-date:

src/components/Footer.native.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const dispatch = useDispatch()
+  const authenticated = useSelector((state: State) => state.authenticated)
+  const user = useSelector((state: State) => state.user)
+  const status = useSelector((state: State) => state.status)
+  // const tutorialStep = useSelector((state: State) => +(getSetting(state, 'Tutorial Step') || 1))
+  // const isPushing = useSelector((state: State) => state.isPushing)
+  // const isTutorialOn = useSelector(isTutorial)
+  const pushQueueLength = useSelector((state: State) => state.pushQueue.length)
+
+  // if (isTutorialOn && tutorialStep !== TUTORIAL2_STEP_SUCCESS) return null
```
raineorshine, line 23: If you switch from `mapStateToProps` to `useSelector`, please change it in both `Footer.native.tsx` and `Footer.tsx`. We want to minimize differences between the native and web components. Some differences are unavoidable, such as native components or DOM access, which we can unify in the future. But accessing the Redux state should be the same. 

You could create a hook to use Redux state for each component, e.g. `Footer.useSelectors.ts`, and share it between web and native to avoid the code duplication 🙂. This would help ensure that changes/fixes on web are propagated to native. Curious your thought on this @Foosballfan!

You are welcome to continue using `mapStateToProps` if that is easier.

src/components/ActionButton.native.tsx
```diff
@@ -40,5 +41,5 @@ const styles = StyleSheet.create({
     marginVertical: 10,
     width: '50%',
   },
-  title: { fontSize: 18 },
+  title: { fontSize: 2 },
```
raineorshine, line 44: Is native `fontSize` a different unit than `px`?

raineorshine: Is there a way we can unify web and native `fontSize` here so they can specified as the same value in the component and converted at a lower-level?

raineorshine (changes requested):

src/components/Footer.native.tsx
Foosballfan: Sounds like a smart idea to me. If everyone starts converting to that convention when we touch components then we will converge on the goal sooner.

I wouldn't take a backwards step by continuing to use map state to props. Just perpetuating technical debt.

raineorshine: Thanks, that feels good to me.

src/components/ActionButton.native.tsx
Foosballfan: React native uses a special unit called DP which is a size that is relative to the pixel density of the device. It scales things so that they approximately relate on devices with different size screens.

Since we are using a text component to adjust the font size according to user's setting we can set a scale that matches somewhat to the Web sizes. 
There is no global equivalent to em on Web which reacts to inherited font size, so we have to do it somewhat manually with a custom component.

src/components/AppComponent.native.tsx
```diff
@@ -87,6 +94,7 @@ const styles = StyleSheet.create({
   container: {
     flex: 1,
     backgroundColor: '#000',
+    // marginTop: RNStatusBar.currentHeight,
```
Foosballfan, line 97: not needed?

src/components/Footer.native.tsx
```diff
@@ -1,39 +1,26 @@
...
-  isTutorialOn,
-  pushQueueLength,
-  tutorialStep,
-  user,
-  status,
-}: ReturnType<typeof mapStateToProps>) => {
+const Footer = () => {
   const dispatch = useDispatch()
+  const { authenticated, user, status, pushQueueLength } = useFooterUseSelectors()
+  // const authenticated = useSelector((state: State) => state.authenticated)
```
Foosballfan, line 17: not needed?

src/components/AppComponent.native.tsx
brunokiafuka: snap missed it.

brunokiafuka: Not in use

src/components/ActionButton.native.tsx
raineorshine: Okay, thanks for explaining

Foosballfan (approved):

raineorshine (approved):

PR #1338, opened 2021-07-10
Optimize and re-enable undo/redo

### LINKED ISSUE #1332: Optimize undo/redo
Undo/Redo currently relies on `deepClone(state)` which is unfortunately too slow. 

I think `immer` will work well instead of `deepClone`. It perform deep updates without a full clone: 
https://immerjs.github.io/immer/

- [ ] Re-enable undo/redo as-is
- [ ] Use the Chrome Profiler to measure the performance *before* (approximate is fine)
- [ ] Replace `deepClone` with Immer's `produce`. Try `applyPatch` on the `draft` object.
- [ ] Use the Chrome Profiler to measure the performance *after*.

anmolarora1:
fixes #1332 

Tried replacing [deepclone](https://github.com/Starcounter-Jack/JSON-Patch#jsonpatchdeepclonevalue-any-any) with immer's [produce](https://immerjs.github.io/immer/produce) for optimizing the undo/redo operations, and it seems to improve the performance by at least 5-6x. Further testing would be helpful for providing detailed insights.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "fp-and-or": "^0.1.3",
     "himalaya": "^1.1.0",
     "html-escaper": "^3.0.3",
-    "immer": "^9.0.3",
+    "immer": "^9.0.5",
     "indexeddbshim": "^7.1.0",
     "ipfs-http-client": "^43.0.1",
     "jex-block-parser": "^1.0.0",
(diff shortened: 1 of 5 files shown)


raineorshine (approved):
No noticeable performance issues with ~1,500 thoughts! Let's go for it!

PR #1342, opened 2021-07-13
alert component + common styles

brunokiafuka:
This PR adds: 

- [x] Alert component 
- [x] Mobile common styles
- [x] Auth modal


src/components/UndoIcon.native.tsx
```diff
@@ -0,0 +1,19 @@
...
+import Svg, { Path } from 'react-native-svg'
+
+// eslint-disable-next-line jsdoc/require-jsdoc
+const mapStateToProps = (state: State) => ({
+  dark: theme(state) !== 'Light',
+})
+
+/** An Undo icon. */
+const UndoIcon = ({ dark, fill, size = 18, style }: Icon) => (
+  <Svg x='0px' y='0px' width={size} height={size} fill={fill || (dark ? 'white' : 'black')} viewBox='0 -50 600 600'>
```
raineorshine, line 14: How about we create `Svg.native.tsx` and `Path.native.tsx`? It looks like everything else is the same between web and mobile.

We usually want to try to fork at the lowest possible level to reduce duplication.

brunokiafuka: We could try to do that but the RN Svg package doesn't support some of the web props like `style | classNames`

raineorshine: Yeah, that would probably be a good idea. I mostly used `style` and `class` to handle different fill/stroke arguments for different icons while respecting `EM/Settings/Theme`.

brunokiafuka: The existing web SVGs would be removed when we move to expo web. 

Yet, if we want to proxy the SVG implementations to reduce the duplications, I believe we can get this fixed by conditionally render the `className` and other web-only props and have them present only on the web, I can work on it on a different PR we decide to implement it.

raineorshine: Okay, thanks. I think we should actually leave it as-is for now given that it will get unified later anyway and there is not a lot of churn on SVG components.

raineorshine (approved):

src/components/Alert.native.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const alert = useSelector((state: State) => state.alert) as NonNullable<Alert>
+  const dispatch = useDispatch()
+
+  // eslint-disable-next-line jsdoc/require-jsdoc
+  const close = () => {
+    dispatch(alertAction(null))
+  }
+
+  return (
+    <View
```
Foosballfan, line 22: cool 👍🏻

Foosballfan (approved):

PR #1343, opened 2021-07-13
provide current state to updateThoughts instead

### LINKED ISSUE #1339: Undo/Redo: Old state recreated after undo
## Steps to Reproduce

1. Create thought `a`
2. Undo

## Current Behavior

Thought `a` is recreated.

An error is logged to the console:

image

## Expected Behavior

Thought `a` should be removed.

anmolarora1:
fixes #1339 

@raineorshine we can see that the error log points to [L170](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/redux-middleware/pushQueue.ts#L170) of `pushQueue` middleware and it's because we're assuming that the [pendingLexemes](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/redux-middleware/pushQueue.ts#L162) would never have `null` entries for thoughtIndexUpdates, as pointed out by @shresthabijay. A quick workaround would involve changing this assumption and skipping the attempt of loading contexts for such lexemes.

However, such a state shouldn't be generated in the first place because the lexeme generated by an undo can never be `pending` so the `pendingLexemes` generated by the `updateThoughts` reducer is incorrect, which is because the state that is being passed to it by  [undoRedoStoreEnhancer](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/redux-enhancers/undoRedoReducerEnhancer.ts#L29) was incorrect. The idea behind the same was to generate the appropriate pushQueue for the undo and redo updates to be propagated to data providers. The existing code tries to do that using the [updated state](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/redux-enhancers/undoRedoReducerEnhancer.ts#L29) (the state created after undo/redo operations) instead of the original state. But to generate the correct `pushQueue` we need the original state and the updates as contextIndexUpdates and thoughtIndexUpdates.
This issue has been addressed [here](https://github.com/anmolarora1/em/blob/8e04c37a779b48e3833fd4ac083b79bff8041085/src/redux-enhancers/undoRedoReducerEnhancer.ts#L21-L39)

However, I've found another issue that needs to be solved to fully fix #1339. 
The current implementation is leading us into an invalid state, due to incorrect values of root context (falsely marked as pending)

[In our initialState](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/util/initialState.ts#L21), we're marking the root thought as pending and the `pending` status is updated through [updateThoughts](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/action-creators/pull.ts#L45) to false only after the children get pulled. But when there are no children, the status doesn't change, leading to an erroneous state - It's the exact state that we reach after hitting undo when there's a single thought under the Root, as described in the [original issue](https://github.com/cybersemics/em/issues/1339). 
That's my finding so far, and I'd like to hear your thoughts before presenting a solution to this one. It might be causing other issues, but we can't say for sure since it's a rare state.


raineorshine:
Nice work

> @raineorshine we can see that the error log points to [L170](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/redux-middleware/pushQueue.ts#L170) of `pushQueue` middleware and it's because we're assuming that the [pendingLexemes](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/redux-middleware/pushQueue.ts#L162) would never have `null` entries for thoughtIndexUpdates, as pointed out by @shresthabijay. A quick workaround would involve changing this assumption and skipping the attempt of loading contexts for such lexemes.

Ah yes. Here's the original discussion with @shresthabijay about that non-null assumption: https://github.com/cybersemics/em/pull/1190#discussion_r640690839. I believe it is a reasonable assumption under most circumstances, though it is fragile. Its safety is not ensured by any guards or the type system, but rather by a contingency of the pull mechanism. Personally I think it may be a good idea to place a guard there since it fails hard if those contingencies change. 

For our purposes on this ticket we should avoid the invalid state generated by the undoRedoEnhancer, as you suggest.

> However, such a state shouldn't be generated in the first place because the lexeme generated by an undo can never be `pending` so the `pendingLexemes` generated by the `updateThoughts` reducer is incorrect, which is because the state that is being passed to it by [undoRedoStoreEnhancer](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/redux-enhancers/undoRedoReducerEnhancer.ts#L29) was incorrect. The idea behind the same was to generate the appropriate pushQueue for the undo and redo updates to be propagated to data providers. The existing code tries to do that using the [updated state](https://github.com/cybersemics/em/blob/4615ad614a7e853f61aa9f13c2f67c8c150a55b6/src/redux-enhancers/undoRedoReducerEnhancer.ts#L29) (the state created after undo/redo operations) instead of the original state. But to generate the correct `pushQueue` we need the original state and the updates as contextIndexUpdates and thoughtIndexUpdates.
> This issue has been addressed [here](https://github.com/anmolarora1/em/blob/8e04c37a779b48e3833fd4ac083b79bff8041085/src/redux-enhancers/undoRedoReducerEnhancer.ts#L21-L39)

Okay, that makes sense.

https://github.com/anmolarora1/em/blob/8e04c37a779b48e3833fd4ac083b79bff8041085/src/redux-enhancers/undoRedoReducerEnhancer.ts#L21-L29

It seems like this would exclude a patch when the thought doesn't exist in the old state. That would be fine for deleting a thought (which exists in the old state), but wouldn't it break an undo operation that was trying to recreate a thought which does not exist in the old state?

anmolarora1:
> 
> https://github.com/anmolarora1/em/blob/8e04c37a779b48e3833fd4ac083b79bff8041085/src/redux-enhancers/undoRedoReducerEnhancer.ts#L21-L29
> 
> It seems like this would exclude a patch when the thought doesn't exist in the old state. That would be fine for deleting a thought (which exists in the old state), but wouldn't it break an undo operation that was trying to recreate a thought which does not exist in the old state?

That's a good catch. This would result in generating invalid states for redo operations. I've fixed it in [this commit](https://github.com/cybersemics/em/pull/1343/commits/f9160d4b637edd53c6f82dd6bd9b61661a746ccf)

raineorshine (approved):

Issue #1354, opened 2021-07-15
Mobile: Bump Thought Down does not set caret


raineorshine:
Bump Thought Down: →←↓

## Steps to Reproduce

```
- a
  - b
```

1. Set the cursor on `a`
2. Close the keyboard
3. Activate `bumpThoughtDown`

## Current Behavior

Mobile keyboard opens but user cannot type.

## Expected Behavior

`bumpThoughtDown` should enter edit mode and place the caret on the new empty thought.

PR #1363, opened 2021-07-23
Set caret after applying bump thought down

onurpolattimur:
fixes: #1354 

-------

We are changing the value of the current thought in bumpThoughtDown.ts and the `useEffect` callback that is related to setSelection logic is not triggered in the `Editable` component.

https://github.com/cybersemics/em/blob/4db13c0b6d84b3531684b6c65e51d4f86b42c409/src/components/Editable.tsx#L360

The useEffect callback is triggered only if the following value change: `isEditing, cursorOffset, state.dragInProgress`

As a solution, I changed the `cursorOffset` in bumpThoughtDown.ts and also set the 'editing' state to true to be able to set caret. I don't want to broke stuff that is related to caret logic and Editable so, that's why I choose this way.  If you have a better idea we can change this logic.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -7,7 +7,6 @@ import classNames from 'classnames'
 import {
   alert,
   cursorBack,
-  editing,
   error,
   editThought,
   importText,
@@ -16,6 +15,7 @@ import {
   setInvalidState,
   tutorialNext,
   newThought,
+  editing as editingAction,
 } from '../action-creators'
 import { isTouch, isSafari } from '../browser'
 import globals from '../globals'
@@ -135,6 +135,7 @@ interface EditableProps {
   */
   transient?: boolean
   onKeyDownAction?: () => void
+  editing?: boolean | null
 }
 
 interface Alert {
@@ -192,6 +193,7 @@ const Editable = ({
   onKeyDownAction,
   dispatch,
   transient,
+  editing,
 }: Connected<EditableProps>) => {
   const state = store.getState()
   const thoughts = pathToContext(simplePath)
@@ -432,7 +434,7 @@ const Editable = ({
       shortcutEmitter.off('shortcut', flush)
       showDuplicationAlert(false, dispatch)
     }
-  }, [isEditing, cursorOffset, state.dragInProgress])
+  }, [isEditing, cursorOffset, state.dragInProgress, editing])
 
   /** Performs meta validation and calls thoughtChangeHandler immediately or using throttled reference. */
   const onChangeHandler = (e: ContentEditableEvent) => {
@@ -618,7 +620,7 @@ const Editable = ({
           !window.getSelection()?.focusNode ||
           !window.getSelection()?.focusNode?.parentElement?.classList.contains('editable')
         ) {
-          dispatch(editing({ value: false }))
+          dispatch(editingAction({ value: false }))
         }
       }
     })
--- a/src/components/StaticThought.tsx
+++ b/src/components/StaticThought.tsx
@@ -31,6 +31,7 @@ const StaticThought = ({
   style,
   simplePath,
   toggleTopControlsAndBreadcrumbs,
+  editing,
 }: ConnectedThoughtProps) => {
   const isRoot = simplePath.length === 1
   const isRootChildLeaf = simplePath.length === 2 && isLeaf
@@ -70,6 +71,7 @@ const StaticThought = ({
         <Editable
           path={path}
           cursorOffset={cursorOffset}
+          editing={editing}
           disabled={!isDocumentEditable()}
           isEditing={isEditing}
           rank={rank}
--- a/src/components/Thought.tsx
+++ b/src/components/Thought.tsx
@@ -106,6 +106,7 @@ interface ThoughtProps {
   style?: React.CSSProperties
   simplePath: SimplePath
   view?: string | null
+  editing?: boolean | null
 }
 
 export type ConnectedThoughtProps = ThoughtProps &
@@ -125,7 +126,7 @@ const getGlobalBullet = (key: string) => GLOBAL_STYLE_ENV[key as keyof typeof GL
 
 // eslint-disable-next-line jsdoc/require-jsdoc
 const mapStateToProps = (state: State, props: ThoughtContainerProps) => {
-  const { cursor, cursorOffset, expanded, expandedContextThought, search, expandHoverTopPath } = state
+  const { cursor, cursorOffset, expanded, expandedContextThought, search, expandHoverTopPath, editing } = state
 
   const { path, simplePath, showContexts, depth } = props
 
@@ -179,6 +180,7 @@ const mapStateToProps = (state: State, props: ThoughtContainerProps) => {
     publish: !search && publishMode(),
     simplePathLive,
     view: attribute(state, contextLive, '=view'),
+    editing,
   }
 }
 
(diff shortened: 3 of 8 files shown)


raineorshine (changes requested):
Good idea, although I think we may want to understand why this doesn't fit the general case. This solution breaks `bumpThoughtDown` in edit mode anyway when `cursorOffset` is 0.

`cursorOffset` may not change, and `state.dragInProgress` is not relevant. I think we want `isEditing` to change.

However, we have a bigger problem: I discovered the lexeme context for the bumped thought gets duplicated, which I just documented in #1364. Let's pause work on the selection behavior until we fix the data issue.

onurpolattimur:
> I think we may want to understand why this doesn't fit the general case that handles setting the caret. 

Because we are changing the value of the selected thought and none of from  `isEditing, cursorOffset, state.dragInProgress` has changed. So, it doesn't trigger the related useEffect callback of `Editable`.

> This solution breaks bumpThoughtDown in edit mode anyway when cursorOffset is 0.

I couldn't see any issue. Can you share the steps to reproduce the issue, please?

> cursorOffset may not change, and state.dragInProgress is not relevant. I think we want isEditing to change.

Yes, I thought so, too. But `isEditing` props coming from Thought and it is related to thought's path and cursor path equality. And this value doesn't change after bumpThoughtDown. Because we are changing the value of the selected thought, it's still 'true' and it doesn't cause to trigger useEffect callback.

> However, we have a bigger problem: I discovered the lexeme context for the bumped thought gets duplicated, which I just documented in #1364. Let's pause work on the selection behavior until we fix the data issue.

I didn't see this issue, I will take a quick look.

raineorshine:
> > This solution breaks bumpThoughtDown in edit mode anyway when cursorOffset is 0.
> 
> I couldn't see any issue. Can you share the steps to reproduce the issue, please?

Sure!

## Steps to Reproduce

```
- a
  - b
```

1. Set the cursor on `a`
2. Set the caret at the beginning of `a` (cursorOffset: 0)
3. Activate `bumpThoughtDown` - <kbd>Meta</kbd> + <kbd>Alt</kbd> + <kbd>d</kbd>

## Current Behavior

Mobile keyboard opens but typing does nothing.

## Expected Behavior

Mobile keyboard should open and edit the empty thought when typing.

onurpolattimur:
I tried on iOS too but, I couldn't reproduce the issue.

> I think we want to move away from trying to use cursorOffset to trigger the useEffect, but feel free to verify the behavior.

Yes, I am just trying to see the issue for the case you mentioned. After checking #1364, I will come back to this issue.
Here is captured screen video. Am I missing something? 

https://user-images.githubusercontent.com/23136437/126871283-ca827a9e-82c0-4b09-9a75-ab9e0b5802a9.MP4

raineorshine:
Apparently if you use a single swipe or tap to move the caret in Step 2, then it works correctly. You can reproduce the issue by doing at least one additional tap before Step 3, e.g. by closing the keyboard and then tapping again, or by tapping at the end before tapping at the beginning.

raineorshine:
Obviously 🙄😂

onurpolattimur:
Hi @raineorshine,
We trigger related useEffect callback of Editable by causing a change for `isEditing` props now. I am not so happy with this solution but I think it is the only way to achieve a change on isEditing props. Because isEditing prop is passed from Thought component and its value depends on the cursor.

https://github.com/cybersemics/em/blob/9af867e54c6cd2dc842a21fef62a409ee7d09783/src/components/Thought.tsx#L136

raineorshine:
Thanks for your solution. I don't think we're quite there yet.

Setting `cursor` to null and back again is going to trigger some unwanted side effects, namely `pull` and `updateUrlHistory`.

It also doesn't feel right to have to artificially trigger the `useEffect` hook. If it's not being triggered and it should be, we need to figure out what dependency should be added to the `useEffect` arguments.

Luckily it's reproducible on desktop if you emulate mobile in Chrome Dev Tools. That should make it a bit easier to investigate.

@shresthabijay Any ideas?

shresthabijay:
> Thanks for your solution. I don't think we're quite there yet.
> 
> Setting `cursor` to null and back again is going to trigger some unwanted side effects, namely `pull` and `updateUrlHistory`.
> 
> It also doesn't feel right to have to artificially trigger the `useEffect` hook. If it's not being triggered and it should be, we need to figure out what dependency should be added to the `useEffect` arguments.
> 
> Luckily it's reproducible on desktop if you emulate mobile in Chrome Dev Tools. That should make it a bit easier to investigate.
> 
> @shresthabijay Any ideas?

I agree with @raineorshine. We need to first look into exactly why the `useEffect` gets triggered in the first place.  Passing `editing` and `offset` to the `setCursor` call on `bumpThoughtDown` and adding `state.editing` as a dependency to `useEffect` seems to solve the problem only when `state.editing` is false. @onurpolattimur Can you check why this doesn't work when `state.editing` is `true`.

onurpolattimur:
Adding `state.editing` to the dependency array isn't enough. When hiding the keyboard, the editing state is changed but because we don't get the `editing` state as props, the Editable component doesn't re-render, and `state.editing` dependency of `useEffect` still has  value of `true` and the next time, the callback isn't triggered because of no change on this dependency. 

> Passing editing and offset to the setCursor call on bumpThoughtDown and adding state.editing as a dependency to useEffect seems to solve the problem only when state.editing is false.

You may have tested it and may have seen it worked but it will work only first onBumpThoughtDown. If you want to try here are the steps of the failure scenario:
- Create thoughts `A/B` and close the keyboard.
- Apply bumpThoughtDown on `A` 
- Edit empty thought and close the keyboard
- Apply bumpThoughtDown

> Did you find why doesn't it work when state.editing is true.  

Yes, in bumpThoughtDown we call` asyncFocus()` and it set selection to hidden input. Because the editing state doesn't change, the useEffect callback isn't triggered and doesn't set caret. Actually, there is no need for that useEffect callback to be triggered because caret already is on the right Editable.  I solve this issue by adding a guard into `bumpThoughtDownShortcut` for now, but I think we should add this guard to asyncFocus()

I made changes and it looks good to me. What do you think @shresthabijay

shresthabijay:
> Adding `state.editing` to the dependency array isn't enough. When hiding the keyboard, the editing state is changed but because we don't get the `editing` state as props, the Editable component doesn't re-render, and `state.editing` dependency of `useEffect` still has value of `true` and the next time, the callback isn't triggered because of no change on this dependency.

I see. It `state.editing` was getting out of sync because we didn't set `editing` back to `false` after blur. Good catch!

shresthabijay:
> Yes, in bumpThoughtDown we call` asyncFocus()` and it set selection to hidden input. Because the editing state doesn't change, the useEffect callback isn't triggered and doesn't set caret. Actually, there is no need for that useEffect callback to be triggered because caret already is on the right Editable. I solve this issue by adding a guard into `bumpThoughtDownShortcut` for now, but I think we should add this guard to asyncFocus()

It seems like we need to be more careful while using `asyncFocus` or changing selection imperatively.  Since the dependencies passed to `useEffect` themselves don't know about actual browser selection, we always need to be sure that dependencies like `editing` are always in proper sync with the browser selection whenever it changes.

shresthabijay:
> I made changes and it looks good to me now 😃. What do you think @shresthabijay?

Great work! Looks good to me too. Please rebase the PR to the latest `dev`.

shresthabijay (approved):
Thank you!

PR #1366, opened 2021-07-25
Preserve only font-weight, font-style and text-decoration style attributes

onurpolattimur:
fixes: #1158


raineorshine (approved):
Thanks!

Issue #1362, opened 2021-07-23
cursorDown on last thought breaks cursor


raineorshine:
## Steps to Reproduce

1. Create thought `a`
2. Activate `cursorDown`

## Current Behavior

Cursor becomes `/a/ROOT`.

## Expected Behavior

Cursor should not change when hitting `cursorDown` on the last thought.

PR #1367, opened 2021-07-25
Don't set cursor on cursorDown if the cursor on last thought

onurpolattimur:
fixes: #1362

### DIFF
--- a/src/reducers/cursorDown.ts
+++ b/src/reducers/cursorDown.ts
@@ -11,7 +11,7 @@ const cursorDown = (state: State) => {
   // if there is a cursor, get the next logical child, sibling, or uncle
   if (cursor) {
     const { nextThoughts, contextChain } = nextThought(state, cursor)
-    return nextThoughts && nextThoughts.length > 0
+    return nextThoughts
       ? setCursor(state, {
           path: nextThoughts,
           contextChain: contextChain || [],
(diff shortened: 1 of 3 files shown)


raineorshine (changes requested):
This does prevent the problem from happening, however it doesn't address the core issue: `nextThoughts` should not return a thought when there are no thoughts following the cursor.

I think `NextThoughtResult` is incorrect to have a non-nullable `Path`:

https://github.com/cybersemics/em/blob/57c8d079a11eaee1105e328e03a17dfd848bc967/src/util/nextThought.ts#L16-L19

i.e. "none" is a valid answer to the question "What is the next thought after this one?"

```ts
interface NextThoughtResult {
  nextThoughts: Path | null
  contextChain: SimplePath[]
}
```

src/reducers/cursorDown.ts
```diff
@@ -11,7 +11,7 @@ const cursorDown = (state: State) => {
   // if there is a cursor, get the next logical child, sibling, or uncle
   if (cursor) {
     const { nextThoughts, contextChain } = nextThought(state, cursor)
-    return nextThoughts.length > 0
+    return nextThoughts && nextThoughts.length > 0
```
raineorshine, line 14: I think you can omit `nextThoughts.length > 0`. A `Path` cannot have a length of 0.

raineorshine (changes requested):

raineorshine (approved):

PR #1369, opened 2021-07-27
[react-native]: multi gesture

brunokiafuka:
This PR adds support to multi gestures on the mobile app, to allow shortcuts to be fired on user gestures.

### DIFF
--- a/src/components/AppComponent.native.tsx
+++ b/src/components/AppComponent.native.tsx
@@ -21,7 +21,7 @@ import ModalAuth from './ModalAuth'
 import ErrorMessage from './ErrorMessage'
 import Content from './Content'
 
-import MultiGesture from './MultiGesture.native'
+import MultiGesture from './MultiGesture'
 import { store } from '../store'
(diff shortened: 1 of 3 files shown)


src/components/AppComponent.native.tsx
```diff
@@ -21,8 +21,26 @@ import ModalAuth from './ModalAuth'
 import ErrorMessage from './ErrorMessage'
 import Content from './Content'
 
+import MultiGesture from './MultiGesture'
+import { store } from '../store'
+import { isGestureHint, inputHandlers } from '../shortcuts'
+
 const { flexOne, darkBackground } = commonStyles
 
+const { handleGestureEnd, handleGestureSegment } = inputHandlers(store ?? {})
```
shresthabijay, line 30: I don't think null coalescing is needed here.

shresthabijay (changes requested):
Looks good to me! Just a small change.

shresthabijay (approved):
Thank you!

PR #1370, opened 2021-07-29
Add new thoughts

brunokiafuka:
This PR adds the ability to users to new thoughts using multi gestures `→↓`


shresthabijay:
@brunokiafuka Thanks for the PR! Using `isMobile` to change the behavior or prevent the execution of web-specific functions may not be a good idea. We will need to think of a better way to make action creators and reducers platforms agnostic. I will try to think of a solution. Please feel free to suggest your ideas.

shresthabijay:
> @brunokiafuka Thanks for the PR! Using `isMobile` to change the behavior or prevent the execution of web-specific functions may not be a good idea. We will need to think of a better way to make action creators and reducers platforms agnostic. I will try to think of a solution. Please feel free to suggest your ideas.

@brunokiafuka We will abstract platform-specific logic as we make progress. Right now most of the web-specific logic is related to the selection. Since we have yet to figure out the selection implementation in the native part, so we do not need to do the abstraction part right away. So we can use `isMobile` flag for now. Some functions like `getTextContentFromHTML` and `decodeThoughtsURL` are browser-specific. So instead of changing how they behave, use flags in the logic where execution of these functions needs to be prevented.

brunokiafuka:
> > @brunokiafuka Thanks for the PR! Using `isMobile` to change the behavior or prevent the execution of web-specific functions may not be a good idea. We will need to think of a better way to make action creators and reducers platforms agnostic. I will try to think of a solution. Please feel free to suggest your ideas.
> 
> @brunokiafuka We will abstract platform-specific logic as we make progress. Right now most of the web-specific logic is related to the selection. Since we have yet to figure out the selection implementation in the native part, so we do not need to do the abstraction part right away. So we can use `isMobile` flag for now. Some functions like `getTextContentFromHTML` and `decodeThoughtsURL` are browser-specific. So instead of changing how they behave, use flags in the logic where execution of these functions needs to be prevented.

it does make sense @shresthabijay.

Foosballfan (approved):

shresthabijay (approved):

Issue #1347, opened 2021-07-14
Mobile: Edit mode not disabled on modal open


raineorshine:
## Steps to Reproduce

```
- a
  - b
```

1. Set the cursor on `b`. 
2. Tap `b` to enter edit mode. 
2. Tap the Export Context toolbar button. 
3. Close the modal. 
4. Tap `a`

## Current Behavior

Keyboard pops up because it is still in edit mode. 

## Expected Behavior

Keyboard should not pop up since the keyboard is down. Edit mode should be disabled as soon as the keyboard goes down. 

## Requirements

Rather than disable edit mode in response to the modal, we should try to come up with a more general solution and disable edit mode when the keyboard is dismissed. 

1. See if there is a way to detect that the keyboard is down (even when there is still a browser selection). 

2. We could disable edit mode when the selection element has been onmounted. That might be expensive to detect though.

3. We could disable edit mode when `Content` is unmounted since we know all thoughts are unmounted. Only slightly more general than disabling edit mode on modal open.

shresthabijay:
@onurpolattimur This issue may have been solved by your recent PR https://github.com/cybersemics/em/pull/1363. Can you check if this issue still exists?

onurpolattimur:
> @onurpolattimur This issue may have been solved by your recent PR #1363. Can you check if this issue still exists?

@shresthabijay yes, this issue still exists. I will look into details.

onurpolattimur:
@shresthabijay we set the editing state to false when hiding the keyboard on Editable. But because some elements except the App and the Modal component are unmounted when we open a modal, the onBlur event of Editable isn't triggered and so, it doesn't set the editing state to false.

> See if there is a way to detect that the keyboard is down (even when there is still a browser selection).

I think, no.

> We could disable edit mode when the selection element has been onmounted. That might be expensive to detect though.

Yes, we can apply this solution by adding some little codes for components that we want to do something after the component is unmounted. I will open PR with this solution.

> We could disable edit mode when Content is unmounted since we know all thoughts are unmounted. Only slightly more general than disabling edit mode on modal open.

Yes, we can also apply this solution but I don't prefer this.

PR #1373, opened 2021-07-30
set editing state to false on Editable on-unmount

onurpolattimur:
fixes: #1347 

I didn't write iOS test as iOS tests take a long time to run. But good news, this issue was producible on mobile-view and I wrote puppeteer test 🐣


shresthabijay (approved):
Works well. Thanks!

PR #1374, opened 2021-07-30
BrowserStack: remove resetSafariFormatSettings

onurpolattimur:
As you know there is an issue with BrowserStack. Sometimes, our app was being loaded as zoomed in tests and we had applied a fix as a workaround. You can see details at #1309 

I got new news from the BrowserStack team.

> Our Engineering team has applied a fix with respect to this behavior and you should not be seeing this behaviour anymore.
> ​
> Could you execute a few tests to confirm the same?

I removed our workaround, we will see how it's going.


shresthabijay (approved):
Seems to be working. Thanks!

PR #1375, opened 2021-07-31
Refactor and optimize Export Modal

raineorshine:
- [x] Use number of lines of exported plain text as number of descendants to avoid extra traversal
- [x] Factor out ExportDropdown component
- [x] Clean up mount/unmount logic
- [x] Update numDescendants in real-time for better user experience


raineorshine:
@shresthabijay Let me know if you have a better idea for updating the number of descendants in real-time besides the emitter. I needed a way for `ModalExport` and `ExportThoughtsPhrase` to share the results from `pull` without `ModalExport` re-rendering each time.

shresthabijay:
> @shresthabijay Let me know if you have a better idea for updating the number of descendants in real-time besides the emitter. I needed a way for `ModalExport` and `ExportThoughtsPhrase` to share the results from `pull` without `ModalExport` re-rendering each time.
> 
> You can reproduce by activating the export modal on a thought with a large number of descendants (1000+).

The issue is that we want to re-render the `ExportThoughtsPhrase` without re-rendering`ModalExport`. So passing the props from the `ModalExport` to `ExportThoughtsPhrase` is out of the question. We can create a React Context that only stores `numDescendants`. `useDescendants` will trigger the change, and `ExportThoughtsPhrase` will rerender. However we need to make sure `useDescendants` itself doesn't re-render because of how context works. To work around that we can either use a selector like this https://www.npmjs.com/package/use-context-selector or create two contexts one for triggering and one for consumption.

shresthabijay (approved):

shresthabijay:
> @shresthabijay Let me know if you have a better idea for updating the number of descendants in real-time besides the emitter. I needed a way for `ModalExport` and `ExportThoughtsPhrase` to share the results from `pull` without `ModalExport` re-rendering each time.
> 
> You can reproduce by activating the export modal on a thought with a large number of descendants (1000+).

I hoisted the pull logic up to a parent component that uses two separate contexts for providing pull status and number of descendants. This way it is easy for nested components to consume necessary data without unecessary re-rerendering.

PR #1376, opened 2021-08-03
[React Native]: prevent scroll on multi gestures.

### LINKED ISSUE #1371: [React Native] Prevent main app scroll and sidebar activation on active multi gesture.
On active multi-gesture, the main content scroll view is still scrollable. Also sidebar can get activated.

**Note: I have only tested this on a IOS simulator.** 
## Expected Behavior.

Ideally, we would want to prevent scrolling of other native elements on active multi-gesture.

brunokiafuka:
Fixes #1371.

### DIFF
--- a/src/components/AppComponent.native.tsx
+++ b/src/components/AppComponent.native.tsx
@@ -1,4 +1,4 @@
-import React, { createRef, useEffect, useRef } from 'react'
+import React, { createRef, useEffect, useRef, useState } from 'react'
 import { ScrollView, View } from 'react-native'
 import { StatusBar } from 'expo-status-bar'
 import Toolbar from './Toolbar'
@@ -49,6 +49,7 @@ const AppComponent: React.FC = () => {
   const multiGestureRef = createRef<MultiGestureRef>()
   const dispatch = useDispatch()
   const { height } = useDimensions().screen
+  const [isScrollEnabled, setIsScrollEnabled] = useState(true)
 
   const showSidebar = useSelector((state: State) => state.showSidebar)
   const showAlert = useSelector((state: State) => state.alert)
@@ -67,6 +68,10 @@ const AppComponent: React.FC = () => {
   useEffect(() => {
     if (showSidebar) return openDrawer()
   }, [showSidebar])
+
+  useEffect(() => {
+    setIsScrollEnabled(multiGestureRef.current?.scrolling || false)
+  }, [multiGestureRef])
 
   const contentHeight = {
     height: height - 125,
@@ -97,7 +102,7 @@ const AppComponent: React.FC = () => {
             onCancel={handleGestureCancel}
           >
             <View style={contentHeight}>
-              <ScrollView scrollEnabled={multiGestureRef.current?.scrolling} nestedScrollEnabled={true} style={flexOne}>
+              <ScrollView scrollEnabled={isScrollEnabled} nestedScrollEnabled={true} style={flexOne}>
                 <Content />
               </ScrollView>
 
--- a/src/components/AppComponent.native.tsx
+++ b/src/components/AppComponent.native.tsx
@@ -1,4 +1,4 @@
-import React, { createRef, useEffect, useRef, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import { ScrollView, View } from 'react-native'
 import { StatusBar } from 'expo-status-bar'
 import Toolbar from './Toolbar'
@@ -15,13 +15,13 @@ import Alert from './Alert'
 import Footer from './Footer'
 import { useDimensions } from '@react-native-community/hooks'
 import { SafeAreaView } from 'react-native-safe-area-context'
-import { State } from '../@types'
+import { Direction, GesturePath, State } from '../@types'
 import { commonStyles } from '../style/commonStyles'
 import ModalAuth from './ModalAuth'
 import ErrorMessage from './ErrorMessage'
 import Content from './Content.native'
 
-import MultiGesture, { MultiGestureRef } from './MultiGesture.native'
+import MultiGesture from './MultiGesture'
 import { store } from '../store'
 import { isGestureHint, inputHandlers } from '../shortcuts'
 
@@ -46,10 +46,10 @@ const handleGestureCancel = () => {
  */
 const AppComponent: React.FC = () => {
   const drawerRef = useRef<DrawerLayout>(null)
-  const multiGestureRef = createRef<MultiGestureRef>()
+
   const dispatch = useDispatch()
   const { height } = useDimensions().screen
-  const [isScrollEnabled, setIsScrollEnabled] = useState(true)
+  const [isGestureActive, setIsGestureActive] = useState(false)
 
   const showSidebar = useSelector((state: State) => state.showSidebar)
   const showAlert = useSelector((state: State) => state.alert)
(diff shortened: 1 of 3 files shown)


src/components/AppComponent.native.tsx
```diff
@@ -21,7 +21,7 @@ import ModalAuth from './ModalAuth'
 import ErrorMessage from './ErrorMessage'
 import Content from './Content'
 
-import MultiGesture from './MultiGesture'
+import MultiGesture, { MultiGestureRef } from './MultiGesture.native'
```
Foosballfan, line 24: isnt this assumed to be native by the compiler?

Foosballfan (approved):

src/components/AppComponent.native.tsx
brunokiafuka: Yes, but the web multi-gesture component doesn't have a`MultiGestureRef` so we need to explicitly define the file in this case.

src/components/MultiGesture.native.tsx
```diff
@@ -40,183 +44,149 @@ const gesture = (p1: Point, p2: Point, threshold: number) =>
...
+    ref,
+  ) => {
+    const [scrolling, setScroll] = useState(true)
+
+    let abandon = false
+    let currentStart: Point | null = null
+    let scrollYStart: number | null = null
+    let sequence: GesturePath = ''
+
+    useImperativeHandle(ref, () => ({ scrolling }))
```
shresthabijay, line 75: I think using ref is not necessary. The `MultiGesture` can prevent the `ScrollView` from scrolling when the gesture is active. Since the `MultiGesture` wraps the `Content`, the nested scroll would be prevented as well. Whenever the gesture is cancelled or abandoned, `ScrollView` can be allowed to scroll again. Please let me know if I am missing something here.

src/components/MultiGesture.native.tsx
```diff
@@ -40,183 +44,149 @@ const gesture = (p1: Point, p2: Point, threshold: number) =>
...
+      // fired when a new gesture is added to the sequence
+      onGesture = noop,
+
+      // fired when all gestures have completed
+      onEnd = noop,
+      ...props
+    },
+    ref,
+  ) => {
+    const [scrolling, setScroll] = useState(true)
```
shresthabijay, line 68: I suggest creating a state called `isGetureActive` instead of `scrolling`.

src/components/MultiGesture.native.tsx
```diff
@@ -40,183 +44,149 @@ const gesture = (p1: Point, p2: Point, threshold: number) =>
...
+              if (g !== sequence[sequence.length - 1]) {
+                sequence += g
+
+                // reset gestures if swipe up or down.
+                if (sequence[0] === 'u' || sequence[0] === 'd') {
+                  reset()
+                  return
+                }
+
+                setScroll(false)
```
shresthabijay, line 162: You can replace this with `setIsGestureActive`.

src/components/MultiGesture.native.tsx
```diff
@@ -40,183 +44,149 @@ const gesture = (p1: Point, p2: Point, threshold: number) =>
...
 
-    // fired when all gestures have completed
-    onEnd: noop,
-  }
-}
+    return (
+      <ScrollView
+        style={commonStyles.flexOne}
+        {...panResponder.panHandlers}
+        scrollEnabled={scrolling}
```
shresthabijay, line 183: Replace this with `!isGestureActive`, preventing scroll when gesture is active.

src/components/AppComponent.native.tsx
```diff
@@ -67,6 +69,10 @@ const AppComponent: React.FC = () => {
     if (showSidebar) return openDrawer()
   }, [showSidebar])
 
+  useEffect(() => {
+    setIsScrollEnabled(multiGestureRef.current?.scrolling || false)
+  }, [multiGestureRef])
```
shresthabijay, line 74: I believe mutable`ref` is not a valid dependency here.

src/components/MultiGesture.native.tsx
brunokiafuka: Because we have a nested scroll view we also need to prevent the children scrolling otherwise we'd still have some scroll interfering with the gestures.

brunokiafuka: Just had a second look, and if we really need to prevent the scroll on gestures we also will need to pass `isGestureActive` to its children being `Content` which is a scrollable view, and its parent `ScrollView`. This way `Content` would then receive a prop to manage its scrolling.

brunokiafuka:
The `ScrollView` component works/behaves independently from its parents when nesting it, thus, we needed to use `isGestureActive` from the `MultiGesture` to manage its children's behavior and prevent the scroll by using the ref value.

src/components/AppComponent.native.tsx
```diff
@@ -67,6 +69,10 @@ const AppComponent: React.FC = () => {
     if (showSidebar) return openDrawer()
   }, [showSidebar])
 
+  useEffect(() => {
+    setIsScrollEnabled(multiGestureRef.current?.isGestureActive || false)
+  }, [multiGestureRef])
```
shresthabijay, line 74: This mutable ref dependency will itself not re-run this effect. It is working because the `showAlert` is causing the `AppComponent` to re-render after every gesture, which is causing effect to work. Ideally, we would want to prevent the whole `AppComponent` to re-render when `alert` state changes.

Since we want a reactive state that syncs with the current status of gesture, `ref` is not an ideal option here. Instead, I think we can attach custom callbacks to the `MultiGesture` instead.

```javascript

/*** State ***/

const [isGestureActive, setIsGestureActive] = useState(false)

/*** Component ***/

  <MultiGesture
      onGesture={(...args) => {
        setIsGestureActive(true)
        handleGestureSegment(...args)
      }}
      onEnd={(...args) => {
        setIsGestureActive(false)
        handleGestureEnd(...args)
      }}
      shouldCancelGesture={shouldCancelGesture}
      onCancel={() => {
        setIsGestureActive(false)
        handleGestureCancel()
      }}
    >
      <ScrollView scrollEnabled={!isGestureActive}/>
    </MultiGesture>
```

src/components/MultiGesture.native.tsx
shresthabijay: Got it!

shresthabijay:
> The `ScrollView` component works/behaves independently from its parents when nesting it, thus, we needed to use `isGestureActive` from the `MultiGesture` to manage its children's behavior and prevent the scroll by using the ref value.

I see. However `ref` is not reactive, so we cannot rely on that to re-render the component. I have come up with a solution here https://github.com/cybersemics/em/pull/1376#discussion_r682470120. Please let me know what you think about this.

src/components/AppComponent.native.tsx
brunokiafuka: I believe this can also work. Also, this would help us prevent the dependency on the `showAlert ` selector changes which as you mentioned causes rerenders.

brunokiafuka:
> > The `ScrollView` component works/behaves independently from its parents when nesting it, thus, we needed to use `isGestureActive` from the `MultiGesture` to manage its children's behavior and prevent the scroll by using the ref value.
> 
> I see. However `ref` is not reactive, so we cannot rely on that to re-render the component. I have come up with a solution here [#1376 (comment)](https://github.com/cybersemics/em/pull/1376#discussion_r682470120). Please let me know what you think about this.

Fixed on [e314fb0](https://github.com/cybersemics/em/pull/1376/commits/e314fb02e853149c3425c1635cdbc5bbdd3b09e9)

shresthabijay:
> > The `ScrollView` component works/behaves independently from its parents when nesting it, thus, we needed to use `isGestureActive` from the `MultiGesture` to manage its children's behavior and prevent the scroll by using the ref value.
> 
> I see. However `ref` is not reactive, so we cannot rely on that to re-render the component. I have come up with a solution here [#1376 (comment)](https://github.com/cybersemics/em/pull/1376#discussion_r682470120). Please let me know what you think about this.

Looks good. Thanks!

PR #1398, opened 2021-08-11
[React-Native]: content editable in webview

brunokiafuka:
Implemented content editable on webview.

### DIFF
--- a/src/components/ContentEditable.native.tsx
+++ b/src/components/ContentEditable.native.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
-import React, { useRef, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import { View, ViewStyle, StyleSheet } from 'react-native'
 import { WebView } from 'react-native-webview'
 
@@ -31,6 +31,8 @@ export interface IKeyDown {
   key: string
 }
 
+const DEFAULT_WEBVIEW_HEIGHT = 25
+
 /**
  * Content Editable Component.
  */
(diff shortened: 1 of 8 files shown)


src/components/ContentEditable.native.tsx
```diff
@@ -1,70 +1,252 @@
...
-    // const event = Object.assign({}, e, {
-    //   target: {
-    //     text: e,
-    //   },
-    // })
-
     props.onChange(e)
   }
 
+  // if (value === undefined) return null
```
Foosballfan, line 183: comments

src/components/ContentEditable.native.tsx
```diff
@@ -1,70 +1,252 @@
...
+            case WEBVIEW_POST_EVENTS.onFocus: {
+              // @ts-ignore
+              window.getSelection = () => contentRef?.current?.injectJavaScript('getSelectionValues()')
+
+              if (props.onFocus) props.onFocus()
+              break
+            }
+
+            case WEBVIEW_POST_EVENTS.onBlur: {
+              //  setValue(html)
```
Foosballfan, line 211: comments

src/components/ContentEditable.native.tsx
```diff
@@ -1,70 +1,252 @@
...
+              if (props.onKeyDown) props.onKeyDown(webEvent as IKeyDown)
+              break
+            }
+
+            case WEBVIEW_POST_EVENTS.onPaste: {
+              if (props.onPaste) props.onPaste(webEvent as IOnPaste)
+              break
+            }
+
+            // TODO: improve onChange event: if we use oninput to listen to changes on the text the component ends up rendering and taking cursor focus and dismissing the keyboard.
```
Foosballfan, line 229: this is done?

src/util/getTextContentFromHTML.ts
```diff
@@ -1,7 +1,12 @@
+import { isMobile } from './isMobile'
+
 /**
  * Returns text without any html tags for the given HTML value. It accurately represents how user sees the text in the browser.
  */
 export const getTextContentFromHTML = (htmlValue: string) => {
+  // temporary hack as shim createElement is causing to exceeded maximum calls
```
Foosballfan, line 7: can we change the method to do this instead? https://css-tricks.com/snippets/javascript/strip-html-tags-in-javascript/

src/components/ContentEditable.native.tsx
```diff
@@ -1,70 +1,245 @@
...
+interface IWebHTMLProps {
+  placeholder: string
+  innerHTML: string
+  isEditing?: boolean
+}
+
+/**
+ * Get html string.
+ */
+const getWebHTML = ({ placeholder, innerHTML, isEditing }: IWebHTMLProps) => `
```
Foosballfan, line 51: We can tidy this up by putting it in adjacent file

src/util/getTextContentFromHTML.ts
brunokiafuka: So this would be: 
```js
if (isMobile()) return htmlValue.replace(/(<([^>]+)>)/gi, "")
```

shresthabijay (commented):
@brunokiafuka I am very happy with the progress. However, these are the things that we need to work on:

- The focus doesn't get applied to thoughts on creating new thoughts.
- The long text doesn't get wrapped.
- Since web view re-renders every time the source changes, everything gets reinitialized in the webview. Maybe we can keep the source the same and change the `innerHTML` by injecting javascript. We can set the focus using the same way.

Please let me know your thoughts.

brunokiafuka:
> @brunokiafuka I am very happy with the progress. However, these are the things that we need to work on:
> 
> * The focus doesn't get applied to thoughts on creating new thoughts.
> * The long text doesn't get wrapped.
> * Since web view re-renders every time the source changes, everything gets reinitialized in the webview. Maybe we can keep the source the same and change the `innerHTML` by injecting javascript. We can set the focus using the same way.
> 
> Please let me know your thoughts.

_* The focus doesn't get applied to thoughts on creating new thoughts._
I was re-testing the focus now, and it's a bit doggy will be improving on it.

_* The long text doesn't get wrapped._
Will wrap the text if it overflows. 

_* Since web view re-renders every time the source changes, everything gets reinitialized in the webview. Maybe we can keep the source the same and change the `innerHTML` by injecting javascript. We can set the focus using the same way._
We are actually preventing the re-renders now, the webview only re-render if you blur the focus to another element/content-editable component. Inject the `innerHTML` will cause the cursor to lose its position.

shresthabijay:
> * We are actually preventing the re-renders now, the webview only re-render if you blur the focus to another element/content-editable component. Inject the `innerHTML` will cause the cursor to lose its position.

I get it now. Thanks!

shresthabijay (approved):

PR #1422, opened 2021-09-03
[React Native]: Toolbar short cuts

brunokiafuka:
- [x] Prose View
- [x] Table View
- [x] Split Sentence


src/components/ContentEditable.native.tsx
```diff
@@ -43,18 +44,33 @@ const ContentEditable = ({
...
+  useEffect(() => {
+    setWebviewHTML(createWebHTML({ innerHTML: html, placeholder, isEditing, isTable }))
   }, [isEditing])
 
+  useEffect(() => {
+    if (html.trim() === innerHTMLValue.trim()) return
+
+    setWebviewHTML(createWebHTML({ innerHTML: html, placeholder, isEditing, isTable }))
+  }, [html])
+
```
shresthabijay, line 73: You can use a single `useEffect` with `isTable`, `isEditing` and `html` as dependencies.

src/components/ContentEditable.native.tsx
```diff
@@ -43,18 +44,33 @@ const ContentEditable = ({
...
   placeholder,
   isEditing,
+  isTable,
   ...props
 }: ContentEditableProps) => {
   const allowInnerHTMLChange = useRef<boolean>(true)
   const { width } = useDimensions().window
   const [height, setHeight] = useState(DEFAULT_WEBVIEW_HEIGHT)
   const contentRef = useRef<WebView>(null)
+  const [innerHTMLValue, setInnerHTMLValue] = useState(html)
```
shresthabijay, line 54: Seems like `innerHTMLValue` doesn't need to be a reactive state. If so, you can use `mutable` ref instead.

shresthabijay (changes requested):
Overall it looks good. Just some minor changes are needed.

src/components/ContentEditable.native.tsx
brunokiafuka: I split into two different effects, `isEditing` and `isTable` are very similar so they are in a single `useEffect`. Whereas `html` effect is independent so it can perform the better.

src/components/ContentEditable.native.tsx
```diff
@@ -51,22 +51,18 @@ const ContentEditable = ({
   const { width } = useDimensions().window
   const [height, setHeight] = useState(DEFAULT_WEBVIEW_HEIGHT)
   const contentRef = useRef<WebView>(null)
-  const [innerHTMLValue, setInnerHTMLValue] = useState(html)
+  const [innerHTMLValue, setInnerHTMLValue] = useState<string>()
```
shresthabijay, line 54: I meant `useRef` instead of `useState`.

shresthabijay (changes requested):
@brunokiafuka  Minor change left.

shresthabijay (approved):

PR #1424, opened 2021-09-07
Styled Components And Twin Macro

shresthabijay:
# Changes

- Added styled-components and [twin.macro](https://github.com/ben-rogerson/twin.macro)
- Refactored `ActionButton`.
- Refactored `ModalComponent`
- Refactored `ModalHelp`, `ModalExport` and `ModalFeedback`
- Added `framer-motion` for modal mount and unmount animation.
- Added `react-icon-kit`


src/TextArea.tsx
```diff
@@ -0,0 +1,10 @@
+import tw from 'twin.macro'
+
+const TextArea = tw.textarea`
```
raineorshine, line 3: Would this file make more sense in `/src/style`?

src/components/ModalComponent.tsx
```diff
@@ -38,144 +42,175 @@ const getSelectionCoordinates = () => {
...
+            opacity: 0,
+          }}
+          animate={{
+            opacity: 1,
+          }}
+          exit={{
+            opacity: 0,
+          }}
+          transition={{ duration: 0.5 }}
+          style={Object.assign(
```
raineorshine, line 174: Let's replace `Object.assign` with the spread operator while we're here.

src/components/ModalExport.tsx
```diff
@@ -205,6 +210,106 @@ const ExportDropdown: FC<ExportDropdownProps> = ({ selected, onSelect }) => {
...
+`
+
+const ActionWrapper = tw.div`
+  flex justify-center
+  mt-14 mb-5
+`
+
+const ClipboardWrapper = tw.div`
+  text-center
+`
```
raineorshine, line 225: In general, we want to name things based on what they do rather than where they are used. Since this component is simply a `<div>` with centered text, we can call it `TextCenter`. This makes is easier to reuse and easier to understand at the call site without having to look up the definition.

If it makes sense, you may be able to apply this principle to other style definitions.

Let me know if there is a Tailwind or styled component convention that I'm not aware of 🙂.

src/components/ModalFeedback.tsx
```diff
@@ -6,6 +6,12 @@ import Modal from './Modal'
 import { alert } from '../action-creators'
 import { AxiosError } from 'axios'
 import { State } from '../@types'
+import TextArea from '../TextArea'
+import TextLink from './TextLink'
+
+// Needs to import twin.macro for tw prop to work
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import tw from 'twin.macro'
```
raineorshine, line 14: Good comment, thank you

raineorshine (changes requested):
Looks good! There are two systemic issues I have questions about.

- Does this approach work on native components as well? We need to have a strategy both in general and for the specific components that are refactored in this PR. The last thing we want is for web and native to diverge even more, and ideally we want an approach that helps move us towards a shared codebase. Even if the styles are completely different, we can still use the same technique and mostly the same component hierarchy.
- It looks like the Dark/Light theming is lost in this PR. What is your strategy for theming?

Also looks like the Puppeteer and Browserstack tests are failing.

src/components/ModalExport.tsx
shresthabijay: I prefer naming things for what they are here. What they do is already defined by their styling. My preference is based on the fact that on most projects I try to find things based on what they are before I try to understand what they do. For example: If I wanted to refactor or make changes to the clipboard related components in `ModalExport` after some days, it would be easier to get the right components based on what they are. I won't be able to find things easily if we name them based on what they do.  I agree that naming things based on what they do is useful for reuseable styled-components. `ClipboardWrapper` happens to be a `div` with centered text. But we would have to change the name if later we add some other styles to it. Also with `tailwind`  we already have access to great reusable CSS utilities, so we don't have to create separate reusable components like `TextCenter`.

I see this approach commonly used in `styled-components` community too.

src/TextArea.tsx
shresthabijay: > * It looks like the Dark/Light theming is lost in this PR. What is your strategy for theming?

For themes, we are using the `tailwind` approach here. https://github.com/shresthabijay/em/blob/cfe81de2bd77a30f96bdc61c246aa173fdfdf853/src/components/ActionButton.tsx#L29-L32

It is compatible with the current theme implementation. If `.dark` class is added to the body, the `dark` variant of the `tailwind` will be applied. 

You can see how `tailwind` handles it https://tailwindcss.com/docs/dark-mode.

shresthabijay:
> * Does this approach work on native components as well? We need to have a strategy both in general and for the specific components that are refactored in this PR. The last thing we want is for web and native to diverge even more, and ideally we want an approach that helps move us towards a shared codebase. Even if the styles are completely different, we can still use the same technique.

I agree. But styles and views are very different on both platforms and I don't think it would be that easy to share those. But we would want to have the same technique for both platforms. I am looking into support of `styled-components` and `tailwind` in react native. I will keep you updated.

src/components/ModalComponent.tsx
shresthabijay: Sure.

shresthabijay:
@raineorshine Sorry I accidentally deleted your comment on keeping styled-components on `src/styles`.  We can keep them in a separate `src/styles`.  But I like to consider them as building block `components` with styles added to them. For example: `TextArea` is not actually a style that we can add to another component. Instead, it is a component with its own style very close to it.

Generally, if a component is large, we can have a separate `style.ts` file for it. Example: 

- components
  -

shresthabijay:
> Also looks like the Puppeteer and Browserstack tests are failing.

Looking into it.

shresthabijay:
## What part of styling we can we share between native and web codebase ?

Since views and styles are very different on both platforms we cannot share views and styles directly. However, we can share themes and design systems between them. I believe we can do the following:

- Create a theme object that can be shared between platforms. (Specially colors). For web we can use CSS variables for theming and in native we can use context API to do it. 
- Design system can be shared by using tailwind on both platforms. Web (`twin.macro`) and Native ('tailwind-react-native-classnames'). 

I think we can try to come up with similar techniques for styling in the codebase but in the end, we can only hope to share design systems and themes between them. It is because we rely on CSS  and CSS variables a lot on the web.

@raineorshine What are your thoughts on this ?

src/components/ModalExport.tsx
raineorshine: > My preference is based on the fact that on most projects I try to find things based on what they are before I try to understand what they do. For example: If I wanted to refactor or make changes to the clipboard related components in `ModalExport` after some days, it would be easier to get the right components based on what they are. I

Okay. So the component names serve as a form of documentation that makes it easier to read the component hierarchy and identify components based on their role.

> But we would have to change the name if later we add some other styles to it.

Or extend/mix-in other styles, though it looks like atomic styled components are more idiomatic.

> Also, it won't be easier to always name things based on what they do, especially when styles are complex.

That's a good point. A lot of components are simply an ad hoc set of styles for that specific situation.

Let's give it a try. I was initially cautious because the "semantic" approach to styling in HTML was a disaster in the late 90's and early 2000's. It caused a proliferation of location-specific classes, and it was the standard approach recommended for years. However styled components are co-located, scoped, and get compiled down in the build step so I think it's not a problem.

raineorshine (changes requested):

raineorshine:
> But styles and views are very different on both platforms and I don't think it would be that easy to share those.

Yes

> But we would want to have the same technique for both platforms. I am looking into support of `styled-components` and `tailwind` in react native. I will keep you updated.

Okay, thanks.

> @raineorshine Sorry I accidentally deleted your comment on keeping styled-components on `src/styles`. We can keep them in a separate `src/styles`. But I like to consider them as building block `components` with styles added to them. For example: `TextArea` is not actually a style that we can add to another component. Instead, it is a component with its own style very close to it.
> 
> Generally, if a component is large, we can have a separate `style.ts` file for it. Example: If a footer has lots of styled components then we can separate them in `style.ts` and assemble them in `index.ts`.
> 
> ```
> - components
>   - footer
>     - index.ts
>     - style.ts
> ```
> 
> However, if a component is very small then we can simply keep everything in one place. Just like with text area component.
> 
> I don't think by any means this is the perfect structure to follow but it keeps it easy and simple. We can always change the structure if we feel friction along the way. Let me know your thoughts on this.

Sounds good! I don't think I had understood at that point that they are actual components, not just style sets.

In that case, how about `/src/components/TextArea`? I'm guessing you didn't intend to place it in `/src`.

> ## What part of styling we can we share between native and web codebase ?
> Since views and styles are very different on both platforms we cannot share views and styles directly. 

Yes, it's impossible to share styles, since views and elements are different. 

> * Create a theme object that can be shared between platforms. (Specially colors). For web we can use CSS variables for theming and in native we can use context API to do it.

Unifying the theme API would be great.

> * Design system can be shared by using tailwind on both platforms. Web (`twin.macro`) and Native ('tailwind-react-native-classnames').

Yes, they should share the same *syntax* to reduce the cognitive gap between web and native.

raineorshine:
Here's what I am seeing in light mode:

## Current

image

## Expected

image

shresthabijay:
> In that case, how about `/src/components/TextArea`? I'm guessing you didn't intend to place it in `/src`.

Yes, I meant in the `src/components/TextArea`.

src/App.css
```diff
@@ -81,7 +84,7 @@ li {
   list-style: none;
 }
 
-input[type='email'],
+/* input[type='email'],
```
raineorshine, line 87: It would be better to remove the lines completely than comment them out if they are not being used anymore. I find that that ends up happening anyway over time, and we always have git blame to find the old code. Let me know if you had something else in mind.

src/TextArea.tsx
```diff
@@ -0,0 +1,11 @@
+import tw from 'twin.macro'
```
raineorshine, line 1: This file should be moved to `/src/components`.

src/components/ModalComponent.tsx
raineorshine: TODO

src/e2e/puppeteer/helpers/waitForHiddenEditable.ts
```diff
@@ -6,6 +6,7 @@ async function waitForHiddenEditable(page: Page, value: string) {
   const editableElement = await waitForEditable(page, value)
   await page.waitForFunction(
     (element: Element) => {
+      // TODO: This fails if app is in light mode
```
raineorshine, line 9: I refactored this to not require DOM access and will merge soon!

raineorshine (changes requested):
The Close button is not working on ModalHelp.

Otherwise looks good!

> > * Does this approach work on native components as well? We need to have a strategy both in general and for the specific components that are refactored in this PR. The last thing we want is for web and native to diverge even more, and ideally we want an approach that helps move us towards a shared codebase. Even if the styles are completely different, we can still use the same technique.
> 
> I agree. But styles and views are very different on both platforms and I don't think it would be that easy to share those. But we would want to have the same technique for both platforms. I am looking into support of `styled-components` and `tailwind` in react native. I will keep you updated.

Let me know what you find out here. This is a critical piece before we move forward with this solution. We need a single syntactical styling system for both Web and React Native, otherwise this only serves to further bifurcate the platforms. It would be an improvement, but not a step forward in the broader goal of a shared component hierarchy.

shresthabijay:
> Let me know what you find out here. This is a critical piece before we move forward with this solution. We need a single syntactical styling system for both Web and React Native, otherwise this only serves to further bifurcate the platforms. It would be an improvement, but not a step forward in the broader goal of a shared component hierarchy.

Tailwind can be used in native with [react-native-classnames](https://github.com/jaredh159/tailwind-react-native-classnames). The syntax is similar to how we use it on the web. However, the main difference is dark mode support. The way we use dark mode on the web is not supported in the native. I don't think we can make both platforms use exactly the same styling interface but this method. But with `styled-components` and `tailwind`, we can have a similar way of defining styles in both platforms. But there will still be platform-specific quirks.

shresthabijay:
@raineorshine I have added `tailwind-react-native-classnames` and have addressed the theming problem with react context and styled component. How we define styles is the same as the web. Let me know your thoughts on this styling system.

I have not addressed your previous code review yet. I will do that after you have approved of this approach.

src/themeProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const ThemeSetterContext = createContext<ThemeSetters>({} as any)
+
+/**
+ * Global Theme Provider.
+ */
+const ThemeProvider: FC = ({ children }) => {
+  const [mode, setMode] = useState<ColorMode>('light')
+  const [fontScale, setFontScale] = useState<number>(1)
```
raineorshine, line 22: I'm skeptical about introducing a completely new method of state management. **em** is a Redux application and all global state is in the store. Access to state is provided through standard actions and reducers. This provider presents a new approach to storing global state and a new approach to modifying it, via hooks. This would be fine for a new application, but I'm not sure it's justified in an existing application with well-established state management norms.

src/components/ActionButton.native.tsx
```diff
@@ -1,14 +1,25 @@
...
+import tw from 'tailwind-react-native-classnames'
+import styled from 'styled-components/native'
+import themeSpecific from '../styled-helpers/themeSpecific'
+
+const Button = styled.TouchableOpacity`
+  ${tw`p-2 my-4 items-center rounded-lg w-6/12` as Record<string, string>}
+  ${themeSpecific({
+    dark: tw`bg-white`,
+    light: tw`bg-black`,
+  })}
```
raineorshine, line 13: This looks good. 

I'm still wrapping my head around how the theme gets passed to this. Where do the props to `themesSpecific` come from?

src/components/App.tsx
```diff
@@ -7,18 +7,24 @@ import AppComponent from './AppComponent'
...
-        <TouchMonitor>
-          <AppComponent />
-        </TouchMonitor>
-      </ErrorBoundaryContainer>
-    </Provider>
-  </DragAndDropContext>
+  <>
+    {/* TODO: Enable reset styles  */}
+    {/* <BaseStyles /> */}
+    <GlobalStyle />
```
raineorshine, line 19: We don't we use `ThemeProvider` on web?

src/components/AppComponent.native.tsx
```diff
@@ -82,13 +85,21 @@ const AppComponent: React.FC = () => {
...
     <>
       <StatusBar
         // eslint-disable-next-line react/style-prop-object
         style='light'
       />
-      <SafeAreaView style={[flexOne, darkBackground]}>
+      <SafeAreaView
+        style={tw.style({
+          'bg-black': theme.mode === 'dark',
+          'bg-white': theme.mode === 'light',
```
raineorshine, line 99: Why does this not use `themeSpecific`? It looks like it is doing the same kind of thing, but maybe I'm confused about where `themeSpecific` can be used.

src/components/ModalWelcome.native.tsx
```diff
@@ -141,16 +161,17 @@ const ModalWelcome = () => {
               }
               style={styles.helperActionButton}
             >
-              <Text style={styles.helperActionButtonText}>This ain’t my first rodeo. Skip it.</Text>
+              <GrayedText>This ain’t my first rodeo. Skip it.</GrayedText>
             </TouchableOpacity>
           }
+          <ThemedButton title='CHANGE THEME' onClick={() => toggleTheme()}></ThemedButton>
```
raineorshine, line 167: The vision of **em** Settings is that they are edited like any other thought. So this is fine for demonstration purposes, but ultimately users will use existing idioms for accessing and editing Settings. We will eventually add a Settings button to the toolbar or footer that links to the `EM/Settings` context where the user can edit settings as thoughts.

src/styled-helpers/themeSpecific.ts
```diff
@@ -0,0 +1,11 @@
+import { Theme } from '../themeProvider'
+
+type ThemeSpecificStyle = Record<Theme['mode'], any>
```
raineorshine, line 3: Can we be more specific than `any`?

src/themeProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+  setMode: (mode: ColorMode) => void
+  setFontScale: (scale: number) => void
+}
+
+export type Theme = {
+  mode: ColorMode
+  fontScale: number
+}
+
+const ThemeSetterContext = createContext<ThemeSetters>({} as any)
```
raineorshine, line 15: Can we be more specific than `any`?

src/util/adjustFontSize.ts
```diff
@@ -13,5 +13,8 @@ export const adjustFontSize = (factor: number, style: TextStyle, originalFontSiz
     : style
   const currentFontSize = mergedStyle.fontSize || originalFontSize
 
+  // Factor should be cakculated by the difference between current base font size and original base font size.
+  const factor = Math.abs(baseFontSize - originalFontSize)
```
raineorshine, line 17: Why `Math.abs`?

raineorshine (changes requested):
> Tailwind can be used in native with [react-native-classnames](https://github.com/jaredh159/tailwind-react-native-classnames). The syntax is similar to how we use it on the web.

Great, that looks good. I will ask Morrow for their opinion.

> However, the main difference is dark mode support. The way we use dark mode on the web is not supported in the native. I don't think we can make both platforms use exactly the same styling interface but this method.

This is a reasonable limitation, but I'd like to understand it better. I understand why we can't use the web method on React Native, since it relies on CSS inheritance, but can you describe the React Native method and why it can't be used on web?

> But with `styled-components` and `tailwind`, we can have a similar way of defining styles in both platforms.

That's great. We're aiming for syntactical/structural similarity and more seamless development across platforms.

> But there will still be platform-specific quirks.

Yes, understood.

shresthabijay:
> This is a reasonable limitation, but I'd like to understand it better. I understand why we can't use the web method on React Native, since it relies on CSS inheritance, but can you describe the React Native method and why it can't be used on web?

We can use the same method we have in react native on the web. But tailwind already has a great and easy solution to dark mode in the web. But it cannot be used in native.

```ts
tw`
 bg-white
 dark:bg-black
`
```

This method uses CSS inheritance internally and is much easier. But we can use the same method i.e `themeSpecific` on the web as well for keeping it similar to native.

src/themeProvider.tsx
shresthabijay: Sure we can push this into the redux state. I did this for a quick demonstration of `themeSpeific` helper.

src/components/ModalWelcome.native.tsx
shresthabijay: I understand. I did not polished these changes on purpose. It was just for demonstration purposes.

src/util/adjustFontSize.ts
shresthabijay: I will change this.

raineorshine:
> So either we can use `themeSpecific` only in native or we can refrain from using `tailwind` dark mode utility classes and use `themeSpecific` on the web as well. I think we can go with it later if we want a consistent styling pattern.

Looks like `tailwind-react-native-classnames` v2 supports the same syntax as web! 

```tsx
import { View, Text } from 'react-native';
import tw from 'twrnc';

const MyComponent = () => (
  <View style={tw`p-4 android:pt-2 bg-white dark:bg-black`}>
    <Text style={tw`text-md text-black dark:text-white`}>Hello World</Text>
  </View>
);
```

https://github.com/jaredh159/tailwind-react-native-classnames/tree/v2#tailwind-react-native-classnames-v2-%EF%B8%8F

src/themeProvider.tsx
raineorshine: Okay, I see. 

In the future, please make explicit which aspects are for demonstration purposes only. I run into this a lot with draft PR's. Neither of us need me to explain things that you already know, but without explicit comments, there is no way for me to infer your intentions about which parts of the code are meant to be evaluated.

src/components/ActionButton.native.tsx
raineorshine: Thank you!

src/components/ModalFeedback.tsx
```diff
@@ -78,19 +84,19 @@ const ModalFeedback = () => {
...
               Cancel
-            </a>
+            </TextLink>
           </div>
         </div>
       )}
     >
-      <p style={{ fontSize: 18, marginBottom: 30 }}>Send us your bugs, hopes, and dreams!</p>
-      <textarea
+      <p tw='text-lg mb-5 text-center'>Send us your bugs, hopes, and dreams!</p>
```
raineorshine, line 98: For my own understanding, is this the same as `` tw.p`text-lg mb-5 text-center` ``? And the same as ``style={tw`text-lg mb-5 text-center` ``?

If so, there appear to be three different syntaxes for some tailwind attributes. I'd like to better understand any reasons there may be to use more than one.

raineorshine (changes requested):
Looks good! I think `twrnc` and `twin.macro` is a great combination and I'm happy to use it going forward.

Light theme looks good.

Here are the UI issues I'm seeing:

- ModalHelp
  - Close button doesn't work
  - Link color should be `#add8e6`, as in `dev`
  - Console warning: Warning: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`
- ModalAuth
  - Font sizes
  - Heading spacing
  - "Work Offline" link color
- ModalExport
  - Font sizes
  - Heading spacing
  - Textarea size
  - Link colors
  - `user-select: none` on "Advanced" dropdown link
- ModalInvites
  - Description text background color
  
After fixing the above issues, you may want to do a side-by-side comparison of each modal with https://em-proto.web.app, to ensure the visual appearance has not changed.

raineorshine:
Is `emotion` compatible with `twin.macro`? And with `twrnc`? Can we still do theming consistently between web and RN?

If you can describe which modules we would need to switch out of that would be helpful, since we arrived at a specific combination that gave us the cross-platform syntax we wanted. Thanks!

shresthabijay:
> Is `emotion` compatible with `twin.macro`? And with `twrnc`? Can we still do theming consistently between web and RN?

. `twin.macro` can use any CSS in js solution for styling. So it compatible with `emotion`. `twrnc` doesn't need emotion to function. However `emotion` also has native support for react native. So it should be as compatible as `styled-components`

raineorshine:
Closing due to inactivity

PR #1425, opened 2021-09-09
[React Native]: thoughts animation

brunokiafuka:
This PR adds the functionality which allows that only current thoughts should be shown based on their current depth of the thoughts. Implementation based of: 

```
    distance-from-cursor-0 fully visible
    distance-from-cursor-1 dimmed
    distance-from-cursor-2 shifted left and hidden
    distance-from-cursor-3 shiifted left and hidden
```


src/components/Subthoughts.native.tsx
```diff
@@ -564,6 +566,8 @@ export const SubthoughtsComponent = ({
   */
   const actualDistance = shouldShiftAndHide || zoom ? 2 : shouldDim ? 1 : distance
 
+  // width - distance * depth;
```
Foosballfan, line 569: comment

Foosballfan (approved):

src/util/getThoughtSpacing.ts
```diff
@@ -0,0 +1,14 @@
...
+
+  if (distance === 0 || (distance === 1 && depth !== 0)) {
+    spacing = 20
+  }
+
+  if (distance === 1 && depth === 0) {
+    spacing = 5
+  }
+
+  return spacing
```
raineorshine, line 13: Please do not use mutations :).

src/util/getThoughtSpacing.ts
```diff
@@ -0,0 +1,14 @@
+/** Get thought space. */
+export const getThoughtSpacing = (distance: number, depth: number): number => {
```
raineorshine, line 2: Since `getThoughtSpacing` is only used in one place, it should be defined in that module. Creating a separate util function unnecessarily increases the surface area of the internal API. Unless it is a very large implementation, it's also better to co-locate functions so that a developer can look at the implementation without leaving the file.

src/components/Thought.tsx
```diff
@@ -106,6 +107,7 @@ interface ThoughtProps {
   simplePath: SimplePath
   view?: string | null
   editing?: boolean | null
+  space?: number
```
raineorshine, line 110: Could you explain the new `space` prop for me?

raineorshine (changes requested):

src/util/getThoughtSpacing.ts
brunokiafuka: Fixed

src/util/getThoughtSpacing.ts
brunokiafuka: Makes sense.

brunokiafuka: Also fixed.

src/components/Thought.tsx
brunokiafuka: We are passing down the `space` prop (maybe needs a better name), so we can access it on the `ContentEditable` component in order to adjust the width of the `WebView` to prevent text overflow on long thoughts.

brunokiafuka: it's the result of `getThoughtSpacing(actualDistance, depth) * depth` that helps us get the space value that needs to be subtracted from the `WebView` container.

brunokiafuka: We did it this way because we aren't able to globally style a component.

src/components/Subthoughts.native.tsx
```diff
@@ -107,6 +108,19 @@ const findFirstEnvContextWithZoom = (
...
+const getThoughtSpacing = (distance: number, depth: number): number => {
+  if (distance === 0 || (distance === 1 && depth !== 0)) {
+    return 20
+  }
+
+  if (distance === 1 && depth === 0) {
+    return 5
+  }
+
+  return 0
```
raineorshine, line 121: A ternary would be cleaner here. Single return statements are easier to reason about than multiple return statements. I try to only use additional return statements when short-circuiting from the beginning of a function.

src/components/Thought.tsx
raineorshine: Okay, I see. Let's name it something like `marginLeftParent` and provide a descriptive comment to make it clearer.

src/components/Subthoughts.native.tsx
brunokiafuka: Code refactored.

raineorshine (approved):

shresthabijay (approved):

raineorshine:
Some conflicts from the PR I just merged (I merge them in the order they are received), then ready to merge!

shresthabijay:
> Some conflicts from the PR I just merged (I merge them in the order they are received

@raineorshine Should we resolve the conflict and merge this PR?

raineorshine:
@brunokiafuka

src/components/Editable.native.tsx
```diff
@@ -133,20 +130,12 @@ const showDuplicationAlert = duplicateAlertToggler()
 // intended to be global, not local state
 let blurring = false
 
-// eslint-disable-next-line jsdoc/require-jsdoc
-const mapStateToProps = (state: State, props: EditableProps) => {
-  return {
-    isCursorCleared: props.isEditing && state.cursorCleared,
-  }
-}
-
```
raineorshine, line 141: I tried this before and it seemed like it made the Editable get re-rendered too often. We can try it again but let's make sure it does not introduce additional re-renders. I believe `connect` shields the child component so that it is not re-rendered when the parent is re-rendered.

Since this is not platform specific, it should be applied to both `Editable.native.tsx` and `Editable.tsx`.

raineorshine (changes requested):

PR #1427, opened 2021-09-10
[React Native]: export thoughts

brunokiafuka:
Added the functionality that will allow users to export/share their thoughts and/or also copy them to the clipboard.

### DIFF
--- a/src/components/ModalExport.native.tsx
+++ b/src/components/ModalExport.native.tsx
@@ -101,6 +101,32 @@ const PullProvider: FC<{ context: Context }> = ({ children, context }) => {
 }
 
 /******************************************************************************
+ * ExportThoughtsPhrase component
+ *****************************************************************************/
+
+interface ExportThoughtsPhraseOptions {
+  context: Context
+  // the final number of descendants
+  numDescendantsFinal: number | null
+  title: string
+}
+
+/** A user-friendly phrase describing how many thoughts will be exported. Updated with an estimate as thoughts are pulled. */
+const ExportThoughtsPhrase = ({ context, numDescendantsFinal, title }: ExportThoughtsPhraseOptions) => {
+  const store = useStore()
+  const state = store.getState()
+
+  // updates with latest number of descendants
+  const numDescendants = useDescendantsNumber()
+
+  const exportThoughtsPhrase = exportPhrase(state, context, numDescendantsFinal ?? numDescendants ?? 0, {
+    value: title,
+  })
+
+  return <Text>{exportThoughtsPhrase}</Text>
+}
+
+/******************************************************************************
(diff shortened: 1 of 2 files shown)


src/components/ModalExport.native.tsx
```diff
@@ -56,18 +126,16 @@ const ModalExport = () => {
...
-  const [exportContent] = useState<string | null>(null)
+  const [selected, setSelected] = useState(exportOptions[1])
+  const [selectedOptionIndex, setSelectedOptionIndex] = useState(0)
+  const [exportContent, setExportContent] = useState<string | null>(null)
   const [shouldIncludeMetaAttributes, setShouldIncludeMetaAttributes] = useState(true)
   const [shouldIncludeArchived, setShouldIncludeArchived] = useState(true)
-  // const [numDescendantsInState, setNumDescendantsInState] = useState<number | null>(null)
-  const numDescendantsInState = 0
+  const [numDescendantsInState, setNumDescendantsInState] = useState<number | null>(null)
+  // const numDescendantsInState = 0
```
Foosballfan, line 136: a few unuseful comments here

src/selectors/exportContext.ts
```diff
@@ -15,6 +16,8 @@ const replaceTitle = (text: string, title: string, format: MimeType) => {
 
 /** Strips out HTML tags when exporting as plain text. */
 const stripHTMLTag = (str: string) => {
+  if (isMobile()) return str
+
```
shresthabijay, line 20: You need to strip out the HTML tags here. Can you see if this library works [`html-to-text`] in native (https://www.npmjs.com/package/html-to-text) ?

src/components/ModalExport.native.tsx
```diff
@@ -360,4 +402,21 @@ const styles = StyleSheet.create({
   },
 })
 
-export default ModalExport
```
shresthabijay, line 363: https://github.com/cybersemics/em/blob/05142c9b35f60f67b10366f9fe2bd1b6f1fbc185/src/components/ModalExport.tsx#L137-L150

You have missed adding this component. `ExportThoughtsPhrase` updates the number of descendants in real time when they are being pulled. 

https://github.com/cybersemics/em/blob/05142c9b35f60f67b10366f9fe2bd1b6f1fbc185/src/components/ModalExport.tsx#L418

shresthabijay (changes requested):

src/selectors/exportContext.ts
brunokiafuka: Used regex instead.
`return str.replace(/(<([^>]+)>)/gi, '')`

shresthabijay (approved):

PR #1430, opened 2021-09-11
Move getCaretPositionDetails within setSelection

raineorshine:
- Hid `getCaretPositionDetails` within `setSelection` since that is the only place it is used
- Replaced `getCaretPositionDetails` in tests with native selection API


raineorshine:
@shresthabijay In hindsight, I'm not sure why we built a DOM traversal algorithm when we only ever set the offset to the beginning or end of a thought.

shresthabijay:
> @shresthabijay In hindsight, I'm not sure why we built a DOM traversal algorithm when we only ever set the offset to the beginning or end of a thought.

I believe there was a case where we needed to set the caret somewhere in the middle after focus is lost. I can't seem to remember where exactly it was. I will look into it.

raineorshine:
I looked and couldn't find it, but perhaps I missed it.

This PR does preserve that functionality anyway.

shresthabijay:
@raineorshine `getCaretPositionDetails` was introduced in this PR https://github.com/cybersemics/em/pull/1203`. The problem is well defined here. Focusing with native selection API is easy but doesn't work well when the value of the `ContentEditable` has rich text features like bold and italics. Since caret offset is not thought in terms of HTML value, the `getCaretPositionDetails` helps bridge the gap and lets us think of the offset in terms of plain text. For example: `hello <b>world</b>`. For setting caret after `w`, the offset we think of is `7` and not `10`. I believe I wrote with function thinking we might need to restore caret anywhere in the middle of thought with HTML value. But I think we don't have a use case for this yet 😅

shresthabijay (approved):
Looks good!

shresthabijay:
> I looked and couldn't find it, but perhaps I missed it.
> 
> This PR does preserve that functionality anyway.

Sure. I believe this functionality is important. At least it gives us confidence that setting caret on HTML value won't be a problem. We already have proper tests for it too.

Issue #1365, opened 2021-07-24
Superscript position is wrong at smaller font size


raineorshine:
## Steps to Reproduce

```
- A
  - I am in continuous creative response to everything that arises
- B
  - I am in continuous creative response to everything that arises
```

- Use mobile device or set screen width so that the subthought text breaks after "response to".
- Set Font Size to 14 using the footer controls.

## Current Behavior

Superscript appears in the middle of the second line. 

![image](https://user-images.githubusercontent.com/750276/126855203-daa59178-5161-41c0-be58-a6d57646d8b9.jpeg)

## Expected Behavior

Superscript should appear at the end of the thought.

ruchijn:
I am not sure but seems we are rendering `ThoughtAnnotation` [here](https://github.com/cybersemics/em/blob/dev/src/components/Thought.tsx#L450). And then `StaticThought` at [here](https://github.com/cybersemics/em/blob/dev/src/components/Thought.tsx#L461). And we render thought inside `ThoughtAnnotation` but hide it. We render superscript in `ThoughtAnnotation` and it is not set to hidden. 

Then we also render thought in `StaticThought` And this thought is visible but superscript is not. So we are rendering thought in separate tag than superscript. Hence the position issue comes up. If thought and superscript are rendered from same tag then the misalignment doesn't happen.

raineorshine:
Awesome thanks!

> It seems we are rendering `ThoughtAnnotation` [here](https://github.com/cybersemics/em/blob/dev/src/components/Thought.tsx?rgh-link-date=2021-09-14T10%3A23%3A14Z#L450). And then `StaticThought` at [here](https://github.com/cybersemics/em/blob/dev/src/components/Thought.tsx?rgh-link-date=2021-09-14T10%3A23%3A14Z#L461). And we render thought inside `ThoughtAnnotation` but hide it. We render superscript in `ThoughtAnnotation` and it is not set to hidden.

Yes, the component hierarchy is like this:

- ThoughtContainer (Thought.tsx)
  - Bullet
  - ThoughtAnnotation
    - StaticSuperscript
  - StaticThought
    - BulletCursorOverlay
    - Editable
    - Superscript
  - Subthoughts
    - ...

> Then we also render thought in `StaticThought` And this thought is visible but superscript is not. So we are rendering thought in separate tag than superscript. Hence the position issue comes up. If thought and superscript are rendered from same tag then the misalignment doesn't happen.

[...]

PR #1432, opened 2021-09-13
Change to fix superscript position on mobile device.

ruchijn:
Fix: #1365 

Before:
Screen Shot 2021-09-12 at 10 04 03 PM

After:
Screen Shot 2021-09-12 at 10 03 07 PM

### DIFF
--- a/src/components/StaticThought.tsx
+++ b/src/components/StaticThought.tsx
@@ -84,7 +84,7 @@ const StaticThought = ({
         />
       )}
 
-      <Superscript simplePath={simplePath} showContexts={showContexts} superscript={!isRoot} />
+      <Superscript simplePath={simplePath} showContexts={showContexts} superscript={false} />
     </div>
   )
 }
--- a/src/components/StaticThought.tsx
+++ b/src/components/StaticThought.tsx
@@ -84,7 +84,7 @@ const StaticThought = ({
         />
       )}
 
-      <Superscript simplePath={simplePath} showContexts={showContexts} superscript={false} />
+      <Superscript simplePath={simplePath} showContexts={showContexts} superscript={!isRoot} />
     </div>
   )
 }
--- a/src/App.css
+++ b/src/App.css
@@ -502,12 +502,17 @@ a.text-link {
   user-select: text;
 }
 
+.thought-container {
+  width: calc(100% - 1em);
+  position: relative;
+}
+
 .thought,
 .thought-annotation {
   width: calc(100% - 1em); /* expand click area. offset content padding to prevent line break */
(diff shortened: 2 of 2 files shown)


src/components/StaticThought.tsx
```diff
@@ -82,7 +82,7 @@ const StaticThought = ({
         />
       )}
 
-      <Superscript simplePath={simplePath} showContexts={showContexts} superscript={false} />
```
ruchijn, line 85: Not sure what was the reason behind passing `false` for `superscript`. I tried testing as much as I could figure out. Please let me know if there was a specific reason in setting this to false.

raineorshine: I did git blame all the way back a few years and still couldn't find the reason for it. I can see that whoever wrote it (which could have been me) wanted to still render the superscript container, even if the superscript was not supposed to be rendered. But I am not sure the reason. 

We may be able to remove it completely, but I would have to do some more rigorous testing to find out for sure.

I would probably leave it as-is until we fix the current issue, just to make sure we are not creating a regression.

raineorshine (changes requested):
The BrowserStack tests are failing and they are correct. It can be a bit hard to find the failed test currently, as I haven't figured out how to turn off the verbose logging, but you can find it by searching for "FAIL ".

I tested briefly on iOS and enabling/disabling edit mode is that main thing that is not working correctly. I suspect that the click area of the ThoughtAnnotation has changed due to the new styles.

```
FAIL src/e2e/iOS/__tests__/caret.ts (264.106 s)
  ✕ Enter edit mode  (11974 ms)
  ✕ Preserve Editing: true (24613 ms)
  ✓ Preserve Editing: false (24172 ms)
  ✓ Tap hidden root thought (23290 ms)
  ✕ Tap hidden uncle (19285 ms)
  ✓ Tap empty content while keyboard up (30648 ms)
  ✕ Tap empty content while keyboard down (23028 ms)
  ✓ Swipe over cursor (38223 ms)
  ✕ Swipe over hidden thought (30407 ms)
  ✓ Bump Thought Down on a thought that has children (26808 ms)
```

Before attempting a solution, please write up a description of the cause in the original issue. Then we can discuss together the best approach.

raineorshine (changes requested):
@ruchijn This BrowserStack test is failing: https://github.com/cybersemics/em/blob/dev/src/e2e/iOS/__tests__/caret.ts#L29

Can you investigate if it's correct before I do my review?

raineorshine:
Unfortunately I'm still seeing the original issue at different screen widths:

image

Issue #1419, opened 2021-09-03
Superscript not rendered on load


raineorshine:
## Steps to Reproduce

```
- a
  - b
- b
``` 

1. Set the cursor to null
2. Refresh the page

## Current Behavior

`b` has no superscript until something causes a reload. 

## Expected Behavior

`b` should have a superscript of 2.

ruchijn:
In `pull.ts` thoughts fetching is async await at https://github.com/cybersemics/em/blob/dev/src/action-creators/pull.ts#L68. 

By the time `state` is being read at https://github.com/cybersemics/em/blob/dev/src/components/ThoughtAnnotation.tsx#L127, all the thoughts haven't been fetched. And hence we see the issue. 

If we make `ThoughtAnnotation` to read `state` again whenever `thoughts` gets updated, the issue is fixed and it works as expected. 
```
const state = useRef(store.getState())
  useEffect(() => {
    state.current = store.getState()
  }, [thoughts])
```

Please let me know if there was a reason behind not making `ThoughtAnnotation` dependent on `thoughts`. 

I ran unit tests locally and it was all good.

PR #1437, opened 2021-09-25
Fix the issue with superscript not rendering when load.

ruchijn:
Fixes: #1419 

Before:
superscript_load_before

After:
superscript_load_after

### DIFF
--- a/src/components/StaticSuperscript.tsx
+++ b/src/components/StaticSuperscript.tsx
@@ -1,14 +1,12 @@
 import React from 'react'
 
 /** Renders a given number as a superscript. */
-const StaticSuperscript = ({ n }: { n: number }) => {
-  return (
-    <span className='superscript-container'>
-      <span className='num-contexts'>
-        <sup>{n}</sup>
-      </span>
+const StaticSuperscript = ({ n }: { n: number }) => (
+  <span className='superscript-container'>
+    <span className='num-contexts'>
+      <sup>{n}</sup>
     </span>
-  )
-}
+  </span>
+)
 
 export default StaticSuperscript
--- a/src/components/ThoughtAnnotation.tsx
+++ b/src/components/ThoughtAnnotation.tsx
@@ -5,7 +5,7 @@ import { store } from '../store'
 import { REGEXP_PUNCTUATIONS } from '../constants'
 import { setCursor } from '../action-creators'
 import { decodeThoughtsUrl, getContexts, getAllChildren, theme, rootedParentOf } from '../selectors'
-import { Connected, Context, Index, SimplePath, State, ThoughtContext, Path } from '../@types'
+import { Connected, Context, Index, SimplePath, State, ThoughtContext, Path, Lexeme } from '../@types'
 import {
   ellipsizeUrl,
   equalPath,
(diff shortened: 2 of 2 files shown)


raineorshine (changes requested):
It works!

We just need to subscribe to only the relevant `Lexeme`. If you subscribe to `thoughtIndex`, then all `ThoughtAnnotations` will re-render every time a `Lexeme` changes.

ruchijn:
How do I subscribe to relevant `Lexeme`? When you said that I thought you meant `thoughtIndex`. 🤔

raineorshine:
> How do I subscribe to relevant `Lexeme`? When you said that I thought you meant `thoughtIndex`. 🤔

`thoughtIndex` is of type `Index<Lexeme>`. It contains every `Lexeme` in memory, including many that are not visible. You don't want a superscript re-rendering unless the actual number for that superscript has changed, and the actual number is determined by a single `Lexeme`. For example, if the thought "apples" appears twice in the full thoughtspace, the `Lexeme` "apple" will contain two contexts. If a different `Lexeme`, e.g. "orange" changes, we don't need to re-render the "apple" superscript.

By subscribe, I mean return from `mapStateToProps`. Whether we use `mapStateToProps`, `useSelector`, or another hook, we are effectively subscribing the component to different conditions that affect when it gets rendered. This is also called reactive programming.

Notice where `numContexts` is being calculated currently in `ThoughtAnnotation`. It calls `getContexts` with `store.getState()` directly, which will not trigger a re-render when the `Lexeme` changes.

Let me know if you need additional clarification!

raineorshine (changes requested):
Since your selector returns a new object reference (the array returned by `getContexts`), it will re-render on every state change. If you react to `numContexts` it won't suffer the same problem, since equal numbers are strictly equal.

ruchijn:
I tested with `getContexts` and `StaticSuperscript`  doesn't re-render. But I will change as per your suggestion.

raineorshine:
> I tested with `getContexts` and `StaticSuperscript` doesn't re-render.

Yes, but `ThoughtAnnotation` re-renders on every state change. Plus, it should be clear that it re-renders every time based on principle. It is returning a new object reference every time.

> And if I react to `numContexts`, then it will not re-render even when state changes.

I'm not sure what you mean. If you react to `numContexts`, then it will re-render when `numContexts` changes. We don't want it to re-render when other parts of the state change.

> Please let me know in case I am missing something. I tried with `numContexts` and it didn't re-render even when it was supposed to.

`mapStateToProps` and `useSelector` don't really work together. If `mapStateToProps` does not change its output, then `render` will not be called at all, and `useSelector` cannot re-calculate. I'm not sure, but my guess is that this is why you weren't seeing it re-render. You need to use one or the other.

raineorshine:
> `mapStateToProps` and `useSelector` don't really work together. If `mapStateToProps` does not change its output, then `render` will not be called at all, and `useSelector` cannot re-calculate. I'm not sure, but my guess is that this is why you weren't seeing it re-render. You need to use one or the other.

I was wrong about this. It appears that `useSelector` works fine when it only uses its internal `state`.

raineorshine (approved):

Issue #1438, opened 2021-09-29
Login screen should be default screen


ruchijn:
When invite system was worked on, it was decided that the default screen for users should be login. No one should be able to access the app unless authenticated. 

_NOTE:_ The app should still be accessible in local environment without logging in.

PR #1439, opened 2021-09-30
Changes made to show login screen by default when user come to app. For local development, we can still access app without authentication.

ruchijn:
Fixes: #1438


raineorshine (changes requested):
- Incorrectly redirects to login page on refresh when already logged in
- Breaks invite links

ruchijn:
When I click on gift icon to go to invites page, it is erroring out for me. I am logged in using one of my gmail accounts. 

Screen Shot 2021-10-03 at 2 58 46 PM
.

ruchijn:
> * Incorrectly redirects to login page on refresh when already logged in


Also I checked for this. In localhost URL, I wasn't able to replicate - chrome regular browser as well as incognito mode. 

login_localhost_incognito
login_localhost_regular

raineorshine:
> > * Incorrectly redirects to login page on refresh when already logged in
> 
> Also I checked for this. In localhost URL, I wasn't able to replicate - chrome regular browser as well as incognito mode.

It would not be visible on localhost since the login page is disabled there. It is reproducible on non-localhost.

> Also I tried checking it on actual iPhone. But I am not able to login. It just freezes when I click on `Sign in with google`. I thought this could be issue with my branch so checked out `dev`. But same issue.

Sign In With Google requires the domain to be whitelisted, and it does not support IP addresses. However, you can create an account with email/password authentication and log in from an IP address to test this functionality.

ruchijn:
@raineorshine - when I conencted to my local firebase database and tried using invite code system, it worked. But when I was using the default database, for existing user I am getting error. So I connected to my local and made the change. Not sure what is wrong with existig user and default database. Plz let me know if you want me to take a look into that. Thanks

raineorshine:
Each database has its own users and its own whitelisted urls. I'm not understanding exactly where you are encountering the error, but I'll try it out for myself and let you know if it works for me on the production db.

ruchijn:
I get the error as soon as I click on the gift icon to see the invitation codes.

raineorshine:
Great, it works and both problems are fixed!

> I get the error as soon as I click on the gift icon to see the invitation codes.

I'm not sure what would be causing that. You may investigate further to see if it is a bug, but if it ends up just being a misunderstanding I would prefer not to log hours for that.

Issue #1440, opened 2021-10-03
Remove invite code for domain outside of localhost


ruchijn:
Currently when we try to open the app using domain outside of localhost, it asks for an invite code. The purpose of this change request is to remove the hardcoded invite as the new invite system is in place. Now user should see the default screen as login screen which is taken care of as part of https://github.com/cybersemics/em/issues/1438. 

This will just clean up the files to remove the code related to hardcode invite code for domain outside of localhosts

PR #1441, opened 2021-10-03
Changes made to remove the hardcode invite for domains outside of localhost.

ruchijn:
Fixes: #1440


src/components/ModalWelcome.tsx
```diff
@@ -125,7 +68,7 @@ const ModalWelcome = () => {
         id='welcome'
         title='Welcome to em'
         className='popup'
-        hideModalActions={!invited}
+        hideModalActions={false}
```
raineorshine, line 71: You can simply omit the property when it is always `false`.

raineorshine (approved):

Issue #1381, opened 2021-08-04
Export Markdown (basic)


raineorshine:
Add Markdown as an export option. Start with a basic implementation that is the same as Plain Text but converts `<b>` and `<strong>` to `**`, and `<i>` and `<em>` to `*`.

ruchijn:
I was looking at if there are any readily available package for this. Seems `https://github.com/thetutlage/html-to-markdown` is a good candidate. 

Before I go down deeper, do you want to have a custom function in our codebase for this or should we entertain over the shelf option that we can find?

raineorshine:
I think a 3rd party library is a great idea. 

`html-to-markdown` only has 9 stars and hasn't been updated in 6 years ago. The main candidates appear to be [showdown](https://github.com/showdownjs/showdown), [turndown](https://github.com/mixmark-io/turndown), and [breakdance](https://github.com/breakdance/breakdance). You can use a tool like [npmcompare](https://npmcompare.com/compare/breakdance,html-to-markdown,showdown,turndown) to get an overview of popularity and activity levels of multiple npm packages. I think `turndown` is the best choice since it has the most recent activity.

ruchijn:
I was playing with different libraries. 

If my HTML export is 
```
<ul>
  <li>__ROOT__  
    <ul>
      <li>a</li>
      <li><b>b</b></li>
      <li><i>c</i></li>
      <li><b><i>d</i></b></li>
    </ul>
  </li>
</ul>
```

it generates output as belwo when I use `html-to-markdown`
```
__ROOT__
  a
  ** b **
  * c *
  ** * d ** *
```

But if I use `turndown`, it generates output as 
```
\_\_ROOT\_\_ a \*\*b\*\* \*c\* \*\*\*d\*\*\*
```

This is with the setting that comes out of box. If you have a preference then plz let me know else I will keep looking if there is a better lib.

raineorshine:
In Markdown, `li`'s are represented as lines starting with `-`, e.g. `- a`, `- b`, `- c`, etc. Our Plain Text export type already exports nested thoughts in this way. We just need to convert each thought's content to markdown.

e.g.

```md
- Normal text
  - <b>Bold</b> text
```

Should output the following Markdown:

```md
- Normal text
  - **Bold** text
```

I could see it being useful to convert up to 6 levels of nested thoughts to headings, but that's out of scope of this feature.

raineorshine:
We need to convert one thought at a time to markdown, not the exported HTML. 

[...]

PR #1442, opened 2021-10-15
Export Markdown

ruchijn:
Fixes: #1381


raineorshine (changes requested):
The code looks good!

Please add a unit test to cover the new behavior of `exportContext`.

We should avoid converting metaprogramming attributes to Markdown since they will not get imported correctly into **em**:

## Steps to Reproduce

- a
  - =pinChildren
    - true
  - **b**old [with the "b" bolded]
    - c

Export Markdown on `a`

## Current Behavior

```
- a
  - \=pinChildren
    - true
  - **b**old
    - c
```

## Expected Behavior

```
- a
  - =pinChildren
    - true
  - **b**old
    - c
```

src/components/ModalExport.tsx
```diff
@@ -498,9 +498,9 @@ const ModalExport: FC<{ context: Context; simplePath: SimplePath; cursor: Path }
                 <input
                   type='checkbox'
                   id={id}
-                  checked={checked}
```
raineorshine, line 500: Thanks for fixing this!

src/selectors/exportContext.ts
```diff
@@ -86,7 +86,9 @@ export const exportContext = (
   // If it is root, then do not convert it to markdown
   const currentThought = head(context)
   const markdownText: string =
-    format === 'text/markdown' && !isHome([currentThought]) ? turndownService.turndown(currentThought) : currentThought
+    format === 'text/markdown' && !isHome([currentThought]) && !currentThought.includes('=')
```
raineorshine, line 89: You should use `!currentThoughts.startsWith('=')` instead, otherwise it will have false positives like "1 + 1 = 2" that are not metaprogramming attributes.

src/selectors/__tests__/exportContext.ts
```diff
@@ -152,3 +152,16 @@ it('export multi-line thoughts as separate thoughts', () => {
...
+
+it('exported as markdown', () => {
+  const text = `Hello <b>wor<i>ld</i></b>`
+
+  const steps = [importText({ text }), setCursor({ path: [{ value: text, rank: 0 }] })]
+
+  // run steps through reducer flow and export as plaintext for readable test
+  const stateNew = reducerFlow(steps)(initialState())
+
+  const exported = exportContext(stateNew, [text], 'text/markdown', { excludeMeta: true })
```
raineorshine: Since `excludeMeta` is not relevant for this test, it should be omitted.

src/selectors/__tests__/exportContext.ts
```diff
@@ -152,3 +152,16 @@ it('export multi-line thoughts as separate thoughts', () => {
       - Hello
       - world`)
 })
+
+it('exported as markdown', () => {
```
raineorshine: This test is good.

We should also a have a test that ensure that metaprogramming attributes are not escaped.

PR #1443, opened 2021-10-16
Preserve only specific style attributes on import single line

ruchijn:
Fixes: #1158


src/reducers/importText.ts
```diff
@@ -84,10 +87,21 @@ const importText = (
...
+    const nodes = parse(textNormalized)
+
+    if (nodes.some(isFormattingTag)) {
+      const tagsToMerge = _.takeWhile(nodes, node => isFormattingTag(node) || node.type === 'text')
+      mergedContent = tagsToMerge.reduce((accum: string, current: HimalayaNode) => {
+        const appendContent =
+          current.type === 'text' ? current.content : current.type === 'comment' ? '' : formattingNodeToHtml(current)
+        return accum + appendContent
+      }, '')
+    }
```
raineorshine, line 100: Looks good. Parsing the input adds complexity but seems necessary to strip out the appropriate style attributes.

A better place for this logic would be inside the `strip` function. It is a good fit semantically since it extends `strip` by stripping out style attributes. This can be opt-in for performance reasons. Then `importText` doesn't need to know about `himalaya`. In general, you want to push complexity downstream.

src/reducers/importText.ts
```diff
@@ -84,10 +87,21 @@ const importText = (
   if (numLines <= 1 && !isRoam && !isRoot(path)) {
     const textNormalized = strip(convertedText, { preserveFormatting: true })
 
+    let mergedContent = textNormalized
+    const nodes = parse(textNormalized)
+
+    if (nodes.some(isFormattingTag)) {
```
raineorshine, line 93: This block of code could use a descriptive comment.

PR #1447, opened 2021-12-07
Remove tags from thoughts instead of whole exported content

### LINKED ISSUE #1446: importText: Breaks on < and />
## Steps to Reproduce

```
- a
- <
- b
- />
- c
```

Set the cursor to null and export.

## Current Behavior

```
- a
- 
- c

```

Note that exporting as HTML works correctly. Only plain text is broken.

## Expected Behavior

```
- a
- <
- b
- />
- c
```

shresthabijay:
fixes #1446 

# Cause of the issue

`exportContext` used regex to remove HTML tags from the whole exported content. So for the given content below

```
- a
- <
- b
- />
- c
```
This part gets removed.

```
<
- b
- />
```

resulting in the final content 

```
- a
- 
- c
```

# Solution

Removed the tags from the individual thought value instead.

### DIFF
--- a/src/selectors/__tests__/exportContext.ts
+++ b/src/selectors/__tests__/exportContext.ts
@@ -173,3 +173,25 @@ it('export as markdown without escaping metaprogramming attributes', () => {
   expect(exported).toBe(`- Hello **wor*ld***
   - =readonly`)
 })
+
+it('export as plain text replacing html tags only from thoughts and not from the whole exported content.', () => {
+  const text = `
+  - a
+  - <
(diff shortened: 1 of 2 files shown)


raineorshine:
Thank you, looks good!

Can you add a test for markdown too?

shresthabijay:
> Thank you, looks good!
> 
> Can you add a test for markdown too?

@raineorshine  I have added the requested test. Please review it. Thanks!

raineorshine (approved):

PR #1452, opened 2021-12-07
Implement cursor persistence

### LINKED ISSUE #1445: Implement cursor persistence with independent editing
# Issue

We only get context from the URL.  With independent editing #1349 , the path is now an array of thought ids. Since we have only context, we don't have an actual path to pull the thoughts for cursor persistence from the URL.

shresthabijay:
fixes #1445

# Changes

- Store cursor path in the URL instead of context.
- Decode URL path, pull and initialize the cursor on app load. 
- Prevent pull until cursor has been initialized.
- Use [nanoid](https://www.npmjs.com/package/nanoid) instead of `uuid` for shorter ids.
- Fix [equalArrays](https://github.com/cybersemics/em/blob/dev/src/util/equalArrays.ts) bug that returned `true` if any of the params was passed as`null`.

# Detailed Explanation

## How the previous implementation worked

- Cursor was stored in the URL as encoded context.
- `cursorInitialized` determined if the cursor decoded from the URL was loaded and initialized as `cursor`.
-  This piece of logic in `updateThoughts` reducer was responsible for the cursor persistence.
https://github.com/cybersemics/em/blob/3dc1a9cf61f05a2a5f824f25173512fad3a90014/src/reducers/updateThoughts.ts#L225-L246
   - When the initial `pull` triggers `updateThoughts`, the cursor is not yet initialized. So it decodes the URL and sets the cursor. The important thing to understand is that the cursor decoded is a ranked path that has not been verified it that actually exists.
   - Since the cursor has been set, `pullQueue` starts pulling the related thoughts.
   - The `pull` triggers the `updateThought` and every time anything is pulled the above-mentioned logic checks if the cursor has been loaded into the state yet. If not it just keeps setting decoded cursor from the  latest URL as the cursor without setting the `cursorInitialized` to `true`. If the `cursor` is found to be loaded in the state, it will set `cursorInitialized` to `true`.
   - Thus if the context in the URL is not found, `cursorInitialized` will remain `false`. The logic will be keep running until we explicitly set the cursor to a valid context and somehow trigger the `updateThoughts`. Only this will set the `cursorInitialized` to `true`.
   
  ## How new implementation works

- The first thing to know is that with independent editing we now access the thoughts using `id` instead of hashed `context`.
- So the cursor is now stored as ids in the URL instead of context.
- After the loading local state, we first decode the URL, pull the cursor, verify if it exists and initialize the cursor with `cursorInitialized` to `true`. 
- `cursorInitialized` is used to prevent the `pullQueue` from pulling anything before the cursor from the url has been decoded, pulled, and verified if it actually exists. 

## Why remove the cursor initialization logic from `updateThoughts`?

Actually, the previous implementation was a bit hacky. It decoded `cursor` from the URL and set it to `cursor` without verifying if it actually existed. The unverified `cursor` in the state was used to trigger `pullQueue`, without being certain if that path actually existed. We had to run the cursor load verification and initialization logic on `updateThoughts` because we never know when it would get pulled. This flow is quite hard to follow and understand.

In the new approach, we call `initializeCursor` which first pulls the cursor path all at once using `pull`, verifies it and then initializes the cursor. If not verified we just set the cursor to the root. Until `cursorInitialized` is set to `true` we prevent `pullQueue` to pull any other thoughts too. With this approach, we don't need to do the cursor load verification on every `updateThoughts` call. It is simple and easy to understand.


shresthabijay:
> https://github.com/cybersemics/em/issues/1445#issuecomment-988095031

@raineorshine I will implement this after you approve of this solution.

shresthabijay:

src/reducers/setCursorInitialized.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Set cursor initialized status and the cursor path after initial pull.
+ */
+const setCursorInitialized = (
+  state: State,
+  { cursor, cursorInitialized }: { cursor?: Path | null; cursorInitialized: boolean },
+) => {
+  return {
+    ...state,
+    cursor,
```
raineorshine, line 12: I'm not sure it's safe to change the cursor directly. We typically dispatch `setCursor` which keeps edit mode, `expanded`, and other cursor-related state in sync. Can we add an `init: boolean` option to `setCursor` instead of creating a separate reducer?

src/initialize.ts
```diff
@@ -101,6 +111,33 @@ export const initFirebase = async (): Promise<void> => {
...
+const initializeCursor = () => {
+  if (!store.getState().cursorInitialized) {
+    const { path } = decodeThoughtsUrl(store.getState())
+
+    // if no path in decoded from the url initialize the cursor with null
+    if (!path || isRoot(path)) {
+      store.dispatch(setCursorInitialized({ cursor: null, cursorInitialized: true }))
+    } else {
+      // pull the path thoughts
+      store.dispatch(pull(path, { force: true })).then(() => {
```
raineorshine, line 127: Is `force` necessary?

raineorshine: Can you use `await` instead of `then`?

raineorshine (commented):
Thanks for the thorough explanation! That's exactly what I was looking for.

After a little digging I figured out that we used to store the URL in local storage and restore it when the app loads at the root (overriding the URL). This was [useful in the PWA](https://github.com/cybersemics/em/commit/69699f8ab4d46df51237e6ea97b41179bf4026a5#diff-06ed7ee35f3408232e9b967df7a5e83858119eb6e5300e9a08a567e15655f41cL14) which frequently gets reset resulting in a fresh load at the root. However it was [removed in this commit](https://github.com/cybersemics/em/commit/69699f8ab4d46df51237e6ea97b41179bf4026a5#diff-06ed7ee35f3408232e9b967df7a5e83858119eb6e5300e9a08a567e15655f41cL14) for reasons that are not clear to me. This more complex behavior was probably the reason for the previous architecture. Unfortunately it is not documented well enough for me to piece together all the reasoning or determine if we should restore the local storage behavior (my fault). I will assume it is not needed unless I encounter the need during PWA usage.

> ## How new implementation works
> * The first thing to know is that with independent editing we now access the thoughts using `id` instead of hashed `context`.
> * So the cursor is now stored as ids in the URL instead of context.
> * After the loading local state, we first decode the URL, pull the cursor, verify if it exists and initialize the cursor with `cursorInitialized` to `true`.
> * `cursorInitialized` is used to prevent the `pullQueue` from pulling anything before the cursor from the url has been decoded, pulled, and verified if it actually exists.

I like eliminating the cursor initialization logic from `updateThoughts` and eliminating the unverified cursor. It was breaking the invariant that the cursor thought always exists. Though it was an elegant way to defer the pull logic to the `pullQueue`.

The downside to the new implementation is in introducing a manual pull to `initialize`. The `pullQueue` is meant to centralize pulls for easier reasoning. Spreading out pulls is not ideal. The only exception is when we call `pull` manually on export. But typically it is called automatically as part of the lifecycle of the pullQueue. However it is driven by the cursor, so it doesn't really work if the cursor is not initialized. We have to make a choice of where to put the cursor initialization logic: in `initialize` when the app loads, or in the `pullQueue` which handles pulls in a centralized place.

src/util/equalArrays.ts
```diff
@@ -6,4 +6,4 @@ export const equalArrays = <T>(a: T[], b: T[]): boolean =>
     a.length === b.length &&
     a.find &&
     // compare with null to avoid false positive for ''
-    a.find((thought, i) => b[i] !== thought)) == null
```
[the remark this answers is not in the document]
shresthabijay: Sure.

src/initialize.ts
```diff
@@ -101,6 +111,33 @@ export const initFirebase = async (): Promise<void> => {
     store.dispatch(statusActionCreator({ value: 'offline' }))
   }, OFFLINE_TIMEOUT)
 }
+
+/**
+ * Decode cursor from url, pull and initialize the cursor.
+ */
+const initializeCursor = () => {
+  if (!store.getState().cursorInitialized) {
```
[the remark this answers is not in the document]
shresthabijay: I don't think so. It will remove this check.

src/reducers/setCursorInitialized.ts
shresthabijay: Sure, that seems reasonable. I will remove this new reducerand add `init` to `setCursor`

src/initialize.ts
shresthabijay: Sure.

shresthabijay:
@raineorshine Please review the changes. I have put the cursor initialization logic in the `pullQueue`.

raineorshine:
Thanks for the update! I'm not quite satisfied with what we have yet. I think it can be better. Initializing a cursor should be rather simple architecturally, but we have ended up with additional state, reducer, action, and middleware logic.

I think the way the three-value cursorInitializationStatus variable is used is less than ideal. The call to `initializeCursor` has been added to the middleware handler which can be invoked multiple times, then the multiple invocations are compensated for by introducing the `loading` state to create idempotency. If the cursor can be initialized somewhere that is not called multiple times, we can avoid this extra state and logic.

Let's take a step back. The cursor initialization logic previous existed in `updateThoughts`. The placement was probably not correct, but setting that aside for now, it mainly faced the problem that the URL was insufficient to determine the `cursor`, since it only contained the context.

https://github.com/cybersemics/em/blob/584060eac6ae903b4a9fd1100e75e592f93859c6/src/reducers/updateThoughts.ts#L224

However with the new URL schema, we do have enough information to determine the `cursor` immediately, before any thoughts are loaded, since the URL is now a proper `Path`. Can we decode the URL and set the cursor in `initialState` and let `pullQueue` pull the thoughts? That would dramatically simplify the logic.

shresthabijay:
> However with the new URL schema, we do have enough information to determine the `cursor` immediately, before any thoughts are loaded, since the URL is now a proper `Path`. Can we decode the URL and set the cursor in `initialState` and let `pullQueue` pull the thoughts? That would dramatically simplify the logic.

That can be done. But we need to 

> Let's take a step back. The cursor initialization logic previous existed in `updateThoughts`. The placement was probably not correct, but setting that aside for now, it mainly faced the problem that the URL was insufficient to determine the `cursor`, since it only contained the context.

The other problem was it was redundantly checking if the cursor has been initialized or not in every update thought. Overall it was very confusing logic.

shresthabijay:
> However with the new URL schema, we do have enough information to determine the `cursor` immediately, before any thoughts are loaded, since the URL is now a proper `Path`. Can we decode the URL and set the cursor in `initialState` and let `pullQueue` pull the thoughts? That would dramatically simplify the logic.

Yes, we can. But I see these problems with this approach. Even though the architecture to load the cursor to the state is gonna be simple with this approach, the cursor validation process after it is gonna be complex.

Let's compare the pros and cons that I can think of the approach we have.

I prefer having the initial approach of pulling the cursor at the beginning on `initialize`. It's pros and cons are as follows:

**Pros**

- Simple and easy to follow cursor validation logic.
- No redundant checks if the cursor has loaded in `updateThoughts`.
- We have the opportunity to warn users or reset the URL to a valid one if the given URL is invalid.
- `state` doesn't have an invalid `cursor`.

**Cons**

- Manual pull.

The approach with first loading the cursor into the `initialState` and then relying on `pullQueue` has following pros and cons.

**Pros**

- No manual pull. Most of the pull is handled by `pullQueue`.

**Cons**

- A complex validation logic.
- Redundant checks if the cursor has loaded in `updateThoughts`.
- No opportunity to warn users or reset the URL to a valid one if the given URL is invalid at the beginning.
- `state` may have an invalid `cursor`.

There are downsides to both approaches, I prefer the first one because its pros far outweigh the cons. It is also a simple and easy-to-understand approach.

raineorshine:
> Still we cannot determine is the cursor is valid just from the URL.

The cursor thought may not exist, but it is properly formed. This wasn't true in the last implementation. We can't determine if the cursor exists until the cursor thought has been successfully pulled. This is an important point, and I believe it justifies a manual pull during initialization.

> The other problem was it was redundantly checking if the cursor has been initialized or not in every update thought. Overall it was very confusing logic to have in `updateThoughts`

Yes, `updateThoughts` is not the right place for it.

> **Pros**
> 
> * Simple and easy to follow cursor validation logic.
> * No redundant checks if the cursor has loaded in `updateThoughts`.
> * We have the opportunity to warn users or reset the URL to a valid one if the given URL is invalid.
> * `state` doesn't have an invalid `cursor`.
> 
> **Cons**
> 
> * Manual pull.

Good points. Additional cons are the added state, reducer, action-creator, and  ternary status with the awkward middleware logic.

> * We have the opportunity to warn users or reset the URL to a valid one if the given URL is invalid.

We should set the cursor to the closest ancestor if the cursor does not exist, either using recursive pulls or pulling all ancestors. This will only work if the full path is in the URL, which may run up again size limits.

> **Pros**
> 
> * No manual pull. Most of the pull is handled by `pullQueue`.
> 
> **Cons**
> 
> * A complex validation logic.
> * Redundant checks if the cursor has loaded in `updateThoughts`.
> * No opportunity to warn users or reset the URL to a valid one if the given URL is invalid at the beginning.
> * `state` may have an invalid `cursor`.

Good points.

> There are downsides to both approaches, I prefer the first one because its pros far outweigh the cons. It is also a simple and easy-to-understand approach.

Fair enough. It is a bit closer in my eyes. I really value encapsulation.

It is tempting to put all the logic in `pullQueue`. While it's reasonable to put the cursor initialization logic there, it doesn't really make sense for `pullQueue` to interact with the URL. `initialize` seems like a better place to decode the URL. So I am in favor of keeping `initializeCursor` in `pullQueue` with the manual `pull`. 

Let's go back to your first approach, but without the extra action and reducer. We can use `setCursor` to set `cursorInitialized`. We may not even need an `init` option. We can probably just have `setCursor` always set `cursorInitialized`.

shresthabijay:
> We should set the cursor to the closest ancestor if the cursor does not exist, either using recursive pulls or pulling all ancestors. This will only work if the full path is in the URL, which may run up again size limits.

I think this is possible. Since `pull` actually pulls every thought for the given array of thought ids. We can iteratively determine the closest path that is valid after the `pull` is complete.

raineorshine:
> > We should set the cursor to the closest ancestor if the cursor does not exist, either using recursive pulls or pulling all ancestors. This will only work if the full path is in the URL, which may run up again size limits.
> 
> I think this is possible. Since `pull` actually pulls every thought for the given array of thought ids. We can iteratively determine the closest path that is valid after the `pull` is complete.

We should probably first pull only the cursor to improve time-till-first-thought, then pull the rest in a separate pull. We can do that in a separate PR though if that's easier.

raineorshine:
> > We should probably first pull only the cursor to improve time-till-first-thought, then pull the rest in a separate pull. We can do that in a separate PR though if that's easier.
> 
> Additional pull may not be required to determine the closest ancestor. We just iterate the decoded cursor path and check from the already pulled data in the state.

If the URL is `/a/b/c/d/e`, then we should only pull `e` initially for performance reasons. We can then pull the other ancestors in a separate pull, at which time we can set the cursor to the closest ancestor.

shresthabijay:
> If the URL is `/a/b/c/d/e`, then we should only pull `e` initially for performance reasons. We can then pull the other ancestors in a separate pull, at which time we can set the cursor to the closest ancestor.

I am a little lost here. So we decode the cursor from the URL, then pull last thought from the path i.e `e`. Then set the decoded cursor ?

raineorshine:
> > If the URL is `/a/b/c/d/e`, then we should only pull `e` initially for performance reasons. We can then pull the other ancestors in a separate pull, at which time we can set the cursor to the closest ancestor.
> 
> I am a little lost here. So we decode the cursor from the URL, then pull last thought from the path i.e `e`. Then set the decoded cursor ?

Yes, we should first load `a/b/c/d/e` rather than `a`, `a/b`, `a/b/c`, `a/b/c/d`, and `a/b/c/d/e` all at once. Then we can load the answers in a separate pull. This can be done in a separate PR though. Let me know if I'm missing something.

shresthabijay:
> Yes, we should first load `a/b/c/d/e` rather than `a`, `a/b`, `a/b/c`, `a/b/c/d`, and `a/b/c/d/e` all at once. Then we can load the answers in a separate pull. This can be done in a separate PR though. Let me know if I'm missing something.

If we load only `e` at the beginning. How

> > > If the URL is `/a/b/c/d/e`, then we should only pull `e` initially for performance reasons. We can then pull the other ancestors in a separate pull, at which time we can set the cursor to the closest ancestor.
> > 
> > 
> > I am a little lost here. So we decode the cursor from the URL, then pull last thought from the path i.e `e`. Then set the decoded cursor ?
> 
> Yes, we should first load `a/b/c/d/e` rather than `a`, `a/b`, `a/b/c`, `a/b/c/d`, and `a/b/c/d/e` all at once. Then we can load the answers in a separate pull. This can be done in a separate PR though. Let me know if I'm missing something.

If we only pull `e` and set the decoded cursor, it would not appear in the tree right? Because its ancestors are missing. Or Do you mean we let the `pullQueue` first pull the root context for improved time till first thought, then pull all the cursor-related thoughts and then set the cursor?

shresthabijay:
@raineorshine  I reverted back to the original implementation. I also removed the extra reducer and action creator. `cursorInitialized` will always be set to `true` on `setCursor`.  I have yet to fully understand the pulling only last thought from the decoded cursor. So we can do that in a separate PR.

src/initialize.ts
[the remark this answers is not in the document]
raineorshine: @shresthabijay

src/initialize.ts
```diff
@@ -101,6 +111,31 @@ export const initFirebase = async (): Promise<void> => {
...
+    const { path } = decodeThoughtsUrl(store.getState())
+
+    // if no path in decoded from the url initialize the cursor with null
+    if (!path || isRoot(path)) {
+      store.dispatch(setCursor({ path: null }))
+    } else {
+      // pull the path thoughts
+      store.dispatch(pull(path, { maxDepth: 0 })).then(() => {
```
raineorshine, line 127: You can use `await` here instead of `then` syntax.

src/redux-middleware/pullQueue.ts
```diff
@@ -193,9 +193,10 @@ const pullQueueMiddleware: ThunkMiddleware<State> = ({ getState, dispatch }) =>
       pullQueue = { ...pullQueue, ...initialPullQueue() }
       updatePullQueue({ forcePull: true })
     }
+    // do not pull before cursor has been initialized
     // do not updatePullQueue if there are syncs queued or in progress
     // this gets checked again in updatePullQueue, but short circuit here if possible
-    else if (!hasPushes(getState())) {
+    else if (getState().cursorInitialized && !hasPushes(getState())) {
```
raineorshine, line 199: Out of curiosity, what happens if this check is removed?

raineorshine (changes requested):

src/redux-middleware/pullQueue.ts
shresthabijay: I think we would want `pullQueue` to start pulling after the cursor has been initialized. Else it will start pulling from the root context right from the first action dispatched. It makes sense to me that `pullQueue` starts pulling only when a cursor is known.

raineorshine: Yes, that makes sense. We don't want redundant pulls.

Note what would happen if `pullQueue` handled the initial pull. Multiple pulls would automatically be deduped with the `lastExpanded` comparison. This is good because instead of it being an edge case, it is handled as a normal case, which reduces complexity. 

In this case the two approaches are pretty close to me, so this doesn't necessitate a change. Just wanted to highlight this as an important architectural consideration.

raineorshine (approved):

PR #1454, opened 2021-12-08
Fix =focus zoom edit mode issue

### LINKED ISSUE #1448: [iOS] =focus: Edit mode broken
## Steps to Reproduce

``` 
- a
- b
  - =focus
    - Zoom
``` 

1. Set the cursor on `b`
2. Tap `b`

## Current Behavior

Cursor back

## Expected Behavior

Enter edit mode on `b`

shresthabijay:
fixes #1448 

# Cause of the issue.

This regression seems to be  introduced in this commit https://github.com/cybersemics/em/commit/e348617679eeb35479b9452a01f135802b3541d5 

When `isHiddenByAutofocus ` was replaced by `isVisible`. Strangely we cannot see the changes in this commit that removed the use of `isHiddenByAutofocus `. But `isHiddenByAutofocus ` is still available on a commit before it. So I assume this was the point of regression.

Since `isVisible` is based on `actualDistance` it doesn't properly determine if the thought is visible or not. Especially in the case of the focus zoom. A cursor that is focused zoomed will always be visible . However, the logic that calculated `isVisible` didn't include it . So this was causing the `cursorBack` issue.

# Solution

I added the necessary logic to make sure that the focus zoomed cursor has `isVisible` true. I also wrote some puppeteer tests for focus zoom.

### DIFF
--- a/src/components/Subthoughts.tsx
+++ b/src/components/Subthoughts.tsx
@@ -728,6 +728,8 @@ export const SubthoughtsComponent = ({
               (attribute(state, [...childContext, '=focus', 'Zoom'], '=bullet') === 'None' ||
                 (!!childContextEnvZoom() && attribute(state, childContextEnvZoom()!, '=bullet') === 'None'))
 
+            const appendedChildPath = appendChildPath(state, childPath, path)
+            const isChildCursor = cursor && equalPath(appendedChildPath, state.cursor)
             /*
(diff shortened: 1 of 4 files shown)


raineorshine (approved):
This is beautiful, Bijay! Perfect fix, tests, and explanation. Great work!

PR #1470, opened 2021-12-23
Fix drop end move

### LINKED ISSUE #1467: Cannot drag from beginning to end of context
## Steps to Reproduce

```
- a
  - b
  - c
```

Drag `b` after `c`

## Current Behavior

Nothing happens.

## Expected Behavior

`b` should be moved after `c`.

shresthabijay:
fixes #1467 

## Changes

- Fix drop end issue
- Fix intermittent failure of BrowserStack tests


raineorshine:
Tests are failing

raineorshine (changes requested):
Tests are failing

shresthabijay:
@raineorshine I was observing that the browser stack tests failed intermittently during github checks. I also tested it from my local machine and it sometimes failed intermittently. I looked into it and came to a convincing reason for it.

So this is how browser stack tests are running.

- Jest environment is started.
- It creates a browser stack local testing connection with a local identifier. A local identifier is used to identify the connection we make to the browser stack.
- webdriverio is used to connect to the browser stack to run the tests.


However, the problem was that the jest ran three tests in parallel. But we for each test we created a browser stack local testing connection with the same local identifier. That meant if one of the tests completes and stops the connection with that local identifier, that connection would probably close for others too. So that caused some tests to fail due to connection errors.

So now the connection is started with the unique local identifier for every test that is running in parallel. I tested it by running the tests multiple times. And it seems to be working without failure. 

However, we need to make sure we don't max out the parallel threads provided by BrowserStack. For now, we have only three tests, so it will work without problem.

Please let me know what you think about it.

src/e2e/config/webdriverio.config.js
```diff
@@ -6,7 +6,10 @@ if (!process.env.BROWSERSTACK_USERNAME) {
 
 const user = process.env.BROWSERSTACK_USERNAME
 
-const config = {
+/**
+ *
```
raineorshine, line 10: Missing JSDOC

src/e2e/config/webdriverio.config.js
```diff
@@ -17,8 +20,8 @@ const config = {
     unicodeKeyboard: true,
     build: process.env.BROWSERSTACK_BUILD_NAME || `Local - ${user} - ${new Date().toISOString().slice(0, 10)}`,
     project: process.env.BROWSERSTACK_PROJECT_NAME || 'em',
-    'browserstack.localIdentifier': process.env.BROWSERSTACK_LOCAL_IDENTIFIER || 'local-' + new Date().getTime(),
```
raineorshine, line 20: It looks like we are not using `BROWSERSTACK_LOCAL_IDENTIFIER` at all any longer. What was the purpose of `BROWSERSTACK_LOCAL_IDENTIFIER` and what is the impact of removing it?

https://docs.travis-ci.com/user/browserstack/#local-identifier

raineorshine (changes requested):
Thanks for looking into this!

> So now the connection is started with the unique local identifier for every test that is running in parallel. I tested it by running the tests multiple times. And it seems to be working without failure.

Awesome!

> However, we need to make sure we don't max out the parallel threads provided by BrowserStack. For now, we have only three tests, so it will work without problem.

How can we limit the number of parallel threads to 3 if we add additional iOS test files?

shresthabijay:
> How can we limit the number of parallel threads to 3 if we add additional iOS test files?

I am not sure about this. For now, the number of threads is equal to the number of ios test files. The limit for the basic plan is 5 parallel threads. To overcome this we can either run tests in a sequence that will increase the total test time. Or we need to find a way to instantiate a connection once before e2e tests run, share it among all the test files and close it after all the tests are executed. I have little experience on BrowserStack so there could possibly be other ways too.

src/e2e/config/webdriverio.config.js
shresthabijay: A local identifier is a unique id that identifies a local connection to the browser stack. If we are using tests in sequential order it is fine to have a single local identifier, which was actually the purpose of the `BROWSERSTACK_LOCAL_IDENTIFIER`. Since we are running the parallelized tests in jests, there were multiple connections to the browser stack that were probably causing the intermittent failure. So I removed the use of a single `BROWSER_STACK_LOCAL_IDENTIFIER` and used random id instead. If we were to run tests in sequential order it would make sense to have a single `BROWSER_STACK_LOCAL_IDENTIFIER`.

raineorshine: That makes sense. Why was it stored as an environment variable though?

raineorshine:
Cool. Thanks for sharing your ideas.

shresthabijay:
> Cool. Thanks for sharing your ideas.
> 
> Ideally we could limit concurrent tests, maybe at the web driver level.

I believe that should be done on jest level. Since it runs tests in parallel. If we could configure jest to use a maximum of 5 test files in parallel, it could be the best solution to this problem. I am not sure if we have that option in jest. However, we can always resort to running tests in sequence using `--runInBand` flag in jest.

src/e2e/config/webdriverio.config.js
shresthabijay: > That makes sense. Why was it stored as an environment variable though?

I have no idea. I think it was supposed to be changed on development so that devs working on it can check their browser stacks logs based on that identifier.

raineorshine:
Could you try `--maxWorkers`?

https://jestjs.io/docs/cli#--maxworkersnumstring

raineorshine (approved):

raineorshine:
> Could you try `--maxWorkers`?
> 
> [jestjs.io/docs/cli#--maxworkersnumstring](https://jestjs.io/docs/cli#--maxworkersnumstring)

I'll create a separate issue for this.

PR #1471, opened 2021-12-27
Add merge thought functionality

### LINKED ISSUE #1465: Merge thoughts not working
## Steps to Reproduce

``` 
- a
  - b
- c
  - a
    - d
``` 

Move `c/a` up a level

## Current Behavior

``` 

- a
  - b
- c
- a
  - b
``` 

## Expected Behavior

Thoughts should be merged recursively. Must work with pending destination thoughts as before. 

``` 
- a
  - b
  - d
- c
```

shresthabijay:
fixes #1465 

## Changes

- Nested duplicate merge
- Fix pending thought merge.
- Fix `importJSON` lexeme entries for multiple thoughts with the same value issue.
- Create separate selectors for `move` and `mergeThoughts`
- Create `mergePending` reducer to handle pending merges.
- Fix all duplicate merge and pending merge tests.

## Implementation of pending merges

`moveThoughts` uses `move` selector to move the thoughts. `move` uses `mergeThoughts` to handle the duplicate merges. When a duplicate merge is found with either `source` or `destination` pending `pendingMerges` is added to the a `pushBatch`. `push` uses `flushPendingMerges` to take the `pendingMerges` and `pull` the associated thoughts and merge using `mergePending`


src/selectors/move.ts
```diff
@@ -0,0 +1,113 @@
...
+interface RecursiveMoveData {
+  thoughtIndexUpdates: Index<Lexeme | null>
+  contextIndexUpdates: Index<Parent | null>
+  pendingMerges: PendingMerge[]
+}
+
+/**
+ * Moves a thought to a destination and handles duplicate merging if found any.
+ */
+const move = (
```
raineorshine, line 16: Why is this a selector rather than a reducer? Selectors are meant to select i.e. read data, so this is a bit confusing.

src/selectors/mergeThoughts.ts
```diff
@@ -0,0 +1,155 @@
...
+        ...updatedThoughtIndex,
+      },
+    },
+  }
+}
+
+/**
+ * Merges two given thoughts within same context and with same value.
+ */
+const mergeThoughts = (
```
raineorshine, line 61: I'm also not clear why this is a selector rather than a reducer? It sounds as if it is meant to change state.

src/reducers/moveThought.ts
```diff
@@ -167,147 +41,11 @@ const moveThought = (state: State, { oldPath, newPath, offset, skipRerank, newRa
...
-
-    return finalUpdates
-  }
-
-  const { thoughtIndexUpdates, contextIndexUpdates } = recursiveMove(movingThoughtId, newParentId, newRank, state)
+  const { thoughtIndexUpdates, contextIndexUpdates, pendingMerges } = move(state, {
+    sourceThoughtPath: oldPath,
+    destinationThoughtPath: rootedParentOf(state, newPath),
+    rank: newRank,
+  })
```
raineorshine, line 48: I'd like to discuss some architectural considerations with the new organization of logic.

What is the difference between `moveThought` and `move`? How are responsibilities divided? Does this result in proper encapsulation and an easy-to-understand low-level api? Compare with `addThought`, `editThought`, and `removeThought` which make up a clear CRUD interface.

At first glance, it looks like all functionality has moved to `move` except the reranking logic. Why not put all the logic in `moveThought` and use that everywhere? Could slightly different behaviors be achieved with an option? Which results in more complexity, the introduction of an option, or splitting out the move logic?

I look forward to hearing your thoughts. Thanks!

raineorshine (changes requested):
Hi Bijay. I've only had a few hours here and there to perform this code review until I am back from vacation, but I figured it would be worth submitting my first pass now in case you get some time to respond. I plan on doing a more in-depth code review once the high level questions are answered and I am back home. Thanks!

src/components/Thought.tsx
```diff
@@ -170,7 +171,7 @@ const mapStateToProps = (state: State, props: ThoughtContainerProps) => {
   const isCursorGrandparent =
     !isExpandedHoverTopPath && !!cursor && equalPath(rootedParentOf(state, parentOf(cursor)), path)
 
-  const isExpanded = !!expanded[headId(path)]
+  const isExpanded = !!expanded[hashContext(path)]
```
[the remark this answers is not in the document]
shresthabijay: Actually this has been introduced due to the new pending merge feature. Due to pending merges we can have same valued children within a same context even though for a while. So the hash of the context is no more a unique identifier. Since path is always unique, I preferred to used that instead. Please let me know if this change doesn't make sense or maybe we could solve this in some other way.

I agree we need to change the name to make more sense or maybe create another function called `hashPath` to that is more suited to hashing path and could be simpler too.

src/reducers/updateThoughts.ts
```diff
@@ -128,12 +130,14 @@ const updateThoughts = (
   const thoughtIndexInvalidated = lexemeCache.addMany(Object.keys(thoughtIndexUpdates))
 
   contextIndexInvalidated.forEach(key => {
+    // @MIGRATION_TODO:  Fix this. state.expanded now uses hash of the path instead of hash of context.
```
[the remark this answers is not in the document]
shresthabijay: Because it would won't work with context view later.

src/selectors/getChildren.ts
```diff
@@ -35,14 +35,19 @@ export const getParent = (state: State, context: Context): Parent | null => {
...
 export const getAllChildrenByContextHash = (state: State, thoughtId: ThoughtId): ThoughtId[] =>
   getThoughtById(state, thoughtId)?.children || noChildren
 
 /** Returns the subthoughts (as Parent) of the given context unordered. . */
 export const getAllChildrenAsThoughts = (state: State, context: Context) =>
   childIdsToThoughts(state, getAllChildren(state, context) || []) ?? []
 
+/** Returns the subthoughts (as Parent) of the given context unordered. . */
+export const getAllChildrenAsThoughtsById = (state: State, id: ThoughtId) =>
+  childIdsToThoughts(state, getAllChildrenByContextHash(state, id) || []) ?? []
```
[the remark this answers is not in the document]
shresthabijay: Got it!

src/selectors/getPathForThought.ts
```diff
@@ -0,0 +1,18 @@
+import { getThoughtById } from './index'
+import { State, ThoughtId } from '../@types'
+import { ROOT_PARENT_ID } from '../constants'
+
+// Note: This function is here for transition to independent editing. Not all low level reducers support opertaions by id instead of whole path.
+/**
+ * Traverses the thought tree upwards from the given thought and returns the rooted path.
+ */
+const getPathForThought = (state: State, thoughtId: ThoughtId): ThoughtId[] | null => {
```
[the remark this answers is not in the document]
shresthabijay: Seems like it. I will delete this.

src/reducers/moveThought.ts
shresthabijay: I was meaning to discuss this with you. The reason I separated `move` and `mergeThoughts` are as follows:


`move` moves thought to a destination and handles duplicate merging if found any. It uses `mergeThoughts` in case of duplicate thoughts. It returns thought indices updates and pending merges are found any.

`mergeThoughts` merges two given thoughts with the same value. It returns the same thing as the `move`. It uses `move` but is the logic is specialized to merge two thoughts.

These two functions are dependent on each other and help create two distinct reducers. `moveThought` and `mergePending`. I agree that they are totally not meant to be selectors. 

I just wanted a way to create these two reducers `moveThought` and `mergePending` without having to duplicate this logic since they are fundamentally dependent on the move and merge logic. However `moveThought` first tries to move the thought and only merges when needed and `mergePending` directly tries to merge the given thoughts and uses `move` logic to achieve that. I found myself spending too much time deciding how to architecture it. So instead I decided to first get it working. That was the original reason for the separation of these `move` and `mergeThought`. I just kept them in `selectors` for the time being. 😅

shresthabijay: Now that I can think of we can have `move` and `mergeThought` logic as part of the main `moveThought` reducer and then export `mergThought` logic from that file and use it in `mergePending` reducer. What do you think of that ?

src/reducers/updateThoughts.ts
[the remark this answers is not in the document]
raineorshine: Ah, I see!

src/components/Thought.tsx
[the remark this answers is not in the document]
raineorshine: Yes, we should have `hashPath` which does a simple `path.join('__SEP__')`. The the types will be correct.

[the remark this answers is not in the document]
shresthabijay: Sure.

package.json
```diff
@@ -20,6 +20,9 @@
     ]
   },
   "main": "mobile/index.tsx",
+  "jest": {
+    "resetMocks": false
+  },
```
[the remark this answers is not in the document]
shresthabijay: I forgot to remove this. I will remove it.

raineorshine:
Okay, I mapped out the relationship between all of the move-related functions. Now I'll think about the architecture.

image

shresthabijay:
> Okay, I mapped out the relationship between all of the move-related functions. Now I'll think about the architecture.
> 
> image

Sure! I will make the changes accordingly.

raineorshine:
> Sure! I will make the changes accordingly.

No, this is a map of what you have currently.

raineorshine:
Here's what I am thinking.

There are two basic operations: one to move a thought and one to merge duplicates. Sometimes a move will result in a duplicate, and sometimes a duplicate will result in a move (i.e. moving children). So they have a reciprocal relationship. Ideally these are represented by two reducers. That would be the most straightforward conceptual structure.

`mergePending` is only used by `pushQueue`, so I am going to assume we can move that into `pushQueue` so that it is module scoped and thus more encapsulated.

That leaves us with `moveThoughts`, `move`, and `mergeThoughts`, which we would like to boil down to `moveThoughts` and `mergeThoughts`. What do you think about:

- Move the `move` logic into `moveThoughts`. All the move logic is then in one place, in a reducer.
- Have `mergeThoughts` call `moveThoughts` instead of `move`

Is this possible? I think the resulting structure would be easier to understand.

shresthabijay:
> > Sure! I will make the changes accordingly.
> 
> No, this is a map of what you have currently.

I meant after you decide about the architecture.

shresthabijay:
> Here's what I am thinking.
> 
> There are two basic operations: one to move a thought and one to merge duplicates. Sometimes a move will result in a duplicate, and sometimes a duplicate will result in a move (i.e. moving children). So they have a reciprocal relationship. Ideally these are represented by two reducers. That would be the most straightforward conceptual structure.
> 
> `mergePending` is only used by `pushQueue`, so I am going to assume we can move that into `pushQueue` so that it is module scoped and thus more encapsulated.
> 
> That leaves us with `moveThought`, `move`, and `mergeThoughts`, which we would like to boil down to `moveThoughts` and `mergeThoughts`. What do you think about:
> 
> * Move the `move` logic into `moveThought`. All the move logic is then in one place, in a reducer.
> * Have `mergeThoughts` call `moveThought` instead of `move`
> 
> Is this possible? What do you think?

Seems like a good plan to me. I think it is possible.

shresthabijay:
@raineorshine I think one problem is that `moveThought` is a reducer and `mergeThoughts` is just a function that returns move updates. If we change `mergeThoughts` to a reducer then we would have redundant `update thoughts` calls. For that, we need to pass an option to skip it.

raineorshine:
I'm open to `mergeThoughts` being implemented as a selector `getMergedThoughtUpdates` if that helps.

shresthabijay:
> I'm open to `mergeThoughts` being implemented as a selector `getMergedThoughtUpdates` if that helps.

Since it would be dependent on the move which would be converted to `moveThought` reducer, it won't be possible to call a reducer from a selector.

raineorshine:
Ah, good point.

raineorshine:
> @raineorshine I think one problem is that `moveThought` is a reducer and `mergeThoughts` is just a function that returns move updates. If we change `mergeThoughts` to a reducer then we would have redundant `update thoughts` calls. For that, we need to pass an option to skip it.

Can you describe the redundancy in more detail? I'd like to understand it better to consider the cost and possible mitigations.

shresthabijay:
> I'm open to `mergeThoughts` being implemented as a selector `getMergedThoughtUpdates` if that helps.
> 
> Or we can just allow the redundant `updateThoughts`.

If we change `mergeThoughts` to a reducer, then we both `moveThoughts` and `mergeThoughts` would be recursive reducers. Since they are basically pure functions, it should be fine.

raineorshine:
Yes, that sounds like a good direction, and I'd like to see the result. Let me know if you get stuck!

shresthabijay:
@raineorshine I have made all the requested changes. All the local tests are passing as well 🎉. Please review the changes. Thanks!

src/reducers/moveThought.ts
```diff
@@ -158,39 +41,118 @@ const moveThought = (state: State, { oldPath, newPath, offset, skipRerank, newRa
...
+   */
+  const duplicateSubthought = () =>
+    childrenOfDestination.find(child => normalizeThought(child.value) === normalizeThought(sourceThought.value))
+
+  // if thought is being moved to the same context that is not a duplicate case
+  const duplicateThought = !sameContext ? duplicateSubthought() : undefined
+
+  const isPendingMerge = duplicateThought && (sourceThought.pending || duplicateThought.pending)
+
+  const newStateAfterMerge =
```
raineorshine, line 67: I feel like this should be part of the `reducerFlow`, but tell me if I'm missing something.

src/reducers/moveThought.ts
```diff
@@ -158,39 +41,118 @@ const moveThought = (state: State, { oldPath, newPath, offset, skipRerank, newRa
...
+  const destinationContext = pathToContext(state, destinationThoughtPath)
+
+  const isArchived = destinationContext?.indexOf('=archive') !== -1
+
+  // if move is used for archive then update the archived field to latest timestamp
+  const archived = isArchived ? timestamp() : destinationThought.archived
+
+  const contextIndexUpdates = newStateAfterMerge
+    ? {}
+    : {
```
raineorshine, line 84: Can you add a comment explaining what the updates are and why they are only needed if `newStateAfterMerge` is null?

src/reducers/moveThought.ts
```diff
@@ -158,39 +41,118 @@ const moveThought = (state: State, { oldPath, newPath, offset, skipRerank, newRa
...
-    updateThoughts({
-      contextIndexUpdates,
-      thoughtIndexUpdates: {},
-      recentlyEdited,
-      // load the children of the conflicted path if it's pending
-      pendingPulls: isNewContextPending ? [{ path: newPath }] : [],
-    }),
+    newStateAfterMerge
+      ? null
+      : updateThoughts({
```
raineorshine, line 124: Similarly, can you add a comment here explaining what is being updated and why it only happens if `newStateAfterMerge` is null?

src/redux-middleware/pushQueue.ts
```diff
@@ -154,6 +163,44 @@ const flushDeletes =
     }
   }
 
+/** Pull children of pending merges and dispatch move to fully merge. */
```
raineorshine, line 166: Maybe preface this with something like "When a thought is moved to another context, duplicates are recursively merged" to add more context.

raineorshine (changes requested):
Looks great!!! Just some small changes and documentation requests.

src/reducers/moveThought.ts
```diff
@@ -158,39 +41,118 @@ const moveThought = (state: State, { oldPath, newPath, offset, skipRerank, newRa
...
+  const childrenOfDestination = getChildrenRankedById(state, destinationThoughtId)
+
+  /**
+   * Find duplicate thought.
+   */
+  const duplicateSubthought = () =>
+    childrenOfDestination.find(child => normalizeThought(child.value) === normalizeThought(sourceThought.value))
+
+  // if thought is being moved to the same context that is not a duplicate case
+  const duplicateThought = !sameContext ? duplicateSubthought() : undefined
```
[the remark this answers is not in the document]
shresthabijay: Got it!

src/reducers/moveThought.ts
shresthabijay: Sure.

src/util/importJSON.ts
```diff
@@ -167,52 +156,56 @@ const saveThoughts = (
...
       const childPath: Path = skipLevel ? path : [...path, getLastAddedChild()!.id]
 
       if (block.children.length > 0) {
         return {
           ...saveThoughts(
             updatedState,
-            contextIndexUpdates,
-            thoughtIndexUpdates,
+            updatedContextIndexUpdates,
+            udpatedThoughtIndexUpdates,
```
raineorshine, line 208: `saveThoughts` takes `contextIndexUpdates` and `thoughtIndexUpdates`, yet the values passed are full `contextIndex` and `thoughtIndex` types, despite their names, as pointed out above. Did you mean to pass `insertUpdates?.contextIndexUpdates`? Or perhaps aggregate all of the updates? 

If you pass the full `contextIndex` and `thoughtIndex` values, I think it would unintentionally reinsert all thoughts.

src/util/importJSON.ts
```diff
@@ -167,52 +156,56 @@ const saveThoughts = (
...
       const getLastAddedChild = () => {
-        const parent = contextIndexUpdates[contextEncoded]
+        const parent = updatedState.thoughts.contextIndex[contextEncoded]
         return (childIdsToThoughts(updatedState, parent.children) ?? []).find(
           child => child.value === nonDuplicateValue,
         )
       }
 
+      const updatedContextIndexUpdates = mergeUpdates(accum.contextIndex, insertUpdates?.contextIndexUpdates || {})
+      const udpatedThoughtIndexUpdates = mergeUpdates(accum.thoughtIndex, insertUpdates?.thoughtIndexUpdates || {})
```
raineorshine, line 199: These names are misleading, because the values are proper `contextIndex` and `thoughtIndex` types, not updates which are typically nullable.

raineorshine (changes requested):

src/util/importJSON.ts
shresthabijay: Since adding new entries to the thought indices are kind of like updates too, so I went with that naming. However we can choose not to use `updates` as the suffix for these variable names.

src/util/importJSON.ts
shresthabijay: Actually passing these two parameters is not needed now. I have refactored it to pass the updated state at every level and just return the new added thought indices entries.

shresthabijay:
@raineorshine I have refactored the `saveThoughts` logic and also renamed the variable names as you suggested. Please review the changes. Thanks!

src/util/importJSON.ts
raineorshine: If the updates are merged into a full `contextIndex` (i.e. taken from `state.thoughts.contextIndex`), then we would consider the result to be a new `contextIndex`. If the updates are merged into another set of updates, then we would consider the result to be strictly updates. 

Note that `mergeUpdates` is *only* used to merge into a full `contextIndex`. Otherwise we would not need to delete falsey values and could simply use `{ ...contextIndexUpdates1, ...contextIndexUpdates2 }`.

raineorshine (approved):
Beautiful! Thank you!!!

PR #1477, opened 2022-01-19
Undo/Redo : Alert

karunkop:
- Added `isInline` as a proptype (defaulted to false) to check for inline styling.
- Updated State type with `undoLastAction` and `redoLastAction` to track latest action that the user performs.
- Added a function called `addLastAction` in `undoRedoReducerEnhancer` file to update state with new parameters(checks for the latestActionType and updates the state).
- Removed the interface `redoAction` and `undoAction` in undo.ts and redo.ts file. Used new dispatch method for alert action.

 ## Reason for this approach:
 - Initially used alert action creator for dispatching but the alert that rendered but it got removed in a matter of ms. Not sure 
    but the reason may be there is this logic for clearing timeout in alert action-creator.

### DIFF
--- a/src/@types/Alert.ts
+++ b/src/@types/Alert.ts
@@ -2,4 +2,5 @@ export type Alert = {
   alertType?: string
   showCloseLink?: boolean
   value: string | null
+  isInline?: boolean
 } | null
--- a/src/@types/State.ts
+++ b/src/@types/State.ts
@@ -84,6 +84,5 @@ export interface State {
   latestShortcuts: Shortcut[]
   enableLatestShorcutsDiagram: boolean
   // state variables to track last actionType executed
-  undoLastAction: string | null | undefined
-  redoLastAction: string | null | undefined
+  lastActionType: string | null | undefined
 }
--- a/src/@types/State.ts
+++ b/src/@types/State.ts
@@ -83,4 +83,7 @@ export interface State {
   user?: User
   latestShortcuts: Shortcut[]
   enableLatestShorcutsDiagram: boolean
+  // state variables to track last actionType executed
+  undoLastAction: string | null | undefined
+  redoLastAction: string | null | undefined
 }
--- a/src/App.css
+++ b/src/App.css
@@ -782,6 +782,22 @@ body.dark {
   z-index: 11; /* above breadcrumbs, toolbar, export modal and status */
 }
 
+.alert-inline {
+  position: absolute;
+  top: 6px;
+  padding: 10px 0;
+  z-index: 11;
+  align-self: center;
+  left: 40%;
+}
+
+.alert-inline-text {
+  padding: 4px 15px;
+  border-radius: 8px;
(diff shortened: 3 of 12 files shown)


raineorshine:
Thanks! I haven't reviewed the code in detail yet; see the request below. 

> - Added `isInline` as a proptype (defaulted to false) to check for inline styling.

Sounds good. 

> - Updated State type with `undoLastAction` and `redoLastAction` to track latest action that the user performs.

You should be able to extract the action type from the undo/redo history which is stored in state.patches. We don't want to add another state variable that that can be derived from information that is already there. 

>  - Initially used alert action creator for dispatching but the alert that rendered but it got removed in a matter of ms. Not sure but the reason may be there is this logic for clearing timeout in alert action-creator.
> 
> 

Do you have steps to reproduce? I have seen this bug before. If so, please create a new issue so we can address this.

karunkop:
> 



> Thanks! I haven't reviewed the code in detail yet; see the requests below first.
> 
> > * Added `isInline` as a proptype (defaulted to false) to check for inline styling.
> 
> Sounds good.
> 
> > * Updated State type with `undoLastAction` and `redoLastAction` to track latest action that the user performs.
> 
> You should be able to extract the action type from the undo/redo history which is stored in state.patches. We don't want to add another state variable that that can be derived from information that is already there.
> 
> > * Initially used alert action creator for dispatching but the alert that rendered but it got removed in a matter of ms. Not sure but the reason may be there is this logic for clearing timeout in alert action-creator.
> 
> Do you have steps to reproduce? I have seen this bug before. If so, please create a new issue so we can address this.



> You should be able to extract the action type from the undo/redo history which is stored in state.patches. We don't want to add another state variable that that can be derived from information that is already there.

True but the reason I did that was because in undoRedoReducerEnhancer we are already taking care of the logic for finding the last actionType. My inital approach was that I wrote the logic again as a selector for finding out the last action type from state.patches and inversePatches. But  later I thought that this is redundant because I am re-writing the same logic which is already handled. So i thought passing last actionType to the state might be the good solution. Please correct me if my thought process is wrong!

raineorshine:
> True but the reason I did that was because in undoRedoReducerEnhancer we are already taking care of the logic for finding the last actionType. My inital approach was that I wrote the logic again as a selector for finding out the last action type from state.patches and inversePatches. But later I thought that this is redundant because I am re-writing the same logic which is already handled. So i thought passing last actionType to the state might be the good solution. Please correct me if my thought process is wrong!

That's great that you identified the logic that already existed in `undoRedoReducerEnhancer`. It's awkward that the logic is "trapped" in the store enhancer. However, adding redundant state is worse than adding redundant logic. Redundant logic can always be factored out into a helper function. 

In this case, the last action type can be calculated with `state.inversePatches[state.inversePatches.length - 1]?.[0].actions[0]`. This is small enough that it doesn't even need to be factored out or shared.

PR #1478, opened 2022-01-21
Undo/Redo : Alert 

### LINKED ISSUE #988: Undo/Redo: Alert
When the user activates undo or redo, show an alert. It should contain a message that contains the action type that was undone or redone, e.g. "Undo: Archive". 

Show the alert in a new inline pill style that does not extend the full width of the screen. The standard alert style would block the user from repeatedly clicking undo or redo. Parameter use the alert reducer to support this new inline style.

karunkop:
Fixes #988

- Added isInline as a proptype (defaulted to false) to check for inline styling.
- Used state.inversePatches history to find out latest action performed.
- Removed the interface redoAction and undoAction in undo.ts and redo.ts file. Used new dispatch method for alert action
- Fixed alert timeout issue by using clearDelay params which was initially ignored

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -784,8 +784,7 @@ body.dark {
 
 .alert-inline {
   width: auto;
-  left: 50%;
-  transform: translateX(-50%) !important;
+  left: 45%;
 }
 
 .dark .alert {
@@ -796,7 +795,6 @@ body.dark {
   padding: 5px 10px;
   background-color: rgba(255, 255, 255, 0.8);
   display: inline-block;
-  text-transform: capitalize;
 }
 
 .dark .alert-text {
--- a/src/App.css
+++ b/src/App.css
@@ -784,15 +784,17 @@ body.dark {
 
 .alert-inline {
   position: absolute;
-  top: 6px;
+  top: 45px;
   padding: 10px 0;
   z-index: 11;
-  align-self: center;
-  left: 40%;
+  color: gray;
+  width: 100%;
+  text-align: center;
+  font-size: small;
 }
 
 .alert-inline-text {
-  padding: 4px 15px;
+  padding: 4px 10px;
   border-radius: 8px;
   border: 1px solid gray;
   text-transform: capitalize;
--- a/src/App.css
+++ b/src/App.css
@@ -783,21 +783,9 @@ body.dark {
 }
 
 .alert-inline {
-  position: absolute;
-  top: 45px;
-  padding: 10px 0;
-  z-index: 11;
-  color: gray;
-  width: 100%;
-  text-align: center;
-  font-size: small;
-}
-
-.alert-inline-text {
-  padding: 4px 10px;
-  border-radius: 8px;
-  border: 1px solid gray;
-  text-transform: capitalize;
+  width: auto;
+  left: 50%;
+  transform: translateX(-50%) !important;
 }
 
 .dark .alert {
@@ -808,6 +796,7 @@ body.dark {
   padding: 5px 10px;
   background-color: rgba(255, 255, 255, 0.8);
   display: inline-block;
+  text-transform: capitalize;
 }
 
 .dark .alert-text {
--- a/src/App.css
+++ b/src/App.css
@@ -782,6 +782,22 @@ body.dark {
   z-index: 11; /* above breadcrumbs, toolbar, export modal and status */
 }
 
+.alert-inline {
+  position: absolute;
+  top: 6px;
+  padding: 10px 0;
+  z-index: 11;
+  align-self: center;
+  left: 40%;
+}
+
+.alert-inline-text {
+  padding: 4px 15px;
+  border-radius: 8px;
+  border: 1px solid gray;
+  text-transform: capitalize;
+}
+
 .dark .alert {
   background-color: rgba(0, 0, 0, 0.8);
 }
--- a/src/action-creators/alert.ts
+++ b/src/action-creators/alert.ts
@@ -8,7 +8,7 @@ interface Options {
   isInline?: boolean
 }
 
-let clearAlertTimeoutId: number | null = null // eslint-disable-line fp/no-let
+let clearAlertTimeoutId: ReturnType<typeof setTimeout> | null = null // eslint-disable-line fp/no-let
 
 /**
  * Dispatches an alert action.
@@ -19,17 +19,14 @@ let clearAlertTimeoutId: number | null = null // eslint-disable-line fp/no-let
  * @param clearDelay Timeout after which alert will be cleared.
  */
 const alert =
-  (
-    value: string | FunctionComponent | null,
-    { alertType, showCloseLink, clearDelay, isInline = false }: Options = {},
-  ): Thunk =>
+  (value: string | FunctionComponent | null, { alertType, showCloseLink, clearDelay, isInline }: Options = {}): Thunk =>
   (dispatch, getState) => {
     const { alert } = getState()
 
     if (clearTimeout) {
       // if clearAlertTimeoutId !== null, it means that previous alert hasn't been cleared yet. In this case cancel previous timeout and start new.
       clearAlertTimeoutId && clearTimeout(clearAlertTimeoutId)
-      setTimeout(() => {
+      clearAlertTimeoutId = setTimeout(() => {
         dispatch({
           type: 'alert',
           alertType,
--- a/src/action-creators/alert.ts
+++ b/src/action-creators/alert.ts
@@ -1,6 +1,5 @@
 import { FunctionComponent } from 'react'
 import { Thunk } from '../@types'
-import { ALERT_TIMEOUT_VALUE } from '../constants'
 
 interface Options {
   alertType?: string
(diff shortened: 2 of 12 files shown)


src/App.css
```diff
@@ -782,6 +782,24 @@ body.dark {
...
+  text-align: center;
+  font-size: small;
+}
+
+.alert-inline-text {
+  padding: 4px 10px;
+  border-radius: 8px;
+  border: 1px solid gray;
+  text-transform: capitalize;
+}
```
raineorshine, line 801: Can we try this with exactly the same styles (position, color, size, etc) as the existing alert, just inline instead of full width? We can add other styles if it needs to stand out more, but I'd like to try the requested change only first.

src/action-creators/alert.ts
```diff
@@ -18,22 +19,26 @@ let clearAlertTimeoutId: number | null = null // eslint-disable-line fp/no-let
...
+    value: string | FunctionComponent | null,
+    { alertType, showCloseLink, clearDelay, isInline = false }: Options = {},
+  ): Thunk =>
   (dispatch, getState) => {
     const { alert } = getState()
 
     if (clearTimeout) {
       // if clearAlertTimeoutId !== null, it means that previous alert hasn't been cleared yet. In this case cancel previous timeout and start new.
       clearAlertTimeoutId && clearTimeout(clearAlertTimeoutId)
-      clearAlertTimeoutId = setTimeout(() => {
```
raineorshine, line 28: Never delete code without providing the following:

- Code to replace the intended functionality
- OR a justification of why the intended functionality is no longer relevant/needed

In this case, we need `clearAlertTimeoutId` to clear the previous timeout, otherwise previous alerts with a longer clear delay could still be dispatched. A new alert should override all previous and pending alerts.

src/action-creators/alert.ts
```diff
@@ -18,22 +19,26 @@ let clearAlertTimeoutId: number | null = null // eslint-disable-line fp/no-let
...
         dispatch({
           type: 'alert',
           alertType,
           showCloseLink,
           value: null,
+          isInline,
         })
         clearAlertTimeoutId = null
-      }, clearTimeout)
+      }, clearDelay)
```
raineorshine, line 41: Good catch! Obviously not what the developer intended.

src/reducers/alert.ts
```diff
@@ -5,18 +5,22 @@ interface Options {
...
-  ...state,
-  alert: value
-    ? {
-        alertType,
-        showCloseLink: showCloseLink !== false,
-        value,
-      }
-    : null,
-})
+const alert = (state: State, { alertType, showCloseLink, value, isInline = false }: Options) => {
```
raineorshine, line 12: You can omit `= false` and let it default to undefined.

src/shortcuts/redo.ts
```diff
@@ -1,20 +1,30 @@
...
   label: 'Redo',
   description: 'Redo',
   svg: RedoIcon,
-  exec: (dispatch: Dispatch<RedoAction>, getState) => {
+  exec: (dispatch, getState) => {
     if (!isRedoEnabled(getState())) return
+
     dispatch({ type: 'redoAction' })
+
+    const { inversePatches } = getState()
```
raineorshine, line 17: Check `patches` for redo history rather than `inversePatches`.

src/shortcuts/redo.ts
```diff
@@ -1,20 +1,30 @@
...
+
+    // Checks the last action type from inverse patch history. Checks from the state after dispatching a redo action.
+    const lastActionType = inversePatches[inversePatches.length - 1]?.[0]?.actions[0]
+
+    if (!lastActionType) return
+
+    // Ignore navigation actions
+    if (NAVIGATION_ACTIONS[lastActionType]) return
```
raineorshine, line 25: Navigation actions get bundled up and undone along with the last modifying action. We still want to dispatch an alert, we just need to be smarter about it to grab the right action type.

For example, if the last actions are `editThought`, `cursorUp`, it will undo both of them and we need to alert "Undo: Edit".

raineorshine (changes requested):

src/App.css
karunkop: Replicating the same style will not work as intended for small size screens as it will still block some of the toolbar shortcuts. Do you want me to leave as it is for now?

src/action-creators/alert.ts
karunkop: Got you! Thanks

src/reducers/alert.ts
karunkop: Sure

src/action-creators/alert.ts
```diff
@@ -5,9 +5,10 @@ interface Options {
   alertType?: string
   showCloseLink?: boolean
   clearDelay?: number
+  isInline?: boolean
 }
 
-let clearAlertTimeoutId: number | null = null // eslint-disable-line fp/no-let
+let clearAlertTimeoutId: ReturnType<typeof setTimeout> | null = null // eslint-disable-line fp/no-let
```
karunkop, line 11: Number type can't be declared for the return type of setTimeout hence replaced it with `ReturnType<typeof setTimeout>`.

src/App.css
raineorshine: Hmmm good point. I'll have to think about that. I don't like how much the new style deviates from the old style, but I can't think of an alternative right now. We can leave this as-is and I can think of some ideas after I have used it in production for a bit. Thanks!

src/action-creators/alert.ts
raineorshine: That seems reasonable. Thanks.

src/App.css
```diff
@@ -782,6 +782,12 @@ body.dark {
   z-index: 11; /* above breadcrumbs, toolbar, export modal and status */
 }
 
+.alert-inline {
+  width: auto;
+  left: 50%;
+  transform: translateX(-50%) !important;
```
raineorshine, line 788: `!important` is considered a bad practice. You can increase the specificity of the selector if you need it to override another selector. Then at least it can be overridden in the future.

src/App.css
```diff
@@ -790,6 +796,7 @@ body.dark {
   padding: 5px 10px;
   background-color: rgba(255, 255, 255, 0.8);
   display: inline-block;
+  text-transform: capitalize;
```
raineorshine, line 799: This would modify the capitalization for all alerts, even ones that need to have uncapitalized words. I suggest capitalizing text in Javascript.

src/shortcuts/redo.ts
```diff
@@ -1,20 +1,33 @@
...
 import { isRedoEnabled } from '../selectors/isRedoEnabled'
+import { alert as alertAction } from '../action-creators'
+import { NAVIGATION_ACTIONS } from '../constants'
 
-interface RedoAction {
-  type: 'redoAction'
+/**
+ * Recursively calculates last action type from patches history if it is one of the navigation actions and finally returns the action.
+ */
+const getActionTypeRecursively = (patches: Patch[], n = 1): string => {
```
raineorshine, line 10: I suggest removing "Recursively" from the name. Function names should not include reference to implementation details.

raineorshine (changes requested):
Looking good! Almost there!

src/App.css
karunkop: The problem is, there is this useSwipeToDismiss hook that provides an inline style of `transform: translateY(0px)` that doesn't allow the inline alert style to be properly centered. So I thought marking the css property important will be the solution. So in this case I am not sure on how to increase the specificity of the selector. May be I could override that with an inline style or is it non recommendable as well?

src/App.css
karunkop: Sure will do!

src/shortcuts/redo.ts
karunkop: Sure, will do!

src/App.css
raineorshine: That's fair. Try an inline style. It is preferable since it can be overridden. If that doesn't work, add a descriptive comment to the `!important` line so that if we need to change it in the future we understand the implications.

src/shortcuts/redo.ts
```diff
@@ -1,20 +1,36 @@
...
 
-interface RedoAction {
-  type: 'redoAction'
+/**
+ * Recursively calculates last action type from patches history if it is one of the navigation actions and finally returns the action.
+ */
+const getLatestActionType = (patches: Patch[], n = 1): string => {
+  const lastActionType = patches[patches.length - n]?.[0]?.actions[0]
+  if (NAVIGATION_ACTIONS[lastActionType]) return getLatestActionType(patches, n + 1)
+  return lastActionType
```
raineorshine, line 14: It's a little cleaner to do a single return statement here, but no worries.

```ts
return NAVIGATION_ACTIONS[lastActionType] ? getLatestActionType(patches, n + 1) : lastActionType
```

raineorshine (approved):

src/shortcuts/redo.ts
karunkop: Sure will keep in mind from next time!

PR #1479, opened 2022-01-21
Added markdown formatting to thoughts for bold and italics.

### LINKED ISSUE #1382: Render Markdown formatting
Render `**bold**` and `*italic*` as **bold** and *italic* when not editing.

Render `**bold**` and `*italic*` as-is while editing (when the caret is on the thought; not just when the cursor is on the thought). 

You can see how the rendered value is conditionally modified here:

https://github.com/cybersemics/em/blob/06012202024a2a6734abcc950669ea0a8aa880cf/src/components/Editable.tsx#L717-L725

ankitkarna99:
Fixes #1382

## Changes
- Created two functions
> - `formatMarkdown`: Takes a string and uses regex to replace and convert `**` and `*` to `<b>` and `<i>` respectively.
> - `removeMarkdownFormatting`: Takes a string and uses regex to invert the change done by `formatMarkdown` function.
- Changed the conditional rendering inside of `Editable` component to use above functions based on `isEditing` current value.
- Set `oldValueRef.current` to values with markdown formatting, since the formatting started to break when moving from active thought to the thought below it.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -249,7 +249,7 @@ const Editable = ({
 
   // side effect to set old value ref to head value from updated simplePath.
   useEffect(() => {
-    oldValueRef.current = value
+    oldValueRef.current = isEditing ? removeMarkdownFormatting(value) : formatMarkdown(value)
   }, [value])
 
   /** Set or reset invalid state. */
@@ -768,10 +768,10 @@ const Editable = ({
           isCursorCleared
           ? ''
           : isEditing
-          ? value
+          ? removeMarkdownFormatting(value)
           : childrenLabel.length > 0
           ? childrenLabel[0].value
-          : ellipsizeUrl(value)
+          : ellipsizeUrl(formatMarkdown(value))
       }
       placeholder={
         isCursorCleared
@@ -796,6 +796,18 @@ const Editable = ({
       style={style || {}}
     />
   )
+}
+
+/** Replaces markdown features like bold and italics via double asterisks and single asterisks with <b> and <i> tags respectively. */
+const formatMarkdown = (value: string) => {
+  return value
+    .replace(/\*\*(.*)\*\*/g, `<b class="md-bold">$1</b>`)
+    .replace(/\*(.*)\*/g, `<i class="md-italics">$1</i>`)
+}
+
+/** Replaces converted bold and italic tags from the html to double asterisks and single asterisks respectivtively. */
+const removeMarkdownFormatting = (value: string) => {
+  return value.replace(/<b class="md-bold">(.*)<\/b>/g, `**$1**`).replace(/<i class="md-italics">(.*)<\/i>/g, `*$1*`)
 }
 
 export default connect(mapStateToProps)(Editable)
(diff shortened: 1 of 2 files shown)


src/components/Editable.tsx
```diff
@@ -249,7 +249,7 @@ const Editable = ({
 
   // side effect to set old value ref to head value from updated simplePath.
   useEffect(() => {
-    oldValueRef.current = value
+    oldValueRef.current = isEditing ? removeMarkdownFormatting(value) : formatMarkdown(value)
```
raineorshine, line 252: There is no need to use `oldValueRef`, as we can just convert the markdown when rendering the ContentEditable. I will push a commit showing this.

src/components/Editable.tsx
```diff
@@ -768,10 +768,10 @@ const Editable = ({
           isCursorCleared
           ? ''
           : isEditing
-          ? value
+          ? removeMarkdownFormatting(value)
```
raineorshine, line 771: We should check `selection.isThought()` within the `isEditing case in order for this to work properly on mobile. On mobile, we only want to render the markdown when the keyboard is up.

You can simulate mobile in Chrome Developer Tools, or by hardcoding `isTouch` to true in [browser.ts](https://github.com/cybersemics/em/blob/11bd31f3a1c3b80a12a0ecdd91534c6002774c6a/src/browser.ts#L6).

raineorshine (changes requested):

src/components/Editable.tsx
ankitkarna99: Thanks for pointing this out. I've updated the code to meet the requirements, but sometimes the markdown doesn't render even if the keyboard is up. In such case, if I tap at the thought one more time, it reveals the markdown. Can you test and guide me through the expected behavior in such cases.

src/components/Editable.tsx
ankitkarna99: Your point is genuine. The current approach is forcing bidirectional conversion inside the `ContentEditable`. I've updated the code to work without updating the `oldValueRef` and applied unidirectional conversion inside of `ContentEditable`. But there was a condition that would reset the `innerHTML` to whatever is there in the `oldValueRef`, possibly to disallow redundant state updates. I've updated the condition to compare the `oldValueRef.current` with `removeMarkdownFormatting(contentRef.current?.innerHTML)`. I also fixed the not re-rendering issue by tracking previous value of `isEditing` inside a new ref.

raineorshine:
Hey Ankit! Your work looks great. Seeing the complexity of the solution (not your fault) has given me a change of heart. Rather than add more local state and complexity to the already overblown `Editable` and deal with potential touch event difficulties, I'd like to take a different approach. Let's convert markdown to HTML on paste. Then thoughts that are being edited don't ever need to be rendered two different ways. <kbd>Ctrl</kbd> + <kbd>B</kbd> works great, and I can even add toolbar buttons for users that are not as familiar with keyboard commands.

I will close this and #1382 and create a new issue for the desired change. You can work on that or a different issue. Just let me know how much time you spend on this for reference.

PR #1480, opened 2022-01-22
Fix for column 2 not initially visible while toggling hidden thoughts

### LINKED ISSUE #1474: showHiddenThoughts: Column 2 not initially visible
## Steps to Reproduce

``` 
- A
  - =view
    - Table
  - B
    - C
  - D
    - E
``` 

Toggle `showHiddenThoughts`. 

## Current Behavior

`A/=view` is visible but `A/=view/Table` is not. 

## Expected Behavior

`A/=view/Table` should be visible. 

Column 2 should always be visible in a table if column 1 is.

nepali-prabhat:
Fixes #1474 

## Preface
Rendering thought depends on `state.expanded` which in-turn depends on `state.showHiddenThoughts`. 
## The problem
`expanded` is not re-computed when toggling hidden thoughts.
## The solution
Recompute the `expanded` object when toggling hidden thoughts.

### DIFF
--- a/src/reducers/toggleHiddenThoughts.ts
+++ b/src/reducers/toggleHiddenThoughts.ts
@@ -1,9 +1,16 @@
 import { State } from '../@types'
+import { expandThoughts } from '../selectors'
 
 /** Shows or hides all hidden and metaprogramming thoughts. */
-const toggleHiddenThoughts = (state: State) => ({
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Awesome! Thanks!

PR #1481, opened 2022-01-22
Support =pin/false in thoughts irrespective of =pinChildren/true in ancestor

### LINKED ISSUE #1243: =pin/false
## Steps to Reproduce

```js
- a
  - =pinChildren
    - true
  - b
    - =pin
      - false
    - c
  - d
``` 

Set the cursor on `a`. 

## Current Behavior

`a/b` is expanded. 

## Expected Behavior

`a/b` should not be expanded. 

`=pin/false` should override all tips of expansion: pinChildren, trailing `:`, only child.

nepali-prabhat:
Fixes #1243 

## Overview
In case of thoughts where children are pinned `i.e. =pinChildren/true` , its descendant's `=pin/false` is not respected.

## The solution
Check for =pin/false while filtering visible children.

### DIFF
--- a/src/selectors/expandThoughts.ts
+++ b/src/selectors/expandThoughts.ts
@@ -146,11 +146,12 @@ function expandThoughtsRecursive(
         const isAncestor = () => isDescendant(childContext, expansionBasePathContext)
 
         /** Check if the path is equal to the expansion path. */
-        const isExpansionBasePath = equalArrays(childContext, expansionBasePathContext)
+        const isExpansionBasePath = () => equalArrays(childContext, expansionBasePathContext)
 
-        if (!isExpansionBasePath && isPinClosed(child.id)) return false
-
-        return !isFunction(value) || isExpansionBasePath || isAncestor()
+        return (
+          (!isFunction(value) || isExpansionBasePath() || isAncestor()) &&
+          (!isPinClosed(child.id) || isExpansionBasePath())
+        )
       })
 
   // expand if child is only child and its child is not url
--- a/src/selectors/expandThoughts.ts
+++ b/src/selectors/expandThoughts.ts
@@ -115,6 +115,9 @@ function expandThoughtsRecursive(
     return attribute(state, context, '=pin')
   }
 
+  /** Returns true if the child should be pinned closed. */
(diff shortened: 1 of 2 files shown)


src/selectors/expandThoughts.ts
```diff
@@ -143,9 +146,11 @@ function expandThoughtsRecursive(
         const isAncestor = () => isDescendant(childContext, expansionBasePathContext)
 
         /** Check if the path is equal to the expansion path. */
-        const isExpansionBasePath = () => equalArrays(childContext, expansionBasePathContext)
+        const isExpansionBasePath = equalArrays(childContext, expansionBasePathContext)
+
+        if (!isExpansionBasePath && isPinClosed(child.id)) return false
```
raineorshine, line 151: Rather than add a separate condition, add it to the final return statement.

src/selectors/expandThoughts.ts
```diff
@@ -143,9 +146,11 @@ function expandThoughtsRecursive(
         const isAncestor = () => isDescendant(childContext, expansionBasePathContext)
 
         /** Check if the path is equal to the expansion path. */
-        const isExpansionBasePath = () => equalArrays(childContext, expansionBasePathContext)
+        const isExpansionBasePath = equalArrays(childContext, expansionBasePathContext)
```
raineorshine, line 149: You'll want to keep `isExpansionBasePath` a function so that `equalArrays` can be short-circuited when possible (due the order of the return condition. This is for performance reasons. `equalArrays` is O(n).

raineorshine (changes requested):

src/selectors/expandThoughts.ts
nepali-prabhat: Sure

src/selectors/expandThoughts.ts
nepali-prabhat: I see. Making the change now.

raineorshine (approved):

raineorshine:
Looks great! Can we get a unit test for this? I can't recall if we have tests for `expandThoughts` directly, or it should be done with RTL or puppeteer. I'm not at my computer now so it's not easy for me to check, but let me know if you can't find the appropriate place for the unit test.

raineorshine (approved):
Excellent!

PR #1482, opened 2022-01-22
Bullet turns into leaf when parent is edited

### LINKED ISSUE #1462: Bullet turns into leaf when parent is edited
https://user-images.githubusercontent.com/750276/145678873-6e31c6a0-e743-4f1e-8d6f-4cde0ef3bce0.MOV

## Steps to Reproduce

``` 
- a
  - b
    - c
``` 

Edit `a` to a different value. 

## Current Behavior

`b`'s bullet becomes a leaf. 

## Expected Behavior

`b`'s bullet should not change.

karunkop:
Fixes #1462 

## Overview
- All the successive descendants children's bullet were turning into leaf when it's ancestor context was edited even though they had children

## Solution
- In bullet.tsx , `leaf` is passed down as a prop but it was never used. This `leaf` was responsible for tracking whether current thought should be bullet or leaf. But here in mapStateToProps same logic is being used again of finding out the children (which seems redundant to me but I have left it as it is) and just used props.leaf to keep the track.

### DIFF
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -1,7 +1,7 @@
 import React, { MouseEvent } from 'react'
 import { connect } from 'react-redux'
 import classNames from 'classnames'
-import { getLexeme, hasChildren, isContextViewActive, isPending } from '../selectors'
+import { getLexeme, isContextViewActive, isPending } from '../selectors'
 import { head } from '../util'
 import { Context, State } from '../@types'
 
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Looks good! Thanks for the overview of the problem!

I removed the other logic, as it didn't seem to be needed any more. If we get a regression, we can always put it back in and comment appropriately.

PR #1484, opened 2022-01-22
Fixed archived counts being counted in superscript.

### LINKED ISSUE #1459: Superscript should not count archived thoughts
## Steps to Reproduce

``` 
- =archive
  - a
- a
``` 

Do not toggle on `showHiddenThoughts`. `=archive` should be hidden. 

## Current Behavior

Superscript of `a` is 2. 

## Expected Behavior

`a` should not have a superscript. 

Archived thoughts should not count towards the superscript contexts when `showHiddenThoughts` is true.

ankitkarna99:
Fixes #1459

## Changes
- Removed `normalizeThought` function from `getAncesstorByValue` utility, since it removed the `=` from `=archive`, resulting in all archived thoughts being counted in the superscript.
- Fixed a typo `=archive` to `=archive` in `ThoughtAnnotation.native.tsx`

I checked all the use-cases of `getAncesstorByValue` and it seems to be used only to find meta values like `=archive`, so it seemed safe to remove the `normalizeThought` from the function.

### DIFF
--- a/src/components/ThoughtAnnotation.native.tsx
+++ b/src/components/ThoughtAnnotation.native.tsx
@@ -150,7 +150,7 @@ const ThoughtAnnotation = ({
   /** Returns true if the thought is not archived. */
   const isNotArchive = (thoughtContext: ThoughtContext) =>
     // thoughtContext.context should never be undefined, but unfortunately I have personal thoughts in production with no context. I am not sure whether this was old data, or if it's still possible to encounter, so guard against undefined context for now.
-    showHiddenThoughts || !getAncestorByValue(state, thoughtContext, '=archieve')
+    showHiddenThoughts || !getAncestorByValue(state, thoughtContext, '=archive')
 
   const numContexts = contexts.filter(isNotArchive).length + (isRealTimeContextUpdate ? 1 : 0)
 
--- a/src/components/__tests__/Superscript.ts
+++ b/src/components/__tests__/Superscript.ts
@@ -1,0 +1,48 @@
+import { screen } from '@testing-library/dom'
+import { importText } from '../../action-creators'
+import { store } from '../../store'
+import createTestApp, { cleanupTestApp } from '../../test-helpers/createRtlTestApp'
+
+beforeEach(createTestApp)
+afterEach(cleanupTestApp)
+
+it('Superscript should count all the contexts in which it is defined.', async () => {
+  store.dispatch([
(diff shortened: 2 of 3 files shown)


src/components/ThoughtAnnotation.native.tsx
```diff
@@ -150,7 +150,7 @@ const ThoughtAnnotation = ({
   /** Returns true if the thought is not archived. */
   const isNotArchive = (thoughtContext: ThoughtContext) =>
     // thoughtContext.context should never be undefined, but unfortunately I have personal thoughts in production with no context. I am not sure whether this was old data, or if it's still possible to encounter, so guard against undefined context for now.
-    showHiddenThoughts || !getAncestorByValue(state, thoughtContext, '=archieve')
+    showHiddenThoughts || !getAncestorByValue(state, thoughtContext, '=archive')
```
raineorshine, line 153: 🤦‍♀️

raineorshine (changes requested):
> I checked all the use-cases of `getAncesstorByValue` and it seems to be used only to find meta values like `=archive`, so it seemed safe to remove the `normalizeThought` from the function.

Indeed it is! Nice work.

Can you update the comment and variable names to remove references to `normalizedValue`? We always want to keep comments in line with the code.

src/components/ThoughtAnnotation.native.tsx
ankitkarna99: Thanks for reminding about the comments. They are updated now.

raineorshine (approved):
Perfect! Ready for unit test.

ankitkarna99:
Which part of the code exactly should we unit test?

raineorshine:
I think we should put the tests in `src/components/__tests__/Superscript.ts`, which doesn't exist yet. We'll need to test at the DOM level. I recommend using RTL for this. You can use https://github.com/cybersemics/em/blob/dev/src/components/__tests__/headings.ts as a starting point. You'll just need a way to select the superscript.

ankitkarna99:
@raineorshine, I have added the test-cases for the superscript component using the React Testing Library. I could not find a way to select the `sup` tag with existing functions without adding the custom test id, so instead I wrote the test other way around. I found the superscript using `screen.getByText` and expected the element to be of type `sup`.

raineorshine (approved):
Looks great!

PR #1487, opened 2022-01-25
Style attributes not stripped after <br>

### LINKED ISSUE #1486: Style attributes not stripped after <br>
When pasted text starts with an unclosed `<br>` tag, subsequent style attributes are not stripped.

There is a bug in the style stripping logic. Use the existing logic, but find out why this fails the parser. Fix the issue and add a test case [after this one](https://github.com/cybersemics/em/blob/5d5fd1285acb0ae33a7aa4a17d1dc33ed6d21e58/src/util/__tests__/importText.ts#L1117-L1130).

## Steps to Reproduce

```
<br><span style="color: pink;">Marcel Duchamp: The Art of the Possible</span>
```

## Current Behavior

The style attribute is preserved and the text is pink.

## Expected Behavior

The style attribute should be stripped and the text should be the default color.

karunkop:
Fixes #1486 

## Problem

- In strip.ts file, there were no default args provided technically although it seemed like the object provided for strip options was default. However this won't work as it will return `undefined` if the parameter is not provided.

## Solution
- Provided default params values for all strip options properties to support stripAttributes logic by default

### DIFF
--- a/src/util/__tests__/importText.ts
+++ b/src/util/__tests__/importText.ts
@@ -1129,6 +1129,21 @@ it('import single line with style attributes', () => {
 </ul>`)
 })
 
+it('import single line with style attributes and a single br tag', () => {
+  const text = `<br><span style="color: pink;">Marcel Duchamp: The Art of the Possible</span>`
+
+  const stateNew = importText(initialState(), { text })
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Great work! I've been stung by that omission before :).

PR #1489, opened 2022-01-26
Allow splitting inside of formatted text

### LINKED ISSUE #742: Split thought breaks on formatting tags

ankitkarna99:
Fixes #742 

## Changes
- Removed `selection.isThought` from the split condition, since it caused no issues with the splitting the content and all the tests were passing.


raineorshine:
Thanks! Will you write a test?

ankitkarna99:
@raineorshine I have no clue regarding which logic to test. Could you guide me?

raineorshine:
I would use the same technique as https://github.com/cybersemics/em/blob/dev/src/components/__tests__/headings.ts  but place the test in a new test file under https://github.com/cybersemics/em/tree/dev/src/shortcuts/__tests__.

ankitkarna99:
@raineorshine I added the test case that explicitly uses `newThought` action creator, but rather than using RTL test, I used **test store** to test the action creator. In the current scenario, `newThought` action splits correctly if `splitResult` is passed into it even if `selection.isThought()` is removed the split condition.

raineorshine (approved):
Looks good!

In hindsight, I think we should have done an E2E test with RTL or Puppeteer, but this is fine. `editingValue` and `splitResult` are really implementation details. Tests should ideally not be aware of implementation details.

I renamed the test to `newThought.ts`. I try to name test files the same as the files they test, as it makes it easier to find them.

ankitkarna99:
Ahh I see. Will keep this in mind while writing tests next time.

PR #1491, opened 2022-01-26
Remove false positive in for multi column mode

### LINKED ISSUE #1488: Multi column table false positive on single nested thought
Blocks #1457

## Background

`=view/Table` causes a context to be rendered in a table view instead of the default bulleted list view. When every child has the same children, then the table automatically is rendered in multi-column mode where grandchildren are interpreted and rendered as column headers. There are still some bugs to work out in multi-column mode. 

## Steps to Reproduce

```
- a
  - =view
    - Table
  - b
    - c
```

## Current Behavior

Table `a` is rendered as if it was a multi-column table. i.e. `b` is rendered as a row header and `c` is rendered as a bold column header.

image

## Expected Behavior

Table `a` should be rendered as a 2-column table. i.e. `b` is rendered as the value in row 1-col 1, and `c` is rendered as the value in row 1-col 2. 

There should be an `isMulticolumn` boolean that is true it all the conditions are met for multi-column mode. It should be only be true if there are at least two children in the parent thought (`a`). 

image

ankitkarna99:
Fixes #1488 

## Changes
- Returned false from `isMultiColumnTable` if `otherChildren` is empty.

## Cause of issue
`otherChildren.every` always returns true if the array is empty. This is most probably why we were getting a false positive from `isMultiColumn()` function.

### DIFF
--- a/src/components/__tests__/Subthoughts.ts
+++ b/src/components/__tests__/Subthoughts.ts
@@ -9,6 +9,7 @@ import Subthoughts from '../Subthoughts'
 import { Context, Path, SimplePath, State } from '../../@types'
 import { setCursorFirstMatchActionCreator } from '../../test-helpers/setCursorFirstMatch'
 import { getAllChildren } from '../../selectors'
+import { screen } from '@testing-library/dom'
 
 // type for Thoughts or Subthoughts component that has a simplePath prop
 interface ThoughtOrSubthoughtsComponent {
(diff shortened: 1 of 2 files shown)


raineorshine (changes requested):
You'll want to add the condition to `isMulticolumn` itself. That makes a lot more sense semantically. 

ankitkarna99:
@raineorshine I've added the test cases that allows multi-column to be enabled only when there is multiple levels of nesting of thoughts.

src/components/__tests__/Subthoughts.ts
```diff
@@ -593,3 +594,53 @@ describe('expand thoughts', () => {
...
+          - Color
+            - Orange
+          - Type
+            - Citrus`,
+      }),
+    ])
+
+    const thoughtRed = screen.getAllByText('Color').filter(element => element.classList.contains('editable'))[0]
+
+    expect(thoughtRed.closest('.is-multi-column')).toBeTruthy()
```
raineorshine, line 644: This is a little loose, but it's okay. It would be better to test that `is-multi-column` is set on the correct node, rather than just any ancestor.

raineorshine (approved):

src/components/__tests__/Subthoughts.ts
ankitkarna99: I thought given the constraints in the input, it was not necessary to test it.

raineorshine: That's fair.

PR #1492, opened 2022-01-28
clearThought + Paste

### LINKED ISSUE #1395: clearThought + Paste
The `clearThought` shortcut puts the cursor editable into a temporary placeholder state that allows the thought to be quickly overwritten. It is primarily used on mobile, but it can be activated on desktop with <kbd>Ctrl</kbd> + <kbd>shift</kbd> + <kbd>alt</kbd> + <kbd>C</kbd>.

## Steps to Reproduce

1. Copy text "b" to the clipboard.
2. Create thought `a`.
3. Activate `clearThought` with <kbd>Ctrl</kbd> + <kbd>shift</kbd> + <kbd>alt</kbd> + <kbd>C</kbd>.
4. Paste at the caret

## Current Behavior

- ba

The imported value gets prepended to the existing thought value. 

## Expected Behavior

- b

The imported value should overwrite the existing thought value, just like typing "b" after activating `clearThought` would do.

karunkop:
Fixes #1395 

## Overview:
- The imported value used to get prepended to the existing thought value whilst clearThought is activated.

## Solution
 - Looking at the code in importText.ts, we are normally just prepending the new imported values to the existing thought. So just checked for the condition where cursorCleared is true. If true make `right` as an empty string, else same as previous.

## Issue 1
- Create any thought 'xyz'.
- Move the cursor to the end of string.
- Paste a certain string from the clipboard (e.g 'a' )
- Imported value gets prepended instead of being appended. Don't think this is an expected behavior, right?

## Issue 2
- Create any thought 'xyz'.
- Delete the text you just created completely but make sure you are still in edit mode
- Paste a random string
- Cursor position doesn't go to end of string rather stays at front. (Potential Issue)
- Now try to delete the string you just pasted either by navigating the cursor to the end and hitting `backspace` or just press `delete` from where the cursor initially is.
- Error is shown  as `Cannot read properties of undefined (reading 'lastUpdated')` 

Let me know your thought on these isssues. If it is valid I will convert them into actual issue in github.

### DIFF
--- a/src/reducers/importText.ts
+++ b/src/reducers/importText.ts
@@ -109,7 +109,8 @@ const importText = (
     // insert the textNormalized into the destValue in the correct place
     // trim after concatenating in case destValue has whitespace
     const left = (destValue.slice(0, replaceStart ?? 0) + textNormalized).trimLeft()
-    const right = destValue.slice(replaceEnd ?? 0).trimRight()
+    // if cursorCleared is true i.e. clearThought is enabled we don't have to use existing thought to be appended
+    const right = state.cursorCleared ? '' : destValue.slice(replaceEnd ?? 0).trimRight()
     const newValue = left + right
 
     const offset = getTextContentFromHTML(left).length


raineorshine (approved):
Works great! Thanks!

PR #1501, opened 2022-01-30
clear timeout for alerts

nepali-prabhat:
Added a timeout of 2000ms for alerts

### DIFF
--- a/src/action-creators/alert.ts
+++ b/src/action-creators/alert.ts
@@ -19,7 +19,10 @@ let clearAlertTimeoutId: ReturnType<typeof setTimeout> | null = null // eslint-d
  * @param clearDelay Timeout after which alert will be cleared.
  */
 const alert =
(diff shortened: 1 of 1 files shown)


raineorshine:
The default of `clearDelay` is supposed to be undefined, i.e. never. What's your intention?

nepali-prabhat:
Oh sorry, I thought it was supposed to have a default. 
The alert for "Duplicate thoughts are not allowed within the same context." didn't have any clearDelay and I couldn't read it properly. Upon further inspection, I thought adding a default there made sense.

raineorshine:
My guess is that what you're seeing is a bug. If no clearDelay is specified, the default is that the alert stays visible until the user closes it (which obviously only works if the close button option is on). 

If you can identify steps to reproduce the bug where the alert disappears as soon as it is shown, then we can create an issue for it. I suspect that something else is inadvertently calling `alert(null)` afterwards.

Issue #1498, opened 2022-01-29
Table view: Bullet and cursor overlay are positioned incorrectly


raineorshine:
There are three different cursor locations where the bullet or cursor is positioned incorrectly. 

## Steps to Reproduce

```
- a
  - =view
    - Table
  - b
    - b1
    - b2
  - d
    - d1
    - d2
```

Change the in-app font size to 24. You can adjust the font size by clicking the large `A` link in the footer.  

## Current Behavior

1. Column 1 bullet with cursor is not vertically centered with its corresponding thought, `b`. 

![image](https://user-images.githubusercontent.com/750276/151663930-4003feb8-72a8-4ccd-b175-48f8f12576f6.jpeg)

2. The cursor overlay bullet in the first thought in column 2 is not centered over the bullet. 

![image](https://user-images.githubusercontent.com/750276/151663956-45f2680a-a7e6-4436-b0a3-8aab35ceb554.jpeg)

3. The cursor overlay bullet in the second thought in column 2 is not centered over the bullet.

![image](https://user-images.githubusercontent.com/750276/151663972-248fce54-666e-4c22-8eff-505bf546f878.jpeg)

## Expected Behavior

Bullet and cursor overlay bullet should be positioned correctly.

## Notes
 
The solution should work on all in-app font sizes on both Chrome and Safari.

nepali-prabhat:
@raineorshine 
We are using `position: "absolute"` for bullet cursor overlay. We could've also added a div below each bullets and toggled its css class. The current approach has a lot of overhead and inconsistencies. 
Is there some performance optimization we are getting from the current approach?

raineorshine:
Could you explain the two approaches in more detail, the existing one, your suggestion, and how exactly they differ? That would help me better evaluate the suggestion.

nepali-prabhat:
## The current scenario
The cursor overlay is being rendered from the `BulletCursorOverlay` component. The component renders a span with `position: absolute`. We are then manipulating the properties `top`, `left`, etc to position the cursor overlay below bullet.

## The problem
Manually setting the position properties has the following caveats:
1. The css rules are extremely difficult to keep track. There are about 61 references of `.bullet-cursor-overlay` in `App.css` most of which involve rules setting top, left, margin, etc.
2. There arise several inconsistencies with style. We are having to write css rule for each case involving inconsistency.

## Proposed solution 1
1. Instead of putting the cursor overlay in StaticThought component, add a new element in the Bullet component itself (like in code example below)
[...]

Issue #1497, opened 2022-01-29
Thought shifts slightly when bullet changes


raineorshine:
## Steps to Reproduce

```
- a
- b
```

Alternately activate indent and outdent on `b`. 

https://user-images.githubusercontent.com/750276/151663646-61115889-b0f0-4fa1-a810-655b6ddd5d93.MOV

## Current Behavior

`a` shifts a few pixels to the right when its bullet changes from a circle to a triangle. 

## Expected Behavior

The position of `a` should not change when its bullet changes shale.

PR #1505, opened 2022-02-01
Implemented cursor overlay in the bullet component itself

nepali-prabhat:
Potential fix for issue #1498 
The PR implements the proposed solution discussed [here](https://github.com/cybersemics/em/issues/1498#issuecomment-1025322736)

## Discussion
While this implementation of the cursor overlay fixes the mentioned issue, it still doesn't perfectly align the bullet and overlay in the center for all font sizes.
The reason for this is that we are using glyphs`•`and `▸` for bullet/triangle and `•` for cursor overlay. These scale at different rate as font size changes. 
One way to solve this is by writing an equation to calculate top/right property of cursor overlay based on different font sizes. 
The other way would be to use SVG for the icons.
I personally think fixing this alignment issue is important as it is something that we interact with all the time in the UI.


src/components/Bullet.tsx
```diff
@@ -66,6 +66,7 @@ const Bullet = ({
         {glyph || (showContexts ? (leaf ? '◦' : '▹') : leaf ? '•' : '▸')}
         <span
           className={classNames('glyph-bg', {
+            expanded: !leaf,
```
raineorshine, line 69: This line doesn't make sense, but maybe it's just the naming.

- A `leaf` is a thought with no children. It can't be expanded.
- A non-leaf is a thought with children. It can be expanded or collapsed, based on the result of `expandThoughts` with the cursor position.

src/App.css
```diff
@@ -1613,6 +1613,19 @@ h1 .num-contexts {
   cursor: pointer;
 }
 
+.glyph-bg {
```
raineorshine, line 1616: Let's use `.bullet-cursor-overlay` if it's not being used any more.

src/components/StaticThought.tsx
```diff
@@ -35,18 +29,13 @@ const StaticThought = ({
...
-  const isRootChildLeaf = simplePath.length === 2 && isLeaf
 
   const state = store.getState()
 
   const { value } = getThoughtByPath(state, simplePath)
 
   return (
     <div className='thought'>
-      {!(publish && (isRoot || isRootChildLeaf)) && !hideBullet && (
-        <BulletCursorOverlay simplePath={simplePath} isDragging={isDragging} />
```
raineorshine, line 47: You can delete the `BulletCursorOverlay` component now.

src/App.css
```diff
@@ -1613,6 +1613,19 @@ h1 .num-contexts {
...
+  color: transparent;
+  position: absolute;
+  transform: translateX(-100%) translateY(3%) scale(3);
+  transition: 0.2s;
```
raineorshine, line 1620: I think I prefer it without a transition.

raineorshine (commented):
Overall the approach seems reasonable to me. I made some small cleanup comments, but you have already had those in mind after the draft was approved.

The main thing I'd like to see is if this approach can keep the cursor overlay bullet centered exactly on the bullet at different font sizes.

nepali-prabhat:
> It sounds like SVG would be the better approach, but I'm open to your opinion.

I went with the SVG approach. 

I've added new css rules to handle dimming of the bullet based by manipulating `fill-opacity` property. We have to do this because we can't use css `color` property to handle dimming like we used to when the glyphs were just fonts. 
I don't know what other regressions might have been caused due to these differences. Would appreciate you pointing those out.
Also, could you please tell me all the platforms/browsers we need to support for testing features like these?

raineorshine:
Chrome (desktop and android), Safari (desktop and iOS), and Safari iOS as a PWA. 

Typically we don't test on every platform, but since we know the bullet positioning is potentially sensitive to platform, we should do so here.

nepali-prabhat:
## More changes:
1. We were rotating `.bullet` when expanding non-leaf nodes. Now, we rotate `.glyph-fg` which is svg path itself.
2. I have deleted css rules that have become obsolete due to the new changes. These include styling related to `.bullet`, `.bullet .glyph`, `.bullet-cursor-overlay`, `bullet-cursor-overlay-highlighted` . But I've not touched any rules involving `.show-context` and `children-new`.
3. Fix for issue: #1497

It seems that vertical alignment of the bullet vary across different font-sizes. As of right now, the variation is low and less noticeable. But we can decide if it's acceptable for now.

raineorshine (changes requested):
Thanks! Here are the issues I'm seeing:

- [x] ~#1497~
- [ ] #1498
  - [ ] Column 1 bullet with cursor is not vertically centered with its corresponding thought, b.
  - [x] ~The cursor overlay bullet in the first thought in column 2 is not centered over the bullet.~
  - [x] T~he cursor overlay bullet in the second thought in column 2 is not centered over the bullet.~
- [ ] The bullet and bullet overlay are too large at the default font size. This is true on both Chrome and Safari, although it is much more pronounced on Safari. They should have the same position and size as in `dev`.
- [ ] The bullet color animation has been lost. It should animate the bullet at the same rate as the thought's text. You can test this by moving the cursor back and forth between two siblings in the root.
- [ ] The triangle has rounded corners now. It should have sharp corners like in `dev`.
- [ ] The bullets are invisible when the Light theme is activated. The default theme is Dark, but a Light theme is supported. While it is not yet possible to edit the theme in the app, you can hardcode the [theme selector](https://github.com/cybersemics/em/blob/bfd5a88c26504bb7dd49913124259e0d8d4ecaa7/src/selectors/theme.ts#L31) to return `'Light'` for testing purposes.

raineorshine (changes requested):
Thanks!

- Triangle glyph looks great
- Vertical centering looks great
- Bullet animation looks great
- Dark mode looks great
- Size and positioning look great at different font sizes (with a few exceptions below)

Here are the issues I am still seeing:

- [ ] The size of the cursor overlay is too large compared to the bullet size. It should exactly match `dev`.
- [x] When in table view and the cursor is on a thought in the first column, there is too much space between the triangle glyph and the thought text, especially at small font sizes.
- [ ] The sizing and vertical centering is wrong on OSX Safari 15.1. It is correct on IOS Safari. What platform you are testing desktop Safari on? Maybe it is only OSX.

Let me know if any of these aren't clear and I can post a screenshot!

nepali-prabhat:
the cursor doesn't act weird on ipadOS and iOS safari; only on OSX. 
I fixed that bug but I'm getting a little horizontal displacement on thought when cursor changes. Its been annoying to debug because none of the parameters that affect the horizontal positioning changes and I suspect its the way OSX safari is rendering the thoughts. 
I want to confirm spending more time fixing this is worth it. Does this happen on your end too?
Side note: I am using safari v15.4; macOS monterey 12.3 beta.

raineorshine (changes requested):
Yes, I'm seeing the same thing. It's worth it for us to spend time on it, since the exact positioning and layout of thoughts in **em** is an essential part of the design.

For what it's worth, the displacement seems to have only been introduced by the last commit, cf2a9e7.

nepali-prabhat:
# The displacement issue
The following behavior would be seen in OSX safari.
ezgif-4-4536405799

# The rotated triangle alignment issue
The following behavior would be while _zooming in/out_ in OSX safari.
ezgif-4-ed4f480cf8

raineorshine:
Thanks Prabhat! I'm glad you were able to find a solution to the displacement and rotation issues. Excellent work. 

There are a couple issues I noticed:

1. On desktop Safari, the cursor overlay is off-center.

2. On mobile Safari, the cursor overlay is centered properly on the bullet, but the overall size and positioning are off. Make sure to compare your branch with the live production app to make sure there are no regressions. 

## Current Behavior

image

## Expected Behavior

image

nepali-prabhat:
Could you say what font size and zoom you're on for both of these?

raineorshine:
I am using the default font size (18) and zoom, unless otherwise noted.

raineorshine:
> I have an Andriod device.

Great! If you could take screenshots of the following examples that would be helpful. 

If you notice any discrepancies, please address them and post new screenshots only when they are all resolved. 

1. With cursor on `a`:

``` 
- a
  - b
``` 

2. With cursor on `c`. 

``` 
- a
  - b
- c
``` 

3. With cursor on `b`. 

``` 
- a
  - =view
    - Table
  - b
    - b1
  - c
    - c1
``` 

4. With cursor on `b1`. 

``` 
- a
  - =view
    - Table
  - b
    - b1
  - c
    - c1
``` 

5. With cursor on `b2`. 

``` 
- a
  - =view
    - Table
  - b
    - b1
    - b2
  - c
    - c1
```

nepali-prabhat:
I have implemented the fix for mobile safari. I have also tested in the following platforms and found no regressions:
- MacOS(12.3 Beta)  safari
- MacOS(12.3 Beta)  chrome
- iPadOS(15.4) chrome
- iOS (14.7.1) safari

However, in **_iPadOS safari_**, I am getting error saying `localStorage.getItem` is insecure. Similarly, https://em-proto.web.app also breaks with the same error and a blank screen.
Here is a screenshot: 
IMG_F3AB781FC138-1

Running the app using expo is showing me the following:
275439313_486594139540781_438654245092985352_n
Notice the three dots. I can't enter thoughts. So I couldn't test the given cases.
I get the same result when running from `dev` branch as well. Could be I'm missing some keys/config files.

Issue #1330, opened 2021-07-09
Zoomed table hides grandchildren 


raineorshine:
## Steps to Reproduce

```js
- A
  - =view
    - Table
  - =focus
    - Zoom
  - B
    - C
  - D
    - E
``` 

Set the cursor on `B`. 

## Current Behavior

 `E` is hidden

## Expected Behavior

`E` should be visible. 

Table column 2 should always be visible. The problem does not occur without `=focus/Zoom`.

raineorshine:
Hold until `=focus/Zoom` is fixed.

PR #1506, opened 2022-02-01
Removed CSS that hid cousins and its children inside of table view

### LINKED ISSUE #1457: Table column 2 not visible with =focus parent
## Steps to Reproduce

``` 
- a
  - =focus
    - Zoom
  - b
    - =pin
      - true
    - c
  - d
    - =pin
      - true
    - =view
      - Table
    - e
      - f
``` 

Set the cursor on `b`. 

## Current Behavior

`f` is hidden by the autofocus.  

## Expected Behavior

`f` should be visible. 

If a table is visible, column 2 should always be visible. 

## Notes

Only a problem with `=focus/Zoom`, `=pin/true`, and `=view/Table`. I suspect that `=focus/Zoom` is overriding the table visibility logic.

May be a duplicate of #1330. I would work on them together.

ankitkarna99:
Fixes #1457 and #1330

Changes:
- Removed CSS that hid cousins and its children inside of column 2.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -1661,20 +1661,6 @@ h1 .num-contexts {
 .distance-from-cursor-1.zoomCursor > .child:not(.editing) > .thought-container {
   color: rgba(0, 0, 0, 0);
 }
-/* hide cousins when cursor is in column 2 */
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Looks like this is a literal reversal of past intentional behavior. That usually doesn't happen! I think it's exactly what we want in this case.

PR #1508, opened 2022-02-03
Thoughts without = should not count as metaprogramming attributes 

### LINKED ISSUE #1507: Thoughts without = should not count as metaprogramming attributes
## Steps to Reproduce

```
- a
  - note
    - b
  - view
    - Table
```

## Current Behavior

`a` is rendered as a table, and has note `b`.

## Expected Behavior

`a` should not be rendered as a table, and not have a note. 

- `note` is not the same as `=note`
- `view` is not he same as `=view`
 
Metaprogramming attributes must be preceded by `=`, i.e. return true from `util/isFunction`.

karunkop:
fixes #1507 

## Overview
- Thoughts without = were  counted as metaprogramming attributes

## Problem
-  In `normalizeThought.ts`, while trimming we are excluding all the characters that are non-word including `=`. Due to this the function returns back a normal string with `=`. Thus even if we passed in `=view` in `attribute.ts` ,normalizedThought would treat `=view` as `view` and hence would work as metaprogramming attribute.

## Solution
- Initially I thought to just omit the use of normalizeThought in `getThoughtByContext.ts` file line no: 63 to be exact, but later thought that this is a function that is used extensively and it might have possible regressions. So I applied a regex expression in `isFunction.ts` to check for possible metaprogramming attribute and then used that function in `normalizedThought.ts` in `trim` function to avoid replacing `=`.

### DIFF
--- a/src/util/getThoughtIdByContext.ts
+++ b/src/util/getThoughtIdByContext.ts
@@ -8,7 +8,6 @@ import { Context, Parent, State, ThoughtId } from '../@types'
 // import { normalizeThought } from './normalizeThought'
 // import { getAllChildren } from '../selectors'
 import { isRoot } from './isRoot'
-import { normalizeThought } from './normalizeThought'
 import { childIdsToThoughts, getThoughtById } from '../selectors'
 
 // const SEPARATOR_TOKEN = '__SEP__'
@@ -60,7 +59,7 @@ const recursiveParentFinder = (
 
   const child = children.find(child => {
     const targetValue = target[targetIndex]
-    return targetValue !== undefined && normalizeThought(target[targetIndex]) === normalizeThought(child.value)
+    return targetValue !== undefined && target[targetIndex] === child.value
   })
 
   if (!child) return null
--- a/src/util/isFunction.ts
+++ b/src/util/isFunction.ts
@@ -1,2 +1,2 @@
 /** Returns true if the given value starts `=`, indicating a metaprogramming attribute i.e =view, =note etc while also considering spaces and word characters. */
-export const isFunction = (s: string) => !!s.match(/^= *\w+ *$/g)
+export const isFunction = (s: string) => !!s.match(/^=[A-Za-z]+$/g)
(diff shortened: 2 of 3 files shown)


raineorshine (changes requested):
`normalizeThought` is working as expected. We want `normalizeThought` to ignore `=`. Its purpose is to ignore exactly those kinds of special characters or punctuation in order to treat related word forms as a single lexeme. The problem is upstream. It seems like there is code that is using `normalizeThought` as part of the logic to check if something is a metaprogramming attribute. However, I don't see any reason to normalize a thought in this case.

karunkop:
> `normalizeThought` is working as expected. We want `normalizeThought` to ignore `=`. Its purpose is to ignore exactly those kinds of special characters or punctuation in order to treat related word forms as a single lexeme. The problem is upstream. It seems like there is code that is using `normalizeThought` as part of the logic to check if something is a metaprogramming attribute. However, I don't see any reason to normalize a thought in this case.
> 
> It's suspicious that `recursiveParentFinder` uses `normalizeThough` instead of doing an exact comparison. Try changing this to an exact comparison and see if the tests still pass. If not, you'll have to dig deeper to determine if it should be an exact comparison, or if a different approach needs to be taken.
> 
> https://github.com/cybersemics/em/blob/f185b80ac21b7a586aeae63fd451de0dd6cd7e15/src/util/getThoughtIdByContext.ts#L63



> `normalizeThought` is working as expected. We want `normalizeThought` to ignore `=`. Its purpose is to ignore exactly those kinds of special characters or punctuation in order to treat related word forms as a single lexeme. The problem is upstream. It seems like there is code that is using `normalizeThought` as part of the logic to check if something is a metaprogramming attribute. However, I don't see any reason to normalize a thought in this case.
> 
> It's suspicious that `recursiveParentFinder` uses `normalizeThough` instead of doing an exact comparison. Try changing this to an exact comparison and see if the tests still pass. If not, you'll have to dig deeper to determine if it should be an exact comparison, or if a different approach needs to be taken.
> 
> https://github.com/cybersemics/em/blob/f185b80ac21b7a586aeae63fd451de0dd6cd7e15/src/util/getThoughtIdByContext.ts#L63

Okay got you! I removed the use of `normalizeThought` in `recursiveParentFinder` function. But I have still replaced the logic in `isFunction.ts` with regex expression to properly check for valid metaprogramming attributes. With this regex, `=  view`  | `=view abc` are not treated as valid metaprogramming attributes

raineorshine (approved):
> Okay got you! I removed the use of `normalizeThought` in `recursiveParentFinder` function.

Great, looks like the tests are still passing.

> But I have still replaced the logic in `isFunction.ts` with regex expression to properly check for valid metaprogramming attributes. With this regex, `= view` | `=view abc` are not treated as valid metaprogramming attributes

Good call!

PR #1509, opened 2022-02-04
Fixed thoughts not spliting after paste.

### LINKED ISSUE #1461: Split thought does not work after paste
https://user-images.githubusercontent.com/750276/145586262-1559ca91-d76c-49af-848a-e394ba0f058e.MOV

## Steps to Reproduce

1. Create a new, empty thought
2. Paste "This is a test"
3. Set the caret in the middle of the thought, e.g. between "is" and "a"
4. Hit Enter

## Current Behavior

``` 
- This is a test
- 
``` 

## Expected Behavior

``` 

- This is 
- a test 

```

## Notes

The issue does not occur if the cursor is moved after the paste.

ankitkarna99:
Fixes #1461

## Problem
`editingValue` was not updated on paste.

## Changes
On the side-effect of updated head-value, we dispatch `setEditingValue` to current value, if the value is different and `isEditing` is active.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -250,7 +250,7 @@ const Editable = ({
   // side effect to set old value ref to head value from updated simplePath. Also update editing value, if it is different from current value.
   useEffect(() => {
     oldValueRef.current = value
-    if (state.editingValue !== value) {
+    if (isEditing && selection.isThought() && state.editingValue !== value) {
       dispatch(setEditingValue(value))
     }
   }, [value])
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
Looks good!

Can you try adding an RTL unit test? Make sure you can properly fail the unit test first, as the JSDOM environment that RTL uses may have some discrepancies from real browser behavior. You can use the `paste` helper as used in the [let-focus test](https://github.com/cybersemics/em/blob/7923c2a29f080e0963e5a3664ea2fc6caf90053b/src/components/__tests__/let-focus.ts#L10-L18). It calls `importText` directly since an actual paste user event cannot be done in RTL [or in puppeteer](https://stackoverflow.com/questions/57101467/how-do-you-paste-text-using-puppeteer). 

I'll run the unit test before and after the fix to make sure it tests what it says it does. If the `paste` helper doesn't work, just let me know. It may not be possible to test this behavior.

Thanks!

PR #1511, opened 2022-02-08
Alert flashes and disappears after move

### LINKED ISSUE #1510: Alert flashes and disappears after move
## Steps to Reproduce

```
- a
- b
```

Move `b` into `a`. 

## Current Behavior

The move alert flashes briefly and then disappears. 

## Expected Behavior

The move alert should disappear after a short delay. 

The delay should already be in the code. There's probably a call to `alert(null)` that is interrupting it. If that is the case, carefully consider the existing use of `alert(null)` to ensure there are no regressions.

karunkop:
Fixes #1510 

## Problem
- `endDrag` handler function in  `DragAndDropThought.tsx` was dispatching a redudant `alert(null)` action.

## Solution
- Removed the `alert(null)` action from endDrag handler function
- Removed the use of manual clearance of alert using `window.setTimeout()`. I'm pretty sure after the latest fix of alert action 
   creator these conditions gets handled automatically.

### DIFF
--- a/src/components/DragAndDropThought.tsx
+++ b/src/components/DragAndDropThought.tsx
@@ -82,7 +82,7 @@ const beginDrag = ({ simplePathLive }: ConnectedThoughtContainerProps) => {
 
 /** Handles drag end. */
 const endDrag = () => {
-  store.dispatch([dragInProgress({ value: false }), dragHold({ value: false }), alert(null)])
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Thanks! That manual `setTimeout` must have been done before the alert supported its own clear delay.

Thank you for providing justification for code deletion.

PR #1512, opened 2022-02-08
Empty thought with formatting tags breaks caret

### LINKED ISSUE #1503: Empty thought with formatting tags breaks caret
## Steps to Reproduce

1. Paste `<b></b>` into a new thought.
2. Hit enter
3. Hit backspace

## Current Behavior

The caret disappears.

## Expected Behavior

The caret should be set on the previous thought after backspace.

karunkop:
Fixes #1503 

## Problem
- Empty formatting tags such as `<b></b>`, `<i></i>` when pasted were resolved even when there is no content.

## Solution
- Used regex expression to replace empty formatting tags with space which will restrict from pasting empty formatting tag ultimately resolving the caret issue.

### DIFF
--- a/src/util/__tests__/strip.ts
+++ b/src/util/__tests__/strip.ts
@@ -45,6 +45,34 @@ it('preserve bold with attribute', () => {
   )
 })
 
+it('empty formatting tags', () => {
+  expect(strip('<b></b>', { preserveFormatting: true })).toBe('')
+})
+
+it('empty formatting tags multiline', () => {
+  expect(
+    strip(
+      `<b>
+  
+      </b>`,
+      { preserveFormatting: true },
+    ),
+  ).toBe('')
+})
+
+it('leading and trailing whitespaces in multiline formatting tags', () => {
+  expect(
+    strip(
+      `<b>
+      
+      Hello
+
+      </b>`,
+      { preserveFormatting: true },
+    ),
+  ).toBe('<b> Hello </b>')
+})
+
 // This fails here, but succeeds on regex101.com (???)
 it.skip('preserve bold with newlines in attribute', () => {
   expect(
--- a/src/util/__tests__/strip.ts
+++ b/src/util/__tests__/strip.ts
@@ -60,19 +60,6 @@ it('empty formatting tags multiline', () => {
   ).toBe('')
 })
 
-it('leading and trailing whitespaces in multiline formatting tags', () => {
-  expect(
-    strip(
-      `<b>
-      
-      Hello
-
-      </b>`,
-      { preserveFormatting: true },
-    ),
-  ).toBe('<b> Hello </b>')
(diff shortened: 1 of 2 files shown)


src/util/strip.ts
```diff
@@ -23,6 +25,7 @@ export const strip = (
     .replace(regexSpanTagOnlyContainsWhitespaces, '$1') // Replace span tags contain whitespaces
     .replace(regexNbsp, ' ')
     .replace(regexDecimalSpace, ' ') // Some text editors use decimal code for space character
+    .replace(regexExpForEmptyFormattingTags, '') // Replace empty formatting tags with space
```
raineorshine, line 28: It removes empty formatting tags, not replace them with space.

src/util/strip.ts
```diff
@@ -12,6 +12,8 @@ type StripOptions = { preserveFormatting?: boolean; preventTrim?: boolean; strip
 
 const regexSpanTagOnlyContainsWhitespaces = /<span[^>]*>([\s]+)<\/span>/gim
 
+const regexExpForEmptyFormattingTags = /<[^/>][^>]*> *<\/[^>]+>/gim
```
raineorshine, line 15: I would change the space to any whitespace to handled cases like `<b>\n\n\n</b>`

src/util/strip.ts
karunkop: True! Btw do we want to remove empty tags or replace them with spaces?

src/util/__tests__/strip.ts
```diff
@@ -45,6 +45,34 @@ it('preserve bold with attribute', () => {
...
+    strip(
+      `<b>
+  
+      </b>`,
+      { preserveFormatting: true },
+    ),
+  ).toBe('')
+})
+
+it('leading and trailing whitespaces in multiline formatting tags', () => {
```
karunkop, line 63: Unit test case for leading/trailing whitespaces inside formatting tags. Apparently app used to crash when pasting such multiline tag with whitespaces.

src/util/strip.ts
```diff
@@ -12,6 +12,10 @@ type StripOptions = { preserveFormatting?: boolean; preventTrim?: boolean; strip
 
 const regexSpanTagOnlyContainsWhitespaces = /<span[^>]*>([\s]+)<\/span>/gim
 
+const regexExpForEmptyFormattingTags = /<[^/>][^>]*>\s*<\/[^>]+>/gim
+
+const regexExpForWhiteSpacesInTags = /[\s]+/gm
```
karunkop, line 17: Regex to replace trailing/leading multiline whitespaces with space. The reason for replacing here with space is to respect certain scenarios like `Hello <b>world</b>` with inline spaces.

raineorshine (approved):

PR #1514, opened 2022-02-11
Alert flashes and disappears after gesture

### LINKED ISSUE #1513: Alert flashes and disappears after gesture
Similar to #1510, just a different triggering action.

## Steps to Reproduce

- Open Chrome Dev Tools and enable the device toolbar to emulate mobile
- Refresh the page
- Initiate any gesture and do not release at the end

## Current Behavior

The gesture alert flashes briefly and then disappears. 

## Expected Behavior

The gesture alert should disappear only after the touch is released.

Note that this is a regression and the logic to close the alert after the touch is released already exists. There's probably a call to `alert(null)` that is interrupting it. If that is the case, carefully consider the existing use of `alert(null)` to ensure there are no regressions.

karunkop:
Fixes #1513 

## Problem

- The main problem lied in the main alert logic i.e `alert.ts` (action-creator). Default behavior of alert should be that if the value for `clearDelay` is not provided explicitly `alert` component shouldn't dismiss at all. But since we are not checking the logic `alert` dismisses soon if `clearDelay` is not provided.

## Solution

- Checked for whether `clearDelay` param has been provided or not. If not then don't dispatch `alert(null)`

### DIFF
--- a/src/action-creators/alert.ts
+++ b/src/action-creators/alert.ts
@@ -24,7 +24,7 @@ const alert =
     const { alert } = getState()
 
     // if clearDelay is not provided i.e undefined alert should not dismiss.
-    if (clearTimeout && clearDelay) {
+    if (clearDelay) {
       // if clearAlertTimeoutId !== null, it means that previous alert hasn't been cleared yet. In this case cancel previous timeout and start new.
       clearAlertTimeoutId && clearTimeout(clearAlertTimeoutId)
       clearAlertTimeoutId = setTimeout(() => {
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Thanks! Looks like we missed this in 74d85afa5b66f5fddd86f269a013ba878ec24497.

karunkop:
> Thanks! Looks like we missed this in [74d85af](https://github.com/cybersemics/em/commit/74d85afa5b66f5fddd86f269a013ba878ec24497).

Yeah! Should have done that before. My bad!

PR #1529, opened 2022-03-10
importJSON error

### LINKED ISSUE #1520: importJSON error
## Steps to Reproduce

```
a
b
c (<sup>d</sup>)
```

Paste the above thoughts.

## Current Behavior

```
Unhandled Rejection (TypeError): Cannot read properties of undefined (reading 'trim')
L138 importJSON.ts
```

## Expected Behavior

```
- a
- b
- c (d)
```

karunkop:
Fixes #1520 

## Solution

- Strip all the unnecessary tags i.e not included as formatting tags from the html itself right at the time when we are importing a text.

### DIFF
--- a/src/util/htmlToJson.ts
+++ b/src/util/htmlToJson.ts
@@ -200,9 +200,35 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
   return result
 }
 
+/** Gets all unique tag lists that is included in the text that we import. */
+const allTagsFinder = (nodes: Element[]): string[] => {
+  const tagLists = nodes.reduce((accum, current) => {
+    if (current?.children?.length) return [...accum, current.tagName, ...allTagsFinder(current.children as Element[])]
+    return current.tagName ? [...accum, current.tagName] : [...accum]
+  }, [] as string[])
+  return _.uniq(tagLists)
+}
+
+const tagsThatIsNotToBeStripped = [...allowedFormattingTags, 'span', 'li', 'ul']
+
+/** Generates dynamic regex expression. */
+const regexGenerator = (tags: string[]): RegExp => {
+  const regexString = tags
+    .reduce((accum, current) => {
+      return [...accum, `<${current}>`, `</${current}>`]
+    }, [] as string[])
+    .join('|')
+  return new RegExp(regexString, 'gim')
+}
+
 /** Parses input HTML and saves in JSON array using Himalaya. */
 export const htmlToJson = (html: string) => {
-  const nodes = parse(html)
-  const blocks = himalayaToBlock(removeEmptyNodesAndComments(nodes))
+  const nodes = parse(html) as Element[]
+
+  const tags = allTagsFinder(nodes)
+  const tagsToBeStripped = tags.filter((tag: string) => !tagsThatIsNotToBeStripped.includes(tag))
+  const regex = regexGenerator(tagsToBeStripped)
+  const strippedHtml = html.replace(regex, '')
+  const blocks = himalayaToBlock(removeEmptyNodesAndComments(parse(strippedHtml)))
   return Array.isArray(blocks) ? blocks : [blocks]
 }


src/util/htmlToJson.ts
```diff
@@ -200,9 +200,35 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
...
+    if (current?.children?.length) return [...accum, current.tagName, ...allTagsFinder(current.children as Element[])]
+    return current.tagName ? [...accum, current.tagName] : [...accum]
+  }, [] as string[])
+  return _.uniq(tagLists)
+}
+
+const tagsThatIsNotToBeStripped = [...allowedFormattingTags, 'span', 'li', 'ul']
+
+/** Generates dynamic regex expression. */
+const regexGenerator = (tags: string[]): RegExp => {
```
raineorshine, line 215: `regexGenerator` is too generic a name for what this function does. It specifically generates a regex to remove tags, so I would call it `generateRegexToMatchTags` or `regexOfTags`.

src/util/htmlToJson.ts
```diff
@@ -200,9 +200,35 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
...
+/** Gets all unique tag lists that is included in the text that we import. */
+const allTagsFinder = (nodes: Element[]): string[] => {
+  const tagLists = nodes.reduce((accum, current) => {
+    if (current?.children?.length) return [...accum, current.tagName, ...allTagsFinder(current.children as Element[])]
+    return current.tagName ? [...accum, current.tagName] : [...accum]
+  }, [] as string[])
+  return _.uniq(tagLists)
+}
+
+const tagsThatIsNotToBeStripped = [...allowedFormattingTags, 'span', 'li', 'ul']
```
raineorshine, line 212: `tagsThatAreNotToBeStripped`

src/util/htmlToJson.ts
```diff
@@ -200,9 +200,35 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
...
+
+/** Generates dynamic regex expression. */
+const regexGenerator = (tags: string[]): RegExp => {
+  const regexString = tags
+    .reduce((accum, current) => {
+      return [...accum, `<${current}>`, `</${current}>`]
+    }, [] as string[])
+    .join('|')
```
raineorshine, line 220: I think this is equivalent, and simpler: `` `</?${tags.join('|')}>` ``

src/util/htmlToJson.ts
```diff
@@ -200,9 +200,35 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
   return result
 }
 
+/** Gets all unique tag lists that is included in the text that we import. */
+const allTagsFinder = (nodes: Element[]): string[] => {
```
raineorshine, line 204: Let's call this `findUniqueTags`.

src/util/htmlToJson.ts
```diff
@@ -200,9 +200,35 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
...
+  const tagLists = nodes.reduce((accum, current) => {
+    if (current?.children?.length) return [...accum, current.tagName, ...allTagsFinder(current.children as Element[])]
+    return current.tagName ? [...accum, current.tagName] : [...accum]
```
raineorshine, line 207: You can write this as a single ternary statement with multiple branches:

```
return a ? a1
	: b ? b1
  : c ? c1
  : d
```

In general you want to avoid multiple return statements in a function, unless it is a small short-circuiting return statement at the beginning of the function. A single exit point is easier to reason about.

raineorshine (changes requested):
Thanks! The loops and dynamic RegExp are pretty inefficient, but it may be okay. We can reexamine it if we notice any performance issues.

Try to name your functions as verbs rather than nouns. e.g. a "generator" sounds like a thing, but a function is not a thing, it's an action. The one exception I can think of is sometimes I give functions a noun name that corresponds to what they return, e.g. `uniqueNodeTags`.

src/util/htmlToJson.ts
karunkop: True! Thanks for the suggestion :)

karunkop: I don't think this will work when there are 'n' number of tags in text.

raineorshine: https://regexr.com/6h79j

karunkop:
> Thanks! The loops and dynamic RegExp are pretty inefficient, but it may be okay. We can reexamine it if we notice any performance issues.
> 
> Try to name your functions as verbs rather than nouns. e.g. a "generator" sounds like a thing, but a function is not a thing, it's an action. The one exception I can think of is sometimes I give functions a noun name that corresponds to what they return, e.g. `uniqueNodeTags`.
> 
> Please add a unit test at your convenience.

Where should I actually put in a unit test case for this? I am little bit confused.

raineorshine:
https://github.com/cybersemics/em/blob/dev/src/util/__tests__/importText.ts

raineorshine (approved):

PR #1536, opened 2022-03-17
Imported/Paste Text prepends despite of cursor position

### LINKED ISSUE #1493: Imported/Paste Text prepends even when irrespective of cursor position
## Steps to reproduce
- Create any thought 'xyz'.
- Move the cursor to the end of string.
- Paste a certain string from the clipboard (e.g 'a' )

## Current Behavior
- Imported value gets prepended instead of being appended.

## Expected Behavior
- Imported value should prepend/append/insert based upon cursor position

karunkop:
Fixes #1493 

## Problem
- When importing/pasting text as a thought, caret position was never updated. Normally when pasting caret position should update and move towards end of pasted string
-  The pasted text was always prepended irrespective of the current caret position.

## Solution
- Instead of just concatenation of `left` and `right` values in`importText.ts`, first calculate the actual current position of the caret. Then based upon that append/prepend/insert operation is done. For this `selection.offset()` method is used to correctly track the caret position. New offset is hence calculated by current caret position + pastedText.length.


src/device/selection.ts
```diff
@@ -231,12 +231,6 @@ export const set = (
   const range = document.createRange()
   const sel = window.getSelection() || new Selection()
 
-  // bail if already selected
```
karunkop, line 234: ## Reason for removing this particular logic
- So this scenario becomes different when we are trying to paste something on the same node(focusNode). With this logic it will never allow the caret position to be set as we will be pasting content on the same node. Hence I don't think this logic is necessary. I am not fully sure at the first place why this logic exists. If you have any objections regarding this please enlighten me!

src/reducers/importText.ts
```diff
@@ -107,14 +107,15 @@ const importText = (
...
     // if cursorCleared is true i.e. clearThought is enabled we don't have to use existing thought to be appended
-    const right = state.cursorCleared ? '' : destValue.slice(replaceEnd ?? 0).trimRight()
-    const newValue = left + right
-
-    const offset = getTextContentFromHTML(left).length
-
+    const replacedDestValue = state.cursorCleared
+      ? ''
+      : destValue.slice(0, replaceStart || 0) + destValue.slice(replaceEnd || 0)
+    const caretPosition = replaceStart || selection.offset() || 0
```
raineorshine, line 114: Can you use `state.cursorOffset` instead of `selection.offset()`? I'm not sure if it will work, but it would be nice to not need the `selection` dependency.

raineorshine (changes requested):
It's not quite working for me yet.

## Steps to Reproduce

1. Create thought `abc`
2. Create thought `xyz`
3. Copy the text `xyz`
4. Delete the text `xyz`
5. Delete the empty thought where `xyz` was, making the caret move to the end of the previous thought
6. Paste `xyz`

https://user-images.githubusercontent.com/750276/158827183-1f6501be-0976-46f1-8f82-6384826f072b.mov

## Current Behavior

- axyzbc

## Expected Behavior

- abcxyz

src/reducers/importText.ts
karunkop: > It's not quite working for me yet.

Yes I saw this. The reason being selection.offset() is not returning correct caret position when caret is placed at the end of string by hitting backspace. Technically it should have provided the value that is equal to the length of string in focused thought. I am still looking at this and finding how do i correctly track the caret position.

raineorshine:
Unfortunately this doesn't fix it in all cases (1. create a new thought, 2. type `a`, 3. paste).

Let's take a step back and try again, since the introduction of global state is undesirable, as you are aware.

> Hey raine! So as you suggested I tried to use  state.cursorOffset instead of selection.offset(). This works just as intended when we switch/set cursor to next thought and then navigate the caret with keyboard and paste the text. **However if  i don't switch and stay on the same thought and then navigate**, the text gets appended instead of pasting it from last set caret position.

Do you mean navigate the cursor to another thought, or navigate the caret within a thought using the keyboard or mouse? Feel free to provide the exact steps to ensure I'm thinking of the same situation.

> I tried to solve this by updating state.cursorOffset state upon onKeyDown event. But this brought a lot of regressions with caret. I am still trying to find the better solution but seems like I could use some help from you if you have any suggestions.

`importText` gets passed `replaceStart` and `replaceEnd`, which determines where the pasted text is inserted:

https://github.com/cybersemics/em/blob/89f9d3e60734e1ffb35454f24b0494c9f7a28996/src/reducers/importText.ts#L44-L48

These are optional arguments, and indeed in this case they are both `undefined`, causing the text to be inserted at the beginning. I'm not entirely clear on the intended semantics when `replaceStart` and `replaceEnd` are not provided. It seems like we need to do one of two things to fix this:

1. **Change the semantics when `replaceStart` and `replaceEnd` are not provided to insert at the caret rather than at the beginning** - This is risky, as there may be code that relies on this behavior, which I think you discovered in your first attempt. I'll also point out that this introduces a browser dependency into a reducer, which is not allowed. Reducers must be pure functions.
2. **Pass a correct `replaceStart` and `replaceEnd` upstream** - Presumably whichever code dispatches `importText` does have access to the caret, and thus can calculate the correct `replaceStart` and `replaceEnd` values to give to `importText`. The caret position might still be wrong, but at least the text gets imported in the right place.

karunkop:
> Unfortunately this doesn't fix it in all cases (1. create a new thought, 2. type `a`, 3. paste).

I actually didn't quite get this scenario. It seems just fine to me or am i missing something here?

raineorshine:
> > Unfortunately this doesn't fix it in all cases (1. create a new thought, 2. type `a`, 3. paste).
> 
> I actually didn't quite get this scenario. It seems just fine to me or am i missing something here?

## Steps to Reproduce

1. Hit Enter to create a new thought
2. Type 'abc'
3. Paste `xyz'

## Current Behavior

https://user-images.githubusercontent.com/750276/159717053-1a514c5a-f058-4441-a183-09e29d9718ca.mov

- xyzabc

## Expected Behavior

- abcxyz

karunkop:
> > > Unfortunately this doesn't fix it in all cases (1. create a new thought, 2. type `a`, 3. paste).
> > 
> > 
> > I actually didn't quite get this scenario. It seems just fine to me or am i missing something here?
> 
> ## Steps to Reproduce
> 1. Hit Enter to create a new thought
> 2. Type 'abc'
> 3. Paste `xyz'
> 
> ## Current Behavior
>  Screen.Recording.2022-03-23.at.10.00.21.AM.mov 
> * xyzabc
> 
> ## Expected Behavior
> * abcxyz

This is so weird. It's working as intended on my end

karunkop:
image

raineorshine:
> This is so weird. It's working as intended on my end

I will double check when I get back to my computer. What browser are you using?

> Case 3: The caret is placed at end of string. But this time, instead of keeping caret manually it is placed as a result of deleting an empty thought sibling. Here as you can see focusOffset is 1 which is actually incorrect.

In the third case, the focusNode is an Element rather than a Text Node. focusOffset has different semantics in this case. To get the caret offset, we may need to calculate it manually. 

> Also with replaceStart and replaceEnd, these are only used when a string is selected and we want to replace that part with imported text. So i don't think we can set a upstream value for them which is usually undefined to solve this problem.

Ah, you're right. I forgot that `replaceStart` and `replaceEnd` were only used when actually replacing text. 

> > I tried to solve this by updating state.cursorOffset state upon onKeyDown event. But this brought a lot of regressions with caret.

Maybe you can update `state.cursorOffset` right before `importText` gets dispatched, or pass it through rather than creating a global variable.

karunkop:
> > This is so weird. It's working as intended on my end
> 
> I will double check when I get back to my computer. What browser are you using?

Chrome!

src/reducers/importText.ts
```diff
@@ -107,13 +107,20 @@ const importText = (
     const textNormalized = strip(convertedText, { preserveFormatting: true })
 
     // insert the textNormalized into the destValue in the correct place
-    // trim after concatenating in case destValue has whitespace
-    const left = (destValue.slice(0, replaceStart ?? 0) + textNormalized).trimLeft()
     // if cursorCleared is true i.e. clearThought is enabled we don't have to use existing thought to be appended
-    const right = state.cursorCleared ? '' : destValue.slice(replaceEnd ?? 0).trimRight()
-    const newValue = left + right
-
-    const offset = getTextContentFromHTML(left).length
```
raineorshine, line 116: If `getTextContentFromHTML` is being removed here, please follow our protocol for deleting existing code. Why did that code exist? In what PR was it introduced? What replaces its functionality? What ensures that removing it doesn't introduce any regressions? You can use `git blame` as I showed you to track down the originating commit and PR.

src/reducers/importText.ts
```diff
@@ -107,13 +107,20 @@ const importText = (
...
+    const replacedDestValue = state.cursorCleared
+      ? ''
+      : destValue.slice(0, replaceStart || 0) + destValue.slice(replaceEnd || 0)
+
+    const caretPosition =
+      (replaceStart || window.getSelection()?.focusNode?.nodeType === Node.TEXT_NODE
```
raineorshine, line 116: In order to maintain compatibility with other platforms (i.e. React Native), we avoid directly accessing the DOM. We use `selection.ts` to hide selection access and allow other platforms to shim their own API.

You can use the existing `isText` method from the `selection` module:

https://github.com/cybersemics/em/blob/d225ba99ee9eed5a92b31913fc87f116f19ed2cd/src/device/selection.ts#L109-L110

src/reducers/importText.ts
```diff
@@ -25,7 +25,7 @@ import sanitize from 'sanitize-html'
 import { getSessionId } from '../util/sessionManager'
 import { ALLOWED_ATTRIBUTES, ALLOWED_TAGS, HOME_PATH } from '../constants'
 import { getAllChildrenAsThoughts } from '../selectors/getChildren'
-import getTextContentFromHTML from '../device/getTextContentFromHTML'
+import * as selection from '../device/selection'
```
raineorshine, line 28: Reducers need to be pure functions, so we shouldn't have any browser dependencies in this file. Instead, move this up a level to the `importText` action-creator and override the cursor offset with a reducer parameter. It increases the complexity of the reducer signature, but it's worth it to maintain the purity of the function.

src/device/selection.ts
raineorshine: @karunkop Does this still need to be removed with the new approach?

raineorshine (changes requested):
Works great! Just some recommendations for best practices.

src/reducers/importText.ts
karunkop: Should I include the reasoning in the PR itself?

karunkop:
https://github.com/cybersemics/em/pull/1536#discussion_r834277859
> Yes this still needs to be removed because we require to update the caret position even when focusElementNew and focusElementOld are same as while importing/pasting text there can also only exist a single node where it's being focused.

src/reducers/importText.ts
karunkop: Yes I got this. So I was thinking overriding the `state.cursorOffset` prior to import text is a bad idea. The reason being it introduces unknown regressions that I faced because of the initial solution I took which included overriding  of `state.cursorOffset`. So what I suggest is instead of using `action-creator` to include the browser selection logic we can add a new payload parameter in `importText` reducer itself. Since we are directly dispatching `importText` reducer from `Editable.tsx` component we don't have to worry about the browser dependencies as well.

src/reducers/importText.ts
karunkop: Thanks! I missed out on this part

src/reducers/importText.ts
raineorshine: Sure, that sounds safer!

src/components/Editable.tsx
```diff
@@ -593,6 +593,7 @@ const Editable = ({
                 replaceEnd: selection.offsetEnd()!,
               }
             : null),
+          caretPosition: (selection.isText() ? selection.offset() || 0 : state.cursorOffset) || 0,
```
raineorshine, line 596: Please provide a comment here, explaining why this is needed. Anything that is not obvious from the code should be explained in a comment. Here, we should note that `state.cursorOffset` may not be up-to-date.

raineorshine (changes requested):

src/reducers/importText.ts
karunkop: The code existed as a part of commit bf06746e4c26efbed6430a983f17135843976977.

It was introduced in Split thought using browser range api #1203

Previously it was just concatenating left and right variables in importText reducer. However with recent changes we require to paste the text exactly from the caret position . I am still using getTextContentFromHtml now as I realized this has to do with when the pasted text in a html string. So this is necessary. So technicallly I am still using this function.

raineorshine: Excellent, thank you!

raineorshine (approved):

PR #1544, opened 2022-03-31
Bold Text disappears

### LINKED ISSUE #1540: Bold text disappears
There is a regression that deletes text as soon as it is made bold.

## Steps to Reproduce

1. Create a thought with text `ab`
2. Select `b` and hit <kbd>Ctrl</kbd> + <kbd>b</kbd> to make it bold.
3. Move the caret to the end of the thought
4. Hit Enter

## Current Behavior

- a
- [cursor]

## Expected Behavior

- a**b**
- [cursor]

karunkop:
Fixes #1540 

## Problem

- The regex expression used for stripping empty formatting tags in `stripEmptyFormattingTags.ts` was considering `b` and `r` also as a valid expression for this case due to which they were being stripped off.

## Solution

- Modified regex expression to avoid above mentioned problem


raineorshine:
Thanks! Please add a unit test.

raineorshine:
I was thinking an integration-level test, i.e. emulating user behavior. `stripEmptyFormattingTags` is pretty fairly self-explanatory from the regexp. 

I would suggest writing a test that either dispatches an `editThought` action or an RTL test that emulates editing the thought component. There are examples of both in some of the `__test__` folders.

raineorshine (changes requested):
I was thinking an integration-level test, i.e. emulating user behavior. `stripEmptyFormattingTags` is pretty fairly self-explanatory from the regexp itself. 

I would suggest writing a test that either dispatches an `editThought` action or an RTL test that emulates editing the thought component. This will provide better coverage of the full stack. There are examples of both in some of the `__tests__` folders.

karunkop:
> I was thinking an integration-level test, i.e. emulating user behavior. `stripEmptyFormattingTags` is pretty fairly self-explanatory from the regexp itself.
> 
> I would suggest writing a test that either dispatches an `editThought` action or an RTL test that emulates editing the thought component. This will provide better coverage of the full stack. There are examples of both in some of the `__tests__` folders.

Sorry i completely missed this comment yesterday. Will do that today. Thanks!

raineorshine (changes requested):
This test is the right idea, although you're going to have to implement it as an integration test (with a functioning DOM) instead of a pure reducer test, because the bug doesn't actually occur in the reducers. Take a look at the tests in `shortcuts` or `components`.

If you try the test on `dev`, you'll notice that it actually passes (without your fix), which tells us that it doesn't test what you think it tests. I recommend always making sure your test fails without the fix. That is just as important as making the test pass with the fix.

karunkop:
> This test is the right idea, although you're going to have to implement it as an integration test (with a functioning DOM) instead of a pure reducer test, because the bug doesn't actually occur in the reducers. Take a look at the tests in `shortcuts` or `components`.
> 
> If you try the test on `dev`, you'll notice that it actually passes (without your fix), which tells us that it doesn't test what you think it tests. When writing tests, I recommend first making sure your test fails without the fix. That is just as important as making the test pass with the fix.

Thanks for the advice, I will right this test with a functioning DOM shortly now

src/components/__tests__/Thought.ts
```diff
@@ -54,6 +54,20 @@ it('create, navigate, and edit thoughts', async () => {
   expect(aSubthoughts.at(1).text()).toBe('a1')
 })
 
+it('create, edit, format and delete empty thought', async () => {
+  windowEvent('keydown', { key: 'Enter' })
+  wrapper.update()
+  const editable = wrapper.find('div.editable')
+  await editable.simulate('change', { target: { value: 'a<b>b</b>' } })
+  windowEvent('keydown', { key: 'Enter' })
+  windowEvent('keydown', { key: 'Backspace' })
```
raineorshine: The Backspace is actually unnecessary.

raineorshine (approved):
Works perfectly! Thanks!

PR #1546, opened 2022-04-06
=note value not loaded on first load 

### LINKED ISSUE #1533: =note value not loaded on first load
## Steps to Reproduce

```
- a
  - b
    - c
      - =note
        - Hello
```

Set the cursor on `a` and refresh the page. 

## Current Behavior

The note under `c` says "Enter a note", indicating that `=note` is pending. 

## Expected Behavior

Metaprogramming attributes should always be loaded with their descendants, and not left pending. 

The note placeholder should not be rendered while still pending.

karunkop:
Fixes #1533 

## Problem

- On initial load of `Note.tsx` component, the `state` is retrieved via `store,getState()` which is initially not the updated one.
- Due to this, `attribute.ts` selector is not able to provide an initial value for note

## Solution
- Instead of retrieving `state` directly from store, passing down via props will accurately render the `Note` component and the descendants `=note` attribute values are loaded.

### DIFF
--- a/src/components/Note.tsx
+++ b/src/components/Note.tsx
@@ -20,6 +20,7 @@ import { Path, State } from '../@types'
 
 interface NoteProps {
   path: Path
+  state: State
 }
 
 /** Sets the cursor on the note's thought as it is being edited. */
@@ -39,8 +40,7 @@ const setCursorOnLiveThought = ({ path }: { path: Path }) => {
 }
 
 /** Renders an editable note that modifies the content of the hidden =note attribute. */
-const Note = ({ path }: NoteProps) => {
-  const state = store.getState()
+const Note = ({ path, state }: NoteProps) => {
   const context = pathToContext(state, path)
   const dispatch = useDispatch()
   const noteRef: { current: HTMLElement | null } = useRef(null)
--- a/src/components/Note.tsx
+++ b/src/components/Note.tsx
@@ -20,7 +20,6 @@ import { Path, State } from '../@types'
 
 interface NoteProps {
   path: Path
-  state: State
 }
 
 /** Sets the cursor on the note's thought as it is being edited. */
@@ -40,7 +39,8 @@ const setCursorOnLiveThought = ({ path }: { path: Path }) => {
 }
 
 /** Renders an editable note that modifies the content of the hidden =note attribute. */
-const Note = ({ path, state }: NoteProps) => {
+const Note = ({ path }: NoteProps) => {
+  const state = store.getState()
   const context = pathToContext(state, path)
   const dispatch = useDispatch()
   const noteRef: { current: HTMLElement | null } = useRef(null)
--- a/src/components/Note.tsx
+++ b/src/components/Note.tsx
@@ -46,12 +46,6 @@ const Note = ({ path }: NoteProps) => {
   const noteRef: { current: HTMLElement | null } = useRef(null)
   const [justPasted, setJustPasted] = useState(false)
 
-  /** Returns true if this context has a non-pending note.. */
-  const hasNote = useSelector((state: State) => {
-    const noteThought = getThoughtByContext(state, [...context, '=note'])
-    return noteThought && !noteThought.pending
-  })
-
   // set the caret on the note if editing this thought and noteFocus is true
   useEffect(() => {
     const state = store.getState()
@@ -61,9 +55,15 @@ const Note = ({ path }: NoteProps) => {
     }
   }, [state.noteFocus])
 
-  if (!hasNote || isContextViewActive(state, context)) return null
+  /** Gets the value of the note. Returns null if no note exists or if the context view is active. */
+  const note: string | null = useSelector((state: State) => {
+    if (isContextViewActive(state, context)) return null
+    const noteThought = getThoughtByContext(state, [...context, '=note'])
+    if (noteThought?.pending) return null
+    return attribute(state, context, '=note')
+  })
 
-  const note = attribute(state, context, '=note')
+  if (note === null) return null
 
   /** Handles note keyboard shortcuts. */
   const onKeyDown = (e: React.KeyboardEvent) => {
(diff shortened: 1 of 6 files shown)


raineorshine:
I think there is something else going on, because `store.getState()` is always up-to-date, by definition.

Note that with Redux you never want to pass the full state to a component. This would cause a component to be re-rendered on every change to the state, even changes that do not affect it. Instead, you'll want to identify exactly which part of the state (or the result of a selector) that you want to subscribe to.

karunkop:
> I think there is something else going on, because `store.getState()` is always up-to-date, by definition.
> 
> Note that with Redux you never want to pass the full state to a component. This would cause a component to be re-rendered on every change to the state, even changes that do not affect it. Instead, you'll want to identify exactly which part of the state (or the result of a selector) that you want to subscribe to.

Yeah I am aware of this. I was finding out other more optimized way to re-render note component. What if instead of calcuating context in `Note.tsx` we provide it via props from `Thought.tsx` ?

raineorshine:
I don't know off the top of my head. Here's what you would need to do:

- show that it fixes the issue
- explain or make obvious why the fix is in the Thought component and not the Note component
- justify that it does not cause unnecessary renders

raineorshine:
Thanks! This is very helpful. 

> > * explain or make obvious why the fix is in the Thought component and not the Note component
> 
> 
> In note component, the only depedency that could cause it to re-render is when `path` (via props) changes. Hence the only solution (to my understanding) was to re-render `Thought` component one extra time when the value of context changes which will ultimately render Note component as well. Hence I think the fix is required in Thought component.

In Redux, `useSelector` (or the pre-hooks `mapStateToProps`), connects to the store independently of the component's parent, and allow fine-grained control of re-renders. While changing the props does cause a re-render, props should only be used for context-specific component properties, i.e. properties that can only be determined by its parent component. Otherwise, we should stick with the standard Redux methods.

So, I'm still not clear why we need to re-render the whole Thought component instead of just re-rendering the Note component. I can take a closer look myself when I get some time. This is going to require that I dig into the problem firsthand. 

> > * justify that it does not cause unnecessary renders
> 
> 
> Thought component renders when any one of the return values in `mapStateToProps` changes. This will apparently lead to `Note` component to render as well. With this particular change , Thought component alongside Note component will render one time more than with the previous code when `context` changes. I think this re-render is necessary because we need to evaluate the value of note attribute when note is not in pending state. Hence, to my understanding it doesn't causes unnecessary re-render.

What does the `context` change from/to? I know the `context` is used to look up the note value with `attribute`, but I'm not quite understanding why the context for a given `Note` would
change after loading. 

Small request: If you do inline responses in a GitHub comment, please make sure to add an extra line before your text so that it breaks out of the quote. I edited your response above to show you the correct formatting. Otherwise it's distracting and harder to read. Thanks!

raineorshine:
From @karunkop:

> In Redux, `useSelector` (or the pre-hooks `mapStateToProps`), connects to the store independently of the component's parent, and allow fine-grained control of re-renders. While changing the props does cause a re-render, props should only be used for context-specific component properties, i.e. properties that can only be determined by its parent component. Otherwise, we should stick with the standard Redux methods.

It's true. Initially I just wanted to re-render note component itself but none of the things that I tried which included re evaluating of context on path change etc didn't seem to be working. I then figure out what's causing the `Note` component to re-render when cursor changes (which causes thought to be re-render), that's when I realized initially too `Thought` should also be re-rendered when context changes.

> > justify that it does not cause unnecessary renders
> `Thought` component renders when any one of the return values in mapStateToProps changes. This will apparently lead to `Note` component to render as well. With this particular change , `Thought` component alongside `Note` component will render one time more than with the previous code when `context` changes. I think this re-render is necessary because we need to evaluate the value of note attribute when note is not in pending state. Hence, to my understanding it doesn't causes unnecessary re-render.
> 
> What does the context change from/to? I know the context is used to look up the note value with attribute, but I'm not quite understanding why the context for a given Note would change after loading.

Thought's context changes hierarchically with cursor position. The answer lies with the `attribute` selector and what exact context it expects for `=note` attribute value to load up correctly. On initial load, the `context` value provided doesn't seem to correct so `attribute` selector returns null. For this exact reason. one extra re-render is necessary.

raineorshine:
The cause of the problem is that `note` was calculated from a stale `State` object, retrieved on render. `note` needs to be calculated with `useSelector` instead so that it can respond to changing `State`, specifically when additional thoughts are loaded. Directly accessing `store.getState` in the `Note` component is not such a good practice (it could have easily been me that added that), and always has the risk of missing re-renders. 

I suspect that the old logic worked as expected before the major thought re-architecture when unique ids were added. I'm a little unclear as to why `pending` is never detected, so there may be some further cleanup that is possible, but I will leave it for now since it isn't causing any problems.

Since we don't want to render the placeholder while the note is loading (which can cause a brief flash), I combined `hasNote` and `note` and made sure to only render the `Note` if there is a valid note value.

From what I can tell, `context` doesn't change. Regardless, if you subscribe the Thought to `contextLive`, it will re-render on every `State` change since `contextLive` is a new array instance every time.`['a', 'b', 'c'] !== ['a', 'b', 'c']`.

raineorshine (approved):

karunkop:
> The cause of the problem is that `note` was calculated from a stale `State` object, retrieved on render. `note` needs to be calculated with `useSelector` instead so that it can respond to changing `State`, specifically when additional thoughts are loaded. Directly accessing `store.getState` in the `Note` component is not such a good practice (it could have easily been me that added that), and always has the risk of missing re-renders.
> 
> I suspect that the old logic worked as expected before the major thought re-architecture when unique ids were added. I'm a little unclear as to why `pending` is never detected, so there may be some further cleanup that is possible, but I will leave it for now since it isn't causing any problems. Calculating the `note` in a `useSelector` fixes the current issue.
> 
> Since we don't want to render the placeholder while the note is loading (which can cause a brief flash), I combined `hasNote` and `note` and made sure to only render the `Note` if there is a valid note value.
> 
> From what I can tell, `context` doesn't change. Regardless, if you subscribe the Thought to `contextLive`, it will re-render on every `State` change since `contextLive` is a new array instance every time.`['a', 'b', 'c'] !== ['a', 'b', 'c']`.

Thanks! Use of selector to evaluate the `attribute` value is a good solution. Sadly I coudn't figure out this. Thanks for the explanation!

PR #1548, opened 2022-04-16
Fixed issue where clearing out imported text used to throw error

### LINKED ISSUE #1494: Imported text when cleared throws an error
## Steps to reproduce
- Create any thought 'xyz'.
- Delete the text you just created completely but make sure you are still in edit mode
- Paste a random string
- Now try to delete the string you just pasted either by navigating the caret to the end and hitting backspace or just press delete 
   from where the cursor initially is.

## Current Behavior
- Error is shown as Cannot read properties of undefined (reading 'lastUpdated')

## Expected Behavior
- No error should be shown while clearing the imported text.

karunkop:
Fixes #1494 

## Problem

- `applyPatch` method of `fast-json-patch` lib was attempting to update/replace a certain `lexemeIndex` which does not existed after importing text. This was due to old set of patches in `inversePatches` (state)

## Solution
- Update inversePatches correctly right after `importText` action dispatch to avoid the mentioned error above


raineorshine (changes requested):
Rather than removing the faulty patch after it's been added, let's try to identify what makes it invalid and fix it at the time it is appended so that undo works.

Adding `importText` to `UNDOABLE_ACTIONS` is good, but then I think we need to make sure a valid patch gets added to `inversePatches`. Enabling `importText` and then removing the faulty patch after the fact doesn't get to the root of the problem.

karunkop:
> Rather than removing the faulty patch after it's been added, let's try to identify what makes it invalid and fix it at the time it is appended so that undo works. Adding `importText` to `UNDOABLE_ACTIONS` is good, but then I think we need to make sure a valid patch gets added to `inversePatches`. Enabling `importText` and then removing the faulty patch after the fact doesn't get to the root of the problem.

The scenario is that  we can't term it as faulty patch to begin with. It's just that there was no proper addition of inverse patches after `importText` action as it was not UNDOABLE previously. I am having a hard time to figure out the expected patches/inverse patches to be created after the `importText` action. This is because of how we are normally implementing undo/redo functionality as a whole and there seems to be some inconsistencies with inverse patches data over each action change. May be we can discuss over the proper implementation of this functionality so that I can be clear on what should be done for this functionality? May be let's start by drawing a rough sketch on the exact scenario with the patches over possible actions that could be dispatched.

raineorshine:
> The scenario is that we can't term it as faulty patch to begin with. It's just that there was no proper addition of inverse patches after `importText` action as it was not UNDOABLE previously. 

That's good that it wasn't a faulty patch. So we can just ensure the correct patch is added on import.

> there seems to be some inconsistencies with inverse patches data over each action change

What inconsistencies specifically? Maybe there is a related bug.

> May be we can discuss over the proper implementation of this functionality so that I can be clear on what should be done for this functionality? May be let's start by drawing a rough sketch on the exact scenario with the patches over possible actions that could be dispatched.

Sure. I will try to explain my thoughts about the problem and suggestions for next debugging steps.

The undo/redo system is designed to work on virtually any action automatically. As you know, it just does a `compare` of two different states, and saves the diff as an inverse patch that can be applied to reverse the change in state. However, there is some special logic to handle navigation.

`importText` uses the following four reducers. The `undoRedoReducerEnhancer` should generate an inverse patch from the state changes produced by these reducers:

- `editThought`
- `setCursor`
- `updateThoughts`
- `editingValue`

As for what is going wrong when you add `importText` to the list of undoable actions, I think we need to investigate further. Maybe you have already explored these questions, in which case please post what you found.

- What tests break when `importText` is made undoable?
- Where do they fail?
- Do the tests have any hidden dependencies on `inversePatches`?

Another line of debugging can be pursued in relation to the point of failure.

 - When is `lexemeIndex/000` removed from the state?
 - Why isn't `lexemeIndex/000` added to the inverse patch by `undoRedoReducerEnhancer`?

    It should look something like this:
    
    ```js
    {
        "op": "add",
        "path": "/thoughts/lexemeIndex/00000000000000000000000000000000",
        "value": {
            "id": "00000000000000000000000000000000",
            "value": "",
            "contexts": [
                "85kVi_7KgGkkMXXLJEjjn"
            ],
            "created": "2022-04-18T14:05:04.379Z",
            "lastUpdated": "2022-04-18T14:05:04.379Z",
            "updatedBy": "85b6fdc7-8314-4b97-a462-ab72d469b3c0"
        },
        "actions": [
            "importText"
        ]
    }
    ```

If you can answer these questions, it should get you closer to the mechanism of failure.

karunkop:
> As for what is going wrong when you add importText to the list of undoable actions, I think we need to investigate further. Maybe you have already explored these questions, in which case please post what you found.

> What tests break when importText is made undoable?

- https://github.com/cybersemics/em/blob/dev/src/shortcuts/__tests__/undo.ts#L46-L73
- https://github.com/cybersemics/em/blob/dev/src/shortcuts/__tests__/undo.ts#L173-L211
- https://github.com/cybersemics/em/blob/dev/src/shortcuts/__tests__/undo.ts#L213-L233
- https://github.com/cybersemics/em/blob/dev/src/shortcuts/__tests__/undo.ts#L347-L356

> Where do they fail?
- Even after `importText` action is made UNDOABLE, the expected behavior that should have happened after importing, editing imported changes is not as intended. This was the same case previously before this fix!

> Do the tests have any hidden dependencies on inversePatches?
- Yes , I think most of the test mentioned above have dependencies on inversePatches. I am not sure if it is hidden or not though.

raineorshine:
> > Where do they fail?
> 
> * Even after `importText` action is made UNDOABLE, the expected behavior that should have happened after importing, editing imported changes is not as intended. This was the same case previously before this fix!

Yes, we're going beyond the original issue as such. There is an error in the console when `importText` is not undoable, and there are regressions when `importText` is undoable. If we're going to fix it, we might as well make it undoable. That's the end goal  anyway. Making it undoable was only a short-term solution to problems that likely arose during the initial implementation of undo/redo.

I'd like to see what you come up with after spending some more time on this. I have explained the undo/redo mechanism and provided a suggested direction, including the expected patch operation for `lexemeIndex/000`. See what you can do, and if you get stuck let me know. If you're not able to make progress, I may give it a shot myself. I think it's beneficial for me to spend time supporting you with discussion as you get to know the codebase, but I also don't want to spend too much of my personal time if we seem to be stuck.

src/redux-enhancers/undoRedoReducerEnhancer.ts
```diff
@@ -212,6 +212,20 @@ const undoRedoReducerEnhancer: StoreEnhancer<any> =
         return newState
       }
 
+      // if an undoable action is of type importText, update inversePatches adding latest patch of actions which contains accurate patches for importText action
```
raineorshine, line 215: This comment is good, but it needs to go a little further. The key question we need to answer for future developers is: Why does `importText` require a special case? What is different from the normal case? What would happen if only the normal case was used?

src/redux-enhancers/undoRedoReducerEnhancer.ts
```diff
@@ -212,6 +212,20 @@ const undoRedoReducerEnhancer: StoreEnhancer<any> =
...
+        // we do not want to update inversePatches  when importText runs initially for tutorial event
+        if (!newState.inversePatches.length) {
+          return newState
+        }
+        lastActionType = actionType
+        const inversePatch = compareWithOmit(newState as Index, state)
+        return {
+          ...newState,
+          inversePatches: [...state.inversePatches.slice(0, -1), addActionsToPatch(inversePatch, [actionType])],
+        }
```
raineorshine, line 226: What happens if you simply remove lines 221–226? Then it falls into the normal case and gets added to `inversePatches` as usual. Seems to work for me.

(While I could turn out to be wrong here, I will note that it was my previous comment that led me to question the need for this code at all. Justifying nontrivial code is therefore useful not just for documentation purposes, but can also help catch redundancies.)

src/redux-enhancers/undoRedoReducerEnhancer.ts
```diff
@@ -212,6 +212,20 @@ const undoRedoReducerEnhancer: StoreEnhancer<any> =
         return newState
       }
 
+      // if an undoable action is of type importText, update inversePatches adding latest patch of actions which contains accurate patches for importText action
+      if (actionType === 'importText') {
+        // we do not want to update inversePatches  when importText runs initially for tutorial event
```
raineorshine, line 217: I recommend adding one more sentence to this comment explaining what will happen if we don't short circuit here. This helps to complete the question of why this code exists, and makes it clear to future developers what depends on this code.

What is a "tutorial event"? I realize that `importText` is executed at some point during the app's initialization, but I don't see the connection to the tutorial. Maybe we can be more specific here.

raineorshine (changes requested):
This works great! Well done!

I have two recommendations for improving the comments, and one question about potentially redundant code.

Thanks!

src/redux-enhancers/undoRedoReducerEnhancer.ts
karunkop: What's the best way to add long descriptive comments @raineorshine ?

raineorshine: It doesn't have to be that long. Just 1-2 additional sentences. Try to answer the questions posed above.

src/redux-enhancers/undoRedoReducerEnhancer.ts
karunkop: Yep, you are correct! This bit of code is actually non trivial because the same logic is easliy handled at the end of function. I will simply just get rid of it! Thanks for the clarification :)

src/redux-enhancers/undoRedoReducerEnhancer.ts
```diff
@@ -212,18 +212,13 @@ const undoRedoReducerEnhancer: StoreEnhancer<any> =
         return newState
       }
 
-      // if an undoable action is of type importText, update inversePatches adding latest patch of actions which contains accurate patches for importText action
+      // importText undoable action type requires a particular check before updating inverse patches as it runs a first time during app initialisation. We want to avoid this because it clutters inverse patches with a lot of unwanted actions that are specifically related with app initialisation and not particularly with NAVIGATION or other UNDOABLE actions.
+      // Cluttering inverse patches wirh unwanted actions can lead to undoing/redoing of actions such as showing up tutorial sections upon further undoing (as it is still executable beacuse of those inverse patches) which we want to avoid here.
```
raineorshine, line 216: I like how you have shifted from describing what the code does, to why the code exists. Now if we can find a way to say the same thing more succinctly, it will be good to go.

src/redux-enhancers/undoRedoReducerEnhancer.ts
```diff
@@ -212,18 +212,13 @@ const undoRedoReducerEnhancer: StoreEnhancer<any> =
...
       if (actionType === 'importText') {
-        // we do not want to update inversePatches  when importText runs initially for tutorial event
+        // we do not want to update inversePatches  when importText runs initially (during app initialisation)
```
raineorshine, line 218: Add this point, this just repeats what was said above, so we can remove it.

raineorshine (approved):

raineorshine:
Take a look at 93b88856f7 to see how I did the comments.

Thanks!

src/shortcuts/__tests__/undo.ts
```diff
@@ -354,3 +354,35 @@ it('non-undoable actions are ignored', () => {
 
   expect(store.getState().inversePatches.length).toEqual(0)
 })
+
+it('undo redo importText action', () => {
```
raineorshine, line 358: I added this test to the already existing https://github.com/cybersemics/em/blob/dev/src/shortcuts/__tests__/undo.ts. 

We follow the convention of naming the test file the same as the shortcut, reducer, or component that it's testing.

PR #1549, opened 2022-04-17
importJSON: Error importing nested list

### LINKED ISSUE #1547: importJSON: Error importing nested list
## Steps to Reproduce

```
<div>
	A
	<ul>
		<li>B</li>
	</ul>
	<ul>
		<li>C</li>
	</ul>
</div>
```

## Current Behavior

```
Unhandled Rejection (TypeError): Cannot read properties of undefined (reading 'trim')
importJSON: L123
```

## Expected Behavior

- a
  - b
  - c

karunkop:
Fixes #1547 

## Problem

- while importing nested list, `htmlToJson` would normally return a single dimensional array. However in some cases, it can return two dimensional array as well, which causes a problem in (https://github.com/cybersemics/em/blob/dev/src/util/importJSON.ts#L123) as it is attempting to trim `undefined`

## Preliminary solution
- The solution is to flatten the array obtained from `htmlToJson` so that `block.scope` is not `undefined`. However this solution doesn't fix the expected formatting of the thoughts as discussed in the issue #1547

### DIFF
--- a/src/util/htmlToJson.ts
+++ b/src/util/htmlToJson.ts
@@ -133,7 +133,7 @@ const joinChildren = (nodes: Block[]) => {
     }),
   )
 
-  return parentsWithChildren.length === 1 ? parentsWithChildren[0] : parentsWithChildren
+  return parentsWithChildren.length === 1 ? parentsWithChildren[0] : parentsWithChildren.flat()
 }
 
 /** Converts an <li> element to a Block. */
@@ -195,7 +195,10 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
 
   // retrieve first chunk, if the first element is Block and the second is Block[], join children (Block[]) with parent (Block), else return blocks as is.
   const [first, rest] = blocks
+  console.log('rest :', rest)
+  console.log('first :', first)
   const result = !Array.isArray(first) && Array.isArray(rest) ? joinChildren(blocks as Block[]) : (blocks as Block[])
+  console.log('result :', result)
 
   return result
 }
--- a/src/util/htmlToJson.ts
+++ b/src/util/htmlToJson.ts
@@ -120,9 +120,9 @@ const handleBr = (nodes: HimalayaNode[], brIndex: number): HimalayaNode[] => {
 }
 
 /** Append children to parent as children property if it's necessary. */
-const joinChildren = (nodes: Block[]) => {
+const joinChildren = (nodes: Block[] | Block[][]) => {
   // split by chunk with size of 2, first element in chunk is Block - parent, the second is Block[] - children
-  const chunks = _.chunk(nodes, 2)
+  const chunks = _.chunk(nodes as Block[], 2)
   const parentsWithChildren = chunks.map(chunk =>
     chunk.reduce((accum, node, index) => {
       if (index === 0) return node
@@ -195,7 +195,8 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
 
   // retrieve first chunk, if the first element is Block and the second is Block[], join children (Block[]) with parent (Block), else return blocks as is.
   const [first, rest] = blocks
-  const result = !Array.isArray(first) && Array.isArray(rest) ? joinChildren(blocks as Block[]) : (blocks as Block[])
+  const result =
+    !Array.isArray(first) && Array.isArray(rest) ? joinChildren(blocks as Block[] | Block[][]) : (blocks as Block[])
 
   return result
 }
--- a/src/util/htmlToJson.ts
+++ b/src/util/htmlToJson.ts
@@ -195,10 +195,7 @@ const himalayaToBlock = (nodes: HimalayaNode[]): Block | Block[] => {
 
   // retrieve first chunk, if the first element is Block and the second is Block[], join children (Block[]) with parent (Block), else return blocks as is.
   const [first, rest] = blocks
-  console.log('rest :', rest)
-  console.log('first :', first)
   const result = !Array.isArray(first) && Array.isArray(rest) ? joinChildren(blocks as Block[]) : (blocks as Block[])
-  console.log('result :', result)
 
   return result
 }
--- a/src/util/htmlToJson.ts
+++ b/src/util/htmlToJson.ts
@@ -120,11 +120,11 @@ const handleBr = (nodes: HimalayaNode[], brIndex: number): HimalayaNode[] => {
 }
 
 /** Append children to parent as children property if it's necessary. */
-const joinChildren = (nodes: Block[] | Block[][]) => {
+const joinChildren = (nodes: (Block | Block[])[]) => {
   // split by chunk with size of 2, first element in chunk is Block - parent, the second is Block[] - children
-  const chunks = _.chunk(nodes as Block[], 2)
+  const chunks = _.chunk(nodes, 2)
   const parentsWithChildren = chunks.map(chunk =>
-    chunk.reduce((accum, node, index) => {
+    chunk.flat().reduce((accum, node, index) => {
       if (index === 0) return node
       return {
         ...accum,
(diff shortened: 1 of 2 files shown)


raineorshine (changes requested):
The return type of `htmlToJson` should be `Block[]`. If it is returning `Block[][]`, then that is a bug. We need to get to the root of the problem rather than fixing it after the fact.

raineorshine (changes requested):
The type of `parentsWithChildren` should be `Block[]`. If something is making it `Block[][]`, then that is a bug. Try to get to the root of the problem that is creating the invalid data to begin with.

karunkop:
> The type of `parentsWithChildren` should be `Block[]`. If something is making it `Block[][]`, then that is a bug. Try to get to the root of the problem that is creating the invalid data to begin with.

`joinChildren` function is expecting Block[], but when we are passing the block, the block can still be as Block[][] from what I have seen. Do you reckon the issue lies initially when we evaluate `blocks`?

raineorshine:
I'm not sure, but it should be pretty easy to trace where `Block[][]` is introduced with some `console.log` statements or breakpoints.

`Block[][]` may be valid after the initial parse, but I'm not sure. At some point it needs to be flattened. However, note that flattening  is a loss of information. You might need that information to achieve the indented structure. Maybe `Block[][]` just needs to be converted into `Block[]` with the inner arrays as `children`. 

The types should be made accurate at each step regardless.

karunkop:
> `Block[][]` may be valid after the initial parse, but I'm not sure. At some point it needs to be flattened. However, note that flattening is a loss of information. You might need that information to achieve the indented structure. Maybe `Block[][]` just needs to be converted into `Block[]` with the inner arrays as `children`.

So far from what I have seen, I think the part where I have done flattening is correct as the intended structure has already been calculated before returning `parentsWithChildren`.  I have thoroughly examined the logic and looks like this is the way to go!

raineorshine:
If that's the case, you still have to fix the types.

karunkop:
> If that's the case, you still have to fix the types.

For `joinChildren` ?

raineorshine:
The type system marks `parentsWithChildren` as `Block[]`, but as you pointed out the actual type at runtime is `Block[] | Block[][]`. We need to ensure that the Typescript types always match the runtime types.

raineorshine:
Here is the problematic cast: https://github.com/cybersemics/em/blob/265e564b88fa72c0828c99d9a912c8b2bb087719/src/util/htmlToJson.ts#L198

karunkop:
> Here is the problematic cast:
> 
> https://github.com/cybersemics/em/blob/265e564b88fa72c0828c99d9a912c8b2bb087719/src/util/htmlToJson.ts#L198
> 
> I wonder if `joinChildren` should be recursive.

I don't think it should be because it's main goal is to just append children to parent as children property. Also the blocks that are passed to this function has been recurssively calculated to get intended structure

src/util/htmlToJson.ts
```diff
@@ -120,9 +120,9 @@ const handleBr = (nodes: HimalayaNode[], brIndex: number): HimalayaNode[] => {
 }
 
 /** Append children to parent as children property if it's necessary. */
-const joinChildren = (nodes: Block[]) => {
+const joinChildren = (nodes: Block[] | Block[][]) => {
   // split by chunk with size of 2, first element in chunk is Block - parent, the second is Block[] - children
-  const chunks = _.chunk(nodes, 2)
+  const chunks = _.chunk(nodes as Block[], 2)
```
raineorshine, line 125: You've just moved the invalid cast from `himilayaToBlock` to `joinChildren`. If the actual runtime type if `Block[] | [][]`, then you can't simply cast it to `Block[]`. You have to create a case if it's `Block[]`, and a case if it's `Block[][]`. (This also assumes that `Block[][][]` is not possible, which would require a recursive type.)

raineorshine (changes requested):

raineorshine:
> I don't think it should be because it's main goal is to just append children to parent as children property. Also the blocks that are passed to this function has been recurssively calculated to get intended structure

Sure, but that only works if `joinChildren` is actually getting `Block[]`. If it is getting nested lists, it either needs to handle them or we need to convert to `Block[]` higher up.

src/util/htmlToJson.ts
karunkop: > > I don't think it should be because it's main goal is to just append children to parent as children property. Also the blocks that are passed to this function has been recurssively calculated to get intended structure
> 
> Sure, but that only works if `joinChildren` is actually getting `Block[]`. If it is getting nested lists, it either needs to handle them or we need to convert to `Block[]` higher up.

That make sense. i will see what I can do here

karunkop:
> > I don't think it should be because it's main goal is to just append children to parent as children property. Also the blocks that are passed to this function has been recurssively calculated to get intended structure
> 
> Sure, but that only works if `joinChildren` is actually getting `Block[]`. If it is getting nested lists, it either needs to handle them or we need to convert to `Block[]` higher up.

That makes sense. I will see what I can do here.

src/util/htmlToJson.ts
karunkop, line 125: Why are we creating chunks of size 2 explicitly?

raineorshine: I believe because `himalaya.parse` returns an array in which the first element is the parent and the second is the children.

karunkop:
> Here is the problematic cast:
> 
> https://github.com/cybersemics/em/blob/265e564b88fa72c0828c99d9a912c8b2bb087719/src/util/htmlToJson.ts#L198
> 
> I wonder if `joinChildren` should be recursive.

Seems like the type casting is done on purpose because the user knows that `nodes:Block[]` could easily be something like both `Block[] |  Block[][]`. This is justified by the comment in line https://github.com/cybersemics/em/blob/dev/src/util/htmlToJson.ts#L124

In that case is it really necessary to reflect type notations strictly?

raineorshine:
> Seems like the type casting is done on purpose because the user knows that `nodes:Block[]` could easily be something like both `Block[] | Block[][]`. This is justified by the comment in line [https://github.com/cybersemics/em/blob/dev/src/util/htmlToJson.ts#L124](https://github.com/cybersemics/em/blob/dev/src/util/htmlToJson.ts?rgh-link-date=2022-04-18T03%3A50%3A16Z#L124)

You are correct that the type casting is done on purpose, though it seems the author didn't anticipate `Block[][]`. After all, the conditional statement checks for `[Block, Block[]]` which is *not* the same as `Block[][]`.

> In that case is it really necessary to reflect type notations strictly?

A cast is correct when there is a narrower type at runtime than how it's annotated *when the type system cannot automatically detect the change*. The type system can perform some basic type inference, but there will always be situations where we know, as developers, that logically the type can be narrowed. This is where a cast is needed.

However, we know that this isn't the circumstance here. We have an unexpected type (`Block[][]`) that was *not* being handled by the existing logic. You don't get a runtime error from a type that is too wide.

There was a runtime error and the type system was passing, saying everything was okay when it wasn't. That is precisely the situation where a strict type notation is most useful :).

The easiest thing would be to just run through the code that was affected by your change, and make sure the runtime type always matches the type annotation. If your fix is correct, then the types will just fall into place. If you run into an inconsistency, it could reveal the need for other changes to the code to properly handle all cases.

karunkop:
> The easiest thing would be to go through the code that was directly affected by your change, and make sure the runtime type always matches the type annotation. If your fix is correct, then the types will just fall into place. If you run into an inconsistency, it could reveal the need for other changes to the code to properly handle all cases.

For one thing, the fix that I have done i.e (https://github.com/cybersemics/em/pull/1549/files#diff-fa6735dde57ad78cd69f61d63a80abffddd32e144e28221d906ae8fc6bd3d1c8R136) is to just match the returnType of `joinChildren` function. I don't understand the point where you said your fix is correct.

src/util/htmlToJson.ts
raineorshine: @karunkop When we paste the offending HTML, the type of `nodes` at runtime is `[Block, Block[], Block[]]`. You can confirm this yourself with a `console.log`. Since this does not match `Block[] | Block[][]`, it tells us that the type annotations are incorrect.

I would recommend tracing the value of `nodes` upstream until you find the first point where the runtime type does not match the type annotation. Then fix the type annotation there, and update the downstream types accordingly.

raineorshine (changes requested):

raineorshine (approved):
Perfect!

PR #1558, opened 2022-04-24
Add feature that shows user last action on undo shortcut description

### LINKED ISSUE #1557: Show undo action in tooltip
## Steps to Reproduce

1. Create a thought
2. Hover the mouse over the Undo button in the toolbar to activate the tooltip.

## Current Behavior

The tooltip shows a description of "Undo."

image

## Expected Behavior

The tooltip should name the action that will be undone: "Undo Import Text"

## Note

The action name can be accessed with `getLastActionType` as done in the [undo shortcut](https://github.com/cybersemics/em/blob/dev/src/shortcuts/undo.ts).

arifqys:
Fixes #1557

Screen Shot 2022-04-24 at 09 58 57

Changes:
- Change undo's message into the function to enable dynamic description based on the last user action
- Add function type declaration for shortcut's message
- Change usage how to display shortcut message

### DIFF
--- a/src/components/ModalHelp.tsx
+++ b/src/components/ModalHelp.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 import { connect } from 'react-redux'
 import { isTouch } from '../browser'
+import { store } from '../store'
 import { formatKeyboardShortcut, globalShortcuts } from '../shortcuts'
 import * as db from '../data-providers/dexie'
 import { makeCompareByProp, sort } from '../util'
@@ -30,7 +31,7 @@ const ShortcutRows = (shortcut: Shortcut, i: number) => (
   <tr key={i}>
     <th>
       <b>{shortcut.label}</b>
-      <p>{shortcut.description}</p>
+      <p>{typeof shortcut.description === 'function' ? shortcut.description(store.getState) : shortcut.description}</p>
     </th>
     <td>
       {isTouch && shortcut.gesture ? (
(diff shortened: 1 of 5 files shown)


raineorshine (approved):
Thanks for your contributions! Works great!

raineorshine:
Would you like to add the same functionality to the `redo` shortcut? `getLatestActionType` only works with `undo`, but you can use `state.patches[state.patches.length - 1][0].actions[0]` to get the action name for `redo`.

arifqys:
Sure I'll try to implement that also later

arifqys:
I updated the code. I used `getLatestActionType(getState().patches)` to get the action name for redo

Screen Shot 2022-04-25 at 13 02 39

Lmk if you have any concerns

src/components/ModalHelp.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 import { connect } from 'react-redux'
 import { isTouch } from '../browser'
+import { store } from '../store'
```
raineorshine, line 4: I'll accept this, but note that importing the `store` directly into a component is considered a bad practice. It is recommended to access the store through `mapStateToProps` or the `useSelector` hook.

src/shortcuts/redo.ts
```diff
@@ -8,7 +8,15 @@ import { getLatestActionType } from '../util/getLastActionType'
 const redoShortcut: Shortcut = {
   id: 'redo',
   label: 'Redo',
-  description: 'Redo',
+  description: getState => {
+    const lastActionType = getLatestActionType(getState().patches)
```
raineorshine, line 12: I was mistaken about `getLatestActionType` only working for `undo`! I thought that `patches` were in reverse order, but I tested it and this works great.

raineorshine (approved):

src/components/ModalHelp.tsx
arifqys: Sure. Thanks for reviewing my code!

PR #1560, opened 2022-04-26
Error deleting second empty thought

### LINKED ISSUE #1543: Error deleting second empty thought
## Steps to reproduce

- Hit enter with an empty thought
- Hit backspace

## Current Behavior

image

## Expected Behavior

- No error should be thrown

karunkop:
Fixes #1543 

## Problem

- Previous thought path was not calculated precisely in `deleteThought` reducer, which was causing to pass incorrect cursor/path to the state.

## Solution

-  Updated `deleteThought` reducer to accurately calculate prev thought path and update cursor in state variable accordingly.


src/reducers/deleteThought.ts
```diff
@@ -229,7 +233,7 @@ const deleteThought = (state: State, { context, thoughtId, orphaned }: Payload)
     state => ({
       ...state,
       contextViews: contextViewsNew,
-      cursor: cursorNew,
+      cursor: prevThought ? ([prevThought.id] as Path) : cursorNew,
```
raineorshine, line 236: This isn't a valid way to construct a Path, as a Path must start at the ROOT. You probably want something like `[...parentOf(state.cursor), prevThought.id]`.

raineorshine: Also, instead of adding the condition here and ignoring `cursorNew`, check the condition when creating `cursorNew` and set the correct value there. 

This line should just be a simple assignment.

raineorshine (changes requested):

src/reducers/deleteThought.ts
karunkop: To answer your question, in `deleteThoughts.ts` reducer `path` is calculated via `rankThoughtsFirstMatch` selector function which is returning the `deleted thought path` for non-empty thought while it's `prev thought path` for an empty thought. So now I am trying to figure out which is the expected behavior. What value should `path` variable in `deleteThoughts` must hold on these two scenarios?

karunkop: My intention with the new logic was to actually replace the existing logic and the way cursor is being calculated. I thought the better way would to be find out the prev thought/sibling and then use it's id as path for next cursor update.

raineorshine: > To answer your question, in deleteThoughts.ts reducer path is calculated via rankThoughtsFirstMatch selector function which is returning the deleted thought path for non-empty thought while it's prev thought path for an empty thought.

Aha! That's definitely the problem. rankThoughtsFirstMatch is a heuristic that can't distinguish siblings. 

Luckily, it's obsolete with the new id-based architecture. Try to find a way to get the Path without having to convert from Context using `rankThoughtsFirstMatch`. Usually you can either build up the Path from the cursor, or you can use one of the selector functions. I can look more closely at this if you need help identifying the correct selector to use. 

> My intention with the new logic was to actually replace the existing logic and the way cursor is being calculated. I thought the better way would to be find out the prev thought/sibling and then use it's id as path for next cursor update.

Fair enough. My comments still hold. You need to solve the problem upstream, and the Path you created is not valid. All Paths must start at the ROOT thought.

karunkop: I have just pushed some changes, however I don't think that it still fully justifies the approach that you mentioned just now. I will have a look once again and build a optimal solution.

karunkop: > Luckily, `rankThoughtsFirstMatch` is largely obsolete with the new id-based architecture. Try to find a way to get the `Path` without having to convert from a `Context` using `rankThoughtsFirstMatch`. Usually you can either build up the `Path` from the cursor, or you can use one of the selector functions. I can look more closely at this if you need help identifying the correct selector to use.

I can see there are number of ways to construct a `path`. I am not sure on what selector function to use for this specific purpose.

raineorshine:
Please add a unit test at your convenience. Thanks!

src/reducers/__tests__/deleteThought.ts
```diff
@@ -80,3 +80,20 @@ it('update cursor after thought deletion', () => {
     },
   ])
 })
+it('deletion of empty thought', () => {
```
raineorshine, line 83: The test case is not about deleting any empty thought, but about deleting a second empty thought. Let's update the test description to be more specific. 

`it('delete the correct empty thought when there are multiple', ...)`

src/reducers/__tests__/deleteThought.ts
```diff
@@ -80,3 +80,20 @@ it('update cursor after thought deletion', () => {
...
+  const steps = [newThought(''), newThought('')]
+
+  const state = initialState()
+  const stateNew = reducerFlow(steps)(state)
+  const rootChildren = stateNew.thoughts.thoughtIndex[HOME_TOKEN].children
```
raineorshine, line 88: Never access `state.thoughts` directly since it is a low-level implementation detail. Instead, use a selector like `getAllChildren`.

src/reducers/__tests__/deleteThought.ts
```diff
@@ -80,3 +80,20 @@ it('update cursor after thought deletion', () => {
...
+
+  // root children before deletion
+  expect(rootChildren.length).toBe(2)
+
+  const stateAfterDeletion = reducerFlow([deleteEmptyThought])(stateNew)
+
+  // root children after deletion
+  const rootChildrenAfterDeletion = stateAfterDeletion.thoughts.thoughtIndex[HOME_TOKEN].children
+
+  expect(rootChildrenAfterDeletion.length).toBe(1)
```
raineorshine, line 98: This only tests the number of children. We want to test that the specific empty thought was deleted. 

This makes a difference if, for example, the empty thoughts had different children.

raineorshine (changes requested):

raineorshine (approved):

PR #1563, opened 2022-05-01
Split thought duplicates lexeme

### LINKED ISSUE #1559: Split thought duplicates lexeme
## Steps to Reproduce

1. Create the thought `*test`
2. Place the caret after `*` and hit Enter to split the thought into `*` and `test`

## Current Behavior

`test` has two lexeme contexts (indicated by the superscript '2')

## Expected Behavior

`test` should only have one lexeme context (no superscript)

Lexeme contexts are stored in `state.thoughts.lexemeIndex[...].contexts`

karunkop:
Fixes #1559 

## Problem

- Everytime a thought is created/edited it gets pushed to IndexedDB. This is handled by `pushQueue` middleware.

- When a normal thought (pure word characters value) is splitted, `pushQueue` middleware runs and it first dispatched `pullPendingLexemes` where it tries to retrieve pending lexemes from IndexedDB if exists. The splitted thought values in this case always produce a unique lexeme value hash hence no pending lexemes are pulled.

- However when a thought (value with non-word + word characters) is splitted in such a way that the caret position is after a non-word character it creates duplication in lexemeValue as `hashThought` provides same hash value for *test, test, ^*test## etc. So in this case when `pullPendingLexemes` is called it returns a pending lexeme from IndexedDB that ultimately gets merged by `mergeConflictingLexemeIndexUpdates`. This is the actual reason that when splitting a certain lexeme can have multiple contexts.

## Solution

When splitting a thought, always remove the current lexeme value (thought to be splitted) from IndexedDB as splitting should always result in creation of new unique lexeme values of the thought.

### DIFF
--- a/src/reducers/splitThought.ts
+++ b/src/reducers/splitThought.ts
@@ -2,7 +2,7 @@ import _ from 'lodash'
 import { HOME_TOKEN } from '../constants'
 import { appendToPath, parentOf, pathToContext, reducerFlow, strip, head } from '../util'
 import { getChildrenRanked, simplifyPath, getThoughtById } from '../selectors'
-import { editableRender, editThought, moveThought, newThought } from '../reducers'
+import { editableRender, editThought, moveThought, newThought, deleteData } from '../reducers'
 import { Path, SplitResult, State } from '../@types'
 
 /** Splits a thought into two thoughts.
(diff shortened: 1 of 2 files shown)


raineorshine (changes requested):
Works great!

raineorshine (approved):
Looks great! Thanks!

PR #1569, opened 2022-05-05
Caret moves to offset 1 after outdent 

### LINKED ISSUE #1568: Caret moves to offset 1 after outdent
## Steps to Reproduce

```
- a
- chicago
```

1. Set the caret at the end of `chicago`
2. Indent
3. Outdent

## Current Behavior

Caret is at offset 1: i.e. `c|hicago`

## Expected Behavior

Caret offset should still be at the end, i.e. i.e. `chicago|`

Caret position should not change on indent or outdent.

karunkop:
Fixes #1568 

## Problem
- When we are calculating offset, we are not considering selection node type which is actually necessary to track accurate caret position.

## Solution

- When calculating cursor offset, check whether the active caret is on which node. Then accordingly apply logic to either take the value from `selection.offset()` or `state.cursorOffset`

### DIFF
--- a/src/e2e/browserEnvironment/helpers/getSelection.ts
+++ b/src/e2e/browserEnvironment/helpers/getSelection.ts
@@ -7,6 +7,7 @@ import { BrowserEnvironment } from '../types'
  * await getSelection().focusOffset
  * await getSelection().focusNode
  * await getSelection().focusNode?.textContent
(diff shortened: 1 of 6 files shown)


raineorshine (approved):
Looks good!

FYI I added a unit test myself as it ended up needing a small addition to the puppeteer test helpers. Feel free to look at the final implementation.

PR #1571, opened 2022-05-06
Merge Databases

raineorshine:
Merges two or more **em** databases together by converting them into a modern schema and importing one into the other, which will merge duplicates.

See `scripts/merge-dbs/README.md` for usage.

## Approach

Loop through each `Parent` and `Lexeme` directly:

- Low memory consumption
- Linear and resumable
  - Script should save as it goes, so that it can be run continuously and interrupted at any point without losing progress.
- Single pass architecture
- No need to export and then import
  - Import is meant to handle unpredictable user data, whereas old databases are highly predictable 
- preserves lastUpdated
- Loop through both `contextIndex` and `lexemeIndex`
  - Looping through `contextIndex` gives accurate `value` and `lastUpdated`, but does not work for legacy `Parents` without a `context` property
  - Looping through `thoughtIndex` [renamed to `lexemeIndex` in latest schema] conflates `value` and `lastUpdated` of hash-matched thoughts, but sometimes exists when a `Parent` lacks `context`.
  - Even modern schemas appear to have both, e.g. `export 2021-12-05.json`
- Requires backfilling ancestors
  - But merges duplicates "for free"

## Running on Node

The em codebase is expected to be run in the browser, so there are some hacks needed to get selectors and util functions to run in node. Also, all react native functionality must be deleted. 

These changes cannot be merged into `dev` without more advanced shimming. Therefore, it's probably be easier to maintain a separate branch. They are simple enough changes that it shouldn't be hard to rebase.

- local storage shim
- remove native components
- remove GestureResponderEvent
- remove react-native-get-random-values

## To Do

- [x] Identify the dev-friendly changes and apply them individually to dev
- [x] Create a new merge-dbs branch
- [x] Improve README documentation
  - usage
  - supported schemas
- [x] Get it to compile
  - Update for unique id schema
  - Erase old import/export/conversion logic
- [x] Loop through each `Parent`
- [x] Loop through each `Lexeme`
- [x] Handle older schemas
- [x] Recreate ancestors
- [x] Save to disk after each file
- [x] Merge TXT
- [x] ~Merge HTML~ (small enough to paste manually)
- [ ] If process grows too long, save to disk every minute and save to `state`: `MIGRATE_LAST_PARENT` or `MIGRATE_LAST_LEXEME` representing the index of the last successfully processed `Parent` or `Lexeme`.

### DIFF
--- a/scripts/merge-dbs/.gitignore
+++ b/scripts/merge-dbs/.gitignore
@@ -1,3 +1,2 @@
 build
 sample
-output
--- a/scripts/merge-dbs/.gitignore
+++ b/scripts/merge-dbs/.gitignore
@@ -1,0 +1,3 @@
+build
+sample
+output
--- a/scripts/merge-dbs/README.md
+++ b/scripts/merge-dbs/README.md
@@ -1,4 +1,6 @@
-Imports one em database into another.
+Merges **em** databases.
+
+Handles old schema and reconstructs all thoughts.
 
 ## Build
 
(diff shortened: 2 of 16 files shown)


raineorshine:
Old databases have been successfully merged. This PR is no longer useful since the database was converted to YJS.

PR #1574, opened 2022-05-14
Import sibling unordered lists

### LINKED ISSUE #1554: Import sibling unordered lists
## Steps to Reproduce

```
<div>
	a
	<ul>
		<li>b</li>
	</ul>
	<ul>
		<li>c</li>
	</ul>
</div>
```

## Current Behavior

- a
  - b
- c

## Expected Behavior

- a
  - b
  - c

## Notes

I stumbled upon an alternative, simpler, implementation to [joinChildren](https://github.com/cybersemics/em/blob/743259e4e24c7b1c6289f168002cf60072428c87/src/util/htmlToJson.ts#L122-L137) which fixes the issue:

```ts
const joinChildren = (nodes: [Block, Block[]]): Block => ({
  ...nodes[0],
  children: [...nodes[0].children, ...nodes.slice(1).flat()],
})
```

However, this causes a regression of #1008, failing to indent the second child. The test that fails is:

https://github.com/cybersemics/em/blob/6883cd6f7f510f74350d87f775762b13d8680679/src/reducers/__tests__/importText.ts#L551-L568

Take into account both test cases and try to devise a general solution.

karunkop:
Fixes #1554 

## Problem

 In `joinChildren` when we are creating chunks we are always creating a chunk of 2 assuming first will be the parent and second one will be the children. The introduction of unwanted  entity (such as empty array) could mess this up and result in unwanted indentation of the thoughts

## Solution (Potential)

Instead of hard coding the value by which the chunks should be created, take in filtered nodes length in account to create chunks. Doing this , it will ensure that sibling unorderd lists are always indented properly. I marked this solution as `potential` just because I am unware about other import scenarios with different indentation and also because there may other optimal way to approach this issue.

### DIFF
--- a/src/reducers/__tests__/importText.ts
+++ b/src/reducers/__tests__/importText.ts
@@ -1136,7 +1136,7 @@ it('import single line with style attributes and a single br tag', () => {
 </ul>`)
 })
 
-it('import nested list', () => {
+it('import plaintext + list as nested list', () => {
   const html = `<div>
   A
   <ul>
--- a/src/reducers/__tests__/importText.ts
+++ b/src/reducers/__tests__/importText.ts
@@ -1151,7 +1151,7 @@ it('import nested list', () => {
   expect(exported).toBe(`- ${HOME_TOKEN}
   - A
     - B
-  - C`)
+    - C`)
 })
 
 it('import raw state', () => {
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Looks good to me!

PR #1575, opened 2022-05-15
Arrow up not working correctly for formatted thoughts

### LINKED ISSUE #1490: Arrow up not working correctly for formatted thoughts
## Steps to reproduce
```
- Cat
- Dog
- **Hello World**
```
Place the cursor anywhere in-between **Hello World** and press arrow-up multiple times. 

## Current Behavior
The cursor does not move to Dog

## Expected Behavior
The cursor moves to Dog

karunkop:
Fixes #1490 

## Problem

-  In `selection.ts` -> `getElementPaddings` function we were splitting computedStyle property value of padding as `px `, which I don't believe is intentional. For formatted text, `getElementPaddings`, the property value of padding was computed as `0px`. The splitting via `px ` was unable to return a desired array.  Due to this when mapping to a Number it returned `NaN` hence the function `isOnFirstLine` returned false as it was unable to perform arithmetic operation due to `NaN` value. Hence `cursorUp` shortcut was not executed on formatted thoughts.

## Solution

- Just removed extra space from `px ` -> `px` while splitting

### DIFF
--- a/src/device/selection.ts
+++ b/src/device/selection.ts
@@ -15,7 +15,7 @@ interface NodeOffset {
 
 /** Gets the padding of an element as an array of numbers. */
 const getElementPaddings = (element: HTMLElement): number[] =>
-  window.getComputedStyle(element, null).getPropertyValue('padding').split('px ').map(Number)
+  window.getComputedStyle(element, null).getPropertyValue('padding').split('px').map(Number)
 
 /** Clears the selection. */
 export const clear = (): void => {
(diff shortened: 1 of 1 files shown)


src/device/selection.ts
```diff
@@ -55,7 +55,7 @@ export const isThought = (): boolean => {
   return isEditable(focusNode) || isEditable(focusNode.parentNode)
 }
 
-/** Returns true if the selection is not on the first line of a multi-line text node. Returns true if there is no selection or if the text node is only a single line. */
+/** Returns true if the selection is  on the first line of a multi-line text node. Returns true if there is no selection or if the text node is only a single line. */
```
raineorshine, line 58: I think this is a mistake. The function returns true if the selection IS on the first line.

raineorshine (approved):
Works great, thanks!

PR #1581, opened 2022-05-22
Insert space when merging recently split thoughts 

### LINKED ISSUE #1576: Insert space when merging recently split thoughts
Merging thoughts is the natural inverse to splitting thoughts. If the user splits a thought preceded by a space, the space should be restored on merge. 

## Steps to Reproduce

https://user-images.githubusercontent.com/750276/168709050-bc5fb128-e2e7-4d79-ac6a-d9803b6dfc19.MOV

1. Create thought `hello world`
2. Set the caret before `world`
3. Hit enter to split the thought 
4. Hit backspace to merge the thoughts 

## Current Behavior

- helloworld

## Expected Behavior

- hello world

## Notes

I recommend adding an optional `splitSource` property to `Thought`. It should store the id of the thought from which it was split. Then on merge, we can check the `splitSource` property to see if a space needs to be re-inserted. It does not need to be persisted to localStorage or Firebase. 

This should have no effect on merged thoughts that did not have an intervening space.

Please include a unit test.

karunkop:
Fix #1576 

## Solution

- Added `splitSource` property in `Thought` type to store in thought id of the one that is going to be splitted.
- `splitSource` property is only passed if when splitting space is needed to be respected.

### DIFF
--- a/src/@types/Thought.ts
+++ b/src/@types/Thought.ts
@@ -15,4 +15,6 @@ export interface Thought {
   // takes precedence over value for sorting purposes
   // used to preserve the sort order of thoughts that are edited to empty instead of moving them back to their insertion point
   sortValue?: string
+  // used to track if a space is required when merging two siblings/thoughts
+  splitSource?: ThoughtId
 }
--- a/src/reducers/__tests__/deleteEmptyThought.ts
+++ b/src/reducers/__tests__/deleteEmptyThought.ts
@@ -15,6 +15,7 @@ import newThought from '../newThought'
 import setCursor from '../setCursor'
 import importTextReducer from '../importText'
 import archiveThought from '../archiveThought'
+import splitThought from '../splitThought'
 
 it('delete empty thought', () => {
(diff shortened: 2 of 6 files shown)


raineorshine (changes requested):
Looks great!

One thing: it should also insert a space if `splitResult.right` has whitespace at the beginning, i.e. `hello| world`.

Make sure to think beyond just the literal issue description. What variations exist? What would the user expect for those variations? If it is not obvious, you can always ask. In this case, it is intuitive to insert a space regardless of whether the thought was split before or after the space in the original thought.

raineorshine (changes requested):
Works great! Will you add a unit test for the other test case (whitespace on right)? Thanks!

karunkop:
> Works great! Will you add a unit test for the other test case (whitespace on right)? Thanks!

I have added that. Didn't you check it out in file tests/deleteEmptyThought?

raineorshine:
Oops! Missed it! Thanks.

raineorshine (approved):

PR #1586, opened 2022-05-28
=pin/false incorrectly hides grandchildren with cursor

### LINKED ISSUE #1527: =pin/false incorrectly hides grandchildren with cursor
Expansion logic is in https://github.com/cybersemics/em/blob/dev/src/selectors/expandThoughts.ts
Tests are in https://github.com/cybersemics/em/blob/dev/src/selectors/__tests__/expandThoughts.ts

## Steps to Reproduce

```
- a
  - =pin
    - false
  - b
    - c
```

Set the cursor on `b`.

## Current Behavior

`c` is hidden.

## Expected Behavior

`c` should be visible.

Children of the cursor should always be visible. This should take precedence over `=pin/false`.

karunkop:
Fixes #1527 

## Problem

- All thoughts were being filtered from the cursor path, due to incorrect logic being used when `=pin/false` was used.

## Solution

-  Correct parameter i.e `child.parentId` is passed to `isPinClosed` function  instead of  `child.id`


src/selectors/__tests__/expandThoughts.ts
```diff
@@ -282,11 +283,27 @@ describe('=pin', () => {
     const stateNew = reducerFlow(steps)(initialState())
 
     expect(isContextExpanded(stateNew, ['a'])).toBeTruthy()
-    expect(isContextExpanded(stateNew, ['a', 'b'])).toBeFalsy()
```
raineorshine, line 285: This test is actually correct as it was. `=pin/false` takes precedence over `=pinChildren/true`.

`=pin/false` is only ignored when the cursor is on the thought, e.g. on `a/b`.

src/selectors/__tests__/expandThoughts.ts
```diff
@@ -282,11 +283,27 @@ describe('=pin', () => {
...
+    expect(isContextExpanded(stateNew, ['a', 'b'])).toBeTruthy()
     expect(isContextExpanded(stateNew, ['a', 'd'])).toBeTruthy()
   })
 })
 
+it('children of cursor should always be visible, it should take precedence over =pin/false', () => {
```
raineorshine, line 291: This test looks good

raineorshine (changes requested):

raineorshine:
There was a strange behavior that caused `b` to incorrectly expand on first load and when moving from a `null` cursor. It was not covered by the tests.

https://user-images.githubusercontent.com/750276/170831255-11a2f250-9cce-4481-9394-1106c9294e53.mov

I updated the logic to use the existing `isAncestor` function. 

Note that in `expandThoughts`, we don't want to access `state.cursor` directly. Various callers invoke `expandThoughts` with different `expansionBasePath` values. They commonly pass `state.cursor`, but this isn't always the case, and even when they do sometimes the state is in the process of being updated.

PR #1588, opened 2022-05-29
deleteThought: Thought not found for id

### LINKED ISSUE #1521: deleteThought: Thought not found for id
## Steps to Reproduce

```
- a
  - b
    - c
      - d
        - e
          - f
            - =archive
              - x
          - g
            - h
              - i
- z
```

1. Wait for the page to load
2. Paste the above thoughts
3. Set the cursor on `a`
4. Refresh the page
5. Permanently delete `a` (<kbd>Ctrl + Shift + Alt + Backspace</kbd>; not archive)

## Current Behavior

```
deleteThought: Thought not found for id
L39 deleteThought.ts
```

## Expected Behavior

`a` and all its descendants should be permanently deleted with no errors.

## Note

The issue does not occur without a page refresh, and does not occur if the cursor is on `z` during the refresh.

karunkop:
fixes #1521 

## Problem

- While creating `pendingDeletes`, there was a duplication for thoughts with metaprogramming attributes. Due to this, when descendant thoughts are recursively deleted, thought id is not found for second time for the duplicated `pendingDelete`.

## Solution

- Ensuring to uniquify  accummulated pending deletes before pushing to `pushQueue` state.

### DIFF
--- a/src/reducers/deleteThought.ts
+++ b/src/reducers/deleteThought.ts
@@ -155,11 +155,11 @@ const deleteThought = (state: State, { context, thoughtId, orphaned }: Payload) 
 
         return {
           ...accum,
-          pendingDeletes: [
+          pendingDeletes: _.uniq([
             ...(accum.pendingDeletes || []),
             ...(accumRecursive.pendingDeletes || []),
             ...(recursiveResults.pendingDeletes || []),
-          ],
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Nice work

PR #1589, opened 2022-05-29
Cursor does not move when note is focused

### LINKED ISSUE #1585: Cursor does not move when note is focused
## Steps to Reproduce

```
- a
  - =note
    - test
- b
```

1. Set cursor on `b`
2. Click on `test`

## Current Behavior

The caret moves to `test` but `state.cursor` stays on `b`.

## Expected Behavior

The cursor should move to `a`.

When the caret moves to a note, the cursor should always move to the note's thought.

karunkop:
Fixes #1585 

## Solution
- removed the use of getEditingPath to calculate simpleLivePath in `Note.tsx` component

## Reason for code remove

- `getEditingPath` and `simplePathLive` are no longer needed. They were needed at a time when the edited value was debounced, so there was a discrepancy between the cursor and the path. Now they are the same and there is no delay, so we can just use simplePath in setCursorOnLiveThought.


raineorshine (changes requested):
Looks good. Can you delete `getEditingPath.ts` completely? `Subthoughts` can use `simplePath` too.

raineorshine (approved):

PR #1593, opened 2022-06-11
Type meta attribute to move cursor

### LINKED ISSUE #1578: Type meta attribute to move cursor
When `showHiddenThoughts` is true, the meta attributes are visible and can be easily edited. When `showHiddenThoughts` is false, it's not so easy. The purpose of the feature is to provide an easy way to edit an existing meta attribute without having to toggle `showHiddenThoughts` on and off. It also avoids creating duplicate meta attributes.

For example, if there is a hidden attribute `=style` and I create a new sibling `=styl` [sic], then nothing happens. There is the hidden `=style` thought and the visible `=styl` thought (visible because the cursor is on it). As soon as the user types `e` on the end of `=styl`, it should recognize that the attribute already exists, delete `=styl`, and move the cursor to the existing `=style` thought. Once the cursor moves there, it will automatically become visible.

## Steps to Reproduce

``` 
- a
  - =style
    - color
      - lightblue
``` 

1. Set the cursor on `a` with hidden thoughts disabled. 
2. Create a new subthought 
3. Type `=style`

## Current Behavior

A new thought is created.

## Expected Behavior

If hidden thoughts are disabled, when the user types a meta attribute that already exists, delete the cursor thought and move the cursor to the end of the existing thought. 

In the above example, `color` and `lightblue` should appear under `=style`.

karunkop:
Fix #1578

## Problem
- Duplicate meta programming thought is created if the similar attribute exists within a context

## Solution

- Check for such condition within context where duplication is found for meta programming attributes
- Delete current cursor thought
- Set cursor to the already existing attribute within a context

### DIFF
--- a/src/reducers/editThought.ts
+++ b/src/reducers/editThought.ts
@@ -1,6 +1,6 @@
 import _ from 'lodash'
 // import { treeChange } from '../util/recentlyEditedTree'
-import { getLexeme, getAllChildren, getThoughtById, parentOfThought, thoughtToPath } from '../selectors'
+import { getLexeme, getAllChildren, getThoughtById, parentOfThought, thoughtToPath, findDescendant } from '../selectors'
 import updateThoughts from './updateThoughts'
 import { Context, Lexeme, SimplePath, State, ThoughtId, Timestamp } from '../@types'
 import {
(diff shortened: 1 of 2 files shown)


src/reducers/editThought.ts
```diff
@@ -32,6 +45,36 @@ const editThought = (
 
   const editedThought = getThoughtById(state, editedThoughtId)
 
+  const editedThoughtValue = editedThought.value
```
raineorshine, line 48: No need for an extra variable here

src/reducers/editThought.ts
```diff
@@ -32,6 +45,36 @@ const editThought = (
...
+
+  const parentThoughtIdsForCollidedThoughts = thoughtCollision?.contexts.map(
+    thoughtId => parentOfThought(state, thoughtId)?.id,
+  )
```
raineorshine, line 52: You should not be using `thoughtCollision` at all. It checks if the new value exists anywhere else in the thoughtspace. This functionality is only concerned with sibling thoughts, and should not be iterating over every other thought in the thoughtspace with the same value.

You can use `Thought.childrenMap` for O(1) lookup of meta attributes, as they are keyed by value:

```ts
const styleThoughtId = thought.childrenMap['=style']
```

raineorshine (changes requested):
Please add a unit test at your convenience. Thanks!

src/reducers/editThought.ts
raineorshine: Even better, use https://github.com/cybersemics/em/blob/dev/src/selectors/findDescendant.ts. That way the implementation details are hidden. It is optimized for attribute children.

src/reducers/editThought.ts
```diff
@@ -32,6 +45,28 @@ const editThought = (
 
   const editedThought = getThoughtById(state, editedThoughtId)
 
+  const parentThoughtIdForEditedThought = parentOfThought(state, editedThoughtId)?.id
+
+  // only calculate decendant thought when current edited thought is a metaprogramming attribute
+
+  const thoughtIdForExistingMetaProgrammingThought =
```
raineorshine, line 52: You should also check that the cursor is on the edited thought. `editThought` is called within various reducer flows, so you can't assume the edited thought is the cursor.

raineorshine (changes requested):
I would suggest moving the new logic below the error checking.  Usually short circuiting logic goes first, and then the rest of the logic proceeds once the inputs have been validated.

src/reducers/editThought.ts
```diff
@@ -49,6 +63,25 @@ const editThought = (
     console.warn(`Missing Lexeme: ${oldValue}`)
   }
 
+  // only calculate decendant thought when current edited thought is a metaprogramming attribute
+  const thoughtIdForExistingMetaProgrammingThought =
+    isFunction(newValue) &&
+    equalArrays(state.cursor || [], thoughtToPath(state, editedThought.id)) &&
```
raineorshine, line 69: Rather than comparing the whole Path, just compare the head thought:

```ts
head(state.cursor) === editedThought.id
```

raineorshine (changes requested):

raineorshine (approved):

PR #1600, opened 2022-06-19
Bullet fades on edit

### LINKED ISSUE #1594: Bullet fades on edit
## Steps to Reproduce

Create thought `a`

## Current Behavior

The bullet changes to gray after a second.

## Expected Behavior

The bullet should not be gray. The bullet should only be gray when the thought is actually missing.

## Notes

Here is where the `.gray` CSS class is applied: https://github.com/cybersemics/em/blob/dev/src/components/Bullet.tsx#L175

The bullet thinks that the thought is missing for some reason.

For the test, you can use RTL to check that `.gray` is not on the bullet ellipsis, i.e. `document.querySelector('.bullet .gray')` should not exist.

karunkop:
Fixes #1594 

## Problem

- `getThoughtById` was being used in `Thought` component level to find recently edited thought.
- however `Bullet` component won't re-render when thought value is being edited hence no recent thought is fetched thus causing `getLexeme` to return undefined.
-  When lexeme is undefined, thought is marked as missing and hence `gray` className is applied to bullet ellipsis.

## Solution

- use `getThoughtById` in `Bullet` component itself to always get recently edited thought with editing value

### DIFF
--- a/src/components/Bullet.native.tsx
+++ b/src/components/Bullet.native.tsx
@@ -1,5 +1,4 @@
 import State from '../@types/State'
-import Thought from '../@types/Thought'
 import React from 'react'
 import { connect } from 'react-redux'
 import getLexeme from '../selectors/getLexeme'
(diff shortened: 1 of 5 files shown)


raineorshine (approved):
Nicely done

PR #1604, opened 2022-06-19
“Archive” and “=archive” should have different hashes

### LINKED ISSUE #1219: “Archive” and “=archive” should have different hashes
## Steps to Reproduce

```
- a
  - =archive
- b
  - Archive
```

*Note: Superscripts are currently disabled for performance reasons. To complete this task, you can re-enable them with `git revert 8e4d37b5ef`.

## Current Behavior

`Archive` has a superscript of 2.

## Expected Behavior

`Archive` should not have a superscript.

`=archive` and `Archive` should not produce the same normalized value with [normalizeThought](https://github.com/cybersemics/em/blob/dev/src/util/normalizeThought.ts). `hashThought` uses `normalizeThought` to produce the key for the `lexemeIndex` table. Change `normalizeThought` to not strip `=` from the beginning of the value.

karunkop:
Fixes #1219 

## Problem

- `trim` method of `normalizeThought` used to strip any non-word character which includes `=` which is a key part of metaprogramming attribute. We should not have same hashes given by `hashThought` for thoughts with value like `=archive` , Archive.

## Solution

- modified `trim` method to preserve `=` for metaprogramming attributes


src/components/__tests__/Superscript.ts
```diff
@@ -46,3 +46,17 @@ it.skip('Superscript should not count archived contexts', async () => {
...
+    importText({
+      text: `
+      - a
+        - =archive
+      - b
+        - Archive`,
+    }),
+  ])
+
+  expect(() => screen.getByText('2')).toThrow('Unable to find an element')
```
raineorshine, line 61: You should always make sure your test fails on the old code first.

This test will always pass because `=archive` is not visible, so it will not contribute to the `Superscript` count. You have to `toggleHiddenThoughts` to make it visible for checking the `Superscript`.

src/util/normalizeThought.ts
```diff
@@ -2,9 +2,17 @@ import _ from 'lodash'
 import * as pluralize from 'pluralize'
 import { REGEXP_TAGS } from '../constants'
 import emojiStrip from 'emoji-strip'
+import isAttribute from './isAttribute'
 
 /** Trims a string. */
-export const trim = (s: string) => s.replace(s.length > 0 && s.replace(/\W/g, '').length > 0 ? /\W/g : /s/g, '')
```
raineorshine, line 7: The function name and comment do not match what it actually does. A developer reading this will think it is a normal trim function, and not realize it is more complicated. Please update the function name and comment to something that matches the functionality.

raineorshine (changes requested):

raineorshine (approved):
Thanks!

I will merge this later as I have to manually migrate my personal database to update the `lexemeIndex` keys.

PR #1610, opened 2022-06-26
Undo/Redo: Cursor restored incorrectly after undo archive

### LINKED ISSUE #1389: Undo/Redo: Cursor restored incorrectly after undo archive
## Steps to Reproduce

1. Create thought `a`
2. Set the cursor to `null`
3. Refresh the page
4. Set the cursor to `a`
5. archiveThought
6. undo

## Current Behavior

Cursor disappears. 

## Expected Behavior

Cursor should be on `a`, since that is where the cursor was before `archiveThought` was activated.

Navigation actions immediately _after_ the last edit should be grouped and undone with the edit. 

Navigation actions immediately _before_ the last edit should not be undone.  

(The reverse is true for redo.)

karunkop:
Fixes #1389 

## Problem

- Refreshing a page generates a new set of undo patches which is different from when we see it normally, i,e without refreshing.  In `undoRedoEnhancer` -> `undoReducer` function has a variable called `undoTwice`. This is used to undo action again if the penultimate  action is of navigation type or `newThought`.  When refreshing a page, there exists a set of penultimate patch due to which `undoTwice` becomes true thus causing undo action to enable two times ultimately causing `setCursor` to be undoed.

## Solution 

- adding a condition to avoid two times undo action when page is refreshed.

## Note
- This may not be a solution you would want because it is just skipping that particular issue. Another possible thing that can be done is to fix the undo patches itself so that it can ressemble the normal condition (i.e without refreshing) but the problem is that as far as I know, patches are generally generated by `compare` (diffState function) which just compares the difference between newState and oldState. So fixing patches means to cleary understand what state is ideal depending upon the situation. Happy to discuss with you regarding this context.

### DIFF
--- a/src/redux-enhancers/undoRedoEnhancer.ts
+++ b/src/redux-enhancers/undoRedoEnhancer.ts
@@ -126,7 +126,9 @@ const undoReducer = (state: State, undoPatches: Patch[]) => {
     penultimateUndoPatch &&
     !!(
       (lastAction && NAVIGATION_ACTIONS[lastAction]) ||
-      (penultimateAction && (NAVIGATION_ACTIONS[penultimateAction] || penultimateAction === 'newThought'))
+      (penultimateAction &&
+        ((NAVIGATION_ACTIONS[penultimateAction] && NAVIGATION_ACTIONS[penultimateAction] !== 'setCursor') ||
+          penultimateAction === 'newThought'))
     )
 
   const poppedUndoPatches = undoTwice ? [penultimateUndoPatch, lastUndoPatch] : [lastUndoPatch]
--- a/src/shortcuts/__tests__/undo.ts
+++ b/src/shortcuts/__tests__/undo.ts
@@ -374,3 +374,32 @@ it('undo redo importText action', () => {
 
   expect(exportedAfterRedo).toEqual(expectedOutputAfterRedo)
 })
+
+it('cursor should restore correctly after undo archive', async () => {
+  timer.useFakeTimer()
+  initialize()
+  await timer.runAllAsync()
+
+  appStore.dispatch([newThought({ value: 'a' }), setCursor({ path: null })])
+  await timer.runAllAsync()
+
+  timer.useFakeTimer()
+  // clear and call initialize again to reload from local db (simulating page refresh)
+  appStore.dispatch(clear())
+  await timer.runAllAsync()
+
+  initialize()
+
+  await timer.runAllAsync()
+
+  appStore.dispatch([setCursorFirstMatchActionCreator(['a']), { type: 'archiveThought' }, { type: 'undoAction' }])
+  await timer.runAllAsync()
+
+  timer.useRealTimer()
+
+  const expectedCursor = [{ value: 'a', rank: 0 }]
+
+  const cursorThoughts = childIdsToThoughts(appStore.getState(), appStore.getState().cursor!)
+
+  expect(cursorThoughts).toMatchObject(expectedCursor)
+})


raineorshine:
Please add a test and I can review. Thanks!

raineorshine (changes requested):
The condition you added is not quite right. The problem is that you are using the check for `setCursor` as a stand-in for a page refresh. However, there are other situations when the second-to-last action would be `setCursor`. Here is one:

1. Create thought `a`
2. Create thought `b`
3. Click on `a`
4. Edit `a` to `aa`
5. Undo

The cursor will not be properly set on `b`, i.e. `undoTwice` should be true in this case to get the desired functionality, but due to the new condition it is false. (Note that there is a separate bug that causes a thought to not be re-rendered after an edit, so if you follow the above steps you won't see `aa` change to `a` on undo but it has been changed in state. But the cursor was not changed in state.)

You expressed an awareness that this may not have been the correct solution, but you can go one step further. When you write a condition, make sure that it is *not too specific* and *not too general*. Writing a condition based on `setCursor` should make you wonder, What other states could result in a penultimate `setCursor`?

> * Another possible thing that can be done is to fix the undo patches itself so that it can ressemble the normal condition (i.e without refreshing) but the problem is that as far as I know, patches are generally generated by `compare` (diffState function) which just compares the difference between newState and oldState. So fixing patches means to cleary understand what state is ideal depending upon the situation. Happy to discuss with you regarding this context.

Why do we need to fix the patches? Are they wrong? Or are they just different? 

It seems like you're on the right track with the condition, since we do want to suppress `undoTwice` in a particular circumstance. We just need a better condition that doesn't result in false positives.

karunkop:
> The condition you added is not quite right. The problem is that you are using the check for `setCursor` as a stand-in for a page refresh. However, there are other situations when the second-to-last action would be `setCursor`. Here is one:
> 
> 1. Create thought `a`
> 2. Create thought `b`
> 3. Click on `a`
> 4. Edit `a` to `aa`
> 5. Undo
> 
> The cursor will not be properly set on `b`, i.e. `undoTwice` should be true in this case to get the desired functionality, but due to the new condition it is false. (Note that there is a separate bug that causes a thought to not be re-rendered after an edit, so if you follow the above steps you won't see `aa` change to `a` on undo even though it has been changed in state. The important part is that the cursor was not changed in state.)

What should it have changed into? Shouldn't it be at thought a even after undo?

raineorshine:
> What should it have changed into? Shouldn't it be at thought a even after undo?

Wow, you're right! The cursor should be on `a` after the undo. Only navigation actions *after* edit actions are grouped together and undone at the same time. I was comparing to `dev`, assuming `dev` was right but it's not.

That complicates things. I think we need to address the grouping issue first without the complication of undoing a null cursor. Will you create a separate issue for the above steps?

karunkop:
> That complicates things. I think we need to address the grouping issue first without the complication of undoing a null cursor. Will you create a separate issue for the above steps?

When you say grouping issue, i am having a little confusion to understand what it exactly means, firstly. Second of all what exact issue should i have to create for? Is it the one where undo action not properly working for editThought?

raineorshine:
> When you say grouping issue, i am having a little confusion to understand what it exactly means, firstly.

- Navigation actions immediately *after* the last edit should be grouped and undone with the edit.
- Navigation actions immediately *before* the last edit should not be undone.
- Navigation actions immediately *after* the last undone edit should not be redone.
- Navigation actions immediately *before* the last undone edit should be grouped and redone with the edit.

> Second of all what exact issue should i have to create for? Is it the one where undo action not properly working for editThought?

I mean the test case mentioned above:

1. Create thought `a`
2. Create thought `b`
3. Click on `a`
4. Edit `a` to `aa`
5. Undo

The cursor should stay on `a` after undo, but currently (in `dev`) it moves back to `b`.

karunkop:
> > When you say grouping issue, i am having a little confusion to understand what it exactly means, firstly.
> 
> 
> 
> - Navigation actions immediately *after* the last edit should be grouped and undone with the edit.
> 
> - Navigation actions immediately *before* the last edit should not be undone.
> 
> - Navigation actions immediately *after* the last undone edit should not be redone.
> 
> - Navigation actions immediately *before* the last undone edit should be grouped and redone with the edit.
> 
> 
> 
> > Second of all what exact issue should i have to create for? Is it the one where undo action not properly working for editThought?
> 
> 
> 
> I mean the test case mentioned above:
> 
> 
> 
> 1. Create thought `a`
> 
> 2. Create thought `b`
> 
> 3. Click on `a`
> 
> 4. Edit `a` to `aa`
> 
> 5. Undo
> 
> 
> 
> The cursor should stay on `a` after undo, but currently (in `dev`) it moves back to `b`.
> 
> 
> 
> Once grouping works for the general case, we can make sure it works in the edge case where the cursor is changed to null (#1389).

Okay will create an issue for this!

raineorshine:
Closed due to inactivity

PR #1670, opened 2022-12-30
Sort meta attributes above emojis

### LINKED ISSUE #1614: Emoji not sorted properly: *️⃣
## Steps to Reproduce

```
- a
  - c
  - *️⃣ Active
  - a
  - ⏸ Inactive
  - =sort
    - Alphabetical
      - Asc
  - b
```

## Current Behavior

`*️⃣ Active` is sorted above the meta attribute `=sort`:

- a
  - *️⃣ Active
  - =sort
    - Alphabetical
      - Asc
  - ⏸ Inactive
  - a
  - b
  - c

## Expected Behavior

`*️⃣ Active` should be sorted below the meta attribute `=sort`.

Meta attributes should always be sorted to the top.
Thoughts that start with emojis should always be sorted below meta attributes and above other thoughts.

- a
  - =sort
    - Alphabetical
      - Asc
  - *️⃣ Active
  - ⏸ Inactive
  - a
  - b
  - c

## Notes

This is the same issue as #1037, but with a different character. Use the same approach and unit testing as #1048.

ankitkarna99:
Fixes #1614 

Changes:
- Added new meta attributes comparator and added it as precedence over emoji comparator.

### DIFF
--- a/src/util/compareThought.ts
+++ b/src/util/compareThought.ts
@@ -5,6 +5,7 @@ import { EMOJI_REGEX, EMOJI_REGEX_GLOBAL } from '../constants'
 import lower from './lower'
 
 const STARTS_WITH_EMOJI_REGEX = new RegExp(`^${EMOJI_REGEX.source}`)
+const STARTS_WITH_META_ATTRIBUTES_REGEX = /^=/
 const IGNORED_PREFIXES = ['the ']
 const CURRENT_YEAR = new Date().getFullYear()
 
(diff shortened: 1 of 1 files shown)


src/util/compareThought.ts
```diff
@@ -52,6 +53,17 @@ export const compareStringsWithEmoji = (a: string, b: string) => {
   return aStartsWithEmoji && !bStartsWithEmoji ? -1 : bStartsWithEmoji && !aStartsWithEmoji ? 1 : 0
 }
 
+/** A comparator that sorts meta-attributes above emojis. */
+export const compareStringsWithMetaAttributes = (a: string, b: string) => {
+  const aStartsWithMetaAttributes = !!a.match(STARTS_WITH_META_ATTRIBUTES_REGEX)
```
raineorshine, line 58: Use https://github.com/cybersemics/em/blob/dev/src/util/isAttribute.ts to check if a string is a meta attribute.

raineorshine (changes requested):
Thanks!

Please add a test to https://github.com/cybersemics/em/blob/dev/src/util/__tests__/compareThought.ts

src/util/compareThought.ts
```diff
@@ -52,6 +53,17 @@ export const compareStringsWithEmoji = (a: string, b: string) => {
   return aStartsWithEmoji && !bStartsWithEmoji ? -1 : bStartsWithEmoji && !aStartsWithEmoji ? 1 : 0
 }
 
+/** A comparator that sorts meta-attributes above emojis. */
```
raineorshine, line 56: This comment isn't quite right. The comparator sorts meta attributes above everything, not just emojis. `compareReasonable` is responsible for ordering them.

ankitkarna99:
## Updates

- Updated comment to properly describe the comparator
- Added test

raineorshine (approved):

PR #1678, opened 2023-01-08
Apply bold, italic, underline to selection, and not toggle attribute

### LINKED ISSUE #1671: Apply bold, italic, underline to selection
When text within a thought is selected, hitting [bold](https://github.com/cybersemics/em/blob/dev/src/shortcuts/bold.tsx) should toggle bold formatting of the selected text.

When there is no browser selection, hitting bold should toggle bold formatting for the entire cursor thought. 

Repeat for [italic](https://github.com/cybersemics/em/blob/dev/src/shortcuts/italic.tsx) and [underline](https://github.com/cybersemics/em/blob/dev/src/shortcuts/underline.tsx).

Use `document.execCommand`. It is deprecated but it is supported in all browsers and is still the best way to apply formatting to a selection in a contenteditable.

```js
document.execCommand('bold')
document.execCommand('italic')
document.execCommand('underline')
```

# 1. Bold/unbold selection

## Steps to Reproduce

1. Create thought "Hello World"
2. Select "World"
3. Hit the [bold](https://github.com/cybersemics/em/blob/dev/src/shortcuts/bold.tsx) toolbar button.

## Current Behavior

=style/fontWeight is inserted but it doesn't seem to affect the thought formatting. This is the wrong approach anyway, as we want to only apply bold to the selected text.

```
- =style
  - fontWeight
    - 700
```

## Expected Behavior

The thought value should change to `Hello <b>World</b>` so that it renders as: "Hello **World**".

If activated again, it should toggle back to unbold (should happen automatically with `execCommand('bold')`).

# 2. Bold/unbold cursor thought

## Steps to Reproduce

1. Create thought "Hello World"
2. Hit the [bold](https://github.com/cybersemics/em/blob/dev/src/shortcuts/bold.tsx) toolbar button.

## Current Behavior

Same as "Current Behavior" above.

## Expected Behavior

Should select all text in the cursor thought and activate `execCommand('bold')`

ankitkarna99:
Fixes #1671 

## Changes
- Created new utility function `emphasizeSelection` to call `document.execCommand` for three operations i.e. bold, italic and underline.
- Removed calls to reducers from shortcuts, but the reducer files for bold, italic and underline are not deleted yet. 
- Made it work in a way that cursor position is not lost after applying or removing emphasis.


src/util/emphasizeSelection.ts
```diff
@@ -0,0 +1,65 @@
...
+    return focusNode
+  } else {
+    return findTargetFocusNode(root.firstChild, savedFocusNodeContents, savedFocusOffset)
+  }
+}
+
+/** Emphasize a selection or entire thought.
+ * To be used inside shortcuts of bold/italics/underline.
+ */
+const emphasizeSelection = (state: State, command: 'bold' | 'italic' | 'underline'): void => {
```
raineorshine, line 26: Let's call this `formatSelection`.

src/util/emphasizeSelection.ts
```diff
@@ -0,0 +1,65 @@
...
+  const thought = getThoughtById(state, _.last(state.cursor)!)
```
raineorshine, line 27: You can use `pathToThought` here.

FYI There are a set of standard named selectors that help convert between paths, thoughts, and contexts:

- `thoughtToContext`
- `thoughtToPath`
- `pathToContext`
- `pathToThought`
- `contextToPath`
- `contextToThought`

Avoid the `context*` functions when possible. We generally don't need to use contexts now that each thought has a unique id.

ankitkarna99:
## Changes

- Refactor `save` and `restore` functions to support nested tags inside of `contenteditable`
- Used `selectAllChildren` instead of `setBaseAndExtent`
- Renamed `emphasizeSelection` to `formatSelection`

src/device/selection.ts
```diff
@@ -140,19 +147,55 @@ export const offsetStart = (): number | null => {
...
+  savedFocusNode: Node,
+  savedFocusOffset: number,
+): ChildNode | undefined => {
+  if (!root) return undefined
+
+  const childNodes = Array.from(root.childNodes)
+  const hasTextNode = childNodes.some(node => node.nodeType === Node.TEXT_NODE)
+  if (hasTextNode) {
+    const focusNode = childNodes.find(node => node.isSameNode(savedFocusNode))
+    return focusNode
```
raineorshine, line 162: `childNodes.find(node => node.isSameNode(savedFocusNode))` will always return `savedFocusNode | undefined`, since you're just finding the child that is `savedFocusNode`. From what I can tell, `findNodeRecursivelyRoot` traverses from a root node to a given node, just to return the given node itself. Am I missing something?

ankitkarna99: Yes, you're right. I added it since it was not working at that moment, and I was using `sel.addRange` instead of `collapse` then. Later, I changed to `collapse` but forgot to check if `savedFocusNode` directly was working or not. Thanks for this.

ankitkarna99: I checked and right now it works directly when using savedFocusNode. Earlier the added HTML and stripped HTML changed the nodes themselves, and I thought it was necessary, but it seems not.

raineorshine: Thanks for the explanation, and for fixing that up. 

Note: As soon as you typed `isSameNode` you should have noticed something was wrong. It should have raised the question, "why am I traversing the DOM just to return something I already have?" I could tell that it was not doing anything just by looking at the code because it was an invariant. No experimental evidence needed. 

The best advice I can give is to pay more attention to what variables are in scope and think about how you might use them without having to do a lot of extra work. This case was obvious, but another more common case is that the function may be necessary, but there is a more efficient/elegant solution. I know that seeing these improvements come with time and experience, so keep working to learn and grow and I'm sure you'll get a knack for it :). 

Lastly, try to get in the mindset that **there is always one best way**, not multiple ways to do the same thing. Looking for the best approach will help guide your mind towards recognizing inefficiencies or clunky implementations.

ankitkarna99: Thanks for the guidance. I will pay more attention to the variables in scope, and focus on the best way to solve them.

raineorshine (approved):

PR #1686, opened 2023-01-13
Persist thoughts in separate yjs doc per thought for lazy loading

raineorshine:


raineorshine:
Close in favor of #1713

PR #1693, opened 2023-01-17
First phase capacitor app

ankitkarna99:
## Changes

- Added capacitor to the project for platforms iOS and Android.
- Updated CSS to fix styling issues with `Alert`, `ErrorMessage`, `Drawer`, and `Toolbar` in Native platforms.
- Updated `StatusBar` to work as a dark mode with the official status bar package.
- Added native copy to clipboard method for `CopyOneDrop`
- Added `NoBounce` to fix bouncing WebView that breaks the styling of the Toolbar.

## How to run the app?

- Make sure python is available via the `python` command rather than `python3`.
- Install dependencies: `yarn install`
- Change `server.url` in `capacitor.config.ts` to your local IP address. (Maybe we can use some NPM package to get the local IP instead of manual typing of IP)
- Sync capacitor files: `yarn cap:sync`
- Open iOS or Android project via: `yarn cap:ios` or `yarn cap:android`
- Build on device of choice.


raineorshine:
I added `ios` and `android` to the `.eslintignore`. We can upgrade that with a more fine-grained ignore list but that should at least allow the linter to run.

> * Change `server.url` in `capacitor.config.ts` to your local IP address. (Maybe we can use some NPM package to get the local IP instead of manual typing of IP)

I updated the config to read environment variables from environment-specific files, following the [create-react-app convention](https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used). This way we don't have environment-specific information checked into source control.

- `.env.development`
- `.env.development.local`
- `.env.production`
- `.env.production.local`

You will need to create `.env.development.local` in the project root and set `CAPACITOR_SERVER_URL=http://192.168.1.65:3000/`.

Additionally, according to the Capacitor docs, we should only use `url` for local development with a live-reload server, not for production: https://capacitorjs.com/docs/config

Therefore we will need to configure a production configuration: https://capacitorjs.com/docs/guides/environment-specific-configurations#dynamically-build-for-different-environments

It is already configured to use `.env.production` when NODE_ENV=production is set. We should probably have two scripts, `yarn cap:sync:dev` and `yarn cap:sync:prod`, to trigger the respective environment build.

ankitkarna99:
I have pushed changes that fix: #1694 and #1695 

I disabled the position absolute fix for iOS and instead used the capacitor's keyboard plugin with `resize: 'none'` in the capacitor config to have a fixed toolbar.

I found that there exists a background color property in the capacitor config specific to iOS which removes the white background. Android has no bouncing effect by default, so only iOS configuration is fine.

src/components/AppComponent.tsx
```diff
@@ -209,16 +211,29 @@ const AppComponent: FC<Props> = props => {
...
     safari: isSafari(),
   })
 
+  useEffect(() => {
+    if (Capacitor.isNativePlatform()) {
+      StatusBar.setStyle({ style: Style.Dark })
+      // Android only, set statusbar color to black.
+      if (Capacitor.getPlatform() === 'android') {
+        StatusBar.setBackgroundColor({
+          color: '#000000',
```
raineorshine, line 226: You should use `colors.bg` to reference the background color from the theme instead of hardcoding it.

src/components/AppComponent.tsx
```diff
@@ -209,16 +211,29 @@ const AppComponent: FC<Props> = props => {
...
+      }
+    }
+  }, [])
+
   const globalStyles = useMemo<[string, React.CSSProperties][]>(
     () => [
       [
         // increase specificity to override .popup .modal-actions
-        'a.button.button.button:hover, a.button.button.button:active',
+        'a.button.button.butt on:hover, a.button.button.button:active',
```
raineorshine, line 236: I think this was a mistake

src/components/AppComponent.tsx
```diff
@@ -209,16 +211,29 @@ const AppComponent: FC<Props> = props => {
...
     android: isAndroid,
+    native: Capacitor.isNativePlatform(),
     'drag-in-progress': dragInProgress,
     chrome: /Chrome/.test(navigator.userAgent),
     safari: isSafari(),
   })
 
+  useEffect(() => {
+    if (Capacitor.isNativePlatform()) {
+      StatusBar.setStyle({ style: Style.Dark })
```
raineorshine, line 222: You should set `Style.Dark` or `Style.Light` depending on the value of the `dark` prop.

raineorshine (changes requested):

src/components/AppComponent.tsx
raineorshine: @ankitkarna99

src/selectors/themeColors.ts
```diff
@@ -3,7 +3,7 @@ import theme from './theme'
 
 const colors = {
   dark: {
-    bg: 'rgba(0, 0, 0, 1)',
+    bg: '#000000',
```
raineorshine, line 6: @ankitkarna99 Is there a reason `rgba` does not work? If so, you should add a comment explaining the constraint. Otherwise a developer will not know that there are special restrictions on the format of the color.

I believe rgba was required for web in a previous version of em, but that is no longer the case.

raineorshine (approved):

Issue #1698, opened 2023-01-23
[Android] Cannot select thought text


raineorshine:
## Steps to Reproduce

1. Create a thought
2. Double tap to select

## Current Behavior

Keyboard closes.

## Expected Behavior

Tapped word should be selected.

PR #1699, opened 2023-01-24
Fix android cannot select thought text

ankitkarna99:
Fixes: #1698

## Problem
- `onContextMenu` is triggered by two ways in Android devices, one being double tap, and other being long press.

## Solution
- I did a `console.log` of the event and found that `PointerEvent` is emitted `onContextMenu` rather than `MouseEvent`.
- On the `PointerEvent` we have a property called `pointerType` which is `mouse` when double tapped and `touch` when long pressed.
- Conditional clearing of selection was then applied.


src/hooks/useLongPress.ts
```diff
@@ -85,11 +86,14 @@ const useLongPress = (
...
   )
 
-  const onContextMenu = useCallback((e: React.MouseEvent) => {
-    e.preventDefault()
-    e.stopPropagation()
-    selection.clear()
-    dispatch(editing({ value: false }))
+  const onContextMenu = useCallback((e: React.PointerEvent) => {
+    // Double tap activation of context menu produces a pointerType of `touch` whereas long press activation of context menu produces pointer type of `mouse`
+    if (!isTouch || (isTouch && e.nativeEvent.pointerType === 'touch')) {
```
raineorshine, line 91: You can remove `isTouch &&` since it will always be true.

raineorshine (changes requested):

src/hooks/useLongPress.ts
ankitkarna99: I see. I forgot about the short-circuiting.

raineorshine (approved):

PR #1825, opened 2024-04-18
Upgrade node to v20

raineorshine:


raineorshine:
- BrowserStack workflow seems to be using an old workflow file (?) because it is incorrectly using `actions/setup-node@v1`
- Puppeteer tests pass locally, but fail in GitHub Action workflow.

PR #1836, opened 2024-04-26
Containerize Puppeteer

raineorshine:
Update 2024-05-04: Puppeteer finally passes in GitHub Actions with the `macos-latest` test runner. This issue is now about running the Puppeteer tests in a Docker container. This would normalize browser rendering across platforms, albeit at the cost of additional test ops complexity.

~Puppeteer currently works locally but fails in GitHub Actions.~

PR #1845, opened 2024-04-28
Puppeteer + MacOS

raineorshine:
Run on Mac since Puppeteer snapshots were generated on a Mac.
Otherwise snapshot tests will fail.

Related: https://github.com/cybersemics/em/pull/1836

### DIFF
--- a/.github/workflows/puppeteer.yml
+++ b/.github/workflows/puppeteer.yml
@@ -14,7 +14,10 @@ env:
 jobs:
   run:
     name: Puppeteer
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


raineorshine:
Started failing with the following error last week:

```
Error launching puppeteer: TimeoutError: Timed out after 30000 ms while trying to connect to the browser! Only Chrome at revision r869685 is guaranteed to work.
```

After further investigation, it appears to be an issue on GitHub's side, and not related to any specific commit in this repo. It has now passed twice in GitHub Actions: once in this PR and once in `main` on merge. Hopefully it is stable. If not, consider containerizing Puppeteer and running it in on `ubuntu-latest` (#1836).

PR #1853, opened 2024-05-04
Migration to Vite

hsmr29:


raineorshine (approved):
Looks great so far. Thanks for the granular commits and good commit messages.

Tested on PWA and confirmed MultiGesture and SwipeableDrawer are working.

raineorshine (approved):

PR #1875, opened 2024-05-15
Sort thoughts with formatting above others

### LINKED ISSUE #1806: Sort bold above punctuation
Sort bold/italic/underline above punctuation. 

Note that markdown gets converted to HTML, so the value that will actually be compared in this case is `<b>B</b>`.

## Steps to Reproduce

```
- =sort
  - Alphabetical
    - Asc
- :
- **B**
- A
```

## Current Behavior

":" is above **B**

## Expected Behavior

**B** should be above ":"

hsmr29:
fix #1806


raineorshine (changes requested):
Thanks, looks good!

Could you add support for `<strike>` as well? That’s also a formatting tag.

raineorshine (approved):

PR #1996, opened 2024-06-03
Fix ServiceWorker

### LINKED ISSUE #1919: Service worker broken in vite
A service worker is used to enable offline functionality when using em as a PWA. This broke during the migration from create-react-app (CRA) to vite. Currently we get the following error in production:

> **Error during service worker registration:**
> SecurityError: Script https://em-proto--staging2-cxls71a0.web.app/service-worker.js load failed

**✓ Before** the conversion to vite, you can see that `service-worker.ts` was properly built and added to the `build` directory:
Screenshot 2024-05-17 at 10 23 11 PM

**✗ After** the conversion to vite:
Screenshot 2024-05-17 at 10 25 35 PM

I made a few changes to get the process started:

- `PUBLIC_URL` was automatically set by CRA, but is not set by vite. It can be safely removed since em is not currently hosted on a CDN (41d9b9cd83).
- em uses [workbox](https://developer.chrome.com/docs/workbox/) for service worker support. I added the missing workbox dependencies (a86e4ba58e).

The error cannot be reproduced on localhost since there is different service worker logic for localhost. I'm not sure the best way to troubleshoot this from localhost. Once we generate `build/service-worker.js` I can redeploy to staging2 and test. If that doesn't work, we'll have to do further troubleshooting.

hsmr29:
Fixes #1919

### DIFF
--- a/package.json
+++ b/package.json
@@ -208,6 +208,7 @@
     "typedoc-plugin-rename-named-parameters": "^1.0.6",
     "typescript": "^5.4.5",
     "vite": "^5.2.12",
+    "vite-plugin-pwa": "^0.20.0",
     "vitest": "^1.6.0",
     "webdriverio": "^7.36.0"
   }
--- a/src/service-worker.ts
+++ b/src/service-worker.ts
@@ -49,7 +49,7 @@ registerRoute(
     // Return true to signal that we want to use the handler.
     return true
   },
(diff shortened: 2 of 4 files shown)


raineorshine (approved):
Thank you! Tested deployment and the service worker is registered and the app works offline again.

Issue #1902, opened 2024-05-16
Drag-and-drop snapshot tests


raineorshine:
Add puppeteer snapshot tests that cover the position of various drag and drop elements. There are two test flags that can be enabled to simulate a drag-and-drop. 

- `simulateDrag` - Simulates a drag in progress by rendering the blue lines that indicates where the thought will be inserted on drop (See: DropHover.tsx).
- `simulateDrop` - Renders the valid areas that a thought can be dropped as color blocks (See: DropUncle.tsx, DropChild.tsx, DropEnd.tsx).

For example, with `simulateDrop` enabled:

<img width="422" alt="image" src="https://github.com/cybersemics/em/assets/750276/e13870c6-8ffa-4636-9f5f-7d077dfbd5f3">

Test flags can be enabled in tests by setting `window.em.testFlags.simulateDrag` or `window.em.testFlags.simulateDrop` and triggering a re-render.

## Subtasks

- [x] a. Verify testFlags functionality in browser
- [x] b. Verify puppeteer snapshot tests on local machine
- [x] c. Add testHelpers method to force re-render from puppeteer test
- Create drag tests:
  - [x] d. [DragAndDropThought](https://github.com/cybersemics/em/blob/main/src/components/DragAndDropThought.tsx?rgh-link-date=2024-05-15T20%3A00%3A37Z).
  - [x] e. [DropEnd](https://github.com/cybersemics/em/blob/main/src/components/DropEnd.tsx?rgh-link-date=2024-05-15T20%3A00%3A37Z)
  - [x] f. [DropChild](https://github.com/cybersemics/em/blob/main/src/components/DropChild.tsx?rgh-link-date=2024-05-15T20%3A00%3A37Z)
  - [x] g. [DropUncle](https://github.com/cybersemics/em/blob/main/src/components/DropUncle.tsx?rgh-link-date=2024-05-15T20%3A00%3A37Z)
- Create drop tests:
  - [x] h. [DragAndDropThought](https://github.com/cybersemics/em/blob/main/src/components/DragAndDropThought.tsx?rgh-link-date=2024-05-15T20%3A00%3A37Z).
  - [x] i. [DropEnd](https://github.com/cybersemics/em/blob/main/src/components/DropEnd.tsx?rgh-link-date=2024-05-15T20%3A00%3A37Z)
[...]

PR #2038, opened 2024-06-15
Puppeteer snapshot, drag and drop test written

Zubair286:
#1902


src/e2e/puppeteer/helpers/simulateDragAndDrop.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/** Simulates Drag and Drop functionality in Puppeteer. */
+const dragAndDrop = async (page: Page, selectorDrag: string) => {
+  await page.evaluate(() => {
+    const thoughts = document.querySelectorAll('[aria-label="thought"]')
+    thoughts.forEach((thought, index) => {
+      // Assign an ID
+      thought.id = `thought-${index}`
+    })
+  })
```
raineorshine, line 11: Let's use the existing [getEditable](https://github.com/cybersemics/em/blob/b20fd295186c0b323c036c70bfa635e9da4a58c5/src/e2e/puppeteer/helpers/getEditable.ts) helper for querying a thought by value.

src/e2e/puppeteer/helpers/snapshotSimulation.ts
```diff
@@ -0,0 +1,16 @@
+import { Page } from 'puppeteer'
+import { WindowEm } from '../../../initialize'
+
+const em = window.em as WindowEm
+
+/** Sets testFlags for simulating drag and drop process . */
+const simaulateSnapshot = async (page: Page): Promise<void> => {
```
raineorshine, line 7: Let's call this `simulateDragAndDrop` and have it accept parameters `{ drag?: boolean, drop?: boolean }` so that we can set them independently.

We could also call it `setDragAndDropColorBlocking`. What do you think?

Please also rename the file to match whatever we decide.

src/e2e/puppeteer/helpers/simulateDragAndDrop.ts
```diff
@@ -0,0 +1,31 @@
+import { Page } from 'puppeteer'
```
raineorshine, line 1: Let's call this file `dragAndDrop`, since it does not simulate a drag-and-drop, but actually performs a drag-and-drop in the browser.

src/e2e/puppeteer/helpers/simulateDragAndDrop.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/** Simulates Drag and Drop functionality in Puppeteer. */
+const dragAndDrop = async (page: Page, selectorDrag: string) => {
```
raineorshine, line 4: This function can be split into a function for general drag-and-drop (of any element) and a function for dragging and dropping a thought.

```ts
dragAndDrop(page: Page, selectorDrag: string, selectorDrop: string)

dragAndDropThought(
  page: Page, 
  sourceValue: string, 
  destValue: string, 
  { position: 'after' | 'before' | 'child' }
)
```

src/e2e/puppeteer/helpers/simulateDragAndDrop.ts
```diff
@@ -0,0 +1,31 @@
...
+  const dragEnd = await page.$eval('#thought-2', el => ({
+    x: el.getBoundingClientRect().x + el.getBoundingClientRect().width / 2,
+    y: el.getBoundingClientRect().y + el.getBoundingClientRect().height / 2,
+  }))
+
+  // Move the mouse to the drag target, then press, move to the drop target, and release
+  await page.mouse.move(dragStart.x, dragStart.y)
+  await page.mouse.down()
+  await page.mouse.move(dragEnd.x, dragEnd.y) // The steps can be adjusted for smoother dragging
+  await page.mouse.up()
```
raineorshine, line 28: I wonder why the snapshot shows that thought `a` is still being dragged even after the mouse is released?

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -290,3 +288,25 @@ describe('Font Size: 22', () => {
...
+    await paste(`
+      - a
+      - b
+      - c
+      - d
+    `)
+
+    await simulateDragAndDrop('[aria-label="bullet"]')
+
+    await sleep(2000)
```
raineorshine, line 307: Can this delay be made any shorter?

raineorshine (changes requested):
Looks great. I'm happy that the color blocking is working and showing up in the snapshot. Just a few suggested changes when you have a chance.

raineorshine:
I figured out why the puppeteer test is failing. The snapshot test is configured to run on `macos`, but the PR only contains the ubuntu snapshot. I can push a commit with the `macos` snapshot once the PR is ready to be merged.

src/e2e/puppeteer/helpers/snapshotSimulation.ts
Zubair286: I think `simulateDragAndDrop` is a fine name for this function.

src/e2e/puppeteer/helpers/simulateDragAndDrop.ts
Zubair286: I think getEditable returns the first matching element, we need to get all the thoughts so we might need to create another helper for querying all matching elements, maybe?

src/e2e/puppeteer/__tests__/snapshot.ts
Zubair286: Sure, I reduced it 500ms, 2000ms was just for testing purpose.

src/e2e/puppeteer/helpers/simulateDragAndDrop.ts
raineorshine: Querying thoughts by name will make the tests more readable. We will need a way to query thoughts that share a value, but I don't particularly see the need to query all thoughts in a single test.

Zubair286: I was using `querySelectorAll` to get the thoughts and add an id at runtime, so any thought could be drag and dropped by using the id. I couldn't find anything else to query the thought with, does the `getEditable` helper use the "value" of the thought?

raineorshine: `getEditable` uses XPath, as you can see here: https://github.com/cybersemics/em/blob/b20fd295186c0b323c036c70bfa635e9da4a58c5/src/e2e/puppeteer/helpers/getEditable.ts

Zubair286: I understand, we might need to paste thoughts with specific values and then drag and drop, then we wont need to query all thoughts,

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -287,6 +285,30 @@ describe('Font Size: 22', () => {
...
+    await paste(`
+      - a
+      - b
+      - c
+      - d
+    `)
+
+    await dragAndDropThought('a', 'd')
+
+    await sleep(1500)
```
raineorshine, line 307: This seems like a long delay. Can you tell me a little more about why this is needed?

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -287,6 +285,30 @@ describe('Font Size: 22', () => {
...
+
+    const image = await screenshot()
+    expect(image).toMatchImageSnapshot()
+
+    await simulateDragAndDrop(false, false)
```
raineorshine, line 312: Please add this to the puppeteer [teardown](https://github.com/cybersemics/em/blob/6275018b29631e5d7939fab42e677b8137b8b00d/src/e2e/puppeteer-environment.ts#L46) function.

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -287,6 +285,30 @@ describe('Font Size: 22', () => {
     await sleep(400)
   })
 
-  // run the snapshot tests at font size 22
+  // // run the snapshot tests at font size 22
```
raineorshine, line 288: typo

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -287,6 +285,30 @@ describe('Font Size: 22', () => {
...
   testSuite()
 })
+
+describe('Drag and Drop simalation tests', () => {
```
raineorshine, line 292: typo: "simalation" → "simulation"

src/e2e/puppeteer/helpers/dragAndDrop.ts
```diff
@@ -0,0 +1,23 @@
+import { Page } from 'puppeteer'
+
+/** PErforms Drag and Drop functionality in Puppeteer browser. */
+const dragAndDrop = async (page: Page, selectorDrag: string, selectDrop: string) => {
```
raineorshine, line 4: Let's call the third parameter `selectorDrop` for consistency.

src/e2e/puppeteer/helpers/dragAndDrop.ts
```diff
@@ -0,0 +1,23 @@
+import { Page } from 'puppeteer'
+
+/** PErforms Drag and Drop functionality in Puppeteer browser. */
```
raineorshine, line 3: Incorrect capitalization

src/e2e/puppeteer/helpers/dragAndDropThought.ts
```diff
@@ -0,0 +1,35 @@
+import { Page } from 'puppeteer'
+import getEditable from './getEditable'
+
+/** PErforms Drag and Drop functionality on a hought in Puppeteer browser. */
```
raineorshine, line 4: Incorrect capitalization and typo: "hought" → "thought".

src/e2e/puppeteer/helpers/dragAndDropThought.ts
```diff
@@ -0,0 +1,35 @@
...
+const dragAndDropThought = async (page: Page, sourceValue: string, destValue: string) => {
```
raineorshine, line 5: You left out the parameter `{ position: 'after' | 'before' | 'child' }` that was suggested in https://github.com/cybersemics/em/pull/2038#discussion_r1641218166. Was this intentional?

You don't have to implement each of the positions yet, but we may as well set it up with the desired API now. The semantics are ambiguous without it.

src/e2e/puppeteer/helpers/dragAndDropThought.ts
```diff
@@ -0,0 +1,35 @@
...
+  const sourceElement = await getEditable(page, sourceValue)
+  const destElement = await getEditable(page, destValue)
+
+  const dragStart = await sourceElement.boundingBox()
+  const dragEnd = await destElement.boundingBox()
+
+  if (!dragStart || !dragEnd) {
+    return
+  }
```
raineorshine, line 14: Since this is in the context of testing, you generally want to `throw` instead of silently returning. Tests should fail hard if any of the preconditions are invalid.

src/e2e/puppeteer/helpers/dragAndDropThought.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const yOffset = 20 // Adding offset to drop destination y position so it is dragged exactly on top of it. Right now it fall short just above the drop target
+
+  // Calculate center positions of the elements
```
raineorshine, line 18: The centering calculations are a bit superfluous since we could just drop it on the top, left corner, but I guess it doesn't hurt.

src/e2e/puppeteer/helpers/simulateDragAndDrop.ts
```diff
@@ -0,0 +1,20 @@
+import { Page } from 'puppeteer'
+import { WindowEm } from '../../../initialize'
+
+const em = window.em as WindowEm
+
+/** Sets testFlags for simulating drag and drop process . */
+const simulateDragAndDrop = async (page: Page, drag?: boolean, drop?: boolean): Promise<void> => {
```
raineorshine, line 7: Can you structure the options as an object, i.e. `{ drag: true, drop: true }`? This will make it a little more readable at the call site.

raineorshine (changes requested):

src/e2e/puppeteer/__tests__/snapshot.ts
Zubair286: I added this so all the ui changes are completely done before taking a snapshot, test was failing sometimes if I did not add this here, maybe we can remove it or decrease the delay, I'll check again if that could work

src/e2e/puppeteer/helpers/dragAndDropThought.ts
Zubair286: So can I just add the parameter for now, and we can get back to it at some later point?

src/e2e/puppeteer/helpers/dragAndDropThought.ts
Zubair286: So should I change it or keep as it is?

src/e2e/puppeteer/__tests__/snapshot.ts
raineorshine: Okay. I know there are some animations that take up to 400ms. I'm not sure what would take 1500ms to update. It would be good to know the reason behind it.

src/e2e/puppeteer/helpers/dragAndDropThought.ts
raineorshine: Yes, add the parameter and `throw` on the unsupported options for now.

src/e2e/puppeteer/helpers/dragAndDropThought.ts
raineorshine: You can keep it as is, I'm just pointing out that it was unnecessary.

src/e2e/puppeteer/__tests__/snapshot.ts
Zubair286: I added this to the teardown function, but it started giving me `window is undefined` for some reason, I gave the first argument a new page using the `browser` but it is giving this error. I am not sure why is this happening, any idea?

raineorshine: Hmmm, I'm not sure. If the page is destroyed in teardown, maybe you can simply omit this.

Zubair286: All right sure, I will remove it from here.

src/e2e/puppeteer/__tests__/snapshot.ts
Zubair286: I have set the delay in the `dragAndDropThought` file of 500ms which seems to work better for some reason than setting it in the test. I'm not really sure why is this though.

raineorshine (approved):

src/e2e/puppeteer/helpers/dragAndDropThought.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  // Offset to add to drop destination y position so it is dragged exactly on top of the drop traget.
+  // Otherwise it falls short, just above the drop target.
+  const yOffset = 20
+
+  // Calculate center positions of the elements
+  const dragPosition = {
+    x: dragStart.x + 1,
+    y: dragStart.y + 1,
+  }
```
raineorshine, line 47: I had to add 1 to the x and y positions to get the drag to trigger on MacOS.

raineorshine (changes requested):
I added ubuntu to the puppeteer workflow, so snapshots are tested in both ubuntu and macos now. I generated the new snapshot for macos. 

Could you update the ubuntu snapshot? It looks different now that the delay was added after the drop.

```
nr test:puppeteer -- snapshot -t Drag -u
```

src/e2e/puppeteer/helpers/dragAndDropThought.ts
Zubair286: Sure thing, updating snapshots and pushing them. Hmm, not sure why is this maybe the rendering difference between the platforms.

Zubair286: All checks have passed, please merge when you get the chance.

raineorshine (approved):

Issue #1826, opened 2024-04-18
Empty context view does not take up space


raineorshine:
## Steps to Reproduce

```
- A
- B
```

1. Set the cursor on `A`.
2. Activate the context view by clicking <img width="26" alt="image" src="https://github.com/cybersemics/em/assets/750276/06a46043-5d0e-4c32-847a-1994f35a824e"> in the toolbar or hitting <kbd>option/alt</kbd> + <kbd>shift</kbd> + <kbd>S</kbd>.

## Current Behavior

`B` overlaps with the empty context view message of `A`.

![image](https://github.com/cybersemics/em/assets/750276/6fc9ced2-e997-4df6-a792-f5af7144487f)

## Expected Behavior

`B` should be rendered after the empty context view message.

The empty context view message should be measured when determining the height of `A` in the [LayoutTree](https://github.com/cybersemics/em/blob/b20fd295186c0b323c036c70bfa635e9da4a58c5/src/components/LayoutTree.tsx).

PR #2070, opened 2024-06-24
Fixed: Empty context view does not take up space

Fenomen-Alex:
Issue 1826 https://github.com/cybersemics/em/issues/1826

I've tracked height calculating from different actions and added `useEffect` dependency to recalculate height of current thought on `toggleContext`.


raineorshine:
I have already hired a developer for this position. Thank you anyway.

FYI The proposed change fixes the issue, but causes all thoughts to re-render, when only the particular thought whose height is updated should re-render.

Issue #2063, opened 2024-06-23
Footer: Breadcrumbs bottom margin is sometimes incorrect


raineorshine:
## Steps to Reproduce

1. Create a thought `hello`.
2. Set the cursor to null.
3. Quickly scroll to the bottom. 
4. Scroll up and repeat. 

## Current Behavior

The safe-area below the breadcrumbs is sometimes preserved when scrolled all the way to the bottom. 

![image](https://github.com/cybersemics/em/assets/750276/54daeffd-2afd-4a0e-80df-065b42f6de1c)

## Expected Behavior

When scrolled to the top, the breadcrumbs should be rendered above the bottom safe area.

When scrolled to the bottom of the document, the bottom margin is not needed because the safe area is not obscured. If the safe area is partially obscured, the bottom margin should adjust dynamically so that there is no extra space in between the breadcrumbs and the safe area at any point during scrolling, i.e. the breadcrumbs should have a smooth transition from being stuck to the bottom of the screen above the safe area to being above the footer.

![image](https://github.com/cybersemics/em/assets/750276/bdf3dcbb-fbba-4e5e-9f45-c2e36666dd4b)

zahidhussain998:
i fixed this #2071

zahidhussain998:
please check my PR which i raised

trevinhofmann:
@raineorshine -- Can you please confirm if this is still an issue, and possibly any other information on how to reproduce it? I haven't been able to get the bottom margin to appear. Here's a video of the simulated mobile device in Chrome, and I have the same results on a physical Android device:

https://github.com/user-attachments/assets/02a3062f-964e-4135-9d4f-1578890e8bd3

raineorshine:
It looks like it now does not respect the bottom safe area at all. Tested on physical iPhone and Simulator.

I think we will need to restore the safe area margin and then see if the irregular spacing is still an issue.

I also missed a step: The cursor has to be set to null, otherwise the breadcrumb are not rendered with position: fixed.

trevinhofmann:
Ah! I didn't realize until now that this was for the Capacitor app.

> It looks like it now does not preserve the bottom safe area at all.

This is most likely a result of: https://github.com/cybersemics/em/pull/2593/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R10

https://developer.mozilla.org/en-US/docs/Web/CSS/env

[...]

PR #2071, opened 2024-06-24
fix(footer): Adjust bottom margin based on scroll position to preserve safe area 2063

zahidhussain998:
This commit fixes the issue of incorrect _bottom margin_ for the footer breadcrumbs during scrolling (#2063). The Footer component now _dynamically_ _adjusts the bottom margin based on the user's scroll position_. A new state variable isScrolledToBottom was introduced to track the scroll position, and a useEffect hook handles scroll events to update this state. The bottom margin is conditionally applied to ensure it's only present when needed

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -105,8 +105,10 @@ const Footer = () => {
     useEffect(() => {
     // Handle scroll position
     const handleScroll = () => {
-      const bottom = window.innerHeight + window.scrollY >= document.documentElement.scrollHeight;
-      setIsScrolledToBottom(bottom);
+      const distance = document.documentElement.scrollHeight - window.innerHeight - window.scrollY;
+      setIsScrolledToBottom(distance);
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
Hi, thanks for your contribution.

There are a couple issues with the proposed solution.

1. The issues specifies "If the safe area is partially obscured, the bottom margin should adjust accordingly." That could be said more clearly, but basically it means that the bottom margin needs to be adjusted based on the distance from the bottom of the page. So `isScrolledToBottom: number` should probably be `distanceFromScrollBottom: number` instead.
2. The `.nav-bottom .nav-container` [already uses --safe-area-inset-bottom](https://github.com/cybersemics/em/blob/58ceb21469a42deaea6a187d09c4dfcb6c437d27/src/App.css#L791). The solution should develop an integrated solution applied to the same element, rather than a separate solution applied to the footer.

raineorshine:
Checking in. A solution needs to be developed that takes into account the existing safe area logic. Are you still interested in this, or should I close?

raineorshine:
Closing due to inactivity

Issue #543, opened 2020-04-12
Fix Divider rendering in table


raineorshine:
# Case 1

Normal view

## Steps to Reproduce

```
- a
  - =children
    - =pin
      - true
  - b
    - b1
    - ---
    - b2
  - c
    - This is a longer thought
    - ---
    - c2
```

## Current Behavior

Divider has a fixed width.

<img width="338" alt="image" src="https://github.com/cybersemics/em/assets/750276/3d9c9c30-e415-4f7b-a457-afd181fe6ee2">

## Expected Behavior

Divider width should be equal to the width of the widest sibling thought.

<img width="312" alt="image" src="https://github.com/cybersemics/em/assets/750276/1a52d07a-701c-4b5a-8b79-11257751e293">

# Case 2

Table View, Column 2, only child divider

## Steps to Reproduce

```
- Roadmap
  - =view
    - Table
  - Meta
    - Keyword Search
  - Adult Development
    - ---
  - Hypertext
    - Future of Text
    - ---
    - Rheingold
    - Landow
    - HT Best Of
    - HT '19
    - HT '18
    - HT '17
  - Semiotics
    - ---
```

## Current Behavior

<img width="389" alt="Brave Browser 2024-07-03 12 35 45" src="https://github.com/cybersemics/em/assets/750276/45392672-9853-42f4-a48e-a2c6b1f0d004">

## Expected Behavior

A divider *that is an only child* in column 2 of a table should be left aligned with the bullets and as wide as the widest col2 thought in the same table. 

In this example, `Adult Development` and `Semiotics` have an only-child divider, so their divider widths should be set to the widest width of all col2 thoughts in the table.

The divider in `Hypertext` is not an only-child, so it should be set to the width of the widest sibling, as usual.

<img width="393" alt="image" src="https://github.com/cybersemics/em/assets/750276/d8cb8e56-9954-49d7-bfe5-aa4eabd07870">

incora-dev:
Estimate: 1.5h

incora-dev:
@raineorshine For now, width for divider calculates like max-width of each subling element. But the problem occurs when we have a group with only divider(in this case we set 85px). What should we do it the case where we have only one divider in the group?

raineorshine:
Ah, that makes sense. If it is the only-child in column 2 of a table, a divider should be the max-width of all of the column 2 thoughts.

Let's hold off on this for now though, it's just too low priority to spend time on.

raineorshine:
[...]

PR #2072, opened 2024-06-25
Initial Table View snapshot tests; optional Docker and VSCode configs

### LINKED ISSUE #1871: Table: Add snapshot test for col1 narrow + no wrap width
```
- X
  - =view
    - Table
  - A
    - Atlanta, Georgia
  - B
    - Boston, Massachusetts
  - C
    - Cincinnati, Ohio
```

image

### LINKED ISSUE #1872: Table: Add snapshot test for col2 narrow + no wrap width
```
- X
  - =view
    - Table
  - All of Us Strangers
    - 4/5
  - Eileen
    - 2/5
  - May December
    - 2/5
```

image

mathiscode:
This PR resolves #1871 and resolves #1872 by adding snapshot test cases for Table View.

Also included is a `docker-compose.yml` to allow easy running, debugging, and testing with Docker to support a more dynamic development experience since port 3000 is the default for a lot of frameworks and can often already be bound. A `.vscode/launch.json` is also included for real-time debugging with VSCode. These are optional and shouldn't affect anyone's current workflow.

To use Docker, run `npm run start:docker` or `docker compose up` and connect as normal.


raineorshine (changes requested):
> This PR resolves #1871 and resolves #1872 by adding snapshot test cases for Table View.

Looks pretty good. Let's just generate a single snapshot for the default font size (18). It's not necessary to render it at multiple font sizes to prove the table view is rendered correctly.

The only issue I see is that the first column in the [no wrap](https://github.com/mathiscode/em/blob/main/src/e2e/puppeteer/__tests__/__image_snapshots__/ubuntu/snapshot-ts-src-e-2-e-puppeteer-tests-snapshot-ts-font-size-18-default-table-view-col-2-narrow-no-wrap-width-1-snap.png) snapshot is wrapping, which is not what we want to test. Maybe this is due to the puppeteer window being too narrow. We could decrease the font size to prevent it from wrapping, but we really ought to use a window width that will pass at the default font size. 

> Also included is a `docker-compose.yml` to allow easy running, debugging, and testing with Docker to support a more dynamic development experience since port 3000 is the default for a lot of frameworks and can often already be bound. A `.vscode/launch.json` is also included for real-time debugging with VSCode. These are optional and shouldn't affect anyone's current workflow.

A lighter weight solution here would be to add a VITE_PORT option to the vite server config. You can see how to read from env in the vite config here: https://stackoverflow.com/questions/66389043/how-can-i-use-vite-env-variables-in-vite-config-js

raineorshine:
FYI I will be merging your changes into the [table2](https://github.com/cybersemics/em/tree/table2) branch during your trial period.

mathiscode:
Dividers should render as expected now, which should resolve #543

raineorshine (approved):

raineorshine (changes requested):
I'm seeing an error in the terminal when I run `npm run test:puppeteer`:

```
WebSocket server error: Port is already in use (x2)
```

I'm not sure if it affects the tests, but I can confirm that it does not occur in `main`.

raineorshine (approved):

PR #2078, opened 2024-06-26
Fix height calculation on context view toggle

baproactive:
#1826 

trigger VirtualThought height update on context view toggle and no context found

### DIFF
--- a/src/components/VirtualThought.tsx
+++ b/src/components/VirtualThought.tsx
@@ -12,6 +12,7 @@ import calculateAutofocus from '../selectors/calculateAutofocus'
 import findDescendant from '../selectors/findDescendant'
 import { findAnyChild, hasChildren } from '../selectors/getChildren'
 import getThoughtById from '../selectors/getThoughtById'
+import isContextViewActive from '../selectors/isContextViewActive'
 import store from '../stores/app'
 import editingValueStore from '../stores/editingValue'
 import equalPath from '../util/equalPath'
@@ -21,8 +22,6 @@ import noteValue from '../util/noteValue'
 import DropChild from './DropChild'
 import DropUncle from './DropUncle'
 import Subthought from './Subthought'
-import isContextViewActive from '../selectors/isContextViewActive'
-import getContexts from '../selectors/getContexts'
 
 /** A resize handler that should be called whenever a thought's height has changed. */
 export type OnResize = (args: {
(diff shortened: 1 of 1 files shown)


src/components/VirtualThought.tsx
```diff
@@ -153,7 +158,7 @@ const VirtualThought = ({
       isVisible: isVisibleNew,
       key: crossContextualKey,
     })
-  }, [crossContextualKey, onResize, path, thought.id])
+  }, [crossContextualKey, onResize, path, thought.id, noOtherContexts])
```
raineorshine, line 161: It's not quite right to add `noOtherContexts` as a dependency to the `updateSize` definition, since the `updateSize` function does not depend on it. It would be better to add it to the `useEffect` hook that *calls* `updateSize`.

The problem is indirectly solved with the current approach because `updateSize` itself is a dependency of the `useEffect` hook that calls `updateSize`.

src/components/VirtualThought.tsx
```diff
@@ -99,6 +101,9 @@ const VirtualThought = ({
   const fontSize = useSelector(state => state.fontSize)
   const note = useSelector(state => noteValue(state, thought.id))
   const ref = useRef<HTMLDivElement>(null)
+  const noOtherContexts = useSelector(
+    state => isContextViewActive(state, simplePath) && getContexts(state, thought.value).length <= 1,
+  )
```
raineorshine, line 106: I think you can select the context view active state and ignore the number of contexts. We probably want to recalculate the height whenever the context view is toggled.

raineorshine (changes requested):

src/components/VirtualThought.tsx
baproactive: Yes, that makes sense. 
I've pushed the update.

src/components/VirtualThought.tsx
baproactive: Yes, I removed the context length check.

src/components/VirtualThought.tsx
```diff
@@ -99,6 +100,7 @@ const VirtualThought = ({
   const fontSize = useSelector(state => state.fontSize)
   const note = useSelector(state => noteValue(state, thought.id))
   const ref = useRef<HTMLDivElement>(null)
+  const noOtherContexts = useSelector(state => isContextViewActive(state, simplePath))
```
raineorshine, line 103: Make sure to update the function name to reflect the new functionality. I've updated it this time.

PR #2080, opened 2024-06-27
Adds snapshot tests for #1871 and #1872

iamplugged:
- Adds snapshots tests for 
- #1871
- #1872

A few observations about #1872:
- The layout is not consistent, the column wraps when we move to something else and come back. Like here: https://www.loom.com/share/21f3671314504c83bfd290d66eb0232a
- The vertical spacing also looks a bit off between the items
- In the snapshot tests, column 1 is wrapped even though if there's enough space. Even after setting the viewport dimensions to 1920 X 1080, the issue persists.


raineorshine (changes requested):
Looks good, thanks.

Let's just generate two snapshots at the default font size. That is sufficient for proving that the table layout is working correctly. If we find table layout bugs that are specific to a different font size in the future, then we can reassess.

iamplugged:
> Looks good, thanks.
> 
> Let's just generate two snapshots at the default font size. That is sufficient for proving that the table layout is working correctly. If we find table layout bugs that are specific to a different font size in the future, then we can reassess.

@raineorshine Updated

raineorshine:
Thank!

I will be merging into the https://github.com/cybersemics/em/tree/devan-table branch during your trial period. I recommend branching from there instead of `main`.

raineorshine:
> I will be merging into the [`devan-table`](https://github.com/cybersemics/em/tree/devan-table?rgh-link-date=2024-06-27T14%3A54%3A33Z) branch during your trial period. I recommend branching from there instead of `main`.

Sorry, I meant to name it [table-devan](https://github.com/cybersemics/em/tree/table-devan). Renamed.

Issue #1712, opened 2023-02-17
Ellipsize url to single line


raineorshine:
## Steps to Reproduce

1. Create a thought that is a url. 
2. Set the cursor to null. 

## Current Behavior

The url ellipsizes after a fixed number of characters, regardless of screen width. 

## Expected Behavior

A non-cursor url thought should ellipsize at the right edge of the screen, regardless of the screen size or the number of characters in the thought. 

You should be able to use the [text-overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) CSS property to avoid having to measure the width. Make the container slightly smaller to leave room for the url icon.

PR #2082, opened 2024-06-27
Ellipsize url to single line with text-overflow

baproactive:
#1712

 - fix box sizing for .editable-annotation and .child .editable to make sure they have the same width
 - remove manual super script position since it will be automatically handled.
 - add .single-line css for auto text ellipse on url. adjust link tag position mismatch due to overflow: none
 - make sure the .editable-annotation-text match with .child .editable

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -550,7 +550,7 @@ a.text-link {
 
 .thought,
 .thought-annotation {
-  width: calc(100% - 1em); /* expand click area. offset content padding to prevent line break */
+  max-width: calc(100% - 1em); /* expand click area. offset content padding to prevent line break */
   /* do not set font-weight or it will override =heading style. */
   margin-top: 0;
   display: inline-block;
@@ -622,6 +622,8 @@ a.text-link {
   margin: -0.5px 0 0 calc(1em - 18px);
   /* create stacking order to position above thought-annotation so that function background color does not mask thought */
   position: relative;
+  /* Prevent the selection from being incorrectly set to the beginning of the thought when the top edge is clicked, and the end of the thought when the bottom edge is clicked. Instead, we want the usual behavior of the selection being placed near the click. */
+  clip-path: inset(0.001px 0 0.1em 0);
   /* Use minimum height to the cover the gap left by clip-path */
   min-height: 1.9em;
 }
--- a/src/App.css
+++ b/src/App.css
@@ -570,7 +570,7 @@ a.text-link {
   display: inline-block;
   width: 100%;
   /* Apply the same padding on .editable_annotation instead of .editable-annotation-text */
-  padding: 0 0.2em 0 0.2em;
+  padding: 0 1em 0 0.333em;
   box-sizing: border-box;
 }
 
@@ -612,7 +612,7 @@ a.text-link {
   /* Add some padding-left otherwise caret is invisible on empty elements. */
   /* Cannot use padding-top on editable, as clicking it causes the selection to go to the beginning. */
   /* Cannot use padding-bottom on editable, as clicking it causes the selection to go to the end. */
-  padding: 0 0.2em 0 0.2em;
+  padding: 0 1em 0 0.333em;
   box-sizing: border-box;
 }
 
@@ -1819,12 +1819,12 @@ h1 .num-contexts {
   text-overflow: ellipsis;
   overflow: hidden;
   white-space: nowrap;
-  max-width: calc(100% - 5rem);
+  max-width: calc(100% - 1em);
 }
 
 /* Fix a tag position after overflow:hidden applined on span tag. */
 .single-line + a {
-  top: -12px;
+  position: absolute;
 }
 
 @media screen and (min-width: 480px) {
--- a/src/App.css
+++ b/src/App.css
@@ -550,7 +550,7 @@ a.text-link {
 
 .thought,
 .thought-annotation {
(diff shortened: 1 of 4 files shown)


src/App.css
```diff
@@ -550,7 +550,7 @@ a.text-link {
 
 .thought,
 .thought-annotation {
-  max-width: calc(100% - 1em); /* expand click area. offset content padding to prevent line break */
+  width: calc(100% - 1em); /* expand click area. offset content padding to prevent line break */
```
raineorshine, line 553: This change has the undesirable effect of extending the click area to the right edge of the screen, which more easily interferes with scrolling and user gestures. Instead, the width of a thought should generally be limited to the width of its content, plus a little padding.

src/App.css
```diff
@@ -603,19 +607,21 @@ a.text-link {
...
+
+.child .editable,
+.editable-annotation-text {
   /* slide editable up so that it overlaps with the previous thought, ensuring there is no dead click zone from clip-path */
-  margin: -0.5px -1em 0 calc(1em - 18px);
+  margin: -0.5px 0 0 calc(1em - 18px);
   /* create stacking order to position above thought-annotation so that function background color does not mask thought */
   position: relative;
-  /* Prevent the selection from being incorrectly set to the beginning of the thought when the top edge is clicked, and the end of the thought when the bottom edge is clicked. Instead, we want the usual behavior of the selection being placed near the click. */
-  clip-path: inset(0.001px 0 0.1em 0);
```
raineorshine, line 618: Was removing this intentional?

raineorshine (changes requested):
Good work, though we have to find a solution that does not extend the width of the thought.

src/App.css
baproactive: Yes, let me use `max-width`

src/App.css
baproactive: Yes, I deleted it to view the annotation text for debugging purpose, but forgot to restore the change. 
I added them back.

raineorshine (changes requested):
The width looks good now, and the ellipsis works as expected. Thank you.

I am seeing two regressions due to the change in margin/padding on the Thought and ThoughtAnnotation.

# 1. Background color padding

When the thought is rendered with a background color (which is applied to the `ThoughtAnnotation`), the color extends too far to the right.

## Steps to Reproduce

```
- a
  - =style
    - color
      - rgba(0, 0, 0, 1)
  - =styleAnnotation
    - backgroundColor
      - yellow
```

## Regression

Brave Browser 2024-06-28 17 58 54

## Expected Behavior

Brave Browser 2024-06-28 17 59 13

# 2. Wrap length

Both URL and non-URL thoughts now wrap at a shorter width than they should.

## Steps to Reproduce

```
- This is a long thought that will wrap if it keeps going and hits the end of the line where it is
- https://github.com/cybersemics/em/01234567890-1234567890-1234567890-1234567890
```

## Regression

Brave Browser 2024-06-28 18 10 23

## Expected Behavior

Brave Browser 2024-06-28 18 09 40

src/App.css
```diff
@@ -603,15 +607,19 @@ a.text-link {
...
 
-.child .editable,
-.editable-annotation-text {
+.child .editable {
   /* Add some padding-right for increased click area. */
   /* Add some padding-left otherwise caret is invisible on empty elements. */
   /* Cannot use padding-top on editable, as clicking it causes the selection to go to the beginning. */
   /* Cannot use padding-bottom on editable, as clicking it causes the selection to go to the end. */
-  padding: 0 1em 0 0.333em;
+  padding: 0 0.2em 0 0.2em;
```
raineorshine, line 615: This padding is necessary to make it easier for users to set the caret at the end of the thought, so unfortunately we can't alter it. 

Make sure to read surrounding comments, as they often contain important information about why a piece of code is needed.

## Regression

image

## Expected Behavior

image

raineorshine (changes requested):
I can confirm the background color issue is fixed. The wrap width needs to be fixed a different way, as noted.

The thought seems to be getting ellipsized to a shorter length than necessary, leaving extra space on the right.

## Current Behavior

There is extra space on the right.

Brave Browser 2024-06-29 11 32 48

## Expected Behavior

The below screenshot shows an ellipsized thought side-by-side with a single line thought that uses all of its available space. In the ellipsized thought, the url icon should end up close to the right edge of the available space. The exact position will vary by about one character width, depending on where the string is ellipsized (here, after `5`; even though there are a few pixels left, it has to ellipsize at `5` or `6`, not in between).

Brave Browser 2024-06-29 11 35 33

src/App.css
baproactive: Yes, I handled these updates.

raineorshine (changes requested):
This looks good, except it brought back the Wrap Width regression. I have my browser width at 945px for the above content.

baproactive:
I have updated the structure a bit to make sure it takes all possible width.

raineorshine (changes requested):
Thanks. I can confirm thoughts are wrapping at the correct width now.

The thought text is overlapping the url icon on multiline urls:
image

Here are the test cases:

- [x] Ellipsize non-cursor urls at the end
- [x] Background color padding
- [x] Preserve wrap width
- [x] Clickable area of thought extends 1em to the right
- [x] No extra space after single-line ellipsized url icon
- [ ] Multiline URL icon should not overlap thought text

raineorshine (approved):
Nice work, everything checks out now. Thanks.

Issue #1889, opened 2024-05-15
Table: Heights are incorrect on initial render


raineorshine:
## Steps to Reproduce

```
- a
  - =view
    - Table
  - =pin
    - true
  - Boston
    - c
  - Chicago
    - d
- x
```

Import the above thoughts and refresh the page.

## Current Behavior

When the page is refreshed, the heights are wrong. When the cursor is set on `a`, the heights are corrected.

https://github.com/cybersemics/em/assets/750276/3ffd13f0-2dae-41ae-a693-932fc3c8646d

## Expected Behavior

Heights should be recalculated and positions updated after thoughts are initially rendered.

emonidi:
Hi There, 
This architecture will only bring you more and more headaches down the line. You need to build a real dom , a nested dom structure and not calculating the height of each element dynamically. I am sorry if I am coming too hard but in 8 years with react I haven't seen such thing. I spent an hour going around the code and again if you do not fix the way you are rendering the tree and use real dom elements and style of display flex this thing will not work. 
Anyway I managed to solve this bug.
Here's a screencast: https://www.loom.com/share/732e2369ae694f7cbfc5b1df48030508?sid=0815b244-c8d0-4a74-88aa-a812639703ae
I am sorry if I am coming too hard, that is not my intetion

raineorshine:
Hi, thanks for your input. The reason for the architecture is that it enables animation between any two states of the tree. For example, if an item is deleted out of the middle of the tree and its descendants are shifted up a level, the descendants can be smoothly animated to their new locations. I'm not sure how that would work if you relied on the normal DOM layout.

emonidi:
> Hi, thanks for your input. The reason for the architecture is that it enables animation between any two states of the tree. For example, if an item is deleted out of the middle of the tree and its descendants are shifted up a level, the descendants can be smoothly animated to their new locations. I'm not sure how that would work if you relied on the normal DOM layout.

The first thing that comes to mind is once you delete the text from a node you would get its calculated height, set (enforce) its height via css and then animate down to zero via js or css transition. When animation is done you remove the node completely from the DOM. That would ensure smooth animation of the nodes below appearing to slide up.

raineorshine:
That may work for a simple animation, but I don't think it would be as easy for a more complex animation that involved many items changing locations. For example, converting a table to a nested list and having all items animate into the correct place.

emonidi:
Have you got a test case to reproduce and see what exactly do you mean?

[...]

PR #2086, opened 2024-06-28
Fix table heights incorrect on initial render

baproactive:
#1889

Root Cause:
Unlike other `Thoughts` `table thoughts` are rendered within a fixed width container. And to expand click area we have style max-width: calc(100% - 1em) Due to this 1em the table cell rendered in 2 lines, and later the width is updated along with other cols. But since the initial render was 2 lines it gets assigned with multiline classname which gives additional padding. <- shows extra space After user selects item or any other render action the multiline class name gets removed causing the cell back to normal height

Solution:
assign max-width 100% instead of calc(100% - 1em) on table col specific cells

### DIFF
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -1,3 +1,4 @@
+import classNames from 'classnames'
 import _ from 'lodash'
 import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { useSelector } from 'react-redux'
@@ -660,6 +661,7 @@ const LayoutTree = () => {
                 // The key must be unique to the thought, both in normal view and context view, in case they are both on screen.
                 // It should not be based on editable values such as Path, value, rank, etc, otherwise moving the thought would make it appear to be a completely new thought to React.
                 key={key}
+                className={classNames({ 'table-col': isTableCol1 })}
                 style={{
                   position: 'absolute',
                   // Cannot use transform because it creates a new stacking context, which causes later siblings' DropChild to be covered by previous siblings'.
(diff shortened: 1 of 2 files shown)


src/components/LayoutTree.tsx
```diff
@@ -660,6 +661,7 @@ const LayoutTree = () => {
                 // The key must be unique to the thought, both in normal view and context view, in case they are both on screen.
                 // It should not be based on editable values such as Path, value, rank, etc, otherwise moving the thought would make it appear to be a completely new thought to React.
                 key={key}
+                className={classNames({ 'table-col': isTableCol1 })}
```
raineorshine, line 664: Let's call the class `table-col1` for clarity.

raineorshine (changes requested):
Works great! Thank you.

Could you add the explanation in a comment in the code? I think it'll be important to document in the code why we need to override the default width of col1 thoughts.

src/components/LayoutTree.tsx
baproactive: Yes.

raineorshine (changes requested):
> Could you add the explanation in a comment in the code? I think it'll be important to document in the code why we need to override the default width of col1 thoughts.

In case you missed this ↑

raineorshine (approved):

raineorshine:
Thanks, looks good.

Can you rebase on the updated [linewrap-joseph](https://github.com/cybersemics/em/tree/linewrap-joseph) branch? It was rebased recently with changes from `main`.

baproactive:
It's rebased now.

raineorshine (approved):

Issue #1842, opened 2024-04-28
Fix Divider rendering


raineorshine:
## Steps to Reproduce

```
- a
- ---
- b
```

## Current Behavior

Divider is rendered too far to the left.

<img width="346" alt="image" src="https://github.com/cybersemics/em/assets/750276/6306c692-c8cc-44d2-99ff-908a77e726be">

## Expected Behavior

Divider should be rendered so that its left edge is flush with the bullets at the same level.

PR #2101, opened 2024-07-01
#1842 #543 Fixes divider alignment and width

iamplugged:
- #1842 , #543


src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    // Get all siblings of the identified parent node
+    const siblings = Array.from(parentTreeNode.parentNode?.children || []) as HTMLElement[]
+
+    const currentIndex = siblings.indexOf(parentTreeNode)
+
+    // Collect consecutive siblings at the same depth
+    let consecutiveSiblings = [
+      ...collectConsecutiveSiblings(siblings, currentIndex, targetDepth, -1),
+      ...collectConsecutiveSiblings(siblings, currentIndex, targetDepth, 1),
```
raineorshine, line 101: Can you explain the concept of `direction` that you are using? I'm not quite getting it. It appears that the collect functions are always called in both directions.

Is it possible to do this with a more straightforward map and filter?

raineorshine (changes requested):
Two behavioral issues I'm seeing:

1. There is a delay after each keystroke before the widths are updated.

https://github.com/cybersemics/em/assets/750276/2c493d4a-0800-4480-aba3-4fa89f0fe2da

2. Widths do not update when new thoughts are added.

https://github.com/cybersemics/em/assets/750276/f0dd61cd-c871-4194-a342-666b80c98db5

It may be worth it for us to discuss possible solutions.

More generally, I have some concerns about the architectural approach. A lot is happening outside React. The use of data attributes is much less common in React, for example. It's usually better to pass values through props, making the dependencies more explicit. In this implementation we have implicit dependencies on `.tree-node`, `[data-depth]`, `.thought`, and `.divider`. I'm worried that we are recreating a solution that is just as fragile as the previous one, and will silently break if any of the DOM structure or classes change. That said, the tight coordination that is necessary to calculate the maximum width across tree-nodes that do not sit side-by-side as simple siblings in the DOM does make this much harder. I don't see an obvious alternative, but I did want to bring up my concern.

src/hooks/useMaxSiblingWidth.ts
iamplugged: @raineorshine 

The problem is that the thought nodes including parents, children, grandchildren all at the same level in the DOM and positioned accordingly. There's no hierarchy.

So to find the sibling nodes, we need to traverse forward and backward until the depth of the node changes. That way we can collect all the sibling nodes.

map and filter cannot be used since we are trying to traverse from a point and only until a specific condition is met.

image

iamplugged:
@raineorshine I've fixed the second issue. For the first issue I've added an animation to make it look smooth.

Lets discuss the other concerns and possibilities

src/hooks/useMaxSiblingWidth.ts
raineorshine: I see, that makes sense now. Thank you for the thorough explanation and video.

So what you are doing is using the absolute depth to reconstruct the parent thought, to determine all of the thoughts with the same parent as the divider. However instead of encoding the depth in the DOM, why not encode the parent path? This can be used to uniquely identify all of the divider's siblings. We already have access to the parent path in React (i.e. `[roadmapId, hypertextId]`, accessed with `parentOf(path)`). You can encode it in the DOM and then query the matching parent path to determine the divider's siblings.

The same approach can be taken for the col2 only-child divider, you just need to store the grandparent path (i.e. the path of the table thought) to identify all col2 thoughts in the same table.

raineorshine:
> @raineorshine I've fixed the second issue. For the first issue I've added an animation to make it look smooth.

The animation helps, but it's really the lag that is the problem. I think it needs to be updated in realtime to look okay.

Try subscribing to the [editingValueStore](https://github.com/cybersemics/em/blob/472472069b4637a88c9bbb14f2ce6918ad742b1e/src/stores/editingValue.ts) to respond to the value in realtime.

src/components/LayoutTree.tsx
```diff
@@ -654,9 +654,14 @@ const LayoutTree = () => {
             // Perform this check here instead of in virtualThoughtsPositioned since it changes with the scroll position (though currently `sizes` will change as new thoughts are rendered, causing virtualThoughtsPositioned to re-render anyway).
             if (belowCursor && !isCursor && y > viewportBottom + height) return null
 
+            const parentPath = parentOf(path)
+            const grandParentPath = parentOf(parentOf(path))
+
             return (
               <div
                 aria-label='tree-node'
+                data-parent-path={parentPath}
```
raineorshine, line 663: Use [hashPath](https://github.com/cybersemics/em/blob/98dff32074ed855a4e9348a2bc00c20cbf3fa42c/src/util/hashPath.ts) for a more explicit conversion of `Path` to a `string`.

src/components/Divider.tsx
```diff
@@ -1,39 +1,34 @@
...
+import useMaxSiblingWidth from '../hooks/useMaxSiblingWidth'
 import fastClick from '../util/fastClick'
 import head from '../util/head'
+import parentOf from '../util/parentOf'
 
 /** A custom horizontal rule. */
 const Divider = ({ path }: { path: Path }) => {
   const dividerSetWidth = React.createRef<HTMLInputElement>()
   const dispatch = useDispatch()
+  const [width, setWidth] = useState(DIVIDER_MIN_WIDTH)
```
raineorshine, line 16: I think you can replace `dividerSetWidth`, `width/setWidth`, and the `useEffect` with a single `useMemo` that depends on `maxSiblingWidth`.

src/components/Divider.tsx
```diff
@@ -44,7 +39,8 @@ const Divider = ({ path }: { path: Path }) => {
         maxWidth: '100%',
         padding: '10px 4px 16px',
         position: 'relative',
-        width: 85,
+        width,
+        transition: 'width 0.3s ease-in-out',
```
raineorshine, line 43: Let's remove the `transition`. Since the width of a thought changes instantaneously while editing, we should update the divider width instantaneously so it matches.

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -267,6 +267,53 @@ describe('Layout Tree: Table View', () => {
     const image = await screenshot()
     expect(image).toMatchImageSnapshot()
   })
+
+  it('Normal View: divider should be as wide as the widest sibling', async () => {
```
raineorshine, line 271: Tests look good, thanks.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+  const path = useMemo(() => parentPath.join(','), [parentPath])
+
+  const updateMaxWidth = useCallback(() => {
+    const nodes = collectNodes(path)
+    setMaxWidth(calculateMaxWidth(nodes))
+  }, [path])
+
+  editingValueStore.subscribe(() => {
+    updateMaxWidth()
+  })
```
raineorshine, line 61: Use `editingValueStore.useEffect(updateMaxWidth)`. It is defined in [react-ministore.ts](https://github.com/cybersemics/em/blob/98dff32074ed855a4e9348a2bc00c20cbf3fa42c/src/stores/react-ministore.ts) and ensures that changes are synced with the React render lifecycle.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  useEffect(updateMaxWidth, [updateMaxWidth, parentPath])
```
raineorshine, line 63: I think this will update on every render since `parentPath` is a new reference every time. You'll need to memoize `parentPath` so it has a stable object reference.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+import parentOf from '../util/parentOf'
+
+/**
+ * Helper function to collect nodes with the specified `data-parent-path`.
+ * If no thought node is found, it continues searching with `data-grand-parent-path`.
+ */
+const collectNodes = (path: string): HTMLElement[] => {
+  /** Collect nodes by provided attribute and value. */
+  const collectByAttribute = (attribute: string, value: string): HTMLElement[] => {
+    return Array.from(document.querySelectorAll<HTMLElement>(`[data-${attribute}='${value}']`)).filter(
```
raineorshine, line 13: Let's add a `tree-node` class name to the LayoutTree node to improve the performance of this attribute selector.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+      node => !node.querySelector('.divider'),
```
raineorshine, line 14: Since this function is performance-sensitive (it is called on every keystroke for each divider), we should avoid the nested querySelector.

I recommend adding a `.divider-thought` class to the `.thought` element in [StaticThought](https://github.com/cybersemics/em/blob/d1781ab3fc3db39a64ab15caba2e8625a5c3bead/src/components/StaticThought.tsx#L160). Then you can easily exclude the divider in the `querySelectorAll`.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+    )
+  }
+
+  let collectedNodes = collectByAttribute('parent-path', path)
+
+  // If no nodes found, fallback to data-grand-parent-path
+  if (!collectedNodes.length) {
```
raineorshine, line 21: The only child status is already available in React, so we could avoid the extra DOM query.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
+import { useCallback, useEffect, useMemo, useState } from 'react'
+import Path from '../@types/Path'
+import editingValueStore from '../stores/editingValue'
+import parentOf from '../util/parentOf'
+
+/**
+ * Helper function to collect nodes with the specified `data-parent-path`.
+ * If no thought node is found, it continues searching with `data-grand-parent-path`.
+ */
+const collectNodes = (path: string): HTMLElement[] => {
```
raineorshine, line 10: Let's type this as `path: Path`. No need to encode it prematurely.

raineorshine (changes requested):
NOTE: I updated `table-devan` with some essential changes and rebased this branch, so make sure to get the updated history at 5e63ce85ac.

Can confirm that the divider width updates in realtime during editing.

A couple cases are not updating in realtime:

# Case 1

Paste a longer thought as a cousin of the divider.

Col2, only-child divider's width is not updated.



# Case 2

Delete the longest thought.

Col2, only-child divider width does not update.

Brave Browser 2024-07-05 15 28 43

src/hooks/useMaxSiblingWidth.ts
iamplugged: Not sure if I understood this correctly. Don't really have a way to figure out if it is the only child due to the structure of all thoughts being flat. Let me know if I am missing something here.

raineorshine: The thoughts are flat only after they have been linearized for rendering. If you go up the component hierarchy to [LayoutTree](https://github.com/cybersemics/em/blob/5ffafd3a22beb062ff2b55b87115fe898aea7870/src/components/LayoutTree.tsx), you have access to hierarchical data in `linearizeTree`. You can make the check there, and then pass it down as a prop to the Divider.

All components also have access to hierarchical thought data through various Redux selectors. It's possible in the Divider to create a selector that uses [findAnyChild](https://github.com/cybersemics/em/blob/5ffafd3a22beb062ff2b55b87115fe898aea7870/src/selectors/getChildren.ts#L172) on `parentOf(path)` to see if any non-divider sibling exists. Let me know if you need additional explanation.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,63 @@
...
+  return widths.length > 0 ? Math.max(...widths) : 0
+}
+
+/**
+ * Hook to calculate the width of a divider, equal to the max width of a sibling
+ * at a particular `data-parent-path`.
+ */
+const useMaxSiblingWidth = (parentPath: Path): number => {
+  const [maxWidth, setMaxWidth] = useState(0)
+  const path = useMemo(() => parentPath, [parentPath])
```
raineorshine, line 49: This memo is not doing anything. If the `parentPath` reference is stable, then `path` will be stable. If the `parentPath` reference is unstable, then `path` will be unstable.

So the first question is, is the `parentPath` reference stable? If so, then you can use it directly as a memo dependency. If it's not stable, you should stabilize it in the caller.

src/shortcuts/delete.ts
```diff
@@ -31,6 +32,7 @@ const exec: Shortcut['exec'] = (dispatch, getState, e) => {
   } else if (findDescendant(state, head(cursor), '=readonly')) {
     dispatch(error({ value: `"${ellipsize(value)}" is read-only and cannot be deleted.` }))
   } else {
+    editingValueStore.update(null)
```
raineorshine, line 35: When a thought is deleted, the cursor moves to the next thought. Unless it is the last thought in the thoughtspace, there is never a point where the editingValue becomes `null`, so this is not correct.

It is triggering a re-render, which fixes the observed issue, but through incorrect means. Setting an invalid editingValue could leads to regressions in other parts of the app that expect the editingValue to reflect the value of the cursor thought that is being edited.

raineorshine (changes requested):
Thanks. Some additional comments below.

As for the paste and delete, it does fix the rendering issue, but only by triggering a change in `editingValueStore` with an incorrect value (see comments). I think what we have here is more of a timing issue than an issue of the `editingValueStore` being incorrect and needing to be updated. After all, the `editingValueStore` *is* being set to the correct value on both paste and delete in the upstream branch. It just seems to be triggering the measurement too early. If a small delay is introduced with a timeout, it indeed gets measured correctly.

We ran into a similar issue in [useMultiline](https://github.com/cybersemics/em/blob/5ffafd3a22beb062ff2b55b87115fe898aea7870/src/components/Editable/useMultiline.ts) which measures the DOM. We had to move some of the effects to `useLayoutEffect` and ensure that the measurement also occurs after the cursor changes. Maybe a similar approach would work here. The key is to trigger the measurement after the cursor has been updated and rendered, but before the next paint.

Try moving the `editingValueStore.update` call in [Editable](https://github.com/cybersemics/em/blob/5ffafd3a22beb062ff2b55b87115fe898aea7870/src/components/Editable.tsx#L163) component into a `useLayoutEffect`.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/**
+ * Hook to calculate the width of a divider, equal to the max width of a sibling
+ * at a particular `data-parent-path`.
+ */
+const useMaxSiblingWidth = (parentPath: Path): number => {
+  const [maxWidth, setMaxWidth] = useState(0)
+  const hasNonDividerSiblings = useSelector(state =>
+    findAnyChild(state, head(parentPath), child => child.value !== '---'),
```
raineorshine, line 52: You can reuse [isDivider](https://github.com/cybersemics/em/blob/e7da6e07a42ca2156a0d6fbf8836e10d858870bb/src/util/isDivider.ts) to check if it is a divider.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+  }, 0)
+}
+
+/**
+ * Hook to calculate the width of a divider, equal to the max width of a sibling
+ * at a particular `data-parent-path`.
+ */
+const useMaxSiblingWidth = (parentPath: Path): number => {
+  const [maxWidth, setMaxWidth] = useState(0)
+  const hasNonDividerSiblings = useSelector(state =>
```
raineorshine, line 51: The name `hasNonDividerSiblings` suggests that it is a boolean, and it is only used as a boolean, so you should coerce it as a boolean here at the source.

A good tip to remember: Type _early_ and format _late_.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+    findAnyChild(state, head(parentPath), child => child.value !== '---'),
+  )
+
+  const updateMaxWidth = useCallback(() => {
+    const nodes = collectNodes(parentPath, !!hasNonDividerSiblings)
+    setMaxWidth(calculateMaxWidth(nodes))
+  }, [parentPath, hasNonDividerSiblings])
+
+  const editingVal = editingValueStore.useSelector()
```
raineorshine, line 60: This will cause the render function to be called on every keystroke, which is probably more than necessary. `updateMaxWidth` can be called outside of the component lifecycle and the component will only re-render if it ends up changing `maxWidth`. Hence the reason for `editingValueStore.useEffect`. Maybe we need to add `editingValueStore.useLayoutEffect`.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  useLayoutEffect(updateMaxWidth, [updateMaxWidth, editingVal])
+  useEffect(updateMaxWidth, [updateMaxWidth, parentPath])
```
raineorshine, line 63: Why both `useEffect` and `useLayoutEffect`?

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+  /** Collect nodes by provided attribute and value. */
+  const collectByAttribute = (attribute: string, value: string): HTMLElement[] =>
+    Array.from(document.querySelectorAll<HTMLElement>(`.tree-node[data-${attribute}='${value}']:not(.divider-thought)`))
+
+  if (hasNonDividerSiblings) {
+    collectedNodes = collectByAttribute('parent-path', hashPath(path))
+  } else {
+    const grandparentPathArray = parentOf(path)
+    // Ensure grandParentPathArray is a non-empty array
+    if (Array.isArray(grandparentPathArray) && grandparentPathArray.length) {
```
raineorshine, line 25: The runtime check of `isArray` is unnecessary since the type system guarantees that `grandparentPathArray` is an array.

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Helper function to collect nodes with the specified `data-parent-path`.
+ * If no thought node is found, it continues searching with `data-grandparent-path`.
+ */
+const collectNodes = (path: Path, hasNonDividerSiblings: boolean): HTMLElement[] => {
+  let collectedNodes: HTMLElement[] = []
+  /** Collect nodes by provided attribute and value. */
+  const collectByAttribute = (attribute: string, value: string): HTMLElement[] =>
+    Array.from(document.querySelectorAll<HTMLElement>(`.tree-node[data-${attribute}='${value}']:not(.divider-thought)`))
```
raineorshine, line 18: Good, this approach is more efficient since querySelectorAll is only called once.

raineorshine (changes requested):

src/hooks/useMaxSiblingWidth.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Hook to calculate the width of a divider, equal to the max width of a sibling
+ * at a particular `data-parent-path`.
+ */
+const useMaxSiblingWidth = (parentPath: Path): number => {
+  const [maxWidth, setMaxWidth] = useState(0)
+  const hasNonDividerSiblings = !!useSelector(state =>
+    findAnyChild(state, head(parentPath), child => !isDivider(child.value)),
+  )
```
raineorshine, line 54: Move the `!!` inside the selector, otherwise the component will re-render every time `findAnyChild` returns a different thought.

raineorshine (changes requested):

src/hooks/useMaxSiblingWidth.ts
iamplugged: Tried adding this to react-ministore.ts

`const useChangeLayoutEffect = (cb: (state: T) => void) => useLayoutEffect(() => store.subscribe(cb), [cb])`

This does not work for the paste and delete cases

raineorshine: Okay. Let me know if you have any other ideas of how to do this without the render function getting called on every keystroke.

iamplugged: `editingValueStore.useEffect(updateMaxWidth)`

@raineorshine For the line above, I think the callback gets called before render. So the calculations don't work until the next cycle. Same for useLayoutEffect.

Some how need to run the logic after the render, because the logic is dependent on calculations from the actual DOM and not the virtual DOM

raineorshine: > @raineorshine For the line above, I think the callback gets called before render. So the calculations don't work until the next cycle. Same for useLayoutEffect.

Yes, that's what I have observed as well unfortunately.

PR #2105, opened 2024-07-01
Fix #543; divider width should match largest sibling

### LINKED ISSUE #543: Fix Divider rendering in table

mathiscode:
This PR fixes #543, causing the divider to now be the same width as the largest sibling.


mathiscode:
This PR has been closed and replaced by #2108

Issue #1681, opened 2023-01-13
Drag-and-drop dead zone between thoughts


raineorshine:
## Steps to Reproduce

Drag a thought slowly across a context with several thoughts. 

## Current Behavior

When dragging directly over a thought, there is a small zone where the drop-hover disappears.  

https://user-images.githubusercontent.com/750276/212223261-20884f28-07cc-4184-abf6-693ad4ceb678.MOV

## Expected Behavior

There should be no gaps between the drop zones.

Test across large and small font sizes.

raineorshine:
This occurs despite there being no visible gap when `simulateDrop` is enabled:

<img width="499" alt="image" src="https://github.com/cybersemics/em/assets/750276/f4e7f63a-e388-4ebd-81f4-0040ffc98f06">

PR #2106, opened 2024-07-02
Fixed drag and drop drop dead zone issue

Zubair286:
#1681 

This issue was occuring because a margin top was assigned to the drag source/target div, which made the drop hover disappear because margin area is not part of the drop target. The margin top is removed, with no visible affect.


src/components/Thought.tsx
```diff
@@ -354,7 +354,7 @@ const ThoughtContainer = ({
           // so that .thought can be sized at 100% and .thought .bullet-cursor-overlay bullet can be positioned correctly.
           position: 'relative',
           // match Editable padding
-          marginTop: '0.501em',
+          // marginTop: '0.501em',
```
raineorshine, line 357: You can remove the line of code instead of commenting it out. Commented out code should only be kept in the codebase if it needs to be included for comparison or to make debugging easier.

Please make sure to remove associated comments as well.

raineorshine (changes requested):
Definitely better. Still getting some dead space near the bullet:
https://github.com/cybersemics/em/assets/750276/b1e4aea8-7952-4c12-a6d8-1c45325cad06

Note: Puppeteer tests are failing.

> This issue was occuring because a margin top was assigned to the drag source/target div, which made the drop hover disappear because margin area is not part of the drop target.

Does that mean our drop target color blocks are incorrect in that case? They are supposed to show exactly where the drop target can be dropped on.

src/components/Thought.tsx
Zubair286: Yes, maybe we can increase the drop target width to take the bullet and padding in as well. As for puppeteer test, not sure why they are failing since only macos ones run right, I can't really change them on my machine, so they should be the updated one after rebase.

raineorshine: Looks like they're failing because there is a visual regression. It appears that thoughts are being rendered `0.501em` above where they should be. Maybe you can add that `0.501em` back on the content's top margin to avoid the regression.

---

While the CI only runs snapshot tests on macos, we keep both macos and ubuntu snapshots now, so you should be able to run them locally to see if any break. If any ubuntu snapshots are missing, you'll need to generate them from `main` and then run them against your branch. I hope to have a more automated solution in place in the future.

Note that you can download and view the test artifacts from the GitHub Actions run, allowing you to view exactly how the macos snapshots have failed. They can be found in the Puppeteer test output all the way at the bottom:
image

Zubair286: Hmm, I am not sure about the marginTop since it was the reason for the dead zone. I have tried setting the margin with a smaller value but dead zone still appears. Maybe there is something else that can be done without adding the margin top, any suggestions?

raineorshine: I'm guessing you can just add the same amount of margin at the top of #content since all the thoughts just got shifted up.

Zubair286: Okay, let me see if that works, if not maybe giving padding bottom to thoughts could work, since I think padding will not cause a dead zone.

src/components/Content.tsx
```diff
@@ -88,6 +88,7 @@ const Content: FC = () => {
         id='content'
         ref={contentRef}
         className={contentClassNames}
+        style={{ marginTop: rootThoughtsLength > 0 ? '0.501em' : '0px' }}
```
raineorshine, line 91: Let's put this in [LayoutTree](https://github.com/cybersemics/em/blob/ac0f5c773f1b6e3312baa9b000d563aa09b2db59/src/components/LayoutTree.tsx) so that we don't need to have the conditional.

src/components/Thought.tsx
```diff
@@ -426,6 +424,8 @@ const ThoughtContainer = ({
               publish={publish}
               simplePath={simplePath}
               thoughtId={thoughtId}
+              // debugIndex={debugIndex}
+              // depth={depth}
```
raineorshine, line 428: Please never comment out code unless it is needed for expository purposes. If it is not necessary, remove it completely.

src/components/Bullet.tsx
```diff
@@ -258,13 +265,13 @@ const Bullet = ({
...
         marginLeft: (fontSize - 9) * 0.5 - 11 - extendClickWidth - (isTableCol1 ? fontSize / 4 : 0),
-        marginBottom: -extendClickHeight - 2,
+        marginBottom: extendClickHeight - 2,
         paddingTop: extendClickHeight,
         paddingLeft: extendClickWidth,
         paddingBottom: extendClickHeight + 2,
         position: 'absolute',
         verticalAlign: 'top',
-        width: 4, // make the bullet wide enough to be clicked, but not enough to encroach on the editable
+        width: 7, // make the bullet wide enough to be clicked, but not enough to encroach on the editable
```
raineorshine, line 274: This width is a legacy of the previous layout engine that used static positioning. Now that the Bullet is absolutely positioned, changing the width will not affect the layout. There is no risk of it encroaching on the editable.

I'm thinking we set the width so that the bullet is flush with the editable. That is, `-marginLeft - paddingLeft`.

src/components/Bullet.tsx
```diff
@@ -258,13 +265,13 @@ const Bullet = ({
         // calculate position of thought for different font sizes
         // Table column 1 needs more space between the bullet and thought for some reason
         marginLeft: (fontSize - 9) * 0.5 - 11 - extendClickWidth - (isTableCol1 ? fontSize / 4 : 0),
-        marginBottom: -extendClickHeight - 2,
+        marginBottom: extendClickHeight - 2,
```
raineorshine, line 268: Since the Bullet is now absolutely positioned, do we need `marginBottom` at all?

Regardless, it does not make sense to reverse the sign here: it is meant to exactly compensate for the `paddingBottom`, i.e. increase the clickable area without increasing the height. Perhaps something else necessitates a change to the margin, but `extendClickHeight` should only be used if it's compensating for paddingBottom.

raineorshine (changes requested):
I can confirm there is no more dead zone. Thanks!

src/components/Bullet.tsx
Zubair286: I think this is unintentional, sorry probably messed with this while undoing the color blocks for bullet

src/components/Thought.tsx
Zubair286: I had planned to use these for color blocks of bullet so I commented it out, I wouldn't have left them otherwise. But if you want me to remove them I will, will just add the code back when adding the color blocks.

raineorshine: Got it, sorry about that. I didn't realize it was intentional. Just leave a note for me so that I know next time.

src/components/Bullet.tsx
```diff
@@ -264,7 +271,7 @@ const Bullet = ({
         paddingBottom: extendClickHeight + 2,
         position: 'absolute',
         verticalAlign: 'top',
-        width: 4, // make the bullet wide enough to be clicked, but not enough to encroach on the editable
+        width: -((fontSize - 9) * 0.5 - 11 - extendClickWidth - (isTableCol1 ? fontSize / 4 : 0)) - extendClickWidth, // make the bullet wide enough to be clicked, but not enough to encroach on the editable
```
raineorshine, line 274: Please factor out the common calculations between `marginLeft` and `width` rather than duplicating code. Duplicating code like this is redundant, unperformant, hard to read, and fragile. It gives the illusion that they are independent, when they are actually tightly coupled. Changing one without changing the other will cause a break, and vice versa.

raineorshine: Please remove the comment, as it is no longer relevant.

raineorshine (changes requested):

src/components/Bullet.tsx
Zubair286: Right, sure sorry about that it skipped my mind, I have fixed this now.

raineorshine (approved):

PR #2108, opened 2024-07-03
Fix #543; divider width should match largest sibling

### LINKED ISSUE #543: Fix Divider rendering in table

mathiscode:
This PR fixes #543 causing the divider to now be the same width as the largest sibling.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -2395,7 +2395,7 @@ code,
 }
 
 .child-divider {
-  margin-left: -125px;
+  margin-left: -16px;
 }
 
 .child-divider .drop-hover {
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -3,12 +3,11 @@ import React, { useEffect, useState } from 'react'
 import { useDispatch } from 'react-redux'
 import Path from '../@types/Path'
 import { setCursorActionCreator as setCursor } from '../actions/setCursor'
-import rootedParentOf from '../selectors/rootedParentOf'
-import store from '../stores/app'
 import editingValueStore from '../stores/editingValue'
 import fastClick from '../util/fastClick'
 import hashPath from '../util/hashPath'
 import head from '../util/head'
+import parentOf from '../util/parentOf'
 
 const DIVIDER_PLUS_PX = 20
 const DIVIDER_MIN_WIDTH = 85
@@ -28,8 +27,7 @@ const Divider = ({ path }: { path: Path }) => {
   /** Get the max width of nearby elements, add DIVIDER_PLUS_PX and set this width for divider. */
   const setStyle = () => {
     if (dividerRef.current) {
-      const state = store.getState()
-      const parentPath = rootedParentOf(state, path)
+      const parentPath = parentOf(path)
 
       const treeNode = dividerRef.current.closest('div.tree-node') as HTMLElement
       if (!treeNode) throw new Error('Divider tree node not found')
@@ -41,7 +39,7 @@ const Divider = ({ path }: { path: Path }) => {
 
       if (elements.length === 1) {
         /** If this divider is an only child in a table, find the widest col2. */
-        const grandparentPath = rootedParentOf(state, parentPath)
+        const grandparentPath = parentOf(parentPath)
         if (!grandparentPath) throw new Error('Divider grandparent not found')
 
         elements = document.querySelectorAll(
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -61,7 +61,10 @@ const Divider = ({ path }: { path: Path }) => {
     }
   }
 
-  useEffect(setStyle, [path, thoughtIndex]) // eslint-disable-line react-hooks/exhaustive-deps
+  // Suppress the exhaustive-deps linter rule so that we can ignore dividerRef. Refs do not work in useEffect dependencies, as mutations do not trigger a re-render.
+  // We could use a callback ref to trigger setStyle every time the DOM node changes, but subscribing to the thoughtIndex (which is included for other thought updates) seems to cover all cases.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(setStyle, [path, thoughtIndex])
   editingValueStore.useEffect(setStyle)
 
   return (
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -36,9 +36,7 @@ const Divider = ({ path }: { path: Path }) => {
       const thoughtPath = treeNode.dataset.path?.split(',')
       if (!thoughtPath) throw new Error('Divider path not found on tree node')
 
-      let elements = document.querySelectorAll(
-        `.tree-node[data-path^="${parentPath}"]:not(.thought-divider)`,
-      ) as NodeListOf<HTMLElement>
+      let elements = document.querySelectorAll(`.tree-node[data-path^="${parentPath}"]:not(.thought-divider)`)
 
       if (elements.length === 1) {
         /** If this divider is an only child in a table, find the widest col2. */
@@ -47,7 +45,7 @@ const Divider = ({ path }: { path: Path }) => {
 
         elements = document.querySelectorAll(
           `.tree-node.table-col2[data-path^="${grandparentPath}"]:not(.thought-divider)`,
-        ) as NodeListOf<HTMLElement>
+        )
       } else if (!elements.length) {
         /** This is a top-level divider, use its siblings instead. */
         elements = document.querySelectorAll('.tree-node.top-level:not(.thought-divider)')
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -52,12 +52,10 @@ const Divider = ({ path }: { path: Path }) => {
 
       /** Find and set the max width of our selected elements. */
       const maxWidth = Math.max(
-        ...Array.from(elements).map(
-          element => (element.querySelector('.thought') as HTMLElement).offsetWidth + DIVIDER_PLUS_PX,
-        ),
+        ...Array.from(elements).map(element => (element.querySelector('.thought') as HTMLElement).offsetWidth),
       )
 
-      setWidth(maxWidth)
+      setWidth(maxWidth + DIVIDER_PLUS_PX)
     }
   }
 
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -61,7 +61,7 @@ const Divider = ({ path }: { path: Path }) => {
     }
   }
 
-  useEffect(setStyle, [dividerRef, path, thoughtIndex])
+  useEffect(setStyle, [path, thoughtIndex]) // eslint-disable-line react-hooks/exhaustive-deps
   editingValueStore.useEffect(setStyle)
 
   return (
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -36,8 +36,8 @@ const Divider = ({ path }: { path: Path }) => {
         ...siblingNodes
           .filter(node => !node.querySelector('.divider'))
           .map(node => {
-            const subs = node.querySelector('.thought') as HTMLElement
-            return subs ? subs.offsetWidth + DIVIDER_PLUS_PX : DIVIDER_PLUS_PX
+            const subthought = node.querySelector('.thought') as HTMLElement
+            return subthought ? subthought.offsetWidth + DIVIDER_PLUS_PX : DIVIDER_PLUS_PX
           }),
       )
 
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -61,7 +61,7 @@ const Divider = ({ path }: { path: Path }) => {
     }
   }
 
-  useEffect(setStyle, [setStyle, thoughtIndex])
+  useEffect(setStyle, [dividerRef, path, thoughtIndex])
   editingValueStore.useEffect(setStyle)
 
   return (
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -3,9 +3,11 @@ import React, { useEffect, useState } from 'react'
 import { useDispatch } from 'react-redux'
 import Path from '../@types/Path'
 import { setCursorActionCreator as setCursor } from '../actions/setCursor'
-import { DIVIDER_MIN_WIDTH, DIVIDER_PLUS_PX } from '../constants'
 import fastClick from '../util/fastClick'
 import head from '../util/head'
+
+const DIVIDER_PLUS_PX = 20
+const DIVIDER_MIN_WIDTH = 85
 
 /** A custom horizontal rule. */
 const Divider = ({ path }: { path: Path }) => {
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -64,7 +64,7 @@ const Divider = ({ path }: { path: Path }) => {
   }
 
   useEffect(setStyle, []) // eslint-disable-line react-hooks/exhaustive-deps
-  editingValueStore.useEffect(setStyle)
+  editingValueStore.useEffect(() => setTimeout(setStyle, 0))
 
   return (
     <div
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -1,10 +1,10 @@
 import classNames from 'classnames'
 import React, { useEffect, useState } from 'react'
-import { useDispatch, useSelector } from 'react-redux'
+import { useDispatch } from 'react-redux'
 import Path from '../@types/Path'
-import State from '../@types/State'
 import { setCursorActionCreator as setCursor } from '../actions/setCursor'
 import rootedParentOf from '../selectors/rootedParentOf'
+import store from '../stores/app'
 import editingValueStore from '../stores/editingValue'
 import fastClick from '../util/fastClick'
 import head from '../util/head'
@@ -16,7 +16,6 @@ const DIVIDER_MIN_WIDTH = 85
 const Divider = ({ path }: { path: Path }) => {
   const dividerRef = React.createRef<HTMLInputElement>()
   const dispatch = useDispatch()
-  const state = useSelector((state: State) => state)
   const [width, setWidth] = useState(DIVIDER_MIN_WIDTH)
 
   /** Sets the cursor to the divider. */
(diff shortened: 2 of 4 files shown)


src/components/Divider.tsx
```diff
@@ -9,27 +9,36 @@ import head from '../util/head'
...
+      const parentNode = dividerRef.current.closest('div.tree-node')
+
+      /** Get the depth of this thought's node. */
+      const parentDepth = parentNode ? parseInt(parentNode.getAttribute('data-depth') || '0') : 0
+      const siblingNodes = Array.from(document.querySelectorAll(`.tree-node[data-depth="${parentDepth}"]`))
+
+      /** Find the largest child node and set the width of the divider to its width plus DIVIDER_PLUS_PX. */
+      const maxWidth = Math.max(
+        ...siblingNodes
+          .filter(node => !node.querySelector('.divider'))
```
raineorshine, line 34: Because this function will be called on every keystroke, I'd like to do a small performance optimization.

Instead of filtering the nodes here, let's add a `thought-divider` class to [StaticThought](https://github.com/cybersemics/em/blob/d1781ab3fc3db39a64ab15caba2e8625a5c3bead/src/components/StaticThought.tsx#L160) when it is a divider thought. Then you can exclude the divider in the `.map` below without an additional DOM query.

src/components/Divider.tsx
```diff
@@ -9,27 +9,36 @@ import head from '../util/head'
...
+          .map(node => {
+            const subs = node.querySelector('.thought') as HTMLElement
```
raineorshine, line 36: Let's call this variable `subthought` (singular) since it is a single thought.

raineorshine (changes requested):
Looks good, just seeing a few issues.

1. The width of the divider should be equal to the width of the widest *sibling*, not cousins.

Here, the first divider should only be as long as `b2`, not `This is a longer thought`:

image

Instead of using `data-depth`, I would suggest storing the parent's `Path`. That way you can uniquely identify all children of the divider's parent.

2. The divider width should update dynamically as the user types. I suggest subscribing to the [editingValueStore](https://github.com/cybersemics/em/blob/472472069b4637a88c9bbb14f2ce6918ad742b1e/src/stores/editingValue.ts).

3. There is one other case that needs to be taken into account. When the divider is the only child in the second column of a table view, it should be set to the max width of all siblings *and cousins* (other col2 thoughts in the same table but a different row). (This was documented in the original issue [here](https://github.com/cybersemics/em/issues/543#issuecomment-613511024); sorry that wasn't clear from the beginning). Technically this case is what the screenshot in the original issue depicts. I will update it so it's clearer.

src/App.css
```diff
@@ -1232,6 +1232,19 @@ h1 .num-contexts {
...
+.placeholder.inverse,
+.thought.inverse [placeholder]:empty::before {
+  color: rgba(255, 255, 255, 0.5);
+}
+.dark .text-note.inverse,
+.dark .placeholder.inverse,
+.dark .thought.inverse [placeholder]:empty::before {
+  color: rgba(7, 7, 7, 0.5);
+}
+
```
raineorshine, line 1247: This is incorrectly showing up in the PR diff for some reason. It appears to be coming from f034a27d60605c0668aefd934eb91099d92f08bf, which is c49618e5478f28cc22fb12161188c54c7f3b7b1d on `main`. It is not in the`table-jay` commit history, so maybe you accidentally branched off `main` instead of `table-jay`?

raineorshine (changes requested):
Hi Jay. I'm noticing there are unrelated changes showing up in the PR diff for some reason. For example, the changes from f034a27d60605c0668aefd934eb91099d92f08bf, which is c49618e5478f28cc22fb12161188c54c7f3b7b1d on `main`, are showing up in the diff for `App.css`. This commit is not in the`table-jay` commit history, so maybe you accidentally branched off `main` instead of `table-jay`?

Once the diff accurately reflects only the changes for this PR I can do a full review.

raineorshine:
I should also mention that I am seeing this warning in the browser console that should not be there:

Brave Browser 2024-07-07 11 24 31

src/components/Divider.tsx
```diff
@@ -1,50 +1,81 @@
...
+      )
+
+      setWidth(maxWidth)
     }
   }
 
-  useEffect(setStyle)
+  useEffect(setStyle, []) // eslint-disable-line react-hooks/exhaustive-deps
+  const unsubscribe = editingValueStore.subscribe(setStyle)
+  useEffect(() => () => unsubscribe(), [unsubscribe])
```
raineorshine, line 67: You can use `editingValueStore.useEffect` provided by the [react-ministore](https://github.com/cybersemics/em/blob/98dff32074ed855a4e9348a2bc00c20cbf3fa42c/src/stores/react-ministore.ts) mixin.

raineorshine (changes requested):

src/App.css
mathiscode: Oops, sorry about that! I wasn't confused about the need for a non-main branch, but what happened is I was originally going to make changes to `StaticThought.tsx` that imported classnames and changed the className prop of the div to use it instead of the string 'thought', and I saw that you had already made it use a classNames object in that commit so I just cherry-picked it to get my working state of that file as close to upstream as possible to work with, not realizing it would affect your review.

I didn't mean to mess up the diff, but I've rebased and force pushed so we should be in a good state now. I'll resolve the other changes below next!

mathiscode:
> I should also mention that I am seeing this warning in the browser console that should not be there:
> 
> Brave Browser 2024-07-07 11 24 31

This should now be fixed with 4e22b98

raineorshine (changes requested):
Diff looks good now, thanks!

A couple cases where the divider width is not updated. I confirmed that `editingValueStore` is updated in these cases, so maybe the effect is firing before React has re-rendered. This would cause `setStyle` to measure the DOM before it has been updated.

# Case 1

Paste a longer thought as a cousin of the divider.

Col2, only-child divider's width is not updated.



# Case 2

Delete the longest thought.

Col2, only-child divider width does not update.

src/components/LayoutTree.tsx
```diff
@@ -1,3 +1,4 @@
+import classnames from 'classnames'
```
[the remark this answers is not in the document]
mathiscode: Oops, my mistake - I had this correct originally and then mistakenly thought I saw it in lowercase somewhere else in the code but didn't verify before that last commit. Fixed now!

src/components/Divider.tsx
```diff
@@ -1,50 +1,81 @@
...
-      const children = parentUl ? (Array.from(parentUl.childNodes) as HTMLElement[]) : []
-      const widths = children.map((child: HTMLElement) => {
-        if (child.classList.contains('child-divider')) return DIVIDER_PLUS_PX
-        const subs = child.getElementsByClassName('subthought') as HTMLCollectionOf<HTMLElement>
-        return subs.length ? subs[0].offsetWidth + DIVIDER_PLUS_PX : DIVIDER_PLUS_PX
-      })
-      const maxWidth = Math.max(...widths)
-      dividerSetWidth.current.style.width = `${maxWidth > DIVIDER_MIN_WIDTH ? maxWidth : DIVIDER_MIN_WIDTH}px`
+    if (dividerRef.current) {
+      const state = store.getState()
```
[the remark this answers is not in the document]
mathiscode: Good catch, thanks!

src/components/Divider.tsx
```diff
@@ -1,50 +1,81 @@
...
+
+      const treeNode = dividerRef.current.closest('div.tree-node') as HTMLElement
+      if (!treeNode) throw new Error('Divider tree node not found')
+
+      const thoughtPath = treeNode.dataset.path
+      if (!thoughtPath || thoughtPath === '') throw new Error('Divider path not found on tree node')
+
+      let elements = document.querySelectorAll(`.tree-node[data-path^="${hashPath(parentPath)}"]:not(.thought-divider)`)
+
+      if (elements.length === 1) {
```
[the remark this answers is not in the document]
mathiscode: Oh sorry about that. My thought was that these classes would be helpful in differentiating the tree-nodes in other cases as well, but I see your point and I'll try to get this refactored ASAP.

src/components/Divider.tsx
```diff
@@ -1,50 +1,81 @@
...
+      const state = store.getState()
+      const parentPath = rootedParentOf(state, path)
+
+      const treeNode = dividerRef.current.closest('div.tree-node') as HTMLElement
+      if (!treeNode) throw new Error('Divider tree node not found')
+
+      const thoughtPath = treeNode.dataset.path
+      if (!thoughtPath || thoughtPath === '') throw new Error('Divider path not found on tree node')
+
+      let elements = document.querySelectorAll(`.tree-node[data-path^="${hashPath(parentPath)}"]:not(.thought-divider)`)
```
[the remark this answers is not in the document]
mathiscode: First we find all tree-nodes with a path starting with the divider's parent which in a table should initially include children in both columns, but exclude any dividers so they aren't used in the width calculation.

When we do this, if elements.length is:

0: The parentPath of a root/top-level divider is empty, so our query doesn't find any elements to compare - we find top-level nodes instead to calculate width in this case `(!elements.length)`

1: Our original query gives us only one other element at this path, meaning this is likely a single divider and we're only finding its col1 counterpart, so we go up another level and use those elements `(elements === 1)`

If elements.length > 1, then we have siblings to check.

A lot of this doesn't really matter now since I'll be doing a more React-centric refactor, but let me know if I misunderstood anything.

src/components/Divider.tsx
```diff
@@ -1,50 +1,81 @@
...
+        const grandparentPath = rootedParentOf(state, parentPath)
+        if (!grandparentPath) throw new Error('Divider grandparent not found')
+
+        elements = document.querySelectorAll(
+          `.tree-node.table-col2[data-path^="${hashPath(grandparentPath)}"]:not(.thought-divider)`,
+        )
+      } else if (!elements.length) {
+        /** This is a top-level divider, use its siblings instead. */
+        elements = document.querySelectorAll('.tree-node.top-level:not(.thought-divider)')
+      }
```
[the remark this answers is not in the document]
mathiscode: There's a little more detail [here](https://github.com/cybersemics/em/pull/2108#discussion_r1671843788), but the parentPath of a root/top-level divider is empty, so our query doesn't find any elements to compare - we find top-level nodes instead to calculate width in this case. This will likely change in the rewrite.

src/components/Divider.tsx
```diff
@@ -1,50 +1,81 @@
...
+        ),
+      )
+
+      setWidth(maxWidth)
     }
   }
 
-  useEffect(setStyle)
+  useEffect(setStyle, []) // eslint-disable-line react-hooks/exhaustive-deps
+  editingValueStore.useEffect(setStyle)
```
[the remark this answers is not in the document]
mathiscode: I did this because we need `setStyle` to be called on the initial render before `editingValueStore` takes over, but providing the expected dependencies causes a loop and it gets called repeatedly. I've created an experimental branch to show this in action:

https://github.com/mathiscode/em/tree/experiment/543-divider-width-exhaustive-deps

I'm definitely open to any suggestions for a more optimal solution, but this does seem to solve the immediate problem.

mathiscode:
Pasting/deleting should now be fixed with 01a5b25. I'll work on replacing the current approach with a more React-centric one.

raineorshine:
Thanks, confirmed. 

Using `setTimeout` works, although it's not totally ideal since it may trigger a second paint on the second render. We [recently switched to `useLayoutEffect`](https://github.com/cybersemics/em/commit/98dff32074ed855a4e9348a2bc00c20cbf3fa42c) in a hook that measures the DOM in order to re-render before the first paint. I played around with it and the only other thing I could get to work was a `useSelector` on the cursor value (which changes on paste and delete) + `useLayoutEffect`. I'm not sure if it's any better for the performance than the `setTimeout`.

src/components/Divider.tsx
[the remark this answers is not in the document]
raineorshine: Ah, I see. I neglected to see the `data-path^=` attribute selector and was mistakenly thinking `data-path=`. That explains why the parent is selected.

src/components/Divider.tsx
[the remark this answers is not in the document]
raineorshine: It looks like `dividerRef` is not a stable reference, even if `dividerRef.current` is. Learned something new! :)

So you should be able to fix it with `[dividerRef.current, path]`.

raineorshine:
I noticed another case where the divider width is not updated: If you set the cursor on the divider or its parent and refresh the page, it assumes the minWidth.

mathiscode:
> I noticed another case where the divider width is not updated: If you set the cursor on the divider or its parent and refresh the page, it assumes the minWidth on load.

Good catch! 8c0df72 should resolve this - it feels a little hacky, we add an additional setTimeout in the case of isCursor, but it seems effective and shouldn't introduce any regressions.

raineorshine:
I'm not really comfortable with a hack here unless we have exhausted all other possibilities. The problem with introducing more timeouts is that it creates more side effects out of the React render cycle, which can lead to more problems in the future. I'm already making an exception for the other `setTimeout`.

We should try to first identify the cause of the problem. Why is the width updated after navigating, but not updated after refresh? What is different about the refresh case? Usually it indicates that there is some other bit of state that needs to be subscribed to. My guess is that the divider is loaded first because it is under the cursor, and there is nothing to make it re-rendered once its siblings load.

mathiscode:
> I'm not really comfortable with a hack here unless we have exhausted all other possibilities. The problem with introducing more timeouts is that it creates more side effects out of the React render cycle, which can lead to more problems in the future. I'm already making an exception for the other `setTimeout`.
> 
> We should try to first identify the cause of the problem. Why is the width updated after navigating, but not updated after refresh? What is different about the refresh case? Usually it indicates that there is some other bit of state that needs to be subscribed to. My guess is that the divider is loaded first because it is under the cursor, and there is nothing to make it re-render once its siblings load.

I'll find another approach and let you know what I come up with.

raineorshine:
That should be fine. We could subscribe to just the thoughts that affect its width (siblings + cousins), but at the cost of selector performance. And there are probably not too many thoughts being rendered that aren't siblings or cousins anyway.

src/components/Divider.tsx
```diff
@@ -1,50 +1,79 @@
...
+      const maxWidth = Math.max(
+        ...Array.from(elements).map(element => (element.querySelector('.thought') as HTMLElement).offsetWidth),
+      )
+
+      setWidth(maxWidth + DIVIDER_PLUS_PX)
     }
   }
 
-  useEffect(setStyle)
+  useEffect(setStyle, [setStyle, thoughtIndex])
```
raineorshine, line 64: `setStyle` is not a stable ref, so this will cause `useEffect` to trigger on every render. This is probably masking a missing dependency.

The `thoughtIndex` dependency is good because at least we're getting the additional render that is needed when the thoughts load.

mathiscode: Oops, I'm not even sure why I did that. I pushed the fix, although it didn't affect the number of times setStyle got called, but everything seems kosher now.

raineorshine: Confusingly, the ref container is not actually a stable reference, only the object inside:

```ts
dividerRef // unstable
dividerRef.current // stable
```

So the number of times `setStyle` is called hasn't changed because it's still getting called on every render.

mathiscode: The only issue with that is I get this complaint:

dividerRef_dependency

raineorshine: If `dividerRef` is unstable, and `dividerRef.current` is an invalid dependency, then what do we do? I must be missing something. 

Maybe you can find the answer online. We can't be the first ones to run into this.

mathiscode: Right. I'll let you know what I come up with.

mathiscode: 56171ea - After some trial and error, I think the cleanest way I can come up with is to just bring back the eslint-disable for exhaustive-deps, and at that point we don't really need to rely on the `dividerRef` at all since it was just added to satisfy the eslint warning. Not having the ref as a dependency calls `setStyle` the same number of times as using `dividerRef.current` so we still have stable dependencies and preserve the correct behavior on page load with the divider selected. Does this seem effective enough to you, or maybe I'm missing something else?

raineorshine: Sure, we can suppress the linter for now. There is a chance that the DOM may update and `setStyle` will not be triggered, but I don't have a test case for it.

FYI It looks like the recommended solution is to switch from an object ref to a callback ref, but my experiments didn't show that it made much difference. It will be good to keep these in mind if we run into this issue in a different context though:

- https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom
- https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780
- https://legacy.reactjs.org/docs/hooks-faq.html?source=post_page-----eb7c15198780--------------------------------#how-can-i-measure-a-dom-node

raineorshine: I added a comment in 7e7e4b5c04 so that future developers can understand why we suppressed the linter.

raineorshine (approved):

raineorshine:
I'm going to merge this, but I'll note the limitations of this implementation here:

1. `react-hooks/exhaustive-deps` is suppressed on the useEffect, which *may* result in `setStyle` not being called when the DOM changes (unable to reproduce)
2. Determining if the divider is an only-child could be done in React, but it is done with direct DOM access, which is more fragile.
3. There is a high performance cost to using multiple DOM selectors in `setStyle` which is called on every keystroke. Accumulating refs and passing them through React would obviate the need for DOM selectors.

Issue #1890, opened 2024-05-15
[macOS] Favorites panel opens after drag-and-drop and app switch


raineorshine:
## Setup

Create a thought `a`.

## Steps to Reproduce

1. Refresh the page.
2. Drag and drop the thought into the same place.
3. Switch tabs away and back.

## Current Behavior

https://github.com/cybersemics/em/assets/750276/db1f2c9b-84dd-4c07-ae6e-087cb6cdb664

## Expected Behavior

Favorites panel should not open when switching apps.

Zubair286:
After debugging this issue, it seems on tab change the onmouseup event fires which cause the sidebar to open up for some reason. I have checked it with onclick and onmousedown, these events work fine but onmouseup for some reason fires. Can we use onmousedown event for this in the fastClick handler?

raineorshine:
> After debugging this issue, it seems on tab change the onmouseup event fires which cause the sidebar to open up for some reason and its inconsistent, sometimes does not open sidebar.

If you do a page refresh at the beginning, it should happen every time.

> I have checked it with onclick and onmousedown, these events work fine but onmouseup for some reason fires.

Just to make sure we're talking about the same thing, are you referring to the fastClick handler in the `HamburgerMenu`?

https://github.com/cybersemics/em/blob/ac0f5c773f1b6e3312baa9b000d563aa09b2db59/src/components/HamburgerMenu.tsx#L83-L88

> Can we use onmousedown event for this in the fastClick handler?

No, that would fundamentally change the behavior of `fastClick`. The intention is to provide an instant response to `onmouseup` that avoids the double click delay that is built into `onclick`. As a drop-in replacement for `onclick`, it should reproduce the familiar behavior of activating the button only when the mouse is released. This allows the user to drag away to cancel the click, not to mention long tap or drag-and-drop functionality.

It seems to me that there must be a bug in `fastClick` or in drag-and-drop that is causing the mouseup to be activated on tab change.

Zubair286:
Yes, it is concerning that onmouseup is being fired at all, the drag and drop has something to do with this but it is far from the HamburgerMenu, I will further look into this and see if I can find a solution.

PR #2109, opened 2024-07-05
Fixed sider opening on tab switch issue

Zubair286:
#1890


raineorshine (approved):
How very strange! Seems like a browser bug. Thank you.

Issue #1887, opened 2024-05-15
Table: Incorrect position of DropHover after last col1 thought


raineorshine:
## Steps to Reproduce

```
- a
  - =view
    - Table
  - =pin
    - true
  - b
    - c
  - d
    - e
- x
```

## Current Behavior

When dragging `x` after `d`, the DropHover is not rendered too far to the right.

https://github.com/cybersemics/em/assets/750276/cc709353-2cd2-4792-8e9a-689154a85d72

## Expected Behavior

The DropHover should be rendered below `d`, with its left edge flush.

Issue #1886, opened 2024-05-15
Table: Incorrect position of DropHover after table


raineorshine:
## Steps to Reproduce

```
- x
- a
  - =view
    - Table
  - =pin
    - true
  - b
    - c
  - d
    - e
```

## Current Behavior

DropHover is positioned too far to the right when dragging `x` after ~~`d`~~ `a`.

https://github.com/cybersemics/em/assets/750276/5f90d0f0-6b29-4771-9bb6-adaa9608796a

## Expected Behavior

DropHover should be positioned correctly at the insertion point, left edge flush with `a`.

## Note

This does not occur when `a` has a next sibling.

raineorshine:
> DropHover is positioned too far to the right when dragging `x` after `d`.

My apologies, that should read "`x` after `a`" as depicted in the video.

PR #2125, opened 2024-07-09
fixed drop-hover position on first column last thought in table view

Zubair286:
#1887

### DIFF
--- a/src/@types/VirtualThoughtProps.ts
+++ b/src/@types/VirtualThoughtProps.ts
@@ -17,6 +17,7 @@ interface VirtualThoughtProps {
   nextChildId?: ThoughtId
   simplePath: SimplePath
   showContexts?: boolean
+  isTableCol2?: boolean
 }
 
 export default VirtualThoughtProps
--- a/src/components/DropEnd.tsx
+++ b/src/components/DropEnd.tsx
@@ -33,6 +33,7 @@ const DropEnd = ({
   // renders the component with additional click area below and to the left since there are no thoughts below to obscure
   last,
   path,
+  isTableCol2,
 }: {
   depth: number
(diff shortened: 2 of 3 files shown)


raineorshine (changes requested):
For #1887, the drop hover needs to be positioned with its left edge flush with the left edge of col1, regardless of the length of thoughts in col1. In this example with longer thoughts in col1, the drop hover is too far to the right:

```
- a
  - =view
    - Table
  - =pin
    - true
  - Boston
    - c
  - Denver
    - e
- x
```

Preview 2024-07-09 10 08 19

Please add a snapshot test for each issue as well. Thanks!

Zubair286:
I think I should make a separate PR for #1886 if the solution is good for that.

raineorshine:
Yes, a separate PR for each issue is preferred. 

Please make sure to describe the cause of the problem as well once you identify. That will help determine the appropriateness of the solution.

Zubair286:
One thing I want to ask is that in the table view initially the columns are too far to the left until x is dragged after d, then the columns shifts to right, my guess is that the drop-hover is rendering to the right before the drag, shouldn't the columns already be in the position move to after drag?

raineorshine:
Hmmm good point. That’s a bug in the table view. We may want to wait until that’s fixed to confirm if this is still an issue.

trevinhofmann:
@Zubair286, it appears that this issue has been resolved (by #2175?):

https://github.com/user-attachments/assets/736f52da-01e6-4c7c-8bbe-81aeaf4aaadf

Can this pull request be closed, or does more need to be done?

Zubair286:
Yes, this pull request can be closed.

Issue #1721, opened 2023-06-21
Disable =drop/top when drop destination is expanded


raineorshine:
## Steps to Reproduce

```
- a
  - =drop
    - top
  - b
  - c
  - d
```

Drag `b` after `d`. 

## Current Behavior

Nothing happens. 

## Expected Behavior

`b` should be moved after `d`. 

`=drop/top` should not apply when the parent _of the drop target_ is expanded.

Zubair286:
If show hidden thought is enabled, then dragging b after d would place it above the =drop/top, otherwise nothing happens. In the scenario when show hidden thoughts is enabled, should the =drop/top not apply when parent is expanded?

raineorshine:
`=drop/top` should not apply when the parent is expanded, regardless of whether `showHiddenThoughts` is enabled.

Zubair286:
I am wondering if `=drop/top` should not be applied to an expanded parents children, then the only point I can think of it working would be at _ROOT_ level, because a thought has to be expanded to drag/drop its children. Please let me know if I am wrong.

raineorshine:
The drop target's parent is the one that is expanded, not the drop source. 

Maybe this example is clearer:

``` 
- x
- a
  - =drop
    - top
  - b
  - c
``` 

1. Set the cursor on `a`. 
2. Drag `x` after `c`.

Zubair286:
I see, in the above example dragging x after c puts the x above b because of the `=drop/top`, but if the parent is expanded this prevents x from being dropped after c., so =drop/top should be disabled when parent is expanded, is that right?

raineorshine:
Yes, that’s correct. `=drop/top` should only be active when dropping into a collapsed thought.

Zubair286:
Okay, I was just a little confused with the first example, I get it now. Thanks

Zubair286:
Please close this issue, as it is fixed.

PR #2126, opened 2024-07-09
disabled =drop/top when thought is expanded

Zubair286:
#1721


src/components/DragAndDropSubthoughts.tsx
```diff
@@ -104,7 +105,9 @@ const drop = (props: VirtualThoughtProps, monitor: DropTargetMonitor) => {
   const parentIdFrom = head(rootedParentOf(state, thoughtsFrom))
   const parentIdTo = head(rootedParentOf(state, pathTo))
   const sameContext = parentIdFrom === parentIdTo
-  const dropTop = attributeEquals(state, parentIdTo, '=drop', 'top')
+  const isExpanded = isThoughtExpanded(state, parentIdTo)
```
raineorshine, line 108: Since we already have the `Path`, it would be better to make an `isPathExpanded` function. This avoids converting the `Path` to a `ThoughtId` and then back to a `Path`:

```ts
const isPathExpanded = (state: State, path: Path) =>
  !!state.expanded[hashPath(path)]

const isExpanded = isPathExpanded(rootedParentOf(state, pathTo))
```

raineorshine (changes requested):
Confirmed that the problem is fixed. One optimization requested. Thanks.

raineorshine (changes requested):
Looks good. Please return `isThoughtExpanded` to `fetchDescendants` now that it is no longer used outside of that module.

Note that Redux selectors go in `/src/selectors` rather than `src/util`.

raineorshine (approved):

Issue #2084, opened 2024-06-28
Multiline URL height is incorrect when cursor is set to null


raineorshine:
## Steps to Reproduce

```
- https://open.substack.com/pub/noahpinion/p/interview-sarah-c-paine?r=7kxga&utm_medium=ios&utm_campaign=post
- a
```

Paste and then set the cursor to null.

## Current Behavior

The URL height is too big. 

If the url is ellipsized and only takes up a single-line, the `updateSize` measurement should kick in and update the height to the actual measured height.

<img width="755" src="https://github.com/cybersemics/em/assets/750276/f54440c1-cb15-4142-9df7-08f167152dc9">

## Expected Behavior

<img width="754" src="https://github.com/cybersemics/em/assets/750276/54d58297-6f73-4726-99c1-a131b5e86a64">

PR #2131, opened 2024-07-10
#2084 fix: assign single line height on truncated url line

baproactive:
#2084

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -618,6 +618,10 @@ This prevents table cols being rendered as multiline content, which causes inaac
   box-sizing: border-box;
 }
 
+.thought {
+  --min-thought-height: 1.9em;
+}
+
 .child .editable,
 .editable-annotation-text {
   /* slide editable up so that it overlaps with the previous thought, ensuring there is no dead click zone from clip-path */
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
First, start with describing the cause of the bug. *Why* is the height incorrect? It is hard for me to evaluate if the solution is appropriate without knowing the cause of the bug. 

I am guessing that `useMultiline` is returning true for the multiline url, and stays that way even when the url is ellipsized. If so, wouldn't we want to correct the problem there? Hard coding the height would seem to only mask the problem.

baproactive:
> First, start with describing the cause of the bug. _Why_ is the height incorrect? It is hard for me to evaluate if the solution is appropriate without knowing the cause of the bug.
> 
> I am guessing that `useMultiline` is returning true for the multiline url, and stays that way even when the url is ellipsized. If so, wouldn't we want to correct the problem there? Hard coding the height would seem to only mask the problem.

It's not affected by multiline flag, it's affected because of the display:inline-block applied to .single-line .thought-annotation .editable-annotation-text
browser calculates the height differently between display:inline-block and display: inline. 

For those annotation boxes under .single-line class name, they have display:inline-block to apply the css text sllipsis style.

raineorshine (changes requested):
Got it, thanks. Let's add an inline comment explaining that. 

Where is the other `1.9em` defined? It might be a good idea to define that in one place if they are tightly coupled. Otherwise that relationship is not explicit and it's more likely that one might be changed accidentally.

src/App.css
```diff
@@ -630,6 +630,15 @@ This prevents table cols being rendered as multiline content, which causes inaac
   min-height: 1.9em;
 }
 
+/** Assign annotation height on single line truncated url */
+.single-line .thought {
+  /**
+    .singl-line element has display: inline-block property, it affects the height calculation of the .thougt div so we need to set the height manually.
```
raineorshine, line 636: Good comment, thanks.

src/App.css
```diff
@@ -630,6 +630,15 @@ This prevents table cols being rendered as multiline content, which causes inaac
...
+    This height 1.9em is the same value as the min-height above: line: 630
```
raineorshine, line 637: We should avoid specifying exact line numbers in comments, as they are quickly invalidated if any code is added or removed above.

src/App.css
```diff
@@ -630,6 +630,15 @@ This prevents table cols being rendered as multiline content, which causes inaac
...
+  */
+  height: 1.9em;
```
raineorshine, line 639: We can leverage CSS variables for a more explicit enforcement of using the same value across multiple properties.

raineorshine (approved):

Issue #2103, opened 2024-07-01
Make browser pointer visible in drag-and-drop tests


raineorshine:
Make browser pointer visible in drag-and-drop tests so that we can more easily see the cause of drop target positioning issues.

1. Use install-mouse-helper to simulate the mouse pointer during snapshots: https://github.com/puppeteer/puppeteer/issues/4378#issuecomment-499726973
2. Change the visible cursor from a circle to a pointer cursor for more precision: https://cursor.in/assets/cursor.svg

PR #2135, opened 2024-07-11
Fixed drop-hover position after table

Zubair286:
#1886

The drop-end is being rendered in the last visible thought, so in table view that would the last thought in column 2, the drop-end styling was not taking into account the table view. Added additional styling for drop-end to render on correct position


src/components/LayoutTree.tsx
```diff
@@ -358,6 +358,8 @@ const LayoutTree = () => {
       : 0,
   )
 
+  const tableCol1WidthRef = useRef(0)
```
raineorshine, line 361: Why use a ref? This value should not be preserved across renders.

src/components/LayoutTree.tsx
```diff
@@ -566,6 +568,10 @@ const LayoutTree = () => {
         ycol1Ancestors.push({ y: yaccum + height, depth: node.depth })
       }
 
+      if (node.isTableCol1) {
+        tableCol1WidthRef.current = Number(tableCol1Widths.get(head(parentOf(node.path)))) / fontSize || 0
```
raineorshine, line 572: The `fontSize` division is unnecessary, as we can just specify it in px instead of em in the CSS `calc` below. Less work and less prone to rounding errors.

src/components/LayoutTree.tsx
```diff
@@ -721,14 +727,18 @@ const LayoutTree = () => {
                       const simplePathEnd =
                         -(cliff + i) < simplePath.length ? (simplePath.slice(0, cliff + i) as SimplePath) : HOME_PATH
                       const cliffDepth = unroot(pathEnd).length
+
+                      const isRootPath = isRoot(pathEnd)
```
raineorshine, line 731: We need a general solution that works for all tables, not just tables in the root. 

```
- z
  - x
  - a
    - =view
      - Table
    - =pin
      - true
    - b
      - c
    - d
      - e
```

src/components/LayoutTree.tsx
```diff
@@ -721,14 +727,18 @@ const LayoutTree = () => {
...
+                      const marginLeftOffset =
+                        isRootPath && (isTableCol1 || isTableCol2) ? Math.round(tableCol1WidthRef.current) : 0
```
raineorshine, line 733: Avoid `Math.round`, as px and em can be specified fractionally.

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -215,6 +215,28 @@ const testSuite = () => {
...
+          - =pin
+            - true
+          - b
+            - c
+          - d
+            - e
+      `)
+
+      await clickThought('x')
+      await dragAndDropThought('x', 'd', { position: 'after', mouseUp: true })
```
raineorshine, line 234: There are a couple problems with this test:

1. If you `mouseUp` here, then this doesn't test the original issue. The issue is about the position of the DropHover, which is shown on drag. If you complete the drop, then the DropHover will no longer be visible. The snapshots you added clearly do not show the expected behavior for the original issue.
2. The code says that the thought is being dragged after d [as a sibling], but the original issue involves dragging after a [below d as an uncle]. This is confusing the semantics of `after` which is meant to be used to drop a thought after a sibling. We need the semantics to align here so that what the test says reflects what is happening in the browser. And we will soon need a test that actually drops `x` after `d` as a sibling anyway.

I think you're going to need a new `position` value to reflect the unique drag-and-drop positioning for tables. #2103 is going to come in helpful here, too.

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -215,6 +215,28 @@ const testSuite = () => {
       const image = await screenshot()
       expect(image).toMatchImageSnapshot()
     })
+
+    it('drop hover after table', async () => {
```
raineorshine, line 219: Let's move this test into the `drag` describe block.

raineorshine (changes requested):

src/e2e/puppeteer/__tests__/snapshot.ts
Zubair286: Yes, you are right. The current way the the `after` prop works, does not take into account the children of the target `thought` if it is expanded, since it drags the thought onto the `[x,y] `values of the target thought. We need to implement a new prop that would work in such a case, which includes table view.

src/components/LayoutTree.tsx
Zubair286: Hmm, I think removing the `isRootPath` conditional would work for a more general solution.

src/components/LayoutTree.tsx
Zubair286: The useMemo hook which is where this ref is being assigned runs after the drag, but we need that during the drag process so the margin can be assigned to the drag-end, so I thought of using ref so I could access the width during the drag.

raineorshine: See if you can do this without introducing additional mutable state. If it doesn't seem possible, let me know and I'll do a closer analysis myself. I have a feeling there is a more elegant solution here.

Zubair286:
About the col1 width variable, I couldn't find a way to make it work without using a ref/state variable. Because if col1 width increases, we also need to update the value, so I think it needs to be mutable.

raineorshine:
The `tableCol1Widths` get set and then used within a single render cycle though. So there is no need to store it in a ref from what I can tell. Unless I'm missing something, you're using the ref to shuttle a value from the `treeThoughtsPositioned` memo calculation to the tsx below. But that is going outside the rendering pipeline unnecessarily. The point of `treeThoughtsPositioned` is to calculate the x,y position of each thought (converting from `TreeThought[]` to `TreeThoughtPositoned[]`) so the render function has everything it needs to render the thoughts. 

If you need to add a new property to `TreeThoughtPositioned` in order for the render function to have everything it needs, feel free.

FYI There is additional documentation on the layout rendering algorithm here that might help: https://github.com/cybersemics/em/wiki/Docs#8-layout-rendering-algorithm

Zubair286:
> The `tableCol1Widths` get set and then used within a single render cycle though. So there is no need to store it in a ref from what I can tell. Unless I'm missing something, you're using the ref to shuttle a value from the `treeThoughtsPositioned` memo calculation to the tsx below. But that is going outside the rendering pipeline unnecessarily. The point of `treeThoughtsPositioned` is to calculate the x,y position of each thought (converting from `TreeThought[]` to `TreeThoughtPositoned[]`) so the render function has everything it needs to render the thoughts.
> 
> If you need to add a new property to `TreeThoughtPositioned` in order for the render function to have everything it needs, feel free.
> 
> FYI There is additional documentation on the layout rendering algorithm here that might help: https://github.com/cybersemics/em/wiki/Docs#8-layout-rendering-algorithm

Yes you are right, the only problem I am facing is that the tableCol1Width value is `undefined` during the drag process, and when drag ends it is assigned. I did start by returning the value from `treeThoughtsPositioned` but like I said it is `undefined` until drag ends, maybe I am missing something, let me see again.

raineorshine:
Okay, so apparently it does go outside a single render cycle. In that case, we should get a better understanding of why that is happening. We would like to find an intervention that is part of the existing render pipeline, or extend the render pipeline to accommodate the needed width in an elegant way, while avoiding ad hoc mutable state. Some additional state appears to be necessary, but how to add it without adding additional complexity is the key.

Zubair286:
> Okay, so apparently it does go outside a single render cycle. In that case, we should get a better understanding of why that is happening, and what the existing architecture allows. We would like to find an intervention that is part of the existing render pipeline, or extend the render pipeline to accommodate the needed width in an elegant way, while avoiding ad hoc mutable state. A render pipeline that cannot handle this case is flawed, and needs to be adjusted in a holistic way. Some additional state appears to be necessary, but how to add it without adding additional complexity is the key.

Right, we really only need to access the `width` actually, don't even need to add another property to `treeThoughtPositioned` but that too is `undefined` during drag, I should check `treeThoughtPositioned`, see why it is working like this.

src/components/LayoutTree.tsx
```diff
@@ -574,6 +578,7 @@ const LayoutTree = () => {
         width: tableCol1Widths.get(head(parentOf(node.path))),
         x,
         y,
+        col1ThoughtWidth,
```
raineorshine, line 581: An unintended consequence of using the last value of `col1ThoughtWidth` is that it gets set for all subsequent nodes, even when the table has long ended. 

It doesn't seem to be affecting the behavior because of the `isTableCol1 || isTableCol2` checks below, but that is only covering up the overreach. If `col1ThoughtWidth` was only set when needed, then `isTableCol1 || isTableCol2` would not be needed at all.

Let's make this more precise by only setting `col1ThoughtWidth` when needed, and getting rid of the `marginLeftOffset` intermediate value.

src/e2e/puppeteer/__tests__/snapshot.ts
raineorshine: I think this test is safe to remove now, since it has been copied to the drag describe block below.

src/e2e/puppeteer/__tests__/snapshot.ts
```diff
@@ -368,6 +390,28 @@ describe('drag', () => {
...
+        - =pin
+          - true
+        - b
+          - c
+        - d
+          - e
+    `)
+
+    await clickThought('x')
+    await dragAndDropThought('x', 'a', { position: 'after', childrenCount: 2 })
```
raineorshine, line 409: I think I would prefer an API of:

```ts
await dragAndDropThought('x', 'd', { position: 'uncle' })
```

which would mean "drag `x` after `d`, as its uncle."

`childrenCount` will not work for multiline thoughts in either col1 or col2.

raineorshine (changes requested):
Hi, thanks for your work.

I found a regression. This is breaking the drop-hover position when dragging into col1 thoughts. Could you add a snapshot test for this case as well?

## Current

ad6d955df6

current

## Expected

`main`

expected

src/components/LayoutTree.tsx
Zubair286: Hmm, the i`sTableCol1 || isTableCol2` are for double checking really since in useMemo its already checking for `node.isTableCol1` before setting the `col1ThoughtWidth` but it is not working it seems, it has to be for col1 for sure, let me see why is it not working.

src/e2e/puppeteer/__tests__/snapshot.ts
Zubair286: Then I think we need to find some other way to find the difference between `dragStart.y` and `dragEnd.y`.

raineorshine: Sure.

Since we already have the `dropUncle` property, maybe this would make more sense:

```ts
await dragAndDropThought('x', 'd', { position: 'after', dropUncle: true })
```

Either way, the point is that the API should reflect the semantics of the caller, not the implementation.

Zubair286: Right, I understand.

src/components/LayoutTree.tsx
Zubair286: Hmm, I see that we need to store the `col1ThoughtWidth` for `DropEnd` after the table, but because both `isTable` and `isTableCol1` are false in `useMemo`, we get 0 for the DropEnd, how about we store the value in a ref and then when both `isTable` and `isTableCol1` are false meaning we are out of the table, we assign that iterations `col1ThoughtWidth`, the refs value and clear the ref maybe.

raineorshine: No, that's adding even more complexity. You've already demonstrated that a ref is not necessary. 

Don't create state and let it "drift" across to the next iteration. Figure out exactly which node you need to set `col1ThoughtWidth` on, and then dynamically build the value on the spot. The nodes are already linearized in the `useMemo`, so `tableCol1Widths` should contain the width of the previous table.

Zubair286: I see that the drop-end which we need shift towards left, is rendered inside the last thought of column 2, it also renders two more drop-end one below last thought of column two and the last thought of column one, so setting `col1ThoughtWidth` when `isTableCol2` is `true` and below in tsx if we check `cliffDepth - depth < 0`, to not include the drop-end below column 2 thought can work, the additional checking is only because of multiple drop-ends rendering on the last thought of column 2.

raineorshine: That could work, though I think it's still covering up a mis/overapplied `col1ThoughtWidth`. `col1ThoughtWidth` should only be set in precisely the circumstance that is needed. Setting it on unrelated nodes (such as the table's other siblings) is just going to create confusion.

Another approach is to not render the drop-end in col2 at all. It's a carryover from non-table rendering. We could make a separate case and render it in a different place in the table view. We still haven't addressed [the regression](https://github.com/cybersemics/em/pull/2135#pullrequestreview-2192702723) created by the current approach. Taking a step back and fixing the cause of the problem in the React rendering tree may be safer than trying to compensate and adjust a component that is fundamentally in the wrong place.

Zubair286: Yes, although instead of applying it on all nodes this narrows down to only col 2 nodes, I see that it still is overapplied because we only really need it on the last node of column 2.

This approach does addresses the regression, the `cliffDepth - depth < 0` check is for that purpose, it will not apply margin-left on the right most drop-end. But I agree for table view drop-end rendering in column 2 seems wrong.

Zubair286: [Screencast from 07-24-2024 09:40:42 PM.webm](https://github.com/user-attachments/assets/11e3b58b-77ff-47e7-b810-9681febba315)

raineorshine: Okay, that's good at least. The behavior looks correct.

I'm happy with the outcome, so if we can narrow `col1ThoughtWidth` so that it is only set when needed, we may be able to move forward with this. I'll let you judge how important it is for #1887 to determine if we take Approach 2 (render `DropEnd` at the end of col1 instead of col2 when in the table view). Either way, I would suggest writing the test now with the working behavior in place.

Zubair286: Right, I already have written tests for both drag after table and column 2 drop-end, I would like to determine if approach 2 yields better result than this, with this behavior do you see #1887 fixed?

raineorshine: > Right, I already have written tests for both drag after table and column 2 drop-end

Yes, pending [this change](https://github.com/cybersemics/em/pull/2135/files#r1687871534).

> I would like to determine if approach 2 yields better result than this to continue with this approach now

Okay

> btw do you see #1887 fixed with this behavior?

Commit ad6d955df6? Yes, #1887 appears to be fixed.

Zubair286: Great, should I add the snapshot test for #1887 in this PR, or should I push it to PR #2125? I think it should be added here.

Zubair286:
I have set the `col1ThoughtWidth` on exactly the last node of column 2, rest is 0. I have also added snapshot test for #1887 to this PR. Please do review when you get the chance @raineorshine

raineorshine (approved):
Take a look at my changes in efeda28dcd. No additional state is necessary. When rendering col2, you can grab the width of col1 directly from the previous `treeThoughtsPositioned` node.

Zubair286:
> Take a look at my changes in [efeda28](https://github.com/cybersemics/em/commit/efeda28dcda5e0348d367178c96f1cad6ffdad75). No additional state is necessary. When rendering col2, you can grab the width of col1 directly from the previous `treeThoughtsPositioned` node.

Oh, I see what you did, I was focusing too much on the useMemo calculation that didn't think this through.I made the code too complex for no reason.. Sorry, this was my oversight.

raineorshine:
Will get this merged as soon as I figure out why the macos snapshots are failing.

Issue #1951, opened 2024-05-25
Two drop targets rendered in expanded context view child


raineorshine:
## Steps to Reproduce

```
- a
  - m
    - x
- b
  - m
    - y
```

1. Set the cursor on `a/m`
2. Activate the context view
3. Set the cursor on `a/m~/a`
4. Drag `a/m~/a/x` below `a/m~/b` and hold to expand. 
5. Drag and hover between `b` and `y`. 

## Current Behavior

Two drop targets are rendered between `b` and `y`.

https://github.com/cybersemics/em/assets/750276/bc57723f-8e35-4bf8-a09d-f5a298e9844c

## Expected Behavior

Only a single drop target should be rendered. Dropping there should move `x` as a sibling of `y`.

PR #2137, opened 2024-07-12
fixed double drop target issue in context view

Zubair286:
#1951 

The issue was happening because the drop-end was not unmounting upon the expanding of the thought 'b', it was not picking up the expansion of the thought, had to give it `path` instead of `simplepath` for it work.


raineorshine (approved):
Perfect, thanks.

PR #2168, opened 2024-07-15
#986 fix space takeup for url link icon

baproactive:


baproactive:
@raineorshine 
Would you review this one? 
I can't find an option to add this PR to you, so just ping you via comment.

raineorshine (changes requested):
It does fix the case in #986, but breaks the url icon on multiline urls:

## Current Behavior

f2ecc3c722693befefe455f3cf6d0b0f0b0b620d

Brave Browser 2024-07-16 08 02 13

## Expected Behavior

`main`

Brave Browser 2024-07-16 08 02 29

src/components/VirtualThought.tsx
```diff
@@ -144,7 +144,12 @@ const VirtualThought = ({
     const autofocusNew = calculateAutofocus(store.getState(), path)
     const isVisibleNew = autofocusNew === 'show' || autofocusNew === 'dim'
     if (!ref.current) return
-    const heightNew = ref.current.getBoundingClientRect().height
+
+    // Need to grab max height between .thought and .thought-annotation since the annotation height might be bigger (due to wrapping link icon).
+    const heightNew = Math.max(
+      ref.current.getBoundingClientRect().height,
+      ref.current.querySelector('.thought-annotation')?.getBoundingClientRect().height || 0,
```
raineorshine, line 151: Does this work if you just measure the height of the ThoughtAnnotation instead of the Thought? (instead of taking the max of the two)

raineorshine: Yes, the Thought height will always be the same or less than ThoughtAnnotation.

raineorshine: I think we should really optimize this query since it is a performance bottleneck. A direct ref to the ThoughtAnnotation would be ideal. Otherwise a fixed DOM traversal (admittedly fragile but probably worth it in this case).

baproactive: We have some cases where we don't have .though-annotation. 
image

baproactive: How about we do this? 
```    let heightNew = ref.current.querySelector('.thought-annotation')?.getBoundingClientRect().height || 0
    if (heightNew == 0) {
      heightNew = ref.current.getBoundingClientRect().height
    }
```

raineorshine: > We have some cases where we don't have .though-annotation.

Good catch.

> How about we do this?

Sure. Let's do that and optimize the `.thought-annotation` selector.

Except I'm now seeing that this does not result in the correct height. (Careful with a quick fix of a px offset here; the height needs to work across all font sizes.)

## Current Behavior

a4f9b63c78



## Expected Behavior

`main`

raineorshine (changes requested):
I'm now seeing that this does not result in the correct height unfortunately. (Careful with a quick fix of a px offset here; the height needs to work across all font sizes.)

## Current Behavior

a4f9b63c78

Height of single-line and ellipsized urls are too large.



## Expected Behavior

`main`

baproactive:
I found fixes and pushed the update. 
The root cause was height difference between .though-annotation and .thought when it's single line. 
It was due to the `overflow: hidden` adding up extra space to the parent. 
It is fixed by adding `vertical-align: top` to the .thought-annotation's span text.
I added inline comment as well.

src/components/VirtualThought.tsx
raineorshine: @baproactive

raineorshine (changes requested):
I can confirm the single-line urls have the correct height now.

I'm noticing that when the cursor is on a multiline thought, the height is too high.

I modified the url snapshot test so that it covers this case. You can get it if you rebase on `main`.

url-1-diff

baproactive:
Found the root cause of this issue - it's due to the a tag modifying the last line of the thought annotation text when it's multiline. 
I updated the a tag position to absolute to make sure it doesn't affect the annotation text lineheight.

raineorshine (changes requested):
The url link has to be part of the document flow and cause the thought to wrap, otherwise it can get cropped by the right edge of the screen on mobile. There is not enough margin-right on mobile when the thought goes to the end of the line.

(You can enable the mobile layout by using the Device Toolbar in Chrome Dev Tools.)



Maybe try limiting the height?

baproactive:
Hi @raineorshine 
I have update the approach slightly. 

position absolute on the thought-annotation a tag causes the mobile issue you mentioned, instead I set the a tag height to make sure it doesn't affect the thought annotation text's height. 
Please check.

raineorshine (changes requested):
The url link shifts position when the cursor is set on it:

https://github.com/user-attachments/assets/586cf942-cf98-456f-90d8-3f02c4bd6cde

raineorshine:
I've split up the url snapshot tests into three tests:

- single-line url (failing)
- multi-line url (failing)
- url child (passing)

That should cover all the cases we have discussed in this thread. I omitted the failing snapshots until they can be generated correctly. Run `yarn test:puppeteer -- snapshot -t url` to run the tests and `-u` to update.

baproactive:
I've fixed the url icon jumping issue. 
it was due to position absolute calculation diff.

raineorshine:
Thanks, I will test once I'm back at my computer. Please update the url snapshot tests as well.

raineorshine (changes requested):
> I've split up the url snapshot tests into three tests:
> 
> * single-line url (failing)
> * multi-line url (failing)
> * url child (passing)
> 
> That should cover all the test cases we have discussed in this thread **except the mobile test case** (normally that would be tested with WebdriverIO but that suite is broken currently).
> 
> I omitted the failing snapshots until they can be generated correctly. Run `yarn test:puppeteer -- snapshot -t url` to run the tests and `-u` to update.

I think my commit may have gotten overwritten by your last force push. Sorry that I wasn't clear I had pushed it to the PR branch. I pushed it again (6b100c2efd).

raineorshine (changes requested):
I can confirm the url jumping issue is fixed. Thanks.

Noticing a couple regressions:

1. At lower font sizes, the url link now incorrectly wraps. I added ce69a0f2b9 to run the url tests at font size 18 and 14 to cover this case.



2. The clickable area of the url link has been diminished. The entire link with a few pixels of padding should be clickable, as in `main`. As of 3881190c43 it's only clickable along the right edge:

https://github.com/user-attachments/assets/b7769dfe-b503-438a-a292-f2a3650d07a5

raineorshine (approved):
This looks good to me. I have manually tested and confirmed the following cases are working:

- [multiline url: url icon should be inline](https://github.com/cybersemics/em/pull/2168#pullrequestreview-2179291013)
- [multiline url with cursor: height should be the same](https://github.com/cybersemics/em/pull/2168#pullrequestreview-2195173567)
- [height of single-line and ellipsized urls should be the same](https://github.com/cybersemics/em/pull/2168#pullrequestreview-2181188572)
- [url icon should not shift position with cursor](https://github.com/cybersemics/em/pull/2168#pullrequestreview-2202799879)
- [at lower font sizes, url icon wraps](https://github.com/cybersemics/em/pull/2168#pullrequestreview-2213218608)
- [[mobile] url icon should not be cropped](https://github.com/cybersemics/em/pull/2168#pullrequestreview-2201654629)

This issue is partially fixed:

- [extended click area](https://github.com/cybersemics/em/pull/2168#pullrequestreview-2213218608) - The top, right, and left sides correctly extend the click area. The bottom does not.

Given the nearly complete coverage of the observed issues, I am satisfied with this PR and think it should be merged. I will open a separate issue for the minor click area issue.

Thanks so much for your persistence! :)

PR #2171, opened 2024-07-19
added mouse pointer to puppeteer tests

Zubair286:
#2103


raineorshine:
Thanks! Please update the snapshot tests as well.

src/e2e/puppeteer/helpers/index.ts
```diff
@@ -73,7 +75,12 @@ const index = <T extends any[]>(...setupArgs: T) => {
   const index = partialWithRef(pageRef, helpers)
 
   beforeEach(async () => {
-    pageRef.current = await setup(...setupArgs)
+    const regex = /\b(drag|drop)\b/i
+
+    // Check if the test name contains 'drag' and 'drop' to determine if the mouse helper should be installed.
+    // TODO: looking for more reliable solution as this forces the use of drag/drop words in the test name.
```
raineorshine, line 81: Yeah, I don't think this is so good. We need a more imperative way to set this explicitly on a per-test basis, such as a testFlag.

raineorshine (changes requested):

src/e2e/puppeteer/helpers/index.ts
Zubair286: Yeah, setting it in testFlag is a better way I think.

Zubair286: Aside from the execution, setting up a testFlag has another issue, because `em.testFlags` can only be accessed inside `page.evaluate`, and `mousePointerHelper` then won't be directly used inside `page.evaluate`, instead we have to serialize the function and `eval` it inside the `page.evaluate`.

raineorshine: How about change `mousePointerHelper` to a function that can be called in the test itself. Then you don't need any test flag.

Zubair286: Yes, I tried that for some reason it does not work unless its called in `setup` function, maybe it has to do something with the `page` instance. I will try to make it work in the test, as that seems to me the most safe option.

raineorshine: The other option is to enable it all the time, but only if that does not affect performance, since it would impact all other tests. 

Otherwise, it would be best to enable it by calling a function in the test itself. It doesn't make sense to do it in the setup function if we are only enabling it on a per-test basis. Once you take it out of the `DOMContentLoaded` handler, I don't see why it wouldn't work outside setup.

Zubair286: Yes, I took it out of the `DOMContentLoaded` and used `evaluate`, and it is working now although the pointer appears at a delay, I think its better if we call it in `dragAndDropThought` function since it will be invoked if we are performing drag/drop.

Zubair286:
The puppeteer test job is skipped, is this intentional? @raineorshine

raineorshine:
No, GitHub does that sometimes and I can't figure out why. Probably because it takes too long.

src/e2e/puppeteer/helpers/mousePointerHelper.ts
```diff
@@ -0,0 +1,87 @@
...
+        background-size: cover;
+        border: none;
+        margin: -10px 0 0 -10px;
+        padding: 0;
+        transition: background .2s, border-radius .2s, border-color .2s;
+      }
+      puppeteer-mouse-pointer.button-1 {
+        transition: none;
+        background: rgba(0,0,0,0.9);
+      }
```
raineorshine, line 32: I think the button logic is part of the original circular pointer, and is no longer relevant.

src/e2e/puppeteer/helpers/mousePointerHelper.ts
```diff
@@ -0,0 +1,87 @@
+import { Page } from 'puppeteer'
+
+/**
+ * Installs mouse helper on the page to display the mouse pointer.
+ */
+async function installMouseHelper(page: Page) {
```
raineorshine, line 6: Let's call this `showMousePointer` and rename the file to match.

raineorshine (changes requested):

src/e2e/puppeteer/helpers/showMousePointer.ts
```diff
@@ -3,51 +3,47 @@ import { Page } from 'puppeteer'
...
-      }
-      puppeteer-mouse-pointer.button-5 {
-        transition: none;
-        border-color: rgba(0,255,0,0.9);
-      }
     `
     document.head.appendChild(styleElement)
+    box.appendChild(container)
     document.body.appendChild(box)
     document.addEventListener(
```
raineorshine, line 48: You can fully remove button-related code since it is no longer used.

raineorshine (changes requested):
Please update the relevant snapshots.

src/e2e/puppeteer/helpers/showMousePointer.ts
Zubair286: Right, I guess only mouseDown and mouseUp event listeners remain which are button related.

raineorshine (changes requested):
The position of the mouse pointer was a bit off, resulting in this:



I fixed the position, and did a larger refactor in order to simplify the implementation.

- Insert svg directly in innerHTML rather than encode and insert as a background image data url.
- Removed unnecessary svg properties.
- Set styles directly on element rather than inserting a style element in the head.

Please update the ubuntu snapshots for the drag-and-drop tests and I will merge. Thanks!

raineorshine (approved):

Issue #1952, opened 2024-05-26
Cliff drop targets obscured by user's finger


raineorshine:
## Steps to Reproduce

```
- a
  - b
    - c
      - d
        - e
          - f
```

Drag `f` onto any of the drop targets over the cliff. 

## Current Behavior

The drag zone is too far to the left, so the user's finger obscures the drop hovers. 

https://github.com/cybersemics/em/assets/750276/942192a8-e29b-49d5-9823-b1074892395d

Note: The video does not show where the user's finger is in contact with the screen, but is included here to clarify the steps to reproduce.

## Expected Behavior

The drag zones should be shifted right so that the left edge of the drop hover is clearly visible.

Zubair286:
Wouldn't shifting right undo the way the drop-end is flush to the thoughts right now?

raineorshine:
The position of the (invisible) drop target should change without changing the position of the (visible) DropHover. How to achieve that of course will depend on the component hierarchy, DOM hierarchy, and CSS, and may require changes in multiple places.

PR #2175, opened 2024-07-22
Adjust position of drop target to prevent user finger obscuring

Zubair286:
#1952 

Adjusted css for drop-end to shift it towards right, so on mobile users finger do not obscure the drop hover


src/App.css
```diff
@@ -2128,7 +2128,7 @@ code,
   position: absolute;
   height: 0.5em;
   /* margin to offset bullet width */
-  margin-left: -1em;
+  margin-left: 0.5em;
```
raineorshine, line 2131: Can you explain this value?

src/components/DropEnd.tsx
```diff
@@ -124,6 +126,7 @@ const DropEnd = ({
           className='drop-hover'
           style={{
             backgroundColor: dropHoverColor,
+            marginLeft: !isRootPath && Number(cliff) < 0 ? 'calc(-0.9em - 13px)' : undefined,
```
raineorshine, line 129: Can you explain this so I understand why these values were chosen?

src/components/DropUncle.tsx
```diff
@@ -44,6 +44,7 @@ const DropUncle = ({
         backgroundColor: testFlags.simulateDrop ? '#52305f' : undefined, // eggplant
         height: '1.9em',
         opacity: 0.9,
+        marginLeft: '-1em',
```
raineorshine, line 47: Does this balance out another value? If so, we need to use a shared variable. Right now all the values appear to be independent.

raineorshine (changes requested):
Thanks, here's what I found:

1. I need a little more explanation of what the values are that you chose, and why you chose them in order to properly evaluate the solution. (Noted in comments below.)
1. This should only change on [touch](https://github.com/cybersemics/em/blob/df3e9748cdc101af433a0401a2271f87ddf39e83/src/browser.ts#L6-L9) devices. The user's finger is not a problem on desktop.
2. We can increase the amount that it shifts over quite a bit, say an additional 3em. My finger still obscures the drop-hover.
3. Why pass `cliff` into `DropEnd` in order to shift the position there, instead of shifting its left position in LayoutTree? https://github.com/cybersemics/em/blob/df3e9748cdc101af433a0401a2271f87ddf39e83/src/components/LayoutTree.tsx#L731
4. One of the snapshot tests is failing. Either it is expected, and you need to update the ubuntu snapshot, or it is unexpected and needs to be investigated.

src/components/DropEnd.tsx
Zubair286: Yeah, so I increased the drop-end margin-left to 0.5em, so with that drop-hover also moved, we needed to move the drop-hover back to its original position relative to the value of margin-left increase in drop-end which ended up being 0.9em for drop-hover.

src/components/DropUncle.tsx
Zubair286: This was actually because DropUncle test was failing because of the changes made in the `App.css`, so I set it to the original value.

Zubair286:
> Thanks, here's what I found:
> 
> 1. I need a little more explanation of what the values are that you chose, and why you chose them in order to properly evaluate the solution. (Noted in comments below.)
> 2. This should only change on [touch](https://github.com/cybersemics/em/blob/df3e9748cdc101af433a0401a2271f87ddf39e83/src/browser.ts#L6-L9) devices. The user's finger is not a problem on desktop.
> 3. We can increase the amount that it shifts over quite a bit, say an additional 3em. My finger still obscures the drop-hover.
> 4. Why pass `cliff` into `DropEnd` in order to shift the position there, instead of shifting its left position in LayoutTree? https://github.com/cybersemics/em/blob/df3e9748cdc101af433a0401a2271f87ddf39e83/src/components/LayoutTree.tsx#L731
> 5. One of the snapshot tests is failing. Either it is expected, and you need to update the ubuntu snapshot, or it is unexpected and needs to be investigated.

1. Explained in the comments below.
2. Yes, thats an oversight on my part, sorry about that. Maybe I should only apply the css changes when `isTouch` is `true`.
3. I increased the `margin-left` of `DropEnd` by `2em`, seemed just enough. Let me increase and check if it works.
4. The reason for passing cliff into `DropEnd` was to make sure to only increment margin-left of `DropHover` on a `cliff`. Otherwise it would have affected on other areas as well.
5.  That is because of the DropTarget being moved towards right, I think should update the snapshot.

src/App.css
Zubair286: It was to shift the DropEnd towards right, `1.5em` seemed just enough.

src/components/DropEnd.tsx
raineorshine: And the 13px?

src/components/DropUncle.tsx
raineorshine: The drop target also needs to be shifted over for `DropUncle`. I can confirm that the user's finger obscures the drop-hover in [this case](https://github.com/cybersemics/em/blob/2dd4f8a0116315c7b2612412727ddc67a9b3e9de/src/e2e/puppeteer/__tests__/snapshot.ts#L346-L371).

raineorshine:
> 4\. The reason for passing cliff into `DropEnd` was to make sure to only increment margin-left of `DropHover` on a `cliff`. Otherwise it would have affected on other areas as well.

So what you are saying is: You are only applying the marginLeft to `drop-hover`, not `drop-end`. If you had adjusted the `left` value of the entire `DropEnd`, then it would have incorrectly shifted the `drop-end` container over as well.

Is that right?

I'd like to see this shift logic unified if possible . It's currently spread out over `DropEnd`, `DropUncle`, and `LayoutTree`.

src/components/DropEnd.tsx
Zubair286: I kept that the same as it was in the drop-hover class, since the 13px was already applied to it. I only calculated the `em` value

Zubair286:
> > 4. The reason for passing cliff into `DropEnd` was to make sure to only increment margin-left of `DropHover` on a `cliff`. Otherwise it would have affected on other areas as well.
> 
> So what you are saying is: You are only applying the marginLeft to `drop-hover`, not `drop-end`. If you had adjusted the `left` value of the entire `DropEnd`, then it would have incorrectly shifted the `drop-end` container over as well.
> 
> Is that right?
> 
> I'd like to see this logic more unified if possible . It's currently spread out over `DropEnd`, `DropUncle`, and `LayoutTree`.

Yes, DropEnd margin-left is already increased from drop-end class, this is why only `DropHover` had to be adjusted here but first I guess I should make it only when `isTouch` is true.

src/components/DropEnd.tsx
raineorshine: Okay, I see now.

It looks like the styles are already split between App.css and inline styles, which is creating added complexity. I will have to address that problem in a separate task. For now, let's carefully each of these changes with inline comments. Otherwise it's going to be very hard for future developers to understand why the values were chosen, and what nonlocal styles they depend on.

raineorshine:
Yeah, and I'm seeing that a lot of the complexity is due to the existing architecture. There may be only so much we can do to simplify the solution.

Either we shift all `drop-end` elements and then compensate the `drop-hover` as you have done, or we shift the `DropEnd` component using `left` in `LayoutTree` and then compensate the contained `drop-end` element. I'm not sure if the latter approach is any better, but I can say that our logic is too spread out and we're applying some of the styles too broadly, which is creating the need for additional compensations that may not be necessary.

Zubair286:
> Yeah, and I'm seeing that a lot of the complexity is due to the existing architecture. There may be only so much we can do to simplify the solution.
> 
> Either we shift all `drop-end` elements and then compensate the `drop-hover` as you have done, or we shift the `DropEnd` component using `left` in `LayoutTree` and then compensate the contained `drop-end` element. I'm not sure if the latter approach is any better. But currently our logic is a bit too spread out, and we're applying some of the styles too broadly, which is creating the need for additional compensations that may not be necessary.

I think shifting the `drop-end` is somewhat better because `LayoutTree` styling is kind of already getting complex with issue #1886 and #1887, which is why I did it. I will try to make it simpler, to not make it affect too broadly.

raineorshine (changes requested):
You may need to rebase on `main` and regenerate the ubuntu drag-and-drop snapshots. Let me know when it's ready for review. Thanks!

Zubair286:
@raineorshine please review the PR, the test cases failing are expected due to some changes I made.

raineorshine (changes requested):
Please generate the ubuntu snapshots so that I can visually confirm the changes before I pull the branch and generate the macos snapshots.

The inline comments help a lot, thanks. 

Next you need to take advantage of variables to reduce code duplication and, more importantly, maintainability. If you specify `calc(-1.9em - 13px)` in two separate files, future developers will not realize that there is a hidden dependency there. Changing one should always change the other. Therefore, specify the value in a `const` that can be imported into each place of usage. You can add global constants to https://github.com/cybersemics/em/blob/58cf75f1f3450ed0718332133cc7adbfeefa8129/src/constants.ts.

Second, there is a relationship between this added value and `margin-left: calc(0.6em - 13px);` in App.css. They are tightly coupled. Changing one without changing the other would cause a regression. Find a way to unify this with code so that we can some compile-time safety. You may need to move the `.drop-end .drop-hover` style into React.

Lastly, the amount that the drop-target is shifted over, `1.5em`, needs to be unified in the same way. Everywhere that relies on that value needs to reference a variable rather than hardcoding the value. And if the `-1.9em` is related, then that needs to be reflected in the code, such as `${-0.4 - fingerShift}em`.

CSS is traditionally untyped and fragile due to all the implicit relationships that exist between styles, properties, and the DOM hierarchy. CSS variables help, but in React we often need to determine the styles to apply at runtime, so it makes sense to move style logic outside of App.css and into components where we have the full power of Typescript. The more we capture style relationships in Typescript and avoid hardcoding values, the more robust the code will be over time, and the less prone to breakage as the component and DOM hierarchy change. Hope that makes sense.

src/components/DropUncle.tsx
```diff
@@ -62,7 +67,16 @@ const DropUncle = ({
...
       {(testFlags.simulateDrag || isHovering) && (
-        <span className='drop-hover' style={{ backgroundColor: dropHoverColor }} />
+        <span
+          className='drop-hover'
+          style={{
+            backgroundColor: dropHoverColor,
+            // move drop-hover to left relative to the shift towards right of drop-target on mobile devices
+            marginLeft: isTouch
+              ? `calc(${DROPHOVER_MARGINLEFT - calculateNewMargin()}em - 13px)`
+              : `calc(${DROPHOVER_MARGINLEFT}em - 13px)`,
```
raineorshine, line 77: Put the ternary inside to remove duplication

src/components/DropEnd.tsx
```diff
@@ -124,6 +135,11 @@ const DropEnd = ({
           className='drop-hover'
           style={{
             backgroundColor: dropHoverColor,
+            // move drop-hover to left relative to the shift towards right of drop-target on mobile devices
+            marginLeft:
+              isTouch && !isRootPath && Number(cliff) < 0
+                ? `calc(${DROPHOVER_MARGINLEFT - calculateNewMargin()}em - 13px)`
+                : `calc(${DROPHOVER_MARGINLEFT}em - 13px)`,
```
raineorshine, line 142: Put the ternary inside to remove duplication

src/components/DropEnd.tsx
```diff
@@ -98,7 +102,14 @@ const DropEnd = ({
...
         height: isRootPath ? '8em' : '1.9em',
-        marginLeft: isRootPath ? '-4em' : last ? '-2em' : undefined,
+        // On mobile move drop-target towards right so user's finger does not obscure the drop-hover
+        marginLeft: isRootPath
+          ? '-4em'
+          : last
+            ? '-2em'
+            : isTouch
+              ? `${DROPEND_MARGINLEFT + calculateNewMargin()}em`
+              : `${DROPEND_MARGINLEFT}em`,
```
raineorshine, line 112: Put the ternary inside to remove duplication

src/constants.ts
```diff
@@ -501,3 +501,6 @@ export const LAYOUT_NODE_ANIMATION_DURATION = 150
 
 export const GESTURE_GLOW_BLUR = 10
 export const GESTURE_GLOW_COLOR: keyof typeof colors.dark = 'gray'
+export const DROPEND_MARGINLEFT = -1
+export const DROPHOVER_MARGINLEFT = 0.6
+export const MARGINLEFT_CHANGE = 2.5
```
raineorshine, line 506: Let's rename `MARGINLEFT_CHANGE` to `DROPHOVER_FINGERSHIFT`

src/util/calculateMargin.ts
```diff
@@ -0,0 +1,22 @@
+import { DROPEND_MARGINLEFT, MARGINLEFT_CHANGE } from '../constants'
+
+/**
+ * Calculates the change in the margin-left of drop-end to compensate for the shift towards the right on mobile devices.
+ */
+function calculateNewMargin() {
```
raineorshine, line 6: This function always returns `MARGINLEFT_CHANGE`...?

raineorshine (changes requested):

src/util/calculateMargin.ts
Zubair286: Yes, it returns the updated change in drop-end margin-left, maybe instead of making it global I should move it inside the function since we don't really need it anywhere else.

raineorshine: No, I mean it always returns `2.5`, no matter what.

src/util/calculateMargin.ts
```diff
@@ -0,0 +1,22 @@
...
+  const changeInValue = MARGINLEFT_CHANGE
+
+  // Calculate the change in drop-end marginLeft from its initial value
+  const dropEndChange = dropEndValue - initialValue
+
+  // Calculate the new changeInMarginLeft
+  const newChangeInMarginLeft = changeInValue + dropEndChange
+
+  return newChangeInMarginLeft
+}
```
Zubair286, line 20: If  we increase the `DROPEND_MARGINLEFT` value to -2 here, than `dropEndChange` calculates the difference which would be 1 in this case, and that is added to `changeInValue` which would result in this case to 3.5, and then return.

raineorshine: Okay, I see. 

Let's make it a constant rather than a function.

Zubair286: I think it would look something like this, 


```
const INITIAL_MARGINLEFT = -1
const INTIAL_DROPHOVER_FINGERSHIFT = 2.5
export const DROPEND_MARGINLEFT = -1
export const DROPHOVER_MARGINLEFT = 0.6

export const DROPHOVER_FINGERSHIFT = DROPEND_MARGINLEFT - INITIAL_MARGINLEFT + INITIAL_DROPHOVER_FINGERSHIFT
```
I was hoping to put the ternary inside the function though.

raineorshine: That’s better. The DROPEND_MARGINLEFT and INITIAL_MARGINLEFT are still a little unclear to me. I'll have to analyze the whole to decide on a better name and an appropriate JSDOC comment. 

Take your best shot at it and I'll review.

Zubair286: The `DROPEND_MARGINLEFT` is the actual value for drop-end, `INITIAL_MARGINLEFT` is what it was initially this is supposed to remain same and not change. As for ternary, I am thinking of keeping the function and putting the ternary operations inside for both drop-end and drop-hover.

raineorshine (commented):
I'm going to do some work on this to see if I can get the abstractions in the right place. While we have reduced duplication, the complexity has increased more than expected.

raineorshine (approved):
In the main branch, the `.drop-end` and `.drop-hover` is fairly spread out, making it hard to easily intervene without causing a regression.

`.drop-hover` is given a different margin-left depending on its container:

https://github.com/cybersemics/em/blob/c10897fd461131acb2f66c58bffbbc3cdf9b5e87/src/App.css#L2105-L2115

https://github.com/cybersemics/em/blob/c10897fd461131acb2f66c58bffbbc3cdf9b5e87/src/App.css#L2152-L2156

https://github.com/cybersemics/em/blob/c10897fd461131acb2f66c58bffbbc3cdf9b5e87/src/App.css#L2455-L2461

`.drop-end` is given a default margin-left in `App.css`, and it is sometimes overridden in the `DropEnd` component:

https://github.com/cybersemics/em/blob/c10897fd461131acb2f66c58bffbbc3cdf9b5e87/src/App.css#L2117-L2122

https://github.com/cybersemics/em/blob/c10897fd461131acb2f66c58bffbbc3cdf9b5e87/src/components/DropEnd.tsx#L92-L101

Consolidating this logic is an advisable refactor, though it probably should be done separately from this PR. Therefore, I opted to apply the finger shift with `translateX` to avoid conflicting with the existing margin-left logic.

Tested on mobile and it appears to work well in all cases, so I removed the `DropUncle` and `cliff` conditional logic.

Issue #2087, opened 2024-06-29
Tutorial: Old tutorial steps accumulate at the bottom of the tutorial


raineorshine:
## Steps to Reproduce

1. Clear site data or open em in an incognito window.
2. Click START TUTORIAL.
3. Click the Next button.

## Current Behavior

The text from the first tutorial step shows up at the end of the second tutorial step. Previous tutorial steps continue to accumulate after each step.

After Step 1:
<img width="752" alt="Brave Browser 2024-06-29 13 28 48" src="https://github.com/cybersemics/em/assets/750276/85d5ffc1-30a6-4919-804f-8b011e651a77">

After Step 2:
<img width="755" alt="Brave Browser 2024-06-29 13 41 46" src="https://github.com/cybersemics/em/assets/750276/e3716257-1d92-4456-a070-c238f50ddf62">

## Expected Behavior

Only the text for the current tutorial step should be shown.

After Step 1:
<img width="752" alt="image" src="https://github.com/cybersemics/em/assets/750276/0e31911e-bd3b-4a72-b810-417c71f2d9eb">

## Note

This may be a regression from the upgrade to Vite a couple months ago. Something appears to be broken with the use of [CSSTransition](https://reactcommunity.org/react-transition-group/css-transition): https://github.com/cybersemics/em/blob/ee9b4ba13adb79d96da52cd4c1eb13e51b4d5aa4/src/components/Tutorial/index.tsx#L39-L51.

michaelperros:
Hi @raineorshine 

I created pull request #2189 to fix this issue.

The issue was caused by the key prop being a special property in React that isn't passed down to child components. I resolved this by renaming the key prop to transitionKey to ensure it is properly passed and used in the CSSTransition component.

Please let me know if there are any questions or if further adjustments are needed.

PR #2189, opened 2024-07-28
Fix tutorial step accumulation issue #2087

michaelperros:
This fixes the issue #2087 

Solution: by renaming the key prop to transitionKey, we ensure that the prop is correctly passed and accessible. Since "key" is a special property in React, it could not keep track of each element being mapped and thus caused the accumulation.

Please review the changes and let me know if further adjustments are needed.


raineorshine (approved):
Good catch! Thanks

michaelperros:
Happy to help! I’d love to assist with the other issues and can start this week if you wish. I’ve sent you a proposal via UpWork. Please let me know if you need any further information for your consideration.

Issue #2088, opened 2024-06-29
Toolbar button stuck down


raineorshine:
## Steps to Reproduce

1. Create thought `a`
2. Click and drag a toolbar button down.

## Current Behavior

The toolbar button stays depressed.

https://github.com/cybersemics/em/assets/750276/68c9d536-92e9-4e56-8362-af4ebad40510

## Expected Behavior

The toolbar button should be reset after the cursor moves off the toolbar, and the command should be cancelled.

michaelperros:
Please see #2196

PR #2196, opened 2024-07-29
Fix stuck toolbar button

michaelperros:
This is a simple fix to issue #2088 

It passes the preexisting state setter setPressingToolbarId to the ToolbarButton component, then calls it when onMouseLeave is triggered AND customize is not true

From a quick glance, it appears your code was attempting to use the drag and drop handlers to essentially do the same thing, but these are actually not called in this edge case (perhaps because customize is false, which disables drag all together)

Please let me know if you have any questions. Would love to have a chance to discuss contract work with you. 

If you need any adjustments here, please let me know.

Thanks!
Michael


raineorshine:
Thanks again for giving this a go.

PR #2202, opened 2024-07-31
Fixed bug that kept the toolbar icon pressed when it was dragged

BlaineOmega:
This fixes the problem of bug #2088. Added a method to reset the isPressed value in the toolbar button when the user drags the toolbar button outside of the toolbar.

### DIFF
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -95,7 +95,7 @@ const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected, handleMouseU
   }, [isDraggingAny, setPressingToolbarId])
 
   /**
-   * update isPressed when the drag is released outside of the toolbar
+   * Udate isPressed when the drag is released outside of the toolbar
    * */
   useEffect(() => {
     const handleMouseUp = event => {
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
Looks good. One small correction.

src/components/Toolbar.tsx
```diff
@@ -94,6 +94,25 @@ const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected }) => {
     }
   }, [isDraggingAny, setPressingToolbarId])
 
+  useEffect(() => {
+    /**
+     * Update isPressed when the drag is released outside of the toolbar.
+     * */
+    const handleMouseUp = event => {
+      if (handleMouseUp) {
```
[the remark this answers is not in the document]
BlaineOmega: Removed the check. Thx.

src/components/Toolbar.tsx
```diff
@@ -30,7 +30,7 @@ interface ToolbarProps {
 }
 
 /** Toolbar component. */
-const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected }) => {
+const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected, handleMouseUp }) => {
```
raineorshine, line 33: Remove unused prop

raineorshine (changes requested):

src/components/Toolbar.tsx
BlaineOmega: That prop is used. Removing it throws an error `handleMouseUp is not defined`

BlaineOmega: Sorry, I got it. Found the issue.

raineorshine (approved):

Issue #1231, opened 2021-06-09
Toolbar button animation tweaks


raineorshine:
## Steps to Reproduce

1. Create thought `a`.
2. Quickly click and release a toolbar button.

## Current Behavior

When the mouse button is quickly released, the dip animation is quickly reversed, showing little change.

In this video, there is a long click (correct) followed by a short click (incorrect).

https://github.com/cybersemics/em/assets/750276/ead547e6-98c8-4efd-b59d-5f8d0af446da

## Expected Behavior

- Complete the full dip animation even if only quickly tapping and releasing.
- Decrease the animation transition duration for a snappier effect.

shresthabijay:
@linghong Any updates?

PR #2206, opened 2024-08-01
1231 icon transition

BlaineOmega:
#1231 bug fix. Also reverted #2088 as it was unneeded with this fix


raineorshine (changes requested):
This looks good for short tap, but not so much for long tap or tap-and-hold. The user still has the mouse button held down, but the toolbar button pops back up. We need the visual feedback from the toolbar button to correspond to the tactile feedback of the mouse click. 

- Short tap: Full animation, even if mouse is released before animation completes.
- Long tap: Reset when mouse is released.
- Tap and hold: Reset when mouse is released.
- Tap and drag away: Reset when mouse is released OR reset when mouse leaves toolbar button (I'm fine with either)

Compared to `main`, we really just want to alter the "Short tap" case.

raineorshine (changes requested):
That looks good. I think we still need to make sure the full animation is completed regardless of how short the click is. 

For example, tapping on a trackpad results in no time in between mousedown and mouseup:

https://github.com/user-attachments/assets/ee1903fb-b02a-4598-8708-b4605967e8f1

raineorshine (changes requested):
The animation behavior works well on both short and long clicks.

Double clicking on a toolbar button sometimes causes it to get stuck down.

src/components/Toolbar.tsx
```diff
@@ -94,21 +121,43 @@ const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected }) => {
...
+     **/
+    const handleTransitionEnd = () => {
+      if (!isMousePressed) {
+        setPressingToolbarId(null)
+      }
+      setIsTransitioning(false)
     }
 
+    // Select all elements with the class 'toolbar-icon'
+    const icons = document.querySelectorAll('.toolbar-icon')
```
raineorshine, line 137: It's usually best to avoid directly accessing the DOM from React components unless absolutely necessary, as changes to the DOM are not captured by the type system.

If we go with this, however, the approach can be slightly improved by using a separate data attribute rather than reusing an existing CSS class.

BlaineOmega:
Closed. Fixed on another PR.

Issue #1967, opened 2024-05-27
Add New Thought to toolbar


raineorshine:
- [x] Add `'newThought'` to the [TOOLBAR_DEFAULT_SHORTCUTS](https://github.com/cybersemics/em/blob/c21cdc17813b6fe06f2e112cfc56693a34ed8b4a/src/constants.ts#L166) so it shows up in the toolbar after undo/redo.
- [x] Replace the `newThought` shortcut [icon](https://github.com/cybersemics/em/blob/c21cdc17813b6fe06f2e112cfc56693a34ed8b4a/src/shortcuts/newThought.tsx#L10) with https://thenounproject.com/icon/plus-5662424/
- [x] Make sure the icon is sized, positioned, and styled like the other toolbar icons.

PR #2207, opened 2024-08-01
1967 updated the new thought icon

BlaineOmega:
#1967 Added correct icon to the toolbar for new thoughts


raineorshine (changes requested):
It looks like it's cropped on the right and bottom edges:



The snapshot tests are failing because they need to be regenerated with the new toolbar. You can update all snapshots with `npm run test:puppeteer -- snapshot -u`, or update specific snapshots with `npm run test:puppeteer -- snapshot -t TEST_TITLE -u`. Just visually check the generated snapshots as a sanity check before pushing to the PR. Thanks!

raineorshine (approved):

Issue #1839, opened 2024-04-28
Upgrade react-dnd to hooks


raineorshine:
The current usage of react-dnd is through a legacy Higher Order Component. 

Upgrade react-dnd to the latest version and switch to hooks.

1. Convert HOC to hooks one component at a time (at v11).
   - [x] [DragAndDropThought](https://github.com/cybersemics/em/pull/2203)
   - [x] [DragAndDropSubThought](https://github.com/cybersemics/em/pull/2220)
   - [x] DragAndDropToolbarButton
   - [x] Favorites
   - [x] ShortcutRow
   - [x] QuickDropIcon
1. Upgrade one major version at a time until v16.
1. Refactor component hierarchy to better reflect new hook usage.
1. Switch enum DragAndDropType back to PascalCase.

raineorshine:
# Drag-and-drop related components

## Toolbar

[DragAndDropToolbarButton](https://github.com/cybersemics/em/blob/main/src/components/DragAndDropToolbarButton.tsx) - Defines `canDrag`, `beginDrag`, `endDrag`, `canDrop` and `drop` for dragging or dropping a toolbar button when customizing the toolbar.

https://github.com/cybersemics/em/assets/750276/4950d843-1a12-4647-a511-5625def7311c

## Thoughts

[DragAndDropThought](https://github.com/cybersemics/em/blob/main/src/components/DragAndDropThought.tsx) - Defines `canDrag`, `beginDrag`, `endDrag`, `canDrop` and `drop` for dragging or dropping a thought. Used by the `Thought` component so that every thought can be dragged, and every thought can serve as a drop target (inserted before, i.e. as the previous sibling).

https://github.com/cybersemics/em/assets/750276/2d6c967c-8cb6-484e-ad66-9cc50cfdd35d

[DropEmpty](https://github.com/cybersemics/em/blob/main/src/components/DropEmpty.tsx) - Allows a thought to be dropped as a subthought of a thought that is empty, i.e. a leaf.

https://github.com/cybersemics/em/assets/750276/93ff603d-31fa-409c-baa4-5c6dabe5887c

[DropEnd](https://github.com/cybersemics/em/blob/main/src/components/DropEnd.tsx) - Allows a thought to be dropped at the end of a list of subthoughts.

https://github.com/cybersemics/em/assets/750276/4f2049b3-1599-4ada-a203-c7ebb0941240

[...]

PR #2215, opened 2024-08-02
fixed issue with snapshot test failing after migrating from react-dnd higher order to hooks

Zubair286:
#1839 

The issue was in `enum DragAndDropType`, `thought and toolbarButton` should be smaller case so it matches the `type` returned by `beginDrag`.

```
enum DragAndDropType {
  Thought = 'thought',
  ToolbarButton = 'toolbarButton',
}
```


raineorshine (changes requested):
Ah, my bad! I didn't realize that my enum had messed it up.

However this raises the question of why the type system didn't catch this. Enums are opaque types, so the runtime value shouldn't matter. `beginDrag` was returning `DragAndDropType.Thought`, not `'thought'`. Where was the lowercase value coming from? Was there a hardcoded instance of `'thought'` that I missed, or does react-dnd convert it to camelCase?

Zubair286:
> the

There does not seem to be any hardcoded instance of `'thought'`, maybe it is react-dnd but I am not too sure about it.

raineorshine:
Okay, makes sense. We can circle back at the end to see if we can switch it back to PascalCase.

raineorshine (approved):

Issue #2060, opened 2024-06-23
Toolbar: Active state of formatting shortcuts should reflect selected text


raineorshine:
When part of a thought is selected, the active state of the toolbar buttons for bold/italic/underline/strikethrought should reflect only the selected text, not the entire thought. 

When no text is selected, the toolbar buttons for bold/italic/underline/strikethrough should be active if and only if all the text in the thought is formatted. 

There are two incorrect cases given below. Please ensure the solution works for all types of formatting: bold, italic, underline, and strikethrough. 

# 1. No text selected with mixed formatting

## Steps to Reproduce

```
Hello **world**
```

## Current Behavior

The toolbar button for bold is active. 

![image](https://github.com/cybersemics/em/assets/750276/2ace88e8-5c36-4305-8208-f22d6cc71323)

## Expected Behavior

The toolbar button for bold should not be active, because the thought has some non-bold characters. 

# 2. Unformatted text selected

## Steps to Reproduce

Select the word 'world'. 

## Current Behavior

Currently, the toolbar button for bold is active, but not for the right reason. Once the above issue is fixed, it will (incorrectly) be inactive. 

![Paste 2024-08-02 14 28 29](https://github.com/user-attachments/assets/31a0e7c1-a122-42b9-9650-72157ec2f7e3)

## Expected Behavior

The toolbar button for bold should be active, because all selected characters are bold.

finkef:
https://github.com/cybersemics/em/pull/2187 partially fixes this by moving tracking of formatting commands to a separate store - triggered by selection changes and commands. In order to fully fix this issue, we would need to remove checks for HTML tags in thought content, e.g. `<i>`, `<b>`, etc.

PR #2218, opened 2024-08-02
2060 formatting button

BlaineOmega:
Addresses issue for #2060 . Leverages fix from @finkef and PR https://github.com/cybersemics/em/pull/2213


raineorshine:
Please avoid using others' solutions during the trial period, as I need to see how you would solve these problems. #2213 is not the correct solution anyway.

I'm going to close this PR, but I've rebased `react-blaine` on `main` so that you have Fabian's commit.

PR #2224, opened 2024-08-05
Migrate ShortcutRow higher order component into hooks

Zubair286:
#1839


src/hooks/useShortcutRow.ts
```diff
@@ -0,0 +1,54 @@
...
+  return { shortcut: shortcut, zone: DragShortcutZone.Remove, type: DragAndDropType.ToolbarButton }
+}
+
+/** Handles drag end. */
+const endDrag = () => {
+  store.dispatch(dragShortcut(null))
+}
+
+/** A draggable and droppable Thought hook. */
+const useShortcutRow = (props: Partial<ShortcutRowProps>) => {
```
raineorshine, line 36: Calling this `useShortcutRow` doesn't really make sense. It's about adding drag-and-drop functionality to the ShortcutRow.

The comment should also be updated.

src/hooks/useShortcutRow.ts
```diff
@@ -0,0 +1,54 @@
...
+  const propsTypes = props as ShortcutRowProps
```
raineorshine, line 37: Try to find a way to do this without `as`. 

I think `Partial<ShortcutRowProps>` is too broad, and should probably be changed to the specific props that are needed.

raineorshine (changes requested):

src/hooks/useShortcutRow.ts
Zubair286: Hmm, I think `Partial` can be removed here altogether since the `ShortcutRowProps` already has everything as optional besides `shortcut`.

raineorshine (changes requested):
Sorry, I didn't provide the way to test this drag-and-drop functionality!

## Current Behavior

b0fe35b038

https://github.com/user-attachments/assets/b37ab3ac-8fe0-44d5-9e1d-60a94218bad9

## Expected Behavior

`main`

https://github.com/user-attachments/assets/dd079cde-976f-4f3d-82b1-efced28e4fb6

Zubair286:
Correct me if I am wrong but are ShortcutRow and DragAndDroptoolbarButton drag and drop interconnected? If so should I implement both of them and push to this PR. @raineorshine

raineorshine:
They have the same item type. Does that means that they need to be upgraded together? I'm not sure. I'll leave it up to you.

raineorshine (approved):
In general it's better to co-locate code instead of splitting it off into separate files and functions. The entire hook instantiation was less than 20 lines of code, so it wasn't necessary to split it into separate functions either. You can see my more compact implementation.

PR #2236, opened 2024-08-05
2060 text formatting button

BlaineOmega:
Addresses #2060 . ~~Removed the "isActive" check from the shortcut and used the commandStateStore to control the state of the button. There was no need for the isActive function, so I removed it.~~


src/@types/Shortcut.ts
```diff
@@ -48,9 +48,6 @@ interface Shortcut {
   /** A readable, internal unique id. */
   id: ShortcutId
 
-  /** A function that returns true if the shortcut should be highlighted in the Toolbar. */
-  isActive?: (getState: () => State) => boolean
-
```
raineorshine, line 53: `isActive` is used by other (non-formatting) shortcuts, so we cannot remove it.

e.g. Try toggling sort:

https://github.com/cybersemics/em/blob/a33035e2076d82d8114459d662b0508ca353d686/src/shortcuts/toggleSort.ts#L38-L45

src/components/ToolbarButton.tsx
```diff
@@ -49,17 +49,18 @@ const ToolbarButtonComponent: FC<DraggableToolbarButtonProps> = ({
...
     throw new Error('The svg property is required to render a shortcut in the Toolbar. ' + shortcutId)
   }
 
   const commandState = commandStateStore.useSelector(state => state[shortcutId])
+
   const isDraggingAny = useSelector(state => !!state.dragShortcut)
   const dragShortcutZone = useSelector(state => state.dragShortcutZone)
-  const isButtonActive =
-    useSelector(state => (customize ? selected : !isActive || isActive(() => state))) || (!customize && commandState)
+  const isButtonActive = useSelector(state => (customize ? selected : commandState))
```
raineorshine, line 62: You have deleted the `customize` logic. When in toolbar customization mode, the active state is set to `selected`.

You can test this out by going to Settings → Customize Toolbar and selecting a toolbar button.

raineorshine (changes requested):

src/components/ToolbarButton.tsx
BlaineOmega: I fixed this issue. Works for both the customize and the regular tool bar. There logic is much simpler :)

src/@types/Shortcut.ts
BlaineOmega: I added this back to the interface.

raineorshine (approved):

raineorshine:
Looks good, thank you!

PR #2240, opened 2024-08-06
Customize regression

BlaineOmega:
Fixed the regression issue for icon not being in the selected state when the user selects an icon during the customize tool bar flow.


BlaineOmega:
Close because it was fixed in PR #2236

Issue #2128, opened 2024-07-09
Fix Bullet expansion tests


raineorshine:
The bullet expansion tests are currently being skipped because they broke when the component hierarchy changed.

Rewrite the selectors to work with the new component hierarchy. You may need to select based on `.tree-node[data-path]` since the tree-nodes are now rendered flat.

https://github.com/cybersemics/em/blob/5ffafd3a22beb062ff2b55b87115fe898aea7870/src/components/__tests__/Bullet.ts#L132

PR #2242, opened 2024-08-06
Fix Bullet expansion tests

jasonribble:
I changed the assignment values to properly execute the button click on the bullet points.

This isn't my favorite solution; though it has the tests running and that's more important.

Solves #2128

### DIFF
--- a/src/components/__tests__/Bullet.ts
+++ b/src/components/__tests__/Bullet.ts
@@ -130,8 +130,6 @@ describe('render', () => {
 
 describe('expansion', () => {
   it('tapping an expanded cursor bullet should collapse the thought by moving the cursor up', async () => {
-    const unrankedPath = ['a', 'b']
-
     await dispatch([
       importText({
         text: `
@@ -141,10 +139,11 @@ describe('expansion', () => {
           - d
       `,
       }),
-      setCursor(unrankedPath),
+      setCursor(['a', 'b']),
     ])
 
-    const context = contextToPath(store.getState(), unrankedPath)
+    const context = contextToPath(store.getState(), ['a', 'b'])
+
     const pathOfThoughtB = context!.join('')
     const bulletOfThoughtB = screen.getByTestId('bullet-' + pathOfThoughtB)
 
@@ -155,8 +154,6 @@ describe('expansion', () => {
   })
 
   it('tapping the cursor bullet on an ancestor should collapse all descendants', async () => {
-    const unrankedPath = ['x', 'a', 'b', 'c']
-
     await dispatch([
       importText({
         text: `
@@ -167,10 +164,10 @@ describe('expansion', () => {
           - d
       `,
       }),
-      setCursor(unrankedPath),
+      setCursor(['x', 'a', 'b', 'c']),
     ])
 
-    const context = contextToPath(store.getState(), unrankedPath)
+    const context = contextToPath(store.getState(), ['x', 'a', 'b', 'c'])
     const pathOfThoughtA = context?.slice(0, 2).join('')
     const bulletOfThoughtA = screen.getByTestId('bullet-' + pathOfThoughtA)
 
--- a/src/components/__tests__/Bullet.ts
+++ b/src/components/__tests__/Bullet.ts
@@ -1,16 +1,14 @@
-import { screen } from '@testing-library/dom'
 import userEvent from '@testing-library/user-event'
 import { importTextActionCreator as importText } from '../../actions/importText'
 import { toggleHiddenThoughtsActionCreator as toggleHiddenThoughts } from '../../actions/toggleHiddenThoughts'
 import { HOME_TOKEN } from '../../constants'
-import contextToPath from '../../selectors/contextToPath'
 import { exportContext } from '../../selectors/exportContext'
 import store from '../../stores/app'
 import createTestApp, { cleanupTestApp } from '../../test-helpers/createRtlTestApp'
 import dispatch from '../../test-helpers/dispatch'
 import { findCursor } from '../../test-helpers/queries/findCursor'
+import { getBullet } from '../../test-helpers/queries/getBullet'
 import { setCursorFirstMatchActionCreator as setCursor } from '../../test-helpers/setCursorFirstMatch'
-import hashPath from '../../util/hashPath'
 
 beforeEach(createTestApp)
 afterEach(cleanupTestApp)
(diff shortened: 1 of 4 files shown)


raineorshine (changes requested):
This is a good start, thanks.

Selecting all thoughts and then indexing into the array is not ideal, because it couples the test to the linear order of thoughts as they are rendered from top-to-bottom on the screen. If a thought is the 2nd or 4th or 10th from the top, that's not very meaningful to the user. While it's not much of a problem when only rendering a few thoughts, it's going to become a problem for tests that involve a larger number of thoughts, or deeply nested thoughts. Pretty soon you have developers counting lines of code in order to figure out which `thoughtContainers[12]` refers to. And adding one thought at the top changes the indexes all the way down.

Instead, we want to find a way to select a thought based on its text. You can use [contextToPath(['a', 'b'])](https://github.com/cybersemics/em/blob/d90f6562fc3cab351aa6aa70adaf1060d51bad4c/src/selectors/contextToPath.ts) to get the path which can be used to select the exact testid of interest. Or you could select by the visible text itself, or embed the thought value in a data attribute. They each have their tradeoffs, but at least they are not dependent on the linear order.

raineorshine:
FYI I switched the PR to merge into `test-jason` instead of `main`.

jasonribble:
Oh okay, I understand the problem. If we wanted to try additional test causes, it would not expand easily. Not DRY.

> You can use [contextToPath(['a', 'b'])](https://github.com/cybersemics/em/blob/d90f6562fc3cab351aa6aa70adaf1060d51bad4c/src/selectors/contextToPath.ts) to get the path which can be used to select the exact testid of interest.

I see a depreciation warning on `contextToPath`

> FYI I switched the PR to merge into `test-jason` instead of `main`.

Oh okay that makes sense where I should be merging. Thank you!

jasonribble:
Would you prefer this 

```js
  it('tapping an expanded cursor bullet should collapse the thought by moving the cursor up', async () => {
    const unrankedPath = ['a', 'b']

    await dispatch([
      importText({
        text: `
        - a
          - b
            - c
          - d
      `,
      }),
      setCursor(unrankedPath),
    ])

    const context = contextToPath(store.getState(), unrankedPath)

    const testIdOfB = context?.join('')

    const bulletOfthoughtB = screen.getByTestId('bullet-' + testIdOfB)

    userEvent.click(bulletOfthoughtB)

    const thoughtCursor = await findCursor()
    expect(thoughtCursor).toHaveTextContent('a')
  })
```

jasonribble:
> ```js 
>     const testIdOfB = context?.join('')
> ```

There's gotta be a better way to get the `hashPath` of B 🤔.

jasonribble:
A better name for the variable would be `pathOfThoughtB`

src/components/__tests__/Bullet.ts
```diff
@@ -128,9 +128,10 @@ describe('render', () => {
   })
 })
 
-// TODO: findSubthoughts is broken after LayoutTree
-describe.skip('expansion', () => {
+describe('expansion', () => {
   it('tapping an expanded cursor bullet should collapse the thought by moving the cursor up', async () => {
+    const unrankedPath = ['a', 'b']
```
raineorshine, line 133: Prefer inlining this value. `['a', 'b']` is a short, readable value, so in tests it's better not to abstract it into a variable and force the user to move their eyes back-and-forth when reading through the test case.

This differs from the DRY approach that is taken on the codebase itself, where we want to avoid duplication and tightly couple dependent values. Tests on the other hand benefit from being as concrete as possible.

src/components/__tests__/Bullet.ts
```diff
@@ -140,18 +141,22 @@ describe.skip('expansion', () => {
...
       }),
-      setCursor(['a', 'b']),
+      setCursor(unrankedPath),
     ])
 
-    const subthoughts = await findSubthoughts('a')
-    const bulletsOfSubthoughtsA = await findAllByLabelText(subthoughts[0], 'bullet')
-    userEvent.click(bulletsOfSubthoughtsA[0])
+    const context = contextToPath(store.getState(), unrankedPath)
+    const pathOfThoughtB = context!.join('')
```
raineorshine, line 148: Use `hashPath`

src/components/__tests__/Bullet.ts
```diff
@@ -140,18 +141,22 @@ describe.skip('expansion', () => {
...
       `,
       }),
-      setCursor(['a', 'b']),
+      setCursor(unrankedPath),
     ])
 
-    const subthoughts = await findSubthoughts('a')
-    const bulletsOfSubthoughtsA = await findAllByLabelText(subthoughts[0], 'bullet')
-    userEvent.click(bulletsOfSubthoughtsA[0])
+    const context = contextToPath(store.getState(), unrankedPath)
```
raineorshine, line 147: `contextToPath` converts a context to a Path, so it would be more appropriate to name this variable `path`. (The context is the input: `['a', 'b']`)

src/components/__tests__/Bullet.ts
```diff
@@ -162,12 +167,14 @@ describe.skip('expansion', () => {
...
       `,
       }),
-      setCursor(['x', 'a', 'b', 'c']),
+      setCursor(unrankedPath),
     ])
 
-    const subthoughts = await findSubthoughts('x')
-    const bulletsOfSubthoughtsX = await findAllByLabelText(subthoughts[0], 'bullet')
-    userEvent.click(bulletsOfSubthoughtsX[0])
+    const context = contextToPath(store.getState(), unrankedPath)
```
raineorshine, line 173: Get the path of the desired thought directly: `contextToPath(store.getState(), ['x', 'a'])`

raineorshine (changes requested):

src/components/__tests__/Bullet.ts
jasonribble: Okay I understand now thank you!

jasonribble: Interesting! Okay cool. Is the `.join` still okay for this case?

src/components/__tests__/Bullet.ts
jasonribble: oh it just clicked

jasonribble: done !

raineorshine (changes requested):
Looks beautiful!

Once we have a helper function `getBullet` we should be good to merge.

jasonribble:
Okay cool; as mentioned something like `getThought(context: string[])` and `getBullet(context: string[])`

raineorshine (approved):

PR #2243, opened 2024-08-07
updates bullet tests for expansion

Cur1yJ:
## Updates broken expansion tests
 - Updates tests to work without adding additional attributes to frontend component
 - No longer relies on the downstream `findSubthoughts` function, instead, utilizes the index of the payloads to be tested
 - Takes a layered approach, first checking the state, then the DOM
 - Keeps the original goals of the tests in-place

### DIFF
--- a/src/components/__tests__/Bullet.ts
+++ b/src/components/__tests__/Bullet.ts
@@ -1,4 +1,4 @@
-import { findAllByLabelText, screen } from '@testing-library/dom'
+import { screen, waitFor } from '@testing-library/dom'
 import userEvent from '@testing-library/user-event'
 import { importTextActionCreator as importText } from '../../actions/importText'
 import { toggleHiddenThoughtsActionCreator as toggleHiddenThoughts } from '../../actions/toggleHiddenThoughts'
@@ -8,7 +8,6 @@ import store from '../../stores/app'
 import createTestApp, { cleanupTestApp } from '../../test-helpers/createRtlTestApp'
 import dispatch from '../../test-helpers/dispatch'
 import { findCursor } from '../../test-helpers/queries/findCursor'
-import { findSubthoughts } from '../../test-helpers/queries/findSubthoughts'
 import { setCursorFirstMatchActionCreator as setCursor } from '../../test-helpers/setCursorFirstMatch'
 
 beforeEach(createTestApp)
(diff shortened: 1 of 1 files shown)


src/components/__tests__/Bullet.ts
```diff
@@ -228,17 +274,33 @@ describe.skip('expansion', () => {
...
+
+    // Wait for the state to update:
+    await waitFor(() => {
+      const newState = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
+
+      // Expect that the state has changed:
+      expect(newState).not.toEqual(initialState)
+
+      // Expect that the entire "pin" section is removed:
+      expect(newState).not.toContain('- =pin\n  - true')
```
raineorshine, line 293: The component tests should actually be testing what is rendered to the DOM. We already have reducer tests that make assertions about the exported thoughts.

src/components/__tests__/Bullet.ts
```diff
@@ -143,12 +141,26 @@ describe.skip('expansion', () => {
...
-    userEvent.click(bulletsOfSubthoughtsA[0])
+    const initialState = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
 
-    const thoughtCursor = await findCursor()
-    expect(thoughtCursor).toHaveTextContent('a')
+    // Find all bullet elements:
+    const bulletElements = await screen.findAllByLabelText('bullet')
+
+    // Click the second bullet directly:
+    await userEvent.click(bulletElements[1]) // Index 1 for the second bullet
```
raineorshine, line 150: Selecting all bullets and then indexing into the array is not ideal, because it couples the test to the linear order of thoughts as they are rendered from top-to-bottom on the screen. If a thought is the 2nd or 4th or 10th from the top, that's not very meaningful to the user. While it's not much of a problem when only rendering a few thoughts, it's going to become a problem for tests that involve a larger number of thoughts, or deeply nested thoughts. Pretty soon you have developers counting lines of code in order to figure out which `thoughtContainers[12]` refers to. And adding one thought at the top changes the indexes all the way down.

Instead, we want to find a way to select a thought based on its text. You can use [contextToPath(['a', 'b'])](https://github.com/cybersemics/em/blob/d90f6562fc3cab351aa6aa70adaf1060d51bad4c/src/selectors/contextToPath.ts) to get the path which can be used to select the exact testid of interest. Or you could select by the visible text itself, or embed the thought value in a data attribute. They each have their tradeoffs, but at least they are not dependent on the linear order.

raineorshine (changes requested):

PR #2244, opened 2024-08-07
Migrate ToolbarButton higher order component to hooks

Zubair286:
#1839


raineorshine (changes requested):
Tested and works well.

Let's clean up the types like we did with the other drag-and-drop components. 

- Remove `Partial` and `as`
- Only define the exact props that are needed. 
- Do not reuse `ToolbarButtonProps` if possible. 
- Inline trivial functions. `drop` is pretty much the only other function that needs to be factored out.

This hook won't be quite as compact as the one in ShortcutRow, but it's still a good reference: https://github.com/cybersemics/em/commit/25de2ee792

src/components/modals/CustomizeToolbar.tsx
```diff
@@ -149,7 +151,7 @@ const ModalCustomizeToolbar: FC = () => {
     >
       <h1 className='modal-title'>Customize Toolbar</h1>
       <p style={{ marginTop: '-1em', marginBottom: '1em' }}>
-        &lt;{' '}
+        &lt{' '}
```
raineorshine: I think this got changed by accident. I fixed it.

src/components/modals/CustomizeToolbar.tsx
```diff
@@ -58,74 +59,75 @@ const drop = (props: unknown, monitor: DropTargetMonitor) => {
...
-
-    return dropTarget(<div>{children}</div>)
-  },
-)
+const DropToRemoveFromToolbar = ({ children }: { children: React.ReactNode }) => {
+  const [{ isHovering, sourceZone }, dropTarget] = useDrop({
+    accept: [DragAndDropType.ToolbarButton, NativeTypes.FILE],
+    drop: (item, monitor) => drop({}, monitor),
+    collect: dropCollect,
+  })
```
raineorshine: This looks good.

Only one that looks off is `drop: (item, monitor) => drop({}, monitor)`. Why are you passing an empty object an an `unknown` param that is never used?

I will remove the unused param.

src/hooks/useDragAndDropToolbarButton.ts
```diff
@@ -19,46 +18,21 @@ import { shortcutById } from '../shortcuts'
 import store from '../stores/app'
 import appendToPath from '../util/appendToPath'
 
-export type DraggableToolbarButtonProps = ToolbarButtonProps & {
+type DraggableToolbarButtonProps = {
   path: Path
   simplePath: SimplePath
```
raineorshine: `path` and `simplePath` are not used by `useDragAndDropToolbarButton` and should be removed.

raineorshine (approved):

src/components/modals/CustomizeToolbar.tsx
Zubair286: Thanks, I was going to remove it guess I forgot, I only added it at the time because I thought the `item` parameter was important so I left it there.

Issue #2144, opened 2024-07-12
Fix toggleSort tests


raineorshine:
The toggleSort tests are currently being skipped because they broke when the component hierarchy changed.

Rewrite the selectors to work with the new component hierarchy. You may need to select based on `.tree-node[data-path]` since the tree-nodes are now rendered flat.

https://github.com/cybersemics/em/blob/f68689f6471af07ba87db97cdd9b68bf42cdc910/src/shortcuts/__tests__/toggleSort.ts

PR #2268, opened 2024-08-12
Fix toggleSort tests 

jasonribble:
- Enable DOM local tests and updated to match new Thought flatten tree structure
- Rename `getBullet` to `getBulletByContext` to be more descriptive
- Create `getThoughtByContext(string[])`
- Create `getSubthoughtsByContext`
- Skipping Global tests for now

Solves #2144

### DIFF
--- a/src/test-helpers/queries/getSubthoughtsByContext.ts
+++ b/src/test-helpers/queries/getSubthoughtsByContext.ts
@@ -1,0 +1,11 @@
+import { screen } from '@testing-library/dom'
+import contextToPath from '../../selectors/contextToPath'
+import store from '../../stores/app'
+
+/** Get the bullet of a thought with the context. */
+export function getSubthoughtsByContext(context: string[]) {
+  const pathOfThoughtA = contextToPath(store.getState(), context)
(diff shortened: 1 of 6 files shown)


src/test-helpers/queries/getSubthoughtsByContext.ts
```diff
@@ -0,0 +1,11 @@
+import { screen } from '@testing-library/dom'
+import contextToPath from '../../selectors/contextToPath'
+import store from '../../stores/app'
+
+/** Get a list of subthoughts of a thought with the context. */
+export function getSubthoughtsByContext(context: string[]) {
```
raineorshine, line 6: Technically this returns all descendants.

jasonribble: Ahh that's a good point 🤔

jasonribble: Either I get rename the function or update it to only get it to a certain character limit (so it is only one level)

raineorshine: I already merged, but I'm open to renaming it. 

We don't really want to add a dependency on the id character length.

PR #2269, opened 2024-08-12
Bump react-dnd from version 13.0.0 to 14.0.0

Zubair286:


raineorshine (approved):
Interesting, I guess they backtracked on the hook API?

Zubair286:
Exactly my thoughts, why bother doing something when you're gonna revert it all in the next major update.

Issue #2253, opened 2024-08-10
Allow right mouse click on thought


raineorshine:
## Steps to Reproduce

Right mouse click on a thought.

## Current Behavior

Nothing happens, and the browser selection disappears.

## Expected Behavior

The context menu that shows up normally on right mouse click should appear.

## Testing

Add a puppeteer snapshot that shows the context menu open.

PR #2286, opened 2024-08-14
fix: Allow right mouse click on thought. 

Krishna2323:
### Details
- Browser's default context menu wasn't opened when right clicking on a thought. 

### Fixed Issue
- https://github.com/cybersemics/em/issues/2253

### RCA
 - Currently we only prevent default when the condition below is matched.
https://github.com/cybersemics/em/blob/31a13f0160fbf65727b4b06c3800ae0490437830/src/hooks/useLongPress.ts#L112-L117

### Changes
- Instead of always calling `e.preventDefault()`, we can call it when `lock` is true, lock will be true when long press event occurs.

### Screenshots/Videos
<details>
<summary>Android: mWeb Chrome</summary>

https://github.com/user-attachments/assets/3946fdf2-d021-4e74-a18f-d34aac2f3277

</details>

<details>
<summary>iOS: mWeb Safari</summary>

https://github.com/user-attachments/assets/a556edfa-65f3-4780-94e8-009ff43a232d

</details>

<details>
<summary>Chrome / Safari</summary>

https://github.com/user-attachments/assets/0f0886e9-708b-417a-9af4-e11956d85821

</details>


raineorshine (changes requested):
Thanks for your contribution. 

Based on the logic, it looks the selection still gets cleared on right click on desktop (untested). Can you confirm? We don't want to clear the selection on desktop. 

In fact, we don't want to use the long press behavior at all on right mouse click. It seem strange to me that there is no code that actually checks the mouse button.

raineorshine:
Exactly. Thanks :)

raineorshine:
Thanks. I'll have to test it myself to confirm the desired behavior. I’m traveling this week, but will review as soon as I return to work.

raineorshine (changes requested):
Okay, I've tested it out on desktop and here's what I'm seeing:

✓ Right-clicking on selected text opens the context menu.
✗ Right-clicking on the bullet clears the browser selection.
✗ Right-clicking on the bullet initiates a drag-and-drop.

No need for a custom context menu, and I don't care whether the context menu works or not on the bullet as long as the above two points are fixed. Thanks!

src/hooks/useLongPress.ts
```diff
@@ -108,9 +108,12 @@ const useLongPress = (
   // Web passes React.MouseEvent
   const onContextMenu = useCallback(
     (e: React.MouseEvent | React.PointerEvent) => {
+      // If lock is true (longpress event occurs) we should prevent opening context menu
+      if (lock) {
+        e.preventDefault()
+      }
```
raineorshine, line 114: Why piggyback on `lock` instead of just checking for right-mouse click? If we're trying to change the behavior only under the condition that the right-mouse button is clicked, it seems like we should test for that directly, unless I'm missing something. Tying it to `lock` tightly couples the right-mouse click behavior to the unrelated lock behavior which was intended to prevent a long press from more than one component at a time.

raineorshine: Try rebasing on `main`. The `packageManager` was added and page-lifecycle was fixed in 10b9d7c90faca1aec451bb104386269b4be53c11.

raineorshine:
It turns out that we mainly just needed to remove the `!isTouch ||` condition in order to avoid disabling the context menu on desktop. This obviates the need to save and restore the selection.

See a2ba727f93

Thank you anyway!

Issue #2267, opened 2024-08-11
Set text color of selection


raineorshine:
## Steps to Reproduce

1. Select part of the text of a thought. 
2. Select a text color. 

## Current Behavior

The entire thought changes color.

![image](https://github.com/user-attachments/assets/1ae14c5e-4d2e-4605-90a2-00e23f355fc0)

## Expected Behavior

Only the selection should change color. 

## Notes

The color formatting implementation needs to change from meta attributes to inline styles.

raineorshine:
Certainly! Happy to have your help.

The following milestones are currently assigned to others, but anything else is fair game.

- https://github.com/cybersemics/em/milestone/18
- https://github.com/cybersemics/em/milestone/21
- https://github.com/cybersemics/em/milestone/30
- https://github.com/cybersemics/em/milestone/31

Let me know if you have any questions about the codebase or how to get started.

raineorshine:
In general you can get started without being assigned, but it's good to check in with me if the instructions are unclear, or if you want to make sure you're taking the right approach.

Note that the above list of milestones are the ones **not** available to work on. They are already assigned to other developers (even if the individual issues say they are not assigned...).

PR #2289, opened 2024-08-15
fix Set text color of selection

MOONSTAR1220:
This pull request implements https://github.com/cybersemics/em/issues/2267.
Fix the setting color to current selection of thought.
Screen


raineorshine:
Thanks! Please fix lint errors and tests, and I'll review. 

We will need to add unit tests to cover the new functionality. An RTL test should allow you to set a browser selection and then set the text color. 

If that doesn't work, we can create a Puppeteer test, but RTL is preferred.

raineorshine (changes requested):
Thanks! Please fix lint errors and tests, and I'll review. 

We will need to add unit tests to cover the new functionality. An RTL test should allow you to set a browser selection and then set the text color. 

If that doesn't work, we can create a Puppeteer test, but RTL is preferred.

MOONSTAR1220:
@raineorshine , now, I'm fixing the lint and test issues.
I'll let you know when it's done.
And for about unit test, I will check about that.
Thanks.

MOONSTAR1220:
Hello @raineorshine . I tried to implement Unit test using RTL.
But there is a problem now.
I used document.execCommand to set the text color and background color.
And document.execCommand can't be used in  jsdom environment used by testing libraries like Vitest, Jest, or similar environments.
I have checked the project and there is no unit test Bold, Italic and other functions 
Please check and let me know the result.
Thanks.

src/util/getCommandState.ts
```diff
@@ -87,6 +89,25 @@ const getCommandState = (value: string): CommandState => {
       matches[command] &&= currentCommandState[command]
     }
   }
+  if (savedValue.length > 0) {
+    const parser = new DOMParser() // new DomParser
```
raineorshine, line 93: `getCommandState` is called on every keystroke, so it's highly performance sensitive. Instantiating a new `DOMParser` on every keystroke is too heavy a solution unfortunately.

We could maybe use this approach if we throttle it.

raineorshine (changes requested):

src/util/getCommandState.ts
MOONSTAR1220: Ok, I will check about that.

MOONSTAR1220: `if (savedValue.length > 0 && (savedValue.match(fontRegex) || savedValue.match(spanRegex)))`

I think if we use this, it will be solved.
**DOMParser** will be created when needed.

raineorshine: That helps, but the problem still exists when color is used. We either need a very efficient parser that is tailor-made for parsing color, or we need to throttle this at least to the next animation frame.

MOONSTAR1220: @raineorshine , I have fixed getCommandState function without using **DOMParser**.
Please check and let me know.

MOONSTAR1220: @raineorshine , What do you think about the current getCommandState function?
I removed DOMParser.

raineorshine: Will take a look tomorrow! I was at a conference this week.

MOONSTAR1220: Ok, I got it.

src/stores/commandStateStore.ts
```diff
@@ -11,6 +11,8 @@ const commandStateStore = reactMinistore<CommandState>({
   italic: false,
   underline: false,
   strikethrough: false,
+  foreColor: 'rgb(227, 227, 227)',
+  backColor: 'rgb(0, 0, 0)',
```
raineorshine, line 15: We can't hardcode these colors in, as they depend on whether the Light or Dark theme is active.

You should probably be one of `false`, `null`, or `undefined` by default.

src/util/__tests__/stripStyleAttribute.ts
```diff
@@ -5,14 +5,14 @@ it('only allow font weight, font-style, and text-decoration', () => {
     stripStyleAttribute(
       'color: red; background-color: white; font-weight: bold; font-size: 14px; font-style: italic; text-decoration: underline;',
     ),
-  ).toBe('font-weight: bold;font-style: italic;text-decoration: underline;')
+  ).toBe('color: red;background-color: white;font-weight: bold;font-style: italic;text-decoration: underline;')
```
raineorshine, line 8: Please update the test title to reflect the new behavior.

src/util/formattingNodeToHtml.ts
```diff
@@ -3,17 +3,38 @@ import _ from 'lodash'
 import stripStyleAttribute from './stripStyleAttribute'
 
 /** Strip span and encode a <i> or <b> element as HTML. */
+const replaceFontToSpan = (node: Element) => {
+  if (node.tagName === 'font') {
```
raineorshine, line 7: Mutating arguments is considered a bad practice. Please find a way to do this in a more functional style.

src/util/formattingNodeToHtml.ts
```diff
@@ -3,17 +3,38 @@ import _ from 'lodash'
...
   const content: string = node.children.reduce((acc, child) => {
     return acc + (child.type === 'text' ? child.content : child.type === 'comment' ? '' : formattingNodeToHtml(child))
   }, '')
-
+  replaceFontToSpan(node)
   if (node.tagName === 'span') {
-    const styleAttribute = _.find(node.attributes, { key: 'style' })
-    if (!styleAttribute) {
+    const styleAttributes = _.filter(node.attributes, { key: 'style' })
+    if (!styleAttributes) {
```
raineorshine, line 29: This condition will always be false, since `_.filter` never returns null or undefined.

src/util/formattingNodeToHtml.ts
```diff
@@ -3,17 +3,38 @@ import _ from 'lodash'
...
       return content
     }
-    const strippedStyleAttribute = stripStyleAttribute(styleAttribute.value)
+    let strippedStyleAttribute = ''
+    styleAttributes.forEach(styleAttribute => {
+      const strippedAttribute = stripStyleAttribute(styleAttribute.value)
+      if (strippedAttribute.length > 0) strippedStyleAttribute += strippedAttribute
+    })
```
raineorshine, line 36: Please use a map and join instead of mutation. Mutation is generally not recommended.

src/util/getCommandState.ts
```diff
@@ -16,6 +16,27 @@ const tags = {
...
+ * Returns an object with foreColor and backColor properties.
+ * If the string does not contain a font or span tag, the default colors are returned.
+ */
+const extractColors = (savedValue: string) => {
+  const foreColorRegex = /<font[^>]*\scolor=["']?([^"']+)["']?[^>]*>/i
+  const backColorRegex = /<span[^>]*\sstyle=["'][^"']*background-color:\s*([^;"']+)/i
+
+  // Attempt to extract the font color
+  const foreColorMatch = savedValue.match(foreColorRegex)
+  const foreColor = foreColorMatch ? foreColorMatch[1].trim() : 'rgb(227, 227, 227)'
```
raineorshine, line 33: The specific foreground/background color we use will depend on the current theme, so we can't hardcode these here unfortunately.

src/util/getCommandState.ts
```diff
@@ -16,6 +16,27 @@ const tags = {
   italic: createTag('i'),
   underline: createTag('u'),
   strikethrough: createTag('strike'),
+  foreColor: createTag('font'),
+  backColor: createTag('span'),
```
raineorshine, line 20: Why does foreColor use `font` and backColor use  `span`?

raineorshine (changes requested):
Thank you! Below you'll find a complete review.

You did a good job fitting into the existing structure, modifying `stripStyleAttribute` and using the `execCommand` infrastructure that was already in place. I didn't know about `foreColor` and `backColor`, so that's great that we can implement this with close to the same logic already used for bold, italic, underline, and strikethrough.

We will need to add puppeteer tests since neither jsdom nor happy-dom supports `document.execCommand`.

It works well, but for one aspect of the behavior: foreground and background colors should not be mixed and matched. Take a look at closer look at the existing functionality in `main`. Each swatch is a foreground+background combo. If you select blue text, the background color is automatically removed. If you select a blue background, the foreground color is automatically removed. This limits the combinations and ensures proper color contrast. Let me know if that needs any clarification.

Otherwise, they were mostly small suggestions, either technical or stylistic. Watch out for code duplication and avoid mutations if at all possible.

Thanks, and let me know if you have any questions!

src/util/getCommandState.ts
raineorshine: That looks much better. If we run into performance issues, then it's a problem with the parsing logic of `getCommandQuery` and can be addressed in a separate issue. This isn't adding major performance overhead any more.

MOONSTAR1220: Hello @raineorshine , When will you check the current functionalities?

raineorshine: https://github.com/cybersemics/em/pull/2289#pullrequestreview-2245841057

src/util/getCommandState.ts
MOONSTAR1220: Because document.execCommand('foreColor') adds <font> tag and document.execCommand('backColor') adds <span> tag.

raineorshine: Can we either use a single span tag with both styles, or two span tags? It seems pretty hacky to mix `font` and `span`.

MOONSTAR1220: Ok, I will check.

src/actions/formatSelection.ts
```diff
@@ -26,10 +29,13 @@ export const formatSelectionActionCreator =
...
+      if (command === 'backColor' || command === 'foreColor') document.execCommand(command, false, color)
+      else document.execCommand(command)
       selection.restore(savedSelection)
     } else {
-      document.execCommand(command)
+      // Apply text and background color to the selected text and apply the command
+      if (command === 'backColor' || command === 'foreColor') document.execCommand(command, false, color)
+      else document.execCommand(command)
       updateCommandState()
     }
```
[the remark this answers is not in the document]
MOONSTAR1220: First one is execute the command to format the entire thought(when nothing is selected) and the second one is to format the selected thought.

MOONSTAR1220:
Thanks for your clear explanation. I will fix this and let you know.

src/actions/formatSelection.ts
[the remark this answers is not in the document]
raineorshine: There is a lot of duplicate code between the two cases. See if you can reorganize it to remove the duplication (without changing the behavior).

MOONSTAR1220:
Hello @raineorshine. I have fixed the feedbacks you mentioned.
Please check and let me know the result.
And for about the pupeteer test, now I'm implementing that.
Thanks.

MOONSTAR1220:
@raineorshine , I have added the initial puppeteer test too.
Please check and let me know the result.
Thanks.

src/util/formattingNodeToHtml.ts
```diff
@@ -3,17 +3,41 @@ import _ from 'lodash'
 import stripStyleAttribute from './stripStyleAttribute'
 
 /** Strip span and encode a <i> or <b> element as HTML. */
+const replaceFontToSpan = (node: Element) => {
```
raineorshine, line 6: Looks good, thank you.

src/util/formattingNodeToHtml.ts
```diff
@@ -3,17 +3,41 @@ import _ from 'lodash'
 import stripStyleAttribute from './stripStyleAttribute'
 
 /** Strip span and encode a <i> or <b> element as HTML. */
```
raineorshine, line 5: Update comment

src/util/formattingNodeToHtml.ts
```diff
@@ -3,17 +3,41 @@ import _ from 'lodash'
...
+  return node
+}
+
+/** Converts a formatting node to HTML. */
 const formattingNodeToHtml = (node: Element) => {
   const content: string = node.children.reduce((acc, child) => {
     return acc + (child.type === 'text' ? child.content : child.type === 'comment' ? '' : formattingNodeToHtml(child))
   }, '')
-
+  node = replaceFontToSpan(node)
```
raineorshine, line 32: Please create a new variable rather than mutating the old one. 

It's never a good idea to mutate arguments, as they are meant to be inputs only.

src/util/getCommandState.ts
```diff
@@ -16,6 +16,27 @@ const tags = {
   italic: createTag('i'),
   underline: createTag('u'),
   strikethrough: createTag('strike'),
+  foreColor: createTag('span'),
+  backColor: createTag('span'),
+}
+
+/** Extracts the foreground and background colors from the given string.
+ * Returns an object with foreColor and backColor properties.
+ * If the string does not contain a font or span tag, the default colors are returned.
```
raineorshine, line 25: If the string does not contain a font or span tag, undefined is returned.

src/e2e/puppeteer/__tests__/color.ts
```diff
@@ -0,0 +1,35 @@
+import helpers from '../helpers'
```
raineorshine, line 1: This looks good.

There are few other tests we need to port over from the old RTL test:

- Clear the text color when selecting white
- Clear background color when selecting text color
- Change color to black when setting background color (this is broken)

raineorshine (changes requested):
Thanks! It's coming along well.

The selection should be preserved when the color is set (just like when using bold, italic, underline, and strikethrough).

You may need to [save and restore](https://github.com/cybersemics/em/blob/9cce8433e479a8c30ccd3496733ecb33b65f50da/src/device/selection.ts#L181-L208) the selection in the action-creator.

raineorshine:
Note that there are some linting errors. (Is your pre-push hook working?)

```sh
$ ls .hooks
post-commit* pre-push*
$ cat .hooks/pre-push
npm run lint
```

MOONSTAR1220:
@raineorshine
Thanks for your review.
I'm traveling right now due to urgent personal matters.
Is it ok to fix above issues next week? Really sorry for that.

raineorshine:
Of course! No problem.

MOONSTAR1220:
Hello @raineorshine .
I'm back.
I will fix the feedbacks and let you know when it's done.
Thanks.

raineorshine:
Thanks, let me know if any questions come up :).

MOONSTAR1220:
I got sick during my long trip.
But I tried to fix these problems first, but I couldn't and now I'm in the hospital.
The doctor said I'll recover in 5-10 days.
Is it okay if I fix the problems after that?
I'll solve them as soon as I get out of the hospital.
Really sorry for that.

raineorshine:
Of course! So sorry to hear you got sick. Hope you get well soon.

MOONSTAR1220:
Hello @raineorshine . I'm back and now I am fixing the puppeteer test code now.
I will let you know when it's done.
Thanks.

MOONSTAR1220:
> Welcome back!
> 
> There have been a lot of changes to `main`, so you will probably run into some merge conflicts.

MOONSTAR1220:
I got it. thanks.

MOONSTAR1220:
Hello @raineorshine. Can we take a call?
I have a problem in puppeteer test.
I would like to discuss about this via call.

raineorshine:
Hey! Sorry, I handle everything through GitHub and Discord. I've found it's beneficial to have a written record that can be referred to in the future, and it's easier to use asynchronous communication as I work with many developers across different time zones. 

You are welcome to post the issue you are running into on the Discord, and either me, Trevin, or Fabian will respond: https://discord.gg/US8A7ybA

src/actions/formatSelection.ts
```diff
@@ -21,15 +20,12 @@ export const formatSelectionActionCreator =
       const thoughtContentEditable = document.querySelector(`[aria-label="editable-${thought.id}"]`)
       if (!thoughtContentEditable) return
 
-      const savedSelection = selection.save()
```
raineorshine, line 24: Removing this seems to be the cause of the selection restoration regression.

src/actions/formatSelection.ts
```diff
@@ -1,13 +1,12 @@
...
 import pathToThought from '../selectors/pathToThought'
 import { updateCommandState } from '../stores/commandStateStore'
 import suppressFocusStore from '../stores/suppressFocus'
 
 /** Format the browser selection or cursor thought as bold, italic, strikethrough, underline. */
 export const formatSelectionActionCreator =
-  (command: 'bold' | 'italic' | 'strikethrough' | 'underline'): Thunk =>
-  (dispatch, getState) => {
+  (command: 'bold' | 'italic' | 'strikethrough' | 'underline' | 'foreColor' | 'backColor', color: string = ''): Thunk =>
+  async (dispatch, getState) => {
```
raineorshine, line 9: Any reason you added `async`? I don't see any awaits that would necessitate it.

src/actions/textColor.ts
```diff
@@ -13,29 +14,17 @@ const textColor = (
...
-      ? setDescendant({
-          path,
-          values: ['=styleAnnotation', 'backgroundColor', backgroundColor === 'inverse' ? colors.fg : backgroundColor],
-        })
-      : deleteAttribute({ path, values: ['=styleAnnotation', 'backgroundColor'] }),
-  ])(state)
+  const thought = pathToThought(state, state.cursor)
+  const thoughtText = thought.value.replace(/<[^>]*>/g, '')
+  // set bullet to text color when the entire thought selected
+  if ((selection.text()?.length === 0 && thoughtText.length !== 0) || selection.text()?.length === thoughtText.length) {
```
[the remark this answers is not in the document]
raineorshine: 👀

src/components/ColorPicker.tsx
```diff
@@ -42,14 +43,45 @@ const ColorSwatch: FC<{
...
+      : undefined
+
+    return !!(
+      (textHexColor && textHexColor === selectedTextHexColor) ||
+      (backHexColor && backHexColor === selectedBackHexColor)
+    )
+  }, [cursorStyle, backgroundColor, color])
   /** Toggles the text color to the clicked swatch. */
-  const toggleTextColor = (e: React.MouseEvent | React.TouchEvent) => {
+  const toggleTextColor = async (e: React.MouseEvent | React.TouchEvent) => {
```
raineorshine, line 67: I don't think `async` is needed here.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +43,45 @@ const ColorSwatch: FC<{
...
     // stop toolbar button dip
     e.stopPropagation()
+
+    // Apply text color to the selection
+    if (selected) dispatch(formatSelection('foreColor', colors.fg))
+    else if (backgroundColor || color !== 'default') dispatch(formatSelection('foreColor', color || 'rgba(0, 0, 0, 1)'))
```
raineorshine, line 73: Default color should be retrieved from the theme rather than hard-coded.

raineorshine (changes requested):
Overall it's looking very good. Code style has improved, and tests look good. Thanks!

A couple behavior issues:

# 1. Background color incorrectly set to white

## Steps to Reproduce

1. Create thought "Hello world".
2. Set the cursor on the thought OR select part of the thought text.
3. Set the foreground color.

## Current Behavior

The background color changes to white.

## Expected Behavior

The background color should not be set (or should be removed if it is already set) when setting the foreground color.

# 2. Selection added to whole thought

## Steps to Reproduce

1. Create thought "Hello world".
2. Set the selection in between "Hello" and "world" (collapsed selection).
3. Activate a formatting command or set the text color.

## Current Behavior

All of the cursor thought's text is selected.

This affects bold, italic, underline, and strikethrough as well.

## Expected Behavior

The selection should not change.

- If nothing is selected before, nothing should be selected after the command.
- If there is a collapsed selection before, there should be a collapsed selection at the same offset after the command.
- If text is selected before, the same text should be selected after the command.

src/actions/textColor.ts
```diff
@@ -2,46 +2,41 @@ import _ from 'lodash'
...
-          values: ['=styleAnnotation', 'backgroundColor', backgroundColor === 'inverse' ? colors.fg : backgroundColor],
-        })
-      : deleteAttribute({ path, values: ['=styleAnnotation', 'backgroundColor'] }),
-  ])(state)
+  // set bullet to text color when the entire thought selected
+  const newState =
+    fullySelected && ((color && color !== 'default') || (backgroundColor && backgroundColor !== 'inverse'))
+      ? [setDescendant({ path, values: ['=bullet', '=style', 'color', backgroundColor! || color!] })]
+      : [deleteAttribute({ path, values: ['=bullet', '=style', 'color'] })]
+  return reducerFlow(newState)(state)
```
raineorshine, line 27: There's no need to use `reducerFlow` any more, as we are no longer composing multiple reducers. You can call the reducers directly:

```ts
return fullySelected && ((color && color !== 'default') || (backgroundColor && backgroundColor !== 'inverse'))
      ? setDescendant(state, { path, values: ['=bullet', '=style', 'color', backgroundColor! || color!] })
      : deleteAttribute(state, { path, values: ['=bullet', '=style', 'color'] })
```

raineorshine (changes requested):
I can confirm that the background color is working as expected now. Thank you.

The selection behavior is not quite right. Carefully read the requirements again and try out the bold/italic/underline commands to see the expected behavior.

> The selection should not change.
> 
> - If nothing is selected before, nothing should be selected after the command.
> - If there is a collapsed selection before, there should be a collapsed selection at the same offset after the command.
> - If [some] text is selected before, the same text should be selected after the command.

raineorshine (changes requested):
# 3. Default color is not removed

## Steps to Reproduce

1. Create a thought.
2. Set the text color to red.
3. Set the color back to white.

## Current Behavior

The thought has embedded 

```
<font style="background-color: rgb(0, 0, 0);" color="#ffffff">Hello world</font>
```

## Expected Behavior

When the text color or background color is reset to the default, it should be removed completely.

- When the Dark theme is active, white text color and black background color should be removed.
- When the Light theme is active, black text color and white background color should be removed.

```
Hello world
```

src/components/ColorPicker.tsx
```diff
@@ -42,14 +43,40 @@ const ColorSwatch: FC<{
...
+    return !!(
+      (textHexColor && textHexColor === selectedTextHexColor) ||
+      (backHexColor && backHexColor === selectedBackHexColor)
+    )
+  }, [cursorStyle, backgroundColor, color])
   /** Toggles the text color to the clicked swatch. */
   const toggleTextColor = (e: React.MouseEvent | React.TouchEvent) => {
     // stop toolbar button dip
-    e.stopPropagation()
+    // e.stopPropagation()
```
raineorshine, line 68: # 4. Allowing the click event to propagate breaks the dip behavior

The dip is the small animation that occurs when a toolbar button is pressed.

## Steps to Reproduce

1. Set the cursor on a thought.
2. Set the text color.

## Current Behavior

af1f1757e1

The text color toolbar button remains dipped.

Brave Browser 2024-09-28 08 38 36


## Expected Behavior

`main`

The animation should complete and the button should not dip down.

Brave Browser 2024-09-28 08 38 11

raineorshine (changes requested):
I can confirm that a collapsed browser selection is now preserved. Thank you.

> When the text color or background color is reset to the default, it should be removed completely.

This is working for the default case and the background swatch case. 

When setting the text color in Dark mode, `background-color: rgb(0, 0, 0);` should not be set, as it is redundant. The same for Light mode, but with a white background color.

# 5. Active swatch not selected on blur

## Steps to Reproduce

1. Create a thought.
2. Blur the browser selection by clicking at the bottom of the screen, below all thoughts.
3. Set the text color.

## Current Behavior

The thought changes color but the swatch is not selected:

Brave Browser 2024-09-28 08 50 19

## Expected Behavior

`main`

The correct color swatch is selected:

image

src/components/ToolbarButton.tsx
```diff
@@ -165,7 +165,7 @@ const ToolbarButton: FC<ToolbarButtonProps> = ({
         // offset top to avoid changing container height
         // marginBottom: isPressing ? -10 : 0,
         // top: isButtonExecutable && isPressing ? 10 : 0,
-        transform: `translateY(${isButtonExecutable && isPressing && !longPress.isPressed && !isDragging ? 0.25 : 0}em`,
+        transform: `translateY(${isButtonExecutable && isPressing && !longPress.isPressed && !isDragging && !showColorPicker ? 0.25 : 0}em`,
```
raineorshine, line 168: I would highly recommend leaving this as-is. The toolbar press uses very carefully designed behavior implemented in #2214.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +47,61 @@ const ColorSwatch: FC<{
...
+    return !!(
+      (textHexColor && textHexColor === selectedTextHexColor) ||
+      (backHexColor && backHexColor === selectedBackHexColor)
+    )
+  }, [cursorStyle, backgroundColor, color])
+
   /** Toggles the text color to the clicked swatch. */
   const toggleTextColor = (e: React.MouseEvent | React.TouchEvent) => {
-    // stop toolbar button dip
-    e.stopPropagation()
```
raineorshine, line 52: Removing this is causing too many problems with the click behavior. Let's take a step back and consider other options. What are you trying to accomplish?

src/actions/formatSelection.ts
```diff
@@ -6,32 +7,34 @@ import suppressFocusStore from '../stores/suppressFocus'
 
 /** Format the browser selection or cursor thought as bold, italic, strikethrough, underline. */
 export const formatSelectionActionCreator =
-  (command: 'bold' | 'italic' | 'strikethrough' | 'underline'): Thunk =>
+  (
+    command: 'bold' | 'italic' | 'strikethrough' | 'underline' | 'foreColor' | 'backColor' | 'removeFormat',
+    color: string = '',
+    colors: any = {},
```
raineorshine, line 13: Can you explain the difference between `color` and `colors`? Setting a proper type will help with the understanding as well.

src/actions/formatSelection.ts
```diff
@@ -6,32 +7,34 @@ import suppressFocusStore from '../stores/suppressFocus'
 
 /** Format the browser selection or cursor thought as bold, italic, strikethrough, underline. */
 export const formatSelectionActionCreator =
-  (command: 'bold' | 'italic' | 'strikethrough' | 'underline'): Thunk =>
+  (
+    command: 'bold' | 'italic' | 'strikethrough' | 'underline' | 'foreColor' | 'backColor' | 'removeFormat',
```
raineorshine, line 11: What does `removeFormat` do? I don't see it referenced elsewhere.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +47,61 @@ const ColorSwatch: FC<{
...
   size = size || fontSize * 1.2
 
+  /** A function that adds an alpha channel to a hex color. */
+  const addAlphaToHex = (hex: string) => {
+    if (hex.length === 7) return hex + 'ff'
+    return hex
+  }
+  const colorRegex = /color="(#[0-9a-fA-F]{6})"/g
+  const bgColorRegex = /background-color:\s*(rgb\(\d{1,3},\s?\d{1,3},\s?\d{1,3}\))/g
+  const selected = useMemo(() => {
```
raineorshine, line 60: Once we get the behavior correct (namely **6. Default background color is not removed**), we should look at how to improve this section. First, it could use some comments to help clarify what it's doing since there's a lot going on. Second, the use of loops and mutations is adding quite a bit of complexity. Maybe we can find a way to simplify and streamline this so that it is easier to reason about.

raineorshine (changes requested):
Thanks, I can confirm that **5. Active swatch not selected on blur** has been fixed.

- **4. Allowing the click event to propagate breaks the dip behavior** - The dip animation is working when the Color Picker is closed, but broken when the Color Picker is open. When clicking on the toolbar button in this case, it does not animate.

# 6. Default background color is not removed

## Steps to Reproduce

Set the text color on a thought.

## Current Behavior

```html
<font color="#00c7e6" style="background-color: rgb(0, 0, 0);">Hello world</font>
```

## Expected Behavior

```html
<font color="#00c7e6">Hello world</font>
```

The general rule is that:

- When the Dark theme is active, white text color and black background color should be removed.
- When the Light theme is active, black text color and white background color should be removed.

This behavior can be observed on `main`.

src/actions/formatSelection.ts
```diff
@@ -6,32 +6,27 @@ import suppressFocusStore from '../stores/suppressFocus'
...
     suppressFocusStore.update(true)
-
     // if there is no selection, format the entire thought by selecting the whole thought
+    const thoughtContentEditable = document.querySelector(`[aria-label="editable-${thought.id}"]`)
+    if (!thoughtContentEditable) return
     if (sel?.toString().length === 0 && thought.value.length !== 0) {
-      const thoughtContentEditable = document.querySelector(`[aria-label="editable-${thought.id}"]`)
-      if (!thoughtContentEditable) return
-
+      // const savedSelection = selection.save()
```
raineorshine, line 20: You can remove this commented out code if it's no longer needed.

src/actions/formatSelection.ts
```diff
@@ -6,32 +6,27 @@ import suppressFocusStore from '../stores/suppressFocus'
...
       const savedSelection = selection.save()
-
       // must suppress focus events in the Editable component, otherwise selecting text will set editing:true on mobile
       sel?.selectAllChildren(thoughtContentEditable)
-
-      document.execCommand(command)
+      document.execCommand(command, false, color)
+      // selection.restore(savedSelection)
```
raineorshine, line 25: You can remove this commented out code if it's no longer needed.

src/components/ColorPicker.tsx
```diff
@@ -1,9 +1,16 @@
...
+import { rgbToHex } from '@mui/material'
+import React, { FC, useLayoutEffect, useMemo, useRef, useState } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
+import Thought from '../@types/Thought'
+import { editThoughtActionCreator as editThought } from '../actions/editThought'
+import { formatSelectionActionCreator as formatSelection } from '../actions/formatSelection'
 import { textColorActionCreator as textColor } from '../actions/textColor'
 import { isTouch } from '../browser'
-import getStyle from '../selectors/getStyle'
+// import * as selection from '../device/selection'
```
raineorshine, line 9: You can remove this commented out code if it's no longer needed.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +49,74 @@ const ColorSwatch: FC<{
   const dispatch = useDispatch()
   const colors = useSelector(themeColors)
   const fontSize = useSelector(state => state.fontSize)
-  const selected =
-    (color && cursorStyle?.color === color) || (backgroundColor && cursorStyle?.backgroundColor === backgroundColor)
+  const currentThought = useSelector(state => !!state.cursor && getThoughtById(state, head(state.cursor))) as Thought
```
raineorshine, line 52: It would be better to subscribe to the value of the thought rather than the whole thought, since that is all that is used. (You can use `head(cursor)` instead of `currentThought.id`).

Side note: `getThoughtById` does not return a stable reference, so as written this will cause the component to re-render on every change to the Redux state. This will be fixed when you subscribe to the value only, since strings are compared by value.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +49,74 @@ const ColorSwatch: FC<{
...
+  const simplePath = useSelector(state => simplifyPath(state, state.cursor!))
```
raineorshine, line 53: `simplifyPath` does not return a stable reference, so this will cause the component to be re-rendered on every change to the Redux state. Set the equality function to https://github.com/cybersemics/em/blob/main/src/util/equalPath.ts to avoid this problem.

It is dangerous to use `!` here. Either handle the `null` case or add a comment that compellingly demonstrates that `state.cursor` could never be null at runtime.

src/components/ColorPicker.tsx
```diff
@@ -111,18 +178,10 @@ const ColorSwatch: FC<{
...
-        ? {
-            // merge =style (which contains color) and =styleAnnotation (which contains backgroundColor)
-            // the style attribute name is not relevant here
-            ...(getStyle(state, head(state.cursor)) || ({} as React.CSSProperties)),
-            ...getStyle(state, head(state.cursor), { attributeName: '=styleAnnotation' }),
-          }
-        : undefined,
-    shallowEqual,
-  )
+  const cursorStyle = {
```
raineorshine, line 181: This reference is not stable, causing the props to ColorSwatch to change on every render, and preventing ColorSwatch from being memoized.

I recommend defining `backgroundColor` and `color` as separate variables and then using `useMemo` to construct a stable ref.

raineorshine (changes requested):
Thanks! Here's what I'm seeing:

✓ 

- **4. Allowing the click event to propagate breaks the dip behavior**
- **6. Default background color is not removed**

✗ 

- **5. Active swatch not selected on blur** - This appears to be broken again.
- The linter is failing.

raineorshine (changes requested):
Thanks, I can confirm that **5. Active swatch not selected on blur** is working now.

I'm seeing a couple edge cases that are not handled properly.

# 7. Wrong color swatch selected on white text next to colored text

## Steps to Reproduce

1. Create thought `Hello World`.
2. Select the text `World` and set the text color to green.
3. Set the browser selection anywhere in the text `Hello`.

## Current Behavior

The green swatch is incorrectly selected.

Brave Browser 2024-10-04 09 43 18

## Expected Behavior

No swatch should be selected (or the default swatch) when the text under the browser selection is the default text color.

This case should use `cursorStyle?.color` but seems to be using different logic.

# 8. Black on black

## Steps to Reproduce

1. Create thought `Hello World`.
2. Set the background color to green.
3. Select the text `World`.
4. Set the text color to green.

## Current Behavior

The text color and background color of `Hello` is black.

Brave Browser 2024-10-04 09 43 55

## Expected Behavior

The text color of `Hello` should be green and the background color should be green.

Brave Browser 2024-10-04 09 44 20

src/components/ColorPicker.tsx
raineorshine: Since `simplePath` is only used when you dispatch `editThought`, I would recommend using a thunk rather than useSelector:

```ts
  dispatch((dispatch, getState) => {
    const state = getState()
    if (!state.cursor) return
    const thought = getThoughtById(state, head(state.cursor))
    const simplePath = simplifyPath(state, state.cursor)
    dispatch(editThought({ 
      oldValue: thought.value, 
      newValue: thoughtContentEditable.innerHTML, 
      path: simplePath 
    })),
  })
```

src/components/ColorPicker.tsx
```diff
@@ -42,14 +49,90 @@ const ColorSwatch: FC<{
...
+  /** A function that adds an alpha channel to a hex color. */
+  const addAlphaToHex = (hex: string) => {
+    if (hex.length === 7) return hex + 'ff'
+    return hex
+  }
+
+  const selected = useMemo(() => {
+    // Define the color and background color regex to get the current color of current thought
+    const colorRegex = /color="(#[0-9a-fA-F]{6})"/g
+    const bgColorRegex = /background-color:\s*(rgb\(\d{1,3},\s?\d{1,3},\s?\d{1,3}\))/g
```
raineorshine, line 68: `color` uses hex and `background-color` uses rgb. Why is that? If there are implicit dependencies here, they need to be made explicit in a comment. It's not obvious from just looking at the code.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +49,90 @@ const ColorSwatch: FC<{
...
+    if (
+      (cursorStyle?.color === undefined && cursorStyle?.backgroundColor === undefined) ||
+      (selectedTextHexColor === '#ccccccff' && selectedBackHexColor === '#333333ff') ||
+      (selectedTextHexColor === addAlphaToHex(rgbToHex(colors.fg)) &&
+        selectedBackHexColor === addAlphaToHex(rgbToHex(colors.bg)) &&
+        !selection.isThought())
+    ) {
+      const colorMatches = currentThoughtValue.match(colorRegex)
+      let matchColor, match
+      // Get the colors and background colors used in current thought's value
```
raineorshine, line 84: This section could use a little more explanation. What does it mean for text to have multiple colors? Why is `matchColor` set to null? Provide an example in a comment if that helps.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +49,90 @@ const ColorSwatch: FC<{
...
+      const fgColors: Set<string> = new Set()
+      if (colorMatches) {
+        colorMatches.forEach(match => fgColors.add(match.slice(7, -1)))
```
raineorshine, line 87: Why `match.slice(7, -1)`? We have `addAlphaToHex` and `rgbToHex` helpers in order to make color operations more semantic. This operation could also use a helper method to make this code more understandable.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +49,90 @@ const ColorSwatch: FC<{
...
+        const elementColor = el.getAttribute('color')
+        const elementStyle = el.getAttribute('style')
+        if (
+          (elementColor && elementColor !== rgbToHex(colors.bg)) ||
+          (label === 'default' && elementStyle?.includes('background-color'))
+        )
+          el.removeAttribute('style')
+      })
+      dispatch(
+        editThought({ oldValue: currentThoughtValue, newValue: thoughtContentEditable.innerHTML, path: simplePath }),
```
raineorshine, line 132: Please explain in a comment why the cursor thought is being edited to its innerHTML, as this is not obvious from looking at the code. What problem does it solve? What would occur if this wasn't done?

src/components/ColorPicker.tsx
```diff
@@ -1,9 +1,16 @@
...
+import { formatSelectionActionCreator as formatSelection } from '../actions/formatSelection'
 import { textColorActionCreator as textColor } from '../actions/textColor'
 import { isTouch } from '../browser'
-import getStyle from '../selectors/getStyle'
+import * as selection from '../device/selection'
+import getThoughtById from '../selectors/getThoughtById'
+import simplifyPath from '../selectors/simplifyPath'
 import themeColors from '../selectors/themeColors'
+import commandStateStore from '../stores/commandStateStore'
+// import equalPath from '../util/equalPath'
```
raineorshine, line 13: This commented out code can be removed if it's no longer used.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +49,90 @@ const ColorSwatch: FC<{
...
+  const simplePath = useSelector(state => simplifyPath(state, state.cursor!))
   size = size || fontSize * 1.2
 
+  /** A function that adds an alpha channel to a hex color. */
+  const addAlphaToHex = (hex: string) => {
+    if (hex.length === 7) return hex + 'ff'
+    return hex
+  }
+
+  const selected = useMemo(() => {
```
raineorshine, line 65: Rather than use `useMemo` with several `useSelector` values as dependencies, wrap all this logic in a single `useSelector`. Then you have access to a fresh State and the component is only re-rendered when the `selected` boolean value changes.

raineorshine (changes requested):
Looks good! I can confirm that (7) and (8) are fixed. That means that all the functional requirements are met! 🎉

The remaining suggestions are provided in order to clean up the implementation.

Lastly, the tests are failing. I haven't investigated why, so take a look and see if they are regressions or tests that need to be updated to reflect the new functionality. Thanks so much :).

src/components/ColorPicker.tsx
```diff
@@ -42,14 +48,104 @@ const ColorSwatch: FC<{
...
+  const addAlphaToHex = (hex: string) => {
+    if (hex.length === 7) return hex + 'ff'
+    return hex
+  }
+
+  const selected = useSelector(state => {
+    const currentThoughtValue = (!!state.cursor && getThoughtById(state, head(state.cursor))?.value) || ''
+    const themeColor = themeColors(state)
+    // Define the color and background color regex to get the current color of current thought
+    // document.execCommand('foreColor') adds the color attribute with hex and document.execCommand('backColor') adds the background-color attribute with the rgb
```
raineorshine, line 64: This explains *what* it is doing, but not *why*. `document.execCommand('foreColor')` and `document.execCommand('backColor')` both work with hex or rgb, so why the difference here?

Maybe it is connected to this comment in themeColors.ts:

```
// Background colors in capacitor app needs to be in hexadecimal codes
```

There may have originally been no repercussions of specifying the background color in hex and the rest of the colors in rgb. But now we have parsing logic that is based on those colors, so that seemingly arbitrary difference is spreading outward and affecting other parts of the app. Ideally, the colors should all be the same format, and just converted to hex for capacitor as needed. In lieu of refactoring the hex background, we must at least leave a comment that clarifies the root cause of this conspicuous difference. 

Hope that makes sense :).

src/components/ColorPicker.tsx
raineorshine: Okay, I think I understand. So `match.slice(7, -1)` extracts the hex value from the first color match. And `matchColor` is only defined if the thought has a single color match, otherwise it is null.

src/components/ColorPicker.tsx
```diff
@@ -42,14 +48,104 @@ const ColorSwatch: FC<{
...
+          (label === 'default' && elementStyle?.includes('background-color'))
+        )
+          el.removeAttribute('style')
+      })
+      dispatch((dispatch, getState) => {
+        const state = getState()
+        if (!state.cursor) return
+        const thought = getThoughtById(state, head(state.cursor))
+        const simplePath = simplifyPath(state, state.cursor)
+        // Edit the current thought with the changed innerHTML
```
raineorshine, line 138: This says *what* the code does, but not *why*. Can you explain? (What problem does it solve? What would occur if this wasn't done?)

src/components/ColorPicker.tsx
```diff
@@ -42,14 +48,104 @@ const ColorSwatch: FC<{
...
+      (backHexColor && backHexColor === selectedBackHexColor)
+    )
+  })
+
   /** Toggles the text color to the clicked swatch. */
   const toggleTextColor = (e: React.MouseEvent | React.TouchEvent) => {
     // stop toolbar button dip
     e.stopPropagation()
+    e.preventDefault()
+    console.log(color, backgroundColor, colors.bg, colors.fg)
```
raineorshine, line 112: Make sure to run the linter before doing a git push, or at least check the CI output after.

raineorshine (changes requested):

raineorshine:
The tests are timing out for some reason. Can confirm on my local machine. The ColorPicker test is passing though, so I'm not sure what the problem is.

raineorshine (changes requested):
Thanks, the remove default color logic looks better now.

The linter and the puppeteer tests are failing. Once those pass we can merge!

raineorshine (approved):

Issue #2284, opened 2024-08-13
Tutorial: Fix small GestureDiagram size and arrow head


raineorshine:
## Steps to Reproduce

Proceed to Step 5 of the Tutorial. 

## Current Behavior

The small GestureDiagram is too thin and the arrow head is too small. 

![image](https://github.com/user-attachments/assets/db7a8ce3-e70e-4b48-91ed-0bd78e923eb8)

## Expected Behavior

The GestureDiagram should be rendered at a size that is easy to read and matches the surrounding text.

PR #2329, opened 2024-08-27
Issue 2282

findusman:
Issue #2284 resolved

Slider removed, fade in and fade out implemented


raineorshine (changes requested):
This will also need to ensure the height is animated when the content height changes from one tutorial step to the next. Maybe measure and apply a height in `useLayout`?

Please rebase on `main` and ensure that only changes related to the associated issue show up in the PR diff. Otherwise it makes it harder to review. Thanks!

Issue #1601, opened 2022-06-19
[mobile] Scroll at edge during drag


raineorshine:
When dragging a thought, if the cursor moves close to the edge of the screen and there are more thoughts out of view, smoothly scroll them into view.

- [x] The active zone should be about a 50px from the edge of the screen. The closer the cursor is to the edge, the faster is should scroll. These should be configurable so that we can tune the user experience.
  - [initial implementation](https://github.com/cybersemics/em/blob/cca37c8f2a2e39aadb465d981b327fd6047ae430/src/util/initEvents.ts#L205-L228)

raineorshine:
The basic functionality is in place, but there was a regression. It does not seem to take the height of the toolbar when scrolling up, and the breadcrumbs when scrolling down.

PR #2335, opened 2024-08-28
Implement scroll feature on drag over the edges

Zubair286:
#1601 

This PR implements feature of scrolling over the edges during a drag process. The toolbar and breadcrumb height is taken into account when dragging over the top and bottom edges.

### DIFF
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -92,7 +92,7 @@ const Sidebar = () => {
       >
         <div
           className='sidebar'
-          data-autoscroll
+          data-scroll-at-edge
           style={{
             userSelect: 'none',
             // must be position:relative to ensure drop hovers are positioned correctly when sidebar is scrolled
--- a/src/components/Sidebar.tsx
+++ b/src/components/Sidebar.tsx
@@ -92,6 +92,7 @@ const Sidebar = () => {
       >
         <div
           className='sidebar'
+          data-autoscroll
           style={{
             userSelect: 'none',
             // must be position:relative to ensure drop hovers are positioned correctly when sidebar is scrolled
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -179,6 +179,7 @@ const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected }) => {
             id='toolbar'
             ref={toolbarRef}
             className='toolbar'
+            {...(customize ? { 'data-autoscroll': true } : {})}
             onScroll={onScroll}
             style={{
               marginLeft: customize ? -3 : 0,
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -179,7 +179,7 @@ const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected }) => {
             id='toolbar'
             ref={toolbarRef}
             className='toolbar'
-            {...(customize ? { 'data-autoscroll': true } : {})}
+            data-scroll-at-edge={customize}
             onScroll={onScroll}
             style={{
               marginLeft: customize ? -3 : 0,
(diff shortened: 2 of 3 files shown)


raineorshine (changes requested):
This task was to fix the existing edge scroll behavior, not implement it from scratch.

Zubair286:
I see, it wasn't working for me at all, so I assumed we had to start from scratch. Sorry for misunderstanding.

raineorshine:
Sorry, that definitely wasn't clear! The `feature` label was from the original issue; I changed it to `bug`. I re-opened the issue in May when it stopped working properly, with this comment:

> The basic functionality is in place, but there was a regression.

The implementation is done in the global `touchmove` event here:

https://github.com/cybersemics/em/blob/cca37c8f2a2e39aadb465d981b327fd6047ae430/src/util/initEvents.ts#L205-L228

Note that it currently only works on mobile. We will need to do a desktop implementation at some point, but that's a separate issue.

I'm open to other implementation strategies, but we should discuss the tradeoffs before jumping in. I imagine the hook-based approach you different advantages and disadvantages.

Zubair286:
Alright, I get it. So this issue is for mobile only, right?

raineorshine:
Correct. You may be able to reproduce with Chrome Developer Tools device toolbar, but I haven't tried.

Zubair286:
It seems that as long as we are only scrolling the window, it works but setting the scrollContainer to either toolbar or sidebar is not working. 

```
  const scroll = () => {
    const scrollLeft = document.documentElement.scrollLeft
    const scrollLeftNew = Math.max(0, scrollLeft + rate.x)
    const scrollTop = document.documentElement.scrollTop
    const scrollTopNew = Math.max(0, scrollTop + rate.y)

    // if we have hit the end, stop autoscrolling
    // i.e. if the next increment would not change scrollTop
    if (scrollLeftNew === scrollLeft && scrollTopNew === scrollTop) {
      autoscrolling = false
      return
    }

    window.scrollTo(scrollLeftNew, scrollTopNew)
    window.requestAnimationFrame(() => {
      if (autoscrolling) {
        scroll()
      }
    })
  }
```

Zubair286:
[Screencast from 08-29-2024 06:35:38 PM.webm](https://github.com/user-attachments/assets/8345d504-c42c-4377-a25d-40d3ae17d61e)

raineorshine:
Okay, thanks for the update. Let me know what you come up with.

Zubair286:
is `scrollContainer` being assigned the toolbar element for the `CustomizeToolbar` drag and drop?

Zubair286:
The problem I see is that, it is always finding the toolbar in all cases, hence the `scrollTop` value never changes because `toolbar` is not being scrolled. We can check for the visibility of the sidebar but for toolbar it wouldn't work since it is always there. is the toolbar in this case refer to the one in `CustomizeToolbar` or main one?

raineorshine:
Drag-and-drop functionality exists in CustomizeToolbar only.

raineorshine:
We need a clean, flexible solution. The current implementation's direct reference to the sidebar and toolbar is questionable. 

One idea is to add a `data-autoscroll` attribute to the sidebar and the toolbar when it's in customize mode. Then we could query the closest `data-autoscroll` above `e.target` in the `onTouchMove` event. Log(n) should be fast enough to avoid performance issues, and then the code in initEvents would be completely agnostic to the drag-and-drop source.

Zubair286:
Yes, don't see the reason to traverse the entire tree for an element on every touch event.

raineorshine:
You can grab the element on dragstart.

Zubair286:
By declaring the scrollContainer as global, and setting it on dragstart event? I see, let me check on that.

raineorshine (approved):
Works great! Thanks.

I renamed autoscroll → scroll-at-edge so as to avoid confusion with https://github.com/cybersemics/em/blob/715208aee47a3c024bfc92d0a88abacb26fff051/src/device/preventAutoscroll.ts.

PR #2344, opened 2024-08-30
PandaCSS modal + anchor button recipes

yangchristina:
- modalRecipe and modalText recipes to replace .popup and .modal classes
- anchorButton recipe to replace a.button


panda.config.ts
```diff
@@ -113,6 +136,21 @@ const globalCss = defineGlobalStyles({
     userSelect: 'none',
     cursor: 'auto',
   },
+  // Sets default link color in recipes/modal color
+  '.modal__root': {
```
raineorshine, line 140: `.modal__root` and `.modal__actions` are build artifacts, so we don't want to code against them. Is there a different we can apply these styles?

src/components/Tutorial/TutorialHint.tsx
```diff
@@ -14,7 +16,15 @@ const TutorialHint: FC<PropsWithChildren> = ({ children }) => {
   return (
     <>
       <a
-        className={`button button-variable-width button-status button-less-padding text-small ${hint ? 'button-dim' : ''}`}
```
raineorshine, line 17: The TutorialHint has a regression:

## Current Behavior

d3d9f02586

The hint anchor button text is blue.

Brave Browser 2024-08-31 09 18 14

## Expected Behavior

The hint anchor button text should be black.

Brave Browser 2024-08-31 09 18 09

src/components/NewThought.tsx
```diff
@@ -78,11 +79,24 @@ const NewThought = ({ path, showContexts, label, value = '', type = 'bullet' }:
...
+                    color: {
+                      base: 'rgba(7, 7, 7, 0.5)',
+                      _dark: 'rgba(255, 255, 255, 0.5)',
+                    },
+                  })
+                : cx(
+                    anchorButton({
+                      variableWidth: true,
+                    }),
+                  )
```
raineorshine, line 95: Don't worry about it for this PR, but in the future try to write objects that only have one or two properties on a single-line like this:

```ts
cx(anchorButton({ variableWidth: true }))
```

Prettier will write them on multiple lines if needed.

Expressions with conditional styles like the `css` statement above are more readable on multiple lines though, as I would continue formatting them as you have been.

src/recipes/modal.ts
```diff
@@ -0,0 +1,162 @@
...
+    opaque: {
+      true: {
+        root: {
+          animation: 'fadein 0.4s',
+        },
+      },
+    },
+  },
+  jsx: ['Devices', 'Help', 'ShortcutTable', 'Auth', 'ModalComponent'],
+  staticCss: ['*'],
```
raineorshine, line 159: Can you explain what `staticCss` does? I've read the [panda docs](https://panda-css.com/docs/guides/dynamic-styling#using-static-css) but I don't totally understand.

src/recipes/modal.ts
```diff
@@ -0,0 +1,162 @@
...
+    },
+    opaque: {
+      true: {
+        root: {
+          animation: 'fadein 0.4s',
+        },
+      },
+    },
+  },
+  jsx: ['Devices', 'Help', 'ShortcutTable', 'Auth', 'ModalComponent'],
```
raineorshine, line 158: What does `jsx` do?

src/recipes/modal.ts
```diff
@@ -0,0 +1,162 @@
...
+          base: 'rgba(0, 0, 0, 0.3)',
+          _dark: 'rgba(255, 255, 255, 0.3)',
+        },
+        borderColor: {
+          base: 'rgba(0, 0, 0, 0.3)',
+          _dark: 'rgba(255, 255, 255, 0.3)',
+        },
+      },
+    },
+    closeX: {
```
raineorshine, line 81: We should do this as inline `css` in the `ModalComponent` since it does not have any variants. It's better to have it co-located with the component whenever possible.

I haven't gone through the other slots in this recipe, but let's inline any others that can be inlined.

src/recipes/modal.ts
```diff
@@ -0,0 +1,162 @@
...
+        },
+        actions: { fontSize: '0.75em' },
+      },
+      home: {
+        // unused, from .modal-home, not sure what this is for
+        root: {
+          marginLeft: '20px',
+          marginTop: '-5px',
+        },
+      },
```
raineorshine, line 110: If it's unused, we should remove it.

raineorshine (changes requested):
Looks good!

src/recipes/modal.ts
yangchristina: `jsx` doesn't really matter, I probably should have only put `ModalComponent` in it. It tells PandaCSS which components use the modal recipe, but it isn't necessary.
Here are the docs to it. https://panda-css.com/docs/concepts/recipes#advanced-jsx-tracking

src/recipes/modal.ts
yangchristina: For recipes, PandaCSS only generates the variants that are being used. And sometimes it can't properly detect whether or not it's being used. For example, since the `center` variant in modal is controlled by a variable, PandaCSS couldn't detect that it was being used, so I had to add `staticCss: ['*']` to force it to generate the `center` class. I could have been more specific in `staticCss` and only named the specific problematic variants, but I figured that almost all variants were being used anyway, and there were multiple variants that used variables, so I might as well generate them all.

panda.config.ts
yangchristina: Descendant selectors won't work here (I tried), so the only way to replace it that I could think of was to add the styles directly to each <a> tag. Since there's a lot, it may be hard remember to be consistent with it, and there would be a lot of instances to replace. 

I do know this isn't a good method, but since it get's put in the @base layer, it can be overridden by any recipe variants. Also, this behaviour is at least documented: https://panda-css.com/docs/concepts/slot-recipes#targeting-slots

Another good option would be to use `defineParts`. This would still require adding it directly to each <a> tag, but would making writing it easier, and would allow any changes to apply to all of the links.
https://panda-css.com/docs/concepts/slot-recipes#defineparts
I was a bit unsure on this, since there was some talk about deprecating it a few years ago, and it isn't really documented, but it's been a while since then, and I don't think they are planning on deprecating it. I believe some component libraries use it under the hood, and I found an issue where the maintainers specifically said to use it over `sva`

raineorshine: Thanks for describing what options we have. 

Since the slot class is not type safe, and defineParts just takes us deeper into proprietary slot functionality, I think the better approach is to apply a recipe directly to each link. That has the benefit of co-location, simplicity, and moving fully away from descendant selectors. 

Since this does require a lot of small changes, I recommend doing it in a separate PR. I'll leave it up to you whether you want to do it after this PR, or circle back later on.

src/recipes/modal.ts
raineorshine: Okay, thanks. 

If we can omit this that would be great. Components can be removed or change names, so I think it's less than ideal to tightly couple the recipe to the components where it is used.

src/components/Tutorial/TutorialHint.tsx
yangchristina: Sorry I missed this

panda.config.ts
yangchristina: Sounds good, I'll circle back to this later on.

src/recipes/link.ts
```diff
@@ -0,0 +1,33 @@
...
+        display: 'inline-block',
+        overflow: 'hidden',
+        position: 'relative',
+        zIndex: 'thoughtAnnotationLink',
+      },
+      footer: {
+        color: { _dark: 'skyblue' },
+      },
+      tutorialText: {
+        color: { _dark: '#87ceeb' },
```
raineorshine, line 27: Let's add `#87ceeb` as the global `_dark` color for `a` tags. Then we can remove the modal, footer, and tutorialText variants completely.

`skyblue` is identical to `#87ceeb`, and `lightblue` is so close that we should probably combine them.

src/recipes/link.ts
```diff
@@ -0,0 +1,33 @@
...
+        color: 'inherit',
+        textDecoration: 'none',
+      },
+      thoughtAnnotation: {
+        height: '1em',
+        display: 'inline-block',
+        overflow: 'hidden',
+        position: 'relative',
+        zIndex: 'thoughtAnnotationLink',
+      },
```
raineorshine, line 22: The `breadcrumbs` and `thoughtAnnotation` variants were added to the recipe but unless I am missing something they are not used anywhere. Is that true? We should aim to replace the old App.css usage in the same commit as adding the new PandaCSS usage.

src/recipes/link.ts
```diff
@@ -0,0 +1,33 @@
...
+  variants: {
+    in: {
+      modal: {
+        color: 'lightblue',
+      },
+      breadcrumbs: {
+        color: 'inherit',
+        textDecoration: 'none',
+      },
+      thoughtAnnotation: {
```
raineorshine, line 16: The only links that appear in the `ThoughtAnnotation` component are the [UrlIconLink](https://github.com/cybersemics/em/blob/587e1d309d9570e5ad06e9bec8eb0570867aa7d4/src/components/ThoughtAnnotation.tsx#L46) and the [EmailIconLink](https://github.com/cybersemics/em/blob/587e1d309d9570e5ad06e9bec8eb0570867aa7d4/src/components/ThoughtAnnotation.tsx#L75), so let's move this style into the `ThoughtAnnotation.tsx` file, preferably as a static object rather than a recipe.

src/components/Tutorial/TutorialNavigation.tsx
```diff
@@ -51,10 +70,7 @@ const TutorialNavigation = ({
             const step = i + (tutorialStep < TUTORIAL2_STEP_START ? 1 : TUTORIAL2_STEP_START)
             return (
               <a
-                className={classNames({
-                  'tutorial-step-bullet': true,
-                  active: step === Math.floor(tutorialStep),
-                })}
+                className={tutorialStepBullet({ active: step === Math.floor(tutorialStep) })}
```
raineorshine, line 73: I can think of a couple ways to do this with inline styles, but correct me if I'm wrong:

```ts
css({
    color: 'inherit',
    textDecoration: 'none',
    fontSize: '32px',
    opacity: step === Math.floor(tutorialStep) ? 0.25 : undefined,
    marginLeft: '1px',
    marginRight: '1px',
    transition: 'all 400ms ease-in-out',
  })
```

or

```ts
css({
    color: 'inherit',
    textDecoration: 'none',
    fontSize: '32px',
    marginLeft: '1px',
    marginRight: '1px',
    transition: 'all 400ms ease-in-out',
  }, {
opacity: step === Math.floor(tutorialStep) ? 0.25 : undefined
})
```

The advantage in my mind is that the styles are closer to where they are being used, it depends less on panda-specific syntax, it avoids a one-time use identifier. Let me know if there's something I'm missing though! I'm still learning about PandaCSS and finding the right balance. We ultimately want to take advantage of the Panda compiler without adopting too many Panda-specific idioms unless they provide a major benefit.

src/recipes/modal.ts
```diff
@@ -0,0 +1,113 @@
...
+    },
+    opaque: {
+      true: {
+        root: {
+          animation: 'fadein 0.4s',
+        },
+      },
+    },
```
raineorshine, line 108: `opaque` does not appear to be used anywhere, so it can be removed.

raineorshine (changes requested):
There are some merge conflicts once you merge `main`. Sorry about that! Let me know if you need any help resolving them. I waited a couple days before I pushed my changes, but they were adding up so I had to push.

src/recipes/link.ts
yangchristina: Ok sounds good.

src/recipes/link.ts
yangchristina: I can look at that. I was planning on just moving the classes that affect a tags into the link recipe first, and deal with them later.

src/components/Tutorial/TutorialNavigation.tsx
yangchristina: I've fixed this now. Yes, doing this makes sense. I've just run into so many cases where panda wasn't statically generating classes with conditional statements that I started automatically turning to variants instead. I did this this though, and confirmed it works. I guess panda can handle the simpler conditional statements.

src/recipes/link.ts
yangchristina: the `Link` component is used within breadcrumbs I believe. I think looking in this is a bit more complex, so I'll leave this for now.

yangchristina:
@raineorshine I've dealt with the merge conflicts. I appreciate that you waited, and it's completely my fault for not working on the PR for a while after making it.

src/components/Tutorial/TutorialNavigation.tsx
raineorshine: Great, thank!

Yeah, I'm still unsure all the cases where static analysis is possible. It's definitely a different way of thinking about the code. PandaCSS can handle simple ternary statements at least when the values are static. It may be worth looking into the generated CSS to see how it compares, but that’s more of an optimization.

I would say that variants are good for well-defined variations on a recipe that is used in multiple modules, whereas inline css is better for ad hoc styles. Others who value standardization over co-colocation may make different architectural decisions though.

raineorshine (approved):

raineorshine:
Thank you!

PR #2347, opened 2024-09-02
Fix #2314: VirtualThought Re-rendering on typing

### LINKED ISSUE #2314: VirtualThought: Re-rendering on typing
> I noticed that typing got increasingly slow as I added many thoughts, so I debugged the rendering with React Dev Tools and Chrome Dev Tools. React Dev Tools pointed me to the biggest rendering bottleneck.
> 
> The problem is VirtualThought is rerendering unnecessarily, so for every keystroke, all the thoughts are being rerendered, which is very expensive. The reason is that even though it uses React.memo, memo performs a shallow comparison by default, and the path and simplePath props are arrays that I'm guessing are being cloned or recalculated, so the references change, provoking a rerender of LayoutTree and all of the thoughts every time (see screenshot 1).
> 
> The solution is adding a deep comparison to the React.memo function:
const VirtualThoughtMemo = React.memo(VirtualThought, (prevProps, nextProps) => isEqual(prevProps, nextProps))
> 
> I tested this with the same 240 thoughts. The typing is smooth, and the rerendering is gone (screenshot 2).
> 
> The screenshots show that LayoutTree and its children's total rendering time decreased from 230.6ms to 2.5ms with this optimization.

image

image

I'm curious why path and simplePathare not stable refs. They shouldn't change when the user edits a thought. Could you investigate? isEqual fixes the problem, but it would be good for us to better understand the root cause.

My (naive) attempt to stabilize the refs before involved memoizing childPath:

https://github.com/cybersemics/em/blob/10b9d7c90faca1aec451bb104386269b4be53c11/src/components/LayoutTree.tsx#L273

...and using isEqual on the linearizeTree selector:

https://github.com/cybersemics/em/blob/10b9d7c90faca1aec451bb104386269b4be53c11/src/components/LayoutTree.tsx#L351

Off the top of my head I can't think of why the entire tree would be recalculated when the current thought changes. If there is a better way to do this I would love to find it.

carloslfu:
This PR contains two changes, as discussed in the issue (#2314):
- It changes the TreeThought to store the ThoughtId instead of the Thought. This makes the selector provoke fewer re-renders. I tested this before and after, and it, in fact, improves rendering performance by not triggering unnecessary re-renders.
- It adds a custom `arePropsEqual` to the `VirtualThoughtMemo` component.

Adding the custom equality function fixes the rerenders when typing:

Screenshot 2024-09-02 at 1 50 19 AM

### DIFF
--- a/src/components/VirtualThought.tsx
+++ b/src/components/VirtualThought.tsx
@@ -240,7 +240,24 @@ const VirtualThought = ({
   )
 }
 
-const VirtualThoughtMemo = React.memo(VirtualThought)
+type VirtualThoughtPropsKeys = keyof typeof VirtualThought
+
+const VirtualThoughtMemo = React.memo(VirtualThought, (prevProps, nextProps) => {
+  let isEqual = true
+
+  for (const key in prevProps) {
+    if (key === 'path' || key === 'simplePath') {
+      isEqual = equalPath(prevProps[key], nextProps[key])
+      if (!isEqual) break
+    } else if (prevProps[key as VirtualThoughtPropsKeys] !== nextProps[key as VirtualThoughtPropsKeys]) {
+      isEqual = false
+      break
+    }
+  }
+
+  return isEqual
+})
+
 VirtualThoughtMemo.displayName = 'VirtualThought'
 
 export default VirtualThoughtMemo
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -61,7 +61,7 @@ type TreeThought = {
   simplePath: SimplePath
   // style inherited from parents with =children/=style and grandparents with =grandchildren/=style
   style?: React.CSSProperties | null
-  thought: Thought
+  thoughtId: string
   // keys of visible children
   // only used in table view to calculate the width of column 1
   visibleChildrenKeys?: string[]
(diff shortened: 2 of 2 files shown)


src/components/VirtualThought.tsx
```diff
@@ -240,7 +240,24 @@ const VirtualThought = ({
   )
 }
 
-const VirtualThoughtMemo = React.memo(VirtualThought)
+type VirtualThoughtPropsKeys = keyof typeof VirtualThought
+
+const VirtualThoughtMemo = React.memo(VirtualThought, (prevProps, nextProps) => {
+  let isEqual = true
+
+  for (const key in prevProps) {
```
raineorshine, line 248: Out of curiosity was your approach a stylistic choice or performance choice? I would have done it like this:

```js
prevProps.every(prop => 
  key === 'simplePath' || key === 'simplePath` 
    ? equalPath(prevProps[key], nextProps[key])
    : prop[key] === nextProps[key]
)
```

raineorshine (approved):

src/components/VirtualThought.tsx
raineorshine: Thanks for clarifying. In this case I don't have a strong preference, just wanted to better understand your reasoning. It's good to know you're aware of more modern alternatives, and I agree that one or two extra function invocations is unlikely to have a measurable effect on performance here. 

I did hire you for performance optimizations after all, so I can't blame you for providing more performant code! :)

Issue #1828, opened 2024-04-18
Drag and drop: Indicate when sorted drop target is out of view


raineorshine:
In this screenshot, `Colorado` is being dragged and hovered over the sorted `States` context, but the drop target is not visible. 

Render an upward pointing blue triangle at the top of screen to communicate to the user that the thought will be inserted above the top of the screeen. 

Render a downward pointing blue triangle at the bottom of the screen when the insertion point is below the bottom of the screen (not pictured). 

Animate the arrows with a gentle up-and-down bobbing motion. 

![image](https://github.com/cybersemics/em/assets/750276/69454f36-d58e-426e-bc1c-c4fd3560d594)

``` 

- US
  - States
    - =pin
      - true
    - =sort
      - Alphabetical
        - Asc
    - Alabama
    - Alaska
    - Arizona
    - Arkansas
    - California
    - Connecticut
    - Delaware
    - Florida
    - Georgia
    - Hawaii
    - Idaho
    - Illinois
    - Indiana
    - Iowa
    - Kansas
    - Kentucky
    - Louisiana
    - Maine
    - Maryland
    - Massachusetts
    - Michigan
    - Minnesota
    - Mississippi
    - Missouri
    - Montana
    - Nebraska
    - Nevada
    - New Hampshire
    - New Jersey
    - New Mexico
    - New York
    - North Carolina
    - North Dakota
    - Ohio
    - Oklahoma
    - Oregon
    - Pennsylvania
    - Rhode Island
    - South Carolina
    - South Dakota
    - Tennessee
    - Texas
    - Utah
    - Vermont
    - Virginia
    - Washington
    - West Virginia
    - Wisconsin
    - Wyoming
  - Colorado
```

hereistopdev:
This is Sheldon from Upwork. I will start work right now to fix this issue

hereistopdev:
https://www.upwork.com/freelancers/~0194d712c9f174416f

Zubair286:
When trying to render a downward arrow for a context such as above `States`, there is an issue with `DropHover` placement. All the thoughts are not rendered until we scroll down so for a thought that will be placed at the very bottom will not have any `DropHover` rendered until we scroll down. Is there any way to detect the position of the  `DropHover` on large context when its rendering at the bottom?

raineorshine:
[...]

PR #2369, opened 2024-09-05
Add indicator arrow on top/bottom of viewport for drop positions outs…

Zubair286:
#1828 

This issue implements a feature that adds an indicator arrow on top/bottom of the viewport, if a thought is dragged over a sorted context and its drop position is outside of viewport. Checking the rank of the target position and comparing the y value of that thought with viewport height + scrollY for bottom arrow, and comparing y with scrollY for top arrow.

--- 

- [x] Bottom arrow when drop target is below the viewport
- [x] Top arrow when drop target is above the viewport
- [x] 1. False negative: hovering slightly above to Colorado
- [x] 2. False positive: hovering over child of sorted context
- [x] 3. False positive when dragging over toolbar or footer
- [x] 4. False positive: Sort near top of screen
- [x] 5. False positive: Sort near bottom of screen
- [x] 6. False positive: Drag to the left (not a drop target)
- [x] 7. False negative: Sort below screen
- [x] 8. False positive when canDrop is false
- [x] 9. Parent highlight is broken


raineorshine (changes requested):
This is a good start, thanks!

You may want to address the behavioral issues below, and then we can move on to refactoring to clean up the design.

1. There is one false negative for the bottom arrow. It doesn't show the indicator arrow even when hovering over the sorted context. Here I am hovering `Colorado` over the Wyoming bullet (which will drop it into the sorted `States` context):

image

2. There is a false positive for both top and bottom arrows. The indicator arrow is incorrectly shown when dragging into a child of the sorted context, in which the DropHover is not outside the viewport:
Screenshot 2024-09-05 at 12 18 56 PM

The other thing to test is that it works on mobile when the keyboard is up. Unfortunately `position: fixed` does not work when the keyboard is up, so you'll have switch to `position: absolute` like in https://github.com/cybersemics/em/blob/587e1d309d9570e5ad06e9bec8eb0570867aa7d4/src/components/Tutorial/TutorialScrollUpButton.tsx.

raineorshine (commented):
All of this new logic should probably go in a custom hook as well to separate it cleanly from the rest of the component logic.

src/components/Content.tsx
```diff
@@ -47,6 +55,66 @@ const Content: FC = () => {
...
+  const newRank = useSelector(state => getSortedRank(state, head(contextParentPath), sourceThought?.value || ''))
+  const scrollY = window.scrollY
+  const viewportHeight = viewportStore.useSelector(state => state.innerHeight)
+  // The arbitrary space above the first thought
+  const spaceAbove = 100
+
+  /** Calculate the visibiltiy of thought being dragged in a sorted context.  */
+  const calculateHoverArrowPosition = useCallback(
+    (thoughts: TreeThoughtPositioned[]) => {
+      const state = store.getState()
```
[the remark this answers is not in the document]
Zubair286: That's right, but I can't use `useSelector` inside a function so I turned to using `getState` method.

src/components/Content.tsx
```diff
@@ -47,6 +55,66 @@ const Content: FC = () => {
     return children.length
   })
   const isAbsoluteContext = useSelector(state => isAbsolute(state.rootContext))
+  const visibilityRef = useRef<'above' | 'below' | null>(null)
```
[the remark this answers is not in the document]
Zubair286: I did not use state because it would cause too many renders during drag but if it can be calculated with existing state maybe I need to think about it again.

src/components/LayoutTree.tsx
```diff
@@ -591,6 +591,11 @@ const LayoutTree = () => {
   // get the scroll position before the render so it can be preserved
   const scrollY = window.scrollY
 
+  // When a thought is hovered over a sorted context, determine the position of arrow.
+  useEffect(() => {
+    calculateThoughtPosition(treeThoughtsPositioned)
+  }, [calculateThoughtPosition, treeThoughtsPositioned])
```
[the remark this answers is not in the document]
Zubair286: I first put it in `LayoutTree` but the arrow was not staying fixed relative to the viewport during scroll. So I put it in `Content`, I really don't like this way as well since I have to pass props to `LayoutTree` which really does not need to be accepting it.

src/components/Content.tsx
```diff
@@ -47,6 +55,66 @@ const Content: FC = () => {
...
+  })
+
+  const newRank = useSelector(state => getSortedRank(state, head(contextParentPath), sourceThought?.value || ''))
+  const scrollY = window.scrollY
+  const viewportHeight = viewportStore.useSelector(state => state.innerHeight)
+  // The arbitrary space above the first thought
+  const spaceAbove = 100
+
+  /** Calculate the visibiltiy of thought being dragged in a sorted context.  */
+  const calculateHoverArrowPosition = useCallback(
```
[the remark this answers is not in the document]
Zubair286: I understand the concern of reiterating the `treeThoughtsPosition` array, if I can put the arrow in the `LayoutTree` without scrolling it out of the viewport, we can use the existing state to remove the unnecessary iterations. Maybe I missed something I will try again with the arrow placement and try to find a fix for this.

src/components/LayoutTree.tsx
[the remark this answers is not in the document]
raineorshine: It's going to have to be switched to `position: absolute` anyway to make it work on mobile Safari.

src/components/Content.tsx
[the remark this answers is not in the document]
raineorshine: `useRef` and `useMemo` can both be used to persiste state across render, but for different reasons.  

`useRef` is used when you need _different_ information from the last render (e.g. the previous cursor location). 

`useMemo` is used when you need the _same_ information from the last render (for performance optimization).

src/components/Content.tsx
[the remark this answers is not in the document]
raineorshine: You would need to add `useSelector` outside the function, or turn the whole callback into a `useSelector`.

src/components/Content.tsx
```diff
@@ -47,6 +55,66 @@ const Content: FC = () => {
...
+    const sourceThoughtId = head(item?.path || [])
+    const sourceThought = getThoughtById(state, sourceThoughtId)
+    return sourceThought
+  })
+
+  const newRank = useSelector(state => getSortedRank(state, head(contextParentPath), sourceThought?.value || ''))
+  const scrollY = window.scrollY
+  const viewportHeight = viewportStore.useSelector(state => state.innerHeight)
+  // The arbitrary space above the first thought
+  const spaceAbove = 100
```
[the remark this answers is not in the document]
Zubair286: Sorry about this, I wanted to calculate the space relative to font size and toolbar height but left this in there accidentally.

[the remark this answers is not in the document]
raineorshine: No problem! Glad you're aware.

trevinhofmann:
Looking good!

@Zubair286 - Have you been able to resolve this?

> There is one false negative for the bottom arrow. It doesn't show the indicator arrow even when hovering over the sorted context. Here I am hovering Colorado over the Wyoming bullet (which will drop it into the sorted States context)

I am still seeing it locally during testing:

https://github.com/user-attachments/assets/af78d0b6-785f-4855-a766-b4047a8187c0

Zubair286:
I am not sure if its a false positive but it still thinks its in `States` while I hover over `Colorado`. As soon as I hover over `Wyoming` it shows though.

[Screencast from 09-18-2024 11:39:58 AM.webm](https://github.com/user-attachments/assets/00a69f00-c7bf-4346-b731-18db894035cc)

Zubair286:
I wonder if its because its hovering over `States` drop end hence it drops even when its not hovering over sorted context since that starts from `Wyoming` thought I believe.

Zubair286:
@raineorshine in this case where we are hovering over `States` drop-end, should we still show the arrow? I am not sure since we are not in the sorted context yet.

raineorshine:
Can you provide steps? I'm not sure what you mean. 

The arrow should be shown whenever the drop target is outside the viewport.

Zubair286:
I meant like in this case, where I am hovering over `Colorado` but it still shows the drop target outside the viewport regardless of hover in sorted context.

[Screencast from 09-18-2024 11:00:52 PM.webm](https://github.com/user-attachments/assets/168299d3-3cba-4974-b12b-0eb3dda92379)

Zubair286:
I meant like in this case, where I am hovering over `Colorado` but it still shows the drop target outside the viewport regardless of hover in sorted context.


[Screencast from 09-18-2024 11:34:47 AM.webm](https://github.com/user-attachments/assets/1166182d-8f7c-4cc3-a53f-b82d8e1d746f)

raineorshine:
Yes, this case should show the arrow. Any hover that drops into `States` when the sorted drop target is outside the viewport should show the arrow. The arrow needs to correspond to the drop behavior.

Zubair286:
Understood, I don't believe we have any selector that tells us that we are hovering over a drop end, since this is only because of that rest of the case are working fine.

Zubair286:
@raineorshine can you please review and confirm the functionality of this PR? Thanks

src/App.css
```diff
@@ -1453,3 +1453,104 @@ a.button-dim {
...
+}
+
+@keyframes bobble {
+  0%, 100% {
+    transform: translateX(-50%) translateY(0);
+  }
+  50% {
+    transform: translateX(-50%) translateY(10px);
+  }
+}
```
trevinhofmann, line 1556: My understanding has been that we want to avoid adding anything to `App.css` in favor of completely switching to Panda for anything new, but the bobble animation here looks fine to me so long as @raineorshine is okay with this.

If you switch to Panda, here is an example of a `pulse` animation defined in `panda.config.ts`: https://github.com/cybersemics/em/blob/66945aea4cf7d9c82a163f39f9822f1ca73f89f9/panda.config.ts#L35-L42

I see you have already defined the `arrowBobbleAnimation` in `panda.config.ts`, so adding the animation keyframe there as well would be my preference for consistency.

src/App.css
```diff
@@ -1453,3 +1453,104 @@ a.button-dim {
...
+
+.copy-icon-wrapper {
+  display: 'inline-flex';
+}
+
+.copy-icon-wrapper svg {
+  cursor: pointer;
+}
```
trevinhofmann, line 1547: Were these checkbox, gift code, and copy icon styles unintentionally added? I don't see how they relate to the pull request, so I assume they were accidentally included when resolving a merge conflict (or something similar).

src/components/LayoutTree.tsx
```diff
@@ -449,22 +484,51 @@ const LayoutTree = () => {
     [fontSize],
   )
 
+  /** Compare ranks between two thoughts. */
```
trevinhofmann, line 487: This is a pure function and can/should be defined outside the context of the `LayoutTree` component.

src/components/LayoutTree.tsx
```diff
@@ -617,6 +722,27 @@ const LayoutTree = () => {
...
+            transform: 'translateX(-50%)',
+            borderBottom: '20px solid rgb(155, 170, 220)',
+            ...(hoverArrowVisibility === 'below' && {
+              bottom: `${arrowBottom}px`,
+              rotate: '180deg',
+            }),
+            animation: `bobble ${token('durations.arrowBobbleAnimation')} infinite`,
+          }}
+        ></div>
+      )}
```
trevinhofmann, line 745: I recommend breaking this into a separate component with its own name such as `<HoverArrow>`. Being defined inline here without a name makes the intended functionality non-obvious to anyone skimming through the code.

src/components/LayoutTree.tsx
```diff
@@ -438,6 +444,35 @@ const LayoutTree = () => {
...
+    // Check if the dragged item is a thought and the drop zone is not a subthought
+    const isThought = item?.zone === 'Thoughts' && state.hoverZone === DropThoughtZone.ThoughtDrop
+    const sourceThoughtId = head(item?.path || [])
+
+    const sourceThought = isThought ? getThoughtById(state, sourceThoughtId) : null
+    return sourceThought
+  })
+
+  const newRank = useSelector(state => getSortedRank(state, head(contextParentPath), sourceThought?.value || ''))
+
```
trevinhofmann, line 475: Rather than overloading the `<LayoutTree>` file with all of this code, I would prefer to see it moved somewhere else in a separate file. @raineorshine recommended creating a custom hook, which is one of the better options for encapsulating this logic.

trevinhofmann (changes requested):
Thank you @Zubair286! The functionality all appears to work as expected when I test it locally, including the [edge case](https://github.com/cybersemics/em/pull/2369#issuecomment-2357606652) I mentioned previously.

I've added a few comments, mostly related to refactoring that should be implemented to avoid polluting the `<LayoutTree>` with this new code which should be encapsulated elsewhere.

src/App.css
Zubair286: Sure, I will add them in `panda.config.ts`

src/App.css
Zubair286: Yes, I believe so they got added during merge conflict during a rebase. I will remove them.

Zubair286:
@trevinhofmann can you review the latest code refactor changes? Thanks

src/util/compareRanks.ts
```diff
@@ -0,0 +1,23 @@
+/** Compare ranks between two thoughts. */
```
trevinhofmann, line 1: Is there a reason for the complexity in this function, where we convert the ranks to lists of numbers (separated by the decimal point) and compare those individually?

The function `compareRanks` is only used in one place to check for a negative number (`compareRanks(newRank, currentRank) < 0`), and it seems to return a negative number if and only if `rankA < rankB`. I might be misunderstanding some edge cases that this handles; can you please explain why we can't compare the ranks simply as `rankA - rankB`?

trevinhofmann (changes requested):
The refactor looks much better, thank you!

@raineorshine -- I noticed one edge case. If the user drags the thought into either the toolbar or footer, the arrow is still displayed even though the thought does not get dropped into the list. Is that functioning as expected?

Demos of the edge case:

https://github.com/user-attachments/assets/63f5a0ce-880e-427d-af2f-c9d8b7b819ca

https://github.com/user-attachments/assets/f25c1439-d02f-4bc1-80f7-e88779775796

src/util/compareRanks.ts
Zubair286: I was not too sure if there could be ranks like 1.1.2, 1.1.3 etc, but given the possibility I had the function written like this. It certainly can be done with just `rankA - rankB`, if there are not any ranks as above, please do correct me if I am wrong here.

trevinhofmann: Since the parameters are of type `number`, values such as `1.1.2` would not be possible.

Zubair286: Hmm, not sure why I assumed it was string, but in that case we can omit the `compareRanks` function and simple do `newRank - currentRank` and this would work just fine.

raineorshine:
Good catch. The arrow should only appear when hovering over a valid drop target.

Zubair286:
The current logic depends on `state.hoveringPath` to figure out where we are hovering, but when hovering over toolbar/footer it gives is the latest thought path, hence the arrow still shows, shouldn't it give `null` or a falsy value if we are not hovering over a valid drop target?

raineorshine:
Yes, I think you're right. `state.hoveringPath` should be null or undefined when not hovering over a valid thought drop target.

Zubair286:
Right, I guess we are gonna need to update its action to return `null` or `undefined` if not hovering over valid drop targets.

Zubair286:
I see `hoveringPath` is being updated in `useHoveringPath` hook. Now, there does not seem to be anyway direct to detect if we are hovering over an invalid drop target, like I mentioned [here](https://github.com/cybersemics/em/pull/2369#issuecomment-2371932118) on toolbar/footer it returns the last hovered path, so I think we can compare the previous path with current path if they are same then we can say that the drop target is not valid. Does that makes sense @raineorshine? Please correct if I am wrong about updating in `useHoveringPath` .

raineorshine:
Comparing the last path with the current path doesn't really make sense to me. 

It seems like the drag-and-drop monitor item should know the hover path.

Zubair286:
The drag and drop monitor item has the information of the dragged item not the drop path I believe.

raineorshine:
Ah, good point. I'm not sure then off the top of my head. It's going to take some investigation.

Zubair286:
Should it be a separate issue as its more of a redux state issue?.

raineorshine:
Sure, that's fair. Even though it affects this issue, it can be seen as out of scope. Presumably as soon as the hoveringPath is fixed, the arrow will show up at the appropriate time.

Zubair286:
Sure, Let me just push the minor code changes @trevinhofmann suggested.

src/components/HoverArrow.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <div
+        style={{
+          width: '0',
+          height: '0',
+          borderLeft: '10px solid transparent',
+          borderRight: '10px solid transparent',
+          position: 'absolute',
+          top: hoverArrowVisibility === 'above' ? scrollTop : undefined,
+          left: '50%',
+          zIndex: '99999',
```
trevinhofmann, line 22: Could you please assign a zIndex from the Panda config? We're trying to keep all of the zIndex values in one place for consistency and to avoid accidental conflicts: https://github.com/cybersemics/em/blob/57bed066075268fc63c1694bc5a05c666f506274/panda.config.ts#L251-L275

trevinhofmann (changes requested):
Looks good to me! I've tested a bit on mobile Safari as well. Thank you. I just have one refactor request regarding the `zIndex` constant (apologies for missing that until now), and then you've got my approval for the PR 👍

src/components/HoverArrow.tsx
Zubair286: Sure, thanks for pointing out.

Zubair286:
@trevinhofmann as suggested, I have moved the `zIndex` property for arrow in `panda.config.css`.

raineorshine (changes requested):
Tested on desktop.

# False positives

The arrow appears in these cases, but it should not.

## 4. False positive: Sort near top of screen

Screenshot 2024-09-27 at 12 28 54 PM

## 5. False positive: Sort near bottom of screen

Screenshot 2024-09-27 at 12 29 17 PM

## 6. False positive: Drag to the left (not a drop target)

Screenshot 2024-09-27 at 12 31 49 PM

# False negatives

The arrow does not appear in this case, but it should.

## 7. False negative: Sort below screen

Screenshot 2024-09-27 at 12 30 10 PM

Zubair286:
Thanks for pointing these cases out @raineorshine , I have an idea of why most of these are behaving like this. Just the false positive number 3 is the one that I think has the same condition as the toolbar/navbar, the `hoveringPath` is not null instead giving previous path. If we drag straight to left then the arrow does not show.

[Screencast from 09-27-2024 07:23:46 PM.webm](https://github.com/user-attachments/assets/56e50e64-d6c7-4296-93da-87efd907a771)

Zubair286:
@raineorshine are we measuring the height of toolbar in anyway in the app that can be reused currently?

raineorshine:
I don't believe so.

Zubair286:
Right, I think we need to calculate it, if just in `LayoutTree`.

Zubair286:
@raineorshine the way we drop thoughts when hovered over a `DropEnd`, I was hoping to use that logic to detect if a thought is hovering over `DropEnd` of sorted context parent, States etc, I don't think the logic used in DropEnd `showDropHover` selector can be applied in sorted context, or can it?

raineorshine:
`showDropHover` is executed within each `DropEnd` to determine whether the `drop-hover` should be shown or not. It doesn't work if the `DropEnd` is off screen, which is exactly when the arrows need to be calculated.

The arrows depend on three pieces of state:

- `state.hoveringPath`
- `state.hoverZone`
- First and last visible thought in the viewport.

With that information you can determine which arrows to show. You're probably going to have to fix `state.hoveringPath` since it is so integral to this.

Zubair286:
@raineorshine  I understand this, but with [this](https://github.com/cybersemics/em/pull/2369#issuecomment-2357606652) specific case, which is also responsible for false positive no. 2, the only problem being that hovering over `DropEnd` of `States` expects the thought to drop in the context. In this case `state.hoveringPath` correctly shows that we are hovering over `States`, but we cannot determine if we are hovering over the `DropEnd` or on top of it, which I think is important if we need this case to pass.

raineorshine:
There must be some way to detect it, because the information is somewhere. It is some combination of:

- Redux State
- Component state
- Drag-and-drop handlers

Once you to determine which information is needed to detect the necessary case, then it's just a question of how to expose/access it. Ideally it's derivable from Redux State, but if you have to add a new Redux State variable or use a ministore on this side that is possible. A ministore seems unlikely to me in this case since this is more related to State, hover state, and viewport.

Zubair286:
From what I have seen, in `useDragAndDropSubThought` `canDrop` function this allows for thought to be dropped on when hovered over `DropEnd`. Specifically the `isHidden` which simply allows for dropping in a subthought, and then we do render the DropEnd based on its rank but this is not enough. I will try to extract the necessary information, hope its not too complex.

`return (!isHidden || isClosestHiddenParent) && !isDescendant && !divider && !showContexts`

raineorshine:
The `useSortedContext` hook does not currently check `state.hoverZone`. Wouldn't that be sufficient to detect the `DropEnd` case?

All of the `canDrop` logic has already been calculated.

`canDrop` → `useDragAndDropSubThought` → `isHovering` → `useHoveringPath` → `dragInProgress`

Zubair286:
In the case of `DropEnd` `state.hoverZone` returns `SubthoughtsDrop`, that could work if we didn't need to cancel out the arrow when hovering over `DropChild` of context children.

raineorshine:
Yes it would be a combination of `state.hoverZone` and some condition on the path.

Zubair286:
Thanks, it works now we need to let the arrow know to not render if `DropHover` is within view if we are hovering over `DropEnd`.

Zubair286:
@raineorshine kindly please do review the latest code changes. Thanks

raineorshine:
Thank you! I will let @trevinhofmann review, starting with the test cases identified above and then looking at the implementation.

Zubair286:
@trevinhofmann please kindly review the changes. Thanks

trevinhofmann (changes requested):
Hi, @Zubair286! I apologize for the delay in getting this review done. I have noticed a few cases that are still incorrectly showing the indicator arrow:

### False positives:

#### 1. Dropping onto the drag-and-drop notice toolbar

https://github.com/user-attachments/assets/1bddafd0-66e8-4591-864e-fb6c4032d08d

#### 2. Dropping to the left of the list

https://github.com/user-attachments/assets/52707c3e-1394-4efe-8c4c-ad2f96f6fdc7

#### 3. Dropping onto the footer

https://github.com/user-attachments/assets/2b4dbd3b-b873-460e-813d-3f74c80e9798

From my testing of the cases in https://github.com/cybersemics/em/pull/2369#pullrequestreview-2333513696, I was unable to reproduce cases 1, 2, and 4 (they appear to be fixed 🎉). Case 3 (Drag to the left) appears to still be an issue, shown in the second video of this comment.

Zubair286:
@trevinhofmann thanks for reviewing, as per these [discussions](https://github.com/cybersemics/em/pull/2369#issuecomment-2373253487) all of the cases you mentioned are due to `state.hoveringPath`, which is separate redux state issue we agreed upon fixing in a separate issue.

raineorshine:
Okay, let's talk through this :). When we find ourselves debating the scope of an issue, it's usually a good idea to take a step back and consider 1) what is needed to properly fix the issue, and 2) how to reconcile that with the scope of work.

By every conceivable measure, the issue is not fixed. There are multiple cases that are not working. At the time that we considered fixing `state.hoveringPath` in a separate issue, it was when there appeared to only be a single case that wasn't covered. It is apparent now that the range of failures is greater than that and deserves a proper solution. It's also nontrivial to confirm that every failing case is the result of the same cause.

I don't believe that what is being asked is beyond the request of the original issue, but I recognize that this has been a considerable endeavor and I'm willing to increase the budget to make sure this gets done right. I recommend recording any additional hours you spend on the issue and I will pay you at your hourly rate to produce a solution to #1828 that covers all cases and evinces a clear design.

Another good reason to do this in the same PR is that it will be easier to evaluate the solution and provide overall design feedback. If we merge this PR and then make changes to `state.hoveringPath` in a separate PR, it may affect how this solution was designed. With this functionality so tightly coupled to `state.hoveringPath`, it's better to design one whole solution than to implement two partial solutions. Hopefully that makes sense, and please let me know if you have additional thoughts on the matter.

As for the particular solution, let's think about how `state.hoveringPath` was implemented at its conception, and what has changed. For its original purpose, it was only used when the user was hovering over a valid drop target. Therefore it didn't matter what the value was when the user was not on a valid drop target. If we are going to use `state.hoveringPath` when it is **not** over a valid drop target, then what event must we subscribe to in order to ensure `state.hoveringPath` is always accurate? Maybe react-dnd's `useDragDropManager' or the native `dragleave` event would work. Also, the `canDrop` handler of `useDragAndDropThought` and `useDragAndDropSubThought` has the necessary information to determine if the drop target is valid, but that seems more indirect. You would need to create some state in a side channel, and predicates ideally should not have side effects. And `canDrop` is called continuously during drag (many times a second), so there are performance considerations.

I'll let you investigate further and choose whichever one is more appropriate. Feel free to reach out for preliminary feedback, but please have it backed up with sound research and provisional suggestions.

Zubair286:
@raineorshine I can definitely see that even if the above cases are because of a different issue, solving it in this PR makes better sense as you mentioned its closely tied to #1828. I want to clear is that, the `state.hoveringPath` is being set in the `useHoveringPath` hook is that correct?

raineorshine:
Great, thanks!

> I want to clear is that, the `state.hoveringPath` is being set in the `useHoveringPath` hook is that correct?

`state.hoveringPath` is currently set in [dragInProgress](https://github.com/cybersemics/em/blob/fe3994ad5c55afbb938cb1ef44bea318e114f5e1/src/actions/dragInProgress.ts#L45). You are not limited to that however, and are welcome to add an action-creator for setting `state.hoveringPath` independently if that is needed.

Zubair286:
Hmm, I was thinking if the `isHovering` property from `useDrop` could be used because it is set to false when we are not hovering over a valid drop target even if dragInProgress is true.

raineorshine:
`useDrop` returns the hovering state of a specific drop target, but the arrows are concerned with the global hovering state. I'm not sure how you would go from individual hovering states to global hovering state. It would be preferable to subscribe to the global hovering state for any Thought item type.

Zubair286:
@raineorshine I wanted to discuss some solutions about the `state.hoveringPath` problem. I have some solutions so I would like some feedback from you on them.

The react-dnd `dragAndDropMonitor` does not provide any sort of data until we hover over valid drop targets so it is not much help so I tried working with drag events, dragleave, dragenter etc. The problem with them is that they will run for each thought, so it creates an unwanted complexity with the hover/arrow behavior. 

One solution that came to my mind was taking refs of `thought-container`, toolbar/bavbar and check if the cursor while dragging is within bounds of these elements, it adds a couple of states and has an edge case with the container where with each level drop there is some dead zone on left, but the arrow/hover will keep on showing until we are out of the container bounds.

Another solution that comes is to check if we are moving behind the x of the currently hovered over path, since we are left with the last hovered thoughts path, maybe we can use that to determine the cursor and that thoughts x position, and update `state.hoveringPath`.

raineorshine:
Thanks for sharing the options you are considering. 

> The react-dnd `dragAndDropMonitor` does not provide any sort of data until we hover over valid drop targets so it is not much help 

So no leave event?

> so I tried working with drag events, dragleave, dragenter etc. The problem with them is that they will run for each thought, so it creates an unwanted complexity with the hover/arrow behavior. 

I think a denounced handler might work here? After the last dragleave if no other thought is hovered over in 100ms then hoveringPath can be set to null. 

> One solution that came to my mind was taking refs of `thought-container`, toolbar/bavbar and check if the cursor while dragging is within bounds of these elements, it adds a couple of states and has an edge case with the container where with each level drop there is some dead zone on left, but the arrow/hover will keep on showing until we are out of the container bounds.

We don't want to tightly couple this behavior to unrelated components like toolbar, navbar, etc. 

> Another solution that comes is to check if we are moving behind the x of the currently hovered over path, since we are left with the last hovered thoughts path, maybe we can use that to determine the cursor and that thoughts x position, and update `state.hoveringPath`.

I'm not sure if x,y coordinates are the right level of operation here. We really just want to be working in terms of drag events. 

Are you sure that react-dnd does not have any leave hook?

Zubair286:
Thanks for the feedback.

Hmm, debounce might help. Let me see if this can give us desired results.

Understood.

Right, I only found this open [issue](https://github.com/react-dnd/react-dnd/issues/2183) in react-dnd which seems to relate to our problem.

raineorshine:
> Right, I only found this open [issue](https://github.com/react-dnd/react-dnd/issues/2183) in react-dnd which seems to relate to our problem.

Yes. Is the proposed hook applicable in our case?

Zubair286:
> > Right, I only found this open [issue](https://github.com/react-dnd/react-dnd/issues/2183) in react-dnd which seems to relate to our problem.
> 
> Yes. Is the proposed hook applicable in our case?

I have implemented it and it seems to at least work when not hovering over an invalid drop target but it also triggers between dragenter and dragleave on thoughts.

raineorshine:
> @raineorshine kindly review the latest fix of state.hoveringPath when you get the chance.

I will let Trevin do the first review.

> On another note, I need to discuss another issue that seem to relate to hovering over dropend of `States` context, rendering an arrow when over `States` dropend is not an issue, the challenge is that the arrow will not know if the hover is out of viewport or not. This is because when hovering over `Colorado`, which in this case is below `States` dropend, the rank that is generated is not of the sorted context because Colorado is not inside the sorted context. So, letting the arrow know if the hover is within viewport is difficult since the rank is not correct in this case. I would like your input on this, maybe there is something I am missing here. Thanks.

I don't think you would use rank in a sorted context. You can use [getChildrenSorted](https://github.com/cybersemics/em/blob/f3a239b55394cf3a0f669e568834deb25f866932/src/selectors/getChildren.ts#L76) to get a list of visible children in sorted order. From there you can determine the relative order of any two children.

Zubair286:
@raineorshine right now there were two solutions to work with for rendering the arrow.

1. By calculating the rank of the dragging thought and then taking the position of the drop point, and calculating if the drop position rank is above or below the viewport. This is how its working right now, hence the issue with drop end hovering.
2. Taking the first and last visible thoughts and calculating if the drop point is above/below the y position of them.

So by using the `getChildrenSorted` and taking the children of `States`, we get all of the children but checking if hovering on thought on `States` dropend 'Colorado' , is the issue here.

raineorshine:
The logic seems fairly straightforward:

1. We are hovering over the DropEnd of a sorted context, States.
2. The first visible child of the sorted context is Oklahoma and the last visible child is Wyoming.
3. The dragged thought, Wolorado, would be sorted (i.e. `getSortedRank`) after Oklahoma and before Wyoming, therefore it is in the viewport.

> I don't think you would use rank in a sorted context.

Never mind this... I forget that ranks are now modified when a context is sorted. So you can use the rank to determine the sort order.

trevinhofmann (changes requested):
In testing this locally, the functionality does appear correct to me on. The arrow is correctly not appearing when hovering the dragged thought over the toolbar, footer, and sides of the page. The arrow appears in the expected cases.

My main concern is around the complexity of the implementation, especially the coupling with the toolbar, and whether it's likely to continue to work as expected with future changes to the project.

src/components/LayoutTree.tsx
```diff
@@ -364,16 +370,24 @@ const LayoutTree = () => {
...
 
   // set the bullet width only during drag or when simulateDrop is true
   useLayoutEffect(() => {
     if (dragInProgress || testFlags.simulateDrop) {
       const bullet = ref.current?.querySelector('[aria-label=bullet]')
+      const toolbar = document.querySelector('#toolbar')
+
+      distanceFromTop.current = (ref.current?.getBoundingClientRect().top ?? 0) + scrollTop
+
+      toolbarRef.current = toolbar?.getBoundingClientRect().height
```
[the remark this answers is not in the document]
Zubair286: Well, yes I avoided using state, since for most of the time `toolbar` and `distanceFropTop` will not be changing unless we change the font size. So, keeping it inside ref and only updating it when `dragInProgress` is true.

src/components/LayoutTree.tsx
```diff
@@ -364,16 +370,24 @@ const LayoutTree = () => {
         state.cursor.length - (hasChildren(state, head(state.cursor)) ? 2 : 3)
       : 0,
   )
+  const scrollTop = scrollTopStore.useState()
 
   const [bulletWidth, setBulletWidth] = useState<number | undefined>(0)
+  const toolbarRef = useRef<number | undefined>(0)
```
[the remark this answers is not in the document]
Zubair286: I undertand, will update it to check for toolbar and remove the typing.

panda.config.ts
```diff
@@ -283,6 +294,7 @@ export default defineConfig({
           bullet: { value: 2 },
           stack: { value: 1 },
           hide: { value: -1 },
+          arrow: { value: 9999 },
```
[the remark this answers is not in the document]
Zubair286: Hmm, not sure if the value is incorrect, and change the value.

src/components/LayoutTree.tsx
```diff
@@ -599,8 +644,36 @@ const LayoutTree = () => {
...
+    if (isSortedContext && insertionY === null && treeThoughtsPositioned.length > 0 && sourceThought) {
+      const lastThought = treeThoughtsPositioned[treeThoughtsPositioned.length - 1]
+      insertionY = lastThought.y + lastThought.height
+    }
+
+    // Determine hoverArrowVisibility based on insertionY
+    if (insertionY !== null) {
+      if (insertionY > maxVisibleY) {
+        hoverArrowVisibility = 'below'
+      } else if (insertionY < scrollTop - (toolbarRef.current ?? 0)) {
```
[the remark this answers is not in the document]
Zubair286: @trevinhofmann  I do have a solution where we match the ranks of first and last thought with the `newRank` instead of checking the y values, while it works the only problem I see is there are edge cases when there is a little space left on top of the first visible thought, just enough for drophover to show, but since the rank will be generated for above the first thought arrow will render, like in the video below. We still would need to use the toolbar height in this case though.

[Screencast from 10-17-2024 01:43:59 PM.webm](https://github.com/user-attachments/assets/b1fb9c5e-8baf-4ca1-924c-ce33dc85a834)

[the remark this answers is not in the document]
raineorshine: I'm okay with that! While technically the hover is in the viewport, it's at the very top of the screen, above all visible thoughts, so the arrow does not feel inappropriate. Thanks for the attention to detail.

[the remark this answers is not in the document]
Zubair286: Understood, there are a couple of more edge cases that also needs to be addressed, like when last visible thought is also the last thought in context, same for first, the drop position rank is after the last/first, even in those cases arrow would render. I believe these needs to be fixed with this approach.

[Screencast from 10-17-2024 04:19:26 PM.webm](https://github.com/user-attachments/assets/10b6c338-9092-47b1-8c72-f578e71cba35)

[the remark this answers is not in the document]
raineorshine: Yes, that one should be addressed. It doesn't look right if the arrow appears when the DropHover is well within the viewport.

src/components/LayoutTree.tsx
```diff
@@ -624,11 +697,13 @@ const LayoutTree = () => {
...
   // Subtract singleLineHeight since we can assume that the last rendered thought is within the viewport. (It would be more accurate to use its exact rendered height, but it just means that there may be slightly more space at the bottom, which is not a problem. The scroll position is only forced to change when there is not enough space.)
   const spaceBelow = viewportHeight - navAndFooterHeight - CONTENT_PADDING_BOTTOM - singleLineHeight
 
+  // Calculate the position of the arrow relative to the bottom of the container.
+  const arrowBottom = totalHeight + spaceBelow - scrollTop - viewportHeight + navAndFooterHeight
```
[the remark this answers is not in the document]
Zubair286: I do feel that there are a lot of variables involved but I can't see a proper positioning of the arrow without using them.

Zubair286:
@trevinhofmann please review the latest changes. Thanks

panda.config.ts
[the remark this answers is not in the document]
trevinhofmann: Thank you. I'd have maybe gone even lower, but `1400` (less than `popup`) seems more reasonable than `9999`.

src/components/LayoutTree.tsx
```diff
@@ -364,16 +370,24 @@ const LayoutTree = () => {
...
 
   // set the bullet width only during drag or when simulateDrop is true
   useLayoutEffect(() => {
     if (dragInProgress || testFlags.simulateDrop) {
       const bullet = ref.current?.querySelector('[aria-label=bullet]')
       setBulletWidth(bullet?.getBoundingClientRect().width)
+
+      const toolbar = document.querySelector('#toolbar')
+
+      if (toolbar) toolbarHeightRef.current = toolbar?.getBoundingClientRect().height
```
trevinhofmann, line 387:
```suggestion
      if (toolbar) toolbarHeightRef.current = toolbar.getBoundingClientRect().height
```

Minor adjustment. We don't need the conditional `?` since the `if` condition has already assured us that the `toolbar` is defined.

Side note, @raineorshine: This ESLint rule could check (and auto-fix) for these: https://typescript-eslint.io/rules/no-unnecessary-condition/

src/components/LayoutTree.tsx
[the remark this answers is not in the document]
trevinhofmann: If we use `useState`, it shouldn't trigger a rerender if the value hasn't changed (React does a shallow comparison), so there shouldn't be any unusual performance hit. I still believe it would make more sense for this to be a state value rather than a ref. Please let me know if I'm missing something.

src/components/LayoutTree.tsx
```diff
@@ -454,6 +468,13 @@ const LayoutTree = () => {
     ),
   )
 
+  const { footerHeight, navbarHeight } = useNavAndFooterHeight()
+  const navAndFooterHeight = navbarHeight + footerHeight
+
+  const maxVisibleY = viewportHeight + scrollTop - ((distanceFromTop.current ?? 0) + navbarHeight)
```
trevinhofmann, line 474: Similar to `toolbarHeightRef`, we can simplify the type of `distanceFromTop` to only be `number` rather than `number | undefined`. It's nice to handle edge cases (the element not being found) upfront rather than letting those unknown/undefined values affect downstream usage as we see here with the `?? 0` default.

src/hooks/useDragLeave.ts
```diff
@@ -0,0 +1,57 @@
+import { debounce } from 'lodash'
+import { useEffect, useRef } from 'react'
+import { useDispatch } from 'react-redux'
+import { updateHoveringPathActionCreator } from '../actions/updateHoveringPath'
+import useSortedContext from './useSortedContext'
+
+const DEBOUNCE_DELAY = 50
+let hoverCount = 0
```
trevinhofmann, line 8: Wouldn't this hook fail to work properly if used in multiple components, since we're using module-scoped variables here? I believe `hoverCount` should be tracked within the scope of the function rather than up at the top here.

src/hooks/useDragLeave.ts
```diff
@@ -0,0 +1,57 @@
...
+    debouncedSetHoveringPath = debounce(() => {
+      // Only set hoveringPath to undefined if hoverCount is still zero
+      dispatch(updateHoveringPathActionCreator({ path: undefined }))
+    }, DEBOUNCE_DELAY)
+  }
+
+  useEffect(() => {
+    if (isDeepHovering && !prevIsDeepHoveringRef.current) {
+      // Cursor has entered a drop target
+      hoverCount += 1
```
trevinhofmann, line 30: I may be misunderstanding `hoverCount`, but should we not also be decrementing whenever leaving the drop target? Here, it looks like we only increment when entering and decrement once during the cleanup.

This _might_ be the reason I'm seeing this issue where the arrow sometimes fails to appear when leaving and reentering the drop target:

https://github.com/user-attachments/assets/2ebbd4ae-ab78-441e-bd98-7ce9635be3d2

src/components/LayoutTree.tsx
```diff
@@ -470,17 +491,29 @@ const LayoutTree = () => {
...
     treeThoughtsPositioned: TreeThoughtPositioned[]
+    hoverArrowVisibility: 'above' | 'below' | null
   } = useMemo(() => {
     // y increases monotically, so it is more efficent to accumulate than to calculate each time
     // x varies, so we calculate it each time
     // (it is especially hard to determine how much x is decreased on cliffs when there are any number of tables in between)
     let yaccum = 0
     let indentCursorAncestorTables = 0
 
+    // Arrow visibility based on insertionY
```
trevinhofmann, line 507: This comment can be updated now that we're using the ranks rather than the `insertionY` calculation.

src/components/LayoutTree.tsx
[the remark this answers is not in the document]
trevinhofmann: The arrow does still appear when the thought above/below the target (Louisiana in this example) is _partially_ in view. Personally, I think that is still acceptable, but I wanted to make sure we're aware of it:

https://github.com/user-attachments/assets/118db3bb-02a8-4bde-beff-5e38f720456f

trevinhofmann (changes requested):
Thank you, @Zubair286! I much prefer this rank/dragLeave approach where we're not so tightly coupled with other components.

src/hooks/useDragLeave.ts
Zubair286: Hmm, well yes it is supposed to be decremented, I did have it in place though, not sure how it got removed. I'll add the decrementing of `hoverCount` in else-if statement. Thanks for pointing out.

src/hooks/useDragLeave.ts
Zubair286: @trevinhofmann  We might have to turn `hoverCount` in a ref or state, so the value isn't reset on every render since it is being updated inside a `useEffect` hook.

raineorshine:
The arrow is incorrectly showing up when `canDrop` is false, such as when dragging a thought over its descendant:

Brave Browser 2024-10-20 18 01 44

I haven't done a full behavior review, but wanted to point this out as it seems to be a new one.

raineorshine:
# Parent highlight is broken

Sometimes the parent thought not highlighted:

## Current Behavior

Brave Browser 2024-10-20 18 06 31

## Expected Behavior

image

raineorshine:
After the latest change requests are made, please test as many edge cases as you can. I added all the test cases from our discussion to the OP so that you can more easily go through them systematically. Once you are satisfied that there are no false positives or false negatives, Trevin can review. Then I will do a final review. Thanks!

Zubair286:
@raineorshine thanks for pointing out these edge cases, I will check for these and any other that needs be tested. This issue seems to touch to a lot of cases, will try and look out for all of them.

src/hooks/useDragLeave.ts
Zubair286: @trevinhofmann should we move the `useDragLeave` hook code into the `Thought` component where its called? The hook likely will not be called into other components, please correct me if I am wrong though. Thanks

Zubair286:
@trevinhofmann please review the latest changes, I have tested the cases mentioned in the OP kindly do on your end as well. Thanks

src/hooks/useDragLeave.ts
trevinhofmann: I would prefer to leave `useDragLeave` as a separate file which and cleaning it up just a bit (e.g. move `hoverCount` and `debouncedSetHoveringPath` into the function, likely as states or refs) so that it _could_ be reused elsewhere. Adding large hooks to a React component's file adds too much complexity in one place for my preference.

@raineorshine -- Do you agree?

trevinhofmann (changes requested):
When I tested, `1-7` and `9` are all working as expected.

However, I was able to reproduce `8. False positive when canDrop is false`:

https://github.com/user-attachments/assets/15b82094-2da4-472c-86f4-8e371977a028

Zubair286:
@trevinhofmann its confusing I can not reproduce this, I have checked logs and the `hoverArrowVisibility` is correctly null in this case.

[Screencast from 10-25-2024 11:19:17 AM.webm](https://github.com/user-attachments/assets/a026bfa7-9854-450c-9ff9-e943e689d398)

Zubair286:
> @trevinhofmann its strange I can not reproduce this, I have checked logs and the `hoverArrowVisibility` is correctly null in this case for me.
> 
>  Screencast.from.10-25-2024.11.19.17.AM.webm

Nevermind this I imported the context again and it reproduced.

I have pushed the fix, can you confirm @trevinhofmann?

trevinhofmann (approved):
Thanks, @Zubair286! All of `1-9` are now working as expected for me.

src/components/HoverArrow.tsx
```diff
@@ -0,0 +1,36 @@
...
+const HoverArrow = ({ hoverArrowVisibility, scrollTop, arrowBottom }: HoverArrowType) => {
+  return (
+    hoverArrowVisibility && (
+      <div
+        style={{
+          width: '0',
+          height: '0',
+          borderLeft: '10px solid transparent',
+          borderRight: '10px solid transparent',
+          position: 'absolute',
```
raineorshine, line 19: Why `position: absolute` instead of `position: fixed`? It seems like `position: absolute` would involve a lot more manual calculation to get it in the viewport, while `position: fixed` does that for you.

src/components/LayoutTree.tsx
```diff
@@ -364,16 +370,27 @@ const LayoutTree = () => {
...
   useLayoutEffect(() => {
     if (dragInProgress || testFlags.simulateDrop) {
       const bullet = ref.current?.querySelector('[aria-label=bullet]')
-      setBulletWidth(bullet?.getBoundingClientRect().width)
+      if (bullet) setBulletWidth(bullet?.getBoundingClientRect().width)
+
+      const toolbar = document.querySelector('#toolbar')
+      if (toolbar) setToolbarHeight(toolbar.getBoundingClientRect().height)
+
+      setDistanceFromTop((ref.current?.getBoundingClientRect().top ?? 0) + scrollTop)
```
raineorshine, line 391: You add `scrollTop` here only to subtract it [below](https://github.com/cybersemics/em/pull/2369/files#diff-ed74b196536d8a76f93f0d172796b6356ed7b7d3775bc7840ce7b15648dfcfb2R477). i.e. I think it cancels out when you calculate `maxVisibleY`. Therefore, you can remove `scrollTop` from these calculations and avoid having to trigger `useLayoutEffect` on scroll.

raineorshine (changes requested):
I have tested all the cases and can confirm they all work! Amazing job.

Now just some optimization and cleanup.

src/hooks/useSortedContext.ts
```diff
@@ -0,0 +1,59 @@
...
+    const dropTargetId = head(state.hoveringPath || [])
+    const isContextParent = attributeEquals(state, dropTargetId, '=sort', 'Alphabetical')
+
+    return state.hoverZone === 'SubthoughtsDrop' && isContextParent
+  })
+
+  const dragDropManager = useDragDropManager()
+
+  // get the source thought and its new rank
+  const sourceThought = useSelector(state => {
```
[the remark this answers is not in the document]
raineorshine: fc2d78b268

src/hooks/useSortedContext.ts
```diff
@@ -0,0 +1,59 @@
...
+    return sourceThought
+  })
+
+  const contextpath = hoveringOnDropEnd ? hoveringPath : contextParentPath
+  const newRank = useSelector(state => getSortedRank(state, head(contextpath), sourceThought?.value || ''))
+
+  return {
+    isSortedContext,
+    newRank,
+    hoveringOnDropEnd,
```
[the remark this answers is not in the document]
raineorshine: 58b3012d99

src/components/LayoutTree.tsx
```diff
@@ -504,6 +540,7 @@ const LayoutTree = () => {
     const tableCol1Widths = new Map<ThoughtId, number>()
     const treeThoughtsPositioned = treeThoughts.map((node, i) => {
       const next: TreeThought | undefined = treeThoughts[i + 1]
+      const state = store.getState()
```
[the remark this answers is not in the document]
raineorshine: 510b80ef3a and d54600f1b3

src/components/LayoutTree.tsx
raineorshine: f7404afb9a

raineorshine:
Comments with commit hashes were minor issues that I fixed. I recommend taking a look at the commits to see the change.

That just leaves two questions:

- https://github.com/cybersemics/em/pull/2369#discussion_r1817901466
- https://github.com/cybersemics/em/pull/2369#discussion_r1817907496

Zubair286:
@raineorshine Thanks for the minor fixes, the codebase looks much better now. As for your questions,

1. Yes, my concern was just that when hovering into context the new rank that is being generated could be negative. It did not occur to me that ranks could also be negative inside sorted context, in which case I think `newRank > 0` should be removed to allow for negative ranks as well.
2. In [this discussion](https://github.com/cybersemics/em/pull/2369#discussion_r1746027535), we agreed to use `position: absolute` to make it work on mobile safari. Previously `position: fixed` was used.

raineorshine:
> 1. Yes, my concern was just that when hovering into context the new rank that is being generated could be negative. It did not occur to me that ranks could also be negative inside sorted context, in which case I think `newRank > 0` should be removed to allow for negative ranks as well.

Sounds good!

> 2\. In [this discussion](https://github.com/cybersemics/em/pull/2369#discussion_r1746027535), we agreed to use `position: absolute` to make it work on mobile safari. Previously `position: fixed` was used.

Ah, yes. I forget. I wish `position: fixed` worked on mobile Safari!

raineorshine (approved):

src/components/LayoutTree.tsx
```diff
@@ -599,8 +657,30 @@ const LayoutTree = () => {
       }
     })
 
-    return { indentCursorAncestorTables, treeThoughtsPositioned }
-  }, [fontSize, sizes, singleLineHeight, treeThoughts])
+    // Determine hoverArrowVisibility based on newRank and the visible thoughts
+    if (newRank > 0 && (isSortedContext || hoveringOnDropEnd)) {
```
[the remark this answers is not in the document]
raineorshine: 1d497d1592

raineorshine:
Thanks for all your hard work ❤️

Issue #1602, opened 2022-06-19
Add DropHover while waiting for drag expand


raineorshine:
## Steps to Reproduce

When dragging a thought, hold the cursor near the top of a thought with hidden parents for a second. The parent will expand to allow you to drag the thought higher up the tree.

https://user-images.githubusercontent.com/750276/174482113-e13d4c59-aee4-4722-83a3-4edc85c8f5eb.mov

## Current Behavior

~~The active hover zone for this behavior is a very small area (indicated in blue).~~

There is no visual feedback when the user is hovering above a parent that can be expanded.

## Expected Behavior

~~The active hover zone should be more generous to make it easier to activate this behavior (indicated in pink).~~

- When the user is dragging a thought and hovering in the expand zone (i.e. the area triggers the drag expand timer), render a [DropHover](https://github.com/cybersemics/em/blob/6275018b29631e5d7939fab42e677b8137b8b00d/src/components/DropHover.tsx) (blue bar) above the thought whose parent will be expanded.
  - In the above screen capture, above `b`.
  - Avoid rendering two DropHovers at once. Consider using the existing DropHover, and changing the rendering condition rather than rendering a new DropHover.
- Add a pulse animation (between light and dark blue) to the DropHover. It should pulse faster compared to the animation in #1954.
- The DropHover should only be rendered while the expansion timer is running, and should be removed as soon as the parent is expanded.
- Lastly, decrease the fade transition duration of the expanded parent.

PR #2372, opened 2024-09-06
Animate hover while waiting for thought to expand

Zubair286:
#1602

### DIFF
--- a/panda.config.ts
+++ b/panda.config.ts
@@ -191,6 +191,12 @@ export default defineConfig({
               _test: '0s',
             },
           },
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Works great! Thanks. 

Two extra `calculateAutofocus` calls on every DropHover is a bit expensive, so I've split the component into two so that we can short circuit it when the DropHover is hidden. See: 1ed11c0f6b

I've also combined a few of the useSelectors for efficiency in 685abd8ad2. Otherwise the component will re-render if any of the intermediate values change.

PR #2373, opened 2024-09-06
PandaCSS - shortcuts table + add conditions

yangchristina:


src/components/ShortcutRow.tsx
```diff
@@ -115,7 +118,8 @@ const ShortcutRow = ({ customize, onSelect, selected, shortcut, indexInToolbar }
           <b>{shortcut.label}</b>
           <p>{description}</p>
         </th>
-        <td>
+        {/* center gesture diagrams on mobile */}
+        <td className={css({ minWidth: { base: '10rem', _mobile: 'auto' }, textAlign: { _mobile: 'center' } })}>
```
raineorshine, line 122: Looks good 👍

raineorshine (changes requested):
Thank you! Here are the issues I'm seeing.

# 1. Command Palette

## Steps to Reproduce

Activate the Command Palette (⌘+P).

## Current Behavior

7cbffa6dd5

Brave Browser 2024-09-06 17 36 53

## Expected Behavior

`main`

Brave Browser 2024-09-06 17 36 38

# 2. Help Screen

## Steps to Reproduce

Click the (?) icon in the footer to open the Help screen.

## Current Behavior

`main` (1821f84951)

Brave Browser 2024-09-06 17 39 10

## Expected Behavior

1821f84951 (or any commit before 1821f84951)

Brave Browser 2024-09-06 17 39 34

# 3. Gesture Library

## Steps to Reproduce

Help → Gesture Library

## Current Behavior

Margins have changed.

Brave Browser 2024-09-06 17 47 56

## Expected Behavior

Layout should be unchanged.

Brave Browser 2024-09-06 17 48 51

yangchristina:
@raineorshine I'm not sure if I have the same as you, but I don't have that showing:
Screenshot 2024-09-06 at 5 20 36 PM

yangchristina:
I added a fix for the modal z-indices. I don't think I was the one to introduce this bug, since I'm pretty sure all modals before where using the `popover` class, and therefore the `popover` z-index, but it also could have been me.

raineorshine (changes requested):
> @raineorshine I cannot reproduce the issues

Thanks for confirming! There must be some vite caching that I'm not used to yet.

> I added a fix for the modal z-indices. I don't think I was the one to introduce this bug, since I'm pretty sure all modals before were using the popover class, and therefore the popover z-index, but it also could have been me. Either way it's not one that really causes problems.

Great, thanks! Could you actually add a z-index in between modal and pop-up for the command palette? It is possible (though uncommon) for all three to be open at once. I didn't anticipate this when I originally gave it the modal class.

yangchristina:
@raineorshine yes there actually is a --z-index-command-palette between the two, it's just not being used (currently command palette uses popup z-index)

raineorshine:
Great. Please go head and change the command palette to use it. I'm not sure what happened there.

raineorshine (approved):
I forgot that the CommandPalette was actually rendered as a `<Popup>`. This should work though!

raineorshine (approved):

PR #2376, opened 2024-09-07
PandaCSS - convert thought, thought-annotation, editable

yangchristina:


raineorshine:
Looks like the Puppeteer snapshot tests are failing. Can you take a look? You can check the diff in the "Upload snapshot diff artifact" workflow step.

yangchristina:
@raineorshine I've fixed it now, thanks!

src/recipes/conditionalMultiline.ts
```diff
@@ -0,0 +1,32 @@
+import { defineRecipe } from '@pandacss/dev'
+
+const conditionalMultilineRecipe = defineRecipe({
```
raineorshine, line 3: I'm not clear why you have `conditionalMultilineRecipe` and `multilineRecipe`. Can you explain?

src/recipes/conditionalMultiline.ts
```diff
@@ -0,0 +1,32 @@
...
+  className: 'multiline',
+  description: 'meant to override other styles when applied',
+  base: {},
+  variants: {
+    apply: {
```
raineorshine, line 8: For applying or not applying an entire recipe, it would be better to just use `base` styles and put the condition in React:

```ts
cx(multiline ? multilineRecipe() : null, ...)
```

raineorshine (changes requested):
Thanks! This is a big one 😅. I'm not able to test each change individually, but it appears you've been meticulous in converting each complex descendant selector. Hoping for the best! 🤞

src/components/Thought.tsx
```diff
@@ -149,23 +150,6 @@ const ThoughtContainer = ({
     return !isExpandedHoverTopPath && !!cursor && equalPath(cursorGrandparent, path)
   })
 
-  // Note: If the thought is the active expand hover top path then it should be treated as a cursor parent. It is because the current implementation allows tree to unfold visually starting from cursor parent.
-  const isCursorParent = useSelector(state => {
```
raineorshine, line 153: The `isCursorParent` definition seems to have been removed completely. Was that a mistake or am I missing something? Thanks!

src/components/Bullet.tsx
```diff
@@ -35,27 +36,256 @@ interface BulletProps {
...
+        opacity: 0.8,
+        transition: 'opacity 0.75s ease-in-out',
+        marginRight: '-5px',
+        left: '4px',
+        fontSize: '28px',
+      },
+    },
+  },
+  variants: {
+    leaf: { true: {} },
```
raineorshine, line 73: Empty variant?

src/components/ErrorBoundaryContainer.tsx
```diff
@@ -27,7 +27,6 @@ const Toggle = ({ children, expand, title }: { children?: any; expand?: boolean;
         >
           <g>
             <path
-              className='glyph-fg triangle'
```
raineorshine, line 30: Where does the `ErrorBoundaryContainer` toggle get styled now?

src/components/Bullet.tsx
yangchristina: It is used in compound variants

src/components/ErrorBoundaryContainer.tsx
yangchristina: I believe it was never styled before, I inspected it, and the classes for glyph-fg and triangle weren't applied, since all the styles where where descendant selectors.

src/recipes/conditionalMultiline.ts
yangchristina: Not deleting multilineRecipe originally was a mistake, but I've changed to using it with base styles like you said below.

src/recipes/conditionalMultiline.ts
yangchristina: Fixed!

src/components/Thought.tsx
yangchristina: Yes sorry about that, I'm not sure how that wound up in there, considering it wasn't even in the right commit

yangchristina:
Sorry the Thought commit turned out so large. There were a lot of descendant selectors involved, and I could have done a better job of splitting things up better.

src/components/ErrorBoundaryContainer.tsx
raineorshine: Okay, thanks! That's very strange, as I did a git blame and it never seemed to be connected even when it was first created 🤷‍♀️

For dead code that is removed, if you're able to call that out in the PR for me that would be great. Will save me time on the review.

raineorshine:
That's okay. The `.thought` conversion may be the messiest one we have due to all the crazy selectors.

src/components/ContextBreadcrumbs.tsx
```diff
@@ -118,11 +119,13 @@ const BreadCrumb = React.memo(
...
             ellipsize(decodeCharacterEntities(value))
           ) : label === HOME_TOKEN ? (
-            <HomeLink color='gray' size={16} />
+            <HomeLink color='gray' size={16} className={css({ position: 'static' })} />
           ) : (
             <Link className='extend-tap-small' simplePath={simplePath} label={label} />
           ))}
-        {!isDeleting && <Superscript simplePath={simplePath} />}
+        {!isDeleting && (
+          <Superscript simplePath={simplePath} css={{ position: 'relative', left: '-2px', top: '-3px' }} />
```
raineorshine, line 127: Passing an inline array or object literal as a prop results in a different ref every render, preventing the component from being memoized. We need to ensure each prop has a stable object reference if the value does not change.

In this case since the styles are completely static, you can define them in a `const` at the module level. This unfortunately moves them further from where they are used, and forces us to have an extra variable, but there is no other way to stabilize the object reference when passing in the `css` object.

Note that `className` does not have this problem because it is a `string` and strings are always compared by value.

src/components/ContextBreadcrumbs.tsx
```diff
@@ -216,7 +219,12 @@ const ContextBreadcrumbs = ({
       - The "c/d" context will render "d" as a thought and "c" as the breadcrumbs.
     */
         !homeContext ? (
-          <HomeLink color='gray' size={16} style={{ position: 'relative', left: -1, top: 2 }} />
+          <HomeLink
+            className={css({ position: 'static' })}
+            color='gray'
+            size={16}
+            iconStyle={{ position: 'relative', left: -1, top: 2 }}
```
raineorshine, line 226: This also has the object reference problem, although it's not as significant since `HomeLink` is rendered rarely, and is a very light component.

src/components/Bullet.tsx
```diff
@@ -206,6 +447,9 @@ const Bullet = ({
     return styles?.color
   })
 
+  const isExpanded = useSelector(state => !!state.expanded[hashPath(path)])
+  const isCursor = isCursorParent || isCursorGrandparent || isEditing || isExpanded
```
raineorshine, line 451: Can you help me understand this `isCursor` variable? It appears to be true when the thought itself (`isEditing`), parent (`isCursorParent`), or grandparent (`isCursorGrandparent`) is the cursor, _or_ when the thought is expanded. I don't understand what those states have in common. Clearly a Bullet can be expanded without having the cursor on it.

src/recipes/thought.ts
```diff
@@ -0,0 +1,48 @@
...
+      marginLeft: { _android: '0.5em' },
+    },
+    '@media (min-width: 560px) and (max-width: 1024px)': {
+      marginTop: { _android: '-0.1px' },
+      marginLeft: { _android: '0.5em' },
+    },
+  },
+  variants: {
+    /** Assign annotation height on single line truncated url. */
+    singleline: {
```
raineorshine, line 23: I know this was the existing name, but can you rename `singleline` to `ellipsizedUrl`? 

I just realized that calling it `singleline` makes it seem like it is the opposite of `multiline`, but it's really a different concept.

src/components/VirtualThought.tsx
```diff
@@ -131,7 +131,7 @@ const VirtualThought = ({
     // Need to grab max height between .thought and .thought-annotation since the annotation height might be bigger (due to wrapping link icon).
     const heightNew = Math.max(
       ref.current.getBoundingClientRect().height,
-      ref.current.querySelector('.thought-annotation')?.getBoundingClientRect().height || 0,
+      ref.current.querySelector('[aria-label="text-annotation"]')?.getBoundingClientRect().height || 0,
```
raineorshine, line 134: I think this should be `[aria-label="thought-annotation"]`.

src/components/ErrorBoundaryContainer.tsx
yangchristina: Sounds good! I'll let you know next time.

raineorshine (changes requested):
Thank you 🙂

src/components/Bullet.tsx
yangchristina: I wasn't really sure what to name it, isExpanded might make more sense, but that is used already.

it comes from these styles
```css
.child.expanded:not(.leaf) > .thought-container > .bullet .glyph .glyph-fg.triangle,
.child.editing:not(.leaf) > .thought-container > .bullet .glyph .glyph-fg.triangle,
.child.cursor-parent:not(.leaf) > .thought-container > .bullet .glyph .glyph-fg.triangle,
.child.cursor-grandparent:not(.leaf) > .thought-container > .bullet .glyph .glyph-fg.triangle 
```

src/recipes/thought.ts
yangchristina: Yes, of course.

src/components/Bullet.tsx
raineorshine: I see! Yes, that's literally if the bullet is expanded. We can call it `isBulletExpanded` for now.

I'm not sure why it's not covered by the first selector `.child.expanded:not(.leaf) > .thought-container > .bullet .glyph .glyph-fg.triangle` since `.editing` should always be `expanded`. But simplifying that is a different task. We'll focus on converting most the styles as-is before we do any serious refactoring or simplifying.

yangchristina: maybe just `expand: true` would work

src/components/ContextBreadcrumbs.tsx
```diff
@@ -110,6 +111,7 @@ const BreadCrumb = React.memo(
       verticalAlign: 1,
       userSelect: 'none',
     }
+    const superscriptCss = { position: 'relative', left: '-2px', top: '-3px' }
```
raineorshine, line 114: Actually this will create a new object reference every call to render. You have to move it to the module scope at the top of the file :).

raineorshine (approved):

Issue #1386, opened 2021-08-04
Extend droppable area when there is empty space below a thought


raineorshine:
If there is empty space below a thought, extend the droppable area of the drop target to make it easier to drop where desired.

## Steps to Reproduce

```
- a
  - b
    - c
    - d
- x
```

1. Set the cursor on `d`
2. Drag `c` and drop it below `d`

## Current Behavior

There is a very small amount of space differentiating the `x`, `a`, and `b` drop targets, making it hard to choose the correct one.

<img width="403" alt="image" src="https://github.com/cybersemics/em/assets/750276/90d83dc8-1247-4684-8aff-b33c28dea547">

## Expected Behavior

We should take advantage of the empty space and increase the size of the drop targets at the bottom of the visible thoughts.

- Extend the height of `b` and `d` by 1em.
- Extend the height of `a` by 1.2em
- Extend the height of `x` by 1.4em
- etc

There can only be one child (`d`) and one sibling (`b`), but many cliff drop targets (`x` and `a`).

![Image 6-20-24 at 1 11 PM](https://github.com/cybersemics/em/assets/750276/87b194f3-9ee9-4cc0-91db-6e1cfac7f854)

Zubair286:
Let me confirm, we need to increase the height of `DropEnd` relative to the depth starting from deepest like b`b`and `d`, do I understand this right?

raineorshine:
I'm not sure exactly if it's `DropEnd`, but yes, the height of the drop targets at the cliff should be increased relative to depth.

PR #2381, opened 2024-09-09
Extend height of drop targets on cliff

Zubair286:
#1386


src/util/cliffThoughtHeight.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Calculate the height that cliff thoughts need to extend.
+ */
+export const calculateCliffThoughtsHeight = ({ deepestDepth, depth }: { deepestDepth?: number; depth?: number }) => {
+  if (depth === undefined) return 0
+
+  // Calculate dynamic height: increase as depth decreases
+  const baseHeight = 1
+  const heightIncrease = 0.2
```
raineorshine, line 9: Let's try a `heightIncrease` of `0.5` actually. `0.2` is a bit tight in practice, and we have lot of space available.

src/util/cliffThoughtHeight.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Calculate the height that cliff thoughts need to extend.
+ */
+export const calculateCliffThoughtsHeight = ({ deepestDepth, depth }: { deepestDepth?: number; depth?: number }) => {
```
raineorshine, line 4: This project uses the convention of a single, default export with the same name as the file.

src/util/cliffThoughtHeight.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Calculate the height that cliff thoughts need to extend.
```
raineorshine, line 2: `Calculates the height of a drop target at the cliff, decreasing with depth depth. This makes it easier to drop a thought at the desired level. Returns height in ems.`

raineorshine (changes requested):
The behavior looks good, thanks.

Let's talk about the implementation. Using a mutable variable within a render function is not recommended. It's creating a situation which is very difficult to reason about—namely that `deepestDepth` increases in each loop, but is being passed to `VirtualThought` in each loop as well, which means that different `VirtualThought` instances are going to have different `deepestDepth` values. So `deepestDepth` seems to converge on the cliff height, but it's not clear which value `VirtualThought` gets and why.

Each node in `treeThoughtsPositioned` has a depth, so it seems like all the information you need is already there. Conceptually you should be looking for how to derive the value you need from the existing data, not build up a new value in the render loop. Plus, you know from other issues that LayoutTree works in two distinct render stages: `treeThoughts` and `treeThoughtsPositioned`. That is where novel layout calculations should go. If it's not a novel layout calculation, it should be expressable in a stateless way further down in the tsx.

On a different note, is there any reason you can't use the cliff height, i.e. `-cliff`? It seems very close to the concept of `deepestDepth`, but tell me if there is a distinction I'm missing.

src/util/cliffThoughtHeight.ts
Zubair286: Alright, I will factor it out in a `util` function

Zubair286:
You are right, `-cliff `really does the trick instead of calculating `deepestDepth`. I am not sure how I missed that.

src/util/cliffThoughtHeight.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Calculates the height of a drop target at the cliff, decreasing with depth depth. This makes it easier to drop a thought at the desired level. Returns height in ems.
+ */
+const calculateCliffThoughtsHeight = ({ deepestDepth, depth }: { deepestDepth?: number; depth?: number }) => {
```
raineorshine, line 4: Let's call this `calculateCliffDropTargetHeight`. It's a little more verbose, but it's more accurate.

We are now using `cliff` for the `deepestDepth`, so let's rename the parameter name to match. It's not good to use multiple identifier names for the same concept.

src/util/cliffThoughtHeight.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (depth === undefined) return 0
+
+  // Calculate dynamic height: increase as depth decreases
+  const baseHeight = 1
+  const heightIncrease = 0.5
+  const dynamicHeight = baseHeight + heightIncrease * ((deepestDepth || depth) - depth)
```
raineorshine, line 10: Once you rename `deepestDepth` to `cliff` you'll need to flip the sign here to `-cliff` for consistency.

I'm regretting that I originally made cliff negative, but if we're going to flip all the signs we should do that in a separate PR.

src/util/cliffThoughtHeight.ts
```diff
@@ -0,0 +1,15 @@
+/**
```
raineorshine, line 1: This file should have same name as the default export.

src/components/DropUncle.tsx
```diff
@@ -9,19 +9,34 @@ import useDragAndDropThought from '../hooks/useDragAndDropThought'
...
   const value = useSelector(state =>
     testFlags.simulateDrop ? getThoughtById(state, head(simplePath))?.value || '' : '',
   )
 
+  const dragInProgress = useSelector(state => state.dragInProgress)
   const { isHovering, dropTarget } = useDragAndDropThought({ path, simplePath })
   useHoveringPath(path, !!isHovering, DropThoughtZone.SubthoughtsDrop)
 
+  // Calculate the height for the uncle thought over cliff
+  const dynamicHeight = calculateCliffThoughtsHeight({ deepestDepth, depth })
```
raineorshine, line 38: Let's call this variable `dropTargetHeight`.

src/components/DropUncle.tsx
```diff
@@ -32,7 +47,7 @@ const DropUncle = ({ depth, path, simplePath }: { depth?: number; path: Path; si
       ref={dropTarget}
       style={{
         backgroundColor: testFlags.simulateDrop ? '#52305f' : undefined, // eggplant
-        height: '1.9em',
+        height: dragInProgress ? `${1.9 + dynamicHeight}em` : '1.9em',
```
raineorshine, line 50: Is there a reason that `dragInProgress` is checked? I think it may work without it. `drop-end` is position: absolute so the height will not affect the layout.

src/components/DropChild.tsx
```diff
@@ -70,6 +75,7 @@ const DropChild = ({ depth, last, path, simplePath }: DropChildProps) => {
           opacity: 0.9,
           // add some additional padding to empty subthought drop targets to avoid gaps in between sibling's subthought drop targets. This provides a smoother experience when dragging across many siblings. The user can still shift left to be clear of the empty subthought drop targets and drop on a child drop target.
           paddingBottom: '1em',
+          height: dragInProgress ? `${0.5 + dynamicHeight}em` : undefined,
```
raineorshine, line 78: Same here, I'm not sure if `dragInProgress` is needed?

src/components/DropChild.tsx
```diff
@@ -53,6 +54,10 @@ const DropChild = ({ depth, last, path, simplePath }: DropChildProps) => {
 
   const { isHovering, dropTarget } = useDragAndDropSubThought({ path, simplePath })
   useHoveringPath(path, !!isHovering, DropThoughtZone.SubthoughtsDrop)
+  const dragInProgress = useSelector(state => state.dragInProgress)
+
+  // Calculate the height for the child thought over cliff
+  const dynamicHeight = calculateCliffThoughtsHeight({ depth })
```
raineorshine, line 60: There is a false positive for DropChild drop targets that are not on a cliff.

## Steps to Reproduce

```
- a
- b
- c
```

1. Set `em.testFlags.simulateDrop = true`.
1. Set the cursor to `null`.
2. Drag `b`.

## Current Behavior

The height of the `a` drop target is too high.

Preview 2024-09-11 18 41 27

## Expected Behavior

The height of the `a` target should not be increased.

Let's add a snapshot test for this case.

image

raineorshine (changes requested):
We're getting close, thanks.

`DropUncle` is not taking into account the cliff height properly. This example makes it clearer:

## Steps to Reproduce

```
- a
  - b
    - c
      - d
        - e
        - f
- x
```

1. Run `em.testFlags.simulateDrop = true` in the console.
1. Set the cursor on `b`.
1. Drag `e`

## Current Behavior

The DropUncle drop target `x` is too short.

image

## Expected Behavior

The DropUncle drop target `x` should be `0.5em` taller than the first cliff drop target (`a`).

Let's add a snapshot test for this case with `simulateDrop` turned on to cover this.

src/components/DropChild.tsx
raineorshine: `c` is the last visible thought, so we should increase the height to take advantage of the extra space available.

src/components/LayoutTree.tsx
```diff
@@ -707,6 +706,9 @@ const LayoutTree = () => {
                   // Do this as padding instead of y, otherwise there will be a gap between drop targets.
                   style={cliff < 0 ? cliffPaddingStyle : undefined}
                   crossContextualKey={key}
+                  cliff={cliff}
+                  index={index}
+                  treeThoughts={treeThoughtsPositioned}
```
raineorshine, line 711: VirtualThought should definitely not need to know about the entire list of `treeThoughtsPositioned`. That breaks the hierarchy and encapsulation we've established. You're going to have to find a way to accomplish this that respects the division of responsibility.

src/components/DropChild.tsx
```diff
@@ -54,6 +56,20 @@ const DropChild = ({ depth, last, path, simplePath }: DropChildProps) => {
...
+  const parentPath = rootedParentOf(
+    useSelector(state => state),
+    path,
+  )
+
+  // Get siblings of the current thought (or root thoughts if no parent)
+  const siblings = useSelector(state => getChildrenSorted(state, head(parentPath)))
+
+  // Check if this thought is the last in its parent's list of children (or in the root context)
+  const isLastChild = head(siblings).id === head(path)
```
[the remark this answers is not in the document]
Zubair286: Hmm, maybe I should've done something like this.

```
const isLastChild = useSelector(state => {
  const parentPath = rootedParentOf(state, path)
  const siblings = getChildrenSorted(state, head(parentPath))

  return head(siblings)?.id === head(path)
})
```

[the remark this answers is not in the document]
raineorshine: Yes, and even including `calculateCliffDropTargetHeight` since it is part of that same overall calculation.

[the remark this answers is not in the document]
Zubair286: Right, and return the `dropTargetHeight` from the usesSelector.

src/components/DropChild.tsx
```diff
@@ -54,6 +56,20 @@ const DropChild = ({ depth, last, path, simplePath }: DropChildProps) => {
...
+
+  // Calculate the height for the child thought over cliff
+  const dropTargetHeight = isLastChild ? calculateCliffDropTargetHeight({ depth }) : 0
```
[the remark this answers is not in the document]
Zubair286: So this would only work if we only have visible children of a single context I believe, unlike in this case where multiple context children are expanded.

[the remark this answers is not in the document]
raineorshine: Yes. I think checking last child of a given parent is the wrong approach. The LayoutTree should probably pass down if it is the last visible thought in the entire rendered tree.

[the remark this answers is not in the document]
Zubair286: I think this was done but it seems commented out and implementation removed? from the `DropEnd` in `LayoutTree`.

```
<DropEnd
  depth={pathEnd.length}
  path={pathEnd}
  cliff={cliff}
  // last={!nextChildId}
/>
```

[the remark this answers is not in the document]
Zubair286: Or I think if it can be calculated in `VirtualThought`, where we are already checking if the thought `isVisible`, we can pass the nextThought to `VirtualThought` and check if its either hidden or null and if current `isVisible` to determine the last rendered thought.

src/components/LayoutTree.tsx
Zubair286: Then I think we can simply take it to `LayoutTree` and pass it from there like this, 

```
const prevThought = treeThoughtsPositioned[index - 1]
const prevCliff = prevThought ? prevThought.cliff : undefined
```

src/components/DropChild.tsx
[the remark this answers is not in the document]
raineorshine: In general we always want functions or components to be given as little information as possible for them to do their job. Doing this reduces the coupling between the function and the rest of the app. Components are on a "need-to-know" basis.

If the logic for `isLastVisible` were all derivable from the Redux state, then at least it would be encapsulable within a selector. However `nextThought` is a product of linearizing and positioning that happens within LayoutTree. `VirtualThought` should not have to know about this linearizing process. The only thing it cares about is whether it is the last visible child in the render tree, and that information is properly determined by the LayoutTree.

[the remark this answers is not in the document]
Zubair286: I see, then it has to be then calculated within the `useMemo` render phase and returned from `treeThoughtPositioned` but of course this logic cannot be completely used there.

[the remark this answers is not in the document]
Zubair286: is the `node.visibleChildrenKeys` not available? it seems missing from the `treeThoughtPositioned` items most of the time.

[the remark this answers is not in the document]
raineorshine: It's only used in table view, [as documented](https://github.com/cybersemics/em/blob/57c332047a3517ff832f037544f9c6438c13f364/src/components/LayoutTree.tsx#L65-L67). I would not recommend using it for other purposes.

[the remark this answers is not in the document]
Zubair286: I think similar logic can be used in `useMemo` map function the only problem I see is the use of `store.getState()` for `calculateAutoFocus`. If it does not do any harm, this could work fine.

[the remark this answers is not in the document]
Zubair286: > It's only used in table view, [as documented](https://github.com/cybersemics/em/blob/57c332047a3517ff832f037544f9c6438c13f364/src/components/LayoutTree.tsx#L65-L67). I would not recommend using it for other purposes.

No I was just asking, it did not seem useful at first glance, but I realize it is not going to be helpful.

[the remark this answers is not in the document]
raineorshine: > I think similar logic can be used in `useMemo` map function the only problem I see is the use of `store.getState()` for `calculateAutoFocus`. If it does not do any harm, this could work fine.

Calling `store.getState()` directly in components is not recommended. We generally need to rely on `useSelector` to ensure the value gets recalculated when the state changes.

`linearizeTree` has access to the state, so you could call `calculateAutofocus` there, though you have to be careful to only call it when needed as it is a heavy function.

Another possibility is to move `calculateAutofocus` up a level to `LayoutTree` instead of `VirtualThought`. It can be calculated in the linearize step. Then you would be able to generate `isLastVisible` there.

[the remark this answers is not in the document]
Zubair286: I was saying the same but in the `useMemo` function not in the `VirutualThought` btw. I did not do it `linearizeTree` but in `useMemo` and it is working fine, but I see that `linearizeTree` has access to `state` so it seems more logical to put it there instead of using `store.getState` in `useMemo`.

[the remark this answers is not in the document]
raineorshine: Yes, exactly

src/util/calculateCliffDropTargetHeight.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Calculates the height of a drop target at the cliff, decreasing with depth depth. This makes it easier to drop a thought at the desired level. Returns height in ems.
+ */
+const calculateCliffDropTargetHeight = ({ cliff = 0, depth }: { cliff?: number; depth?: number }) => {
+  if (depth === undefined) return 0
+
+  // Calculate dynamic height: increase as depth decreases
+  const baseHeight = 1
+  const heightIncrease = 0.5
+  const dropTargetHeight = baseHeight + heightIncrease * ((-cliff || depth) - depth)
```
trevinhofmann, line 10: This is just a minor suggestion, but I believe the functionality of this line would be clearer with a ternary instead of `(-cliff || depth) - depth)`:

```suggestion
  const dropTargetHeight = cliff === 0
    ? baseHeight
    : baseHeight + heightIncrease * (-cliff - depth)
```

trevinhofmann (approved):
Thanks, this is looking good to me! I added a small suggestion.

src/util/calculateCliffDropTargetHeight.ts
Zubair286: Thanks, I updated the code.

src/components/LayoutTree.tsx
```diff
@@ -343,7 +346,19 @@ const linearizeTree = (
     return [...accum, node, ...descendants]
   }, [])
 
-  return thoughts
+  // Generate isLastVisible property here because determinig the next thought is difficult in the recursive call. So we do it after tree of thought is completely generated.
+  const treeThoughtsWithIsLastVisible = thoughts.map((node, index, thought) => {
+    const nextNode = thought[index + 1]
+    const nextAutofocus = nextNode ? calculateAutofocus(state, nextNode.path) : null
```
raineorshine, line 352: This needs to be written with a bit more performance sensitivity, as `calculateAutofocus` is a heavy function.

If you're going to call `calculateAutofocus` on every node in `linearizeTree`, you must as well store the value and pass it down the component hierarchy to avoid repeating the same calculation further down.

See our previous discussion here: https://github.com/cybersemics/em/pull/2381#discussion_r1759746404.

src/components/LayoutTree.tsx
```diff
@@ -665,6 +680,12 @@ const LayoutTree = () => {
             // Perform this check here instead of in virtualThoughtsPositioned since it changes with the scroll position (though currently `sizes` will change as new thoughts are rendered, causing virtualThoughtsPositioned to re-render anyway).
             if (belowCursor && !isCursor && y > viewportBottom + height) return null
 
+            // Get the next and previous thoughts
```
raineorshine, line 683: This comment does not seem to match the code. I think it's just getting the previous thought, not the next and previous thoughts.

src/util/calculateCliffDropTargetHeight.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Calculates the height of a drop target at the cliff, decreasing with depth depth. This makes it easier to drop a thought at the desired level. Returns height in ems.
+ */
+const calculateCliffDropTargetHeight = ({ cliff = 0, depth }: { cliff?: number; depth?: number }) => {
+  if (depth === undefined) return 0
+
+  // Calculate dynamic height: increase as depth decreases
+  const baseHeight = 1
+  const heightIncrease = 0.5
+  // const dropTargetHeight = baseHeight + heightIncrease * ((-cliff || depth) - depth)
```
raineorshine, line 10: Remove commented out code.

raineorshine (changes requested):
1. ~~[This case](https://github.com/cybersemics/em/pull/2381#discussion_r1759414455) is not fixed yet.~~

2. ~~When any thought is dragged, there is a warning in the console:~~

   ~~`Selector unknown returned the root state when called. This can lead to unnecessary rerenders.
Selectors that return the entire state are almost certainly a mistake, as they will cause a rerender whenever *anything* in state changes.`~~

   Brave Browser 2024-09-18 16 20 04

3. You have a couple duplicate snapshots. Maybe you changed the name of a snapshot test and forgot to delete the old snapshot?

Zubair286:
1. [This case](https://github.com/cybersemics/em/pull/2381#discussion_r1759414455) seems to work fine for me.
2. I'm not sure but I do not get these warnings in the console.
3. I strangely had an ubuntu snapshot folder, which contained some duplicate snapshots.

[Screencast from 09-18-2024 11:00:52 PM.webm](https://github.com/user-attachments/assets/33784db1-07d4-4958-9f93-d95d0b8ab0ec)

src/components/LayoutTree.tsx
Zubair286: So I guess you are saying it should be calculated once, instead of mapping through the `treeThought` array, and then assigned to the `treeThought` array elements. Then we should pass it down the hierarchy?

raineorshine:
Sorry about that! I was on the wrong commit.

I can confirm that (1) and (2) are not an issue.

src/components/LayoutTree.tsx
Zubair286: Yes, I forgot to update it, I was getting the `next` as well there when I wrote that comment before removing the `next`.

Zubair286:
Got it, probably changed the name of snapshot test after generating snapshots first, will remove it now,

src/components/LayoutTree.tsx
Zubair286: I had to map the `treeThought` because it wasn't going to work out in `reduce` method, and since we have to calculate it for every thought I needed to map over the `treeThought` array to calculate it.

raineorshine: Yes, that should work.

Zubair286:
@raineorshine can you please review and confirm the functionality of this PR? Thanks

trevinhofmann (approved):
Thank you @Zubair286 and @raineorshine! The changes in this pull request are (now) quite streamlined and well-designed.

There is a bit of prop drilling with `isLastVisible`:
* `LayoutTree` => `VirtualThought` => `DropChild`
* `LayoutTree` => `DropEnd`

And with `prevCliff`:
* `LayoutTree` => `VirtualThought` => `DropUncle`

However, I don't think that two levels of prop drilling is excessive, so I would be comfortable with merging this design/implementation.

Regarding performance, the only part that _could_ be concerning is the frequent calls to `calculateCliffDropTargetHeight` when each of these components render. However, the calculation in that function is O(1) and has extremely low overhead (just a few basic arithmetic calculations). With some rough performance testing, each call to `calculateCliffDropTargetHeight` is in the ballpark of 0.000005ms on a V8 engine. This is not at all concerning, even being calculated on every render.

src/components/LayoutTree.tsx
```diff
@@ -75,6 +78,7 @@ type TreeThoughtPositioned = TreeThought & {
   width?: number
   x: number
   y: number
+  isLastVisible?: boolean
```
raineorshine, line 81: `TreeThoughtPositioned` already inherits `isLastVisible` from `TreeThoughts`.

src/components/LayoutTree.tsx
```diff
@@ -567,12 +573,16 @@ const LayoutTree = () => {
         ycol1Ancestors.push({ y: yaccum + height, depth: node.depth })
       }
 
+      const nextAutofocus = next?.autofocus
+      const nextIsLastVisible = nextAutofocus === 'show' || nextAutofocus === 'dim'
```
raineorshine, line 577: This variable is called `nextIsLastVisible`, but it actually just checks `nextIsVisible`. There is no logic here related to it being the last thought or last visible thought.

src/components/LayoutTree.tsx
```diff
@@ -567,12 +573,16 @@ const LayoutTree = () => {
...
+
       return {
         ...node,
         cliff,
         height,
         singleLineHeightWithCliff,
         width: tableCol1Widths.get(head(parentOf(node.path))),
+        isLastVisible: !nextIsLastVisible,
```
raineorshine, line 585: Won't this incorrectly match any thought that precedes a visible thought? We only want to match the last visible thought.

raineorshine (changes requested):
> There is a bit of prop drilling with `isLastVisible`:
> 
> * `LayoutTree` => `VirtualThought` => `DropChild`
> * `LayoutTree` => `DropEnd`
> 
> And with `prevCliff`:
> 
> * `LayoutTree` => `VirtualThought` => `DropUncle`
> 
> However, I don't think that two levels of prop drilling is excessive, so I would be comfortable with merging this design/implementation.

Yes, I'm fine with that!

> Regarding performance, the only part that _could_ be concerning is the frequent calls to `calculateCliffDropTargetHeight` when each of these components render. However, the calculation in that function is O(1) and has extremely low overhead (just a few basic arithmetic calculations). With some rough performance testing, each call to `calculateCliffDropTargetHeight` is in the ballpark of 0.000005ms with my machine using the V8 JS engine. This is not at all concerning, even being calculated on every render.

The performance concern was with `calculateAutofocus`, as [mentioned here](https://github.com/cybersemics/em/pull/2381#discussion_r1765281957). However setting aside performance, it's still a design issue. `calculateAutofocus` was supposed to be moved higher up the hierarchy so it could be utilized for the `isLastVisible` calculation, but instead it was just added. The [old instance](https://github.com/cybersemics/em/blob/66945aea4cf7d9c82a163f39f9822f1ca73f89f9/src/components/VirtualThought.tsx#L99) was not removed. You just need to follow through on the refactor and pass the autofocus down to `VirtualThought` rather than recalculating it below. Then you get a cleaner design and better performance with no downsides (that I can see).

src/components/LayoutTree.tsx
Zubair286: The idea was that if the next thought is `undefined` or autofocus 'hide' then the current is the last visible thought here.

raineorshine: As written, this will evaluate to true for more than just the last visible thought. Here's one case (but there may be others):

```
- a
  - b
    - c
- x
```

Set the cursor on `c`.

`isLastVisible` evaluates to true for both `c` and `x`.

Zubair286: Right, I understand now. We can add a check for `node.autofocus !== 'hide'` but if there are more thoughts below the last visible their isLastVisible would still be set to `true`.

raineorshine: Only one thought can be visible and followed by a hidden thought, as **em** does not render non-contiguous visible thoughts. Once you have the desired logic set up, I would suggest testing it on a few different cases to ensure it works as expected.

Zubair286:
@raineorshine in DropHover we have some `calculateAutofocus` calls, should those also be replaced by the autofocus prop?

raineorshine:
I would probably hold off on that, as it would require adding `autofocus` as a prop to `Thought` which could change the render behavior. If we into performance issues with `DropHover` we can revisit it.

Zubair286:
@raineorshine can you review the latest changes? Thanks

src/components/LayoutTree.tsx
```diff
@@ -567,12 +580,22 @@ const LayoutTree = () => {
...
       }
 
+      // check if the previous thought is already the last visible thought
+      const prevIsVisible = isVisible(prev?.autofocus, node.autofocus)
+
+      // check if the next thought is visible or not
+      const nextIsLastVisible = isVisible(next?.autofocus, node.autofocus)
+
+      // if previous thought is last visible then set isLastVisible of every next thought to false
+      const isLastVisible = !prevIsVisible ? false : !nextIsLastVisible
```
raineorshine, line 590: I think this can be simplified to the following, or am I missing a case?

```ts
const isLastVisible =
  (node.autofocus === 'dim' || node.autofocus === 'show') &&
  !(next?.autofocus === 'dim' || next?.autofocus === 'show')
```

src/components/LayoutTree.tsx
```diff
@@ -707,6 +737,9 @@ const LayoutTree = () => {
                   // Do this as padding instead of y, otherwise there will be a gap between drop targets.
                   style={cliff < 0 ? cliffPaddingStyle : undefined}
                   crossContextualKey={key}
+                  prevCliff={prevCliff}
```
raineorshine, line 740: This should just be inlined as `treeThoughtsPositioned[index - 1]?.cliff`. It's short and readable and only used in one place, so it doesn't need to be extracted into a variable.

raineorshine (changes requested):
Looks good now, just a little cleanup.

src/components/LayoutTree.tsx
Zubair286: Yes, it works great. Thanks

raineorshine (approved):

PR #2386, opened 2024-09-09
Fixed validation for min and max font size in Settings modal

### LINKED ISSUE #2340: Able to (easily) set the font-size to tiny values, including 0
## Steps to Reproduce

1. Click into the Settings page from the footer
2. Select the font size input so the cursor is at the end of the `18` input
3. Press backspace/delete once, changing the font size to `1`
4. Press backspace/delete again, changing the font size to `0`.

## Current Behavior

While actively updating the font size value in the input, the actual font size of the entire app is changing. This makes for a very jarring experience, as the appearance of the input drastically changes while trying to update it.

Further, it is possible to set the font size all the way to `0` by simply trying to delete the existing value before typing in the desired font size. This makes the input and all other text completely disappear.

https://github.com/user-attachments/assets/faefb941-c836-423a-b430-61b1172fdba4

## Expected Behavior

~I believe it would make more sense for the font size of the app to only update after the user completes editing of the input (i.e. `onBlur` instead of `onChange`).~

Additionally, it would likely make sense to have a minimum font size to avoid the user setting it to a value so small that they can't properly navigate the application to change it.

MiguelHG2351:
Fixes #2340

### DIFF
--- a/src/components/modals/Settings.tsx
+++ b/src/components/modals/Settings.tsx
@@ -1,4 +1,4 @@
-import { FC, PropsWithChildren } from 'react'
+import { FC, PropsWithChildren, useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { fontSizeActionCreator } from '../../actions/fontSize'
 import { showModalActionCreator as showModal } from '../../actions/showModal'
@@ -47,9 +47,10 @@ const Setting: FC<
 /** A font size control. */
 const FontSize = () => {
   const dispatch = useDispatch()
-  const fontSize = useSelector(state => state.fontSize)
+  const fontSizeSelector = useSelector(state => state.fontSize)
+  const [fontSize, setFontSize] = useState<number>(fontSizeSelector)
   const label =
-    fontSize <= MIN_FONT_SIZE
+    fontSizeSelector <= MIN_FONT_SIZE
       ? 'minimum reached'
       : fontSize >= MAX_FONT_SIZE
         ? 'maximum reached'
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
Thanks for your first contribution!

The minimum font size is now being enforced, thank you. However, it should still update instantly rather than onBlur. If the user entered font size is less than the minimum it's okay to reset it to the minimum onBlur, but any valid changes to the font size should update onChange.

trevinhofmann (approved):
Thank you for fixing this!

While testing, I noticed some other minor details that could be further improved, such as:
* The user should be able to fully delete the input value and then type in a new font size. At the moment, deleting the current value leaves a `0` in the input which makes entering a new font size awkward.
* It might be better to clamp to the minimum/maximum value, e.g. setting the font size to 8 rather than ignoring the input when the user enters `4`

That said, this pull request is already a standalone improvement, and I believe it should be merged as-is.

raineorshine (approved):
Looks good!

> * The user should be able to fully delete the input value and then type in a new font size. At the moment, deleting the current value leaves a `0` in the input which makes entering a new font size awkward.

I agree, but we can do that in a separate PR. It will require some numerical validation if the input type is changed from number to text.

> * It might be better to clamp to the minimum/maximum value, e.g. setting the font size to 8 rather than ignoring the input when the user enters `4`

I don't want to mess with the value while the user is typing, but I do think it should reset on blur.

I've added this:

```ts
	// reset on blur if font size is out of bounds
	onBlur={() => {
	  if (fontSize < MIN_FONT_SIZE || fontSize > MAX_FONT_SIZE) {
	    setFontSize(fontSizeSelector)
	  }
	}}
```

Issue #2170, opened 2024-07-19
🐼 PandaCSS


raineorshine:
Convert em's CSS styles to [PandaCSS](https://panda-css.com/).

## Background

**em** uses a combination of inline styles (slow, limited) and one big App.css (untyped, non-local). In 2021 there was an effort to switch to [emotion](https://github.com/emotion-js/emotion) + [twin.macro](https://github.com/ben-rogerson/twin.macro) (#1424), but it was never completed. The technology has evolved since then, and compile-time CSS-in-TS is the new best-in-class. In the lead is PandaCSS, which combines the best of many of other libraries like Stitches and Vanilla Extract. (CSS Modules is also well-liked, though I am opting for a CSS-in-TS approach for type safety and co-location.)

## Tasks

- [x] Setup PandaCSS
- [x] Convert inline styles
- [x] Convert App.css
- [x] Convert global styles to [defineGlobalStyles](https://panda-css.com/docs/concepts/writing-styles#global-styles)
- [x] Remove classNames library
- [x] Convert key frames 
- [x] Convert CSSTransition class
- [x] Convert specificity selectors
- [x] Convert classNames used for DOM behavior to aria-label or data attributes
- [x] Convert theme to [design tokens](https://panda-css.com/docs/theming/tokens)
- [x] Convert z-index variables to design tokens
- [x] Convert animation durations to design tokens
- [x] Write snapshot tests for theme
- [x] Fix theme

## Requirements

Avoid some PandaCSS-specific syntax for future compatibility:

  - Avoid [shorthand properties](https://panda-css.com/docs/concepts/writing-styles#shorthand-properties) e.g. `p`, `bg`
  - Avoid [exclamation syntax](https://panda-css.com/docs/concepts/writing-styles#important-styles) (and avoid `!important` if at all possible)
  - Avoid [pseudo properties](https://panda-css.com/docs/concepts/writing-styles#using-pseudo-props)
  - Avoid [combining shorthand properties with longhand](https://panda-css.com/docs/concepts/writing-styles#property-conflicts), e.g. `padding` and `paddingTop` together
  - Avoid [deep merging](https://panda-css.com/docs/concepts/writing-styles#merging-styles) multiple styles
  - Avoid [JSX style props](https://panda-css.com/docs/concepts/style-props)

## Commit hygiene

[...]

Issue #2390, opened 2024-09-10
Dev server styles break when checking out a different branch


raineorshine:
## Steps to Reproduce

1. Close the dev server if it's running.
2. `git checkout main` and start the dev server.
3. Switch branches.

## Current Behavior

1. The Text Color icon in the toolbar is not styled correctly:
   <img width="21" alt="Brave Browser 2024-09-10 11 33 59" src="https://github.com/user-attachments/assets/a0198fd1-cb59-4011-91c3-8409dbc07b34">

2. The Sort icon in the toolbar is not positioned correctly. On further inspection, it has the css class `trf_translateY\(4px\)` in the HTML, but its styles do not exist in the styled-system that is loaded.

The only thing that fixes the style is to kill the dev server and restart it. These interventions **do not** fix the problem:

✗ Running `panda codegen`
✗ Refreshing the page.
✗ Clearing the browser cache and refreshing the page.
✗ Hitting 'r' in the cli of the running dev server to restart.
✗ Running vite with `--force`

## Expected Behavior

The styles should refresh automatically when `panda codegen` is run and the dev server is running.

<img width="27" alt="image" src="https://github.com/user-attachments/assets/93e5f952-a073-4c35-b313-892b5fd87781">

raineorshine:
Seems to be related to vite HMR functionality.

`src/index.css` has a variable `__vite__css` that gets automatically updated after `panda codegen`. However it is missing styles for some reason.

Issue #2371, opened 2024-09-05
Snapshot tests fail with missing PandaCSS styles after component changes


raineorshine:
When I modify a component with the puppeteer tests running, and then hit (r) to re-run, the tests correctly pick up on the change. However, when I undo the code change and hit (r) again, I see slight differences in the snapshot, exactly where certain PandaCSS styles would apply. 

Probably the same as #2390, though it's not clear why only some font sizes fail. 

(Note that this problem existed before puppeteer was dockerized in #2365.)

## Steps to Reproduce

1. Start the dev server. 
1. `yarn test:puppeteer render -t load`
2. Add `return null` to Toolbar.tsx.
3. Hit (r) in the tests to re-run. They will fail as expected.
4. Remove `return null` from the component.
5. Hit (r) in the tests to re-run.

## Current Behavior

The snapshot tests for Font Size 18 and 22 fail. 

![font-size-18-default-initial-load-1-diff](https://github.com/user-attachments/assets/443cc6a0-e636-49f3-b2d7-744705fa1aac)

## Expected Behavior

All tests should pass.

Notice that if you quit the puppeteer tests and run `yarn test:puppeteer render -t load` from scratch *without changing any code* the tests will now pass.

trevinhofmann:
I have not been able to reproduce this issue locally.

https://github.com/user-attachments/assets/85c98066-a2a3-4fac-a643-a6bb9e1d7b56

raineorshine:
Fixed in #2420

PR #2397, opened 2024-09-14
Pandacss element selectors

yangchristina:
#2170

### DIFF
--- a/src/components/Checkbox.tsx
+++ b/src/components/Checkbox.tsx
@@ -85,7 +85,7 @@ const Checkbox: FC<
           overflow: 'auto',
         })}
       >
-        {title && <div style={{ lineHeight: children ? 1.2 : 1.5 }}>{title}</div>}
+        {title && <div className={css({ lineHeight: children ? 1.2 : 1.5 })}>{title}</div>}
         {children && <p className={css({ marginTop: '0.5em', fontSize: 'md', color: 'dim' })}>{children}</p>}
       </div>
 
--- a/src/components/Checkbox.tsx
+++ b/src/components/Checkbox.tsx
@@ -1,4 +1,5 @@
 import React, { FC, PropsWithChildren } from 'react'
+import { css } from '../../styled-system/css'
 import fastClick from '../util/fastClick'
 
 /** A static checkbox component with em styling. */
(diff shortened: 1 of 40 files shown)


yangchristina:
@trevinhofmann Puppeteer tests passed here, but they aren't passing locally for me (including for main). Do you have any idea why?

src/components/Tutorial/index.tsx
```diff
@@ -168,7 +179,13 @@ const Tutorial: FC = () => {
             cursorHeadValue.toLowerCase() === TUTORIAL_CONTEXT[tutorialChoice].toLowerCase())) &&
         gesture() ? (
           <div className='tutorial-trace-gesture'>
-            <GestureDiagram path={gesture()!} size={160} strokeWidth={10} arrowSize={5} className='animate-pulse' />
+            <GestureDiagram
+              path={gesture()!}
```
trevinhofmann, line 183: This comment is not blocking, and likely out of scope for this pull request. However, would it be safer to assign a variable such as `const path = gesture()` so that we can call the function once* and not need to include the `!` assertion here (instead, `path={path}`)?

\* I see that we are using the `once` utility to memoize the result and avoid doing the calculation multiple times per render, but would that still serve a purpose if we simply computed the `path` variable once during the render? As it is now, it seems like the `once` utility and `!` assertion are somewhat codependent and could be simplified by straightforwardly assigning the `path` variable once.

trevinhofmann:
> @trevinhofmann Puppeteer tests passed here, but they aren't passing locally for me (including for main). Do you have any idea why?

Locally, I am currently seeing only a single timeout failure on both the `main` branch and your `pandacss-element-selectors` branch:

>  FAIL  |puppeteer-e2e| src/e2e/puppeteer/__tests__/ContextView.ts > load buffered ancestors of contexts when context view is activated

Are you seeing any other puppeteer failures locally, @yangchristina?

trevinhofmann (approved):
This pull request touches quite a few parts of the application and the styling changes appear relatively safe, so I won't spend too much time digging into every affected component. However, I have spent some time poking through the application to view the tutorial, tips, modals, etc., and everything does look and function as I would expect.

I added one suggestion and a follow-up question about the Puppeteer test, but these changes look ready to me. Thank you!

yangchristina:
@trevinhofmann yes I was getting a timeout failure at some point, there also tests failing from having the dev server running at the same time which according to Raine should be related to https://github.com/cybersemics/em/issues/2371, https://github.com/cybersemics/em/issues/2390

src/components/Tutorial/index.tsx
```diff
@@ -108,8 +109,20 @@ const Tutorial: FC = () => {
 
   const cursorHeadValue = useSelector(state => state.cursor && headValue(state, state.cursor))
 
+  const gesturePath = gesture()
```
trevinhofmann, line 112: Thank you!

trevinhofmann (approved):

src/components/ContextBreadcrumbs.tsx
```diff
@@ -123,7 +125,17 @@ const BreadCrumb = React.memo(
           ) : label === HOME_TOKEN ? (
             <HomeLink color='gray' size={16} className={css({ position: 'static' })} />
           ) : (
-            <Link className='extend-tap-small' simplePath={simplePath} label={label} />
+            <Link
+              cssRaw={css.raw(
+                {
+                  margin: '-0.25em',
+                  padding: '0.25em',
+                },
```
raineorshine, line 133: Since we already have the `extendTap` recipe, let's make a `small` variant for this case.

src/components/Checkbox.tsx
```diff
@@ -31,34 +60,33 @@ const Checkbox: FC<
...
-        }}
+        })}
       >
-        {title && (
-          <div className='checkbox-label' style={{ lineHeight: children ? 1.2 : 1.5 }}>
-            {title}
-          </div>
-        )}
-        {children && <p className='checkbox-description text-medium dim'>{children}</p>}
+        {title && <div style={{ lineHeight: children ? 1.2 : 1.5 }}>{title}</div>}
```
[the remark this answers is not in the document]
yangchristina: Yes, sorry I missed this.

raineorshine (approved):

PR #2405, opened 2024-09-17
Factor out usePositionFixed hook

### LINKED ISSUE #2393: Refactor positionFixed utility
Original issue for the toolbar: #1695
Newer issue for handling other position fixed components: #2387

## Current Behavior

Several components switch from position fixed to position absolute on mobile Safari due to a Safari limitation. This causes the component position to be choppy during scroll, but it is the only way to get the components to appear relative to the viewport while the keyboard is up.

The switching behavior has been partially abstracted out into the [positionFixed](https://github.com/cybersemics/em/blob/c217ebb4b95435e24d7f6953b0953f84df1d6dcf/src/components/Editable/positionFixed.ts) utility, but it's not very well designed.

## Expected Behavior

- Propose a better architecture that decouples the position switching logic from the components that use it.
- Refactor positionFixed so that the does not have explicit references to DOM elements that are position fixed. 
- Consider creating a custom hook that components can use to register as position fixed, with the position absolute fallback built in.

raineorshine:
Closes #2393

Waiting for #2397 which was submitted first. Merge conflicts expected.


raineorshine:
The toolbar right arrow shifted right for some reason. The position was already slightly off, and it will probably need to be changed after #2408 anyway, so I just updated the snapshot for now.

PR #2407, opened 2024-09-21
Pandacss anchor buttons

yangchristina:
#2170 
This is the last of the element selectors in App.css
- remove unused `ul ul.context-chain` from App.css
- removed className='tutorial-text', which doesn't have any styles

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -23,10 +23,6 @@
   z-index: var(--z-index-scroll-zone);
 }
 
-ul ul.context-chain {
-  margin-left: 1.6em;
-}
-
 .clear {
   clear: both;
 }
--- a/src/App.css
+++ b/src/App.css
@@ -53,11 +53,6 @@
     text-align: center;
   }
 }
-.upper-right {
-  position: absolute;
-  top: 0;
-  right: 0;
-}
 
 .text-small {
   font-size: 80%;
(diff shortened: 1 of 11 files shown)


yangchristina:
I failed a puppeteer test earlier: https://github.com/cybersemics/em/actions/runs/10983149534/job/30492188860?pr=2407

I make some small changes that wouldn't make any difference really, and the tests are passing now. I also couldn't find any connection between the failing snapshot and the code in my last commit, which seems odd.

src/components/AppComponent.tsx
```diff
@@ -260,7 +189,6 @@ const AppComponent: FC = () => {
 
   return (
     <div className={componentClassNames}>
-      <GlobalStyles styles={globalStyles} />
```
trevinhofmann, line 263: Awesome! 🎉

src/App.css
```diff
@@ -23,10 +23,6 @@
   z-index: var(--z-index-scroll-zone);
 }
 
-ul ul.context-chain {
-  margin-left: 1.6em;
-}
```
trevinhofmann, line 28: While we're removing the unused `.context-chain`, now would be a good time to clean up the other references to it by removing these comments? https://github.com/cybersemics/em/blob/25cbb95a8ac01a42540d93a459eaef5e9d4f1498/src/App.css#L685-L713

trevinhofmann (changes requested):
Aside from the one suggestion about `.context-chain` cleanup, this pull request looks good to me. Thanks, @yangchristina!

trevinhofmann (approved):

raineorshine (approved):

Issue #1792, opened 2024-01-30
Table row height is incorrect


raineorshine:
## Steps to Reproduce

```
- “Unity of apperception”
  - =view
    - Table
  - Object
    - For some reason I was afraid to look this up
  - Assumption
    - 
  - Negative prediction
    - ?
  - Rationalizations
  - Outcome
```

## Current Behavior

Height of "Assumption" and "Rationalizations" rows are too great.

![image](https://github.com/cybersemics/em/assets/750276/9b7ba1ce-5ece-42bb-b3e7-284c0a041c56)

## Expected Behavior

Rows should be sized to fit the content.

PR #2411, opened 2024-09-22
Fix order of calculating accumulated thought heights.

kevinstubbs:
Fixes https://github.com/cybersemics/em/issues/1792 The issue was caused by double-counting the col1 heights, because we were first adding the height to the accumulation and then recording the y position of the column as the yaccum (which we just increased) and then adding our height to it again.

Before
image

After
image


trevinhofmann (changes requested):
Thank you @kevinstubbs! The adjusted ordering in height calculations seems to make sense, but I am still seeing the issue during the initial render. The layout seems correct after selecting a thought to trigger the recalculation, but there is still an excessive height for "Assumption" during the first render:

https://github.com/user-attachments/assets/3744a744-b754-471a-bb18-4481aa60d156

Additionally, there is a small merge conflict which needs to be resolved.

kevinstubbs:
Hi @trevinhofmann, good catch. Looks like the issue you found is one that's unrelated to the yaccum bug. This other issue is caused by the height of the thought being incorrectly calculated, I haven't found the root cause but basically there is something which is changing the height to what we actually see as the end result, but it happens after the setSize has been called from the layout effect in VirtualThought. That's why it fixes itself when you click on something (updateSize is called in VirtualThought when selectCursor changes).

I'm still investigating, but these are my findings so far.

raineorshine:
This is a tricky one! Just wanted to check in and see if we can help at all.

PR #2412, opened 2024-09-22
Fix inconsistent width of table col1 

### LINKED ISSUE #1795: Table: Col1 width is different on load
## Steps to Reproduce

```
- A
  - =view
    - Table
  - ✓
    - Boston, Massachusetts
  - ✗
    - Cincinnati, Ohio
- B
```

1. Paste the above thoughts.
2. Set the cursor to `A`.
3. Refresh the page.
4. Set the cursor to null.
5. Pause one second.
6. Set the cursor back to `A`.

## Current Behavior

The width of col1 is different between steps 3 (after refresh) and 5 (after moving the cursor away and back).

https://github.com/cybersemics/em/assets/750276/d147993e-067d-4adb-9886-e29e7d282c8c

## Expected Behavior

The width of col1 should be consistent even if the cursor has moved away and back.

kevinstubbs:
fixes #1795 

This was caused by ignoring column width updates (as the thoughts tree populates with information that the parent is a table and therefore should use the true width rather than 3em min-width).

If this could caused a performance issue, then perhaps we can think about removing the min-width of 3em, or not using it until we're sure that the view attribute of the parent node in the path has been populated.

Demo following repro steps from https://github.com/cybersemics/em/issues/1792

[video](https://github.com/user-attachments/assets/d432be03-589d-4990-9220-8bfc58a20736)

### DIFF
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -145,7 +145,7 @@ const useSizeTracking = () => {
         sizeRemovalTimeouts.current.delete(key)
 
         setSizes(sizesOld =>
-          height === sizesOld[key]?.height && isVisible === sizesOld[key]?.isVisible
+          height === sizesOld[key]?.height && width === sizesOld[key]?.width && isVisible === sizesOld[key]?.isVisible
             ? sizesOld
             : {
                 ...sizesOld,


trevinhofmann (approved):
Thank you, @kevinstubbs! I was able to reproduce the issue on the `main` branch, and the issue is not present after this change. The solution you present seems like a self-explanatory and necessary change, ensuring that we update the width when the calculated value has changed from `sizesOld`.

There are no immediately obvious performance concerns to me. My first consideration might be if there are situations with deeply nested thoughts with some parent/child dependencies on the width (e.g. a new width for a child triggers a new width calculation for the parents, triggering a new width calculation for the grandparent, etc.). I'm less familiar with that possibility, so I'll give @raineorshine a chance to comment on any possible performance concerns.

raineorshine (approved):

raineorshine:
Looks good to me! This was likely an oversight when width was added to the size calculation. I think it was originally just height.

The DOM measuring is one of the slower parts of the front-end. However, it used to be based on ResizeObservers on every thought which was extremely slow. The initial estimated height/width followed by the DOM measurement should be acceptable performance with the expected number of thoughts on screen.

PR #2419, opened 2024-09-24
fix #2398: strip slows down expandThoughts

### LINKED ISSUE #2398: strip slows down expandThoughts
> I found that a significant part of scripting and HTML parsing is done at the start of the interactions due to the `expandThoughtsRecursive` function using the `strip` function to perform sanitization on the values. In my tests, this accounts for at least 15-20% of the interaction time, with 1000 thoughts processed. This could be fixed by moving the sanitization to another stage that's not after every interaction. It is to be noted that if window-like virtualization is implemented, the impact of optimizing this would be reduced as the number of thoughts processed per interaction would be greatly reduced. WDYT?
> 
> Brave Browser 2024-09-15 19 23 20

It looks like `strip` was added to handle bold/italic/strikethrough. Thoughts that end in `:` should be pinned open, including `<b>:</b>`. See #1073. Solved in #1090 but note the unit tests appear to be false positives now.

We can remove `strip` and replace it with a much faster alternative that just strips closing tags.

carloslfu:
This PR replaces the use of `strip`  inside `expandThoughts` with a regex that strips closing tags and matches them. After testing it, I noticed I could feel the difference while using the app, and the data shows this approach is much more performant.

I profiled creating a new thought with 1000 thoughts already created. I did it by scrolling down to the final thoughts and then hitting Enter.

This change reduces the `newThought` execution times from 110-130ms to 3-5ms.

Before:
Screenshot 2024-09-24 at 12 35 08 PM

After:
Screenshot 2024-09-24 at 12 26 44 PM

### DIFF
--- a/src/selectors/expandThoughts.ts
+++ b/src/selectors/expandThoughts.ts
@@ -20,12 +20,13 @@ import isURL from '../util/isURL'
 import keyValueBy from '../util/keyValueBy'
 import parentOf from '../util/parentOf'
 import publishMode from '../util/publishMode'
-import strip from '../util/strip'
 import unroot from '../util/unroot'
 import childIdsToThoughts from './childIdsToThoughts'
 import { anyChild, getAllChildrenAsThoughts } from './getChildren'
 import getContexts from './getContexts'
 import pinned from './isPinned'
+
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Amazing! Love these big, early wins 😁

carloslfu:
Closes #2398.

PR #2427, opened 2024-09-26
Pandacss common recipes

yangchristina:


yangchristina:
@raineorshine I found a visual regression I introduced in 23fe15a, I am really sorry about this
I passed in `…css`, which was a function instead of the prop I meant to pass in.
It's only a few pixels in difference, so it shouldn't really be noticeable
Should have been: 
left: '-2px', top: '-3px' 
Instead was:
top: '-1px',
left: '1px',

How should I deal with this? I committed the fix to this PR (5c7ba7b)

raineorshine:
Good catch! I don't have a specific recommendation for how to fix it, as long as it is documented. Normally I would fix it in a separate commit. Which component did it affect exactly?

yangchristina:
Here's the line number of the original mistake: https://github.com/cybersemics/em/commit/23fe15a494144048108ba7c14ff5c1be37641981#diff-d16fd5a3a438fecd6fa3a04a27c4fb06cb6f9b5abaa3efdfc092be6160560923R25
(line 25)

This is the affected section: https://github.com/cybersemics/em/commit/10aad6e3dc4afc28862acb06453fd64577d4a6be#diff-a8d22c2c9737dbc3c4be4c3dbadba4b5c798d9fd12884736fcce4f66b5af0054

This is the commit of the fix, which is done in a separate commit: https://github.com/cybersemics/em/pull/2427/commits/d52bc4c2bd051b65d891ce388f9931fe848b4a29

trevinhofmann (changes requested):
Thank you, @yangchristina! This is looking great to me. I tested a bit on desktop and mobile, and didn't notice any regressions. I just have one comment/request about completing a migration from `style` to `className`/`cx`/`css`.

src/components/DropUncle.tsx
```diff
@@ -40,33 +41,33 @@ const DropUncle = ({
...
-        'drop-end': true,
-      })}
+      className={cx(
+        dropEnd(),
+        css({
+          backgroundColor: testFlags.simulateDrop ? '#52305f' : undefined, // eggplant
+          opacity: 0.9,
+        }),
+      )}
+      style={{ height: `${1.9 + dropTargetHeight}em` }}
```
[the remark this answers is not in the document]
yangchristina: No, because `dropTargetHeight` is a dynamic variable, and className is only for styles statically analyzable at build-time.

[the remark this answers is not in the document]
trevinhofmann: Aha, this makes perfect sense. Thank you for clarifying!

trevinhofmann (approved):

src/components/Bullet.tsx
```diff
@@ -413,6 +414,7 @@ const Bullet = ({
     const isHolding = state.draggedSimplePath && head(state.draggedSimplePath) === head(simplePath)
     return isHolding || isDragging
   })
+  const value = useSelector(state => getThoughtById(state, thoughtId)?.value)
```
raineorshine, line 417: It's better to move `isDivider` into the selector here. Then the component will only be re-rendered when the boolean value changes, not every time the thought value changes.

src/components/DropHover.tsx
```diff
@@ -88,19 +89,28 @@ const DropHover = ({ simplePath }: { simplePath: SimplePath }) => {
     const autofocusParent = calculateAutofocus(state, parent)
     return autofocus === 'dim' && autofocusParent === 'hide'
   })
+  const thoughtId = head(simplePath)
+  const value = useSelector(state => getThoughtById(state, thoughtId)?.value)
```
raineorshine, line 93: Same here

src/components/StaticThought.tsx
```diff
@@ -180,7 +182,7 @@ const StaticThought = ({
           // left, top are eyeballed for different font sizes
           <HomeIcon style={{ position: 'relative', left: fontSize - 14, top: fontSize / 4 - 1 }} />
         ) : isDivider(value) ? (
-          <Divider path={simplePathLive} />
+          <Divider cssRaw={css.raw({ borderColor: 'rgb(204, 204, 204)' })} path={simplePathLive} />
```
raineorshine, line 185: Can we move the border color into the Divider component itself? Every Divider should have this border color applied from what I can tell, so there is no reason to parameterize it.

src/components/Divider.tsx
```diff
@@ -39,22 +40,26 @@ const Divider = ({ path }: { path: Path }) => {
...
     >
       <div
         aria-label={'editable-' + head(path)}
-        className={classNames({
-          divider: true,
-          // requires editable-hash className to be selected by the cursor navigation via editableNode
-        })}
+        className={css(
+          {
+            // divider: true,
```
raineorshine, line 57: Remove commented out code

src/components/Note.tsx
```diff
@@ -123,6 +123,7 @@ const Note = React.memo(({ path }: { path: Path }) => {
       <ContentEditable
         html={note || ''}
         innerRef={noteRef}
+        aria-label='note-editable'
         className={'note-editable'}
```
raineorshine, line 127: Can we remove the className now that it uses `aria-label`?

src/e2e/puppeteer/helpers/getEditingText.ts
```diff
@@ -5,7 +5,7 @@ import { Page } from 'puppeteer'
  */
 const getEditingText = (page: Page) =>
   page.evaluate(() => {
-    return document.querySelector('.editing .editable')?.innerHTML
+    return document.querySelector('[data-editing=true] .editable')?.innerHTML
```
raineorshine, line 8: `.editable` should probably be pulled dynamically from the `editable` recipe rather than hardcoded here.

raineorshine (changes requested):

src/e2e/puppeteer/helpers/getEditingText.ts
raineorshine: This was updated in https://github.com/cybersemics/em/blob/51c8fa94f41ef99c466a9e659ce216a4b38913fd/src/e2e/iOS/helpers/getEditingText.ts but not https://github.com/cybersemics/em/blob/51c8fa94f41ef99c466a9e659ce216a4b38913fd/src/e2e/puppeteer/helpers/getEditingText.ts.

Like in the iOS helper, I'm curious if a PandaCSS will even work inside the puppeteer browser page environment. It's worth a try.

src/e2e/iOS/helpers/getEditingText.ts
```diff
@@ -1,9 +1,10 @@
 import { Browser } from 'webdriverio'
+import { editable } from '../../../../styled-system/recipes'
 
 /** Get the thought value that cursor on. */
 const getEditingText = (browser: Browser<'async'>): Promise<string | undefined> => {
   return browser.execute(() => {
-    return document.querySelector('.editing .editable')?.innerHTML
+    return document.querySelector(`[data-editing=true] ${editable()}`)?.innerHTML
```
raineorshine, line 7: I think this should be `[data-editing=true] .${editable()}` (added: `.`)

I'm curious if this will work inside the webdriver's browser context.

raineorshine (changes requested):
Looks good! Just `getEditingText` that needs to be fixed.

yangchristina:
@raineorshine do you have any idea how to deal with this? `waitForEditable` is now timing out, I tried to see if it would work if I increased the timeout, but it was still timing out. I did log the output of `editable()` to make sure that it is returning the string `'editable'`

raineorshine:
Think of the code inside `page.evaluate` as a completely different JS environment. It is possible to pass serializable values from the test environment to the puppeteer browser environment through the second argument. Take a look at the puppeteer helpers to see how that technique is used. I can take a stab at it if you can't figure it out.

yangchristina:
@raineorshine Thanks! that worked.

raineorshine (changes requested):
I think I missed the bigger picture in my review yesterday 😅...

Quick review of how we use different HTML attributes:

- `data-testid` for testing
- `aria-label` or `data-*` for runtime behavior
- `className` for css

So using `.${editable()}` for runtime behavior is not really correct. Recipes should just be used for styling. 

Since `aria-label` is already used, we could add a `data-editable` boolean attribute. Or we could move the current area-label to `data-thoughtid` and set `aria-label='editable'`. Let me know what you think.

I know that this doesn't exactly fit the current milestone, but if we're going to the trouble of injecting `editable()` into query selectors, we should probably take a step back and do it the right way. It makes sense to me to do it now, but let me know if you prefer moving it to a different block of work.

yangchristina:
I can add a data-editable attribute. I kept with className, since `editable()` is used in two places, so it would mean that every time you used `editable`, you'd have to add `data-editable` as well. However I'm not sure if any of the .editable selectors actually mean to select the .editable in Search.tsx. If the selectors are all only for Editable.tsx, then yes I do think data-editable would make more sense.

raineorshine:
I'm okay with the redundancy, as avoiding using classNames for behavior is a higher priority.

Search.tsx is broken and needs to be fixed anyway, so we can ignore it for now. Thanks!

raineorshine (approved):

PR #2428, opened 2024-09-28
Update eslint rules to prefer default export

### LINKED ISSUE #2346: eslint: Enforce default export
- Add `import/prefer-default-export` to eslint rules
- Remove `import/prefer-default-export` ts override
- Convert single named exports to default exports (one commit per conversion)
- Inline functions if they are only used in one file

```ts
    // Enforce default export
    // https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/prefer-default-export.md
    'import/prefer-default-export': [
      2,
      {
        // any: Any exporting file must contain a default export.
        // single: When there is only a single export from a module, prefer using default export over named export.
        target: 'any',
      },
    ],
```

Examples:

9922909732 Remove device/router.
37af57d9b6 emojiRegex: Convert to default export.
63c7bee4e1 useSelectors: Move into Footer module.
14718399d8 Remove old migrations.
d3f83f15f6 metaSettings and hashKeys: Disable import/prefer-default-export.
0d7f2c5fb7 isUndoEnabled: Convert to default export.
8e66ec3675 serviceWorkerRegistration: Disable import/prefer-default-export.
b8db740c52 shortcuts: Disable import/prefer-default-export.
b53efdd8e8 shortcuts/index: Disable import/prefer-default-export.
299ffceaa0 Rename gestureEnum and convert to default export.
0dcfd95f68 constants: Disable import/prefer-default-export.
ba0a947ecf findAllThoughtsByText: Convert to default export.
6a59651eb5 Remove createMockStore.
154a143e3a createTestStore: Convert to default export.
8c3210b3e1 Remove unused mock-debounce-throttle.
39db7e8097 findSubthoughts: Convert to default export.
0a74b99908 getClosestByLabel: Convert to default export.
4b2b2ab397 Remove unused queryAllThoughtsByText.
d8890cd959 queryThoughtByText: Convert to default export.
649ce8259a validateRoam test: No need to export testData.
790300ec35 compareThought: disable import/prefer-default-export
ed5c5bc9f6 isMarkdown: Convert to default export.
ede06f22b5 markdownToText: Convert to default export.
2a2791c7f1 Refactor memoizeResolvers.
62399263b9 text-block-parser: disable import/prefer-default-export
6f652cefa4 react-dnd: disable import/prefer-default-export

mstrofbass:
Adds the prefer-default-export eslint rule and fixes the resulting errors.

Closes #2346


mstrofbass:
@raineorshine Please review when you have a chance. I added a few overrides where it made sense *to me* so feel free to let me know where you disagree. The primary ones where I'm unsure about are:

importFiles.ts b53efdd8e861e7e3f5a9ed90f35456400df05600
positionFixed.ts a4da92c019359ae11ba098f7be4b9551d16de544
disableScroll.ts 003b1b4c20c1200738eff35179d42d0f6b08948e
serviceWorkerRegistration.ts 769193cb059dfb42725735dc5e0d2074903ec1ee

...but you should take a look at the others regardless.

src/actions/importFiles.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/prefer-default-export */
```
raineorshine, line 1: This is fine. `deleteResumableFIle` is a somewhat hacky solution and should probably be refactored separately.

raineorshine (changes requested):
Thank you, looks good.

I realize I made an oversight when I wrote the requirements, which I take full responsibility for. I had forgotten that reducers and action-creators are co-located in `/actions` now. By convention, the reducer is the default export and the action-creator is a named export. This has to be handled as an exception. Otherwise the default export of some action files is a reducer and in others is an action-creator.

Because this is my fault, please record any time it takes you to correct this and I will pay you at your hourly rate. Sorry for the trouble!

src/device/disableScroll.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/prefer-default-export */
```
raineorshine, line 1: Another way to do this would be to have a single default export that takes a boolean argument, e.g. `allowScrolling(false)`.

src/components/Editable/positionFixed.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable import/prefer-default-export */
```
raineorshine, line 1: This is fine, as this functionality is getting a major refactor in #2405.

src/device/disableScroll.ts
mstrofbass: I went ahead and refactored this since it took minimal effort.

mstrofbass:
@raineorshine changes to the action creators were rolled back. I did not see any reducers in the files so all the changes were just adding the named export back and that's it. Didn't take enough time to justify billing separately.

Let me know if you spot anything else.

raineorshine (approved):

PR #2430, opened 2024-09-28
Remove blur from snapshot tests and lower failure threshold

raineorshine:
Remove blur from snapshot tests and lower failure threshold now that they are in browserless and consistent across platforms.


raineorshine:
Consistently passed after 6 runs in the CI and on my local machine, so going to merge.

PR #2432, opened 2024-09-29
pass className and cssRaw into Icons

yangchristina:


raineorshine:
Thanks!

The Welcome modal is the first thing the user sees when starting the app for the first time, then it is dismissed permanently by flipping a flag in localStorage. You can test the Welcome modal by clearing all site data.

yangchristina:
@raineorshine Oh I see thanks! I've confirmed that there is indeed no HomeIcon, 
Screenshot 2024-09-29 at 11 37 12 PM

so in this part `let width = logoEls[0] && logoEls[0].width`, `width` is actually `undefined`, which when properly typed makes line 43 problematic.

Screenshot 2024-09-29 at 11 39 34 PM

raineorshine:
Must be some old functionality that didn't get removed properly!

yangchristina:
@raineorshine I also found another regression I introduced previously which I've fixed in the previous commit ddea822, sorry I missed this
I removed .modal-content, but missed where it was used as a querySelector
Realistically this wouldn't have impacted anyone, since it only affects when you make your screen basically as small in height as it can go.
The commit that introduced this was eecd97f, line 104 in ModalComponent.tsx

yangchristina:
@raineorshine I can quickly remove the logo stuff if you'd like? It's just removing lines 48, 43, and 30 in the image above, as well as the logo selector

raineorshine:
I'd appreciate that!

src/shortcuts/splitSentences.tsx
```diff
@@ -16,8 +16,16 @@ import parentOf from '../util/parentOf'
 import splitSentence from '../util/splitSentence'
 
 // eslint-disable-next-line jsdoc/require-jsdoc, react-refresh/only-export-components
-const Icon = ({ fill = 'black', size = 20, style }: IconType) => (
-  <svg version='1.1' width={size} height={size} fill={fill} style={{ marginTop: -2, ...style }} viewBox='0 0 110 115'>
+const Icon = ({ fill = 'black', size = 20, style, className, cssRaw }: IconType) => (
```
trevinhofmann, line 19: Same question about `cssRaw` not being used.

trevinhofmann (changes requested):

trevinhofmann (approved):
Thank you very much, @yangchristina! These changes all look good to me, and I haven't noticed any regressions in the icons when testing on desktop and mobile (Safari).

raineorshine (commented):
> The value of `cssRaw` is not used within this component. Should we be using something like `className={cx(css(cssRaw), className)}` similar to the other components?

I'm also curious about this. It looks like all of the icons accept both `cssRaw` and `className`, but it's not clear to me why both are needed, or when a developer should use one over the other. The ToolbarButton component only uses `className` and `style`.

yangchristina:
@raineorshine cssRaw lets you override existing styles properly, but it cannot contain recipes. So classNames is really for recipes.

raineorshine:
> cssRaw lets you override existing styles properly, but it cannot contain recipes. So className is really for recipes.

That makes sense. What about className? Can you just pass `css(...)` to className? I'm guessing it affects how styles are overridden, but maybe you can clarify that for me.

Having every component accept `cssRaw` and `className` is certainly the most versatile, but we lose specificity. In practice, there are probably only a few variants of custom styles that are passed to a given component. A better interface would be to define a prop for the variant and allow the component itself to choose the correct styles. Or, if we need to pass a runtime value to a component, pass the raw value and only convert it into css inside the component. These are my general thoughts, but we need to think about how this might apply to Icons.

> I don't recall any other icon using a recipe other than `icon()`, so I can probably remove className if you'd like? I remember at least one icon that already had existing styling on it, which was why I added cssRaw

Let's drill down on this a little more. Which icons currently use `className`? Which icons currently use `cssRaw`? Once we review the usage, we can consider a narrower implementation.

Note that ToolbarButton passes `style` to every icon via https://github.com/cybersemics/em/blob/main/src/%40types/Icon.ts#L8. It would be nice to convert that to PandaCSS, but it does contain some runtime values.

yangchristina:
`TextColor.tsx` already has existing styling to it, so passing in className has the potential to cause conflicts
```tsx
className={cx(
          css(
            {
              borderRadius: 5,
              display: 'inline-block',
              marginLeft: 2,
              marginRight: 2,
              textAlign: 'center',
            },
            cssRaw,
          ),
          className,
)}
```
I'd like to pass the statically analyzable styles in, in `ToolbarButton.tsx`
```css
<SVG
        size={fontSize}
        cssRaw={css.raw({
          position: 'relative' as const,
          cursor: isButtonExecutable ? 'pointer' : 'default',
          opacity: dropToRemove ? 0 : 1,
          transition: 'opacity 200ms ease-out',
        })}
        style={style}
      />
```
There are no conflicts here, since the styles are different, but if you used `classNames`, whenever you changed the styling for either component, you'd have to check to make sure there are not conflicts.

I haven't converted `splitSentences.tsx` yet, but it will have existing styles in className too (`marginTop: -2`) 

I'm not sure if there's other cases I'm forgetting, but using `className` instead of `cssRaw` may make future styling less flexible, since anytime you adjust the styling for an icon, you'll have to confirm that it doesn't conflict with what is passed in in `ToolbarButton`, as well as any other className prop passed into that icon.

Removing `className` might be a better idea, since it just removes the ability to use recipes for any icon, but no Icons part of <ToolbarButton> at least are using recipes (only HomeIcon, and that is not part of shortcuts). Using recipes with these icons wouldn't be a good idea anyway, since most already have the `icon()` recipe, and using multiple recipes also has the potential to cause conflicts.

raineorshine:
Thank you for the explanation! I can see that changing recipe or component styles when using `className(cx(...))` can result in conflicts if we're not careful.

I might be missing something obvious, but can you explain why `cssRaw` does not suffer the same problem? It seems like if I set the same css property in both places it would also be a conflict and the type system would not catch that for me.

Anyway, let's remove `className` from the icons as you suggested and just rely on `cssRaw`. That simplifies the design.

yangchristina:
`cx` is just a string concatenator, it doesn't actually merge anything, while `css` will [deeply merge](https://panda-css.com/docs/concepts/writing-styles#merging-styles) its arguments. If you want it to merge with `cx`, you can possibly consider using something like this: https://github.com/brendon1555/panda-cx-deduplicator/

Sounds good! I'll remove classNames

raineorshine:
`css` will deeply merge its arguments, but it still does so *silently*. That is, there is no type error/warning if a conflicting css property is added. I think the main difference is that the order of arguments to `css` provides deterministic priority to styles, whereas multiple `classNames` will ultimately rely on browser css precedence (though I think layers makes this a bit more predictable, as recipes are applied in a separate layer...?).

raineorshine (approved):

PR #2438, opened 2024-09-30
1873: Initial implementation of narrow col1 wrap logic.

### LINKED ISSUE #1873: Table: Col1 narrow + wrap
## Steps to Reproduce

```
- X
  - =view
    - Table
  - Currently Reading
    - # The Ecological Approach to Visual Perception
  - To Read
    - # Explaining Understanding
  - Read
    - # Jamgon Kongtrul’s Retreat Manual
```

## Current Behavior

Col2 wraps before Col1. 

## Expected Behavior

Col1 should wrap after "Currently".

Col1 should wrap before Col2 when the average thought length is smaller.

kevinstubbs:
Resolves #1873 by implementing the formula `col1width = avg1/(avg1+avg2) * totalWidth` with 7em as the minimum col1 width and 20em as the maximum.

Before
image

After
image

An example where col 2 widths are very small.
- X
  - =view
    - Table
  - Currently Reading
    - a
  - To Read
    - e
image


raineorshine:
Thanks! I'll let @trevinhofmann do a first pass. 

Please update the puppeteer snapshots by running `yarn test:puppeteer -u`. Check any snapshots that were updated and make sure they look okay. If all looks good, you can push the new snapshots in a new commit to the PR. Then the CI will pass and the new snapshots will show up in the diff for our review.

trevinhofmann (changes requested):
Thank you, @kevinstubbs!

Here are some results from my testing:

When the right column is quite narrow, it is now wrapping the first column unnecessarily. If both columns can fit without wrapping, I would expect them to:

https://github.com/user-attachments/assets/b3474fe0-4e4e-4248-b956-c6a3989b5ca9

When both columns are relatively wide, some of the rows begin to overlap:

https://github.com/user-attachments/assets/1f727fb8-0e73-4ac9-94ff-c63da75be93d

src/components/LayoutTree.tsx
```diff
@@ -518,10 +527,50 @@ const LayoutTree = () => {
...
+          // at least the 1st column's node.
+          let numCol2Entities = 0
+          let tableCol2Width = 0
+
+          node.childrenValues?.forEach(x => {
+            col1Sum += x.col1Value?.length || 0
+
+            if (x.col2Value?.length) {
+              col2Sum += x.col2Value?.length || 0
+              numCol2Entities++
```
trevinhofmann, line 553: The calculation of the width of column 2 does not seem to account for any subthoughts in the column except for the first one of each row. I believe we should look at the length of the longest subthought in the column's row rather than only the first of each row.

In the video below, you can see that the length of the "Baz" subthought affects the layout, but the "Foo" subthought does not:

https://github.com/user-attachments/assets/ad735afa-52a8-4640-be32-93cc31e0c5e4

raineorshine:
Thanks for the initial review @trevinhofmann. Looks like we may have to tweak the algorithm a bit to find a good balance. 

> When both columns are relatively wide, some of the rows begin to overlap

This might be a separate bug, just made apparent through the new table calculation. Can it be reproduced in `main`? I don't want to add increase the scope of this issue unless we know it was definitely caused by the particular implementation that was used here.

raineorshine (changes requested):
After reviewing the snapshots, the main issue I'm seeing is the way that the table layout has changed for tables with very narrow, single-letter values in col1. This is a common use case for tables in **em**. In this case, col1 should be as narrow as possible. The implementation in `main` currently handles this well: https://github.com/kevinstubbs/em/blob/main/src/e2e/puppeteer/__tests__/__image_snapshots__/table-view/table-view-col-1-narrow-no-wrap-width-1.png

The solution for #1873 should reduce the width of col1 when its values are shorter than col2 without increasing the width of col1 when it contains very short, single-letter values.

PR #2440, opened 2024-10-05
Refactor puppeteer helpers to use global browser page

### LINKED ISSUE #2363: Refactor puppeteer helpers to use global browser page
The puppeteer helpers were designed to avoid the use of global variables. To access the browser page, they used a convoluted process of currying the page object on setup. 

https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/helpers/index.ts

In practice, this offers little benefit. It adds complexity to understanding and using helpers, as well as friction adding new helpers. 

While the principle of decoupling tests from puppeteer specifics is good, the helpers themselves should be able to access the global browser object as needed. This will result in a simpler and easier-to-understand design. 

- Refactor all the puppeteer helpers to use the global browser object.
- Remove the helper index and import helpers directly in tests.

mstrofbass:
The changes include:

1. Moving the `beforeEach` and `afterEach` functions from the `index.ts` file to a `setup.ts` file
2. Updating the `vitest` puppeteer workspace to add the new `setup.ts` file as a `setupFile`
3. Updating the `setup` function to set `global.page` to the newly generated page created prior to each test.
4. Added an `emulate` helper because we lose the ability to pass the device info to the `setup` function
5. Updated each of the helpers to refer to the `global.page` reference instead of accepting the `page` as a parameter.
6. Updated each of the tests to import the helpers directly
7. Removed references to the `index.ts` file
8. Deleted the `index.ts` file

Problems:

1. As I noted in #2363, this implementation works because the individual test files are isolated from each other and the tests within a particular file run sequentially. Any implementation changes that allow the tests within a file to run concurrently will break.
2. I could not figure out a way to make the `global` module definition more ...global... and this particular problem is beyond my Typescript skills. Open to suggestions on how to improve this.
3. This was difficult to break into small commits since I wasn't sure if it would work and a lot of the stuff is intertwined. Sorry if that makes it a bit difficult to review.

Closes #2363


mstrofbass:
@raineorshine this is ready for review but I am expecting you to request some changes so just let me know.

raineorshine (commented):
> 1. As I noted in [Refactor puppeteer helpers to use global browser page #2363](https://github.com/cybersemics/em/issues/2363), this implementation works because the individual test files are isolated from each other and the tests within a particular file run sequentially. Any implementation changes that allow the tests within a file to run concurrently will break.

Thanks, noted. I [commented](https://github.com/cybersemics/em/issues/2363#issuecomment-2395378362) on the issue and would be interested in your opinion. I will get second opinions from the rest of the team as well.

> 2. I could not figure out a way to make the `global` module definition more ...global... and this particular problem is beyond my Typescript skills. Open to suggestions on how to improve this.

Yes, I think reproducing the type extension in every file is too much. If we decided to stick with the global approach, we would need to figure this out. Trevin or Fabian could possibly help.

> 3. This was difficult to break into small commits since I wasn't sure if it would work and a lot of the stuff is intertwined. Sorry if that makes it a bit difficult to review.

No problem!

mstrofbass:
@raineorshine since `page` is loaded when `setup` it has to be initialized as `null`. In order to keep the use clean, I created a `fetchPage` function that will throw an error if `page` hasn't been initialized so we don't have to check for null everywhere else. This is mostly a typescript issue inasmuch as we should never have a situation where a helper is trying to access the page before it's been initialized.

mstrofbass:
Can we get this merged and then do any refining under new tasks? Happy to include any refinements under the same milestone, it's just blocking some of the other big tasks and it'll be a lot easier to get the big ones taken care of and come back and do clean up afterwards.

raineorshine (approved):
> Can we get this merged and then do any refining under new tasks? Happy to include any refinements under the same milestone, it's just blocking some of the other big tasks and it'll be a lot easier to get the big ones taken care of and come back and do clean up afterwards.

Sure, I can do that. Normally we need to handle testing, regressions, and refactors in the original PR in order to keep everything in one place and avoid merging incomplete code into `main`. However given that this is only in need of refactoring and it is blocking other tasks I will make an exception this time.

> @raineorshine since `page` is loaded when `setup` it has to be initialized as `null`. In order to keep the use clean, I created a `fetchPage` function that will throw an error if `page` hasn't been initialized so we don't have to check for null everywhere else. This is mostly a typescript issue inasmuch as we should never have a situation where a helper is trying to access the page before it's been initialized.

This looks pretty good. I have an idea for how to eliminate this extra step which I will articulate in a new issue.

raineorshine:
I did some experimenting to see if the `page` object could be exported directly from the setup rather than having the extra step of `fetchPage`. Once the behavior was confirmed it didn't take much finish the refactor. See #2451.

Issue #2053, opened 2024-06-23
URL icon is not vertically aligned with text baseline


raineorshine:
## Steps to Reproduce

Paste a url. 

## Current Behavior

The url icon is too high.

![image](https://github.com/cybersemics/em/assets/750276/c1764bf1-a936-4968-85f3-3f2fa3277c06)

## Expected Behavior

The url icon should vertically align with the baseline of the text. 

The solution should work across font sizes.

## Testing

Snapshot test

PR #2441, opened 2024-10-07
2053: Improve visual vertical alignment of URL icon

kevinstubbs:
Improved vertical alignment of the URL icon, resolving #2053 The difference is minor, but can be seen in the updated snapshot images.

Vertical-align: text-top is supported by all major browsers: https://caniuse.com/mdn-css_properties_vertical-align_text-top


raineorshine (approved):
Looks good to me!

And the existing url snapshot tests seem to cover this pretty well, so I don't think we need a new test.

Issue #1942, opened 2024-05-24
Command Library Search


raineorshine:
## Steps to Reproduce

1. Click the (?) icon in the footer 
2. Click Command Library. 

## Current Behavior

There is no way to search the commands. 

## Expected Behavior

Add a search bar to the top of the Command Library. As the user types, the list of commands should be filtered down to only matches. Commands should match on label or description. 

This should use the same logic as the Command Palette (Command + P). You might need to do some refactoring so they can share code.

PR #2448, opened 2024-10-08
Command Library Search

MiguelHG2351:
feature https://github.com/cybersemics/em/issues/1942


raineorshine (approved):
Thanks for your contribution! This looks great. I like how you factored out `useShortcut` so that it can be shared across components.

There was one regression related to mobile functionality. On mobile, the filtered list should only include shortcuts with `shortcut.gesture` defined.

I have one recommendation regarding the design of the hook API. In the current implementation, `keyboardInProgress`/`setKeyboardInProgress` and `recentCommands`/`setRecentCommands` are all passed to the hook. The hook is responsible for storing the state, but not responsible for managing it, which is a bit confusing.

Conceptually, it's better to think of `keyboardInProgress` and `recentCommands` as inputs to `useShortcut` that affect the filtering. The state management is trivial, so there's not really any complexity that needs to be hidden away from the components. Therefore, we can move the state management up and get a hook that has a simpler API and clearer distinction between inputs and outputs:

```ts
const shortcuts = useShortcut(keyboardInProgress)
```

And with some options:

```ts
const shortcuts = useShortcut(keyboardInProgress, {
  recentCommands,
  sortActiveCommandsFirst: true,
})
```

PR #2449, opened 2024-10-09
Pandacss: convert thought related components

yangchristina:
8af4755 fixes e31045c line 19 (extra comma inside style)
- This doesn't seem like a regression, since when running main, the border still shows up. However I have confirmed that adding the comma is causing the puppeteer tests to fail, but it looks like the expected test is wrong (doesn't have the border)

The left is missing a border, but a border should exist
font-size-18-default-initial-load-1-diff
@trevinhofmann I think this may be a problem with the test, I'm not too familiar with how these snapshot tests work, so I was wondering if you could give me some advice on how to deal with this?

25ce846 fixes 65ac386 removed line 107 of `components/Bullet.tsx` (fix selector in puppeteer test) (regression in `load a child after a parent is expanded` test, causing it to always pass)


yangchristina:
Interesting, I didn't change anything, just merged with main, and all the tests are passing now.

trevinhofmann:
> @trevinhofmann I think this may be a problem with the test, I'm not too familiar with how these snapshot tests work, so I was wondering if you could give me some advice on how to deal with this?

> Interesting, I didn't change anything, just merged with main, and all the tests are passing now.

From the diff, the failure seemed related to icons. It probably isn't worth diving into exactly why that was, but it's a good rule of thumb to keep the PR relatively up-to-date with `main`, and I'm glad that this worked out this time. Thank you!

src/components/QuickDropIcon.tsx
```diff
@@ -97,7 +98,7 @@ const QuickDropIcon = ({
           size={fontSize * 1.5}
           fill={isHovering ? (dark ? 'lightblue' : 'royalblue') : dark ? 'white' : 'black'}
           // disable default .icon transition so that highlight is immediate
-          style={{ cursor: 'move', transition: 'none', verticalAlign: 'middle' }}
+          cssRaw={css.raw({ cursor: 'move', transition: 'none', verticalAlign: 'middle' })}
```
trevinhofmann, line 101: Could you please explain why `className={css({...})}` is used in most cases, but some such as this one use `cssRaw={css.raw({ ... })}`?

Pardon my ignorance on this one. I found this bit of documentation, but I'm not sure why `css.raw` applies here: https://panda-css.com/docs/guides/dynamic-styling#alternative

src/components/Thought.tsx
```diff
@@ -413,17 +413,14 @@ const ThoughtContainer = ({
...
+        data-testid={'thought-' + hashPath(path)}
         className={css({
           /* Use line-height to vertically center the text and bullet. We cannot use padding since it messes up the selection. This needs to be overwritten on multiline elements. See ".child .editable" below. */
           /* must match value used in Editable useMultiline */
-          lineHeight: '1.72',
-        })}
-        data-testid={'thought-' + hashPath(path)}
-        style={{
-          // ensure that ThoughtAnnotation is positioned correctly
+          lineHeight: '1.72', // ensure that ThoughtAnnotation is positioned correctly
```
trevinhofmann, line 420:
```suggestion
          lineHeight: '1.72',
          // ensure that ThoughtAnnotation is positioned correctly
```

I believe this comment applies to the next line (`position: 'relative',`). It's also not a very useful comment, but that's not in the scope of this change.

trevinhofmann (changes requested):
Thank you very much, @yangchristina! I sincerely apologize for the extended delay on this review.

The changes look great overall, and everything seems to work as expected when testing a bit locally. I just have one very minor suggestion about moving a comment, and two questions.

src/components/QuickDropIcon.tsx
yangchristina: This is because `css.raw({ ... })` can be passed into className like so: `className={css(cssRaw)}`. This is sometimes better than passing in className directly, since `className={cx(css({...}), className)}` doesn't merge in a reliable way (ex. if you have conflicting styles, you don't know whether the conflicting style in `className` or `css({...})` will apply, but if you do `className={cx(css({...}, cssRaw))}` instead, you know that the styles in `cssRaw` will override the `{...}` part.

yangchristina: Here are some docs on this: https://panda-css.com/docs/concepts/merging-styles#merging-css-objects

src/components/DropDownMenu.tsx
```diff
@@ -17,19 +16,31 @@ interface DropDownMenuProps {
...
+          flexDirection: 'column',
+          justifyContent: 'space-evenly',
+          position: 'absolute',
+          right: '3px',
+          padding: '0px 5px',
+          zIndex: 'stack',
+          backgroundColor: 'bg',
+          border: `1px solid {colors.fg}`,
+        })}
+        style={style}
```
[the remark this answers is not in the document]
yangchristina: Yes I can do this!

yangchristina:
Thanks @trevinhofmann! No worries.

raineorshine (approved):

PR #2453, opened 2024-10-13
Disable yjs connecting status in tests2 #2111

### LINKED ISSUE #2111: Disable YJS connecting status in tests
When there are no thoughts, the UI is kept blank or says "Connecting..." until the YJS websockets connects or time out after 5 sec. This slows down the tests considerably. 

Disable this loading screen when in test mode and go straight to the thoughts.

Remove the [sleep](https://github.com/cybersemics/em/blob/5ffafd3a22beb062ff2b55b87115fe898aea7870/src/e2e/puppeteer/helpers/setup.ts#L68-L70) in test setup and make sure the tests pass.

mstrofbass:
Updates the OfflineStatusStore to use a timeout of `0` before going into offline mode, thus hiding the loading screen immediately when we're in the e2e tests. The corresponding sleep time is removed from the e2e tests setup function, shaving a significant amount of time off of their runtime.

Closes #2111

### DIFF
--- a/src/e2e/puppeteer/setup.ts
+++ b/src/e2e/puppeteer/setup.ts
@@ -1,7 +1,7 @@
 /* eslint-disable import/prefer-default-export */
 import chalk from 'chalk'
 import { Browser, ConsoleMessage, Device, Page } from 'puppeteer'
-import sleep from '../../../util/sleep'
+import sleep from '../../util/sleep'
 
 // eslint-disable-next-line @typescript-eslint/no-namespace
 declare module global {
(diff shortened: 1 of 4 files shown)


mstrofbass:
I'm not sure how safe the use of `navigator.webdriver` is here since there may be legitimate reasons that we want to use automation and have the site behave as normal. Maybe we can set a variable on the window that we can check?

raineorshine (changes requested):
Thanks, looks good! You can make `isE2E` a constant variable rather than a function, since it is always determined synchronously on startup and never changes. 

Since we're only using 'navigator.webdriver` for e2e tests now, and it's trivial to search for all instances in the codebase in the future, then I think it's not necessary to create another of abstraction at this time.

raineorshine (approved):
Going to merge this now since it offers such a dramatic improvement to the test performance, and will help the rest of the team immediately!

Issue #2316, opened 2024-08-21
Add Light mode toggle to Settings


raineorshine:
Add a toggle to the Settings modal for Light Mode.

```js
dispatch(setting({ key: 'Theme', value: 'Light' }))
```

raineorshine:
Hi, thanks for your interest. You are welcome to work on any issue that is not assigned to anyone, but it's a good idea to check in with me first to confirm. Just fork and clone the repo, run `yarn`, and you'll be up and running :). Let me know if you have any questions!

PR #2459, opened 2024-10-15
add theme toggle at setting page

RED-ROSE515:
I added theme toggle(switch) to setting page.

Issue : https://github.com/cybersemics/em/issues/2316 (Add Light mode toggle to Settings)


src/components/ThemeSwitch.tsx
```diff
@@ -0,0 +1,41 @@
+import { FC, useEffect, useState } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
+import { settingsActionCreator as setting } from '../actions/settings'
+import theme from '../selectors/theme'
+
+/** A ThemeSwitch Component with a title and description. */
+const ThemeSwitch: FC = () => {
+  const dispatch = useDispatch()
+  const currentTheme = useSelector(state => theme(state))
+  const [active, setActive] = useState(currentTheme === 'Light')
```
raineorshine, line 10: It would be better to use Redux as the source of truth rather than creating duplicate state within the component. 

This will eliminate the need for the `useEffect`.

raineorshine (changes requested):
Thank you! Works great.

The sun and moon toggle is cool, but we need something that is more on brand for this app. For now, I would suggest using the existing https://github.com/cybersemics/em/blob/main/src/components/Checkbox.tsx. Eventually I will have our designer do a proper mockup.

Something I should have told you about head of time is that the project uses [PandaCSS](https://panda-css.com/) for styles. App.css is deprecated and is currently being converted to PandaCSS, so unfortunately I can't accept new additions to that file. Take a look at how the [css](https://github.com/cybersemics/em/blob/f3a239b55394cf3a0f669e568834deb25f866932/src/components/Checkbox.tsx#L19) function is used in other components, and read the PandaCSS documentation if you need more advanced features.

I will add a comment to App.css that makes this clearer for other developers.

raineorshine:
That's fine, thanks.

You'll need to update the Settings modal snapshot as well.

```
yarn test:puppeteer -u modal
```

raineorshine:
Sorry, didn't mean to push to the branch :)

raineorshine (approved):
Great, thanks :)

raineorshine (approved):

raineorshine:
You can use https://discord.gg/qGzrH9Bt for general questions and ephemeral discussion, but please keep specific issue-related discussion on GitHub. That way we have a clear record associated with the issue.

PR #2461, opened 2024-10-17
implement lettercase feature

### LINKED ISSUE #2424: Command: Lettercase
Add a new command that provides a drop down menu to set the letter case of the cursor thought:

- lowercase
- UPPERCASE
- Sentence case (only the first letter capitalized)
- [Title Case](https://github.com/blakeembrey/change-case/tree/main/packages/title-case)

The dropdown menu should match the behavior of the [ColorPicker](https://github.com/cybersemics/em/blob/main/src/components/ColorPicker.tsx) component, but with the four options displayed in a vertical list. Factor out common functionality to avoid code duplication.

RED-ROSE515:
Closes #2424

I implemented lettercase features( LowerCase, UpperCase, SentenceCase, TitleCase) to the ToolBar.

These are the screenshots.
image
image
image
image


raineorshine (changes requested):
Thank you! This is a great stat. Thanks for adding a unit test as well.

Please always reference the associated issue number in the PR description. That way GitHub will link the PR to the issue. I've added it above.

It looks like the Linter, Tests, and Puppeteer tests are all failing in the CI. Once these are all passing I can do a proper review.

I will note that `useWindowOverlay` has been duplicated in the ColorPicker and LetterCasePicker component. As mentioned in the issue, and following the DRY principle, we want to avoid code duplication. Code duplication doubles the amount of code to maintain and leads to subtle discrepancies over time. Instead, factor out the common code and use the appropriate level of abstraction to share common functionality.

src/actions/formatLetterCase.ts
```diff
@@ -0,0 +1,50 @@
...
+        case 'LowerCase':
+          return value.toLowerCase()
+        case 'UpperCase':
+          return value.toUpperCase()
+        case 'SentenceCase': {
+          const sentenceCaseRegex = /(^\w|\.\s*\w)/gi
+          return value.toLowerCase().replace(sentenceCaseRegex, match => match.toUpperCase())
+        }
+        case 'TitleCase':
+          return titleCase(value.toLowerCase())
```
raineorshine, line 33: Why convert it to lowercase first before converting it to TitleCase?

src/actions/formatLetterCase.ts
```diff
@@ -0,0 +1,50 @@
...
+import { titleCase } from 'title-case'
+import Thunk from '../@types/Thunk'
+import * as selection from '../device/selection'
+import pathToThought from '../selectors/pathToThought'
+import simplifyPath from '../selectors/simplifyPath'
+import { editThoughtActionCreator as editThought } from './editThought'
+
+/** Format the browser selection or cursor thought based on the specified letter case change. */
+export const formatLetterCaseActionCreator =
+  (command: string): Thunk =>
```
raineorshine, line 11: You can narrow the type of `command` to just the strings that are acceptable.

src/actions/formatLetterCase.ts
```diff
@@ -0,0 +1,50 @@
...
+    const state = getState()
+    const cursor = state.cursor
+
+    if (!cursor) return
+
+    const thought = pathToThought(state, cursor)
+    const originalThoughtValue = thought.value
+    const savedSelection = selection.save()
+    /** Util function to apply the appropriate transformation based on the command. */
+    const getUpdatedThoughtValue = (command: string, value: string): string => {
```
raineorshine, line 22: Let's call this function `applyLetterCase`.

src/shortcuts/lettercase.ts
```diff
@@ -0,0 +1,27 @@
...
+const lettercase: Shortcut = {
+  id: 'lettercase',
+  label: 'Lettercase',
+  description: 'Change the Letter case.',
+  svg: Icon,
+  canExecute: getState => isDocumentEditable() && !!getState().cursor,
+  multicursor: {
+    enabled: false,
+    error: () => 'Cannot change text color with multiple thoughts.',
+  },
```
raineorshine, line 15: This shortcut should have multicursor support. Start with `multicursor: true` and see if that works. If not, we can discuss other multicursor strategies.

src/shortcuts/lettercase.ts
```diff
@@ -0,0 +1,27 @@
...
+  exec: (dispatch, getState) => {
+    const state = getState()
+    dispatch({
+      type: 'toggleLettercase',
+      path: state.cursor,
```
raineorshine, line 20: I think `path` is leftover from another shortcut and can be removed.

src/components/__tests__/LetterCasePicker.ts
```diff
@@ -0,0 +1,59 @@
+// import { importTextActionCreator as importText } from '../../actions/importText'
```
raineorshine, line 1: Remove commented out code.

src/components/LetterCasePicker.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const dispatch = useDispatch()
+  const overflow = useWindowOverflow(ref)
+
+  /** Toggles the Letter Case to the clicked swatch. */
+  const toggleLetterCase = (command: string, e: React.MouseEvent | React.TouchEvent) => {
+    e.stopPropagation()
+    e.preventDefault()
+    dispatch(formatLetterCase(command))
+  }
+  const selected = useSelector(state => {
```
raineorshine, line 30: This selector uses a lot of the same logic as `getUpdatedThoughtValue` in the action-creator. This could lead to discrepancies in the future. Instead, move `getUpdatedThoughtValue` into a separate util function that both modules can use.

src/components/LetterCasePicker.tsx
```diff
@@ -0,0 +1,94 @@
...
+    if (value === titleCase(value.toLowerCase())) return 'TitleCase'
+    return ''
+  })
+  const casingTypes = ['LowerCase', 'UpperCase', 'SentenceCase', 'TitleCase']
+
+  return (
+    <div style={{ userSelect: 'none' }}>
+      <div
+        ref={ref}
+        style={{
```
raineorshine, line 45: This project uses [PandaCSS](https://panda-css.com/) for styling. The `style` attribute should only be used for styles with dynamic runtime values. Otherwise you should use Panda's `css` function. See other components for usage, or check out https://panda-css.com/docs/concepts/writing-styles.

src/actions/formatLetterCase.ts
```diff
@@ -0,0 +1,50 @@
...
+        default:
+          return value
+      }
+    }
+
+    const updatedThoughtValue = getUpdatedThoughtValue(command, originalThoughtValue)
+    const simplePath = simplifyPath(state, cursor)
+
+    dispatch(
+      editThought({
```
raineorshine, line 43: Let's add `force: true` to this call. I'm seeing some instances where the thought value is updated but the ContentEditable is not. `force: true` is a bit heavy-handed, but it should fix it for now.

raineorshine (changes requested):
Thank you! Looks very good. I've offered some tips below to clean up the code a bit.

There are just a couple behavioral issues I am seeing:

1. Let's prevent both the ColorPicker dropdown and the LetterCase dropdown from being open at the same time. If one is clicked, the other one should close.
2. The browser selection (i.e. text cursor) is not being preserved when it starts at the end of the thought. I'm guessing that `selection.restore` does not work in this case because `savedSelection.node` is not the same after the thought re-renders.

raineorshine:
When you rebase on `main`, note the signature for `isDropdownOpen` has changed: https://github.com/cybersemics/em/commit/9b340a258c6bd4152df6c4253e1de285ebf97b88

src/shortcuts/letterCase.ts
```diff
@@ -0,0 +1,23 @@
...
+const letterCase: Shortcut = {
+  id: 'letterCase',
+  label: 'LetterCase',
+  description: 'Change the Letter case.',
+  svg: Icon,
+  canExecute: state => isDocumentEditable() && !!state.cursor,
+  multicursor: {
+    enabled: false,
+    error: () => 'Cannot change text color with multiple thoughts.',
+  },
```
raineorshine, line 15: You'll want to set this to `multicursor: ignore` and then implement each option as a separate shortcut to enable multicursor support.

I considered the possibility of implementing multicursor support inline to the reducer, but that would miss out on the existing cursor handling that happens automatically with `execShortcutWithMulticursor`. Plus, we need each letter case option to be a separate shortcut anyway so that they show up in the Command Palette. Separate shortcuts is a bit verbose, but it provides the necessary functionality.

raineorshine (changes requested):
Hi, thanks for the update. The changes that you made look good.

> 1. Let's prevent both the ColorPicker dropdown and the LetterCase dropdown from being open at the same time. If one is clicked, the other one should close.

The LetterCasePicker is closed when the ColorPicker is selected, but not the other way around. You also need to close the ColorPicker when the LetterCasePicker is opened.

> 2\. The browser selection (i.e. text cursor) is not being preserved when it starts at the end of the thought. I'm guessing that `selection.restore` does not work in this case because `savedSelection.node` is not the same after the thought re-renders.

This is still not working for me. When the caret is at the end of the thought and I apply a letter case, the caret incorrectly moves to the beginning of the thought.

> This project uses [PandaCSS](https://panda-css.com/) for styling. The `style` attribute should only be used for styles with dynamic runtime values. Otherwise you should use Panda's `css` function. See other components for usage, or check out [panda-css.com/docs/concepts/writing-styles](https://panda-css.com/docs/concepts/writing-styles).

Do you need any help with this? The style attributes need to be replaced with calls to PandaCSS's `css` function.

raineorshine:
> In my opinion, there is no need to save the caret offset.

From the developer's perspective, preserving the caret offset can feel like extra work, and maybe it's not that important so it could be ignored. However from the user's perspective, having the caret move to the beginning of the thought is the extra work. The default to them is that the caret stays in the same place. To move the caret to the beginning of the thought, we would need evidence that this provides additional affordance to the user.

Maybe the user is typing a thought, decides halfway through that they need to make the thought title case, then they resume writing. In this case, the caret should definitely be where they left it after they activate title case. It is less likely that the user's intention is to edit the beginning of a thought when their caret was at the end.

> I tried to implement PandaCSS on my side, it doesn't work.

No problem, let's troubleshoot. Where did it stop working? What was the actual and expected behavior?

raineorshine (changes requested):
> > 2. The browser selection (i.e. text cursor) is not being preserved when it starts at the end of the thought. I'm guessing that `selection.restore` does not work in this case because `savedSelection.node` is not the same after the thought re-renders.
> 
> This is still not working for me. When the caret is at the end of the thought and I apply a letter case, the caret incorrectly moves to the beginning of the thought.

Better. Try activating a few different letter case options in a row. The caret is preserved the first time, but then moves on subsequent activations. 

> You'll want to set this to `multicursor: ignore` and then implement each option as a separate shortcut to enable multicursor support.
> 
> I considered the possibility of implementing multicursor support inline to the reducer, but that would miss out on the existing cursor handling that happens automatically with `execShortcutWithMulticursor`. Plus, we need each letter case option to be a separate shortcut anyway so that they show up in the Command Palette. Separate shortcuts are a bit verbose, but it provides the necessary functionality.
> 
> Let me know if anything about the multicursor functionality needs clarification.

The letter case options still need to be split into separate shortcuts and given multicursor support, but for simplicity we can do that in a separate PR. The current functionality is adequate for an initial implementation.

raineorshine (approved):
Looks good!

Note that the selection offset only refers to the character offset when focusNode.nodeType is a text node (`3`). When focusNode.nodeType is an element type (`1`), then offset actually changes meaning and refers to the index of the node. This caused the caret to incorrectly move to offset 1 in the following case:

1. Create a new thought.
2. Type `Hello` and leave the caret at the end of the thought.
3. Activate lowercase from the letter case dropdown.
4. Activate UPPERCASE from the letter case dropdown.

Luckily `selection.offsetThought` normalizes the editable offset. I added a fix in 588aec4e5c.

raineorshine:
> > The letter case options still need to be split into separate shortcuts and given multicursor support, but for simplicity we can do that in a separate PR. The current functionality is adequate for an initial implementation.
> 
> I got it. I will split letter case options into separate shortcuts.

That would be great. This PR or separate PR? Up to you.

Issue #2408, opened 2024-09-22
Lottie icons


raineorshine:
Update the toolbar icons to the 59 new Lottie icons and animate on ~click~ execute.

## Assets

- [lottie-icons.zip](https://github.com/user-attachments/files/17088824/lottie-icons.zip)
  - Includes SVG and Lottie JSON files
  - `icons1` contains 27 icons that are shown in the toolbar by default
  - `icons2` contains 32 icons that can be added to the toolbar via Settings/Customize Toolbar.

![presentation](https://github.com/user-attachments/assets/5d807e12-c470-40f2-84ca-9f6a4b531587)
![presentation](https://github.com/user-attachments/assets/39e3a30b-25fc-4804-b534-21927cd4514e)

### Milestone 1

- Replace existing toolbar svg’s with new static svg’s
- Pixel perfect positioning

### Milestone 2

- Animate on ~click~ execute (except active → inactive)
    - Complete animation regardless of mousedown time, i.e. even if the mouse is released before the animation duration
    - Smooth appearance when clicked again before the first animation has completed
- Preserve drag-and-drop behavior in customize mode
    - Do not animate on drag

### Milestone 3

- Scale with state.fontSize
- Preserve extended click area
- Desktop and mobile
- Dark and Light theme
- Preserve highlighting and dimming when icon is active and inactive

### Milestone 4

- Special handling of ternary button: Sort
    - The sort button has three states: Ascending, Descending, and Manual
    - The icon needs to animate smoothly from one state to the next on click
        - Manual → Ascending
        - Ascending → Descending
        - Descending → Manual

## Notes

- The only new functionality is replacing the existing "press" animation with the new Lottie animation. The rest of the requirements are about preserving the existing behavior. Play around with the existing app to get familiar with the existing toolbar behavior.
- Each command has an [svg](https://github.com/cybersemics/em/blob/66945aea4cf7d9c82a163f39f9822f1ca73f89f9/src/%40types/Shortcut.ts#L72-L73).
- The toolbar has a [customize](https://github.com/cybersemics/em/blob/66945aea4cf7d9c82a163f39f9822f1ca73f89f9/src/components/modals/CustomizeToolbar.tsx#L145) mode that allows users to add and remove commands from the toolbar. This can be accessed from the Settings. The click behavior is slightly different in order to accommodate drag-and-drop functionality. Familiarize yourself with customize mode and make sure that that functionality is preserved.

Issue #1953, opened 2024-05-26
Add blue arrow indicating off screen drop in sorted context


raineorshine:
## Steps to Reproduce

```
- US
  - New York
  - States
    - =pin
      - true
    - =sort
      - Alphabetical
        - Asc
    - Alaska
    - Arizona
    - Arkansas
    - California
    - Colorado
    - Connecticut
    - Delaware
    - Florida
    - Georgia
    - Hawaii
    - Idaho
    - Illinois
    - Indiana
    - Iowa
    - Kansas
    - Kentucky
    - Louisiana
    - Maine
    - Maryland
    - Massachusetts
    - Michigan
    - Minnesota
    - Mississippi
    - Missouri
    - Montana
    - Nebraska
    - Nevada
    - New Hampshire
    - New Jersey
    - New Mexico
    - North Carolina
    - North Dakota
    - Ohio
    - Oklahoma
    - Oregon
    - Pennsylvania
    - Rhode Island
    - South Carolina
    - South Dakota
    - Tennessee
    - Texas
    - Utah
    - Vermont
    - Virginia
    - Washington
    - West Virginia
    - Wisconsin
    - Wyoming
  - Alabama
```

a) Drag `New York` as if to drop it into `States`. 
b) Drag `Alabama` as if to drop it into `States`. 

## Current Behavior

`States` is highlighted, but the drop hover is off screen. 

## Expected Behavior

a) When the drop target is below the bottom of the screen, render a downward facing blue arrow at the bottom of the screen. 

b) When the drop target is above the top of the screen, render an upward facing blue arrow at the top of the screen. 

Animate the arrow with a gentle up-and-down bobbing motion.

raineorshine:
Duplicate of #1828

PR #2462, opened 2024-10-18
Lottie icons 2408 (Milestone 2)

ohrytskov:
#2408 

- Animate on ~click~ execute (except active → inactive)
  - Complete animation regardless of mousedown time, i.e. even if the mouse is released before the animation duration
  - Smooth appearance when clicked again before the first animation has completed
- Preserve drag-and-drop behavior in customize mode
  - Do not animate on drag


src/@types/Icon.ts
```diff
@@ -1,13 +1,28 @@
...
   height?: number
   size?: number
   style?: React.CSSProperties
   width?: number
+  animated?: boolean
+  animationComplete?: () => void
 }
 
-export default Icon
+export interface AnimatedIconType {
```
raineorshine, line 15: I think you can use `interface AnimatedIconType extends IconType` here for inheritance.

By convention on this project we generally export a single, default export per file, with the filename matching the type name. So `IconType` would go in `IconType.ts` and `AnimatedIconType` would go in `AnimatedIconType.ts`.

src/components/icons/AnimatedIcon.tsx
```diff
@@ -0,0 +1,44 @@
...
+  animated,
+  animationData,
+  children,
+  animationComplete,
+}: AnimatedIconType) => {
+  const newSize = size * ICON_SCALING_FACTOR
+  const color = style.fill || fill || token('colors.fg')
+
+  // Local state to manage animation status
+  const [isAnimated, setIsAnimated] = useState(animated)
```
raineorshine, line 24: Maybe you're setting up for future functionality, but is there a reason you are creating local state instead of just using the `animated` prop?

src/components/icons/AnimatedIcon.tsx
```diff
@@ -0,0 +1,44 @@
...
+  useEffect(() => {
+    setIsAnimated(animated)
+  }, [animated])
+
+  /** Calls the animationComplete callback if provided. */
+  const handleAnimationComplete = () => {
+    if (animationComplete) {
+      animationComplete()
+    }
+  }
```
raineorshine, line 35: Can you pass `animationComplete` directly to `<LottieAnimation>` instead of having the wrapper?

raineorshine (commented):
The animations looks great, and the behavior when clicking multiple times is perfect.

# 1. Only animate inactive → active

Note that for commands that have an active/inactive state (pin, pinChildren, table view, bold, italic, etc) the animation should only occur when going from inactive → active. It should not be animated from active → inactive, but rather just go gray.

For shortcuts that do not define an `isActive` function, then animation should be performed whenever the button is clicked.

# 2. Preserve click area

You may already be aware, but the click area of the toolbar icons has gotten smaller. In particular the extra padding below.

## Current Behavior

61890345d8

Screenshot 2024-10-22 at 7 17 51 PM


## Expected Behavior

`main`

Screenshot 2024-10-22 at 7 15 32 PM

src/components/icons/AnimatedIcon.tsx
ohrytskov: You're right that we could directly use the animated prop for the current functionality. However, the reason I’ve set up local state (isAnimated) is to allow for more flexible control over the animation in future updates.

src/@types/Icon.ts
ohrytskov: Good idea to use interface AnimatedIconType extending IconType. 
Also, I will separate the types into their own files following the project convention.

src/components/icons/AnimatedIcon.tsx
ohrytskov: That makes sense. If more control over the animation becomes necessary in future updates, we can always revisit it and reintroduce the state as needed.

src/components/icons/AnimatedIcon.tsx
ohrytskov: That’s a good point; we can pass animationComplete directly to <LottieAnimation> for now. I initially set it up with a wrapper to accommodate future enhancements, like ensuring a smooth reappearance of the animation.

ohrytskov:
Regarding the 'Preserve click area' point, it seems to be related to the scaling adjustments planned for milestone 3.

package.json
```diff
@@ -85,6 +85,7 @@
     "@mui/material": "^6.1.0",
     "algoliasearch": "^5.2.4",
     "axios": "^1.7.7",
+    "canvas": "^2.11.2",
```
raineorshine, line 88: Is `canvas` a new peer dependency requirement? I'm not seeing where this used.

ohrytskov: The canvas package is essential for resolving HTMLCanvasElement.prototype.getContext errors in our test environment. Without this package, running yarn test would result in numerous errors, as the canvas-related functionalities required for simulations would be missing.

raineorshine: Okay. It's not listed as a peer dependency of `lottie-react`, although I do see `jest-canvas-mock`. Is the `canvas` package related to that? Just trying to understand what depends on this new package, as it was not needed before.

ohrytskov: The react-lottie library, which is used for rendering Lottie animations in React, does rely on the canvas element through the bodymovin library. Bodymovin (part of lottie-web) uses canvas or SVG as its renderer, depending on the animation's complexity and settings. When adding react-lottie, this canvas dependency can require additional setup in testing environments like Jest or Vitest, particularly for mock or DOM rendering support.

trevinhofmann:
@ohrytskov -- There's still just one failing test case: https://github.com/cybersemics/em/actions/runs/11522295561/job/32077852690?pr=2462

When this is ready for review, please just let me know and I'll be happy to take a look right away. Thanks!

ohrytskov:
@trevinhofmann, all checks have passed.

src/components/icons/BoldTextIcon.tsx
```diff
@@ -1,46 +1,36 @@
...
-
-/** Bold icon. */
-const BoldTextIcon = ({ fill, size = 20, style = {}, cssRaw }: IconType) => {
-  const newSize = size * ICON_SCALING_FACTOR // Ensure sizing follows scaling factor
-  const strokeColor = style.fill || fill || token('colors.fg') // Calculate stroke color
+import IconType from '../../@types/IconType'
+import AnimatedIcon from './AnimatedIcon'
+import animationData from './animations/13-bold_4.json'
 
+/** Bold Text Icon with Conditional Lottie Animation. */
```
trevinhofmann, line 5: From this comment: https://github.com/cybersemics/em/pull/2462#pullrequestreview-2385991894

> 1. Only animate inactive → active

The Lottie icons with an active/inactive state are still being animated when going from active → inactive:

https://github.com/user-attachments/assets/21928723-4116-4930-b06a-dc9f3ef4d77a

src/components/icons/StrikethroughIcon.tsx
```diff
@@ -1,42 +1,38 @@
...
-
-/** Strikethrough icon. */
-const StrikethroughIcon = ({ fill, size = 20, style = {}, cssRaw }: IconType) => {
-  const newSize = size * ICON_SCALING_FACTOR
-  const strokeColor = style.fill || fill || token('colors.fg')
+import IconType from '../../@types/IconType'
+import AnimatedIcon from './AnimatedIcon'
+import animationData from './animations/16-strikethrough_3.json'
 
+/** Strikethrough Icon with Conditional Lottie Animation. */
```
trevinhofmann, line 5: The original bounce animation was implemented to cancel early if the user drags the cursor away from the button (cancelling the action). @raineorshine, should the Lottie animations also cancel early in this case? Currently, the bounce animation cancels but the Lottie animation does not:

https://github.com/user-attachments/assets/121a74e9-1202-47a7-8a58-f4eef1044114

yarn.lock
```diff
@@ -12,87 +12,149 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@algolia/client-abtesting@npm:5.2.4":
-  version: 5.2.4
-  resolution: "@algolia/client-abtesting@npm:5.2.4"
+"@algolia/client-abtesting@npm:5.9.1":
```
trevinhofmann, line 15: @raineorshine, do we have a standard approach for manually adding/updating dependencies? This PR only added `lottie-react` and `canvas`, yet the lockfile has many, many unrelated updates. I see we have `lockfile-lint` set up, but that is limited in what it can cover. It would be nice to have more narrow updates to the lockfile so that they can at least be somewhat reviewed for any possible concerns. Changes like this are also likely to lead to merge conflicts.

trevinhofmann (changes requested):
Thank you very much, @ohrytskov! There are a few issues that I found, but the animations in general look very nice.

src/components/icons/TextColor.tsx
```diff
@@ -1,28 +1,24 @@
...
-
-/** TextColor Icon. */
-const TextColorIcon = ({ fill, size = 20, style = {}, cssRaw }: IconType) => {
-  const newSize = size * ICON_SCALING_FACTOR
-  const strokeColor = style.fill || fill || token('colors.fg')
+import IconType from '../../@types/IconType'
+import AnimatedIcon from './AnimatedIcon'
+import animationData from './animations/17-text-color_2.json'
 
+/** TextColor Icon with Conditional Lottie Animation. */
```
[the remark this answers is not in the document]
raineorshine: This is a regression. I did fix the animation duration in e75cc36e5b, but both the bounce and lottie animations should complete in full.

yarn.lock
raineorshine: I don't have a process for that, but I'm open to suggestions.

raineorshine:
Important: The lottie animation should occur when the command is executed, not on touchstart/mousedown. The animations were designed to correspond to a successful state change. For example, pinning a thought causes the pin icon to wobble. They don't make sense to be animated at the start of the button press.

I'm sorry I didn't make that clear in the requirements. I spent so much time on that with the designer that I took it for granted!

ohrytskov:
Thank you for the feedback. I'm currently implementing adjustments based on the comments.

src/components/icons/BoldTextIcon.tsx
ohrytskov: @raineorshine, the shortcut is missing an isActive function here.

ohrytskov:
Concerning the shift that occurs after clicking and when the animation starts, this is related to scaling and is planned for milestone 3.

src/components/icons/BoldTextIcon.tsx
ohrytskov: @trevinhofmann, it seems @raineorshine is suggesting that for shortcuts missing an isActive function, the animation should trigger whenever the button is clicked:
"For shortcuts that do not define an isActive function, then animation should be performed whenever the button is clicked."

raineorshine: That’s correct. But change that to "when executed."

ohrytskov:
@trevinhofmann , could you review everything again and let me know if there’s anything else needed?

src/components/icons/BoldTextIcon.tsx
ohrytskov: @raineorshine , should the transition from bold text to normal be animated? Would this represent a state change successfully?

src/components/icons/ItalicTextIcon.tsx
```diff
@@ -1,39 +1,23 @@
...
-
-/** Italic icon. */
-const ItalicTextIcon = ({ fill, size = 20, style = {}, cssRaw }: IconType) => {
-  const newSize = size * ICON_SCALING_FACTOR // Ensure sizing follows scaling factor
-  const strokeColor = style.fill || fill || token('colors.fg') // Calculate stroke color
+import IconType from '../../@types/IconType'
+import AnimatedIcon from './AnimatedIcon'
+import animationData from './animations/14-italic_4.json'
 
+/** Italic Text Icon with Conditional Lottie Animation. */
```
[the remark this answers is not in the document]
trevinhofmann: > Concerning the shift that occurs after clicking and when the animation starts, this is related to scaling and is planned for milestone 3.

src/components/icons/OutdentIcon.tsx
```diff
@@ -1,86 +1,83 @@
...
-
-/** Outdent icon. */
-const OutdentIcon = ({ fill, size = 14, style = {}, cssRaw }: IconType) => {
-  const newSize = size * ICON_SCALING_FACTOR
-  const strokeColor = style.fill || fill || token('colors.fg')
+import IconType from '../../@types/IconType'
+import AnimatedIcon from './AnimatedIcon'
+import animationData from './animations/05-outdent_2.json'
 
+/** Outdent Icon with Conditional Lottie Animation. */
```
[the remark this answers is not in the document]
trevinhofmann: > Concerning the shift that occurs after clicking and when the animation starts, this is related to scaling and is planned for milestone 3.

src/components/icons/TextColor.tsx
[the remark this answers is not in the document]
trevinhofmann: This regression appears to have been resolved. Thanks @ohrytskov!

https://github.com/user-attachments/assets/34484656-8b0a-46c3-b1bf-380359ab3a9b

src/components/icons/BoldTextIcon.tsx
raineorshine: The transition from bold to normal would not be animated, as that represents a transition from active → inactive.

trevinhofmann: I believe these icons are now functioning as expected. My understanding is that the animation should trigger when activating bold/itallic/underline/strikethrough, and not when going back to normal, but I'll let @raineorshine answer your questions as well to be sure.

https://github.com/user-attachments/assets/6e94b28a-6a38-4fe2-a1e9-16a6769592cd

src/components/icons/StrikethroughIcon.tsx
trevinhofmann: The lottie animations no longer activate in this case, as expected 👍

src/components/icons/BoldTextIcon.tsx
trevinhofmann: Ah, there is one small issue with this still @ohrytskov.

When activating the icon, the animation plays (as expected). When deactivating the icon after the animation completes, the animation does not play (as expected).

However, when double-clicking to quickly activate and then deactivate the icon, the animation plays twice (not expected) instead of only once when activating:

https://github.com/user-attachments/assets/e111dedb-c469-4dbb-b897-744b1bc0b29f

yarn.lock
trevinhofmann: I'll look into this, but I'm positive there is a simple solution. Before I can do that, though, I need to figure out why my own `yarn` seems to be somewhat incompatible with what this project is using. When I update dependencies myself, the lockfile heading changes from:

```
# This file is generated by running "yarn install" inside your project.
# Manual changes might be lost - proceed with caution!
```

To:

```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1
```

Along with other formatting changes causing **_every_** line to get updated in the file 😅

I'm guessing this is a conflict between the old yarn installed with `npm` and the new yarn installed via `corepack enable`. I tried switching from the old one to the new one (assuming that is the issue), but haven't had any luck getting a different result yet. I'll debug that on my own time later.

trevinhofmann (changes requested):
Thank you for the update, @ohrytskov! The animations are now functioning much more as I'd expect. Can you please resolve the merge conflicts and also look into this double-clicking (quickly activating -> deactivating) issue? https://github.com/cybersemics/em/pull/2462#discussion_r1824250245

yarn.lock
raineorshine: Yes, it does sound like a yarn version mismatch. Maybe the executable is not correct in your PATH?

I'm on yarn `v4.4.1`.

ohrytskov:
The formatting icons store their active state in a different Redux state compared to other icons. Therefore, the animation has been adjusted.

raineorshine (changes requested):
I found some issues with the animation behavior. 

I have not reviewed code yet.

# 3. Animation persists active → inactive

The animation sometimes continues after the button enters the inactive state:

https://github.com/user-attachments/assets/750527fa-164b-43b0-b3c7-79faa7419bd6

# 4. Missing animation inactive → active

## 4a.

Sometimes the animation is not activated when entering the active state:

https://github.com/user-attachments/assets/ca3f03f3-a319-4750-9bcd-55cf7c419055

## 4b.

Sometimes the animation is not activated when executing a command without an active state:

https://github.com/user-attachments/assets/a3a6631f-1239-405a-bc5c-21913dadbc97

# 5. Disappearing button

Sometimes the button disappears and fades back in:

https://github.com/user-attachments/assets/a589f0e9-f2a0-4170-8be4-5e2d53194e12

yarn.lock
trevinhofmann: It was a journey, but I'm now on yarn `4.4.1` as well. Looks like `yarn add` typically only adds the required dependencies without doing miscellaneous updates.

@ohrytskov, I see you've cleaned up the unrelated `yarn.lock` updates when merging in the main branch with https://github.com/cybersemics/em/pull/2462/commits/936e4018eee49772a91a0cbdd8aae38ed809f106. Thank you very much for that!

Did you do anything other than `yarn add [package]` when initially adding the dependencies? I'm just trying to understand why there were so many changes previously.

ohrytskov:
@raineorshine, could you please run git pull to update the repository and then execute the yarn command? After that, remove the website cache, refresh the page, and retest everything. Additionally, note that opacity handling has been removed from the AnimatedIcon component.

raineorshine:
I can confirm that **5. Disappearing button** has been fixed. 

I was already working off the latest commit and issues **3** and **4** remain.

yarn.lock
ohrytskov: I'm also using Yarn v4.4.1 and Node.js v20.17.0.

ohrytskov: With corepack enabled.

ohrytskov:
@raineorshine, could you please review the new version?

raineorshine:
Thanks! I'm going to let @trevinhofmann review and make sure the behavior is 100% correct before I review again.

src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -172,7 +174,7 @@ describe('all platforms', () => {
 describe('mobile only', () => {
   beforeEach(async () => {
     await emulate(KnownDevices['iPhone 11'])
-  }, 5000)
+  }, 20000)
```
trevinhofmann, line 177: Did this timeout need to increase from 5s to 20s because of the additional bundle size with the animations? I'm guessing this is unavoidable, but just wanted to ask because we'd like to keep the test execution as brief as possible.

trevinhofmann (changes requested):
Nearly everything is working as expected including issues (1-5). For example, the active/inactive and double-click behavior works as intended:

https://github.com/user-attachments/assets/d2f88c6a-96e2-4ce4-bd47-62322c475715

---

@ohrytskov - I noticed one potential issue. The bounce animation sometimes freezes briefly at the bottom for certain buttons. It only seems to happen when there are more than a few thoughts, such as with the test case from https://github.com/cybersemics/em/issues/1953. I assume this may be related to the time it takes for the thoughtspace to be reorganized, but ideally that operation would not interfere with the animations:

https://github.com/user-attachments/assets/ea162f18-bca9-4318-8ecb-f9df2e7709f3

---

@raineorshine - The "Table View" button does not have an active/inactive state. Since it is also this way in the `main` branch, I assume that the issue (if it is an issue) is out of the scope of this PR. Just wanted to make a note, as I would expect it to be active when the table view is enabled for the selected thought:

https://github.com/user-attachments/assets/be5223cd-bff7-41cd-aba2-e2dfe15b2887

src/e2e/puppeteer/__tests__/caret.ts
ohrytskov: The temporary increase in Puppeteer test timeout to 20000ms was solely for debugging purposes. I’ve now restored it to the original setting.

raineorshine:
> @raineorshine - The "Table View" button does not have an active/inactive state. Since it is also this way in the `main` branch, I assume that the issue (if it is an issue) is out of the scope of this PR. Just wanted to make a note, as I would expect it to be active when the table view is enabled for the selected thought:

Yes, this is a regression I think from 79815dc. I can fix it separately. Thanks!

ohrytskov:
@trevinhofmann, commenting out the executeShortcutWithMulticursor line in ToolbarButton.tsx resolves the freezing issue, indicating that it’s not related to animation.

raineorshine:
I'd like to clarify some things about the button press getting stuck. Does that behavior exist in `main`? If not, what is causing it to appear now? 

Understanding that connection better will help us determine the best place to address it.

src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -187,10 +189,13 @@ describe('mobile only', () => {
...
     await clickBullet('b')
 
+    await scrollIntoView('[aria-label="Subcategorize"]')
+    await scrollBy('#toolbar', 50, 0)
+
     await click('[aria-label="Subcategorize"]')
 
     const textContext = await getSelection().focusNode?.textContent
-    expect(textContext).toBe('')
+    expect([undefined, '']).toContain(textContext)
```
trevinhofmann, line 198: @ohrytskov -- Could you please explain this change? I would expect the value to consistently be either `''` or `undefined` rather than needing to use `[].toContain`. And, if the value has changed from `''` to `undefined` with this PR, it's not obvious to me why that would be.

trevinhofmann (approved):
@raineorshine - The button also sticks the same way in `main`, yes. For that reason, I would be more likely to consider this a performance (?) issue outside the scope of this PR, but I'll leave that up to you.

`main`:

https://github.com/user-attachments/assets/801d2c50-3a55-43e2-9d09-2ccdd4cbb222

`ohrytskov/Lottie-icons-2408-Milestone-2`:

https://github.com/user-attachments/assets/bcf707eb-2cb8-44c3-ab15-2b093e312e7e

---

Aside from that, I'm not seeing any other issues with the functionality. The code changes look good to me as well. I have one question for @ohrytskov about the change to the `caret.ts` test, but I don't think it is a blocking issue.

I'm also not the biggest fan of committing large/unreviewable JSON files for the animations, but I assume there's no simple way around that for now 👍

raineorshine (approved):
> @raineorshine - The button also sticks the same way in `main`, yes. For that reason, I would be more likely to consider this a performance (?) issue outside the scope of this PR, but I'll leave that up to you.

Great, thanks! Yes, if it's a problem in `main` then I think we should create a separate issue for it.

> Aside from that, I'm not seeing any other issues with the functionality. The code changes look good to me as well.

I reviewed and it all looks good. Click and animation timing is perfect 👌

Thank you both ❤️

src/e2e/puppeteer/__tests__/caret.ts
ohrytskov: This change is temporary and is aimed at improving the stability of tests.

ohrytskov: I'm already working on milestone 3, where I'll create a revised and clearer version of this test with some slight improvements.

PR #2463, opened 2024-10-19
Add puppeteer video recorder helper

mstrofbass:
Proposal to add a Puppeteer video recording helper to make it easier to debug puppeteer tests.

I'm going to be using this regardless of whether it's merged or not; I just figured I'd offer it up since it may prove to be useful for others.

### DIFF
--- a/src/e2e/puppeteer/helpers/recordVideo.ts
+++ b/src/e2e/puppeteer/helpers/recordVideo.ts
@@ -2,7 +2,7 @@ import { PuppeteerScreenRecorder } from 'puppeteer-screen-recorder'
 import { page } from '../setup'
 
 /** Initialize recorder. */
-async function init() {
+function init() {
   const recorder = new PuppeteerScreenRecorder(page)
 
   /** Start recording. */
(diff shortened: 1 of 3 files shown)


mstrofbass:
This has puppeteer-screen-recorder pinned to version 3.0.3 because 3.05 seems to have a bug with the build/packaging that breaks importing. It should be unpinned once that is resolved in the dependency.

src/e2e/puppeteer/helpers/recordVideo.ts
```diff
@@ -0,0 +1,20 @@
+import { PuppeteerScreenRecorder } from 'puppeteer-screen-recorder'
+import { page } from '../setup'
+
+/** Initialize recorder. */
+async function init() {
```
raineorshine, line 5: `init` should not be async

raineorshine (changes requested):
Sure! This looks useful.

src/e2e/puppeteer/helpers/recordVideo.ts
mstrofbass: Good catch!

mstrofbass:
When I start using it a little more I'll look into configuring a directory and then adding that directory to the .gitignore. We could probably just add `*.mp4` to the .gitignore file for now as well, but I should have a more robust solution soon enough.

raineorshine:
FYI I decided to remove puppeteer-screen-recorder because the ffmpeg dependency was a bit heavy and it's not being actively used by the team.

Issue #2110, opened 2024-07-06
Remove puppeteer test delays


raineorshine:
1. Disable all CSS transitions ~~on autofocus, thoughts, modals, and alerts~~ during tests. 

   Having to wait for the animations slows down snapshot tests considerably.

   The work has been started in #2343. Take a look at how the PandaCSS _test condition is utilized to conditionally set the animation durations to 0s. Any remaining CSS transitions that slow down the tests (namely modals and alerts) need to be converted.

2. Once animations are disabled, remove the artificial delays that were added to the puppeteer tests due to the need to wait for the animations. 
  
   If a delay cannot be removed but can be reduced, reduce it as much as possible. Instead of `sleep`, consider waiting for an element to become visible. 
  
   Make sure there are no intermittent failures by running the tests multiple times.

mstrofbass:
@raineorshine a couple of questions on this.

1. Do the CSSTransition React components need any modification under this issue?
2. Is there a better way to identify the particular locations that need to be updated besides searching for all references to the `transition` CSS property?
3. Should all instances of the `transition` property that have non-zero timings be set to 0 or will there be some transitions that still need a delay? If the latter, how do I identify those?
4. It's not clear to me which transitions should be shared. More particularly, some transition timings are the same between distinct CSS elements and it's not immediately clear whether they should use the same setting or they should be unqiue values.

I think #4 is most easily solvable by me just doing distinct settings for each and then letting you review the changes and tell me which can be combined (if any) since that puts the onus on me to find them first.

raineorshine:
I think this answers a couple questions: *All* transitions should be disabled during the tests. That is the simplest and most effective approach. There is never a case where an automated test would cover the animation itself, just the beginning and end states. I originally wrote this issue before the PandaCSS conversion (which is still in progress). The `_test` conditions makes this much cleaner. 

As for shared transitions, I agree let's keep them separate for now (except the ones that have already been combined). We can combine later in a refactor, but I would say that’s a separate scope from this work.

mstrofbass:
One clarification needed:

The CSSTransition classes are React components that take an integer timeout value in milliseconds. The PandaCSS tokens are string values and include the units (e.g., `300ms`, `0.1s`). We can still use the PandaCSS tokens and apply them to the CSSTransition classes by (1) excluding the units and (2) using `parseInt`, but it's a bit ugly and not consistent. Is there a better way to handle the CSSTransition classes?

Example PandaCSS:

```javascript
 durations: {
[...]

PR #2464, opened 2024-10-19
Move first tranche of transition durations to PandaCSS 

mstrofbass:
More work on #2110 

This moves most of the CSS transition durations to `panda.config.ts` and then configures the transitions based on the value in `panda.config.ts`. The durations in `panda.config.ts` are set to `0ms` when in testing.

There are a few remaining durations that have not been moved but may be causing problems.


raineorshine (approved):
This looks great!

raineorshine (changes requested):
The code looks great, but some of the animations seem to be broken:

1. Alert - Delete a non-empty thought. The alert should fade in and after a delay fade out.
2. Toolbar buttons - Should animate when pressed.
3. NavBar breadcrumbs - Should fade out when user starts typing in a thought.

PR #2482, opened 2024-10-20
Make transition durations dynamic Part 1

mstrofbass:
This replaces PR #2464 and makes progress on #2110 

The changes here are:

1. We now have a top-level config object for the duration timings.
2. We have a helper that wraps the config with getters that return `0` instead of the duration when we're in the e2e tests.
3. The PandaCSS config is generated based on the top-level config.
4. The CSSTransition components use the helper to get the appropriate duration.

NOTE: I wanted to be able to test the helper under testing conditions so I had to add some provisions to allow for overriding the state and force it to operate as if it was in tests. We could make this a bit simpler if we brought back the `isE2E` helper (in it's functional form) now that we're checking this in multiple places; then `isE2E` could be easily mocked so we can test both paths.

### DIFF
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -2,6 +2,7 @@ import React, { useCallback, useRef } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { css, cva, cx } from '../../styled-system/css'
 import { bullet } from '../../styled-system/recipes'
+import { token } from '../../styled-system/tokens'
 import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import ThoughtId from '../@types/ThoughtId'
@@ -54,7 +55,7 @@ const glyph = cva({
         position: 'relative',
         marginLeft: '-16.8px',
         opacity: 0.9,
-        transition: 'opacity 0.75s ease-in-out',
+        transition: `opacity ${token('durations.bulletOpacityDuration')} ease-in-out`,
         marginRight: '-5px',
         left: '3px',
         fontSize: '16px',
@@ -65,7 +66,7 @@ const glyph = cva({
         position: 'relative',
         marginLeft: '-16.8px',
         opacity: 0.8,
-        transition: 'opacity 0.75s ease-in-out',
+        transition: `opacity ${token('durations.bulletOpacityDuration')} ease-in-out`,
         marginRight: '-5px',
         left: '4px',
         fontSize: '28px',
@@ -208,7 +209,7 @@ const glyph = cva({
 
 const glyphFg = cva({
   base: {
-    transition: 'transform 0.1s ease-out, fill-opacity 0.5s ease-out',
+    transition: `transform ${token('durations.bulletFgTransformDuration')} ease-out, fill-opacity ${token('durations.bulletFgOpacityDuration')} ease-out`,
   },
   variants: {
     gray: {
@@ -524,7 +525,7 @@ const Bullet = ({
           },
           '@media (min-width: 560px) and (max-width: 1024px)': {
             _android: {
-              transition: 'transform 0.1s ease-in-out',
+              transition: `transform ${token('durations.bulletFgTransformDuration')} ease-in-out`,
               marginLeft: '-3px',
             },
           },
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -2,7 +2,6 @@ import React, { useCallback, useRef } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { css, cva, cx } from '../../styled-system/css'
 import { bullet } from '../../styled-system/recipes'
-import { token } from '../../styled-system/tokens'
 import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import ThoughtId from '../@types/ThoughtId'
@@ -55,7 +54,7 @@ const glyph = cva({
         position: 'relative',
         marginLeft: '-16.8px',
         opacity: 0.9,
-        transition: `opacity ${token('durations.bulletOpacityDuration')} ease-in-out`,
+        transition: `opacity {durations.bulletOpacityDuration} ease-in-out`,
         marginRight: '-5px',
         left: '3px',
         fontSize: '16px',
@@ -66,7 +65,7 @@ const glyph = cva({
         position: 'relative',
         marginLeft: '-16.8px',
         opacity: 0.8,
-        transition: `opacity ${token('durations.bulletOpacityDuration')} ease-in-out`,
+        transition: `opacity {durations.bulletOpacityDuration} ease-in-out`,
         marginRight: '-5px',
         left: '4px',
         fontSize: '28px',
@@ -209,7 +208,7 @@ const glyph = cva({
 
 const glyphFg = cva({
   base: {
-    transition: `transform ${token('durations.bulletFgTransformDuration')} ease-out, fill-opacity ${token('durations.bulletFgOpacityDuration')} ease-out`,
+    transition: `transform {durations.bulletFgTransformDuration} ease-out, fill-opacity {durations.bulletFgOpacityDuration} ease-out`,
   },
   variants: {
     gray: {
@@ -526,7 +525,7 @@ const Bullet = ({
           },
           '@media (min-width: 560px) and (max-width: 1024px)': {
             _android: {
-              transition: `transform ${token('durations.bulletFgTransformDuration')} ease-in-out`,
+              transition: `transform {durations.bulletFgTransformDuration} ease-in-out`,
               marginLeft: '-3px',
             },
           },
--- a/src/components/ErrorBoundaryContainer.tsx
+++ b/src/components/ErrorBoundaryContainer.tsx
@@ -26,7 +26,7 @@ const Toggle = ({ children, expand, title }: { children?: any; expand?: boolean;
               className={css({
                 transform: expanded ? 'rotate(90deg) translateX(10px)' : '',
                 transformOrigin: 'center center',
-                transition: 'transform 0.1s ease-out',
+                transition: `transform ${token('durations.triangleToggleTransformDuration')} ease-out`,
               })}
             />
           </g>
--- a/src/components/ErrorBoundaryContainer.tsx
+++ b/src/components/ErrorBoundaryContainer.tsx
@@ -26,7 +26,7 @@ const Toggle = ({ children, expand, title }: { children?: any; expand?: boolean;
               className={css({
                 transform: expanded ? 'rotate(90deg) translateX(10px)' : '',
                 transformOrigin: 'center center',
-                transition: `transform ${token('durations.triangleToggleTransformDuration')} ease-out`,
+                transition: `transform {durations.triangleToggleTransformDuration} ease-out`,
               })}
             />
           </g>
--- a/src/components/Favorites.tsx
+++ b/src/components/Favorites.tsx
@@ -123,7 +123,7 @@ const FavoritesOptions = ({
             style={{
               display: 'inline-block',
               transform: `rotate(${showOptions ? 90 : 0}deg)`,
-              transition: 'transform 150ms ease-out',
+              transition: `transform ${token('durations.favoritesSlidedownDuration')} ease-out`,
               // avoid position:absolute to trivially achieve correct vertical alignment with text
               marginLeft: '-1em',
             }}
--- a/src/components/Favorites.tsx
+++ b/src/components/Favorites.tsx
@@ -4,6 +4,7 @@ import { useDispatch, useSelector } from 'react-redux'
 import CSSTransition from 'react-transition-group/CSSTransition'
 import { css, cx } from '../../styled-system/css'
 import { dropHover } from '../../styled-system/recipes'
+import { token } from '../../styled-system/tokens'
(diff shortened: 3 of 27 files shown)


src/durations.config.ts
```diff
@@ -0,0 +1,45 @@
+export type DurationConfig = Record<string, number>
+
+const durationsMillis: DurationConfig = {
```
raineorshine, line 3: This would be better typed `as const` so that we can extract the keys as compile time.

src/util/durations.ts
```diff
@@ -0,0 +1,41 @@
+import durationsConfig, { type DurationConfig } from '../durations.config'
+
+/** Initialize duration helper. */
+function init() {
+  let inTest: boolean = !!navigator.webdriver
+
+  /** Returns the duration for a particular key or undefined if it doesn't exist. */
+  const get = (key: string): number => {
```
raineorshine, line 8: Since the duration keys can be determined at compile-time, we can use Typescript to restrict the type here instead of doing a runtime check. This will provide better DX through autosuggest as well.

raineorshine (changes requested):
Looks good! I can confirm the animations are still working.

Just one type suggestion then ready to merge.

> NOTE: I wanted to be able to test the helper under testing conditions so I had to add some provisions to allow for overriding the state and force it to operate as if it was in tests. We could make this a bit simpler if we brought back the `isE2E` helper (in it's functional form) now that we're checking this in multiple places; then `isE2E` could be easily mocked so we can test both paths.

I think the benefit we get from this additional testing is probably not worth the additional complexity. It turns what is essentially an if-statement that zeros out the duration into an object-oriented class with its own interface and set of tests. The duration logic is not likely to change in the future and not likely to develop regressions. 

I would generally reserve testing for 1) high-level integration testing of user behavior, or 2) low-level unit testing of nontrival functionality. Testing the test framework has diminishing returns, especially when the added test complexity exceeds the complexity of the original logic.

Maybe I'm wrong about this and the long-term benefit exceeds the additional complexity. Or perhaps I'm underestimating the possibility of regressions in the duration switch. Anyway, just wanted to share my perspective. I'm fine to leave this in since it is already done. Thanks!

raineorshine:
@mstrofbass I resolved some merge conflicts, so make sure to pull the latest (dbc8719a2e).

mstrofbass:
> Looks good! I can confirm the animations are still working.
> 
> Just one type suggestion then ready to merge.
> 
> > NOTE: I wanted to be able to test the helper under testing conditions so I had to add some provisions to allow for overriding the state and force it to operate as if it was in tests. We could make this a bit simpler if we brought back the `isE2E` helper (in it's functional form) now that we're checking this in multiple places; then `isE2E` could be easily mocked so we can test both paths.
> 
> I think the benefit we get from this additional testing is probably not worth the additional complexity. It turns what is essentially an if-statement that zeros out the duration into an object-oriented class with its own interface and set of tests. The duration logic is not likely to change in the future and not likely to develop regressions.
> 
> I would generally reserve testing for 1) high-level integration testing of user behavior, or 2) low-level unit testing of nontrival functionality. Testing the test framework has diminishing returns, especially when the added test complexity exceeds the complexity of the original logic.
> 
> Maybe I'm wrong about this and the long-term benefit exceeds the additional complexity. Or perhaps I'm underestimating the possibility of regressions in the duration switch. Anyway, just wanted to share my perspective. I'm fine to leave this in since it is already done. Thanks!

It's a philosophical difference that's not a quick discussion but a few reasons I prefer it:

1. 90% of the bugs I encounter are really simple, low-level bugs that this type of testing catches but is easily missed with e2e and integration tests;
2. These really simple tests are really fast to write and have low cognitive load, so you can get broad coverage with little mental effort...integration tests require a lot more in-depth thought about how components work together so it's easier to miss things
3. e2e testing generally captures most of the stuff integration tests target, so integration tests don't really provide much extra except for certain scenarios
4. Low-level unit tests make refactoring significantly easier because it's super simple to figure out if the refactored code works just like the old; 
5. For an open-source project, low level unit tests should make it easier to be confident that the contributor did proper testing because you should be able to look through 

It's also widely considered a best practice...this is a bit of an appeal to authority that I don't care for but it's worth pointing out that there are a lot of justifications for it.

My general preference is full coverage via true unit tests, as full coverage as I can get via e2e tests, and integration tests for really complex portions that are difficult to test via e2e tests. Most of my testing is in backend stuff, so this paradigm gets a bit iffy in UI land but the general philosophy is still applicable.

I'm fine either way, I just default to writing low-level tests unless someone tells me not to since it's easier to write them when writing the code than it is to come back later and write them. I will also modify my style to your preferences; it will just take some probing here and there to figure out what they are!

mstrofbass:
Will get to the requested changes later this evening!

src/components/ErrorBoundaryContainer.tsx
```diff
@@ -26,7 +26,7 @@ const Toggle = ({ children, expand, title }: { children?: any; expand?: boolean;
               className={css({
                 transform: expanded ? 'rotate(90deg) translateX(10px)' : '',
                 transformOrigin: 'center center',
-                transition: 'transform 0.1s ease-out',
+                transition: `transform ${token('durations.triangleToggleTransformDuration')} ease-out`,
```
yangchristina, line 29:
```suggestion
                transition: `transform {durations.triangleToggleTransformDuration} ease-out`,
```

src/components/Favorites.tsx
```diff
@@ -128,7 +129,7 @@ const FavoritesOptions = ({
             className={css({
               display: 'inline-block',
               transform: showOptions ? `rotate(90deg)` : `rotate(0deg)`,
-              transition: 'transform 150ms ease-out',
+              transition: `transform ${token('durations.favoritesSlidedownDuration')} ease-out`,
```
yangchristina, line 132:
```suggestion
              transition: `transform {durations.favoritesSlidedownDuration} ease-out`,
```

src/components/LatestShortcutsDiagram.tsx
```diff
@@ -32,11 +34,11 @@ const LatestShortcutsDiagram: FC<LatestShortcutsDiagramProps> = ({ position = 'm
         in={latestShortcuts.length > 0}
         classNames={{
           enter: css({ opacity: 0 }),
-          enterActive: css({ opacity: 1, transition: 'opacity 400ms' }),
+          enterActive: css({ opacity: 1, transition: `opacity ${token('durations.latestShortcutsOpacityDuration')}` }),
```
yangchristina, line 37:
```suggestion
          enterActive: css({ opacity: 1, transition: `opacity {durations.latestShortcutsOpacityDuration}` }),
```

src/components/NewThought.tsx
```diff
@@ -74,7 +75,9 @@ const NewThought = ({ path, showContexts, label, value = '', type = 'bullet' }:
   }, [dispatch, distance, path, value])
 
   return show ? (
-    <ul className={css({ transition: 'all 0.75s ease-out', marginTop: 0 })}>
+    <ul
+      className={css({ transition: `all ${token('durations.newThoughtTransitionDuration')} ease-out`, marginTop: 0 })}
```
yangchristina, line 79:
```suggestion
      className={css({ transition: `all {durations.newThoughtTransitionDuration} ease-out`, marginTop: 0 })}
```

src/components/Bullet.tsx
```diff
@@ -65,7 +66,7 @@ const glyph = cva({
         position: 'relative',
         marginLeft: '-16.8px',
         opacity: 0.8,
-        transition: 'opacity 0.75s ease-in-out',
+        transition: `opacity ${token('durations.bulletOpacityDuration')} ease-in-out`,
```
yangchristina, line 69:
```suggestion
        transition: `opacity {durations.bulletOpacityDuration} ease-in-out`,
```

src/components/Bullet.tsx
```diff
@@ -208,7 +209,7 @@ const glyph = cva({
 
 const glyphFg = cva({
   base: {
-    transition: 'transform 0.1s ease-out, fill-opacity 0.5s ease-out',
+    transition: `transform ${token('durations.bulletFgTransformDuration')} ease-out, fill-opacity ${token('durations.bulletFgOpacityDuration')} ease-out`,
```
yangchristina, line 212:
```suggestion
    transition: `transform {durations.bulletFgTransformDuration} ease-out, fill-opacity {durations.bulletFgOpacityDuration} ease-out`,
```

src/components/Bullet.tsx
```diff
@@ -525,7 +526,7 @@ const Bullet = ({
           },
           '@media (min-width: 560px) and (max-width: 1024px)': {
             _android: {
-              transition: 'transform 0.1s ease-in-out',
+              transition: `transform ${token('durations.bulletFgTransformDuration')} ease-in-out`,
```
yangchristina, line 529:
```suggestion
              transition: `transform {durations.bulletFgTransformDuration} ease-in-out`,
```

src/components/Tips/Tip.tsx
```diff
@@ -19,7 +20,7 @@ const Tip: FC<
         // disable pointer revents when hidden, otherwise it will block clicks on the NavBar
         pointerEvents: display ? 'auto' : 'none',
         transform: display ? 'translateY(0)' : 'translateY(100%)',
-        transition: 'transform 200ms ease-in-out, opacity 200ms ease-in-out',
+        transition: `transform ${token('durations.tipTransitionDuration')} ease-in-out, opacity ${token('durations.tipTransitionDuration')} ease-in-out`,
```
yangchristina, line 23:
```suggestion
        transition: `transform {durations.tipTransitionDuration} ease-in-out, opacity {durations.tipTransitionDuration} ease-in-out`,
```

src/components/Tutorial/TutorialNavigation.tsx
```diff
@@ -58,7 +59,7 @@ const TutorialNavigation = ({
                   fontSize: '32px',
                   marginLeft: '1px',
                   marginRight: '1px',
-                  transition: 'all 400ms ease-in-out',
+                  transition: `all ${token('durations.tutorialStepNavigationDuration')} ease-in-out`,
```
yangchristina, line 62:
```suggestion
                  transition: `all {durations.tutorialStepNavigationDuration} ease-in-out`,
```

src/components/modals/Export.tsx
```diff
@@ -664,7 +665,7 @@ const ModalExport: FC<{ simplePaths: SimplePath[] }> = ({ simplePaths }) => {
                 userSelect: 'none',
                 display: 'flex',
                 position: 'relative',
-                transition: 'opacity 100ms ease-in-out',
+                transition: `opacity ${token('durations.advancedSettingsLinkOpacityDuration')} ease-in-out`,
```
yangchristina, line 668:
```suggestion
                transition: `opacity {durations.advancedSettingsLinkOpacityDuration} ease-in-out`,
```

yangchristina (commented):
`token()` is for using inside the `style` prop. For `cva` and `css`, you can use ` '...{durations.tokenName} ...'` directly, and if the property only consists of `'{duration.tokenName}'`, you can put in `'tokenName'` directly.

raineorshine:
Thanks for expanding on that!

You make great points, so I want to soften my position. Unit tests are essential for preventing regressions, and I think we should strive for as much coverage as possible. The existing unit tests have been indispensable for several major refactors, and I've never found myself wanting fewer. Integration tests should not be thought of as a replacement for unit tests, but rather as complementary. All the reasons you provided I am in deep agreement with. 

To narrow it down, I guess what was most relevant here was mocking, and how it can add complexity if we're not careful. The `isE2E` would allow for easy mocking as you pointed out. There might also be a way to mock `navigator.webdriver` directly, but I don't know off the top of my head. 

When mocking and/or testing the test coverage comes up again, let's discuss so we can consider the tradeoffs. Otherwise, I'm all for full test coverage, especially for `src/actions`, `src/selectors`, and `src/util` :).

src/durations.config.ts
mstrofbass: Good call! This is not a TypeScript feature I've used before.

src/util/durations.ts
mstrofbass: I think I got this right (I at least got the error to go away...)

mstrofbass:
> To narrow it down, I guess what was most relevant here was mocking, and how it can add complexity if we're not careful. The `isE2E` would allow for easy mocking as you pointed out. There might also be a way to mock `navigator.webdriver` directly, but I don't know off the top of my head.

I tend to avoid mocking global or language stuff like that because it's frequently significantly more work than just wrapping it. I went ahead and tried to mock `navigator.webdriver` but got an error saying that `webdriver` was not defined. I was copying examples from the web so it should work since we're using `jsdom` but I gave up after a bit. Happy to circle back around to this after we get some of the bigger stuff knocked out.

src/util/durations.ts
```diff
@@ -0,0 +1,39 @@
+import durationsConfig from '../durations.config'
+
+/** Initialize duration helper. */
+function init() {
+  let inTest: boolean = !!navigator.webdriver
+
+  /** Returns the duration for a particular key or undefined if it doesn't exist. */
+  const get = (key: keyof typeof durationsConfig): number => {
```
raineorshine, line 8: Exactly!

PR #2487, opened 2024-10-21
1901: [Safari] Table: Col1 cursor bullet incorrectly rendered on the right

### LINKED ISSUE #1901: [Safari] Table: Col1 cursor bullet incorrectly rendered on the right
Safari only

## Steps to Reproduce

``` 

- X
  - =view
    - Table
  - Atlanta
    - A1
  - Boston
    - =note
      - This is a note
    - B1
```

Set the cursor on `Boston`. 

## Current Behavior

The cursor bullet is rendered in the wrong position. This only occurs when the cursor thought has a note. 

image

## Expected Behavior

Cursor bullet should always be rendered to the left of the thought.

kevinstubbs:
Fixes #1901

The issue was caused by margin CSS properties being ignored by Safari when the element is absolutely positioned.

Screenshot from safari.
image

### DIFF
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -541,9 +541,8 @@ const Bullet = ({
         }),
       )}
       style={{
-        marginTop: -extendClickHeight,
-        marginLeft: -extendClickWidth + marginLeft,
-        marginBottom: -extendClickHeight - 2,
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Tested and works well! The extended click area still looks good.

Issue #2425, opened 2024-09-26
Thought not trimmed


raineorshine:
## Steps to Reproduce

1. Create a new thought with leading or trailing spaces. 
2. Set the cursor to null. 

## Current Behavior

The spaces are preserved. 

## Expected Behavior

The thought should be trimmed on blur. 

## Notes

This is a regression so start by identifying the faulty commit and posting it here. 

Add a unit test to prevent this from happening again.

PR #2488, opened 2024-10-22
fix thought trim error

RED-ROSE515:
Issue : Thought not trimmed #2425
Fix the issue by updating the contentRef with the trimmed value.

Faulty commit :   Handle HTML escape on paste and type (#2348)  - e488361b8e8a7395f402f4fc860486bad79c2f6a

### DIFF
--- a/src/e2e/puppeteer/__tests__/thought-trim.ts
+++ b/src/e2e/puppeteer/__tests__/thought-trim.ts
@@ -1,0 +1,27 @@
+import click from '../helpers/click'
+import getEditingText from '../helpers/getEditingText'
+import paste from '../helpers/paste'
+import press from '../helpers/press'
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+it('when cursor is null, thought will be trimmed.', async () => {
+  const importText = `
+  -    Hello World.      `
+
+  await paste(importText)
+  await click('#content')
+  const editingText = await getEditingText()
+  expect(editingText).toBe(`Hello World.`)
+})
+
+it('when change the row, thought will be trimmed.', async () => {
+  const importText = `
+  -       Hello World.      `
+
+  await paste(importText)
+  await press('Enter')
+  await press('ArrowUp')
+  const editingText = await getEditingText()
+  expect(editingText).toBe(`Hello World.`)
+})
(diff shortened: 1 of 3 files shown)


raineorshine:
This was a regression, as mentioned in the issue, so we have to start by identifying the faulty commit. That will inform which solution we decide to pursue. 

I recommend using git bisect.

raineorshine (changes requested):
This was a regression, as mentioned in the issue, so we have to start by identifying the faulty commit. That will inform which solution we decide to pursue.

I recommend using git bisect.

src/e2e/puppeteer/__tests__/thought-trim.ts
```diff
@@ -0,0 +1,27 @@
...
+import press from '../helpers/press'
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+it('when cursor is null, thought will be trimmed.', async () => {
+  const importText = `
+  -    Hello World.      `
+
+  await paste(importText)
+  await click('#content')
```
raineorshine, line 13: Let's avoid tightly coupling this test to the unrelated `#content` element. Use `selection.clear` or activate cursorBack instead.

src/e2e/puppeteer/__tests__/thought-trim.ts
```diff
@@ -0,0 +1,27 @@
...
+  const editingText = await getEditingText()
+  expect(editingText).toBe(`Hello World.`)
+})
+
+it('when change the row, thought will be trimmed.', async () => {
+  const importText = `
+  -       Hello World.      `
+
+  await paste(importText)
```
raineorshine, line 22: It would be good to have one test for paste and one test that creates a thought and types a value with spaces at the beginning/end. Otherwise we're only testing the paste case.

raineorshine (changes requested):

raineorshine:
On my side is correctly trims when the thought is edited and the cursor moves away, but not on import (paste).

## Steps to Reproduce

Create a new thought and paste `    one`.

## Current Behavior

Thought is not trimmed.

## Expected Behavior

Thought should be trimmed on import.

raineorshine (approved):
Thanks!

PR #2489, opened 2024-10-22
Panda css general

yangchristina:
No styles exist for .shortcut-tab-wrapper, .toolbar-popup
Unused classes: expand-click-area-left, expand-click-area-right, subthought-highlight, down-chevron, up-chevron, add-new-thought, add-new-thought-link, has-other-contexts, center-in-content, ellipsized-path, modal-icon-inner, modal-icon, center-desktop


src/components/Loader.tsx
```diff
@@ -19,21 +21,24 @@ const rippleLoaderChild = {
 /**
  * Loading component.
  */
-const Loader = ({ size = 32, style }: LoaderProps) => {
+const Loader = ({ size = 32, style, cssRaw }: LoaderProps) => {
```
trevinhofmann, line 24: The `style` prop is now unused in the only place that calls `<Loader>`. Can we remove it?

Additionally, that passed style (`marginRight: '15px'`) used to apply to the outer div, but has been moved to the inner div. I tested the difference locally, and it looks like `marginRight` doesn't have any effect on the inner div, but it did on the outer div.

src/components/ColorPicker.tsx
```diff
@@ -144,73 +147,67 @@ const ColorSwatch: FC<{
...
       )}
     </span>
   )
 }
 
 /** Text Color Picker component. */
-const ColorPicker: FC<{ fontSize: number; style?: React.CSSProperties }> = ({ fontSize, style }) => {
+const ColorPicker: FC<{ fontSize: number; style?: React.CSSProperties; cssRaw?: SystemStyleObject }> = ({
+  fontSize,
+  style,
```
trevinhofmann, line 181: The `style` prop is no longer passed by the only component using `<ColorPicker>`. Can we remove it?

https://github.com/cybersemics/em/blob/3d25ccaa6270bf91b2f032fc3dc6d8079d97f4ae/src/components/icons/TextColorWithColorPicker.tsx#L43

src/components/HighlightedText.tsx
```diff
@@ -21,13 +18,7 @@ const HighlightedText: FC<{ value: string; match: string; disabled?: boolean }>
...
         return (
           <span key={i}>
-            <span
-              style={{
-                color: !disabled && isMatch ? colors.vividHighlight : undefined,
-              }}
-            >
-              {char}
-            </span>
+            <span className={css({ color: !disabled && isMatch ? 'vividHighlight' : undefined })}>{char}</span>
```
trevinhofmann, line 21: Just a note for posterity: I thought that dynamic styles like these needed to be passed to `style` instead of generating a className, but it looks like Panda is able to generate multiple classNames for conditionals like this.

https://panda-css.com/docs/guides/dynamic-styling#runtime-conditions

I tested it out on another component, and it does indeed generate multiple classNames and switch between them as expected. Very cool.

trevinhofmann (changes requested):
Looks great! Thank you, @yangchristina.

It's somewhat challenging to review when the PR title/description don't fully explain what the goal of the PR is. Regardless, this is clearly great progress in the refactor toward Panda CSS.

I tested the app some, and everything generally seems to still be working/styled as expected 👍

There are just a few comments, some of which do not need to be addressed.

src/components/Loader.tsx
```diff
@@ -19,21 +21,24 @@ const rippleLoaderChild = {
...
+        className={css(
+          {
+            width: '100%',
+            height: '100%',
+            position: 'relative',
+            transform: 'translateZ(0) scale(0.49)',
+            backfaceVisibility: 'hidden',
+            transformOrigin: '0 0',
+          },
+          cssRaw,
```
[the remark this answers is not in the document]
raineorshine: Just to offer my perspective: I do tend to side with the counter argument when it comes to margin. The Loader should probably not be responsible for how it is positioned in the parent component. So it's about separation of concerns. 

Perhaps the better solution here would be to wrap the Loader in a div and assign the margin to that. Then you don't have to pass any css across the component boundary.

src/components/HighlightedText.tsx
raineorshine: Yes, I went through the same discovery process! As long as the resulting value is static (in this case the string 'vividHighlight' or undefined), then PandaCSS handles the ternary just fine.

src/components/Loader.tsx
yangchristina: Thanks for catching this! @raineorshine, there's been a good number of merge requests since the Loader commit, is it okay if I don't squash the fix into the original commit?

raineorshine: Sure, that's fine!

FYI For the future I do prefer rebasing on `main` instead of merge commits unless a branch has fallen really far behind. That will create a cleaner commit history, which I want to preserve in the case of the PandaCSS migration so that we can trace regressions more granularly.

yangchristina: Sounds good!

raineorshine (approved):

PR #2490, opened 2024-10-22
Fix URL icon sometimes not appearing

### LINKED ISSUE #2481: Url icon does not appear on valid url
## Steps to Reproduce

```
https://test.com/
```

## Current Behavior

The url icon does not appear.

The behavior seems to be inconsistent. Sometimes after editing and trying a few different urls the icon shows up.

## Expected Behavior

The url icon should be rendered next to the url.

kevinstubbs:
Fixes #2481 by ignoring HTML-like tags when checking if a given string is a URL.

image

I also noticed the same tag stripping regex was being used in the bulletColor action, so I have refactored them both into a separate utility function called `stripTags`.

### DIFF
--- a/src/util/isURL.ts
+++ b/src/util/isURL.ts
@@ -1,9 +1,13 @@
 import isEmail from './isEmail'
+import stripTags from './stripTags'
 
 const REGEX_URL =
   /^(?:http(s)?:\/\/)?(www\.)?[a-zA-Z@:%_\\+~#=]+[-\w@:%_\\+~#=.]*[\w@:%_\\+~#=]+[.:][\w()]{2,6}((\/[\w-()@:%_\\+~#?&=.]*)*)$/i
 
 /** Checks if a string is a URL. */
-const isURL = (s: string) => REGEX_URL.test(s) && !isEmail(s)
+const isURL = (s: string) => {
+  const strippedValue = stripTags(s)
+  return REGEX_URL.test(strippedValue) && !isEmail(strippedValue)
+}
 
 export default isURL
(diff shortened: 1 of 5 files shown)


src/util/isURL.ts
```diff
@@ -1,9 +1,13 @@
...
 
 const REGEX_URL =
   /^(?:http(s)?:\/\/)?(www\.)?[a-zA-Z@:%_\\+~#=]+[-\w@:%_\\+~#=.]*[\w@:%_\\+~#=]+[.:][\w()]{2,6}((\/[\w-()@:%_\\+~#?&=.]*)*)$/i
 
 /** Checks if a string is a URL. */
-const isURL = (s: string) => REGEX_URL.test(s) && !isEmail(s)
+const isURL = (s: string) => {
+  const strippedValue = stripTags(s)
+  return REGEX_URL.test(strippedValue) && !isEmail(strippedValue)
+}
```
trevinhofmann, line 11: I'm a little skeptical of `isURL` automatically stripping out the tags, rather than stripping the tags before passing them to `isURL`. If I were to import and use `isURL` elsewhere, I would expect it to only check if the string is a URL, and not necessarily also return `true` if it's a URL enclosed in tags.

However, at the moment, it looks like this utility is only used in places where it would make sense to do this. I don't think this is a blocking issue, but I just wanted to point it out.

trevinhofmann (approved):
> I also noticed the same tag stripping regex was being used in the bulletColor action, so I have refactored them both into a separate utility function called stripTags.

Thank you for this! Great work.

I added a couple of comments worth considering, but I don't see any blocking issues to prevent this from being merged as-is.

Screenshot 2024-10-23 at 1 12 23 AM

raineorshine (approved):

PR #2491, opened 2024-10-23
Remove Puppeteer Test Delays Part 1

mstrofbass:
This PR removes the bulk of the `sleep` calls from the puppeteer e2e tests resulting in a clean run of the e2e tests taking only ~38 seconds instead of ~60 on my machine and gets us pretty close to completing #2110.

These removals were the ones that were "easy" (i.e., didn't require any additional work beyond removing the `sleep` call itself). I suspect that it is highly likely we will get intermittent failures but they are rare enough that I cannot tell how often they will appear; it's entirely possible that they will be environment specific (e.g., they may appear more often or less often depending on the speed of the machine). 

IMO we've cut the time down by 1/3 so even if we have to rerun a test or two occasionally it's unlikely to negate the gains in this PR.

### DIFF
--- a/src/e2e/puppeteer/__tests__/ContextView.ts
+++ b/src/e2e/puppeteer/__tests__/ContextView.ts
@@ -29,19 +29,10 @@ it('load buffered ancestors of contexts when context view is activated', async (
 
   await waitForEditable('m')
 
-  // wait for a re-render in case the lexeme was loaded after the parent
-  // getEditingText will return undefined if we don't wait
-  // we don't currently have a way to tell if a lexeme is missing or just loading
-  await sleep(100)
-
   await clickThought('m')
   await scrollIntoView('[data-testid="toolbar-icon"][aria-label="Context View"]')
   await scrollBy('#toolbar', 50, 0)
   await click('[data-testid="toolbar-icon"][aria-label="Context View"]')
-
-  // allow ancestors to be loaded
-  // may not be practically necessary, but there could be a delay on slower machines
-  await sleep(10)
 
   // assert that c is loaded
   await waitForEditable('c')
--- a/src/e2e/puppeteer/__tests__/ContextView.ts
+++ b/src/e2e/puppeteer/__tests__/ContextView.ts
@@ -1,4 +1,3 @@
-import sleep from '../../../util/sleep'
 import click from '../helpers/click'
 import clickThought from '../helpers/clickThought'
 import paste from '../helpers/paste'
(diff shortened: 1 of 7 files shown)


mstrofbass:
The sleep in the `escape-html.ts` sleep appears to be causing a problem more frequently than I thought. As far as I can tell, this sleep is necessary for the clipboard to register the HTML. 

I tried to see if I could resolve this real quick but will need more time. This whole test takes less than 10 seconds so the reversion doesn't add time back to the tests.

mstrofbass:
`escape-html.ts` may have been an easy fix, actually...the `navigator.clipboard.write` methods are, apparently, async and were not being awaited. I'm not familiar with Puppeteer's `evaluate` method so I'm not 100% sure this is actually a complete solution but it appears to have worked.

https://github.com/cybersemics/em/pull/2491/commits/fba71ea097e0af19c1fb5b1f374d0299796d99a7

src/e2e/puppeteer/__tests__/ContextView.ts
```diff
@@ -29,20 +28,11 @@ it('load buffered ancestors of contexts when context view is activated', async (
 
   await waitForEditable('m')
 
-  // wait for a re-render in case the lexeme was loaded after the parent
-  // getEditingText will return undefined if we don't wait
-  // we don't currently have a way to tell if a lexeme is missing or just loading
-  await sleep(100)
```
raineorshine, line 35: Deleting code is my favorite... 🤤

raineorshine (approved):
> This PR removes the bulk of the `sleep` calls from the puppeteer e2e tests resulting in a clean run of the e2e tests taking only ~38 seconds instead of ~60 on my machine and gets us pretty close to completing #2110.

That's a great improvement! Good to get some significant gains now before our tests inevitably grow in number.

> These removals were the ones that were "easy" (i.e., didn't require any additional work beyond removing the `sleep` call itself). I suspect that it is highly likely we will get intermittent failures but they are rare enough that I cannot tell how often they will appear; it's entirely possible that they will be environment specific (e.g., they may appear more often or less often depending on the speed of the machine).

Understood. We can keep an eye out for failures in case we need to selectively add minimal sleeps back in, or find a way to wait for the expected state.

> `escape-html.ts` may have been an easy fix, actually...the `navigator.clipboard.write` methods are, apparently, async and were not being awaited. I'm not familiar with Puppeteer's `evaluate` method so I'm not 100% sure awaiting them is actually a complete solution but it appears to have worked well enough that I haven't gotten an error during the tests yet.

We pass an async function to `evaluate` in the no longer used [waitForState](https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/helpers/waitForState.ts), so it should work! Good catch.

Issue #2480, opened 2024-10-20
Sort does not work on fragmentary bibliography


raineorshine:
I found a case that is not sorted properly.

There are [many rules](https://github.com/cybersemics/em/blob/main/src/util/compareThought.ts) when it comes to sorting. First identify the cause. I am guessing they are falsely identified as dates. If so, propose a narrower rule for date detection that excludes this case without excluding the others. 

There is good test coverage for sorting so you don't have to worry about regressions as long as the tests pass.

## Steps to Reproduce

```
- Caracciolo, Peter. And T.s. Eliot. 1989.
- Bautze-Picron C. 2008.  Image. Oxford: Archaeopress.
- Ducœur, G. (2019) religions. [Online] 236 (1), 11–40.
```

Activate Sort Ascending.

## Current Behavior

Not sorted:

```
- Caracciolo, Peter. And T.s. Eliot. 1989.
- Bautze-Picron C. 2008.  Image. Oxford: Archaeopress.
- Ducœur, G. (2019) religions. [Online] 236 (1), 11–40.
```

## Expected Behavior

Should be sorted alphabetically:

```
- Bautze-Picron C. 2008.  Image. Oxford: Archaeopress.
- Caracciolo, Peter. And T.s. Eliot. 1989.
- Ducœur, G. (2019) religions. [Online] 236 (1), 11–40.
```

raineorshine:
Great, that makes sense. 

I'm not sure how changing the priority of compareDateAndOther fixes the problem. It seems like we should use a Regex to narrow the strings that are parsed as dates.

PR #2495, opened 2024-10-24
Fix sort logic which doesn't work on fragmentary bibliography

RED-ROSE515:
Issue : Sort does not work on fragmentary bibliography #2480
Fix the order of `compareDateAndOther` function.


raineorshine:
Thanks! Please add the test case to the existing unit tests. 

It would also be helpful to explain why this works, as it's not obvious (I still don't quite understand it).

raineorshine (approved):
Thanks!

PR #2497, opened 2024-10-26
Add toolbarPointerEvents recipe for typesafe pointer-events override

raineorshine:
# Problem

In cases where `pointer-events: none` needs to be overridden, there is an implicit relationship between the base and the override. For example, Toolbar needs `pointer-events: none` and ToolbarButton needs to override it with `pointer-events: auto` (in order to avoid blocking clicks in the editor when a dropdown menu is open). 

This may produce the correct behavior for the user, but the relationship between the base and the override are not enforced by the type system if implemented in a simple `css` call. This makes it easy for one value to be inadvertently changed in the future, when in fact they are tightly coupled.

# Solution

A recipe is created to apply `pointer-events: none` and override it with `pointer-events: auto` as needed. This creates an explicit link between their usage and provides a traceable type for developers to understand the impact of changing either one.

# Alternatives

The same effect could be accomplished with a pair of Javascript constants without the use of Panda.

### DIFF
--- a/panda.config.ts
+++ b/panda.config.ts
@@ -17,6 +17,7 @@ import modalTextRecipe from './src/recipes/modalText'
 import multilineRecipe from './src/recipes/multiline'
 import textNoteRecipe from './src/recipes/textNote'
 import thoughtRecipe from './src/recipes/thought'
+import toolbarPointerEventsRecipe from './src/recipes/toolbarPointerEvents'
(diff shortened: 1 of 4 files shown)


raineorshine:
@yangchristina Interested to get your feedback on this pattern. It's a bit complex for what it is, but it's the most typesafe way I could think to tightly couple the base with the override.

Issue #2467, opened 2024-10-19
Strip external colors


raineorshine:
## Steps to Reproduce

Copy "19th–early 20th century" from https://www.brooklynmuseum.org/opencollection/objects/124722 and paste into **em**.

## Current Behavior

Text is black on white background.

## Expected Behavior

Text should be the default white on black with no styles applied.

- Color should be stripped from thoughts imported from external sources.
- Color should be preserved from thoughts imported from **em** (e.g. copying and pasting a thought into a different context).

## Notes

How will you differentiate internal from external imports? Should we do a simple regex on the format that **em** uses, or should we add a `data-source='em'` attribute?

raineorshine:
The problem is a bit broader than that. In the vast majority of cases, users only want to copy the content from other websites, not the style. Colors from the source document probably only look good on the original website. This has happened to many times when using **em** myself. Every time it copies colors, I wish it would have just copied the text.

For this reason, the default behavior should be to strip all colors from text that is pasted from external sources. If text with colors is copied and pasted within **em**, then we should preserve the colors, since we know that they were intentionally added by the user.

In the future we could potentially add the option to opt-in to copying colors, if there is demand.

raineorshine:
Great :)

The main question I have right now is how we should detect internal vs external text in the clipboard. Let me know what you think would be a good approach.

PR #2499, opened 2024-10-27
fix Strip colors internal

RED-ROSE515:
Fix Strip external colors #2467

Use clipboardData.setData('text/em') to check if the text is internal or external.


raineorshine:
Please take a look at the failing test in the CI. Once that is fixed I can review.

src/components/ContentEditable.tsx
```diff
@@ -67,9 +68,34 @@ const ContentEditable = React.memo(({ style, html, disabled, innerRef, ...props
     props.onChange(event)
   }
 
+  // eslint-disable-next-line jsdoc/require-jsdoc
+  const handleCopy = (event: React.ClipboardEvent) => {
+    const currentText = selection.text()
+    const currentHtml = selection.html()
+    const clipboardData = event.clipboardData
+    clipboardData.setData('text/plain', currentText!)
+    clipboardData.setData('text/em', currentText!)
```
raineorshine, line 77: Let's just set `text/em` to true rather than duplicate the copied text.

src/components/ContentEditable.tsx
```diff
@@ -67,9 +68,34 @@ const ContentEditable = React.memo(({ style, html, disabled, innerRef, ...props
     props.onChange(event)
   }
 
+  // eslint-disable-next-line jsdoc/require-jsdoc
```
raineorshine, line 71: Please avoid disabling the linter. The function is not as self-evident as it seems; you can explain that it sets `text/em` so that the clipboard data can be later identified as coming from **em**.

src/device/selection.ts
```diff
@@ -375,6 +375,21 @@ export const split = (el: HTMLElement): SplitResult | null => {
...
 
 /** Returns the selection text, or null if there is no selection. */
 export const text = () => window.getSelection()?.toString() ?? null
+
+/** Returns the selection html, or null if there is no selection. */
+export const html = () => {
+  const selection = document.getSelection()
+  const range = selection?.getRangeAt(0)
+  const div = document.createElement('div')
+  div.appendChild(range!.cloneContents())
```
raineorshine, line 392: Return `null` if range is undefined. `!` is usually used when there are other runtime conditions that guarantee the value is truthy, and should be used sparingly.

raineorshine (changes requested):
There is one puppeteer test that is failing in the CI related to paste.

raineorshine (changes requested):

src/components/Editable/useOnPaste.ts
```diff
@@ -29,6 +29,8 @@ const useOnPaste = ({
       // mobile Safari copies URLs as 'text/uri-list' when the share button is used
       const plainText = e.clipboardData.getData('text/plain') || e.clipboardData.getData('text/uri-list')
       const htmlText = e.clipboardData.getData('text/html')
+      // if the browser is controlled by automation, or the clipboard comes from em, set true.
```
raineorshine, line 32: This deserves a bit more explanation. *Why* do we need to check for browser automation? **Because puppeteer does not allow setData with other MIME types**. That's the key piece of information.

src/components/Editable/useOnPaste.ts
```diff
@@ -65,10 +67,10 @@ const useOnPaste = ({
         }
 
         const text = htmlText
-          ? // Clean HTML from clipboard
-            strip(htmlText, { preserveFormatting: true })
-          : // Escape plain text from clipboard
-            escapeHtml(plainText.trim())
+          ? emText
+            ? strip(htmlText, { preserveFormatting: true })
+            : strip(htmlText) // Clean HTML from clipboard
```
raineorshine, line 72: Simpler: `strip(htmlText, { preserveFormatting: emText }`

src/components/ContentEditable.tsx
```diff
@@ -67,9 +68,34 @@ const ContentEditable = React.memo(({ style, html, disabled, innerRef, ...props
...
+    const currentText = selection.text()
+    const currentHtml = selection.html()
+    const clipboardData = event.clipboardData
+    clipboardData.setData('text/plain', currentText!)
+    clipboardData.setData('text/em', 'true')
+    clipboardData.setData('text/html', currentHtml!)
+    event.preventDefault()
+  }
+
+  /** Copies the selection text, sets a text/em flag in the clipboard data to detect the source on paste and removes the current selection. */
```
raineorshine, line 82: `Cuts`

raineorshine (changes requested):
Tested and works well so far.

There are two additional cases that need to be accounted for.

# 1. External bold text should copy and paste

## Steps to Reproduce

1. Copy some **bold** text from a different app.
2. Paste into **em**.

## Current Behavior

Bold is removed.

## Expected Behavior

Bold/italic/strikethrough should be preserved. Colors and other HTML should be removed.

You will probably need to modify `strip` support this other option.

# 2. Copy Cursor command

The `copyCursor` shortcut programmatically copies the cursor thought to the clipboard. We should also set the `text/em` flag in this case to allow pasting colors.

See:

- https://github.com/cybersemics/em/blob/main/src/shortcuts/copyCursor.ts
- https://github.com/cybersemics/em/blob/main/src/shortcuts/__tests__/copyCursor.ts

## Steps to Reproduce

1. Set the cursor on a thought .
3. Enter <kbd>Control/⌘</kbd> + <kbd>C</kbd> to copy the entire thought.
4. Paste into another thought.

## Current Behavior

Plain text is pasted.

## Expected Behavior

HTML with formatting and colors should be pasted.

src/actions/__tests__/importText.ts
```diff
@@ -1114,7 +1114,7 @@ it('import single line with style attributes', () => {
   expect(exported).toBe(`<ul>
   <li>__ROOT__${EMPTY_SPACE}
     <ul>
-      <li><span style="color: rgb(255, 255, 255);font-weight: bold;background-color: rgb(0, 0, 0);">Atonement</span></li>
+      <li><span style="font-weight: bold;background-color: rgb(0, 0, 0);">Atonement</span></li>
```
raineorshine, line 1117: Shouldn't this also strip the background color if this is an external source?

src/components/ContentEditable.tsx
```diff
@@ -1,5 +1,10 @@
 import React, { useEffect, useRef } from 'react'
 import { useSelector } from 'react-redux'
+import * as selection from '../device/selection'
+import exportContext from '../selectors/exportContext'
+import simplifyPath from '../selectors/simplifyPath'
+import store from '../stores/app'
```
raineorshine, line 6: We generally want to avoid importing the global store. Instead, stick to `useSelector` so that we can respect the normal flow of data in a React-Redux app. Otherwise it can lead to stale state.

src/components/ContentEditable.tsx
```diff
@@ -67,9 +72,44 @@ const ContentEditable = React.memo(({ style, html, disabled, innerRef, ...props
     props.onChange(event)
   }
 
+  /** Copies the selection text and sets a text/em flag in the clipboard data to detect the source is 'em' on paste. */
+  const handleCopy = (event: React.ClipboardEvent) => {
+    event.preventDefault()
+
+    const state = store.getState()
+    const simplePath = simplifyPath(state, state.cursor!)
```
raineorshine, line 80: ContentEditable is supposed to be a very low-level component, not aware of the store or thought structure. The only state it has access to is `state.editableNonce` and that is clearly an escape hatch. 

Additionally, putting this logic here adds the hidden invariant that this ContentEditable is within the cursor thought. That may be circumstantially true, but is probably not something we want to rely on. We would like to avoid implicit invariants sneaking in.

Try taking custom `onCopy` and `onCut` handlers like `onPaste`. Then this logic can move up to the `Editable` component which has access to the Redux state and is aware of thought structure. You could also add a custom `useOnCopy`/`useOnCut` hook like the existing `useOnPaste`. Both of these approaches would result in better separation of concerns.

src/shortcuts.ts
```diff
@@ -282,7 +282,7 @@ export const inputHandlers = (store: Store<State, any>) => ({
     shortcutEmitter.trigger('shortcut', shortcut)
 
     if (!shortcut.canExecute || shortcut.canExecute(store.getState())) {
-      e.preventDefault()
+      if (shortcut.id !== 'copyCursor') e.preventDefault()
```
raineorshine, line 285: This is okay as a temporary measure, but we'll have to think about how to expand the Shortcut structure to accommodate this in a more general way. Having shortcut.ts have to know about specific shortcut behavior is not a good idea. We want shortcut-specific behavior to be encapsulated within the appropriate shortcut file.

raineorshine (changes requested):
The overall approach looks good, and testing the behavior it works much better now! Great to be able to copy formatting and color internally now.

I have left a few comments for requested change, but left other parts of the code that looked in progress.

There are a couple places in the UI where formatting tags were supposed to be stripped out, and they are not anymore. I added unit test coverage that didn't exist, so if you rebase on upstream `main` you will see the failing tests and consequently the steps to reproduce.

- Alert: c0cc20e45c
- NavBar: 5206391bf6

As for the tests that are still failing, let me know if you need any guidance. I realize that some of the tests will need to be updated since the behavior is changing. I will evaluate changed tests closely to ensure there are no regressions in behavior that we actually want to keep.

Thank you!

raineorshine (changes requested):
Works great!

Just a few code quality questions.

src/components/Editable/useOnCut.ts
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import * as selection from '../../device/selection'
+
+/** Returns an onCopy handler that cuts the selection text, sets a text/em flag in the clipboard data to detect the source on paste and removes the current selection. */
+const useOnCut = (event: React.ClipboardEvent) => {
```
raineorshine: As a hook, this should return a callback that takes the ClipboardEvent. It's not really a hook if it is used directly as the event handler itself.

src/components/ContentEditable.tsx
```diff
@@ -70,6 +71,8 @@ const ContentEditable = React.memo(({ style, html, disabled, innerRef, ...props
   return (
     <div
       {...props}
```
raineorshine: `onCopy` and `onCut` are both already added via `{...props}`.

raineorshine (approved):

raineorshine:
Thanks, good to go!

Note the refactor I did to address my last comments: 7596523168

PR #2515, opened 2024-10-31
Remove puppeteer test delays #2110

### LINKED ISSUE #2110: Remove puppeteer test delays

mstrofbass:
This PR:

* Updates the empty thoughtspace text to use a configurable duration
* Updates the hamburger menu to use a configurable duration
* Removes the `sleep` from the render thoughts tests
* Removes the `sleep` from the table-view tests
* Restricts the use of `sleep` in the drag and drop helper to only get used on drops

Closes #2110


trevinhofmann (changes requested):
Other than the one question about `src/e2e/puppeteer/helpers/computedStyles.ts`, this change looks fine to me. Thanks, @mstrofbass!

@raineorshine, please note that there are still two sleeps remaining, explained here: https://github.com/cybersemics/em/issues/2110#issuecomment-2450742719

> `cursor.ts` - This appears to be needed for waiting until the cursor position is saved to persistent state prior to refreshing. There's probably a way to verify the state before refreshing but it's going to require some additional investigation into the architecture that I think is better done later.

> `dragAndDrapThought.ts` - This sleep is needed so the notes have time to update after dropping. I made it so the sleep call only fires on drop, which means it only applies to a subset of the tests. Obv we can optimize this and probably get rid of the sleep altogether but, at this point, we're going to get more speed increases by splitting the drag and drop tests into two separate files that run in parallel.

mstrofbass:
Yes and no; I was using on that branch when I was debugging the transitions and was using that to assist in the debugging. Forgot to remove it.

raineorshine (approved):

trevinhofmann (approved):

raineorshine:
Two drag-and-drop tests failed after the merge: https://github.com/cybersemics/em/actions/runs/11626764992/job/32378961668

PR #2516, opened 2024-11-01
Capacitor build script fixes

fbmcipher:
There were some issues with the Capacitor config file (`capacitor.config.ts`) and the build scripts for Capacitor in `package.json` that were stopping me from getting a working build of Capacitor on Android. Opening the app would just yield a black screen.

This was happening because of a line in the Capacitor config file that says to use a development server defined in `.env` instead of loading from local built assets (if `NODE_ENV=development`)

I have resolved this with some changes in `capacitor.config.ts` and some new scripts in `package.json`:

* yarn cap:android:dev – dev build that uses live server
* yarn cap:android:static – dev build that uses static assets
* yarn cap:android – dev build that uses live server (default)

And all the same for iOS.

(As a bonus – this also means we have proper build scripts ready for CI/CD!)


raineorshine (changes requested):
Thank you!

Unfortunately, I now see the blank screen on iOS :). When running `yarn cap:ios` on `main` the Xcode console says `capacitor://localhost` (correct) and on this branch it says `http://192.168.1.8:3000/` (incorrect, at least when in development mode).

Aside from that, maybe you can explain more about BUILD_MODE. Our original goal for configuring the build environment was to have everything branch off `NODE_ENV`, which can be `development` (default), `production`, or `test`. 

It's been a while since another developer set up Capacitor on this project, but it looks like the intention is to serve the app from the vite dev server in `development` mode and serve the app from the internal Capacitor sever in `production` mode. I suspect the Android issue is actually a problem connecting to `localhost` rather than a misconfiguration of the environment per se.

fbmcipher:
Apologies for the regression! Let me explain the rationale behind these changes. I'll work with a fresh clone of both `main` and `capacitor-fixes` to ensure no environment settings cause inconsistent behaviour.

## Evaluating the behaviour in `main`

`yarn cap:ios` on `main` does two things:
* `yarn cap:sync` – this copies built web assets from `build` directory to the iOS project.
* `cap open ios` – opens the project in Xcode to build.

When you Build and Run the project in Xcode, the console shows the internal Capacitor server `capacitor://localhost`, which loads the built web assets – and the app works as intended. **So `yarn cap:ios` is copying the web build artifacts, and then building a static version of the app** – not load from vite!

`yarn cap:android` on `main` does the same two things:
* `yarn cap:sync` – copies built web assets from `build` directory to the Android project
* `cap open android` – opens the project in AndroidStudio to build

However, when you build and run this project on Android, you'll see this errors in the console:
* `E: Provided server url is invalid: no protocol: `
* `W: Loading app at null`

That the Capacitor URL is being logged as `null` indicated to me that the server URL wasn't being properly set. This makes sense, as there is a line in `capacitor.config.ts` that redefines the server URL based on the `NODE_ENV`:

```
const serverConfig =
  nodeEnv === 'development'
    ? {
        server: {
          url: process.env.CAPACITOR_SERVER_URL,
          cleartext: true,
        },
      }
    : {}
```

Of course, in a fresh clone, `CAPACITOR_SERVER_URL` is blank. So that causes the null server URL, and causes the blank screen to show on Android. This indicates to me that the problem is at least in part a configuration issue. Even if I set CAPACITOR_SERVER_URL in `.env.development.local`, my changes would not cascade correctly – which was the rationale behind ab97bbbe92e6de16064c5266b887f0b907f1acf8.

(I have no idea why this only happens on Android and not iOS, as really the behaviour should be identical across both platforms!).

**Revisiting the initial intention:** There's an ambiguity. `yarn cap:android` and `yarn cap:ios` in `main` are set up to generate _static builds_ (given the fact that they sync the built web assets to the native project, and that the iOS build points to the internal Capacitor server `capacitor://localhost` which correctly loads the built web assets).

**But** the code in `capacitor.config.ts`, and the behaviour of the Android build, tell us that when we make a development build we should be loading from a vite dev server (as why else would we be redefining `CAPACITOR_CONFIG_URL` from its default?)

So my conclusions were that either that:
* **The iOS build is improperly configured** (because it should be loading from vite), _and/or_
* **The Android build is improperly configured** (because it should be loading from in-built assets as the iOS build does)

Either way, we don't have the ability to build *both* a static dev build *and* a live dev build – and either way there does seem to be a misconfiguration somewhere.

## `BUILD_MODE`
So we have two kinds of development build:

* `static` – containing built web assets from cap:sync. This is the current behaviour of `yarn cap:ios` in `main`.
* `server` – running off of the vite dev server. As you say, this does seem to be the intention given the explicit code in `capacitor.config.ts`.

That's why I introduced `BUILD_MODE` and the new build scripts – to resolve the ambiguity and ensure dev builds work as explicitly intended on both iOS and Android.

In `capacitor-fixes` `yarn cap:android` and `yarn cap:ios` are aliased to `yarn cap:android:dev` and `yarn cap:ios:dev` – so a dev build running of the vite server is the default. The user can opt to `yarn cap:android:static` if they want a static dev build for testing purposes.

**If you wanted to work solely off of `NODE_ENV`, we could work it out as follows:**
* `development` builds run off of the Vite dev server => `yarn cap:ios`
* `test` builds run off of static web assets in development/debug mode => `yarn cap:ios:test`
* `production` builds run off of production web assets => `yarn cap:ios:prod`

## Testing `capacitor-fixes` on a fresh clone:
* `yarn cap:ios:dev`: Loads app from Vite dev server at `localhost:3000` (the new default `CAPACITOR_SERVER_URL`), and works as intended. It's interesting that you see `http://192.168.1.8:3000`. I guess this is set in your env file? It should work if you have a vite server running there.
* `yarn cap:ios:static`: Generates a static build using built web assets (no vite). **This is the current behaviour of cap:ios in `main`.** Works as intended, loads from `capacitor://localhost`.
* `yarn cap:android:dev`: Loads from Vite dev server. I did need to reconfigure `CAPACITOR_SERVER_URL` with an absolute IP address, but then it works as intended.
* `yarn cap:android:static`: Generates a static build using built web assets (no vite). Works as intended.

---

I hope this explains the situation in more detail, and I also hope I understood your concerns correctly! Please let me know how you want to proceed.

raineorshine:
> When you Build and Run the project in Xcode, the console shows the internal Capacitor server `capacitor://localhost`, which loads the built web assets – and the app works as intended. **So `yarn cap:ios` is copying the web build artifacts, and then building a static version of the app** – not load from vite!

Ah, I see! I misunderstood the localhost part.

> Of course, in a fresh clone, `CAPACITOR_SERVER_URL` is blank. So that causes the null server URL, and causes the blank screen to show on Android. This indicates to me that the problem is at least in part a configuration issue. Even if I set CAPACITOR_SERVER_URL in `.env.development.local`, my changes would not cascade correctly – which was the rationale behind [ab97bbb](https://github.com/cybersemics/em/commit/ab97bbbe92e6de16064c5266b887f0b907f1acf8).

Yes, that makes sense.

> * `static` – containing built web assets from cap:sync. This is the current behaviour of `yarn cap:ios` in `main`.
> * `server` – running off of the vite dev server. As you say, this does seem to be the intention given the explicit code in `capacitor.config.ts`.

The static assets that are generated with `yarn build` and added to `/build` is the production build. The live dev server is what I refer to as development. So I think of it like this:

`static` → `production
`server` → `development`

I don't think we need to worry about `test` at the moment, since we don't ever check `NODE_ENV === 'test'` and we don't have any tests that run against the capacitor builds.

Do you think this would work? Or do you think static/server + production/development are orthogonal? I'm confused about what a dev build of static assets would mean... 🤔

> * `yarn cap:ios:dev` (now aliased to `yarn cap:ios`): Loads app from Vite dev server at `localhost:3000` (the new default `CAPACITOR_SERVER_URL`), and works as intended. It's interesting that you see `http://192.168.1.8:3000`. I guess this is set in your env file? It should work if you have a vite server running there.

This does not work on my side, and `192.168.1.8` is not in any of my env files. I just get a blank screen.

> * `yarn cap:ios:static`: Generates a static build using built web assets (no vite). **This is the current behaviour of cap:ios in `main`.** Works as intended, loads from `capacitor://localhost`.

This works for me! I hadn't tried it before.

Are you able to run the Capacitor app against the live dev server? If so, it must be an issue on my computer and I can work on that separately. I'll be interested to see if it works for the developer that's coming in to work on the Capacitor build.

fbmcipher:
> Do you think this would work? Or do you think static/server + production/development are orthogonal? I'm confused about what a dev build of static assets would mean... 🤔

Okay, I think I can see the confusion!

Here's how the Capacitor starter project does it by default.

```
    "build:native": "ionic build && npx cap copy",
    "build:native:prod": "ionic build --prod && npx cap copy",
    "install:android": "npm run build:native && cd android && gradlew installDebug",
    "install:android:prod": "npm run build:native:prod && cd android && gradlew installDebug",
```

In the stock config, there isn't even a live server. You get two builds:
* *development/debug* native app build, with development/unminified web code
* *production* native app build, with production/minified web code

Like you pointed out, our static asset development builds contain a production build from vite + development build of the native app. _This is an oversight_, the builds should be as follows:

* `cap:android:static` – development/unminified web code with debug native app build
* `cap:android:dev` – live server with debug native app build

I do hear what you're saying – the vite server *is* more appropriate for actual development.

The main advantage behind the static development build is that they'd enable development tools (like web inspect) to attach to the app. That could come in if a dev/tester's using the app away from the computer, they run into a bug, and need the ability to connect up to a computer to investigate more deeply. Of course, we couldn't do this with a production build.

I can either:
* update our static dev build to use development/unminified web code
* remove the static dev build if you feel we don't need it

Hope that makes sense 😅 Aiming to get this closed & merged ASAP so you can hand over to your Capacitor dev.

raineorshine:
Thanks for your response.

> The main advantage behind the static development build is that they'd enable development tools (like web inspect) to attach to the app. That could come in if a dev/tester's using the app away from the computer, they run into a bug, and need the ability to connect up to a computer to investigate more deeply. Of course, we couldn't do this with a production build.

My takeaway from this is that the live dev server has different benefits from a static development build, and that these are mutually exclusive:

- Live dev server - Hot reload allows developer to see UI changes immediately when developing on local machine.
- Static dev build - Allows development tools (like web inspect) to attach to the app and view the console, set breakpoints, etc.

Is that accurate?

I'm a little out of my element here, and I'm not sure what to do.

@msdewitt Would you mind taking a look at this discussion and offering your opinion since you're our new Capacitor guy? :)

> > Are you able to run the Capacitor app against the live dev server? If so, it must be an issue on my computer and I can work on that separately. I'll be interested to see if it works for the developer that's coming in to work on the Capacitor build.
> 
> Yes – this works for me on a fresh clone if I start the vite server in one tab, then run `cap:ios:dev` in another. Have you tried in a fresh clone?

I'm reluctant to try a fresh clone, as it's a time-consuming process. It's also important to me that I can rely on git for version control and trust that checking out a different version is consistent across platforms. So a fresh clone is not something I typically reach for. But I can try it this time since this appears to only be an issue on my machine.

fbmcipher:
Thanks for the review Trevin :)

After a couple of changes, I can confirm this works from a fully fresh clone on both iOS and Android – both live + static builds. I'm running macOS 15.0.1 for these tests FWIW.

Please find below a full list of the steps I took to get the live server build working.

### Steps
1. clone repo
2. checkout branch `fbmcipher:capacitor-fixes`
3. `yarn install`
4. Edit `CAPACITOR_SERVER_URL` to http://[your-computers-ip]:3000
5. `yarn cap:android:dev` (note – we need to run this command again each time we change CAPACITOR_SERVER_URL)
6. `yarn start`
7. Build project in Android Studio
8. It should work!

`yarn cap:android:static`, `yarn cap:ios:static` and `yarn cap:ios:dev` also work at this point with no further intervention.

### Question
Could you paste the `logcat` when you get the black screen error on Android? It's very strange that you're getting that with the static build. I could _understand_ it with the server build (as it would indicate a connection error) but there should be no trouble loading the static assets.

### Note
> For Android, the app builds (though I needed to change the SDK Version in [variables.gradle](https://github.com/cybersemics/em/blob/main/android/variables.gradle#L3-L4). Likely a Gradle/Android/Java configuration mismatch on my machine) but I also get the white screen for the emulator.

Interestingly, I had to do this too, but didn't have to when I first wrote this PR. Fixed in c9b949756126ecd652bff96000501f2bb0bda493.

trevinhofmann (changes requested):
Thank you very much!

I upgraded to the latest Android Studio and was able to run all 4 on a fresh clone 🎉

* `yarn cap:android:static`
* `yarn cap:android:dev`
* `yarn cap:ios:static`
* `yarn cap:ios:dev`

Considering that they all seem to work fine, I'd be happy to include all of them.

I can see where it would be nice to have the dev/server build for quicker prototyping/testing with the Capacitor app. I was working on an iOS Safari-specific bug earlier today. Being able to use that same dev server if we have a similar Capacitor-specific bug/feature to work on could be a nice time saver.

---

My only request now would be adding some documentation regarding the differences between those 4 to the `README.md` along with any notes for running the Capacitor build (e.g. `CAPACITOR_SERVER_URL`).

trevinhofmann (approved):
Looking great to me! Thank you for the updates to the README.

raineorshine:
Thanks so much and sorry for the delay! I able to get the Capacitor app working with both the live server and static builds.

I'm still trying to understand the difference between the [static development build](https://github.com/fbmcipher/em/blob/capacitor-fixes/README.md#static-development-builds) and [production build](https://github.com/fbmcipher/em/blob/capacitor-fixes/README.md#production-builds).

> It's more like this:
> 
> * Live dev server - Hot reload allows developer to see UI changes immediately when developing on local machine _and allows development tools_.
> * Static dev build - Development build of the app – allows development tools
> * Production build - production build of the app – development tools not allowed.

It's the concept of "development build" that I am not understanding, what exactly would allow/disallow development tools here. `cap:android:static` and `cap:ios:static` both run `yarn build`, which [by definition](https://vite.dev/guide/build) is a production build. Where is there a development build?

> Here's how the Capacitor starter project does it by default.
> 
> ```
>     "build:native": "ionic build && npx cap copy",
>     "build:native:prod": "ionic build --prod && npx cap copy",
>     "install:android": "npm run build:native && cd android && gradlew installDebug",
>     "install:android:prod": "npm run build:native:prod && cd android && gradlew installDebug",
> ```
> 
> In the stock config, there isn't even a live server. You get two builds:
> 
> * _development/debug_ native app build, with development/unminified web code
> * _production_ native app build, with production/minified web code

The difference between the dev build and the production build makes sense in the Capacitor starter project because there are separate `ionic build` and `ionic build --prod` commands. However, it doesn't seem analogous to our setup, since `yarn build` is always a production build.

fbmcipher:
Yes, looking back at this with a fresh pair of eyes, it is a little bit confusing.

In the *static development build*, you're right that the web assets are built in production mode. The development build refers to the *native iOS/Android project being built in debug mode rather than in release mode*.

The debug build has some properties that the release mode does not – of course, I can attach a native code (Java/Swift) debugger to the debug build. But on both iOS and Android, I can *also* inspect the underlying WebView during development via `chrome://inspect` or Safari's remote Web Inspector in debug builds.

When I made this PR, I thought it would be useful to have a **build of em that would operate independently of a server, but still be inspectable in Safari/Chrome if bugs ever came up that needed that extra debugging ability to investigate**.

But ultimately, in development, it hasn't come up. We've used `chrome://inspect` and Safari's Web Inspector *heavily* to debug, but *always* with the live server build.

> The difference between the dev build and the production build makes sense in the Capacitor starter project because there are separate ionic build and ionic build --prod commands. However, it doesn't seem analogous to our setup, since yarn build is always a production build.

Yes, good catch.

I hope this helps explain the difference – ultimately, I think if it's causing more confusion than legitimate solutions we should pull it out. I'm totally fine with having live dev build + production build – especially as you noted the static dev build is just using production code anyway.

raineorshine:
Okay, thanks. Maybe my trouble understanding is just lack of experience. I haven't used `chrome://inspect` before.

Can you clarify what triggers "debug mode" and what triggers "release mode"? It's not clear from the npm scripts where this is happening.

When you say "inspect the underlying WebView", do you mean inspect DOM elements, native functionality, or something else?

It's possible that we just need a different name for it. It doesn't seem right to call it a "development build" when it is running the production `/build` output. Maybe debug mode?

fbmcipher:
> Can you clarify what triggers "debug mode" and what triggers "release mode"? It's not clear from the npm scripts where this is happening.

Yes, AFAIK this actually happens during the `yarn cap:sync` stage, during which the profile for the native project in Xcode/Android Studio gets set depending on NODE_ENV.

> When you say "inspect the underlying WebView", do you mean inspect DOM elements, native functionality, or something else?

I'm referring to the ability to inspect DOM elements inside Capacitor's WebView, which isn't possible in a standard production build.

> It's possible that we just need a different name for it. It doesn't seem right to call it a "development build" when it is running the production /build output. Maybe debug mode?

Perhaps – `debug` does feel a little bit clearer! I am happy with either renaming it to `debug` or removing it entirely.

raineorshine (approved):
Got it, thanks for clarifying!

I've decided to stick with the scripts as they are, with a couple small changes that hopefully clarify the situation:

- Remove `cap:ios:dev` and `cap:android:dev` (i.e. just use `cap:ios` and `cap:ios`). This brings down the number of `yarn cap` scripts to a slightly more manageable 8.
- Describe the development modes as "live" vs "static" and move production copy/sync scripts into the less prominent "Other Scripts". This prioritizes the scripts that developers will most often be interacting with, the development scripts.
- Add a table for reference.

We can change this in the future if needed, but hopefully this retains the flexibility of the different builds but simplifies it conceptually for new developers. Open to feedback.

---

*Excerpt from new README:*

### Modes

There are two development modes that determine where the Capacitor build loads the web resources from. Each mode can be run on iOS or Android.

| Mode   | Description                                          | iOS                   | Android                   |
| ------ | ---------------------------------------------------- | --------------------- | ------------------------- |
| live   | Connect to live dev server at `CAPACITOR_SERVER_URL` | `yarn cap:ios`        | `yarn cap:android`        |
| static | Build production app and output to `/build`          | `yarn cap:ios:static` | `yarn cap:android:static` |

#### Live (default)

Runs off the local Vite development server. Ideal for rapid prototyping, debugging, and hot-reloading your changes.

1. Set the development server URL
   Update your local development environment file (`.env.development.local`) by setting the following variable to the URL of your Vite dev server.

   ```sh
   CAPACITOR_SERVER_URL=http://192.168.x.x:3000
   ```

   **Note**: Use your machine's local network IP (e.g., http://192.168.x.x:3000) rather than localhost (http://localhost:3000).

2. Start the Vite development server

   ```sh
   yarn start
   ```

3. Synchronise the native project and open it in the IDE

   ```sh
   yarn cap:ios       # runs cap:sync and opens iOS project in Xcode
   yarn cap:android   # runs cap:sync and opens Android project in Android Studio
   ```

   **Notes**:

   - Make sure to run these commands again any time you change `CAPACITOR_SERVER_URL`.
   - If you see a blank screen when the app starts, then either `CAPACITOR_SERVER_URL` is wrong or you need to disable your VPN/firewall.

#### Static

Outputs the production build to `/build` and syncs capacitor to the static build output. This will test the production build output, but allow inspecting the WebView and attaching a debugger.

```sh
yarn cap:ios:static     # runs cap:sync pointing to static /build output
yarn cap:android:static # runs cap:sync pointing to static /build output
```

fbmcipher:
I'm happy with all of those changes – no notes! Thank you!

PR #2524, opened 2024-11-01
Move modal styles in globalCss out

yangchristina:
#2170


yangchristina:
@trevinhofmann 
I want to move this directly into modal.root, however then it will conflict with another `description` in recipes/modalText
Screenshot 2024-11-01 at 2 49 03 PM
The easiest way I can think of is to change description from a <p> to something else, like a <span> or something.
The second method, is that since `description` is only used twice (and one of those is not in use), in Devices.tsx and Invites.tsx, I can put it inline instead of in the recipe, that way it would override <p> styles.

trevinhofmann:
Hi @yangchristina -- Apologies for the delay in responding. I saw the PR was a draft and didn't notice that I was mentioned in the comment.

I'm not sure that I have a complete understanding of the issue, but my suggestion would like be the second option of adding some inline styles. It is generally safer and preferable to not make changes to the semantic structure of the HTML (e.g. changing a `<p>` into a `<span>`). Please let me know if you disagree, but I would favor focusing on the CSS specificity which might include the use of inline styles if necessary.

yangchristina:
Hi @trevinhofmann, no worries, it's my fault for not following up.

I agree the second might be best, I'm just concerned about code duplication, since it's splitting up the old .description class into two individual instances. Another option may be to import an `cva` instead of a recipe, since we can't use recipes for this, although I'm not sure where I'd put the `cva` (maybe still in the `recipes` folder?)

yangchristina:
Actually I think I was overcomplicating this in my mind, I've just imported `css()` from the recipes folder. Wasn't sure if I should make a new folder since it technically isn't a recipe.

src/recipes/modal.ts
```diff
@@ -40,6 +40,7 @@ const modalRecipe = defineSlotRecipe({
         base: 'all 0.4s ease-out',
         _test: 'none',
       },
+      '& p': { margin: '0 0 1em 0' }, // modalDescriptionClass overrides this
```
trevinhofmann, line 43: I am okay with this, but I'm uncertain about the value of this comment. Is it uncommon to override styles with more specificity? If we are overriding a style, I think it would be more useful to note that in the place that we override it.

src/recipes/modalDescriptionClass.tsx
```diff
@@ -0,0 +1,9 @@
+import { css } from '../../styled-system/css'
+
+// Overrides the default paragraph margin in the modal recipe.
+const modalDescriptionClass = css({
+  fontSize: '18px',
+  marginBottom: '30px',
+})
+
+export default modalDescriptionClass
```
trevinhofmann, line 9: Can this be renamed to `modalDescriptionClass.ts`? The `.tsx` extension should only be required for files that use the JSX syntax.

trevinhofmann (changes requested):
> Actually I think I was overcomplicating this in my mind, I've just imported css() from the recipes folder. Wasn't sure if I should make a new folder since it technically isn't a recipe.

For consistency, I am in favor of creating a new directory for reused `css( ... )` classes such as this, and reserving the `recipes` directory exclusively for recipes. Do you expect that there will be more files like `modalDescriptionClass.tsx`?

---

I noticed one regression when testing locally. This PR seems to have shifted the toolbar to the far left of the page:

## `move-modal-styles-out-of-globalcss`:

Screenshot 2024-11-09 at 3 12 55 AM

## `main`:

Screenshot 2024-11-09 at 3 11 48 AM

src/recipes/modal.ts
yangchristina: Yes, that is true, thanks for bringing it up!

yangchristina:
@trevinhofmann I'm getting this regression on main actually, so I don't think it's related to this PR.

raineorshine:
Thanks, that one is my bad! Regression slipped into `main` in ec3e432e71fd83e8261462dfddc335c01367af7f. Easy fix of restoring `right: 0`.

yangchristina:
> > Actually I think I was overcomplicating this in my mind, I've just imported css() from the recipes folder. Wasn't sure if I should make a new folder since it technically isn't a recipe.
> 
> For consistency, I am in favor of creating a new directory for reused `css( ... )` classes such as this, and reserving the `recipes` directory exclusively for recipes. Do you expect that there will be more files like `modalDescriptionClass.tsx`?
> 

@trevinhofmann sorry I missed this earlier, I can create a new directory in my next PR. I don't really expect to find more cases like `modalDescriptionClass.tsx`, since normally you can just make another recipe, but it definitely could happen.

Issue #1660, opened 2022-09-10
Add tests for multiline caret movement


raineorshine:
Add puppeteer tests to assert the expected multiline cursor/selection movement.

```
- a
  - b
    - c
    - Beautiful antique furnishings fill this quiet, comfortable flat across from the Acropolis museum. AC works great. It is in an heavily touristic area, but the convenience can't be beat. Highly recommended.
    - d
```

- [x] 1. Cursor down from `a` with the caret at the beginning: should move the cursor to `b` with the caret at the beginning.
- [x] 2. Cursor down from `a` with the caret at the end: should move the cursor to `b`, with the caret at the beginning.
- [x] 3. Cursor up from `b` with the caret at the beginning: should move the cursor to `a`, with the caret at the beginning.
- [x] 4. Cursor up from `b` with the caret at the end: should move the cursor to `a`, with the caret at the beginning.
- [x] 5. Cursor down from `c` and then immediately cursor down from  `Beautiful antique...`: the cursor should stay on the same thought and caret should move the cursor down one line. **(test currently failing; can be written and skipped)**
- [x] 6. Cursor down from `Beautiful antique...` with the caret in the middle of the first line: the cursor should stay on the same thought and the caret should move down one line.
- [x] 7. Cursor down from `Beautiful antique...` with the caret in the middle of the last line: should move the cursor to `d` with the caret at the beginning.
- [x] 8. Cursor up from `Beautiful antique...` with the caret in the middle of the first line: should move the cursor to `c` with the caret at the beginning.
- [x] 9. Cursor up from `Beautiful antique...` with the caret in the middle of the last line: the cursor should stay on the same thought and the caret should move up one line.
- [x] 10. Cursor up from `d`: should move the cursor to `Beautiful antique...` with the caret at the beginning of the line.

PR #2543, opened 2024-11-04
Multiline caret tests for start/end of single line

ethan-james:
#1660

### DIFF
--- a/src/e2e/puppeteer/__tests__/caret-multiline.ts
+++ b/src/e2e/puppeteer/__tests__/caret-multiline.ts
@@ -7,6 +7,7 @@ import waitForEditable from '../helpers/waitForEditable'
 vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
 
 describe('all platforms', () => {
+  // test case 1
   it('on cursorDown, the caret should move from the beginning of a cursor to the beginning of the new cursor.', async () => {
     const importText = `
     - a
@@ -30,6 +31,7 @@ describe('all platforms', () => {
     expect(offset).toBe(0)
   })
 
+  // test case 2
   it('on cursorDown, the caret should move from the end of a cursor to the beginning of the new cursor.', async () => {
     const importText = `
   - a
@@ -53,6 +55,7 @@ describe('all platforms', () => {
     expect(offset).toBe(0)
   })
 
+  // test case 3
   it('on cursorUp, the caret should move from the beginning of a cursor to the beginning of the new cursor.', async () => {
     const importText = `
     - a
@@ -76,6 +79,7 @@ describe('all platforms', () => {
     expect(offset).toBe(0)
   })
 
+  // test case 4
   it('on cursorUp, the caret should move from the end of a cursor to the beginning of the new cursor.', async () => {
     const importText = `
   - a
@@ -99,6 +103,7 @@ describe('all platforms', () => {
     expect(offset).toBe(0)
   })
 
+  // test case 5
   it.skip('on 2x cursorDown, the caret should move from the end of a cursor into the 2nd line of the new multi-line cursor.', async () => {
     const importText = `
   - a
@@ -125,6 +130,7 @@ describe('all platforms', () => {
     expect(offset).toBeGreaterThan(70)
   })
 
+  // test case 6
   it('on cursorDown, the caret should move from the beginning of a multi-line cursor into the 2nd line of the same cursor.', async () => {
     const importText = `
   - a
@@ -138,7 +144,7 @@ describe('all platforms', () => {
     const editableNodeHandle = await waitForEditable(
       "Beautiful antique furnishings fill this quiet, comfortable flat across from the Acropolis museum. AC works great. It is in an heavily touristic area, but the convenience can't be beat. Highly recommended.",
     )
-    await click(editableNodeHandle, { edge: 'left' })
+    await click(editableNodeHandle, { offset: 2 })
 
     await press('ArrowDown')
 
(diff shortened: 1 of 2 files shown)


ethan-james:
@raineorshine This is a work in progress, but please feel free to comment if you want the tests to have a different format.

raineorshine:
Looks good so far!

ethan-james:
@raineorshine Now it's ready for a look.

src/e2e/puppeteer/__tests__/caret-multiline.ts
```diff
@@ -0,0 +1,254 @@
...
+  - a
+    - b
+      - c
+      - ${multiLineCursor}
+      - d`
+
+    await paste(importText)
+
+    const editableNodeHandle = await waitForEditable(multiLineCursor)
+    await click(editableNodeHandle, { offset: multiLineCursor.length - 2 })
```
raineorshine, line 168: Nice. I see how since the browser won't line wrap in the middle of the word "recommended", offset `multilineCursor.length - 2` is guaranteed to be on the last line.

src/e2e/puppeteer/__tests__/caret-multiline.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    await press('ArrowUp')
+
+    // the focus must be in the middle of the multi-line cursor after cursor up
+    const textContext = await getSelection().focusNode?.textContent
+    expect(textContext).toBe(multiLineCursor)
+
+    const offset = await getSelection().focusOffset
+    expect(offset).toBeGreaterThan(70)
```
raineorshine, line 227: Similar to the above, this will depend on the browser width.

This corresponds to test case (9).

src/e2e/puppeteer/__tests__/caret-multiline.ts
```diff
@@ -0,0 +1,254 @@
...
+    const textContext = await getSelection().focusNode?.textContent
+    expect(textContext).toBe(
+      "Beautiful antique furnishings fill this quiet, comfortable flat across from the Acropolis museum. AC works great. It is in an heavily touristic area, but the convenience can't be beat. Highly recommended.",
+    )
+
+    const offset = await getSelection().focusOffset
+    expect(offset).toBeGreaterThan(70)
+  })
+
+  it('on cursorDown, the caret should move from the beginning of a multi-line cursor into the 2nd line of the same cursor.', async () => {
```
raineorshine, line 128: I believe this should start "near the beginning"  (say offset `2`) rather than at the beginning (offset `0`)?

I've numbered the test cases in #1660 to make them easier to refer to. This would correspond to (6) I believe.

raineorshine (changes requested):
Looks very good! Thanks for your thoroughness and attention to detail.

Just a couple questions below.

Note: You can add the related issue # to the PR description and GitHub will create a handy link trail for us. (Added.)

src/e2e/puppeteer/__tests__/caret-multiline.ts
```diff
@@ -0,0 +1,254 @@
...
+    await press('ArrowDown')
+
+    // the focus must be in the middle of the multi-line cursor after cursor down
+    const textContext = await getSelection().focusNode?.textContent
+    expect(textContext).toBe(
+      "Beautiful antique furnishings fill this quiet, comfortable flat across from the Acropolis museum. AC works great. It is in an heavily touristic area, but the convenience can't be beat. Highly recommended.",
+    )
+
+    const offset = await getSelection().focusOffset
+    expect(offset).toBeGreaterThan(70)
```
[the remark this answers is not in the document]
ethan-james: I think that setting a fixed width is pretty likely to break down the line, so I will favor option 2 unless it's something where a fixed width is necessary.

ethan-james:
@raineorshine This one is ready for another look, it just has the one flaky drag & drop test. Do you ever have trouble with that, and if so do you have a preferred way to get it to pass?

raineorshine:
Thanks! Will review soon. 

@mstrofbass is working on fixing that flaky test ☺️

src/e2e/puppeteer/__tests__/caret-multiline.ts
```diff
@@ -0,0 +1,266 @@
...
+
+    await press('ArrowDown')
+
+    // the focus must be in the middle of the multi-line cursor after cursor down
+    const textContext = await getSelection().focusNode?.textContent
+    expect(textContext).toBe(multiLineCursor)
+
+    const offset = await getSelection().focusOffset
+    expect(offset).toBeGreaterThan(multiLineCursor.length / 3)
+    expect(offset).toBeLessThan((multiLineCursor.length * 2) / 3)
```
raineorshine, line 158: I think this math is misleading, as it gives the impression that Line 1 corresponds to the first third of the string, Line 2 corresponds to the second third of the string, etc. In actuality that would only be true at a specific screen width, which we have not set. I don't want this test to have an implicit dependency on a puppeteer browser width that we have not explicitly set.

I'd like to suggest something along the lines of `offsetEnd > offsetStart && offsetEnd < multilineCursor.length`. This correctly asserts that the caret has moved to the right, but not to the very end, which by all accounts means it has moved down one line. It does this without making any assumptions about the length of the string or the width of the browser, i.e. where the line wraps.

If I am missing something or you have an alternative suggestions please let me know!

raineorshine (changes requested):

src/e2e/puppeteer/__tests__/caret-multiline.ts
ethan-james: I think you're right. It seemed like a clever solution, but it's still too fiddly.

Issue #1598, opened 2022-06-18
Merge two thoughts without intervening edit does not work


raineorshine:
## Steps to Reproduce

```
- a
- b
- c
```

1. Set the caret at the beginning of `c` and hit Backspace
2. Set the caret at the beginning of `bc` and hit Backspace

## Current Behavior

Nothing happens

## Expected Behavior

`bc` should get merged into `a` to create `abc`

PR #2544, opened 2024-11-04
Cursor offset should not block deleteEmptyThought

ethan-james:
#1598 

Some notes for the future 🚀

* `state.cursorOffset` is not updated by keyboard/mouse events. It is set manually and infrequently, and this seems like a good opportunity to reduce its responsibilities
* [deleteEmptyThoughtOrOutdent](https://github.com/cybersemics/em/blob/main/src/shortcuts/deleteEmptyThoughtOrOutdent.tsx) decides whether the `deleteEmptyThought` action can be dispatched, and it directly checks the browser selection API to see if the caret is at the beginning of the cursor
* storing the `caretOffset` in `ministore` probably makes sense, but we can put off doing it for now because this is not a reactive context, and we would currently only access the store from a single place

### DIFF
--- a/src/actions/__tests__/deleteEmptyThought.ts
+++ b/src/actions/__tests__/deleteEmptyThought.ts
@@ -126,6 +126,24 @@ describe('normal view', () => {
   - ab`)
   })
 
+  it('merge thoughts multiple times', () => {
+    // set the cursor
+    const steps = [
+      newThought('a'),
+      newThought('b'),
+      newThought('c'),
+      setCursor(['c']),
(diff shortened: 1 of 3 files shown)


ethan-james:
@raineorshine OK, I think I've done my due diligence. The `deleteEmptyThought` action is only dispatched from `deleteEmptyThoughtOrOutdent`, which will short-circuit if `canExecuteDeleteEmptyThought` returns `false`, which will happen if `selection.offset() !== 0`

raineorshine (approved):

Issue #2469, opened 2024-10-19
Remove empty <font> element after setting text color to default


raineorshine:
## Steps to Reproduce

1. Create thought `Hello`
2. Set the text color to blue.
3. Set the text color back to the default.

## Current Behavior

```
- <font>Hello</font>
```

## Expected Behavior

```
- Hello
```

## Testing

Add a Puppeteer test that covers this case.

(It has to be a Puppeteer test because the text color functionality is based on `document.execCommand` which does not work in JSDOM.)

PR #2546, opened 2024-11-05
ColorPicker: remove empty <font> element

RED-ROSE515:
Remove empty <font> element after setting text color to default.
Issue : #2469 
Remove the empty <font> and <span> elements in formatSelection.
Also fix this issue : Black background color not stripped from ContentEditable #2548

### DIFF
--- a/src/actions/formatSelection.ts
+++ b/src/actions/formatSelection.ts
@@ -47,6 +47,29 @@ export const formatSelectionActionCreator =
           if (elementColor && isDifferentColor) return true
           return false
         }
+
+        /** Function to collect tag names without significant attributes. */
+        const collectTagsWithoutAttributes = (text: string, pattern: RegExp): string[] =>
+          Array.from(text.matchAll(pattern))
+            // Filter out tags that lack meaningful attributes
+            .filter(([, , attributes]) => {
+              const meaningfulAttributes = ['style=', 'class=', 'color=', 'background-color=']
+              // Return true if attributes are absent or do not contain any meaningful attributes
+              return !attributes || !meaningfulAttributes.some(attr => attributes.includes(attr))
+            })
+            // Map to extract the tag names from matches
+            .map(([, tagName]) => tagName)
+
+        /** Function to create a new text string with specified tags and their content removed. */
+        const removeTags = (text: string, tags: string[]): string =>
+          // Use reduce to accumulate a new string without the unwanted tags
+          tags.reduce((acc, tagName) => {
+            const openingTagPattern = new RegExp(`<${tagName}(\\s[^>]*)?>`, 'gi')
+            const closingTagPattern = new RegExp(`</${tagName}>`, 'gi')
+            // Replace both opening and closing tags with an empty string
+            return acc.replace(openingTagPattern, '').replace(closingTagPattern, '')
+          }, text)
+
         dispatch((dispatch, getState) => {
           const state = getState()
           if (!state.cursor) return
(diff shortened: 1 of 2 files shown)


src/e2e/puppeteer/__tests__/color.ts
```diff
@@ -121,3 +106,23 @@ it('Clear the text color when setting background color', async () => {
   expect(extractStyleProperty(cursorText!)?.backgroundColor).toBe('rgb(170, 128, 255)')
   expect(extractStyleProperty(cursorText!)?.color).toBe('#000000')
 })
+
+it('Empty <font> or <span> element will be removed after setting color to default.', async () => {
```
raineorshine, line 110: Should just say `Empty <span>` since this test doesn't actually test empty `<font>`.

src/e2e/puppeteer/__tests__/color.ts
```diff
@@ -31,10 +33,12 @@ it('Set the text color of the text', async () => {
   await click('[data-testid="toolbar-icon"][aria-label="Text Color"]')
   await click('[aria-label="text color swatches"] [aria-label="blue"]')
 
+  await press('ArrowUp')
+  await press('ArrowDown')
```
raineorshine, line 37: It should not be necessary to blur and refocus. If so, that probably indicates a bug. Can you explain?

src/e2e/puppeteer/__tests__/color.ts
```diff
@@ -2,6 +2,8 @@ import click from '../helpers/click'
 import clickThought from '../helpers/clickThought'
 import getEditingText from '../helpers/getEditingText'
 import paste from '../helpers/paste'
+import press from '../helpers/press'
+// import refresh from '../helpers/refresh'
```
raineorshine, line 6: 👀

src/actions/formatSelection.ts
```diff
@@ -47,24 +47,49 @@ export const formatSelectionActionCreator =
...
+          const tagWithoutStylePattern = /<(span|font)(\s[^>]*)?>/gi
 
           //Replace style attributes based on the conditions
           const newThoughtValue = thought.value.replace(styleAttrPattern, match => {
             if (shouldRemoveStyle(match)) return ''
             return match
           })
-
+          const tagsToRemove = collectTagsWithoutAttributes(newThoughtValue, tagWithoutStylePattern)
+          const newValue = removeTags(newThoughtValue, tagsToRemove)
```
raineorshine, line 88: We need better naming here. `newValue` and `newThoughtValue` are too similar. You should probably rename `newThoughtValue` if there is something newer.

src/actions/formatSelection.ts
```diff
@@ -47,24 +47,49 @@ export const formatSelectionActionCreator =
           if (elementColor && isDifferentColor) return true
           return false
         }
+
+        /** Function to collect tag names without significant attributes. */
+        const collectTagsWithoutAttributes = (text: string, pattern: RegExp): string[] =>
```
raineorshine, line 52: Good use of pure functions in these helpers 👍

src/actions/formatSelection.ts
```diff
@@ -47,24 +47,49 @@ export const formatSelectionActionCreator =
...
+          Array.from(text.matchAll(pattern))
+            // Filter out tags that lack meaningful attributes
+            .filter(([, , attributes]) => {
+              const meaningfulAttributes = ['style=', 'class=', 'color=', 'background-color=']
```
raineorshine, line 56: We already have `ALLOWED_ATTR`, so let's use those. We want to avoid hardcoding values as if they are independent, when they clearly are tightly coupled to other parts of the app that handle parsing, stripping, and formatting.

https://github.com/cybersemics/em/blob/6e4fa5311c6b9e5ddb4baa50e352d33aa58b2f8b/src/constants.ts#L279

raineorshine (changes requested):
The overall behavior looks good, and I can confirm the empty `<font>` is now being stripped.

src/e2e/puppeteer/__tests__/color.ts
RED-ROSE515: I got it. I will split this test to `background color test` and `text color test`.

Issue #2185, opened 2024-07-24
Upgrade @sinonjs/fake-timers


raineorshine:
Upgrade the `@sinonjs/fake-timers` dependency to the latest versoin.

This is a breaking change and requires upgrading syntax/fixing broken tests.

mstrofbass:
We use `@sinonjs/fake-timers` in `testTimer.js` (which is then used throughout the tests) but we use vitest fake timers in `createTestApp`. Is there a reason for this? If not, would it make more sense to try to swap `testTimer.ts` to use vitest fake timers and then get rid of the `@sinonjs/fake-timers` dependency?

raineorshine:
100%! `@sinonjs/fake-timers` is a legacy from the Jest days.

mstrofbass:
This was already completed at some point and we just didn't notice because all the relevant tests are skipped. I believe all the tests are going to be handled as part of the next tranche of issues so we can probably close this one.

I have also learned that Vitest uses the Sinon fake timers under the hood but fails to expose all of the functionality so we might not be able to do a drop in replacement. I'll get a better look when I start touching those tests.

PR #2547, opened 2024-11-05
DRAFT: Upgrade @testing-library/user-event to v14.5.2

### LINKED ISSUE #2183: Upgrade @testing-library/user-event
Upgrade the `@testing-library/user-event` dependency from `^13.5.0` to the latest `14.x`. 

This is a potentially breaking change and may require upgrading syntax/fixing broken tests.

mstrofbass:
This PR updates the `@testing-library/user-event` library to the latest version (^14.5.2). 

The primary breaking change between user-event v13 and v14 is that the calls are now asynchronous. This has two notable impacts that break some of the tests:

1. Action-related function calls (e.g., `click` and `type`) return promises and must be `awaited` or otherwise handled like a promise
2. Fake timers impacts action-related function calls.

There are likely two possible solutions for fake timers: 

1. Advance the timers; or
2. Use `userEvent.setup({ delay: null })`

I opted for the second option primarily because we are planning on upgrading the fake timers and I really don't want to make that more work than necessary at this point. A secondary consideration is that it it can be called once and used for each action function instead of having to advance the timers after each action function call. This is demonstrated in the `Divider.ts` test.

I am definitely open to advancing the fake timers if we think that makes the most sense but would prefer to defer until after the upgrade in #2185 is completed.

Closes #2183


mstrofbass:
This PR assumes that PR #2545 has been merged and can be removed from draft status and merged once #2545 is merged.

raineorshine (approved):
Thanks for the thorough explanation!

> There are likely two possible solutions for fake timers:
> 
> 1. Advance the timers; or
> 2. Use `userEvent.setup({ delay: null })`
> 
> I opted for the second option primarily because we are planning on upgrading the fake timers and I really don't want to make that more work than necessary at this point. A secondary consideration is that it it can be called once and used for each action function instead of having to advance the timers after each action function call. This is demonstrated in the `Divider.ts` test.
> 
> I am definitely open to advancing the fake timers if we think that makes the most sense but would prefer to defer until after the upgrade in #2185 is completed.

That sounds reasonable.

PR #2552, opened 2024-11-05
Focus on isEditing Editable when sidebar closes

### LINKED ISSUE #2062: Selection offset is lost when Sidebar opens
## Steps to Reproduce

1. Create a thought. 
2. Set the caret at the end of the thought. 
3. Open and close the Sidebar. 

## Current Behavior

The caret is moved to the beginning of the thought. 

https://github.com/cybersemics/em/assets/750276/52cbc277-e4b4-4a81-bc33-8b932a783bb6

## Expected Behavior

The caret should be restored to its previous position at the end of the thought.

ethan-james:
#2062 

When clicking on the sidebar, the browser blurs the current cursor and any browser selection will disappear. The selection persists in the cursor where it was, so calling `focus` on a cursor where `isEditing` is `true` will restore the caret along with any selection.


src/e2e/puppeteer/__tests__/toggleSidebar.ts
```diff
@@ -0,0 +1,31 @@
...
+    const importText = `
+    - ${cursor}`
+
+    await paste(importText)
+
+    const editableNodeHandle = await waitForEditable('Hello')
+    await click(editableNodeHandle, { offset: 1 })
+
+    await click('[aria-label=menu]')
+    await page.waitForSelector('[aria-label=sidebar] .fade-enter-done')
```
raineorshine, line 21: `.fade-enter-done` is an animation class that we probably don't want to couple our test to. Can you say more about this?

Relatedly, tests must be only select based on `aria-label` or `data-testid`. This allows us to separate behavior from styles.

raineorshine (changes requested):
Thank you!

There are three ways to close the Sidebar. From my testing this fix works in two of the three:

- [x] Hit Escape
- [x] Click hamburger menu
- [ ] Click on empty space in the editor

As an aside, it's often helpful to start with a description of the cause of the bug. That can help us better understand the problem, help inform the solution, and it's useful from a documentation perspective. In this example, I'm curious if the editing thought is getting re-rendered or useEditMode is triggering again when the Sidebar closes.

ethan-james:
> Thank you!
> 
> There are three ways to close the Sidebar. From my testing this fix works in two of the three:
> 
> * [x]  Hit Escape
> * [x]  Click hamburger menu
> * [ ]  Click on empty space in the editor
> 
> As an aside, it's often helpful to start with a description of the cause of the bug. That can help us better understand the problem, help inform the solution, and it's useful from a documentation perspective. In this example, I'm curious if the editing thought is getting re-rendered or useEditMode is triggering again when the Sidebar closes.

OK, I found a fix for the empty space method. And yes, useEditMode triggers when `showSidebar` changes, and again when `hadSidebar` changes. There appears to be 3 re-renders in each thought (editing or not) and I'm not sure where the 3rd one comes from.

ethan-james:
I think that the 3rd re-render is happening because of `const state = store.getState()` in `Editable.tsx`. I don't think any DOM elements are changing, so it may not be a big deal.

raineorshine:
Yes, I'm not sure why it re-renders 3 times. That should just be an optimization issue and not affect the behavior though, right?

It wouldn't be `const state = store.getState()` because that does not trigger a re-render. Only `useSelector` triggers a re-render. 

(Aside: That must be an old bit of code, as direct access of the store in a component is a bad practice. It should either use `useSelector` or be wrapped in a thunk action.)

ethan-james:
Right, the re-renders should be very quick as long as the number of editable nodes remains manageable.

You're probably right about `getStore()` not triggering a re-render. I was using the [useWhatChanged](https://github.com/simbathesailor/use-what-changed) hook but it may have been picking up on a new state object that was produced during a render that was triggered elsewhere.

src/e2e/puppeteer/__tests__/toggleSidebar.ts
```diff
@@ -0,0 +1,29 @@
+import click from '../helpers/click'
+import getSelection from '../helpers/getSelection'
+import paste from '../helpers/paste'
+import waitForEditable from '../helpers/waitForEditable'
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+describe('all platforms', () => {
```
trevinhofmann, line 8:
```suggestion
describe('toggle sidebar', () => {
```

The `describe` title should more accurately describe the set of test cases.

src/e2e/puppeteer/__tests__/toggleSidebar.ts
```diff
@@ -0,0 +1,29 @@
...
+  it('on toggleSidebar, the caret offset should be preserved once it closes.', async () => {
```
trevinhofmann, line 9: It might be worthwhile to write 3 tests for this, covering the 3 different ways to close the sidebar:
* Hit Escape
* Click hamburger menu
* Click on empty space in the editor

trevinhofmann (changes requested):
Hi, @ethan-james. Thank you for the PR! The first two cases are still working as expected.

The third case, `Click on empty space in the editor`, is still losing the cursor selection within the thought. 

Demo:

https://github.com/user-attachments/assets/87de641b-fe7b-4048-82e8-2ba73528cabd

src/components/Editable/useEditMode.ts
```diff
@@ -136,6 +138,18 @@ const useEditMode = ({
...
 
+  // Resume focus if sidebar was just closed and isEditing is true
+  useEffect(() => {
+    if (showSidebar) {
+      setHadSidebar(true)
+    } else {
+      if (isEditing && hadSidebar) {
+        contentRef.current?.focus()
+      }
+      setHadSidebar(false)
```
trevinhofmann, line 149: Could the two `setHadSidebar` statements be simplified into one?

```
setHadSidebar(showSidebar);
```

---

Additionally, this effect hook gets triggered twice because `hadSidebar` is a state value. It gets updated within the hook, then triggers it a second time because it has updated. Could we change it to `const hadSidebar = useRef(false)` and update it with `hadSidebar.current = showSidebar` at the end of the effect? Then, it can be removed as a dependency and only trigger the effect hook once without breaking the functionality.

---

Lastly, there is another PR proposing a `usePrevious` hook. I don't think refactoring this PR to make use of a `usePrevious` hook is necessary, but it would be nicer (and more legible) to simply have:

```
const showSidebar = useSelector(state => state.showSidebar)
const hadSidebar = usePrevious(showSidebar)
```

src/e2e/puppeteer/__tests__/toggleSidebar.ts
```diff
@@ -0,0 +1,74 @@
...
+    await click('[data-testid=sidebar] [aria-hidden=true]')
+
+    // the focus must be after 'l' after cursor down
+    const textContext = await getSelection().focusNode?.textContent
+    expect(textContext).toBe(cursor)
+
+    const offset = await getSelection().focusOffset
+    expect(offset).toBe(3)
+  })
+})
```
trevinhofmann, line 74: I like the three tests covering each way to close the menu. Nice work!

src/e2e/puppeteer/helpers/waitForTransitionEnd.ts
```diff
@@ -0,0 +1,18 @@
+import { page } from '../setup'
+
+/** Wait for a transition to finish. */
+async function waitForTransitionEnd(selector: string) {
+  await page.evaluate(selector => {
+    return new Promise<void>(resolve => {
+      const transition = document.querySelector(selector)
+      if (!transition) throw new Error(`'${selector}' selector not found`)
+      const onEnd = function () {
```
trevinhofmann, line 9:
```suggestion
      const onEnd = () => {
```

I would prefer to see this as an arrow function for consistency. This project uses arrow functions almost everyone, and actual `function` blocks have some odd quirks/footguns.

trevinhofmann (changes requested):
Thank you! The updated PR now functions as expected for all three ways of closing the sidebar.

I just have a few minor refactoring requests.

src/e2e/puppeteer/helpers/waitForTransitionEnd.ts
```diff
@@ -0,0 +1,14 @@
+import { page } from '../setup'
+
+/** Wait for a transition to finish. */
+async function waitForTransitionEnd(selector: string) {
```
trevinhofmann, line 4: I would also prefer this to be an arrow function, but I see that a few of the other helpers are also defined this way, so this shouldn't be a blocking issue for the PR.

@raineorshine -- It would be nice if we refactor to consistently use arrow functions IMO (and enforce with ESLint).

trevinhofmann (approved):
Thank you, @ethan-james!

There are a couple of minor refactoring questions/suggestions, but I don't think any are blocking.

The implementation is still working as expected 👍

raineorshine (approved):

raineorshine (approved):
Great work!

ethan-james:
> Great work!

Thank you!

Issue #2514, opened 2024-10-31
[Capacitor] Top safe area is covered


raineorshine:
Tested in Xcode Simulator.

Should also be launched and tested on physical device.

## Steps to Reproduce

Run **em** in the Xcode Simulator.

## Current Behavior

Top safe area is covered by the UI:

![Simulator Screenshot - iPhone 16 Pro - 2024-10-31 at 19 25 20](https://github.com/user-attachments/assets/a7f81388-3154-4704-b24f-549bfb734ffc)

## Expected Behavior

UI should appear below top safe area in the native app.

Make sure the position on web and PWA does not change.

Issue #1700, opened 2023-01-24
[Capacitor] Set overscroll background color from theme


ankitkarna99:
There was a default white background in the capacitor WebView. It has been temporarily fixed for Dark theme by using ios configuration inside of `capacitor.config.ts` as:

```
ios: {
  backgroundColor: '000000',
},
```

But we want the `backgroundColor` to be able to configurable based on the app's theme Light or Dark.

## Steps to Reproduce
- Sync capacitor
- Run iOS build
- Change to Light theme
- Try to scroll past the top or bottom of the screen.

## Current Behavior
- A white background appears. 

## Expected Behavior
- The background should be white or black depending upon the app's theme.

ankitkarna99:
## Proposed Solution

- Omit the use of `backgroundColor` in capacitor config and instead add `scrollEnabled: false` as:
```
ios: {
  scrollEnabled: false,
},
```
- Make the app scrollable by updating `html, body, #root, #app` selector in App.css to:
``` 
html,
body,
#root,
#app {
  height: 100%;
  font-size: 16px;
  overflow-y: auto;
  clip-path: inset(0 0 0 0);
}
```
- **[Not Tested]** Can remove the EnableBounce.m objective-c add-on if it doesn't affect the bouncing scroll.

msdewitt:
@raineorshine I have some other changes going in first which will affect this issue. If we can merge [this pr,](https://github.com/cybersemics/em/pull/2555) then I am confident I can make some changes with Capacitor StatusBar plugin.

PR #2555, opened 2024-11-06
Fix: [Capacitor] Top safe area is covered

msdewitt:
#2514

This is a small fix for the overlap issue caused on IOS without the contentInset being set. By default, the app will overlap without this configuration.

Screenshots after fix on ios iPhone 15 emulator:

Screenshot 2024-11-06 at 12 35 26 PM

Screenshot 2024-11-06 at 12 36 57 PM


trevinhofmann (changes requested):
Thank you, @msdewitt!

This change works when the user is at the top of the page, but the toolbar area is still hidden after scrolling down.

## iOS Simulator:

https://github.com/user-attachments/assets/f3615e6c-2e30-4e4f-9aab-dd07ded238f5

## Physical iOS:

https://github.com/user-attachments/assets/88933041-83ab-4d87-b6b9-c627f0081246

msdewitt:
@trevinhofmann I'm not sure if that issue is covered in the bug card attached to this fix. This fix is only to fix the overlap of the top bar icon interacting with the phone icons. Do you have a link to the bug issue for the navbar not sticking to the top that you mentioned above? Thanks

msdewitt:
@raineorshine I see now. I was confused because I thought the tool bar was not supposed to scroll with the movement, but I see it hiding after it scrolls in the video from watching it a few times. This is indeed an issue. Let me see what I can do.

raineorshine:
Thanks! The toolbar is position fixed and expected to not scroll (i.e. it should stay at the top of the viewport regardless of the scroll position). For reference, it should behave the same as on desktop or Mobile Safari.

msdewitt:
There is an update for this fix now which can handle the scrolling.

https://github.com/user-attachments/assets/4901bbfd-d737-4be0-9e0f-d2fe67c76a34


Please review when you can @trevinhofmann  and @raineorshine

trevinhofmann:
Thank you, @msdewitt! I am away from my office at the moment, but I will be able to test the latest changes with physical devices tomorrow.

raineorshine:
Thanks! That seems to fix it for the toolbar at least.

However, the content is still overlapping with the safe area when you scroll up. This is a frame from your video:

image

No part of the UI should overlap with the safe area, including the toolbar, menu, editor, scroll zone, footer, etc. The safe area should just be the background color, otherwise it interferes with the clock, connection, and battery icons that are rendered by iOS.

msdewitt:
@raineorshine Actually, its not that it overlaps, but that by default the ios status bar will change from a solid color to transparent when you scroll down. This is a default behavior which we can change by using Capacitor StatusBar plugin. It's just that those changes overlap with another issue I am working on and was preparing to merge later. Do you want me to handle the Capacitor StatusBar changes in this fix or in this [issue](https://github.com/cybersemics/em/issues/1700)?

index.html
```diff
@@ -24,7 +24,7 @@
     <title>em</title>
 
   </head>
-  <body>
+  <body style="padding-top: env(safe-area-inset-top);">
```
raineorshine, line 27: Let's apply this style in [panda.config.ts](https://github.com/cybersemics/em/blob/82c1f68a31df4086867f3e809c563cdfe2e84ac4/panda.config.ts#L159) for consistency.

raineorshine (changes requested):
I see! Thanks for clarifying. Let's add the StatusBar plugin in a separate PR.

index.html
msdewitt: This is actually unused and I removed it in the latest commit. Its good to go now.

trevinhofmann:
@msdewitt -- Can you please have a look at the commits in this PR and get those cleaned up? It seems that there was some issue with a rebase/merge, because there are now many unrelated commits and file changes. Thank you!

msdewitt:
@trevinhofmann Yes, I got too annoyed with the rebase on this branch and just made a new pr after closing this one. Please see: [Link](https://github.com/cybersemics/em/pull/2593) for further discussions. Its an easier review now.

PR #2557, opened 2024-11-06
Upgrade capacitor and plugins.

msdewitt:
No changes in android or ios emulators when upgrading from Capacitor 4 to 6.

It's all manageable and not a drastic change such as going from Capacitor 3 to 4.


raineorshine (changes requested):
Thanks!

The yarn.lock indicates that you are on yarn classic (v1), which is no longer supported. Yarn is no longer hosted on `npm` and modern yarn (currently v4) requires installation with Corepack:

https://yarnpkg.com/getting-started/install

msdewitt:
@raineorshine Is this pr good to merge for the capacitor updates? It may cause other outstanding prs to update their package.json files.

trevinhofmann:
@msdewitt -- Just a heads up that there are merge conflicts in the `yarn.lock` which will need to be resolved.

msdewitt:
@trevinhofmann Do you think I should just delete the lock file then? I was pushing it because it contains the new versions for the capacitor v6. I wasn't sure if this was a good idea to push up for an open source project or not.

raineorshine:
Sorry about the conflict! 

For some additional background: PR's are generally merged in the order that that they pass their code review. We try to make this as close as possible to the order the PR's are received, but it can vary depending on review times and the time it takes for change requests to be resolved.

The best way I've found to resolve lock file conflicts is to revert to the [upstream lock file](https://github.com/cybersemics/em/blob/main/yarn.lock) after the rebase, then run `yarn`. The (merged) `package.json` on your branch will have the necessary changes and yarn will update the lock file appropriately.

This is a safer approach than deleting the lock file and regenerating it from scratch.

.yarnrc.yml
```diff
@@ -1,6 +1,7 @@
 logFilters:
-  # react-native-web overrides react-native, so we don't need to have react-native installed
```
raineorshine, line 2: Why remove this helpful comment?

.yarnrc.yml
```diff
@@ -1,6 +1,7 @@
...
   - level: discard
     text: "em@workspace:. doesn't provide react-native (p52384), requested by react-native-gesture-handler."
 
 nodeLinker: node-modules
+
+yarnPath: .yarn/releases/yarn-4.5.1.cjs
```
raineorshine, line 7: Can you explain this and the addition of the release file?

raineorshine (changes requested):
Thanks!

raineorshine:
Just a small request in the future: When pulling changes from upstream main, please **rebase** instead of doing a merge. Rebase "replays" all of your commits on top of the upstream main. Thus there is no merge commit and it results in a cleaner commit history.

```
git remote add upstream https://github.com/cybersemics/em
git fetch upstream main
git rebase upstream/main
git push origin main --force
```

.yarnrc.yml
msdewitt: I'll put it back in. I didn't know it got lost.

.yarnrc.yml
msdewitt: Ah, this thing is still showing up even though I switched to yarn 4! Annoying

msdewitt: This issue seems to happen when an older version of yarn is still installed despite having the newest version running.

raineorshine (changes requested):
Looks like the lock file is not aligned yet. Try reverting to the lock file from `main`, and then running `yarn` with the new package.json. Thanks!

raineorshine (approved):
Perfect

Issue #2399, opened 2024-09-16
Slow virtualization of thoughts


raineorshine:
> I found that at 1000 thoughts, the interactions start to feel sluggish. This happens with the last thoughts, and it doesn't happen with the first thoughts. It starts at around ~500. After testing and looking at the code I figured this is because of the virtualization. The virtualization algorithm cuts the last thoughts but not the thoughts at the start. This makes the app performant for the start of the list but slow for cases where there are many thoughts and the user has scrolled there.
> 
> I tested with a hacked version of window virtualization (with hardcoded values), and it solved the slow interactions. This could really be the main thing to make em super performant. At first glance, it seems such a change is significant but tractable. You know more than I do about the algorithm. Do you see any roadblocks to making this happen?

Good find! I think there are probably some good opportunities to optimize the virtualization algorithm.

In the VirtualThought component, `shimHiddenThought` will remove a thought from the DOM after it becomes hidden (and its fade out animation completes):

https://github.com/cybersemics/em/blob/57c332047a3517ff832f037544f9c6438c13f364/src/components/VirtualThought.tsx#L101-L104

However, when I look at the DOM it does not seem to be consistently working. So part of what we may be dealing with is a bug in the current virtualization algorithm. It might be good to check if `shimHiddenThought` is working at all in your test case.

# Working

## Steps to Reproduce

```
- US
  - States
    - Alabama
    - Alaska
    - Arizona
    - Arkansas
    - California
    - Colorado
    - Connecticut
    - Delaware
    - Florida
    - Georgia
    - Hawaii
    - Idaho
    - Illinois
    - Indiana
    - Iowa
    - Kansas
    - Kentucky
    - Louisiana
    - Maine
    - Maryland
    - Massachusetts
    - Michigan
    - Minnesota
    - Mississippi
    - Missouri
    - Montana
    - Nebraska
    - Nevada
    - New Hampshire
    - New Jersey
    - New Mexico
    - New York
    - North Carolina
    - North Dakota
    - Ohio
    - Oklahoma
    - Oregon
    - Pennsylvania
    - Rhode Island
    - South Carolina
    - South Dakota
    - Tennessee
    - Texas
    - Utah
    - Vermont
    - Virginia
    - Washington
    - West Virginia
    - Wisconsin
    -  Wyoming
      - a
        - b
```

1. Set the cursor on US → States → Wyoming → a → b
2. Refresh the page

## Current & Expected Behavior

[...]

PR #2559, opened 2024-11-07
Fix: Slow virtualization of thoughts

dangold604:
This resolves [2399](https://github.com/cybersemics/em/issues/2399).

### DIFF
--- a/src/components/VirtualThought.tsx
+++ b/src/components/VirtualThought.tsx
@@ -1,11 +1,13 @@
 import React, { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react'
 import { shallowEqual, useSelector } from 'react-redux'
+import { css } from '../../styled-system/css'
 import Autofocus from '../@types/Autofocus'
 import LazyEnv from '../@types/LazyEnv'
 import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import State from '../@types/State'
 import ThoughtId from '../@types/ThoughtId'
+import useChangeRef from '../hooks/useChangeRef'
 import useDelayedAutofocus from '../hooks/useDelayedAutofocus'
 import useSelectorEffect from '../hooks/useSelectorEffect'
 import { hasChildren } from '../selectors/getChildren'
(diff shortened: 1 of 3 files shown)


src/components/VirtualThought.tsx
```diff
@@ -107,7 +107,7 @@ const VirtualThought = ({
   const isVisible = zoomCursor || autofocus === 'show' || autofocus === 'dim'
   const shimHiddenThought = useDelayedAutofocus(autofocus, {
     delay: 750,
-    selector: autofocusNew => autofocus === 'hide' && autofocusNew === 'hide' && !!height,
+    selector: autofocusNew => ((autofocus === 'hide' && autofocusNew === 'hide') || (autofocus === 'hide-parent' && autofocusNew === 'hide-parent')) && !!height,
```
raineorshine: I'll point out the comment a few lines above:

`// Only shim 'hide', not 'hide-parent', thoughts, otherwise hidden parents snap in instead of fading in when moving up the tree.`

raineorshine (changes requested):
Thanks!

Unfortunately, this causes the parent to fade in too quickly when navigating up the hierarchy.

## Steps to Reproduce

```
- a
  - b
    - c
      - d
```

1. Set the cursor on `c`
2. Set the cursor on `b`

## Current Behavior

7bbb59f

`a` fades in too quickly:

https://github.com/user-attachments/assets/899ae2dd-f28f-444f-9137-eb47bdcddefa

## Expected Behavior

`main`

`a` fades slowly (layoutSlowShiftDuration):

https://github.com/user-attachments/assets/544f2661-4122-4bb6-9ea1-edb94973c859

src/components/VirtualThought.tsx
dangold604: Hi @raineorshine Thanks for your feedback.
To smoothly animate the thoughts, change the duration time of the opacity transition of each thought from `150` to `1100` and adjust the transition-timing-function from `ease-in` to `ease-out`.

https://github.com/cybersemics/em/blob/f055c3a4a425ebe311e002c5950af5c1fa7906d4/src/components/Subthought.tsx#L132

https://github.com/cybersemics/em/blob/f055c3a4a425ebe311e002c5950af5c1fa7906d4/src/durations.config.ts#L28

raineorshine: This animation intentionally uses a slow ease-out when the autofocus changes, and a fast ease-in otherwise, e.g. when thoughts are expanded. Altering the animation duration and timing function in the way you are suggesting will break the expected animation behavior.

## Suggested

https://github.com/user-attachments/assets/a6cf5e09-1929-48cc-8410-31144808832a

## Expected

`main`

https://github.com/user-attachments/assets/8c0f431b-1dab-4a9d-aad2-0097ef39c557

PR #2560, opened 2024-11-07
Set overscroll background color from theme

msdewitt:
[Issue 1700](https://github.com/cybersemics/em/issues/1700)

This is the suggested fix in the issue.
Simulator Screen Recording - iPhone 15 Pro - 2024-11-07 at 11 39 37

### DIFF
--- a/capacitor.config.ts
+++ b/capacitor.config.ts
@@ -24,7 +24,7 @@ const config: CapacitorConfig = {
   webDir: 'build',
   ...serverConfig,
   ios: {
-    backgroundColor: '000000',
+    scrollEnabled: false,
   },
   plugins: {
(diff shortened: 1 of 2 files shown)


trevinhofmann:
Hi, @msdewitt! It looks like there are a few snapshot tests failing. Could you please update those with `yarn test:puppeteer --update`?

raineorshine:
Actually, you would only update the snapshot tests after visually confirming that there was no visual regression. Please check the [visual diff](https://github.com/cybersemics/em/wiki/Testing#5-visual-snapshot-tests) to see where these are failing. 

A quick check of the GitHub Actions upload artifact shows that these indeed indicate a regression and should not be updated.

Issue #2548, opened 2024-11-05
Black background color not stripped from ContentEditable


raineorshine:
The background color is incorrectly set to black when the default text color it selected ~however this only manifests after another character is typed~.

This suggests to me that the thought value is being updated correctly, but the innerHTML of the ContentEditable component is not, so as soon as the user starts editing again the value is updated. Indeed, the ContentEditable [has a mechanism](https://github.com/cybersemics/em/blob/6e4fa5311c6b9e5ddb4baa50e352d33aa58b2f8b/src/components/ContentEditable.tsx#L35-L47) to prevent the innerHTML from changing during editing. This may need to be bypassed or overridden in order to ensure that it aligns with the thought value after formatting. Untested.

**Do not simply strip the background color in a followup edit.** We want to fix this at the root of the problem rather than piling on good edits to fix bad edits.

## Steps to Reproduce

```
Hello
```

1. Set the caret at the end of the thought.
2. Set the text color to blue.
3. Set the text color to default (white).

## Current Behavior

```
<font style="background-color: rgb(0, 0, 0);">Hello</font>
```

## Expected Behavior

```
Hello
```

## Testing

Add a puppeteer test that covers this case.

raineorshine:
#2546 partially fixed this, although I'm still seeing an incorrect innerHTML which I would like to fix.

Basically, the thought value is correct, but the innerHTML is not. Since this can be difficult to observe in Chrome Dev Tools without triggering a blur event, I recommend temporarily adding the following code to the [editThoughtActionCreator](https://github.com/cybersemics/em/blob/0a46b573f9536d372a7f86d752c513f62a6aa5b0/src/actions/editThought.ts#L195-L199):

```ts
    setTimeout(() => {
      console.log({
        innerHTML: document.querySelector('[data-editing] [contenteditable]')?.innerHTML,
        value: payload.newValue,
      })
    }, 400)
```

This will show you both the persisted thought value and the innerHTML of the contenteditable, with enough of a delay to ignore any intermediate changes.

## Current Behavior

After following the steps to reproduce (above), the following is logged:

```ts
{
  innerHTML:  "<font style=\"background-color: rgb(0, 0, 0);\">Hello</font>"
  value: "Hello"
}
```

## Expected

```ts
{
  innerHTML:  "Hello"
  value: "Hello"
}
```

raineorshine:
It appears I was on the wrong commit! Sorry about that.

[...]

PR #2561, opened 2024-11-07
Run document.execCommand only when necessary

RED-ROSE515:
Fix this issue : Black background color not stripped from ContentEditable #2548


raineorshine (approved):
Works great! Thanks.

Issue #2498, opened 2024-10-26
moveThoughtDown: Thought animates up but not down


raineorshine:
## Steps to Reproduce

```
- a
- b
```

1. Set the cursor on `a`
2. moveThoughtDown
3. moveThoughtUp
4. moveThoughtDown
5. etc

## Current Behavior

The bottom thought animates up, but the top thought does not animate down.

At 4x slowdown: `layoutNodeAnimationDuration: 600`

https://github.com/user-attachments/assets/6bfa4a20-01b6-45da-9076-f86d7005e758

## Expected Behavior

`a` and `b` should both animate into place.

## Notes

Since the `tree-node` animation does not change, this is likely related to how React persists HTMLElements in the DOM. The `key` does not change so I am not sure why the top thought seems to be remounted (and thus not animated).

raineorshine:
Hi @zukilover. Unfortunately I have to reopen as I found one small issue after I merged #2564. One of the thoughts disappears for a single frame.

If you step through this video you can see that `b` disappears for a single frame before moving back down. I confirmed this in a Chrome performance recording as well.

https://github.com/user-attachments/assets/3baba885-b1e8-4f31-8712-7efb9ac91253

PR #2564, opened 2024-11-08
Fix: thought animates up but not down #2498

zukilover:
This update ensures that the previous y position for a given thoughtId is used for calculating smooth transitions in React, and provides a fallback to the current y value if the previous one is unavailable. This is important for the following reasons:

Ensuring Smooth CSS Transitions:

React may reuse DOM nodes and update them asynchronously, which can lead to jarring transitions if the current y value is directly used in the transition.

By utilizing the previous y value, we allow for a smoother transition between the old and new positions when the component re-renders, improving the user experience.

In cases where the previous y value is not available (e.g., on the first render or if the thoughtId is missing), the current y value is used as a fallback.

This ensures that the transition always has a valid reference to work with, preventing issues such as undefined or null values that could break the transition logic.

Result:

[screencast-localhost-3000-2024-11-08-18-08-13_fb6SKTPk.webm](https://github.com/user-attachments/assets/26d2b194-0754-4681-a80f-2495089b649e)


trevinhofmann (changes requested):
Thank you for the PR, @zukilover! This proposal appears to fix the issue in the very simple case with two thoughts:

https://github.com/user-attachments/assets/4e2794d4-9ac8-45a8-a979-312c58f9f1af

However, it causes strange behavior in most other cases.

---

# 1. Thoughts "lag" in their positioning

## Steps to Reproduce

```
- foo
- bar
- baz
```

1. Set the cursor on `foo`
2. `moveThoughtDown`
3. Set the cursor on `bar`

## Current Behavior

After (2), the thoughts are not rearranged. After (3), they rearrange.

https://github.com/user-attachments/assets/c07d08af-ebd7-4235-929d-f3b2e7413764

## Expected Behavior

The thoughts should rearrange after (2) `moveThoughtDown` rather than being delayed until (3).

---

# 2. New thoughts are initially displayed too low

## Steps to Reproduce

```
- foo
- bar
- baz
```

1. Set the cursor on `baz`
2. Hit `enter` to add a new thought
3. Enter text into the new thought

## Current Behavior

After (2), new thought is positioned slightly too low. After (3), it repositions to where it should be in the layout.

https://github.com/user-attachments/assets/4aa0b4ea-c46d-429c-9d49-00f4d51765ad

## Expected Behavior

The new thought should initially be positioned correctly so that it does not reposition after entering text (causing the rerender).

---

I have noticed similar issues, but they all belong to the same category. Due to the use of `usePrevious` for determining `y`, various actions lead to thoughts being incorrectly positioned on the y-axis until some other action triggers a rerender.

There are also a large number of snapshots changed in this PR which seem like they should not be affected. My best guess is that these subtle difference are due to similar slightly incorrect y-axis positioning from using the previous value.

You can attempt to push forward with `usePrevious` if you believe it is the best solution, but my own suggestion would be to focus on a different approach. I believe it would be more fruitful to look into the component's lifecycle to see how the animation can be completed.

zukilover:
Hi @trevinhofmann, thank you for the feedback.

Regarding the first issue you mentioned, **1. Thoughts "lag" in their positioning**, I haven't observed this behavior on my end. I compared it with the main branch, and the positioning seems the same. Could you kindly double-check?

For the second issue, **2. New thoughts are initially displayed too low**, I’ve identified the cause. It was due to a `paddingBottom` being added to new thoughts in this section of the code: 
https://github.com/cybersemics/em/blob/4ac9b8e81e09a62603548058cfe3dad23c39856d/src/components/LayoutTree.tsx#L496

I’ve addressed this by ensuring that the animation optimization occurs when thoughts are rearranged. The only flag I could use to handle this is `dragInProgress`. Please let me know if this approach makes sense.

As for the failed snapshot tests, I believe they aren't related to the changes in this PR. The errors seem to be caused by a very subtle difference, as indicated here:

```
Error: Expected image to match or be a close match to snapshot but was 0.16770833333333335% different from snapshot (805 differing pixels).
```

Here’s one of the screenshots that illustrate the issue:
font-size-18-default-collapsed-subthought-1-diff

raineorshine:
Thank you for your response @zukilover!

> I’ve addressed this by ensuring that the animation optimization occurs when thoughts are rearranged. The only flag I could use to handle this is `dragInProgress`. Please let me know if this approach makes sense.

I just wanted to jump in here to point out that there are a variety of ways that a command like `moveThoughtDown` may be activated:

1. Clicking the toolbar button
2. Activating the keyboard shortcut (Ctrl/Cmd + Shift + Down)
3. Selecting the command from the Command Palette (Ctrl/Cmd + P)
4. Drag-and-drop
5. Activating Undo/Redo

Therefore, `state.dragInProgress` is not sufficient to capture the range of ways that a user may activate a command. 

It is possible to read from the undo history to see the last command that was executed, which you may indeed need for some of the other CSS Animation tasks that involve adding command-specific animations. However, I would not suggest a command-specific approach on this issue. We would expect thoughts to be able to trade places smoothly with the general (unspecialized) animation algorithm that animates y0 to y1.

To add on to what @trevinhofmann was saying, it's not clear to me why `usePrevious` would be needed at all. After all, we know that `y` is correct before and after the command is executed, because the thoughts are rendered in the correct position in both cases. CSS should simply animate between the two. Why isn't that happening? Even stranger, why is it working when a thought moves up, but not down? It may be beneficial to do some additional investigation to better understand the current behavior. That will likely strongly influence the chosen solution.

---

Just to let you know, you've got a merge conflict to resolve before the CI will run again. Please use `git rebase` rather than `git merge`, as that will result in a cleaner commit history. I recommend pulling upstream and rebasing every time you sit down to work on an open PR to prevent it from falling behind. New changes are continuously being added to `main`. Thanks!

zukilover:
Hi @raineorshine, thanks for the insights.

The reason this was fixed with usePrevious hook is because of the render cycle in React that is executed in batch. This is why the transition works only for the first updated thought and not for subsequent ones.

Let me revisit the fix without using usePrevious.


It'd be great if you could point out other potential root causes that I might overlook.

raineorshine:
You are definitely more familiar with the React render cycle than me! Maybe you can help me out. I'm afraid I don't understand how the render cycle would cause a thought moving up to be animated, but a thought moving down to not. Maybe you can include a deeper explanation with your solution, as I would love to understand this better.

zukilover:
@raineorshine React batches state updates, e.g. in a case when you drag an item, it queues up the changes and only re-renders the component once, after all changes are applied. In our case, this causes an issue because React doesn't apply the changes in a way that triggers CSS transitions.

CSS transitions require a change in the style property (like top) over time. If React's batched state updates apply the new values instantly, the transition won’t be able to animate between the old and new values. The change appears to happen too fast if you will, for CSS to animate it.

I was trying to use `requestAnimationFrame` because it is more performant for animations, as it aligns with the browser’s native rendering loop. However, this approach would require me to refactor how we render `treeThoughtsPositioned` items here: 

https://github.com/cybersemics/em/blob/4ac9b8e81e09a62603548058cfe3dad23c39856d/src/components/LayoutTree.tsx#L743

(i.e. create a component for the mapped items)

raineorshine:
Thanks, that’s helpful! So is it multiple re-renders that causes the value to be updated instantly and not have time to transition? I would expect there to be just two values of `top`: before and after. I don't know why it wouldn't animate between those if `top` was not otherwise changing 🤔

How does rAF help here?

The refactor you mentioned does not sound bad.

zukilover:
`top` was actually changing as the `tree-node` is keyed by the thought id. Now, all of the `tree-node` are re-rendered (due to the changes of `top`) at once because of React's internal re-render logic (the batch).

Using `requestAnimationFrame` will "force" these updates to be done in sync browser's rendering cycle.

@raineorshine.

raineorshine:
Interesting. So the idea is that changes to the x,y position should be delayed until the next animation frame, and that will cause them to animate smoothly because they won't be interrupted by a re-render?

zukilover:
If by "delayed" you mean synchronizing with the browser’s repaint cycle, then yes, that's the goal.

Here’s what I’ve tried so far:

1. **Implemented `usePrevious` hook**:  
   This introduced some complications, so I need a more robust way to handle these changes across various actions:
   - Clicking the toolbar button
   - Activating the keyboard shortcut (Ctrl/Cmd + Shift + Down)
   - Selecting a command from the Command Palette (Ctrl/Cmd + P)
   - Drag-and-drop interactions
   - Triggering Undo/Redo

2. **Added `will-change` property**:  
   This didn’t have the desired effect.

3. **Refactored with `requestAnimationFrame`**:  
   I’m still working on this, but so far, it hasn’t resolved the animation issues.

zukilover:
I'm thinking, rather than covering all of the 'reorder' scenarios, what if we exclude this previous y calculation for a new thought creation (that is the Enter to add a thought). I wonder if this condition is a good idea @raineorshine :

```
// Only refer to calculatedY for existing thought
top: cliff < 0 ? y : calculatedY,
```

raineorshine:
It makes sense that a newly created thought would not have a previous `y` value to use.

I don't see the connection between new thought creation and the cliff. The cliff is the additional padding that is added after nested thoughts:

(Greatly exaggerated for clarity)

Brave Browser 2024-11-10 08 30 51

zukilover:
@raineorshine @trevinhofmann I forced pushed to this PR with a completely different approach by simply refactoring the way we map `treeThoughtsPositioned`. It may look much but it is a simple as converting `.map(fn)` into `.map(Component)`.

This should fix the original issue mentioned in this PR without introducing side effects.

src/components/LayoutTree.tsx
```diff
@@ -364,6 +386,167 @@ const linearizeTree = (
...
+  } = props
+
+  const [y, setY] = useState(_y)
+
+  useEffect(() => {
+    if (y !== _y) {
+      // When y changes React re-renders the component with the new value of y. It will result in a visual change in the DOM.
+      // Because this is a state-driven change, React applies the updated value to the DOM, which causes the browser to recognize that
+      // a CSS property has changed, thereby triggering the CSS transition.
+      setY(_y)
```
trevinhofmann, line 434: I will admit that I am confused on how/why this approach fixes the issue. It feels too hacky for my comfort, but I will let @raineorshine decide if it is acceptable.

src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/font-size-13-superscript-1.png
trevinhofmann, line 1: The other snapshots need to be updated, but can you please look into why this `b` thought is slightly shifted by these changes? I wouldn't expect any side effects such as this from fixing the thought animation.

src/components/LayoutTree.tsx
```diff
@@ -364,6 +386,167 @@ const linearizeTree = (
...
+    setSize,
+    cliffPaddingStyle,
+    dragInProgress,
+    autofocusDepth,
+  } = props
+
+  const [y, setY] = useState(_y)
+
+  useEffect(() => {
+    if (y !== _y) {
```
trevinhofmann, line 430: This check shouldn't be necessary, since `setY` won't actually trigger a rerender/update if `y === _y`, right?

src/components/LayoutTree.tsx
```diff
@@ -364,6 +386,167 @@ const linearizeTree = (
...
+      // When y changes React re-renders the component with the new value of y. It will result in a visual change in the DOM.
+      // Because this is a state-driven change, React applies the updated value to the DOM, which causes the browser to recognize that
+      // a CSS property has changed, thereby triggering the CSS transition.
+      setY(_y)
+    }
+  }, [y, _y])
```
trevinhofmann, line 436: Slightly concerning, but won't we trigger an extra render and call to the hook with this, since `y` is a dependency and also gets updated within the effect? Perhaps that extra render is a part of this fix, but I admittedly do not understand.

trevinhofmann (changes requested):
The latest changes have made the animation work as expected:

https://github.com/user-attachments/assets/e8fc4c7a-5fda-481e-8e53-ffcd8f67f7c2

---

The snapshots in the PR are outdated, but there appears to still be one that fails:

`src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/__diff_output__/font-size-13-superscript-1-diff.png`:
font-size-13-superscript-1-diff

The side effect seems to be that the `b` thought is now shifted up by 1 pixel.

src/components/LayoutTree.tsx
zukilover: This check guarantees that it will update only and only if the previous value changes. This at the same time answers your concern about unnecessary re-render.

src/components/LayoutTree.tsx
zukilover: The root cause as we discussed was due to a batched update in React lifecycle, that is why it only animated the first sequence of the renders (top to bottom transition in this case).

useEffect serves a purpose of getting a better control at how we orchestrate its render cycle.

src/components/LayoutTree.tsx
raineorshine: I was hoping you would know something I don't 😅

I don't totally understand why this works, but apparently it does cause the DOM to be updated in a way that the CSS can be transitioned properly. 

My main concern with this is performance. It looks like every node with a changed `y` will be rendered twice. Given the performance heavy LayoutTree and multiple animations happening at once, this may be significant. That said, correctness is more important than performance. 

So it's really question of whether we do some performance measurements now and get a sense of the impact, or kick the can down the road and potentially make life harder for our front-end performance engineer, @dangold604 😇. Since measuring performance is not too time intensive, I would suggest we do this now and then make a decision based on the results. Maybe we'll get lucky and the impact is not significant!

raineorshine: @zukilover @trevinhofmann Would love to hear your thoughts on the potential performance impact of this change, and how we might mitigate that.

zukilover: @raineorshine with the new refactor, React will only triggers a re-render when the y value changes, preventing unnecessary re-renders of other items in the list. This also allows for better CSS transition handling, as React applies the updated state to the DOM, which in turn triggers smooth animations.

React's diffing algorithm is efficient at updating only the affected components, so we shouldn't see significant overhead from this change.

zukilover:
@raineorshine need your advise here as `yarn test:puppeteer -u` still results in failure, even though onion skin doesn't show any differences.

src/components/LayoutTree.tsx
raineorshine: > @raineorshine with the new refactor, React will only triggers a re-render when the y value changes, preventing unnecessary re-renders of other items in the list.

Right, but the question here is, Does the component render twice when the y values changes, and does that affect performance?

> This also allows for better CSS transition handling, as React applies the updated state to the DOM, which in turn triggers smooth animations.

Great, that's definitely what we're looking for.

> React's diffing algorithm is efficient at updating only the affected components, so we shouldn't see significant overhead from this change.

This is true in general, but based on my experience with **em** we actually have to be a bit careful. The LayoutTree component is the slowest component in the hierarchy. I know it to be the case that the `LayoutTree` render function takes multiple milliseconds, so extraneous renders do have an effect on performance.

Maybe the performance is actually *better* because of the new timing of the CSS transitions! It's just hard for me to tell without measuring before/after.

Ultimately this is a moot point since this is a correctness fix, not a performance optimization. I don't see any visible performance issues, so for the sake of moving forward, I will set this concern aside. Thanks for the discussion.

raineorshine:
I'll take a look. It's possible that it is a false positive.

Some background: Even docker and browserless have not eliminated some rendering inconsistencies with the puppeteer snapshot tests. Unfortunately, there are sometimes actual visual regressions on the scale of just a few pixels difference, so we can't afford to loosen the failure threshold. A failed snapshot test forces us to do a careful comparison to ensure it is not a regression. It's annoying, but in this case it's good to be on the safe side.

raineorshine (approved):

raineorshine:
Thank you, great work.

raineorshine:
FYI I had to change `useEffect` to `useLayoutEffect` for the snapshots to render consistently (c6f79c9)

PR #2565, opened 2024-11-08
Lottie icons 2408 (Milestone 3)

ohrytskov:
#2408 

- Scale with state.fontSize
- Preserve extended click area
- Desktop and mobile
- Dark and Light theme
- Preserve highlighting and dimming when icon is active and inactive


src/components/icons/AnimatedIcon.tsx
```diff
@@ -16,11 +28,27 @@ const AnimatedIcon = ({
...
+  const [dynamicColor, setDynamicColor] = useState<string | null>(null)
+
+  // Effect to calculate the color from the parent div
+  useEffect(() => {
+    if (divRef.current) {
+      const computedStyle = getComputedStyle(divRef.current)
+      const hexColor = rgbToHex(computedStyle.color) // Convert to hex
+      setDynamicColor(hexColor)
+    }
+  }, [isLightTheme, style, cssRaw]) // Dependencies that might affect the final style
```
trevinhofmann, line 47: Is this `useEffect` needed? I think we should be able to change `dynamicColor` from being state and just declare it as a normal variable. For reference: https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

src/components/icons/LottieAnimation.tsx
```diff
@@ -1,36 +1,128 @@
...
+  const [key, setKey] = useState(0)
   const lottieRef = useRef<LottieRefCurrentProps | null>(null)
 
+  /**
+   * Converts hex color to RGBA.
+   *
+   * @param hex - The hex color code.
+   * @returns The RGBA values.
+   */
+  const hexToRGBA = useCallback((hex: string): number[] => {
```
trevinhofmann, line 44: Since this is a pure function, we should be able to declare it as a regular function outside of the component rather than using `useCallback` for it.

src/components/icons/LottieAnimation.tsx
```diff
@@ -1,36 +1,128 @@
...
+   * fill and updates its color using the provided RGBA array.
+   * If the item contains a nested items array, it recursively
+   * calls itself to ensure that all relevant shapes within the
+   * structure are updated with the new color.
+   *
+   * @param item - The shape item object which may contain
+   * stroke or fill color properties.
+   * @param rgbaArray - The RGBA values to set as new color.
+   */
+  const updateColorsInItem = useCallback((item: any, rgbaArray: number[]) => {
```
trevinhofmann, line 70: This function could also be declared outside of the scope of the component without `useCallback`.

src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/font-size-13-initial-load-1.png
trevinhofmann, line 1: The icons were shifted off-center (too high). Can this be corrected so that they remain vertically centered?

trevinhofmann (changes requested):
Thank you for the PR, @ohrytskov!

--

## Issue 1: Icons shift during animation

Some of the animations still cause the icon to shift in place. I believe this was postponed until Milestone 3:

https://github.com/user-attachments/assets/26bea373-2d3f-425d-95da-717a86d6d666

---

@raineorshine -- I have a question about the "Preserve extended click area" requirement. I understand that there should be some additional clickable space around the icon, but is it expected to be this large? Currently, the entire green region is clickable, and there is no unclickable gap between the buttons:

Screenshot 2024-11-13 at 12 24 30 AM

raineorshine:
> @raineorshine -- I have a question about the "Preserve extended click area" requirement. I understand that there should be some additional clickable space around the icon, but is it expected to be this large? Currently, the entire green region is clickable, and there is no unclickable gap between the buttons:

Yes, that's right. The click region in `main` is correct and there should be no unclickable gap between the buttons.

src/components/icons/LottieAnimation.tsx
ohrytskov: @trevinhofmann, can we move these functions to utils? What do you think?

trevinhofmann: For now, I would be okay with either keeping them in the component file (outside of the component definition) or moving them to the `util` directory. If it is likely that they would be reusable in other files, then it would definitely make sense to move them to the `util` directory.

trevinhofmann:
@ohrytskov -- Just a heads up that there are some merge conflicts to resolve and puppeteer tests failing.

Please let me know if you have any questions, and let me know when you would like me to review the PR again. Thank you!

src/components/icons/LottieAnimation.tsx
raineorshine: I think you can use `hexToRgb` provided by `@mui/material` which is already included in the project.

---

FYI I follow a simple rule when it comes to helper functions: If it's only used in one file, define it in that file. If it's used by multiple files, define it in the `util` directory.

This reduces surface area and provides a more deterministic rule-of-thumb that avoids developers having to anticipate future needs.

src/components/icons/AnimatedIcon.tsx
```diff
@@ -1,10 +1,22 @@
...
 import { css, cx } from '../../../styled-system/css'
 import { icon } from '../../../styled-system/recipes'
 import { token } from '../../../styled-system/tokens'
 import AnimatedIconType from '../../@types/AnimatedIconType'
 import { ICON_SCALING_FACTOR } from '../../constants'
+import theme from '../../selectors/theme'
 import LottieAnimation from './LottieAnimation'
 
+/** Function to Convert RGB Color to Hex. */
+const rgbToHex = (color: string): string => {
```
raineorshine, line 12: You can use `rgbToHex` provided by `@mui/material` which is already included in the project.

src/components/icons/LottieAnimation.tsx
```diff
@@ -1,36 +1,125 @@
...
+ * fill and updates its color using the provided RGBA array.
+ * If the item contains a nested items array, it recursively
+ * calls itself to ensure that all relevant shapes within the
+ * structure are updated with the new color.
+ *
+ * @param item - The shape item object which may contain
+ * stroke or fill color properties.
+ * @param rgbaArray - The RGBA values to set as new color.
+ */
+const updateColorsInItem = (item: any, rgbaArray: number[]) => {
```
raineorshine, line 44: Let's properly type all Typescript code and avoid the use of `any`. We lose the benefits of the type checker if we use `any`.

src/components/icons/LottieAnimation.tsx
```diff
@@ -1,36 +1,125 @@
...
   style,
 }) => {
+  const [key, setKey] = useState(0)
   const lottieRef = useRef<LottieRefCurrentProps | null>(null)
 
   useEffect(() => {
     if (lottieRef.current) {
       lottieRef.current.setSpeed(speed)
+      if (animationData && color) {
+        changeLineColor(animationData, color)
```
raineorshine, line 114: `changeLineColor` mutates the inside of `animationData` without the object reference changing. This results in the awkward external `key` required notify the component that the value has actually changed.

Instead, use `useMemo` to create a derived `animationDataWithColor` that can be passed to `Player`. Then the Player component will update naturally when the `animationDataWithColor` reference changes. It's easier to reason about because it avoids mutations, and it's friendlier to React since it follows norms around object references.

src/components/icons/LottieAnimation.tsx
```diff
@@ -1,36 +1,125 @@
...
+ */
+const updateColorsInItem = (item: any, rgbaArray: number[]) => {
+  if ((item.ty === 'st' || item.ty === 'fl') && item.c) {
+    // Check for animated color property
+    if (item.c.a === 1) {
+      item.c.k.forEach((keyframe: any) => {
+        keyframe.s = rgbaArray
+      })
+    } else {
+      item.c.k = rgbaArray
```
raineorshine, line 52: Please note that it is dangerous and considered a bad practice to mutate arguments. Arguments are the input to a function and should generally not be modified. Use the return value as the output. Pure functions are easier to reason about and more self-contained. Modifying the argument has upstream implications and can affect anything above it in the call stack.

I won't make you rewrite the function this time, but please keep this recommendation in mind for the future. Thanks!

raineorshine (changes requested):

src/components/icons/AnimatedIcon.tsx
ohrytskov: The custom rgbToHex function was successfully replaced with the @mui/material utility.

src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/font-size-13-initial-load-1.png
ohrytskov: Sure, I will make sure the icons stay centered.

src/components/icons/LottieAnimation.tsx
ohrytskov: Thank you for highlighting the importance of functional purity and immutability. I completely agree that avoiding the mutation of arguments is crucial for maintaining code readability and reliability. It ensures that functions remain predictable and easier to test. 
However, I also considered that copying a large object, like the JSON animation data, might significantly affect performance. 
To address this, I enhanced the handling of Lottie animation data by implementing immutability and memoization strategies.

raineorshine: Yes, I think memoization is the right call to ensure good performance here!

ohrytskov:
All checks have successfully passed, and the necessary adjustments have been made. The PR for milestone 3 is ready for review.

src/components/icons/LottieAnimation.tsx
trevinhofmann: I'd slightly prefer to use something like `map` to avoid mutating within `updateColorsInItem` at all. That said, I can confirm that this is only used by `changeLineColor` which passes a clone, and the result is now memoized with `useMemo`. Thank you, @ohrytskov!

src/@types/LottieTypes.ts
```diff
@@ -0,0 +1,56 @@
...
+  shapes?: (ShapeItem | NestedShape)[] // Combine both types for flexibility
+  [key: string]: unknown // Allow for any additional properties
+}
+
+/** The main Lottie data interface consisting of multiple layers. */
+interface LottieData {
+  layers: ShapeLayer[] // Main layers array, which may contain various types
+}
+
+export default LottieData
```
trevinhofmann, line 56: This project is using a convention where each type file in `@types` only exports a single default type. If multiple types need to be exported, please create a separate file for each. Where types are derived from others, you can import and use them (e.g. import `ShapeLayer` in the `LottieData` type file).

src/components/icons/AnimatedIcon.tsx
```diff
@@ -16,11 +20,21 @@ const AnimatedIcon = ({
...
+  const isLightTheme = useSelector(state => theme(state) === 'Light')
+  const lottieColor = isLightTheme ? '#000000' : '#FFFFFF'
   const newSize = size * ICON_SCALING_FACTOR
   const color = style.fill || fill || token('colors.fg')
 
+  // Create a ref to the parent div
+  const divRef = useRef<HTMLDivElement | null>(null)
+
+  // Calculate dynamic color directly
+  const computedStyle = divRef.current ? getComputedStyle(divRef.current) : null
```
trevinhofmann, line 32:
```suggestion
  const computedStyle = divRef.current ? window.getComputedStyle(divRef.current) : null
```

This isn't too important, but nice for consistency. All other calls to this in the project explicitly use `window.getComputedStyle`.

src/components/icons/LottieAnimation.tsx
```diff
@@ -1,38 +1,146 @@
...
   style?: React.CSSProperties
 }
 
+/**
+ * Converts hex color to RGBA.
+ *
+ * @param hex - The hex color code.
+ * @returns The RGBA values.
+ */
+const hexToRGBA = (hex: string): RGBA => {
```
trevinhofmann, line 21: Would it make sense to also use `hexToRgb` from `@mui/material`? It does support rgba, but I'm not sure if it would be worth extracting the values from the string if we need an array of numbers:

```
> hexToRgb('#ffffffaa')
'rgba(255, 255, 255, 0.667)'
```

trevinhofmann (changes requested):
Thank you for fixing the position shift that happens for some animations.

I'm still seeing a very minor scaling issue for some. For example, the Italic button is slightly taller at the end of the animation and then shrinks suddenly when the animation completes:


https://github.com/user-attachments/assets/32251545-f70a-482e-85a8-e22205b8b23f

This _may_ be a small enough difference to be negligible/ignored, but I'll defer to @raineorshine for that decision.

---

There are some merge conflicts in the snapshots. Can you please resolve those, @ohrytskov?

raineorshine:
> This _may_ be a small enough difference to be negligible/ignored, but I'll defer to @raineorshine for that decision.

We'll want to fix this, as it is noticeable. Thanks!

src/components/icons/LottieAnimation.tsx
ohrytskov: Parsing hex colors directly is generally simpler because hex values are compact, predictable, and easier to process programmatically.

raineorshine: If we are using `rgbToHex` from the `@mui/material` library, we should also use `hexToRgb` from the library for consistency. It doesn't matter how simple the implementation is when it has already been implemented for us. 

If implementing it ourselves meant we could remove `@mui/material` as a dependency, that would be a compelling case. As it stands there does not seem to be anything to gain.

ohrytskov:
I've addressed the feedback, and the PR is ready for review.

ohrytskov:
I’ve made a few additional updates.

src/components/icons/LottieAnimation.tsx
```diff
@@ -1,38 +1,164 @@
...
+ * @param item - The shape item object which may contain
+ * stroke or fill color properties.
+ * @param rgbaArray - The RGBA values to set as new color.
+ */
+const updateColorsInItem = (item: ShapeItem, rgbaArray: RGBA) => {
+  if ((item.ty === 'st' || item.ty === 'fl') && item.c) {
+    // Check for animated color property
+    if (item.c.a === 1) {
+      if (Array.isArray(item.c.k)) {
+        const keyframes = item.c.k as AnimationKeyframe[]
```
trevinhofmann, line 69: Why do we need to assert the type here (and why is it safe to assume)? The type of `item.c.k` is `RGBA | number[] | RGBA[] | AnimationKeyframe[]`. How are we sure that the type here will be `AnimationKeyframe[]` and not potentially `RGBA`, `number[]`, or `RGBA[]`? If we know that the type will always be `AnimationKeyframe[]`, perhaps the type for `ShapeItem` needs to be updated accordingly so that we don't need to assert this type here.

trevinhofmann (changes requested):
Thank you for making those changes, @ohrytskov. The PR is looking great overall.

The shifting seems to be resolved at the default font size:


https://github.com/user-attachments/assets/0c1aa06d-f080-4d51-8e31-6c5c7e48f1da

However, there is still a shift at different font sizes.

Font size 40:

https://github.com/user-attachments/assets/e7ecd04f-61c2-4327-b39b-4d651fb97a7c

Font size 8:

https://github.com/user-attachments/assets/fddb86ff-5846-49f4-a4ef-9d26b1a5a07c

src/components/icons/LottieAnimation.tsx
ohrytskov: We need to assert the type because item.c.k can be different types based on whether the color property is static or animated. Earlier, ColorProperty had a broad type for k (RGBA | RGBA[] | number[] | AnimationKeyframe[]) because it was assumed there could be various types during testing and debugging. However, in this context, it's determined that when the color property is animated, item.c.k is specifically an AnimationKeyframe[]. Therefore, it's safe to assert this type here. Adjusting ColorProperty helps handle both static and animated cases appropriately, but the type assertion remains necessary because item.c.k can vary in different instances.

ohrytskov:
PR updated and ready for the next review.

src/components/icons/LottieAnimation.tsx
trevinhofmann: I see that the type has been narrowed a bit since that last comment, and the TS compiler assumes that the type of `item.c.k` is `RGBA | AnimationKeyframe[]`. Just trying to avoid 

I'm still not sure if I understand the typing. Is it because `item.ty === 'st' || item.ty === 'fl'` that we can assume the type of `item.c.k` is `AnimationKeyframe[]`? If that's the case, we _could_ update the `ShapeItem` definition to account for that. I'm just hoping to avoid type assertions because making these assumptions (if ever false) can lead to unexpected runtime errors that would have otherwise been caught by TS.

This isn't the only type assertion we have in the project, though, so maybe @raineorshine can give some direction on what is preferable.

trevinhofmann (changes requested):
Thank you, @ohrytskov. The `italic` button seems to have the correct scaling at each font size. Some of the other buttons still have minor scaling issues, though. Below are a few examples.

The `strikethrough` animation is slightly smaller than the icon, so it shifts to a larger size when the animation completes (default font size):

https://github.com/user-attachments/assets/d83191ad-b63d-4246-b58b-f67950342f30

The `swap parent` animation is slightly larger than the icon, so it shifts to a smaller size when the animation completes (default font size):

https://github.com/user-attachments/assets/fe641110-7c19-4871-8dba-f2f575aecdae

The `subcategorize` animation is slightly taller than the icon, so it shortens when the animation completes (default font size):

https://github.com/user-attachments/assets/191cbd86-0101-47f9-903e-82e225ed4803

The `pin` icon shifts up and to the left after the animation, at a font size of 40 (may also happen at the default font size, but not as noticeable):

https://github.com/user-attachments/assets/f27797d5-f54c-4bef-a41d-f54e354faf43

There may be other icons affected as well. Can you please retest each animation at the minimum (8), default (18), and maximum (40) font size to check for scaling inconsistencies before the next review?

---

Somewhat unrelated, but I've also noticed that the (inactive->active) animations are playing when no thought is selected. Is that the intended behavior, @raineorshine?

If not, I would recommend addressing it in a separate PR as it is unrelated to Milestone 3.

https://github.com/user-attachments/assets/245cf078-43dc-4530-a030-0e2b70a58e0a

raineorshine:
> Somewhat unrelated, but I've also noticed that the (inactive->active) animations are playing when no thought is selected. Is that the intended behavior, @raineorshine?

No, that's not intentional. The lottie animation should not play if the command is not executed successfully. This generally is prevented when `canExecute` returns false, though in some cases `canExecute` can return true and `exec` can return early.

> If not, I would recommend addressing it in a separate PR as it is unrelated to Milestone 3.

I think this needs to happen as part of this PR, otherwise it will introduce a regression into `main`.

src/components/icons/LottieAnimation.tsx
ohrytskov: Thank you for bringing up this concern. You're absolutely right to question whether we can safely assume the type of item.c.k as AnimationKeyframe[] based on item.ty === 'st' || item.ty === 'fl'. Making unchecked assumptions can lead to unexpected runtime errors, especially if the Lottie JSON structure evolves or if we’re working with incomplete or inaccurate typings.
The current assumption is grounded in the Lottie JSON specification. For shape items of type 'st' (stroke) or 'fl' (fill), the c property generally represents color. This property can take two forms:

- A static RGBA array ([r, g, b, a]), or
- A series of animation keyframes (AnimationKeyframe[]) if the animation flag (a) is set to 1.

Thus, the conditional check (item.ty === 'st' || item.ty === 'fl') narrows the scope, making this assumption valid within this specific context. However, as you pointed out, relying on type assertions (item.c.k as AnimationKeyframe[]) can mask potential edge cases or changes in the JSON schema.

Recent Updates:

In the recent commit, I addressed this by refactoring the ColorProperty type into a discriminated union. This ensures that the type of k is correctly tied to the value of a:

```tsx
type ColorProperty =
  | { a: 0; k: RGBA } // Static color
  | { a: 1; k: AnimationKeyframe[] } // Animated color

```

With this structure, TypeScript now guarantees that item.c.k is of type AnimationKeyframe[] whenever item.c.a === 1, eliminating the need for type assertions like item.c.k as AnimationKeyframe[].

The recent changes in the commit enhance type safety by removing unnecessary type assertions and leveraging TypeScript's capabilities. However, input from @raineorshine or other team members on the broader strategy for balancing type safety and development efficiency across the project would be valuable. While type assertions can be acceptable as a temporary solution, they should be mitigated with thorough documentation and robust test coverage. The ultimate goal is to achieve a balance between immediate development needs and long-term maintainability.

ohrytskov:
@trevinhofmann, the updated version of the PR is ready for review.

src/components/icons/LottieAnimation.tsx
ohrytskov: Thank you so much for your feedback!
I've taken your advice and refactored the code to use type predicates. Specifically:

- Split `ColorProperty` into `StaticColor` and `AnimatedColor` types, each in their own file for better modularity.
- Added the predicate functions `isStaticColor` and `isAnimatedColor` to leverage TypeScript's type narrowing.
- Modified `LottieAnimation.tsx` to utilize the updated types and improved color update logic.

ohrytskov:
The forEach loops were replaced with map to enhance immutability.

trevinhofmann:
Thank you very micj, @ohrytskov! Could you please resolve the merge conflict in `src/components/icons/LottieAnimation.tsx`? Excited to review this again once that is ready 🙂

ohrytskov:
**Regarding the inactive to active animations that are playing when no thought is selected:**

Here is the code snippet from `ToolbarButton.tsx`:

```tsx
if (!customize && isButtonExecutable && !disabled && !scrolled && isPressing) {
  executeShortcutWithMulticursor(shortcut, { store, type: 'toolbar', event: e })

  if ((!isActive && !commandState) || (isActive && !isButtonActive)) {
    setIsAnimated(true)
  } else {
    setIsAnimated(false)
  }

  // prevent Editable blur
  if (isTouch) {
    e.preventDefault()
  }
}
```

This code shows that the animation is triggered in the same `if` block as `executeShortcutWithMulticursor`. The variables in the `if` statement allow both the command and animation to run together. Fixing it at the command level will resolve this, so the issue is unrelated to this PR.

ohrytskov:
@trevinhofmann, the merge conflicts have been resolved. Looking forward to your review.

raineorshine:
> Also, apologies if I missed it, but I don't see any changes related to this in the recent commits at a quick glance:
> 
> > Somewhat unrelated, but I've also noticed that the (inactive->active) animations are playing when no thought is selected.
> 
> > I think this needs to happen as part of this PR, otherwise it will introduce a regression into `main`.

@trevinhofmann Could you identify steps to reproduce? Maybe I am misunderstanding, because what I'm seeing is:

- When there is no cursor: animations do not play (correct)
- When there is a cursor and active browser selection: animations play (correct)
- When there is a cursor without an active browser selection: animations play, commands are executed, but toolbar icons do not stay highlighted (incorrect, but present in `main`).

I had suggested fixing it in this PR because I assumed the issue was introduced here, but now that I test it out I realize I don't understand. Once I'm clear on what is being reported I can provide a better response to @ohrytskov. Thanks!

trevinhofmann:
Yes, this is the case that I reproduced and which is also present in `main`:

1. Click outside of the editable thoughts
2. Click on the Bold/Italic/Underline/Strikethrough button

The thought's formatting does update and the button animation plays, but the state of the button does not change after the animation.

I'm not sure what the desired behavior is, but noticed the inconsistency with the button animation playing even though its state does not update.

Since it is also present in `main`:

> I would recommend addressing it in a separate PR as it is unrelated to Milestone 3.

raineorshine:
Got it! I originally misunderstood. Yes, we can create a separate issue for that.

trevinhofmann (approved):
All of the icon animations that I could find look correctly scaled to me at various font sizes. Thank you very much for your perseverance on this one, @ohrytskov!

---

@raineorshine, I noticed some off-center issues with the blue/gray bars/circles on the Customize Toolbar page, but those are unrelated to the animations and already present in `main`. I suppose that I should start filing separate issues for those things (and making sure they aren't already tracked as issues) instead of cluttering PR reviews with tangentially-related items. I'll plan to do that going forward 😄

https://github.com/user-attachments/assets/a086cff7-f1db-4329-93a8-79f1da93f7bf

raineorshine:
> @raineorshine, I noticed some off-center issues with the blue/gray bars/circles on the Customize Toolbar page, but those are unrelated to the animations and already present in `main`.

Good catch! Thank you. I don't think we have an issue for that yet. However, the UI is going to be going through a small redesign, so we actually have a new mockup to apply for this screen.

> I suppose that I should start filing separate issues for those things (and making sure they aren't already tracked as issues) instead of cluttering PR reviews with tangentially-related items. I'll plan to do that going forward 😄

I like that plan :).

src/components/icons/AnimatedIcon.tsx
```diff
@@ -16,11 +20,21 @@ const AnimatedIcon = ({
   children,
   animationComplete,
 }: AnimatedIconType) => {
+  const isLightTheme = useSelector(state => theme(state) === 'Light')
```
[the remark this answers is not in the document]
raineorshine: Thanks for the suggestion. Personally I don't love prefixing with `get` as I don't find it adds any semantic value, but I see the practical use in differentiating the selector from the local variable.

The logic is so minimal here, that factoring out `getIsLightTheme` just to abstract out the `===` sign is arguably not worth the effort, but it's a close call. With even the slightest bit more complexity I would be inclined to factor it out.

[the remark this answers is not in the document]
ohrytskov: I've improved AnimatedIcon by simplifying the useSelector call for isLightTheme.

src/components/icons/LottieAnimation.tsx
```diff
@@ -35,15 +164,26 @@ const LottieAnimation: React.FC<LottieAnimationProps> = ({
...
-    if (lottieRef.current) {
+    if (lottieRef.current && animationDataWithColor) {
       lottieRef.current.setSpeed(speed)
+      setKey(prevKey => prevKey + 1) // Forces the re-render to apply the color change
     }
-  }, [speed])
+  }, [speed, animationDataWithColor])
+
+  if (!animationDataWithColor) {
+    console.warn('animationData is not provided.')
```
raineorshine, line 174: Instead of `console.warn`, we likely either want to fully allow `null` (by not throwing a warning) or fully disallow `null` (by not allowing it in the type).

src/components/RedoIcon.tsx
```diff
@@ -9,7 +9,7 @@ const RedoIcon = ({ fill, size = 18, style = {}, cssRaw, animated, animationComp
       <svg
         xmlns='http://www.w3.org/2000/svg'
         viewBox='0 0 24 24'
-        style={{ ...style, width: '100%', height: '100%' }}
+        style={{ ...style, width: '100%', height: '100%', transform: 'scale(0.96, 0.96)' }}
```
raineorshine, line 12: My only concern with scaling is that it will cause line widths no longer be uniform. All of the icons were custom designed to have the same stroke width.

Can you tell me a little about what the scaling is solving?

src/components/SearchIcon.tsx
```diff
@@ -3,14 +3,14 @@ import AnimatedIcon from './icons/AnimatedIcon'
...
-const SearchIcon = ({ fill, size, style = {}, cssRaw, animated, animationComplete }: IconType) => {
+const SearchIcon = ({ fill, size = 18, style = {}, cssRaw, animated, animationComplete }: IconType) => {
   return (
     <AnimatedIcon {...{ fill, size, style, cssRaw, animated, animationData, animationComplete }}>
       <svg
         xmlns='http://www.w3.org/2000/svg'
         viewBox='0 0 24 24'
         fill='none'
-        style={{ ...style, width: '100%', height: '100%' }}
+        style={{ ...style, width: '100%', height: '100%', transform: 'translate(0, -2%) scale(1, 1)' }}
```
raineorshine, line 13: `scale(1, 1)` is the default, so this shouldn't be needed.

src/components/icons/BumpThoughtDownIcon.tsx
```diff
@@ -3,14 +3,14 @@ import AnimatedIcon from './AnimatedIcon'
...
-const BumpThoughtDownIcon = ({ fill, size, style = {}, cssRaw, animated, animationComplete }: IconType) => {
+const BumpThoughtDownIcon = ({ fill, size = 18, style = {}, cssRaw, animated, animationComplete }: IconType) => {
   return (
     <AnimatedIcon {...{ fill, size, style, cssRaw, animated, animationData, animationComplete }}>
       <svg
         xmlns='http://www.w3.org/2000/svg'
         viewBox='0 0 24 24'
         fill='none'
-        style={{ ...style, width: '100%', height: '100%' }}
+        style={{ ...style, width: '100%', height: '100%', transform: `translate(-1%, 2%) scale(1, 1)` }}
```
raineorshine, line 13: `scale(1, 1)` is the default, so this shouldn't be needed.

src/components/icons/NormalTextIcon.tsx
```diff
@@ -3,14 +3,14 @@ import AnimatedIcon from './AnimatedIcon'
...
-const NormalTextIcon = ({ fill, size, style = {}, cssRaw, animated, animationComplete }: IconType) => {
+const NormalTextIcon = ({ fill, size = 18, style = {}, cssRaw, animated, animationComplete }: IconType) => {
   return (
     <AnimatedIcon {...{ fill, size, style, cssRaw, animated, animationData, animationComplete }}>
       <svg
         xmlns='http://www.w3.org/2000/svg'
         viewBox='0 0 24 24'
         fill='none'
-        style={{ ...style, width: '100%', height: '100%' }}
+        style={{ ...style, width: '100%', height: '100%', transform: 'translate(0, -3%) scale(1, 1)' }}
```
raineorshine, line 13: `scale(1, 1)` is the default, so this shouldn't be needed.

raineorshine (changes requested):
This looks great! Thanks for the detailed comments and careful attention to detail.

ohrytskov:
All required adjustments have been made, and the PR is now ready for review.

src/components/RedoIcon.tsx
raineorshine: Okay, thanks. They were all made by the same designer to the same specifications, so I'm not sure why some icons would need to be scaled up while others need to be scaled down. But since none of them are scaled more than 5–10%, I don't think the difference in stroke width will be noticeable. They look good to me across font sizes now, so it seems to be working.

src/components/icons/LottieAnimation.tsx
```diff
@@ -1,28 +1,157 @@
...
+import StaticColor from '../../@types/lottie/StaticColor'
 
+/** The props for the LottieAnimation component. */
 interface LottieAnimationProps {
-  animationData: any
+  animationData: LottieData | null
   speed?: number
+  color?: string
   onComplete?: () => void
   style?: React.CSSProperties
```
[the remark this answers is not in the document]
raineorshine: Fixed in 83137f68ad3

src/components/icons/LottieAnimation.tsx
```diff
@@ -35,15 +164,26 @@ const LottieAnimation: React.FC<LottieAnimationProps> = ({
...
+  if (!animationDataWithColor) {
+    console.warn('animationData is not provided.')
+    return null // Early return if animationData is null
+  }
 
   return (
     <Player
-      style={style}
-      animationData={animationData}
+      key={key} // Force re-render to apply theme changes, refreshing cached animation data with updated color settings
```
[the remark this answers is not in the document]
raineorshine: @ohrytskov Thoughts about `key`? https://github.com/cybersemics/em/pull/2565/files#r1862679211

src/components/icons/LottieAnimation.tsx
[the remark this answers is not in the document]
raineorshine: Fixed in 83137f68ad3

raineorshine:
Thanks! Just let me know your thoughts about https://github.com/cybersemics/em/pull/2565#discussion_r1862679211

ohrytskov:
The key is no longer necessary as animationData now updates with a fresh reference whenever it changes, so it has been removed.

PR #2566, opened 2024-11-09
Colors to design tokens

yangchristina:
#2170 

I added `transparent` as a design token to replace `rgba(255, 255, 255, 0)`, which seemed quite unnecessary since using just 'transparent' would be the same thing. Let me know if it's better to get rid of it.

I moved the styles on body meant to prevent white flashes onto html, since it breaks light theme. Also you may get dark flashes on light theme if you change it programmatically instead of using the toggle, since it depends on localStorage
The problem with this still is that if you change the theme, then scroll, the html bg (at the edges) won't change, you have to refresh for that.

### DIFF
--- a/src/colors.config.ts
+++ b/src/colors.config.ts
@@ -47,6 +47,9 @@ const colors = {
     transparent: 'transparent',
     bgTransparent: 'rgba(0, 0, 0, 0)',
     fgTransparent: 'rgba(255, 255, 255, 0)',
+    modalColor: 'rgba(227, 227, 227, 1)', // #e3e3e3
+    quickDropBgHover: 'rgba(40,40,40,0.8)',
+    quickDropBg: 'rgba(30,30,30,0.8)',
   },
   light: {
     // Background colors in capacitor app needs to be in hexadecimal codes
@@ -96,6 +99,9 @@ const colors = {
     transparent: 'transparent',
     fgTransparent: 'rgba(0, 0, 0, 0)',
     bgTransparent: 'rgba(255, 255, 255, 0)',
+    modalColor: 'rgba(28, 28, 28, 1)', // #1c1c1c
+    quickDropBgHover: 'rgba(215, 215, 215, 0.8)',
+    quickDropBg: 'rgba(225, 225, 225, 0.8)',
   },
 } as const
 
--- a/src/colors.config.ts
+++ b/src/colors.config.ts
@@ -81,7 +81,7 @@ const colors = {
     gray15: 'rgba(38, 38, 38, 1)', // #262626
     gray33: 'rgba(85, 85, 85, 1)',
     gray50: 'rgba(128, 128, 128, 1)', // #808080 (gray)
-    gray66: 'rgba(169, 169, 169, 1)', // #a9a9a9
+    gray66: 'rgba(169, 169, 169, 1)', // #a9a9a9 (darkgray)
     gray: 'rgba(169, 169, 169, 1)', // #a9a9a9
     green: 'rgba(0, 214, 136, 1)', // #00d688
     highlight: 'rgba(65, 105, 225, 1)', // #4169e1 (royalblue)
--- a/src/colors.config.ts
+++ b/src/colors.config.ts
@@ -29,13 +29,14 @@ const colors = {
     vividHighlight: '#63c9ea',
     white: 'rgba(255, 255, 255, 1)',
     yellow: 'rgba(255, 208, 20, 1)', // #ffd014
-    inputBorder: 'rgba(153, 153, 153, 1)', // #999
+    inputBorder: 'rgba(153, 153, 153, 1)', // #999, also used in navBar
+    breadcrumbs: 'rgba(153, 153, 153, 1)',
     link: 'rgba(135, 206, 235, 1)', // #87ceeb
     dim: 'rgba(255, 255, 255, 0.5)',
     dimInverse: 'rgba(7, 7, 7, 0.5)',
     bullet: 'rgba(217, 217, 217, 1)',
     codeBg: 'rgba(51, 51, 51, 1)', // #333333
-    modalExportUnused: 'rgba(68, 68, 68, 1)', // #444
+    modalExportUnused: 'rgba(68, 68, 68, 1)', // #444, used in dropdown menu
     divider: 'rgba(204, 204, 204, 1)',
     bgMuted: 'rgba(51, 51, 51, 1)', // #333
     footerBg: 'rgba(26, 26, 26, 1)', // #1a1a1a
@@ -50,6 +51,12 @@ const colors = {
     modalColor: 'rgba(227, 227, 227, 1)', // #e3e3e3
     quickDropBgHover: 'rgba(40,40,40,0.8)',
     quickDropBg: 'rgba(30,30,30,0.8)',
+    bulletGray: 'rgba(102, 102, 102, 1)', // #666
+    midPink: 'rgba(255, 123, 195, 1)', // #ff7bc3
+    purpleEggplant: '#32305f', // purple-eggplant
+    commandSelected: 'rgba(33, 33, 33, 1)', // #212121
+    eggplant: 'rgba(82, 48, 95, 1)', // eggplant
+    checkboxForm: 'rgba(62, 62, 62, 1)', // #3e3e3e
   },
   light: {
     // Background colors in capacitor app needs to be in hexadecimal codes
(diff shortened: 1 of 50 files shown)


raineorshine:
> I added `transparent` as a design token to replace `rgba(255, 255, 255, 0)`, which seemed quite unnecessary since using just 'transparent' would be the same thing. Let me know if it's better to get rid of it.

That would behave differently with `transition: color`.

yangchristina:
> > I added `transparent` as a design token to replace `rgba(255, 255, 255, 0)`, which seemed quite unnecessary since using just 'transparent' would be the same thing. Let me know if it's better to get rid of it.
> 
> That would behave differently with `transition: color`.

Ah I see, that makes sense, thanks!

raineorshine:
I vaguely recall that I needed to use the alpha channel for fading out thoughts (years ago) because I was running into z-indexing issues with opacity. However, that is probably no longer the case with the new layout engine that renders all thoughts as siblings in the DOM. 

So it's possible that is no longer needed but we would have to analyze it to determine if that’s the case. Probably best to keep it as-is for this conversion and then refactor later.

yangchristina:
I believe I've gotten the hex and rgb values outside of tests.
@raineorshine should I convert named colors into design tokens too?

raineorshine:
Yes, please!

yangchristina:
The icon disappears when active, didn't see any hardcoded colors, so maybe it's svg related?
Screenshot 2024-11-09 at 2 24 21 AM

Also clicking an icon in the toolbar is causing it to disappear for a second and reappear (weird flash), haven't been able to pinpoint where it is coming from yet.

yangchristina:
I think I've got all the named colors, but not fully sure if I missed any.

yangchristina:
I changed both colors `transparent` and `rgba(255,255,255,0)` to use the transparent color token, which may not be intended behaviour. 
@raineorshine would you prefer I make two transparent variables to deal with transition? However I'm also assuming plain `transparent` variables wouldn't have to deal with transition anyway (otherwise they should be using rgba), so it wouldn't affect anything.

raineorshine:
I think `transparent` and `rgba(255,255,255,0)` are different things and should use different tokens.

I just did a quick browser test, and apparently a literal `transparent` value *can* animate with `transition: color`. For example `white` → `transparent` behaves similarly to `rgba(255,255,255,1)` → `rgba(255,255,255,0)`. 

I think it would be desirable to give them separate design tokens and not make additional changes at this time. We can analyze their use in color animations and consider simplifying or refactoring at another time.

yangchristina:
> I think `transparent` and `rgba(255,255,255,0)` are different things and should use different tokens.
> 
> I just did a quick browser test, and apparently a literal `transparent` value _can_ animate with `transition: color`. For example `white` → `transparent` behaves similarly to `rgba(255,255,255,1)` → `rgba(255,255,255,0)`.
> 
> I think it would be desirable to give them separate design tokens and not make additional simplifications at this time. We can analyze their use in color animations and consider refactoring at another time.

Sounds good, I've split transparent into 3 tokens, transparent, bgTransparent, and fgTransparent

src/selectors/theme.ts
```diff
@@ -23,6 +23,7 @@ import publishMode from '../util/publishMode'
 
 /** Gets the theme. */
 const theme = (state: State) => {
+  // return 'Dark'
```
trevinhofmann, line 26: Is this comment intentional? I assume it was added for debugging and not meant to be included.

trevinhofmann (changes requested):
Thank you for the PR, @yangchristina!

---

## 1: Breadcrumbs are too light in Light Mode

The breadcrumb text seems to have been inadvertently brightened in Light Mode with this change, making them very difficult to read.

Screenshot 2024-11-13 at 1 05 14 AM

---

## 2: Toolbar action information text is illegible in Light Mode

Either the box or the text inside of it needs to be made lighter, because they are both currently dark in Light Mode which makes it illegible.

Screenshot 2024-11-13 at 1 12 20 AM

src/selectors/theme.ts
yangchristina: You are right, thanks for catching this!

yangchristina:
@trevinhofmann thanks! I wasn't originally going to fix light mode in this PR, since that's part of the next milestone, but I've fixed it for all the gray colors for now. Checking the rest will take some more time, if you think having them all done now is beneficial, let me know.

src/colors.config.ts
```diff
@@ -3,18 +3,21 @@ const colors = {
...
     fg: 'rgba(255, 255, 255, 1)',
     fgOverlay80: 'rgba(20, 20, 20, 0.8)',
     fgOverlay90: 'rgba(20, 20, 20, 0.9)',
+    fgOverlay50: 'rgba(255, 255, 255, 0.5)',
     gray15: 'rgba(38, 38, 38, 1)', // #262626
     gray33: 'rgba(85, 85, 85, 1)',
     gray50: 'rgba(128, 128, 128, 1)', // #808080 (gray)
     gray66: 'rgba(169, 169, 169, 1)', // #a9a9a9
-    gray: 'rgba(169, 169, 169, 1)', // #a9a9a9
+    gray: 'rgba(169, 169, 169, 1)', // #a9a9a9, this is used for disabled things + text color, unlike gray66 it is the same for both light and dark
```
trevinhofmann, line 20: Panda tokens have support for both a `value` and `description`. It would require some refactoring (`convertColorsToPandaCSS`), but do you think it would be worth considering using the `description` instead of comments for notes like this one?

This is just a question, not a blocking issue.

src/colors.config.ts
```diff
@@ -26,22 +29,59 @@ const colors = {
...
+    thoughtAnnotation: 'rgba(34, 34, 34, 1)', // #222
+    transparent: 'transparent',
+    bgTransparent: 'rgba(0, 0, 0, 0)',
+    fgTransparent: 'rgba(255, 255, 255, 0)',
+    modalColor: 'rgba(227, 227, 227, 1)', // #e3e3e3
+    quickDropBgHover: 'rgba(40,40,40,0.8)',
+    quickDropBg: 'rgba(30,30,30,0.8)',
+    bulletGray: 'rgba(102, 102, 102, 1)', // #666
+    midPink: 'rgba(255, 123, 195, 1)', // #ff7bc3
+    purpleEggplant: '#32305f', // purple-eggplant
```
trevinhofmann, line 57:
```suggestion
    purpleEggplant: '#32305f',
```

src/colors.config.ts
```diff
@@ -26,22 +29,59 @@ const colors = {
...
+    commandSelected: 'rgba(33, 33, 33, 1)', // #212121
+    eggplant: 'rgba(82, 48, 95, 1)', // eggplant
```
trevinhofmann, line 59:
```suggestion
    eggplant: 'rgba(82, 48, 95, 1)',
```

src/colors.config.ts
```diff
@@ -54,6 +94,40 @@ const colors = {
...
+    thoughtAnnotation: 'rgba(221, 221, 221, 1)', // #ddd
+    transparent: 'transparent',
+    fgTransparent: 'rgba(0, 0, 0, 0)',
+    bgTransparent: 'rgba(255, 255, 255, 0)',
+    modalColor: 'rgba(28, 28, 28, 1)', // #1c1c1c
+    quickDropBgHover: 'rgba(215, 215, 215, 0.8)',
+    quickDropBg: 'rgba(225, 225, 225, 0.8)',
+    bulletGray: 'rgba(153, 153, 153, 1)', // #999999
+    midPink: 'rgba(255, 123, 195, 1)',
+    purpleEggplant: '#32305f', // purple-eggplant
```
trevinhofmann, line 122:
```suggestion
    purpleEggplant: '#32305f',
```

src/colors.config.ts
```diff
@@ -54,6 +94,40 @@ const colors = {
...
+    commandSelected: 'rgba(222, 222, 222, 1)',
+    eggplant: 'rgb(85, 51, 98)', // eggplant
```
trevinhofmann, line 124:
```suggestion
    eggplant: 'rgb(85, 51, 98)',
```

trevinhofmann (approved):
> I wasn't originally going to fix light mode in this PR, since that's part of the next milestone, but I've fixed it for all the gray colors for now.

Thank you! If there is light mode work planned in the next milestone, I don't think we need to make those changes now. I just noticed and reported (1) and (2) as they seemed to be regressions which caused some text to be illegible.

---

LGTM, thank you @yangchristina!

Sorry, I missed a few of the redundant color name comments in my last review if you wouldn't mind removing that also. I don't want to hold up the PR over these, though.

src/colors.config.ts
yangchristina: I think this would be worth doing, so also moving things so it is:
```js
gray: {
  dark: ...,
  light: ...,
  description: ...,
}
```
I can refactor this in a future PR @raineorshine if you'd like

raineorshine: I'm not sure. PandaCSS descriptions are nice in theory, but do not have any editor support as far as I know. JSDOC seems more reliable and widely supported.

raineorshine (approved):
Looks good, thank you!

This is a big step towards more consistent theming in **em**.

Issue #2182, opened 2024-07-23
Upgrade @testing-library/react


raineorshine:
Upgrade the `@testing-library/react` dependency from `^12.1.5` to the latest `16.x`. 

This is a breaking change and requires upgrading syntax/fixing broken tests.

PR #2568, opened 2024-11-10
Upgrade testing library react to v14 #2182

mstrofbass:
Partial progress on upgrading react-testing-library to the latest version. This gets us to v14.

The upgrade from v14 to v15/v16 is a bit more challenging and taking additional time to troubleshoot so I wanted to get this merged.

One of the upgrades (to v13, I believe), switched to the new `createRoot` thing that results in some operations switching from synchronous to asynchronous, necessitating the forwarding of the fake timers in a few places. This may be a temporary change.

Note that some dependencies may have been updated to be peer dependencies, which may necessitate the removal of the `node_modules` directory prior to running `yarn install`.

Relates to #2182

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,7 @@
     "@sinonjs/fake-timers": "^13.0.4",
     "@testing-library/dom": "^10.4.0",
     "@testing-library/jest-dom": "^6.6.2",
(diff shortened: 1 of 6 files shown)


raineorshine (approved):
Thank you!

PR #2570, opened 2024-11-10
Upgrade react testing library to v16 #2182

### LINKED ISSUE #2182: Upgrade @testing-library/react

mstrofbass:
**IMPORTANT**: This upgrade changed some of the react-testing-library dependencies to be peer dependencies. This may require the deletion of the `node_modules` folder prior to doing `yarn install` in existing installations. It also required deleting and recreating the `yarn.lock` file. 

Upgrades @testing-library/react to latest (v16.0.1).

Closes #2182

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,7 @@
     "@sinonjs/fake-timers": "^13.0.4",
     "@testing-library/dom": "^10.4.0",
     "@testing-library/jest-dom": "^6.6.2",
-    "@testing-library/react": "^15.0.0",
+    "@testing-library/react": "^16.0.1",
     "@testing-library/user-event": "^14.5.2",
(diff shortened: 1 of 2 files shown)


mstrofbass:
I would recommend someone pull this down, delete their `node_modules` directory, run `yarn install` and then make sure the tests run.

They ran for me but I ran into so many seemingly disparate problems that I'd like someone to validate that it works for them

raineorshine (approved):
This worked for me with both an in-place `yarn install` and a `yarn install` after deleting `node_modules`.

Issue #2526, opened 2024-11-02
Animation: Fade out unmounted thought


raineorshine:
The example here uses a collapsing context, but the fade out animation should be the default for any thought node that becomes unmounted.

## Steps to Reproduce

```
- a
  - b
- c
```

1. Set the cursor on `a`.
2. Set the cursor to null.

## Current Behavior

`b` disappears immediately.

https://github.com/user-attachments/assets/86cda20a-dbc2-42a8-8c5c-e41cadf80174

## Expected Behavior

`b` should fade out with `veryFastDuration` and `ease-out`.

## Note

The collapsed node is unmounted, so there is currently nothing to animate. You will need the [CSSTransition](https://reactcommunity.org/react-transition-group/css-transition) component which you can see used throughout the project to make the thought stay mounted until the animation completes.

PR #2581, opened 2024-11-12
Animation: Fade out unmounted thought

zukilover:
#2526

This update implements the use of `CSSTransition` to apply the `.fade` CSS classes when items are unmounted. 

Please note that the current implementation uses the existing `.fade` classes, which have a transition duration of `200ms`, while the `timeout` for the transition is set to `veryFastDuration` (80ms).

https://github.com/cybersemics/em/blob/9e7f8a1e438468f3b1b8f330ffa8701f39074635/src/App.css#L21-L24

Let me know if you'd prefer to create a new CSS class to match the durations more closely.

### DIFF
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -464,7 +464,9 @@ const TreeNode = ({
     >
       <FadeTransition
         id={thoughtKey}
-        duration='nodeOpacity'
+        // The FadeTransition is only responsible for fade out on unmount.
+        // See autofocusChanged for normal opacity transition.
+        duration='nodeFadeOut'
         nodeRef={fadeThoughtRef}
         in={transitionGroupsProps.in}
         unmountOnExit
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -412,6 +412,7 @@ const TreeNode = ({
   autofocusDepth: number
 } & Pick<CSSTransitionProps, 'in'>) => {
   const [y, setY] = useState(_y)
+  const fadeThoughtRef = useRef<HTMLDivElement>(null)
 
   useLayoutEffect(() => {
     if (y !== _y) {
@@ -437,7 +438,6 @@ const TreeNode = ({
   const xCol2 = isTableCol1 ? nextThought?.x || previousThought?.x || 0 : 0
   // Increasing margin-right of thought for filling gaps and moving the thought to the left by adding negative margin from right.
   const marginRight = isTableCol1 ? xCol2 - (width || 0) - x - (bulletWidth || 0) : 0
-  const fadeThoughtRef = useRef<HTMLDivElement>(null)
 
   return (
     <div
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -1,6 +1,8 @@
 import _ from 'lodash'
 import React, { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react'
 import { useSelector } from 'react-redux'
+import { TransitionGroup } from 'react-transition-group'
+import { CSSTransitionProps } from 'react-transition-group/CSSTransition'
 import { css } from '../../styled-system/css'
 import Autofocus from '../@types/Autofocus'
 import Index from '../@types/IndexType'
@@ -39,6 +41,7 @@ import parseLet from '../util/parseLet'
 import safeRefMerge from '../util/safeRefMerge'
 import unroot from '../util/unroot'
 import DropEnd from './DropEnd'
+import FadeTransition from './FadeTransition'
 import HoverArrow from './HoverArrow'
 import VirtualThought, { OnResize } from './VirtualThought'
 
@@ -395,6 +398,7 @@ const TreeNode = ({
   cliffPaddingStyle,
   dragInProgress,
   autofocusDepth,
+  ...transitionGroupsProps
 }: TreeThoughtPositioned & {
   thoughtKey: string
   index: number
@@ -406,7 +410,7 @@ const TreeNode = ({
   cliffPaddingStyle: { paddingBottom: number }
   dragInProgress: boolean
   autofocusDepth: number
-}) => {
+} & Pick<CSSTransitionProps, 'in'>) => {
   const [y, setY] = useState(_y)
 
   useLayoutEffect(() => {
@@ -433,6 +437,7 @@ const TreeNode = ({
   const xCol2 = isTableCol1 ? nextThought?.x || previousThought?.x || 0 : 0
   // Increasing margin-right of thought for filling gaps and moving the thought to the left by adding negative margin from right.
   const marginRight = isTableCol1 ? xCol2 - (width || 0) - x - (bulletWidth || 0) : 0
+  const fadeThoughtRef = useRef<HTMLDivElement>(null)
 
   return (
     <div
@@ -457,31 +462,35 @@ const TreeNode = ({
         textAlign: isTableCol1 ? 'right' : undefined,
       }}
     >
-      <VirtualThought
-        debugIndex={testFlags.simulateDrop ? indexChild : undefined}
-        depth={depth}
-        dropUncle={thoughtId === cursorUncleId}
-        env={env}
-        indexDescendant={indexDescendant}
-        // isMultiColumnTable={isMultiColumnTable}
-        isMultiColumnTable={false}
-        leaf={leaf}
-        onResize={setSize}
-        path={path}
-        prevChildId={prevChild?.id}
-        showContexts={showContexts}
-        simplePath={simplePath}
-        singleLineHeight={singleLineHeightWithCliff}
-        // Add a bit of space after a cliff to give nested lists some breathing room.
-        // Do this as padding instead of y, otherwise there will be a gap between drop targets.
-        // In Table View, we need to set the cliff padding on col1 so it matches col2 padding, otherwise there will be a gap during drag-and-drop.
-        style={cliff < 0 || isTableCol1 ? cliffPaddingStyle : undefined}
-        crossContextualKey={thoughtKey}
-        prevCliff={treeThoughtsPositioned[index - 1]?.cliff}
-        isLastVisible={isLastVisible}
-        autofocus={autofocus}
-        marginRight={isTableCol1 ? marginRight : 0}
-      />
+      <FadeTransition id={thoughtKey} duration='nodeOpacity' nodeRef={fadeThoughtRef} in={transitionGroupsProps.in}>
+        <div ref={fadeThoughtRef}>
+          <VirtualThought
+            debugIndex={testFlags.simulateDrop ? indexChild : undefined}
+            depth={depth}
+            dropUncle={thoughtId === cursorUncleId}
+            env={env}
+            indexDescendant={indexDescendant}
+            // isMultiColumnTable={isMultiColumnTable}
+            isMultiColumnTable={false}
+            leaf={leaf}
+            onResize={setSize}
+            path={path}
+            prevChildId={prevChild?.id}
+            showContexts={showContexts}
+            simplePath={simplePath}
+            singleLineHeight={singleLineHeightWithCliff}
+            // Add a bit of space after a cliff to give nested lists some breathing room.
+            // Do this as padding instead of y, otherwise there will be a gap between drop targets.
+            // In Table View, we need to set the cliff padding on col1 so it matches col2 padding, otherwise there will be a gap during drag-and-drop.
+            style={cliff < 0 || isTableCol1 ? cliffPaddingStyle : undefined}
+            crossContextualKey={thoughtKey}
+            prevCliff={treeThoughtsPositioned[index - 1]?.cliff}
+            isLastVisible={isLastVisible}
+            autofocus={autofocus}
+            marginRight={isTableCol1 ? marginRight : 0}
+          />
+        </div>
+      </FadeTransition>
 
       {/* DropEnd (cliff) */}
       {dragInProgress &&
@@ -907,26 +916,28 @@ const LayoutTree = () => {
           marginRight: `${-indent + (isTouch ? 2 : -1)}em`,
         }}
       >
-        {treeThoughtsPositioned.map((thought, index) => (
-          <TreeNode
-            {...thought}
-            index={index}
-            // Pass unique key for the component
-            key={thought.key}
-            // Pass the thought key as a thoughtKey and not key property as it will conflict with React's key
-            thoughtKey={thought.key}
-            {...{
-              viewportBottom,
-              treeThoughtsPositioned,
-              bulletWidth,
-              cursorUncleId,
-              setSize,
-              cliffPaddingStyle,
-              dragInProgress,
-              autofocusDepth,
-            }}
-          />
-        ))}
+        <TransitionGroup>
+          {treeThoughtsPositioned.map((thought, index) => (
+            <TreeNode
+              {...thought}
+              index={index}
+              // Pass unique key for the component
+              key={thought.key}
+              // Pass the thought key as a thoughtKey and not key property as it will conflict with React's key
+              thoughtKey={thought.key}
+              {...{
+                viewportBottom,
+                treeThoughtsPositioned,
+                bulletWidth,
+                cursorUncleId,
+                setSize,
+                cliffPaddingStyle,
+                dragInProgress,
+                autofocusDepth,
+              }}
+            />
+          ))}
+        </TransitionGroup>
       </div>
     </div>
   )
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -462,7 +462,13 @@ const TreeNode = ({
         textAlign: isTableCol1 ? 'right' : undefined,
       }}
     >
-      <FadeTransition id={thoughtKey} duration='nodeOpacity' nodeRef={fadeThoughtRef} in={transitionGroupsProps.in}>
+      <FadeTransition
+        id={thoughtKey}
+        duration='nodeOpacity'
+        nodeRef={fadeThoughtRef}
+        in={transitionGroupsProps.in}
+        unmountOnExit
+      >
         <div ref={fadeThoughtRef}>
           <VirtualThought
             debugIndex={testFlags.simulateDrop ? indexChild : undefined}
--- a/src/components/VirtualThought.tsx
+++ b/src/components/VirtualThought.tsx
@@ -94,7 +94,7 @@ const VirtualThought = ({
   const cursorLeaf = useSelector(state => !!state.cursor && !hasChildren(state, head(state.cursor)))
   const cursorDepth = useSelector(state => (state.cursor ? state.cursor.length : 0))
   const fontSize = useSelector(state => state.fontSize)
-  const note = useSelector(state => noteValue(state, thought.id))
+  const note = useSelector(state => noteValue(state, thought?.id))
   const ref = useRef<HTMLDivElement>(null)
 
   /***************************
@@ -149,11 +149,11 @@ const VirtualThought = ({
     onResize?.({
       height: heightNew,
       width: widthNew,
-      id: thought.id,
+      id: thought?.id,
       isVisible: isVisibleNew,
       key: crossContextualKey,
     })
-  }, [crossContextualKey, onResize, thought.id, autofocus])
+  }, [crossContextualKey, onResize, thought?.id, autofocus])
 
   // Recalculate height when anything changes that could indirectly affect the height of the thought. (Height observers are slow.)
   // Autofocus changes when the cursor changes depth or moves between a leaf and non-leaf. This changes the left margin and can cause thoughts to wrap or unwrap.
@@ -183,7 +183,7 @@ const VirtualThought = ({
   // TODO: useLayoutEffect does not work for some reason, causing the thought to briefly render at the incorrect height.
   const value = useSelector(state => {
     const thoughtId = head(simplePath)
-    return thoughtId ? getThoughtById(state, thoughtId).value : null
+    return thoughtId ? getThoughtById(state, thoughtId)?.value : null
   })
   useEffect(updateSize, [updateSize, value])
 
@@ -191,11 +191,11 @@ const VirtualThought = ({
   useEffect(
     () => {
       return () => {
-        onResize?.({ height: null, width: null, id: thought.id, isVisible: true, key: crossContextualKey })
+        onResize?.({ height: null, width: null, id: thought?.id, isVisible: true, key: crossContextualKey })
       }
     },
     // these should be memoized and not change for the life of the component, so this is effectively componentWillUnmount
-    [crossContextualKey, onResize, thought.id],
+    [crossContextualKey, onResize, thought?.id],
   )
 
   // Short circuit if thought has already been removed.
--- a/src/recipes/fadeTransition.ts
+++ b/src/recipes/fadeTransition.ts
@@ -29,7 +29,6 @@ const fadeTransitionRecipe = defineSlotRecipe({
         exitActive: { transition: `opacity {durations.slow} ease 0ms` },
       },
       nodeOpacity: {
-        enterActive: { transition: `opacity {durations.nodeOpacity} cubic-bezier(0.4, 0, 0.6, 1)` },
         exitActive: { transition: `opacity {durations.nodeOpacity} ease-out` },
       },
     },
--- a/src/recipes/fadeTransition.ts
+++ b/src/recipes/fadeTransition.ts
@@ -28,9 +28,9 @@ const fadeTransitionRecipe = defineSlotRecipe({
         enterActive: { transition: `opacity {durations.distractionFreeTyping} ease 0ms` },
         exitActive: { transition: `opacity {durations.slow} ease 0ms` },
       },
-      nodeOpacity: {
+      nodeFadeOut: {
         enter: { opacity: 1 },
-        exitActive: { transition: `opacity {durations.nodeOpacity} ease-out` },
+        exitActive: { transition: `opacity {durations.nodeFadeOut} ease-out` },
       },
     },
   },
--- a/src/recipes/fadeTransition.ts
+++ b/src/recipes/fadeTransition.ts
@@ -8,6 +8,7 @@ const fadeTransitionRecipe = defineSlotRecipe({
     enterActive: { opacity: 1 },
     exit: { opacity: 1 },
     exitActive: { opacity: 0 },
+    exitDone: { opacity: 0 },
   },
   variants: {
     duration: {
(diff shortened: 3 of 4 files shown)


raineorshine:
Thanks! I'll have to see what duration feels best in this scenario, but presumably we always want the CSSTransition timeout to match the animation duration.

Before we can review, there are a couple things that need to be addressed:

- Please reference the related issue number in the PR description.
- CI is failing

raineorshine:
The Unit tests pass! We will look into the flaky test.

The Puppeteer test is still failing after three attempts, so there might be more going on there. However, we can do QA first to ensure that it is working as intended.

zukilover:
@raineorshine I pushed a new update that should fix the puppeteer test.

src/components/VirtualThought.tsx
```diff
@@ -149,11 +149,11 @@ const VirtualThought = ({
     onResize?.({
       height: heightNew,
       width: widthNew,
-      id: thought.id,
+      id: thought?.id,
```
trevinhofmann, line 152: @raineorshine - It would be worth considering enabling `noUncheckedIndexedAccess` for TS. We currently look up the thought with `state.thoughts.thoughtIndex[id]`, and TS assumes that the index exists. With the flag enabled, the type of `thought` would be corrected from `Thought` to `Thought | undefined`.

trevinhofmann (approved):
Thank you, @zukilover! This fix is working as expected:

https://github.com/user-attachments/assets/e453a1ce-266a-4b44-901e-4722a4d26549

raineorshine:
Thanks @trevinhofmann!

I ran out of time this morning, but will review and test soon.

src/components/VirtualThought.tsx
raineorshine: Yes, that's a good idea. We will have to consider the impact of that, as it could touch a lot of files. Would definitely help with `state.thoughts.thoughtIndex[id]`.

raineorshine (changes requested):
I tested and can confirm the new unmount animation. Looks good!

I added a `very-fast-fade` animation class to match the `veryFastDuration`.

This PR effectively adds both a fade-in and fade-out animation via the CSSTransition. However, thoughts already faded in on mount. This makes me think that we are applying a redundant fade-in animation somewhere else. This will have an impact on performance, and it's probably better from a code design perspective to avoid redundant animations. Can you find the old fade-in animation and suggest a way to eliminate the redundancy.

Lastly, we should confirm the easing function and duration of the old fade-in animation and make sure that is preserved.

Thanks!

zukilover:
The existing fade animation likely come from these lines:

https://github.com/cybersemics/em/blob/560ff541f3f5ddefe95320bf144f51b0c2b989cf/src/components/Subthought.tsx#L127-L130

I will check further.

zukilover:
@raineorshine  I reviewed the line here:

https://github.com/cybersemics/em/blob/560ff541f3f5ddefe95320bf144f51b0c2b989cf/src/components/Subthought.tsx#L127-L130

The thought does indeed fade in on mount, but this effect is immediately overridden by the `tree-node`'s `very-fast-fade-enter` animation when the component mounts. In other words, it shouldn't impact the initial mount behavior on the first render.

Additionally, the specific line mentioned controls the opacity when the autofocus changes on a thought. Please let me know if this aligns with your understanding.

raineorshine (approved):
Thanks for investigating. So if I understand correctly, you're saying that these two opacity transitions are completely independent. 

- The CSSTransition controls the thought opacity when mounting/unmounting.
- The Subthought controls the thought opacity when autofocus changes.

raineorshine:
I removed the opacity effect so the CSSTransition can now do all the work. (The only thing I changed behaviorally was change 80ms ease-out to 150ms linear, but that is just a slight tweak to the appearance.) `isEmpty` had to be added to `TreeThought` to preserve the behavior that that empty thoughts snap in instead of animate, since the animation shifted from VirtualThought to LayoutTree. However, the complete removal of the useEffect is an overall decrease in complexity.

@zukilover @trevinhofmann I would appreciate both of your reviews to confirm that what I did makes sense, and if you think this is ready to merge. Thanks!

zukilover:
Thanks for the update, @raineorshine. I've resolved the conflict. @trevinhofmann please let me know if this looks good to you.

raineorshine:
#2595 was just merged which eliminated App.css, creating a nontrivial merge conflict. `very-fast-fade` needs to be converted to the new `FadeTransition`. I tried, but I can't seem to get the animation to work correctly. 

https://github.com/user-attachments/assets/bd41cc83-afbe-405b-aa0d-f52390dd824a

@yangchristina Could you help with this?

I'm also seeing an error in the console (before merge), but we should probably resolve the merge first before investigating:

Brave Browser 2024-11-18 18 18 23

src/components/LayoutTree.tsx
```diff
@@ -878,6 +893,9 @@ const LayoutTree = () => {
   // Subtract singleLineHeight since we can assume that the last rendered thought is within the viewport. (It would be more accurate to use its exact rendered height, but it just means that there may be slightly more space at the bottom, which is not a problem. The scroll position is only forced to change when there is not enough space.)
   const spaceBelow = viewportHeight - navAndFooterHeight - CONTENT_PADDING_BOTTOM - singleLineHeight
 
+  // Create a ref for each node. This helps to avoid the `findDOMNode is deprecated` warning.
+  //const nodeRef = createRef<HTMLDivElement>()
```
trevinhofmann, line 897: Will this remain unused, or is this a WIP?

If `nodeRef` is not needed, it would be better to remove it instead of leaving it commented.

zukilover:
@yangchristina @raineorshine I've implemented the new `FadeTransition` and added new durations: `no` (0) and `veryFast`. Let me know what you think of the updates.

src/App.css
raineorshine, line 1: App.css should be removed completely now.

src/components/LayoutTree.tsx
```diff
@@ -878,6 +893,8 @@ const LayoutTree = () => {
   // Subtract singleLineHeight since we can assume that the last rendered thought is within the viewport. (It would be more accurate to use its exact rendered height, but it just means that there may be slightly more space at the bottom, which is not a problem. The scroll position is only forced to change when there is not enough space.)
   const spaceBelow = viewportHeight - navAndFooterHeight - CONTENT_PADDING_BOTTOM - singleLineHeight
 
+  // Create a ref for each node. This helps to avoid the `findDOMNode is deprecated` warning.
```
raineorshine, line 896: I'm still getting the warning on 65274eada58.

raineorshine (changes requested):
> @yangchristina @raineorshine I've implemented the new `FadeTransition` and added new durations: `no` (0) and `veryFast`. Let me know what you think of the updates.

Thank you @zukilover! That's exactly what I had in mind.

raineorshine:
Looks like the fade animation is not working. Tested e3311159edb516b29ea90a848c4344279f1cf20e.

zukilover:
@raineorshine, the issue appears to be caused by the tree-node already having top and left transitions defined here:
https://github.com/cybersemics/em/blob/8760cae0efa8ecd9dd4760dfc4fd7b6f75da43e3/src/components/LayoutTree.tsx#L442-L444

This conflicted with the `opacity` transition from the FadeTransition component, preventing it from working properly.
To fix this, I updated the transition property to `all` and ensured the duration matched the fade animation.

I also tried wrapping the `tree-node` in a `div` to separate the fade transition from the `tree-node` transition, but this approach didn't work. Only the `opacity` transition was applied, and the `top` and `left` transitions were ignored.

Also, @yangchristina, I added a base class `appearDone` to handle the initial mounting of the thought. Without this, it was producing an `undefined` class.

zukilover:
This is the latest result showing opacity worked with the changes above:

fade-frame

src/components/LayoutTree.tsx
```diff
@@ -361,175 +366,185 @@ const linearizeTree = (
...
+      cliffPaddingStyle,
+      dragInProgress,
+      autofocusDepth,
+    },
+    ref,
+  ) => {
+    const [y, setY] = useState(_y)
+
+    useLayoutEffect(() => {
+      if (y !== _y) {
```
trevinhofmann, line 425: > If the new value you provide is identical to the current `state`, as determined by an [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will **skip re-rendering the component and its children**. This is an optimization. Although in some cases React may still need to call your component before skipping the children, it shouldn’t affect your code.

https://react.dev/reference/react/useState#setstate-caveats

I still think the `if (y !== _y)` check is unnecessary because React will not re-render if `y === _y` regardless. However, it doesn't hurt to be explicit with the intent. I am okay with this as-is.

trevinhofmann (changes requested):
Thank you, @zukilover.

The fade out animation appears to be working correctly:

https://github.com/user-attachments/assets/c15c8b3f-d352-457c-a35b-15183e784ac1

However, I am still seeing the `findDOMNode` warning in the console on 088a57c6db173c8183a32927ed6e355a58c861a1.

Screenshot 2024-11-21 at 3 21 44 AM

raineorshine (changes requested):
Thanks for the update!

I've tested it out, and I can see that the animation timings have changed quite a bit. I understand that you have some constraints to work with, but we will need to keep the node transition and fade-in the same as it is on `main` (timing and easing function). They have been carefully tuned from real-world use. The only thing that should visibly change is the new fade out animation on unmount.

zukilover:
@raineorshine in this case, I've added back the original layout duration which was `layoutNodeAnimationDuration`. 

While we could make the tree node's top/left transition independent of the fade in/out effect for the thought by wrapping the `VirtualThought` with the `FadeTransition`, this wouldn’t be the ideal approach. The reason is that the tree-node won't be unmounted after fading out.

I hope this makes sense.

trevinhofmann (approved):
Thank you, @zukilover. The console warning for `findDOMNode` is no longer appearing, and the fade out animation is working as I would expect:

https://github.com/user-attachments/assets/f10d7adf-138d-4486-9ea7-4b43d0e16556

For comparison, here are the same actions in the `main` branch:

https://github.com/user-attachments/assets/fdda1908-05d9-4415-a1b9-d5b5b800c8e1

---

I'll leave the review of the other (previously) affected transition animations to @raineorshine, as I'm a bit less familiar with that topic.

trevinhofmann (changes requested):
Apologies. After re-reviewing the demos between `fix/fade-on-unmount` and `main`, it appears that the fade-in animation when expanding the thought `b` is not currently working on this branch as it was in `main`.

zukilover:
@trevinhofmann @raineorshine This has been updated by moving the `FadeTransition` one level down to `VirtualThought`, with the following changes:

- The `tree-node` Y-axis transitions are now separated from the thought’s `opacity` transition, ensuring they are mutually exclusive for better control over each transition.
- This change makes sense since `VirtualThought` already manages a `refOpacity`, which can now be directly passed from the `FadeTransition`'s `nodeRef`.
- With the transition logic now handled by `FadeTransition`, all CSS overrides related to opacity and transitions for `autofocus: show` are skipped by setting them to `undefined`, allowing `FadeTransition` to manage them correctly.

raineorshine (changes requested):
Thanks @zukilover. I can confirm that the fade-in and fade-out animations are present.

Here are the issues I'm still seeing in acfd20e2283.

# 1. Fade-in timing

I'm seeing some discrepancies in the fade-in animation timing compared to `main`. They are not visible at 1x speed, but become clear when stepping through the animations frame by frame.

Overall what I'm seeing is that the fade in starts later and ends earlier in acfd20e2283 than in `main`.

Here is a frame near the beginning of the fade-in. You can see that the newly mounted thought fades in faster than it should in acfd20e2283:

(`main` is on the left; acfd20e2283 is on the right.)

QuickTime Player 2024-11-23 20 43 47

Here is a frame towards the end of the fade-in. In `main`, the fade-in completes at exactly the same time that `c` arrives in its final y position. In acfd20e2283, `b` is at full opacity too early.

QuickTime Player 2024-11-23 20 44 52

# 2. Ancestors snap in on cursorUp

I'm not 100% certain when this was introduced, so it might be worth tracking down the offending commit.

## Steps to Reproduce

```
- a
  - b
    - c
      - d
        - e
          - f
```

1. Set the cursor on `f`
2. Quickly activate cursorUp repeatedly.

## Current Behavior

https://github.com/user-attachments/assets/7ddcc402-6ef0-448e-b083-848be60ab8ee

## Expected Behavior

https://github.com/user-attachments/assets/4edfc9b2-2701-421b-80f8-18b7e046901e

zukilover:
@raineorshine the fade-in timing difference happens due to 2 reasons:

1. Previously, `VirtualThought` fades in in `layoutNodeAnimationDuration` which was 150ms, and now it is using `veryFastDuration` which is 80ms:

https://github.com/cybersemics/em/blob/29fb6c88672c18ff23b72cf3fa02ab3b2e1cdc42/src/components/VirtualThought.tsx#L253

2. Still related to the code above, the new `fadeTransition` uses `ease-out` instead of `ease-in` here:

https://github.com/cybersemics/em/blob/acfd20e2283ca3d29638f5bc696f32f4ed0082ac/src/recipes/fadeTransition.ts#L20-L21

Proposed solution:
Change duration of `veryFastDuration` to `150ms` so that it matches with `layoutNodeAnimationDuration` and change the animation timing function to `ease-in`.

raineorshine:
`veryFastDuration` is a standard duration that is used in other places, so we don't want to change that specifically. However, we could add a special duration just for this instance since it has unique timing needs (similar to the poorly named `mediumBoth`).

Note that we can also use a custom bezier curve if `ease-in` and `ease-out` do not work and we need to precisely fit the fade within the 150ms.

zukilover:
This has been fixed with a new animation timing function. Here's the result at the same frame:

 
data_image_jpg;base…

raineorshine:
Thanks. I'll do a detailed review at the end. Will you investigate **2. Ancestors snap in on cursorUp**? That seems to be more than a timing issue.

zukilover:
@raineorshine the `Ancestors snap in on cursorUp` issue is fixed by refactoring the `autofocus` transition now that we use `FadeTransition`. From my latest check, it is working as expected. 

I will double check to make sure.

zukilover:
I have checked the latest update, here's how it works:

[screencast-localhost_3000-2024_11_26-12_39_48.webm](https://github.com/user-attachments/assets/352352d7-79cb-4778-a84d-349f7b582492)

We can see from individual frames that the opacity works as expected on autofocus:
c

b

a

raineorshine:
Unfortunately that's not the expected behavior. Thoughts should not snap to full opacity as soon as the cursor moves to them.

Take a closer look at the difference between Current and Expected shown in **2. Ancestors snap in on cursorUp**: https://github.com/cybersemics/em/pull/2581#pullrequestreview-2456479250.

zukilover:
@raineorshine I think the opacity inline style within `VirtualThought` need to be refactored into `fadeTransition`. Will @yangchristina be able to help with this? I've been struggling with it quite some time, but couldn't come up with the correct setup.

I mapped opacities into these transitions:

```
nodeAutofocusHideOpacity: {
        enter: { transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
        enterActive: { opacity: 0, transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
        enterDone: { opacity: 0, transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
        exit: { opacity: 0 },
        exitActive: { transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
      },
      nodeAutofocusDimOpacity: {
        enter: { opacity: 0.5, transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
        enterActive: { opacity: 0.5, transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
        enterDone: { opacity: 0.5, transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
        exit: { opacity: 0.5 },
        exitActive: { transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
      },
      nodeAutofocusShowOpacity: {
        enter: { opacity: 0.5 },
        enterActive: { opacity: 1, transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
        enterDone: { opacity: 1, transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
        exitActive: { transition: `opacity {durations.layoutSlowShiftDuration} ease-out` },
      },
```

but still no luck.

raineorshine:
FadeTransition is used to fade in/out elements on mount/unmount. Autofocus opacity is applied to thoughts that are already mounted, so FadeTransition doesn't seem like a good fit here.

It seems like the FadeTransition opacity is overriding the autofocus opacity, but we want it the other way around. Maybe opacity: 1 does not need to be applied during some phases of the animation? Unfortunately I'm not able to get too much deeper into the technical details, as I have a lot of other responsibilities at the moment. 

I would suggest backtracking to the last point where autofocus was working. Building on a broken autofocus is probably not a good idea. Then find out how to animate on unmount from there.

zukilover:
Hi @raineorshine, I've pushed an update that separates the `FadeTransition` for the thought while preserving the existing autofocus behavior.

This update addresses the three issues we discussed earlier:

1. Ensures the thought fades out when unmounted.
2. Keeps the autofocus behavior unchanged.
3. Synchronizes the fade-in transition with the `layoutNodeAnimation` by introducing a new `nodeOpacity` duration.

raineorshine:
Thanks so much! I'll let @trevinhofmann do the first review.

trevinhofmann (approved):
Thank you, @zukilover! The latest implementation looks quite clean and simple, and everything appears to be behaving as expected.

Both fade in and fade out are functioning as I would expect:

https://github.com/user-attachments/assets/c3f3f234-da59-4797-82c3-342f6cb62ce4

The "Ancestors snap in on cursorUp" issue is also resolved:

https://github.com/user-attachments/assets/8ca8bb82-cbe6-40b1-92e6-91b0b804b6bf

raineorshine (changes requested):
Thanks! This is looking very good.

Here is what I'm noticing.

# 3. Empty thought fades in

There used to be a special case for the empty thought so it would snap in rather than animate in. This was lost in the new implementation.

## Current Behavior

https://github.com/user-attachments/assets/b1554fd2-5c4b-4752-919c-76bc3de6df2c

## Expected Behavior

https://github.com/user-attachments/assets/6440e735-08e7-449d-b7a6-8a586e6fa469

# 4. Long fade-in

## Steps to Reproduce

```
- a
  - b
- c
```

Use Ctrl/Cmd + ArrowUp and Ctrl/Cmd + ArrowDown to move the cursor between `a` and `c`.

## Current Behavior

Notice how `b` takes longer than expected (at least 500ms) to reach full opacity.

https://github.com/user-attachments/assets/1fc4bcc1-df14-4865-9d72-6b674dc2d364

## Expected Behavior

https://github.com/user-attachments/assets/393ca4eb-90cf-4a55-9732-7127bed18ad5

## Notes

I'm not sure why the fade-in exceeds the `nodeOpacity` duration. It seems as if it is fading in with the `layoutSlowShift` duration which is what the Subthought autofocus [opacity transition](https://github.com/cybersemics/em/blob/bb6b138dd26c37911786e5c7b3b111af26d4e792/src/components/Subthought.tsx#L129) uses. But the Subthought component has not changed at all, so I'm not sure how that could be implicated.

Could it be that the thought is not being unmounted, and thus the [autofocusChanged] ref is not being renewed as expected? Should we being setting `unmountOnExit` on the `FadeTransition`?

# 5. Layered fade-in

This is more speculative, as I don't have a clearly observable test case. It may be masked by **<span>#</span>4** above.

If the goal is to only add a fade-out animation, why apply a fade-in animation with FadeTransition at all? CSSTransition can be configured to only animate on exit. This is relates to an older comment I made about [redundant fade-in animations](https://github.com/cybersemics/em/pull/2581#pullrequestreview-2438597537). It seems that layering fade-in animations may lead to unexpected behavior. (Not to mention composing easing functions makes it harder to design animation timing.)

zukilover:
@raineorshine you are correct, we shouldn't touch mounting thought animation. I've pushed updates so that it focuses on the exit transition. I've also added `unmountOnExit` that should fix the long fade-in issue. Kindly re-review.

src/recipes/fadeTransition.ts
```diff
@@ -27,6 +28,9 @@ const fadeTransitionRecipe = defineSlotRecipe({
         enterActive: { transition: `opacity {durations.distractionFreeTyping} ease 0ms` },
         exitActive: { transition: `opacity {durations.slow} ease 0ms` },
       },
+      nodeOpacity: {
+        exitActive: { transition: `opacity {durations.nodeOpacity} ease-out` },
```
raineorshine, line 32: Should this also set `enter: { opacity: 1 },` to completely avoid the `enter` animation?

raineorshine (changes requested):
Thanks, looks great now! That totally took care of issues 3,4,5.

I discovered an error when navigating a larger thoughtspace. Confirmed that this does not occur in `main`.

## Steps to Reproduce

I'm sorry I wasn't able to reproduce it with a more minimal test case.

```
- Buddhist Art
  - Bodhisattvas
    - Avalokiteshvara
      - “One who who hears the cries of the world”
      - Korean: Guanyin
      - Japanese: Kannon
      - Often has a small Amithaba figure in the crown
      - Compassion
      - Lotus bloom
      - Medicine flask
      - Wish-fulfilling jewel
        - Possibly develops in Paekche
      - Eleven heads
        - Given by Amithaba to help tend to all those who suffer
      - Usually the center of a mandorla triad
    - Green Tara
      - Abundance
      - Auspiciousness
      - Action/Force
    - Kshitigarbha
      - Japanese: Jizo
      - Korean: Jijang
      - Monk
      - Close-cropped hair
      - Staff
    - Manjushri
      - Japanese: Monju
      - Sword and noose
        - Cuts through ignorance
      - Prajna
        - Wisdom
        - Insight
      - Rides a blue lion
      - Resides in the eastern Pure Land
    - Samantabhadra
      - “Universal Worthy”
      - Japanese: Kugen
      - 12th–14th c.
        - Heian period
      - Connected to the Lotus Sutra
      - Six-tusked white elephant
      - Often serves as Shakyamuni’s attendant with:
        - Manjushri
      - Worshipped independently in Japan
      - Tendai school
    - Vajrapani
      - “vajra holder”
      - Power
    - Vajrasattva
      - Esoteric form of Vajrapani
  - Buddhas
    - Akshoba
      - Presides over the cementerial
    - Amithaba
      - Korean: Amita
      - Pure land Buddhism
      - Buddha of Infinite Light
      - Fingers curled and knuckles pressed together
    - Bhaishajyaguru
      - Buddha of Healing
    - Dipankara
      - Past Buddha that has been forgotten
    - Maitreya
      - Future Buddha
      - Standing on a lotus base instead of sitting suggests that Maitreya resides in heaving, awaiting his final rebirth
      - Water flask (globe) of an ascetic
      - Pensive pose
        - hanka shii
      - Originated with images of Shakyamuni as Prince
      - Emerged with Pure Land Buddhism during a time of war social chaos
        - Northern Wei split violently into Eastern and Western Wei
    - Shakyamuni
      - Double bun
        - Aristocratic East Asian youth, 6th c.
    - Vairochana
      - Formed from the mind-made buddha created upon enlightenment
      - Jewels and crown representing Vairochana as a living Buddha presiding over the heavens
      - Conceptual map of the universe
      - Sometimes represented by:
        - Bumhisparsha
  - Development
    - *Alternative structure*
    - 1st–2nd c. Anthropomorphic representations
    - 3rd c. — China
    - 4–5th c. — Korea
      - Introduced through foreign diplomatic relations
      - Aristocratic class only
    - 5–6th c. — Japan
    - 7–9th c. — Vajrayana → Japan
    - 6–7th c. — Vajrayana → S.E. Asia
      - Maritime routes
    - 11th c. — Himalayas
    - 11th–13th c. — Mahayana/Vajrayana spread
  - Episodes/Motifs
    - Parinirvana
      - Final nirvana
      - Late 11th c. — Lying on his back
      - Early 14th c. — Lying on side with right arm as pillow
  - Medium
    - Dry lacquer
      - Few extant examples
      - Clay core with layers of fabric strips coated in resin (lacquer)
      - Multiple layers of lacquer with thickened including oil and burned bone
  - Mudras
    - Abhaya Mudra
      - Hand raised
      - Approachability
    - Bumhisparsha
      - Touching the earth
      - Moment of enlightenment
    - Vitarka
      - Thumb and index finger touching, palm out
      - Teaching
      - Exposition
    - Varada
      - Palm out, fingers down
      - Granting boons
      - Wish granting
    - Dhyana
      - Meditation
      - One hand covering the other in front of hara
      - Does not occur in Japan?
    - Sovereignty
      - Right arm raised, left arm lowered
      - Newborn Buddha
      - Supremacy over all that exists above or below the heavens
    - Wisdom-fist
      - Two stacked fists
      - Mahavairocana
        - Esoteric Buddhism
      - Supreme Buddha of the Avatamsaka teaching
        - Korea
  - Poses &amp; Motifs
    - King Aśoka Buddha
      - Northern Wei → Koguryo → Japan
      - Japan
        - More solemn face
        - Shawl crosses lower near the ankles
        - More often holds wish-fulfilling jewel than the “heart-shaped attribute”
          - Paekche
        - High crown
          - Paekche
    - Hanka shii
      - 7th c. Japan
      - Pensive
      - hanka
        - “Legs half-crossed”
      - shii
        - Meditation
        - Finger gently touching the cheek
      - Shou-cordon
        - Decorative cloth hanging from the waist
        - Fashioned after Chinese court attire
        - Jade ring
        - Tucked under the figure’s bottom, bottom braids separate from top
          - Silla
        - Hanging straight down, too and bottom braids are connected
          - Paekche
        - Derived from:
          - Northern Qi
          - Sui
          - Early Tang
      - Origin
        - India
        - Originally represented Shakyamuni and then Maitreya
    - Single mandorla triad
      - Distinctly Korean
        - Scrolling vine and concentric circles within head halo
        - Crest of the head halo
          - Wish-fulfilling jewels
        - Head halo light ray pattern
        - Attendants extend beyond the edge of the mandorla
        - Interval petals
          - Artichoke style
        - Space between feet
        - Placement of Transformation Buddhas
        - Eyebrows chiseled after casting
      - vs Chinese
        - “Snake-dragon” flame pattern
        - Wider halos
        - Attendants are smaller and more distant from the central Buddha
      - Influenced by:
        - Northern Wei
    - Oversized head
```

1. Set the cursor on `Buddhist Art`
2. Set the cursor on `Bodhisattvas`
3. Set the cursor on `Avalokiteshvara`.

## Current Behavior

Error:

```
Rendered fewer hooks than expected
```

## Expected Behavior

No error.

zukilover:
@raineorshine I have fixed the `Rendered fewer hooks than expected` issue by moving the `useRef` hook to the top of the render logic.

raineorshine:
Great, thanks!

Just awaiting your reply here: https://github.com/cybersemics/em/pull/2581#discussion_r1870456347?

zukilover:
@raineorshine Yes, to avoid the enter animation, I've updated the fadeTransition to include the full opacity at enter.

src/recipes/fadeTransition.ts
zukilover: Yes, I've added `enter: { opacity: 1 }` to avoid the enter animation.

raineorshine (approved):
Thanks for your persistence. Everything is working great!

PR #2582, opened 2024-11-12
Fix: Slow virtualization of thoughts

dangold604:
## Problem
This pull request resolves a bug causing slow virtualization of thoughts, ensuring smooth animation performance as expected. The issue occurred because the animation did not run smoothly when hidden thoughts were re-rendered; this was due to the animation being applied directly to the hiding DIV element.

## Solution
The solution is to remove the animation from the hiding DIV element and instead apply it to the parent DIV element, which maintains the animation's smoothness and performance.

Related to [#2399](https://github.com/cybersemics/em/issues/2399)

### DIFF
--- a/src/components/Subthought.tsx
+++ b/src/components/Subthought.tsx
@@ -1,12 +1,10 @@
-import React, { useEffect, useMemo, useRef } from 'react'
+import React, { useMemo, useRef } from 'react'
 import { shallowEqual, useSelector } from 'react-redux'
-import { css } from '../../styled-system/css'
 import Autofocus from '../@types/Autofocus'
 import LazyEnv from '../@types/LazyEnv'
 import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import ThoughtId from '../@types/ThoughtId'
-import useChangeRef from '../hooks/useChangeRef'
 import attributeEquals from '../selectors/attributeEquals'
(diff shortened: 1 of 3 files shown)


raineorshine (changes requested):
Thanks for your submission!

It looks like the Puppeteer tests are failing. The change has impacted the drag-and-drop functionality. Can you take a look?

dangold604:
Hi @raineorshine,
I’ve resolved the issue with the Puppeteer test. The command `yarn test:puppeteer` ran successfully on my local machine.
Thank you!

raineorshine (changes requested):
Thanks!

Looking through the snapshot diffs, I can see that many of the drag-and-drop tests have broken. There are too many for me to point out individually, but they should be fairly clear with a quick visual comparison in the GitHub diff. 

The most common case seems to be the child drop targets which are no longer above the other drop targets. Maybe the stacking order inadvertently changed when the animation was moved? The other one is that the drop hover line has disappeared in some tests.

dangold604:
Hi @raineorshine 
I updated this PR now.
Thanks.

raineorshine (approved):
Great work! I can confirm that animations are all working as expected, and uncles are no longer rendered after they have been hidden by the autofocus.

raineorshine:
Unfortunately I had to revert this, as it caused a rendering discrepancy in Safari: #2650

Perhaps we can circle back to this later and see if there is another approach that could work.

Issue #1631, opened 2022-07-21
Changes are not persisted after undo


raineorshine:
## Steps to Reproduce

1. Create thought `a`. 
2. Create thought `b` after `a`. 
3. Undo.
4. Create thought `c` after `a`. 
5. Refresh the page. 

## Current Behavior

`c` was not persisted. 

No changes are persisted after the undo. 

``` 
- a
``` 

## Expected Behavior

All changes should be persisted. 

```
- a
- c
```

raineorshine:
Possibly caused by undo.

raineorshine:
Revert 40f1a7c after this is fixed.

RED-ROSE515:
This is why this unexpected behavior happens.
Because, current lexemeId will be removed from dockeys in **deleteThought** function(thoughtspace.ts).
And return error in checking if the dockeys contain the cxid.
Because of that `paused ` flag set as true and changes are not persisted after undo.

raineorshine:
Thanks for investigating!

Can you clarify what is triggering `deleteThought`? Which step?

I understand that when the `taskQueue` is paused from the error, it never resumes, and that's why new edits are not persisted. I'm not sure exactly which `taskQueue` you are referring to though.

One thing that would be helpful would be for you to link to some of the relevant line numbers in the code. That will make it easier for me to connect your explanation to the specific places in the code that you are referring to.

Next, what are the incorrect assumptions that the code makes? Normally, dockeys are cached so that we can always access a thought from its parents. What is different about undo/redo?

What do you suggest as a solution?

Thank you!

RED-ROSE515:
Currently, if you don't create a new thought before exec `undo`, it works well.
But according to these steps, it doesn't work properly.
 1. Create thought a.
 2. Create thought b after a.
 3. Undo.

Because, when exec undo, it executes deleteThought function and it will delete the thought id from dockeys.
 [https://github.com/cybersemics/em/blob/9e7f8a1e438468f3b1b8f330ffa8701f39074635/src/data-providers/yjs/thoughtspace.ts#L942](url)

raineorshine:
Okay, thanks. One question that brings for me is, why does it break on undo, but not if you manually activate `deleteThought` on `b`? The effect should basically be the same.

That is, on the surface there doesn't appear to be anything wrong with undo triggering `deleteThought`, since the end goal is to delete `b`. There must be something more to it.

raineorshine:
[...]

PR #2588, opened 2024-11-15
Fix : Changes are not persisted after undo

RED-ROSE515:
This Pull Request resolves the issue : #1631.

First, there was a TypeError in updateLexeme.
   `TypeError: str is undefined
    _writeVarStringNative encoding.js:307
    writeString UpdateEncoder.js:72
    write Item.js:691
    writeStructs encoding.js:67
    writeClientsStructs encoding.js:100
    writeClientsStructs encoding.js:99
    writeStructsFromTransaction encoding.js:368
    writeUpdateMessageFromTransaction Transaction.js:135
    cleanupTransactions Transaction.js:361
    transact Transaction.js:439
    transact Doc.js:185
    updateLexeme thoughtspace.ts:611`
    
    So, I added checking for value if it has changed before setting that.
    
    Second, remove `onLexemeChange` part to prevent updateLexeme performed after deleteLexeme.


raineorshine (changes requested):
Thanks! I can confirm that changes are persisted after undo now. Thanks for restoring the shortcuts and toolbar buttons!

Just a few questions about the implementation. Please update the PR description with a good overview as previously advised. It will help me better understand how your solution solves the issue.

raineorshine:
Thank you for the general write up. I have unresolved three of the questions I had, as I don't feel that they have been adequately answered yet. Please respond individually at your convenience.

raineorshine (approved):

raineorshine:
Note: The original stated solution should be ignored.

The new solution appears to make the following changes:

- Add use of `lexemeKeyToDb` when updating non-context props on the YJS doc.
- Execute updates then deletes rather than in parallel.
- Remove pending Lexeme logic that triggers `onLexemeChange`.

PR #2589, opened 2024-11-15
Fix/fade in new thought

### LINKED ISSUE #1790: Animation: Fade in new thought
New Thought has a visual problem in that the bullet and placeholder briefly overlap the next thought until it shifts down.

It has a perceptual problem in that the animation duration is the same duration as larger changes, such as collapse context or toggle table view. For the right "feel", the animation duration should be proportional to the "mass" of the changes, i.e. a T-Rex moves slower than a chicken. 

## Steps to Reproduce

``` 
- a
- c
``` 

1. Set the cursor on `a`
2. Activate the New Thought command

## Current Behavior

When a new thought is created, the bullet and placeholder briefly overlap the next thought until it shifts down.

*Note: The placeholder ("Add a thought") is only shown in training mode, so is not captured in the video.*

https://github.com/user-attachments/assets/f86cbd39-3de6-417a-ad09-94c5d43d03e8

## Expected Behavior

1. Speed up the tree-node's transition (normally `layoutNodeAnimationDuration`) by 50% on New (Sub)Thought only.
1. Fade in the bullet and placeholder of the new thought with `fastDuration` or `veryFastDuration` and `ease-in-out` so the overlap is less noticeable.

zukilover:
Fixes #1790 

This PR is branched off of #2581 because it depends on the new `CSSTransition` for `tree-node` element.
Added new property `value` for the `TreeThought` to distinguish new thoughts.

<details>
<summary>End result:</summary>

https://github.com/user-attachments/assets/e92bf6f0-6256-4e8a-9dff-fc3f1c5e6a80

</details>

### DIFF
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -57,6 +57,7 @@ type TreeThought = {
   // index among all visible thoughts in the tree
   indexDescendant: number
   isCursor: boolean
+  isEmpty: boolean
   isInSortedContext: boolean
   isTableCol1: boolean
   isTableCol2: boolean
@@ -73,7 +74,6 @@ type TreeThought = {
   thoughtId: string
   isLastVisible?: boolean
   autofocus: Autofocus
-  value: string
   // keys of visible children
   // only used in table view to calculate the width of column 1
   visibleChildrenKeys?: string[]
@@ -307,6 +307,7 @@ const linearizeTree = (
       belowCursor = true
     }
 
+    const isEmpty = child.value === ''
     const isTable = attributeEquals(state, child.id, '=view', 'Table')
     const isTableCol1 = attributeEquals(state, head(simplePath), '=view', 'Table')
     const isInSortedContext = attributeEquals(state, head(simplePath), '=sort', 'Alphabetical')
@@ -321,6 +322,7 @@ const linearizeTree = (
       indexChild: i,
       indexDescendant: virtualIndexNew,
       isCursor,
+      isEmpty,
       isInSortedContext,
       isTableCol1,
       isTableCol2,
@@ -339,7 +341,6 @@ const linearizeTree = (
       simplePath: contextViewActive ? thoughtToPath(state, child.id) : appendToPathMemo(simplePath, child.id),
       style,
       thoughtId: child.id,
-      value: child.value,
       ...(isTable
         ? { visibleChildrenKeys: getChildren(state, child.id).map(child => crossContextualKey(contextChain, child.id)) }
         : null),
@@ -384,6 +385,7 @@ const TreeNode = ({
   indexChild,
   indexDescendant,
   isCursor,
+  isEmpty,
   isTableCol1,
   isTableCol2,
   thoughtKey,
@@ -409,7 +411,6 @@ const TreeNode = ({
   cliffPaddingStyle,
   dragInProgress,
   autofocusDepth,
-  value,
   ...transitionGroupsProps
 }: TreeThoughtPositioned & {
   thoughtKey: string
@@ -425,8 +426,10 @@ const TreeNode = ({
 } & Pick<CSSTransitionProps, 'in'>) => {
   const [y, setY] = useState(_y)
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
-  const lastPatches = useSelector(state => state.undoPatches[state.undoPatches.length - 1])
-  const isLastActionNewThought = lastPatches?.some(patch => patch.actions[0] === 'newThought')
+  const isLastActionNewThought = useSelector(state => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    return lastPatches?.some(patch => patch.actions[0] === 'newThought')
+  })
 
   useLayoutEffect(() => {
     if (y !== _y) {
@@ -486,7 +489,7 @@ const TreeNode = ({
         // The FadeTransition is only responsible for fade out on unmount;
         // or for fade in on mounting of a new thought.
         // See autofocusChanged for normal opacity transition.
-        duration={value === '' ? 'nodeFadeIn' : 'nodeFadeOut'}
+        duration={isEmpty ? 'nodeFadeIn' : 'nodeFadeOut'}
         nodeRef={fadeThoughtRef}
         in={transitionGroupsProps.in}
         unmountOnExit
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -73,6 +73,7 @@ type TreeThought = {
   thoughtId: string
   isLastVisible?: boolean
   autofocus: Autofocus
+  value: string
   // keys of visible children
   // only used in table view to calculate the width of column 1
   visibleChildrenKeys?: string[]
@@ -338,6 +339,7 @@ const linearizeTree = (
       simplePath: contextViewActive ? thoughtToPath(state, child.id) : appendToPathMemo(simplePath, child.id),
       style,
       thoughtId: child.id,
+      value: child.value,
       ...(isTable
         ? { visibleChildrenKeys: getChildren(state, child.id).map(child => crossContextualKey(contextChain, child.id)) }
         : null),
@@ -407,6 +409,7 @@ const TreeNode = ({
   cliffPaddingStyle,
   dragInProgress,
   autofocusDepth,
+  value,
   ...transitionGroupsProps
 }: TreeThoughtPositioned & {
   thoughtKey: string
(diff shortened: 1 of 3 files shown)


src/components/LayoutTree.tsx
```diff
@@ -739,149 +743,167 @@ const LayoutTree = () => {
...
+              const previousThought = isTableCol1 ? treeThoughtsPositioned[index - 1] : null
+
+              // Adjust col1 width to remove dead zones between col1 and col2, increase the width by the difference between col1 and col2 minus bullet width
+              const xCol2 = isTableCol1 ? nextThought?.x || previousThought?.x || 0 : 0
+              // Increasing margin-right of thought for filling gaps and moving the thought to the left by adding negative margin from right.
+              const marginRight = isTableCol1 ? xCol2 - (width || 0) - x - (bulletWidth || 0) : 0
+
+              // Speed up the tree-node's transition by 50% on New (Sub)Thought only.
+              const isNewThought = value === ''
+              const layoutDuration = isNewThought ? `layoutNodeAnimationFastDuration` : `layoutNodeAnimationDuration`
```
trevinhofmann, line 798: This duration does not appear to have the intended effect (or any effect?). Since the duration is so short, I updated `layoutNodeAnimationDuration` to `5000` ms in the durations config to test it locally. Doing so doesn't seem to have lengthened the duration of any animations for a new subthought:

https://github.com/user-attachments/assets/1f82445d-fc0f-48ba-a76e-56955c74221b

I also increased the value of `layoutNodeAnimationDuration` with the same unchanged results.

trevinhofmann (changes requested):
@zukilover / @yangchristina / @raineorshine, please let me know whether you would rather proceed with this PR or with #2601 which resolved the merge conflicts but may have made other changes in doing so.

---

# 1. Disappearing thoughts

Thoughts are unexpectedly disappearing from the end of the list when inserting new thoughts before them

## Steps to reproduce

```
- foo
- bar
- baz
```

1. Set the cursor on `foo`
2. `newThought` several times

## Current Behavior

The `bar` and `baz` thoughts disappear:

https://github.com/user-attachments/assets/3e8f0e0d-9cc8-4444-a219-0f3e301c2390

## Expected Behavior

The thoughts should not disappear. Demo from the `main` branch:

https://github.com/user-attachments/assets/1382db8d-602b-4698-910c-64952edac38d

raineorshine:
Marking as draft until #2581 is merged. Stay tuned.

zukilover:
@raineorshine @trevinhofmann I've made a full update to the approach in this PR:

1. The tree-node transition is now twice as fast (using the new `layoutNodeAnimationFast` duration) when the previous thought is a new one.
2. The new thought fades in with a new transition, `nodeFadeIn`, making the naming more consistent since the opposite transition is now `nodeFadeOut`. Both transitions have a duration of 80ms, which corresponds to the `veryFast` timing.

trevinhofmann (changes requested):
Thank you for the update, @zukilover.

Since this is only affecting the transition during for the subsequent thought, any other thoughts below it still transition more slowly. This causes a noticeable inconsistency between them, with thought `two` moving faster than the rest. When inserting new thoughts, `two` overlaps with `three`. When deleting them, `two` moves away from the other thoughts until they catch up.

https://github.com/user-attachments/assets/4c0cc37e-29cc-4fdd-b3f7-af04dbd7e469

I believe it would make sense to adjust the transition speed for all of the thoughts so that we don't have this inconsistency.

raineorshine:
Thanks for the initial review @trevinhofmann. 

Note that checking that the value of any given thought is `=== ''` is not going to tell you if the last command was New Thought or New Subthought. With your current approach it will alter the transition merely with the presence of an empty thought, regardless of the last command. An empty thought that already exists should obviously not alter the transition duration for unrelated commands. 

You need to check the undo history to see what the last command was: `state.undoPatches`.

zukilover:
I've updated the approach to check the undo history via `state.undoPatches` instead. Thanks @raineorshine  for pointing that out.

src/components/LayoutTree.tsx
```diff
@@ -414,6 +417,8 @@ const TreeNode = ({
 } & Pick<CSSTransitionProps, 'in'>) => {
   const [y, setY] = useState(_y)
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
+  const lastPatches = useSelector(state => state.undoPatches[state.undoPatches.length - 1])
+  const isLastActionNewThought = lastPatches?.filter(patch => patch.actions[0] === 'newThought')
```
trevinhofmann, line 421: Since `.filter` leaves `isLastActionNewThought` as an array, it will always be truthy even if empty (i.e. where we use it in the ternary below to determine `layoutTransition`).

```suggestion
  const isLastActionNewThought = lastPatches?.some(patch => patch.actions[0] === 'newThought')
```

trevinhofmann (changes requested):
Thank you, @zukilover. After the fix in my comment below, it looks like using `undoPatches` will be a more accurate way to detect `newThought` than looking at the previous value being an empty string.

However, I am noticing a strange lag causing overlap when quickly inserting multiple new thoughts:

https://github.com/user-attachments/assets/983bea32-1dce-436d-98df-0fa92668a054

src/components/LayoutTree.tsx
zukilover: I've updated the logic with `.some` but didn't experience that lagging problem.
[screencast-localhost_3000-2024_12_12-07_25_33.webm](https://github.com/user-attachments/assets/103cac82-6fbd-442a-8049-fbffff0580b1)

I'll double check. 

I'm wondering, is it happening on the main branch too? @trevinhofmann

trevinhofmann: On `main`, we do have the overlap due to the animation being too slow (which this PR addresses):

https://github.com/user-attachments/assets/506cc0c6-f3d0-4af2-a90d-9c405a7bdd72

However, I don't experience the same lag issue which is still happening on this PR's branch:

https://github.com/user-attachments/assets/63aaaaff-a567-4eb4-8e50-f9e0202ed0cb

It isn't super consistent, but happens fairly often for me. Try holding the `enter` key to create new thoughts more quickly than just clicking the toolbar button.

zukilover: I can confirm that the issue mentioned above can also be reproduced in the main branch. 
I doubt this is related to the changes in this PR, but I'll figure out a solution to fix it.

https://github.com/user-attachments/assets/a34c887a-76c2-479b-b9a8-6e4663afd49d

trevinhofmann: That is interesting. I'll need to retest, because I hadn't seen it in `main` before. If it is happening in `main` and not a new regression, I'd rather not fix it in this PR.

trevinhofmann: I'm still unable to reproduce on `main`:

https://github.com/user-attachments/assets/fe7e2fc0-afaa-4b70-a53e-b843c35abd3b

And seeing it consistently on `fix/fade-in-new-thought`:

https://github.com/user-attachments/assets/9cf7c842-f9f2-4cef-83b2-b1188a9e34aa

src/components/LayoutTree.tsx
```diff
@@ -330,6 +331,7 @@ const linearizeTree = (
       simplePath: contextViewActive ? thoughtToPath(state, child.id) : appendToPathMemo(simplePath, child.id),
       style,
       thoughtId: child.id,
+      value: child.value,
```
raineorshine, line 334: Rather than pass `value: child.value`, pass `isEmpty: child.value === ''` so that `treeThoughts` is not recalculated every time a thought value changes.

In general, prefer narrower API's that pass the minimum data needed by the downstream components. This will decrease the surface area for change and in React often increase performance.

src/components/LayoutTree.tsx
```diff
@@ -414,6 +417,8 @@ const TreeNode = ({
 } & Pick<CSSTransitionProps, 'in'>) => {
   const [y, setY] = useState(_y)
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
+  const lastPatches = useSelector(state => state.undoPatches[state.undoPatches.length - 1])
+  const isLastActionNewThought = lastPatches?.some(patch => patch.actions[0] === 'newThought')
```
raineorshine, line 421: I recommend combining `lastPatches` and `isLastActionNewThought` into a single `useSelector` for performance reasons. Then the render function will only be re-triggered when `isLastActionNewThought` changes, instead of every time `lastPatches` changes.

In general aim for "thick" selectors in Redux for better performance. In many cases it won't be relevant, but we know from testing that `LayoutTree` is a heavy component and takes several milliseconds to render.

raineorshine (changes requested):
Works great, thanks!

Just a couple tweaks below.

> I can confirm that the issue mentioned above can also be reproduced in the main branch.
> I doubt this is related to the changes in this PR, but I'll figure out a solution to fix it.

I was also able to reproduce the overlap due to lagged initial render in `main` (23f4f9904f), so it's not a regression from this PR. I will create a separate for it.

trevinhofmann:
> I was also able to reproduce the overlap due to lagged initial render in main (https://github.com/cybersemics/em/commit/23f4f9904f766e4a60f3915697713c0af261b609), so it's not a regression from this PR. I will create a separate for it.

Great, thanks for testing that! Not sure why I'm seeing different results, but it's nice to confirm that this is not a new regression.

src/components/LayoutTree.tsx
zukilover: Thanks for the suggestion. I've updated the logic so that it passes `isEmpty` instead of `value`.

src/components/LayoutTree.tsx
zukilover: `isLastActionNewThought` is now updated in a single `useSelector`.

trevinhofmann (approved):
The functionality still appears correct to me, and the recent refactoring looks good. Thanks @zukilover!

raineorshine (approved):

PR #2590, opened 2024-11-15
fix additional eslint v9 dependencies and error

digitliofficial:


digitliofficial:
@raineorshine 

1. updated all the errors and warnings. 
2. updated few plugins version
3. eslint-config-react-app is not supported with v9 and there is no any new version available.

review and share your feedback if any.

raineorshine (approved):
Looks good, thanks!

PR #2593, opened 2024-11-16
Fix: [Capacitor] Top safe area is covered

msdewitt:
#2514

This is a small fix for the overlap issue caused on IOS without the contentInset being set. By default, the app will overlap without this configuration.

Screenshots after fix on ios iPhone 15 emulator:

Screenshot 2024-11-06 at 12 35 26 PM

Screenshot 2024-11-06 at 12 36 57 PM


raineorshine (approved):
Tested and confirmed! Thanks.

PR #2594, opened 2024-11-16
Fix: [Capacitor] Set overscroll background color from theme

msdewitt:
[Issue 1700](https://github.com/cybersemics/em/issues/1700)

I added a way to eliminate the ios bounce outside the boundary of the app while adding a scroll feature within the controllable range of the app code instead of the device.

### DIFF
--- a/capacitor.config.ts
+++ b/capacitor.config.ts
@@ -24,7 +24,7 @@ const config: CapacitorConfig = {
   webDir: 'build',
   ...serverConfig,
   ios: {
-    backgroundColor: '000000',
+    scrollEnabled: true,
   },
   plugins: {
     Keyboard: {
--- a/src/App.css
+++ b/src/App.css
@@ -3,7 +3,15 @@
   This file is deprecated. All styles should be added inline using PandaCSS.
   See: https://panda-css.com/docs/concepts/writing-styles
 */
-
+html,
+body,
+#root,
+#app {
+  height: 100%;
+  font-size: 16px;
+  overflow-y: auto;
+  clip-path: inset(0 0 0 0);
+}
 /* react-transition-group animation: fade */
 .fade-enter {
   opacity: 0;
--- a/ios/App/App/EnableBounce.m
+++ b/ios/App/App/EnableBounce.m
@@ -11,6 +11,6 @@
 @implementation UIScrollView (NoBounce)
 - (void)didMoveToWindow {
    [super didMoveToWindow];
-   self.bounces = YES;
+   self.bounces = NO;
 }
 @end


msdewitt:
@raineorshine The difference in puppeteer tests on this pr is caused by turning off the bounce feature of ios and taking inspiration from @ankitkarna99's proposed solution in the issue: [Issue 1700](https://github.com/cybersemics/em/issues/1700). 

Originally, it is suggested to set scroll to false, but this caused issues with the webView. Instead, I opted for a solution similar to this, disabling the scroll bounce within the ios webview and setting the scrolling power to within our app itself.

We cannot set the background color dynamically on the ios webview as we do with javascript when a user changes themes, so we have to opt for a solution like this which either hides or limits the user from seeing the webview background behind the app. 

With these changes, there is a drawback. By disabling the scroll bounce, it disables the scroll refresh feature which naturally comes with the Ios Webview.

@raineorshine please tell me your thoughts on this issue based on this information.

src/App.css
```diff
@@ -3,7 +3,15 @@
   This file is deprecated. All styles should be added inline using PandaCSS.
   See: https://panda-css.com/docs/concepts/writing-styles
 */
-
```
raineorshine, line 6: Can you add these styles to `panda.config.ts`? This file is actually deprecated.

src/App.css
```diff
@@ -3,7 +3,15 @@
...
+html,
+body,
+#root,
+#app {
+  height: 100%;
+  font-size: 16px;
```
raineorshine, line 11: Font size does not seem related to the problem we are trying to solve. Is that needed?

capacitor.config.ts
```diff
@@ -24,7 +24,7 @@ const config: CapacitorConfig = {
   webDir: 'build',
   ...serverConfig,
   ios: {
-    backgroundColor: '000000',
```
raineorshine, line 27: Remove this `backgroundColor` is causing a white flash on launch. Can we retain this, or does it interfere with the solution?

Even if it does not change with the theme, it would be better to have a black flash than a white flash.

raineorshine (changes requested):
Thanks for the update! This looks like a promising approach. The overscroll background color now correctly matches the theme background color based on my testing.

The only issue I am seeing is that programmatic scroll functionality is broken. Here are some examples, although there are likely others if `window.scrollTo` is not working everywhere.

- Scroll zone parallax - When the user scrolls, the scroll zone on the right should scroll independently, at a reduced rate.
- scrollCursorIntoView - When the cursor moves to a thought that is outside the viewport, it should automatically scroll to the thought. To reproduce, activate New Subthought on a thought with a long list of children that exceeds the height of the viewport. It should scroll down to the new thought.
- etc

capacitor.config.ts
msdewitt: Good idea.

raineorshine:
> It looks like it's working.

Hmmm sorry about that! Maybe my branch was not up-to-date, or I failed to build the Capacitor app correctly when I tested it. I do see now that the scroll is working as expected.

Now what I'm seeing in the XCode simulator is that scroll bounce on overscroll is disabled completely. The scroll stops when it reaches the top. Just to confirm, is that what you have as well?

msdewitt:
@raineorshine Are you talking about where you are scroll to the very bottom and create a new sub-thought?


Like this: https://github.com/user-attachments/assets/835d24fc-3395-4f39-b3e2-f606535c3dfa

I also thought this was not expected, but when I double-checked if the same behavior happens on main I found that it's exactly the same as my branch.

raineorshine:
I'm referring to the "scroll bounce" or "elastic scroll" behavior that occurs when scrolling past the top of the document. 

https://github.com/user-attachments/assets/9b89a29a-700b-49e3-85b3-a1f2ecf19509

msdewitt:
@raineorshine yes, that scroll bounce is actually a part of the UIScrollView of the IOS device. It is disabled in IOS in order to prevent the overscroll background from being shown since we can't dynamically alter the color of that (Its not javascript.)

Here is a video of the scroll bounce enabled and the UIScrollView background color set to white for visibility:

https://github.com/user-attachments/assets/71515cfb-caa5-4f7c-890a-5cbedd9aa739

raineorshine:
I see, thanks. Unfortunately we can't disable the scroll bounce. I've tried it both ways, and it feels really unnatural without it.

I'm open to other ideas, though I realize we are somewhat limited given the behavior of the WebView!

Maybe it's possible to add scroll bounce to a container within the body.

msdewitt:
@raineorshine I think that's possible. However, I'm not as familiar with adding animation and transformations with React as I am with Angular. 

I don't think I can complete this ticket since we can't remove the scroll bounce.

Do you want to reopen a new ticket to add a scroll bounce to the native JavaScript?

msdewitt:
@raineorshine Have you seen javascript emulation examples that work well?  Originally I thought this was a big change but if its much less than I thought, maybe I can still complete it here.

raineorshine:
> @raineorshine I mean the javascript emulation. It looks needlessly complicated to implement and might interact weirdly with the UIScrollView of IOS.

I've looked at all the Javascript emulations that exist for scroll bounce and unfortunately none of them perform well enough to replace the native scroll bounce. We're going to have to stick to native browser scroll bounce. Since this appears to not be possible on the WebView without losing control of the overscroll background, then it should be on an element in the DOM. I think we should explore this latter possibility.

> In fact, by disabling the scroll bounce on IOS it is made consistent with all the other versions of the app. There is no scroll bounce similar on web right now and this would be adding a new change across the app.

Not true. We currently have scroll bounce in Chrome, Safari, and Mobile Safari PWA, which is how I use the app personally on a daily basis. **em** has always had scroll bounce, at least on Apple devices.

msdewitt:
@raineorshine Alright, I managed to fix this issue after a lot of pain. I created a small custom plugin for capacitor and attached it to the project to interact with the native Webview. 

@raineorshine How do you want to store this small plugin? I originally just placed it as a folder within the project, but then lint was angry at me for adding a file to the install path and called it a security issue.

em_plugins/webview/package.json
```diff
@@ -0,0 +1,78 @@
...
+  "files": [
+    "android/src/main/",
+    "android/build.gradle",
+    "dist/",
+    "ios/Sources",
+    "Package.swift",
+    "WebviewBackground.podspec"
+  ],
+  "author": "Matheu DeWitt",
+  "license": "MIT",
```
raineorshine, line 18: Please omit the license so that it is included under the parent package's license.

em_plugins/webview/package.json
```diff
@@ -0,0 +1,78 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/msdewittdev.git"
+  },
+  "bugs": {
+    "url": "https://github.com/msdewittdev/issues"
+  },
```
raineorshine, line 25: `repository` and `bugs` should be removed or point to https://github.com/cybersemics/em.

em_plugins/webview/src/web.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import type { WebviewBackgroundPlugin } from './definitions';
+
+export class WebviewBackgroundWeb extends WebPlugin implements WebviewBackgroundPlugin {
+  // async echo(options: { value: string }): Promise<{ value: string }> {
+  //   console.log('ECHO', options);
+  //   return options;
+  // }
+  async changeBackgroundColor(options: { color: string }): Promise<void>{
+    console.log('Color changed: ', options.color)
```
raineorshine, line 11: Remove console.log

android/capacitor.settings.gradle
```diff
@@ -10,3 +10,6 @@ project(':capacitor-keyboard').projectDir = new File('../node_modules/@capacitor
 
 include ':capacitor-status-bar'
 project(':capacitor-status-bar').projectDir = new File('../node_modules/@capacitor/status-bar/android')
+
+include ':webview-background'
+project(':webview-background').projectDir = new File('../node_modules/webview-background/android')
```
raineorshine, line 15: Is there a reason the plugin is loaded from `node_modules` rather than directly from the source, since it's part of the same repo?

em_plugins/webview/.eslintignore
```diff
@@ -0,0 +1,3 @@
+build
+dist
+example-app
```
raineorshine, line 3: `example-app` appears to be leftover from the starter code.

em_plugins/webview/.prettierignore
```diff
@@ -0,0 +1,3 @@
+build
+dist
+example-app
```
raineorshine, line 3: `example-app` appears to be leftover from the starter code.

src/components/ThemeSwitch.tsx
```diff
@@ -9,6 +10,12 @@ const ThemeSwitch: FC = () => {
   const dispatch = useDispatch()
   const light = useSelector(state => theme(state) === 'Light')
 
+  if (light) {
+    WebviewBackground.changeBackgroundColor({ color: '#ffffff' })
```
raineorshine, line 14: Please use the PandaCSS design token for `color.bg` rather than hardcoding hex colors.

panda.config.ts
```diff
@@ -148,8 +148,9 @@ const globalCss = defineGlobalStyles({
         userSelect: 'none',
       },
     },
-  },
-  'html, body, #root, #app': { height: '100%', fontSize: '16px' },
+  }, // eslint-disable-next-line
```
raineorshine, line 151: In order to disable a lint rule, a couple things are required:

1. Disable rules explicitly by name. Lone `// eslint-disable-next-line` that disables all rules is not allowed.
2. Include an inline comment that explains why disabling the linter is necessary/reasonable.

In general, disabling the linter is not recommended, and will only be allowed if a sufficient argument is made.

em_plugins/webview/dist/esm/web.js
```diff
@@ -0,0 +1,11 @@
+import { WebPlugin } from '@capacitor/core';
+export class WebviewBackgroundWeb extends WebPlugin {
+    // async echo(options: { value: string }): Promise<{ value: string }> {
+    //   console.log('ECHO', options);
+    //   return options;
+    // }
```
raineorshine, line 6: You can remove commented out code if no longer needed.

em_plugins/webview/dist/esm/index.d.ts
raineorshine, line 1: Is `dist` meant to be under source control?

raineorshine (changes requested):
Awesome! Nice work. That seems like the right solution. Then we don't have to mess with the scroll bounce at all.

Should we call `changeBackgroundColor` after the theme is loaded from localStorage? Otherwise I think the WebView background color will still be the default.

I don't have a lot of experience with monorepos, but if we want to avoid publishing the plugin separately I would be fine storing it in `/packages`. I suggest `/packages/webview-background` so that the directory name matches the plugin name.

It looks like the yarn lockfile is out of sync, which is causing the CI to fail. Maybe related to importing the plugin from the local file system?

Rather than removing `node_modules` in subsequent commits, please rebase and ensure that they never enter the git history at all. Otherwise there will be many thousands of files that are part of the commit history even though they do not need to be under source control.

android/capacitor.settings.gradle
msdewitt: So I added it like this so that it can flow through the same capacitor build flow as normal plugins. Its also easier to add changes to this custom plugin in the future if we need too. The code is already available here.

It first builds and then adds to the project through the normal flow after the plugin is added to the app by `yarn add ./em_plugins/webview`

em_plugins/webview/.eslintignore
msdewitt: Yes, I'll remove it. It auto generates when using a cli to create it from scratch.

android/capacitor.settings.gradle
raineorshine: Okay, sounds good!

em_plugins/webview/dist/esm/index.d.ts
raineorshine: Okay, so we include it in source control so that other developers don't have to rebuild the plugin, but can just run `cap:copy` after fetching updates to the repo?

msdewitt: Yes, we can do exactly that. And since the plugin itself is small, if any additional plugins are needed in the future, its not difficult to add more functions. I already laid out the framework that others can follow. The goal would be to have any difficulties like this, 

(Where we run into a situation where Capacitor doesn't have a plugin with the functionality we need), 

in the future can just be handled by this custom plugin.

src/components/modals/ModalComponent.tsx
```diff
@@ -69,6 +69,7 @@ class ModalComponent extends React.Component<ModalProps> {
   close = () => {
     this.animateAndClose!()
     this.props.onClose?.()
+    window.location.reload()
```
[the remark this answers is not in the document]
raineorshine: Yikes! That is a pretty drastic action.

If that is absolutely the only way to change the background color of the Webview, then we have to constrain it as much as possible:

1. Only refresh if running the Capacitor build.
2. Only refresh if the theme has changed.
3. Add an Alert and a message beneath the Close button that notifies the user that the page will refresh upon close in order to update the theme.

raineorshine:
@msdewitt It looks like the CI is failing due to a lock file issue. Could you take a look?

src/components/modals/ModalComponent.tsx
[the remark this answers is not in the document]
msdewitt: @raineorshine alright, good idea to limit it to the capacitor build.

msdewitt:
> @msdewitt It looks like the CI is failing due to a lock file issue. Could you take a look?

I took a look at this and it seems to be a bug with yarn itself for `file:` and `git:` type dependencies. I'm not sure how to fix this as it's not even resolved over there. I have a general idea of what's happening, I think the different file systems it is pulling from is generating a different integrity check. The integrity check for the `file:` is completely useless in this regard and I'm looking how to disable it, but I can't find anything online.

@raineorshine  Would you know how to handle this?

[Link to this issue](https://github.com/yarnpkg/berry/issues/6105)

raineorshine:
No, I'm afraid I haven't used `file:` with yarn before.

The issue mentions `checkSumBehavior: ignore`, but I'm not sure if it's possible to limit that to a single dependency. 

@trevinhofmann Any ideas?

trevinhofmann:
For now, I've just updated the checksum from my Mac. This will at least unblock the CI from installing the dependencies and running.

Not sure what the solution will be for developers on Windows. I'm not using Windows at the moment, but I could start a VM to look further into this if you'd like. From the linked yarn issues, it seems like the most likely culprit is related to how git is configured to handle CRLF line endings on Windows.

packages/webview/dist/esm/web.js
```diff
@@ -0,0 +1,11 @@
+import { WebPlugin } from '@capacitor/core';
+export class WebviewBackgroundWeb extends WebPlugin {
+    // async echo(options: { value: string }): Promise<{ value: string }> {
+    //   console.log('ECHO', options);
+    //   return options;
+    // }
```
raineorshine, line 6: This commented out code can be removed.

packages/webview/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "lint": "npm run eslint && npm run prettier -- --check && npm run swiftlint -- lint",
+    "fmt": "npm run eslint -- --fix && npm run prettier -- --write && npm run swiftlint -- --fix --format",
+    "eslint": "eslint . --ext ts",
+    "prettier": "prettier \"**/*.{css,html,ts,js,java}\" --plugin=prettier-plugin-java",
+    "swiftlint": "node-swiftlint",
+    "docgen": "docgen --api WebviewBackgroundPlugin --output-readme README.md --output-json dist/docs.json",
+    "build": "npm run clean && npm run docgen && tsc && rollup -c rollup.config.mjs",
+    "clean": "rimraf ./dist",
+    "watch": "tsc --watch",
+    "prepublishOnly": "npm run build"
```
raineorshine, line 43: Since we use `yarn` on this project, can we switch to `yarn` instead of `npm`? I think that would be more consistent.

src/components/modals/ModalComponent.tsx
```diff
@@ -69,6 +70,9 @@ class ModalComponent extends React.Component<ModalProps> {
   close = () => {
     this.animateAndClose!()
     this.props.onClose?.()
+    if (Capacitor.isNativePlatform()) {
+      window.location.reload()
+    }
```
raineorshine, line 75: This is good that it only executes on the Capacitor build, but let's also avoid refreshing when the theme has not changed. As mentioned previously, we only want to refresh the app if we absolutely have to, since it is highly undesirable from a UX perspective. Hope you can understand.

You should have an easier time accessing Settings-specific logic by tapping into the `onClose` hook in [ModalSettings](https://github.com/cybersemics/em/blob/e39ebc414b562bd1b6a8131cdea6977da51bd273/src/components/modals/Settings.tsx#L117). Actually, this needs to be moved there anyway, because as-written it will refresh the app whenever any modal is closed.

src/components/modals/Settings.tsx
```diff
@@ -119,7 +120,16 @@ const ModalSettings = () => {
...
       actions={({ close }) => (
         <div className={css({ textAlign: 'center' })}>
-          <ActionButton key='close' title='Close' {...fastClick(() => close())} />
+          <ActionButton
+            key='close'
+            title='Close'
+            {...fastClick(() => {
+              close()
+              if (Capacitor.isNativePlatform()) {
+                window.location.reload()
```
raineorshine, line 129: I think this already refreshes since it calls the modal `close()` function above?

raineorshine: Yes, but it calls the ModalComponent's `close` function, which already does the reload.

src/components/modals/ModalComponent.tsx
msdewitt: Gotcha, I added a workaround to make it refresh as minimally as possible. It will only work onClose for the settings modal menu in this new version.

src/components/modals/Settings.tsx
msdewitt: Fixed.

msdewitt:
I finished up the changes for this merge, but now I ran into the same issue with yarn build for local builds again. I'm a bit confused why this is occurring again after the changes you pushed up @raineorshine . This pr is almost ready to merge just after this yarn build fix is solved again.

raineorshine:
Thanks @msdewitt!

I was able to run `yarn` and it updated the checksum:

```diff
diff --git a/yarn.lock b/yarn.lock
index 582c719a2e..ab60cf4560 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -17570,10 +17570,10 @@ __metadata:
 
 "webview-background@file:./packages/webview::locator=em%40workspace%3A.":
   version: 0.0.2
-  resolution: "webview-background@file:./packages/webview#./packages/webview::hash=f103c4&locator=em%40workspace%3A."
+  resolution: "webview-background@file:./packages/webview#./packages/webview::hash=d157bb&locator=em%40workspace%3A."
   peerDependencies:
     "@capacitor/core": ^6.0.0
-  checksum: 10c0/c3cae6281c2b684fd3f59db661c0b1d40cab4daca1be5681a4938ad58ba393c66ffd00242ba0c298e1ea72d35c5c0cd67e634aed1599671c24c9fac4125e9fe5
+  checksum: 10c0/3682d2b107e32e2ea9d42ceacf961336ed986d1f8ae3669e23290b7a6e1c89f92e06fd7896259de77574024ea42cf5d4362c3df3bb766c8c2464522e026506cb
   languageName: node
   linkType: hard
 

```

However, I get an error when running `yarn cap:ios`:

```sh
$ yarn cap:ios
Configuring development build
✔ Copying web assets from build to android/app/src/main/assets/public in 10.28ms
✔ Creating capacitor.config.json in android/app/src/main/assets in 558.08μs
✔ copy android in 45.65ms
✔ Updating Android plugins in 4.97ms
[info] Found 4 Capacitor plugins for android:
       @capacitor/clipboard@6.0.2
       @capacitor/keyboard@6.0.2
       @capacitor/status-bar@6.0.1
       webview-background@0.0.2
node:events:498
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, lstat '/Users/raine/projects/em/node_modules/webview-background/android/src/main'
Emitted 'error' event on Walker instance at:
    at /Users/raine/projects/em/node_modules/@ionic/utils-fs/dist/index.js:258:22
    at callback (/Users/raine/projects/em/node_modules/graceful-fs/polyfills.js:306:20)
    at FSReqCallback.oncomplete (node:fs:198:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'lstat',
  path: '/Users/raine/projects/em/node_modules/webview-background/android/src/main'
}
```

msdewitt:
Yes, you need to build the app before it cap:ios or cap:android is run.

raineorshine:
Thanks for your reply. I ran `npm install` and `npm run build` in the `/packages/webview` directory, but `tsc` fails with 43 errors. Running  Typescript `v4.1.6` as indicated in the package.json.

msdewitt:
@raineorshine  I can try and update the plugin typescript to that version, but I wasn't getting the same tsc errors that you are mentioning when I try to reproduce. Can you try deleting your local dependencies from yarn and node_modules folder and reinstalling locally to see if it fixes that issue?

I had this issue a while back and this fixed it.

raineorshine:
I tried removing node_modules and reinstalling, but I still get the same Typescript errors when I run `npm run build` in the `/packages/webview` directory.

I also get the following warning on `npm install` in the `webview` project:

```sh
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: cosmiconfig@9.0.0
npm warn Found: typescript@4.1.6
npm warn node_modules/typescript
npm warn   dev typescript@"~4.1.5" from the root project
npm warn   1 more (tsutils)
npm warn
npm warn Could not resolve dependency:
npm warn peerOptional typescript@">=4.9.5" from cosmiconfig@9.0.0
npm warn node_modules/cosmiconfig
npm warn   cosmiconfig@"^9.0.0" from swiftlint@2.0.0
npm warn   node_modules/swiftlint
npm warn
npm warn Conflicting peer dependency: typescript@5.7.2
npm warn node_modules/typescript
npm warn   peerOptional typescript@">=4.9.5" from cosmiconfig@9.0.0
npm warn   node_modules/cosmiconfig
npm warn     cosmiconfig@"^9.0.0" from swiftlint@2.0.0
npm warn     node_modules/swiftlint
```

@trevinhofmann Could you try building the `webview` package from this PR branch on your machine?

@msdewitt Can you confirm your node, npm, and Typescript versions? This is what I am on:

node v22.0.0
npm v10.9.1
Typescript v4.1.6

Thank you!

msdewitt:
Oh, I understand now. You are installing packages inside the new plugin.



My environment is:
Node: v18.13.0
Npm: 8.19.3
Tsc: 4.1.5

msdewitt:
@trevinhofmann I have finished adding build fixes to this, but once again the checksum has caused build issues. I need to update the checksum to the same one that is run on the github actions machine. 

Any idea where I can find the checksum it is using? I tried to look at the github actions report in the Test check below, but it didn't show the checksum.

trevinhofmann:
I've updated the checksums locally, and the CI is installing dependencies correctly again.

@msdewitt -- The checksums are visible in the `Install npm dependencies` step of the CI builds: https://github.com/cybersemics/em/actions/runs/12506619819/job/34891830241

Screenshot 2024-12-27 at 7 07 11 PM

I'm not _certain_ that the CRLF line endings are the issue, but it might be worth trying to adjust this git setting to see if it does affect the checksums for you on Windows: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#global-settings-for-line-endings

msdewitt:
@trevinhofmann Thank you! These changes should be good to merge then after that push. @raineorshine can review as well.

The changes in that custom plugin will most likely not be moved often in the future except for specific capacitor functionalities that don't already have a plugin on the market.

raineorshine:
Thanks @trevinhofmann! The CI is passing and I am not getting tsc errors any more in `/packages/webview`.

Do we need to add an additional step to the `cap:copy` or `cap:sync` npm script? I see how `webview` is included as a dependency in the root project, but I'm not clear how it gets built in the first place without the developer manually cd'ing into the `/packages/webview` directory and installing and building from there. We should automate this for better DX, especially given that these additional steps are not listed in the README.

Regarding the behavior, it is working when the app first loads. When I enable the Light theme, kill the app, and open it again however the background is black. Tested in Xcode Simulator.

msdewitt:
@raineorshine It is pulling the webview dep directly from the build directory folder. When a new version of the plugin is released, we just need to push the saved new build in the directory folder up to git. Yarn will automatically install the 'latest' version we have built in the local and we shouldn't need to add any additional flows into the capacitor build pipeline.

Every time you build a new 'webview' plugin version, its the same as if you are 'saving a new version' of the build to a repository, its just that the repository is always local.

raineorshine:
Okay, I think I understand. So we have the `webview` package build output included in version control, so we don't need to do anything besides `git pull` and `yarn` to use it.

msdewitt:
Exactly

msdewitt:
@raineorshine Everything looks good on my end.
Here is a screen recording of me changing the theme, canceling it out, and reloading it.

Try testing again, but don't build or do anything with the plugin version. Just `yarn` install in the main directory and `npm run build` and `npm run cap:ios` and it should be good to go.


https://github.com/user-attachments/assets/c2845805-6293-4fbe-87bf-6790e3629dbf

raineorshine:
Thanks for the screen recording. You can see that when you reopen the app it has the wrong background color (black) when the theme is Light.

msdewitt:
@raineorshine I understand, you are talking about the splash screen.

I can't update this initial loading of the splash screen dynamically due to the limits of Capacitor until after the app loads.

When an app using Capacitor loads, it first pulls the capacitor configuration file at the root of the mobile project and won't look at any variables in the app during start-up. We can only update the backgroundColor after the app has loaded through the plugin. It is not possible to update the configuration file of Capacitor before the app starts so we can only make do with a static background color at the start of the app each time.

This is the best we can do and I don't see any way to make further changes to alter the configuration file dynamically.

raineorshine:
No, not the splash screen actually. I'm referring to the overscroll background, the same background indicated in the original issue. You can see it in your video at the very end, after the splash screen goes away. It is set to black, even though the theme is Light.

msdewitt:
@raineorshine Well, maybe I didn't mean splash screen. So this is what I mean.
Screenshot 2024-12-29 at 4 25 50 PM
This file is the configuration file with a preset file for capacitor. It is configured during installation as the set backgroundColor.

During each startup of the app, the app looks at this configuration file and immediately fills in the background color with that value regardless of any changes to the background color from the previous apps changes. 

This happens before the app starts, so I can't even make changes inside the mobile app or pull from app storage to see what the personal settings of the background color are set too.

Now after the app is loaded, only then can we update the background color through the plugin.

Here is the plugin code example:
Screenshot 2024-12-29 at 4 28 41 PM


So at the start of each startup of a capacitor app, the webviewBackground is set to the color in the Capacitor configuration file.

The capacitor configuration file is not designed to be updated dynamically and it should not be stored anywhere accessible on the IOS device itself to update it. I think the configuration values are set as the default during the build of the app in CapacitorBridge file.

So my conclusion is that this change is not possible due to how capacitor is setup.

raineorshine:
Thanks for the additional explanation.

Would it be possible to set the overscroll background color after the app has loaded, and we have had time to read the theme?

Here's what I'm seeing:

https://github.com/user-attachments/assets/3481c903-0cb9-47ba-be62-5dcc234d3b8d

msdewitt:
For some reason that mp4 is not working for me.

We do set the background color right after the app loads, which happens right after the webview loads.

In capacitor when the app launches it first loads -> Webview -> App
The backgroundColor default is first set in the Webview and then update when the App loads to the users personal settings. So that if you change it to white background and close the app, it will be a white background when you open it again. Its just that the white background is updated at the [App] part of the loading configuration so if the default of the [Webview] is black, then it will first load a black webview and change to a white after.

raineorshine:
How about this? What I'm seeing is that the overscroll background color is black when the Theme is white after the app loads and after the splash screen is gone, as the user is using the app.

https://github.com/user-attachments/assets/9464a660-09e3-4b05-ad47-e322b3f56e9f

msdewitt:
Wow, that's really strange. I'm glad you caught that. That shouldn't be happening. When I pulled the latest code and pushed up a fix for it. For some reason, the setting of the theme was not happening and I needed read it back in in the AppComponent.

src/components/modals/Settings.tsx
```diff
@@ -119,7 +119,16 @@ const ModalSettings = () => {
...
         <div className={css({ textAlign: 'center' })}>
-          <ActionButton key='close' title='Close' {...fastClick(() => close())} />
+          <ActionButton
+            key='close'
+            title='Close'
+            {...fastClick(() => {
+              close(true)
+              // if (Capacitor.isNativePlatform()) {
+              //   window.location.reload()
+              // }
```
trevinhofmann, line 129: Is this comment unintentional, likely just used during testing? I believe it should be removed now.

packages/webview/dist/esm/index.js
```diff
@@ -0,0 +1,7 @@
+import { registerPlugin } from '@capacitor/core';
+const WebviewBackground = registerPlugin('WebviewBackground', {
+    web: () => import('./web').then((m) => new m.WebviewBackgroundWeb()),
+});
+export * from './definitions';
+export { WebviewBackground };
+//# sourceMappingURL=index.js.map
```
trevinhofmann, line 7: I would strongly prefer not to include `dist` files in the repository, and to include building this package as part of building the main project. There are a variety of reasons (repository bloat, merge conflicts more likely, built code more difficult to review for security issues, difficulty verifying that the build matches the source, etc.) for this, but they're relatively unimportant for this particular package due to its small size. I'll defer to @raineorshine who [previously indicated](https://github.com/cybersemics/em/pull/2594#issuecomment-2564403642) that this is probably fine 👍

packages/webview/package.json
```diff
@@ -0,0 +1,75 @@
...
+  "name": "webview-background",
+  "version": "0.0.2",
+  "description": "A small plugin to change the background of a webview.",
+  "main": "dist/plugin.cjs.js",
+  "module": "dist/esm/index.js",
+  "types": "dist/esm/index.d.ts",
+  "unpkg": "dist/plugin.js",
+  "files": [
+    "android/src/main/",
+    "android/build.gradle",
```
trevinhofmann, line 11: Is there an `android` directory? I don't see one.

packages/webview/README.md
```diff
@@ -0,0 +1,35 @@
+# webview-background
+
+A small plugin to change the background of a webview.
+
+## Install
+
+```bash
+npm install webview-background
```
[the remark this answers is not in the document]
raineorshine: Yes, it would be preferable to use the same package manager throughout the project.

Originally suggested here: https://github.com/cybersemics/em/pull/2594#discussion_r1864315059

packages/webview/dist/esm/index.js
raineorshine: That was my intuition as well, so I'm glad you said something. I know that @msdewitt did this to simplify the `webview` package release process, but I would love to discuss the feasibility of building it from source within the root project's build pipeline.

packages/webview/package.json
msdewitt: Its not needed for this plugin. We can add it later if needed. I deleted it since the fix seemed to only be for IOS.

packages/webview/dist/esm/index.js
msdewitt: Awesome

packages/webview/package.json
trevinhofmann: Makes sense. We can remove the `android` files from this list in `package.json`.

msdewitt:
Alright, I put in the fixes you requested @trevinhofmann . I had to go and change my approach to hide the contentInset on IOS with an ios only CSS adjustments.

I hope to get the approved soon after these changes. I'll rebase after the changes are confirmed in this rework.

packages/webview/package.json
trevinhofmann: Still seeing these which can be removed:

```
    "android/src/main/",
    "android/build.gradle",
```

packages/webview/dist/esm/index.js
trevinhofmann: @msdewitt -- Would it be feasible to build from source (as part of the main project's `yarn build`) rather than committing the `dist` build to the repo? I imagine we could update the main `package.json` `build` command to something like:

```
    "build": "yarn build:webview && yarn build:styles && vite build",
```

Then add the `build:webview` script in there as well to build this package from source.

packages/webview/README.md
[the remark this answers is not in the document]
trevinhofmann: @msdewitt -- We still have the `package-lock.json` / npm configuration in this package. Can we please switch to yarn?

src/e2e/puppeteer-environment.ts
```diff
@@ -29,8 +29,6 @@ const PuppeteerEnvironment: Environment = {
       .catch((err: Error) => {
         // using `console.log` here to avoid errors or logs being swallowed by vitest
         // all of `console.error`, `console.warn` and `console.info` don't show up in the terminal
-        // eslint-disable-next-line no-console
-        console.log('Could not connect to browserless.')
```
trevinhofmann, line 33: Is this change relevant to the PR?

src/e2e/puppeteer/setup.ts
```diff
@@ -50,8 +50,6 @@ const setup = async ({
         break
       case 'info':
       case 'log':
-        // eslint-disable-next-line no-console
-        console[messageType](text)
```
trevinhofmann, line 54: Is this change relevant to the PR?

packages/webview/dist/esm/web.js
```diff
@@ -0,0 +1,6 @@
+import { WebPlugin } from '@capacitor/core';
+export class WebviewBackgroundWeb extends WebPlugin {
+    async changeBackgroundColor(options) {
+        console.log('Color changed: ', options.color);
```
trevinhofmann, line 4: Looks like this log was removed [as requested](https://github.com/cybersemics/em/pull/2594#discussion_r1859023507) from the source (`packages/webview/src/web.ts`) but then not rebuilt and included in the build here.

One example of why we should build from source rather than including the `dist` in the repo :)

src/components/modals/ModalComponent.tsx
```diff
@@ -66,9 +68,12 @@ class ModalComponent extends React.Component<ModalProps> {
     window.addEventListener('keydown', this.onKeyDown, true)
   }
 
-  close = () => {
+  close = (fullReload: any = false) => {
     this.animateAndClose!()
     this.props.onClose?.()
+    if (Capacitor.isNativePlatform() && fullReload) {
+      window.location.reload()
```
trevinhofmann, line 75: I don't believe we should be doing a full page reload here. The only cases where we have those at the moment are for unrecoverable error handling.

Is there a particular reason for doing this? If I understand correctly, it's doing a full reload every time the Settings modal is closed.

src/components/Toolbar.tsx
```diff
@@ -221,9 +222,11 @@ const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected }) => {
           }),
         )}
         style={{
+          // height: fontSize + 300,
```
trevinhofmann, line 225: Unintentional comment? This can be removed.

trevinhofmann (changes requested):
Thank you, @msdewitt!

These issues appear to be resolved:
> 1. Initial black overscroll is not resolved until first interaction
> 2. Thoughts displayed above toolbar

The `node_modules` commits will still need to be rebased/squashed for (3).

---

## 4. Toolbar has been shifted down

The toolbar at the top of the page has been shifted down considerably and needs to be positioned back to its original location.

https://github.com/user-attachments/assets/60f0e2f8-757e-4772-aa53-1e68aedeb9a2

src/e2e/puppeteer-environment.ts
msdewitt: Yes, for some reason lint didn't like it after some changes. Most likely after I merged with main.

src/e2e/puppeteer/setup.ts
msdewitt: Same as above. Lint didn't like it.

src/components/modals/ModalComponent.tsx
msdewitt: We have already had discussions about this in this pr. Its unfortunate that we have to do this, but it is necessary. We can only limit the changes down to the very minimal as possible. (I.E. Only in the settings menu with a flag set as true.)

raineorshine: This is due to a limitation in changing the background color through the WebView. It's bad, but it's the only way to update the overscroll background color from what I understand.

It should reload the page only when the theme setting has changed.

msdewitt:
@trevinhofmann  Please tell me the IOS version you are using to test that toolbar being moved down. In order to cope with your previous review changes about not showing the notes when scrolling down, I had to add a feature which alters the CSS to make the navbar look like it is showing so that it is not covered by the native IOS statusBar. 

For some reason, the device you are using is showing that change, but the native IOS StatusBar already has a cover, therefor showing it as too big.

src/e2e/puppeteer-environment.ts
raineorshine: We probably want to keep this logging in, as it informs the user if there is any issue with connecting to browserless. The linter does not complain in `main`, so I suspect that this is a regression.

trevinhofmann:
I am using an iPhone 14 Pro with iOS version 18.1.1.

msdewitt:
Alright, maybe its an issue for anything below Ios 15 I need to check.

msdewitt:
@trevinhofmann Could you try and run the simulator with IOS version 17? I think the issue happens with versions later than 18.0, but I need to confirm that. I tried to replicate it locally with an Iphone 14 but it didn't show the same screen error as your screen capture showed.

raineorshine:
I can confirm the toolbar is shifted down on iOS 18.1 and iOS 17.5 in XCode Simulator.

msdewitt:
@raineorshine  How about IOS 17.4? I am using 17.4 on Iphone 15.

raineorshine:
Each Simulator runtime is ~6-8 GB so I'm not able to download additional minor versions to test. However, I am able to reproduce the issue on the following versions:

- iOS 16.4 on iPhone 14 Pro (Simulator)
- iOS 17.5 on iPhone 15 (Simulator)
- iOS 17.5 on iPhone 15 Pro (Simulator)
- iOS 18.1 on iPhone 16 (Simulator)
- iOS 18.1 on iPhone 16 Pro (Simulator)
- iOS 18.1 on iPhone 15 Pro (physical device)

trevinhofmann:
@msdewitt -- Thank you again for the PR. The toolbar reposition appears to be reproducible on several iOS versions. Please let me know if you need any additional assistance.

trevinhofmann:
Keeping this branch/PR for posterity, but I've rebased and squashed to open a new PR here: #2836

PR #2596, opened 2024-11-17
Fix/blur shrink ondelete

### LINKED ISSUE #2527: Animation: Blur and shrink on delete/collapseContext
- [x] Blocked by: #2526

## Steps to Reproduce

```
- a
- b
- c
```

Delete `b`.

## Current Behavior

`b` fades out according to the default unmount animation.

## Expected Behavior

The thought should fade out, blur, and shrink (scale3d) when the following commands are executed:

- Delete
- Collapse Context

zukilover:
Fixes #2527 

Changed animation name from `very-fast-fade` to `very-fast-morph` as it also includes other transformation and filter as well, i.e. `scale3d` and `blur`.

This PR depends on https://github.com/cybersemics/em/pull/2581 since it needs the newly added `CSSTransition`.

### DIFF
--- a/src/components/Subthought.tsx
+++ b/src/components/Subthought.tsx
@@ -116,12 +116,28 @@ const Subthought = ({
     ref.current.style.opacity = opacity
   })
 
+  /**
+   * Cleans up editable classes from the provided HTML string.
+   *
+   * @param htmlString - The HTML string to clean up.
+   * @returns The cleaned HTML string.
+   */
+  const cleanUpEditableClasses = (htmlString: string) => {
+    const container = document.createElement('div')
+    container.innerHTML = htmlString
+    const editableElements = container.querySelectorAll('.editable--preventAutoscroll_true')
+    editableElements.forEach(element => {
+      element.classList.remove('editable--preventAutoscroll_true')
+    })
+    return container.innerHTML
+  }
+
   // Capture the static HTML string when the thought is first rendered
   useEffect(() => {
     if (thought && ref.current) {
-      cachedHTMLRef.current = ref.current.innerHTML
+      cachedHTMLRef.current = cleanUpEditableClasses(ref.current.innerHTML)
     }
-  }, [thought])
+  }, [thought, ref])
 
   // If the thought is deleted, return the cached static HTML from the ref
   if (!thought && cachedHTMLRef.current) {
--- a/src/components/Subthought.tsx
+++ b/src/components/Subthought.tsx
@@ -6,6 +6,7 @@ import LazyEnv from '../@types/LazyEnv'
 import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import ThoughtId from '../@types/ThoughtId'
+import useCachedNode from '../hooks/useCachedNode'
 import useChangeRef from '../hooks/useChangeRef'
 import attributeEquals from '../selectors/attributeEquals'
 import findFirstEnvContextWithZoom from '../selectors/findFirstEnvContextWithZoom'
@@ -58,9 +59,9 @@ const Subthought = ({
 }) => {
   const state = store.getState()
   const ref = useRef<HTMLDivElement>(null)
-  // Cache the DOM before it is deleted
-  const cachedHTMLRef = useRef<string | null>(null)
   const thought = useSelector(state => getThoughtById(state, head(simplePath)), shallowEqual)
+  // Cache the DOM before it is deleted
+  const cachedHTMLRef = useCachedNode({ thought, elementRef: ref })
   const noOtherContexts = useSelector(
     state => thought && isContextViewActive(state, simplePath) && getContexts(state, thought.value).length <= 1,
   )
@@ -115,29 +116,6 @@ const Subthought = ({
     // start opacity at 0 and set to actual opacity in useEffect
     ref.current.style.opacity = opacity
   })
-
-  /**
-   * Cleans up editable classes from the provided HTML string.
-   *
-   * @param htmlString - The HTML string to clean up.
-   * @returns The cleaned HTML string.
-   */
-  const cleanUpEditableClasses = (htmlString: string) => {
-    const container = document.createElement('div')
-    container.innerHTML = htmlString
-    const editableElements = container.querySelectorAll('.editable--preventAutoscroll_true')
-    editableElements.forEach(element => {
-      element.classList.remove('editable--preventAutoscroll_true')
-    })
-    return container.innerHTML
-  }
-
-  // Capture the static HTML string when the thought is first rendered
-  useEffect(() => {
-    if (thought && ref.current) {
-      cachedHTMLRef.current = cleanUpEditableClasses(ref.current.innerHTML)
-    }
-  }, [thought, ref])
 
   // If the thought is deleted, return the cached static HTML from the ref
   if (!thought && cachedHTMLRef.current) {
--- a/src/components/Subthought.tsx
+++ b/src/components/Subthought.tsx
@@ -58,6 +58,8 @@ const Subthought = ({
 }) => {
   const state = store.getState()
   const ref = useRef<HTMLDivElement>(null)
+  // Cache the DOM before it is deleted
+  const cachedHTMLRef = useRef<string | null>(null)
   const thought = useSelector(state => getThoughtById(state, head(simplePath)), shallowEqual)
   const noOtherContexts = useSelector(
     state => thought && isContextViewActive(state, simplePath) && getContexts(state, thought.value).length <= 1,
@@ -113,6 +115,18 @@ const Subthought = ({
     // start opacity at 0 and set to actual opacity in useEffect
     ref.current.style.opacity = opacity
   })
+
+  // Capture the static HTML string when the thought is first rendered
+  useEffect(() => {
+    if (thought && ref.current) {
+      cachedHTMLRef.current = ref.current.innerHTML
+    }
+  }, [thought])
+
+  // If the thought is deleted, return the cached static HTML from the ref
+  if (!thought && cachedHTMLRef.current) {
+    return <div dangerouslySetInnerHTML={{ __html: cachedHTMLRef.current }} />
+  }
 
   // Short circuit if thought has already been removed.
   // This can occur in a re-render even when thought is defined in the parent component.
--- a/src/components/Subthought.tsx
+++ b/src/components/Subthought.tsx
@@ -6,7 +6,7 @@ import LazyEnv from '../@types/LazyEnv'
 import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import ThoughtId from '../@types/ThoughtId'
-import useCachedNode from '../hooks/useCachedNode'
+import useCachedThoughtHtml from '../hooks/useCachedThoughtHtml'
 import useChangeRef from '../hooks/useChangeRef'
 import attributeEquals from '../selectors/attributeEquals'
 import findFirstEnvContextWithZoom from '../selectors/findFirstEnvContextWithZoom'
@@ -60,8 +60,8 @@ const Subthought = ({
   const state = store.getState()
   const ref = useRef<HTMLDivElement>(null)
   const thought = useSelector(state => getThoughtById(state, head(simplePath)), shallowEqual)
-  // Cache the DOM before it is deleted
-  const cachedHTMLRef = useCachedNode({ thought, elementRef: ref })
+  // Cache the thought HTML before it is deleted so that we can animate on unmount
+  const cachedThoughtHtmlRef = useCachedThoughtHtml({ thought, elementRef: ref })
   const noOtherContexts = useSelector(
     state => thought && isContextViewActive(state, simplePath) && getContexts(state, thought.value).length <= 1,
   )
@@ -117,14 +117,12 @@ const Subthought = ({
     ref.current.style.opacity = opacity
   })
 
-  // If the thought is deleted, return the cached static HTML from the ref
-  if (!thought && cachedHTMLRef.current) {
-    return <div dangerouslySetInnerHTML={{ __html: cachedHTMLRef.current }} />
+  // If the thought has unmounted, return the cached static HTML from the ref so that it can animate out.
+  if (!thought) {
+    return cachedThoughtHtmlRef.current ? (
+      <div dangerouslySetInnerHTML={{ __html: cachedThoughtHtmlRef.current }} />
+    ) : null
   }
-
-  // Short circuit if thought has already been removed.
-  // This can occur in a re-render even when thought is defined in the parent component.
-  if (!thought) return null
 
   return (
     <>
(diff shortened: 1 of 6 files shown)


raineorshine:
Thanks! Let's wait for #2581 before reviewing since merge conflicts are expected. Please rebase on upstream main and resolve merge conflicts once ready.

zukilover:
@raineorshine I'm still trying to figure out how to capture the thought data while it's being deleted. The short-circuit here prevents animating during the delete process, causing it to appear empty before the animation finishes:

https://github.com/cybersemics/em/blob/be45dec8c0e7dcd4b1012cc7045e35f884821a4d/src/components/VirtualThought.tsx#L201-L203

raineorshine:
Removed the short circuit in 0d71558434. Hope that helps!

zukilover:
Since the `Subthought` component currently short-circuits and returns `null` when the `thought` is `undefined`, I've implemented an alternative approach by caching the DOM of the `Subthought` before it gets deleted. This allows the cached DOM to remain visible during the 80ms `nodeDissolve` transition, enabling a smooth animation effect. By caching the DOM, we ensure that the transition is visible without triggering any store-related logic or reactivity, keeping the animation independent of the state management process.

src/components/Subthought.tsx
```diff
@@ -115,6 +117,18 @@ const Subthought = ({
...
+  // Capture the static HTML string when the thought is first rendered
+  useEffect(() => {
+    if (thought && ref.current) {
+      cachedHTMLRef.current = ref.current.innerHTML
+    }
+  }, [thought])
+
+  // If the thought is deleted, return the cached static HTML from the ref
+  if (!thought && cachedHTMLRef.current) {
+    return <div dangerouslySetInnerHTML={{ __html: cachedHTMLRef.current }} />
```
trevinhofmann, line 129: Personally, I'm generally okay with carefully using `dangerouslySetInnerHTML`. In this case, I don't see much risk related to e.g. XSS vulnerabilities since both the input and rendered HTML come from the user.

It could _maybe_ be argued that a user could be tricked into pasting and then deleting a thought with a malicious `<script>` in it, but we seem to be escaping `<` and `>` in thoughts regardless.

Just to be sure, have you considered if there are other options?

trevinhofmann (changes requested):
The shrink/blur/fade animation for deletion generally seems to work as expected, although it's a bit difficult to notice at full speed:

https://github.com/user-attachments/assets/5003faff-dca6-4c02-a2a7-32cffe15e086

Slowing the animation down frame-by-frame confirms that the animations are working, except for one issue. At the beginning of the animation, the thought is empty for ~1 frame:

https://github.com/user-attachments/assets/07a4b37b-1236-4b8f-97db-386087add3fe

---

#2527 requests that this animation apply for both deletions and collapsing of context. Do you plan to add the animations for collapsing context in this PR as well?

src/components/Subthought.tsx
zukilover: @trevinhofmann I've tried a few different approaches before including:
- Storing thought directly to `useRef`, but it was ineffective since `Subthought` has a short-circuit (see below).
- Removing the short-circuit in `Subthought`, turned out to be too complicated since the `Thought` component still relies on the redux store.

zukilover: I have fixed the missing frame. This came from `editable--preventAutoscroll_true` class that animates the editable element's opacity. Since it's just a cache and it should always be visible, I perform a cleanup for that class before rendering it to the cached HTML.

I've also removed the `isLastActionDeleteThought` check so that it applies to the collapsing of context too.

trevinhofmann (changes requested):
Thank you very much, @zukilover!

The animation is now being applied correctly to collapsing context 👍

These changes also resolved the initial "blink" before the animation starts. However, I am now noticing that the other thoughts do not animate up until the blur/shrink/fade animation completes:

https://github.com/user-attachments/assets/8d71c850-a977-4a7d-b83c-3ddeeb3a0409

I also noticed that the thoughts do not get cleaned up / removed from the DOM after they've been deleted. Here are three deleted thoughts still in the DOM, which should probably be removed once the animation completes:

Screenshot 2024-12-27 at 6 54 53 PM

zukilover:
Hi @trevinhofmann I don't see how it's possible to animate the next thought up before the previous one has not been completely deleted because `transitionGroupsProps.in` —which triggers the animation— will be effective when `treeThoughtsPositioned` changes. If we compare it to `main` branch, next node will move up when `nodeFadeOut` is completed too. Let me know if this makes sense.

Also the DOM not being removed also happens in the `main` branch, I wonder if this should be treated a separate issue.

raineorshine:
When a thought is deleted, `treeThoughtsPositioned` should immediately be updated and bump the y positions of subsequent thoughts, triggering the translate animation. You can see the other thoughts moving up even as `b` is fading out in this video from `main`. This was discussed and refined in https://github.com/cybersemics/em/pull/2581#pullrequestreview-2480043247 (4):

https://github.com/user-attachments/assets/e1866e6c-126f-4ea9-900d-ad07109a7a15

When I was setting up the video I did notice that there were some cases where the fade out animation did not occur at all, which would seem to be either a regression or bug in #2581. The above video shows the expected behavior with the fade out. The intention of #2527 is to replace the fade out animation with fade out + shrink + scale without affecting the timing or other animations like other thoughts translating up.

@trevinhofmann @zukilover Could you both confirm that the fade out animation is working for you on delete in `main`? If #2581 is buggy it could make solving this issue more difficult.

> Also the DOM not being removed also happens in the `main` branch, I wonder if this should be treated a separate issue.

Thanks for confirming that. Created #2751.

trevinhofmann:
> @trevinhofmann @zukilover Could you both confirm that the fade out animation is working for you on delete in `main`? If #2581 is buggy it could make solving this issue more difficult. I recommend deleting a few thoughts in a row, from the beginning, middle, and end of a list to test it out.

Testing `main` (42b1ef55abd5457ecbbbc39512be3dc6a1682fb5), I see:
A) ✅ Thoughts fade when collapsing context
B) ❌ Thoughts do **not** fade when deleting
C) ✅ Subsequent thoughts immediately shift up when collapsing context
D) ✅ Subsequent thoughts immediately shift up when deleting

Testing `fix/blur-shrink-ondelete` (2a5174240bd9a3a6b00a836d646ab4beae217d99), I see:
A) ✅ Thoughts fade/blur/shrink when collapsing context
B) ✅ Thoughts fade/blur/shrink when deleting
C) ✅ Subsequent thoughts immediately shift up when collapsing context
D) ✅ Subsequent thoughts immediately shift up when deleting

I believe #2581 and #2526 were specifically for collapsing context, and I'm not seeing issues there. Deleted thoughts are not fading on `main`, but appear to be fading/blurring/shrinking correctly now in this PR.

> Hi @trevinhofmann I don't see how it's possible to animate the next thought up before the previous one has not been completely deleted because `transitionGroupsProps.in` —which triggers the animation— will be effective when `treeThoughtsPositioned` changes. If we compare it to `main` branch, next node will move up when `nodeFadeOut` is completed too. Let me know if this makes sense.

I'm not sure if that makes sense to me. The (C) and (D) cases work as expected for me on `main`, but were not working as expected in this PR when I made [this comment](https://github.com/cybersemics/em/pull/2596#pullrequestreview-2524454939).

That said, (C) and (D) appear to have been fixed in the current version of this PR 🎉. Have you made any recent changes that were expected to resolve this?

> Also the DOM not being removed also happens in the `main` branch, I wonder if this should be treated a separate issue.

Good catch! Apologies, and I agree with you and @raineorshine that this should be a separate issue (#2751) in that case.

trevinhofmann (approved):
Marking this as approved from my end. With the current changes, the code still looks fine to me, and the fade/blur/shrink animations (and shifting up of subsequent thoughts) are now working as expected for me on every deletion and collapsing context. The animation is quite fast, so you may need to step frame-by-frame to confirm:

https://github.com/user-attachments/assets/f5cea4d3-7d40-41d9-af5d-2536ad2b44c3

---

@raineorshine -- Can you please retest the cases where the fade was not working on `main` to ensure that the fade/blur/shrink works as expected for you with this PR? I haven't been able to reproduce the missing animation here.

@zukilover -- Can you please explain what changes were made since the last review? Due to the rebase (which is fine), I can't see what exactly has changed since then. It would be nice to know the changes because [your comment](https://github.com/cybersemics/em/pull/2596#issuecomment-2566203724) indicated that you weren't able to fix the shifting of subsequent thoughts, but the latest changes seem to have fixed it.

Thanks!

zukilover:
@trevinhofmann, the only change in the latest rebase was the removal of the type assertion here:

https://github.com/cybersemics/em/blob/1402dca1cbaf33b2eebe32f17d9ffc231d1314a8/src/components/Subthought.tsx#L63-L64

Additionally, I assume the anomaly you pointed out here:
https://github.com/cybersemics/em/pull/2596#pullrequestreview-2524454939 was caused by slowing down the duration (likely layoutNodeAnimation), which made the transition appear out-of-sync.

Here's the latest result from this PR. Please confirm if it works as expected:

https://github.com/user-attachments/assets/67854401-2491-43d7-98eb-0f16c36bd284

src/components/Subthought.tsx
```diff
@@ -114,6 +116,34 @@ const Subthought = ({
...
+  /**
+   * Cleans up editable classes from the provided HTML string.
+   *
+   * @param htmlString - The HTML string to clean up.
+   * @returns The cleaned HTML string.
+   */
+  const cleanUpEditableClasses = (htmlString: string) => {
+    const container = document.createElement('div')
+    container.innerHTML = htmlString
+    const editableElements = container.querySelectorAll('.editable--preventAutoscroll_true')
```
raineorshine, line 128: `.editable--preventAutoscroll_true` is autogenerated by PandaCSS so it's dangerous to hardcode it into the component. If the [recipe variant](https://github.com/cybersemics/em/blob/42b1ef55abd5457ecbbbc39512be3dc6a1682fb5/src/recipes/editable.ts#L41) changes, this will break without us knowing.

I believe PandaCSS has a way to programmatically get the className of a recipe variant, although I don't know the technique off the top of my head. Could you use that here? If you can't find it let me know.

src/components/Subthought.tsx
```diff
@@ -114,6 +116,34 @@ const Subthought = ({
...
+    })
+    return container.innerHTML
+  }
+
+  // Capture the static HTML string when the thought is first rendered
+  useEffect(() => {
+    if (thought && ref.current) {
+      cachedHTMLRef.current = cleanUpEditableClasses(ref.current.innerHTML)
+    }
+  }, [thought, ref])
```
raineorshine, line 140: What do you think about putting the `useEffect`, `cleanUpEditableClasses`, and refs in a custom hook? That might help keep this component tidy.

raineorshine (changes requested):
@trevinhofmann Thanks for the thorough testing. I can also confirm that the fade/shrink/blur is working as expected now. The cases where there is no fade/shrink/blur are explainable:

- The animation happens so fast that sometimes my screen recorder doesn't even capture it. Upon closer review I can see the small deleted thought very faintly in one of the frames which confirms that it is animating.
- Deleting the very last thought in the thoughtspace does not trigger the fade/shrink blur as it causes the "Hit the Enter key to add a new thought" message to appear. This will never happen in practice once the user has at least one thought, so we can ignore this case.

> I believe #2581 and #2526 were specifically for collapsing context, and I'm not seeing issues there. Deleted thoughts are not fading on `main`, but appear to be fading/blurring/shrinking correctly now in this PR.

#2526 was intended to create a fade out for all unmounted thoughts, not just collapse. This PR overrides that behavior for delete.

raineorshine (changes requested):
Thanks for making the requested changes! The new approach is more robust and type safe.

Looking over the implementation again, I have some reservations about the way the DOM is utilized in this performance critical part of the application. `document.createElement`, `document.querySelectorAll`, and `element.classList.remove` could all be avoided with a simple `.replace`. I have to say that I would prefer that approach. It's just not worth the risk of impacting the performance here.

zukilover:
Sure, @raineorshine. I've refactored the function and use a regular expression for string replacement, which can be faster and more efficient than iterating through an array of CSS classes.

For most practical scenarios, the performance difference is likely to be negligible.
Thanks.

src/hooks/useCachedNode.ts
```diff
@@ -0,0 +1,58 @@
...
+      editableRecipe({
+        preventAutoscroll: true,
+      }),
+    ).split(' ')
+
+    // Filter classes containing '--' to identify variant classes
+    const modifiers = editableClass.filter(cls => cls.includes('--'))
+
+    // Create a regular expression to match and remove BEM modifier (variant) classes
+    const regex = new RegExp(`\\b(${modifiers.join('|')})\\b`, 'g')
```
raineorshine, line 41: Would it be possible to instantiate this regex outside the function, or even at compile-time? `new RegExp` is known to be slow, and the className will never change in a given build.

raineorshine (changes requested):
Thank you!

Yes, I totally agree. You'll have to trust me that this is one of those exceptions! Performance of thought rendering is critical in this application. There are already noticeable performance issues, so we don't want to inadvertently make it worse.

src/hooks/useCachedNode.ts
zukilover: I've moved the calculation out of the hook, to optimize the computation and reduce unnecessary overhead.

src/hooks/useCachedNode.ts
```diff
@@ -0,0 +1,57 @@
...
+const regex = new RegExp(`\\b(${modifiers.join('|')})\\b`, 'g')
+
+/**
+ * Custom hook to capture and cache the static HTML string of a node.
+ *
+ * @param thought - The thought object.
+ * @param elementRef - The ref of the node.
+ * @returns The cached HTML string.
+ */
+const useCachedNode = ({
```
raineorshine, line 26: I would suggest a more specific name for this hook, e.g. `useCachedThoughtHtml`, since it has several thought-specific pieces of functionality.

We can also add to the JSDOC to explain a little more about what it does and where it's useful.

src/hooks/useCachedNode.ts
```diff
@@ -0,0 +1,57 @@
...
+   *
+   * @param htmlString - The HTML string to clean up.
+   * @returns The cleaned HTML string.
+   */
+  const cleanUpEditableClasses = (htmlString: string) => {
+    // Replace BEM modifier classes with an empty string
+    return htmlString.replace(regex, '').trim()
+  }
+
+  // Capture the static HTML string when the thought is first rendered
```
raineorshine, line 47: The commend needs a slight tweak. It captures the HTML not just when the thought is first rendered, but whenever the `thought` changes.

src/hooks/useCachedNode.ts
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * Cleans up editable classes from the provided HTML string.
+   *
+   * @param htmlString - The HTML string to clean up.
+   * @returns The cleaned HTML string.
+   */
+  const cleanUpEditableClasses = (htmlString: string) => {
+    // Replace BEM modifier classes with an empty string
+    return htmlString.replace(regex, '').trim()
+  }
```
raineorshine, line 45: Now that this is just a single `.replace`, it doesn't need to be wrapped in a function.

raineorshine (approved):
Looks good, thanks!

I did some minor refactoring and renaming. I recommend taking a quick look over the changes.

PR #2601, opened 2024-11-18
PR 2589 merged with main

yangchristina:
#2589 merged with main

Cannot put CSSTransition inside wrapper component, must be in `map`

Changes: 
- timeout uses fastDuration instead of veryFastDuration
- replace `left {durations.${layoutDuration}} ease-out,top {durations.${layoutDuration}} ease-out`, with
```js
transition: isNewThought
          ? `left {durations.layoutNodeAnimationFastDuration} ease-out,top {durations.layoutNodeAnimationFastDuration} ease-out`
          : `left {durations.layoutNodeAnimationDuration} ease-out,top {durations.layoutNodeAnimationDuration} ease-out`,
```
(PandaCSS cannot statically extract`{durations.${layoutDuration}}`)


src/components/LayoutTree.tsx
```diff
@@ -434,14 +441,20 @@ const TreeNode = ({
   // Increasing margin-right of thought for filling gaps and moving the thought to the left by adding negative margin from right.
   const marginRight = isTableCol1 ? xCol2 - (width || 0) - x - (bulletWidth || 0) : 0
 
+  // Speed up the tree-node's transition by 50% on New (Sub)Thought only.
+  const isNewThought = value === ''
```
trevinhofmann, line 445: @raineorshine - Can you please clarify what is intended in this part of #1790? I believe the current functionality is incorrect, but it isn't obvious to me what the desired animations/durations are.

> Speed up the tree-node's transition (normally layoutNodeAnimationDuration) by 50% on New (Sub)Thought only.

This logic currently affects the animation duration when indenting to convert an empty thought into a subthought. It does not affect the animation when indenting a non-empty thought or when using the `newSubthought` action. Here is a demo of the three, with `layoutNodeAnimationDuration` increased for a more noticeable difference:

https://github.com/user-attachments/assets/e052d3b8-df04-4b64-b9e6-d10e7d358584

trevinhofmann (changes requested):
Thank you, @yangchristina!

This PR does not appear to have the same `1. Disappearing thoughts` issue as #2589 👍

However, I am not seeing a fade-in animation when creating a new thought. To be sure it wasn't just too fast for me to notice, I also tried changing the animation speed to `'slow'`.

https://github.com/user-attachments/assets/39ef4eb1-da86-4007-b3ee-d1582424096c

---

I don't believe this is being enforced (yet), but just want to also note that rebasing is generally preferred instead of merge commits.

yangchristina:
@raineorshine @trevinhofmann The fade-in animation works if the `div` with `nodeRef` is moved into LayoutTree. However, then this check `if (belowCursor && !isCursor && y > viewportBottom + height) return null` would not use the same `y_`, which seems to have been added for CSSTransitions.

raineorshine:
I'm sorry for the confusion, but it was actually #2581 that needed help with the merge, not #2589. I mentioned this in #2581, but I was not specific in my personal communication to you @yangchristina on Discord, so my apologies.

@zukilover made some progress on merging #2581, so we might not need this. I will mark this as draft for now and then we will merge the PR's one at a time, in order, until we get the desired result. Thanks for your patience.

yangchristina:
@raineorshine Sorry about this, I should have checked with you first. It's good it's been resolved now though!

PR #2603, opened 2024-11-20
Cursor moves up instead of caret moving up one line

### LINKED ISSUE #2512: Cursor moves up instead of caret moving up one line
## Steps to Reproduce

1. Paste the following multiline text into a new thought:
> As <b>em</b> is a highly optimized custom editor, control of the browser selection or text cursor ("caret") is critical. This document defines consistent terminology and provides an overview of browser selection usage in the codebase.
2. Create thought `a` above the multiline thought.
3. Set the caret at the beginning of the second line of the multiline text. (In the screenshot below it would be before the word "cursor", but it will vary depend on the width of your browser where it breaks.)
4. ArrowUp

## Current Behavior

The cursor moves to the previous thought.

image

## Expected Behavior

The cursor should not change, and the caret should move up one line.

image

## Notes

The complete expected multiline behavior is documented in #1660

ethan-james:
Fixes #2512

If the caret is at the beginning of a line, the selection seems to span from the end of the previous line to the start of the new line. This causes `window.getSelection().getRangeAt(0).getClientRects()` to return 2 rects, where the first has a y coordinate that corresponds to the previous line.

### DIFF
--- a/src/e2e/puppeteer/__tests__/caret-multiline.ts
+++ b/src/e2e/puppeteer/__tests__/caret-multiline.ts
@@ -263,4 +263,27 @@ describe('all platforms', () => {
     const offset = await getSelection().focusOffset
     expect(offset).toBe(0)
   })
+
+  // test case 11
+  it('on cursorUp, the caret should move from the start of the 2nd line in a multi-line cursor to the start of the 1st line in the same cursor.', async () => {
+    const multiLineCursor =
+      "Beautiful antique furnishings fill this quiet, comfortable flat across from the Acropolis museum. AC works great. It is in an heavily touristic area, but the convenience can't be beat. Highly recommended."
+    const importText = `
+  - a
+  - ${multiLineCursor}`
+
+    await paste(importText)
+
+    const editableNodeHandle = await waitForEditable(multiLineCursor)
+    await click(editableNodeHandle, { y: 1 })
+
+    await press('ArrowUp')
+
+    // the focus must be at the start of the multi-line cursor after cursor up
+    const textContext = await getSelection().focusNode?.textContent
+    expect(textContext).toBe(multiLineCursor)
+
+    const offset = await getSelection().focusOffset
+    expect(offset).toBe(0)
+  })
 })
--- a/src/device/selection.ts
+++ b/src/device/selection.ts
@@ -93,7 +93,7 @@ export const isOnFirstLine = (): boolean => {
   const clientRects = selection.getRangeAt(0).getClientRects()
   if (!clientRects?.length) return true
 
-  const { y: rangeY } = clientRects[0]
+  const { y: rangeY } = clientRects[clientRects.length - 1]
   if (!rangeY) return true
 
   const baseNodeParentEl = baseNode?.parentElement as HTMLElement


trevinhofmann (approved):
Thank you for the PR, @ethan-james.

This fix does resolve #2512 when testing locally:

https://github.com/user-attachments/assets/2827c5a6-ec94-44a8-a096-1aa5b113b897

---

@raineorshine, I noticed during testing that another case in #1660 is not working as expected:

> Cursor down from c and then immediately cursor down from `Beautiful antique...`: the cursor should stay on the same thought and caret should move the cursor down one line.

Rather than moving to the next line, the cursor moves to the next thought. This same issue appears in `main`, not caused by this PR.

https://github.com/user-attachments/assets/12ca7186-21f6-4429-bd77-a71e19b518eb

raineorshine:
Does that correspond to the skipped test? If so, it's a known issue. There should be an issue open for it, but I would have to track it down as it's not linked to #1660.

trevinhofmann:
Ah, yes. I believe that issue is covered by the skipped test.

src/e2e/puppeteer/__tests__/caret-multiline.ts
```diff
@@ -263,4 +263,27 @@ describe('all platforms', () => {
...
+    const multiLineCursor =
+      "Beautiful antique furnishings fill this quiet, comfortable flat across from the Acropolis museum. AC works great. It is in an heavily touristic area, but the convenience can't be beat. Highly recommended."
+    const importText = `
+  - a
+  - ${multiLineCursor}`
+
+    await paste(importText)
+
+    const editableNodeHandle = await waitForEditable(multiLineCursor)
+    await click(editableNodeHandle, { y: 1 })
```
raineorshine, line 278: `y` is specified in pixels, so this would click on the editable at (1,1), which I think would place the selection at the start of the 1st line, no?

Yet the caret is somehow still on the multiline thought after ArrowUp. What am I missing?

Note: You could ArrowDown first to get the caret onto the second line.

ethan-james: Oh, I guess I misunderstood how it worked a little bit. From `helpers/click.ts`:

```
  const coordinate = !offset
    ? {
        x: boundingBox.x + (edge === 'left' ? 1 : edge === 'right' ? boundingBox.width - 1 : boundingBox.width / 2),
        y: boundingBox.y + boundingBox.height / 2,
      }
    : await offsetCoordinates()
```

So it's clicking on the left of the bounding box, halfway down plus 1 pixel. Which in my case is the correct position. It sounds like ArrowDown is the way to go, thanks for pointing that out.

raineorshine: You're right!!! My bad. So it would click on (1, height/2 + 1). This would indeed click on the second line IF the thought wraps in the correct place. 

It would be great to make this test independent of the width of the browser window as we did in the other tests.

ethan-james: I updated it so that it uses ArrowDown. There are some test errors that appear unrelated to anything I changed.

raineorshine (approved):

Issue #1705, opened 2023-02-12
[mobile] Only disable gestures performed near the browser selection


raineorshine:
## Current Behavior

Currently, all gestures are disabled if any browser selection exists. 

This is done to avoid false positives while trying to move the beginning or end of a selection. 

## Expected Behavior

Only disable the gesture if it is performed ~~100px~~ 1em away from the selection (above or below), where there is no chance of false positive.

PR #2607, opened 2024-11-22
Don't cancel gesture if touch is > 2em pixels away

ethan-james:
#1705

Get the bounding client rect for the current selection (if applicable) and only abandon a gesture if the `touchstart` coordinates are within a box that extends ~~100 pixels~~ 2em out from the client rect.


src/e2e/puppeteer/__tests__/gestures.ts
```diff
@@ -0,0 +1,24 @@
...
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+const iPhone = KnownDevices['iPhone 15 Pro']
+
+describe('gestures', () => {
+  it.skip('when starting a gesture, the command palette will open', async () => {
+    const cursor = 'Hello'
+    const importText = `
+    - ${cursor}`
+
+    await page.emulate(iPhone)
```
ethan-james, line 15: @raineorshine @trevinhofmann I cannot for the life of me get puppeteer to simulate a selection while emulating a mobile browser. I'll give it another try on Monday, but please let me know if you have any experience getting it to work.

raineorshine: Thanks for trying! If it's difficult/impossible I'm okay skimping on test coverage here.

raineorshine: Did you try creating a selection with touch events, or by directly accessing the browser selection API?

ethan-james: I did try directly accessing the browser selection API, and found that it wouldn't hold onto the selection once it left the `page.evaluate` block. That might be worth another try, but I didn't have much luck the first time.

As for creating the selection with touch events, I tried to use a physical device to figure out how that would work, and it seemed like the only way to get a selection was to tap and then use the tooltip menu to "Select All". I don't think the tooltip menu appears in the Puppeteer emulator. If you know another, more reliable way to tap to create a a selection, please do let me know.

raineorshine: A double tap selects the word under the touch. I imagine you will run into the same issue though in which the selection is lost after the `page.evaluate`.

We will have to rely on manual testing for this one 👍

ethan-james: Double tap worked!

trevinhofmann (changes requested):
Thanks, @ethan-james! Overall, this is a nice implementation. I just have one possible issue with the cancellation zone, and the tests need to be updated/removed.

src/device/selection.ts
```diff
@@ -440,3 +440,29 @@ export const html = () => {
...
+
+  const rect = getBoundingClientRect()
+  if (!rect) return false
+
+  const left = rect.left - 100
+  const right = rect.right + 100
+  const top = rect.top - 100
+  const bottom = rect.bottom + 100
+
+  return x >= left && y >= top && x <= right && y <= bottom
```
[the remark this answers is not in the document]
raineorshine: Yes, I was thinking it would be a full width zone. That said, I don't mind the 100px on all sides at all. I think it still accomplishes the goal of preventing gestures from interfering with text selection.

raineorshine:
> Thank you for attempting to add this to the automated tests, but @raineorshine mentioned that it is okay to not include these if emulating a selection is unnecessarily challenging.

Yes, I think if the test is not working it's best to remove it from the codebase. We always have this PR to refer back to.

src/e2e/puppeteer/__tests__/gestures.ts
ethan-james: I actually wasn't able to get the `browserless/chrome` v1 Docker image working on my Macbook M1 (presumably due to issues with the ARM chipset) so I've been running the v2 `ghcr.io/browserless/chromium` image. Maybe this is the first time that I've found a difference between the two versions. I'm planning to get a new computer next year, and I'll choose something with fewer Docker compatibility issues. That said, upgrading the pipeline to use the new image might also be a worthwhile project.

src/device/selection.ts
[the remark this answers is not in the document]
ethan-james: I can make the change quickly if you decide you want it to be full width.

trevinhofmann (approved):
Thanks, @ethan-james!

raineorshine (approved):
Thanks! Works great and code looks good. Good separation of concerns.

FYI After playing with it on my phone, I made the distance dynamic and tightened it up to 2em. That should still be enough to avoid false positives, and it will be more forgiving for users that use a higher font size.

Issue #2510, opened 2024-10-28
Gap between thoughts


raineorshine:
There is a gap between thoughts that makes an undesirable space to click.

- [x] ~Identify the commit that caused the regression so that we can reverse it. This will have less risk than adding a new style.~

I traced the bug back as far as 6f36c5227c0e97a5b54820d8139f921c72929c15 and could not find a good commit. Either the bug has existed for a very long time, or it was introduced in a new version of Chrome. Either way, it makes sense now to troubleshoot on `main` and treat it as a new bug rather than a regression.

## Steps to Reproduce

```
- a
- b
- c
```

Slowly move the mouse from one thought to another and you will see the cursor change when in between thoughts.

## Current Behavior

There is a gap between thoughts.

<img width="331" alt="image" src="https://github.com/user-attachments/assets/089a5746-b75f-45db-bd4b-96eae894459a">

## Expected Behavior

Thoughts should be rendered flush so that there is no dead space in between. Clicking in between two thoughts should always set the caret on one of the two thoughts.

PR #2609, opened 2024-11-22
fix : gap between thoughts

RED-ROSE515:
Issue : Gap between thoughts #2510

Currently there is unexpected gaps between thoughts.
And it's because of clipPath in editable recipe.
So, I removed clipPath.


raineorshine (changes requested):
`clipPath` is there for a reason, which is documented in the inline comment. You will need to investigate what functionality it is providing, what is lost by its removal, and what an adequate replacement could be.

If it is no longer needed, you will need to provide a convincing argument as to why this has changed.

RED-ROSE515:
Yeah, currently, the clipPath is used to prevent incorrect selection behavior when clicking on the edges of an editable element.

RED-ROSE515:
@raineorshine , What do you think about adding margin top and bottom, instead of removing clipPath?

raineorshine:
I'm not sure. You will have test it and see if the desired selection behavior is preserved across platforms and font sizes.

RED-ROSE515:
@raineorshine , Could you tell me what is the original one, and what is new generated one?
drag-drop-hover-after-first-thought-of-column-one-1-diff

raineorshine:
If margin doesn't work, another idea is to shift the `y` positioning up since thoughts are absolutely positioned.

RED-ROSE515:
I think margin is good right now.

raineorshine:
Closing in favor of #2656.

Issue #2392, opened 2024-09-12
Capacitor: Haptic feedback


raineorshine:
Use the Capacitor Haptics API to provide physical feedback to the user.

https://capacitorjs.com/docs/apis/haptics

Haptic options:

- [impact](https://capacitorjs.com/docs/apis/haptics#impact) - light | medium | heavy
- [notification](https://capacitorjs.com/docs/apis/haptics#notification) - success | warning | error
- [vibrate](https://capacitorjs.com/docs/apis/haptics#vibrate) - duration
- [selectionStart](https://capacitorjs.com/docs/apis/haptics#selectionstart)
- [selectionChanged](https://capacitorjs.com/docs/apis/haptics#selectionchanged)
- [selectionEnd](https://capacitorjs.com/docs/apis/haptics#selectionend)

Haptic requirements (first iteration):

- Gesture
  - start: `selectionStart`
  - swipe: `selectionChanged`
  - end: `selectionEnd`
    - *except* Back/Forward: `impact` (light)
    - *except* Delete: `notification` (warning)
    - *except* Delete permanently: `notification` (warning)
  - cancel: `notification` (warning)
- Drag-and-drop
  - start: `selectionStart`
  - drop target changes: `selectionChanged`
  - end: `selectionEnd`
  - cancel: `notification` (warning)
- Hamburger menu
  - toggle: `impact` (light)
- Link
  - click: `impact` (light)

msdewitt:
@raineorshine I've been looking into this issue for a bit now and I have a few questions we can discuss about it. The majority of these Haptrics changes involve impact (light) and I found that we can make it uniform by adding it during the fastClick.ts file which is used throughout the app.

Is this something you are interested in? It adds a light vibration at almost every 'button action' throughout the app, including links, when you tap the screen anywhere to scroll when you open the hamburger menu, and in most other places.


The selectionStart and selectionEnd probably can't be handled in a generic way like this, but its still nice to find this. Tell me what you think.

raineorshine:
Yes, agreed!

The only one that I'm unsure of is "when you tap the screen anywhere to scroll". Do you mean when you tap the scroll zone on the right?

msdewitt:
@raineorshine, I added selectionChanged to the MultiGesture feature when a user uses the canvas, but I saw that it triggers each time the canvas is updated. Was this what you considered when you listed it on this ticket? Or were you thinking it would be just a selectionStart and selectionEnd event after the user finishes drawing?
[...]

Issue #2474, opened 2024-10-19
[mobile] First drag-and-drop does not work.


raineorshine:
Confirmed on:

- iOS Safari
- Chrome device emulator

## Steps to Reproduce

```
- a
- b
```

1. Paste the above thoughts and refresh the page.
2. Try to drag `a` below `b`.

## Current Behavior

Drag-and-drop is not activated on the first attempt. 

Thereafter it works as expected.

## Expected Behavior

Drag-and-drop should work on the first attempt.

PR #2611, opened 2024-11-22
Haptics

msdewitt:
This PR is one which will change as we test to further perfect the acceptance criteria mentioned in [issue-2392](https://github.com/cybersemics/em/issues/2392).


src/util/fastClick.ts
```diff
@@ -27,7 +28,7 @@ const fastClick = isTouch
           const y = e.touches[0].clientY
           touchStart = { x, y }
         }
-
+        Haptics.impact({ style: ImpactStyle.Light })
```
raineorshine, line 31: It might make more sense to place the hapic in `onTouchEnd`, since that is where the handler fires. Then the haptic feedback will occur at the same time as the execution of the handler. If done in `onTouchStart`, there will be a gap between the haptic and the execution, and haptics will unintentionally occur on cancelled clicks (such as a drag or scroll).

msdewitt: Great idea. I immediately saw the touch triggers happening at regular intervals after this change.

msdewitt:
@raineorshine I have managed to add the haptic changes in at each of the listed places. I have a list below of those mentioned in the ticket. The ones with _Italics_ are completed, while the others are either not completed due to being unable to reproduce the functionality or because it is still being discussed (Like vibrations on scrolling.)

@trevinhofmann If you have time to test some of these that would be great.

* _Gesture
    * start: selectionStart
    * swipe: selectionChanged
    * end: selectionEnd
        * except Back/Forward: impact (light)
        * except Delete: notification (warning)
        * except Delete permanently: notification (warning)
    * cancel: notification (warning)_
* Scroll
    * Continuous impact? Like old iPod or Nest wheel.
_* Long tap
    * start: selectionStart_
* Drag-and-drop
    _* start: selectionStart_
    * drop target changes: selectionChanged ?
    _* end: selectionEnd_
    * cancel: notification (warning)
_* Hamburger menu
    * toggle: impact (light)
* Link
    * click: impact (light)
* Button
    * click: impact (light)_


Changes and optimizations:
Instead of manually picking impacts: It was changed to adding an impact(light) in the ticket discussion. Now all touches should have a light impact added to them.

raineorshine:
> @raineorshine I have managed to add the haptic changes in at each of the listed places. I have a list below of those mentioned in the ticket. The ones with _Italics_ are completed, while the others are either not completed due to being unable to reproduce the functionality or because it is still being discussed (Like vibrations on scrolling.)

Great, thanks!

> Instead of manually picking impacts: It was changed to adding an impact(light) in the ticket discussion. Now all touches should have a light impact added to them.

Can you explain "all touches"? Is that all touchstart events, all touchend events, or both? My understanding was that we discussed adding impact(light) to `fastClick` execution.

> Needs discussion: Scroll Continuous impact? Like old iPod or Nest wheel.
> 
> Q: I don't think we should add a continuous impact. We already have light impacts after each tap.

I'm not really sure what is possible with Capacitor haptics as I haven't experienced yet myself, but the idea here was to add a very tactile experience similar to the old iPod or Nest wheel where there is continuous haptic feedback that is proportional to velocity. We may have to experiment a bit to get it right. I'm open to any ideas you have based on your initial experience with the Haptics API.

msdewitt:
@raineorshine 
What I mean by all touches is what we discussed with 'fastclick'. Not all touchStart events. Its just that so many places in the code use fastclick that a light impact will generate when lightly touching the scroll, the menu, the buttons, and the links.

Pretty much everywhere at that point.

As for the tactile feedback, Maybe we could add an impact when a user scrolls an interval distance. Therefore, the faster they scroll the faster the interval of the vibration. I don't recommend adding this until after further testing is done on how it feels to use in day-to-day. I figure that adding too many vibrations and impacts when touching and scrolling will cause the user to feel overwhelmed.

If you are not worried bout this, please ignore my recommendation and I can implement the scroll feature as I just described, based on interval impacts.

raineorshine:
> What I mean by all touches is what we discussed with 'fastclick'. Not all touchStart events. Its just that so many places in the code use fastclick that a light impact will generate when lightly touching the scroll, the menu, the buttons, and the links.
> 
> Pretty much everywhere at that point.

Great, sounds good. I look forward to trying it out!

> As for the tactile feedback, Maybe we could add an impact when a user scrolls an interval distance. Therefore, the faster they scroll the faster the interval of the vibration. I don't recommend adding this until after further testing is done on how it feels to use in day-to-day. I figure that adding too many vibrations and impacts when touching and scrolling will cause the user to feel overwhelmed.
> 
> If you are not worried bout this, please ignore my recommendation and I can implement the scroll feature as I just described, based on interval impacts.

Sure, I'm happy to pause that feature until I get my hands on the first iteration. Let's keep it on the table though as I have gotten a lot of interest in high tactile feedback from my beta testers, especially neurodivergent users.

--

I'm noticing some merge conflicts, so I recommend rebasing on `main` and resolving them at your convenience. That will also allow the CI to run. Thanks!

msdewitt:
@raineorshine Looks good to go.

raineorshine:
Thanks! I will let @trevinhofmann do an initial review of code + behavior, then I will review and get a feel for how the haptics feel.

src/actions/deleteThought.ts
```diff
@@ -53,6 +55,9 @@ const deleteThought = (state: State, { local = true, pathParent, thoughtId, orph
 
   // See: Payload.local
   const persist = local || remote
+  if (Capacitor.isNativePlatform()) {
+    Haptics.notification({ type: NotificationType.Warning })
+  }
```
trevinhofmann, line 60: One suggestion might be creating our own utility for haptics which checks for `Capacitor.isNativePlatform()` internally so that we don't need to include so many new `if` blocks throughout the rest of the code.

e.g.:

```suggestion
  haptics.notification({ type: NotificationType.Warning })
```

A reasonable counterargument is that we want to expressly show within the hooks/actions/components that the haptic event is only triggered on native.

I would be okay with leaving this as it is, but slightly prefer creating a utility.

src/util/fastClick.ts
```diff
@@ -27,7 +29,6 @@ const fastClick = isTouch
           const y = e.touches[0].clientY
           touchStart = { x, y }
         }
-
```
trevinhofmann, line 30: This is fine, but just a gentle reminder to avoid including irrelevant changes.

trevinhofmann (changes requested):
Thank you for the PR, @msdewitt!

Overall, the code looks very reasonable to me.

> - Gesture
>   - start: `selectionStart`
>   - swipe: `selectionChanged`
>   - end: `selectionEnd`
>     - *except* Back/Forward: `impact` (light)
>     - *except* Delete: `notification` (warning)
>     - *except* Delete permanently: `notification` (warning)
>   - cancel: `notification` (warning)

⚠️

1. I am not noticing any feedback when starting a gesture (except, generally, when the gesture menu appears).
2. I am noticing the feedback for "selection changed", but only when waiting long enough for the gesture menu to appear. If I quickly make a gesture, it does not provide feedback throughout the gesture.
3. For the "Back" gesture, there is feedback at the end (and only the end).
4. For the "Forward" gesture, there is feedback when the gesture menu appears and when the gesture ends.

Most of these can be grouped as "gesture feedback only enabled when the gesture menu is opened". @raineorshine -- Is that the intended functionality?

> - Long tap
>   - start: `selectionStart`

✅ This appears to be working as expected but is difficult for me to test, because I believe it conflicts with / is redundant with native haptics.

> - Drag-and-drop
>   - start: `selectionStart`
>   - drop target changes: `selectionChanged`
>   - end: `selectionEnd`
>   - cancel: `notification` (warning)

As previously discussed, drag-and-drop functionality is not working on `main`, and partly captured in #2474. When trying to drag a thought, it switches from drag-and-drop to starting a new gesture. When trying to drag a button on the Customize Toolbar page, it just scrolls the toolbar.

I'll skip testing this 👍

> - Hamburger menu
>   - toggle: `impact` (light)

✅ This appears to be working as expected.

> - Link
>   - click: `impact` (light)

✅ This appears to be working as expected.

> - Button
>   - click: `impact` (light)

✅ This appears to be working as expected.

---

Currently having some difficulty getting my Android test device connected, but I'll resolve that soon on my own time and be ready to test again next time with that device.

src/actions/deleteThought.ts
raineorshine: We definitely want to abstract this out since the condition is present with every Haptics call. It should always NOOP on other platforms.

msdewitt:
@raineorshine Alright, it sounds like most of the feedback is good on this first round of testing. Did you also want to get the app in your hands to test and changes? I will go back and make changes to the gesture portion now.

I guess that I can move the Haptics SelectionStart and SelectionEnd portions to the MultiGesture class. Is there a faster place to detect a new selectionChanged for Gestures other than the GestureMenu?

raineorshine:
Thanks! I would like to wait till the gesture haptics timing is fixed, since that is a key part of the experience. 

The `onGesture` callback (or its call site) is where to look to respond to new swipes as soon as they occur.

raineorshine:
Thanks! I'll take a look asap.

raineorshine (changes requested):
I had a chance to try it myself, and it works great! Really exciting to have this feature in **em** 😁

Here are my thoughts on each of the haptics we have now:

- [x] **Gesture/start** ~~`selectionStart`~~ - Remove... feels extraneous
- [ ] **Gesture/onGesture** ~~`selectionChanged`~~ - Let's bump this up to `impact(light)`, as `selectionChanged` doesn't feel strong enough. We may go impact(medium) but let's test impact(light) first.
- [ ] **Gesture/end** ~~`selectionEnd`~~ - Remove... then the number of impacts will exactly match the number of swipes of the gesture
- [ ] **Gesture/cancel** `notification` (warning) - - I'm not feeling this one, but it might be overlapping with the universal touchend. I'll test again after the next round.
- [ ] **Scroll/Continuous** ~~`impact`~~ `selectionChanged` Like old iPod or Nest wheel. - The `selectionChanged` haptic is very light and I think would make great scrolling feedback. Let's move forward with this. You can generate `selectionChanged` haptics proportional to the velocity (i.e. every X pixels scrolled).
- [ ] **Long tap/start** `selectionStart` - This one seems to be missing. It should trigger as soon as the long tap fires, i.e. when the bullet turns the blue highlight. See [useLongPress](https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/hooks/useLongPress.ts).
- [ ] **Drag-and-drop** - HOLD
- [x] **Hamburger menu** - toggle: `impact` (light) ✓ 
- [x] **Link** - click: `impact` (light) ✓ 
- [ ] **Button** - click: `impact` (light) - Active buttons work well. Disabled buttons should not trigger haptics. For example, if the cursor is null then the Bold button is disabled.

Tapping the blank area below the thoughts should **not** result in any haptics. This may be triggered by the fastClick in [Content.tsx#L69](https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/components/Content.tsx#L69). Adding a `preventHaptics` option to `fastClick` might solve this.

msdewitt:
> I had a chance to try it myself, and it works great! Really exciting to have this feature in **em** 😁
> 
> Here are my thoughts on each of the haptics we have now:
> 
> * [ ]  **Gesture/start** ~`selectionStart`~ - Remove... feels extraneous
> * [ ]  **Gesture/onGesture** ~`selectionChanged`~ - Let's bump this up to `impact(light)`, as `selectionChanged` doesn't feel strong enough. We may go impact(medium) but let's test impact(light) first.
> * [ ]  **Gesture/end** ~`selectionEnd`~ - Remove... then the number of impacts will exactly match the number of swipes of the gesture
> * [ ]  **Gesture/cancel** `notification` (warning) - - I'm not feeling this one, but it might be overlapping with the universal touchend. I'll test again after the next round.
> * [ ]  **Scroll/Continuous** ~`impact`~ `selectionChanged` Like old iPod or Nest wheel. - The `selectionChanged` haptic is very light and I think would make great scrolling feedback. Let's move forward with this. You can generate `selectionChanged` haptics proportional to the velocity (i.e. every X pixels scrolled).
> * [ ]  **Long tap/start** `selectionStart` - This one seems to be missing. It should trigger as soon as the long tap fires, i.e. when the bullet turns the blue highlight. See [useLongPress](https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/hooks/useLongPress.ts).
> * [ ]  **Drag-and-drop** - HOLD
> * [x]  **Hamburger menu** - toggle: `impact` (light) ✓
> * [x]  **Link** - click: `impact` (light) ✓
> * [ ]  **Button** - click: `impact` (light) - Active buttons work well. Disabled buttons should not trigger haptics. For example, if the cursor is null then the Bold button is disabled.
> 
> Tapping the blank area below the thoughts should **not** result in any haptics. This may be triggered by the fastClick in [Content.tsx#L69](https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/components/Content.tsx#L69). Adding a `preventHaptics` option to `fastClick` might solve this.

Awesome review. Thank you for this through review. Its very detailed. I can make those changes you requested now.

src/components/MultiGesture.tsx
```diff
@@ -216,6 +230,11 @@ class MultiGesture extends React.Component<MultiGestureProps> {
             y: gestureState.moveY,
           }
           this.props.onEnd?.({ sequence: this.sequence, clientStart: this.clientStart!, clientEnd, e })
+          if (Capacitor.isNativePlatform()) {
+            for (let i = 0; i < this.sequence.length; i++) {
+              Haptics.impact({ style: ImpactStyle.Light })
```
raineorshine, line 235: Just to clarify, the onEnd haptics should be removed completely. The one impact per swipe is already achieved by onGesture above.

msdewitt: Gotcha, I can do that. Thank you

msdewitt:
@raineorshine I've finished adjusting the changes you had in code review for testing tonight. Please take a look at it this morning to see how it feels after the adjustments. 

Specifically, see how the toolbar at the top feels with the 'scroll' feature enabled.

If you can, please give some feedback for any further adjustments.

src/hooks/useLongPress.ts
```diff
@@ -41,6 +42,7 @@ const useLongPress = (
         clientCoords.current = { x: e.touches?.[0]?.clientX, y: e.touches?.[0]?.clientY }
       }
       onTouchStart?.()
+      Haptics.selectionStart()
```
raineorshine, line 45: This should go next to `onLongPressStart`, where the long press actually starts.

`start` only starts the timer.

raineorshine (changes requested):
Thanks! It's looking good. 

Here's are the changes I think should happen.

## Gestures

Gestures are better now. I'm usually getting one haptic per swipe.

I'm still getting two impacts sometimes. Try slowly activating Back (→) and Forward (←) between a, b, c:

```
- a
  - b
    - c
```

## Scroll

This is a good start! I can see a couple things in need of change to get a more satisfying feel.

### 1. Rate

To create the most realistic feedback, the rate of haptic feedback should ideally be proportional to the velocity of the scroll, i.e. the number of haptics should be proportional to the number of pixels scrolled. However, the `scroll` event does not fire every pixel when the user scrolls quickly. This means that currently we're getting the opposite of what we want: the rate of haptic feedback is inversely proportional to the scroll velocity.

So basically you have to measure how many pixels have been scrolled each scroll event and determine whether haptics should fire or not. I would aim for about one haptic every 10 pixels.

We could also try firing multiple haptics when a large scroll is detected in a single event, though that might create a weird delay.

Note: we are limited to 32 Hz (31.25 haptics per second) so we should keep that in mind. Additional haptics above that are dropped, so I don't think we have to do anything special with this rate limiting though unless it affects performance.

### 2. Touchmove

Since haptics parallel touch events to create a tactile experience, it's clear that we don't want them to fire when the user's finger it not on the screen. So just firing haptics on scroll is a problem because of momentum scrolling (when it keeps scrolling after the user has released their finger from the screen).

It's not pretty, but we actually have a global `touching` variable that you should be able to easily check:

https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/globals.ts#L9-L11

## Long press

This one is still missing. It should trigger as soon as the long tap fires, i.e. when the bullet turns the blue highlight.

## Button

I'm still getting a false positive on disabled toolbar buttons:

1. Set the cursor to null.
2. Tap `bold` toolbar button.

msdewitt:
> Thanks! It's looking good.
> 
> Here's are the changes I think should happen.
> 
> ## Gestures
> Gestures are better now. I'm usually getting one haptic per swipe.
> 
> I'm still getting two impacts sometimes. Try slowly activating Back (→) and Forward (←) between a, b, c:
> 
> ```
> - a
>   - b
>     - c
> ```
> 
> ## Scroll
> This is a good start! I can see a couple things in need of change to get a more satisfying feel.
> 
> ### 1. Rate
> To create the most realistic feedback, the rate of haptic feedback should ideally be proportional to the velocity of the scroll, i.e. the number of haptics should be proportional to the number of pixels scrolled. However, the `scroll` event does not fire every pixel when the user scrolls quickly. This means that currently we're getting the opposite of what we want: the rate of haptic feedback is inversely proportional to the scroll velocity.
> 
> So basically you have to measure how many pixels have been scrolled each scroll event (and probably since the start of the scroll) and determine whether haptics should fire or not. I would aim for about one haptic every 10 pixels.
> 
> We could also try firing multiple haptics when a large scroll is detected in a single event, though that might create a weird delay.
> 
> Note: we are limited to 32 Hz (31.25 haptics per second) so we should keep that in mind. Additional haptics above that are dropped, so I don't think we have to do anything special with this rate limiting though unless it affects performance.
> 
> ### 2. Touchmove
> Since haptics parallel touch events to create a tactile experience, it's clear that we don't want them to fire when the user's finger it not on the screen. So just firing haptics on scroll is a problem because of momentum scrolling (when it keeps scrolling after the user has released their finger from the screen).
> 
> It's not pretty, but we actually have a global `touching` variable that you should be able to easily check:
> 
> https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/globals.ts#L9-L11
> 
> ## Long press
> This one is still missing. It should trigger as soon as the long tap fires, i.e. when the bullet turns the blue highlight.
> 
> ## Button
> I'm still getting a false positive on disabled toolbar buttons:
> 
> 1. Set the cursor to null.
> 2. Tap `bold` toolbar button.

So I have some news then regarding haptics and the behavior you want for quick consecutive triggers of haptics, right now it is already firing the haptics response as fast as possible and it might not be built to handle what you want with the fast triggers based on scrolling velocity. Right now, it is set to trigger every 15 pixels in the code. But because each Haptics vibration needs to contact the native device, it runs in an async thread. When we try and trigger them many times consecutively, there is a buffer delay that cannot be overcome when translating to the native code and back along with the asyn behavior of many threads spawning.

@raineorshine If you have ideas to overcome this it would be great.

raineorshine:
> So I have some news then regarding haptics and the behavior you want for quick consecutive triggers of haptics, right now it is already firing the haptics response as fast as possible and it might not be built to handle what you want with the fast triggers based on scrolling velocity.

Yes, that makes sense. I'm not suggesting we try to speed it up. I'm suggesting we slow down the haptics when the user only scrolls a tiny amount.

> Right now, it is set to trigger every 15 pixels in the code.

Can you link to the line of code for this? I can't seem to find it.

Also just to confirm, you have only added scroll haptics to the Toolbar so far, right?

> But because each Haptics vibration needs to contact the native device, it runs in an async thread. When we try and trigger them many times consecutively, there is a buffer delay that cannot be overcome when translating to the native code and back along with the asyn behavior of many threads spawning.

Yes, I understand that is a constraint we have to work with.

Can scrolling the body be detected in the Capacitor plugin? If so, we could possibly control the scroll haptics from the WebView whether there is no latency from the js bridge.

src/components/Toolbar.tsx
```diff
@@ -137,6 +138,7 @@ const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected }) => {
 
       if (scrollDifference >= 5) {
```
msdewitt, line 139: Sorry the difference isn't 15px but 5 pixels right now.

msdewitt:
@raineorshine Yes, its only added to the toolbar right now, just to work out these issues. Adding these changes to the IOS Webview scroll bar will be even more troublesome since it's not in the javascript code.

raineorshine:
> @raineorshine Yes, its only added to the toolbar right now, just to work out these issues.

Great, just wanted to make sure I was looking at the same commit as you.

> @raineorshine Yes, its only added to the toolbar right now, just to work out these issues. Adding these changes to the IOS Webview scroll bar will be even more troublesome since it's not in the javascript code.

I suggest that we consider all our options, estimate the work effort, and then determine what is in scope for the current contract. There is far more harm in dismissing a possible course of action prematurely. I'm sure that between the two of us we can make a good judgment about the best way to solve this. I respect your time and I'm not going to make you do anything you don't want to do.

> Sorry the difference isn't 15px but 5 pixels right now.

Got it, I see. It will only trigger the haptics when a scroll event has exceeded 5px.

Stepping back from the implementation, I want to make sure that you understand and are able to reproduce the behavior I am describing. Try scrolling very slowly, then medium speed, then very quickly. When scrolling very slowly, the haptics fire very frequently. When scrolling at medium speed, the haptics fire less frequently. And when scrolling very quickly, the haptics don't fire very much at all. Due to the latency I understand that there may not be anything we can do about the high rate of scroll. Therefore, let's focus on the slow scroll. At the moment there are too many haptics. If we can bring that rate down, I think it will feel a lot more natural.

While I'm not able to guide the exact implementation, I am happy to further clarify the behavior. I bet with a little experimentation you can get it closer to a more proportional and natural feeling rate of scroll, at least at the slower speeds.

msdewitt:
@raineorshine Gotcha, while I was testing here I was able to see each trigger happen more frequently when scrolling faster, but the device itself might not be able to handle so many vibrations and the native components might prevent it, causing the faster speeds to seem like there are less impacts.
Screenshot 2024-11-28 at 6 27 34 PM

And thank you for respecting my time. We can talk more in Upwork chat about that when needed.

raineorshine:
> @raineorshine Gotcha, while I was testing here I was able to see each trigger happen more frequently when scrolling faster, but the device itself might not be able to handle so many vibrations and the native components might prevent it, causing the faster speeds to seem like there are less impacts.

Yes, that could be. Let's focus on the slower scrolling for now. I don't think the 5px threshold is working the way you are expecting. Probably because `lastScrollLeft` is not what you think it is (and admittedly poorly named). I scroll 1px and I get haptic feedback.

trevinhofmann (changes requested):
> Secondly, I added a small impact light at the start of any and all Gestures to see if the experience improves at all. @trevinhofmann I know you mentioned that you didn't feel anything at the start of most Gestures, so I wanted to see if this makes it feel better.

I see additional changes have been made since, but still wanted to confirm my current experience with gestures on the latest commit (aa23971b72850827b27591acd8c3ae83f72c6000).

The haptics for gestures are no longer dependent on the gesture menu's appearance, and I am usually observing the haptic feedback when starting a gesture now.

---

I have also noticed what @raineorshine pointed out:

> I'm still getting two impacts sometimes. Try slowly activating Back (→) and Forward (←) between a, b, c:

There are usually two rather than one whenever making the gesture slowly.

---

I also noticed that there is feedback for every direction change when gesturing, including when cancelling the gesture. For example, if you just draw multiple circles for the gesture, the haptic feedback will continue (4 per circle) even after recognizing that the gesture is cancelled.

That _may_ be the intended/desired functionality, but #2392 only mentions "cancel: `notification` (warning)", so I would probably expect the haptic notification feedback only once when the gesture is cancelled. @raineorshine may be able to clarify what is expected.

raineorshine:
Good point. Subsequent swipes after a gesture has been cancelled should not trigger haptics.

msdewitt:
Sigh, so annoying. Let me fix these conflicting files in a bit.

raineorshine:
Yes, sorry about that! `shortcuts` is in the process of being renamed to `commands`.

msdewitt:
@trevinhofmann Hey Trevin, Is it possible for you to help me rebase this branch with main? I've been having strange issues on my local git for the past week and have grown very frustrated with making no progress just syncing up the branches. It would really help me out if you could help me with this.

trevinhofmann:
Hi, @msdewitt. I started to rebase on `main`, but there were far too many conflicts related to `shortcuts` being renamed to `commands`. For now, I added a merge commit (6d1ff83eba2c86d4c583231069861402125bb8ff) from `main` to resolve the conflicts. A rebase is preferred over a merge commit, it would have taken an unreasonable amount of time in this case.

For future PRs, I would recommend keeping to rebase instead of merge and doing so relatively often to stay up-to-date with `main`.

For this PR, it might be worth squashing all of the commits into one since there have been so many changes throughout the history of the branch. The current diff is relatively simple compared to the entirety of changes in the 42 commits, so it may be worth cleaning up the commit history by squashing. This will also remove the merge commits and make it easy to rebase this branch going forward if needed.

trevinhofmann:
Hi, @msdewitt! Just wanted to check in and see if there is anything I can help with for this PR. Thank you!

msdewitt:
@trevinhofmann I do need some help with figuring out how to reliably trigger Haptics at a set interval during scrolling. This is the last issue I need to resolve for this pull request, but no matter how I try to word it in the code, it doesn't seem to want to play nice.

I've tried a method which tracks the number of pixels traveled since the 'scroll starting point', but the problem is that this 'scroll starting point' only updates after the scroll action comes to a complete stop. So the OnScroll listener seems like its a problem.

Do you have any ideas?

src/components/Toolbar.tsx
```diff
@@ -132,15 +133,19 @@ const Toolbar: FC<ToolbarProps> = ({ customize, onSelect, selected }) => {
       setRightArrowIsShown(el.offsetWidth + el.scrollLeft < el.scrollWidth - 20)
     }
   }, [])
-
   /** Handles toolbar scroll event. */
   const onScroll = useCallback(
     (e: React.UIEvent<HTMLElement>) => {
+      let lastScrollY = 0
```
trevinhofmann, line 139: One issue here is that `lastScrollY` is always reset to 0 whenever this callback gets called.

Here is one suggested approach that seems trigger the haptic feedback for every ~50 pixels scrolled. You might want to use that as a starting point: https://github.com/cybersemics/em/commit/c0dabee32ff736d7f45b1e0492dfa4e2a7ecb7c6

src/actions/deleteThought.ts
trevinhofmann: Refactored in the new PR here: a8ab9aec4919b3983d394df433b52a57b9e3402d

trevinhofmann:
Due to the large number of commits and several merge conflicts, I have squashed the changes and reopened a new PR at #2827.

src/util/fastClick.ts
```diff
@@ -13,6 +15,9 @@ const fastClick = isTouch
       // triggered on mouseup or touchend
       // cancelled if the user scroll or drags
       tapUp: (e: React.TouchEvent) => void,
+
+      isHaptics: boolean = false,
+
```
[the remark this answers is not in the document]
trevinhofmann: In the new PR, I have refactored the options to named parameters: 1e49e848d92ea8788c3b7641d7040d7a2fdd2dae

I left the first parameter (`tapUp`) as a required first parameter, and changed `isHaptics` (renamed to `enableHaptics`), `tapDown`, `tapCancel`, and `touchMove` to named parameters in an options object.

I believe this is the cleanest API of the two suggestions (not needing to ever pass `undefined` parameters, and not worrying about the parameter ordering of these options), and it required minimal refactoring in other parts of the repository since the vast majority of places using it only pass the first `tapUp` parameter.

PR #2612, opened 2024-11-23
Fix light theme + light theme snapshot tests

yangchristina:


src/colors.config.ts
```diff
@@ -5,7 +5,7 @@ const colors = {
     bgOverlay80: 'rgba(0, 0, 0, 0.8)',
     bgOverlay50: 'rgba(0, 0, 0, 0.5)',
     bgOverlay30: 'rgba(0, 0, 0, 0.3)',
-    black: 'rgba(0, 0, 0, 1)',
+    black: 'rgba(0, 0, 0, 1)', // should be unused now
```
trevinhofmann, line 8: From a quick search, this looks like the only place still using `'black'`:

https://github.com/cybersemics/em/blob/31c4ea737db3b453f38799759a5c26198e573a2f/src/components/ColorPicker.tsx#L95

Should this be updated?

trevinhofmann (changes requested):
Thank you for the PR and thorough tests, @yangchristina! I was a bit concerned that the additional tests might increase the test run time significantly, but the difference seems relatively small and well worth the additional coverage.

The only issue I'm still seeing with the light theme is that the button animations are not visible:

https://github.com/user-attachments/assets/48cfada0-506d-405e-aa76-61db3e165b51

src/colors.config.ts
yangchristina: Yes, sorry, I'll remove that later, I added that afterwards, and forgot to remove the comment

trevinhofmann: Thanks! Are you planning to make that change in this PR or a future one?

yangchristina: Sorry about that, I've just changed it.

src/e2e/puppeteer/__tests__/render-thoughts.ts
```diff
@@ -14,6 +16,20 @@ expect.extend({
...
+  await paste(`
+    - a
+      - m
+    - b
+      - m
+  `)
+
+  await press('ArrowUp')
+
+  return await screenshot()
```
raineorshine, line 30: FYI `return await` can always be simplified to `return`. The `await` is implied by the async function.

src/e2e/puppeteer/__tests__/render-thoughts.ts
```diff
@@ -116,16 +138,14 @@ const testSuite = () => {
     })
 
     it('superscript', async () => {
```
raineorshine, line 140: This may not be relevant after moving the light tests out of `testSuite`, but note that you can take advantage of nested describe blocks for clean organization of test variants:

## Option 1

```ts
	describe('initial load', () => {
	  it('dark', async () => {
	    ...
	  })
	  it('light', async () => {
	    ...
	  })
	})
```

## Option 2

```ts
	describe('dark', () => {
	  it('initial load', async () => {
	    ...
	  })
	})
	
	describe('light', () => {
	  it('initial load', async () => {
	    ...
	  })
	})
```

src/e2e/puppeteer/__tests__/render-thoughts.ts
```diff
@@ -27,6 +43,12 @@ const testSuite = () => {
       const image = await screenshot()
       expect(image).toMatchImageSnapshot()
     })
+
+    it('initial load on light theme', async () => {
```
raineorshine, line 47: I actually vote for taking all the Light tests out of the `testSuite` that gets run at different font sizes. Theme and font size are orthogonal, so we don't need to test different themes on different font sizes. Doing so will just slow down the tests.

What do you think?

src/e2e/puppeteer/__tests__/sidebar.ts
```diff
@@ -0,0 +1,42 @@
...
+const fullyOpenSidebar = async () => {
+  await click('[aria-label=menu]')
+  await new Promise(resolve => setTimeout(resolve, 2000))
+}
+
+describe('sidebar', () => {
+  it('empty sidebar on dark theme', async () => {
+    await fullyOpenSidebar()
+
+    expect(await screenshot()).toMatchImageSnapshot({ customSnapshotIdentifier: 'sidebar-empty' })
```
raineorshine, line 25: In the sidebar tests, could you hide the toolbar buttons like I did in https://github.com/cybersemics/em/blob/36edb98e159d91721b610db1e1d4e512f886d690/src/e2e/puppeteer/__tests__/ColorPicker.ts? That will prevent churn when the toolbar changes. Visual regressions of the toolbar in Light mode will get picked up by render-thoughts initial load.

```ts
await hideVisibility('[data-testid="toolbar-icon"]')
```

raineorshine (changes requested):
Thanks!

src/e2e/puppeteer/__tests__/render-thoughts.ts
yangchristina: I agree, will take it out, thanks!

src/e2e/puppeteer/__tests__/render-thoughts.ts
yangchristina: Good to know, thanks!

src/e2e/puppeteer/__tests__/ColorPicker.ts
```diff
@@ -3,19 +3,9 @@ import sleep from '../../../util/sleep'
 import configureSnapshots from '../configureSnapshots'
 import click from '../helpers/click'
 import clickThought from '../helpers/clickThought'
+import hideVisibility from '../helpers/hideVisibility'
 import paste from '../helpers/paste'
 import screenshot from '../helpers/screenshot'
-import { page } from '../setup'
-
-/** Hides all DOM nodes that match the selector from the DOM by setting visibility: hidden. Do nothing if the selector is empty. */
-const hideVisibility = async (selector: string) => {
```
raineorshine, line 11: Thanks for factoring this out 🙇‍♀️

src/e2e/puppeteer/__tests__/sidebar.ts
```diff
@@ -0,0 +1,42 @@
...
+expect.extend({
+  toMatchImageSnapshot: configureSnapshots({ fileName: path.basename(__filename).replace('.ts', '') }),
+})
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+/** Open sidebar and wait for it to slide all the way open. */
+const fullyOpenSidebar = async () => {
+  await click('[aria-label=menu]')
+  await new Promise(resolve => setTimeout(resolve, 2000))
```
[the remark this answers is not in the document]
yangchristina: @raineorshine I put the MUI default transition values into durations.config, but I can also just use fastDuration (200ms) for both the enter and exit, to standardize everything
```js
  // MUI duration defaults https://mui.com/material-ui/customization/transitions/#durations
  muiEnteringScreen: 225,
  muiLeavingScreen: 195,
```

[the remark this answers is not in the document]
yangchristina: Actually I'll just change it to use fastDuration, let me know if you prefer MUI's defaults.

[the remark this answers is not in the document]
yangchristina: Also, @raineorshine would it be okay if I refactored durations.config to remove the `Duration` suffix for ever token?
Right now in `FadeTransition`, I don't like that `timeout={durations.get(`${duration}Duration`)}` will throw a typeerror if any one of the duration keys do not end in `Duration`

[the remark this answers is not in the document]
raineorshine: Yes, thanks, that seems cleaner!

src/e2e/puppeteer/__tests__/sidebar.ts
```diff
@@ -0,0 +1,46 @@
...
+import setTheme from '../helpers/setTheme'
+import type from '../helpers/type'
+
+expect.extend({
+  toMatchImageSnapshot: configureSnapshots({ fileName: path.basename(__filename).replace('.ts', '') }),
+})
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+/** Open sidebar and wait for it to slide all the way open. */
```
raineorshine, line 16: Update comment

src/e2e/puppeteer/__tests__/sidebar.ts
```diff
@@ -0,0 +1,46 @@
...
+const openSidebar = async () => {
+  await click('[aria-label=menu]')
+  await new Promise(resolve => setTimeout(resolve, 100))
```
raineorshine, line 19: Rather than setting an arbitrary timeout, I think we can wait for an element in the sidebar to appear. That will generally be faster.

```ts
await page.locator('[data-testid="TEST_ID"]').wait()
```

raineorshine (changes requested):
Thanks, looks good!

Regarding the Puppeteer tests, has this been rebased recently on the latest `main`? I fixed some flaky drag-and-drop snapshot tests yesterday that might be affecting the CI result.

The modal and render-thought snapshot tests however are reliable. I will re-run the CI just in case, but any remaining failing tests should be addressed. If there are no perceptible differences, then you are safe to update the snapshot. When it passes in the CI and on your local machine then we'll know it's safe to merge.

src/e2e/puppeteer/__tests__/sidebar.ts
yangchristina: Thanks! Also all the failing tests are light mode related, so I'll just update them.

yangchristina:
@raineorshine is this a mistake? tutorialStepNavigationDuration does not exist as far as I'm aware (in TutorialNavigation.tsx)
```js
transition: `all {durations.tutorialStepNavigationDuration} ease-in-out`
```

raineorshine:
Yes, it must be. Can you use git blame to look back a few commits? If not I can look for it. Maybe just got missed when I was standardizing the durations.

yangchristina:
@raineorshine this token was removed in ea31e30

raineorshine:
Great. I think you can see how to fix it, yes?

yangchristina:
@raineorshine yes, I've added the fix.

I'm getting a flaky test with sidebar-recently-edited-diff
sidebar-recently-edited-diff
sidebar-recently-edited-diff

I'm not sure why the hamburger and TriangleRight are sometimes being removed, since they don't have `data-testid="toolbar-icon"`

raineorshine:
Hmmm that’s strange. And the scroll arrow. It's intermittent? Local machine too, or just in the CI?

raineorshine:
There were two deeper issues that were causing the snapshot failures.

First, there will still some hardcoded durations in the codebase. In particular, the icon color transition was being triggered on theme change. The tests would fail intermittently depending on how many frames were rendered before the snapshot was taken and if they exceeded the pixel diff failure threshold. I converted the hardcoded durations to PandaCSS durations in 746af2cce3 so that they are zeroed out during puppeteer tests.

Second, the hamburger menu and toolbar arrow were sometimes hidden in the Recently Edited Thoughts snapshot because of a feature called Distraction Free Typing which hides the toolbar on desktop after the user starts typing. While the `distractionFreeTyping` fade duration was getting zeroed out in tests, there was a hardcoded 100ms throttle duration that was not. I moved it to the durations config in 7a1baf8b4b so that it now gets zeroed out so the snapshots will consistently reflect the change.

raineorshine (approved):

yangchristina:
@raineorshine Thank you for explaining this! I didn’t consider how the transition played out in tests. I’ve learned something new and appreciate your clarification!

raineorshine:
It was a learning experience for me, too 🙂

Issue #2605, opened 2024-11-21
[Android] Gesture incorrectly activated in scroll zone


raineorshine:
Android Chrome only.

## Steps to Reproduce

Scroll by dragging in the scroll zone.

## Current Behavior

Sometimes the TraceGesture is visible and a gesture is activated.

https://github.com/user-attachments/assets/1e5e6e7d-3319-4782-9697-a404625e835e

## Expected Behavior

Gestures should not be able to be activated when a touch starts in the scroll zone.

snqb:
Not the gesture issue itself, but scroll being choppy and slow is not an android exclusive, it looks like.
From a very preliminary research I can conclude that it's the fact that we virtualize thoughts, that leads to choppiness.

Attaching a chromium browser test on Windows.
I guess, in the end we want to have some better kind of logic behind virtualization.
If this thing is choppy on a powerful PC, then it's gonna be(and is) atrocious on less powerful Android devices.

https://github.com/user-attachments/assets/bacac01d-8206-4f8a-86d0-203a7a7245ff

snqb:
And concerning the gesture: I think I'll just tweak the `PanResponder` and it should be gone.
It's just wrongly capturing scroll motions, but I just haven't gotten to that yet.

raineorshine:
> Not the gesture issue itself, but scroll being choppy and slow is not an android exclusive, it looks like. From a very preliminary research I can conclude that it's the fact that we virtualize thoughts, that leads to choppiness.
> 
> Attaching a chromium browser test on Windows. I guess, in the end we want to have some better kind of logic behind virtualization. If this thing is choppy on a powerful PC, then it's gonna be(and is) atrocious on less powerful Android devices.

Thanks, that's good to know. We have an existing issue to track this problem: https://github.com/cybersemics/em/issues/2399

> And concerning the gesture: I think I'll just tweak the `PanResponder` and it should be gone.
> It's just wrongly capturing scroll motions, but I just haven't gotten to that yet.

Okay, sounds good!

snqb:
A bit more of updates.
After more digging, I realized, that it's mostly the `TraceGesture` that has no notion scroll vs gesture. So a lot of times it could paint a gesture, while we're scrolling, whilst not triggering testures.

I think it's wrong, and we should only show the trace on definite gestures.
Preliminarily turning on `this.scrolling` on responder events and rendering `TraceGesture` conditionally helps, but is kinda ugly.

[...]

PR #2617, opened 2024-11-23
Check for gesture zone in `TraceGesture` too

snqb:
Fixes https://github.com/cybersemics/em/issues/2605

Turns out, the problem is not Android-only, it's also reproducible for iOS, see the video attached.
(android video is in the original issue)

https://github.com/user-attachments/assets/ba06ed28-19cb-49a5-960b-5ee28911aab2

Thing is, we have the notion of scroll vs gesture zone in the `MultiGesture`, while we don't track it in `TraceGesture`, thus leading to half-painted canceled traces. Surely, it doesn't affect functionality, but is kinda misleading, since we have this strong separation of scroll vs gesture zones.
This PR moves out the gesture zone logic to a util file, and reuses it in both `MultiGesture` and `TraceGesture` to share the logic.

### DIFF
--- a/src/components/MultiGesture.tsx
+++ b/src/components/MultiGesture.tsx
@@ -4,7 +4,7 @@ import Direction from '../@types/Direction'
 import GesturePath from '../@types/GesturePath'
 import { noop } from '../constants'
 import gestureStore from '../stores/gesture'
-import viewportStore from '../stores/viewport'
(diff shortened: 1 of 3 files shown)


raineorshine (approved):
Great, this makes a lot of sense now that we see it is not Android-specific. 

Nice troubleshooting and factoring out the shared functionality.

Thank you!

Issue #2604, opened 2024-11-21
[Mobile] App crashes with "problem repeatedly occurred"


raineorshine:
- Bad on current staging2
- Freezes before scroll zones appears, then flashes and crashes
- Unable to reproduce on localhost or servebuild

![image](https://github.com/user-attachments/assets/92638b7b-dd2b-4428-90db-1fc00b8f0028)

raineorshine:
Seems to be a very long paint event:

<img width="605" alt="Safari 2024-11-24 17 38 12" src="https://github.com/user-attachments/assets/5539b7c0-7b8d-4ddd-9382-c17096e9130f">

PR #2621, opened 2024-11-23
repeat the scroll stars bg with smaller size

snqb:
In progress, not ready for merge.

Had to crop the image to have smooth transitions between repeats + get rid off the too large size + repeat the pattern.

| Before | After |
| - | - |
| image | image |


public/img/scroll-zone/stars_cropped.png
raineorshine, line 1: Please replace the existing image rather than adding a new one.

src/components/ScrollZone.tsx
```diff
@@ -30,7 +28,8 @@ const ScrollZone = ({ leftHanded }: { leftHanded?: boolean } = {}) => {
       })}
       style={{
         transform: `translateY(calc(-${scrollTop / 4 + 400}px))`,
-        width: scrollZoneWidth,
+        // as per viewportStore.scrollZoneWidth = 0.39, but we just go with CSS-based calc
+        width: '39vmin',
```
raineorshine, line 32: I don't recommend decoupling the width from `scrollZoneWidth`. That creates future risk that the values will diverge without us knowing. We want the code to enforce the tight coupling.

raineorshine (changes requested):

src/components/ScrollZone.tsx
snqb: yep, that's why i didn't really wanna do this :D
count this a question

snqb:
but overall, does this look like what we wanted?

public/img/scroll-zone/stars_cropped.png
snqb: I tried going .jpg/.png at 400-500kb to .webp and got to 44kb without much loss in quality
.webp is supported in 99% of mobile browsers which we target anyways, and we don't use much images to not too care about their footrprint

snqb:
and besides the pure static look I'm fixing, there is also an aspect of the parallax effect being kind of laggy
should I try to address this here? (i'd rather not)

src/components/ScrollZone.tsx
```diff
@@ -7,17 +7,16 @@ import viewportStore from '../stores/viewport'
 
 /** An overlay for the scroll zone that blocks pointer events. */
 const ScrollZone = ({ leftHanded }: { leftHanded?: boolean } = {}) => {
-  const hideScrollZone = useSelector(state => state.showModal || getUserSetting(state, Settings.hideScrollZone))
   const scrollTop = useScrollTop()
   const scrollZoneWidth = viewportStore.useSelector(state => state.scrollZoneWidth)
+  const hideScrollZone = useSelector(state => state.showModal || getUserSetting(state, Settings.hideScrollZone))
```
snqb, line 12: moved to be near "if hideScrollZone"

raineorshine:
Overall it looks pretty good. 

There appears to be a black gap between each repeat that we should probably fix so the tiling is less visible.

raineorshine (approved):
Thank you!

Turns out this fixes #2604 as well.

PR #2622, opened 2024-11-24
Fix pull queue tests 

### LINKED ISSUE #2139: Fix pullQueue tests
The pullQueue tests are currently being skipped due to a data provider issue.

https://github.com/cybersemics/em/blob/f68689f6471af07ba87db97cdd9b68bf42cdc910/src/redux-middleware/__tests__/pullQueue.ts

mstrofbass:
First problem here is that we can't `await getContext` when using fake timers and replication because `getContext` calls `getLexemeById` which then calls `await` itself, causing the code to sit there and wait because nothing is forwarding the timers. Thus, we must grab the `getContext` promise, run the timers forward, and _then_ `await` the getContext promise.

Second problem appears that `thoughtspace.ts` has at least one instance where it awaits an operation and then creates async operations afterwards. This means that `runOnlyPendingTimers` won't work since that simply forwards the pending timer and won't forward past the subsequent async operation(s). 

Closes #2139

### DIFF
--- a/src/redux-middleware/__tests__/pullQueue.ts
+++ b/src/redux-middleware/__tests__/pullQueue.ts
@@ -283,8 +283,6 @@ it('edit thought with buffered descendants', async () => {
 
   await refreshTestApp()
 
-  // await act(vi.runOnlyPendingTimersAsync)
-
   // // edit thought with buffered descendants
   await dispatch(editThought(['a'], 'k'))
 
--- a/src/redux-middleware/__tests__/pullQueue.ts
+++ b/src/redux-middleware/__tests__/pullQueue.ts
@@ -35,7 +35,7 @@ const matchContextsChildren = async (provider: DataProvider, context: Context, c
 }
 
 /**
- * getContext but automatically runs the timers forward
+ * Calls getContext but automatically runs the timers forward.
  */
 const getContextAsync = async (provider: DataProvider, context: Context) => {
   const ctxPromise = getContext(provider, context)
--- a/src/redux-middleware/__tests__/pullQueue.ts
+++ b/src/redux-middleware/__tests__/pullQueue.ts
@@ -41,7 +41,7 @@ const getContextAsync = async (provider: DataProvider, context: Context) => {
   const ctxPromise = getContext(provider, context)
   await vi.runAllTimersAsync()
 
-  return await ctxPromise
+  return ctxPromise
 }
 
 beforeEach(createTestApp)
@@ -51,7 +51,6 @@ it('disable isLoading after initialize', async () => {
   expect(store.getState().isLoading).toBe(false)
 })
 
-// y-indexeddb breaks tests
 it('load thought', async () => {
   // create a thought, which will get persisted to local db
   await dispatch(newThought({ value: 'a' }))
@@ -106,7 +105,6 @@ it('do not repopulate deleted thought', async () => {
   expect(parentEntryChild).toBe(null)
 })
 
-// y-indexeddb breaks tests
 it('load buffered thoughts', async () => {
   await dispatch(
     importText({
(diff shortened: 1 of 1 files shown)


raineorshine:
Looks good, thanks!

> First problem here is that we can't `await getContext` when using fake timers and replication because `getContext` calls `getLexemeById` which then calls `await` itself, causing the code to sit there and wait because nothing is forwarding the timers. Thus, we must grab the `getContext` promise, run the timers forward, and _then_ `await` the getContext promise.

Is there any way we can automate this? It seems like this is a frequent gotcha, and there are a lot of discussions online of people running into similar problems of multiple tranches of timers not being run to completion.

src/redux-middleware/__tests__/pullQueue.ts
```diff
@@ -34,6 +34,16 @@ const matchContextsChildren = async (provider: DataProvider, context: Context, c
...
 }
 
+/**
+ * Calls getContext but automatically runs the timers forward.
+ */
+const getContextAsync = async (provider: DataProvider, context: Context) => {
+  const ctxPromise = getContext(provider, context)
+  await vi.runAllTimersAsync()
+
+  return await ctxPromise
```
raineorshine, line 44: `return await` is redundant. You can always just return the promise.

src/redux-middleware/__tests__/pullQueue.ts
```diff
@@ -271,21 +281,20 @@ it.skip('edit thought with buffered descendants', async () => {
   await matchContextsChildren(db, ['a', 'b', 'c', 'd'], [{ value: 'e' }])
   await matchContextsChildren(db, ['a', 'b', 'c', 'd', 'e'], [])
 
-  await act(vi.runOnlyPendingTimersAsync)
-
   await refreshTestApp()
 
-  // edit thought with buffered descendants
+  // // edit thought with buffered descendants
```
raineorshine, line 286: Extra comment

raineorshine (changes requested):

src/redux-middleware/__tests__/pullQueue.ts
```diff
@@ -139,7 +149,7 @@ it('load buffered thoughts', async () => {
 })
 
 // y-indexeddb breaks tests
```
raineorshine, line 151: Old comment no longer needed.

src/redux-middleware/__tests__/pullQueue.ts
```diff
@@ -170,11 +180,11 @@ it.skip('delete thought with buffered descendants', async () => {
...
-  expect(await getContext(db, ['a', 'b', 'c', 'd'])).toBeFalsy()
-  expect(await getContext(db, ['a', 'b', 'c', 'd', 'e'])).toBeFalsy()
+  expect(await getContextAsync(db, ['a'])).toBeFalsy()
+  expect(await getContextAsync(db, ['a', 'b'])).toBeFalsy()
+  expect(await getContextAsync(db, ['a', 'b', 'c'])).toBeFalsy()
+  expect(await getContextAsync(db, ['a', 'b', 'c', 'd'])).toBeFalsy()
+  expect(await getContextAsync(db, ['a', 'b', 'c', 'd', 'e'])).toBeFalsy()
 })
 
 // y-indexeddb breaks tests
```
raineorshine, line 190: Old comment no longer needed.

src/redux-middleware/__tests__/pullQueue.ts
```diff
@@ -245,7 +255,7 @@ it('move thought with buffered descendants', async () => {
...
```
raineorshine, line 257: Old comment no longer needed.

raineorshine (changes requested):

raineorshine (approved):

Issue #2583, opened 2024-11-12
Uncaught Exception: ReferenceError: document is not defined


zukilover:
Here's a log from vitest:

```
Vitest caught 2 unhandled errors during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.

⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯
ReferenceError: document is not defined
❯ Timeout._onTimeout src/redux-middleware/scrollCursorIntoView.ts💯30
❯ listOnTimeout node:internal/timers:581:17
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
❯ processTimers node:internal/timers:519:7
```

raineorshine:
https://github.com/cybersemics/em/actions/runs/11801187295/job/32876029673?pr=2582

mstrofbass:
Taking a first pass at debugging this and this might be causing the problem:

https://github.com/cybersemics/em/blob/67457b39353e4a0f3540b00cbefdb38bd4026dfe/src/redux-middleware/scrollCursorIntoView.ts#L108-L115

My guess is that some state change during the tests is triggering the setTimeout and the delayed function isn't being called prior to the test completing. 

I think that part of the difficulty with this in the tests is that all of the store operations _look_ synchronous (because they don't return promises) but in this case there may be an asynchronous call generated anyways (via setTimeout), so the test writer doesn't know that they should either use fake timers or clear timeouts/intervals. 

I might be able to recreate this by increasing the setTimeout delay a lot but since it (1) is not cropping up every time and thus (2) is unlikely to hide a test failure (i.e., be a false positive), I'll probably have to look into this on Friday.

One cheap way to fix this and also decrease the chances it will crop back up may be to set the duration used by the setTimeout to zero (like we do in the e2e tests).

raineorshine:
Thanks, that's informative. Dealing with asynchronous timing seems to be a real challenge in tests.

> One cheap way to fix this and also decrease the chances it will crop back up may be to set the duration used by the setTimeout to zero (like we do in the e2e tests).

Yes, I think the animation durations should definitely be zero'd out in the RTL tests just like the E2E tests.

--

There are a couple troubleshooting questions that come to mind:

1. Since this is a recent regression, which commit introduced it?
[...]

PR #2623, opened 2024-11-25
Reference error document undefined

mstrofbass:
This is a possible, temporary solution for the defect discussed in #2583.

Simply checking if `document` exists should clear up the error.


raineorshine (changes requested):
Yes, I suppose that hiding the issue is better than having to revert the previous test changes that introduces the async issue.

Could you add a `console.warn` when `document` is undefined? That will avoid entirely brushing the problem under the rug and allow us to notice if it is still an issue in the future.

PR #2624, opened 2024-11-25
Use fake timers for `jump.ts` tests

mstrofbass:


raineorshine:
I'm going to get this merged so that the issue is ameliorated on `main`, but happy to have more discussion as needed. Thanks!

PR #2625, opened 2024-11-25
Freeze VirtualThought width during indent shift

### LINKED ISSUE #4: [iOS] Caret drifts during navigation
When navigating, the content is smoothly translated, but the caret drifts.

## Current Behavior

On iOS, the caret drifts out of place during the animation. 

Video 1:
https://github.com/cybersemics/em/assets/750276/931e766d-291c-495b-83b7-bebde4625dcb

Video 2:
https://youtu.be/Poo61duk41w

## Expected Behavior

The browser cursor should stay in place (not move relative to its contenteditable).

If this is a bug in Safari and there is no fix, see if you can hide the caret during the transition (without removing the selection, closing the keyboard, or preventing the user from typing).

ethan-james:
Fixes #4

Animating `transform` doesn't update the caret position correctly in iOS Safari. Animating `margin` as well will pull the caret along in the correct direction, fixing the caret drift. It also causes the layout to update continuously during the animation, so it is beneficial to calculate the width (adding or subtracting new margins) that the thought will eventually occupy, and then apply that static width to the `VirtualThought` for the duration of the transition.

### DIFF
--- a/src/components/VirtualThought.tsx
+++ b/src/components/VirtualThought.tsx
@@ -7,14 +7,17 @@ import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import State from '../@types/State'
 import ThoughtId from '../@types/ThoughtId'
+import { isTouch } from '../browser'
 import useChangeRef from '../hooks/useChangeRef'
 import useDelayedAutofocus from '../hooks/useDelayedAutofocus'
+import usePrevious from '../hooks/usePrevious'
 import useSelectorEffect from '../hooks/useSelectorEffect'
 import { hasChildren } from '../selectors/getChildren'
 import getThoughtById from '../selectors/getThoughtById'
 import isContextViewActive from '../selectors/isContextViewActive'
 import editingValueStore from '../stores/editingValue'
 import viewportStore from '../stores/viewport'
+import durations from '../util/durations'
 import equalPath from '../util/equalPath'
 import head from '../util/head'
 import noteValue from '../util/noteValue'
@@ -63,6 +66,7 @@ const VirtualThought = ({
   isLastVisible,
   autofocus,
   marginRight,
+  indent,
 }: {
   // contextChain is needed to uniquely identify thoughts across context views
   debugIndex?: number
@@ -86,6 +90,7 @@ const VirtualThought = ({
   isLastVisible?: boolean
   autofocus: Autofocus
   marginRight: number
+  indent: number
 }) => {
   // TODO: Why re-render the thought when its height changes? This information should be passively passed up to LayoutTree.
   const [height, setHeight] = useState<number | null>(singleLineHeight)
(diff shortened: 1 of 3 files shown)


ethan-james:
Getting the width and then the virtual height to update properly has been a bit of a journey, and it's not quite perfect yet because the height still overflows sometimes on iOS Safari. I'll take another crack at making it perfect, probably on Tuesday.

src/components/VirtualThought.tsx
```diff
@@ -165,7 +182,9 @@ const VirtualThought = ({
 
   // Recalculate height when anything changes that could indirectly affect the height of the thought. (Height observers are slow.)
   // Autofocus changes when the cursor changes depth or moves between a leaf and non-leaf. This changes the left margin and can cause thoughts to wrap or unwrap.
-  useLayoutEffect(updateSize, [
+  useLayoutEffect(() => {
+    window.requestAnimationFrame(updateSize)
+  }, [
```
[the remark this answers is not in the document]
ethan-james: Right, that should have been removed after I determined that it didn't do anything. Sorry about that, I was in a little too much of a hurry to get things committed.

ethan-james:
> Yes, this is very tricky! Changing the layout calculations, render cycle timing, and effect triggers has a significant chance of introducing regressions.
> 
> I am admittedly nervous about going down this path. Even if it does fix the caret drift, I can already see that the additional renders and effect triggers are going to affect performance. I'm starting to think that simply hiding the caret during animation may be a better choice overall. After all, the caret is mainly useful when typing or at rest, and state.cursor will still be visible when navigating.

Yeah, I was excited to get it as close to working as I did, but it has already been through a few iterations to try and squash fiddly little issues and there are no doubt scenarios that I didn't know to test. Once I've researched the other comments, maybe I'll recommend hiding the caret as well.

src/hooks/usePrevious.ts
```diff
@@ -1,14 +1,18 @@
...
+import { useEffect, useState } from 'react'
 
 /** Retains & returns the previous value passed in each time the hook is invoked. */
-function usePrevious<T>(value: T): T | undefined {
-  const ref = useRef<T>()
+function usePrevious<T>(value: T): T {
+  const [current, setCurrent] = useState(() => value)
+  const [previous, setPrevious] = useState(() => value)
 
   useEffect(() => {
```
ethan-james, line 8: @raineorshine One thing that I didn't expected to see: this hook gets unmounted (a returned callback is called) at various points during the render cycle. I believe that caused it to run again even if `value` hadn't changed.

src/hooks/usePrevious.ts
```diff
@@ -1,14 +1,18 @@
-import { useEffect, useRef } from 'react'
+import { useEffect, useState } from 'react'
 
 /** Retains & returns the previous value passed in each time the hook is invoked. */
-function usePrevious<T>(value: T): T | undefined {
```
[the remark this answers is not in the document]
ethan-james: Good point.

src/hooks/usePrevious.ts
ethan-james: It doesn't seem to be the entire component, just the effect inside the hook. Maybe I was misunderstanding something.

raineorshine: Yeah, I'm not sure.

ethan-james:
Since we're not going ahead with this solution, I'll open a new pull request rather than adding more commits that revert this work.

raineorshine:
Sure, sounds good!

You are also welcome to force push the new commits, even if everything up to this point is discarded. Sometimes it's nice to keep all the history in one PR, and I can mark comments as `outdated` to reduce clutter.

Issue #2549, opened 2024-11-05
Apply bullet color automatically instead of with metaprogramming attribute


raineorshine:
## Steps to Reproduce

```
- Hello
```

Set the text color to blue.

## Current Behavior

Bullet color is set to blue with the `=bullet` metaprogramming attribute.

```
- <font color="#00c7e6">Hello</font>
  - =bullet
    - =style
      - color
        - blue
```

## Expected Behavior

Bullet color should automatically be set to blue when the entire thought is blue without adding a metaprogramming attribute.

```
- <font color="#00c7e6">Hello</font>
```

PR #2627, opened 2024-11-25
Apply bullet color automatically instead of metaprogramming.

RED-ROSE515:
Issue : Apply bullet color automatically instead of with metaprogramming attribute https://github.com/cybersemics/em/issues/2549

In this PR, I have removed the using the metaprogamming attributes when setting text color with the color picker.
I changed the fix function in Bullet.tsx.
I checked thought value and get the current foreColor and backColor from the thought value.
And set the bullet color with that color.

### DIFF
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -443,6 +443,14 @@ const Bullet = ({
 
   /** Check if the entire value is wrapped in a single font or span tag. */
   const isWrappedInSingleTag = (str: string): boolean => {
+    // Check if there's any text before the first tag or after the last tag
+    const firstTagIndex = str.indexOf('<')
+    const lastTagIndex = str.lastIndexOf('>')
+
+    if (firstTagIndex > 0 || lastTagIndex < str.length - 1) {
+      return false
+    }
+
     const tagStack: string[] = []
     let currentPos = 0
 
@@ -481,7 +489,7 @@ const Bullet = ({
     }
 
     // Valid if exactly one tag remains in stack (the outer tag)
-    return tagStack.length === 1 && tagStack[0] === 'font'
+    return tagStack.length === 1 && (tagStack[0] === 'font' || tagStack[0] === 'span')
   }
 
   const fill = useSelector(state => {
@@ -489,7 +497,6 @@ const Bullet = ({
     if (!thought) return undefined
 
     const value = thought.value
-
     // Check if the entire value is wrapped in a single font or span tag
     const isWrappedInTag = isWrappedInSingleTag(value)
     if (isWrappedInTag) {
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -16,7 +16,6 @@ import attributeEquals from '../selectors/attributeEquals'
 import findDescendant from '../selectors/findDescendant'
 import { getAllChildrenAsThoughts, getChildren } from '../selectors/getChildren'
 import getLexeme from '../selectors/getLexeme'
-import getStyle from '../selectors/getStyle'
 import getThoughtById from '../selectors/getThoughtById'
 import isContextViewActive from '../selectors/isContextViewActive'
 import isMulticursorPath from '../selectors/isMulticursorPath'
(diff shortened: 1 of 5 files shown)


src/e2e/puppeteer/helpers/setSelection.ts
```diff
@@ -0,0 +1,22 @@
+import { page } from '../setup'
+
+/** Sets the selection of the text in the editable. */
+const setSelection = async (start: number, end: number) => {
+  await page.evaluate(
+    (start: number, end: number) => {
+      const selection = window.getSelection()
+      const range = document.createRange()
+      const textNode = document.querySelector('[contenteditable="true"]')?.firstChild
```
raineorshine, line 9: This will only set the selection on the cursor thought if it is the first thought.

To make this more flexible, use `document.querySelector('[data-editing=true] [data-editable]')`.

src/e2e/puppeteer/helpers/setSelection.ts
```diff
@@ -0,0 +1,22 @@
...
+      if (textNode) {
```
raineorshine, line 10: If there is no `textNode`, this function will return silently. This is a bad practice in tests. In tests, we want to fail hard when values are not what they are expected to be. Therefore, you should throw an exception if there is no text node.

src/components/Bullet.tsx
```diff
@@ -442,11 +441,78 @@ const Bullet = ({
...
-    return styles?.color
+    const thought = getThoughtById(state, thoughtId)
+    if (!thought) return undefined
+
+    const value = thought.value
+    // Check if the entire value is wrapped in a single font or span tag
+    const isWrappedInTag = isWrappedInSingleTag(value)
+    if (isWrappedInTag) {
+      // Check for background-color in style attribute
+      const bgColorMatch = value.match(/style="[^"]*background-color:\s*([^;"'>]+)[^"]*"/i)
```
raineorshine, line 504: You can instantiate static regexes at the top of the module for efficiency.

src/components/Bullet.tsx
```diff
@@ -442,11 +441,78 @@ const Bullet = ({
...
+      if (bgColorMatch) return bgColorMatch[1]
+
+      const colorMatch = value.match(/style="[^"]*color:\s*([^;"'>]+)[^"]*"/i)
+      if (colorMatch) return colorMatch[1]
```
raineorshine, line 508: `bgColorMatch` and `colorMatch` can be combined:

```ts
value.match(/style="[^"]*(background-)?color:\s*([^;"'>]+)[^"]*"/i)
```

src/components/Bullet.tsx
```diff
@@ -442,11 +441,78 @@ const Bullet = ({
     return children.length < Object.keys(thought.childrenMap).length
   })
 
-  // fill =bullet/=style override
+  /** Check if the entire value is wrapped in a single font or span tag. */
+  const isWrappedInSingleTag = (str: string): boolean => {
```
raineorshine, line 445: This seems so provide similar functionality as https://github.com/cybersemics/em/blob/main/src/util/getCommandState.ts. Is there a reason that can't be used?

raineorshine (changes requested):
Thanks! I can confirm that the behavior is as expected, and both of the cases described in https://github.com/cybersemics/em/pull/2606#pullrequestreview-2456371067 have been fixed.

src/components/Bullet.tsx
RED-ROSE515: I got it.

src/components/Bullet.tsx
RED-ROSE515: Literally, `isWrappedInSingleTag` checks if the thought value is wrapped in a sing tag.
Like this : <font><span>He</span><font>ll</font>o</span>   : true
               <font>Hel</font><span>lo</span>  : false
And `getCommandState` function determines which commands (bold, italic, underline ....) applied to the thought value.
So, I can't use `getCommandState` in `Bullet.tsx`

raineorshine: Are you saying that `getCommandState` also returns the color of substrings, not just the whole string? I think we're saying the same thing, just in different ways, but correct me if I'm wrong.

RED-ROSE515: Yes, it also returns the color of substrings.
So, I can't use `getCommandState` in Bullet.

raineorshine (approved):
Thank you!

Issue #2431, opened 2024-09-29
Error style regression


raineorshine:
Regression: 3dcf1dce18b46f30431b7b9d813ee814bb1f1ce2

## Steps to Reproduce

```js
em.store.dispatch({ type: 'error', value: 'test' })
```

## Current Behavior

<img width="733" alt="error-regression" src="https://github.com/user-attachments/assets/cdccb276-8a56-49cc-bbf4-a3d7008a4fc1">

## Expected Behavior

<img width="734" alt="error-expected" src="https://github.com/user-attachments/assets/cf0d7ddc-7d51-4791-abec-0787e54599ae">

raineorshine:
Putting this on hold for now since the error component will likely change with the redesign.

PR #2628, opened 2024-11-25
fix UI issue on error popup

deepwork0714:
Fix UI issue for https://github.com/cybersemics/em/issues/2431

Issue happened due to the padding
so I set position value instead of padding so whatever error content is long or not, close icon is always at the top-right corner of pop up.


raineorshine (changes requested):
Hi, thanks for your contribution.

There are two issues with the proposed solution.

First, padding is necessary so the tap area of the close button is enlarged. If you just reposition it, then the tap area will be diminished and harder to hit on mobile.

Second, the existing implementation adds padding proportional to the font size. Your solution adds fixed padding in pixels which will not look right across font sizes. This is why the [puppeteer snapshot tests](https://github.com/cybersemics/em/wiki/Testing#5-visual-snapshot-tests) are failing.

Note: The unit tests are failing on `main`, but should be resolved soon. You can safely ignore `document is not defined` errors in the unit test till #2624  is merged.

Let me know if I can assist further. Thanks!

Issue #2619, opened 2024-11-23
[Mobile] The Stars background is crudely attached.


snqb:
Is this the correct look for the stars on the background?
It looks cut off: on Android, on iOS and even on desktop browser when in smaller size mode.

## Steps to reproduce
Just add like 50-100 thoughts and look at the Gesture Zone.
![{6571B0F2-80CE-47D1-A842-225676D6EDFD}](https://github.com/user-attachments/assets/499de666-1870-4100-86a2-2a94c57fc073)

## Desired behavior
Feels like this should be a long and continous starry canvas, without being cut off like this.

raineorshine:
Thanks, you're right. It needs to be tiled.

snqb:
![image](https://github.com/user-attachments/assets/9d39d908-3394-4cba-85bc-63919cf0a2fe)
does this look alright?

raineorshine:
Looks good from here. I'll need to use the live version and scroll a bit to know for sure.

snqb:
https://github.com/cybersemics/em/pull/2621 if you can try it here

Issue #2651, opened 2024-11-25
[Mobile] ScrollZone parallax does not take into account autofocus cropping


raineorshine:
## Steps to Reproduce

``` 
- A
- B
- C
- D
- E
- F
- G
- H
- I
- J
- K
- L
- M
- N
- O
- P
- Q
- R
- S
- T
- U
- V
- X
- Y
- Z
  - 1
    - 2
``` 

Move the cursor between `Z` and `1`. 

## Current Behavior

The scroll zone background jumps. 

https://github.com/user-attachments/assets/64cadb28-5a29-4adb-9928-920e7d11c952

## Expected Behavior

The scroll zone background position should not move. It should not be affected by the autofocus cropping.

snqb:
The only thing that came up to my mind that is not gonna be too complex is just not moving it for smaller jumps

Since the `ScrollZone` just listens for `y`, and then moves the bg by `y / 4 + 400` pixels, it's hard not to trigger the move.
A sophisticated and complex solution could mean trying to differentiate between autofocus and scroll, but do we really want that?
A simpler one could be just comparing latest Y positions and try to judge whether that's a focus thing or not.

snqb:
@raineorshine on a side-note, I wasn't really planning to merge the bg thing, as I was looking for better backgrounds(ours was crudely cut imo)

I have this on my local branch right now, isn't it a bit prettier?
https://github.com/user-attachments/assets/dcc8e1a0-bff5-4c8f-8239-615fe75d595b


concerning the autofocus scroll, I'm looking into it

snqb:
> No, there's no need to guess or approximate here. We already have the exact scroll amount that is jumped for autofocus, so we just need to adjust the parallax to use the autofocus-adjusted y value instead of the actual scrollY.

I think I've found a solution that allows us not to do calculations https://github.com/cybersemics/em/pull/2652

snqb:
so, it should not move when we have autofocus crop(when we go from Z to 1), a level deeper
but should when we go J -> K or A -> B or D -> C by tapping the cursor

right?

raineorshine:
I'm sorry, I don't understand. Can you clarify the question?

snqb:
Let me rephrase it, yup.
So we don't want the background to move on "autofocus crop". It means Z -> 1 should not trigger any bg move.

But should we trigger a bg move if I switch from A to J on the other side of the screen?
Illustrating this:

[...]

PR #2652, opened 2024-11-26
Better parallax

snqb:
Expands on https://github.com/cybersemics/em/issues/2619

- uses other background(8kb), which also has starry background, taken from https://www.transparenttextures.com
- adds transition for less intrusive, smoother bg scroll

feel free to tweak things to your liking, but it came out pretty neat

there is one visible side-effect of using "touchmove" -- you can encounter it after scrolling down, then coming back back to the top of the list. But to my liking, it creates a rubberband-like experience, and basically is feature, not a bug 🤓 
feel free to criticize this, i surely can find a solution of this, but I like it :D
https://github.com/user-attachments/assets/4ae390f3-f012-45bc-9b45-6ff951895b8e


raineorshine:
Hi, thanks for this!

So, I'm going to have to ask you to break this down into one PR per issue. It's too hard to keep track of multiple threads of conversation related to multiple issues in a single PR.

> * subscribes to "touchmove" rather than "scroll" to calculate `y` position -- thus autofocus does not trigger "parallax" - [[Mobile] ScrollZone parallax does not take into account autofocus cropping #2651 (comment)](https://github.com/cybersemics/em/issues/2651#issuecomment-2501163854)

That would fail to trigger parallax when a programmatic scroll is triggered (e.g. when the cursor is automatically scrolled into view).

The proper solution will make the parallax based on the autofocus-adjusted scroll position rather than the exact scroll position.

raineorshine (approved):
Looks good! I like the new image.

After playing around with the transition, I decided I didn't like the delay. It feels better to me if the ScrollZone pans immediately, proportional to the number of pixels scrolled. If it feels choppy, it's probably because the frame rate is dropping.

I'm open to further discussion if it continues to be a problem.

Issue #2618, opened 2024-11-23
Scrolling works in gesture zone on Chromium/Android.


snqb:
While investigating [wrong gesture detection](https://github.com/cybersemics/em/issues/2605) on the scroll zone, I've also encountered a Chromium issue that allows for scrolling on the Gesture Zone.

## Steps to Reproduce
Scroll by dragging in the Gesture Zone on Chromium either on Android or Windows(if you have a touch screen like I do)
 
## Current Behavior
It allows scroll on the Gesture Zone on Android & Chrome.
https://github.com/user-attachments/assets/8b5c3ebf-dbfc-4d33-a53b-d278d920ca3b

## Expected Behavior
It plainly shouldn't allow for scroll there.

snqb:
Reuploading the video, as it attached improperly above.

https://github.com/user-attachments/assets/94ed9b34-acec-4318-a1c8-3fec09f33864

raineorshine:
Thanks for the update. I don't see any comments explaining why `e.cancelable` is checked, so removing it worth a try!

PR #2654, opened 2024-11-26
remove cancelable from condition to prevent touch

snqb:
fixes https://github.com/cybersemics/em/issues/2618

some touch events are not cancelable, thus we don't prevent those
perhaps, it's for times when you scroll over big chunks of text(that's where I've seen the most of not-being-canceled)
but if I remove the && e.cancelable part, it does prevents the event just fine, so it also claims that falsely.

the behavior of events being cancelable or not are not consitent on Android browsers, e.g. sometimes they are in Chrome, but not on Vivaldi.

### DIFF
--- a/src/components/MultiGesture.tsx
+++ b/src/components/MultiGesture.tsx
@@ -109,7 +109,7 @@ class MultiGesture extends React.Component<MultiGestureProps> {
     document.body.addEventListener(
       'touchmove',
       e => {
-        if (this.disableScroll && e.cancelable) {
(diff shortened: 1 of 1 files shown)


snqb:
@raineorshine hey, do we merge this? android is not good with scrolling on the left, messes up a lot of gesture

raineorshine (approved):
Looks like it was added in bb35db828a3c39527860b0a705fe45f280422c93 but with no explanation why.

Issue #2452, opened 2024-10-13
Layout shifts when cursor is on url thought


raineorshine:
Please make sure to identify and clearly describe the cause of the bug before proposing a solution.

## Steps to Reproduce

```
- a
  - test.com
- b
```

Move the cursor between `a` and `test.com`

## Current Behavior

The position of `b` shifts slightly. It looks as if the height of the url thought is slightly too small when it has the cursor.

https://github.com/user-attachments/assets/0a33125e-16cc-47be-ac1d-3e75e70ff75e

## Expected Behavior

The thoughts should not shift position.

## Testing

Add a puppeteer snapshot test that covers this case. Make sure the test fails properly before the fix since this is such a small layout difference.

PR #2655, opened 2024-11-27
fix : layout shifts when cursor is on url thought

RED-ROSE515:
Issue : Layout shifts when cursor is on url thought #2452

There are two methods to figure this issue out.
**First Solution** : Currently, Ellipsize URL when not in edit mode, even if the URL can't be ellipsized.
                        So, I add the code to ellipsize url only when the URL can be ellipsized.
**Second Solution** : In `EllipsizedURL` recipe,  set the height as minThoughtHeight. Because of this, unexpected behavior happend.
                             So, I can add minHeight to `baseThought` recipe as minThoughtHeight.
                             
In this PR, I implemented First Solution.


raineorshine (changes requested):

`URL_MAX_CHARS` is actually an old technique that is no longer used. The `ellipsizeUrl` helper should be removed from the project. As you can see, it's not actually referenced anywhere in the code.

Url ellipsis is accomplished with [textOverflow: 'ellipsis'](https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/components/StaticThought.tsx#L214).

Therefore the solution should not assume ellipsis at a certain number of characters. Rather it is handled automatically when the thought exceeds the width of its container by the text-overflow property.

> **Second Solution** : In `EllipsizedURL` recipe, set the height as minThoughtHeight. Because of this, unexpected behavior happend.
> So, I can add minHeight to `baseThought` recipe as minThoughtHeight.

This sounds promising, and deserves some more investigation. Is `minThoughtHeight` incorrect? Does it scale properly with font size?

Adding a reference to the relevant inline documentation here:

https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/recipes/editable.ts#L21-L22

https://github.com/cybersemics/em/blob/efcc745b83bb029c08b5f3dd31ec8caf02af1b86/src/recipes/thought.ts#L22-L31

You're on the right track! We just need to get more specific about what exactly is wrong with `minThoughtHeight` and under what conditions.

> .single-line element has display: inline-block property, it affects the height calculation of the .thought div so we need to set the height manually.

---

Note: It took me a second to find it, but you have included the cause of the bug in your writeup to Second Solution. Instead, I recommend starting your PR description with the cause of the bug. You can also post the cause of the bug to the original issue even before your come up with a solution. The cause always goes before the solution (i.e. a diagnosis alway comes before the treatment plan).

src/recipes/thought.ts
```diff
@@ -25,7 +25,6 @@ const thoughtRecipe = defineRecipe({
         /**
           .single-line element has display: inline-block property, it affects the height calculation of the .thought div so we need to set the height manually.
          */
-        height: 'minThoughtHeight',
```
raineorshine: Make sure adjust inline comments when you change code. Otherwise they won't make sense anymore.

src/components/StaticThought.tsx
```diff
@@ -210,7 +210,6 @@ const StaticThought = ({
                 ...(isTableCol1 && { maxWidth: '100%' }),
                 ...(isAttribute(value) && { fontFamily: 'monospace' }),
                 ...(ellipsizedUrl && {
-                  display: 'inline-block',
```
raineorshine, line 213: I've tested this and can't find any regressions, so I will tentatively say it's safe to remove.

`inline-block` was added in the original url ellipsis implementation in  https://github.com/cybersemics/em/pull/2082/files#diff-60f5dcfc15327d5dd812d9df394c217efbedb4aa33dca782ed69d39dce811972R1818. It's not clear why, since it appears to work fine with the default `display: block`.

raineorshine (approved):

Issue #2466, opened 2024-10-19
splitSentence: Remove leading 'and'


raineorshine:
Add another case to the `splitSentence` command.

## Steps to Reproduce

```
one, two, and three
```

## Current Behavior

```
- one
- two
- and three
```

## Expected Behavior

```
- one
- two
- three
```

PR #2658, opened 2024-11-28
"and" as in-sentence splitter

snqb:
Fixes https://github.com/cybersemics/em/issues/2466

Now "and" counts just like a comma in separating <= 1 sentences

https://github.com/user-attachments/assets/e21abf29-56f2-4e69-96e8-6601ff19827a

### DIFF
--- a/src/util/splitSentence.ts
+++ b/src/util/splitSentence.ts
@@ -55,26 +55,21 @@ const splitSentence = (value: string): string[] => {
   // pattern2, multiple symbols: ?! !!! ...
   const mainSplitRegex = /[.;!?]+/g
 
-  const splitters = value.match(mainSplitRegex)
-
-  // When it cannot be split by the main spliter, spliter by ','
-  if (!splitters)
-    return value
-      .split(',')
-      .filter(s => s !== '')
-      .map(s => s.trim())
+  const sentenceSplitters = value.match(mainSplitRegex)
 
   /**
-   * Checks if the value has no other main split characters  except one period at the end.
+   * Checks if the value has no other main split characters  except one period at the end, i.e. value is just one sentence.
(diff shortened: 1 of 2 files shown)


src/util/splitSentence.ts
```diff
@@ -55,26 +55,21 @@ const splitSentence = (value: string): string[] => {
...
 
-  if (hasOnlyPeriodAtEnd())
+  // if we're sub-sentence or in one sentence territory, split by comma and "and"
+  // e.g. "john, johnson, and john doe" -> "- john - johnson - john doe"
+  if (!sentenceSplitters || hasOnlyPeriodAtEnd())
     return value
-      .split(',')
+      .split(/,|and/i)
+      .map(s => s.trim())
       .filter(s => s !== '')
```
snqb, line 72: swapping those allows me to have a simple regex that doesn't care about space

this commit shows that we either do a more complex regex with `\s*` in it, or just trim first, then filter :D
imo, this is way more concise
https://github.com/cybersemics/em/pull/2658/commits/36d59071a6e1ed0a55afbac15bdce07696491399

src/util/splitSentence.ts
```diff
@@ -55,26 +55,21 @@ const splitSentence = (value: string): string[] => {
...
 
-  const splitters = value.match(mainSplitRegex)
-
-  // When it cannot be split by the main spliter, spliter by ','
-  if (!splitters)
-    return value
-      .split(',')
-      .filter(s => s !== '')
-      .map(s => s.trim())
+  const sentenceSplitters = value.match(mainSplitRegex)
```
snqb, line 58: pardon me the liberty for renaming it
but feels like stressing the fact that it's a sentence splitter(as vs `,|and` which are splitters too)

PR #2659, opened 2024-11-28
Fix: Collapsing via the arrow icon causes a performance slowdown

dangold604:
## Problem

This pull request resolve a bug causing slow action(expand/collapse) of thoughts, speed of action performance as expected.
This issue occurred because the cursor was not activated when clicking on the expanded root thought bullet.

## Solution

The solution is to keep the cursor when clicking on the expanded root thought bullet.
Related to [#2649](https://github.com/cybersemics/em/issues/2649)

## Demo

https://github.com/user-attachments/assets/5aa41593-431e-4d2b-bd11-0b37a459a741


raineorshine:

Tapping on the bullet is only one of many ways to set the cursor to null. For example, if you hit escape or click the home icon the cursor will also be set to null. In that case, the virtualization is still broken. So the virtualization bug is not related to bullets or expansion states per se. I can see that virtualization breaks when the cursor becomes null though.

Note that at the design level, having the cursor thought be expanded is an invariant, so the suggested solution would violate that. The expansion state is determined by the location of the cursor, the surrounding thoughts, and if any thoughts are pinned open/closed. In **em**, unlike other outliners, you typically just navigate around, and thoughts expand/collapse automatically as the autofocus adjusts to the cursor location.

dangold604:
@raineorshine 
Are you suggesting that we should make virtualization work even when the cursor is null?

raineorshine:
Yes

code-withme-biz:
@raineorshine 
I’ve identified the reason why virtualization doesn’t work when the cursor is null. 

https://github.com/cybersemics/em/blob/a378795937eef25a712363435faca403ce9f060a/src/components/LayoutTree.tsx#L427

The condition checks whether the cursor is null or not, and based on that, decides whether to render the thoughts. I found that removing the `bellowCursor `variable from the condition ensures virtualization works correctly even when the cursor is null.

However, I would like to understand the original reasoning behind the creation of the `bellowCursor `variable. Once the purpose of this is clear, I believe we can resolve the issue of virtualization being disabled when the cursor is null.

dangold604:
@raineorshine 

I’ve identified the reason why virtualization doesn’t work when the cursor is null.

https://github.com/cybersemics/em/blob/a378795937eef25a712363435faca403ce9f060a/src/components/LayoutTree.tsx#L427

The condition checks whether the cursor is null or not, and based on that, decides whether to render the thoughts. I found that removing the `bellowCursor `variable from the condition ensures virtualization works correctly even when the cursor is null.

However, I would like to understand the original reasoning behind the creation of the `bellowCursor `variable. Once the purpose of this is clear, I believe we can resolve the issue of virtualization being disabled when the cursor is null.

raineorshine:
This discussion should probably go in #2649, not in this PR.

raineorshine (approved):

PR #2660, opened 2024-11-29
Fix Modal Export Tests

mstrofbass:
With the updates to fix the `thoughtspace` replication, these just needed to run the timers appropriately. We might be able to combine the sequential dispatches but I don't want to increase complexity until we confirm there aren't any timing issues.

If we encounter any intermittent errors in these tests then we might need to run the timers forward between the calls to `dispatch`.


raineorshine (approved):
Looks good!

mstrofbass:
This is ready to merge whenever.

PR #2672, opened 2024-12-03
upgrade to redux-thunk 3

snqb:
the only breaking change that affects us seems to be `export default thunk` -> `export thunk`, so just re-importing via named works.

other things they did were dropping UMD, going ESM, removing some type helpers, which we don't use.
all tests pass, and app works fine.
they recommend upgrading to Redux Toolkit as well, to use "Modern Redux", but I guess we don't really want that
so the changes are this simple :D

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7769,7 +7769,7 @@ __metadata:
     redux: "npm:^4.2.1"
     redux-devtools-extension: "npm:^2.13.9"
     redux-mock-store: "npm:^1.5.5"
-    redux-thunk: "npm:^2.4.2"
+    redux-thunk: "npm:^3.1.0"
     reselect: "npm:^5.1.1"
     tcp-port-used: "npm:^1.0.2"
     text-block-parser: "npm:^1.1.1"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -14931,12 +14931,12 @@ __metadata:
...
-"redux-thunk@npm:^2.4.2":
-  version: 2.4.2
-  resolution: "redux-thunk@npm:2.4.2"
+"redux-thunk@npm:^3.1.0":
+  version: 3.1.0
+  resolution: "redux-thunk@npm:3.1.0"
   peerDependencies:
-    redux: ^4
-  checksum: 10c0/e202d6ef7dfa7df08ed24cb221aa89d6c84dbaa7d65fe90dbd8e826d0c10d801f48388f9a7598a4fd970ecbc93d335014570a61ca7bc8bf569eab5de77b31a3c
+    redux: ^5.0.0
```
snqb, line 14938: I guess maybe we should upgrade redux to 5 first before this, working on that

raineorshine (approved):
Beautiful. Thanks for the explanation!

yarn.lock
raineorshine: Yes, though given that all the tests pass it seems to not be an issue.

raineorshine:
I'm going to merge so that the package-lock does not conflict after you upgrade redux.

PR #2673, opened 2024-12-03
Lottie icons 2408 (Milestone 4)

ohrytskov:
#2408 

- Special handling of ternary button: Sort
  - The sort button has three states: Ascending, Descending, and Manual
  - The icon needs to animate smoothly from one state to the next on click
    - Manual → Ascending
    - Ascending → Descending
    - Descending → Manual


ohrytskov:
I was reviewing the code in src/shortcuts/toggleSort.ts and came across the following snippet:

```typescript
// Show an error if the ranks do not match the sort condition.
// This is only needed for migrating to permasort, and can be removed after the migration is complete.
error: state => {
  if (!state.cursor || isRoot(state.cursor)) return null

  const simplePath = simplifyPath(state, state.cursor)
  const id = head(simplePath)
  const sortPreference = getSortPreference(state, id)
  if (sortPreference.type === 'None') return null

  const childrenSorted = getAllChildrenSorted(state, id)
  const childrenRanked = getChildrenRanked(state, id)
  return childrenSorted.length === childrenRanked.length &&
    !childrenRanked.every((childRanked, i) => childRanked.id === childrenSorted[i].id)
    ? 'Ranks do not match sort condition'
    : null
},
```

The comment mentions that:

> *"This is only needed for migrating to permasort, and can be removed after the migration is complete."*

I'm wondering if it's appropriate to remove this `error` function now.

Removing it could simplify the codebase and eliminate any unnecessary error messages that users might encounter.

What do you think? Would it be safe to proceed with removing this section, or are there any remaining considerations we should address before doing so?

src/selectors/getSortPreference.ts
```diff
@@ -1,31 +1,62 @@
...
-        type: childrenSort[0].value,
-        direction: getSortDirection(childrenSort[0].value, state, id),
-      }
-    : getGlobalSortPreference(state)
+  let sortId = findDescendant(state, id, ['=sort'])
+  let childrenSort = sortId ? getAllChildrenAsThoughts(state, sortId) : []
+
+  // If no direct sort is found, iteratively check parent context
+  let currentId: ThoughtId | null = id
+  while ((!sortId || childrenSort.length === 0) && currentId && currentId !== '__ROOT__') {
```
raineorshine, line 39: Can you explain the the while loop? It looks like it checks every ancestor, but sort is only supposed to be determined by the parent's meta attributes.

raineorshine:
> What do you think? Would it be safe to proceed with removing this section, or are there any remaining considerations we should address before doing so?

It can be removed, but in a separate PR.

src/selectors/getSortPreference.ts
ohrytskov: The while loop was initially designed to traverse up the ancestor chain to check for a =sort meta attribute. However, you're correct that sorting should only depend on the immediate parent's =sort meta attribute.

In the current implementation:

1. The loop checks every ancestor context until it either finds a =sort meta attribute with children or reaches the root.
2. This logic attempts to handle cases where no direct =sort is found for the current context or its immediate parent.

To align with the intended behavior - restricting sort determination to the immediate parent's =sort - the logic can be simplified to:

- Check the current context for =sort.
- If not found or empty, check the immediate parent's =sort meta.

The while loop is no longer necessary for this behavior. I’ve updated the logic accordingly in the revised version. Let me know if there’s any edge case you think might still require the loop.

raineorshine: Yes, that's closer to the expected behavior.

Maybe I'm missing something, but what was wrong with the original `getSortPreference`? It has been working fine as far as I know.

ohrytskov: Previously, getCursorSortDirection used cursorId directly, which didn't yield the correct sort direction. By updating it to use the parent of cursorId, the function now returns the correct sort direction. The new version works as intended.

ohrytskov: Since getCursorSortDirection in Sort.tsx has been updated to use parentId, the previous adjustments to getSortPreference are unnecessary.

ohrytskov:
All checks have passed. The PR is now ready for your review.

src/components/icons/Sort.tsx
```diff
@@ -1,15 +1,36 @@
...
+import getThoughtById from '../../selectors/getThoughtById'
 import head from '../../util/head'
 import AnimatedIcon from './AnimatedIcon'
 import animationData from './animations/10-sort_4.json'
 import animationDataDesc from './animations/11-sort-descanding_7.json'
 
+/** Find the parent ID of a given thought ID. */
+const findParentId = (state: State, thoughtId: ThoughtId): ThoughtId | null => {
+  const thought = getThoughtById(state, thoughtId)
+  return thought ? thought.parentId : null
```
trevinhofmann, line 16: This is fine as-is, but a small suggestion:
```suggestion
  return thought?.parentId ?? null
```

src/components/icons/Sort.tsx
```diff
@@ -1,15 +1,36 @@
...
+}
+
 /** Cursor Sort Direction. */
-const getCursorSortDirection = (state: State) => getSortPreference(state, head(state.cursor || HOME_PATH)).direction
+const getCursorSortDirection = (state: State) => {
+  const cursorId = head(state.cursor || HOME_PATH)
+
+  // Find the parent of cursorId
+  const parentId = findParentId(state, cursorId)
```
trevinhofmann, line 24: I could definitely be misunderstanding some nuances of `state.cursor` (for example, possible issues with the Context View), but wouldn't `parentId` generally be the second-to-last ID in the list (if there is one)?

```suggestion
  // Find the parent of cursorId
  const parentId = state.cursor[state.cursor.length - 2]
```

src/components/icons/Sort.tsx
```diff
@@ -1,15 +1,36 @@
...
+
+  // Get the sort preference using the parentId
+  const sortPref = getSortPreference(state, parentId!)
```
trevinhofmann, line 27: I haven't noticed any runtime errors during testing, but some of the types in this file seem off. My first concern is the non-null assertion operator for `parentId!`. I'd typically rather avoid using assertions based on assumptions that may or may not be true (now, or in the future).

Working backward from there, I don't see why `findParentId` necessarily needs to return `ThoughtId | null` rather than just `ThoughtId`. According to the existing types, `getThoughtById` should always return a non-null `Thought`, and every thought should have a `parentId` -- including the `HOME_PATH` with a `parentId` of itself.

That said, I'm not sure if `getThoughtById` _should_ be guaranteeing that it always returns a non-null `Thought` since it's looking up the given ID in a list which _may_ not contain it. 🤷‍♂️

src/components/icons/Sort.tsx
```diff
@@ -148,8 +169,12 @@ const IconDesc = ({ fill, size = 18, style = {}, cssRaw, animated, animationComp
 /** Sort Icon Component with Conditional Lottie Animation. */
 const SortIcon = ({ size = 18, style = {}, cssRaw, animated, animationComplete }: IconType) => {
   const direction = useSelector(getCursorSortDirection)
+
+  // Disable animation for HOME_PATH or no direction
+  const isAnimated = direction === 'Desc' ? true : animated && direction !== null
```
trevinhofmann, line 174: An OR might be simpler than the ternary here:

```suggestion
  const isAnimated = direction === 'Desc' || (animated && direction !== null)
```

src/selectors/findDescendant.ts
```diff
@@ -12,10 +12,14 @@ const findDescendant = (state: State, thoughtId: ThoughtId | null, values: strin
   if (!thoughtId || values.length === 0) return thoughtId
   if (!Array.isArray(values)) values = [values]
   // if the value is a meta attribute, use childrenMap for O(1) lookup
+
+  const currentThought = getThoughtById(state, thoughtId)
```
trevinhofmann, line 16: Since the refactoring in this file is not relevant to the purpose of the PR, I would prefer see it in a separate PR.

However, it is a minor change (and a good one!), so I would be okay with this.

trevinhofmann (changes requested):
Thanks for the PR, @ohrytskov! I have a few small comments regarding possible refactoring, but the implementation looks fairly clean and the basic animations seem to be working as expected:

https://github.com/user-attachments/assets/31085f8d-279e-43d1-967a-16c7ca125c84

---

## 1: Animation triggered when navigating from `Unsorted` or `Ascending` to `Descending`

### Overview

When navigating from any list which is either unsorted or sorted in ascending order to any list which is sorted in ascending order, the Sort button's animation is triggered.

### Steps to Reproduce

```
- fruits
  - apple
  - orange
  - banana
  - pear
```

1. Set the cursor on `apple`
2. Click to sort `Descending`
3. Click to sort `Ascending`
4. Set the cursor on `fruits`
5. Set the cursor on `apple`

### Current Behavior

The `Sort` button's animation is triggered.

https://github.com/user-attachments/assets/8cdcc1d8-a03e-4c23-af0e-2bf1069c585d

### Expected Behavior

The animation should not be triggered when navigating. It should only be triggered when actively changing the sorting of a list.

---

> > What do you think? Would it be safe to proceed with removing this section, or are there any remaining considerations we should address before doing so?

> It can be removed, but in a separate PR.

I created a separate issue to track this: #2683

While testing, I found a condition which triggers the error state. I will post those details on #2683 shortly.

raineorshine:
Thanks @trevinhofmann!

@ohrytskov I'm noticing that the arrow is not pointed in the correct direction on Sort Descending. It should point down in that case. 

It is shown correctly in the screenshot at the top of #2408. Maybe it was incorrect in the Lottie JSON? Either way, the arrow should reverse between Sort Ascending and Sort Descending. Thanks so much.

ohrytskov:
All suggested changes have been implemented. If there are no further concerns, this PR is ready for final review.

src/components/ToolbarButton.tsx
```diff
@@ -81,17 +83,40 @@ const ToolbarButton: FC<ToolbarButtonProps> = ({
...
+  const direction = useSelector((state: State) => {
+    if (shortcutId !== 'toggleSort') return null
+    return getCursorSortDirection(state)
+  })
+
+  /** Determines whether to animate the toolbar button based on the shortcut and state. */
+  const shouldAnimate = () => {
+    // Handle the 'toggleSort' shortcut separately
+    if (shortcutId === 'toggleSort') {
+      return direction === null || direction === 'Asc'
```
trevinhofmann, line 96: The animations are functioning correctly (nice work!), but the reasoning is not obvious to me while reading the code.

@ohrytskov - Would you please explain why animating when there is "no direction" or "ascending" makes sense? I would think there should be an animation when changing to "descending" or "ascending", but I must be missing something.

Edit: Thinking about this more, I guess the reason is that we animate when changing _from_ these states. I.e., when changing from "no direction" or "ascending", we are changing _to_ "ascending" or "descending" and expect the animation. Is that accurate?

src/util/getCursorSortDirection.ts
```diff
@@ -0,0 +1,38 @@
...
+import getSortPreference from '../selectors/getSortPreference'
+import head from '../util/head'
+
+/** Retrieves the parent ID from the state's cursor. */
+const getParentIdFromCursor = (state: State): ThoughtId | null => {
+  const cursor = state.cursor
+
+  if (!cursor || cursor.length < 2) {
+    // No cursor or cursor has less than 2 items (no parent)
+    return head(HOME_PATH)
```
trevinhofmann, line 13: Can this be simplified? I don't think this is worth holding up the PR, but a minor suggestion:

```suggestion
    return HOME_TOKEN
```

trevinhofmann (approved):
Thanks for the update, @ohrytskov! The latest changes look great to me, and everything appears to be functioning as expected. The issue `1: Animation triggered when navigating from Unsorted or Ascending to Descending` is no longer happening, and the arrow direction for descending has also been fixed. I haven't been able to find any other issues with the functionality.

I have one question about the `shouldAnimate` logic (which I may have already answered myself) and one very small refactoring suggestion, but neither issue is blocking. Thanks!

https://github.com/user-attachments/assets/f54b4641-5f7a-4b77-8898-3367b2583fb7

src/components/ToolbarButton.tsx
ohrytskov: Yes, that's accurate. We animate the state change when transitioning from "no direction" or "ascending" to "ascending" or "descending".

ohrytskov:
@raineorshine, regarding the descending animation - yes, you are correct - the arrow for "Sort Descending" was not pointing in the correct direction due to an issue with the Lottie JSON file. I've fixed it by rotating the "Sort Ascending" JSON animation by 180 degrees. Everything should now look correct with the arrow pointing in the required direction for descending sort.

raineorshine:
Hey, thanks. We actually just need the arrow flipped, not the entire icon. Should I reach out to the Lottie designer or is that something you can do?

ohrytskov:
@raineorshine, I can definitely take care of flipping the arrow. Could you please clarify whether you're referring to the arrow in the Lottie animation or the SVG icon? Knowing this will help me make the appropriate adjustments. Thanks!

ohrytskov:
Lottie animations are JSON-based and describe animations with multiple shapes, keyframes, and transformations, unlike static SVGs. While the animation enriches the visual with motion, SVGs can provide a static representation, offering fallback or placeholder functionality. Using both SVG and Lottie together ensures improved performance (SVGs load instantly, while JSON animates on interaction), enhanced interactivity (static visuals transition into dynamic animations), and compatibility (SVGs serve as a fallback for unsupported environments). This approach balances simplicity and engagement: SVGs handle immediate visuals, while Lottie animations add dynamic, responsive elements, creating a seamless and polished user experience.

ohrytskov:
Could you take a closer look at the current implementation? Are you sure it’s not already set up? Let me know if there’s something specific you’d like clarified!

raineorshine:
Thank you for the explanation. I still have some general questions about the Lottie + SVG strategy for my own understanding, but I can save those for a side discussion. The structure is in place and working.

After watching Trevin's video, I see the transition between Ascending and Descending is not quite right. The bars snap abruptly from ascending to descending rather than animating as they should.

I have reached out to the Lottie designer to get a visual demo of the correct behavior from Manual → Ascending → Descending → [repeat]. This way you can have a clear target. I will get that to you asap. Thanks!

raineorshine:
I'm still waiting on the Lottie designer to get back to me with a demo of the expected behavior. This is necessary since the issue is not just with the arrow. I will let you know as soon as I have something to show you.

raineorshine (changes requested):
Hi. I reviewed the animation more closely and it actually looks very good! The three states are exactly what I'm looking for. My apologies, I thought that the bars on the right were not animating, but that was only when going from Descending → Manually, which is expected! 

The only issue I'm seeing is that the arrow head is filled when the arrow is at rest. 

## Current

image

## Expected

Brave Browser 2024-12-16 14 18 34

src/components/ToolbarButton.tsx
```diff
@@ -81,17 +83,40 @@ const ToolbarButton: FC<ToolbarButtonProps> = ({
...
+    }
+
+    // General condition for other shortcuts
+    if ((!isActive && !commandState) || (isActive && !isButtonActive)) {
+      return true
+    }
+
+    // If none of the above conditions are met, do not animate
+    return false
```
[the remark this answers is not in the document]
ohrytskov: Thank you for the insightful suggestions! I agree that simplifying the logic and inlining the `shouldAnimate` function improves readability and reduces complexity. I've updated the code as per your recommendations:

- Simplified the return statement to use nested ternaries.
- Inlined the `shouldAnimate` logic directly into the `setIsAnimated` call.
- Added comments for clarity as you've suggested.

ohrytskov:
@raineorshine, the PR appears to be finalized.

raineorshine (approved):

Issue #2580, opened 2024-11-11
Upgrade Redux to v5


raineorshine:
There are breaking changes: #2576

Release notes: https://github.com/reduxjs/redux/releases/tag/v5.0.0

PR #2675, opened 2024-12-03
Upgrade to Redux 5

snqb:
fixes https://github.com/cybersemics/em/issues/2580

As per https://github.com/reduxjs/redux/releases/tag/v5.0.0

- Now we use `UnknownAction` instead of `AnyAction`. It's what the redux team recommends insistently. Why not, it didn't affect anything.
- We can't assume(as recommended) the `action` in middlewares being an action -- it's `unknown` now, so we have to check it via `isAction` or narrow the types. This affected a couple files, where I had to add checks.
- renamed the imports to use `legacyCreateStore` because seeing strikethrough is not pleasant, and they're being very insistent on deprecating ~`createStore`~ and going for rtk. 
- `useStore` now started giving an `{}` state instead of `State`, so i had to add global types for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "react-signature-pad-wrapper": "^3.4.0",
     "react-split-pane": "^0.1.92",
     "react-transition-group": "^4.4.5",
-    "redux": "^4.2.1",
+    "redux": "^5.0.1",
     "redux-devtools-extension": "^2.13.9",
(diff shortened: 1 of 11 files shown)


snqb:
please can you see 64c40c93a8629ab685af4137cde4083927321440

raineorshine (approved):
Looks good! Thanks!

Issue #2364, opened 2024-09-03
Tutorial: Refactor GestureDiagram condition


raineorshine:
## Steps to Reproduce

Start the Tutorial and got to Tutorial Step 2 on mobile.

Can be reproduced using the device toolbar in Chrome Dev Tools.

## Current Behavior

The condition specifying all the tutorial steps in which the GestureDiagram should be rendered is defined in `/src/components/Tutorial/index.tsx`, separate from the tutorial step components.

## Expected Behavior

The GestureDiagram condition should be co-located with the tutorial step components.

Note that the GestureDiagram needs to be rendered below the `.tutorial` element, whose height depends on the screen size of the device. The tutorial step component just contains the content of the tutorial step, not the paginator or Next/Prev buttons. You will need to do some refactoring to render the GestureDiagram in the right place while ensuring the condition is co-located in the tutorial step component.

PR #2679, opened 2024-12-04
Hide tutorial gesture steps logic behind a portal & colocation

snqb:
Closes https://github.com/cybersemics/em/issues/2364

After considering:
- additional gesture exports per file tutorial file
- another map of step -> gesture
- esoteric shit like `TutorialStepSecondThought.gesture = ...`
- Portals

I decided to go with Portals. Now after finishing the stuff, I really like the decision.
- the gesture thing depends on an inner state of many possible components, and renders kinda independently -- isn't it what Portals are for?
- it allowed very subtle additions of `<TutorialGestureDiagram gesture={commandById('newThought')?.gesture} />` into already existing `<TutorialHint />` conditional component
- no export/map shenanigans, no new abstractions
- lots of conditional variables were already used in the step components, so most of files are just +2 lines(import and jsx)
- we got less additions than deletions, and got much more clarity

also, feels like I changed no behaviors(judging via human brain static analysis). Sadly, we don't have automated tests for this. I did manual testing though, and it all seems to work as it should.
Thought of adding tests, but it'd take me quite a bit more time.

### DIFF
--- a/src/components/Tutorial/TutorialGestureDiagram.tsx
+++ b/src/components/Tutorial/TutorialGestureDiagram.tsx
@@ -1,0 +1,27 @@
+import { createPortal } from 'react-dom'
+import { css } from '../../../styled-system/css'
+import Command from '../../@types/Command'
+import GesturePath from '../../@types/GesturePath'
+import GestureDiagram from '../GestureDiagram'
+
+/** A portal for conditionally rendering Gesture hints. Used to show an appropriate gesture  below the tutorial UI up top, thus Portal. */
+const TutorialGestureDiagram = ({ gesture }: { gesture: Command['gesture'] }) => {
+  const target = document.getElementById('tutorial-gesture-diagram-portal')
+
+  return (
+    gesture &&
+    target &&
+    createPortal(
+      <GestureDiagram
+        path={gesture as GesturePath}
+        size={160}
+        strokeWidth={10}
+        arrowSize={5}
+        cssRaw={css.raw({ animation: 'pulse 1s infinite alternate' })}
+      />,
+      target,
+    )
+  )
+}
+
+export default TutorialGestureDiagram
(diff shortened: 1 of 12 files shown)


raineorshine:
Thanks! Looks promising. Will review soon.

src/components/Tutorial/TutorialGestureDiagram.tsx
```diff
@@ -0,0 +1,27 @@
+import { createPortal } from 'react-dom'
+import { css } from '../../../styled-system/css'
+import Command from '../../@types/Command'
+import GesturePath from '../../@types/GesturePath'
+import GestureDiagram from '../GestureDiagram'
+
+/** A portal for conditionally rendering Gesture hints. Used to show an appropriate gesture  below the tutorial UI up top, thus Portal. */
+const TutorialGestureDiagram = ({ gesture }: { gesture: Command['gesture'] }) => {
+  const target = document.getElementById('tutorial-gesture-diagram-portal')
```
raineorshine, line 9: One thing we'll want to do is put this ID in a shared constant. This will help prevent one changing without the other.

src/components/Tutorial/index.tsx
```diff
@@ -148,34 +102,9 @@ const Tutorial: FC = () => {
...
-            cursor &&
-            cursorHeadValue &&
-            cursorHeadValue.toLowerCase() === TUTORIAL_CONTEXT2_PARENT[tutorialChoice].toLowerCase()) ||
-          (tutorialStep === TUTORIAL2_STEP_CONTEXT2_SUBTHOUGHT_HINT &&
-            cursor &&
-            cursorHeadValue &&
-            cursorHeadValue.toLowerCase() === TUTORIAL_CONTEXT[tutorialChoice].toLowerCase())) &&
-        gesturePath ? (
+        {isTouch && (
           <div
```
raineorshine, line 106: The container needs to be hidden when the TutorialGestureDiagram is not rendered, otherwise a transparent black bar will appear. It may not be visible since it happens to be rendered in the empty space above the first thought, but it's probably better for it not to be rendered at all when there is no gesture being displayed.

raineorshine (changes requested):

snqb:
I've addressed the issues

raineorshine (approved):

Issue #2535, opened 2024-11-02
Second browser back does not work after cursorDown


raineorshine:
## Steps to Reproduce

```
- a
- b
```

1. Set the cursor on `a`.
2. cursorDown
3. Hit the back button in the browser.
4. cursorDown
5. Hit the back button in the browser.

## Current Behavior

Nothing happens.

## Expected Behavior

Cursor should move back to `a`.

PR #2682, opened 2024-12-05
Fix back button doesn't work properly after cursorDown.

RED-ROSE515:
Issue : Second browser back does not work after cursorDown #2535

Reason : The problem occurs because the code is checking for path `equality (equalPath(lastPath, path))` in the `onPopstate` handler, which is preventing proper navigation when you revisit the same path.

Solution : So, I updated the setting lastPath logic.   : lastPath = state.cursor

1. When user is on 'a' and navigates to 'b':
   lastPath becomes 'a'
2. When user hits back button:
   state.cursor is 'b', so lastPath becomes 'b'
   Navigation goes back to 'a'
3. When user navigates to 'b' again:
   lastPath is 'b' (preserved from previous step)
4. When user hits back button:
   Since lastPath is 'b', we know we came from 'b'

### DIFF
--- a/src/e2e/puppeteer/__tests__/browserBack.ts
+++ b/src/e2e/puppeteer/__tests__/browserBack.ts
@@ -1,0 +1,53 @@
+import clickThought from '../helpers/clickThought'
+import paste from '../helpers/paste'
+import press from '../helpers/press'
+import waitForEditable from '../helpers/waitForEditable'
+import { page } from '../setup'
+
+vi.setConfig({ testTimeout: 20000 })
+
+it('browser back button should work correctly after cursor movements', async () => {
+  const importText = `
+  - a
+  - b`
+
+  await paste(importText)
+
+  await waitForEditable('b')
+  await clickThought('a')
+  await press('ArrowDown')
+  await page.goBack()
+  await waitForEditable('a')
+  await press('ArrowDown')
+  await page.goBack()
+  const editable = await waitForEditable('a')
+  expect(editable).toBeTruthy()
+})
+
+it('browser back button should work correctly with complex cursor navigation', async () => {
+  const importText = `
+  - a
+  - b
+  - c`
+
+  await paste(importText)
+
+  await waitForEditable('c')
+  await clickThought('c')
+  await press('ArrowUp')
+  await waitForEditable('b')
+  await press('ArrowUp')
+  await waitForEditable('a')
+  await page.goBack()
+  await waitForEditable('b')
+  await page.goBack()
+  await waitForEditable('c')
+  await press('ArrowUp')
+  await waitForEditable('b')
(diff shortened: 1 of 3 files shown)


raineorshine (changes requested):
Thanks for your submission. I can confirm that the original case is fixed.

The previous implementation determined the current path from the url. I'm not sure exactly why, as theoretically we should just use `state.cursor`, as you did. If that's true, we should switch over to using `state.cursor` completely. It doesn't make sense to use the decoded `path` sometimes and `state.cursor` others. It needs to be a unified implementation. Maybe git blame will reveal more about the intentions of the decoded url path.

In any case, I did find a regression, so we'll need to keep working on this. Given the likelihood of regressions, adding proper test coverage is essential. Could you add puppeteer tests for the two cases under discussion currently? #2535 and the below regression case. We can add more cases if we encounter any other regressions.

## Steps to Reproduce

```
- a
- b
- c
```

1. Set the cursor on `c`
2. cursorUp *(cursor is on `b`)*
3. cursorUp *(cursor is on `a`)*
4. Back *(cursor is on `b`)*
5. Back *(cursor is on `c`)*
6. cursorUp *(cursor is on `b`)*
7. cursorUp *(cursor is on `a`)*
8. Back

## Current Behavior

Cursor does not move.

## Expected Behavior

Cursor should move from `b` to `a`.

src/util/initEvents.ts
```diff
@@ -113,26 +113,35 @@ const scrollAtEdge = (() => {
   return { start, stop }
 })()
 
+let lastPath: Path | null
+
+/** Update the LastPath. */
+export const updateLastPath = (path: Path) => {
+  lastPath = path
+}
```
raineorshine, line 121: A mutable global variable is considered a bad practice and should be avoided. A mutable variable with a limited scope (under closure) can occasionally be justified. For state that spans multiple files, we really should stick to proper state management solutions, like Redux or our internal ministore.

src/e2e/puppeteer/__tests__/browserBack.ts
```diff
@@ -0,0 +1,53 @@
...
+  await paste(importText)
+
+  await waitForEditable('b')
+  await clickThought('a')
+  await press('ArrowDown')
+  await page.goBack()
+  await waitForEditable('a')
+  await press('ArrowDown')
+  await page.goBack()
+  const editable = await waitForEditable('a')
```
raineorshine, line 23: `waitForEditable` just finds a thought on the page, not the cursor thought.

You can use https://github.com/cybersemics/em/blob/16fd8ed8e1bb6d54e6c0a8253a9e9e15ff7b1e40/src/e2e/puppeteer/helpers/getEditingText.ts to find the cursor thought.

PR #2690, opened 2024-12-08
Remove asyncFocus from swapNote command

### LINKED ISSUE #2290: [mobile] swapNote: Keyboard opens but selection is not set
## Steps to Reproduce

```
- a
  - b
```

1. Set the cursor on `b`.
2. Activate `swapNote`. 
3. Close the keyboard. 
4. Activate `swapNote`.

## Current Behavior

The keyboard is opened on steps 2 and 4. 

## Expected Behavior

The keyboard should stay closed.

`state.editing` should be unchanged on swapNote.

ethan-james:
Fixes #2290 

The `swapNote` command explicitly calls `asyncFocus()` which create a browser selection and opens the keyboard. Removing this function call stops it from taking control of the selection & keyboard.

### DIFF
--- a/src/commands/swapNote.ts
+++ b/src/commands/swapNote.ts
@@ -1,7 +1,6 @@
 import Command from '../@types/Command'
 import { swapNoteActionCreator } from '../actions/swapNote'
 import ConvertToNoteIcon from '../components/icons/ConvertToNoteIcon'
-import asyncFocus from '../device/asyncFocus'
 import hasMulticursor from '../selectors/hasMulticursor'
 import isDocumentEditable from '../util/isDocumentEditable'
 
@@ -17,7 +16,6 @@ const swapNote: Command = {
   },
   svg: ConvertToNoteIcon,
   exec: dispatch => {
-    asyncFocus()
     dispatch(swapNoteActionCreator())
   },
 }
--- a/src/actions/swapNote.ts
+++ b/src/actions/swapNote.ts
@@ -6,6 +6,7 @@ import findDescendant from '../selectors/findDescendant'
 import { anyChild } from '../selectors/getChildren'
 import getRankAfter from '../selectors/getRankAfter'
 import isContextViewActive from '../selectors/isContextViewActive'
+import pathToThought from '../selectors/pathToThought'
 import simplifyPath from '../selectors/simplifyPath'
 import appendToPath from '../util/appendToPath'
 import ellipsize from '../util/ellipsize'
@@ -70,6 +71,8 @@ const swapNote = (state: State) => {
             const oldPath = appendToPath(cursor, noteId, noteChildId)
             const newPath = appendToPath(cursor, noteChildId)
             const newRank = getRankAfter(state, appendToPath(simplePath, noteId))
+            const note = pathToThought(state, oldPath)
+
             return reducerFlow([
               moveThought({ oldPath, newPath, newRank }),
               // delete =note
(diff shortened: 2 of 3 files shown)


src/commands/swapNote.ts
```diff
@@ -17,7 +16,6 @@ const swapNote: Command = {
   },
   svg: ConvertToNoteIcon,
   exec: dispatch => {
-    asyncFocus()
```
ethan-james, line 20: @raineorshine Presumably there was a reason why this was called in the first place. I found the change in cybersemics/em@1286df98513de8ba3be37cd3d4d152cf6a74406d but I wasn't able to find an associated PR or other comments on how this came about. Please let me know if there is other behavior that you want to see.

raineorshine: Thanks for investigating. asyncFocus is used to open the virtual keyboard, so you're right that is not desired when it starts closed. 

Can you confirm that if the virtual keyboard is open at the beginning of the test case, it stays open through all steps? I want to make sure this fix doesn't cause a regression in editing mode.

ethan-james: Yes, it does stay open. Drawing the gesture doesn't seem to affect the browser selection.

trevinhofmann (approved):
Looks good to me! Thank you, @ethan-james.

From my testing, the issue did not affect Android/Chrome. For iOS/Safari, this change fixes the issue for me.

The keyboard also continues to stay open if it was already open during the gesture.

raineorshine (changes requested):
Thanks!

The only case that is not working for me is when the caret starts in the note.

## Steps to Reproduce

```
- a
  - b
```

1. Set the cursor on `b`.
2. Activate swapNote.
3. Set the caret in the note `b`.
4. Activate swapNote.

## Current Behavior

The virtual keyboard closes.

## Expected Behavior

The virtual keyboard should stay open. The caret should either stay at the same character offset or move to the end of the thought.

In general the virtual keyboard should stay open if it is already open, and stay closed if it is already closed (consistent with `state.editing`).

ethan-james:
@raineorshine I think I fixed it by adding `[aria-label=note-editable]` elements to the `selection.isEditable()` function. The only thing that I notice is that the caret jumps from the end of the note to the beginning of the thought after the former is converted to the latter. I feel like I've noticed that in other contexts, is there already an issue open for that?

raineorshine:
No, I don't believe so. Can you try setting `state.cursorOffset`? I'm not sure if that will work, but it's worth trying. Control over the selection offset is limited, but it would be nice to keep the offset at the end of the thought.

raineorshine:
Either preserving the offset or placing the caret at the end of the thought would be fine. I find that placing it at the beginning is less useful in practice when editing.

ethan-james:
Here's the relevant line in the `setCursor` action:

```
const updatedOffset = offset ?? (editingValueStore.getState() !== null ? 0 : null)
```

I can change that `0` to `length - 1` if you want to impact all `setCursor` actions, or I can add `cursorOffset: length - 1` to `setCursor({ path: parentNoteChildId ? appendToPath(parentOf(cursor), parentNoteChildId) : parentOf(cursor) })` in `swapNote` if you only want this to affect `swapNote`.

I would do `cursorOffset: typeof state.cursorOffset === 'number' ? state.cursorOffset : length - 1` but it'll already be set to `0` rather than `null` (by `updatedOffset` above) so it will still place the caret at the beginning in that case.

raineorshine:
Let's do it just in `swapNote`. Thanks!

raineorshine (approved):

PR #2693, opened 2024-12-10
Disable hiddenInput unless asyncFocus is invoked

### LINKED ISSUE #2258: [mobile] Keyboard opens after swiping sidebar closed
## Steps to Reproduce

1. Create thought `a`.  
2. Close the keyboard. 
3. Tap the hamburger menu to open the sidebar. 
4. Swipe the sidebar closed. 

## Current Behavior

The keyboard opens. 

## Expected Behavior

The keyboard should not open.

ethan-james:
Fixes #2258 

When the sidebar opens, it fires a `focus` event in iOS Safari with a `relatedTarget` attribute containing the `hiddenInput` element, indicating that this hidden element is losing focus. When the sidebar is closed, as best I can tell it returns focus to this hidden element. Disabling the element except when `asyncFocus` is called prevents it from re-opening the keyboard.


ethan-james:
There is a second piece to this that I still need to complete. Currently, when the sidebar closes, `useEditMode` is restoring focus to the element where `isEditing` is true. This behavior isn't appropriate when the keyboard was closed, so I need to figure out a reliable way of tracking `focus` state within the editable to determine whether it should restore focus.

raineorshine:
> When the sidebar opens, it fires a `focus` event in iOS Safari with a `relatedTarget` attribute containing the `hiddenInput` element, indicating that this hidden element is losing focus. When the sidebar is closed, as best I can tell it returns focus to this hidden element. Disabling the element except when `asyncFocus` is called prevents it from re-opening the keyboard.

Thanks, this looks good.

> There is a second piece to this that I still need to complete. Currently, when the sidebar closes, `useEditMode` is restoring focus to the element where `isEditing` is true. This behavior isn't appropriate when the keyboard was closed, so I need to figure out a reliable way of tracking `focus` state within the editable to determine whether it should restore focus.

`state.editing` is meant to be true when the virtual keyboard is open, and false when it is closed. `useEditMode` will set `shouldSetSelection` to false and thus not restore the selection [if `state.editing` is false](https://github.com/cybersemics/em/blob/16fd8ed8e1bb6d54e6c0a8253a9e9e15ff7b1e40/src/components/Editable/useEditMode.ts#L64), i.e. when the virtual keyboard is closed.

I suggest we investigate why `state.editing` is true in this case if the virtual keyboard really is closed. 

FYI The idea behind `useEditMode` is that it handles cursor caret restoration automatically without having to micromanage it in individual commands/components.

ethan-james:
Great, thanks for that context! I wasn't looking forward to intercepting `focus` & `blur` events directly.

ethan-james:
Although, my understanding is that you can still be "editing" a node even if there's no selection. The bullet stays highlighted and you can draw gestures while the keyboard is closed. Are these 2 different pieces of state?

raineorshine:
Most commands are relative to the cursor thought: `state.cursor`. The bullet is highlighted with a gray circle around it.

`state.editing` specifically corresponds to whether the virtual keyboard is up or not. It is always true on desktop.

Let me know if that doesn't answer your question though.

ethan-james:
Nope, that's perfect, thanks. I think I already knew that but needed to be reminded.

ethan-james:
When the sidebar opens, it pulls focus from the `Editable` and triggers its `onBlur` handler. It's simple enough to ignore that blur when `state.showSidebar` is true and stop it from dispatching `editingAction({ value: false })`. This ensures that the keyboard state is preserved (open or closed) when you tap on the backdrop to dismiss the sidebar.

Tapping on the hamburger menu to close the sidebar is another story. In iOS Safari, it does not trigger a `click` event when it is tapped to open the sidebar, but it does trigger one when it closes. This causes the `Editable` to blur some time after `toggleSidebar(false)` runs and sets `state.editing` to false after the sidebar closes. So the keyboard always stays closed after tapping the hamburger menu twice to open & close the sidebar.

You can see that it's easy to reproduce this behavior on `main` without using `clearThought`:

1. Create a thought
2. Put the caret inside the thought
3. Tap the hamburger menu to open the sidebar
4. Tap the hamburger menu to close the sidebar

ethan-james:
Delaying the closing of the sidebar until after the `click` event fires allows the `Editables` to ignore all sidebar-related `blur` events. I think this is a good idea as long as it doesn't cause any unwanted side effects.

trevinhofmann (approved):
Thanks, @ethan-james! The implementation seems reasonable to me, and is working as expected (the keyboard does not appear, regardless of whether I close the sidebar by clicking the hamburger button or clicking outside of the sidebar). I haven't noticed any regressions from these changes.

Demo on iOS Safari:

https://github.com/user-attachments/assets/0498e4c4-86af-4e76-8056-e792b64f6de8

raineorshine (changes requested):
Hi, thanks. This is an interesting approach. However, the added delay is quite noticeable. The purpose of `fastClick` is to avoid such a delay. You'll need to find a way to solve this without losing the `fastClick` timing.

raineorshine:
FYI I re-wrote the original issue with simpler steps to reproduce. I don't know why I originally thought it was related to clearThought.

ethan-james:
I managed to stop the `click` event by adding `preventDefault` to the `touchend` event. Unfortunately, it still doesn't restore the browser selection. It seems to be triggering the hook in `useEditMode`, so I just need to figure out a more reliable way to re-create the selection.

raineorshine:
Can you remind me, why is `useEditMode` restoring the selection when the keyboard is down? I may have missed this point.

Also just to clarify, the keyboard should always be closed after the sidebar is closed.

ethan-james:
In order to fix #2552, I set it up to restore focus on the current element when the sidebar is closed.

It should always be closed? So...

1. click on a thought to open keyboard
2. open sidebar
3. close sidebar
4. keyboard closed?

If so, then it's already doing that 😄

raineorshine:
Yes, I realize I am contradicting myself. Sorry for the confusion. 

Truthfully I would be okay with either of the following:

1. If the keyboard starts open, keep it open and preserve the caret offset. 
2. Or, close the keyboard and keep it closed. 

The reason is that the keyboard does close when the sidebar is opened, which could be interpreted as editing mode ending. I'm relaxing the requirement if that results in a more elegant solution, as I think it's still intuitive for the user.

If this is the current behavior, I can review shortly. Thanks!

ethan-james:
The Material UI modal attempts to restore focus to the editable when it closes, and I think that would be the truly elegant solution. However, it's not actually restoring the browser selection, so the keyboard stays closed. I think I'd like to:

1. Briefly try to figure out what's blocking the focus restoration, so that I can point to a specific cause
2. If it's happening for a good reason (or an inelegant-to-fix reason), remove my previous changes to `useEditMode` so that the keyboard consistently stays closed

Let me know how that sounds to you.

raineorshine:
Sure! That would be the ideal solution, at least when edit mode is enabled. Thanks for investigating. Let me know what you come up with.

ethan-james:
I haven't had too much success finding a root cause, but I have found a variety of disparate things:

In the `onFocus` handler in `Editable`, `setCursorOnThought` gets short-circuited by this line:

```
if ((!editing || editingMode) && equalPath(cursor, path)) return
```

but that doesn't really matter because, in `useEditMode`, this line stops it from calling `setSelectionToCursorOffset`:

```
const cursorWithoutSelection = editingCursorOffset !== null || !selection.isActive()
```

because there's no active selection due to the component being blurred when the sidebar opens. `setSelectionToCursorOffset` would call `selection.set()` which would (hopefully) programmatically create a browser selection.

My gut feeling is that the restoration of the `focus` event that is performed by Material UI's modal component should trigger the selection natively, but it doesn't, and the things that I can think of to try don't change that. Namely:

1. calling `asyncFocus()` first
2. Removing the `onFocus` handler entirely from `Editable`

It does seem to work in this much different example: https://codesandbox.io/embed/mjtznp?module=/src/Demo.tsx&fontsize=12 but I see the same behavior in em in desktop Chrome, meaning that it doesn't restore focus after closing the sidebar.

That's about all the time I can spare to look into it at the moment, as I need to pack and then fly tomorrow. If you'd like, I can quickly remove the workarounds from `Editable` and `useEditMode` so that the keyboard stays closed and `editable` is correctly set to `false`. If we do figure out how to make Material UI's focus restoration work properly in the future, then those workarounds shouldn't really be necessary.

raineorshine:
Thanks so much, this is useful background at least. The sidebar case presents an unexpected situation for the existing selection logic. In general it requires two taps to enter edit mode on mobile (first to set the cursor and second to open the keyboard). It never expects the selection to be restored by another component. Thus I think many of the assumptions made for `onFocus`, `setCursorToThought`, and `useEditMode` are invalidated by the sidebar behavior.

We can add exceptions as needed, but as you have seen, this touches several files and starts to grow difficult to manage. If `state.editing` and `state.cursorOffset` are set before the sidebar restores the focus, it *may* be possible to set up the conditions so that `useEditMode` triggers correctly. But without further investigation that is purely speculative.

My sense at this point is that it would be better to keep the keyboard closed after the sidebar is activated. It is reasonable behavior from the user's perspective, and the restoring the caret properly does not seem to be worth the added complexity.

src/components/Sidebar.tsx
```diff
@@ -89,6 +89,10 @@ const Sidebar = () => {
       disableSwipeToOpen={!isTouch}
       ref={containerRef}
       transitionDuration={durations.get('fast')}
+      // On iOS Safari, restoring focus works when tapping the backdrop to close the sidebar, but not when tapping the hamburger
+      // menu to close the sidebar. Hopefully the hamburger menu can be fixed and focus can be restored properly in all cases.
+      // Until then, letting the backdrop (correctly) restore focus results in inconsistent behavior.
+      ModalProps={{ disableRestoreFocus: true }}
```
raineorshine, line 95: Let's only disable this on mobile, since desktop works as expected.

src/components/HamburgerMenu.tsx
```diff
@@ -90,12 +91,16 @@ const HamburgerMenu = () => {
...
+        {...fastClick(e => {
           // TODO: Why does the sidebar not open with fastClick or onTouchEnd without a setTimeout?
           // onClick does not have the same problem
           setTimeout(() => {
             dispatch(toggleSidebar({}))
           }, 10)
+          // The Editables need to ignore the click event and its corresponding blur, fired after all of the touch events.
+          if (isTouch && isSafari()) {
+            e.preventDefault()
+          }
```
raineorshine, line 103: Can you remind me why this is needed? What happens if it is omitted?

It's often helpful in an inline comment for a special case to include an "Otherwise..." clause to make it super clear.

Thanks!

src/components/Sidebar.tsx
```diff
@@ -89,6 +89,10 @@ const Sidebar = () => {
       disableSwipeToOpen={!isTouch}
       ref={containerRef}
       transitionDuration={durations.get('fast')}
+      // On iOS Safari, restoring focus works when tapping the backdrop to close the sidebar, but not when tapping the hamburger
+      // menu to close the sidebar. Hopefully the hamburger menu can be fixed and focus can be restored properly in all cases.
+      // Until then, letting the backdrop (correctly) restore focus results in inconsistent behavior.
```
raineorshine, line 94: Good comment 👍

raineorshine (changes requested):

src/components/HamburgerMenu.tsx
ethan-james: If it's omitted, Safari fires a click event when the sidebar is closed, despite not firing one when the sidebar is opened. It's not necessary for this implementation that we've settled on, but it will block future efforts at focus restoration. Up to you whether it stays in or gets removed for now. I like it because it makes the events symmetrical, and there are fewer of them.

raineorshine:
The browser selection is still not being restored for me on fdc54c0ec8 on desktop the way that it is in `main`. Are you getting the same thing?

ethan-james:
> The browser selection is still not being restored for me on [fdc54c0](https://github.com/cybersemics/em/commit/fdc54c0ec89c1e0711626865c8df9955da46146f) on desktop the way that it is in `main`. Are you getting the same thing?

Yes, I took out the code that was restoring focus on desktop because I assumed that we would want to keep it consistent. I can put it back if you want, although I'm not able to test on the iPad until I get back on a network where it can connect to my laptop dev server. We're likely to end up in back in the situation where the backdrop restores focus but the hamburger menu does not.

ethan-james:
> > The browser selection is still not being restored for me on [fdc54c0](https://github.com/cybersemics/em/commit/fdc54c0ec89c1e0711626865c8df9955da46146f) on desktop the way that it is in `main`. Are you getting the same thing?
> 
> Yes, I took out the code that was restoring focus on desktop because I assumed that we would want to keep it consistent. I can put it back if you want, although I'm not able to test on the iPad until I get back on a network where it can connect to my laptop dev server. We're likely to end up in back in the situation where the backdrop restores focus but the hamburger menu does not.

And actually, the desktop behavior is not correct because it never sets `editing` to `false`. So if you blur the selection, then open/close the sidebar, it will erroneously restore focus to the editable. Maybe that's not something you were ever worried about, maybe that was always me.

raineorshine:
Thank you for clarifying the current functionality.

Desktop is a different situation because it does not have edit mode the same way that mobile does. While mobile has to balance editing ease with screen space taken up the virtual keyboard, desktop has no such tradeoff. There is no downside to having an active selection on desktop, so we want to restore the selection on desktop as much as possible. 

At the moment desktop works perfectly on `main`. The goal is to fix #2258 without changing the desktop behavior.

> I can put it back if you want, although I'm not able to test on the iPad until I get back on a network where it can connect to my laptop dev server.

No rush. You can either post an update that I can test on mobile, or I can wait until you're able to test both platforms.

> We're likely to end up in back in the situation where the backdrop restores focus but the hamburger menu does not.

That doesn't seem to be the case now on desktop on `main`. I'd like to try to just make the changes to fix #2258 on mobile. If you run into a wall and are not able to do this without introducing desktop regressions, please let me know and I can try it myself.

> And actually, the desktop behavior is not correct because it never sets `editing` to `false`. So if you blur the selection, then open/close the sidebar, it will erroneously restore focus to the editable. Maybe that's not something you were ever worried about, maybe that was always me.

I'm fine with that, since an active browser selection is almost always a good thing on desktop.

ethan-james:
OK, thanks for the clarification! I'll sort it out when I get a chance.

raineorshine (approved):
👍

Issue #2687, opened 2024-12-07
Thoughts marked `=done` should be collapsed


raineorshine:
Thoughts marked `=done` should be collapsed by default like `=pin/false`.

## Steps to Reproduce

```
- a
  - =children
    - =pin
      - true
  - b
    - c
  - d
    - =done
    - e
```

## Current Behavior

`d` is expanded.

## Expected Behavior

`d` should be collapsed (unless the cursor is on it or one of its ancestors).

The behavior should be the same as `=pin/false`.

## Testing

Add appropriate tests to https://github.com/cybersemics/em/blob/1acd87a79df8a9c929a3d72650cc98451a1c32d4/src/selectors/__tests__/expandThoughts.ts. You can copy and adapt the `=pin/false` that are relevant.

PR #2694, opened 2024-12-10
Thoughts marked =done should be collapsed

RED-ROSE515:
This PR resolves this issue : Thoughts marked =done should be collapsed #2687.
I think there is no need to explain the cause of bug. It's closer to implementing the function.

**Solution** : 

 - Added a new helper function isDone that checks if a thought has the =done attribute
 - Modified the hasOnlyChild condition to check for =done in addition to =pin/false
 - Modified the expansion filter to also check for =done when determining if a child should be expanded
 - These changes will make thoughts marked with =done behave like thoughts marked with =pin/false - they will be collapsed by 
    default unless the cursor is on them or one of their ancestors or they are explicitly pinned with =pin/true.


raineorshine (approved):
Perfect. Thanks!

Issue #2629, opened 2024-11-25
formatSelection: IndexSizeError on set whole thought color after selection color


raineorshine:
I found a regression with `formatSelection` when changing the text color of the entire thought after setting the text color of a selection.

## Steps to Reproduce

1. Create thought `Hello world`.
2. Select the text `world`.
3. Open the ColorPicker and click the blue swatch.
5. Click the blue swatch again.

## Current Behavior

`Uncaught IndexSizeError: Failed to execute 'collapse' on 'Selection': There is no child at offset 2.`

<img width="735" alt="image" src="https://github.com/user-attachments/assets/63f4f88a-9d81-4ffb-91d7-feffb91e84c1">

## Expected Behavior

There should be no error, and the whole thought should be blue.

## Note

I have not verified, but this looks to me like `window.getSelection().focusOffset` is not handled correctly when `window.getSelection().focusNode.nodeType` is an ELEMENT_NODE.

> Offset: If the focus node is a TEXT_NODE, then offset refers to the character offset of the caret. If the focus node is an ELEMENT_NODE, then offset is the index of the element before the caret. For example, offset: 1 on an ELEMENT_NODE means the caret is after the focus node. Thus, do not mistakenly assume the offset is always a character offset.

RED-ROSE515:
@raineorshine , I can't find the regression here.
It works well on my side.

RED-ROSE515:
I will check about that.

raineorshine:
Added: #2684

PR #2696, opened 2024-12-12
Fix : Uncaught IndexSizeError

RED-ROSE515:
This PR resolves the issue : formatSelection: IndexSizeError on set whole thought color after selection color #2629

The error is actually occurring in the restore function when trying to collapse the selection.
Cause of bug :  We're directly saving the `focusNode` and `focusOffset` from the selection without validating what type of node it is or if the offset is valid for that node type. When the saved node is an `ELEMENT_NODE` and the saved offset is greater than the number of child nodes that element has, we got this error.

So, I handle this by checking the node type and adjusting the offset accordingly.

Solution : So, I have changed the `restore` function to handle the selection restoration more safely.
 - I added proper validation of the selection object.
 - And I added type-specific offset validation.
 - Wrapped the operation in a try-catch block with a fallback to clear the selection if restoration fails

### DIFF
--- a/src/device/selection.ts
+++ b/src/device/selection.ts
@@ -57,14 +57,8 @@ export const isCollapsed = (): boolean => !!window.getSelection()?.isCollapsed
 export const isActive = (): boolean => !!window.getSelection()?.focusNode
 
 /** Returns true if the Node is an editable. */
-const isEditable = (node?: Node | null) => {
-  const element = node as HTMLElement
-  return (
-    !!element &&
-    element.nodeType === Node.ELEMENT_NODE &&
-    (element.hasAttribute('data-editable') || element.ariaLabel === 'note-editable')
-  )
-}
+const isEditable = (node?: Node | null) =>
+  !!node && node.nodeType === Node.ELEMENT_NODE && !!(node as HTMLElement).hasAttribute?.('data-editable')
 
 /** Returns true if the selection is on a thought. */
 // We should see if it is possible to just use state.editing and selection.isActive()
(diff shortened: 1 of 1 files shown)


src/device/selection.ts
```diff
@@ -205,10 +199,31 @@ export const restore = (savedSelection: SavedSelection | null): void => {
   }
 
   const sel = window.getSelection()
+  if (!sel) return
+
+  try {
```
raineorshine, line 204: You can remove the try-catch since you handled the IndexSizeError.

raineorshine (changes requested):
Thanks, looks good. I can confirm the error no longer occurs.

There is a case where the offset is wrong after switching the color a couple times, but I will save that for another issue. It will involve novel logic for preserving the perceived offset even when the HTML changes. This PR successfully addresses the unhandled exception.

> The error is actually occurring in the restore function when trying to collapse the selection.
> Cause of bug : We're directly saving the `focusNode` and `focusOffset` from the selection without validating what type of node it is or if the offset is valid for that node type. When the saved node is an `ELEMENT_NODE` and the saved offset is greater than the number of child nodes that element has, we got this error.

Thank you. That's a good explanation of what is happening in general, though it is missing an explanation of why the focusOffset would be out of range in this particular case. Presumably it is because the contents of the focusNode have changed, but `selection.save()` does not have a mechanism to update its internal representation of the caret. Since you have exact Steps to Reproduce, you could provide the specific offset that is causing the error. That would help illuminate the issue by providing an example that proves the general argument you are making. No need at this point, but something to keep in mind.

RED-ROSE515:
Ah, I got it. I will enhance my explanation with the specific case.

PR #2701, opened 2024-12-14
dragAndDropThought: Wait for highlighted bullet to unrender

raineorshine:


raineorshine:
Test failed in the CI with the highlight bullet visible and the QuickDrop panel visible. The QuickDrop panel is explicitly hidden before the snapshot is taken, so I'm not sure why it would be visible.

raineorshine:
384e58e timed out

Closing as neither of these attempts consistently makes the test pass.

Issue #2313, opened 2024-08-21
Tutorial: Test completion of tutorial


raineorshine:
Add a ~puppeteer~ RTL test that walks through the entire tutorial, ensuring that it can be completed.

Note that createTestApp currently [skips the tutorial](https://github.com/cybersemics/em/blob/12a8c54ddf1f396dd864a91728436f387509645e/src/test-helpers/createTestApp.tsx#L37) so you will need to add an option to override this. I suggest `{ tutorial: true }`.

trevinhofmann:
Just curious -- Why the change from puppeteer to RTL? Wondering if puppeteer screenshots/snapshots would be valuable for the tutorial.

raineorshine:
I decided to go with RTL tests for test suite performance reasons. Since the main goal here is to test that the tutorial can successfully be completed, I didn't think it was worth it to cover the visuals given the added test time.

PR #2702, opened 2024-12-15
Tests for Tutorial

snqb:
closes https://github.com/cybersemics/em/issues/2313

A `userEvents`-only run through the Tutorial.
Assumes desktop, as relies on keyboard shortctuts.
Goes through _most_ steps(excluding mostly hint steps), ensures those and faithfully replicates a user going through the instructions.
0 dispatches used.

{C0A862A8-D2FD-46F0-8491-5A8911420188}

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,8 +7,5 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[typescriptreact]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
-  "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
+  }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,5 @@
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[typescriptreact]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,4 +8,7 @@
   "[typescriptreact]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
+  "[typescript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
 }
--- a/src/components/Tutorial/Tutorial2StepContextViewOpen.tsx
+++ b/src/components/Tutorial/Tutorial2StepContextViewOpen.tsx
@@ -46,7 +46,7 @@ const Tutorial2StepContextViewOpen = () => {
       {TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}" or "{TUTORIAL_CONTEXT2_PARENT[tutorialChoice]}" to show it again.
     </p>
   ) : contextViewClosed ? (
-    <p>Oops, somehow the context view was closed. Select "Relationships".</p>
+    <p>Oops, somehow the context view was closed. Select "{TUTORIAL_CONTEXT[tutorialChoice]}".</p>
   ) : (
     <>
       <p>
--- a/src/components/Tutorial/TutorialNavigationButton.tsx
+++ b/src/components/Tutorial/TutorialNavigationButton.tsx
@@ -14,7 +14,7 @@ const TutorialNavigationButton = React.forwardRef<
 >(({ clickHandler, value, disabled = false, classes }, ref) => (
   <a
     className={cx(anchorButtonRecipe({ variableWidth: true, smallGapX: true }), classes)}
-    onClick={clickHandler}
+    // onClick={clickHandler}
     {...{ disabled }}
     {...fastClick(clickHandler)}
     ref={ref}
--- a/src/components/Tutorial/TutorialNavigationButton.tsx
+++ b/src/components/Tutorial/TutorialNavigationButton.tsx
@@ -14,7 +14,6 @@ const TutorialNavigationButton = React.forwardRef<
 >(({ clickHandler, value, disabled = false, classes }, ref) => (
   <a
     className={cx(anchorButtonRecipe({ variableWidth: true, smallGapX: true }), classes)}
-    // onClick={clickHandler}
     {...{ disabled }}
     {...fastClick(clickHandler)}
     ref={ref}
--- a/src/components/Tutorial/__tests__/Tutorial.ts
+++ b/src/components/Tutorial/__tests__/Tutorial.ts
@@ -121,7 +121,7 @@ describe('Tutorial 2', async () => {
   beforeEach(() => createTestApp({ tutorial: true }))
   afterEach(cleanupTestEventHandlers)
   afterAll(cleanupTestApp)
-  it('we learn about context indicators', async () => {
+  it('we learn about context view', async () => {
     await user.click(screen.getByText('Learn more'))
 
     expect(screen.getByText(/shows a small number to the right of the thought, for example/)).toBeInTheDocument()
@@ -141,29 +141,29 @@ describe('Tutorial 2', async () => {
   })
 
   describe('step context 1 - create a "Home" to-do list', () => {
-    it('we create a Home thought', async () => {
+    it('we create a "Home" thought', async () => {
       await user.keyboard('{Enter}')
       await user.type(lastEmptyThought(), 'Home')
       await act(vi.runOnlyPendingTimersAsync)
       expect(screen.getByText(/Add a thought with the text "To Do"/)).toBeInTheDocument()
     })
 
-    it('we add a Home->To Do subthought', async () => {
+    it('we add a "Home" • "To Do" subthought', async () => {
       await user.keyboard('{Control>}{Enter}{/Control}')
       await user.type(lastEmptyThought(), 'To Do')
       await act(vi.runOnlyPendingTimersAsync)
       expect(screen.getByText(/Now add a thought to “To Do”/)).toBeInTheDocument()
     })
 
-    it('we add a Home->To Do->or to not subthought sub-subthought', async () => {
+    it('we add a "Home" • "To Do" • "Or to not subthought" sub-subthought', async () => {
       await user.keyboard('{Control>}{Enter}{/Control}')
-      await user.type(lastEmptyThought(), 'or to not')
+      await user.type(lastEmptyThought(), 'Or to not')
       await act(vi.runOnlyPendingTimersAsync)
 
       /* thoughts:
       - Home
         - To Do
-          - or to not
+          - Or to not
       */
       expect(screen.getByText(/Nice work!/)).toBeInTheDocument()
     })
@@ -191,18 +191,18 @@ describe('Tutorial 2', async () => {
       expect(screen.getByText(/Now add a thought with the text "To Do"/)).toBeInTheDocument()
     })
 
-    it('we add another Work->To Do thought', async () => {
+    it('we add another "Work" • "To Do" thought', async () => {
       await user.keyboard('{Control>}{Enter}{/Control}')
       await user.type(lastEmptyThought(), 'To Do')
       expect(screen.getByText('Imagine a new work task. Add it to this “To Do” list.'))
     })
 
-    it('we see context indicators', async () => {
+    it('we see contexts with a superscript', async () => {
       expect(screen.getAllByRole('superscript')[0]).toHaveTextContent('2')
       expect(screen.getByText(/This means that “To Do” appears in two places/)).toBeInTheDocument()
     })
 
-    it('we add a Work->To Do->Text boss', async () => {
+    it('we add a "Work" • "To Do" • "Text boss" thought', async () => {
       await user.keyboard('{Control>}{Enter}{/Control}')
       await user.type(lastEmptyThought(), 'Text boss')
       await act(vi.runOnlyPendingTimersAsync)
--- a/src/components/Tutorial/__tests__/Tutorial.ts
+++ b/src/components/Tutorial/__tests__/Tutorial.ts
@@ -1,140 +1,145 @@
 import { getByText, screen } from '@testing-library/dom'
 import userEvent from '@testing-library/user-event'
 import { act } from 'react'
-import { importTextActionCreator as importText } from '../../../actions/importText'
-import { newThoughtActionCreator as newThought } from '../../../actions/newThought'
 import {
   HOME_TOKEN,
   TUTORIAL2_STEP_CHOOSE,
+  TUTORIAL2_STEP_CONTEXT1,
+  TUTORIAL2_STEP_CONTEXT1_PARENT,
+  TUTORIAL2_STEP_CONTEXT1_SUBTHOUGHT,
+  TUTORIAL2_STEP_CONTEXT2_PARENT,
+  TUTORIAL2_STEP_CONTEXT_VIEW_EXAMPLES,
+  TUTORIAL2_STEP_CONTEXT_VIEW_OPEN,
+  TUTORIAL2_STEP_CONTEXT_VIEW_SELECT,
+  TUTORIAL2_STEP_START,
   TUTORIAL_STEP_AUTOEXPAND,
+  TUTORIAL_STEP_AUTOEXPAND_EXPAND,
+  TUTORIAL_STEP_FIRSTTHOUGHT,
   TUTORIAL_STEP_FIRSTTHOUGHT_ENTER,
   TUTORIAL_STEP_SECONDTHOUGHT,
+  TUTORIAL_STEP_SECONDTHOUGHT_ENTER,
+  TUTORIAL_STEP_START,
   TUTORIAL_STEP_SUBTHOUGHT,
+  TUTORIAL_STEP_SUBTHOUGHT_ENTER,
   TUTORIAL_STEP_SUCCESS,
 } from '../../../constants'
 import exportContext from '../../../selectors/exportContext'
 import store from '../../../stores/app'
-import createTestApp, { cleanupTestApp, createTestAppWithTutorial } from '../../../test-helpers/createTestApp'
-import dispatch from '../../../test-helpers/dispatch'
-import { setCursorFirstMatchActionCreator as setCursorFirstMatch } from '../../../test-helpers/setCursorFirstMatch'
+import createTestApp, { cleanupTestApp, cleanupTestEventHandlers } from '../../../test-helpers/createTestApp'
 
 // as per https://testing-library.com/docs/user-event/options/#advancetimers
 // we should avoid using { delay: null }, and use jest.advanceTimersByTime instead
 const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime })
 
-describe.skip('Tutorial test keyboard events', () => {
-  beforeEach(createTestApp)
-  afterEach(cleanupTestApp)
-  it('2 enters', async () => {
-    await user.keyboard('{Enter}')
-    await user.keyboard('{Enter}')
-  })
+/** Get last created thought in the document, used mostly after `user.keyboard('{Enter}')`. */
+const lastThought = () => Array.from(document.querySelectorAll('[contenteditable="true"]')).at(-1)!
 
-  it('another 2 enters', async () => {
-    await user.keyboard('{Enter}')
-    await user.keyboard('{Enter}')
-  })
-
-  it('final 2 enters', async () => {
-    await user.keyboard('{Enter}')
-    await user.keyboard('{Enter}')
-
-    const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
-    console.log(exported)
-  })
-
-  afterAll(() => {
-    const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
-    console.log(exported)
-  })
-})
+/** Ensure we are at the given step of the tutorial. */
+const ensureStep = (step: number) => {
+  expect(store.getState().storageCache?.tutorialStep).toBe(step)
+}
 
 describe('Tutorial 1', async () => {
-  beforeEach(createTestAppWithTutorial)
+  beforeEach(() => createTestApp({ tutorial: true }))
+  afterEach(cleanupTestEventHandlers)
   afterAll(cleanupTestApp)
   describe('step start', () => {
     it('shows the welcome text', () => {
+      ensureStep(TUTORIAL_STEP_START)
       expect(screen.getByText('Welcome to your personal thoughtspace.')).toBeInTheDocument()
     })
 
     it('shows the "Next" button, that leads to the first step', async () => {
       await user.click(screen.getByText('Next'))
-      expect(
-        screen.getByText('First, let me show you how to create a new thought', { exact: false }),
-      ).toBeInTheDocument()
+      expect(screen.getByText(/First, let me show you how to create a new thought/)).toBeInTheDocument()
     })
   })
 
   it('step first thought - how to create a thought ', async () => {
+    ensureStep(TUTORIAL_STEP_FIRSTTHOUGHT)
     expect(screen.getByText('Hit the Enter key to create a new thought.')).toBeInTheDocument()
-
-    await act(() => user.keyboard('{Enter}'))
+    await user.keyboard('{Enter}')
     expect(screen.getByText('You did it!')).toBeInTheDocument()
-    expect(screen.getByText('Now type something. Anything will do.')).toBeInTheDocument()
 
-    await dispatch(newThought({ value: 'my first thought' }))
+    ensureStep(TUTORIAL_STEP_FIRSTTHOUGHT_ENTER)
+    expect(screen.getByText('Now type something. Anything will do.')).toBeInTheDocument()
+    await user.type(lastThought(), 'my first thought')
+    await user.keyboard('{Enter}')
     await act(vi.runOnlyPendingTimersAsync)
-    expect(screen.getByText('Click the Next button when you are ready to continue.')).toBeInTheDocument()
-    expect(store.getState().storageCache?.tutorialStep).toBe(TUTORIAL_STEP_FIRSTTHOUGHT_ENTER)
   })
 
   it('step second thought - prompts to add another thought', async () => {
-    await dispatch(newThought({ value: 'a thought' }))
-    await act(vi.runOnlyPendingTimersAsync)
-    expect(store.getState().storageCache?.tutorialStep).toBe(TUTORIAL_STEP_SECONDTHOUGHT)
+    expect(screen.getByText('Well done!')).toBeInTheDocument()
+    expect(screen.getByText(/Try adding another thought/)).toBeInTheDocument()
+    ensureStep(TUTORIAL_STEP_SECONDTHOUGHT)
 
-    expect(screen.getByText('Try adding another thought. Do you remember how to do it?')).toBeInTheDocument()
-    await dispatch(newThought({ value: 'a new thought' }))
+    await user.type(lastThought(), 'my second thought')
+    await user.keyboard('{Enter}')
     await act(vi.runOnlyPendingTimersAsync)
+
+    ensureStep(TUTORIAL_STEP_SECONDTHOUGHT_ENTER)
+
     expect(screen.getByText('Good work!')).toBeInTheDocument()
+    expect(screen.getByText('Now type some text for the new thought.')).toBeInTheDocument()
+    await user.type(lastThought(), 'third thought')
+    await user.keyboard('{Enter}')
+    await act(vi.runOnlyPendingTimersAsync)
   })
 
   it('step subthought - how to create a thought inside thought', async () => {
-    await user.click(screen.getByText('Next'))
-    expect(store.getState().storageCache?.tutorialStep).toBe(TUTORIAL_STEP_SUBTHOUGHT)
-    await dispatch([
-      newThought({ value: 'uncle' }),
-      newThought({
-        value: 'parent',
-      }),
-      newThought({
-        value: 'child',
-        insertNewSubthought: true,
-      }),
-    ])
+    ensureStep(TUTORIAL_STEP_SUBTHOUGHT)
+    expect(exportContext(store.getState(), [HOME_TOKEN], 'text/plain')).toBe(
+      `- __ROOT__
+  - my first thought
+  - my second thought
+  - third thought
+  - `,
+    )
+
+    // Now I am going to show you how to add a thought within another thought.
+    expect(screen.getByText(/Now I am going to show you how to add a thought/)).toBeInTheDocument()
+
+    // since we have cursor on empty thought
+    expect(screen.getByText(/Hit the Delete key to delete the current blank thought/)).toBeInTheDocument()
+    expect(screen.getByText(/Then hold the Ctrl key and hit the Enter key/)).toBeInTheDocument()
+    await user.keyboard('{Backspace}')
+    await user.keyboard('{Control>}{Enter}{/Control}')
     await act(vi.runOnlyPendingTimersAsync)
 
-    expect(screen.getByText(`As you can see, the new thought "child" is nested`, { exact: false })).toBeInTheDocument()
-    expect(getByText(screen.getByTestId('tutorial-step'), 'parent', { exact: false })).toBeInTheDocument()
+    await user.type(lastThought(), 'child')
+    await act(vi.runOnlyPendingTimersAsync)
+    expect(exportContext(store.getState(), [HOME_TOKEN], 'text/plain')).toBe(
+      `- __ROOT__
+  - my first thought
+  - my second thought
+  - third thought
+    - child`,
+    )
+    // as you can see, the new thought "child" is nested within "third thought"
+    expect(screen.getByText(/As you can see, the new thought "child" is nested/)).toBeInTheDocument()
+    expect(getByText(screen.getByTestId('tutorial-step'), /"third thought"/)).toBeInTheDocument()
+
+    ensureStep(TUTORIAL_STEP_SUBTHOUGHT_ENTER)
+    await user.click(screen.getByText('Next'))
   })
 
-  describe('step autoexpand - shows that thoughts expand and collapse on outside click', async () => {
-    it('tells us about thoughts being hidden when clicked away', async () => {
-      await user.click(screen.getByText('Next'))
-
-      expect(store.getState().storageCache?.tutorialStep).toBe(TUTORIAL_STEP_AUTOEXPAND)
-      expect(
-        screen.getByText('thoughts are automatically hidden when you click away', { exact: false }),
-      ).toBeInTheDocument()
+  describe('step autoexpand', async () => {
+    it('thoughts are automatically hidden when you click away', async () => {
+      ensureStep(TUTORIAL_STEP_AUTOEXPAND)
+      expect(screen.getByText(/thoughts are automatically hidden when you click away/)).toBeInTheDocument()
     })
 
-    it('tells us to click on uncle to hide the child', async () => {
-      await dispatch([
-        importText({
-          text: `
-      - parent
-        - child
-      - uncle
-      `,
-        }),
-      ])
-      await act(() => user.click(screen.getByText('uncle')))
+    it('click on "uncle" thought to hide child', async () => {
+      await user.click(screen.getByText('my second thought'))
       expect(screen.getByText('Notice that "child" is hidden now.', { exact: false })).toBeInTheDocument()
+      ensureStep(TUTORIAL_STEP_AUTOEXPAND_EXPAND)
     })
 
-    it('tells us to click on the parent to reveal the child', async () => {
-      await act(() => user.click(Array.from(screen.getAllByText('parent')).at(-1)!))
+    it('click back on a "parent" thought to reveal child', async () => {
+      expect(screen.getByText(/Click "third thought" to reveal its subthought "child"/)).toBeInTheDocument()
+      await user.click(screen.getAllByText('third thought').at(-1)!)
       expect(screen.getByText('Lovely. You have completed the tutorial', { exact: false })).toBeInTheDocument()
+      ensureStep(TUTORIAL_STEP_SUCCESS)
     })
   })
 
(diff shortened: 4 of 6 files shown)


src/test-helpers/createTestApp.tsx
```diff
@@ -47,6 +47,9 @@ const createTestApp = async () => {
   })
 }
 
+/** Set up testing and mock document and window functions, and show tutorial. */
+export const createTestAppWithTutorial = () => createTestApp({ tutorial: true })
```
snqb, line 51: this, to preserve `beforeEach(createTestApp)` syntax

snqb:
in proccess of fixing various errors, not much ready for review yet
will be soon, though

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,317 @@
...
+    await dispatch(newThought({ value: 'my first thought' }))
+    await act(vi.runOnlyPendingTimersAsync)
+    expect(screen.getByText('Click the Next button when you are ready to continue.')).toBeInTheDocument()
+  })
+
+  describe('step second thought - create another thought', () => {
+    it('prompts to type some text to the created thought, then congratulates on typing', async () => {
+      await dispatch([
+        newThought({ value: 'already created thought' }),
+        tutorialStep({ value: TUTORIAL_STEP_SECONDTHOUGHT }),
```
trevinhofmann, line 57: The PR description says that manual clicking is usually prioritized over store-side creation, which I think I would also prefer. However, there are quite a few places here where we dispatch actions to the store rather than interacting with the page. Is there a reason to do that instead of e.g. clicking the "Next" button like we do elsewhere (`await user.click(screen.getByText('Next'))`)?

Same for e.g. `await dispatch(newThought({ value: 'my first thought' }))` rather than typing in the thought.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,317 @@
...
+
+    expect(screen.getByText(`As you can see, the new thought "child" is nested`, { exact: false })).toBeInTheDocument()
+    expect(getByText(screen.getByTestId('tutorial-step'), 'parent', { exact: false })).toBeInTheDocument()
+  })
+
+  describe('step autoexpand - show that thoughts expand and collapse on outside click', async () => {
+    beforeAll(() => dispatch(tutorialStep({ value: TUTORIAL_STEP_AUTOEXPAND })))
+    it('tells us about thoughts being hidden when clicked away', async () => {
+      expect(
+        screen.getByText('thoughts are automatically hidden when you click away', { exact: false }),
```
trevinhofmann, line 89: Is there a reason to use `{ exact: false }` for all of these?

https://testing-library.com/docs/queries/about/#precision

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,317 @@
...
+      await act(vi.runOnlyPendingTimersAsync)
+      expect(screen.getByText("Add a subthought and I'll show you.", { exact: false })).toBeInTheDocument()
+
+      await dispatch(newThought({ value: 'child', insertNewSubthought: true }))
+      await act(vi.runOnlyPendingTimersAsync)
+      expect(
+        screen.getByText(`Try clicking on thought "uncle" to hide subthought "child".`, { exact: false }),
+      ).toBeInTheDocument()
+
+      await dispatch(setCursorFirstMatch(['uncle']))
```
trevinhofmann, line 114: Another example where I think it might be preferable to actually click on the thought.

trevinhofmann (changes requested):
Looking good overall, @snqb. Thank you for the PR!

Just a few initial questions.

src/components/Tutorial/__tests__/Tutorial.ts
snqb: yeah, I kinda lied in the end...
I had these spans of statements to find the correct thought in the space, click on it, trigger a keyboard shortcut to add a thought by `user.type`, but in the end it felt so pointless, as I found myself the core-est functionality of the app, whilst I was tasked with testing a tutorial. And that added some complexity, removing the clarity I tried to achieve, since it's one of the bigger test files in the project
and it felt like dispatching new thoughts was as productive as clicking on things, as the tutorial's logic relies on redux state a lot.

and as per `user.click(screen.getByText('Next'))`, there are two reasaons:
1. a lot of times we don't have the `next` button, the next step is triggered conditionally from store values
2. the test cases became too dependant on each other. E.g. we remove one test case -- and have a cascade of failures, since we didn't increment the step on the removed case. One could argue it's fine, but the ergonomics of being able to do `it.only` or `it.skip` wherever we want to isolate cases, was too tempting.
And I've kinda tested this `Next` thing in `shows the "Next" button, that leads to the first step` case, and its functionality doesn't differ in other cases

src/components/Tutorial/__tests__/Tutorial.ts
snqb: reason's simple -- a lot of times in the tutorial the text is quite long, spanning 3-4-5 lines
I tried to find the most crucial pieces of the text on a given step, that clearly indicates what's going on, without going through the pain of doing multi-line matches

src/components/Tutorial/__tests__/Tutorial.ts
snqb: yeah, right
at first it indeed was a click, but again, due to me feeling like I'm testing the core functionality of expanding thoughts, covered already in `src/selectors/__tests__/expandThoughts.ts`, I went for a simpler approach to preserve brevity
I can of course rewrite this, but I feel like yesterday-me had a point :D

src/components/Tutorial/__tests__/Tutorial.ts
raineorshine: Thanks for the explanation. While I can see the desire to simplify the tests, I think we lose something essential by not imitating the user's exact actions. Whenever you dispatch an action directly, you open a gap in the test coverage. How do you know that the user can trigger the action? The only way to ensure that the user can complete the tutorial is by automating the exact steps that the user would take.

> 1. a lot of times we don't have the `next` button, the next step is triggered conditionally from store values

This is a good example of my point. Enabling the conditions for the Next button to appear should be under test coverage. This is acutally the area where the tutorial has experienced the most regressions. If a Redux selector changes unintentionally, then the tutorial step is not testing the correct condition any more, preventing the user from advancing. My main reason for #2313 was to put the tutorial advancement under test coverage. Manually advancing defeats this purpose.

> 2\. the test cases became too dependant on each other. E.g. we remove one test case -- and have a cascade of failures, since we didn't increment the step on the removed case. One could argue it's fine, but the ergonomics of being able to do `it.only` or `it.skip` wherever we want to isolate cases, was too tempting.

The root dependency is that the tutorial steps are dependent on each other. If there is a bug in step 3, then the user naturally cannot advance to step 4. If the test for step 3 could fail but the test for step 4 still pass, that indicates a gap in test coverage.

It's possible to jump to the correct tutorial step at the beginning of each test, but then you would have to recreate the start state exactly as it was at the end of the previous step. Otherwise, there would be a gap in the test coverage.

I hope that helps explain the importance of imitating the user's actions in order to ensure that the tutorial can be completed. If a certain set of actions seem overly complex or difficult to abstract, feel free to consult with Trevin to find a clean solution.

snqb: yeah, now I totally agree with "next" example, thank you

but still, I feel like in some places(like autoexpand) I shouldn't do manual clicks I feel like, since it's already covered

let me see again which parts are lacking attention

raineorshine: The expansion behavior itself is covered in other tests, but we would still want to test that the tutorial can detect that the thought has been collapsed or expanded per the instructions to the user.

snqb:
Hey, I've added some showcase of keyboard issues I've got.
Apparently, it's kind of tricky to cleanly create thoughts via `userEvent`.
We get insane results like this tree(got it via `exportContext`) after `user.keyboard` and `user.type` manipulations.
{9C748D62-8535-4102-A643-0E23E38BA25B}

You can just run the tests via 
` yarn vitest src/components/Tutorial/ --project unit` and tinker around with what's going on.

Below we have an almost-working test suite for the tutorial, where I'm stuck with this keyboard behavior, so you might want to have a look there, removing the `.only` from the first `describe`.



{1D445E4D-613B-4FDA-A8FF-96CB9EDE24BC}

trevinhofmann:
@snqb -- I've started tinkering with this. It seems that regardless of whether we use `user.keyboard` or `user.type` for the enter key, it's creating two thoughts:

```
(before)
- __ROOT__

(after)
- __ROOT__
  - 
  -
```

I also gave `fireEevent.keyPress` a try, but it doesn't seem to create a new thought at all:

```
(before)
- __ROOT__

(after)
- __ROOT__
```

This has me a bit perplexed, so I will take a look at it tomorrow with a fresh perspective.

trevinhofmann:
Hi @snqb, I've been playing around with this issue without much success yet. Have you had a chance to look into it further?

snqb:
I am really doubtful I can do more than I did, because this thing almost got me into a mental breakdown

What I suggest is: maybe we just do redux-way for these kinds of cases, preserving the rest `userEvents` style, and make a new issue on investigating why we have such problems with creating thoughts. And also, it generally feels like something worth a separate issue.
Thus, I could just fix this and, I believe one more issue, and we're good to mark this complete.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,356 @@
...
+
+    const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
+
+    // eslint-disable-next-line no-console
+    console.debug('subthoughts', exported)
+  })
+})
+
+beforeEach(createTestAppWithTutorial)
+afterAll(cleanupTestApp)
```
trevinhofmann, line 103: Sorry it took forever to find, but these lines seem to be the issue.

I'm assuming these `beforeEach`/`afterAll` hooks were meant to be included within the `describe` blocks.

Having `beforeEach(createTestAppWithTutorial)` at the top level and `beforeEach(createTestApp)` within the `'A curious case of me not being able to grasp keyboard'` test at the top was triggering both before the test ran.

I had confirmed that the keyboard event handler was being called twice for every `await user.keyboard('{Enter}')` and spent quite a bit of effort looking into why React Testing Library was triggering it twice. Looking at it from another angle, I realized we were actually initializing the event handlers twice, and each handler was only triggered once.

The calls to `createTestApp` and `createTestAppWithTutorial` each call `initialize`->`initEvents`->`window.addEventListener('keydown', keyDown)`, and we end up with two handlers.

@snqb -- If you could please fix this test file to ensure that the before/after hooks are configured in the correct blocks, the user events seem to be handled correctly. To quickly verifyy, I simply removed the other tests/hooks altogether:

```
Before
- __ROOT__

Enter key was pressed!

After
- __ROOT__
  -
```

snqb: hey, sorry for late response -- I guess my body resisted opening this issue too much :D

but again, the events were failing me, no matter how I put the `createTestApp` and `cleanupTestApp`. In `before` and `after` `all` and `each` in `describes` and inside the individual cases.

so until I drowned in insanity again, I decided to see if the events handlers are being cleaned up alright:

and I've added a logger inside `initEvents`  to see how the `keydown` thingy is working inside
https://github.com/cybersemics/em/blob/128153531a41c49eb19daf7eb784a40f768a1020/src/util/initEvents.ts#L329
became
```tsx
  window.addEventListener('keydown', (e: KeyboardEvent) => {
    console.log('keydown', e.key)
    keyDown(e)
  })
```

and then I've made myself a small `describe` suite

```jsx
describe.only('Tutorial test keyboard events', () => {
  beforeEach(createTestApp) // we do these things
  afterEach(cleanupTestApp) // in almost every test file

  it('2 enters', async () => {
    await user.keyboard('{Enter}')
    await user.keyboard('{Enter}')
  })

  it('another 2 enters', async () => {
    await user.keyboard('{Enter}')
    await user.keyboard('{Enter}')
  })

  it('final 2 enters', async () => {
    await user.keyboard('{Enter}')
    await user.keyboard('{Enter}')
  })
})
```

and I expected to see 3 individual runs that rigger 3*2 = 6 keydown enters
```
keydown Enter
keydown Enter
.
.
keydown Enter
keydown Enter
.
.
keydown Enter
keydown Enter
```

but got(voila):
{0AF88215-177A-4876-A6EA-64B144DC4D95}


it feels to me, this is something that should not happen, or I don't get it :D

an initial inspection has shown that the `keyUp` and `keyDown` handlers do _exist_ when the `cleanup` is called, but event handlers are persisted across cleanups. 

I can dig deeper and see what's going on, but even if I should, it feels like something deserving a separate issue.

raineorshine: Hey @snqb, thanks for bringing this to our attention. Definitely looks like the event handlers are not being removed on cleanup. I would like you to try to identify the cause of this if you can. Feel free to address it in a separate PR or in this one if that’s easier, as it does not seem to be affecting other issues at the moment. 

Thanks for all your effort.

snqb: Ok, I guess I messed up and this behavior was because of me altering a function that is supposed to be cleaned up xdddddd
glad I didn't spend much time on that

So, my problem with this testing env is still in the fact that I don't want to call `cleanupTestApp` after each case, as I want to have consecutive runs to fully emulate the tutorial. And I feel like I've found the way: i guess I'll just call the `cleanup`  separately from all of those `store.dispatch(clear)` and `db.clear()` stuff, to turn off the event listeners, but preserve the state.

i'm so sorry for the disinformation above 🤣

src/components/Tutorial/Tutorial2StepContextViewOpen.tsx
```diff
@@ -46,7 +46,7 @@ const Tutorial2StepContextViewOpen = () => {
       {TUTORIAL_CONTEXT1_PARENT[tutorialChoice]}" or "{TUTORIAL_CONTEXT2_PARENT[tutorialChoice]}" to show it again.
     </p>
   ) : contextViewClosed ? (
-    <p>Oops, somehow the context view was closed. Select "Relationships".</p>
+    <p>Oops, somehow the context view was closed. Select "{TUTORIAL_CONTEXT[tutorialChoice]}".</p>
```
snqb, line 49: it was a bug I've found while testing

.vscode/settings.json
```diff
@@ -8,4 +8,7 @@
   "[typescriptreact]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
+  "[typescript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
```
snqb, line 13: files keep registering as `typescript`, not `typescriptreact`

is it fine we add `typescript` too?

src/components/Tutorial/TutorialNavigationButton.tsx
```diff
@@ -14,7 +14,6 @@ const TutorialNavigationButton = React.forwardRef<
 >(({ clickHandler, value, disabled = false, classes }, ref) => (
   <a
     className={cx(anchorButtonRecipe({ variableWidth: true, smallGapX: true }), classes)}
-    onClick={clickHandler}
```
snqb, line 17: it's duplicated by `fastClick`, which caused a skip of step I've painfully experienced multiple times while testing

src/test-helpers/createTestApp.tsx
```diff
@@ -80,4 +79,13 @@ export const refreshTestApp = async () => {
   await act(vi.runOnlyPendingTimersAsync)
 }
 
+/** Clear exisiting event listeners, but without clearing the app. */
+export const cleanupTestEventHandlers = async () => {
```
snqb, line 83: had to add this, as I really want not to clear the whole app between the test runs of the tutorial

.vscode/settings.json
raineorshine: Sure. I haven't run into any issues myself with the current configuration, but I don't think it can hurt.

src/components/Tutorial/TutorialNavigationButton.tsx
raineorshine: Yes, good catch. It doesn't make sense to have both.

snqb:
concerning the failing test:
it seems to be resolved if we manually `await createTestApp`
```tsx
beforeEach(async () => {
  await createTestApp()
})
```

so the problem is with this case:
```tsx
it('disable isLoading after initialize', async () => {
  expect(store.getState().isLoading).toBe(false)
})
```

`isLoading` keeps being `undefined`, while I haven't found a single use in our app where `undefined` is kinda possible via assignment, though it's possible judging from types in `tutorial` reducer.
```tsx
/** Sets the Tutorial setting value. */
const tutorial = (state: State, { value }: { value?: boolean }) => {
  return ({
    ...settings(state, {
      key: 'Tutorial',
      value: value ? 'On' : 'Off',
    }),
    // disable isLoading when dismissing the tutorial, since we can assume this is a new thoughtspace or the thoughtspace has already been loaded
    isLoading: state.isLoading && value // <--- here `value` is `value?`, so it could leak 
  })
}```
but I haven't found at a glance a realistic way this could happen
so I judged, it's a create test thing

btw it really puzzles me that explicitly `await`-ing does lead to a different result
I assumed it kinda handles promises, as it works with all other cases.
```tsx
beforeEach(async () => {
  await createTestApp()
})
```

raineorshine:
Vitest passes a test context object to the `beforeEach` callback when written in point-free style: `beforeEach(createTestApp)`. It was being ignored before, but now with the options object that was added, `tutorial` gets accidentally set to undefined. This gets passed down to `isLoading` through the `tutorial` action.

The point-free style is nice, so I would suggest just dispatching `false` explicitly when the tutorial should be closed:

```ts
    if (!tutorial) {
      store.dispatch({ type: 'tutorial', value: false })
    }
```

trevinhofmann:
@snqb -- Thank you again for the PR, as well as the latest updates. Please let me know when this is ready for review again, and I'll be sure to take a look right away.

snqb:
@trevinhofmann hey, I think you can have a look.
I'm pretty satisfied with the final code, if we set the goal to have a very straightforward user-events based walkthrough.

trevinhofmann (approved):
Thank you, @snqb. Nice work! The updated tests cover all of the steps for the full tutorial (Parts I & II) with direct user interaction rather than manipulating the store. I've walked through each of the tests manually with the tutorial in the live app, and I'm satisfied with the test coverage.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+// as per https://testing-library.com/docs/user-event/options/#advancetimers
+// we should avoid using { delay: null }, and use jest.advanceTimersByTime instead
+const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime })
+
+/** Get last created thought in the document, used mostly after `user.keyboard('{Enter}')`. */
+const lastThought = () => Array.from(document.querySelectorAll('[contenteditable="true"]')).at(-1)!
+
+/** Ensure we are at the given step of the tutorial. */
+const ensureStep = (step: number) => {
+  expect(store.getState().storageCache?.tutorialStep).toBe(step)
```
raineorshine, line 39: Using `storageCache` tightly couples the tests to an implementation detail of a caching mechanism. The tests should not have to be aware of the storageCache. It would be better to add a `data-testid` or `aria-label` for the tutorial step.

Another approach is to not make an assertion about the tutorial step at all, and instead just assert the text that the user sees. That has the benefit of being less dependent on the implementation, but is more likely to break if the tutorial text changes. What do you think?

Separately, avoid moving `expect` statements into helper methods. Test helpers can make queries or return values, but it's better to have the assertions front-and-center in the tests themselves.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+import exportContext from '../../../selectors/exportContext'
+import store from '../../../stores/app'
+import createTestApp, { cleanupTestApp, cleanupTestEventHandlers } from '../../../test-helpers/createTestApp'
+
+// as per https://testing-library.com/docs/user-event/options/#advancetimers
+// we should avoid using { delay: null }, and use jest.advanceTimersByTime instead
+const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime })
+
+/** Get last created thought in the document, used mostly after `user.keyboard('{Enter}')`. */
+const lastThought = () => Array.from(document.querySelectorAll('[contenteditable="true"]')).at(-1)!
```
raineorshine, line 35: Let's use `[aria-label="thought"] [contenteditable]` or `[data-editable="true"]` to be a bit more specific. There happen to be no other contenteditable's on the screen, but it's a bit safer to not assume that.

Even better would be to specifically select the empty thought.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+    expect(screen.getByText('Now type something. Anything will do.')).toBeInTheDocument()
+    await user.type(lastThought(), 'my first thought')
+    await user.keyboard('{Enter}')
+    await act(vi.runOnlyPendingTimersAsync)
+  })
+
+  it('step second thought - prompts to add another thought', async () => {
+    expect(screen.getByText('Well done!')).toBeInTheDocument()
+    expect(screen.getByText(/Try adding another thought/)).toBeInTheDocument()
+    ensureStep(TUTORIAL_STEP_SECONDTHOUGHT)
```
raineorshine, line 74: Actions should generally be performed in the test that depends on them. So instead of typing and hitting enter in  one test and asserting the result in the next, move the typing and hitting enter to the next test where it is asserted.

As a general rule, a test should end with an `expect`. Anything after that should probably belong to the next test.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+      await user.click(screen.getByText('Next'))
+      expect(screen.getByText(/First, let me show you how to create a new thought/)).toBeInTheDocument()
+    })
+  })
+
+  it('step first thought - how to create a thought ', async () => {
+    ensureStep(TUTORIAL_STEP_FIRSTTHOUGHT)
+    expect(screen.getByText('Hit the Enter key to create a new thought.')).toBeInTheDocument()
+    await user.keyboard('{Enter}')
+    expect(screen.getByText('You did it!')).toBeInTheDocument()
```
raineorshine, line 62: Test actions look good and are very readable 👍

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  it('step subthought - how to create a thought inside thought', async () => {
+    ensureStep(TUTORIAL_STEP_SUBTHOUGHT)
+    expect(exportContext(store.getState(), [HOME_TOKEN], 'text/plain')).toBe(
+      `- __ROOT__
+  - my first thought
+  - my second thought
+  - third thought
+  - `,
+    )
```
raineorshine, line 97: Good thinking. While we could make a smaller assertion, I think showing the full thoughtspace at this point makes the tests easier to understand.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+
+    // Now I am going to show you how to add a thought within another thought.
```
raineorshine, line 99: This comment is redundant and can safely be removed.

Likewise with similar redundant comments below.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+    expect(screen.getByText(/Now I am going to show you how to add a thought/)).toBeInTheDocument()
+
+    // since we have cursor on empty thought
+    expect(screen.getByText(/Hit the Delete key to delete the current blank thought/)).toBeInTheDocument()
+    expect(screen.getByText(/Then hold the Ctrl key and hit the Enter key/)).toBeInTheDocument()
```
raineorshine, line 104: Let's stick with a single `screen.getByText` for a given point in the tutorial. Testing the minimal unique text at the given step will give us effectively the same test coverage, but will be less maintenance if the tutorial text is changed in the future.

Likewise in similar occurrences below.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+      `- __ROOT__
+  - Home
+    - To Do
+      - or to not
+  - Work
+    - To Do`,
+    )
+    expect(screen.getByText('Very good!')).toBeInTheDocument()
+
+    expect(screen.getAllByRole('superscript')[0]).toHaveTextContent('2')
```
raineorshine, line 241: This is a good way to select the superscript 👍

raineorshine (changes requested):
Thanks, I think we have the right approach now!

src/components/Tutorial/__tests__/Tutorial.ts
trevinhofmann: > Using `storageCache` tightly couples the tests to an implementation detail of a caching mechanism. The tests should not have to be aware of the storageCache. It would be better to add a `data-testid` or `aria-label` for the tutorial step.
>
> Another approach is to not make an assertion about the tutorial step at all, and instead just assert the text that the user sees. That has the benefit of being less dependent on the implementation, but is more likely to break if the tutorial text changes. What do you think?

I agree with this, and would lean toward relying only on the page's text. It might be relevant for a unit test to specifically check the internal state of the storage cache, but these tutorial tests are more along the lines of functional tests. I would prefer the second option ("not make an assertion about the tutorial step at all, and instead just assert the text that the user sees").

For the same reason that we did not want to directly manipulate the internal state of the redux store, we shouldn't need to verify the internal state of the storage cache. It makes sense for the tests to act as a user would and verify what a user would see.

Breaking when the tutorial text changes is inevitable and, in my opinion, acceptable.

trevinhofmann: > Separately, avoid moving `expect` statements into helper methods. Test helpers can make queries or return values, but it's better to have the assertions front-and-center in the tests themselves.

I'm personally more on the fence about whether or not it is reasonable to have helper methods with `expect` statements, but I believe the common approach is not to do so. I think it would be reasonable to enforce that in **em**.

Ignoring the other comment about not using `storageCache` for a moment, this could reasonably be refactored directly within the tests to:

```
expect(getTutorialStep()).toBe(TUTORIAL_STEP_START)
```

Then, the helper method could simply be:

```
const getTutorialStep = () =>  store.getState().storageCache?.tutorialStep
```

src/components/Tutorial/__tests__/Tutorial.ts
trevinhofmann: > Even better would be to specifically select the empty thought.

I considered this during my previous review, but didn't mention it since the last thought and the empty thought were coincidentally always the same.

It would be nicer to explicitly select the thought to be edited (the empty thought), and easier to update in the future if we need to edit another thought during testing.

src/components/Tutorial/__tests__/Tutorial.ts
trevinhofmann: This goes against the "don't test internal state" concept we advocated for in [this comment](https://github.com/cybersemics/em/pull/2702#discussion_r1916722734). However, I agree that the value it provides to clearly see and test the entire thoughtspace is worth an exception.

If we wanted to keep this while not inspecting the internal state of the store, we could implement a helper function to retrieve the thoughts actually displayed in the DOM and verify those. I'm not sure how complex that helper function might be to implement, though.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  it('step context 2 - we create "Work" thought with children', async () => {
+    await user.click(screen.getByText('Next'))
+    ensureStep(TUTORIAL2_STEP_CONTEXT2_PARENT)
+    expect(screen.getByText(/Now we are going to create a different "To Do" list./)).toBeInTheDocument()
+
+    // we created a new thought on 3rd level, so we shift-tab our way back to root
+    await user.keyboard('{Enter}')
+    await user.keyboard('{Shift>}{Tab}{/Shift}')
+    await user.keyboard('{Shift>}{Tab}{/Shift}')
```
trevinhofmann, line 223: I don't mind this, but it might be more natural to select a thought at the root and create the new thought from there with one `enter` (and not use `shift+tab`).

Thoughts, @raineorshine?

raineorshine: Yes, I agree. Shift + Tab activates the Outdent command. It would be better to stick to simple navigation and New Thought commands for completing the tutorial.

src/test-helpers/createTestApp.tsx
```diff
@@ -13,10 +13,9 @@ import storage from '../util/storage'
 let cleanup: Await<ReturnType<typeof initialize>>['cleanup']
 
 /** Set up testing and mock document and window functions. */
-const createTestApp = async () => {
+const createTestApp = async ({ tutorial }: { tutorial?: boolean } = {}) => {
```
trevinhofmann, line 16:
```suggestion
const createTestApp = async ({ tutorial = false }) => {
```

See: https://github.com/cybersemics/em/pull/2702/files#r1919837703

src/components/Tutorial/__tests__/Tutorial.ts
raineorshine: Yes, that would be a better approach in general. However, I think having to parse the hierarchy out of the flat rendered TreeNodes would be unnecessarily complex in this case.

src/test-helpers/createTestApp.tsx
```diff
@@ -33,8 +32,8 @@ const createTestApp = async () => {
     )
 
     store.dispatch([
-      // skip tutorial
-      { type: 'tutorial', value: false },
+      // there are cases where we want to show tutorial on test runs, whilst mostly we don't
+      { type: 'tutorial', value: !!tutorial },
```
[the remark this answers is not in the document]
raineorshine: FYI This is a small concession to allow point-free style in `beforeEach`. 

See: https://github.com/cybersemics/em/pull/2702#issuecomment-2585880527

src/components/Tutorial/__tests__/Tutorial.ts
snqb: So, I arrived at 
```tsx
/** Get last created empty thought in the document. Gets the last thought that is not empty. Mostly used after `user.keyboard('{Enter}')` to get the new thought. */
const lastThought = () =>
  Array.from(document.querySelectorAll('[data-editable="true"]'))
    .filter(it => !it.textContent)
    .at(-1)!
```

I think it works. Perhaps, a more refined solution could be out there, but it works :D
I'd prefer to modify this thing and move out to `test-helpers` at some point if we hit the wall, but not now.

src/components/Tutorial/__tests__/Tutorial.ts
snqb: I thought about creating a helper to give us a diff-able thought tree, but it felt a step too far for me, since I really just wanna "document" the hierarchy, so the test suite is easier to follow for the developer.
So I feel like, this is an okay-ish exception :D

src/components/Tutorial/__tests__/Tutorial.ts
snqb: totally makes sense, thanks
now it's
```jsx
await user.click(screen.getByText('Home'))
await user.keyboard('{Enter}')
```

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,242 @@
...
+import exportContext from '../../../selectors/exportContext'
+import store from '../../../stores/app'
+import createTestApp, { cleanupTestApp, cleanupTestEventHandlers } from '../../../test-helpers/createTestApp'
+
+// as per https://testing-library.com/docs/user-event/options/#advancetimers
+// we should avoid using { delay: null }, and use jest.advanceTimersByTime instead
+const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime })
+
+/** Get last created empty thought in the document. Gets the last thought that is not empty. Mostly used after `user.keyboard('{Enter}')` to get the new thought. */
+const lastThought = () =>
```
raineorshine, line 14: Let's call this `lastEmptyThought` to better reflect the new functionality.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,242 @@
...
+import { HOME_TOKEN } from '../../../constants'
+import exportContext from '../../../selectors/exportContext'
+import store from '../../../stores/app'
+import createTestApp, { cleanupTestApp, cleanupTestEventHandlers } from '../../../test-helpers/createTestApp'
+
+// as per https://testing-library.com/docs/user-event/options/#advancetimers
+// we should avoid using { delay: null }, and use jest.advanceTimersByTime instead
+const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime })
+
+/** Get last created empty thought in the document. Gets the last thought that is not empty. Mostly used after `user.keyboard('{Enter}')` to get the new thought. */
```
raineorshine, line 13: > Gets the last thought that is not empty.

Maybe a typo on this part?

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,242 @@
...
+    expect(screen.getByText(/Try adding another thought/)).toBeInTheDocument()
+    await user.type(lastThought(), 'my second thought')
+    await user.keyboard('{Enter}')
+    await act(vi.runOnlyPendingTimersAsync)
+
+    expect(screen.getByText('Good work!')).toBeInTheDocument()
+    expect(screen.getByText('Now type some text for the new thought.')).toBeInTheDocument()
+    await user.type(lastThought(), 'third thought')
+    await user.keyboard('{Enter}')
+    await act(vi.runOnlyPendingTimersAsync)
```
raineorshine, line 56: There are still a few places where tests do not end in `expect`.

See: https://github.com/cybersemics/em/pull/2702#discussion_r1916750269

src/components/Tutorial/__tests__/Tutorial.ts
raineorshine: @snqb Still needs attention

raineorshine (changes requested):

snqb:
ok, I've rethought the thing and spent some time
but now almost everywhere we have one-expect-per suite
and it's kinda cleaner, pls have a look
I think it's all resolved

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    it('we can exit tutorial mode', async () => {
+      await user.click(screen.getByText('Play on my own'))
+      await act(vi.runOnlyPendingTimersAsync)
+      expect(() => screen.getByTestId('tutorial-step')).toThrow('Unable to find an element')
+    })
+
+    it('we can continue to advanced tutorial', async () => {
+      expect(screen.getByText('Learn more')).toBeInTheDocument()
+    })
```
raineorshine, line 116: There's a contradiction here: If the user just exited the tutorial, then how could the "Learn more" button be visible?

This seems to rely on the tutorial inadvertently being re-launched on `beforeEach`, which does not reflect real-world usage. We should make this more explicit, as it's not super obvious what's happening here, and closer to end-user behavior. I recommend one of the following options:

1. Don't click "Play on my own" at all. Then the test can click "Learn more" just as the user would. (less test coverage)
2. Click "Play on my own". Then navigate to the Help and start the second Tutorial as the user would. (more work, but more test coverage)

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,260 @@
...
+  })
+})
+
+describe('Tutorial 2', async () => {
+  beforeEach(() => createTestApp({ tutorial: true }))
+  afterEach(cleanupTestEventHandlers)
+  afterAll(cleanupTestApp)
+  it('we learn about context indicators', async () => {
```
raineorshine, line 124: It's more accurate to say "contexts" or "context view" than "context indicators". Let's update this to be more consistent with **em** terminology.

Likewise below.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,260 @@
...
+
+  describe('step context 1 - create a "Home" to-do list', () => {
+    it('we create a Home thought', async () => {
+      await user.keyboard('{Enter}')
+      await user.type(lastEmptyThought(), 'Home')
+      await act(vi.runOnlyPendingTimersAsync)
+      expect(screen.getByText(/Add a thought with the text "To Do"/)).toBeInTheDocument()
+    })
+
+    it('we add a Home->To Do subthought', async () => {
```
raineorshine, line 151: Please use `Home/To Do` or `Home • To Do` to indicate subthought relationships. That will be a bit more consistent with our internal documentation, since `->` is not used.

Likewise below.

raineorshine (changes requested):
Thank you! I like the new structure. I think one `expect` at the end of each test is very clean.

src/components/Tutorial/__tests__/Tutorial.ts
```diff
@@ -0,0 +1,260 @@
...
+const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime })
+
+/** Gets the last empty thought in the document. Mostly used after `user.keyboard('{Enter}')` to get the new thought. */
+const lastEmptyThought = () =>
+  Array.from(document.querySelectorAll('[data-editable="true"]'))
+    .filter(it => !it.textContent)
+    .at(-1)!
+
+describe('Tutorial 1', async () => {
+  beforeEach(() => createTestApp({ tutorial: true }))
```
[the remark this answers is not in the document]
snqb: I really have not much idea how to make it work with one `beforeAll`, since if I don't create a new test app for each case, I'm left with empty-ish DOM
I think, the whole setup relies on re-creating an isolated testing env for each case, as it's done on every other test suite.

right now it's run with one db/store/state/whatever, but with cleaning up event listeners that keep attaching
but it's not one sequential run per se, though I'd argue it's a lot like reloading page at each step :D

I'm really not sure how to make it work with one `beforeAll`, really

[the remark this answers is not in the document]
snqb: > What do you mean by "empty-ish DOM"?
```
html
<body
  data-browser="safari"
  data-color-mode="dark"
  data-device="desktop"
  data-drag-in-progress="false"
  data-native="false"
  data-platform="other"
/>
```

> Are you saying that the current implementation recreates an isolated testing environment for each case? Because my understanding is that these tests depend on the current test step being preserved between tests.

we have everything in db, store etc., but dom is empty if we don't call the `createTestApp` on each test case
so the current approach, and me attaching and detaching events, is:
1) we `createTestApp`, get correct dom state, built from store/db.
2) we run tests
3) we remove event listeners created

Just runing a `createTestApp` once does not work, and it's a problem of our general testing environment.
It's easy to demonstrate if we try to do this:
1. we `createTestApp`
2. we run one test, where we create a thought "first"
3. we should have "first" thought present in our dom

but it does not work like this at all.
the test suite below gets 2/2 green, but it should not.
```tsx
describe.only('T_T', () => {
  beforeAll(createTestApp)

  it('we create "first" thought', async () => {
    await user.keyboard('{Enter}')
    await user.type(lastEmptyThought(), 'first')
    await user.keyboard('{Enter}')

    expect(exportContext(store.getState(), HOME_TOKEN)).toEqual(`<ul>
  <li>__ROOT__  
    <ul>
      <li>first</li>
      <li></li>
    </ul>
  </li>
</ul>`)
    expect(screen.getByText('first')).toBeInTheDocument()
  })

  it('we expect to see "first" thought', () => {
    console.log(exportContext(store.getState(), HOME_TOKEN))

    screen.debug(document.querySelector('.content')!) /* <-- prints this
    <body
      data-browser="safari"
      data-color-mode="dark"
      data-device="desktop"
      data-drag-in-progress="false"
      data-native="false"
      data-platform="other"
    />
    */

    expect(() => screen.getByText('first')).toThrow()
  })
})
```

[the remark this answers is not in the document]
snqb: maybe we should "fix" this, but will it break other tests?

i generally feel like this pr now achieves what we want it to, given the current testing environment
maybe we should create another issue for it?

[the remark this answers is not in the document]
raineorshine: Thanks. I'd love to figure out why the DOM is empty. Without knowing the answer to that, it's hard to know what the right call is here. 

However, since all the tests pass and they are quite readable, I'm okay with merging the current implementation so that we have a good checkpoint in the commit history. We can investigate further in a separate issue. 

If you could address the small stylistic changes below, then I'd be happy to merge. Thanks!

src/components/Tutorial/__tests__/Tutorial.ts
snqb: since I can do it, given the circumstances, I left it as is
at least it tests something that can break easily
and if we'll change our setup to have correct dom each time, this case will fail right away

snqb:
I've updated the stylistics

raineorshine (approved):

Issue #2676, opened 2024-12-04
prevThought


raineorshine:
## Steps to Reproduce

```
- a
  - =pin
  - b
- c
```

1. Set the cursor on `c`
2. ArrowUp

## Current Behavior

The cursor moves to the previous sibling, `a`.

## Expected Behavior

The cursor should move to the previous visible thought, `b`.

ArrowUp (`cursorUp`) should move the cursor to the previous visible thought regardless of the structure of thoughts, sort preferences, which thoughts are expanded, or open context views.

i.e. We need a `prevThought` that is analogous to https://github.com/cybersemics/em/blob/main/src/selectors/nextThought.ts.

## Testing

Add tests to https://github.com/cybersemics/em/blob/be48a7235fae311089585bb6c23413d945ea8d05/src/actions/__tests__/cursorUp.ts.

PR #2703, opened 2024-12-16
implement prevThought

RED-ROSE515:
This PR implements the feature : prevThought #2676 

I implemented `prevThought.ts` to get the prevThought of current thought.
In this function, I need to check if the prevThought is expanded or not.
I can see that there's already a function called `expandThoughts`. 
So, I use this function to check if the thought is expanded or not.

1. The `prevThought` function first gets the previous sibling using prevSibling.
2. If there is a previous sibling, it uses lastVisibleDescendant which internally.
3. Checks if the thought is expanded using isExpanded. 
    - If expanded, it returns the last visible child recursively.
    - If not expanded, it returns the thought itself
4. If no previous sibling exists and not at root level, return the parent.

It's the main logic of `prevThought` function.

And I added tests for this in cursorUp.

### DIFF
--- a/src/actions/cursorUp.ts
+++ b/src/actions/cursorUp.ts
@@ -1,36 +1,31 @@
 import State from '../@types/State'
 import Thunk from '../@types/Thunk'
 import setCursor from '../actions/setCursor'
-import { HOME_PATH, HOME_TOKEN } from '../constants'
+import { HOME_TOKEN } from '../constants'
 import { getChildrenSorted } from '../selectors/getChildren'
-import prevSibling from '../selectors/prevSibling'
-import rootedParentOf from '../selectors/rootedParentOf'
-import appendToPath from '../util/appendToPath'
-import isRoot from '../util/isRoot'
-import parentOf from '../util/parentOf'
+import prevThought from '../selectors/prevThought'
 
-/** Moves the cursor to the previous sibling. Works in normal or context view. If there is no cursor, sets the cursor on the last thought of in the home context. */
+/** Moves the cursor to the previous visible thought in visual order. If there is no cursor, sets the cursor on the last thought in the home context. */
 const cursorUp = (state: State, { preserveMulticursor }: { preserveMulticursor?: boolean } = {}) => {
   const { cursor } = state
-  const path = cursor || HOME_PATH
-  const pathParent = rootedParentOf(state, path)
 
-  const prevThought = cursor
-    ? // if cursor exists, get the previous sibling
-      prevSibling(state, cursor)
+  const path = cursor
+    ? // if cursor exists, get the previous thought in visual order
+      prevThought(state, cursor)
     : // otherwise, get the last thought in the home context
-      getChildrenSorted(state, HOME_TOKEN).slice(-1)[0]
-
-  const prevPath = prevThought
-    ? // non-first child path
-      appendToPath(parentOf(path), prevThought.id)
-    : // when the cursor is on the first child in a context, move up a level
-      !isRoot(pathParent)
-      ? pathParent
+      getChildrenSorted(state, HOME_TOKEN).slice(-1)[0]?.id
+      ? [getChildrenSorted(state, HOME_TOKEN).slice(-1)[0].id]
       : null
 
   // noop if there is no previous path, i.e. the cursor is on the very first thought
-  return prevPath ? setCursor(state, { path: prevPath, preserveMulticursor }) : state
+  return path && path.length > 0
+    ? setCursor(state, {
+        path: [path[0], ...path.slice(1)],
+        cursorHistoryClear: true,
+        editing: true,
+        preserveMulticursor,
+      })
+    : state
 }
 
 /** Action-creator for cursorUp. */
--- a/src/actions/cursorUp.ts
+++ b/src/actions/cursorUp.ts
@@ -1,3 +1,4 @@
+import Path from '../@types/Path'
 import State from '../@types/State'
 import Thunk from '../@types/Thunk'
 import setCursor from '../actions/setCursor'
@@ -13,16 +14,14 @@ const cursorUp = (state: State, { preserveMulticursor }: { preserveMulticursor?:
     ? // if cursor exists, get the previous thought in visual order
       prevThought(state, cursor)
     : // otherwise, get the last thought in the home context
-      getChildrenSorted(state, HOME_TOKEN).slice(-1)[0]?.id
-      ? [getChildrenSorted(state, HOME_TOKEN).slice(-1)[0].id]
+      getChildrenSorted(state, HOME_TOKEN).at(-1)?.id
+      ? ([getChildrenSorted(state, HOME_TOKEN).at(-1)!.id] as unknown as Path)
       : null
 
   // noop if there is no previous path, i.e. the cursor is on the very first thought
   return path && path.length > 0
     ? setCursor(state, {
-        path: [path[0], ...path.slice(1)],
-        cursorHistoryClear: true,
-        editing: true,
+        path: path,
         preserveMulticursor,
       })
     : state
--- a/src/actions/cursorUp.ts
+++ b/src/actions/cursorUp.ts
@@ -15,7 +15,7 @@ const cursorUp = (state: State, { preserveMulticursor }: { preserveMulticursor?:
       prevThought(state, cursor)
     : // otherwise, get the last thought in the home context
       getChildrenSorted(state, HOME_TOKEN).at(-1)?.id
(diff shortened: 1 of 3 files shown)


src/selectors/prevThought.ts
```diff
@@ -0,0 +1,76 @@
...
+const lastVisibleDescendant = (state: State, path: Path): Path => {
+  const id = head(path)
+
+  // Only traverse children if the thought or its parent is expanded
+  if (!isExpanded(state, path)) return path
+
+  // When showHiddenThoughts is true, show all children, otherwise filter hidden ones
+  const children = state.showHiddenThoughts
+    ? getAllChildrenAsThoughts(state, id)
+    : getAllChildrenAsThoughts(state, id).filter(child => isVisible(state, child))
```
raineorshine, line 29: You can use `getChildren`, which gets only visible thoughts by default.

```diff
diff --git a/src/selectors/prevThought.ts b/src/selectors/prevThought.ts
index 5cc7d01a73..7a9f1fed29 100644
--- a/src/selectors/prevThought.ts
+++ b/src/selectors/prevThought.ts
@@ -1,7 +1,7 @@
 import Path from '../@types/Path'
 import State from '../@types/State'
 import expandThoughts from '../selectors/expandThoughts'
-import { getAllChildrenAsThoughts, isVisible } from '../selectors/getChildren'
+import getChildren, { isVisible } from '../selectors/getChildren'
 import prevSibling from '../selectors/prevSibling'
 import rootedParentOf from '../selectors/rootedParentOf'
 import appendToPath from '../util/appendToPath'
@@ -24,9 +24,7 @@ const lastVisibleDescendant = (state: State, path: Path): Path => {
   if (!isExpanded(state, path)) return path
 
   // When showHiddenThoughts is true, show all children, otherwise filter hidden ones
-  const children = state.showHiddenThoughts
-    ? getAllChildrenAsThoughts(state, id)
-    : getAllChildrenAsThoughts(state, id).filter(child => isVisible(state, child))
+  const children = getChildren(state, id)
 
   // If no children, return current path
   if (children.length === 0) return path
```

src/selectors/prevThought.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // If no children, return current path
+  if (children.length === 0) return path
+
+  const lastChild = children[children.length - 1]
+  if (!lastChild) return path
```
raineorshine, line 35: It's not too significant,  but you can remove this check since `lastChild` is guaranteed to exist. You already checked `children.length` above.

src/selectors/prevThought.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  const childPath = appendToPath(path, lastChild.id)
+  return lastVisibleDescendant(state, childPath)
+}
+
+/** Gets the previous visible sibling, recursively skipping hidden thoughts if needed. */
+const getPreviousVisibleSibling = (state: State, path: Path, currentSibling: Path | null = null): Path | null => {
```
raineorshine, line 42: It might just be a naming issue, but I'm having a little trouble understanding this function. It says in the name and JSDOC that it gets the previous sibling, but it calls `lastVisibleDescendant` internally, which clearly is not limited to siblings.

Can you explain whether this is indeed supposed to only select siblings (i.e. thoughts with the same parent), or if it has a different purpose? If it has a different purpose, we may need to come up with a better name.

src/selectors/prevThought.ts
```diff
@@ -0,0 +1,76 @@
...
+    const prevSiblingPath = appendToPath(parentOf(path), prevSiblingThought.id)
+    return lastVisibleDescendant(state, prevSiblingPath)
+  }
+
+  const prevPath = appendToPath(parentOf(path), prevSiblingThought.id)
+  return getPreviousVisibleSibling(state, path, prevPath)
+}
+
+/** Gets the previous thought in visual order. */
+const prevThought = (state: State, path?: Path): Path | null => {
```
raineorshine, line 60: This function appears to be simplifiable to:

```ts
/** Gets the previous thought in visual order. */
const prevThought = (state: State, path?: Path): Path | null =>
  path ? getPreviousVisibleSibling(state, path) || rootedParentOf(state, path) : null
```

Does that work for you, or am I missing something?

raineorshine (changes requested):
This looks very good. Works as expected, and handles pinned thoughts, hidden thoughts, and context view.

I just found one case involving the context view that does not work as expected. Here is the test case for it:

```ts
  it('move cursor from normal view into context view child', () => {
    const text = `
      - =children
        - =pin
      - a
        - m
          - x
      - b
        - m
          - y
    `

    const steps = [importText({ text }), setCursor(['a', 'm']), toggleContextView, setCursor(['b']), cursorUp]

    expect(pathToContext(stateNew, stateNew.cursor!)).toEqual(['a', 'm', 'b'])
  })
```

Let me know if you need any additional assistance understanding the context view. There is some documentation here: https://github.com/cybersemics/em/wiki/Docs#332-context-view

Thank you!

src/actions/cursorUp.ts
```diff
@@ -1,36 +1,31 @@
...
+      ? [getChildrenSorted(state, HOME_TOKEN).slice(-1)[0].id]
       : null
 
   // noop if there is no previous path, i.e. the cursor is on the very first thought
-  return prevPath ? setCursor(state, { path: prevPath, preserveMulticursor }) : state
+  return path && path.length > 0
+    ? setCursor(state, {
+        path: [path[0], ...path.slice(1)],
+        cursorHistoryClear: true,
+        editing: true,
```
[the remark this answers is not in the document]
RED-ROSE515: Currently, no need to pass those values.

src/selectors/prevThought.ts
RED-ROSE515: - A
  - A1
    - A1a
- B (current thought)

When called on B, it will return A1a, not A.

So, I changed like this : getLastVisibleDescendantOfPreviousSibling

src/selectors/prevThought.ts
```diff
@@ -0,0 +1,124 @@
...
+  if (!path) return null
+
+  const pathParent = rootedParentOf(state, path)
+  const isRootLevel = isRoot(pathParent)
+
+  // Check if we're in a context view
+  const inContextView = isContextViewActive(state, pathParent)
+  const hasActiveContextViews = Object.keys(state.contextViews).length > 0
```
raineorshine, line 91: This checks if *any* context view is active, even if it is far away from the path. You would need to devise a more specific condition.

src/selectors/prevThought.ts
```diff
@@ -0,0 +1,124 @@
...
+  if (prevSiblingPath) return prevSiblingPath
+
+  // If at root level, stop here - don't traverse up to parent
+  if (isRootLevel) return null
+
+  // If not at root level and no previous sibling, check if parent is in context view
+  if (hasActiveContextViews && path.length === 1) {
+    const parentContextViewRoot = findContextViewRoot(state, pathParent)
+    if (parentContextViewRoot) return lastContext(state, parentContextViewRoot)
+  }
```
raineorshine, line 118: This block is identical to one just a few lines above. There is probably a way to rearrange this code to eliminate the redundancy.

src/selectors/prevThought.ts
```diff
@@ -0,0 +1,124 @@
...
+  if (!state.expanded[hashPath(appendToPath(path, lastContext.parentId))])
+    return appendToPath(path, lastContext.parentId)
+  return lastVisibleDescendant(state, appendToPath(path, lastContext.parentId))
+}
+
+/**
+ * Gets the last visible descendant of the previous sibling.
+ * If the previous sibling is hidden, recursively tries the next previous sibling.
+ */
+const getLastVisibleDescendantOfPreviousSibling = (
```
raineorshine, line 56: It would probably be cleaner to separate this function into two:

1. Get the previous visible sibling
2. Get the last visible descendant of the previous sibling

raineorshine (approved):
Thanks for making the requested changes! That helps a lot.

There is a problem with the context-view related functionality. `isInContextView` and `hasContextView` are both based on checking if a context view is active *anywhere* in the thoughtspace. This happens to work in the simple test case we defined, but is not going to work in all cases. We need a more specific condition to know if the context view is active in the particular path where it matters.

I have assessed the context view test case and determined that there is a high risk that we either won't find a solution that effectively covers all cases, or it will increase the complexity too much. Therefore I've removed the test case and related functionality.

The other major change I made is to replace `getLastVisibleDescendantOfPreviousSibling` with a simpler process of selecting the previous visible sibling and then selecting its last visible descendant. Let me know if you anticipate any issues with this, but the tests all pass.

Issue #2677, opened 2024-12-04
Collapse selected thoughts during multiselect


raineorshine:
## Steps to Reproduce

```
- a
  - b
  - c
    - d
```

1. Set the cursor on `b`.
2. Shift + ArrowDown. `b` and `c` are now selected.

## Current Behavior

`c` is expanded.

## Expected Behavior

`c` should be collapsed.

Selected thoughts should be collapsed during multiselect.

## Testing

Add tests to https://github.com/cybersemics/em/blob/main/src/selectors/__tests__/expandThoughts.ts

raineorshine:
I've given this some additional thought, and have decided that the real problem is the way the cursor changes on Shift + ArrowDown. I think a better solution will be to keep the cursor where it is, and have multiselect work independently. This will allow descendants to be selected, which could be useful.

Thanks for your help.

PR #2704, opened 2024-12-16
Collapse selected thoughts during multiselect

RED-ROSE515:
The PR implements this feature : Collapse selected thoughts during multiselect #2677

In this PR, I implement the feature in this logic.

1. First, change the order of `addMulticursor` and `setCursor`. In `setCursor`, `expandThoughts` will be implemented. So, before executing the `expandThoughts`, addMulticursor.
2. In `expandThoughts` function, check if the current path is in multicursor.
    If yes, do not expand the thought.
3. And I added the test to cover this feature.

### DIFF
--- a/src/actions/addMulticursor.ts
+++ b/src/actions/addMulticursor.ts
@@ -23,7 +23,10 @@ const addMulticursor = (state: State, { path, ignoreCursor }: { path: Path; igno
         // on desktop, add the cursor to the multicursor set if it's empty
         ...(isEmpty && !ignoreCursor && state.cursor && !isTouch ? { [hashPath(state.cursor)]: state.cursor } : {}),
       },
-      // update expanded state to include both cursor and new multicursor path
+    }),
+    // update expanded state after multicursors have been updated
+    state => ({
+      ...state,
       expanded: expandThoughts(state, path),
     }),
     // on desktop, set the cursor to the new multicursor if none exists
--- a/src/actions/addMulticursor.ts
+++ b/src/actions/addMulticursor.ts
@@ -3,11 +3,12 @@ import Path from '../@types/Path'
 import State from '../@types/State'
 import Thunk from '../@types/Thunk'
 import { isTouch } from '../browser'
+import expandThoughts from '../selectors/expandThoughts'
 import hashPath from '../util/hashPath'
 import reducerFlow from '../util/reducerFlow'
 import setCursor from './setCursor'
 
-/** Adds a cursor to the multicursor set. */
+/** Adds a cursor to the multicursor set and updates expanded state. */
 const addMulticursor = (state: State, { path, ignoreCursor }: { path: Path; ignoreCursor?: boolean }): State => {
   const isEmpty = !Object.keys(state.multicursors).length
 
@@ -22,6 +23,8 @@ const addMulticursor = (state: State, { path, ignoreCursor }: { path: Path; igno
         // on desktop, add the cursor to the multicursor set if it's empty
         ...(isEmpty && !ignoreCursor && state.cursor && !isTouch ? { [hashPath(state.cursor)]: state.cursor } : {}),
       },
+      // update expanded state to include both cursor and new multicursor path
+      expanded: expandThoughts(state, path),
     }),
     // on desktop, set the cursor to the new multicursor if none exists
     !state.cursor && !isTouch ? setCursor({ path, preserveMulticursor: true }) : null,
--- a/src/actions/clearMulticursors.ts
+++ b/src/actions/clearMulticursors.ts
@@ -1,47 +1,18 @@
 import State from '../@types/State'
 import Thunk from '../@types/Thunk'
 import expandThoughts from '../selectors/expandThoughts'
-import { getAllChildrenAsThoughts } from '../selectors/getChildren'
-import getThoughtById from '../selectors/getThoughtById'
-import head from '../util/head'
 
-/** Clears all multicursors and updates expanded state based on current cursor and previously selected thoughts. */
+/** Clears all multicursors and updates expanded state based on current cursor. */
 const clearMulticursors = (state: State): State => {
-  // Get all previously selected paths
-  const selectedPaths = Object.values(state.multicursors)
-
-  // Create a new state without multicursors
   const stateWithoutMulticursors = {
     ...state,
     multicursors: {},
   }
 
-  // Expand based on cursor first
-  let newExpanded = expandThoughts(stateWithoutMulticursors, state.cursor)
-
-  // Then expand each previously selected path
-  selectedPaths.forEach(path => {
-    // Get the thought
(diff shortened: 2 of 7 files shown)


raineorshine (changes requested):
Thanks for your submission!

I'm seeing an issue where the expanded thoughts are not reset properly after the multicursor is exited. There are different ways to trigger it, but here's one:

## Steps to Reproduce

(Same steps as #2677 + Escape.)

```
- a
  - b
  - c
    - d
```

1. Set the cursor on `b`.
2. Shift + ArrowDown. `b` and `c` are now selected.
3. Escape

## Current

Screenshot 2024-12-16 at 9 22 08 PM

## Expected
Screenshot 2024-12-16 at 9 22 14 PM

## Notes

Since changing the multicursor now affects the expansion state, you will need to update `state.expanded` in [addMulticursor](https://github.com/cybersemics/em/blob/main/src/actions/addMulticursor.ts), [removeMulticursor](https://github.com/cybersemics/em/blob/main/src/actions/removeMulticursor.ts), and [clearMulticursors](https://github.com/cybersemics/em/blob/main/src/actions/clearMulticursors.ts).

```ts
  expanded: expandThoughts(state, state.cursor)
```

Once you add that then maybe you won't need to change the order of `setCursor` in the cursorUp/Down commands.

RED-ROSE515:
**The key changes I made to fix the expansion state:** 

##When clearing multicursors, we now:
  First get all previously selected paths from state.multicursors.
  Create a temporary state without multicursors to avoid the multiselection expansion block.
  Expand based on the current cursor first.
  Then expand each previously selected path that has children.
  Merge all the expanded states together.

  Use a clean state without multicursors to avoid the multiselection block.

src/actions/addMulticursor.ts
```diff
@@ -22,6 +23,8 @@ const addMulticursor = (state: State, { path, ignoreCursor }: { path: Path; igno
         // on desktop, add the cursor to the multicursor set if it's empty
         ...(isEmpty && !ignoreCursor && state.cursor && !isTouch ? { [hashPath(state.cursor)]: state.cursor } : {}),
       },
+      // update expanded state to include both cursor and new multicursor path
+      expanded: expandThoughts(state, path),
```
raineorshine, line 27: `state` is the old state before the new multicursor is added. You need to use the most up-to-date state. I recommend adding a new reducer to the flow after `multicursors` is updated.

src/actions/removeMulticursor.ts
```diff
@@ -2,15 +2,18 @@ import _ from 'lodash'
...
-/** Removes a cursor from the multicursor set. */
+/** Removes a cursor from the multicursor set and updates expanded state. */
 const removeMulticursor = (state: State, { path }: { path: Path }): State => {
   const { [hashPath(path)]: _, ...remainingMulticursors } = state.multicursors
 
   return {
     ...state,
     multicursors: remainingMulticursors,
+    // Update expanded state based on remaining cursor or multicursors
+    expanded: expandThoughts(state, state.cursor),
```
raineorshine, line 16: `state` is the old state before the multicursor is removed. You need to pass it an up-to-date state.

src/commands/cursorDown.tsx
```diff
@@ -91,7 +91,6 @@ const cursorDownShortcut: Command = {
       const isNextPathMulticursor = nextPath && isMulticursorPath(state, nextPath)
 
       dispatch([
-        setCursor({ path: nextPath, preserveMulticursor: true }),
```
raineorshine, line 94: It doesn't seem correct to change `cursorDown` when the `cursorDown` behavior is not changing. Just the expansion behavior is changing.

src/actions/clearMulticursors.ts
```diff
@@ -1,11 +1,47 @@
...
+  const selectedPaths = Object.values(state.multicursors)
+
+  // Create a new state without multicursors
+  const stateWithoutMulticursors = {
+    ...state,
+    multicursors: {},
+  }
+
+  // Expand based on cursor first
+  let newExpanded = expandThoughts(stateWithoutMulticursors, state.cursor)
```
raineorshine, line 20: Can you explain why `newExpanded` is not enough on its own?

`expandThoughts(state, state.cursor)` should generally create the entire expansion state from the given state. Maybe I'm missing something though.

raineorshine (changes requested):

src/selectors/__tests__/expandThoughts.ts
```diff
@@ -823,3 +824,27 @@ describe('=done', () => {
     expect(isContextExpanded(stateNew, ['a', 'd'])).toBeFalsy()
   })
 })
+
+describe('multicursor', () => {
```
raineorshine, line 828: Let's also add a test for the clearMulticursors case.

src/actions/clearMulticursors.ts
RED-ROSE515: The key is that this function is handling the cleanup of multicursors while preserving the expanded state of previously selected paths. 

Here's why both parts are needed.
`expandThoughts`(stateWithoutMulticursors, state.cursor) expands the path to the current cursor location.
However, the user might have previously expanded other paths while they had multiple selections (multicursors). 
If we only kept the cursor's expansion, those other expanded sections would collapse when clearing multicursors, which would be jarring for the user.

raineorshine: Thanks. Can you provide a test or steps to reproduce? I'm open to the idea, but since it was not something originally proposed I will have to see what it's like with and without this additional change.

My concern is that this custom logic can cause `state.expanded` to be different even with the same state, depending on whether `clearMulticursors` was called or not. `state.expanded` should be completely deterministic so that the expansion state is a pure function of the cursor, multicursor, and surrounding thoughts.

raineorshine: Thanks. However, I don't think it would be possible to expand `c` in this case, since multicursor thoughts are always collapsed.

src/selectors/__tests__/expandThoughts.ts
RED-ROSE515: I think we need to add puppeteer test for expandThoughts. (Because we need to add keydown event)

raineorshine: A puppeteer test is a little heavy for this, but you could add an RTL test with a test Redux store.

RED-ROSE515: Ok, I will check about that.

raineorshine: e.g. https://github.com/cybersemics/em/blob/0d71558434e9a7dbe09b9396c440084950a87045/src/commands/__tests__/cursorNext.ts

raineorshine:
Closing per https://github.com/cybersemics/em/issues/2677#issuecomment-2557847765

Issue #2303, opened 2024-08-20
Upgrade react-signature-pad-wrapper


raineorshine:
Latest is broken on mobile Safari.

https://github.com/szimek/signature_pad/releases/tag/v5.0.0

```
undefined is not an object (evaluating signaturePad._handlePonterStart.bind')
```

Note that we add a shadow to the SignaturePad canvas context by accessing a private member on the signaturePad class. It's not ideal, but it always us to add the necessary styling to the gesture trace. If that private member has changed in the new major version, it could be causing it to break.

snqb:
this happens because of https://github.com/szimek/signature_pad/commit/29a8b5a2a19387782c3eec1055389011eb7f9699 which removed the `_handlePointerStart` thingy
And changing it to `_handlePointerDown` in our code breaks gestures a bit, so still researching

snqb:
hey, provided a draft that works fine if we switch to touches
I'll gladly research why `pointer` events are not working right, but decided to show that it works fine with `touch` events if we fancy that approach

raineorshine:
Makes sense, let's give touch events a try.

PR #2706, opened 2024-12-17
Update gesture pad & use touch events

snqb:
fixes https://github.com/cybersemics/em/issues/2303

Long story short: they had  `handlePointerStart` and `handlePointerEnd` functions that did some custom stuff, differing from very standard `pointerdown` and `pointerup` events. So they dumped that, and made two new functions: `handlePointerDown` and `handlePointerUp`, which led to many unexpected problems if we use them. 
https://github.com/szimek/signature_pad/commit/29a8b5a2a19387782c3eec1055389011eb7f9699

After spending considerable amount of time trying to trick it to work the same way, I ended up just using `touch` events, as suggested in the original issue.
There's just two big differences between the original implementation:
| before | after |
| - | - |
| custom pointer start/end functions handled native `down/up` events |  now we manually call `touchend = pointerup` |
| on pointer move we'd monkey-patch `preventDefault` | we only apply it in the gesture zone |

### DIFF
--- a/src/components/TraceGesture.tsx
+++ b/src/components/TraceGesture.tsx
@@ -83,10 +83,10 @@ const TraceGesture = ({ eventNodeRef }: TraceGestureProps) => {
     })
 
     eventNode?.addEventListener('touchmove', e => {
-      const isGesturizing = gestureStore.getState().length > 0
+      const isGestureInProgress = gestureStore.getState().length > 0
       const touch = e.touches[0]
 
-      if (isGesturizing && isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
+      if (isGestureInProgress && isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
         handleTouchMove(e)
       }
     })
--- a/src/components/TraceGesture.tsx
+++ b/src/components/TraceGesture.tsx
@@ -83,9 +83,10 @@ const TraceGesture = ({ eventNodeRef }: TraceGestureProps) => {
     })
 
     eventNode?.addEventListener('touchmove', e => {
-      if (!show) return
+      const isGesturizing = gestureStore.getState().length > 0
       const touch = e.touches[0]
-      if (isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
+
+      if (isGesturizing && isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
         handleTouchMove(e)
       }
     })
@@ -111,7 +112,7 @@ const TraceGesture = ({ eventNodeRef }: TraceGestureProps) => {
       eventNode?.removeEventListener('touchend', handleTouchEnd)
       signaturePad.removeEventListener('beginStroke', onBeginStroke)
     }
-  }, [eventNodeRef, onBeginStroke, leftHanded, show])
+  }, [eventNodeRef, onBeginStroke, leftHanded])
 
   return (
     <div
--- a/src/components/TraceGesture.tsx
+++ b/src/components/TraceGesture.tsx
@@ -65,23 +65,36 @@ const TraceGesture = ({ eventNodeRef }: TraceGestureProps) => {
     const signaturePad = signaturePadRef.current['signaturePad']
     const eventNode = eventNodeRef?.current
 
-    // Attach pointer handlers to a provided node rather than the signature pad canvas.
+    // Attach touch handlers to a provided node rather than the signature pad canvas.
     // See: eventNodeRef
-    const handlePointerStart = signaturePad._handlePointerStart.bind(signaturePad)
-    const handlePointerMove = signaturePad._handlePointerMove.bind(signaturePad)
+    const handleTouchStart = signaturePad._handleTouchStart.bind(signaturePad)
+    const handleTouchMove = signaturePad._handleTouchMove.bind(signaturePad)
+    const handleTouchEnd = signaturePad._handleTouchEnd.bind(signaturePad)
 
-    eventNode?.addEventListener('pointerdown', e => {
+    eventNode?.addEventListener('touchstart', e => {
       // Make preventDefault a noop otherwise tap-to-edit is broken.
       // e.cancelable is readonly and monkeypatching preventDefault is easier than copying e.
       e.preventDefault = noop
-      const shouldActivateGesture = isInGestureZone(e.clientX, e.clientY, leftHanded)
-      if (shouldActivateGesture) {
-        handlePointerStart(e)
+
+      const touch = e.touches[0]
+      if (isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
+        handleTouchStart(e)
       }
     })
-    eventNode?.addEventListener('pointermove', e => {
+
+    eventNode?.addEventListener('touchmove', e => {
+      const touch = e.touches[0]
+      if (isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
+        handleTouchMove(e)
+      }
+    })
+
+    eventNode?.addEventListener('touchend', e => {
+      // Make preventDefault a noop otherwise tap-to-edit is broken.
+      // e.cancelable is readonly and monkeypatching preventDefault is easier than copying e.
       e.preventDefault = noop
-      handlePointerMove(e)
+
+      handleTouchEnd(e)
     })
 
     signaturePad.addEventListener('beginStroke', onBeginStroke)
@@ -92,8 +105,9 @@ const TraceGesture = ({ eventNodeRef }: TraceGestureProps) => {
     signaturePad.canvas.height = signaturePad.canvas.offsetHeight
 
     return () => {
-      eventNode?.removeEventListener('pointerdown', handlePointerStart)
-      eventNode?.removeEventListener('pointermove', handlePointerMove)
+      eventNode?.removeEventListener('touchstart', handleTouchStart)
+      eventNode?.removeEventListener('touchmove', handleTouchMove)
+      eventNode?.removeEventListener('touchend', handleTouchEnd)
       signaturePad.removeEventListener('beginStroke', onBeginStroke)
     }
   }, [eventNodeRef, onBeginStroke, leftHanded])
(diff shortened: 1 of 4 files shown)


snqb:
> Makes sense, let's give touch events a try.

https://github.com/cybersemics/em/issues/2303#issuecomment-2548524913

trevinhofmann (changes requested):
Thank you for the PR, @snqb!

This change does correctly upgrade `react-signature-pad-wrapper` and `signature_pad` while avoiding the `undefined is not an object` error. However, I am seeing at least two regressions.

## 1. Page scrolls while attempting to draw a gesture

### Current Behavior

When drawing a gesture, the page scrolls. This causes the gesture drawing to move a bit wildly.

https://github.com/user-attachments/assets/2b776868-4187-4052-8e32-f9d0ffed0184

### Expected Behavior

The page should not scroll while a gesture is in progress.

https://github.com/user-attachments/assets/b4d1cd3a-817d-425d-adc9-1e4218bae198

## 2. Unable to click to edit thoughts

### Current Behavior

Clicking a thought selects it, but does not place the cursor for editing.

https://github.com/user-attachments/assets/afd603fa-2391-46a1-8265-eba2a9ef2d60

### Expected Behavior

Clicking a selected thought should place the cursor and open the keyboard for editing.

https://github.com/user-attachments/assets/645b87c8-955a-4c7b-a0bf-0d0cdabbb3fe

src/components/TraceGesture.tsx
```diff
@@ -65,23 +65,36 @@ const TraceGesture = ({ eventNodeRef }: TraceGestureProps) => {
...
+      const touch = e.touches[0]
+      if (isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
+        handleTouchStart(e)
+      }
+    })
+
+    eventNode?.addEventListener('touchmove', e => {
+      const touch = e.touches[0]
+      if (isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
+        handleTouchMove(e)
```
snqb, line 88: we should check for this too, as blindly monkey-patching `event.preventDefault` stops our scroll prevention

but this way we can just trigger the trace only on our gesture side, so it works very well

snqb:
hey, I've fixed those. I could repro them on my android, but now I can't.
Very sorry for not being able to test on ios myself for now

trevinhofmann (changes requested):
Thank you, @snqb! These issues have been resolved when testing the latest changes:

>1. Page scrolls while attempting to draw a gesture
>2. Unable to click to edit thoughts

Could you please investigate the failing puppeteer test at `src/e2e/puppeteer/__tests__/sidebar.ts > sidebar > recently edited thoughts`?

snqb:
@trevinhofmann I just rebased it against latest master, now it's green. It was something on the upstream, and got fixed I believe.

src/components/TraceGesture.tsx
```diff
@@ -65,23 +65,36 @@ const TraceGesture = ({ eventNodeRef }: TraceGestureProps) => {
...
+    const handleTouchStart = signaturePad._handleTouchStart.bind(signaturePad)
+    const handleTouchMove = signaturePad._handleTouchMove.bind(signaturePad)
+    const handleTouchEnd = signaturePad._handleTouchEnd.bind(signaturePad)
 
-    eventNode?.addEventListener('pointerdown', e => {
+    eventNode?.addEventListener('touchstart', e => {
       // Make preventDefault a noop otherwise tap-to-edit is broken.
       // e.cancelable is readonly and monkeypatching preventDefault is easier than copying e.
       e.preventDefault = noop
-      const shouldActivateGesture = isInGestureZone(e.clientX, e.clientY, leftHanded)
```
trevinhofmann, line 77: I slightly prefer keeping this intermediate variable (and including one in the `touchmove` event listener) because it clarifies the intent via its name. It's always nice to read `if (booleanWithClearName) {`. However, these handlers are already relatively simple, and I don't feel strongly enough to request changing it.

trevinhofmann (approved):
Thank you for rebasing and resolving that puppeteer test!

Everything appears to be working correctly for me, and all checks pass 🎉

raineorshine (changes requested):
Thanks! Looks good.

The only problem I'm seeing is that the gesture trace stops at the edge of the scroll zone. As long as the gesture is initiated in the gesture zone, it should be able to go anywhere on the screen, as in `main`.

snqb:
> Thanks! Looks good.
> 
> The only problem I'm seeing is that the gesture trace stops at the edge of the scroll zone. As long as the gesture is initiated in the gesture zone, it should be able to go anywhere on the screen, as in `main`.
> 

I couldn't reproduce it on windows touch, android and latest ios + safari, iphone 14pro
the gesture does get to the scroll zone with no hassle, the commands are activated fine, and the look-and-feel is just like main on my end

image

trevinhofmann:
I tried, but am also unable to reproduce the issue with the gesture not extending into the scroll zone. It is working fine for me on iOS. Which device are you using?

iOS Capacitor:

hello

iOS Safari:

hello safari

raineorshine:
Hmmm, I'll try again! I was testing on iOS Safari.

src/components/TraceGesture.tsx
raineorshine: Correct me if I'm wrong, but we shouldn't have to check if the touch is in the gesture zone on every `touchmove` event. Once the user starts gesturing the signature should work anywhere.

This would explain the behavior I was seeing earlier where the trace stopped at the boundary of the scroll zone, though confusingly I can no longer reproduce it :/. There's something else going on here I can't quite figure out.

If you need this to only fire when a gesture is in progress you can check `show`. That might help with scroll performance, too.

raineorshine (changes requested):
I tested it again and can no longer reproduce the boundary behavior. I'm not sure why it was doing that! 😅

(I don't know what is more annoying—when things don't work, or when things work but can't be explained!)

snqb:
previously, in `pointermove`: we'd capture the whole view's `pointermoves`, and pass it to the signature pad
so the naive update would be
```tsx
eventNode?.addEventListener('pointermove' -> `touchmove`, e => {
  e.preventDefault = noop
  handlePointerMove -> handleTouchMove(e)
})
```

1. but, now we can't `noop` the thing and pass it, since `touchmove` is a very important event
```tsx
// Multigesture.tsx
document.addEventListener 'touchmove' e
if (this.disableScroll) {
  e.preventDefault()
}
```

2. also, since `eventNode` is basically the whole screen, we can't also
```tsx
eventNode?.addEventListener('pointermove' -> `touchmove`, e => {
  handlePointerMove(e)
})
```
it breaks the scroll, because everything gets fed to the signature pad. 
- yes, we won't be able to gesturize on the scrollzone, because the `event` is not nooped, and then cancelled successfully in MutiGesture,
- but we also can't scroll, since we are sending all of those events down to `handlePointerMove`, which is not meant to be scrollable.

so in the end, we are doing this: let signature pad capture events in the gestures zone only:
```tsx
  if (isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
    handleTouchMove(e)
  }
```

the behavior which you say yesterday can't be explained by this, as the `beginstroke` is called only on `pointerdown`, `mousedown` and ` touchstart` events
{12D980C2-8486-4810-92DA-2097E46DB449}


so in the end: I have one question -- why we used to `noop` the event in `pointermove` in the first place xd
it doesn't affect `tap-to-edit` as `-start` and `-end` events do
and I don't really get why we should noop it there otherwise

snqb:
I think too that `noop` for `move` was a mistake. But it's absolutely necessary for the `start` and `end` ones, I've tested it.

And if we don't wanna invoke the `isInGestureZone` on each `touchmove`, I'd say it's wise.

What I can propose after some lengthy ruminations is to limit the gesturepad's container using the `viewportStore`'s data.
The gestures go outside the box just fine, as it's what they've allowed in the latest version of the gesture pad.
But the events are not registered to the right, and only registered on the red zone.
So it works beautifully. Not sure if it's what you might want, but to me it makes a lot of sense.
Also it could mean, this could simplify some things, but let's see what you think.

{6D654E49-4954-495F-BE52-73F1FE1300E5}

raineorshine:
Thanks for your analysis. It's an interesting challenge since `handleTouchMove` messes up the scroll zone.

Limiting the dimensions of the signature pad element seems like a plausible solution. Let's try it. If it doesn't work, `isInGestureZone` is just a little bit of math so *probably* won't have an impact on performance.

snqb:
so, setting up this way makes me have to create another div to use as `eventNodeRef` and make it `absolute` or `fixed` to give it a size of `width - scrollZone` and position it on the "gesture" part of the screen.

And that leads to z-index issues. If I put a z-index above `content`, the content touch is broken. E.g. tap-to-edit.
If I set it below, it just doesn't register gestures xdd

I tried to get around this a lot of different ways, but can't seem to find the way to fix this.
{E611F058-C20E-4B16-9C87-942334E9B1FA}

It sounded so good at first, but now I'm thinking maybe the calculations were fine.

raineorshine:
Thanks for giving it a try. It you're running into difficulties, it's probably best to go with the existing gesture zone calculation. 

Not sure if you've tried this yet, but you may also be able to short circuit early if there is no gesture in progress.

snqb:
i added the `show` thing as a short-circuiting mechanism

had to add it as a dep for the effect, but i guess it's fine

raineorshine (changes requested):
> i added the `show` thing as a short-circuiting mechanism

Thank you!

> had to add it as a dep for the effect, but i guess it's fine

Oops, I didn't think of that! It's not ideal, as it means all the event handlers are removed and re-added every time a gesture starts. Could you access `gestureStore.getState()` directly instead? We don't need the reactive value.

snqb:
yeah, removed it
works very well like this
and I like this statement
```tsx
  if (isGesturizing && isInGestureZone(touch.clientX, touch.clientY, leftHanded)) {
    handleTouchMove(e)
  }
```

hope you don't mind the variable name :D

raineorshine (approved):
> I hope you don't mind the variable name :D

I'd prefer to call it `isGestureInProgress`. I think that will be more understandable to the average developer.

I appreciate the creativity though :)

PR #2710, opened 2024-12-18
[iOS] Dictation

### LINKED ISSUE #2119: [iOS] Speech-to-text v2
Speech-to-text, i.e. dictation, support was originally added in f3560eb5eb939720c038ba08a375e99d645b531c but is now broken. An alternative design is now proposed. 

## Steps to Reproduce

1. Create a new thought. 
2. Enter Dictation mode by pressing the microphone icon in the bottom right corner of the virtual keyboard. 
3. Say "Hello"
4. Say "newline"
5. Say "Goodbye"

image

## Current Behavior

Newlines are deleted. 

``` 
- HelloGoodbye
``` 

## Expected Behavior

When "newline" is dictated:

1. Create a new thought above with all the text before the newline. 
2. Delete the text up to and including the newline in the cursor thought. 
3. If there is any text after the newline (which can occur when speaking quickly), preserve it. 

``` 
- Hello
- Goodbye
```

mykhailonehelia:
fixes #2119


src/components/Editable.tsx
```diff
@@ -346,6 +346,20 @@ const Editable = ({
         ),
       )
 
+      if (newValue.toLowerCase().indexOf(' new thought') > 0) {
+        const [prev, next] = newValue.split('new thought')
```
raineorshine, line 350: The split should also be case insensitive.

src/components/Editable.tsx
```diff
@@ -346,6 +346,20 @@ const Editable = ({
...
+        newValue = prev.trim()
+
+        setTimeout(() => {
+          dispatch(
+            newThought({
+              at: rootedParentOf(state, path),
```
raineorshine, line 356: You should omit `at`. It will create the new thought after the cursor thought by default. Creating the new thought at the end of the parent doesn't make sense.

raineorshine (changes requested):
Thanks for the PR! Looks like a good start.

The only issue is that this will incorrectly trigger when the user types "new thought" on the keyboard. We don't want this triggering when the user types or edits a thought. It should only trigger during dictation.

src/components/Editable.tsx
mykhailonehelia: got it

src/components/Editable.tsx
mykhailonehelia: got it

src/components/Editable.tsx
```diff
@@ -346,6 +347,19 @@ const Editable = ({
         ),
       )
 
+      if (inputMethod === 'dictation' && newValue.toLowerCase().indexOf(' new thought') > 0) {
```
raineorshine, line 350: This would result in a false positive when "new thought" is typed on the keyboard and then the user enters dictation mode. The intended behavior is for "new thought" and other commands to only be transformed when the user is dictating.

raineorshine (changes requested):
Thanks for the update! This is an improvement, as typed "new thought" no longer triggers the dictation command. However, I'm finding that it fails to detect "new thought" in the case where dictation inserts "new" and "thought" in separate events, which is quite common. When I practiced dictating I found that I got a lot of false negatives.

I think you will need to look at the full value of the thought to pick up on these cases where "new" and "thought" are input by the dictation feature as separate words. However, you also need to avoid false positives and ignore "new thought" that was previously typed. Therefore, a better approach might be to record the selection offset when dictation starts so that you can exclude existing text from detection.

e.g. If you have thought `Hello| World` (with the caret in between "Hello" and "World") and you start dictating, then `value.replace(/^Hello/, '').replace(/ world$/, '')` will give you the dictated text excluding text that was already there.

While the expected functionality is fairly intuitive when using the app, it can be helpful to spell it out more explicitly so that you can review each of these cases and ensure that your solution works as expected:

- Trigger the New Thought command when "new thought" is dictated.
- Trigger the New Thought command when "new [pause] thought" is dictated.
- Do not trigger New Thought when "new thought" is typed.
- Do not trigger New Thought when dictating new text on a thought that already contains the text "new thought".

Thanks and let me know if I can clarify anything else!

Issue #2686, opened 2024-12-06
Cursor Up does not work correctly in sorted context with bold thought


raineorshine:
Related to #2685

## Steps to Reproduce

```
- fruits
  - apple
  - orange
  - banana
  - pear
```

1. Set the cursor on `apple`
2. Click to sort `Ascending`
3. Click to sort `Descending`
4. Make `apple` bold
5. Set the cursor on `pear`
6. ArrowUp

## Current Behavior

The cursor moves to `fruits`.

## Expected Behavior

The cursor should move to `apple`.

## Note

If you move the cursor around, you'll notice that it thinks the subthoughts are in the following order:

```
- pear
- orange
- banana
- apple
```

raineorshine:
Fixed in #2720

PR #2719, opened 2024-12-23
fix cursor move tests

RED-ROSE515:
This PR resolves the issue : Cursor Up does not work correctly in sorted context with bold thought #2686

The cause of Issue : The problem is that when the list is sorted in descending order, the `prevSibling` function is still looking for the previous sibling in the internal order, not the visual order.

When you sort in descending order, the internal order and visual order of thouths become :
   Internal order:                   
   - pear                                 
   - orange                              
   - banana                             
   - apple

   Visual order:     
   - apple 
   - banana  
   - orange                            
   - pear
   
   So, when cursorUp and cursorDown, it doesn't work properly.
   
Solution for this Issue : To get the visual order of thoughts, I used `getChildrenRanked` instead of `getChildrenSorted` in 
    prevSibling and nextSibling. And it has been solved.

### DIFF
--- a/src/actions/__tests__/cursorDown.ts
+++ b/src/actions/__tests__/cursorDown.ts
@@ -1,13 +1,16 @@
-import State from '../../@types/State'
+import { act } from 'react'
 import cursorDown from '../../actions/cursorDown'
 import importText from '../../actions/importText'
+import { importTextActionCreator as importTextAction } from '../../actions/importText'
 import newSubthought from '../../actions/newSubthought'
 import newThought from '../../actions/newThought'
-import toggleAttribute from '../../actions/toggleAttribute'
 import toggleContextView from '../../actions/toggleContextView'
-import contextToPath from '../../selectors/contextToPath'
+import toggleSortShortcut from '../../commands/toggleSort'
+import createTestStore from '../../test-helpers/createTestStore'
 import expectPathToEqual from '../../test-helpers/expectPathToEqual'
+import { setCursorFirstMatchActionCreator as setCursorAction } from '../../test-helpers/setCursorFirstMatch'
 import setCursor from '../../test-helpers/setCursorFirstMatch'
+import executeCommand from '../../util/executeCommand'
 import initialState from '../../util/initialState'
 import pathToContext from '../../util/pathToContext'
 import reducerFlow from '../../util/reducerFlow'
@@ -64,19 +67,25 @@ describe('normal view', () => {
     expectPathToEqual(stateNew, stateNew.cursor, ['b'])
   })
 
-  it('work for sorted thoughts', () => {
-    const steps = [
-      newThought('a'),
-      newSubthought('n'),
-      newThought('m'),
-      setCursor(['a']),
-      (state: State) =>
-        toggleAttribute(state, { path: contextToPath(state, ['a']), values: ['=sort', 'Alphabetical'] }),
-      cursorDown,
-    ]
-
-    const stateNew = reducerFlow(steps)(initialState())
-
+  it('work for sorted thoughts', async () => {
+    const store = createTestStore()
+    act(() => {
+      store.dispatch([
+        importTextAction({
+          text: `
+              - a
+                - n
+                - m
+          `,
+        }),
+        setCursorAction(['a', 'n']),
+      ])
+    })
+    act(() => executeCommand(toggleSortShortcut, { store }))
+    act(() => {
+      store.dispatch([setCursorAction(['a'])])
+    })
+    const stateNew = cursorDown(store.getState())
     expectPathToEqual(stateNew, stateNew.cursor, ['a', 'm'])
   })
 })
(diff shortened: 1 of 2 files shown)


src/actions/__tests__/cursorDown.ts
```diff
@@ -65,10 +65,16 @@ describe('normal view', () => {
...
 
   it('work for sorted thoughts', () => {
+    const text = `
+      - a
+        - =sort
+          - Alphabetical
+            - Asc
+        - m
+        - n
+    `
```
raineorshine, line 75: I'm fine with replacing the `newThought` and `newSubthought` calls with a single `importText`. However, we should start with the thoughts unordered and then sort them in the test. Starting them in sorted order with `=sort` already applied misses this.

```
- a
  - n
  - m
```

However, it also looks like there was a bug in the test already, as it uses `toggleAttribute` directly rather than `toggleSort`. We should switch to `toggleSort`.

src/actions/__tests__/cursorDown.ts
```diff
@@ -79,6 +85,33 @@ describe('normal view', () => {
...
+  it('should navigate correctly in descending sorted list with formatted text', () => {
+    const text = `
+      - fruits
+        - =sort
+          - Alphabetical
+            - Desc
+        - apple
+        - orange
+        - banana
+        - pear
```
raineorshine, line 98: It's not possible to have `=sort/Alphabetical/Desc` on a list of unsorted thoughts with normal use of the app. Activating Sort Descending would permanently rearrange the list.

src/actions/__tests__/cursorDown.ts
```diff
@@ -79,6 +85,33 @@ describe('normal view', () => {
 
     expectPathToEqual(stateNew, stateNew.cursor, ['a', 'm'])
   })
+
+  it('should navigate correctly in descending sorted list with formatted text', () => {
```
raineorshine, line 89: Avoid using the word "correctly" in a test, as it does not describe *what* the correct behavior is. Instead, describe the expected behavior.

`it('move cursor from formatted thought to first unformatted thought in descending order', ...)`

src/commands/__tests__/cursorNext.ts
```diff
@@ -75,25 +75,17 @@ describe('normal view', () => {
...
-            path: contextToPath(getState(), ['SORT']),
-            values: ['=sort', 'Alphabetical'],
-          }),
-        ),
-      setCursor(['SORT', 'a']),
-      cursorNext(),
     ])
-
+    setCursor(['a'])
+    executeCommandWithMulticursor(toggleSortShortcut, { store })
```
raineorshine, line 85: Great! This is better because it triggers the toggleSort shortcut rather than directly setting meta attributes.

src/actions/__tests__/cursorUp.ts
```diff
@@ -171,6 +173,33 @@ describe('normal view', () => {
     const cursor = state.cursor
     expect(pathToContext(state, prevThought(state, cursor!)!)).toEqual(['a'])
   })
+
+  it('should navigate correctly in descending sorted list with formatted text', () => {
```
raineorshine, line 177: See similar comments in `cursorDown`.

raineorshine (changes requested):
Thanks for the PR!

It looks like this was a bug from an incomplete migration of the sort functionality. It used to be possible to have a context rendered in sorted order without changing the ranks. However, toggleSort was changed to re-rank the entire list so that sorting is permanent. This was done because `=sort` is loaded later than the list itself, causing thoughts to get briefly rendered unordered, which was visually disruptive. `getChildrenSorted` is a remnant from the previous implementation. Perhaps we can factor out it out completely in a separate refactor.

Good job identifying the appropriate functions to fix the sort order.

src/e2e/puppeteer/__tests__/cursor-move.ts
```diff
@@ -0,0 +1,44 @@
+import click from '../helpers/click'
+import clickThought from '../helpers/clickThought'
+import getEditingText from '../helpers/getEditingText'
+import paste from '../helpers/paste'
+import press from '../helpers/press'
+import waitForEditable from '../helpers/waitForEditable'
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+it('move cursor from formatted thought to first unformatted thought in descending order', async () => {
```
raineorshine, line 10: Any reason you changed this to a puppeteer test? We really want to use RTL tests when possible since puppeteer tests are so slow.

RED-ROSE515: @raineorshine , I tried to implement using RTL but we can't run document.execCommand in RTL.

raineorshine:
It looks like #2686 was fixed by #2720 so I think this is no longer needed. However, the unit tests would still be helpful.

src/e2e/puppeteer/__tests__/cursor-move.ts
RED-ROSE515: toggleSort works in RTL but I need to check the bold thought goes to first of children in this test.

raineorshine:
Let's remove the new sort logic if it is no longer needed and move the puppeteer test to existinghttps://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/__tests__/cursor.ts. Then I think this is good to merge. Thanks!

raineorshine (approved):

Issue #2685, opened 2024-12-06
Sort error on bold thought


raineorshine:
## Steps to Reproduce

```
- fruits
  - apple
  - orange
  - banana
  - pear
```

1. Set the cursor on `apple`
2. Click to sort `Ascending`
3. Click to sort `Descending`
4. Make `apple` bold

## Current Behavior

The thoughts are sorted correctly, but there is a toolbar button error: `Ranks do not match sort condition`.

https://github.com/user-attachments/assets/0b744e13-e242-4ca9-b54e-b6fe7b4d1aac

## Expected Behavior

There should be no error, and the thoughts should still be sorted correctly.

- fruits
  - **apple**
  - pear
  - orange
  - banana

RED-ROSE515:
Hello @raineorshine , I have checked `toggleSort.ts`.
Currently, checking the order of `getAllChildrenSorted` and `getChildrenRanked`.
`getAllChildrenSorted` checks the sort preference and can return either alphabetically sorted thoughts (using `getChildrenSortedAlphabetical`) or rank-sorted thoughts (using `getChildrenRanked`).

`getChildrenRanked` always sorts by rank using compareByRank function.

And now, the sortType is **Alphabetical**. So those two functions also return different order of children.
So, `toggleSort` causes error.

raineorshine:
Thanks, that makes sense. Now, the question is, why does the rank order not match the sort order? Toggle Sort should re-rank thoughts in sorr order. Clearly the **bold** thought is handled differently. You will need to identify why bold is handled differently and indicate which code is wrong. 

Please explain it both specifically (in terms of the exact thoughts in the test case) and generally (in terms of the algorithm).

RED-ROSE515:
I found the best solution for this issue.
This issue casued because of `compareThougtDescending`.
Currently, in `compareThoughtDescending` function, return the reverse of `compareThought`.

So, I changed the `compareThoughtDescending` function to check if either thought has HTML formatting. And `compareFormatting` first.
This resolves the current Issue.

RED-ROSE515:
But there is another regression in sort logic now.

https://github.com/user-attachments/assets/e31d748b-5ed1-479e-b01b-8ff5f522f78f

As you can see in the video, when bold the **apple**, it's okay to set `apple` text to first.
But when remove `bold`, **apple** needs to go to last. It doesn't work properly now.

raineorshine:
[...]

PR #2720, opened 2024-12-24
fix sort error on bold thought

RED-ROSE515:
This PR resolves the issue : Sort error on bold thought #2685

This issue casued because of `compareThougtDescending`.
Currently, in `compareThoughtDescending` function, return the reverse of `compareThought`.

I found the best solution for this issue.
So, I changed the `compareThoughtDescending` function to check if either thought has HTML formatting. And `compareFormatting` first.
This resolves the current Issue.

### DIFF
--- a/src/selectors/getSortedRank.ts
+++ b/src/selectors/getSortedRank.ts
@@ -1,15 +1,21 @@
 import State from '../@types/State'
 import ThoughtId from '../@types/ThoughtId'
-import { compareReasonable } from '../util/compareThought'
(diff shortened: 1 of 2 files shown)


raineorshine:
Thanks for identifying the issue. I can't accept a solution that introduces a [regression](https://github.com/cybersemics/em/issues/2685#issuecomment-2560717420), so it seems like you have some more work to do.

raineorshine (approved):
Looks good! Thanks!

Issue #2723, opened 2024-12-24
Context View: `=pin` should not be applied on context


raineorshine:
## Steps to Reproduce

```
- a
  - m
    - x
- b
  - m
    - =pin
    - y
```

Activate context view on `a/m`.

## Current Behavior

`a/m~/b` is expanded.

<img width="201" alt="Screenshot 2024-12-24 at 9 32 03 AM" src="https://github.com/user-attachments/assets/eea791f7-d031-4eb3-b324-59e72cf1da42" />

## Expected Behavior

`a/m~/b` should not be expanded.

`=pin` should not be applied on contexts in the context view.

<img width="201" alt="Screenshot 2024-12-24 at 9 32 23 AM" src="https://github.com/user-attachments/assets/36845fdf-646b-4ae5-8a7c-700c6076a7b9" />

## Testing

Add test to https://github.com/cybersemics/em/blob/7d812c7c6dd0ebb325dc5431d86411ca501f3424/src/selectors/__tests__/expandThoughts.ts.

PR #2731, opened 2024-12-25
=pin should not be applied on context

RED-ROSE515:
This PR resolves the Issue : Context View: =pin should not be applied on context #2723

In this PR, I've modified the `expandThoughtsRecursive` function to only apply pin expansion when not in context view.
Added a condition !showContexts to check if we're not in context view before applying pin expansion.


raineorshine (approved):
👍

Issue #2711, opened 2024-12-20
getThoughtById: Return undefined when thought does not exist


raineorshine:
`getThoughtById` currently returns `Thought` even though the thought by may not exist anymore. The only time state.thoughts.thoughtIndex[id] is only undefined in rare circumstances after an async operation or unmount, e.g. CSSTransition, but it is possible nonetheless and should be typed accordingly.

When I explicitly set the return type to `Thought | undefined` there are 170 type errors, so this touches many files. I suggest making these changes incrementally. In most cases it will simply require detecting if the thought does not exist and short circuiting, but the exact handling will depend on the context:

```ts
if (!thought) return
```

Related: #2591

snqb:
hey, so if I change the `getThoughtById`'s signature to return `undefined` too, should I go over the usages and mark them as 
`@ts-expect-error` before fixing them little by little? can't think of other way to suspend the errors we'll have

raineorshine:
Yes, that would work. You could also change the return type back to non-null for partial PR's and we can change it permanently at the end when all call sites have been updated.

snqb:
Hey, I've almost finshed. There are cases that I feel need discussion

#### the heads:
```typescript
const headRank = (state: State, path: Path) => getThoughtById(state, head(path)).rank // -> number | undefined
const headValue = (state: State, path: Path) => getThoughtById(state, head(path)).value // -> string | undefined
```
previously we relied it on them being defined
- i don't want to put a default value here to preserve type, or I am not sure what number to put
- should it then become also undefined-able? then ok, but these touch a lot of places
- do we yolo `!` it and see how it goes?

not sure what to decide

snqb:
besides the third bunch, there is only `headRank` and `headValue` left, btw.

![{C3CCB0C1-0EC5-433B-B983-402AD129D014}](https://github.com/user-attachments/assets/79947238-19c4-499b-8b96-c20d59191fab)

raineorshine:
Thanks! I think `headRank` and `headValue` have to return undefined. That's the only way to make them type safe.

PR #2732, opened 2024-12-25
First bunch of possibly undefined thoughts

snqb:
https://github.com/cybersemics/em/issues/2711

went through 14 files out of (?)
as suggested, to have as little conflicts as possible, I'm gonna go through with small bunches
mostly returning state, and sometimes putting warnings where the file has those already, or where I found it could be very confusing if we dind't have ones

### DIFF
--- a/src/actions/archiveThought.ts
+++ b/src/actions/archiveThought.ts
@@ -79,6 +79,8 @@ const archiveThought = (state: State, options: { path?: Path }): State => {
 
   if (!thought) {
     console.error(`achiveThought: Thought not found for id ${head(simplePath)}`)
+
+    return state
   }
   const thoughts = pathToContext(state, simplePath)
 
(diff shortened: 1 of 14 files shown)


src/actions/archiveThought.ts
```diff
@@ -79,6 +79,8 @@ const archiveThought = (state: State, options: { path?: Path }): State => {
 
   if (!thought) {
     console.error(`achiveThought: Thought not found for id ${head(simplePath)}`)
+
+    return state
```
snqb, line 83: we relied on `thought` below, as if it's defined

src/actions/cursorForward.ts
```diff
@@ -31,7 +31,7 @@ const cursorForward = (state: State) => {
   else {
     const simplePath = simplifyPath(state, cursor)
     const firstChild = firstVisibleChild(state, head(simplePath))
-    isValidChild = cursorFromHistory && !!getThoughtById(state, head(cursor)).childrenMap[head(cursorFromHistory)]
+    isValidChild = cursorFromHistory && !!getThoughtById(state, head(cursor))?.childrenMap[head(cursorFromHistory)]
```
snqb, line 34: used as a boolean predicate, so we're fine with `?`

raineorshine (approved):
This approach looks good to me. Thanks!

PR #2734, opened 2024-12-25
Second bunch of possibly undefined thoughts

snqb:
https://github.com/cybersemics/em/issues/2711

continuation of https://github.com/cybersemics/em/pull/2732

trying to be as tidy as possible, preserving a given file's error conventions, returning the same things returned for empty cases, e.g. somewhere it's `null`, sometimes it's `0` e.g. with ranks, sometimes it's `undefined` being passed on, as the function type already had `undefined` as a possibliity, sometimes `?.` access is enough for predicates, and sometimes it's `{}` and `false`.

there is one thing that requires discrete attention(i can revert it separately :D): 34e6d63c562733e96a39ee7eb39388aae153ce53
we used to have
`const isPending = (state: State, thought: Thought | null): boolean => ...`

it expects a `Thought` or `null` strangely, since every of 4 usages of `isPending` use the immediate result of `getThoughtById`, which has no hints of returning null whatsoever.
so I made it expect `undefined` also
`const isPending = (state: State, thought: Thought | null | undefined): boolean => ...`


src/selectors/getContextsSortedAndRanked.ts
```diff
@@ -28,12 +29,13 @@ const getContextsSortedAndRanked = (state: State, value: string): Thought[] => {
         ...(thought || {
           id: cxid,
           childrenMap: {},
-          parentId: '', // ???
+          parentId: '' as ThoughtId, // ???
```
snqb, line 32: I should've not touched this maybe, but 
`thought` comes from `getThoughtById` and the types don't match: we need `created` and `parentId: ThoughtId` as a "replacement" for our "missing" thought, so I thought I gotta fix this TS issue

src/selectors/getThoughtById.ts
```diff
@@ -2,7 +2,10 @@ import State from '../@types/State'
 import ThoughtId from '../@types/ThoughtId'
 
 /**
- * Gets a Thought by its ThoughtId.
+ * Gets a Thought by its ThoughtId. Can be undefined if the Thought is not in the state.
+ * The only time state.thoughts.thoughtIndex[id] is only undefined in rare
+ * circumstances after an async operation or unmount, e.g. CSSTransition,
+ * but it is possible nonetheless and should be typed accordingly.
```
snqb, line 8: since, I've left a couple `see getThoughtById` comments where I've found things not to be clear, I felt that adding why it could be undefined, makes sense
if it does make sense, maybe I should even put the issue link here too, right?

src/selectors/isPending.ts
```diff
@@ -3,7 +3,7 @@ import Thought from '../@types/Thought'
 import { getAllChildrenAsThoughts } from '../selectors/getChildren'
 
 /** Returns true if the context has not been loaded from the remote yet, or if the children are not loaded. */
-const isPending = (state: State, thought: Thought | null): boolean => {
+const isPending = (state: State, thought: Thought | null | undefined): boolean => {
```
snqb, line 6: this is the "controversial" part I guess
explained in the PR itself

snqb:
also, same goes for 
src/util/isDivider.ts

```typescript
/** Returns true if the value starts with multiple dashes and should be interpreted as a divider. */
const isDivider = (s: string | null) => s !== null && (s.startsWith('---') || s.startsWith('—-'))

export default isDivider
```

a lot of types, we put a `thought.value`, which if `thought` is undefined, is also undefined.
`const isDivider = (s: string | null | undefined) => !!s && (s.startsWith('---') || s.startsWith('—-'))`
could lead to lots of these issues getting "fixed", but is it gonna be correct logically?

snqb:
`  if (value == null) return null`

hey, I've found this piece of code, it's an early return for `Thought.tsx`.
So, judging from `==`, not `===`, it feels like, we used to have no notion of `undefined` vs `null`.
And then this makes me really think that every function that expects `Thought | null` actually means `Thought | undefined`, and the `null` notion is kind of the artifact of the past :D

src/selectors/getContextsSortedAndRanked.ts
raineorshine: That's fine. Clearly `''` is not a valid `ThoughtId`, but it is being treated as one already, so there is no harm in adding the type assertion.

src/selectors/getThoughtById.ts
raineorshine: Good idea

src/selectors/isPending.ts
raineorshine: I think `Thought | undefined` should cover it. I'm not sure why it was using `Thought | null`. Thanks!

raineorshine:
> So I propose allowing an `undefined` here too. If we assumed `null` possible, could we assume `undefined` too? `const isDivider = (s: string | null | undefined) => (s?.startsWith('---') || s?.startsWith('—-'))`
> 
> then we won't have to check for the `thought` existence in many places, since a lot of times, it's a cosmetic thing in a react component.

I agree with your judgment.

> ` if (value == null) return null`
> 
> hey, I've found this piece of code, it's an early return for `Thought.tsx`. So, judging from `==`, not `===`, it feels like, we used to have no notion of `undefined` vs `null`. And then this makes me really think that every function that expects `Thought | null` also means `Thought | undefined`

Yes. `state.cursor` can be null, but `thought?.value` can never be null.

raineorshine (changes requested):
I have re-run the failing Puppeteer snapshot test and it still fails, so I think may have a small regression. Can you investigate?

snqb:
indeed, a regression
I didn't account for `value` being `''`
fixed it, sorry

src/actions/splitThought.ts
```diff
@@ -57,8 +58,9 @@ const splitThought = (state: State, { path, splitResult }: { path?: Path; splitR
 
     // move children
     state => {
-      // we can safely assume that the cursor has been set to the newly created thought that contains valueRight
       const childNew = getThoughtById(state, head(state.cursor!))
+      if (!childNew) return state
+      // we can safely assume that the cursor has been set to the newly created thought that contains valueRight
```
raineorshine, line 63: This comment should go above the line with `state.cursor!` since it attempts to explain the `!`.

src/commands/delete.ts
```diff
@@ -26,7 +26,8 @@ const exec: Command['exec'] = (dispatch, getState) => {
   // otherwise Chrome/Brave will open "Clear browsing data"
   if (!cursor) return
 
-  const value = getThoughtById(state, head(simplifyPath(state, cursor))).value
+  const value = getThoughtById(state, head(simplifyPath(state, cursor)))?.value
+  if (!value) return
```
raineorshine, line 30: Careful... empty string is falsey! :)

```
if (value === undefined) return
```

src/components/CopyOneDrop.tsx
```diff
@@ -13,6 +13,10 @@ import CopyClipboard from './icons/CopyClipboard'
 /** Copy the thought on drop. */
 const drop = (state: State, { simplePath }: DragThoughtItem) => {
   const value = getThoughtById(state, head(simplePath))?.value
+  if (!value) {
```
raineorshine, line 16: `value === undefined`

src/components/DeleteDrop.tsx
```diff
@@ -15,6 +15,10 @@ import DeleteIcon from './icons/DeleteIcon'
 /** Copy the thought on drop. */
 const drop = (state: State, { simplePath, path, zone }: DragThoughtItem) => {
   const value = getThoughtById(state, head(simplePath))?.value
+  if (!value) {
```
raineorshine, line 18: `value === undefined`

src/components/ExportDrop.tsx
```diff
@@ -15,6 +15,10 @@ import ShareIcon from './icons/ShareIcon'
 /** Export the thought on drop. */
 const drop = (state: State, { simplePath }: DragThoughtItem) => {
   const value = getThoughtById(state, head(simplePath))?.value
+  if (!value) {
```
raineorshine, line 18: `value === undefined`

src/components/LayoutTree.tsx
```diff
@@ -268,6 +268,8 @@ const linearizeTree = (
 
   const thoughtId = head(path)
   const thought = getThoughtById(state, thoughtId)
+  // if the thought is suddenly undefined, we opt for a blank tree, see getThoughtById
+  if (!thought) return []
```
raineorshine, line 272: Let's move this into the ternary [below](https://github.com/cybersemics/em/blob/1de83fc608b916d44f83caf39032d3f5442b7b01/src/components/LayoutTree.tsx#L275) since `value` is only used there. That will have a narrower impact.

src/components/Thought.tsx
```diff
@@ -212,7 +212,7 @@ const ThoughtContainer = ({
   // true if the thought has an invalid option
   const invalidOption = useSelector(state => {
     const thought = getThoughtById(state, thoughtId)
-    if (!thought) return false
+    if (!thought || !value) return false
```
raineorshine, line 215: `value === undefined`

src/hooks/useHideBullet.ts
```diff
@@ -36,6 +36,8 @@ const useHideBullet = ({
     const thought = getThoughtById(state, thoughtId)
     // bullet may be set from =children or =grandchildren and passed as a prop
     if (hideBulletProp) return true
+    // and if the thought is somehow not found, we just pass the `undefinedness` up the chain
+    if (!thought) return
```
raineorshine, line 40: Let's return false to preserve the boolean return type.

src/selectors/expandThoughts.ts
```diff
@@ -80,6 +80,10 @@ function expandThoughtsRecursive(state: State, expansionBasePath: Path, path: Pa
   const simplePath = !path || path.length === 0 ? HOME_PATH : simplifyPath(state, path)
   const thoughtId = head(path)
   const thought = getThoughtById(state, thoughtId)
+  if (!thought) {
+    console.error(new Error('expandThoughtsRecursive: Invalid Thought; No thought found with id ' + thoughtId))
+    return {}
+  }
```
raineorshine, line 86: Since `thought` is only used in one place, let's only short circuit in that case rather than short circuiting the whole function:

```ts
  const childrenUnfiltered = showContexts
    ? thought ? childIdsToThoughts(state, getContexts(state, thought.value)) : []
    : ...
```

raineorshine (changes requested):
Overall looks very good! You did a good job returning the "default" value in a variety of circumstances. See full review below.

src/actions/join.ts
```diff
@@ -28,14 +28,19 @@ const join = (state: State, { paths }: { paths?: Path[] } = {}) => {
   const children = paths
     ? paths.map(path => getThoughtById(state, head(path)))
     : getAllChildrenSorted(state, parentId).filter(child => !isAttribute(child.value))
+
+  // getThoughtById -> Thought | undefined, so we need to filter out possible undefined children
+  // and here we ignore missing thoughts, as every function below assumes that the thought exists
+  // it's very unlikely that thought is missing, but we wanna be safe, see getThoughtById
+  const onlyExistingChildren = children.filter(Boolean)
```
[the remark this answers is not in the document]
snqb: Hey, thanks. I gave it a second thought and I think even moving the filter inside the original ternary is cleaner.
```typescript
  const children = paths
    ? // getThoughtById -> Thought | undefined, so if we (unlikely) get undefined, we filter it out, see getThoughtById
      paths.map(path => getThoughtById(state, head(path))).filter(Boolean)
    : getAllChildrenSorted(state, parentId).filter(child => !isAttribute(child.value))
```
Bummer I didn't realize that `getAllChildrenSorted` already handles empty thoughts at first.

src/components/LayoutTree.tsx
snqb: I've found this approach to be purest if we put it in ternary.
```typescript
  const children = contextViewActive
    ? thought ? getContextsSortedAndRanked(state, thought.value) : []
    : // context children should render the children of a specific Lexeme instance to avoid repeating the Lexeme.
      // See: contextId (above)
      getChildrenRanked(state, contextId || thoughtId)
      ```

File: src/hooks/useHideBullet.ts (hunk 11 above)
Comment (snqb) [reply src/hooks/useHideBullet.ts:40]: I had a rationale, but keeping the type signature makes sense to me too 🙇

File: src/commands/delete.ts (hunk 5 above)
Comment (snqb) [reply src/commands/delete.ts]: yeah, that was clumsy...
thanks!

File: src/actions/join.ts [hunk 14]
```diff
@@ -25,9 +25,13 @@ const join = (state: State, { paths }: { paths?: Path[] } = {}) => {
...
   const simplePath = simplifyPath(state, path)
   const parentId = head(parentOf(simplePath))
+
   const children = paths
-    ? paths.map(path => getThoughtById(state, head(path)))
+    ? // getThoughtById -> Thought | undefined, so if we (unlikely) get undefined, we filter it out, see getThoughtById
+      paths.map(path => getThoughtById(state, head(path))).filter(Boolean)
     : getAllChildrenSorted(state, parentId).filter(child => !isAttribute(child.value))
+
+  // const children = childrenUnchecked.filter(Boolean)
```
raineorshine, line 34: I think you can remove this line now. Maybe left in my accident?

src/components/LayoutTree.tsx
```diff
@@ -268,11 +268,14 @@ const linearizeTree = (
 
   const thoughtId = head(path)
   const thought = getThoughtById(state, thoughtId)
+  // if the thought is suddenly undefined, we opt for a blank tree, see getThoughtById
```
raineorshine, line 271: This can safely be removed now.

raineorshine (approved):

Issue #2721, opened 2024-12-24
Note on context child: Error: [] is not a valid path


raineorshine:
## Steps to Reproduce

```
- a
  - m
    - x
- b
  - m
    - y
```

1. Activate context view on `a/m`.
2. Set the cursor on `a/m~/a/x`.
3. Activate Note.

## Current Behavior

`Error: [] is not a valid path. Use [__ROOT__].`

## Expected Behavior

No error.

PR #2735, opened 2024-12-25
Fix failing note cursor

snqb:
fixes https://github.com/cybersemics/em/issues/2721, is ad-hoc fix, needs a deeper think to be not-ad-hoc

it was `onFocus` of a newly created note that triggered an incorrect dispatch of `setCursor`:

```
const simplePath = simplifyPath(state, path) || path
.
.
.
setCursor({
  path: simplePath,
  .
  .
}
```

the problem was: `simplifyPath` would return an empty array at times like this.
The and we'd get `simplifyPath(state, path) || path) = `[] || path` = `[]` , so the fallback never really worked 😄 

After putting a more elaborate condition, and using `path` as a fallback, we get broken editor behavior: the part of the tree collapses(we go to a wrong `path`)

In the end, when I just did not allow `[]` paths, it fixed the thing. And the app, from my manual tests, works just fine. The cursor is there, everything is smooth.
`setDescendant` action puts the appropriate state, from what I observed.

and on the reasons why we even get to the case of `simplifyPath` returning `[]` here, I'm not really sure, and feels like, we should dig there, if we want more than just a ad-hoc bug fix.

### DIFF
--- a/src/components/Note.tsx
+++ b/src/components/Note.tsx
@@ -7,35 +7,16 @@ import Path from '../@types/Path'
 import { cursorDownActionCreator as cursorDown } from '../actions/cursorDown'
 import { deleteAttributeActionCreator as deleteAttribute } from '../actions/deleteAttribute'
 import { editingActionCreator as editing } from '../actions/editing'
-import { setCursorActionCreator as setCursor } from '../actions/setCursor'
 import { setDescendantActionCreator as setDescendant } from '../actions/setDescendant'
 import { setNoteFocusActionCreator as setNoteFocus } from '../actions/setNoteFocus'
 import { toggleNoteActionCreator as toggleNote } from '../actions/toggleNote'
 import { isSafari, isTouch } from '../browser'
 import * as selection from '../device/selection'
-import simplifyPath from '../selectors/simplifyPath'
 import store from '../stores/app'
 import equalPathHead from '../util/equalPathHead'
 import head from '../util/head'
 import noteValue from '../util/noteValue'
 import strip from '../util/strip'
-
-/** Sets the cursor on the note's thought as it is being edited. */
-const setCursorOnLiveThought = ({ path }: { path: Path }) => {
-  const state = store.getState()
-  const simplePath = simplifyPath(state, path)
-
-  if (simplePath?.length > 0) {
-    store.dispatch(
-      setCursor({
-        path: simplePath,
-        cursorHistoryClear: true,
-        editing: true,
-        noteFocus: true,
-      }),
-    )
-  }
-}
 
 /** Renders an editable note that modifies the content of the hidden =note attribute. */
 const Note = React.memo(({ path }: { path: Path }) => {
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
Hi, thanks for your submission and for identifying that `simplifyPath` returns `[]`.

As you know we maintain high code quality standards on this project, so I'm not able to accept an ad-hoc fix. For all bugs, we expect to identify the root cause which will inform an appropriate solution. In rare cases we may determine that the effort to create a proper solution is not justified, but that is not the case here. I would encourage you to trace `simplifyPath` to determine what is causing it to return `[]`. 

`[]` is not a valid Path, but it's not covered by the type system so we have to manually trace where it's coming from.

trevinhofmann:
Hi, @snqb. Thank you very much for the initial pull request. Please let me know if you need any further guidance to proceed.

snqb:
yeah, the first submission was just seeing the thing, then I got distracted with other tasks
sorry for having this open for a while, I guess it's not a good practice to keep things open for so long

The more I looked at the issue, the more it seeems to me that doing a `setCursor` is wrong, simplifying the path was wrong too
and I don't quite get why we want to clear the cursor history while focusing on a note.

so the end solution revolves just around manually dispatching `editing + setNoteFocus`
the cursor is already handled anyways
```tsx
/** Sets the cursor on the note's thought as it is being edited. */
  const onFocus = () => store.dispatch([editing({ value: true }), setNoteFocus({ value: true })])
```
and it kinda rhymes with 

```tsx

  const onBlur = () => {
    if (isTouch && !selection.isActive()) {
      setTimeout(() => dispatch(editing({ value: false })))
    }
  }
```


but since also, the `noteFocus` action does change the `editing`
```tsx
...
  noteFocus: value,
  // always enter edit mode when there is note focus
  // it will be set in the Note's onFocus anyway, but set it here so that we are not as dependent on what happens there
  ...(value ? { editing: true } : null),
```

the end function could be just
`const onFocus = () => store.dispatch(setNoteFocus({ value: true }))`

From the preliminary clicking around: it works just fine, and also solves the original issue.
Also, tests seem to pass fine too. Let me know if I missed something, but the solution sounds logical to me.

src/components/Note.tsx
```diff
@@ -111,6 +94,9 @@ const Note = React.memo(({ path }: { path: Path }) => {
     }
   }
 
+  /** Sets the cursor on the note's thought as it is being edited. */
+  const onFocus = () => store.dispatch(setNoteFocus({ value: true }))
```
snqb, line 98: diff from the original fn:

setCursor({
      ~path: simplePath~, // does not simplify a path, we don't need it
      ~cursorHistoryClear: true~, // why do we need it?
      editing: true, // does it inside `setNoteFocus`
      noteFocus: true, // does it via `setNoteFocus`
    })

raineorshine (commented):
Thanks for your submission. We need the `setCursor` actually. If the user clicks on the note of a non-cursor thought, the cursor needs to move to that thought.

raineorshine:
As I was investigating the code to form a proper reply, it became clear that the use of `simplePath` was incorrect and the Note just needed to use `path` instead. I made the fix in e8dbc42b9e.

The case could still use an RTL test. Is this something you could do?

Issue #2724, opened 2024-12-24
Context View: `=table` should not be applied on context


raineorshine:
## Steps to Reproduce

```
- a
  - a1
    - m
      - x
- b
  - =view
    - Table
  - b1
    - m
      - y
```

Activate context view on `a/a1/m`.

## Current Behavior

`a/a1/m~/b1` does not have a bullet and the margin is off.

<img width="201" alt="Screenshot 2024-12-24 at 9 35 54 AM" src="https://github.com/user-attachments/assets/2373d0a6-d6ea-4953-9eeb-62b6e78b0bec" />

## Expected Behavior

`a/a1/m~/b1` should have a bullet and the margin should be the same as `a/a1/m~/a1`.

`=table` should not be applied on contexts in the context view.

<img width="201" alt="Screenshot 2024-12-24 at 9 36 18 AM" src="https://github.com/user-attachments/assets/909d591c-5115-4402-a10d-1e66f36d19b1" />

## Testing

Add an RTL test.

PR #2736, opened 2024-12-26
=table should not be applied on context

RED-ROSE515:
This PR resolves the issue : Context View: =table should not be applied on context #2724.

In this PR, I have modified the `useHideBullet` to not hide bullets(=table) in context view.
And I added a function to check if the current thought is in context view.
If the current thougt is in context view, show bullet.

### DIFF
--- a/src/components/Thought.tsx
+++ b/src/components/Thought.tsx
@@ -179,7 +179,16 @@ const ThoughtContainer = ({
     attributeEquals(state, head(rootedParentOf(state, parentOf(simplePath))), '=view', 'Table'),
   )
 
-  const hideBullet = useHideBullet({ children, env, hideBulletProp, isEditing, simplePath, thoughtId })
+  const isInContextView = useSelector(state => {
+    /** Check if the thought is in the context view. */
+    const checkContextView = (currentPath: Path): boolean => {
+      if (currentPath.length <= 1) return false
+      return isContextViewActive(state, currentPath) || checkContextView(parentOf(currentPath))
+    }
+    return checkContextView(path)
+  })
+
+  const hideBullet = useHideBullet({ children, env, hideBulletProp, isEditing, simplePath, isInContextView, thoughtId })
   const style = useThoughtStyle({ children, env, styleProp, thoughtId })
   const styleAnnotation = useSelector(
     state =>
@@ -436,6 +445,7 @@ const ThoughtContainer = ({
             publish={publish}
             simplePath={simplePath}
             thoughtId={thoughtId}
+            isInContextView={isInContextView}
             // debugIndex={debugIndex}
             // depth={depth}
           />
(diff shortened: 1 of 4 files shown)


src/hooks/useHideBullet.ts
```diff
@@ -42,9 +44,14 @@ const useHideBullet = ({
       thought.value !== '=grandchildren' && attributeEquals(state, head(simplePath), '=bullet', 'None')
 
     /** Returns true if the bullet should be hidden because it is in table column 1 and is not the cursor. */
-    const hideBulletTable = () =>
-      !equalPath(simplePath, state.cursor) &&
-      attributeEquals(state, head(rootedParentOf(state, simplePath)), '=view', 'Table')
+    const hideBulletTable = () => {
+      // Don't hide bullets in context view
+      if (isInContextView) return false
```
raineorshine, line 49: It would be a bit cleaner to place this condition in the existing return statement.

I try to avoid multiple consecutive return statements.

src/components/Thought.tsx
```diff
@@ -179,7 +179,16 @@ const ThoughtContainer = ({
     attributeEquals(state, head(rootedParentOf(state, parentOf(simplePath))), '=view', 'Table'),
   )
 
-  const hideBullet = useHideBullet({ children, env, hideBulletProp, isEditing, simplePath, thoughtId })
+  const isInContextView = useSelector(state => {
+    /** Check if the thought is in the context view. */
+    const checkContextView = (currentPath: Path): boolean => {
+      if (currentPath.length <= 1) return false
+      return isContextViewActive(state, currentPath) || checkContextView(parentOf(currentPath))
```
raineorshine, line 186: I don't believe this should be recursive. Only the parent thought needs to be checked to determine if a thought is in the context view.

src/components/__tests__/Bullet.ts
```diff
@@ -155,6 +158,38 @@ describe('render', () => {
     const bullets = document.querySelectorAll('[data-bullet="parent"]')
     expect(bullets.length).toBe(1)
   })
+
+  it('render bullets in context view entries even when parent has =view/Table', async () => {
```
raineorshine, line 162: Test looks good

raineorshine (changes requested):
Behavior looks good, thanks! Just a couple recommendations.

src/components/Thought.tsx
```diff
@@ -179,7 +179,16 @@ const ThoughtContainer = ({
...
 
-  const hideBullet = useHideBullet({ children, env, hideBulletProp, isEditing, simplePath, thoughtId })
+  const isInContextView = useSelector(state => {
+    /** Check if the thought is in the context view. */
+    const checkContextView = (currentPath: Path): boolean => {
+      if (currentPath.length <= 1) return false
+      return isContextViewActive(state, parentOf(currentPath))
+    }
+    return checkContextView(path)
+  })
```
raineorshine, line 189: This can. be simplified to:

```ts
const showContexts = useSelector(state => isContextViewActive(state, rootedParentOf(state, path)))
```

or even:

```ts
const showContexts = useSelector(state => isContextViewActive(state, parentOf(path)))
```

raineorshine (approved):

src/components/Thought.tsx
RED-ROSE515: I got it. I will regard that.

Issue #2733, opened 2024-12-25
Cursor up/down does not work on empty thought in sorted context


raineorshine:
## Steps to Reproduce

```
- x
  - b
  - a
  - =sort
    - Alphabetical
      - Desc
```

1. Set the cursor on `x`. 
2. Activate New Subthought Above (Meta + Shift + Enter). 
3. Move cursor up/down.

## Current Behavior

Cursor acts as if the empty thought is at the end of the list.

## Expected Behavior

Cursor up should move the cursor from the empty thought to `x`.
Cursor down should move the cursor from the empty thought to `b`.

## Testing

Add an RTL test.

## Notes

Empty thoughts are generally ignored in sorted lists. This is to avoid re-sorting the list when the user activates New Thought or New Subthought. It will only re-sort when the user starts typing in the thought.

PR #2737, opened 2024-12-26
fix Cursor up/down on empty thought in sorted context

RED-ROSE515:
This PR resolves the issue : Cursor up/down does not work on empty thought in sorted context #2733

The cause of bug : This issue caused because of mismatching results from `getAllChildrenSorted` and `getChildrenRanked`.

So, I updated the `compareReasonableDescending` function to check compareEmpty first.
And this issue resolved.

### DIFF
--- a/src/actions/__tests__/cursorDown.ts
+++ b/src/actions/__tests__/cursorDown.ts
@@ -5,6 +5,7 @@ import { importTextActionCreator as importTextAction } from '../../actions/impor
 import newSubthought from '../../actions/newSubthought'
 import newThought from '../../actions/newThought'
 import toggleContextView from '../../actions/toggleContextView'
+import newSubthoughtTopShortcut from '../../commands/newSubthoughtTop'
 import toggleSortShortcut from '../../commands/toggleSort'
 import createTestStore from '../../test-helpers/createTestStore'
 import expectPathToEqual from '../../test-helpers/expectPathToEqual'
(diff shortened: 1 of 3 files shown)


src/actions/__tests__/cursorDown.ts
```diff
@@ -88,6 +89,28 @@ describe('normal view', () => {
     const stateNew = cursorDown(store.getState())
     expectPathToEqual(stateNew, stateNew.cursor, ['a', 'm'])
   })
+
+  it('move cursor from empty thought to next thought', () => {
```
raineorshine: We can make this test description a little more specific. See 45d582df01.

raineorshine (approved):
This does seem to solve it. I'm not entirely sure that empty thoughts *should* be sorted to the top, but I will leave that for a separate PR. This might be punting the problem down the road, but at least we have the test coverage now so we are safe to make changes to the sorting in the future without introducing a regression. Thanks!

src/actions/__tests__/cursorUp.ts
```diff
@@ -171,6 +178,28 @@ describe('normal view', () => {
     const cursor = state.cursor
     expect(pathToContext(state, prevThought(state, cursor!)!)).toEqual(['a'])
   })
+
+  it('move cursor from empty thought to previous thought in context sorted in descending order', () => {
```
raineorshine, line 182: Is there a reason you did a full shortcut test instead of a reducer test? Generally reducer tests are going to be faster and more minimal because they do not require `act`.

src/actions/__tests__/cursorDown.ts
RED-ROSE515: I got it.

src/actions/__tests__/cursorUp.ts
RED-ROSE515: For accuracy, I did full shortcut test.

Issue #1801, opened 2024-02-07
splitSentence: Parenthetical


raineorshine:
Add another case to the `splitSentence` util, only for thoughts that end in `)` or `).` 

## Steps to Reproduce

```
This is a thought (and a subthought)
```

## Current Behavior

No-op

## Expected Behavior

```
- This is a thought
  - and a subthought
```

## Testing

Add a reducer test.

PR #2740, opened 2024-12-28
add new split sentence feature : parenthetical

RED-ROSE515:
This PR implements this feature : splitSentence: Parenthetical #1801

In this PR, I have implemented new parenthetical split sentence feature.
First, I check if the thought ends with parenthetical content using the regex pattern.
And capture both the main thought and the content inside parentheses.
Return them as separate thoughts after trimming whitespace.

And in this PR, I have added `insertNewSubThought` parameter to `splitSentence` function.
For about the abbreviation test case, it doesn't work with the new feature.
So,  I modified the test to match with the new parenthetical feature.

### DIFF
--- a/src/actions/__tests__/splitSentences.ts
+++ b/src/actions/__tests__/splitSentences.ts
@@ -297,8 +297,8 @@ describe('abbreviations', () => {
     const exported = splitThought(value)
 
     expect(exported).toBe(`- ${HOME_TOKEN}
-  - One.
-  - Two ( U.N.)`)
+  - One. Two
+    - U.N.`)
   })
 
   it('split thought as expected if the dot comes from an abbreviation followed by empty spaces and a quotation mark', () => {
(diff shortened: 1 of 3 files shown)


src/actions/__tests__/splitSentences.ts
```diff
@@ -297,8 +297,8 @@ describe('abbreviations', () => {
     const exported = splitThought(value)
 
     expect(exported).toBe(`- ${HOME_TOKEN}
-  - One.
-  - Two ( U.N.)`)
+  - One. Two
+    - U.N.`)
```
raineorshine, line 301: Makes sense, thanks 👍

src/actions/__tests__/splitSentences.ts
```diff
@@ -647,3 +647,31 @@ describe('complicated cases', () => {
   - IPv4: 11.11.11.111`)
   })
 })
+
+describe('parenthetical content', () => {
```
raineorshine, line 651: Tests look great and cover all relevant cases, positive and negative. Thank you!

raineorshine (approved):

PR #2741, opened 2024-12-28
Fix undo persistence test

### LINKED ISSUE #2145: Fix undo-redo tests
Two undo-redo tests are currently broken and being skipped.

https://github.com/cybersemics/em/blob/f68689f6471af07ba87db97cdd9b68bf42cdc910/src/shortcuts/__tests__/undo-redo.ts#L188-L245

mstrofbass:
This fixes the remaining undo/redo test. 

**Problem #1:** It looks like we must use fake timers if we want the `store` state to be updated based on database operations (e.g., if we use `initialize()` to reload the state). I think this is because the `thoughtspace` operations are asynchronous and don't call the store operations prior to the test ending. (I'm not sure why we didn't get other errors that made this clear.)

**Problem #2**: There appears to be some store state that isn't cleaned up between tests that persists from the previous tests and breaks this one. The other tests probably aren't effected because they all call `createTestStore`, which presumably creates a brand new store. 

This second issue is somewhat ancillary but might come up when mixing tests that do/don't require `createTestStore` and should figure out the root cause at some point. My initial guess is that we'll just need to figure out a way reverse whatever happens in `createTestStore`. 

Closes #2145

### DIFF
--- a/src/commands/__tests__/undo-redo.ts
+++ b/src/commands/__tests__/undo-redo.ts
@@ -1,3 +1,4 @@
+import { act } from 'react'
 import { archiveThoughtActionCreator as archiveThought } from '../../actions/archiveThought'
 import { clearActionCreator as clear } from '../../actions/clear'
 import { cursorBackActionCreator as cursorBack } from '../../actions/cursorBack'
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Great, thanks for the explanation. I will merge this as a solution for #2145 and we can try to spot the deeper issue with test store isolation.

Issue #1633, opened 2022-07-22
Context View: Group contexts from identical parent contexts


raineorshine:
When a thought exists in multiple contexts that have the same parent, group them together under a single ContextBreadcrumbs rather than repeating the breadcrumbs for each thought. 

This is a simple change that reduces visual clutter and does not face the more complex design implications of hierarchical contexts. 

![image](https://user-images.githubusercontent.com/750276/180343766-88a9f935-ed70-4a9b-9c3e-c3c75de87faa.jpeg)

raineorshine:
Design needed to determine how this fits with:

- editable (categorizable) context view
- editable breadcrumbs
- expandable breadcrumbs

PR #2743, opened 2024-12-28
Feature: Context View: Group contexts in breadcrumbs hierarchy

raineorshine:
#1633


raineorshine:
The current context view that renders contexts as a flat list proved to be a poor fit for hierarchical grouping. Closing in favor of a fully hierarchical approach using the existing linearizeTree algorithm.

PR #2744, opened 2024-12-28
Third bunch of possible undefined thoughts

### LINKED ISSUE #2711: getThoughtById: Return undefined when thought does not exist

snqb:
#2711

### DIFF
--- a/src/actions/formatLetterCase.ts
+++ b/src/actions/formatLetterCase.ts
@@ -17,7 +17,9 @@ export const formatLetterCaseActionCreator =
     if (!cursor) return
 
     const thought = pathToThought(state, cursor)
-    const originalThoughtValue = thought.value
+    const originalThoughtValue = thought?.value
+
+    if (originalThoughtValue === undefined) return
 
     const updatedThoughtValue = applyLetterCase(command, originalThoughtValue)
     const simplePath = simplifyPath(state, cursor)
--- a/src/actions/formatLetterCase.ts
+++ b/src/actions/formatLetterCase.ts
@@ -16,18 +16,18 @@ export const formatLetterCaseActionCreator =
     const cursor = state.cursor
     if (!cursor) return
 
-    const originalThoughtValue = pathToThought(state, cursor)?.value
-    if (originalThoughtValue === undefined) return
+    const thought = pathToThought(state, cursor)
+    if (!thought) return state
 
-    const updatedThoughtValue = applyLetterCase(command, originalThoughtValue)
+    const oldValue = thought.value
+    const newValue = applyLetterCase(command, oldValue)
     const simplePath = simplifyPath(state, cursor)
-
     const offset = selection.offsetThought()
 
     dispatch(
       editThought({
-        oldValue: originalThoughtValue,
-        newValue: updatedThoughtValue,
+        oldValue,
+        newValue,
         path: simplePath,
         force: true,
       }),
--- a/src/actions/formatLetterCase.ts
+++ b/src/actions/formatLetterCase.ts
@@ -16,9 +16,7 @@ export const formatLetterCaseActionCreator =
     const cursor = state.cursor
     if (!cursor) return
 
-    const thought = pathToThought(state, cursor)
-    const originalThoughtValue = thought?.value
-
+    const originalThoughtValue = pathToThought(state, cursor)?.value
     if (originalThoughtValue === undefined) return
 
     const updatedThoughtValue = applyLetterCase(command, originalThoughtValue)
--- a/src/hooks/useDragAndDropSubThought.ts
+++ b/src/hooks/useDragAndDropSubThought.ts
@@ -125,15 +125,18 @@ const drop = (props: DroppableSubthoughts, monitor: DropTargetMonitor) => {
     return
   }
 
-  if (thoughtTo) {
-    store.dispatch(
-      moveThought({
-        oldPath: thoughtsFrom,
-        newPath: pathTo,
-        newRank: (dropTop ? getPrevRank : getNextRank)(state, thoughtTo.id),
-      }),
-    )
+  if (!thoughtTo) {
+    console.warn(`Cannot drop ${thoughtFrom} on itself. Aborting drop.`)
+    return
   }
+
+  store.dispatch(
+    moveThought({
+      oldPath: thoughtsFrom,
+      newPath: pathTo,
+      newRank: (dropTop ? getPrevRank : getNextRank)(state, thoughtTo.id),
+    }),
+  )
 
   // alert user of move to another context
   if (!sameContext && thoughtTo && thoughtFrom) {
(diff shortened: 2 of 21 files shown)


snqb:
a couple test cases failing, will fix later

src/selectors/parentOfThought.ts
```diff
@@ -8,7 +8,7 @@ const parentOfThought = (state: State, thoughtId: ThoughtId): Thought | null =>
   const thought = getThoughtById(state, thoughtId)
   if (!thought) return null
   const parentThought = getThoughtById(state, thought.parentId)
-  return parentThought
+  return parentThought ?? null
```
snqb, line 11: preserve fn signature

src/stores/commandStateStore.ts
```diff
@@ -36,7 +36,7 @@ export const updateCommandState = () => {
   const action =
     selection.isActive() && selection.isOnThought()
       ? getCommandState(selection.html() ?? '')
-      : getCommandState(pathToThought(state, state.cursor).value)
+      : getCommandState(pathToThought(state, state.cursor)?.value ?? '')
```
snqb, line 39: on that end we have a `path.length === 0` default case, so fallback is fine here

src/util/importJSON.ts
```diff
@@ -233,10 +234,10 @@ const importJSON = (
   { lastUpdated = timestamp(), updatedBy = clientId, skipRoot = false }: ImportJSONOptions = {},
 ) => {
   const destThought = pathToThought(state, simplePath)
-  const destEmpty = destThought.value === '' && !anyChild(state, head(simplePath))
+  const destEmpty = destThought?.value === '' && !anyChild(state, head(simplePath))
   // use getNextRank instead of getRankAfter because if dest is not empty then we need to import thoughts inside it
-  const rankStart = destEmpty ? destThought.rank : getNextRank(state, head(simplePath))
-  const rankIncrement = getRankIncrement(state, blocks, destThought, rankStart)
+  const rankStart = destEmpty ? destThought?.rank : getNextRank(state, head(simplePath))
+  const rankIncrement = destEmpty ? getRankIncrement(state, blocks, destThought, rankStart) : 1
```
snqb, line 240: 1 is a fallback in getRankIncrement too, so that's my judgment

src/test-helpers/deleteThoughtAtFirstMatch.ts
```diff
@@ -19,6 +19,8 @@ const getThoughtAndParentPath = (state: State, at: string[]): [Thought, Path] =>
 
   const thought = pathToThought(state, path)
 
+  if (!thought) throw new Error(`Thought not found for path: ${path}`)
```
snqb, line 22: it's a test file, and we throw an error for empty path above, so thought why not

src/test-helpers/contextToThought.ts
```diff
@@ -9,7 +9,7 @@ import getThoughtById from '../selectors/getThoughtById'
  */
 const contextToThought = (state: State, context: Context): Thought | null => {
   const id = contextToThoughtId(state, context)
-  return id ? getThoughtById(state, id) : null
+  return id ? (getThoughtById(state, id) ?? null) : null
```
snqb, line 12: to preserve type for now

but there's an itch to make it `Thought | null | undefined`
wdyt?

src/util/importJSON.ts
```diff
@@ -63,6 +63,7 @@ const insertThought = (
   },
 ) => {
   const thoughtOld = getThoughtById(state, id)
+  if (!thoughtOld) return
```
raineorshine, line 66: Let's return `null` here since `insertUpdates` is nullable below.

raineorshine (changes requested):
Thanks!

src/actions/formatLetterCase.ts
```diff
@@ -17,7 +17,9 @@ export const formatLetterCaseActionCreator =
     if (!cursor) return
 
     const thought = pathToThought(state, cursor)
-    const originalThoughtValue = thought.value
+    const originalThoughtValue = thought?.value
+
+    if (originalThoughtValue === undefined) return
```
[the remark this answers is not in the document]
snqb: I'm not sure if I got you right, but here's what I've ended up with

```typescript
const cursor = state.cursor
if (!cursor) return

const originalThoughtValue = pathToThought(state, cursor)?.value
if (originalThoughtValue === undefined) return
```

src/test-helpers/contextToThought.ts
snqb: ok, I made it `Thought | undefined`, and it broke nothing

src/actions/formatLetterCase.ts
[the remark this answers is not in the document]
raineorshine: `originalThoughtValue` can only be undefined due to the optional chaining operator. It doesn't really make sense to use it here in order to check for undefined, when what you really want to check is if the thought is undefined. `thought.value` is always a string. There's no reason to get it mixed up with the undefined thought check.

```
const thought = pathToThought(state, cursor)
if (!thought) return
```

src/hooks/useDragAndDropSubThought.ts
```diff
@@ -118,13 +118,18 @@ const drop = (props: DroppableSubthoughts, monitor: DropTargetMonitor) => {
...
     store.dispatch(
       error({ value: `Cannot move the ${isEM(thoughtsFrom) ? 'em' : 'home'} context to another context.` }),
     )
     return
   }
 
+  if (!thoughtTo) {
+    console.warn(`Cannot drop ${thoughtFrom} on itself. Aborting drop.`)
+    return
+  }
```
raineorshine, line 131: This `console.warn` is not accurate, as dropping a thought on itself is already detected and short circuited using the `equalPath` check above. Please update the message or remove the `console.warn` completely, combining it with the first return statement.

src/hooks/useDragAndDropSubThought.ts
```diff
@@ -134,7 +139,7 @@ const drop = (props: DroppableSubthoughts, monitor: DropTargetMonitor) => {
   )
 
   // alert user of move to another context
-  if (!sameContext) {
+  if (!sameContext && thoughtTo && thoughtFrom) {
```
raineorshine, line 142: This isn't necessary due to the short circuit above.

raineorshine (changes requested):

Issue #2729, opened 2024-12-25
Help: Question mark gesture


raineorshine:
Add a question mark gesture to the Help command: →↓←↓

Unlike other commands, any gesture ending with →↓←↓ should open Help.

The question mark gesture should always be displayed in the gesture menu, at the end. 

i.e. 

→↓←↓ 
↑→↓←↓
↑↑→↓←↓
↑↑↑→↓←↓
...

PR #2745, opened 2024-12-29
Add question mark gesture to Help command

RED-ROSE515:
This PR implements the new feature : [mobile] Add question mark gesture to Help command. #2729

In this PR, I implented the code like this.
In `commands.ts`'s `handleGestureEnd` function, I first check if the sequence ends with the help gesture (rdld).
If it does end with rdld, I use the help command regardless of what came before.
If it doesn't end with rdld, I use the normal command lookup logic.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -221,7 +221,8 @@ export const inputHandlers = (store: Store<State, any>) => ({
   handleGestureEnd: ({ sequence, e }: { sequence: GesturePath | null; e: GestureResponderEvent }) => {
     const state = store.getState()
 
-    // Get the help command and its gesture
+    // Get the command from the command gesture index.
+    // When the command palette  is displayed, disable gesture aliases (i.e. gestures hidden from instructions). This is because the gesture hints are meant only as an aid when entering gestures quickly.
     const helpCommand = commandIdIndex['help']
     const helpGesture = helpCommand?.gesture as string
 
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -151,6 +151,13 @@ let commandPaletteGesture: number | undefined
 
 const { commandKeyIndex, commandIdIndex, commandGestureIndex } = index()
 
+/** Gets the canonical gesture of the command as a string, ignoring aliases. Returns an empty string if the command does not have a gesture. */
+export const gestureString = (command: Command): string =>
+  (typeof command.gesture === 'string' ? command.gesture : command.gesture?.[0] || '') as string
+
+/** Get a command by its id. */
+export const commandById = (id: CommandId): Command => commandIdIndex[id]
+
 /**
  * Keyboard and gesture handlers factory function that binds the store to event handlers.
  *
@@ -223,8 +230,9 @@ export const inputHandlers = (store: Store<State, any>) => ({
 
     // Get the command from the command gesture index.
     // When the command palette  is displayed, disable gesture aliases (i.e. gestures hidden from instructions). This is because the gesture hints are meant only as an aid when entering gestures quickly.
-    const helpCommand = commandIdIndex['help']
-    const helpGesture = helpCommand?.gesture as string
+
+    const helpCommand = commandById('help')
+    const helpGesture = gestureString(helpCommand)
 
     // If sequence ends with help gesture, use help command
     // Otherwise use the normal command lookup
@@ -323,10 +331,3 @@ export const inputHandlers = (store: Store<State, any>) => ({
     }
   },
 })
-
-/** Gets the canonical gesture of the command as a string, ignoring aliases. Returns an empty string if the command does not have a gesture. */
-export const gestureString = (command: Command): string =>
-  (typeof command.gesture === 'string' ? command.gesture : command.gesture?.[0] || '') as string
-
-/** Get a command by its id. */
-export const commandById = (id: CommandId): Command => commandIdIndex[id]
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -221,12 +221,17 @@ export const inputHandlers = (store: Store<State, any>) => ({
   handleGestureEnd: ({ sequence, e }: { sequence: GesturePath | null; e: GestureResponderEvent }) => {
     const state = store.getState()
 
-    // Get the command from the command gesture index.
-    // When the command palette  is displayed, disable gesture aliases (i.e. gestures hidden from instructions). This is because the gesture hints are meant only as an aid when entering gestures quickly.
-    const command =
-      !state.showCommandPalette || !commandGestureIndex[sequence as string]?.hideFromHelp
+    // Get the help command and its gesture
+    const helpCommand = commandIdIndex['help']
+    const helpGesture = helpCommand?.gesture as string
+
+    // If sequence ends with help gesture, use help command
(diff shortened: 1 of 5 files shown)


src/commands.ts
```diff
@@ -221,10 +221,15 @@ export const inputHandlers = (store: Store<State, any>) => ({
   handleGestureEnd: ({ sequence, e }: { sequence: GesturePath | null; e: GestureResponderEvent }) => {
     const state = store.getState()
 
-    // Get the command from the command gesture index.
-    // When the command palette  is displayed, disable gesture aliases (i.e. gestures hidden from instructions). This is because the gesture hints are meant only as an aid when entering gestures quickly.
```
raineorshine, line 225: I think this comment was deleted accidentally. It explains the use of `hideFromHelp`.

raineorshine (changes requested):
Thanks! This is a great start. I can confirm that the gesture works correctly. `handleGestureEnd` seems like a reasonable place to do this.

Just one more requirement. I see that this is a draft so you are probably working on it!

> The question mark gesture should always be displayed in the gesture menu, at the end.

## Steps to Reproduce

Swipe ↓ and hold.

## Current Behavior

The gesture menu only shows Move Thought Down.

## Expected Behavior

The help command should always be displayed at the end of the gesture menu since it can always be activated.

src/commands.ts
RED-ROSE515: I got it.

raineorshine (changes requested):
Looks good! We just need to preserve the swipe highlight and selection behavior. There are two cases to note here.

# 1. Partially entered

## Steps to Reproduce

Swipe ←↑ and hold.

## Current Behavior

The Help command is correctly shown at the end of the gesture menu, but the first two swipes (→↓) are incorrectly highlighted.



## Expected Behavior

The Help command gesture diagram should not have any swipes highlighted here, since the gesture in progress ←↑  does not match the beginning of of the help gesture →↓←↓. Only the swipes that have been input should be highlighted, as you can see in the other commands. This will require some special handling to support Help's unique ability to be entered at the end of any gesture.

e.g.

- If the gesture in progress is →↓, then nothing should be highlighted in the Help gesture diagram.
- If the gesture in progress is →, then → should be highlighted.
- If the gesture in progress is →↓, then →↓  should be highlighted.
- If the gesture in progress is ←→, then → should be highlighted.
- If the gesture in progress is ←→↓, then →↓  should be highlighted.
- If the gesture in progress is ←→↓←, then →↓←  should be highlighted.
- If the gesture in progress is ←→↓←↓, then →↓←↓  should be highlighted.
- etc

raineorshine:
# 2. Selected

## Steps to Reproduce

Swipe ←↑→↓←↓ and hold.

## Current Behavior

The Help command gesture diagram is highlighted correctly, but the command item is not shown as selected.

Screenshot 2024-12-30 at 12 37 40 PM

## Expected Behavior

The Help command item should be selected since the gesture in progress ends in →↓←↓ and releasing will activate Help.

For example, this is what the Subcategorize item looks like when selected:

Screenshot 2024-12-30 at 12 14 17 PM copy

src/commands.ts
```diff
@@ -223,8 +223,14 @@ export const inputHandlers = (store: Store<State, any>) => ({
 
     // Get the command from the command gesture index.
     // When the command palette  is displayed, disable gesture aliases (i.e. gestures hidden from instructions). This is because the gesture hints are meant only as an aid when entering gestures quickly.
-    const command =
-      !state.showCommandPalette || !commandGestureIndex[sequence as string]?.hideFromHelp
+    const helpCommand = commandIdIndex['help']
```
raineorshine, line 226: Let's use `commandById` rather than accessing `commandIdIndex` directly. We already have `commandById`, so we might as well use and it and hide `commandIdIndex` as an implementation detail.

https://github.com/cybersemics/em/blob/75b8c516e70922bdfa44d29a8cc47e33c942e865/src/commands.ts#L325-L326

src/commands.ts
```diff
@@ -223,8 +223,14 @@ export const inputHandlers = (store: Store<State, any>) => ({
...
+    const helpGesture = helpCommand?.gesture as string
```
raineorshine, line 227: You can use the `gestureString` helper to get the normalized gesture string:

https://github.com/cybersemics/em/blob/75b8c516e70922bdfa44d29a8cc47e33c942e865/src/commands.ts#L321-L323

src/components/CommandPalette.tsx
```diff
@@ -197,7 +204,22 @@ const CommandRow: FC<{
...
             <GestureDiagram
               color={disabled ? token('colors.gray') : undefined}
-              highlight={!disabled ? gestureInProgress.length : undefined}
+              highlight={
+                !disabled
+                  ? shortcut.id === 'help'
+                    ? // For help command, only highlight the matching portion at the end
+                      (() => {
+                        const helpGesture = gestureString(shortcut)
+                        const subsequences = Array.from({ length: helpGesture.length }, (_, i) =>
```
raineorshine, line 213: This is acceptable, but note that there is a way to do this more efficiently by iterating from the end of the gesture. You shouldn't have to generate all the subsequences.

src/components/CommandPalette.tsx
```diff
@@ -421,18 +447,26 @@ const CommandPalette: FC = () => {
...
-                onClick={onExecute}
-                onHover={onHover}
-                selected={!isTouch ? shortcut === selectedShortcut : gestureInProgress === shortcut.gesture}
-                shortcut={shortcut}
-              />
-            ))}
+            {shortcuts.map(shortcut => {
+              // Check if the current gesture sequence ends with help gesture
+              const isHelpMatch =
+                shortcut.id === 'help' && (gestureInProgress as string)?.toString().endsWith(helpGesture)
```
raineorshine, line 453: Let's use [gestureString](https://github.com/cybersemics/em/blob/75b8c516e70922bdfa44d29a8cc47e33c942e865/src/commands.ts#L321-L323) and [commandById](https://github.com/cybersemics/em/blob/75b8c516e70922bdfa44d29a8cc47e33c942e865/src/commands.ts#L325-L326) here as well.

src/components/CommandPalette.tsx
```diff
@@ -291,6 +313,10 @@ const CommandPalette: FC = () => {
 
   const [selectedShortcut, setSelectedShortcut] = useState<Command>(shortcuts[0])
 
+  // Get help command info for gesture matching
+  const helpCommand = useMemo(() => globalCommands.find((cmd: Command) => cmd.id === 'help'), [])
```
raineorshine, line 317: We have [commandById](https://github.com/cybersemics/em/blob/75b8c516e70922bdfa44d29a8cc47e33c942e865/src/commands.ts#L325-L326) which operates in O(1) rather than O(n).

This can either be inlined below or taken completely outside the component and put at the module level. We shouldn't have to store or memoize it at the component level since it is not connected to the component's lifecycle.

raineorshine (changes requested):
Thank you! Both cases are solved and it works perfectly now. I just have a few suggestions below about using the existing helpers for accessing the help gesture by id.

PR #2746, opened 2024-12-30
V1.5 Alert

yangchristina:


src/colors.config.ts
```diff
@@ -65,6 +65,8 @@ const colors = {
     exportTextareaColor: 'rgba(170, 170, 170, 1)', // #aaa, also used in anchorButton
     inherit: 'inherit',
     currentColor: 'currentColor',
+    alertBorder: 'rgba(36, 36, 36, 1)',
+    alertBg: 'rgba(23, 23, 23, 1)', // #171717
```
raineorshine, line 69: I'd like to start making design tokens more general, like `bg` and `fg`, so that they can be easily reused. This will also make theming easier. What do you think about `panelBg` and `panelBorder`? This would describe any component that has a background color and border that are slightly off from `bg`.

src/e2e/puppeteer/__tests__/render-thoughts.ts
```diff
@@ -167,7 +168,7 @@ describe('Font Size: 13', () => {
     await click('[data-testid=decrease-font]') // 13
 
     // close alert
-    await click('[data-testid=close-button]')
+    await hide('[data-testid=alert]')
```
raineorshine, line 171: Could we add an ✗ that appears in the upper right corner on hover to dismiss the alert? I realize the mockup did not account for the close functionality on desktop.

I'm okay if this is + time.

raineorshine (changes requested):
Thanks for your submission!

The main Alert component is represented by https://github.com/cybersemics/em/blob/main/src/components/Popup.tsx. https://github.com/cybersemics/em/blob/main/src/components/Alert.tsx is a HOC that mainly just handles fade in/out. By circumventing Popup.tsx we're losing some key functionality:

- Cross-browser support for position: fixed
- Swipe-to-dismiss functionality
- Multicursor support
- Import Files special handling.

The goal is to re-style the alert without losing any of the existing functionality. I would suggest duplicating https://github.com/cybersemics/em/blob/main/src/components/Popup.tsx and using that as a starting point. The only other component that uses it is https://github.com/cybersemics/em/blob/42b1ef55abd5457ecbbbc39512be3dc6a1682fb5/src/components/CommandPalette.tsx#L469, and that will be re-styled as well so we could make them separate components to decouple them.

Let me know if this makes sense!

yangchristina:
> Thanks for your submission!
> 
> The main Alert component is represented by https://github.com/cybersemics/em/blob/main/src/components/Popup.tsx. https://github.com/cybersemics/em/blob/main/src/components/Alert.tsx is a HOC that mainly just handles fade in/out. By circumventing Popup.tsx we're losing some key functionality:
> 
> * Cross-browser support for position: fixed
> * Swipe-to-dismiss functionality
> * Multicursor support
> * Import Files special handling.
> 
> The goal is to re-style the alert without losing any of the existing functionality. I would suggest duplicating https://github.com/cybersemics/em/blob/main/src/components/Popup.tsx and using that as a starting point. The only other component that uses it is
> 
> https://github.com/cybersemics/em/blob/42b1ef55abd5457ecbbbc39512be3dc6a1682fb5/src/components/CommandPalette.tsx#L469
> 
> , and that will be re-styled as well so we could make them separate components to decouple them.
> Let me know if this makes sense!

Should font size and padding still follow these?
```js
  const padding = useSelector(state => state.fontSize / 2 + 2)
 ```

File: src/e2e/puppeteer/__tests__/render-thoughts.ts (hunk 2 above)
Comment (yangchristina) [reply src/e2e/puppeteer/__tests__/render-thoughts.ts:171]: Yes I can add it.

**Comment by yangchristina** [comment]
I'll update the puppeteer snapshot once the styling has been confirmed.

**Comment by raineorshine** [comment]
> Should font size and padding still follow these?
> 
> ```js
>   const fontSize = useSelector(state => state.fontSize)
>   const padding = useSelector(state => state.fontSize / 2 + 2)
> ```

Yes, the alert should scale with `state.fontSize` and its padding should be proportionate. i.e. the ratio of fontSize to padding in the Figma file should be preserved.

@trevinhofmann @yangchristina Do either of you have an opinion about using `transform: scale(...)` vs adjusting `fontSize` and `padding`? The goal is for our UI components to look good at different font sizes. "Font size" is a bit of a misnomer as it affects more than just the font size. It is a user setting that controls the overall size of the UI to fit the user's preference.

**Comment by trevinhofmann** [comment]
> Do either of you have an opinion about using `transform: scale(...)` vs adjusting `fontSize` and `padding`?

Would the `transform: scale(...)` be applied to ~everything?

Initially, I lean more toward adjusting `fontSize`, `padding`, etc. to have more control and avoid unintended side effects. However, my opinion on this is somewhat uninformed and I'm not sure what the unintended side effects may be, unless you would like me to take some time exploring it further.

**Comment by raineorshine** [comment]
> > Do either of you have an opinion about using `transform: scale(...)` vs adjusting `fontSize` and `padding`?
> 
> Would the `transform: scale(...)` be applied to ~everything?

At the moment I'm just considering applying it to the Alert component, but it's worth considering if it should be applied to other UI components, like modals, command palette, settings, etc. However most of them already scale the font size and specify padding in em units.

> Initially, I lean more toward adjusting `fontSize`, `padding`, etc. to have more control and avoid unintended side effects. However, my opinion on this is somewhat uninformed and I'm not sure what the unintended side effects may be, unless you would like me to take some time exploring it further.

Yes, that's what I'm thinking.

The main thing I can think of is that `transform` does not affect the document flow. With the Alert component it may not matter as long as the transform origin is set correctly, but with other components it could break their positioning on the screen.

> Would "Zoom" be a better name, with 1.0 as the default? The setting might need a different input type (selection from a dropdown?), if that were to be adopted.

I had considered it, but my designer didn't think it was ideal since "Zoom" suggests that you are narrowing the field of focus, like a camera zoom. Of course, "Zoom" in Chrome browser settings does change the size of everything without narrowing the field of focus. So I guess there are two different semantics floating around. Probably deserves some user testing to determine how the average person understands it.

**Comment by yangchristina** [comment]
> At the moment I'm just considering applying it to the Alert component, but it's worth considering if it should be applied to other UI components, like modals, command palette, settings, etc. However most of them already scale the font size and specify padding in em units.

It would be fine for the Alert component, since it doesn't affect the layout of any other elements, and it would ensure that if anything else was added to the alert component, it would scale appropriately.

Like @trevinhofmann, I don't know a lot about this topic, so I can't say much otherwise.

However, I also don't see the problem with using em units.

**Comment by raineorshine** [comment]
Let's go with scaling the `fontSize` and em units for margin and padding. If we use `scale` on the Alert component but not other components, I think it's going to cause small aspects like borders to be inconsistent.

**Comment by yangchristina** [comment]
> Let's go with scaling the `fontSize` and em units for margin and padding. If we use `scale` on the Alert component but not other components, I think it's going to cause small aspects like borders to be inconsistent.

Sounds good!

**Review by trevinhofmann — CHANGES_REQUESTED** [review]
Thanks for the PR, @yangchristina!

Quick note: Please include a PR description explaining the goals of the change. I have the design document for these UI/UX changes, but it would still be nice to know a bit more about which particular parts the PR is changing.

---

## 1. The "x" button does not close the alert

The "x" button appears correctly on hover (although it's a bit small, IMO), but clicking it does not dismiss the alert popup.

https://github.com/user-attachments/assets/6657df75-b526-43f2-9da5-39b7bcb3c6a3

---

## 2. Confusing cursor pointer on icons

This might be intentional and a non-issue, but it was confusing to me that the cursor changes to a pointer when hovering over the icon. Normally, the pointer is reserve for clickable items that cause an action, but this icon is not clickable.

https://github.com/user-attachments/assets/1adecc07-8961-4efe-8246-1575ed5fe04d

---

On mobile, I have a few questions that I would just like to clarify (cc @raineorshine).

A) Should the alert be centered, or is it intended to be on the right side of the screen? I don't see a mobile design for the alert in Figma.
B) Should the alert still fade out / dismiss while I'm actively dragging it?
C) Should the alert be dismissed by dragging it up/down, or left/right? I'm fine with either, but my initial "training" as a user made me assume that it would be dismissed by dragging it left/right.

https://github.com/user-attachments/assets/40b0acdf-7f10-4de1-8f79-cd953b8b3c0f

File: src/components/Alert.tsx [hunk 3]
```diff
@@ -37,37 +37,9 @@ const Alert: FC = () => {
...
-        <RedoIcon
-          size={iconSize}
-          fill={token('colors.bg')}
-          cssRaw={css.raw({ position: 'relative', top: '0.25em', left: '0.25em' })}
-        />
-      </a>
-    </div>
-  ) : null
+  const Icon =
+    alert?.alertType && alert.alertType in alertToIcon ? alertToIcon[alert.alertType as keyof typeof alertToIcon] : null
```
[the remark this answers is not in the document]
yangchristina: I've cleaned it up a bit, is this better?

yangchristina:
@trevinhofmann Here's the [figma](https://www.figma.com/design/g0osrpkb5dhY0qPxW9mWjB/Em-UX?node-id=20-244&t=CKREofqyquLlCP1v-4), it is supposed to centred.

There aren't mocks for the 'x', so I didn't know how large to make it. I can make it larger.

yangchristina:
@raineorshine for the 'x' button, should it only show when `alert?.showCloseLink` is true?

yangchristina:
Screenshot 2025-01-07 at 9 25 16 PM

Does this look around right? The icon in the mocks is thinner than the text, but `CloseButton` uses text instead of an svg icon.

Spacing wise, using text is less exact.
Screenshot 2025-01-07 at 9 27 15 PM

@raineorshine what do you think of switching the close button to an svg icon instead of the text character?

raineorshine:
That does look close to the mockup, although I didn't anticipate how it would look with a single-line alert. I don't think the vertical centering looks great for a close button, which is expected to be in the upper right.

I was thinking more like this:

401014607-e5df5427-0e50-4dd7-ae9b-bcb7703a45ce copy 2

Or we could move it to the corner in a circle, which I think looks a bit better:

401014607-e5df5427-0e50-4dd7-ae9b-bcb7703a45ce

I could ask Ahmad to do a mockup of the latter so you have more exact proportions to work with.

I'm fine with using SVG for it if that helps.

yangchristina:
> That does look close to the mockup, although I didn't anticipate how it would look with a single-line alert. I don't think the vertical centering looks great for a close button, which is expected to be in the upper right.
> 
> I was thinking more like this:
> 
> 401014607-e5df5427-0e50-4dd7-ae9b-bcb7703a45ce copy 2
> 
> Or we could move it to the corner in a circle, which I think looks a bit better:
> 
> 401014607-e5df5427-0e50-4dd7-ae9b-bcb7703a45ce
> 
> I could ask Ahmad to do a mockup of the latter so you have more exact proportions to work with.
> 
> I'm fine with using SVG for it if that helps.

In that case will the mockup for tip be changed too?

raineorshine:
I don't see any difference between Tip and Alert, so they should probably be the same component. It just need to support HTML and icons.

yangchristina:
https://github.com/user-attachments/assets/70f01bc4-131f-46df-9d6d-f764e02e3e43

https://github.com/user-attachments/assets/3a8c2f69-c484-4a10-a823-890f7358c1a1

src/hooks/usePositionFixed.ts
```diff
@@ -39,7 +43,7 @@ const usePositionFixed = (): {
     position: position ?? 'fixed',
     overflowX: position === 'absolute' ? 'hidden' : 'visible',
     /* spacing.safeAreaTop applies for rounded screens */
-    top: position === 'absolute' ? `${scrollTop}px` : token('spacing.safeAreaTop'),
+    top: disableTop ? undefined : position === 'absolute' ? `${scrollTop}px` : token('spacing.safeAreaTop'),
```
yangchristina, line 46: @raineorshine @trevinhofmann Not sure what the right way to do this should be. I've anchored it to the bottom with `bottom: '36px',`, so I can't have to `top` style here.

raineorshine: `usePositionFixed` currently only supports `position: fixed` elements on the top of the screen, so we'll need to add support for elements fixed to the bottom of the screen. 

This is a fallback for iOS Safari in which `position: fixed` does not work when the keyboard is up. That’s why we have to emulate it with `position: absolute` + `scrollTop`. 

To position an element absolutely at the bottom of the screen, it will use something like `scrollTop + viewportHeight`.

Let me know if I can clarify anything further.

yangchristina: Is this okay? Not really sure how to test on ios safari.

trevinhofmann:
@yangchristina - Thank you very much for making changes. Please let me know if (or when) the PR is ready for review, and I'll be sure to take a look right away.

yangchristina:
> @yangchristina - Thank you very much for making changes. Please let me know if (or when) the PR is ready for review, and I'll be sure to take a look right away.

It's ready to review, thanks for looking at it!

src/hooks/usePositionFixed.ts
raineorshine: The easiest way to test it is to remove the `isTouch && isSafari() && !isIOS` condition. The `position: absolute` fallback is only necessary on Safari, but it does work on other platforms for testing purposes. If the toolbar jitters during scroll you know it is using the `position: absolute` fallback.

yangchristina: It works for the most part, but there is a bug that occurs when you very quickly toggle font sizes, where the alert gets stuck on the bottom. I thought it could be related to the fact that I'm using `top` when position absolute and `fromBottom`, which doesn't take into account the height of the alert.

I tried to fix it with this:
```
if (position === 'absolute') {
    if (fromBottom) {
      const fullHeight = Math.max(
        document.body.scrollHeight,
        document.body.offsetHeight,
        document.documentElement.scrollHeight,
        document.documentElement.getBoundingClientRect().height,
      )
      bottom = `${fullHeight - window.innerHeight - scrollTop + offset}px`
    } else {
      top = `${scrollTop + offset}px`
    }
  }
```

However, html and body both do not span the entire scroll, not sure why.
Screenshot 2025-01-14 at 11 12 58 AM

raineorshine: What if you subscribe the hook to `state.fontSize` to force a re-render?

src/components/Alert.tsx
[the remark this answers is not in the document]
trevinhofmann: Thank you! I do like the new type much more.

trevinhofmann (changes requested):
Thank you for the updates, @yangchristina! The new `x` close button looks great and works as expected. The issues `1. The "x" button does not close the alert` and `2. Confusing cursor pointer on icons` are both resolved 👍

---

> > B) Should the alert still fade out / dismiss while I'm actively dragging it?
>
> I don't think it needs to fade out while actively dragging it. I'm basing this on the current behavior in main.

@raineorshine -- Apologies, I don't think I phrased that clearly. What I am seeing now is: If I press down on an alert and start dragging it to dismiss it, it will soon disappear due to the timeout even while I am still pressing down and dragging it. Should it instead remain active/visible while I am still holding/dragging it?

---

## 3. The alert is not centered on mobile

On mobile, the alert is still positioned on the right side of the screen rather than being centered.

foo-center

## 4. The alert is sometimes displayed too low on mobile

There are certain cases where the alert is not positioned correctly at the bottom on mobile. It is shifted down, partially off of the visible screen. In the demo below, I was able to trigger this by clicking in the empty portion of the thoughtspace after the alert was open.

https://github.com/user-attachments/assets/108ff279-5804-486d-95b7-0f88cf01b02a

## 5. The alert can only be swiped up to dismiss

It should be possible to dismiss an alert on mobile by dragging it down. At the moment, it can only be dismissed by dragging it up.

> Since it shows up at the bottom of the screen, my tendency is to dismiss it by dragging it down.

https://github.com/user-attachments/assets/bb25f2a1-7d9a-440e-adf0-de6b1f4ae5ef

yangchristina:
@trevinhofmann How did you run the mobile version? Is it the capacitor version?

Web but smaller screen size:
Screenshot 2025-01-16 at 8 20 42 PM

src/hooks/usePositionFixed.ts
yangchristina: Maybe but I think I still need to deal with height for edge cases like largest font size. Will just pass in a height prop.

yangchristina:
> @trevinhofmann How did you run the mobile version? Is it the capacitor version?
> 
> Web but smaller screen size: Screenshot 2025-01-16 at 8 20 42 PM

Actually I think this may be due to the `transform` property on `useSwipeToDismissProps` interfering with the `transform` used to center the alert.

yangchristina:
@trevinhofmann The alert should be centered now. Do you mind checking?

> ## 4. The alert is sometimes displayed too low on mobile

This should also be fixed now.

yangchristina:
@raineorshine Can you please go over the requirements for swipe to dismiss? I haven't changed anything for that functionality, and assumed previously that it was already implemented. It is, than perhaps it could be a change in `usePositionFixed` interfering with it.

raineorshine:
The only requirement of `swipeToDismiss` that’s relevant here is that it should swipe down to dismiss instead of up. That should be a simple configuration change, as we use a third-party plugin. 

I recommend checking the props and styles that `swipeToDismiss` applies to ensure that they are not being overridden. You may need to take a different approach to centering the alert or wrap the element in a div so that you can apply styles independently.

yangchristina:
Managed to run it on ios. Changed it to swipe down.
Screenshot 2025-01-18 at 7 51 06 PM

src/components/PopupBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+import useCombinedRefs from '../hooks/useCombinedRefs'
+import usePositionFixed from '../hooks/usePositionFixed'
+import useSwipeToDismiss from '../hooks/useSwipeToDismiss'
+import syncStatusStore from '../stores/syncStatus'
+import fastClick from '../util/fastClick'
+import CloseButton from './CloseButton'
+
+export type PopupBaseProps = PropsWithChildren<
+  {
+    // used to cancel imports
```
trevinhofmann, line 19: Changing this to a JSDoc comment like the others will include this documentation in editors for features like autocomplete.

```suggestion
    /** used to cancel imports */
```

trevinhofmann (changes requested):

All of the following issues appear to be resolved:
* `1. The "x" button does not close the alert`
* `2. Confusing cursor pointer on icons`
* `3. The alert is not centered on mobile`
* `4. The alert is sometimes displayed too low on mobile`
* `5. The alert can only be swiped up to dismiss`

---

## 6. Gesture alerts sometimes fade immediately

**Edit: This issue exists in `main`, so a separate issue has been created: https://github.com/cybersemics/em/issues/2784**

When a gesture is made slowly, the resulting notification (e.g. `New Thought`) is displayed for the normal amount of time. However, when quickly making a gesture, the notification appears and disappears almost immediately.

https://github.com/user-attachments/assets/02f60a27-a60d-4623-895b-0f38a0481a57

---

@raineorshine -- I noticed one other potential issue, but I would ask you to clarify what the requirement is.

On mobile (testing on Android's Chrome, specifically), the alert can be covered by the keyboard. Do we want to leave the alert at the bottom of the "page" when the keyboard is open, or should it be shifted above the keyboard and displayed in the visible portion of the page?

https://github.com/user-attachments/assets/285e488e-a6d9-4acf-85be-6a910041aa73

raineorshine:
> ## 6. Gesture alerts sometimes fade immediately
> When a gesture is made slowly, the resulting notification (e.g. `New Thought`) is displayed for the normal amount of time. However, when quickly making a gesture, the notification appears and disappears almost immediately.

@trevinhofmann I think this is a problem in `main` actually. If so, could you create a new issue for it?

> On mobile (testing on Android's Chrome, specifically), the alert can be covered by the keyboard. Do we want to leave the alert at the bottom of the "page" when the keyboard is open, or should it be shifted above the keyboard and displayed in the visible portion of the page?

That seems to be an Android only problem, as I believe the alert shows up above the virtual keyboard in mobile Safari. I think the alert will need to be in the visible part of the screen to be useful. For example, if the user accidentally deletes a thought, it's important that they see the alert so they know what happened.

@yangchristina I'm not sure what the workarounds are on Android for making position:fixed elements appear above the virtual keyboard, but I imagine others have run into this before. Maybe you can investigate a bit to determine the effort involved in this fix?

src/components/PopupBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+import useCombinedRefs from '../hooks/useCombinedRefs'
+import usePositionFixed from '../hooks/usePositionFixed'
+import useSwipeToDismiss from '../hooks/useSwipeToDismiss'
+import syncStatusStore from '../stores/syncStatus'
+import fastClick from '../util/fastClick'
+import CloseButton from './CloseButton'
+
+export type PopupBaseProps = PropsWithChildren<
+  {
+    /** used to cancel imports */
```
trevinhofmann, line 19:
```suggestion
    /** Used to cancel imports. */
```

Apologies for missing this in my previous suggestion, but the ESLint rules require JSDoc comments to begin with a capital letter and end with a period.

trevinhofmann (changes requested):
> @trevinhofmann I think this is a problem in `main` actually. If so, could you create a new issue for it?

Ah, good catch! I have created a new issue for this: https://github.com/cybersemics/em/issues/2784

> I think the alert will need to be in the visible part of the screen to be useful. For example, if the user accidentally deletes a thought, it's important that they see the alert so they know what happened.

@yangchristina -- I believe the only remaining issues are the merge conflict and (7) below:

---

## 7. Alert hidden by keyboard

On mobile, the alert can be covered by the keyboard. When the on-screen keyboard is open, the alert should be shifted up to still be included at the bottom of the visible portion of the page.

https://github.com/user-attachments/assets/285e488e-a6d9-4acf-85be-6a910041aa73

yangchristina:
I don't think there's a simple way to move the alert up when the keyboard opens. However I can make it so that if the keyboard is open, you can see the alert. However if the alert was there before the keyboard opened, it will get covered once the keyboard opens.

https://github.com/user-attachments/assets/00f62e6e-c73c-41d1-962b-0741c6155f58

yangchristina:
@raineorshine I've also updated the close button svg. Does this look right?

raineorshine:
> I don't think there's a simple way to move the alert up when the keyboard opens. However I can make it so that if the keyboard is open, you can see the alert. However if the alert was there before the keyboard opened, it will get covered once the keyboard opens. Is this okay @raineorshine?

The alert is also used for ongoing states such as drag-and-drop, multicursor, import progress, etc. I'm afraid that if the keyboard covers the alert, the user won't have this feedback.

Another problem is the inverse: If the user triggers an alert while the keyboard is up, and then closes the keyboard, the alert is left floating in the middle of the screen.

Instead of using `window.visualViewport`, which is non-reactive (and [buggy](https://github.com/cybersemics/em/blob/875843163108c256489a938a421c75de54216597/src/stores/viewport.ts#L22-L24)), my suggestion is to use our [viewportStore](https://github.com/cybersemics/em/blob/main/src/stores/viewport.ts).

```ts
const { innerHeight, virtualKeyboardHeight } = viewportStore.useState()
```

> @raineorshine I've also updated the close button svg. Does this look right?

I haven't tested it at multiple font sizes, but looks great so far! A few suggestions:

- I think the padding of the close button in the mockup looks a little small. I bumped the width from 20 to 24 in the mockup. Could you increase the padding accordingly?
- Let's make it desktop only. Mobile fires a hover event on click ("click intent") which we don't want in this case.
- Expand the click area about 10px on all sides to make it easier to click. I typically do this with a combination of padding and negative margin.
- Ensure that the expanded click area is available at all times, not just on hover. Otherwise hovering a few pixels above the upper-right corner where the ✗ should appear won't do anything initially.

trevinhofmann (commented):
Hi, @yangchristina! I will delay reviewing for now until you have a chance to address the latest feedback from @raineorshine. Please let me know when this is ready again for review, and I'll be happy to have another look. Thank you!

yangchristina:
Screenshot 2025-01-29 at 10 58 14 PM

@raineorshine `viewportStore` doesn't update when keyboard is up, so it doesn't fix this. I think I need an `touchmove` listener to update whether or not keyboard is up

raineorshine:
Thanks for the update. I didn't realize `viewportStore` was not updated when the virtual keyboard activated. We should be able to update it in response to the `selectionchange` event, when `selection.isActive()` changes. Then `innerHeight` will be accurate. 

If that doesn't work, or if you need additional guidance, @trevinhofmann can work with you on a solution.  Thanks!

yangchristina:
@raineorshine actually it kinda works sometimes? There was one point where the alert was actually working, but now it's back to not really working. (didn't make any code changes, just seems really inconsistent)

https://github.com/user-attachments/assets/26caa19b-6841-457f-baa5-4afd75ce2bdc

yangchristina:
Here's another example of it being really inconsistent. I'm not sure if it's just a problem with the simulator. From the logs, `viewportStore` does actually update correctly, but changes don't always show, or show inconsistently

https://github.com/user-attachments/assets/6cd541ba-eea2-43b7-b9ce-0d34701bf9e8

raineorshine:
That is strange. Some ideas:

1. Confirm that XCode Simulator does not behave differently from physical phone. 
2. Confirm viewportStore.update is called whenever the keyboard changes. 
3. Confirm Alert is re-rendered correctly when the window size changes. 

There is also a question of timing. We don't want the Alert position to be delayed, and it's probably not possible for it to smoothly animate down with the keyboard. Our best bet is probably to have the Alert snap to the bottom as soon as the keyboard is closed. The keyboard should cover the alert while animating out of view. Then the Alert will be in its place at the bottom of the screen. 

However, it's not really possible to design the correct behavior when you are getting inconsistent behavior, so I would focus on understanding the cause of the inconsistency first. There is usually another variable we are not aware of that is affecting the behavior. 

Hope that helps!

yangchristina:
@raineorshine I don't really have a way of running this on an actual phone. (I don't use an iphone and even if I did I don't know how I'd connect that phone to em's localhost)

raineorshine:
No problem, one of us can verify that. It's unlikely that it's the problem, but it doesn't hurt to verify base assumptions when inconsistent behavior is observed.

yangchristina:
@raineorshine that would be great, thanks!

yangchristina:
> 2. Confirm viewportStore.update is called whenever the keyboard changes.
Can confirm this. It does update when keyboard changes.
> 3. Confirm Alert is re-rendered correctly when the window size changes.
Is there another way to check if a re-rendering is happening at all? It clearly looks like it's not rendering correctly most of the time.

trevinhofmann:
Running the latest on a physical iPhone, I see the same functionality. It looks close to working, but there is a lagging update to the positioning when the keyboard is opened/closed. I will be able to look into this more later today.

https://github.com/user-attachments/assets/d651a3d8-538e-4afe-babc-a173ec6b5e26

raineorshine:
> How can I resize on ios devices?

You would rotate from portrait to landscape. XCode Simulator has an option to do this also.

raineorshine:
> Running the latest on a physical iPhone, I see the same functionality. It looks close to working, but there is a lagging update to the positioning when the keyboard is opened/closed. I will be able to look into this more later today.

When the keyboard is closed, you may need to "anticipate" the upcoming screen height in order to render the alert at the bottom instead of animating it down. That’s the tricky part.

trevinhofmann:
For the sake of debugging, I:
* added a permanent alert to show the keyboard height from the viewport store
* added a secondary alert at the top (to always have debugging visible)
* added the `usePositionFixed` styles for the regular alert, to both alerts

https://github.com/user-attachments/assets/5bb8c6c3-a799-4c2b-b451-aff942b83b66

I made the style text quite small to ensure it fits, but it should still be visible in the video when expanded.

This debugging revealed that:
1. The keyboard height is always accurate (though a bit delayed when closing the keyboard, which may be unavoidable)
2. The `usePositionFixed` for the alert switches from `position: fixed` to `position: absolute` when the keyboard is opened, and this positioning seems to prevent it from rendering correctly above the keyboard

I'm not well informed on the history/reasoning for the `position: absolute` functionality. The `git blame` history is a bit muddled, but the comments indicate this is a fix for mobile Safari. Considering the `position: fixed` seems to consistently work on iOS Safari (likely thanks to the changes in this PR), perhaps we can remove it? I would just need to do some more testing to make sure there are no unintended side effects for other components.

As a (very tentative) fix, this commit removes the `position: absolute` functionality from the `usePositionFixed` hook: https://github.com/cybersemics/em/commit/6b4edd797e6e272160cbb46d7fb694e5bfa0ba6c

With that change added (and making the alert permanent for demo purposes), the alert seems to be correctly positioned. There is still a delay when closing the keyboard, but that may be unavoidable. It seems that Safari does not trigger the resize event until the keyboard fully closes.

https://github.com/user-attachments/assets/bf4f9941-243e-4b96-bd9f-31e9af877ab9

raineorshine:
Thank you for the additional testing @trevinhofmann!

> I'm not well informed on the history/reasoning for the `position: absolute` functionality. The `git blame` history is a bit muddled, but the comments indicate this is a fix for mobile Safari. Considering the `position: fixed` seems to consistently work on iOS Safari (likely thanks to the changes in this PR), perhaps we can remove it? I would just need to do some more testing to make sure there are no unintended side effects for other components.

As far as I know, `position: fixed` doesn't work at all on mobile Safari when the virtual keyboard is up. The element will stay fixed to the bottom of the document instead of the bottom of the viewport.

Demo: https://s5ntpd.csb.app/

Let me know if you are seeing something different. I've worked on the problem at various times over the years and have not been able to find a proper solution. Most recently I generalized and factored out the the position: fixed functionality into the `usePositionFixed` hook so it can be used in the Toolbar, Alert, and other components as needed.

> There is still a delay when closing the keyboard, but that may be unavoidable. It seems that Safari does not trigger the resize event until the keyboard fully closes.

That's too bad. I don't think the delay is really acceptable from a UX perspective. The alert floating in the middle of the screen looks pretty bad.

I was thinking about the challenge of "anticipating" the next height, and perhaps the better approach is to simply hide the alert while the virtual keyboard is closing. We could hide it on blur (or global `selectionchange`) and then show it again when the `resize` fires.

trevinhofmann:
> As far as I know, `position: fixed` doesn't work at all on mobile Safari when the virtual keyboard is up. The element will stay fixed to the bottom of the document instead of the bottom of the viewport.

Ah! That makes sense. I see the issue when scrolling. In that case, we'll need to fix the calculations for `position: absolute` in `usePositionFixed` (rather than removing them), as that seems to be the problem.

> I was thinking about the challenge of "anticipating" the next height, and perhaps the better approach is to simply hide the alert while the virtual keyboard is closing. We could hide it on blur (or global `selectionchange`) and then show it again when the `resize` fires.

That makes sense to me!

---

@yangchristina -- Would you like to work on the above two changes? Otherwise, I will be available to attempt them later today.

yangchristina:
> > As far as I know, `position: fixed` doesn't work at all on mobile Safari when the virtual keyboard is up. The element will stay fixed to the bottom of the document instead of the bottom of the viewport.
> 
> Ah! That makes sense. I see the issue when scrolling. In that case, we'll need to fix the calculations for `position: absolute` in `usePositionFixed` (rather than removing them), as that seems to be the problem.
> 
> > I was thinking about the challenge of "anticipating" the next height, and perhaps the better approach is to simply hide the alert while the virtual keyboard is closing. We could hide it on blur (or global `selectionchange`) and then show it again when the `resize` fires.
> 
> That makes sense to me!
> 
> @yangchristina -- Would you like to work on the above two changes? Otherwise, I will be available to attempt them later today.

@trevinhofmann Sorry for responding late. Position absolute is already being handled though, so that shouldn't be the issue.

```
  if (position === 'absolute') {
    top = fromBottom ? `${scrollTop + innerHeight - (height ?? 0) - offset}px` : `${scrollTop + offset}px`
  } else if (fromBottom) {
    bottom = `calc(${token('spacing.safeAreaBottom')} + ${offset + currentKeyboardHeight}px)`
  } else {
    top = `calc(${token('spacing.safeAreaTop')} + ${offset}px)`
  }
```

updatePositionFixed shouldn't be called considering this is on an ios device.
```
if (isTouch && isSafari() && !isIOS) {
    updatePositionFixed()
    document.addEventListener('selectionchange', updatePositionFixed)
  }
```

trevinhofmann:
> @trevinhofmann Sorry for responding late. Position absolute is already being handled though, so that shouldn't be the issue.

No problem! I'm suspicious that some part of the `position: absolute` calculation is incorrect since that seems to be when the alert is not visible (from the first video in https://github.com/cybersemics/em/pull/2746#issuecomment-2635178658).

I will let you take the first attempt at resolving this, but please let me know if you are stuck and would like some help.

Thank you!

yangchristina:
> > @trevinhofmann Sorry for responding late. Position absolute is already being handled though, so that shouldn't be the issue.
> 
> No problem! I'm suspicious that some part of the `position: absolute` calculation is incorrect since that seems to be when the alert is not visible (from the first video in https://github.com/cybersemics/em/pull/2746#issuecomment-2635178658).
> 
> I will let you take the first attempt at resolving this, but please let me know if you are stuck and would like some help.
> 
> Thank you!

Sounds good, thanks! Do you happen to have the code for the debugging somewhere? If it's not too inconvenient,  would love to see how you did this.

yangchristina:
Fixed! Yes, my calculations for `position: absolute` were wrong. `innerHeight` in`viewportStore` said "Height of the viewport, not including the virtual keyboard.", which I assumed meant the full height minus the keyboard. I've updated the jsdoc comment to "Height of the viewport, including the virtual keyboard." if that's okay. I can either do that or subtract keyboard height from `innerHeight` and keep the jsdoc comment as is.

> I was thinking about the challenge of "anticipating" the next height, and perhaps the better approach is to simply hide the alert while the virtual keyboard is closing. We could hide it on blur (or global selectionchange) and then show it again when the resize fires.

UX wise it seems fine to me to have a delay when you are opening the keyboard, but not when you are closing it. The delay when opening is essentially the same as hiding the alert a bit before showing it again. But for closing, having it float in the middle of the screen is more bothersome and looks like a bug.

The original scope of this alert was just supposed to include styling, so changing the way it shows on keyboard close is out of scope.

yangchristina:
I'm not sure why, but ColorPicker tests aren't failing for me locally.

raineorshine:
> Fixed! Yes, my calculations for `position: absolute` were wrong. `innerHeight` in`viewportStore` said "Height of the viewport, not including the virtual keyboard.", which I assumed meant the full height minus the keyboard. I've updated the jsdoc comment to "Height of the viewport, including the virtual keyboard." if that's okay. I can either do that or subtract keyboard height from `innerHeight` and keep the jsdoc comment as is.

You are correct! Sorry for the confusion. I clearly documented that wrong at some point. `innerHeight` **does** include the virtual keyboard. `virtualKeyboardHeight` can be subtracted from it to get the visible viewport height.

This might be the source of other bugs if there is any more code that is assuming the wrong behavior of `innerHeight`, but I will have to investigate that separately.

> UX wise it seems fine to me to have a delay when you are opening the keyboard, but not when you are closing it. The delay when opening is essentially the same as hiding the alert a bit before showing it again. But for closing, having it float in the middle of the screen is more bothersome and looks like a bug.

Yes, I agree.

> The original scope of this alert was just supposed to include styling, so changing the way it shows on keyboard close is out of scope.

Thanks, agreed.

@trevinhofmann Could you possibly do this part since you are familiar with the problem? I would prefer do it all in one PR so that we can avoid the layout problem getting into `main`.

> I'm not sure why, but ColorPicker tests aren't failing for me locally.

Nothing changed with it, but I've been noticing a lot more GitHub Actions failures due to this snapshot test. I have temporarily disabled it in d936436632b until we can get to the bottom of it. Thanks.

trevinhofmann:
@raineorshine -- It seems that iOS Safari [does not provide any events](https://martijnhols.nl/blog/how-to-detect-the-on-screen-keyboard-in-ios-safari) to detect when the keyboard is open/closed or opening/closing. That explains why we have the workaround of using `selection.isActive()` to detect when the keyboard is open, but it limits our ability to know when the keyboard is _finished_ closing.

My proposed solution (e06af0a3aaa97d56294892ec2a7e769d313c78a4) refactors the keyboard detection out of `usePositionFixed` into a dedicated `safariKeyboardStore` and uses a [fixed timeout](https://github.com/cybersemics/em/pull/2746/commits/e06af0a3aaa97d56294892ec2a7e769d313c78a4#diff-f4502f3b781eeb6822bff24fc3cb27d97c06e429cec8a10e2ac2abee7f7ea2bdR26) to determine when the keyboard has completed closing.

Please let me know if you have any better ideas, but I haven't been able to find any way to definitively determine when the keyboard finishes closing.

Demo:

https://github.com/user-attachments/assets/bc4a0d8f-d3ed-42ee-ba94-909c7ec16f1c

trevinhofmann:
One potential enhancement would be to have the timeout gradually update the state of the "keyboard height" for iOS Safari to mimic it reducing in height as it closes. This way, the modal could slide down with the keyboard rather than temporarily disappearing. Again, though, there appears to be no supported way of determining the exact timing/height of the iOS Safari keyboard. The calculated/estimated keyboard hight might be slightly different from the actual keyboard height.

If you'd like me to try this approach, please just let me know.

src/stores/safariKeyboardStore.ts
```diff
@@ -0,0 +1,38 @@
...
+    }, VIRTUAL_KEYBOARD_CLOSE_DURATION)
+  }
+  safariKeyboardStore.update({
+    open: keyboardIsVisible,
+  })
+}
+
+if (isTouch && isSafari() && !isIOS) {
+  updateKeyboardState()
+  document.addEventListener('selectionchange', updateKeyboardState)
```
raineorshine, line 35: I prefer to initialize document and events in a dedicated loading phase rather than in the module scope. It creates more predictable timing, and allows event listeners to be removed in a cleanup phase if needed.

Therefore, I recommend exporting `updateKeyboardState` and calling it in the existing global `onSelectionChange`:

https://github.com/cybersemics/em/blob/71545c5dbe3eda5204cf00a02590cafc48b2f58d/src/util/initEvents.ts#L164-L171

src/hooks/usePositionFixed.ts
```diff
@@ -1,45 +1,50 @@
...
+  const safariKeyboard = safariKeyboardStore.useState()
+  const position = safariKeyboard.open ? 'absolute' : 'fixed'
   const scrollTop = useScrollTop({ disabled: position === 'fixed' })
+  const { innerHeight, currentKeyboardHeight } = viewportStore.useState()
 
-  useEffect(initEventHandler, [])
+  let top, bottom
+  if (position === 'absolute') {
+    top = fromBottom
+      ? `${scrollTop + innerHeight - currentKeyboardHeight - (height ?? 0) - offset}px`
```
raineorshine, line 32: Can we use `innerHeight - virtualKeyboardHeight` instead of `currentKeyboardHeight`?

src/hooks/usePositionFixed.ts
```diff
@@ -1,45 +1,50 @@
...
+      : `${scrollTop + offset}px`
+  } else if (fromBottom) {
+    bottom = `calc(${token('spacing.safeAreaBottom')} + ${offset + currentKeyboardHeight}px)`
```
raineorshine, line 35: Wouldn't `currentKeyboardHeight` always be 0 in this case, since the keyboard is closed?

raineorshine (changes requested):
Thanks @trevinhofmann!

> @raineorshine -- It seems that iOS Safari [does not provide any events](https://martijnhols.nl/blog/how-to-detect-the-on-screen-keyboard-in-ios-safari) to detect when the keyboard is open/closed or opening/closing. That explains why we have the workaround of using `selection.isActive()` to detect when the keyboard is open, but it limits our ability to know when the keyboard is _finished_ closing.

Yes, it's unfortunate there isn't a native event for this. We have [state.editing](https://github.com/cybersemics/em/blob/71545c5dbe3eda5204cf00a02590cafc48b2f58d/src/%40types/State.ts#L74-L75) in the Redux store which basically tracks the virtual keyboard on mobile. However, it's a bit more indirect in that it's tied in with the editor. It is also causative, actually determining whether the virtual keyboard is opened or not depending on certain commands (e.g. `newThought`, `clearThought` always turn editing mode on). It may be safer to maintain a separate implementation for the position fixed functionality, although I don't love the redundancy. Thoughts?

> My proposed solution ([e06af0a](https://github.com/cybersemics/em/commit/e06af0a3aaa97d56294892ec2a7e769d313c78a4)) refactors the keyboard detection out of `usePositionFixed` into a dedicated `safariKeyboardStore` and uses a [fixed timeout](https://github.com/cybersemics/em/pull/2746/commits/e06af0a3aaa97d56294892ec2a7e769d313c78a4#diff-f4502f3b781eeb6822bff24fc3cb27d97c06e429cec8a10e2ac2abee7f7ea2bdR26) to determine when the keyboard has completed closing.

That seems fine.

> One potential enhancement would be to use a `setInterval` rather than `setTimeout` to gradually update the state of the "keyboard height" for iOS Safari to mimic it reducing in height as it closes. This way, the modal could slide down with the keyboard rather than temporarily disappearing. Again, though, there appears to be no supported way of determining the exact timing/height of the iOS Safari keyboard. The calculated/estimated keyboard height might be slightly different from the actual keyboard height as it closes.

I don't trust being able to accurately track the virtual keyboard position.

We could "anticipate" the new height and render the alert in the correct place ahead of time, but I think that should be explored in a separate PR, if at all.

---

I have a few comments/questions on the code before I test.

src/hooks/usePositionFixed.ts
trevinhofmann: @yangchristina might be better able to answer this, but the `virtualKeyboardHeight` seems to be a cached estimate in this case, whereas `currentKeyboardHeight` is correctly calculated. In testing, the `virtualKeyboardHeight` was ~290px and `currentKeyboardHeight` was ~268px. Using the `currentKeyboardHeight` seems to give a better positioning in this case.

src/hooks/usePositionFixed.ts
trevinhofmann: Yep! Refactored that one out here: 99419f66bf875679e633a2322e1cc819e1873560

src/hooks/usePositionFixed.ts
raineorshine: To clarify, I meant the state variable, `state.virtualKeyboardHeight`, which is: 

```ts
        isViewportValid
          ? virtualKeyboardHeight
          : isPortrait
            ? virtualKeyboardHeightPortrait
            : virtualKeyboardHeightLandscape`
```

I didn't realize `virtualKeyboardHeight` was the name of both a local variable and a state variable, so my statement was ambiguous.

> @yangchristina might be better able to answer this, but the `virtualKeyboardHeight` seems to be a cached estimate in this case, whereas `currentKeyboardHeight` is correctly calculated.

The main problem is that `currentKeyboardHeight` uses `window.visualViewport.height` as-is, which has a bug in iOS Safari. When the phone is rotated with the keyboard open, rotated back, and then the keyboard closed, the value of `window.visualViewport.height` is wrong. We have to override it to the known portrait or landscape height based on the device dimensions. That is the intention behind `state.virtualKeyboardHeight`.

The [estimates](https://github.com/yangchristina/em/blob/2887108096eec3876ef575bb7cff4a9a7cb2b12f/src/stores/viewport.ts#L6) are only used until a valid viewport is detected, then they [permanently change to the measured height](https://github.com/yangchristina/em/blob/2887108096eec3876ef575bb7cff4a9a7cb2b12f/src/stores/viewport.ts#L31C104-L38).

> In testing, the `virtualKeyboardHeight` was ~290px and `currentKeyboardHeight` was ~268px. Using the `currentKeyboardHeight` seems to give a better positioning in this case.

I'm not sure if you're talking about the local variable or state variable (again, my bad on the ambiguity), but I'm assuming these measurements were not taken under the condition of the rotation bug, therefore `window.innerHeight - window.visualViewport.height` will be correct.

`state.virtualKeyboardHeight` should be preferred over `window.innerHeight - window.visualViewport.height` due to the rotation bug. If it's giving you bad numbers, and the existing code has a bug, I can only speculate as to the cause:

- It is properly switching from the estimated height to the measured height?
- Are the hardcoded ratios for ` virtualKeyboardHeightPortrait` incorrect for some devices?
- Is `isViewportValid` correct?

trevinhofmann:
> Yes, it's unfortunate there isn't a native event for this. We have [state.editing](https://github.com/cybersemics/em/blob/71545c5dbe3eda5204cf00a02590cafc48b2f58d/src/%40types/State.ts#L74-L75) in the Redux store which basically tracks the virtual keyboard on mobile. However, it's a bit more indirect in that it's tied in with the editor. It is also causative, actually determining whether the virtual keyboard is opened or not depending on certain commands (e.g. newThought, clearThought always turn editing mode on). It may be safer to maintain a separate implementation for the position fixed functionality, although I don't love the redundancy. Thoughts?

When changing the landscape/portrait rotation, I noticed that `selection.isActive()` does not provide the correct value when closing the keyboard after the rotation.

This is resolved by switching to `state.editing`, and I've made that change here: c16fd5c5e2bf4108676640b7516edf518562fc25

src/hooks/usePositionFixed.ts
trevinhofmann: There may be an issue with `virtualKeyboardHeight`.

Here are screenshots showing an alert with the current values of `currentKeyboardHeight` and `virtualKeyboardHeight` from the store, using each of those values as either:

```
      ? `${scrollTop + innerHeight - currentKeyboardHeight - (height ?? 0) - offset}px`
```

```
      ? `${scrollTop + innerHeight - virtualKeyboardHeight - (height ?? 0) - offset}px`
```

| Using `currentKeyboardHeight`  | Using `virtualKeyboardHeight` |
| ------------- | ------------- |
| current  | virtual  |

Also, this seems to be the only place that we are using `currentKeyboardHeight`. If we _do_ replace it here in the calculation, I believe we could remove `currentKeyboardHeight` from the store entirely.

---

Regarding the portrait/landscape rotation, the current implementation doesn't seem to cause issues when using `currentKeyboardHeight` (after changing to `state.editing`):

https://github.com/user-attachments/assets/2e627c1b-7744-4b7e-a851-c49d0c8cc027

raineorshine:
I tested the supposed keyboard rotation bug and was not able to reproduce it. (A good lesson: no matter how clearly documented something is, without steps to reproduce, current, and expected behavior it can be very hard to evaluate in the future.)

I will have to think more about whether `currentKeyboardHeight` is the right abstraction. It kind of represents a composite value: the height of the virtual keyboard + whether the keyboard is open. State should generally be orthogonal.

Will revisit soon.

src/stores/safariKeyboardStore.ts
```diff
@@ -0,0 +1,37 @@
+import store from './app'
+import reactMinistore from './react-ministore'
+
+const VIRTUAL_KEYBOARD_CLOSE_DURATION = 800
+
+const safariKeyboardStore = reactMinistore<{
+  open: boolean
+  closing: boolean
```
raineorshine, line 8: `ministore` will infer the generic type from the initialized value, so it's not necessary to explicitly provide it.

raineorshine (changes requested):
Thank you! Looks great. Tested at different font sizes, screen sizes, and device rotation.

raineorshine:
- [ ] Close ✗ button should only appear on desktop (`!isTouch`). It is currently showing up on mobile on hover intent (e.g. swipe to dismiss the delete alert when the keyboard is down).

raineorshine:
- [ ] Clickable area of ✗ button should exist on desktop even when not visible. Otherwise it creates a jarring user experience where the hover zone changes depending on the hover state.
  Preview 2025-02-27 14 33 30

raineorshine:
- [ ] When opening the keyboard, the alert floats in the middle of the page. We should avoid this by either animating the alert up, delaying its new coordinates while they the keyboard is opening, or hiding it while the keyboard is opening. 
  - I'm leaning towards animating it up, even though I was opposed to it earlier. Snapping in is too abrupt and fading in doesn't make sense because the alert is already visible, it's just moving.

raineorshine:
- When closing the keyboard, I would like to render the alert at the anticipated position (i.e. where it will be once the keyboard is fully closed) rather than hiding it for a fixed delay. I find the current behavior too jarring, and so what I had previously deemed a mere optimization I see now more as a necessary change in behavior.

src/components/PopupBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+    /** If defined, will show a small x in the upper right corner. */
+    onClose?: () => void
+    anchorFromBottom?: boolean
+    anchorOffset?: number
+    /** Only for if `anchorFromBottom = true`. For calculating position on mobile safari. */
+    calculatedHeight?: number
+    cssRaw?: SystemStyleObject
+    circledCloseButton?: boolean
+    showXOnHover?: boolean
+    swipeDownToDismiss?: boolean
```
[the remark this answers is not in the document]
yangchristina: `showXOnHover` is false for tip, since in the design the x always shows.
PopupBase is also used in Popup component, which is used by command palette. `swipeDownToDismiss` not true there.

src/components/CloseButton.tsx
```diff
@@ -1,34 +1,96 @@
...
       ✕
-    </a>
+    </BaseCloseButton>
   )
 }
 
+/** A close button with an ✕. */
+const CloseButton = ({ circled, ...props }: CloseButtonProps & { circled?: boolean }) => {
+  return circled ? <CircledCloseButton {...props} /> : <LetterCloseButton {...props} />
+}
```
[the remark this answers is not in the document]
yangchristina: `circled` is not true in `ErrorMessage`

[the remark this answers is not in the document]
yangchristina: `circled=false` is the old close button. `circled=true` is the new svg. It's basically a different variant of the same component.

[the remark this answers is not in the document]
raineorshine: Got it, thanks! I missed the `ErrorMessage` usage.

src/components/PopupBase.tsx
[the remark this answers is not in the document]
raineorshine: I see, thank you.

trevinhofmann:
I've updated this so that the alert slides up/down as the keyboard opens/closes: 805eb0b5c49c3ffd441fd890c7c34e7a4c4c7ba0

trevinhofmann:
I've ensured that the close button only appears on desktop: 86391f453bdc8bd5b213a570340dc2047d439872

And preserved the hover zone of the close button even when not visible: 99d3740c24b3399f2660704654320bc8bf3a3bbe

src/stores/safariKeyboardStore.ts
trevinhofmann: Fixed by b13e1bc90c7d77215dc6969434d0cb68f9f6bcef

src/components/PopupBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+            onClose={onClose}
+            disableSwipeToDismiss
+          />
+        ) : null}
+      </div>
+    )
+  },
+)
+
+PopupBase.displayName = 'Popup'
```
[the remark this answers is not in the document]
trevinhofmann: Fixed: 8c0ef8a1ed05f20d47c0f9ea5921d3ea2aa585b4

src/components/Alert.tsx
```diff
@@ -77,11 +41,44 @@ const Alert: FC = () => {
...
+              right: 0,
+              width: 'max-content',
+            })}
+            ref={popupRef}
+            // Specify a key to force the component to re-render and thus recalculate useSwipeToDismissProps when the alert changes. Otherwise the alert gets stuck off screen in the dismiss state.
+            key={value}
+            circledCloseButton
+            showXOnHover
+            onClose={alert.showCloseLink && !isTouch ? onClose : undefined}
+            calculatedHeight={popupRef.current?.getBoundingClientRect().height || 50}
```
raineorshine, line 64: `getBoundingClientRect` has performance implications because it forces a layout recalculation before returning up-to-date element dimensions and position. Therefore it should only be called on user action or in `useEffect` (or `useLayoutEffect` if critical for the initial paint).

src/components/CloseButton.tsx
```diff
@@ -1,33 +1,94 @@
...
 import haptics from '../util/haptics'
 
-/** A close button with an ✕. */
-const CloseButton = ({ onClose, disableSwipeToDismiss }: { onClose: () => void; disableSwipeToDismiss?: boolean }) => {
-  const fontSize = useSelector(state => state.fontSize)
-  const padding = fontSize / 2 + 2
+type CloseButtonProps = {
+  onClose: () => void
+  disableSwipeToDismiss?: boolean
+  cssRaw?: SystemStyleObject
```
raineorshine, line 12: Similar question here.

I would probably steer clear of allowing both `cssRaw` and `style` in the component interface as well. `css` and `style` are often used side-by-side on a DOM element when dealing with runtime values that the panda compiler can't handle, but this hidden constraint can quickly get lost when the values originate up the component hierarchy.

src/components/CloseButton.tsx
```diff
@@ -1,33 +1,94 @@
...
-        css({
-          fontSize: 'sm',
-          color: 'inherit',
-          right: '0',
-          textDecoration: 'none',
-          top: '0',
-        }),
+        css(
+          {
+            // inherit not yet supported by plugin
```
raineorshine, line 31: This comment (x3) can be removed since we have [whitelisted](https://github.com/cybersemics/em/blob/f010868894f21341c49bfcbf7e5a84e8b699c3dd/eslint.config.js#L194) inherit now.

raineorshine (changes requested):
Thank you for taking the extra time to add the new behavior! I think it looks great.

We're getting closer, I just have a few more questions about data modeling in the store and component props design.

src/stores/safariKeyboardStore.ts
```diff
@@ -0,0 +1,64 @@
...
+// An approximation of how long it takes the keyboard to open
+const VIRTUAL_KEYBOARD_OPEN_DURATION = 200
+// An approximation of how long it takes the keyboard to close
+const VIRTUAL_KEYBOARD_CLOSE_DURATION = 150
+
+/** A store that tracks the state of the Safari virtual keyboard. */
+const safariKeyboardStore = reactMinistore({
+  open: false,
+  opening: false,
+  closing: false,
```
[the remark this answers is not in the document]
trevinhofmann: Ah, those were partly carryovers from previous functionality and partly kept in case it would be useful later to know if the keyboard is opening/closing. Since we switched to the sliding behavior for `usePositionFixed`, we don't actually use `opening`/`closing` at all anymore.

I'll remove them for now since they are unused, but I agree that `state: 'open' | 'opening' | 'closing'` would be better if we decide to include them later.

Removed here: a65c520a0247741ac07f06e45ca27284e01a4b09

src/stores/safariKeyboardStore.ts
```diff
@@ -0,0 +1,64 @@
...
+  // A timeout is necessary to ensure the editing state is updated after the selection change.
+  // This places the function call in the next event loop, after the state has been updated.
+  setTimeout(() => {
+    const { virtualKeyboardHeight } = viewportStore.getState()
+    const editing = store.getState().editing
+    const keyboardIsVisible = editing === true
+    if (keyboardIsVisible && !safariKeyboardStore.getState().open) {
+      // Animate the keyboard opening
+      const start = Date.now()
+      const interval = setInterval(() => {
```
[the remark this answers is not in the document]
trevinhofmann: As soon as the keyboard starts opening or closing, we start an interval at 60 FPS (`16.666` is used elsewhere in the repo for this; might be worth a small refactor to put that in a constant somewhere but I figured that would be outside the scope of this work) to gradually update the height based on `VIRTUAL_KEYBOARD_OPEN_DURATION` / `VIRTUAL_KEYBOARD_CLOSE_DURATION` and the `virtualKeyboardHeight`. The interval clears itself once the duration is reached.

With the way `usePositionFixed` works, I think a CSS animation would be tricky. I'm sure there's a way to make it work, but we'd also have the edge case of an element (e.g. alert) appearing in the middle of the keyboard opening/closing. For that reason, I think it's useful for the store to have the current (estimated) height of the keyboard at all times.

src/components/PopupBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+export type PopupBaseProps = PropsWithChildren<
+  {
+    /** Used to cancel imports. */
+    importFileId?: string
+    /** If defined, will show a small x in the upper right corner. */
+    onClose?: () => void
+    anchorFromBottom?: boolean
+    anchorOffset?: number
+    /** Only for if `anchorFromBottom = true`. For calculating position on mobile safari. */
+    calculatedHeight?: number
```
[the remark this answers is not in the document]
trevinhofmann: Yes, that should work. Refactored here: b68265d1c8476fd32402dc25088fb2215b8f36dc

src/components/PopupBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+import CloseButton from './CloseButton'
+
+export type PopupBaseProps = PropsWithChildren<
+  {
+    /** Used to cancel imports. */
+    importFileId?: string
+    /** If defined, will show a small x in the upper right corner. */
+    onClose?: () => void
+    anchorFromBottom?: boolean
+    anchorOffset?: number
```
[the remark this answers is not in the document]
trevinhofmann: Added: 4aacfd098e49318fe6b8568a2aea8ec2dcc7328e

src/components/PopupBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+    /** Only for if `anchorFromBottom = true`. For calculating position on mobile safari. */
+    calculatedHeight?: number
+    cssRaw?: SystemStyleObject
+    circledCloseButton?: boolean
+    showXOnHover?: boolean
+    swipeDownToDismiss?: boolean
+  } & Omit<React.HTMLAttributes<HTMLDivElement>, 'className'>
+>
+
+/** A popup component that can be dismissed. */
```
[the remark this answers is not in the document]
trevinhofmann: Clarified here: ef976005b2acd96b9ce472f1dcbf5979df63acbd

src/components/PopupBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+  {
+    /** Used to cancel imports. */
+    importFileId?: string
+    /** If defined, will show a small x in the upper right corner. */
+    onClose?: () => void
+    anchorFromBottom?: boolean
+    anchorOffset?: number
+    /** Only for if `anchorFromBottom = true`. For calculating position on mobile safari. */
+    calculatedHeight?: number
+    cssRaw?: SystemStyleObject
```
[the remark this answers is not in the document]
trevinhofmann: Passed by `<Popup>`:
```
          boxShadow: 'none',
          border: 'none',
          display: 'block',
          padding: '8%',
          boxSizing: 'border-box',
          zIndex: 'popup',
          backgroundColor: 'bg',
          width: '100%',
          color: 'gray50',
          overflowY: 'auto',
          maxHeight: '100%',
          maxWidth: '100%',
```

Along with the combined CSS Raw from `CommandPalette`:
```
          zIndex: 'commandPalette'
```

Passed in `<Alert>`:
```
          boxSizing: 'border-box',
          background: 'panelBg',
          border: '1px solid {colors.panelBorder}',
          borderRadius: '8px',
          zIndex: 'popup',
          marginInline: 'auto',
          left: 0,
          right: 0,
          width: 'max-content',
```

If we're open to exceptions for passing `cssRaw`, I think `PopupBase` would be a reasonable one since it seems to be more of a utility component that leaves the styling up to the parent.

If you'd like me to look into refactoring this regardless, please let me know.

src/components/CloseButton.tsx
trevinhofmann: Removed: 82afe88e8fe1fd23130193724cd09c13ccdb4ee5

src/components/Alert.tsx
trevinhofmann: Moved into a `useEffect` (after moving the calculation into `PopupBase`): 869af85b3f63c61c957fbc15f7e72ab4b55623c7

src/components/CloseButton.tsx
trevinhofmann: For now, I've moved the `opacity: 0` style to a `transparent` prop so that none of the consumers of `CloseButton` are passing in `cssRaw` or `style` now: eafcb7599030ffe74c243fe42712b036ed41ebb4

The `cssRaw` styling passed internally by `CircledCloseButton` has quite a few styles that I'm worried would be excessive to include in `BaseCloseButton`, but I removed a couple of the redundant styles: 64d05919ed97161b3262209a5cda89ffcff8e357

Please let me know if you'd like me to refactor this further with e.g. `circled` and `letter` props that conditionally add the relevant styles so that the `CloseButtonProps` do not need to include `cssRaw` and `style` at all. Thanks!

trevinhofmann:
@raineorshine, do you know if this is one of the flaky puppeteer tests? It's passing locally for me, but failed 3 times in a row in the CI:

```
FAIL  |puppeteer-e2e| src/e2e/puppeteer/__tests__/drag-and-drop.ts > drag > should show alert and quick drop panels when long pressing after drag operation
Error: Expected image to match or be a close match to snapshot but was 0.3822916666666667% different from snapshot (1835 differing pixels).
```

src/stores/safariKeyboardStore.ts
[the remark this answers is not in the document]
raineorshine: Thanks. I think can keep this for now since it appears to be working as expected. We can consider a single CSS animation if we notice performance or other issues.

src/components/CloseButton.tsx
raineorshine: Thank you for the update! That is an improvement.

Here are the concerns I still have:

- As you mentioned, `CloseButtonProps` includes `cssRaw` and `style`, which end up exposed on the `CloseButton` props even though they are never used there. I think reusing `CloseButtonProps` on all of the variant components was probably done for convenience more than correctness. It would be better to narrow those component interfaces rather than leaving them wide open.
- By injecting styles into the base component, it adds complexity to the props and leaves the types too wide. My question is: Can these style be applied to a wrapper div in CircledCloseButton/LetterCloseButton instead? Why do they need to be merged into the `<a>` tag? Applying the styles to a wrapper div would simplify the interface between the components, be more type safe, and reduce the risks of style merging.
- Setting `opacity: 0` is more about completely hiding the close button rather than transparency. I don't see any CSS animations on `opacity` for a fade effect that would require us to render it at opacity 0. Could `PopupBase` can just add `showXOnHover` to the condition for rendering `CloseButton`, and then we could remove the `transparent` prop completely?
- `upperRightRecipe` appears to only be used in one place now, and no longer needs to be a recipe.

src/components/PopupBase.tsx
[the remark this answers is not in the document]
raineorshine: Yes, let's try to find a better way to do this. I have similar concerns as mentioned for `CloseButton`.

I understand the idea of an abstract, functional component that does not define any styles in theory. react-dnd takes this approach. But something doesn't sit right about it here.

- Exposing `cssRaw` to the `CommandPalette` just to set the `zIndex` doesn't seem right.
- Why are styles injected into the base component, adding complexity to the props and making them less type safe, rather applying them to a wrapper div? What styles actually need to be merged in the base component?

Thanks for looking into this further.

raineorshine (changes requested):
Thank you!

> @raineorshine, do you know if this is one of the flaky puppeteer tests? It's passing locally for me, but failed 3 times in a row in the CI:
> 
> ```
> FAIL  |puppeteer-e2e| src/e2e/puppeteer/__tests__/drag-and-drop.ts > drag > should show alert and quick drop panels when long pressing after drag operation
> Error: Expected image to match or be a close match to snapshot but was 0.3822916666666667% different from snapshot (1835 differing pixels).
> ```

I can confirm that the puppeteer tests pass on my local machine. However, this seems like an environment issue that should be investigated. The drag-and-drop tests have not been failing consistently in [recent](https://github.com/cybersemics/em/actions/runs/13882176154) [checks](https://github.com/cybersemics/em/actions/runs/13876112552) on `main`.

trevinhofmann:
Hi, @raineorshine! I've rebased to resolve conflicts and refactored to remove the `cssRaw` and `style` props from:
* `PopupBase`: 8ec6373c1b68423fc9f029a877592da374bbd2d0
* `CloseButton`: a58a22ce4175aa26d44c78d7bc12b30a7e8d19b0

> Why are styles injected into the base component, adding complexity to the props and making them less type safe, rather applying them to a wrapper div? What styles actually need to be merged in the base component?

I've tried to apply styles to wrapper divs, but most of them do not function as expected unless applied directly to the popup element (e.g. border, background).

> Setting opacity: 0 is more about completely hiding the close button rather than transparency. I don't see any CSS animations on opacity for a fade effect that would require us to render it at opacity 0. Could PopupBase can just add showXOnHover to the condition for rendering CloseButton, and then we could remove the transparent prop completely?

The `opacity: 0` is applied to resolve https://github.com/cybersemics/em/pull/2746#issuecomment-2689108442. In order to keep the hover area effective, we need to render the element but make it fully transparent unless hovered.

> upperRightRecipe appears to only be used in one place now, and no longer needs to be a recipe.

Removed: bd8578392e3cf57d717473ea60ec45112cde3a80

--

I believe this addresses the most recent comments. Please let me know what you think. Thanks!

raineorshine:
Is there a reason that `Popup` is a separate abstraction from `CommandPalette`?

- CommandPalette → Popup → PopupBase
- Alert → PopupBase

The Popup/PopupBase distinction could maybe be removed in favor of the simpler:

- CommandPalette → Popup
- Alert → Popup

Let me know what you think!

trevinhofmann:
@yangchristina - Would you like to explain your reasoning for creating the `Popup` / `PopupBase` distinction? Otherwise, I can look into this and give my opinion as well.

yangchristina:
> @yangchristina - Would you like to explain your reasoning for creating the `Popup` / `PopupBase` distinction? Otherwise, I can look into this and give my opinion as well.

Popup used to be used for other things as well I think, but now since it's only Command Palette it makes sense to remove the extra layer. Popup had specific styling, ex. taking up full screen, close button on top right, swipe up. 

I was originally made an entirely separate component for Alert, but I needed a lot of the Popup functionality [(reference above)](https://github.com/cybersemics/em/pull/2746#pullrequestreview-2525857892), so I created an abstraction. I agree that a distinction between Popup/PopupBase is no longer necessary. 

My only concern is the number of merge conflicts this will cause, since Command Palette work is in progress, but that can also be dealt with when it happens.

raineorshine (approved):
Thanks for your input.

I'll merge this now and we can come back to refactor `PopupBase` after the Command Palette changes are in.

PR #2752, opened 2025-01-02
Bullet/Editable: Use fastClick for clickHandler

### LINKED ISSUE #1691: [iOS] setCursor is delayed in edit mode
## Steps to Reproduce

1. Create a few thoughts.
2. Stay in edit mode.
3. Move the cursor around by tapping other thoughts. 

## Current Behavior

There is a delay after `touchend` before the cursor moves. This delay only exists in edit mode. 

## Expected Behavior

On `touchend`, the cursor should immediately move. 

(On `touchstart` the cursor cannot move immediately since we have to wait for drag-and-drop.)

ethan-james:
Fix #1691 

The `Bullet` component was using `onClick` to dispatch its `setCursor` action, which has a noticeable delay on iOS. Wrapping the click handler in `fastClick` solves the issue.

In `Editable`, the `onTap` handler was ignoring taps that would cause the keyboard to open, and falling back to the `onFocus` handler after a 300ms delay. It was necessary to intercept those taps and set the cursor with a manual offset to put the caret back at the end of the thought on iOS Safari. There was also some flakiness where `selection.set` wasn't always opening the keyboard, so now it calls `contentRef.current.focus()` explicitly.

### DIFF
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -20,6 +20,7 @@ import getThoughtById from '../selectors/getThoughtById'
 import isContextViewActive from '../selectors/isContextViewActive'
 import isMulticursorPath from '../selectors/isMulticursorPath'
 import rootedParentOf from '../selectors/rootedParentOf'
+import fastClick, { type FastClickEvent } from '../util/fastClick'
 import hashPath from '../util/hashPath'
 import head from '../util/head'
 import isDivider from '../util/isDivider'
@@ -531,16 +532,15 @@ const Bullet = ({
   // expand or collapse on click
   // has some additional logic to make it work intuitively with pin true/false
   const clickHandler = useCallback(
-    (e: React.MouseEvent) => {
-      // stop click event from bubbling up to Content.clickOnEmptySpace
-      e.stopPropagation()
+    (e: FastClickEvent) => {
       // short circuit if dragHold
       // useLongPress stop is activated in onMouseUp but is delayed to ensure that dragHold is still true here
       // stopping propagation from useLongPress was not working either due to bubbling order or mismatched event type
       if (dragHold) return
 
       // short circuit if toggling multiselect
-      if (!isTouch && (isMac ? e.metaKey : e.ctrlKey)) {
+      // e should always be a MouseEvent if !isTouch, but getting the type system to believe it is another story
+      if (!isTouch && e instanceof MouseEvent && (isMac ? e.metaKey : e.ctrlKey)) {
         dispatch(toggleMulticursor({ path }))
         return
       }
@@ -605,7 +605,9 @@ const Bullet = ({
         paddingBottom: extendClickHeight + 2,
         width,
       }}
-      onClick={clickHandler}
+      {...fastClick(clickHandler)}
+      // stop click event from bubbling up to Content.clickOnEmptySpace
+      onClick={e => e.stopPropagation()}
     >
       <svg
         className={cx(
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -539,8 +539,7 @@ const Bullet = ({
       if (dragHold) return
 
       // short circuit if toggling multiselect
-      // e should always be a MouseEvent if !isTouch, but getting the type system to believe it is another story
-      if (!isTouch && e instanceof MouseEvent && (isMac ? e.metaKey : e.ctrlKey)) {
+      if (!isTouch && (isMac ? e.metaKey : e.ctrlKey)) {
         dispatch(toggleMulticursor({ path }))
         return
       }
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -18,7 +18,7 @@ import { setCursorActionCreator as setCursor } from '../actions/setCursor'
 import { toggleColorPickerActionCreator as toggleColorPicker } from '../actions/toggleColorPicker'
 import { toggleLetterCaseActionCreator as toggleLetterCase } from '../actions/toggleLetterCase'
 import { tutorialNextActionCreator as tutorialNext } from '../actions/tutorialNext'
-import { isMac, isSafari, isTouch } from '../browser'
+import { isMac, isTouch } from '../browser'
 import { commandEmitter } from '../commands'
 import {
   EDIT_THROTTLE,
@@ -203,8 +203,6 @@ const Editable = ({
 
       // set offset to null to allow the browser to set the position of the selection
       let offset = null
-
-      if (isTouch && isSafari() && contentRef.current?.innerHTML.length) offset = contentRef.current.innerHTML.length
 
       // if running for the first time, restore the offset if the path matches the restored cursor
       if (!cursorOffsetInitialized) {
@@ -635,6 +633,7 @@ const Editable = ({
 
         onPaste(e)
       }}
+      role='button'
       style={style}
     />
   )
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -548,7 +548,7 @@ const Editable = ({
       if (
         disabled ||
         // dragInProgress: not sure if this can happen, but I observed some glitchy behavior with the cursor moving when a drag and drop is completed so check dragInProgress to be safe
-        (isTouch && isSafari() && !globals.touching && !state.dragInProgress && !state.dragHold)
+        (!globals.touching && !state.dragInProgress && !state.dragHold && (!editingOrOnCursor || !isVisible))
       ) {
         // do not set cursor on hidden thought
         e.preventDefault()
@@ -560,7 +560,7 @@ const Editable = ({
 
           if (state.showLetterCase) dispatch(toggleLetterCase({ value: false }))
         } else {
-          setCursorOnThought({ editing: editingOrOnCursor })
+          setCursorOnThought()
 
           // When the the cursor is first set on a thought, prevent the default browser behavior to avoid activating edit mode.
           // Do not reset until the long tap is definitely over.
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -49,6 +49,7 @@ import addEmojiSpace from '../util/addEmojiSpace'
 import containsURL from '../util/containsURL'
 import ellipsize from '../util/ellipsize'
 import equalPath from '../util/equalPath'
+import fastClick from '../util/fastClick'
 import head from '../util/head'
 import isDivider from '../util/isDivider'
 import strip from '../util/strip'
@@ -614,7 +615,7 @@ const Editable = ({
       placeholder={placeholder}
       // stop propagation to prevent default content onClick (which removes the cursor)
       onClick={stopPropagation}
-      onTouchEnd={onTap}
+      {...fastClick(onTap)}
       // must call onMouseDown on mobile since onTap cannot preventDefault
       // otherwise gestures and scrolling can trigger cursorBack (#1054)
       onMouseDown={onTap}
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -545,7 +545,7 @@ const Editable = ({
       if (
         disabled ||
         // dragInProgress: not sure if this can happen, but I observed some glitchy behavior with the cursor moving when a drag and drop is completed so check dragInProgress to be safe
-        (!globals.touching && !state.dragInProgress && !state.dragHold && (!editingOrOnCursor || !isVisible))
+        (isTouch && !globals.touching && !state.dragInProgress && !state.dragHold)
       ) {
         // do not set cursor on hidden thought
         e.preventDefault()
@@ -557,7 +557,7 @@ const Editable = ({
 
           if (state.showLetterCase) dispatch(toggleLetterCase({ value: false }))
         } else {
-          setCursorOnThought()
+          setCursorOnThought({ editing: editingOrOnCursor })
 
           // When the the cursor is first set on a thought, prevent the default browser behavior to avoid activating edit mode.
           // Do not reset until the long tap is definitely over.
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -18,7 +18,7 @@ import { setCursorActionCreator as setCursor } from '../actions/setCursor'
 import { toggleColorPickerActionCreator as toggleColorPicker } from '../actions/toggleColorPicker'
 import { toggleLetterCaseActionCreator as toggleLetterCase } from '../actions/toggleLetterCase'
 import { tutorialNextActionCreator as tutorialNext } from '../actions/tutorialNext'
-import { isMac, isTouch } from '../browser'
+import { isMac, isSafari, isTouch } from '../browser'
 import { commandEmitter } from '../commands'
 import {
   EDIT_THROTTLE,
@@ -202,6 +202,8 @@ const Editable = ({
 
       // set offset to null to allow the browser to set the position of the selection
       let offset = null
+
+      if (isTouch && isSafari() && contentRef.current?.innerHTML.length) offset = contentRef.current.innerHTML.length
 
       // if running for the first time, restore the offset if the path matches the restored cursor
       if (!cursorOffsetInitialized) {
@@ -545,7 +547,7 @@ const Editable = ({
       if (
         disabled ||
         // dragInProgress: not sure if this can happen, but I observed some glitchy behavior with the cursor moving when a drag and drop is completed so check dragInProgress to be safe
-        (isTouch && !globals.touching && !state.dragInProgress && !state.dragHold)
+        (isTouch && isSafari() && !globals.touching && !state.dragInProgress && !state.dragHold)
       ) {
         // do not set cursor on hidden thought
         e.preventDefault()
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -633,6 +633,8 @@ const Editable = ({
 
         onPaste(e)
       }}
+      // iOS Safari delays event handling in case the DOM is modified during setTimeout inside an event handler,
+      // unless it is given a hint that the element is some sort of form control
       role='button'
       style={style}
     />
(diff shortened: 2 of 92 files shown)


src/components/Editable.tsx
```diff
@@ -203,6 +203,8 @@ const Editable = ({
       // set offset to null to allow the browser to set the position of the selection
       let offset = null
 
+      if (isTouch && isSafari() && contentRef.current?.innerHTML.length) offset = contentRef.current.innerHTML.length
```
ethan-james, line 206: The caret was getting positioned at the end of the thought somehow `onFocus`, so I wanted to preserve that behavior `onTap`.

src/components/Editable/useEditMode.ts
```diff
@@ -51,6 +51,7 @@ const useEditMode = ({
           selection.clear()
         } else {
           selection.set(contentRef.current, { offset: editingCursorOffset || 0 })
+          if (isTouch && isSafari()) contentRef.current?.focus()
```
ethan-james, line 54: I'm not quite sure why this is necessary, but it was having the following problem on iOS Safari:

1. Tap a thought - cursor moves, keyboard stays closed
2. Tap it again - keyboard opens
3. Close keyboard
4. Tap it again - keyboard does not open

raineorshine: Thanks, good catch. 

I'll create a separate issue to get to the bottom of this.

src/components/Bullet.tsx
```diff
@@ -531,16 +532,15 @@ const Bullet = ({
...
+    (e: FastClickEvent) => {
       // short circuit if dragHold
       // useLongPress stop is activated in onMouseUp but is delayed to ensure that dragHold is still true here
       // stopping propagation from useLongPress was not working either due to bubbling order or mismatched event type
       if (dragHold) return
 
       // short circuit if toggling multiselect
-      if (!isTouch && (isMac ? e.metaKey : e.ctrlKey)) {
+      // e should always be a MouseEvent if !isTouch, but getting the type system to believe it is another story
+      if (!isTouch && e instanceof MouseEvent && (isMac ? e.metaKey : e.ctrlKey)) {
```
trevinhofmann, line 543: Just checking, why do we need to convince the type system that `e` is a `MouseEvent`? I'm not seeing any issues with the following, but I may be missing something:

```suggestion
      if (!isTouch && (isMac ? e.metaKey : e.ctrlKey)) {
```

trevinhofmann (approved):
Thank you, @ethan-james!

Switching to `fastClick` seems to have done the trick, when I test locally:

https://github.com/user-attachments/assets/0f0c4376-7964-46c0-8cea-7202c80b1c41

src/components/Bullet.tsx
ethan-james: It looks OK. I tried a couple of different definitions for `FastClickEvent` so this one may not need the type guard, as long as it passes CI.

raineorshine (changes requested):
Thanks for the submission! I can confirm that the timing is correct now.

The only problem I'm seeing is that the selection offset is not correct when tapping in the middle of a non-cursor thought in edit mode. This behavior relies on the default browser behavior to set the selection underneath the user's finger.

## Steps to Reproduce

```
- Hello World
- x
```

1. Set the cursor on `x`.
2. Tap `x` to open the keyboard if it's not open already.

## Current Behavior

a81cfd88b8

The caret is placed at the end of the thought, i.e. `Hello World|`

## Expected Behavior

`main`

The caret should be placed in between `Hello` and `World`, i.e. `Hello| World`.

ethan-james:
When I tested it on the `main` branch, I saw that `onFocus` would place the caret at the end of the thought, so that's the behavior that I tried to emulate. I'll see if instead I can come up with a reliable way to determine the selection offset from a `touchend` event.

raineorshine:
Do you know why the `onTap` handler is ignoring attempts to open the keyboard. Is one of the `e.preventDefault`s being called? 

My sense is that it would be more reliable to fix the root cause of the keyboard not appearing rather than trying to restore the selection and offset after the fact. But let me know what you find out!

ethan-james:
> Do you know why the `onTap` handler is ignoring attempts to open the keyboard. Is one of the `e.preventDefault`s being called?
> 
> My sense is that it would be more reliable to fix the root cause of the keyboard not appearing rather than trying to restore the selection and offset after the fact. But let me know what you find out!

It used to call `onTap` which would fall back to `onFocus` in this circumstance (cursor already on thought, keyboard closed). `onFocus` had a ~300ms delay so we're not using that anymore (by including `e.preventDefault` and manually managing the selection).

All of the advice on Google for eliminating click/focus delays is 10 years old and doesn't work, but I did just now put together a CodePen (https://codepen.io/ethan-james-the-sasster/pen/ByBYqrL) showing that the delay doesn't occur on simple `contenteditable` divs. So maybe there is hope for eliminating the delay at the source and going back to using `onFocus`, which I agree is a better solution.

Last thing to point out, though, is that on the iPad I'm using, the native focus behavior places the caret at the end of the editable text, and there doesn't seem to be a way to tap it into a different location. I've already emulated that behavior, although I agree that letting the browser control it natively is better.

ethan-james:
Also, I wasn't able to find a way to get the offset from a `touchend` event. If we do decide to pursue manually setting the offset, I think we have to try one of the usual suspects:

1. Estimating the width of each character and dividing by the touch x/y position
2. Inserting a hidden element and measuring its size with different combinations of characters

raineorshine:
> All of the advice on Google for eliminating click/focus delays is 10 years old and doesn't work, but I did just now put together a CodePen ([codepen.io/ethan-james-the-sasster/pen/ByBYqrL](https://codepen.io/ethan-james-the-sasster/pen/ByBYqrL)) showing that the delay doesn't occur on simple `contenteditable` divs. So maybe there is hope for eliminating the delay at the source and going back to using `onFocus`, which I agree is a better solution.

Yes, exactly. The delay is not inherent to contenteditables, but is being created through our own event handling. There is admittedly a lot of complexity with the selection handling, including automatically setting the selection when a thought re-renders, via `useEditMode`. However, theoretically we shouldn't need a lot of complex functionality for setting the caret when we are already in edit mode. The default browser behavior should set the caret under the user's finger on `touchend`.

> Last thing to point out, though, is that on the iPad I'm using, the native focus behavior places the caret at the end of the editable text, and there doesn't seem to be a way to tap it into a different location. I've already emulated that behavior, although I agree that letting the browser control it natively is better.

Interesting. My last few iPhones have correctly set the caret in the middle of the thought in these cases. I haven't tested on iPad recently, but I would assume with the same version of iOS it would behave the same. What version of iOS are you running? Do you possibly have another device or way to reproduce the default [caret behavior](https://github.com/cybersemics/em/pull/2752#pullrequestreview-2537798125)? It might be difficult to troubleshoot this otherwise.

> Also, I wasn't able to find a way to get the offset from a `touchend` event. If we do decide to pursue manually setting the offset, I think we have to try one of the usual suspects:
> 
> 1. Estimating the width of each character and dividing by the touch x/y position
> 2. Inserting a hidden element and measuring its size with different combinations of characters

Yes, that's what I was worried about. It's plausible, but strikes me as high effort + high risk. Plus, inserting a DOM element and measuring it may introduce a delay itself, negating the benefit of the fastClick.

ethan-james:
Here's what I see on an iPhone running 18.1.1. You can see that I can tap to the end of the editable div, or to the beginning, or select all.

https://github.com/user-attachments/assets/a0663115-4c8e-4499-be08-7e1d7696aa75

ethan-james:
I just tried it with longer text in the editable div, and I see that you can also tap to the end of any given word.

raineorshine:
> I just tried it with longer text in the editable div, and I see that you can also tap to the end of any given word. 🤦

Yes, that too! In that case, I really don't recommend trying to programmatically recreate the default caret offset behavior on tap.

ethan-james:
> > I just tried it with longer text in the editable div, and I see that you can also tap to the end of any given word. 🤦
> 
> Yes, that too! In that case, I really don't recommend trying to programmatically recreate the default caret offset behavior on tap.

Yes, I would like to avoid it as well! I'm looking into why the `focus` delay exists, so hopefully that will bear fruit.

raineorshine:
Thanks. Let me know if any of the existing selection logic needs clarification. I may not be able to explain all the details that have built up over time, but I'll do my best to elucidate anything that is particularly confusing.

ethan-james:
I found out that `react-dnd` was using this option: `{ delayTouchStart: TIMEOUT_LONG_PRESS_THOUGHT }` which was causing about a 300ms delay before focus. I removed it and I think drag & drop is still working OK, although there may be specific reasons why that delay was included in the first place.

There still may be about a 100ms delay between touch events & focus.

Screenshot 2025-01-09 at 10 56 22

It looks a bit different from the CodePen

Screenshot 2025-01-09 at 11 20 52

but there could be React or app-related overhead baked into that.

raineorshine:
Good catch! Interesting that that changed the `touchend` timing. We want the long press and drag-and-drop timing to remain unchanged from the user's perspective, so hopefully this doesn't affect that.

ethan-james:
Long press timing should be unaffected, and I tested drag-and-drop some more. Without any `delayTouchStart`, drag-and-drop works well but takes control immediately and doesn't allow drawing gestures on top of editable elements.

If I set `delayTouchStart` to 100ms, then everything seems to feel similar to how it was before. Gestures can be drawn, presumably the window for allowing them is smaller but I can't really tell from playing around with it.

There was always a 100ms delay (even with `delayTouchStart` set to `0`) and I haven't been able to locate the cause of that. It seems remarkably consistent in its timing, so presumably it's intentional on some level. I tried to remove `react-dnd` entirely to see if the delay went away, and as best I could tell it did not. That was a messy process, so I'm not terribly confident in the result that I got. This delay also affects non-editable elements such as the hamburger menu. The 300ms delay that was plaguing the hamburger menu is also down to 100ms, but it is still affected.

So as things stand, I think that setting `delayTouchStart: 100` is the way to go. In addition to the possible inevitability of that delay, it also restores behavior like drawing gestures on top of thoughts. I haven't committed that change yet, in case you have a different view.

raineorshine:
Thanks for your thoroughness and excellent troubleshooting. 

Apologies if this question has been answered already, but why is `touchend` short circuited? Maybe if the default `touchend` was allowed then `onFocus` wouldn't be needed. However I haven't reviewed the intricacies of the different events that are fired in this case so I may be missing something. 

While your proposed path forward seems reasonable given the limitations and challenges, I have two concerns:

1. We still don't actually know the cause of the 100ms delay.
2. When `touchend` fires, drag-and-drop should not be relevant anymore. A short tap won't trigger a long press or drag-and-drop. Thus it's not clear why drag-and-drop would be interfering with touchend timing. (Or is it interfering with `onFocus` timing?)

If you can offer any insight into the above, it would be greatly appreciated. Also please let me know if the time you have invested in this issue has exceeded reasonable expectations for the original scope. I'd like to get to the bottom of this, and I'm willing to allocate more resources to make it happen.

ethan-james:
If I understand the question correctly, it was being short circuited before when I was calling `preventDefault` inside of `onTap` to prevent `onFocus` from ever firing. Now it's not being short circuited, because it isn't preventing the default behavior, which is the continuation of the event chain. I'm not sure if `touchend` does anything by itself, except maybe open the little context menu. `focus` is where the browser selection is created natively.

`touchend` is firing right away, it's the `mousedown`/`mouseup` & subsequent `focus` event that are being delayed. I see your point about not really understanding why drag & drop would be interfering with non-touch events, and the only evidence I have is that changing `delayTouchStart` definitely changes the gap between touch & mouse/focus events. That delay allows other consumers to use the touch events, such as gesture drawing.

I think I can manage one more deep dive into this before we run up against reasonable expectations for scope. It looks like the Safari profiler contains pretty good timing information about when the various events are triggered, and hopefully I can decipher that output to get a sense for what's causing the delay.

ethan-james:
So it doesn't look like it was actually related to touch events after all. Using the profiler, I found out that there was a 100ms timer being set in `initEvents.ts`:

```
  /** MouseMove event listener. */
  const onMouseMove = _.debounce(
    () => distractionFreeTypingStore.update(false),
    durations.get('distractionFreeTypingThrottle'),
    {
      leading: true,
    },
  )
```

that was preventing `mousemove` from ending and moving on to `mousedown`. You can see in the first profile log where it says `Timer 61 installed - 100ms delay` and then `mousedown` doesn't fire until after `Timer 61 fired` 100ms later.

Screenshot 2025-01-11 at 12 25 14

You can see that without that debounced event listener, there's more like a 30ms delay between `mousemove` and `mousedown`. I don't see anything obvious that would indicate that this 30ms delay is intentional. I had `delayTouchStart` set to 100ms locally (to facilitate drawing gestures on top of thoughts) but setting that to 0 still results in a 30-35ms delay between `mousemove` & `mousedown`.

Screenshot 2025-01-11 at 12 23 20

raineorshine:
> If I understand the question correctly, it was being short circuited before when I was calling `preventDefault` inside of `onTap` to prevent `onFocus` from ever firing. Now it's not being short circuited, because it isn't preventing the default behavior, which is the continuation of the event chain. I'm not sure if `touchend` does anything by itself, except maybe open the little context menu. `focus` is where the browser selection is created natively.

Okay, thanks! The question is not so relevant if `onFocus` is where the magic happens.

> So it doesn't look like it was actually related to touch events after all. Using the profiler, I found out that there was a 100ms timer being set in `initEvents.ts`:
> ...
> that was preventing `mousemove` from ending and moving on to `mousedown`. You can see in the first profile log where it says `Timer 61 installed - 100ms delay` and then `mousedown` doesn't fire until after `Timer 61 fired` 100ms later.

I see the results, but I'm not sure how much I trust them. What it's showing doesn't seem possible to me. `_.debounce` sets a timeout for the callback and returns immediately. There's no way that it could prevent another event handler.

I may be confused as to why `mousemove` is being called at all. Doesn't this task only involve a tap? There is no `mousemove` or `touchmove` that fires when just tapping an element.

Another idea is to set `delayTouchStart: 0` and find a way to emulate the delay in our code rather than letting react-dnd do it.

ethan-james:
Changing `delayTouchStart` should not affect long press behavior (as long as we're not re-using the same constant) and if we reduce `delayTouchStart` to 0 then we won't be able to draw gestures on top of thoughts anymore. Maybe that's OK if it's intentional.

As for `mousemove`, it seems to always fire on a clickable element. From https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html:

Screenshot 2025-01-15 at 13 42 57

raineorshine:
> Changing `delayTouchStart` should not affect long press behavior (as long as we're not re-using the same constant) and if we reduce `delayTouchStart` to 0 then we won't be able to draw gestures on top of thoughts anymore. Maybe that's OK if it's intentional.

No, we still want to be able to gesture anywhere.

> As for `mousemove`, it seems to always fire on a clickable element. From [developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html):

Thanks, that's good to know.

ethan-james:
As for decoupling drag-and-drop behavior from onFocus, I think we're limiting the scope of that discussion to

1. options supported by `react-dnd` or
2. replacing/overriding `react-dnd` behavior

Is that what you meant? It seems to me that we consider `touchend` to be more important than `touchstart` because `touchend` signals that neither a drag-and-drop nor a gesture is occurring. The `react-dnd` options are here: https://react-dnd.github.io/react-dnd/docs/backends/touch and I'll try to see if any of the options other than `delayTouchStart` are helpful.

raineorshine:
Yes, theoretically it should be possible to trigger the tap as soon as `touchend` fires, since we know that a drag-and-drop is not occurring. I'm not sure what react-dnd supports out of the box for this though.

ethan-james:
I can't find any discussion of this online, but I modified my CodePen to show pretty conclusively that simply setting a timeout in the `touchstart` handler delays execution of the `focus` event (after it delays `mouse` events)

https://codepen.io/ethan-james-the-sasster/pen/ByBYqrL

Let me know whether or not you see the delay!

ethan-james:
Even more amazingly, it seems to delay events up until 400ms, and then not above that threshold. I updated the CodePen again to demonstrate.

ethan-james:
Here is some other poor soul who went through the same thing and dug way deeper into WebKit to figure out what was going on: https://cldfire.dev/blog/when-settimeout-is-blocking/

raineorshine:
Wow! This is one of the craziest browser behaviors I've seen in a long time. It actually makes sense now in terms of a way to make hover events backwards compatible with touch screens. Fascinating. 

> I can't find any discussion of this online, but I modified my CodePen to show pretty conclusively that simply setting a timeout in the `touchstart` handler delays execution of the `focus` event (after it delays `mouse` events)
> 
> 
> 
> https://codepen.io/ethan-james-the-sasster/pen/ByBYqrL
> 
> 
> 
> Let me know whether or not you see the delay!

Yes, indeed. 401 is fast, 400 is slow. Crazy. 

> Here is some other poor soul who went through the same thing and dug way deeper into WebKit to figure out what was going on: https://cldfire.dev/blog/when-settimeout-is-blocking/

And here's a juicy discovery:

> This is handy stuff, and ultimately what I was able to use to get out of the debacle I started with. role="button" and various other ARIA roles (full list contained in the isARIAControl method) can be used to easily opt out of this content observation behavior, eliminating the delay involved in waiting on the setTimeout callback to finish executing.

I added `role=button` to the slow contenteditable in the CodePen and the delay went away! Could you try to apply this in **em**?

ethan-james:
Looks like it worked!

raineorshine:
Wowwwww. What a payoff!!! 😄

In fact, it's so good that I don't think `fastClick` is even needed any more! 😮 I tested `role='button'` with `onClick` and it's lightning fast ⚡️

Honestly, this is a huge win. It's going to make the app really feel native. Thank you 🙏

raineorshine (approved):

PR #2754, opened 2025-01-05
Editable: cancel tap if it moves beyond threshold

### LINKED ISSUE #1453: [iOS PWA] clearThought on top of thought does not set browser selection
## Steps to Reproduce

1. Create a thought 
2. Swipe right-left on top of the thought to activate `clearThought`

## Current Behavior

Keyboard opens, but browser selection is not set. Typing will do nothing. 

## Expected Behavior

`clearThought` should work regardless of where the swipe occurs.

ethan-james:
Fixes #1453

When a touch begins on an `Editable`, the `onTap` handler fires even if the touch has moved far away before it ends. `fastClick` already contains a `MOVE_THRESHOLD` that will cancel the touch after it moves a certain distance.

Executing `clearThought`, then closing the keyboard, then executing `clearThought` again will result in the keyboard failing to open the 2nd time. This is the same issue that popped up in #2752 and is fixed by that PR. Specifically, by calling `contentRef.current?.focus()` after `selection.set`.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -49,6 +49,7 @@ import addEmojiSpace from '../util/addEmojiSpace'
 import containsURL from '../util/containsURL'
 import ellipsize from '../util/ellipsize'
 import equalPath from '../util/equalPath'
+import fastClick from '../util/fastClick'
 import head from '../util/head'
 import isDivider from '../util/isDivider'
 import strip from '../util/strip'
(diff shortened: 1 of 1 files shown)


trevinhofmann (approved):
Thank you, @ethan-james. This change appears to have resolved the issue for me on iOS (both Safari and Capacitor):

https://github.com/user-attachments/assets/8963decf-ebe6-4182-96e5-86c6636a46d8

raineorshine (approved):
Thank you!

PR #2755, opened 2025-01-05
Forurth(final) bunch of undefined thought refactor

snqb:
needs some discussions


src/actions/archiveThought.ts
```diff
@@ -93,7 +93,8 @@ const archiveThought = (state: State, options: { path?: Path }): State => {
   /** Gets the previous sibling context in the context view. */
   const prevContext = () => {
     const thoughtsContextView = thoughtsEditingFromChain(state, path)
-    const contexts = showContexts ? getContextsSortedAndRanked(state, headValue(state, thoughtsContextView)) : []
+    const head = headValue(state, thoughtsContextView)
+    const contexts = showContexts && head ? getContextsSortedAndRanked(state, head) : []
```
snqb, line 97: my thinking is that if head value is either `undefined` or `''`, we don't want to `getContextsSortedAndRanked` it, and just go with `[]`

src/actions/cursorForward.ts
```diff
@@ -21,8 +21,9 @@ const cursorForward = (state: State) => {
 
   // context view
   let cursorNew, isValidChild
-  if (showContexts) {
-    const contexts = getContextsSortedAndRanked(state, headValue(state, cursor))
+  const cursorValue = headValue(state, cursor)
+  if (showContexts && cursorValue) {
```
snqb, line 25: and here I'm uncertain, as I don't like to put this here, but feels like `contexts[0]` below it already implies we shouldn't go there with a falsy cursorValue, but dunno

src/actions/deleteEmptyThought.ts
```diff
@@ -147,7 +147,11 @@ export const deleteEmptyThoughtActionCreator: Thunk = (dispatch, getState) => {
   const uneditable = contextOfCursor && findDescendant(state, head(cursor), '=uneditable')
 
   if (prevThought && uneditable) {
-    dispatch(error({ value: `'${ellipsize(headValue(state, cursor))}' is uneditable and cannot be merged.` }))
+    dispatch(
+      error({
+        value: `'${ellipsize(headValue(state, cursor) ?? 'MISSING_THOUGHT')}' is uneditable and cannot be merged.`,
```
snqb, line 152: i dunno, feels like these cases could be just `!`-ed, if we don't want to give some information token like `~missing/empty/??~/

- i don't want to create a variable for this
- fallbacking like this in ellipsize, and making it go through it also is an eye-sore, but i dont want to extract a variable, again
- I also thought maybe an `ellipsize` could take `str | undefined`, but it sounds so evil

so wdyt? :D

src/components/Tutorial/TutorialStepAutoExpand.tsx
```diff
@@ -86,9 +86,9 @@ const TutorialStepAutoExpand = () => {
               <>
                 {' '}
                 to hide
-                {(isCursorLeaf ? cursorValue : cursorChildValue).length === 0 && ' the empty '} subthought
+                {(isCursorLeaf ? cursorValue : cursorChildValue)?.length === 0 && ' the empty '} subthought
                 {ellipsize(
-                  isCursorLeaf && cursorValue.length > 0
+                  isCursorLeaf && cursorValue && cursorValue.length > 0
```
snqb, line 91: I'm not sure about places like this
previously we assumed `string` and it was wise to be like `length > 0`
but now it's `string | undefined` and `isCursorLeaf && cursorValue`  could be viable, though implicit. Maybe sounds hacky, but `cursorValue && cursorValue.length > 0` is so ugly and meaningless

src/components/Tutorial/TutorialStepSecondThoughtEnter.tsx
```diff
@@ -13,7 +13,7 @@ const TutorialStepSecondThoughtEnter = () => {
   const ready = useSelector(state => {
     if (!state.cursor) return true
     const headCursorValue = headValue(state, state.cursor)
-    return headCursorValue?.length > 0
+    return headCursorValue && headCursorValue?.length > 0
```
snqb, line 16: `headCursorValue?.length > 0` is not something we can compare,
so again, ugly

src/actions/archiveThought.ts
raineorshine: We should just check for `undefined`. Checking for empty string is outside the scope of this PR.

src/actions/cursorForward.ts
raineorshine: Similar to above, we want to check explicitly for undefined. Falsey check includes empty string, which we don't want to do. In general, `contexts` should default to `[]` if the thought no longer exists.

src/actions/archiveThought.ts
snqb: ok, that totally makes sense, thanks
I've also changed all other places to check against `value !== undefined`, so we can allow "" thoughts

src/components/Tutorial/TutorialStepAutoExpand.tsx
raineorshine: This approach looks fine, just make sure to check for undefined explicitly.

src/components/Tutorial/TutorialStepSecondThoughtEnter.tsx
raineorshine: That's fine.

src/actions/archiveThought.ts
```diff
@@ -93,7 +93,8 @@ const archiveThought = (state: State, options: { path?: Path }): State => {
   /** Gets the previous sibling context in the context view. */
   const prevContext = () => {
     const thoughtsContextView = thoughtsEditingFromChain(state, path)
-    const contexts = showContexts ? getContextsSortedAndRanked(state, headValue(state, thoughtsContextView)) : []
+    const head = headValue(state, thoughtsContextView)
```
raineorshine, line 96: Let's just call this `value`. I don't want it to be confused with the `head` function.

src/actions/archiveThought.ts
```diff
@@ -110,7 +111,8 @@ const archiveThought = (state: State, options: { path?: Path }): State => {
   /** Gets the next sibling context in the context view. */
   const nextContext = (): ThoughtContext => {
     const thoughtsContextView = thoughtsEditingFromChain(state, path)
-    const contexts = showContexts ? getContextsSortedAndRanked(state, headValue(state, thoughtsContextView)) : []
+    const head = headValue(state, thoughtsContextView)
```
raineorshine, line 114: Let's just call this `value`. I don't want it to be confused with the `head` function.

src/actions/deleteThoughtWithCursor.ts
```diff
@@ -25,10 +25,11 @@ import parentOf from '../util/parentOf'
 import reducerFlow from '../util/reducerFlow'
 
 /** Given a path to a thought within the context view (a/m~/b), find the associated thought (b/m). This is nontrivial since the associated thought (b/m) is a different Lexeme instance than the context view thought (a/m). */
-const getContext = (state: State, path: Path) =>
-  getContexts(state, headValue(state, parentOf(path))).find(
-    cxid => getThoughtById(state, cxid)?.parentId === head(path),
-  )
+const getContext = (state: State, path: Path) => {
+  const contextValue = headValue(state, parentOf(path))
+  const context = contextValue !== undefined ? getContexts(state, contextValue) : []
```
raineorshine, line 30: This should be `contexts` (plural)

raineorshine (changes requested):
Looks good!

src/actions/archiveThought.ts
raineorshine: @snqb

raineorshine (changes requested):
Missed one

raineorshine (approved):

raineorshine:
Nice work, thanks!

Looks like my estimate was pretty accurate, too.

Issue #2338, opened 2024-08-28
Consolidate Redux test store with global app store


raineorshine:
Depends on: #2336, #2337

Once the `updateUrlHistoryMiddleware` and `storageCacheStoreEnhancer` have been fixed in the Redux test store, it should be combined with the [global app store](https://github.com/cybersemics/em/blob/cca37c8f2a2e39aadb465d981b327fd6047ae430/src/stores/app.ts). This will prevent drift between the tests and production app if any middleware or store enhancers are added in the future.

The logic to skip the tutorial can be moved into a separate `skipTutorial` test helper.

mstrofbass:
My solution to this was to keep `createTestStore` and and change it to:

1. Load the global store
2. Call the `clear` action
3. Dispatch the tutorial skipping actions
4. Return the global store

#2 ended up being necessary because the store persists between tests in the same file. This is not great but it's not immediately clear to me how best to completely recreate the store between tests since it is created at import time.

raineorshine:
Thanks! I'm glad you found a solution to the enhancers not working.

I like the idea of going all-in on the global store, but I think the proposed architecture has a couple downsides:

1. The function is called `createTestStore`, but it only creates a store indirectly by importing the global store. When called multiple times, it does not create the store at all. Thus there is a mismatch between the name and the behavior.
2. `createTestStore` does some additional setup by skipping the tutorial. This is okay, but it's not entirely obvious. The original issue suggests separating this functionality to make this clearer.

Both of these concerns make `createTestStore` a bit counterintuitive. While #2742 eliminates the redundancy between the global store and the test store with the minimal amount of code change, I don't think it improves the overall organization of the codebase. The developer still needs to understand the difference between the global store and `createTestStore`. It's not immediately obvious why `createTestStore` needs to be used instead of just importing the global store.

I think following test conventions of a `setup` and `teardown` step makes more sense here. `skipTutorial` fits perfectly into setup and `clear` fits into teardown. We can even move `skipTutorial` into the setup function like we do in the [puppeteer setup](https://github.com/cybersemics/em/blob/1de83fc608b916d44f83caf39032d3f5442b7b01/src/e2e/puppeteer/setup.ts#L67-L77). Using more conventional setup/teardown steps will be familiar to most developers and using the global store as-is eliminates the `createTestStore` confusion.

[...]

Issue #2337, opened 2024-08-28
Enable storageCacheStoreEnhancer in Redux test store


raineorshine:
Enable the `storageCacheStoreEnhancer ` in the Redux test store and fix any errors.

The test should ideally use the same store that is used in production. Missing features in the test environment should be mocked/shimmed so the tests run on the normal application code.

https://github.com/cybersemics/em/blob/3409ccc29b521c444fe4561e1c968621aade7f1c/src/test-helpers/createTestStore.ts#L43-L44

mstrofbass:
Okay, it looks like we're getting a different error now. What appears to be happening is that we're getting into `app.ts` because different files (e.g., `commandStateStore`, `prettyPath.ts`, etc.) end up loading `src/stores/app.ts`, which calls `createStore` and throws the following unhelpful message:

```
TypeError: b is not a function
 ❯ node_modules/redux/src/compose.ts:36:53
 ❯ node_modules/redux/src/compose.ts:36:51
 ❯ node_modules/redux/src/compose.ts:36:51
 ❯ node_modules/redux/src/compose.ts:36:51
 ❯ Module.createStore node_modules/redux/src/createStore.ts:89:12
 ❯ src/stores/app.ts:49:15
     47| )
     48| 
     49| const store = createStore(
       |               ^
     50|   appReducer,
     51|   composeEnhancers(
```

I'm not really sure what's going on here, but should we be concerned that we're calling `createStore` in `src/stores/app.ts` while also trying to call `createStore` in `createTestStore`?

Issue #2336, opened 2024-08-28
Enable updateUrlHistoryMiddleware in Redux test store


raineorshine:
Enable the `updateUrlHistoryMiddleware` in the Redux test store and fix any errors.

The test should ideally use the same store that is used in production. Missing features in the test environment should be mocked/shimmed so the tests run on the normal application code.

https://github.com/cybersemics/em/blob/3409ccc29b521c444fe4561e1c968621aade7f1c/src/test-helpers/createTestStore.ts#L29-L30

PR #2756, opened 2025-01-06
Consolidate test store with global app store 

mstrofbass:
My theory is that the reason the enhancers wouldn't work with the test store is because of some weird import issues that resulted in the real createStore being called in src/stores/app.ts. We can bypass fixing https://github.com/cybersemics/em/issues/2336 and https://github.com/cybersemics/em/issues/2337 by just going straight to the global store.

This change does the following:

1. Adds a helper for clearing the store
2. Adds a helper for skipping the tutorial
3. Adds a helper for initializing the store that defaults to clearing the store and skipping the tutorial
4. Replaces all `store` references that originate with `createTestStore` with references directly to the store
5. Calls `initStore` in the `beforeEach` when appropriate.

Closes https://github.com/cybersemics/em/issues/2336
Closes https://github.com/cybersemics/em/issues/2337
Closes https://github.com/cybersemics/em/issues/2338


src/test-helpers/clearStore.ts
```diff
@@ -0,0 +1,9 @@
+import { clearActionCreator as clear } from '../actions/clear'
+import store from '../stores/app'
+
+/** Dispatches actions on the global store in an act block. */
```
raineorshine, line 4: I would suggest removing this file and inlining `store.dispatch(clear())` to reduce the surface area of the internal API. If we need to factor out it out in the future we can.

src/test-helpers/skipTutorial.ts
```diff
@@ -0,0 +1,14 @@
+import store from '../stores/app'
+
+/** Dispatches actions on the global store in an act block. */
```
raineorshine, line 3: Let's inline this function as well since it's not used anywhere else.

src/test-helpers/initStore.ts
```diff
@@ -0,0 +1,18 @@
+import clearStore from './clearStore'
+import skipTutorialHelper from './skipTutorial'
+
+interface Params {
+  persist?: boolean
```
raineorshine, line 5: You can add JSDOCs to `persist` and `allowTutorial`. Then they will provide helpful popups when using IntelliSense.

raineorshine (changes requested):
Thanks, this looks good to me. It accomplishes the goal of consolidating the test store with the global app store.

At this time there appear to be little difference between a single `beforeEach(initStore)` and a paired `beforeEach(skipTutorial)` and `afterEach(clear)`. We can keep them combined until there is a need otherwise.

raineorshine:
Hi @mstrofbass. Just checking in on this. Could you take a look at the few small changes requested above? Then we should be able to get this merged. Thanks!

src/test-helpers/clearStore.ts
mstrofbass: Yeah, I was on the fense about this one.

mstrofbass:
Okay, should be good to go now.

raineorshine (approved):

Issue #1785, opened 2024-01-07
Toolbar box-shadow mask disappears when keyboard is up


raineorshine:
Note: This is related to a known problem on mobile Safari with `position: fixed`.

- **Mobile Safari** - Broken ✗ 
- **Xcode Simulator** - Works ✓ 
- **iPhone** - Works ✓
- **Android** - Works ✓

## Steps to Reproduce

Set the caret on a thought to enter edit mode and open the virtual keyboard.

## Current Behavior

The drop shadow on the toolbar disappears when the toolbar is switched from `position: fixed` to `position: absolute`.

https://github.com/cybersemics/em/assets/750276/985acfdb-4813-4bfb-8319-8e92fefb5a00

## Expected Behavior

Drop shadow should always be rendered on the toolbar.

msdewitt:
@raineorshine I'm unable to replicate this error in my local.

Could you provide more information?

![Simulator Screen Recording - iPhone 15 Pro - 2024-11-07 at 12 11 11](https://github.com/user-attachments/assets/3767cc7a-56be-40bb-8d45-27fa5e5f6b8f)

raineorshine:
Platform?

msdewitt:
Safari, Xcode simulator: Iphone 15 pro

raineorshine:
Thanks. It's actually already tested on iPhone Simulator where it works correctly (see original issue description above). We know that it is broken in mobile Safari, but it has not yet been tested on a physical iPhone or Android.

If it's only broken on mobile Safari, then we can focus troubleshooting on that platform, but first we need to test iPhone and Android.

msdewitt:
I'll test it on my android browser.

raineorshine:
@trevinhofmann Could you test this out on a physical iPhone?

trevinhofmann:
@raineorshine - The drop shadow is subtle with the new background displayed on the right side, but I have tested with these results:

✅ Physical Android device (Chrome):

https://github.com/user-attachments/assets/1f1bc8df-2e32-4260-b2e9-059229d04618

❌ Physical iOS device (Safari):

https://github.com/user-attachments/assets/be3a5d0f-85b5-4b5d-8a60-09c5cae4ab98

raineorshine:
Thanks!

In the iOS screenshot, I see a url bar... is that the Capacitor build? Mobile Safari has already been tested, so we're looking to test the Capacitor iOS build on a physical iPhone. 

[...]

PR #2760, opened 2025-01-07
Toolbar box-shadow mask disappears when keyboard is up

msdewitt:
Fix for issue: [1785](https://github.com/cybersemics/em/issues/1785)

Screenshot 2025-01-07 at 12 02 27 PM

Need to keep an eye on these changes to make sure no regression is introduced, but otherwise this seems to be a solid fix for keeping the box shadow on safari browser.

### DIFF
--- a/src/e2e/puppeteer-environment.ts
+++ b/src/e2e/puppeteer-environment.ts
@@ -29,7 +29,7 @@ const PuppeteerEnvironment: Environment = {
       .catch((err: Error) => {
         // using `console.log` here to avoid errors or logs being swallowed by vitest
         // all of `console.error`, `console.warn` and `console.info` don't show up in the terminal
-        // eslint-disable-next-line no-console
+
         console.log('Could not connect to browserless.')
         throw err
       })
--- a/src/e2e/puppeteer-environment.ts
+++ b/src/e2e/puppeteer-environment.ts
@@ -29,8 +29,6 @@ const PuppeteerEnvironment: Environment = {
       .catch((err: Error) => {
         // using `console.log` here to avoid errors or logs being swallowed by vitest
         // all of `console.error`, `console.warn` and `console.info` don't show up in the terminal
-
-        console.log('Could not connect to browserless.')
         throw err
       })
 
--- a/src/e2e/puppeteer-environment.ts
+++ b/src/e2e/puppeteer-environment.ts
@@ -29,6 +29,8 @@ const PuppeteerEnvironment: Environment = {
       .catch((err: Error) => {
         // using `console.log` here to avoid errors or logs being swallowed by vitest
         // all of `console.error`, `console.warn` and `console.info` don't show up in the terminal
+        // eslint-disable-next-line no-console
+        console.log('Could not connect to browserless.')
         throw err
       })
 
--- a/src/e2e/puppeteer/setup.ts
+++ b/src/e2e/puppeteer/setup.ts
@@ -50,7 +50,6 @@ const setup = async ({
         break
       case 'info':
       case 'log':
-        console[messageType](text)
         break
       // ConsoleMessage 'warning needs to be converted to native console 'warn'
       case 'warn':
--- a/src/e2e/puppeteer/setup.ts
+++ b/src/e2e/puppeteer/setup.ts
@@ -50,7 +50,6 @@ const setup = async ({
         break
       case 'info':
       case 'log':
-        // eslint-disable-next-line no-console
         console[messageType](text)
         break
       // ConsoleMessage 'warning needs to be converted to native console 'warn'
--- a/src/e2e/puppeteer/setup.ts
+++ b/src/e2e/puppeteer/setup.ts
@@ -50,6 +50,8 @@ const setup = async ({
         break
       case 'info':
       case 'log':
+        // eslint-disable-next-line no-console
+        console[messageType](text)
         break
       // ConsoleMessage 'warning needs to be converted to native console 'warn'
       case 'warn':
--- a/src/hooks/usePositionFixed.ts
+++ b/src/hooks/usePositionFixed.ts
@@ -27,7 +27,7 @@ const initEventHandler = once(() => {
 /** Emulates position fixed on mobile Safari with positon absolute. Returns { position, overflowX, top } in absolute mode. */
 const usePositionFixed = (): {
   position: 'fixed' | 'absolute'
-  overflowX?: 'hidden' | 'visible'
+  overflowX?: 'visible'
   top: string
 } => {
   const position = positionFixedStore.useState()
@@ -37,7 +37,7 @@ const usePositionFixed = (): {
 
   return {
     position: position ?? 'fixed',
-    overflowX: position === 'absolute' ? 'hidden' : 'visible',
+    overflowX: 'visible',
     /* spacing.safeAreaTop applies for rounded screens */
     top: position === 'absolute' ? `${scrollTop}px` : token('spacing.safeAreaTop'),
   }
--- a/src/hooks/usePositionFixed.ts
+++ b/src/hooks/usePositionFixed.ts
@@ -27,7 +27,6 @@ const initEventHandler = once(() => {
 /** Emulates position fixed on mobile Safari with positon absolute. Returns { position, overflowX, top } in absolute mode. */
 const usePositionFixed = (): {
   position: 'fixed' | 'absolute'
-  overflowX?: 'visible'
   top: string
 } => {
   const position = positionFixedStore.useState()
@@ -37,7 +36,6 @@ const usePositionFixed = (): {
 
   return {
     position: position ?? 'fixed',
-    overflowX: 'visible',
     /* spacing.safeAreaTop applies for rounded screens */
     top: position === 'absolute' ? `${scrollTop}px` : token('spacing.safeAreaTop'),
   }
--- a/src/hooks/usePositionFixed.ts
+++ b/src/hooks/usePositionFixed.ts
@@ -24,7 +24,7 @@ const initEventHandler = once(() => {
   }
 })
 
-/** Emulates position fixed on mobile Safari with positon absolute. Returns { position, overflowX, top } in absolute mode. */
+/** Emulates position fixed on mobile Safari with position absolute. Returns { position, top } in absolute mode. */
 const usePositionFixed = (): {
   position: 'fixed' | 'absolute'
   top: string
--- a/index.html
+++ b/index.html
@@ -25,6 +25,7 @@
 
   </head>
   <body>
+    <div style="position: relative; overflow: hidden;">
     <div id="root"></div>
 
     <script>
@@ -45,5 +46,6 @@
       document.body.setAttribute('data-env', navigator.webdriver ? 'test' : '%MODE%')
     </script>
     <script type="module" src="/src/index.tsx"></script>
+  </div>
   </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -25,7 +25,6 @@
 
   </head>
   <body>
-    <div style="position: relative; overflow: hidden;">
     <div id="root"></div>
 
     <script>
@@ -46,6 +45,5 @@
       document.body.setAttribute('data-env', navigator.webdriver ? 'test' : '%MODE%')
     </script>
     <script type="module" src="/src/index.tsx"></script>
-  </div>
   </body>
 </html>
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -109,7 +109,7 @@ const NavBar = ({ position }: { position: string }) => {
                       className={css({
                         position: 'relative',
                         zIndex: 'stack',
-                        bottom: '30px',
+                        bottom: '35px',
                         ...(position === 'top' && { cssFloat: 'left', marginRight: '2px' }),
                         ...(position === 'bottom' && { position: 'absolute', left: '-2px' }),
                       })}
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -109,6 +109,7 @@ const NavBar = ({ position }: { position: string }) => {
                       className={css({
                         position: 'relative',
                         zIndex: 'stack',
+                        bottom: '30px',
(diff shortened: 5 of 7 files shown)


raineorshine:
Thanks for the submission! It would be nice if it was this easy, though I have my doubts. We know it was added for a reason. I think we should track down the commit message that describes the original use of `overflowX`.

Other behavior to test on mobile Safari with keyboard open and closed:

- Toolbar scrolling
- Overscroll
- Color picker dropdown

msdewitt:
@raineorshine Alright, I finally found a solution to this issue after a few hours of work.

I managed to solve the problem with overflow by adding a relative bounds div wrap everything within the body tag, with an overflow: hidden property.

Everything seems to be working so far and I haven't found any issues with this solution so far. If you guys can find any regressions of bugs introduced by this fix, please let me know.

https://github.com/user-attachments/assets/c6edf7c5-b6d9-4fcf-b82a-2bc146ba4905

msdewitt:
@trevinhofmann I think things are working with this fix, but the puppeteer tests are all messed up.

src/e2e/puppeteer-environment.ts
```diff
@@ -29,8 +29,6 @@ const PuppeteerEnvironment: Environment = {
       .catch((err: Error) => {
         // using `console.log` here to avoid errors or logs being swallowed by vitest
         // all of `console.error`, `console.warn` and `console.info` don't show up in the terminal
-        // eslint-disable-next-line no-console
-        console.log('Could not connect to browserless.')
```
trevinhofmann, line 33: Please avoid irrelevant changes such as these.

src/e2e/puppeteer/setup.ts
```diff
@@ -50,8 +50,6 @@ const setup = async ({
         break
       case 'info':
       case 'log':
-        // eslint-disable-next-line no-console
-        console[messageType](text)
```
trevinhofmann, line 54: Please avoid irrelevant changes such as these.

src/hooks/usePositionFixed.ts
```diff
@@ -37,7 +36,6 @@ const usePositionFixed = (): {
 
   return {
     position: position ?? 'fixed',
-    overflowX: position === 'absolute' ? 'hidden' : 'visible',
```
trevinhofmann, line 40: Can you please explain this change? I'm a bit worried there will be unintended side effects from removing this attribute in the several places that use `usePositionFixed`.

trevinhofmann (changes requested):
Testing locally, this proposal does seem to resolve the toolbar box-shadow issue. It also appears to introduce regressions.

> the puppeteer tests are all messed up

Yes, there are several failing puppeteer tests. They are (mostly?) related to the Home icon being shifted down. This appears to be an unintended side effect and will need to be resolved.

Example:

font-size-18-default-initial-load-1-diff

raineorshine:
Adding this comment here, since this appears to be the bigger blocker on the suggested change:

> @raineorshine Yes, I followed it further and I think I found where and why it was added. The reason overflowX is set to hidden when it is set to absolute is because the app starts acting strange with the repositioning, giving a horizontal scroll bar on the bottom of the screen where there should be none.
> 
>  [Simulator.Screen.Recording.-.iPhone.15.-.2025-01-07.at.12.50.14.mp4](https://github.com/user-attachments/assets/ddf1e66e-50d6-41aa-9e2f-c8c6ea0d20d2)

https://github.com/cybersemics/em/issues/1785#issuecomment-2576004362

src/hooks/usePositionFixed.ts
msdewitt: This change is needed to be removed and was discussed with @raineorshine  in the ticket [1785](https://github.com/cybersemics/em/issues/1785). The reason that overflowX needed to be set when we switch to position absolute, was because there were no 'relative bounds' surrounding the app. When we switch to absolute, it moves part of the page off screen horizontally, thus giving the app a horizontal scroll bar that we don't need.

You can see the videos above attached to this pr, showing the horizontal scroll bar appear. This was a patchwork fix in the past and it needed to be reworked, because everytime the overflowX was set to hidden while the position was absolute, the boxShadow would disappear.

In order to resolve the issue of the boxShadow disappearing we can only find a way to remove this overflowX change (regarding the horizontal scrollbar), and to resolve this we need to add a relative bounds and overflow setting around all the apps which display in the root.

msdewitt:
@raineorshine and @trevinhofmann, I think I am confused about what regressions this change has introduced. I can address them as soon as I know what they are.

src/hooks/usePositionFixed.ts
raineorshine: Great explanation, thanks.

raineorshine:
Great! If the horizontal scroll bar issue is fixed, then it should just be @trevinhofmann's comments above that require addressing:

- Puppeteer failure - If your change altered the layout, you will need to adjust the layout to compensate for it.
- console.log - Those are essential logging functions for test errors.

I will test at the end to ensure there are no other regressions!

msdewitt:
@raineorshine I am not familiar with puppeteer, where should I update the snapshots with the new version in the code? I have been searching for it.

raineorshine:
I don't believe you should update any snapshots in this case. If a snapshot is failing, it means a visual regression was introduced. You can see in the diff that the home icon got shifted down. The code should be fixed so that the puppeteer tests pass and match the original snapshots.

You can find more information here: https://github.com/cybersemics/em/wiki/Testing#5-visual-snapshot-tests

msdewitt:
Then, I cannot put this change in. Because this change adds a relative div box around the app to stabilize changes between position:absolute. The entire app's screenshot is shifted slightly. Technically, this is a 'regression' because of the change compared to the reference screenshot.

raineorshine:
Maybe you could find places to adjust the CSS to compensate for the layout shifts? We ideally want to preserve the existing layout while fixing the box-shadow issue. If the `position: relative` fix causes unintended side effects, it makes sense that they may need to be compensated for.

msdewitt:
So this change is necessary because there originally was no 'relative bounds' within the app when anything was changed to position:absolute. This caused issues with another pull request of mine that will be also fixed by this pr. Because it didn't have any relative bounds, the IOS status bar would overlap with the apps. I originally put in a fix for this and was planning to push in another one. 

However, this pr fixes that issue.

Adding this relative bounds prevents position:absolute from extending the app horizontally. In the current app, we just hid the extra horizontal panels by setting OverflowX: hidden. But this is not really a good fix compared to preventing the app from extending hozontially when an element's 'position' changes.

Long story short, this is a fix to a root problem of the overflowX. 

If we don't want to push this in, I have not found another way to remove overflowX. and remove the horizontal app additions with a position change.

msdewitt:
@raineorshine Do you know any advice on how to handle updating the patch regressions then? I don't know how to make any more progress on this pr.

raineorshine:
I'm not sure what you mean by updating the patch regressions. This looks to be a CSS issue. Are you able to reproduce it in Chrome?

msdewitt:
@raineorshine No, this issue is caused by what I mentioned earlier in this thread.
Screenshot 2025-01-16 at 2 58 18 PM

`<div style="position: relative; overflow: hidden;">`

This div is newly added as a bounds around the entire app in order to fix the problem with `position: absolute` causing a horizontal shift occurring due to there not being any relative bounds within the app.

The only problem is, I can't move this CSS styling to any other places without causing other styling changes within the app. Its best as it is right now. However, because of this new div surrounding everything, the puppeteer tests are failing. 

So here comes my dilemma about what I had been asking about. I can either discard this fix and this pr completely, because the fix is relative niche as it only occurs on safari. Or we can update the saved screenshots of what the CSS used to look like to match these on coming changes.

msdewitt:
And I don't know how to update those saved screenshots with puppeteer. I am stuck.

raineorshine:
I think you might be missing the point here. Updating the screenshots is like turning off the fire alarm off when there is a fire. The puppeteer tests are trying to tell us something. In this case, it's not a fire, but a flaw in the construction. It's like the contractor neglected to put a vent in the bathroom. But in order to put in the vent, we need to move the sink and re-paint.

I can reproduce the regression very easily in Chrome. It's clearly a CSS issue. There are some other styles that might need to be adjusted or elements moved to accommodate the new relative container. Is this something you're interested in doing? If not, that's totally okay. It's up to you whether you want to continue on this. The success criteria is resolution of the box-shadow issue without any regressions. I can confidently say that it is possible. If you're not up for it, I totally understand.

msdewitt:
@raineorshine I do realize the point of the tests. However, I might be confused about how to update new CSS changes if the puppeteer service prevents differences of a certain amount.

There are 6 tests that are failing due to less than 1% difference in pixels.

I may not be able to handle this final push to fix these changes due to my lack of understanding on this aspect. I understand that puppeteer is comparing differences and warning about it, but it shouldn't be like traditional tests like unit tests where it tests the logic.

You are right that I may not be up for it, because I have been thinking about this for a bit too long to continue to admit that I can finish it according you the code review comments I was given.

raineorshine:
If these were imperceptible differences, I would totally understand that. But the visual regression is quite clear. It is a CSS issue that can be troubleshooted like any CSS issue. If you start looking at the home icon and the nav bar, I bet you would start to get some clues.

It's totally up to you though. I've appreciated your help up to this point!

## Current Behavior

4cdb704431

PR Screenshot 2025-01-16 at 10 05 58 PM

## Expected Behavior

`main`

main Screenshot 2025-01-16 at 10 05 10 PM

msdewitt:
Oh, if this is the issue it's not a difficult fix.
Is the home button supposed to be 'sticky' at the bottom of the left screen?

Currently, it exists but you need to scroll down the page to see it.

Screenshot 2025-01-16 at 4 18 30 PM


Screenshot 2025-01-16 at 4 18 25 PM

raineorshine:
> Oh, if this is the issue it's not a difficult fix.

Yes, this is what the puppeteer snapshot tests show us. You should be able to see it in the diff both on your machine when you run `yarn test:puppeteer` and in the diff artifacts that can be downloaded from GitHub Actions.

> Is the home button supposed to be 'sticky' at the bottom of the left screen?

Yes, exactly. You can test on `main` or on [staging](https://em-proto--staging2-cxls71a0.web.app/) to see the exact behavior.

raineorshine:
Hi @msdewitt. Now that we're on the same page with the snapshot tests, are you able to look into the sticky home positioning? Thanks!

msdewitt:
No problem, I'll take a look.

trevinhofmann:
Hi, @msdewitt! Thank you for the most recent changes. There are still [some failing puppeteer tests](https://github.com/cybersemics/em/actions/runs/12914334550/job/36013688879?pr=2760) which appear related to the position of the "home" icon:

```
❯ src/e2e/puppeteer/__tests__/commandPalette.ts:24:30
❯ src/e2e/puppeteer/__tests__/render-thoughts.ts:45:21
❯ src/e2e/puppeteer/__tests__/render-thoughts.ts:45:21
❯ src/e2e/puppeteer/__tests__/render-thoughts.ts:45:21
❯ src/e2e/puppeteer/__tests__/render-thoughts.ts:203:19
❯ src/e2e/puppeteer/__tests__/render-thoughts.ts:230:32
```

Please let me know if you have any questions and/or when the pull request is ready again for review.

trevinhofmann:
There were 4 changes in the latest version of the PR. I inspected each to see which ones cause regressions and which ones are needed for the fix.

| File  |  Change  | Result |
| ------------- | ------------- | ------------- |
| `index.html`  | Add a `<div style="position: relative; overflow: hidden;">` wrapper to everything | ❌ Causes the breadcrumbs to be hidden |
| `src/components/NavBar.tsx`  | Add `bottom: '35px'` style to the `<HomeLink>` | ❌ Breaks the breadcrumbs layout |
| `src/components/modals/ModalComponent.tsx`  | Add `overflow: 'visible'` and `position: 'relative'` styles | ❌ Seems to accomplish nothing (?)
| `src/hooks/usePositionFixed.ts`  | Remove `overflowX` style | ✅ Fixes the issue #1785. |

After testing, it seems that we only needed to remove the `overflowX` to fix the issue, and the other changes were causing the regressions. I have resolved this in 67bc8508626689e0732f0f3ec6994de3d6a1ad0e.

trevinhofmann (approved):

raineorshine:
Thanks for the update!

> @raineorshine -- I'm surprised that the fix would be this simple, so please test on iOS Safari to make sure I haven't missed something. In my testing, though, this one change is sufficient to resolve the toolbar box-shadow mask when the keyboard is open:

It is true that removing `overflowX` fixes the box-shadow issue, but it causes a horizontal scrollbar to appear in some circumstances. This is described in https://github.com/cybersemics/em/issues/1785#issuecomment-2576004362, although exact steps were not provided. I suspect that this was the reason `overflowX` was added to begin with.

If you're not able to reproduce, let me know and I'll give a try.

trevinhofmann:
I've been trying to reproduce the horizontal scrollbar, but haven't been able to yet. I've tried longer thoughts, rotating between landscape and portrait, etc. but nothing is causing the horizontal scrollbar to appear for me on the current branch.

Would you mind giving it a try, @raineorshine? Thank you!

trevinhofmann:
Aha!

I'm not sure why I wasn't able to reproduce that before, but I can now. There are some apparent quirks with (iOS) Safari, and we can resolve this one by simply limiting the element with the box-shadow to `maxWidth: '100%'`.

At first, I tried setting `overflowX: 'hidden'` only on the parent element. While that does work to resolve the excessive width of the box-shadow element, it also (strangely) clips the box-shadow element as well. To get around that, we'd need to add a `paddingBottom` as well, but then that parent element would be the focused element when clicking in the box shadow (when a user might be trying to click to edit a thought which is only partially obscured, for example), so that wasn't a viable option.

I dug just a bit more into it, and it seems the box-shadow element is the only one with an excessive width, so adding the `maxWidth: '100%'` to it prevents that element from causing the horizontal scrollbar to appear.

Please retest and let me know if that resolved the issue for you as well. Thanks!

raineorshine (approved):
When I attempted to fix this, I also ran into the problem of one change always breaking another. 

Glad the simple solution of `maxWidth` solves it!

Issue #2738, opened 2024-12-26
Cancel gesture diagram


raineorshine:
## Steps to Reproduce

1. Simulate mobile using the device toolbar from Chrome Dev Tools.
2. Start swiping

## Current Behavior

Cancel does not appear in the Command Palette (gesture menu).

Once the current sequence of swipes matches no gesture, the gesture menu switches to an alert that says "✗ Cancel Gesture".

## Expected Behavior

- Cancel should always appear at the end of the gesture menu.
- Once the current sequence of swipes matches no gesture, do not switch the gesture menu to an alert that says "✗ Cancel Gesture". Instead, highlight the text and svg of the the cancel gesture just like other gestures when they are selected. Otherwise, render the cancel gesture diagram in the foreground color.
- Pass null to `path` in the GestureDiagram component to indicate the Cancel gesture. Document in prop JSDOC. https://github.com/cybersemics/em/blob/5bfeca0d0823f3863d68e6b2202ccef6db8251b5/src/components/GestureDiagram.tsx#L18
- Render the cancel gesture with the same proportions and stroke width as the other gesture diagrams.
- Render with correct proportions at all font sizes.
- Add the cancel gesture to https://github.com/cybersemics/em/blob/5bfeca0d0823f3863d68e6b2202ccef6db8251b5/src/components/modals/TestGestureDiagram.tsx.

A mockup of the Cancel gesture diagram:
![image](https://github.com/user-attachments/assets/c5ad6152-dc75-4128-9f8a-fa7bccf0eb0e)

Cancel gesture diagram SVG:

`<svg width="26" height="22" viewBox="0 0 26 22" fill="none" xmlns="http://www.w3.org/2000/svg">
[...]

PR #2762, opened 2025-01-09
Cancel gesture diagram

Howlmaster:
I fixed the issue : [mobile] Cancel gesture diagram #2738

I took a look at https://github.com/cybersemics/em/pull/2745 and I added the cancel gesture to the last the gesture menu.
And now, it meets all expected behaviors.


src/commands/cancel.ts
```diff
@@ -0,0 +1,17 @@
...
+  label: 'Cancel',
+  description: 'Cancel the current gesture.',
+  gesture: undefined,
+  keyboard: { key: ',', meta: true },
+  multicursor: 'ignore',
+  svg: () => null,
+  exec: dispatch => {
+    alert('cancel')
+  },
+  allowExecuteFromModal: true,
```
raineorshine, line 14: We can omit `allowExecuteFromModal` since gestures cannot be activated when a modal is open.

src/commands/cancel.ts
```diff
@@ -0,0 +1,17 @@
+import Command from '../@types/Command'
+
+const cancelShortcut: Command = {
+  id: 'cancel',
+  label: 'Cancel',
+  description: 'Cancel the current gesture.',
+  gesture: undefined,
+  keyboard: { key: ',', meta: true },
```
raineorshine, line 8: I'm not sure why there is a keyboard command assigned. Maybe it was used for testing?

src/commands/cancel.ts
```diff
@@ -0,0 +1,17 @@
...
+  multicursor: 'ignore',
+  svg: () => null,
+  exec: dispatch => {
+    alert('cancel')
+  },
```
raineorshine, line 13: If `exec` is not used, you can use a [noop](https://github.com/cybersemics/em/blob/db00102d003112ea1d8cb04993dce3e5f77e7c85/src/constants.ts#L133). We wouldn't want to do `window.alert`.

src/components/CommandPalette.tsx
```diff
@@ -437,11 +438,18 @@ const CommandPalette: FC = () => {
             })}
           >
             {shortcuts.map(shortcut => {
-              // Check if the current gesture sequence ends with help gesture
               const helpCommand = commandById('help')
               const isHelpMatch =
                 shortcut.id === 'help' && (gestureInProgress as string)?.toString().endsWith(gestureString(helpCommand))
 
+              // Check if current gesture matches any command's gesture
+              const hasMatchingCommand = shortcuts.some(cmd => gestureInProgress === cmd.gesture)
```
raineorshine, line 446: Calling `shortcuts.some` inside `shortcuts.map` is going to result in an O(n^2) running time. You should move this check outside `map` for efficiency.

Alternatively, you might be able to use `!selectedShortcut && !isHelpMatch`. That should correspond to the cancel gesture.

raineorshine (changes requested):
Thanks for your submission! Works great. Thanks for covering all the requirements.

Below are the issues I found. Please let me know if you have any questions. Thanks!

raineorshine:
## Current Behavior

When selected in the command palette, "Cancel" is rendered with the foreground color (white).

Paste 2025-01-09 12 26 55

## Expected Behavior

The selected command should always be rendered with the highlight color (blue).

raineorshine:
## Current Behavior

The Cancel GestureDiagram is slightly too large and is positioned too close to the Help diagram.

Paste 2025-01-09 12 33 21

## Expected Behavior

The Cancel GestureDiagram should be slightly smaller and positioned evenly between the commands that come before and after, just like other commands.

raineorshine:
## Current Behavior

The Cancel GestureDiagram is highlighted (blue) even when a valid gesture (New Thought) is selected.

Paste 2025-01-09 12 33 21

## Expected Behavior

The Cancel GestureDiagram should be dim (gray) when not selected. It should only be highlighted when selected.

raineorshine:
## Steps to Reproduce

1. Decrease the font size using the font controls in the footer.
2. Activate the Command Palette by inputing a gesture.

## Current Behavior

The Cancel GestureDiagram does not scale with font size.

Paste 2025-01-09 12 25 01

## Expected Behavior

The Cancel GestureDiagram should scale with font size like the other GestureDiagrams.

raineorshine:
## Steps to Reproduce

```
yarn test:puppeteer commandPalette
```

Note that docker is required to run the Puppeteer tests.

## Current Behavior

The Puppeteer test are failing because a snapshot is out of date.

## Expected Behavior

The Puppeteer tests should all pass.

To update the snapshots for this test, run: 

```
yarn test:puppeteer commandPalette -u
```

raineorshine:
## Current Behavior

The Cancel diagram does not show up in https://github.com/cybersemics/em/blob/5bfeca0d0823f3863d68e6b2202ccef6db8251b5/src/components/modals/TestGestureDiagram.tsx.

## Expected Behavior

The Cancel diagram should be added to https://github.com/cybersemics/em/blob/5bfeca0d0823f3863d68e6b2202ccef6db8251b5/src/components/modals/TestGestureDiagram.tsx as requested in the original issue. The snapshot test should be updated once added.

raineorshine:
@Howlmaster Just checking in! Let me know if you have any questions about the requested changes.

raineorshine:
Thanks!

Looks like the `yarn.lock` was inadvertently changed.

src/e2e/puppeteer/__tests__/__image_snapshots__/commandPalette/commandPalette-light.png
raineorshine, line 1: Let's add `hideFromCommandPalette: true` to the new Cancel command to avoid having it added to the Command Palette.

raineorshine (changes requested):
I see that all of the comments were marked resolved from my previous review, yet I don't see any changes in the diff. I've unresolved them for now. Are you still working on these?

Howlmaster:
Hi @raineorshine,

I’ve resolved all the issues we discussed and pushed the updated code. You can check the git history for the changes. Please let me know if you encounter any more issues or need further adjustments.

Thank you!

src/components/CommandPalette.tsx
```diff
@@ -436,32 +446,49 @@ const CommandPalette: FC = () => {
...
-                  shortcut={shortcut}
-                />
-              )
-            })}
-            {shortcuts.length === 0 && <span className={css({ marginLeft: '1em' })}>No matching commands</span>}
+            {shortcuts.length > 0 ? (
+              <>
+                {(() => {
+                  const helpCommand = commandById('help')
+                  // const cancelCommand = commandById('cancel')
```
raineorshine, line 453: You can remove this commented out code.

src/e2e/puppeteer/__tests__/__image_snapshots__/commandPalette/commandPalette.png
raineorshine, line 1: It looks like the new highlight logic for mobile is accidentally overriding the desktop highlight logic.

Here is an example that makes it clearer:

## Steps to Reproduce

1. Activate the Command Palette on desktop (Meta + P)
2. Type "Jump"

## Current Behavior

a780a7e4aa

Screenshot 2025-01-19 at 9 54 09 AM

## Expected Behavior

`main`

Screenshot 2025-01-19 at 9 53 55 AM

raineorshine (changes requested):
Thanks addressing the requested changes! I can confirm that the spacing is fixed, the O(n^2) implementation was changed to O(2n), and several other issues were corrected.

I found two issues that do not seem to be resolved yet, and one regression. Could you take another look? Let me know if there is anything that doesn't make sense!

- [#issuecomment-2580925625](https://github.com/cybersemics/em/pull/2762#issuecomment-2580925625)
- [#issuecomment-2580945906](https://github.com/cybersemics/em/pull/2762#issuecomment-2580945906)
- regression (see below)

Note: You can ignore the [font size issue](https://github.com/cybersemics/em/pull/2762#issuecomment-2580925508), as I realized that scaling by font size is broken in `main`.

Thank you!

Howlmaster:
Hi @raineorshine.
How can I check this https://github.com/cybersemics/em/pull/2762#issuecomment-2580945906 working properly.
I mean after update this file https://github.com/cybersemics/em/blob/5bfeca0d0823f3863d68e6b2202ccef6db8251b5/src/components/modals/TestGestureDiagram.tsx.
Then where can I check this one work properly?
Could you give me some tips?

src/components/CommandPalette.tsx
```diff
@@ -210,17 +209,30 @@ const CommandRow: FC<{
                         )
                       })()
                     : // For other commands, use normal highlighting
-                      gestureInProgress.length
+                      shortcut.id === 'cancel' && selected
+                      ? 1 // Highlight the cancel gesture when any gesture is in progress
+                      : shortcut.id === 'cancel'
+                        ? undefined
+                        : gestureInProgress.length
```
raineorshine, line 216: This would be better expressed as:

```ts
    shortcut.id === 'cancel' 
      ? (selected ? 1 : undefined)
      : gestureInProgress.length
```

That way the cancel logic is consolidated and you avoid the duplication of `shortcut.id === 'cancel'`.

src/components/CommandPalette.tsx
```diff
@@ -231,7 +243,15 @@ const CommandRow: FC<{
...
-              color: disabled ? 'gray' : gestureInProgress === shortcut.gesture ? 'vividHighlight' : 'fg',
+              color: disabled
+                ? 'gray'
+                : isTouch
+                  ? selected
+                    ? 'vividHighlight'
+                    : 'fg'
+                  : gestureInProgress === shortcut.gesture
+                    ? 'vividHighlight'
+                    : 'fg',
```
raineorshine, line 254: You can eliminate the duplication here by converting the two separate inner conditions into a single union:

```ts
    : isTouch
      ? selected || gestureInProgress === shortcut.gesture
        ? 'vividHighlight'
        : 'fg',
```

raineorshine (changes requested):
Thanks! I can see the reported issues were fixed.

Could also take a look at the regression that occurs on a new, empty thoughtspace? It was the last comment here: https://github.com/cybersemics/em/pull/2762#issuecomment-2601784519

src/components/modals/TestGestureDiagram.tsx
```diff
@@ -17,6 +18,15 @@ const ModalTestGestureDiagram = () => {
       <GestureDiagram path='rdrul' size={48} arrowSize={12} />
       <GestureDiagram path='rl' size={48} arrowSize={12} />
       <GestureDiagram path='rld' size={48} arrowSize={12} />
+      <svg width={24} height={24} viewBox='0 0 24 24' style={{ marginBottom: 11, marginLeft: 2 }}>
```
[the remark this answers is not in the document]
raineorshine: Yes, that’s fine.

Howlmaster:
@raineorshine 
Thanks for your reply. I updated the code based on your feedback and pushed.

Howlmaster:
@raineorshine 
Thanks for your reply. 
I've updated and pushed the code based on your feedback.

raineorshine:
I tested this again, and I can confirm that it is a regression. I was able to reproduce it in Chrome in device emulation mode and in mobile Safari. It is not an issue in `main`. Would you mind taking another look at this?

src/components/EmptyThoughtspace.tsx
```diff
@@ -58,6 +58,9 @@ const EmptyThoughtspace = ({ isTutorial }: { isTutorial?: boolean }) => {
                 <span>
                   Swipe{' '}
                   <GestureDiagram
+                    cssRaw={css.raw({
+                      marginBottom: '-10px !important',
```
raineorshine, line 62: Thank you!

Unfortunately, this causes the entire EmptyThoughtspace component to shift up 10px, which is not desired. The empty thoughtspace message is positioned where it is so that it matches the positioning of the first thought once it is rendered. Please ensure that the positioning is unchanged from `main`.

## Current

Screenshot 2025-01-22 at 1 57 21 PM

## Expected

Screenshot 2025-01-22 at 1 57 26 PM

raineorshine: Note that `!important` is considered a bad practice, as it makes it very hard to control CSS layering once it is used. I made an exception for TestGestureDiagram since it is only for testing purposes, but I cannot make the same concession here. You will have to find a different way to accomplish this.

raineorshine (changes requested):

src/components/GestureDiagram.tsx
```diff
@@ -182,7 +204,7 @@ const GestureDiagram = ({
     <svg
       width={width || '100'}
       height={height || '100'}
-      className={css(inGestureContainer && { position: 'relative', top: '10px' }, cssRaw)}
```
raineorshine, line 185: Removing this `top` value is what broke the spacing in EmptyThoughtspace. Can you share why this was removed? Maybe we can find an alternative approach that doesn't have unintended side effect.

src/components/EmptyThoughtspace.tsx
```diff
@@ -55,9 +55,12 @@ const EmptyThoughtspace = ({ isTutorial }: { isTutorial?: boolean }) => {
           <>
             <span className={css({ userSelect: 'none' })}>
               {isTouch ? (
-                <span>
+                <span style={{ display: 'inline-block', marginTop: '10px' }}>
                   Swipe{' '}
                   <GestureDiagram
+                    cssRaw={css.raw({
+                      marginBottom: '-10px',
+                    })}
```
raineorshine, line 63: While this does eliminate `!important`, I feel like we're getting further away from the correct solution. We shouldn't have to override multiple layers of styles in order to shift the diagram down. The original `position: relative` and `top: 10px` achieved this with fewer side effects.

raineorshine (changes requested):

src/components/GestureDiagram.tsx
Howlmaster: @raineorshine 
I accidently removed top: '10px' in the first commit.
I updated this one working properly.

raineorshine (approved):
That fixed it!

Thanks for your contribution!

PR #2763, opened 2025-01-09
Table View 1842 (Milestone 1) 

ohrytskov:
#1842

### DIFF
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -1,12 +1,16 @@
 import React, { useEffect } from 'react'
 import { useDispatch } from 'react-redux'
+import { useSelector } from 'react-redux'
 import { css } from '../../styled-system/css'
 import { SystemStyleObject } from '../../styled-system/types'
 import Path from '../@types/Path'
 import { setCursorActionCreator as setCursor } from '../actions/setCursor'
 import { DIVIDER_MIN_WIDTH, DIVIDER_PLUS_PX } from '../constants'
+import attributeEquals from '../selectors/attributeEquals'
+import rootedParentOf from '../selectors/rootedParentOf'
 import fastClick from '../util/fastClick'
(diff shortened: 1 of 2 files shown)


ohrytskov:
This pull request is now ready for review.

trevinhofmann (approved):
Thank you for the pull request, @ohrytskov!

This appears to have resolved the divider positioning on desktop and mobile (also tested at various font sizes).

Screenshot 2025-01-15 at 4 03 16 AM

src/components/Divider.tsx
```diff
@@ -36,12 +40,19 @@ const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) =
 
   useEffect(setStyle)
 
+  const isTableView = useSelector(state => {
+    const parentId = head(parentOf(path)) || head(rootedParentOf(state, path))
```
raineorshine, line 44: `parentOf` and `rootedParent` do the same thing, so we don't need both.

The only difference is that `parentOf` only works if `path.length > 1`. `rootedParentOf` works on any path and can differentiate between different root contexts. I will update the JSDOC to make this more clear!

raineorshine (changes requested):

src/components/Divider.tsx
```diff
@@ -36,12 +40,19 @@ const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) =
...
+    const result = attributeEquals(state, parentId, '=view', 'Table')
+    return result
```
raineorshine, line 46: Avoid creating an additional variable here and just return the statement directly.

raineorshine (changes requested):
Otherwise looks good!

raineorshine (approved):
Thank you!

PR #405, opened 2020-03-05
set divider width

ArtMyWeb:
Fix #346
- get max width of list elements that are nearby for divider, add 30px and set this width to divider

PR #2764, opened 2025-01-09
Update divider width in real-time in table view

ohrytskov:
#543 
Built on https://github.com/cybersemics/em/pull/2763. Ignore commit https://github.com/cybersemics/em/commit/99c19d92f28dca8fcc4aca8353b9098b561f27b8 to review current changes only.

### DIFF
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -1,4 +1,4 @@
-import React, { useLayoutEffect, useRef, useState } from 'react'
+import React, { useLayoutEffect, useMemo, useRef, useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import { css } from '../../styled-system/css'
 import { SystemStyleObject } from '../../styled-system/types'
@@ -77,12 +77,18 @@ const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) =
     dispatch(setCursor({ path }))
   }
 
+  // Derive values from the arrays to use in the dependency array, including .value changes
+  const childrenData = useMemo(() => children.map(child => `${child.id}:${child.value}`).join('|'), [children])
+  const thoughtsAtSameDepthData = useMemo(
+    () => thoughtsAtSameDepth.map(thought => `${thought.id}:${thought.value}`).join('|'),
+    [thoughtsAtSameDepth],
+  )
+
   useLayoutEffect(() => {
     /** Calculates and updates the Divider's width based on sibling thought widths. */
     const updateDividerWidth = () => {
       if (!dividerRef.current) return
 
-      // If the divider is the only child and we're not in Table View
       if (isOnlyChild && !isTableView) {
         setDividerWidth(DIVIDER_MIN_WIDTH)
         return
@@ -115,7 +121,13 @@ const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) =
 
     updateDividerWidth()
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [isOnlyChild, isTableView])
+  }, [
+    isOnlyChild,
+    isTableView,
+    dividerId,
+    childrenData, // Include the derived data that reflects .value changes
+    thoughtsAtSameDepthData, // Include the derived data that reflects .value changes
+  ])
 
   return (
     <div
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useLayoutEffect, useRef, useState } from 'react'
-import { useDispatch, useSelector, shallowEqual } from 'react-redux'
+import { shallowEqual, useDispatch, useSelector } from 'react-redux'
 import { css } from '../../styled-system/css'
 import { SystemStyleObject } from '../../styled-system/types'
 import Path from '../@types/Path'
@@ -18,38 +18,34 @@ import isDivider from '../util/isDivider'
 
 /** Custom hook to fetch thought IDs that affect the divider width. */
 const useWidthDependentThoughtIds = (path: Path): ThoughtId[] => {
-  return useSelector(
-    (state: State) => {
-      const parentPath = rootedParentOf(state, path)
-      const parentId = head(parentPath)
-      const grandParentPath = parentId ? rootedParentOf(state, parentPath) : null
-      const grandParentId = grandParentPath ? head(grandParentPath) : null
-      const children = parentId ? getAllChildrenAsThoughts(state, parentId) : []
-      const childrenWithoutDividers = children.filter(child => !isDivider(child.value))
-      const isOnlyChild = childrenWithoutDividers.length === 0
-      const isTableView =
-        attributeEquals(state, parentId, '=view', 'Table') ||
-        attributeEquals(state, grandParentId, '=view', 'Table')
+  return useSelector((state: State) => {
+    const parentPath = rootedParentOf(state, path)
+    const parentId = head(parentPath)
+    const grandParentPath = parentId ? rootedParentOf(state, parentPath) : null
+    const grandParentId = grandParentPath ? head(grandParentPath) : null
+    const children = parentId ? getAllChildrenAsThoughts(state, parentId) : []
+    const childrenWithoutDividers = children.filter(child => !isDivider(child.value))
+    const isOnlyChild = childrenWithoutDividers.length === 0
+    const isTableView =
+      attributeEquals(state, parentId, '=view', 'Table') || attributeEquals(state, grandParentId, '=view', 'Table')
 
-      const dependentThoughtIds = isOnlyChild
-        ? isTableView && grandParentId
-          ? // If the thought is the only child and in a table view, get the grandchildren's IDs
-            getAllChildrenAsThoughts(state, grandParentId)
-              .filter(child => !isDivider(child.value))
-              .flatMap(parent =>
-                (parent.id ? getAllChildrenAsThoughts(state, parent.id) : [])
-                  .filter(child => !isDivider(child.value))
-                  .map(child => child.id),
-              )
-          : // If the thought is the only child but not in a table view, return an empty array
-            []
-        : // If the thought is not the only child, get the sibling thought IDs
-          childrenWithoutDividers.map(child => child.id)
+    const dependentThoughtIds = isOnlyChild
+      ? isTableView && grandParentId
+        ? // If the thought is the only child and in a table view, get the grandchildren's IDs
+          getAllChildrenAsThoughts(state, grandParentId)
+            .filter(child => !isDivider(child.value))
+            .flatMap(parent =>
+              (parent.id ? getAllChildrenAsThoughts(state, parent.id) : [])
+                .filter(child => !isDivider(child.value))
+                .map(child => child.id),
+            )
+        : // If the thought is the only child but not in a table view, return an empty array
+          []
+      : // If the thought is not the only child, get the sibling thought IDs
+        childrenWithoutDividers.map(child => child.id)
 
-      return dependentThoughtIds
-    },
-    shallowEqual,
-  )
+    return dependentThoughtIds
+  }, shallowEqual)
 }
 
 /**
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useLayoutEffect, useRef, useState } from 'react'
-import { useDispatch, useSelector } from 'react-redux'
+import { useDispatch, useSelector, shallowEqual } from 'react-redux'
 import { css } from '../../styled-system/css'
 import { SystemStyleObject } from '../../styled-system/types'
 import Path from '../@types/Path'
@@ -48,7 +48,7 @@ const useWidthDependentThoughtIds = (path: Path): ThoughtId[] => {
 
       return dependentThoughtIds
     },
-    (prev, next) => prev.length === next.length && prev.every((id, index) => id === next[index]),
+    shallowEqual,
   )
 }
 
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -116,8 +116,6 @@ const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) =
         widths = getThoughtWidths(siblingThoughts)
       }
 
-      //const maxWidth = widths.length > 0 ? Math.max(...widths) : DIVIDER_MIN_WIDTH
-      //setDividerWidth(Math.round(Math.max(maxWidth, DIVIDER_MIN_WIDTH)))
       setDividerWidth(Math.round(Math.max(...widths, DIVIDER_MIN_WIDTH)))
     }
 
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -1,5 +1,5 @@
-import React, { useCallback, useRef, useState } from 'react'
-import { useDispatch } from 'react-redux'
+import React, { useEffect, useRef, useState } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
 import { css } from '../../styled-system/css'
 import { SystemStyleObject } from '../../styled-system/types'
 import Path from '../@types/Path'
@@ -7,13 +7,50 @@ import State from '../@types/State'
 import ThoughtId from '../@types/ThoughtId'
 import { setCursorActionCreator as setCursor } from '../actions/setCursor'
 import { DIVIDER_MIN_WIDTH } from '../constants'
-import useSelectorEffect from '../hooks/useSelectorEffect'
 import attributeEquals from '../selectors/attributeEquals'
 import { getAllChildrenAsThoughts } from '../selectors/getChildren'
+import editingValueStore from '../stores/editingValue'
 import fastClick from '../util/fastClick'
 import head from '../util/head'
 import isDivider from '../util/isDivider'
 import parentOf from '../util/parentOf'
+
+/** Custom hook to fetch divider-related data from the state. */
+const useDividerData = (path: Path) => {
+  const dividerId = head(path)
+  const parentPath = parentOf(path)
+  const parentId = head(parentPath)
+  const grandParentPath = parentOf(parentPath)
+  const grandParentId = head(grandParentPath)
+
+  return useSelector((state: State) => {
+    const children = parentId ? getAllChildrenAsThoughts(state, parentId) : []
+    const childrenWithoutDividers = children.filter(child => !isDivider(child.value))
+    const isOnlyChild = childrenWithoutDividers.length === 0
+    const isTableView =
+      (parentId && attributeEquals(state, parentId, '=view', 'Table')) ||
+      (grandParentId && attributeEquals(state, grandParentId, '=view', 'Table'))
+
+    let thoughtsAtSameDepth: { id: ThoughtId; value: string }[] = []
+
+    if (isTableView && isOnlyChild && grandParentId) {
+      const parentSiblings = getAllChildrenAsThoughts(state, grandParentId)
+      thoughtsAtSameDepth = parentSiblings.flatMap(parent => {
+        const childrenOfParent = parent.id ? getAllChildrenAsThoughts(state, parent.id) : []
+        return childrenOfParent.filter(child => !isDivider(child.value))
+      })
+    }
+
+    return {
+      dividerId,
+      parentId,
+      isOnlyChild,
+      isTableView,
+      children,
+      thoughtsAtSameDepth,
+    }
+  })
+}
 
 /** Helper function to get widths of thoughts by querying the DOM. */
 const getThoughtWidths = (thoughts: { id: ThoughtId }[]) => {
@@ -21,7 +58,7 @@ const getThoughtWidths = (thoughts: { id: ThoughtId }[]) => {
     const innerThoughtElement = document.querySelector(
       `[aria-label="tree-node"][data-id="${thought.id}"] [aria-label="thought"]`,
     )
-    return innerThoughtElement ? innerThoughtElement.getBoundingClientRect().width : 0
+    return innerThoughtElement?.getBoundingClientRect().width ?? 0
   })
 }
 
@@ -31,48 +68,19 @@ const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) =
   const dividerRef = useRef<HTMLDivElement>(null)
   const [dividerWidth, setDividerWidth] = useState<number>(DIVIDER_MIN_WIDTH)
 
-  const dividerId = head(path)
-  const parentPath = parentOf(path)
-  const parentId = head(parentPath)
-  const grandParentPath = parentOf(parentPath)
-  const grandParentId = head(grandParentPath)
+  // Use the custom hook to get necessary data
+  const { dividerId, isOnlyChild, isTableView, children, thoughtsAtSameDepth } = useDividerData(path)
+  const editingValue = editingValueStore.useSelector(state => state)
 
-  /** Selector function to fetch necessary data from the state. */
-  const selectDividerData = useCallback(
-    (state: State) => {
-      const children = parentId ? getAllChildrenAsThoughts(state, parentId) : []
-      const childrenWithoutDividers = children.filter(child => !isDivider(child.value))
-      const isOnlyChild = childrenWithoutDividers.length === 0
-      const isTableView =
-        (parentId && attributeEquals(state, parentId, '=view', 'Table')) ||
-        (grandParentId && attributeEquals(state, grandParentId, '=view', 'Table')) ||
-        false
+  /** Sets the cursor to the divider. */
+  const setCursorToDivider = (e: React.MouseEvent | React.TouchEvent) => {
+    e.stopPropagation()
+    dispatch(setCursor({ path }))
+  }
 
-      let thoughtsAtSameDepth: { id: ThoughtId; value: string }[] = []
-
-      if (isTableView && isOnlyChild && grandParentId) {
-        const parentSiblings = getAllChildrenAsThoughts(state, grandParentId)
-        thoughtsAtSameDepth = parentSiblings.flatMap(parent => {
-          const childrenOfParent = parent.id ? getAllChildrenAsThoughts(state, parent.id) : []
-          return childrenOfParent.filter(child => !isDivider(child.value))
-        })
-      }
-
-      return {
-        isOnlyChild,
-        isTableView,
-        children,
-        thoughtsAtSameDepth,
-      }
-    },
-    [parentId, grandParentId],
-  )
-
-  /** Updates the Divider's width based on sibling thought widths. */
-  const updateDividerWidth = useCallback(
-    (dividerData: ReturnType<typeof selectDividerData>) => {
-      const { isOnlyChild, isTableView, children, thoughtsAtSameDepth } = dividerData
-
+  useEffect(() => {
+    /** Calculates and updates the Divider's width based on sibling thought widths. */
+    const updateDividerWidth = () => {
       if (!dividerRef.current) return
 
       if (isOnlyChild && !isTableView) {
@@ -89,7 +97,9 @@ const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) =
         }
         widths = getThoughtWidths(thoughtsAtSameDepth)
       } else {
+        // Non-Table View or Divider is not the only item
         const siblingThoughts = children.filter(child => child.id !== dividerId && !isDivider(child.value))
+
         if (siblingThoughts.length === 0) {
           setDividerWidth(DIVIDER_MIN_WIDTH)
           return
@@ -97,20 +107,15 @@ const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) =
         widths = getThoughtWidths(siblingThoughts)
       }
 
+      // Determine the maximum width
       const maxWidth = widths.length > 0 ? Math.max(...widths) : DIVIDER_MIN_WIDTH
       setDividerWidth(Math.round(Math.max(maxWidth, DIVIDER_MIN_WIDTH)))
-    },
-    [dividerRef, dividerId],
-  )
+    }
 
-  // Use useSelectorEffect to watch for changes in the necessary state slices
-  useSelectorEffect(updateDividerWidth, selectDividerData)
+    updateDividerWidth()
 
-  /** Sets the cursor to the divider. */
-  const setCursorToDivider = (e: React.MouseEvent | React.TouchEvent) => {
-    e.stopPropagation()
-    dispatch(setCursor({ path }))
-  }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isOnlyChild, isTableView, editingValue, dividerId])
 
   return (
     <div
--- a/src/components/Divider.tsx
+++ b/src/components/Divider.tsx
@@ -1,11 +1,10 @@
-import React, { useEffect } from 'react'
-import { useDispatch } from 'react-redux'
-import { useSelector } from 'react-redux'
+import React, { useLayoutEffect, useRef, useState } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
 import { css } from '../../styled-system/css'
 import { SystemStyleObject } from '../../styled-system/types'
 import Path from '../@types/Path'
 import { setCursorActionCreator as setCursor } from '../actions/setCursor'
-import { DIVIDER_MIN_WIDTH, DIVIDER_PLUS_PX } from '../constants'
+import { DIVIDER_MIN_WIDTH } from '../constants'
 import attributeEquals from '../selectors/attributeEquals'
 import rootedParentOf from '../selectors/rootedParentOf'
 import fastClick from '../util/fastClick'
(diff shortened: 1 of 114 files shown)


raineorshine:
Note: Built on #2763. Ignore commit 99c19d92f28dca8fcc4aca8353b9098b561f27b8 to review current changes only.

ohrytskov:
This PR is ready for review.

ohrytskov:
Both of the two latest commits successfully resolve the current issue. In the most recent commit, I've implemented state-based logic using selectors wherever possible instead of relying solely on DOM queries. However, since we still need to use DOM queries to measure text widths, does it make sense to combine state-based logic with selectors and DOM queries in this way? Or would it be better to use only DOM queries, as in the previous commit?

ohrytskov:
Just to clarify, the two commits I'm referring to are:

Commit 919dd24: Implemented using only DOM queries.
Commit ad33c61: Switched to state-based logic using selectors wherever possible.

src/components/Divider.tsx
```diff
@@ -1,61 +1,141 @@
...
-        return subs.length ? subs[0].offsetWidth + DIVIDER_PLUS_PX : DIVIDER_PLUS_PX
-      })
-      const maxWidth = Math.max(...widths)
-      dividerSetWidth.current.style.width = `${maxWidth > DIVIDER_MIN_WIDTH ? maxWidth : DIVIDER_MIN_WIDTH}px`
+  useLayoutEffect(() => {
+    /** Calculates and updates the Divider's width based on sibling thought widths. */
+    const updateDividerWidth = () => {
+      if (!dividerRef.current) return
+
+      // If the divider is the only child and we're not in Table View
```
trevinhofmann, line 85: Not a blocking issue, but this comment is unnecessary. Since the variables are clearly named, the comment is redundant and does not add additional information.

`if (isOnlyChild && !isTableView)` reads somewhat literally as `if is-only-child and is-not-table-view`, which is almost identical to the comment

src/components/Divider.tsx
```diff
@@ -1,61 +1,141 @@
...
+
+      // Determine the maximum width
+      const maxWidth = widths.length > 0 ? Math.max(...widths) : DIVIDER_MIN_WIDTH
+      setDividerWidth(Math.round(Math.max(maxWidth, DIVIDER_MIN_WIDTH)))
     }
-  }
 
-  useEffect(setStyle)
+    updateDividerWidth()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
trevinhofmann, line 117: @ohrytskov - Can you please explain why the dependencies `children`, `dividerId`, and `thoughtsAtSameDepth` are not required?

trevinhofmann: _Possibly_ related to `1. Divider widths do not update`.

trevinhofmann (changes requested):
Thank you for the PR, @ohrytskov!

This seems to work when pasting the given examples from the issue:

<details>
<summary>Case 1</summary>

Screenshot 2025-01-15 at 4 34 42 AM
</details>

<details>
<summary>Case 2</summary>

Screenshot 2025-01-15 at 4 34 52 AM
</details>

## 1. Divider widths do not update

The divider widths do not update while editing the thoughts.

https://github.com/user-attachments/assets/b2969d44-793a-450d-8636-6551749bf56f

ohrytskov:
Including children and thoughtsAtSameDepth in the useLayoutEffect dependency array caused an infinite re-render loop because these arrays were recreated with new references on each render. Since the hook's dependency array relies on referential equality, and the arrays were derived from selectors that created new instances of the Redux state values, the effect ran repeatedly. This triggered state updates and re-renders in a loop. To resolve this, the arrays were memoized with useMemo, ensuring their references only changed when their contents did.

ohrytskov:
@trevinhofmann, I've made the following updates:

- Refactored the `Divider` component to use `useSelectorEffect` for more efficient state updates.
- Memoized `updateDividerWidth` with `useCallback` to prevent redundant re-renders.

src/components/Divider.tsx
```diff
@@ -1,70 +1,136 @@
...
+import isDivider from '../util/isDivider'
+import parentOf from '../util/parentOf'
+
+/** Helper function to get widths of thoughts by querying the DOM. */
+const getThoughtWidths = (thoughts: { id: ThoughtId }[]) => {
+  return thoughts.map(thought => {
+    const innerThoughtElement = document.querySelector(
+      `[aria-label="tree-node"][data-id="${thought.id}"] [aria-label="thought"]`,
+    )
+    return innerThoughtElement ? innerThoughtElement.getBoundingClientRect().width : 0
```
trevinhofmann, line 24:
```suggestion
    return innerThoughtElement?.getBoundingClientRect().width ?? 0
```

This line is fine as-is, but I'll make this suggestion as personal preference for slightly better readability.

trevinhofmann (changes requested):
Thank you for the updates, @ohrytskov!

This has resolved `1. Divider widths do not update`, and Cases 1 & 2 from #543 appear to still be working correctly. I do see a couple issues related to the updating divider widths, however.

---

## 2. Divider width updates are choppy/laggy

When editing a thought, the width of the divider has quite choppy/laggy updates.

https://github.com/user-attachments/assets/ac2b3a7e-3bc7-494c-a92f-9028dfbf9174

---

## 3. Divider width does not update after deleting a space

The width of the divider appears to update after deleting non-space characters, but it does not update after deleting a space character. In the example below, the divider becomes shorter after deleting each letter in `and`, but not after deleting the space before the word.

https://github.com/user-attachments/assets/a43e086e-9361-49f2-a380-c5b9cbd3c0f7

raineorshine:
You might have to use the DOM value or `editingValueStore` to get the real-time value of the cursor. Value updates to Thoughts in the Redux store are throttled. This could be what is creating the appearance of laggy updates.

ohrytskov:
@trevinhofmann, real-time updates of the divider width have been implemented. Do we really need to account for trailing spaces in width calculations, considering that these spaces are not stored in the Redux store?

raineorshine:
It would be nice to ignore the trailing spaces, but unfortunately it's causing slightly different behavior between adding and deleting.

- When a space is typed at the end of a thought (e.g. `Hello` → `Hello `), the divider width matches the **trimmed value**. 
- When a non-space character is deleted from the end of a thought (e.g. `Hello World` → `Hello `), the divider width matches the **untrimmed value**.

This is particularly noticeable when moving the cursor away after the latter case. The divider width changes even though the thought value does not change.

I would be okay with the width of either the trimmed value or the untrimmed value, as long as it is consistent. Thanks!

---

Also noticing some warnings in the console. Could you look into those?

ohrytskov:
The divider now dynamically adjusts its width to match thoughts with trailing spaces.

ohrytskov:
Regarding the console warnings, thank you for bringing them to my attention. I'll investigate and resolve any issues causing those warnings.

ohrytskov:
Following up on my earlier message - the issue causing the console warnings has been addressed in the latest commit: 8270579.

ohrytskov:
@trevinhofmann, could you please review the latest changes when you have a moment?

trevinhofmann (changes requested):
Thank you for the updates, @ohrytskov!

The issue `2. Divider width updates are choppy/laggy` has been resolved. It is much smoother now, thanks!

I am still seeing the issue `3. Divider width does not update after deleting a space`, however:

https://github.com/user-attachments/assets/930ef77a-6479-4b59-8eae-7a545a464ce0

---

@raineorshine, the divider width functionality when wrapping to a new line looks good to me. Just wanted to show you a demo clip and make sure that it looks right to you also:

https://github.com/user-attachments/assets/d61ce99a-0837-4efa-9ec5-a7ceec07bbc6

---

@raineorshine, one more question. Should divider width be affected by sibling thoughts' widths outside of tables as well, or is this functionality only wanted for tables?

Screenshot 2025-01-22 at 12 44 35 AM

raineorshine:
> @raineorshine, the divider width functionality when wrapping to a new line looks good to me. Just wanted to show you a demo clip and make sure that it looks right to you also:

Yes, that looks good to me.

> @raineorshine, one more question. Should divider width be affected by sibling thoughts' widths outside of tables as well, or is this functionality only wanted for tables?

Yes, divider width should be set to the widest sibling outside of tables. This was implemented in #405, but apparently it has stopped working. 

@ohrytskov How hard would it be to add sibling support for non-tables? They don't need any of the uncle logic. If it's too much we can create a separate issue for it, but I imagine it could be a small change if it the logic is already there for tables.

ohrytskov:
Following the feedback, all issues have been resolved, and the divider width now accurately reflects the widest sibling outside of tables.

ohrytskov:
@trevinhofmann, could you please review the latest changes?

trevinhofmann (changes requested):
Thank you for the updates, @ohrytskov!

I have a few refactoring questions. Otherwise, this looks great to me in terms of functionality!

ohrytskov:
Thank you for the feedback! I understand the concern that using a store specifically for the editing timestamp can feel hacky. The reasoning behind this approach is that it simplifies tracking when the editing value is updated, without causing unnecessary dependency churn within the component. The store acts as a lightweight mechanism to avoid re-renders due to unrelated state updates, which would occur if we relied on Redux or selectors in this case.

After consideration, I've switched to using the ResizeObserver to directly observe size changes in sibling thought elements. This approach:

- Provides real-time updates, keeping the Divider's width in sync with its siblings.
- Eliminates the need for a store, simplifying state management.
- Optimizes performance by observing only relevant elements.

**Implementation**
The updated code uses ResizeObserver to attach listeners to sibling thoughts, recalculates width on size changes, and ensures proper cleanup.

ohrytskov:
@trevinhofmann, I've addressed the comments, and the PR is ready for the next review.

src/components/Divider.tsx
```diff
@@ -1,70 +1,182 @@
...
+          resizeObservers.current.push(resizeObserver)
+        }
       })
-      const maxWidth = Math.max(...widths)
-      dividerSetWidth.current.style.width = `${maxWidth > DIVIDER_MIN_WIDTH ? maxWidth : DIVIDER_MIN_WIDTH}px`
     }
-  }
 
-  useEffect(setStyle)
+    observeResize()
```
trevinhofmann, line 152: Is there a reason for creating an `observeResize` function and then immediately calling it only once? I think it would make sense to simplify this by removing the `observeResize` function declaration and call, and unindenting the contents of the function by one.

src/components/Divider.tsx
```diff
@@ -1,70 +1,182 @@
...
+      setDividerWidth(Math.round(Math.max(...widths, DIVIDER_MIN_WIDTH)))
+    }
+
+    updateDividerWidth() // Initial call to set the width
+
+    /** Attach resize observers to relevant thought elements. */
+    const observeResize = () => {
+      // Clear any existing observers
+      resizeObservers.current.forEach(observer => observer.disconnect())
+      resizeObservers.current = []
```
trevinhofmann, line 128: Do we need to disconnect each observer and empty the array here? This seems redundant since it's already handled in the returned cleanup function.

src/components/Divider.tsx
```diff
@@ -1,70 +1,182 @@
...
+}
 
 /** A custom horizontal rule. */
 const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) => {
-  const dividerSetWidth = React.createRef<HTMLInputElement>()
   const dispatch = useDispatch()
+  const dividerRef = useRef<HTMLDivElement>(null)
+  const [dividerWidth, setDividerWidth] = useState<number>(DIVIDER_MIN_WIDTH)
+  const { dividerId, isOnlyChild, isTableView, children, thoughtsAtSameDepth } = useDividerData(path)
+  const resizeObservers = useRef<ResizeObserver[]>([])
```
trevinhofmann, line 83: If we can remove the redundant cleanup at the start of `observeResize`, we should also be able to avoid this `useRef`. Instead, we can declare/instantiate a new `resizeObservers` each time the `useEffect` hook is called and disconnect/cleanup in the returned callback function.

src/redux-middleware/updateEditingValue.ts
```diff
@@ -13,7 +13,7 @@ const updateEditingValue: ThunkMiddleware<State> = ({ getState }) => {
     const thought = state.cursor ? getThoughtById(state, head(state.cursor)) : null
     const value = thought?.value ?? null
 
-    editingValueStore.update(value)
+    editingValueStore.update(value?.trim() ?? null)
```
trevinhofmann, line 16: Could you please give a quick explanation for this change? It seems like it might have been related to previous attempts to resolve the whitespace handling for divider widths, but I don't see how it affects the current solution.

trevinhofmann (changes requested):
Thank you for refactoring this, @ohrytskov! I much prefer the observers rather than using the timestamp-from-a-store approach. This seems much cleaner to me.

I just have a few more simplifications for refactoring the new observers and one question about the `updateEditingValue` middleware, then I believe this will be ready for my approval.

ohrytskov:
Thank you for the feedback, @trevinhofmann! All points have been addressed: the redundant cleanup and useRef for resizeObservers were removed, the observeResize function was simplified and inlined. The PR is now ready for the next review.

raineorshine:
Thanks for all your work! I'm just seeing these new updates.

I'm sorry to say that ResizeObservers are *extremely slow* 😔. We tried them a long time ago in **em** in the `useMultiline` hook when we needed updated thought heights to determine the multiline status. Setting a ResizeObserver on multiple thoughts caused a massive slowdown.

If you can prove (empirically) that ResizeObservers here are more performant than the other approach I would be happy to change my mind. However my base assumption based on past experience is that they are untenable.

ohrytskov:
Thank you for your feedback and for sharing your experience with ResizeObserver. I understand your concerns, especially given the past performance issues with the useMultiline hook. While I initially explored using ResizeObserver for real-time size detection, perhaps revisiting our previous approach based on the untrimmed value might be more suitable here.

I'm referring to this commit: [7df6c65](https://github.com/cybersemics/em/pull/2764/commits/7df6c656ba9f55a799677c2bad40dae17e77690b).

Do you think returning to that method could be a better solution in this context? I'm open to re-evaluating our options to ensure we find the best approach. Thank you again for highlighting this—I appreciate your insights and look forward to resolving this together.

Also, maybe this approach is a better alternative to the ResizeObserver solution:

```tsx
    const eventListeners: { element: HTMLElement; handler: () => void }[] = []

    const thoughtsToObserve =
      isTableView && isOnlyChild
        ? thoughtsAtSameDepth
        : children.filter(child => child.id !== dividerId && !isDivider(child.value))

    thoughtsToObserve.forEach(thought => {
      const editableElement = document.querySelector(
        `[aria-label="tree-node"][data-id="${thought.id}"] [data-editable="true"]`,
      ) as HTMLElement | null

      if (editableElement) {
        const handleInput = () => {
          updateDividerWidth()
        }
        editableElement.addEventListener('input', handleInput)
        eventListeners.push({ element: editableElement, handler: handleInput })
      }
    })

    return () => {
      eventListeners.forEach(({ element, handler }) => {
        element.removeEventListener('input', handler)
      })
    }
```

raineorshine:
I will let you and @trevinhofmann figure out the best non-ResizeObserver approach and I will review at the end. Unfortunately I'm not able to be super involved in the implementation details at this stage. 

Note that there are several past attempts at solutions from other developers that you can learn from as well. They are linked from
#543.

ohrytskov:
@raineorshine, can you please review [commit `2b3d434`](https://github.com/cybersemics/em/pull/2764/commits/2b3d434035b9700d208e45de59539d49e06cde99)? It contains the latest changes related to our discussion.

raineorshine:
I'm not able to review specific commits at this stage in the process. Trevin's role is to perform the initial review and work with you until you have a well-designed, working solution. I only chimed in early because I have past experience with ResizeObservers that I felt was important to share so that we stay on the right track. Please continue working with Trevin to evaluate the current approach and I will be happy to review and provide feedback at the end.

ohrytskov:
@trevinhofmann, could you take a look at the latest updates when you have a moment?

src/components/Divider.tsx
```diff
@@ -1,70 +1,185 @@
...
+            return
+          }
+          widths = getThoughtWidths(siblingThoughts, font, editingValueUntrimmed, editingThoughtId)
+        }
+
+        setDividerWidth(Math.round(Math.max(...widths, DIVIDER_MIN_WIDTH)))
+      }
+
+      updateDividerWidth() // Initial call to set the width
+    }, // eslint-disable-next-line react-hooks/exhaustive-deps
```
trevinhofmann, line 151: It would be helpful for future reference to include a comment here with [the explanation you provided previously](https://github.com/cybersemics/em/pull/2764#issuecomment-2594037898) for why this rule is ignored.

trevinhofmann (changes requested):
Hi, @ohrytskov! Thank you for making the latest updates. I apologize for leading you to believe that `ResizeObserver` was a viable solution; the code was certainly nice and readable.

Thank you for switching away from the `ResizeObserver`. At a high level, I believe the current design of recalculating divider widths when the dependencies change is acceptable. However, I am noticing some inconsistencies now with the functionality.

---

## 4. Divider widths are longer than thoughts

Although the divider widths do correctly respect whitespace, they are now longer than the widths of the widest sibling thought. They should be equal to the widest sibling thought.

https://github.com/user-attachments/assets/90f77e43-bf3a-4939-8132-ef4a1f46920a

---

## 5. Divider widths affected by selected thought

The divider widths (in tables) are now being affected by the currently active thought. The current selection should not affect the widths of any dividers.

https://github.com/user-attachments/assets/8aeab24c-6ec0-49d1-bbeb-c79669ad3f58

src/stores/editingValueUntrimmed.ts
```diff
@@ -0,0 +1,28 @@
...
+ * ```typescript
+ * const editingValueUntrimmed = editingValueUntrimmedStore.useSelector(state => state);
+ * ```
+ *
+ * This will provide your component with real-time access to the untrimmed value as the user types, allowing
+ * for immediate reactions to user input without waiting for the overall application state to update.
+ */
+const editingValueUntrimmedStore = reactMinistore<string | null>(null)
+
+export default editingValueUntrimmedStore
```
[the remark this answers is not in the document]
ohrytskov: I think it's a good idea to refactor editingValue into a single store that includes the untrimmed version. However, this should be done in a separate PR. It's a great idea overall!

ohrytskov:
@trevinhofmann, I've addressed the issues you mentioned:

- **Divider widths exceeding thought widths**: The widths now align with the widest sibling thought and no longer extend unnecessarily.
- **Divider widths affected by the selected thought**: Divider widths are now consistent and no longer change based on the selected thought.

The PR is ready for your next review.

src/components/Divider.tsx
ohrytskov: Good point! I've added a comment explaining why `eslint-disable-next-line react-hooks/exhaustive-deps` is used. 
It now states: 
_'eslint-disable-next-line react-hooks/exhaustive-deps` is used to exclude `editingThoughtId` because it is not essential for this effect's dependencies._

This should provide clarity for future reference.

src/stores/editingValueUntrimmed.ts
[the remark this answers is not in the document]
trevinhofmann: On one hand, I think it would be ideal to do that in this PR rather than set ourselves up for needing a separate refactoring PR.

On the other hand, doing that would add considerably to the diff in this PR, and it would be cleaner in this case to review one functionality PR and one refactoring PR.

I'm okay with having 2 PRs, if it makes sense to Raine.

src/components/Divider.tsx
ohrytskov: All other required dependencies have been included.

src/components/Divider.tsx
```diff
@@ -1,70 +1,172 @@
...
+}
+
+/**
+ * Calculates the width of a single text string, considering trailing spaces.
+ */
+const calculateTextWidth = (text: string, font: string): number => {
+  initializeCanvas()
+  if (!context) return 0
+  context.font = font
+  return Math.ceil(context.measureText(text).width)
```
trevinhofmann, line 75: I believe this is the cause of `6. Divider continues to grow after text wraps`. Here, we are looking at the width of the entire thought text without considering line wrapping. We would need to either account for text that wraps, or set a maximum width after calculating the full text width.

src/components/Divider.tsx
```diff
@@ -1,70 +1,172 @@
...
+
+/**
+ * Initializes the canvas and context for text width calculation.
+ */
+const initializeCanvas = () => {
+  if (!canvas) {
+    canvas = document.createElement('canvas')
+    context = canvas.getContext('2d')
+  }
+}
```
trevinhofmann, line 66: For readability, I would prefer to simply return a new context for each function call:

```suggestion
/**
 * Initializes the canvas and context for text width calculation.
 */
const getCanvasContext = () => {
  const canvas = document.createElement('canvas')
  return canvas.getContext('2d')
}
```

However, if we want to cache the canvas/context for performance reasons*, we can still simplify this by removing the `initializeContext` function altogether.

\* Running some really rough benchmarks on my machine, it seems to take about 0.01ms to create a canvas context and about 0.005ms to measure the text. This isn't _super_ concerning, but this function could be executed many times in parallel for large thoughtspaces with many dividers. For that reason, I agree with your choice to cache the canvas context (i.e. the second suggestion in this comment).

```suggestion
/** Initialize a canvas context for text width calculation */
let canvas = document.createElement('canvas')
let context = canvas.getContext('2d')
```

trevinhofmann (changes requested):
Thank you, @ohrytskov! The latest changes resolved the issues `4. Divider widths are longer than thoughts` and `5. Divider widths affected by selected thought`.

I am seeing one other issue now.

---

## 6. Divider continues to grow after text wraps

When a thought's text wraps to a new line, the divider continues to grow wider. The divider should reach a maximum width.

https://github.com/user-attachments/assets/eed9422d-b723-44a8-a042-d41c5f521ea6

ohrytskov:
@trevinhofmann, the PR is ready for the next review. Let me know if any further changes are needed.

ohrytskov:
The new solution now uses DOM elements to measure the maximum width of thoughts instead of relying on canvas text measurement. Since accounting for text wrapping requires accessing the actual element, we no longer need canvas for width calculations.This also simplifies the logic by returning to DOM-based measurements, making it more reliable for real-time UI updates.

trevinhofmann (approved):
Thank you, @ohrytskov!

The previous issues are all fixed, and the functionality on desktop and mobile appears to be working as expected.

src/stores/editingValueUntrimmed.ts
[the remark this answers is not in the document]
trevinhofmann: cc @raineorshine

src/components/Divider.tsx
```diff
@@ -1,70 +1,141 @@
...
+      } else {
+        const siblingThoughts = children.filter(child => !isDivider(child.value))
+        if (siblingThoughts.length === 0) {
+          setDividerWidth(DIVIDER_MIN_WIDTH)
+          return
+        }
+        widths = getThoughtWidths(siblingThoughts)
+      }
+
+      setDividerWidth(Math.round(Math.max(...widths, DIVIDER_MIN_WIDTH)))
```
raineorshine, line 114: Can we omit `Math.round`? I believe browsers handle subpixel rendering just fine.

src/components/Divider.tsx
```diff
@@ -1,70 +1,141 @@
...
-      const widths = children.map((child: HTMLElement) => {
-        if (child.querySelector('[data-divider=true]')) return DIVIDER_PLUS_PX
-        const subs = child.getElementsByClassName('subthought') as HTMLCollectionOf<HTMLElement>
-        return subs.length ? subs[0].offsetWidth + DIVIDER_PLUS_PX : DIVIDER_PLUS_PX
-      })
-      const maxWidth = Math.max(...widths)
-      dividerSetWidth.current.style.width = `${maxWidth > DIVIDER_MIN_WIDTH ? maxWidth : DIVIDER_MIN_WIDTH}px`
-    }
-  }
+  useEffect(() => {
```
raineorshine, line 88: This should probably be `useLayoutEffect` instead of `useEffect`. Otherwise it may paint the divider with the wrong width once before repainting with the correct width. Better to block rendering and paint it once at the correct width.

src/components/Divider.tsx
```diff
@@ -1,70 +1,141 @@
...
+    /** Calculates and updates the Divider's width based on sibling thought widths. */
+    const updateDividerWidth = () => {
```
raineorshine, line 90: The multiple return statements and calls to `setDivideWidth` make this function look more complicated than it actually is. Try refactoring it so that there is a single `setDividerWidth` at the end and no early returns. `Math.max(...widths, DIVIDER_MIN_WIDTH))` naturally returns `DIVIDER_MIN_WIDTH` when `widths = []`.

src/components/Divider.tsx
```diff
@@ -1,70 +1,141 @@
...
+      if (!dividerRef.current) return
```
raineorshine, line 91: Note that this initial short circuit to make sure the component is still mounted is fine, and does not need to be removed like the others.

src/components/Divider.tsx
```diff
@@ -1,70 +1,141 @@
...
+      prev.children.every((child, index) => child.value === next.children[index].value) &&
+      prev.thoughtsAtSameDepth.length === next.thoughtsAtSameDepth.length &&
+      prev.thoughtsAtSameDepth.every((thought, index) => thought.value === next.thoughtsAtSameDepth[index].value),
+  )
+}
+
+/**
+ * Calculates the width of multiple thoughts by measuring their rendered widths in the DOM.
+ */
+const getThoughtWidths = (thoughts: { id: ThoughtId }[]): number[] => {
```
raineorshine, line 64: Since this function only needs ids, it would be better to reduce the surface area of the parameters and just take `ids: ThoughtId[]`.

src/components/Divider.tsx
```diff
@@ -1,70 +1,141 @@
...
 
 /** A custom horizontal rule. */
 const Divider = ({ path, cssRaw }: { path: Path; cssRaw?: SystemStyleObject }) => {
-  const dividerSetWidth = React.createRef<HTMLInputElement>()
   const dispatch = useDispatch()
+  const dividerRef = useRef<HTMLDivElement>(null)
+  const [dividerWidth, setDividerWidth] = useState<number>(DIVIDER_MIN_WIDTH)
+  const { isOnlyChild, isTableView, children, thoughtsAtSameDepth } = useDividerData(path)
+  const editingThoughtId = useSelector((state: State) => state.cursor && head(state.cursor))
+  const editingValueUntrimmed = editingValueUntrimmedStore.useSelector(editingValue => editingValue)
```
raineorshine, line 80: Shorthand for selecting the whole state: `editingValueUntrimmedStore.useState()`

src/components/Divider.tsx
```diff
@@ -1,70 +1,141 @@
...
+  )
+}
+
+/**
+ * Calculates the width of multiple thoughts by measuring their rendered widths in the DOM.
+ */
+const getThoughtWidths = (thoughts: { id: ThoughtId }[]): number[] => {
+  return thoughts.map(thought => {
+    const innerThoughtElement = document.querySelector(
+      `[aria-label="tree-node"][data-id="${thought.id}"] [aria-label="thought"]`,
```
raineorshine, line 67: Would it work to use the existing editable element for measurement?

```ts
document.querySelector(`[aria-label="editable-${id}"]`)
```

src/components/LayoutTree.tsx
```diff
@@ -476,9 +476,14 @@ const TreeNode = ({
...
     : `left {durations.layoutNodeAnimation} ease-out,top {durations.layoutNodeAnimation} ease-out`
 
+  const parentId = path.length > 1 ? path[path.length - 2] : null
+
   return (
     <div
       aria-label='tree-node'
+      data-id={thoughtId}
+      data-parent-id={parentId}
+      data-depth={depth}
```
raineorshine, line 486: It looks like `data-depth` and `data-parent-id` are not used. 

`data-id` may also be able to be removed if the sibling elements are selected by editable `aria-label` as mentioned above.

raineorshine (changes requested):
Thank! I really like the new approach. It is lightweight since it avoids ResizeObserver and Canvas, and almost all of the logic is contained within the Divider component. Nice work.

There are a couple more dependencies that need to trigger `updateDividerWidth` for it to be updated when the font size changes and the window is resized:

```ts

  viewportStore.useSelectorEffect(updateDividerWidth, state => state.innerWidth)
```

---

Overall I like the separation between `useDividerData`, `getThoughtWidths`, and `updateDividerWidth`. `useDividerData` selects from the Redux state, `getThoughtWidths` interacts with the DOM, and `updateDividerWidth` calculates the divider width.

However, notice that `updateDividerWidth` is really just an extension of `useDividerData`, determining which thoughts' widths are relevant based on the Redux state. By splitting up `useDividerData` and `updateDividerWidth` into separate hooks, it exposes more variables than necessary to the component. 

Combining `useDividerData` and `updateDividerWidth` into a single custom hook that returns `ThoughtId[]` would reduce the `{ isOnlyChild, isTableView, children, thoughtsAtSameDepth }` return object down to a single `ThoughtId[]`, eliminate the custom equality function, and reducer the number of re-renders.

Let me know if you see any issues with this suggested change!

ohrytskov:
Thank you for your feedback, @raineorshine! I’ve carefully reviewed your suggestions and have made all the necessary changes accordingly. The PR has been updated with the requested modifications, and it is now ready for your next review. Please let me know if there are any further adjustments needed.

src/components/Divider.tsx
```diff
@@ -1,70 +1,113 @@
...
+import { getAllChildrenAsThoughts } from '../selectors/getChildren'
 import rootedParentOf from '../selectors/rootedParentOf'
+import editingValueUntrimmedStore from '../stores/editingValueUntrimmed'
+import viewportStore from '../stores/viewport'
 import fastClick from '../util/fastClick'
 import head from '../util/head'
+import isDivider from '../util/isDivider'
+
+/** Custom hook to fetch relevant thought IDs for calculating divider width. */
+const useRelevantThoughtIds = (path: Path): ThoughtId[] => {
```
raineorshine, line 20: Let's call this `useWidthDependentThoughtIds`. I try to avoid words like "relevant" or "data" in variable names since they don't explain anything.

src/components/Divider.tsx
```diff
@@ -1,70 +1,113 @@
...
+                (parent.id ? getAllChildrenAsThoughts(state, parent.id) : [])
+                  .filter(child => !isDivider(child.value))
+                  .map(child => child.id),
+              )
+          : []
+        : childrenWithoutDividers.map(child => child.id)
+
+      return relevantThoughtIds
+    },
+    (prev, next) => prev.length === next.length && prev.every((id, index) => id === next[index]),
```
raineorshine, line 48: You can use drop in `shallowEqual` from react-redux or `isEqual` from lodash rather than providing a custom equality function.

src/components/Divider.tsx
```diff
@@ -1,70 +1,113 @@
...
+      const parentPath = rootedParentOf(state, path)
+      const parentId = head(parentPath)
+      const grandParentPath = parentId ? rootedParentOf(state, parentPath) : []
+      const grandParentId = head(grandParentPath)
+      const children = parentId ? getAllChildrenAsThoughts(state, parentId) : []
+      const childrenWithoutDividers = children.filter(child => !isDivider(child.value))
+      const isOnlyChild = childrenWithoutDividers.length === 0
+      const isTableView =
+        (parentId && attributeEquals(state, parentId, '=view', 'Table')) ||
+        (grandParentId && attributeEquals(state, grandParentId, '=view', 'Table'))
```
raineorshine, line 32: You can make this a little more terse by omitting `parentId &&` since `attributeEquals` handles null.

src/components/Divider.tsx
```diff
@@ -1,70 +1,113 @@
...
 import head from '../util/head'
+import isDivider from '../util/isDivider'
+
+/** Custom hook to fetch relevant thought IDs for calculating divider width. */
+const useRelevantThoughtIds = (path: Path): ThoughtId[] => {
+  return useSelector(
+    (state: State) => {
+      const parentPath = rootedParentOf(state, path)
+      const parentId = head(parentPath)
+      const grandParentPath = parentId ? rootedParentOf(state, parentPath) : []
```
raineorshine, line 25: It won't affect the behavior here, but `[]` is not a valid `Path`. It should probably be set to `null` if there is no `parentId`.

src/components/Divider.tsx
```diff
@@ -1,70 +1,113 @@
...
+      const grandParentId = head(grandParentPath)
+      const children = parentId ? getAllChildrenAsThoughts(state, parentId) : []
+      const childrenWithoutDividers = children.filter(child => !isDivider(child.value))
+      const isOnlyChild = childrenWithoutDividers.length === 0
+      const isTableView =
+        (parentId && attributeEquals(state, parentId, '=view', 'Table')) ||
+        (grandParentId && attributeEquals(state, grandParentId, '=view', 'Table'))
+
+      const relevantThoughtIds = isOnlyChild
```
raineorshine, line 34: Thanks, this is much better.

Let's call this `dependentThoughtIds`. 

Could you also add a couple inline comments that explain the two branches?

raineorshine (changes requested):
Looks great! Just a few touchups, then we should be good to merge.

ohrytskov:
This appears to be ready for the next review—awaiting your feedback, @raineorshine.

raineorshine (approved):
Thank you!

Issue #1628, opened 2022-07-19
[iOS] Prevent autoscroll to focused editable


raineorshine:
## Steps to Reproduce

```
- a
  - b
  - c
  - d
  - e
  - f
  - g
  - h
```

1. Set cursor to `null` and scroll so that all thoughts are within the viewport.
2. Tap on `h` and enter edit mode

## Current Behavior

The window scrolls down so that the focused thought is in the middle of the screen. The top thoughts are now obscured.

https://user-images.githubusercontent.com/750276/212205968-d020bc5d-fe3d-49a8-89e7-a64773f10df0.MOV

## Expected Behavior

Autoscroll should scroll the minimum amount necessary to bring the cursor thought into view. 

## To Do

- [x] ~~`el.focus({ preventScroll })`~~
	- Needs to work on tap, without programmatic focus
- [x] ~~`onTouchEnd`~~
	- e.preventDefault does stop anything from happening, but any attempt to focus causes the scroll.
- [x] ~~body-scroll-lock~~
- [x] ~~[once scroll scrollTo](https://stackoverflow.com/a/39584526/480608)~~
	- scroll position is maintained, but screen jumps
- [x] ~~[focus focusIn](https://stackoverflow.com/questions/6740253/disable-scrolling-when-changing-focus-form-elements-ipad-web-app#comment61708382_31472878)~~
- [x] ~~setting contenteditable after touch~~
- [x] ~~override scrollIntoView~~
- [x] ~~position: fixed~~
- [x] ~~[transform](https://stackoverflow.com/questions/47800185/jquery-mobile-how-to-stop-auto-scroll-to-focused-input-in-mobile-web-browser/47805387#47805387)~~
- [x] ~~[add readonly ontouchstart, and remove ontouchend](https://stackoverflow.com/questions/6740253/disable-scrolling-when-changing-focus-form-elements-ipad-web-app/7209710#7209710)~~
- [x] ~~[keyframes](https://stackoverflow.com/questions/6740253/disable-scrolling-when-changing-focus-form-elements-ipad-web-app/60712537#60712537)~~
- [x] ~~[opacity: 0](https://stackoverflow.com/questions/37835638/focus-on-input-inside-modal-causes-ios-to-scroll-down/74905873#74905873) [gist](https://gist.github.com/kiding/72721a0553fa93198ae2bb6eefaa3299)~~
- [x] ~~overflow: hidden~~
[...]

PR #2765, opened 2025-01-10
Fix missing frame when animating thought up/down

zukilover:
This PR fixes the missing frame issue at #2498 , which occurs when the editable `preventAutoscroll` class is applied, causing the thought to render with zero opacity:
https://github.com/cybersemics/em/blob/13f92afd31c80983a291561e5ab6d6261296aae3/src/components/Editable.tsx#L595-L597

The solution ensures that the `preventAutoscroll` class is omitted during the `y` transition of the tree node. To achieve this, I refactored the `y` animation into `TreeNodeAnimation.tsx` and introduced a new `isAnimating` flag. Since this flag needs to be passed down through multiple layers (5 levels from `LayoutTree` to `Editable`), I implemented a React context to manage the state more effectively.


zukilover:
@raineorshine, could you please re-run the failed test? It seems like a pipeline issue to me.

raineorshine:
Thanks!

Can you confirm that you tested `preventAutoscroll` after this change and it still successfully prevents autoscroll in mobile Safari? I want to make sure we don't introduce any regressions. 

We should also do a performance benchmark before/after since this potentially introduces a lot of state changes during a critical rendering phase.

zukilover:
I tested preventAutoscroll by simulating an up/down movement animation and then focusing on one of the subthoughts. You can view the screen recording here:

https://github.com/user-attachments/assets/99f3c38e-98d4-4ae2-9076-8e1293655ae5

I've confirmed that the flag resets once the transition is complete. You can check the relevant code here:

https://github.com/cybersemics/em/blob/54ff6d745aadc331aab46e443d1f22b288def43e/src/components/TreeNodeAnimation.tsx#L33-L38

Additionally, I’ve optimized the code by using `useRef` instead of `useState` to prevent unnecessary state updates:

https://github.com/cybersemics/em/blob/54ff6d745aadc331aab46e443d1f22b288def43e/src/components/TreeNodeAnimation.tsx#L24

Previously, using `useState` not only caused unnecessary state changes, but also led to the entire thought rendering with opacity: 0 (triggered by preventAutoScroll due to a re-render).

raineorshine:
That sounds great. I'll let @trevinhofmann review and test first, but looking forward to trying it out!

src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/font-size-18-default-superscript-1.png
trevinhofmann, line 1: Can you please investigate why these snapshots were affected? The contents of the last thought were shifted slightly up in each of them.

src/components/LayoutTree.tsx
trevinhofmann, line 1: Note for @raineorshine - The diff for this file is **much** easier to review after hiding whitespace changes.

trevinhofmann (changes requested):
Thank you, @zukilover! This appears to resolve the animation for thoughts moved down:

https://github.com/user-attachments/assets/ccc91dcd-4e14-4fac-96ec-0bfaa10172fc

I've also tested on mobile (iOS Safari), and the animations work there as well.

I just have one question regarding the snapshots, otherwise this looks good to me.

src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/font-size-18-default-superscript-1.png
raineorshine: Yes, these look significant to me.

src/components/LayoutTree.tsx
raineorshine: I have not used this feature nearly enough!!! Thank you

src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/font-size-18-default-superscript-1.png
zukilover: I'm still trying to understand why there are changes in the snapshots. I’ve done manual testing and compared the main branch with this PR, but I’m not seeing any differences in the screenshots.

MERIDIQ_before_after_50_50

raineorshine: @trevinhofmann Could you test this and see if the layout shift is happening on your machine? I'd like you and @zukilover to try to get to the bottom of this. Thanks!

zukilover: I still haven’t figured out the cause, but I took an extra step to double-check if it works in the real app. I simulated a `paste` and then manually pressed the up arrow. The result was consistent: the last thought didn’t shift up.

https://github.com/user-attachments/assets/8fa81b24-bba1-4ece-ba5c-6ab8d28c1d91


thought


What’s strange is that the test suites pass for font sizes 13 and 22, but not for 18.

trevinhofmann: Testing manually at font size 18 in the real app, there is no shift in the positioning:

comparison

---

Running the puppeteer tests locally, I get the same diff output as the CI which shows the positioning shift:

color-theme-superscript-on-light-theme-1-diff

font-size-18-default-superscript-1-diff

multiline-font-size-13-1-diff

trevinhofmann: This doesn't reduce the surface area by much, but the initial commit 6363d14e92875e1ba41e9dfe00fe088152b8a053 has the snapshot failures. The second commit, 69df6850de2ffd69585717a27f8cc5b94c8c9e18, is not the cause of the issue.

zukilover: The isAnimating flag is the core change in this PR and is indeed what triggered the failed tests.

I will try to simulate real DOM interactions in the tests (replacing the current paste method) to see if this fixes the issue.

trevinhofmann: Sounds good. Thanks, @zukilover! Looking forward to seeing what you learn from that.

raineorshine:
Thank you both for your work on this tricky issue.

I was finally able to reproduce the layout shift on `main` outside of the puppeteer tests. See a detailed explanation in #2787. While the instigating cause is not 100% clear, my best guess is that the changes in this PR change the animation timing enough to trigger the rendering discrepancy. It is likely that the [paste](https://github.com/cybersemics/em/blob/875843163108c256489a938a421c75de54216597/src/e2e/puppeteer/helpers/paste.ts) helper injects the thoughts in a way that creates a different render behavior in the tests than when manually testing. 

Given this, I will tentatively say that the layout issue is not a regression caused by this PR and can be safely ignored. Needless to say, this does not evoke confidence in the puppeteer tests. But I think that is out of scope for this issue, and can be addressed in the new issue I created.

---

Now, considering the current approach. I do have concerns with the **complexity** and potential **performance impact** of this change. It involves `requestAnimationFrame`, adding/removing event handlers on every thought, and `cloneElement`. There is a lot going on here to fix a missing frame. 

All of this seems to center around the `preventAutoscroll` behavior, yet I don't see a lot of research into the history and purpose of the original functionality. The new solution is clearly thought out and well tested. Nevertheless, I think it would be worth some due diligence before introducing what I would consider to be a fairly heavy, high risk change.

Question: Is the `preventAutoscroll` style still needed?

Note that there are two different autoscroll scenarios to be aware of. I regret that they use the same name, as this ambiguity only confuses the matter:

- The `preventAutoscroll` editable recipe variant (i.e. `animation: 0.01s 1 preventAutoscroll`) was originally added in 24afba87b5fad2f2629363861c9c26edecfee606 in December 2020. Based on this [StackOverflow answer](https://stackoverflow.com/a/60712537/480608). While it does not link to an issue, the commit messages points out that it was specifically added to prevent autoscroll during the tutorial: "Tutorial: Prevent autoscroll to editable after new thought." The tutorial covers the top part of the screen and the user can interact with the editor below, so the purpose of this is to avoid scrolling the tutorial out of view when the user follows the instructions to create a new thought.
- The `preventAutoscroll` utility was originally added in 95bc8aaa859a3a32a2d74a63ae271a13af641b88 in 2023 to fix #1628. This was a more challenging problem to solve and involved a JS solution, explained in this [StackOverflow answer](https://stackoverflow.com/a/76731361/480608). Presumably this is not related to *this issue* (#2498), but I thought it was worth distinguishing in order to avoid confusion. (Though it does raise the question... If the `animation` trick works in the tutorial, why doesn't it work here?)

Next steps:

- [ ] Confirm that the `preventAutoscroll` *style* is still needed to prevent autoscroll when creating a new thought on mobile in the Tutorial.
- [ ] If it is still needed, do some performance testing to evaluate the impact of this change. e.g. Compare the expansion of a large number of thoughts on `main` vs in this PR with 4x CPU throttling.)

zukilover:
@raineorshine, here's my conclusion after comparing the behavior of the main branch with this PR:

I tested it on iOS Safari, and it’s clear that if the purpose of adding the hack-ish `preventAutoscroll` keyframe is to prevent auto-scrolling when focusing on the editable, it doesn’t achieve that effectively. As shown in the videos below, there isn’t much difference between the main branch and this PR. Specifically, when creating a thought, the editable is focused, but parts of the tutorial still go out of view.

The other notable difference is that, in the main branch, adding a subthought on mobile causes the virtual keyboard to cover the editable. However, in this PR, I can focus on the editable without it being obscured by the keyboard.

Main branch:

https://github.com/user-attachments/assets/8cd96576-efb1-4cb8-9c04-9b45c0fadc1c


This PR:

https://github.com/user-attachments/assets/7b484e3c-5002-45a5-ae94-010d2d60907b

I’m wondering if we should move forward with this PR by removing the style entirely?

raineorshine:
Thanks for testing it out. The videos look close to me, though it's a bit hard to compare frame by frame. It even appears that the preventAutoscroll utility (2) may be overriding the preventAutoscroll style (1). Could you try disabling the utility and seeing if the tutorial still autoscrolls? I just want to establish firmly that the style does not have any effect and is no longer needed. @trevinhofmann can also validate. 

Thanks so much!

trevinhofmann:
I did some quick testing on iOS Safari to compare `main`, this PR, and disabling each of the `preventAutoScroll` functionalities (it is indeed unfortunate that we have two with the same name).

<details>
<summary>main</summary>

https://github.com/user-attachments/assets/282ab074-726c-4144-8966-d430207df7ea
</details>

<details>
<summary>this pull request</summary>

https://github.com/user-attachments/assets/b1485055-d5d2-47a5-89a6-bbda66adf425
</details>

<details>
<summary>this pull request, with preventAutoScroll style disabled</summary>

https://github.com/user-attachments/assets/f66ef9fb-f316-4d8d-ac97-2db1347e2f3a
</details>

<details>
<summary>this pull request, with preventAutoScroll utility disabled</summary>

https://github.com/user-attachments/assets/e8693d34-26bd-46df-bd4c-7150909e59d3
</details>

It seems that disabling the `preventAutoScroll` utility causes it to scroll most of the tutorial out of view.

raineorshine:
Thanks @trevinhofmann! I think we're good to remove the `preventAutoscroll` style then and call this solved.

While existing code is usually there for a reason, this is a good reminder that it doesn't hurt to test those base assumptions. It's a rare exception, nevertheless, and I'm glad we were cautious.

Issue #2167, opened 2024-07-15
Error importing empty parent in root


raineorshine:
## Steps to Reproduce

```
- 
  - a
``` 

## Current Behavior

Error importing.

## Expected Behavior

Should import as an empty thought with child `a`.

PR #2766, opened 2025-01-10
fix error importing empty parent in root

RED-ROSE515:
This PR resolves the issue : Error importing empty parent in root #2167

Cause of the issue : There is a problem in handling the empty parents in import process. When we have an empty parent node, the context-to-path conversion is failing because it's treating the empty string as an invalid value.

Solution : I've modified the `importFiles.ts` file to handle empty parent nodes correctly. So, when I detect an empty parent node, I use the current path instead of trying to convert the empty context to a path.

Thanks.


raineorshine:
Thanks for your submission!

If the cause of the issue is related to `contextToPath`, I'm curious why you added a special case to `importFiles` rather than fixing the root cause.

PR #2771, opened 2025-01-14
LayoutTree: no hideCaret animation for 1st thought

### LINKED ISSUE #2708: Caret hidden on New Thought in root
I found a false positive for #4. 

## Steps to Reproduce

Create a new thought. 

## Current Behavior

The caret is hidden for 750ms even though there is no indentation transition.  

## Expected Behavior

The caret should not be hidden.

ethan-james:
Fixes #2708 

Any `hideCaret` animation value will trigger the animation when it is first applied. If there's only one thought, use `'none'` as the animation value, which will not play an animation. There will not be any indentation changes while only one thought exists, so there will be no need to hide the caret.

### DIFF
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -937,7 +937,12 @@ const LayoutTree = () => {
   return (
     <div
       // the hideCaret animation must run every time the indent changes on iOS Safari, which necessitates replacing the animation with an identical substitute with a different name
-      className={cx(css({ marginTop: '0.501em' }), hideCaret({ animation: getHideCaretAnimationName(indentDepth) }))}
+      className={cx(
+        css({ marginTop: '0.501em' }),
+        hideCaret({
+          animation: treeThoughts.length > 1 ? getHideCaretAnimationName(indentDepth) : 'none',
+        }),
+      )}
       style={{
         // add a full viewport height's space above to ensure that there is room to scroll by the same amount as spaceAbove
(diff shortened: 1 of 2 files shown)


trevinhofmann:
@ethan-james or @raineorshine -- Can you please clarify how to reproduce the original bug described in #2708? When testing `main` on Desktop (Chrome) and iOS (Safari), I am not seeing the carrot be hidden for 750ms when creating a new thought (or subthought).

raineorshine:
When creating a new though in an empty thoughtspace on mobile Safari, the caret is initially hidden.

I can test this one, no problem. I can't think of what the difference might be!

https://github.com/user-attachments/assets/9a7d1ac9-6223-4e96-986b-c58d95c10aac

raineorshine (changes requested):
Thanks! I confirm it works for the first thought.

It looks like there is a regression where the caret is now hidden when creating a sibling thought or subthought. I confirmed it does not occur in `main`. Could you like into this?

## Steps to Reproduce

1. Create a new thought `a`.
2. Close the keyboard.
3. Activate New Thought or New Subthought.

## Current Behavior

Caret is hidden.

## Expected Behavior

Caret should not be hidden. Caret should only be hidden when animating indent.

ethan-james:
I think I came up with a better CSS-based solution that doesn't rely on changing the animation name when new thoughts are added.

raineorshine (approved):
Interesting. Seems to work!

Thanks for the explanatory comment.

Issue #2753, opened 2025-01-04
selection.set does not always open keyboard


raineorshine:
The goal of this issue is to:

- Determine what is wrong with `selection.set` and why `focus()` is needed (if possible). 
- Determine if it is safe to move the `focus()` call into `selection.set`. 
- Remove `focus()` from `useEditMode` to restore the invariant that all selection behavior occurs via the `selection` module.

> I'm not quite sure why this is necessary, but it was having the following problem on iOS Safari:
> 
> 
> 
> 1. Tap a thought - cursor moves, keyboard stays closed
> 
> 2. Tap it again - keyboard opens
> 
> 3. Close keyboard
> 
> 4. Tap it again - keyboard does not open

_Originally posted by @ethan-james in https://github.com/cybersemics/em/pull/2752#discussion_r1902247165_

Related: #2770 #2772

raineorshine:
As pointed out in #2772, `selection.isText()` returns false when the `focusNode` is set to the editable div. This stems from an annoying native browser behavior where the same selection at the end of an input may be represented in two different ways:

- `focusNode` is a TEXT_NODE with `focusOffset` equal to the number of characters in the text.
- `focusNode` is an ELEMENT_NODE with `focusOffset equal to 1 (the number of nodes in the parent)

This seems to be throwing off https://github.com/cybersemics/em/blob/650c3d2e38e109ec4e6b3f0c9a2b4e73ece43789/src/device/asyncFocus.ts.

raineorshine:
Possibly part of the solution:

```diff
diff --git a/src/device/asyncFocus.ts b/src/device/asyncFocus.ts
index 77873951a5..c49e1acbba 100644
--- a/src/device/asyncFocus.ts
+++ b/src/device/asyncFocus.ts
@@ -32,8 +32,8 @@ export const AsyncFocus: () => () => void = () => {
 
   document.body.prepend(hiddenInput)
   return () => {
-    // do not set the selection if it is already on a text node
-    if (!selection.isText()) {
+    // do not set the selection if it is already on a contenteditable
+    if (!selection.isContentEditable()) {
       hiddenInput.disabled = false
       hiddenInput.focus()
       // the hidden input should not be a valid focus target unless this function was invoked
diff --git a/src/device/selection.ts b/src/device/selection.ts
index af42d834b6..047afe1beb 100644
--- a/src/device/selection.ts
+++ b/src/device/selection.ts
[...]

PR #2772, opened 2025-01-15
cursorCleared: remove asyncFocus

### LINKED ISSUE #2705: clearThought does not work after cursorForward
## Steps to Reproduce

```
- a
  - b
```

1. Set the cursor on `a`.
2. Set the caret on `a` so the keyboard opens.
3. Swipe ← to activate cursorForward.
4. Swipe →←→ to activate clearThought.

## Current Behavior

Nothing happens.

## Expected Behavior

clearThought should work as normal.

ethan-james:
Fixes #2705 

Calling `asyncFocus()` after `clearThought` triggers the editable's `onBlur` which dispatches `cursorCleared({ value: false })`, which undoes the whole `clearThought` operation.

Screenshot 2025-01-15 at 10 23 33

Without `asyncFocus()`, `clearThought` can proceed as intended.

Screenshot 2025-01-15 at 10 24 26

`asyncFocus()` was triggered on any `clearThought`, and the reason why it's failing in this specific case is because `selection.isText()` return `true` on a basic `clearThought`, and `false` after `cursorForward`. The reason for this is because the native selection sets `focusNode` to the text node within the editable, while `cursorForward` is setting `focusNode` to the editable `div`.

### DIFF
--- a/src/components/Editable/useEditMode.ts
+++ b/src/components/Editable/useEditMode.ts
@@ -51,6 +51,7 @@ const useEditMode = ({
           selection.clear()
         } else {
           selection.set(contentRef.current, { offset: editingCursorOffset || 0 })
+          if (isTouch && isSafari()) contentRef.current?.focus()
         }
       }
 
--- a/src/actions/cursorCleared.ts
+++ b/src/actions/cursorCleared.ts
@@ -1,6 +1,5 @@
 import State from '../@types/State'
 import Thunk from '../@types/Thunk'
-import asyncFocus from '../device/asyncFocus'
 
 /**
  * Sets state.cursorCleared which controls a special state in which the cursor is rendered as an empty string. In this state the thought can be deleted or edited, but if the user navigates away the thought is restored to its previous value.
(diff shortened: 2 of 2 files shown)


src/components/Editable/useEditMode.ts
```diff
@@ -51,6 +51,7 @@ const useEditMode = ({
           selection.clear()
         } else {
           selection.set(contentRef.current, { offset: editingCursorOffset || 0 })
+          if (isTouch && isSafari()) contentRef.current?.focus()
```
ethan-james, line 54: I guess this is the same fix from #2752 that is already captured in #2753.

In this case, the steps to reproduce that bad focus behavior are:

1. `clearThought` gesture on the active cursor
2. Close the keyboard
3. `clearThought` gesture again
4. Keyboard doesn't open

raineorshine (approved):
Thanks so much for the clear explanation. It looks like I have been a bit too liberal with `asyncFocus`! It works great when there is no focus, but causes problems if the editable already has the focus, and it definitely doesn't work well with `clearThought`.

ethan-james:
Right on, I tested it when there is no focus and it seemed OK. I know `asyncFocus` is in there to solve a problem, but I couldn't find any case that would break without it here.

I'd still like to figure out why the `selection.isText()` thing is breaking, but like I say, I didn't see any example where leaving `asyncFocus` in would improve things.

raineorshine:
Yep. It's possible we might be able to implement a more general fix, but I've opened #2753 for that.

Issue #2396, opened 2024-09-14
LayoutTree: Factor out DropCliff component


raineorshine:
This code should be factored out into a new component called `DropCliff`:

https://github.com/cybersemics/em/blob/54f5a65c70defb4d5700411e25202232bacfe106/src/components/LayoutTree.tsx#L712-L751

PR #2785, opened 2025-01-23
DropCliff component

snqb:
closes https://github.com/cybersemics/em/issues/2396

### DIFF
--- a/src/components/DropCliff.tsx
+++ b/src/components/DropCliff.tsx
@@ -1,0 +1,62 @@
+import { css } from '../../styled-system/css'
+import Path from '../@types/Path'
+import { isTouch } from '../browser'
+import { HOME_PATH } from '../constants'
+import head from '../util/head'
+import unroot from '../util/unroot'
+import DropEnd from './DropEnd'
+
+/** Renders a drop target for dropping thoughts after a cliff. */
+const DropCliff = ({
+  cliff,
+  depth,
+  path,
+  isTableCol2,
+  prevWidth = 0,
+  isLastVisible,
+}: {
+  cliff: number
+  depth: number
+  path: Path
+  isTableCol2: boolean
+  prevWidth?: number
+  isLastVisible?: boolean
+}) => {
+  if (cliff < 0) {
+    return Array(-cliff)
+      .fill(0)
+      .map((_, i) => {
+        const pathEnd = -(cliff + i) < path.length ? (path.slice(0, cliff + i) as Path) : HOME_PATH
+        const cliffDepth = unroot(pathEnd).length
+        const depthDiff = cliffDepth - depth
+        const dropEndMarginLeft = isTableCol2 && depthDiff < 0 ? prevWidth : 0
+        return (
+          <div
+            key={'DropEnd-' + head(pathEnd)}
+            className={css({
+              position: 'relative',
+              top: '-0.2em',
+              transition: `left {durations.fast} ease-out`,
+              zIndex: 'subthoughtsDropEnd',
+            })}
+            style={{
+              left: `calc(${cliffDepth - depth}em - ${dropEndMarginLeft}px + ${isTouch ? -1 : 1}px)`,
+            }}
+          >
+            <DropEnd
+              depth={pathEnd.length}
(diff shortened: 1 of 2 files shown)


src/components/LayoutTree.tsx
```diff
@@ -540,50 +539,18 @@ const TreeNode = ({
...
-                  // last={!nextChildId}
-                />
-              </div>
-            )
-          })}
+        autofocusDepth - depth < 2 && (
+          <DropCliff
+            cliff={cliff}
+            depth={depth}
+            path={path}
```
snqb, line 549: these feel ok drill down, and none of them can't actually be extracted out of the loop

src/components/LayoutTree.tsx
```diff
@@ -540,50 +539,18 @@ const TreeNode = ({
...
+            isTableCol2={isTableCol2}
+            prevWidth={treeThoughtsPositioned[index - 1]?.width}
```
snqb, line 551: but these 2 maybe suck

I also initially did a 
`marginLeft={isTableCol2 ? treeThou...` and applied it inside the `DropCliff` with a condition

but it kinda felt like more vague, though half as verbose

trevinhofmann: We could pass a raw CSS object to the `DropCliff` which has the `left` margin calculated, but I believe it makes more sense to let the `DropCliff` calculate the styling on its own -- especially because the component is only used in one place and doesn't need to let the parent component determine this.

I agree that it doesn't feel ideal to pass values such as `isTableCol2` and `prevWidth`, but it is the right choose in this case in my opinion.

src/components/LayoutTree.tsx
trevinhofmann: Similar to [the other thread](https://github.com/cybersemics/em/pull/2785#discussion_r1927396150), I would _generally_ like to avoid prop drilling but consider it appropriate in this particular case 👍

trevinhofmann (approved):
Thank you for the refactor, @snqb! I haven't found any regressions when testing on desktop and mobile, and I believe the prop drilling is acceptable and reasonably considered.

src/components/LayoutTree.tsx
```diff
@@ -540,50 +539,19 @@ const TreeNode = ({
...
-        autofocusDepth - depth < 2 &&
-        Array(-cliff)
-          .fill(0)
-          .map((x, i) => {
-            const pathEnd = -(cliff + i) < path.length ? (path.slice(0, cliff + i) as Path) : HOME_PATH
-            const cliffDepth = unroot(pathEnd).length
-
-            // After table col2, shift the DropEnd left by the width of col1.
-            // This correctly positions the drop target for dropping after the table view.
-            // Otherwise it would be too far to the right.
```
raineorshine, line 557: I think this comment got dropped. Could we keep it in?

src/components/DropCliff.tsx
```diff
@@ -0,0 +1,62 @@
...
+  isLastVisible,
+}: {
+  cliff: number
+  depth: number
+  path: Path
+  isTableCol2: boolean
+  prevWidth?: number
+  isLastVisible?: boolean
+}) => {
+  if (cliff < 0) {
```
raineorshine, line 25: Minor point, but I tend to prefer the negation for short circuiting statements. It results in a smaller indent level and fewer curly braces to deal with:

```ts
if (cliff >= 0) return null

return ...
```

src/components/DropCliff.tsx
```diff
@@ -0,0 +1,62 @@
+import { css } from '../../styled-system/css'
+import Path from '../@types/Path'
+import { isTouch } from '../browser'
+import { HOME_PATH } from '../constants'
+import head from '../util/head'
+import unroot from '../util/unroot'
+import DropEnd from './DropEnd'
+
+/** Renders a drop target for dropping thoughts after a cliff. */
```
raineorshine, line 9: This component actually renders multiple drop targets: one for every level of the cliff.

raineorshine (changes requested):

raineorshine (approved):
Thank you!

Issue #2157, opened 2024-07-12
importHtml: WorkFlowy import with notes


raineorshine:
Restore functionality covered by this skipped test: 

https://github.com/cybersemics/em/blob/f68689f6471af07ba87db97cdd9b68bf42cdc910/src/util/__tests__/importHtml.ts#L257-L289

raineorshine:
It looks like all HTML importing is broken. Further investigation is needed before WorkFlowy HTML can be troubleshooted.

PR #2788, opened 2025-01-27
initial workflowy update

RED-ROSE515:
This PR resolves the issue importHtml: WorkFlowy import with notes #2157.
This issue occurs because currenlty strip the pasted html.
Current strip function removes all elements from pasted html from WorkFlowy.

So, in this PR, I have modified the strip function and updated the htmlToJson function to effectlively converts html to Json.

### DIFF
--- a/src/components/Editable/useOnPaste.ts
+++ b/src/components/Editable/useOnPaste.ts
@@ -1,3 +1,4 @@
+import { parse } from 'himalaya'
 import { escape as escapeHtml } from 'html-escaper'
 import React, { useCallback } from 'react'
 import { useDispatch } from 'react-redux'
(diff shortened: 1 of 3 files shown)


raineorshine:
Thank you for the draft PR!

I just wanted to check in and ask, Do you think that `strip` is the right place for this logic? The purpose of strip is to remove formatting, html, etc. It seems odd to make that function responsible for doing complex parsing or restructuring.

Maybe this requires a bit of refactoring to make sure that the logic is split appropriately, ensuring that each function has a limited scope and purposeful name.

RED-ROSE515:
Thanks for your feedback.
I got your point.
I will integrate new function to handle new logic.

raineorshine:
It appears that all HTML importing is broken, so we'll have to get to the bottom of that before addressing WorkFlowy HTML. 

Regardless, we don't want to write a new HTML parsing algorithm or create a separate algorithm to maintain just for WorkFlowy.

Thank you for your time.

Issue #2774, opened 2025-01-15
GestureDiagram: Custom Help diagram


raineorshine:
Create a custom Help diagram with a nicer looking question mark.

Note: This *only* affects the appearance of the GestureDiagram; the actual gesture behavior is already (implicitly) supported by the MultiGesture component which parses curved gestures in the nearest cardinal directions.

## Steps to Reproduce

Swipe →↓←↓ and hold to select the Help command from the Command Palette.

## Current Behavior

The default cardinal diagram is rendered:

<img width="398" alt="Screenshot 2025-01-15 at 9 28 33 PM" src="https://github.com/user-attachments/assets/130acd7b-4a77-4d61-bb6d-172ca0709336" />

## Expected Behavior

<img width="41" alt="Pixelmator Pro 2025-01-15 22 08 05" src="https://github.com/user-attachments/assets/23371c06-0c9f-497e-8c84-1cccbcf38b5e" />

- Render a custom GestureDiagram on mobile for the help gesture (→↓←↓).
- Ensure the size and positioning are correct at all font sizes.
- When a gesture is in progress, separate segments should be highlighted on each swipe:

→
<img width="123" alt="Pixelmator Pro 2025-01-15 22 39 22" src="https://github.com/user-attachments/assets/b9b57080-5e59-4cee-8af4-d9926220a6c6" />

→↓
<img width="136" alt="Pixelmator Pro 2025-01-15 22 39 02" src="https://github.com/user-attachments/assets/22a76215-3b7a-4d73-8822-8d96d9046626" />

→↓←
<img width="131" alt="Pixelmator Pro 2025-01-15 22 38 51" src="https://github.com/user-attachments/assets/51805990-91bc-4ad2-a376-e1eaef911143" />

→↓←↓
<img width="131" alt="image" src="https://github.com/user-attachments/assets/ed84181d-44ef-4793-921c-dd6d6c91284b" />

PR #2789, opened 2025-01-28
custom help diagram edited

Howlmaster:
#2774


src/components/GestureDiagram.tsx
```diff
@@ -243,7 +243,17 @@ const GestureDiagram = ({
         const { x, y } = positions[i]
         return (
           <path
-            d={`M ${x} ${y} l ${segment.dx} ${segment.dy}`}
+            d={
+              path === 'rdld'
```
raineorshine, line 247: An inline comment would help clarify what's going on here, as it might not be completely obvious to other developers.

src/components/GestureDiagram.tsx
```diff
@@ -258,7 +268,7 @@ const GestureDiagram = ({
             strokeLinecap='round'
             strokeLinejoin='round'
             fill='none'
-            markerEnd={i === pathSegments.length - 1 ? `url(#${id})` : undefined}
+            markerEnd={path === 'rdld' ? undefined : i === pathSegments.length - 1 ? `url(#${id})` : undefined}
```
raineorshine, line 271: This would be better expressed as:

```ts
i === pathSegments.length - 1 && path !== 'rdld' ? `url(#${id})` : undefined
```

raineorshine (approved):
Thanks so much! Looks great.

Issue #2751, opened 2024-12-31
TreeNode not removed from DOM after thought is deleted


raineorshine:
As reported in https://github.com/cybersemics/em/pull/2596#pullrequestreview-2524454939.

We have [unmountOnExit](https://github.com/cybersemics/em/blob/42b1ef55abd5457ecbbbc39512be3dc6a1682fb5/src/components/LayoutTree.tsx#L497) so I'm not sure why it's still in the DOM.

## Steps to Reproduce

```
- a
- b
- c
```

Delete the thought `c`.

## Current Behavior

The TreeNode instance (`tree-node`) is not removed from the DOM after a thought is deleted.

![Screenshot 2024-12-27 at 6 54 53 PM](https://github.com/user-attachments/assets/95e64e0a-80dd-45aa-9bad-675ed57f216b)

## Expected Behavior

TreeNode elements should be removed from the DOM after a thought is deleted and the deletion animation completes.

raineorshine:
Fixed in ad18586a9b88b257df5e4265805a77a0488e1580

PR #2800, opened 2025-01-30
position faux caret onFocus and onChange

### LINKED ISSUE #2709: Render faux caret during indentation animation

ethan-james:
Fixes #2709 

Here is a rough draft of how a faux caret would work. I can see the 1 additional layout call from `getBoundingClientRect`, which I call twice, once to get the offset parent and once to get the selection rect itself. It seems that they are batched into a single layout call, hopefully.

Here is a simple example, tapping within the same thought.

main:
Screenshot 2025-01-30 at 14 26 18

this branch:
Screenshot 2025-01-30 at 14 25 07

And a more complex example where I tapped from one thought to a thought at a different indent level. I got quite a wide range of results for this example, but overall I think it's possible to use the faux caret while only adding 1 layout call.

main:
Screenshot 2025-01-30 at 14 40 00

this branch:
Screenshot 2025-01-30 at 14 36 33

The approach that I used is to change the faux caret `onTap` or `onChange`, since it will need to move as they type. This results in layout calls during typing. This approach would be combined with the `hideCaret` animations to show/hide the faux caret when indent levels change (and make it blink, of course) but it does mean that the event handlers that control the faux caret are always on whether or not the indent level is changing.

It would also be possible to enable/disable the event handlers so that they are only causing layout calls while the `hideCaret` animation is active, at the cost of additional code complexity. I *think* I can see a path to doing it without adding additional renders, but it's hard to say until I get into it.

And of course, there are all sorts of weird positioning issues that are caused by different margins or whatever that still need to be ironed out.


src/components/Editable.tsx
```diff
@@ -88,6 +88,7 @@ interface EditableProps {
   */
   transient?: boolean
   onEdit?: (args: { path: Path; oldValue: string; newValue: string }) => void
+  onThoughtFocus?: () => void
```
raineorshine, line 91: If you are looking for a way to avoid all the props drilling, perhaps you can create a new ministore that the FauxCaret component can subscribe to, that can be directly updated from `Editable`. Or maybe you have another idea.

src/components/Editable.tsx
```diff
@@ -420,6 +422,8 @@ const Editable = ({
         throttledChangeRef.current.flush()
         thoughtChangeHandler(newValue, { rank, simplePath })
       } else throttledChangeRef.current(newValue, { rank, simplePath })
+
+      requestAnimationFrame(() => onThoughtFocus?.())
```
raineorshine, line 426: Can you explain the reasoning behind `requestAnimationFrame`? 

While this is good for performance, it seems like it *might* cause the measurement to not occur in time for the first paint of the faux caret.

raineorshine (commented):
> And a more complex example where I tapped from one thought to a thought at a different indent level. I got quite a wide range of results for this example, but overall I think it's possible to use the faux caret while only adding 1 layout call.

Okay, that seems reasonable.

> The approach that I used is to change the faux caret `onTap` or `onChange`, since it will need to move as they type.

I don't think we should try to entirely replace the caret, as that opens up too much space for discrepancies from the native functionality. It will probably also save us from special case handling for typing, selection, etc.

How can we only show the faux caret when the native caret is invisible? The challenge is getting it to line up with the `hideCaret` functionality. `transitionend` might be helpful here? As for typing, we can always hide the faux caret immediately when the user starts typing.

I don't believe we will need to make it blink if it is just visible during the indentation animation.

src/components/Editable.tsx
ethan-james: As far as I can tell, the selection ranges aren't updated during the focus handler, so it needs to be delayed until the end of the frame using `requestAnimationFrame` or `setTimeout`.

src/components/LayoutTree.tsx
```diff
@@ -447,6 +454,39 @@ const TreeNode = ({
...
+  // Hide the faux caret when typing occurs.
+  editingValueStore.subscribe(() => {
+    if (isTouch && isSafari() && caretRef.current) {
+      caretRef.current.style.display = 'none'
+    }
+  })
+
+  // If the thought isCursor and edit mode is on, position the faux cursor at the point where the
+  // selection is created.
+  useEffect(() => {
```
ethan-james, line 466: It seems like I can get everything I need from the props. Let me know if you can think of any impact this would have on performance.

ethan-james:
I haven't figured out how to style the faux caret the same as the real caret, but I think that this is ready for review in terms of the approach I've taken.

Also, I feel like there should be a CSS variable like `--webkit-caret-color` or something, but I don't see one. Do you have any information about the caret styles?

raineorshine (changes requested):
Thanks for the update!

This seems to work on tap, but it is not triggered on other commands that change the indentation level such as Cursor Forward, Cursor Back, New Thought, Subcategorize One, Undo, Redo, etc. (Fwiw, I don't think that explicitly whitelisting all the commands that could trigger an indentation change is a good idea. Hopefully we can find a general condition that is a good fit.)

`selectionchange`? `transitionstart`? It's admittedly tricky as any false negatives or false positives can throw it off.

> I haven't figured out how to style the faux caret the same as the real caret, but I think that this is ready for review in terms of the approach I've taken.
> 
> Also, I feel like there should be a CSS variable like `--webkit-caret-color` or something, but I don't see one. Do you have any information about the caret styles?

We will eventually be setting the caret to a custom color with [::selection](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection), so we don't need to match the native color.

src/components/LayoutTree.tsx
```diff
@@ -446,6 +452,40 @@ const TreeNode = ({
...
+    // The selection ranges aren't updated until the end of the frame when the thought is fcoused.
+    requestAnimationFrame(() => {
+      if (isTouch && isSafari() && caretRef.current) {
+        if (editing && isCursor) {
+          const offset = fadeThoughtRef.current?.getBoundingClientRect()
+
+          if (offset) {
+            const rect = getBoundingClientRect()
+
+            if (rect?.x || rect?.y) {
```
ethan-james, line 474: I fixed at least one issue with gestures/commands, but I think the deeper issue is that `getBoundingClientRect` tends to return `0,0` when the parent element node is selected rather than its text node child. This is something that we discussed before here: https://github.com/cybersemics/em/issues/2753#issuecomment-2593940306.

Has any work been done to investigate or resolve the discrepancy between selections that are element nodes and selections that are text nodes? It would be nice to know if there's somewhere that I can look.

raineorshine: I'm not sure it's possible to resolve the discrepancy exactly. The browser has two representations for the same state, so we will always need to handle both. However we can hide the discrepancy within the selection module.

The difference can be detected with `selection.focusNode.nodeType === Node.ELEMENT_NODE`. For ELEMENT_NODE, you should be able to call `getBoundingClientRect` on the `focusNode` rather than the selection range.

ethan-james: Things are looking better now with the test cases you listed above. Do you have any more test cases for me to check?

There are a couple of issues that I haven't tackled yet:

1. empty thoughts have different padding from thoughts with text in them, so the caret offset is a little different and there's more of a jump between the faux caret and the real caret
2. when a thought is deleted the cursor moves to the previous thought, and the faux caret shows at the beginning of the thought while the real caret is positioned at the end

raineorshine:
Nice!

Hopefully these cases don't all need special handling, but I'm not sure how general your solution is yet so I thought I would just list as many cases as I could think of that might affect the indentation and caret. 

- Undo/Redo
- Indent/Outdent
- Table View
- Sort
- Jump Back
- Notes
- Split
  - By hitting the enter key while the caret it in the middle of the thought
- Merge
  - By hitting backspace at the beginning of a thought with a previous sibling
- Drag-and-drop
  - Caret can safely be hidden in this case if necessary

src/components/LayoutTree.tsx
ethan-james: > 2. when a thought is deleted the cursor moves to the previous thought, and the faux caret shows at the beginning of the thought while the real caret is positioned at the end

I discovered that `selection.focusOffset` is `1` when an ELEMENT_NODE is selected and the caret is at the end. Unfortunately, this isn't very helpful because `getBoundingClientRect` only gives the width & height of the bounding box, whereas the caret will often be positioned at the end of a line that doesn't stretch the full width of the box.

The only thing I can think of to fix this is to hide the faux caret in those circumstances. What do you think?

ethan-james: > Editables have padding-right of `1em` and min-width of `3em`, so for inline thoughts that do not take up the full width, the end of the thought is always a fixed distance from the right edge of the bounding box.
> 
> If that doesn't help, could you provide a screenshot or steps to reproduce? I'm not quite able to visualize the problem.

Sorry, I should have done that from the start. I mean multi-line thoughts like the following:

Screenshot 2025-02-12 at 09 55 45

That's on Chrome because it's easier to screenshot on my computer, but it's the same issue in Safari. The selected range looks like:

Screenshot 2025-02-12 at 10 04 02

and the selection as a whole looks like:

Screenshot 2025-02-12 at 10 07 44

So I don't know of a way to get a client rect that matches the caret position. Let me know if you do!

raineorshine: Thanks! I understand now. 

How about getting the position of the superscript? That will always be right after the caret in the ELEMENT_NODE offset 1 case.

ethan-james: Good thought. It looks like `ThoughtAnnotation` relies on an `inline-block` wrapper with a copy of the thought's value in it. Currently it looks like it only exists for thoughts with one of the annotations (`showSuperscript || url || email || styleAnnotation`) so if we went that route, the hidden copy would need to exist on all thoughts. Does that seem reasonable?

I think then I would control it using another CSS variable like `--faux-caret-line-end-opacity` that could be animated by `hideCaret` and overridden on `div[aria-label=tree-node]` if the normal faux caret should show, or not overridden if the line end caret should show.

raineorshine: Yes, that should be okay. I imagine there are more efficient approaches to `ThoughtAnnotation` itself, such as cloning the DOM node, but that's a separate issue. 

If you believe this is the best way to move forward, let's give it a try. Thanks!

ethan-james: This seemed to work out OK. Let me know what you think.

raineorshine: Could you provide exact steps to reproduce the ELEMENT_NODE with offset 1 case? I want to make sure I'm testing the same thing as you.

raineorshine:
After testing 77a6634, it looks like it's working quite well. I think the approach is successful, and the extra attention to detail will really help **em** look and feel like a native app. 

I noticed that the positioning of the faux caret is sometimes slightly off, and the Puppeteer tests show some visual regressions with the layout, but you are probably already aware.

We can do a more thorough review, starting with Trevin's review, whenever you think it's ready. Thanks so much!

src/components/LayoutTree.tsx
ethan-james: Creating a new sub-thought, then deleting it using backspace should do it. As long as the child is at a greater indent level than the parent so that the animations play.

ethan-james:
OK, I fixed the Puppeteer regressions and opted to add a line-start faux caret rather than using the method of calculating the start of the bounding box plus padding. This covers cases where `focusNode` is an ELEMENT_NODE and `focusOffset` is 0. That allowed me to position the faux caret better inside empty thoughts.

There still appear to be some cases where the faux caret shifts by a pixel or two, but I think it's ready for a more thorough review so that we can see when that is noticeable.

src/components/LayoutTree.tsx
```diff
@@ -446,6 +452,55 @@ const TreeNode = ({
...
+    if (isTouch && isSafari() && caretRef.current) {
+      caretRef.current.style.display = 'none'
+      setShowLineStartFauxCaret(false)
+      setShowLineEndFauxCaret(false)
+    }
+  })
+
+  // If the thought isCursor and edit mode is on, position the faux cursor at the point where the
+  // selection is created.
+  useEffect(() => {
```
raineorshine, line 469: I think it would be a good idea to factor the bulk of this logic out into a custom hook. It would be nice to put it all in a custom component, but `showLineStartFauxCaret` and `showLineEndFauxCaret` are used in `TreeNode` itself, so it seems like a custom hook is the best approach. I would suggest stuffing `showLineStartFauxCaret` and `showLineEndFauxCaret` into the hook and having it return an opaque style object since `TreeNode` doesn't really need to know the implementation details as long as `--faux-caret-line-start-opacity` and `--faux-caret-line-end-opacity` are applied correctly. Let me know what you think.

src/components/LayoutTree.tsx
```diff
@@ -446,6 +452,55 @@ const TreeNode = ({
...
+    if (isTouch && isSafari() && caretRef.current) {
```
raineorshine, line 470: Rather than wrap the entire effect in a conditional, I prefer to invert the condition and short circuit at the top. This avoids the extra indentation level and makes it a bit easier to read.

```ts
if (!isTouch || !isSafari() || !caretRef.current) return
```

src/components/LayoutTree.tsx
```diff
@@ -446,6 +452,55 @@ const TreeNode = ({
...
+      if (editing && isCursor) {
+        // The selection ranges aren't updated until the end of the frame when the thought is focused.
+        setTimeout(() => {
+          if (caretRef.current) {
```
raineorshine, line 474: Similar opportunity to short circuit. I would probably add `fadeThoughtRef` as well:

```ts
if (!caretRef.current || !fadeThoughtRef.current) return
```

src/components/LayoutTree.tsx
```diff
@@ -446,6 +452,55 @@ const TreeNode = ({
...
+
+            if (offset) {
+              const rect = getBoundingClientRect()
+
+              if (rect) {
+                const { x, y } = rect
+
+                caretRef.current.style.display = 'inline'
+                caretRef.current.style.top = `${y - offset.y}px`
+                caretRef.current.style.left = `${x - offset.x}px`
```
raineorshine, line 485: Is there a reason to use direct DOM manipulation here instead of the typical PandaCSS styling? `setShowLineStartFauxCaret` will trigger a React re-render anyway, so I don't think it saves us a re-render, but correct me if I'm wrong.

src/components/LayoutTree.tsx
```diff
@@ -446,6 +452,55 @@ const TreeNode = ({
     return lastPatches?.some(patch => deleteActions.includes(patch.actions[0]))
   })
 
+  const [showLineEndFauxCaret, setShowLineEndFauxCaret] = useState(false)
+  const [showLineStartFauxCaret, setShowLineStartFauxCaret] = useState(false)
+
+  // Hide the faux caret when typing occurs.
+  editingValueStore.subscribe(() => {
```
raineorshine, line 459: Can we use `editingValueStore.useEffect`, or is the raw subscribe needed for some reason?

`useEffect` handles unsubscribing, which is missing here.

src/components/LayoutTree.tsx
```diff
@@ -551,6 +608,21 @@ const TreeNode = ({
               prevWidth={treeThoughtsPositioned[index - 1]?.width}
             />
           )}
+        <span
+          className={css({
+            color: 'blue',
```
raineorshine, line 613: You can add a new color definition (`caret`? `selection`?) to https://github.com/cybersemics/em/blob/main/src/colors.config.ts.

raineorshine (changes requested):
Thanks for the update! I have to say, I'm impressed with the result. It is surprisingly consistent across many different cursor movements. Thanks to the general solution it is robust in the face of different commands, different directions of movement, and splitting/merging/moving thoughts.

I found one case of Undo/Redo where the faux caret is missing, but I'm not sure why. 

Notes and Table View require special handling, but since the base case works so well I am confident that we will be able to add support to those cases.

- [ ] Minor pixel differences at different font sizes
- [x] Indent/Outdent
- [x] Sort
- [x] Jump Back
- [x] Split
- [x] Merge
- [x] Drag-and-drop
- [ ] Undo/Redo
- [ ] Notes
- [ ] Table View

## Minor pixel differences 

https://github.com/user-attachments/assets/8c8495c9-05f4-4ed9-92a5-752640eb5a18

## Undo/Redo

https://github.com/user-attachments/assets/cc0fb933-0b32-4399-9d59-6a58e265628c

## Notes

TODO

## Table View

- Toggle table view on/off
- Cursor Forward from parent
- Cursor Forward from col1
- Cursor Back from col2
- Cursor Back from col1

https://github.com/user-attachments/assets/844e8b6a-735b-410f-aca2-c85cd15379dd

src/components/LayoutTree.tsx
ethan-james: Yeah, I was originally going with a renderless approach, but it got more and more complicated so I scrapped it. I'll clean up the DOM manipulation code.

ethan-james:
The refactor into a component & a hook is complete, and I fixed the faux caret when toggling table view.

Notes are rendered in a full-width `ContentEditable`, and have no equivalent of a `ThoughtAnnotation`. The caret is often (perhaps always?) placed at the end of the note, so in order to use the line end faux caret method, either a `NoteAnnotation` will need to be created, or else the `ContentEditable` width will need to be dynamic. Do you think either or these approaches will work, or do you have another suggestion?

raineorshine:
Making notes dynamic width would be an improvement. They should still have some right padding to make them easy to click. 

It's possible for the user to click and place the caret anywhere in the note, not just at the end. 

It would be preferable to position the caret without an additional
`visibility: hidden` element as ThoughtAnnotation does.

ethan-james:
I changed the `ContentEditable` inside `Note` to be `display: inline` on mobile Safari, which seems to have worked. I removed the 1em of right padding from the content editable (also only on mobile Safari) and applied it to the container instead. This has the effect of removing 1em of clickable area, which may have been the padding you meant, but it seems like if we want a larger clickable area, the right thing to do would be to add an `onFocus` handler to the container. Please let me know what you had in mind for clickable area.

As for clicking anywhere inside the note, I think that's only possible once the cursor is already on that thought. When moving from one indent level to another, which is the case I'm concerned with here, the caret always seems to go to the end. As a result, I only added a "line end" `FauxCaret` and not a "line start". If you know of test cases where the caret goes to the beginning or middle when changing indent depth, I can accommodate that as well.

I think the only other thing left to look at is the slight shift of the caret position that happens sometimes. It looks like it's relatively consistent when transitioning left vs transitioning right so I may be able to fix it. Or it might be related to the delay introduced by `setTimeout` in order to get the correct result from `window.getSelection` which could make it more difficult to nail down.

raineorshine:
> I changed the `ContentEditable` inside `Note` to be `display: inline` on mobile Safari, which seems to have worked. I removed the 1em of right padding from the content editable (also only on mobile Safari) and applied it to the container instead. This has the effect of removing 1em of clickable area, which may have been the padding you meant, but it seems like if we want a larger clickable area, the right thing to do would be to add an `onFocus` handler to the container. Please let me know what you had in mind for clickable area.

I havent tried the `onFocus` approach before, but if you have had success with that method feel free. I have only used padding. 

The end result should be that clicking within 1em from the right edge of the note should set the caret at the end of the note. This is the same as the thought behavior. 

> As for clicking anywhere inside the note, I think that's only possible once the cursor is already on that thought. When moving from one indent level to another, which is the case I'm concerned with here, the caret always seems to go to the end. As a result, I only added a "line end" `FauxCaret` and not a "line start". If you know of test cases where the caret goes to the beginning or middle when changing indent depth, I can accommodate that as well.

I think this will result in the caret being set to the middle of the note as the indent depth is changing:

1. Set the cursor on `d`. 
2. Click the note in between `Hello` and `world`. 

``` 
- a
  - b
    - =note
      - Hello world
    - c
      - d
``` 

> I think the only other thing left to look at is the slight shift of the caret position that happens sometimes. It looks like it's relatively consistent when transitioning left vs transitioning right so I may be able to fix it. Or it might be related to the delay introduced by `setTimeout` in order to get the correct result from `window.getSelection` which could make it more difficult to nail down.

Great, let me know what you find out!

ethan-james:
Rather than depending on the direction or the distance of indent (left vs right, near vs far) it seems like the offset problem is related to the text itself. You can see from this video that "Three", "Eight" and maybe "Nine" are just about perfect, while "Seven" is the worst.

https://github.com/user-attachments/assets/1d137671-56c2-452e-9851-f5e6dcadca62

I compared the parent offset (`fadeThoughtRef.current.getBoundingClientRect()`) to the selection bounding box (`window.getSelection().getRangeAt(0).getBoundingClientRect()`) and you can see from the screenshot that the caret bounding box is always some number of whole pixels away from its parent. This seems like a potential culprit, if the number that comes back from `getBoundingClientRect()` is in fact an approximation.

Screenshot 2025-02-26 105633

Performing the same comparison in Chrome, you can see that they're not always separated by whole pixels, although the differences are very small and I'm not totally sure that accounts for the discrepancy.

Screenshot 2025-02-26 110735

I think all I've accomplished so far is to discount the theories that it could be a fixed nudge based on left vs right indent, or that it's related to imperfect movement of the container from one frame to the next. Currently I don't have any bright ideas to resolve this, but let me know if you have any.

ethan-james:
To clarify, those rects in the screenshots are only from the main, positioned caret rather than from the new "line start" or "line end" carets, which don't get repositioned.

src/components/ThoughtAnnotation.tsx
```diff
@@ -148,7 +150,6 @@ const ThoughtAnnotation = React.memo(
           pointerEvents: 'none',
           userSelect: 'none',
           boxSizing: 'border-box',
-          width: '100%',
```
ethan-james, line 151: I originally scoped this change to `isMobileSafari()` but then realized that it impacts the existing annotations as well. Here is a before & after in Chrome:

With `width: '100%'` in a table:
Screenshot 2025-02-28 105701

Without `width: '100%'`:
Screenshot 2025-02-28 105732

I don't see any situations where removing the width causes problems, but let me know if you know of any. There's a bug when expanding the width of a thought in column 1, but it's unrelated to `width: '100%'` because it occurs either way:
Screenshot 2025-02-28 105242

raineorshine: I've traced the `width: 100%` back to #2168. Unfortunately it's connected with some sensitive layout behavior for the url icon. We have some limited test coverage with the puppeteer snapshot tests, but I will have to test manually.

Given that `width: 100%` was recently added, I assume that removing it will cause a regression, but I'm not sure what that is yet.

raineorshine:
Could you clarify, `useEffect` in which component?

ethan-james:
> Could you clarify, `useEffect` in which component?

Inside the `TreeNode` component in LayoutTree.tsx, the `useEffect` on line 465 or any other `useEffect` in that component doesn't run when the thought is re-added.

raineorshine:
You might need to merge `main`. #2751 reported that TreeNodes were not properly removed from the DOM. This could explain why `useEffect` was not firing again. That was fixed last week.

ethan-james:
I think that I already had that commit merged into my branch, but I am seeing better results now. I can get `useEffect` to fire when `fadeThoughtRef.current` changes, which should give me what I need.

ethan-james:
> 
> Do you notice any differences across font sizes?
> 

> 
> * Is there any padding that is affecting the measurement? `getBoundingClientRect` includes padding by default.
>   
>   * The `.editable` element has `box-sizing: border-box`. That shouldn't affect the measurement of `fadeThoughtRef`, though it may affect the selection's bounding box.
> * The `.editable` element is clipped with `-webkit-clip-path: inset(0.001px 0 0.1em 0)` to prevent the selection from being incorrectly set to the beginning of the thought when the top edge is clicked, and the end of the thought when the bottom edge is clicked. Could this be affecting one of the measurements?
> 

Thanks, I'll test these and share what I find.

> 
> Re: Note + Table View: I'm noticing that the caret is not visible when creating a new note, and the table view has some discrepancies depending on the selection offset. Are you still working on these?

I replaced `width: '100%'` on the thought annotation and added `textAlign: 'left'`, which fixed the incorrect behavior that I was seeing in the `Font Size: 13 - superscript` test. Now `Font Size: 18 (default) > superscript` is failing but it looks like an extremely minor discrepancy and might be flaky.

I'll look into Note + Table View, and any other discrepancies that affect the line start & end carets, which should have proper positioning independent of any issues with `getBoundingClientRect` measurements.

ethan-james:
I added a line start faux caret to Note, and that should fix empty notes in table view or otherwise.

I didn't notice any difference with or without `box-sizing: border-box` or `-webkit-clip-path: inset(0.001px 0 0.1em 0)`. At different font sizes, the discrepancy changes but looks like it's following the same general principle. Additionally, the offsets of the line start & line end faux carets get messed up. I tried changing them to use relative `em` values instead of `px`, but that didn't quite work either.

raineorshine:
Okay, thanks. I don't have any other ideas right now, but I think we should move forward with this. It's tolerable, and this is a big improvement on the invisible caret.

Let me know when you have tested it thoroughly and it's ready for a final review. Thanks!

ethan-james:
Great, I've tested pretty thoroughly and made a few tweaks. The only bug that I see is that sometimes the selection doesn't actually change when the cursor changes. Meaning that you can have the caret in B, then tap C, and the cursor will change to C but the selection will remain in B. This results in the caret being hidden during the animation, since the faux caret will only show on a thought if `isCursor` is true. If you can think of a better way to track where the selection actually is, please let me know, but this feels like a bug that's only semi-related to what I've been doing.

It looks like the puppeteer test that is failing is in fact flaky. This time the one that's failing is unrelated. I think you can proceed with your review, and hopefully everything works well!

panda.config.ts
```diff
@@ -59,12 +59,27 @@ const durations = Object.entries(durationsConfig).reduce(durationsReducer, {})
 
 const hideCaret = {
   '0%': {
+    '--faux-caret-opacity': 1,
+    '--faux-caret-line-start-opacity': 1,
+    '--faux-caret-line-end-opacity': 1,
+    '--faux-caret-note-line-end-opacity': 1,
+    '--faux-caret-note-line-start-opacity': 1,
```
raineorshine, line 66: Can you clarify why the note needs its own line start/end vars rather than using `--faux-caret-note-line-start-opacity`? I suspect that it is to ensure that the faux caret is rendered when shifting from a note to a thought, and a thought to a note, but please confirm.

src/hooks/useFauxCaretCssVars.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  // If the thought isCursor and edit mode is on, position the faux cursor at the point where the
+  // selection is created.
+  useEffect(() => {
+    if (!isMobileSafari()) return
+    if (editing && isCursor) {
+      // The selection ranges aren't updated until the end of the frame when the thought is focused.
+      setTimeout(() => {
+        if (editing && isCursor) {
+          if (isNote()) {
```
raineorshine, line 39: You should be able to use `store.getState().noteFocus` since we're already tracking the note state in the Redux store.

Either approach is reasonable, but less dependence on the DOM is usually preferred. Using `state.noteFocus` will eliminate the direct dependency between `selection` and the Note's aria-label.

src/device/selection.ts
```diff
@@ -466,7 +488,7 @@ export const getBoundingClientRect = () => {
   const selection = window.getSelection()
 
   if (selection && selection.rangeCount) {
-    return selection.getRangeAt(0).getBoundingClientRect()
+    if (isText()) return selection.getRangeAt(0).getBoundingClientRect()
```
raineorshine, line 491: You can add `isText()` to the existing `if` for a bit less complexity.

src/device/selection.ts
```diff
@@ -147,6 +153,22 @@ export const isOnLastLine = (): boolean => {
 /** Returns true if the browser selection is on a text node. */
 export const isText = (): boolean => window.getSelection()?.focusNode?.nodeType === Node.TEXT_NODE
 
+/** Returns true if the browser selection is on an element node and focus offset is 0.
+ * This represents a case where the browser will render the caret at the start of the text node's content.
+ */
+export const isStartOfElementNode = (): boolean => {
+  const selection = window.getSelection()
+  return !isText() && !!selection && selection.focusOffset === 0
+}
```
raineorshine, line 162: In my experience, there's usually no need to differentiate this case since `focusOffset === 0` in both cases and the semantics are the same in both cases (caret is at the beginning of the thought). It's only when `focusOffset === 1` that the semantics are different, as it refers to a node offset instead o a character offset.

Therefore, it *may* be possible to remove this function and simply use `focusOffset === 0`. However, maybe I'm missing something and the node type matters here.

Not a high priority, but might reduce some complexity if we only need `isEndOfElementNode`.

raineorshine (changes requested):
I have tested this to my best ability and the behavior appears to be correct in all relevant cases. Thanks so much!

I have documented the bugs with the hidden caret that exist in `main` in #2840, #2841, and #2842.

I have moved on to reviewing the design and architecture. The overall approach looks good, and some complexity is necessary. I think with some additional work, we can clean up the implementation and encapsulate more of this in the `FauxCaret` component, keeping `LayoutTree` as clean as possible. Happy to discuss any of the points below and weigh alternatives. Thanks!

src/hooks/useFauxCaretCssVars.ts
```diff
@@ -0,0 +1,68 @@
...
+  // but fadeThoughRef is re-created and will trigger an update.
+  fadeThoughtElement: HTMLDivElement | null,
+  isCursor: boolean,
+  isTableCol1: boolean,
+  path: Path,
+) => {
+  const [showLineEndFauxCaret, setShowLineEndFauxCaret] = useState(false)
+  const [showLineStartFauxCaret, setShowLineStartFauxCaret] = useState(false)
+  const [showNoteLineStartFauxCaret, setShowNoteLineStartFauxCaret] = useState(false)
+  const [showNoteLineEndFauxCaret, setShowNoteLineEndFauxCaret] = useState(false)
```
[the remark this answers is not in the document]
ethan-james: No, now that I have all of the different permutations nailed down, I can use something like an enum to specify one of the four should be shown. I'll do that now.

src/components/LayoutTree.tsx
```diff
@@ -427,8 +430,13 @@ const TreeNode = ({
   dragInProgress: boolean
   autofocusDepth: number
 } & Pick<CSSTransitionProps, 'in'>) => {
+  const editing = useSelector(state => state.editing)
```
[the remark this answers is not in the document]
ethan-james: Great point.

src/device/selection.ts
ethan-james: I believe the rationale was that there's padding on the element node itself, so that would need to be taken into account and we would still need to distinguish between an element node with `focusOffset === 0` and a text node with `focusOffset === 0`.

ethan-james:
I refactored so that the `editing`, `isCursor` & `isTableCol1` props are defined with `useSelector` inside the `PositionedFauxCaretWrapper`. I tried to do something similar to encapsulate `StaticFauxCaret` so that they're no longer reliant on CSS vars set in `LayoutTree` but I ran into an issue where there's no element ref to check in `ThoughtAnnotation`. This affects the edge case where a thought is deleted, then recreated using undo or redo. I'll take another pass tomorrow and see if I can either create a ref in `ThoughtAnnotation` and `Note` or else think of something cleverer.

ethan-james:
I got pretty far into refactoring the faux carets so that the opacity is set inside each one rather than at the `LayoutTree` level. I ran into an issue with the calls to `setTimeout`, where in certain circumstances one would see `window.getSelection()` returning an element node, and the other would see a text node. In that case, neither (or potentially both) carets would show.

I'll sleep on it and see if I can think of a solution, but in the meantime it might be worth reviewing the consolidated `LayoutTree` version that's currently committed.

src/components/LayoutTree.tsx
```diff
@@ -551,13 +559,17 @@ const TreeNode = ({
...
+          <FauxCaret opacity='var(--faux-caret-opacity)' />
+        </PositionedFauxCaretWrapper>
       </div>
     </FadeTransition>
   )
 }
 
 /** Lays out thoughts as DOM siblings with manual x,y positioning. */
 const LayoutTree = () => {
+  const editing = useSelector(state => state.editing)
```
raineorshine, line 572: Let's push this down to `useFauxCaretCssVars`, closer to where it used.

Generally we want to place selectors as close as possible to their use and avoid prop drilling. The only exception might be for performance reasons, but even then memoization may be a better solution. However, `state => state.editing` is trivial so that isn't a concern here.

src/components/LayoutTree.tsx
```diff
@@ -508,6 +554,7 @@ const TreeNode = ({
           width: isTableCol1 ? width : `calc(100% - ${x}px + 1em + 10px)`,
           ...style,
           textAlign: isTableCol1 ? 'right' : undefined,
+          ...fauxCaretCssVars,
```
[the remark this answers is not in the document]
raineorshine: If I'm understanding correctly, `fauxCaretCssVars` needs to be applied to the `TreeNode` so that it applies to `FauxCaret` instances within descendant `Note` and `ThoughtAnnotation` components and within the `TreeNode` itself. Is that right?

src/components/LayoutTree.tsx
```diff
@@ -690,11 +702,13 @@ const LayoutTree = () => {
   // We need to do this in a second pass since we do not know the height of a thought until it is rendered, and since we need to linearize the tree to get the depth of the next node for calculating the cliff.
   const {
     indentCursorAncestorTables,
+    tableDepth,
     treeThoughtsPositioned,
     hoverArrowVisibility,
   }: {
     // the global indent based on the depth of the cursor and how many ancestors are tables
     indentCursorAncestorTables: number
+    tableDepth: number
```
raineorshine, line 711: Can you add a JSDOC to document `tableDepth`?

src/components/LayoutTree.tsx
```diff
@@ -770,6 +785,7 @@ const LayoutTree = () => {
       // Calculate the cursor ancestor table width when we are on the cursor node.
       // This is used to animate the entire tree to the left as the cursor moves right.
       if (node.isCursor) {
+        tableDepth = node.isTableCol1 ? 1 : node.isTableCol2 ? 3 : 0
```
raineorshine, line 788: Is there a reason that `tableDepth` has moved from a `useSelector` to the iterative layout algorithm? It looks like it can be determined statically from the `State`.

src/components/PositionedFauxCaretWrapper.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { isSafari, isTouch } from '../browser'
+import { getBoundingClientRect } from '../device/selection'
+import attributeEquals from '../selectors/attributeEquals'
+import editingValueStore from '../stores/editingValue'
+import equalPath from '../util/equalPath'
+import head from '../util/head'
+
+type Props = PropsWithChildren<{
+  path: Path
+  wrapperElement: HTMLDivElement | null
```
raineorshine, line 14: Let's add JSDOC's for each prop (particularly `wrapperElement`).

src/components/PositionedFauxCaretWrapper.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * the selection occurred. This differs from what happens when an element node is selected through the use
+ * of commands (cursor down, undo/redo, etc.) where another faux caret is shown/hidden in a static position.
+ *
+ * See FauxCaret.tsx for more information.
+ */
+const PositionedFauxCaretWrapper = ({ children, path, wrapperElement }: Props) => {
+  const editing = useSelector(state => state.editing)
+  const isCursor = useSelector(state => equalPath(path, state.cursor))
+  const isTableCol1 = useSelector(state => attributeEquals(state, head(path), '=view', 'Table'))
+  const [styles, setStyles] = useState({})
```
raineorshine, line 30: This will need to be explicitly typed since Typescript cannot infer the type from `{}`.

src/hooks/useFauxCaretCssVars.ts
```diff
@@ -0,0 +1,57 @@
...
+  })
+
+  // If the thought isCursor and edit mode is on, position the faux cursor at the point where the
+  // selection is created.
+  useEffect(() => {
+    if (!isTouch || !isSafari()) return
+    if (editing && isCursor) {
+      // The selection ranges aren't updated until the end of the frame when the thought is focused.
+      setTimeout(() => {
+        if (editing && isCursor) {
```
raineorshine, line 36: I *think* `editing` and `isCursor` will always be true here, as the setTimeout callback has closure over the props. `editing` and `isCursor` in the outer scope will update since they are part of the memo dependencies, but I don't believe that will extend to the `setTimeout` callback.

If you need to ensure that `editing` and `isCursor` are still true after the `setTimeout` callback has fired, then you will need to import the `store` and do a fresh `store.getState()`.

src/hooks/useFauxCaretCssVars.ts
```diff
@@ -0,0 +1,57 @@
...
+            setFauxCaretType(isStartOfElementNode() ? 'noteStart' : isEndOfElementNode() ? 'noteEnd' : 'none')
+          } else {
+            setFauxCaretType(isStartOfElementNode() ? 'thoughtStart' : isEndOfElementNode() ? 'thoughtEnd' : 'none')
+          }
+        }
+      })
+    } else {
+      setFauxCaretType('none')
+    }
+  }, [editing, fadeThoughtElement, isCursor, isTableCol1, noteFocus, path])
```
raineorshine, line 47: `fadeThoughtElement`, `isTableCol1`, and `noteFocus` are listed as dependencies, yet they do not appear in the `useEffect` callback. If this is intentional, we should add a comment explaining that there is a hidden DOM dependency indirectly connected to these props that requires re-triggering the effect.

raineorshine (changes requested):
Thank you! I think we're almost there.

The overall footprint of `useFauxCaretCssVars`, `PositionedFauxCaretWrapper`, `FauxCaret`, and `hideCaret` is complex, but I think reasonable given the size of the task. Emulating the native caret and handling visibility, animation, and positioning is no small task, and I think we have done remarkably well.

I think we can hide/encapsulate a few more of the implementation details so that `LayoutTree`, `Note`, and `ThoughtAnnotation` can use the faux caret while knowing as little as possible. I've pointed these opportunities below. Please let me know if you have any questions/concerns!

There is one regression related to the available tap area of the Note, described below.

Thanks!

ethan-james:
Thanks for the review! I took care of some of the simpler items, and will revisit the others tomorrow.

src/components/LayoutTree.tsx
ethan-james: I needed to add `isTableCol1` as an additional check and figured that I could get them for free inside the layout algorithm. I'll put it back the way it was, with the additional check to `isRootedParent`.

src/components/LayoutTree.tsx
ethan-james: > If I'm understanding correctly, fauxCaretCssVars needs to be applied to the TreeNode so that it applies to FauxCaret instances within descendant Note and ThoughtAnnotation components and within the TreeNode itself. Is that right?

That's right. As a refresher, I tried to move the logic into the faux caret itself, but the `setTimeout` callbacks were returning different selections.

As a result, each `TreeNode` needs its own `fauxCaretCssVars` to show/hide its static faux carets depending on whether `isCursor` is true for that thought. I think I can do a little better by encapsulating the following logic into a `StaticFauxCaret`:

```
setOpacity(editing && isCursor ? 'var(--faux-caret-<caretType>-opacity)' : '0'
```

and then set `fauxCaretCssVars` globally on `LayoutTree`, so that only the layout tree and faux carets re-render when `editing` or `isCursor` changes rather than the `TreeNode` components themselves. Does that sound like it's worth pursuing?

raineorshine: Apologies for the delay!

> > If I'm understanding correctly, fauxCaretCssVars needs to be applied to the TreeNode so that it applies to FauxCaret instances within descendant Note and ThoughtAnnotation components and within the TreeNode itself. Is that right?
> 
> That's right. As a refresher, I tried to move the logic into the faux caret itself, but the `setTimeout` callbacks were returning different selections.
> 
> As a result, each `TreeNode` needs its own `fauxCaretCssVars` to show/hide its static faux carets depending on whether `isCursor` is true for that thought

Got it, thanks!

> I think I can do a little better by encapsulating the following logic into a `StaticFauxCaret`:
> 
> ```
> setOpacity(editing && isCursor ? 'var(--faux-caret-<caretType>-opacity)' : '0'
> ```
> 
> and then set `fauxCaretCssVars` globally on `LayoutTree`, so that only the layout tree and faux carets re-render when `editing` or `isCursor` changes rather than the `TreeNode` components themselves. Does that sound like it's worth pursuing?

I *think* so... but it's hard for me to imagine at the moment because I think the `setOpacity` interface is the wrong fit. I think we should abstract away the css variables, as described [here](https://github.com/cybersemics/em/pull/2800#discussion_r2001086238). Once that is refactored, I'll be able to better visualize the suggested change. If you have an intuition about simplifying it though, feel free to pursue.

src/components/LayoutTree.tsx
raineorshine: Perfect. I think the performance difference is negligible, so we should prefer the static/declarative approach when possible, limiting the layout algorithm to only the calculations that must be done procedurally/accumulatively.

src/components/LayoutTree.tsx
ethan-james: By pushing it down, do you mean back into `TreeNode`? That's where it was originally, until [this].(https://github.com/cybersemics/em/pull/2800#discussion_r1984999925)

src/components/Note.tsx
```diff
@@ -134,14 +135,17 @@ const Note = React.memo(({ path }: { path: Path }) => {
         marginLeft: fontSize - 14,
       }}
     >
+      <span className={css({ position: 'absolute', margin: '-4px 0 0 -4px' })}>
+        <FauxCaret fontSize='1.2em' opacity='var(--faux-caret-note-line-start-opacity)' />
```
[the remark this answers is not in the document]
ethan-james: I think that putting `fontSize` on the wrapper works fine, and also that encapsulating `opacity` inside `FauxCaret` is also a good idea.

src/components/LayoutTree.tsx
raineorshine: Hmmm... clearly there is some tension between performance and prop drilling. Thanks for your patience.

I will spare you my equivocation! It's fine as it is now. I may be guilty of premature optimization.

src/components/LayoutTree.tsx
raineorshine: Okay, thanks for trying! I will just make a note of what you tried so that I don't repeat it.

Can you clarify which wrapper you are referring to?

raineorshine:
I think this may have gotten merged the wrong way, as the diff is showing changes that are already in `main`.

e.g. The commit for `300.30.72` in `main` has commit hash 6cd78f3725f, but in this PR it has been rewritten to ada4632.

ethan-james:
> I think this may have gotten accidentally merged the wrong way, as the diff is showing changes that are already in `main`.
> 
> e.g. The commit for `300.30.72` in `main` has commit hash [6cd78f3](https://github.com/cybersemics/em/commit/6cd78f3725f884bbadea8eeb8eb1624053f01666), but here it has been rewritten to [ada4632](https://github.com/cybersemics/em/commit/ada4632130e46882c311b4783ccc7a830a40ec00).

Yuck, sorry. I did the merge in VS Code which is a relatively new workflow for me, but I don't think I did anything weird.

raineorshine:
Hmm. The main two ways of merging are either rebasing onto `main` or merging `main` into the branch with a merge commit. Neither of these approaches would change the hashes of commits that are already in `main`.

The problem with the current history is there is no way for me to get a clean diff for reviewing, as the commits are now interspersed. Maybe you could revert to 3308cef, and take another stab at it?

image

ethan-james:
I'm trying to get it to force push to the branch as we speak.

ethan-james:
OK, I think it's fixed now. I'll try the merge again more carefully.

src/device/selection.ts
raineorshine: Reflecting on this some more, I'm not quite convinced. When `focusOffset` is 0, the caret looks and behaves exactly the same whether it is on an element node or text node, doesn't it? My understanding is that it was overdetermined. 

Hypothetically, if we removed the `start` variants, what would break?

The only case that I could imagine causing problems is if there are descendant element nodes within a thought, like `Hello <b>World</b>` and the caret was at the beginning of the `<b>` element node. I'm not sure exactly how to reproduce that though.

raineorshine:
Perfect! I'm signing off now, but will review again asap.

src/device/selection.ts
ethan-james: Here is how `window.getSelection()` & `window.getSelection().getRangeAt(0).getBoundingClientRect()` look for text nodes (top 2 lines) and element nodes (bottom 2 lines).

Screenshot 2025-03-24 093805

In the case of an element node, there's still only 1 range, so that 0,0,0,0 rect is all that we get using this method. It could be used to place the positioned caret at the start of the element, but accommodations need to be made including:

1. setting the height of the faux caret to a default value rather than using the height of the rect
2. getting the padding of the thought to use as an offset rather than subtracting the thought's offset parent's rect as we do for text nodes

raineorshine: Thanks for looking into this. I see that even though `focusOffset` is the same, the bounding rect is different for element nodes. 

Given that additional accommodations would need to be made, I'm thinking it doesn't make sense to refactor this. I would only want a refactor here if it allowed us to eliminate code and reduce complexity.

raineorshine (changes requested):
Hi! This is looking great. We should be able to merge soon.

FYI I'm looking for a way to encapsulate the different pieces of the FauxCaret functionality, as they are tightly coupled but spread out. I'm leaning towards breaking my "one default export" rule in order to co-locate them all in a single module, and make a  few name changes to provide some more conceptual unity. I will let you know how it goes and give you an opportunity to provide feedback. 

In the mean time, I did find one rendering bug. Or maybe a regression, as I didn't notice this before. Could you look into this?

## Steps to Reproduce

Tested on mobile Safari.

```
- a
  - b
    - c
```

1. Set the cursor on `c`. 
2. Enter edit mode to open the virtual keyboard. 
3. Swipe →↓ to create a new subthought. 

## Current Behavior

The FauxCaret is rendered below the correct position for one frame. 

image

https://github.com/user-attachments/assets/79606ed7-5076-4700-a305-bf3d8055f89f

## Expected Behavior

The FauxCaret should be rendered in the correct position.

panda.config.ts
```diff
@@ -57,14 +57,37 @@ const zIndexDescending = (keys: string[]) => keyValueBy(keys.reverse(), (key, i)
...
     caretColor: 'transparent',
   },
+  /** The distributed asynchronous nature of the faux caret leads to situations where the animation
+   * begins before the caret is correctly repositioned. Delaying its appearance for at least 1 frame
+   * will eliminate these glitches.
+   */
+  '2.5%': {
+    '--faux-caret-opacity': 1,
+  },
```
ethan-james, line 76: Let me know if this at least fixes the flash of misplaced caret. It might not be the best solution, but I think any solution will involve some brittle async trickery and this was the simplest so I think it's a good jumping-off point for discussion.

raineorshine: Does the job!

raineorshine (approved):
Approved. I will do a final refactor before merging.

raineorshine:
First, I combined `PositionedFauxCaret` and `FauxCaret`.

Second, I renamed `hideCaret` and `useFauxCaretCssVars`, using the concept of a "Provider" which is analogous to React HOC components that are used to enable functionality for descendant components. While not perfectly parallel (one is a hook, the other is not; one uses className, the other uses style), I hope the naming makes them a little more conceptually unified. The are indeed tightly coupled, so the names are supposed to reflect that.

- `fauxCaretTreeProvider` - Provides the root FauxCaret CSS vars and animations for the LayoutTree. Applied to `className`.
- `useFauxCaretNodeProvider` - Provides FauxCaret CSS vars for specific nodes. Applied to `style`.

I have tested the changes to make sure I didn't break anything, and all seems good. If you could do a final review and let me know what you think, and if you have any additional feedback, that would be great!

src/components/LayoutTree.tsx
```diff
@@ -435,7 +435,13 @@ const TreeNode = ({
   // Since the thoughts slide up & down, the faux caret needs to be a child of the TreeNode
   // rather than one universal caret in the parent.
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
-  const fauxCaretCssVars = useFauxCaretCssVars(editing, fadeThoughtRef.current, isCursor, isTableCol1, path)
+  const fauxCaretCssVars = useFauxCaretCssVars({
```
ethan-james: This will create a new object during each render cycle and cause the hook to run. The expensive part is the useEffect which is still watching the primitives, so there's probably no issue here.

src/components/FauxCaret.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  // Hide the positioned faux caret when typing occurs.
+  editingValueStore.useEffect(() => {
+    if (!isTouch || !isSafari() || caretType !== 'positioned') return
+    setStyles({ display: 'none' })
+  })
+
+  // If the thought isCursor and edit mode is on, position the faux cursor at the point where the selection is created.
+  useEffect(() => {
+    if (!isTouch || !isSafari() || caretType !== 'positioned') return
```
ethan-james, line 59: This short-circuits for line start/end faux carets and avoids the expensive layout call below, which was my main concern about combining them. Looks good.

ethan-james (commented):
This would be approved, but I'm the original author so I can't.

src/components/LayoutTree.tsx
raineorshine: All hooks re-run every time the component re-renders, so I don't think passing the arguments individually or as an object will make any difference, unless I am missing something. Hooks are not memoized.

PR #2801, opened 2025-02-01
Factor out useOnPaste logic into importData action creator

RED-ROSE515:
In this PR, I have created a new action creator called importData.
And shift `useOnPaste`'s main logic into `importData` action creator.

### DIFF
--- a/src/actions/importData.ts
+++ b/src/actions/importData.ts
@@ -1,0 +1,117 @@
+import SimplePath from '../@types/SimplePath'
+import Thunk from '../@types/Thunk'
+import * as selection from '../device/selection'
+import rootedParentOf from '../selectors/rootedParentOf'
+import isMarkdown from '../util/isMarkdown'
+import strip from '../util/strip'
+import timestamp from '../util/timestamp'
+import { importFilesActionCreator as importFiles } from './importFiles'
+import { importTextActionCreator as importText } from './importText'
+import { newThoughtActionCreator as newThought } from './newThought'
+
+interface ImportDataPayload {
+  path: SimplePath
+  text: string
+  html: string | null
+  rawDestValue: string
+  transient?: boolean
+  isEmText?: boolean
+}
+
+/** Action-creator for importData. This is an action that handles importing content
+ * into the application, choosing between importText and importFiles based on the content type.
+ *
+ * This action is primarily used when:
+ * - Pasting content (plain text, formatted text, or HTML) into the application.
+ * - Importing content from external sources.
+ * - Handling clipboard data with mixed formats.
+ *
+ * Key features:
+ * - Handles both single-line and multi-line content.
+ * - Supports markdown content.
+ * - Can preserve or strip text formatting based on configuration.
+ *
+ * Differences from other import actions:
+ * - importText: Used for single-line content or markdown. Handles inline text insertion and replacement.
+ * - importFiles: Used for multi-line, non-markdown content. Creates new thought structures for each line.
+ *
+ * @param payload - Configuration object for importing data.
+ * @param payload.path - The path where the data should be imported.
+ * @param payload.text - The plain text content to be imported.
+ * @param payload.html - HTML content to be imported, or null if importing plain text.
+ * @param payload.rawDestValue - The untrimmed destination value to preserve whitespace when combining with existing content.
+ * @param payload.transient - If true, creates a new empty thought before importing the text.
+ * @param payload.isEmText - If true, preserves formatting when stripping HTML. Default: false.
+ *
+ * @returns A Thunk that handles importing the data based on whether it is multiline or markdown.
+ */
+export const importDataActionCreator = ({
+  path,
+  text,
+  html,
+  rawDestValue,
+  transient,
+  isEmText = false,
+}: ImportDataPayload): Thunk => {
+  return (dispatch, getState) => {
+    const state = getState()
+
+    // If transient first add new thought and then import the text
+    if (transient) {
+      dispatch(
+        newThought({
+          at: rootedParentOf(state, path),
+          value: '',
+        }),
+      )
+    }
+
+    const processedText = html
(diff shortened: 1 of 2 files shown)


src/actions/__tests__/onPaste.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    // Get the root children
+    const rootChildren = getAllChildren(store.getState(), HOME_TOKEN)
+
+    // Verify the structure
+    expect(rootChildren.length).toBe(2)
+
+    // Find 'a' and 'b' thoughts
+    const aThought = rootChildren.find(id => getThoughtById(store.getState(), id)?.value === 'a')
+    const bThought = rootChildren.find(id => getThoughtById(store.getState(), id)?.value === 'b')
```
raineorshine, line 59: To test the thought structure, use `exportContext` and assert the entire output. This is easier to read than walking the thought tree and asserting each level:

```ts
  const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')

  - a
    - a1
      - a2`)
```

src/actions/__tests__/onPaste.ts
```diff
@@ -0,0 +1,85 @@
...
+import { HOME_TOKEN } from '../../constants'
+import { getAllChildren } from '../../selectors/getChildren'
+import getThoughtById from '../../selectors/getThoughtById'
+import store from '../../stores/app'
+import { importDataActionCreator as importData } from '../importData'
+
+describe('onPaste', () => {
+  beforeEach(() => {
+    // Reset store to initial state before each test
+    store.dispatch({ type: 'CLEAR_THOUGHT_CACHE' })
```
raineorshine, line 11: Maybe this was left in from previous testing?

src/actions/importData.ts
```diff
@@ -0,0 +1,88 @@
...
+interface ImportDataPayload {
+  path: SimplePath
+  text: string
+  html: string | null
+  rawDestValue: string
+  transient?: boolean
+  isEmText?: boolean
+}
+
+/** Action-creator for importData. */
```
raineorshine, line 21: It would be great to have a more descriptive JSDOC that explain in more detail what this action does. Some questions it can answer are:

- What does it do?
- When would it be used?
- How does it different from `importText` and `importFiles`?

src/actions/importData.ts
```diff
@@ -0,0 +1,88 @@
...
+import { importTextActionCreator as importText } from './importText'
+import { newThoughtActionCreator as newThought } from './newThought'
+
+interface ImportDataPayload {
+  path: SimplePath
+  text: string
+  html: string | null
+  rawDestValue: string
+  transient?: boolean
+  isEmText?: boolean
```
raineorshine, line 18: Parameters should be described with JSDOC comments to make it easier for other developers to understand the use of this action.

src/components/Editable/useOnPaste.ts
```diff
@@ -43,86 +36,26 @@ const useOnPaste = ({
...
-        // Check if the text is markdown, if so, prefer importText over importFiles
-        const markdown = isMarkdown(text)
-        // Resumable imports (via importFiles) import thoughts one at a time and can be resumed if the page is refreshed or there is another interruption. They have a progress bar and they allow duplicates pending descendants to be loaded and merged.
-        // Non-resumable imports (via importText), in contrast, are atomic, fast, and preserve the browser selection. Due to the lack of support for duplicates pending descendants, they are only used for single line imports.
-        if (!multiline || markdown) {
-          dispatch(
-            importText({
-              // use caret position to correctly track the last navigated point for caret
-              // calculated on the basis of node type we are currently focused on. `state.cursorOffset` doesn't really keep track of updated caret position when navigating within single thought. Hence selection.offset() is also used depending upon which node type we are on.
-              caretPosition: (selection.isText() ? selection.offset() || 0 : cursorOffsetState) || 0,
```
raineorshine, line 85: It looks like some of the inline comments have been lost in this refactor. Could you go back through and make sure that all comments are preserved?

raineorshine (changes requested):
Thank you for your submission!

> This issue is caused because we're stripping the HTML structure too aggressively in `importData` action creator.
> So, I have modified the `importData` action to better handle nested HTML structures.

Could you be more specific? What structure needs to be preserved?

> I added a check for nested list structure using /
> 
> .*?</ul>/s.test(html).
> When we detect a nested list structure, we preserve the HTML structure and only clean up meta tags.
> For non-nested HTML content, we continue with the existing strip behavior.

My main concern with this approach is that there is overlap in functionality between the two code branches, and that it doesn't utilize the existing functionality.

`strip` does some initial processing before the html is passed to [importFiles](https://github.com/cybersemics/em/blob/0294fa4a4664150441634dfc9292b3b473bbc313/src/actions/importFiles.ts#L399) and eventually to [textToHtml](https://github.com/cybersemics/em/blob/0294fa4a4664150441634dfc9292b3b473bbc313/src/util/textToHtml.ts) and [htmlToJson](https://github.com/cybersemics/em/blob/0294fa4a4664150441634dfc9292b3b473bbc313/src/util/htmlToJson.ts). Despite the name, `textToHtml` actually can take HTML as input. It normalizes the HTML, preparing it for the full parse by himalaya.

Given the well-established pipeline that exists, it seems like we should be identifying the bug and finding a solution within that rather than bypassing it completely. Bypassing it may work for the particular case that we are testing here, but it skips a lot of other functionality and will likely break other use cases. The existing pipeline really *ought* to work on any HTML input. 

Handling Notion HTML should amount to a minimal transformation that nudges the structure into something the existing HTML parsing algorithm understands, not creating a separate branch of code to maintain. The overly aggressive `strip` points to either a bug or a design problem in the existing import pipeline that needs to be addressed. I suggest analyzing the existing import pipeline and making a more targeted intervention.

---

I might suggest just refactoring `useOnPaste` into `importData` in this PR, and opening a separate PR for the import fix. Then we can get the refactor merged and focus on the bug fix.

raineorshine (approved):
Thanks!

Issue #2747, opened 2024-12-30
Import nested bulleted list from Notion


raineorshine:
## Steps to Reproduce

```
- a
    - m
        - x
- b
    - m
        - y
```

1. Paste the above content into [Notion](https://www.notion.com/).
2. Copy the content from Notion into **em**.
   - Make sure to select as blocks rather than just selecting the text
   <img width="585" alt="image" src="https://github.com/user-attachments/assets/03145334-2376-449a-b0e6-c7dfe2079fb7" />

## Current Behavior

```
- a
- m
- x
- b
- m
- y
```

The text/html content is:

```
<meta charset='utf-8'><ul>
<li>a
<ul>
<li>m
<ul>
<li>x</li>
</ul>
</li>
</ul>
</li>
<li>b
<ul>
<li>m
<ul>
<li>y</li>
</ul>
</li>
</ul>
</li>
</ul>
<!-- notionvc: dc8eb4c5-e3d2-4bce-b414-86b8ea5744ec -->
```

## Expected Behavior

```
- a
    - m
        - x
- b
    - m
        - y
```

## Testing

Add unit test.

raineorshine:
Fixed in #2814

PR #2806, opened 2025-02-05
Fix notion nested bulleted list import error

RED-ROSE515:
This PR resolves the issue : https://github.com/cybersemics/em/issues/2747

This issue is occured because `strip` function removes the nested list and return a list.

So, in this PR, I added a new `processHtmlContent` function that checks if the HTML has nested list structure.
If it does, preserves the structure while only cleaning up unnecessary artifacts
If not, falls back to the original stripping behavior.

I use this function instead of directly calling `strip` on the HTML content.

### DIFF
--- a/src/actions/importData.ts
+++ b/src/actions/importData.ts
@@ -32,8 +32,7 @@ const processHtmlContent = (html: string, isEmText: boolean): string => {
   }
 
   // Otherwise strip the HTML
-  return strip(html, { preserveFormatting: isEmText, stripColors: !isEmText })
-    .replace(/\n\s*\n+/g, '\n')
+  return strip(html, { preserveFormatting: isEmText, stripColors: !isEmText }).replace(/\n\s*\n+/g, '\n')
 }
 
 /** Action-creator for importData. This is an action that handles importing content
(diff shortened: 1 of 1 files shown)


raineorshine (commented):
Hi, thanks for the new submission.

Here are my thoughts on this approach:

- `strip` does a lot of work, including converting <br> tags, decimal codes, formatting tags, html sanitization, colors, and empty formatting attributes. `processHtmlContent` does not do any of this, and there is no test coverage for these cases. It is assuming that Notion-like HTML will not need of this extra handling, which I don't think is safe.
- Two separate code pathways to maintain
- Lacking historical context: Notion HTML worked at one point. When was the regression introduced, and what caused it? Identifying the regression should probably have been the first thing to happen, long before a proposed solution.
- Lastly, as we recently discovered, HTML importing in general is broken in `main` right now. It perhaps broke when `importFiles` was added, but further research is needed. That is a blocker for this issue. It doesn't make sense to fix a subset of HTML imports when the general case is not working.

So, unfortunately I think there are bigger questions to answer before we can propose a solution to Notion HTML.

I'm going to put this PR on hold until we have a better understanding of the original regression and consider how this may or may not be fixed within the existing `strip` function.

raineorshine:
Closing in favor of #2814. It restored general HTML import by removing the early `strip`, at the expense of some other import cases. See PR for details.

Issue #2528, opened 2024-11-02
Animation: swapParent


raineorshine:
## Steps to Reproduce

```
- a
  - b
```

1. Set the cursor on `b`
2. Swap Parent

## Current Behavior

No animation.

https://github.com/user-attachments/assets/cdf4b34a-a3fc-4083-9deb-b5f6e1dbd0ce

## Expected Behavior

- [x] `a` should curve left and `b` should curve right (counter-clockwise) as they animate into their new positions.
- [x] The greater thought (via https://github.com/cybersemics/em/blob/main/src/util/compareThought.ts#L188) should curve right, while the lesser thought should curve left.

## Notes

I recommend [layered animation](https://tobiasahlin.com/blog/curved-path-animations-in-css/) to simulate animating along a curve.

zukilover:
Currently, swapping the parent does not trigger position changes. I'm still working on finding a way to replicate the `moveThought` functionality and integrate it into the swapParent action.

raineorshine:
Yes, that's tricky. I wonder if you could alter the `key` to trick React into swapping the parent and child, thus causing them to have an animated position change.

zukilover:
I added a `value` to the key, which isn't the ideal solution, but I did it just to test if it would trigger the animation. The result was that it triggered the thought fade transition, but not the x and y transitions for the tree nodes. I’m considering an approach where we first swap the nodes, then only swap the thought IDs (not the values) to trigger the x and y transitions. I’m open to any suggestions or alternative ideas.

raineorshine:
I'm not sure off the top of my head. Since the ids and thus keys are "wrong" due to the `swapParent` logic, it seems like we will need to swap DOM nodes with no animation either at the beginning or end of the swap in order. Then the animation will at least start or end with the correct key.

PR #2810, opened 2025-02-08
Set up animation for swapParent action

zukilover:
#2528


zukilover:
@raineorshine Here are the changes I made in this PR:  

1. Replaced `editThought` with `moveThought` when swapping a parent. This establishes a solid foundation for node transitions, but a structural change to `LayoutTree` is required for the animation to work correctly. Otherwise, it will only animate the left and top positions in a linear manner.  

   https://github.com/cybersemics/em/blob/888e626bfa17a70a5ca186769e8bd5eed96ebfe9/src/actions/swapParent.ts#L44-L50

2. In the next set of updates, I separated the `left` and `top` transitions by applying them to the `tree-node` div and its new child div. This follows the layered animation approach: the parent div (`tree-node`) handles the left transition, while its child div handles the top transition.  

   https://github.com/cybersemics/em/blob/6b8583a4d863ffe24e05c78ac23aac01837c9008/src/components/LayoutTree.tsx#L508-L511

https://github.com/cybersemics/em/blob/6b8583a4d863ffe24e05c78ac23aac01837c9008/src/components/LayoutTree.tsx#L528-L530

   However, this change caused a test failure:  
  https://github.com/cybersemics/em/blob/6b8583a4d863ffe24e05c78ac23aac01837c9008/src/e2e/puppeteer/__tests__/table-view.ts#L41-L42

   The snapshot comparison indicated that the test failed because the table columns appeared to shrink. However, comparing the real app from this branch with the main branch showed no visible difference.  

   Another issue is that when dragging, the node gets selected instead of initiating the actual drag action.  

3. In my latest attempt, I replaced the child div's `top` transition with a `transform` transition:  
  https://github.com/cybersemics/em/blob/363ca6e769684b95646e72572bdabbb6bb48ea79/src/components/LayoutTree.tsx#L532-L533

   This works by shifting the parent's `translateY` by `-y(px)`, then negating it in the child using `translateY(y(px))` (since `-y + y = 0`).  

   However, this resulted in the worst test failures, breaking all tests related to drag-and-drop behavior. It also caused the same draggable issue as Solution 2.  

Now, I'm trying to apply the layered transition only when the last action was `swapParent`. The challenge is that I can't find an effective way to target the parent node in this case, as its last action could be anything other than `swapParent`.

raineorshine:
Thank you for the update and the additional background on the approaches you have tried.

> 1. Replaced `editThought` with `moveThought` when swapping a parent. This establishes a solid foundation for node transitions, but a structural change to `LayoutTree` is required for the animation to work correctly. Otherwise, it will only animate the left and top positions in a linear manner.

Thanks. To clarify, does this mean that the rewritten `swapParent` is sufficient to cause the the parent and child to be animated as they switch positions, albeit without the counter-clockwise rotation described in the original issue?

> 2\. In the next set of updates, I separated the `left` and `top` transitions by applying them to the `tree-node` div and its new child div. This follows the layered animation approach: the parent div (`tree-node`) handles the left transition, while its child div handles the top transition.

That sounds reasonable.

When splitting up `left` and `top`, I wonder if the axis that is not defined should be zeroed out. e.g. `left: x, top: 0` and `top: y, left: 0`. I'm not sure, but I have run into issues before omitting top or left from `position: absolute` elements.

Similarly with width, I'm not sure if `width: 100%` or omitting width makes a difference here. Just something to think about.

> The snapshot comparison indicated that the test failed because the table columns appeared to shrink. However, comparing the real app from this branch with the main branch showed no visible difference.

Table view is pretty buggy right now, making this a bit hard to test. We can follow up on this specific case once we have a solid approach.

> Another issue is that when dragging, the node gets selected instead of initiating the actual drag action.

Interesting. Do you have any thoughts on why this might occur? None of the drag-and-drop refs have changed. The only thing that has changed is the extra absolutely positioned div ancestor.

> 3\. In my latest attempt, I replaced the child div's `top` transition with a `transform` transition

Yes, I would expect this to be a problem, as noted here:

https://github.com/cybersemics/em/blob/a714a89e17cdc359c25c681aa66f83822c72f22d/src/components/LayoutTree.tsx#L489

`transform` would be better for performance, but the stacking context issue is a problem.

> Now, I'm trying to apply the layered transition only when the last action was `swapParent`. The challenge is that I can't find an effective way to target the parent node in this case, as its last action could be anything other than `swapParent`.

Can you talk a bit about how you think this approach improves on the previous one? Does it address the drag-and-drop issues observed?

---

Another idea is to leave `top/left` as-is, and layer an additional container with its own `top/left` that can "nudge" the element into the desired rotation. Layered animations are easiest to reason about when each axis has its own layer (x and y), but technically the browser can composite two layers, each with their own x,y animations.

zukilover:
> Thanks. To clarify, does this mean that the rewritten `swapParent` is sufficient to cause the the parent and child to be animated as they switch positions, albeit without the counter-clockwise rotation described in the original issue?

This is almost correct. The `swapParent` refactor alone is sufficient to animate the child but not the parent during the swap.

> When splitting up `left` and `top`, I wonder if the axis that is not defined should be zeroed out. e.g. `left: x, top: 0` and `top: y, left: 0`. I'm not sure, but I have run into issues before omitting top or left from `position: absolute` elements.

Zeroing out these properties does not affect the animation, so I have implemented the suggestion.

> Similarly with width, I'm not sure if `width: 100%` or omitting width makes a difference here. Just something to think about.

This cannot be done, as it will cause an issue when dragging. Omitting the width will result in the dragging state displaying a narrow blue line instead of a full one. See the video below for reference. 

https://github.com/user-attachments/assets/1873332f-072b-4dac-a5e2-80dba260980a


> Table view is pretty buggy right now, making this a bit hard to test. We can follow up on this specific case once we have a solid approach.

I still am not sure how to fix the test issue. Any suggestions would be greatly appreciated.


> Interesting. Do you have any thoughts on why this might occur? None of the drag-and-drop refs have changed. The only thing that has changed is the extra absolutely positioned div ancestor.

This seems to be caused by the fact that the `y` position has now been moved to the child div instead of the `tree-node`. Inspecting the element reveals that all parent elements are stacked at the very top of the `LayoutTree`'s root element.

The drag and drop is working as expected from the last commit.

> Yes, I would expect this to be a problem, as noted here:

I have now implemented the layered transition using the top and left properties instead of transform. 

> Can you talk a bit about how you think this approach improves on the previous one? Does it address the drag-and-drop issues observed?

The idea is to apply these structural changes only while swapping nodes. I assume this would preserve the original structure and prevent the tests from breaking.

> Another idea is to leave `top/left` as-is, and layer an additional container with its own `top/left` that can "nudge" the element into the desired rotation. Layered animations are easiest to reason about when each axis has its own layer (x and y), but technically the browser can composite two layers each with their own x,y animations.

I have tried this approach, unfortunately this breaks more tests than before. 


The latest updates in this PR have achieved the desired animation for swapping the parent.
I'm still trying to figure out how to fix the broken test for the table view.


https://github.com/user-attachments/assets/69a760ec-43f3-4b5a-9e0d-b542a736e0b7

raineorshine:
Thanks for the additional information. The animation looks great!

Drag-and-drop appears to be working, so I won't comment on this further unless a problem comes up. I will test more thoroughly at the end to make sure there are no regressions. (I'm noticing drag-and-drop bugs in `main` that incorrectly activate multiselect so I will make sure to ignore those for the purpose of this PR.)

> I still am not sure how to fix the test issue. Any suggestions would be greatly appreciated.

I can confirm that this is a regression and I am able to reproduce it locally. Pasting a table with wide col1 thoughts results in an overly narrow col1 the first few times it is rendered.

The rendering algorithm roughly works like this:

1. Normal TreeNodes are rendered with a dynamic css width of [calc(100% - ${x}px + 1em + 10px)](https://github.com/cybersemics/em/blob/c14011d197dc2116c3398b42cdc80d228fcabb65/src/components/LayoutTree.tsx#L508), i.e. the remaining width available up to the right edge of the screen.
2. Col1 TreeNodes are initially rendered with `width: undefined` which should default to 100% for the block level `tree-node` dev.
    - `width` is initially undefined because none of the `sizes` have been set yet. The thoughts have to get rendered once before their widths can be [measured](https://github.com/cybersemics/em/blob/c14011d197dc2116c3398b42cdc80d228fcabb65/src/components/VirtualThought.tsx#L144) and sizes set.
4. After the first render, all TreeNodes in col1 are rendered with the [max measured width](https://github.com/cybersemics/em/blob/c14011d197dc2116c3398b42cdc80d228fcabb65/src/components/LayoutTree.tsx#L752-L761) of thoughts in that column.

The problem is that the change in this PR is changing the timing/number of renders/paints such that col1 gets an incorrect width measurement. Take a look at the value of `tableCol1Widths` in `main` vs in this PR when the sample case is pasted:

`main`:

Brave Browser 2025-03-01 12 12 12

`eea4d6bfea9`:

Brave Browser 2025-03-01 12 14 12

A few points of explanation:

- The stepped increase in width is kind of a chain reaction resulting from the initial incorrect width. Once a thought is measured with width `23.984`, the fixed width will be set for all col1 thoughts, constraining their width even if they need more space. Yet each render will result in the width increasing as it is rendered and measured with padding (tentatively). Col1 thoughts are supposed to be first rendered with unconstrained width so that the max width can be properly measured and then applied in the next render to all col1 thoughts.
- [useLayoutEffect](https://github.com/cybersemics/em/blob/c14011d197dc2116c3398b42cdc80d228fcabb65/src/components/VirtualThought.tsx#L162) ensures that the size is measured and the thought is re-rendered before the next paint. However, any change to the timing of renders or DOM measurements could potentially disrupt this.

My guess it that the initial `width: undefined` is no longer rendering the col1 thought with width 100% on the first render. Without an initial render with unconstrained width, the first measurements of col1 thoughts will be incorrect. You could try changing the `isTableCol1` [condition](https://github.com/cybersemics/em/blob/c14011d197dc2116c3398b42cdc80d228fcabb65/src/components/LayoutTree.tsx#L508) to `isTableCol1 && width` when the TreeNode width is set so that it defaults to the dynamic remaining width rather than `undefined`.

Another thing to try is changing the inner div from `position: absolute` to `position: relative`. This will allow the element to affect the container's width, which may fix the `width: undefined` case.

Or, the timing/number of renders/paints is throwing off the initial measurement for different reasons.

Hope that helps!

src/actions/swapParent.ts
```diff
@@ -20,28 +20,48 @@ const swapParent = (state: State) => {
...
+
+  const childThought = getThoughtById(state, childId)
+  const parentThought = getThoughtById(state, parentId)
+  if (!childThought || !parentThought) return state
+
+  // Get children of the child thought that will be moved
+  const childChildren = getChildrenRanked(state, childId)
+
+  // First move the child to replace its parent's position
+  let stateNew = moveThought(state, {
```
raineorshine, line 34: (Feel free to ignore if this was intentionally temporary!)

I recommend using utilities such as `reducerFlow` to compose reducers. Mutation is considered a bad practice, and this project adheres to functional programming principles.

src/actions/swapParent.ts
```diff
@@ -20,28 +20,48 @@ const swapParent = (state: State) => {
...
+    cursor: [childId] as [ThoughtId, ...ThoughtId[]],
+    offset: childThought.value.length,
+  }
 }
 
 /** Action-creator for swapParent. */
 export const swapParentActionCreator = (): Thunk => dispatch => dispatch({ type: 'swapParent' })
 
-export default _.curryRight(swapParent)
+export default swapParent
```
raineorshine, line 67: Assuming you removed currying for testing purposes, but this should be restored before merging.

raineorshine (changes requested):
FYI The reducer isn't quite working as expected. I've added test coverage in e2d1d384b43.

zukilover:
@raineorshine @trevinhofmann 

I've modified the width calculation logic for table column 1 thoughts to ensure proper initial width measurements:
- Only applies fixed width to col1 thoughts when we have actual measurements (`width !== undefined`)
- Otherwise uses the dynamic calculation that extends to the right edge

Please kindly review.

src/components/Editable.tsx
```diff
@@ -581,6 +581,7 @@ const Editable = ({
 
   return (
     <ContentEditable
+      data-rank={rank}
```
raineorshine, line 584: Unused?

src/actions/swapParent.ts
```diff
@@ -20,25 +22,67 @@ const swapParent = (state: State) => {
...
+  const grandparent = parentOf(parent)
+
+  // Get siblings (other children of parent excluding the child being swapped)
+  const parentChildren = getChildrenRanked(state, parentId)
+  const siblings = parentChildren.filter(sibling => sibling.id !== childId)
 
-  return reducerFlow(reducers)(state)
+  return reducerFlow([
+    // First move the child to replace its parent's position
+    state =>
```
raineorshine: Currying really shines in situations like this where `state` is passed through the flow functions, but is not needed for the arguments.

src/actions/swapParent.ts
```diff
@@ -20,25 +22,67 @@ const swapParent = (state: State) => {
...
+          newRank: sibling.rank,
+        })
+      }, state),
+
+    // Keep cursor on the child at its new position, preserving the full path
+    state => ({
+      ...state,
+      cursor: [...grandparent, childId, parentId] as [ThoughtId, ...ThoughtId[]],
+      offset: childThought.value.length,
+    }),
```
raineorshine: Generally we want to compose over the `setCursor` reducer to ensure the cursor invariants are maintained and `state.expanded` is updated appropriately.

There was a bug where the parent thought was not expanded when just setting `state.cursor` directly. It didn't manifest because it was incidentally expanded as an only-child.

raineorshine (changes requested):
The animation works beautifully!

I found another `swapParent` functional regression. Added test coverage in 3f70828b50c. Could you look into this?

raineorshine:
This animation should be limited to the `swapParent` command.

## Steps to Reproduce

```
- a
  - b
- c
```

1. Indent `c`
2. Outdent `c`

## Current Behavior

`c` animates on a curve.

## Expected Behavior

`c` should animate in a straight line.

Thoughts should only animate on the curved line on `swapParent`.

raineorshine:
There is one more requirement from the original issue:

> The greater thought (via [main/src/util/compareThought.ts#L188](https://github.com/cybersemics/em/blob/main/src/util/compareThought.ts?rgh-link-date=2024-11-02T18%3A04%3A33.000Z#L188)) should curve right, while the lesser thought should curve left.

src/components/Editable.tsx
zukilover: Yes, this has been removed now.

src/actions/swapParent.ts
zukilover: Thanks, that makes sense. I'll keep that in mind.

src/actions/swapParent.ts
zukilover: Makes sense. I appreciate the explanation.

zukilover:
> This animation should be limited to the `swapParent` command.
> 
> ## Steps to Reproduce
> ```
> - a
>   - b
> - c
> ```
> 
> 1. Indent `c`
> 2. Outdent `c`
> 
> ## Current Behavior
> `c` animates on a curve.
> 
> ## Expected Behavior
> `c` should animate in a straight line.
> 
> Thoughts should only animate on the curved line on `swapParent`.

I've restricted the curved animation to `swapParent` while refactoring `TreeNode` in [074d761](https://github.com/cybersemics/em/pull/2810/commits/074d7617dcba01e0e318d362f4a6919d34de0a2e).

zukilover:
I'm still working on the last acceptance criteria: The greater thought (via https://github.com/cybersemics/em/blob/main/src/util/compareThought.ts#L188) should curve right, while the lesser thought should curve left.

zukilover:
> There is one more requirement from the original issue:
> 
> > The greater thought (via [main/src/util/compareThought.ts#L188](https://github.com/cybersemics/em/blob/main/src/util/compareThought.ts?rgh-link-date=2024-11-02T18%3A04%3A33.000Z#L188)) should curve right, while the lesser thought should curve left.

This has been fixed in https://github.com/cybersemics/em/pull/2810/commits/580370ed87b4b9278a401a29bbfee784c7177555.

I have to create nested ternaries otherwise it will throw ` @pandacss/no-property-renaming` error.

Let me know your thoughts.

trevinhofmann (changes requested):
Thanks for the updates, @zukilover!

The basic swap parent animation seems to be working correctly. I would expect a bit more of a curve in the animation (it's difficult to notice), but I'll defer to @raineorshine for deciding that.

<details>
<summary>Normal Speed</summary>

https://github.com/user-attachments/assets/cfb7ac72-89cd-4d1f-80a2-771c6036cf26
</details>

<details>
<summary>0.1x Speed</summary>

https://github.com/user-attachments/assets/27a8b1b2-0e58-4b74-a0d3-9d8821fcfec4
</details>

---

With the curve as it is, it's difficult for me to tell which thought curves left and which thought curves right. Slowing it down, it _looks like_ the child-becoming-parent thought is always curving to the right while the parent-becoming-child thought always curves to the left. From the second criteria of #2528, I think the direction should be dependent on which thought is "greater".

In the demo below, you can see that `c` curves to the right of `b` when `c` is becoming the parent, and `b` curves to the right of `c` when `b` is becoming the parent. If I understand the criteria correctly, `c` is the "greater" thought and should always curve right.

<details>
<summary>Frame by frame demo</summary>

https://github.com/user-attachments/assets/df6ea65f-0b06-47cb-92c6-c0e78ca2ea5b
</details>

---

## Issue: Additional jump animation after the swap parent animation

### Steps to Reproduce

```
- a
  - b
    - c
```

1. Focus `c`
2. Swap parent
3. Either continue typing on thought `b` or click thought `c`

### Current Behavior

`b` and `c` jump up and back down

<details>
<summary>Clicking</summary>

https://github.com/user-attachments/assets/703b2676-bb23-439d-bc90-d9d79d00f131
</details>

<details>
<summary>Typing</summary>

https://github.com/user-attachments/assets/b1a5f59b-d4ea-42b2-b512-02a216029cb3
</details>

### Expected Behavior

The swap animation should already be complete, so no additional animations should happen.

raineorshine (changes requested):
Yeah, the animation curve doesn't look quite right now. It's almost as if the child animates directly towards its parent and then starts curving upward later. It needs to appear as a single, arced curve. The animation curve in c603365 is a good reference.

## Current Behavior

0c3fe82c4f0

Brave Browser 2025-03-29 08 14 05

https://github.com/user-attachments/assets/e8b870f7-7aa4-4ab3-930d-5a10efe2155b

## Expected Behavior

c603365



https://github.com/user-attachments/assets/407d6045-8afc-4649-8268-bdb3f7eec6c7

src/components/LayoutTree.tsx
```diff
@@ -428,12 +532,24 @@ const TreeNode = ({
...
     const lastPatches = state.undoPatches[state.undoPatches.length - 1]
     return lastPatches?.some(patch => patch.actions[0] === 'newThought')
   })
 
+  // true if the last action is swapParent or if the last action is not moveThought
+  const isLastActionSwapParent = useSelector(state => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    return (
+      lastPatches?.some(patch => patch.actions[0] === 'swapParent') ||
+      lastPatches?.every(patch => patch.actions[0] !== 'moveThought')
```
raineorshine, line 549: I'm not really sure what the intention is here, but these conditions are not orthogonal. If the last action is `swapParent`, then `!== 'moveThought'` is necessarily true.

raineorshine (changes requested):
Thanks, the animation curve looks good now!

---

I think there is some confusion about the expected behavior of the second requirement, the direction of the curved path.

`isLastActionSwapParent` is the only condition that should determine *if* the curve is animated. `isGreaterThought` should have no affect on *whether or not* the thought follows a curved path. It only affects which curved path to use.

If `isLastActionSwapParent` is true, then `isGreaterThought` is used to determine the path of the curve. If the child is greater than the parent, it should curve right. If the child is less than the parent, it should curve left.

Screenshot 2025-03-30 at 9 35 30 AM

Screenshot 2025-03-30 at 9 35 26 AM

In other words, if the child value is greater than the parent value, the thoughts should appear to swap places in a counter-clockwise motion. Otherwise, clockwise.

Please let me know if that makes sense, or if I can clarify it further!

src/components/LayoutTree.tsx
zukilover: Good catch. This does not make sense. I will fix it.

zukilover: This has been reverted. I've fixed the unexpected animation after swapping parent by checking if there is any `swapParent` action from the last patches.

src/actions/__tests__/swapParent.ts
```diff
@@ -31,7 +34,11 @@ it('no-op if cursor is a root thought', () => {
     - b
      - c`
 
-  const steps = [importText({ text }), setCursor(['a']), swapParent()]
+  const steps = [
+    importText({ text }),
+    (state: State) => setCursor({ path: contextToPath(state, ['a'])! })(state),
```
trevinhofmann, line 39: I generally dislike using `!` (non-null assertion operator) in TS as it's an unsafe escape hatch from the types we've set up, but I'm less worried in this case since we're only using it in the test suite. This comment is non-blocking and does not need to be addressed 👍

src/components/Editable.tsx
trevinhofmann: `data-rank={rank}` is still in the diff and appears to be unused.

src/components/LayoutTree.tsx
```diff
@@ -378,6 +379,126 @@ const linearizeTree = (
   return thoughts
 }
 
+/** Renders a thought node with proper positioning and transitions. */
+const ThoughtNode = ({
```
trevinhofmann, line 383: @zukilover - Could you please explain this change of moving out `ThoughtNode`? I assume the justification is similar to the one for `TreeNode` and might be a bit outside of my understand, but I would still like to ask. I'm mostly concerned about potential performance impacts on the `LayoutTree` with this change (though I don't see anything in particular that is concerning 👍) while not understand why it is necessary. Thank you!

src/components/LayoutTree.tsx
```diff
@@ -446,6 +593,19 @@ const TreeNode = ({
...
+  // 2. Create L-shaped animation paths by controlling when each movement starts/ends
+  // 3. Prevent interference between transitions that could occur in a single node
+
+  // Track x position changes. When x prop changes, update state to trigger CSS transition.
+  // State updates ensure smooth transitions by forcing a re-render that applies the new position.
+  useLayoutEffect(() => {
+    if (x !== _x) {
+      setX(_x)
+    }
+  }, [x, _x])
```
trevinhofmann, line 607: I believe we discussed this before in a related change, but I'm still fairly confident that the `if x !== _x` check is unnecessary because the `setX` call will be a no-op if the values are the same.

https://react.dev/reference/react/useState#setstate-caveats

> If the new value you provide is identical to the current `state`, as determined by an [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison, React will **skip re-rendering the component and its children.** This is an optimization. Although in some cases React may still need to call your component before skipping the children, it shouldn’t affect your code.

src/components/LayoutTree.tsx
```diff
@@ -471,87 +631,59 @@ const TreeNode = ({
...
-            leaf={leaf}
-            onResize={setSize}
-            path={path}
-            prevChildId={prevChild?.id}
-            showContexts={showContexts}
-            simplePath={simplePath}
-            singleLineHeight={singleLineHeightWithCliff}
-            // Add a bit of space after a cliff to give nested lists some breathing room.
-            // Do this as padding instead of y, otherwise there will be a gap between drop targets.
-            // In Table View, we need to set the cliff padding on col1 so it matches col2 padding, otherwise there will be a gap during drag-and-drop.
```
trevinhofmann, line 531: Was it intentional to remove some of these comments? Unless I'm missing something, this seems like an example of one which is still relevant.

trevinhofmann (changes requested):
Thank you for the latest updates, @zukilover! The animation seems much better, and everything is functioning as described in #2528.

I have a few minor comments on the code changes, but it looks great overall!

src/components/LayoutTree.tsx
zukilover: That’s exactly the issue we’re addressing here. We need to ensure the DOM update is applied synchronously so the transition works correctly. The alternatives are using flushSync, which isn't ideal, or useEffect, as we’ve implemented here. This approach should not impact performance, it’s simply a way to better orchestrate updates for smoother animations.

src/components/LayoutTree.tsx
zukilover: This was unintentional. I've added back comments.

src/actions/__tests__/swapParent.ts
zukilover: Thanks, I will keep this in mind.

.vscode/settings.json
```diff
@@ -6,6 +6,6 @@
   // otherwise it will override prettier.requireConfig in projects that do not use Prettier
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[typescriptreact]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
```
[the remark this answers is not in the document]
zukilover: This was unintentional. I've reverted the change.

src/components/LayoutTree.tsx
zukilover: This was part of figuring out the best way to stack the layered animations. I decided to keep it separate for better readability.

src/components/LayoutTree.tsx
raineorshine: I think @trevinhofmann is asking about the condition rather than the overall approach. If it's truly a no-op when `x === _x`, then the condition is (by definition) not needed. So if it is needed, then `setX` must have some kind of side effect in that case.

zukilover: Oh I see now. I’ll double-check whether setX introduces any side effects and get back to you. 

If there are no unintended consequences, I’ll remove the extra check.

src/components/LayoutTree.tsx
```diff
@@ -426,14 +560,39 @@ const TreeNode = ({
...
+  cursorIsGreaterThanPrevSibling: boolean
 } & Pick<CSSTransitionProps, 'in'>) => {
   const [y, setY] = useState(_y)
+  const [x, setX] = useState(_x)
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
+
+  const isParentSwapped = useSelector(state => {
+    const allPatches = state.undoPatches
+    return allPatches?.flat().some(patch => patch.actions[0] === 'swapParent')
+  })
```
raineorshine, line 572: This condition is true if `swapParent` was activated *at any point in the undo history*, not just the last action. I'm not sure what the intention is here, but clearly the behavior should not depend on whether the user activated `swapParent` previous to the last action.

src/components/LayoutTree.tsx
```diff
@@ -426,14 +560,39 @@ const TreeNode = ({
...
   const isLastActionNewThought = useSelector(state => {
     const lastPatches = state.undoPatches[state.undoPatches.length - 1]
     return lastPatches?.some(patch => patch.actions[0] === 'newThought')
   })
 
+  // true if the last action is swapParent
+  const isLastActionSwapParent = useSelector(state => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    return lastPatches?.some(patch => patch.actions[0] === 'swapParent')
+  })
```
raineorshine, line 584: `isLastActionSwapParent` should probably be defined in `ThoughtNode`, since that is the only place it is used.

src/components/LayoutTree.tsx
```diff
@@ -426,14 +560,39 @@ const TreeNode = ({
...
+
+  const isLastActionEditThought = useSelector(state => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    return lastPatches?.some(patch => patch.actions[0] === 'editThought')
+  })
```
raineorshine, line 589: This can't be right. Nowhere in the intended behavior does it say that `swapParent` should behave differently if the last action was `editThought`.

src/components/LayoutTree.tsx
```diff
@@ -426,14 +560,39 @@ const TreeNode = ({
...
+
+  const isLastActionSetCursor = useSelector(state => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    return lastPatches?.some(patch => patch.actions[0] === 'setCursor')
+  })
```
raineorshine, line 594: This also can't be right. Nowhere in the intended behavior does it say that `swapParent` should behave differently if the last action was `setCursor`.

This also leads to bizarre behavior where setting the cursor with `keydown` causes a faulty animation, but setting the cursor with `click` does not.

https://github.com/user-attachments/assets/ddb800b3-e8eb-4380-a383-f2cb68f24eac

src/components/LayoutTree.tsx
```diff
@@ -890,6 +1041,26 @@ const LayoutTree = () => {
     treeThoughts,
   ])
 
+  // Calculate isGreaterThought at LayoutTree level
+  const cursorIsGreaterThanPrevSibling = useSelector(state => {
```
raineorshine, line 1045: Doesn't this just need to compare the thought with its parent? No need to use `treeThoughtsPositioned` to access the render order. 

If so, this can go in `ThoughtNode`.

src/components/LayoutTree.tsx
```diff
@@ -890,6 +1041,26 @@ const LayoutTree = () => {
...
+      // Get the current thought and previous thought to compare
+      const index = treeThoughtsPositioned.findIndex(thought => thought.isCursor)
+      const prevThought = index > 0 ? getThoughtById(state, head(treeThoughtsPositioned[index - 1].path)) : null
+
+      // Compare the current thought with the previous thought to determine the animation curve
+      const thoughtComparison: number = prevThought && cursorThought ? compareThought(cursorThought, prevThought) : 0
+      return thoughtComparison > 0
+    } catch (error) {
+      return false
+    }
```
raineorshine, line 1061: try-catch is generally only for unhandleable runtime errors. Instead, detect the appropriate condition and return false in that case.

Note that try-catch also comes with a performance cost, so that's another reason not to use it except in the case of unexpected runtime faults.

src/components/LayoutTree.tsx
```diff
@@ -378,6 +379,138 @@ const linearizeTree = (
...
+            : transition,
+      })}
+      style={outerDivStyle}
+    >
+      <div
+        className={css({
+          position: 'absolute',
+          transition: isLastActionSwapParent
+            ? cursorIsGreaterThanPrevSibling
+              ? isLastActionNewThought
```
raineorshine, line 484: The last action cannot be both `swapParent` and `newThought` at the same time. Thus, a nested ternary with these two conditions doesn't make sense.

raineorshine (changes requested):
Thanks @zukilover. The animation looks good now, and you got the clockwise/counterclockwise behavior perfectly.

Some of the conditions used to determine the animation behavior are clearly wrong in principle. I'm not sure what the desired outcome was, so I couldn't provide more direction, but I tried to point out the logical fallacy. Please think through these a bit more and make sure that they not only behave correctly, but make sense in theory. If you need assistance, let us know what you are trying to accomplish, and we can work together to identify the correct condition.

In addition, I found one regression in the context view, which I added a test for.

Thanks so much for your time and effort.

src/components/LayoutTree.tsx
zukilover: @raineorshine  This change addresses the issue where a node continues to use the non-layered animation even after the swapParent action has occurred. That’s why we're combining this with the other two actions, setCursor and editThought. Although they might seem unrelated, they play a role in this behavior, as shown in this line:
https://github.com/cybersemics/em/blob/f35fdc188915eeacdd13281b8f72c096c6762d52/src/components/LayoutTree.tsx#L436

With this approach, the swapped node avoids the unnecessary linear Y-axis movement that occurs after the parent swap.

src/components/LayoutTree.tsx
zukilover: Please see this comment: https://github.com/cybersemics/em/pull/2810#discussion_r2030008163

src/components/LayoutTree.tsx
zukilover: Please see this comment: https://github.com/cybersemics/em/pull/2810#discussion_r2030008163, I'll check how I can handle the keydown event in this case.

src/components/LayoutTree.tsx
raineorshine: It sound like there is a problem to solve in terms of which animation to use after `swapParent` has occurred, or how to "reset" the animation state.

The current approach perhaps works in some specific cases, but it is most definitely incorrect as a general solution. It is logically incorrect on multiple levels:

1. Checking for `swapParent` anywhere in the undo history means that 0 swaps can behave differently than 2 swaps, even though they represent exactly the same state.
2. The steps to reproduce in the original issue sets the cursor on `b` as a minimal test case, but there are an *infinite number of actions* that could lead to the cursor ending up on `b`. On purely theoretical grounds, `isLastActionEditThought ` and `isLastActionSetCursor`  cannot possibly be causal determinants of the desired `swapParent` animation behavior.

I think some additional thought needs to go into this to come up with a general solution.

src/components/LayoutTree.tsx
raineorshine: Any action could exist before `swapParent` is activated. This needs to work regardless of the previous action.

src/components/LayoutTree.tsx
zukilover: Understood. I’ll work on a more comprehensive solution to resolve the issue.

trevinhofmann:
@zukilover, just a heads up that there are merge conflicts since TreeNode and several hooks were recently refactored out of LayoutTree.tsx. It may be worth considering if we should extract the ThoughtNode out as well.

raineorshine:
FYI I removed the special case animation for `newThought` in f0a496e9416 based on real-world testing. Just a heads up as it will result in a small merge conflict.

zukilover:
Thanks @raineorshine. I'm still trying to get this test to pass in [`swapParent.ts`](https://github.com/cybersemics/em/blob/c844351f80c6e909c0c4684ef69670d34eafc7cf/src/actions/__tests__/swapParent.ts#L158)

with this initial structure:
```
- a
    - m
      - x
  - b
    - m
      - y1
        - y
```

When I swap y to its parent, the path would be `[ 'b', 'm', 'y1', 'y' ]` which I think is expected because we're now in a context view. But the new test case expects the paths to be `['a', 'm', 'b', 'y1', 'y']`.

Could you explain why would 'b' come after 'm', and does 'a' need to be visible when it's in context view?

raineorshine:
Sure! Let me see if I can help clarify.

> with this initial structure:
> 
> ```
> - a
>     - m
>       - x
>   - b
>     - m
>       - y1
>         - y
> ```

I think this was pasted incorrectly, so I'm going to go off the test you referenced:

```
- a
  - m
    - x
- b
  - m
    - y
      - y1
```

> When I swap y to its parent, the path would be `[ 'b', 'm', 'y1', 'y' ]` which I think is expected because we're now in a context view.

I assume you mean swap `y1` with its parent, since that is what the test says.

`[ 'b', 'm', 'y1', 'y' ]` would be the `Path` if `swapParent` was activated on `y1` in normal view. In context view, the `Path` would be different, as I describe below.

> Could you explain why would 'b' come after 'm'

This has to do with how thought ids are appended to the cursor when navigating into the context view. First the context view is activated on cursor `['a', 'm']`. Then, instead of showing `m`'s children, it will show `m`'s contexts (i.e. all of the thoughts in which the lexeme 'm' appear). These are: `a` and `b`. When you navigate to one of the contexts, it will be appended to the cursor. So if you navigate to the `b` context in context view, the cursor will be `['a', 'm', 'b']`. If you navigate one deeper, the cursor will be `['a', 'm', 'b', 'y']`. And one deeper will be `['a', 'm', 'b', 'y', 'y1']`. Notice that this corresponds exactly to the visible "path" of ancestors (seen in the breadcrumbs at the bottom of the screen.)

This is the difference between `SimplePath` and `Path`. `SimplePath` must traverse from the root of thoughtspace; `Path` can traverse across context views. `state.cursor` is a `Path`, so we have to be careful not to convert it to a `SimplePath`, otherwise we will lose our place in nested context views.

For more information, see: https://github.com/cybersemics/em/wiki/Docs#332-context-view

zukilover:
I've updated swapParent to correctly preserve context view paths, which should maintain the intended context view navigation structure. Let me know if this approach works better for you.

zukilover:
I just noticed the Puppeteer test failed for some reason. I'll take a look into it now.

src/actions/swapParent.ts
```diff
@@ -1,14 +1,16 @@
 import _ from 'lodash'
 import State from '../@types/State'
 import Thunk from '../@types/Thunk'
-import editThought from '../actions/editThought'
+import moveThought from '../actions/moveThought'
+import { getChildrenRanked } from '../selectors/getChildren'
 import getThoughtById from '../selectors/getThoughtById'
-import simplifyPath from '../selectors/simplifyPath'
```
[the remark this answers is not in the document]
zukilover: Understood. The latest changes maintain `simplifyPath` for the `moveThought` operations while preserving the full Path information for cursor navigation. This keeps the context view chain intact while still correctly targeting the proper thought locations during the swap.

src/actions/__tests__/swapParent.ts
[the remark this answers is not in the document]
zukilover: Got it, thank you.

src/actions/swapParent.ts
```diff
@@ -20,25 +22,62 @@ const swapParent = (state: State) => {
...
     }),
-    editThought({
-      oldValue: parentValue,
-      newValue: childValue,
-      path: simplifyPath(state, parent),
+
+    // Then move the parent under the child
+    moveThought({
+      oldPath: parent,
+      newPath: appendToPath([childId], parentId),
```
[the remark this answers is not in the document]
zukilover: Thanks to point this out. We're now using `simplifyPath` instead of `appendToPath`.

src/components/LayoutTree.tsx
zukilover: I've updated this so that non-layered animations are only applied during thought movement. This should resolve the unexpected extra animation that occurred when a node was swapped. Let me know if this makes sense.

src/components/LayoutTree.tsx
zukilover: This has been moved to `TreeNode`.

src/components/LayoutTree.tsx
zukilover: I've fixed the approach in https://github.com/cybersemics/em/pull/2810/commits/a3fd82890102922d18a4bf362b590f2ba2fa18e1

src/components/LayoutTree.tsx
zukilover: This has been fixed in https://github.com/cybersemics/em/pull/2810/commits/a3fd82890102922d18a4bf362b590f2ba2fa18e1

src/components/LayoutTree.tsx
zukilover: `isLastActionNewThought` is no longer used here.

src/actions/__tests__/swapParent.ts
```diff
@@ -58,4 +60,134 @@ it('swaps child thought with parent', () => {
...
+      - x
+  - b
+    - m
+      - y1
+        - y`)
+
+    expectPathToEqual(stateNew, stateNew.cursor, ['a', 'm', 'b', 'y1', 'y'])
+  })
+
+  it('swaps child with parent in context view', () => {
```
trevinhofmann, line 161: I added this test as an example of the following case:

<details>
<summary>Error when swapping parent</summary>

https://github.com/user-attachments/assets/99ebadb7-ab60-445d-9d2d-7332419fba57
</details>

What is actually the expected behavior when attempting to swap with a parent which is the focus of the context view? cc @raineorshine

I put my best guess in the expectations of this test case, but it may by incorrect. Please feel free to update this test accordingly.

Regardless of the expected functionality, it shouldn't be throwing this error at `const stateNew = reducerFlow(steps)(initialState())`:

```
 FAIL  |unit| src/actions/__tests__/swapParent.ts > context view > swaps child with parent in context view
Error: moveThought: sourceThought undefined not found
 ❯ moveThought src/actions/moveThought.ts:61:11
     59|   if (!sourceThought) {
     60|     console.error({ oldPath, newPath, offset, skipRerank, newRank })
     61|     throw new Error(`moveThought: sourceThought ${sourceThoughtId} not…
       |           ^
     62|   }
     63| 
 ❯ wrapper node_modules/lodash/lodash.js:5255:19
 ❯ src/util/reducerFlow.ts:15:24
 ❯ src/util/reducerFlow.ts:14:14
 ❯ swapParent src/actions/swapParent.ts:106:4
 ❯ apply node_modules/lodash/lodash.js:488:27
 ❯ wrapper node_modules/lodash/lodash.js:5101:16
 ❯ src/util/reducerFlow.ts:15:24
```

src/actions/swapParent.ts
```diff
@@ -1,44 +1,109 @@
...
-    editThought({
-      oldValue: childValue,
-      newValue: parentValue,
-      path: simplifyPath(state, cursor),
-      force: true,
+    // Create new cursor path by combining:
+    // 1. All segments except the last one
+    // 2. All elements from the last segment except the last two
+    // 3. The swapped elements in reverse order
+    const combinedPath = [...contextChain[0], ...lastSegment.slice(0, -2), swapElements[1], swapElements[0]]
```
trevinhofmann, line 62: Pardon my limited familiarity with the context feature (and context chains), but is it possible for the `contextChain` to have a length greater than 2? If so, it seems we are omitting any elements of the `contextChain` except the first and last when creating `combinedPath`.

Please let me know if I am misunderstanding this part of the code for creating the new cursor.

trevinhofmann (changes requested):
Thank you for the updates @zukilover. Functionality still seems to be working in most of the "happy paths", but I have a couple of concerns in the threads below about the context view.

src/actions/__tests__/swapParent.ts
raineorshine: Thanks for pointing out this case.

`swapParent` should be disallowed on a context in the context view, as there is no sensible outcome. (The context view shows all of the parents across the entire thoughtspace in which a lexeme appears. "Swapping" something in place of the lexeme would be sort of like renaming all its parents across the entire thoughtspace, a widespread and destructive action, and certainly far from the intuitive idea of a parent-child swap.)

Activating `swapParent` in this case should alert the user that `swapParent` is not allowed in the context view, and otherwise do nothing.

Note that this does not preclude the use of `swapParent` on deeper descendants. They should exactly the same as normal view. The only care that must be taken is to preserve the context chain, and not accidentally collapse cursor into a `SimplePath`.

src/actions/swapParent.ts
raineorshine: Yes, `contextChain` can have length greater than 2, i.e. there can be multiple nested context views. 

Anyway, this action should probably not be doing low level context chain operations.

src/actions/swapParent.ts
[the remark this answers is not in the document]
raineorshine: I'm sorry if I pointed you in the wrong direction. The Wiki was just to provide some background information on Context View. You probably shouldn't be operating on context chains in the action itself. That is a low level structure.

As long as the cursor is preserved and the moved paths are simplified, no special case is needed to handle context view descendants.

See: c78b688beab

src/actions/__tests__/swapParent.ts
raineorshine: Added in [0325a39](https://github.com/cybersemics/em/pull/2810/commits/0325a39aad29db622698ed3b73bd716beacb26ab)

raineorshine: The case of `swapParent` on a child of a context in the context view is actually nontrivial as well. While it *may* be possible to create a sensible outcome for this action, for the sake of expediency I have disallowed it for now (01a6e6677c6).

src/components/TreeNode.tsx
```diff
@@ -81,14 +84,39 @@ const TreeNode = ({
...
+  // true if the last action is swapParent
+  const isLastActionSwapParent = useSelector(state => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    return lastPatches?.some(patch => patch.actions[0] === 'swapParent')
+  })
+
+  const isLastActionMoveThought = useSelector(state => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    return lastPatches?.some(patch => patch.actions[0] === 'moveThought')
+  })
```
raineorshine, line 96: You're probably going to run into trouble with this. Any action can come before the swap, including actions that use `moveThought` internally. For example, `categorize` uses `moveThought` internally, but its action type is `categorize`. Additionally, `undo/redo` breaks the rule, as the temporally last action may come from `undoPatches` or `redoPatches`.

Basically, any condition that checks for a specific action before the swap is going to have breaking cases. A swap may be preceded by *any* action. We cannot rely on a specific preceding action.

raineorshine (changes requested):
Hi, thanks for the update.

I am seeing an issue with other animations after `swapParent` occurs.

## Steps to Reproduce

```
- a
  - b
    - c
  - d
```

1. Set the cursor on `a/b`
2. Activate `swapParent`
3. Navigate between `a` and `b`

## Current Behavior

`d` no longer animates.

https://github.com/user-attachments/assets/a56583d3-4b31-495a-b53c-e64be01ad1ed

## Expected Behavior

`d` should smoothly animate.

https://github.com/user-attachments/assets/0a78e1aa-4e72-47c1-bddf-81470623511f

src/components/TreeNode.tsx
zukilover: I'm still working on this. I'm now approaching it from a different angle to see if the dynamic top positions are what's causing the unexpected trailing animation. The goal is to avoid relying on any conditions based on undoPatches.

zukilover: I've fixed this by setting the `top` position of the outer div to `auto` by default. The root cause was that it had been set to 0, which caused the animation to start at the top of the viewport, resulting in an unexpected trailing transition.

The final result now looks like this:

https://github.com/user-attachments/assets/e77bcfa1-8a8b-4672-8212-8f86ae277bf1

src/components/TreeNode.tsx
```diff
@@ -81,14 +84,29 @@ const TreeNode = ({
...
-      // When y changes React re-renders the component with the new value of y. It will result in a visual change in the DOM.
-      // Because this is a state-driven change, React applies the updated value to the DOM, which causes the browser to recognize that
-      // a CSS property has changed, thereby triggering the CSS transition.
-      // Without this additional render, updates get batched and subsequent CSS transitions may not work properly. For example, when moving a thought down, it would not animate.
-      setY(_y)
-    }
+    // Track x position changes. When x prop changes, update state to trigger CSS transition.
+    // State updates ensure smooth transitions by forcing a re-render that applies the new position.
+    setX(_x)
+  }, [x, _x])
```
raineorshine, line 102: `x` is no longer a dependency, so it can be removed.

src/components/TreeNode.tsx
```diff
@@ -81,14 +84,29 @@ const TreeNode = ({
...
+
+  useLayoutEffect(() => {
+    // When y changes React re-renders the component with the new value of y. It will result in a visual change in the DOM.
+    // Because this is a state-driven change, React applies the updated value to the DOM, which causes the browser to recognize that
+    // a CSS property has changed, thereby triggering the CSS transition.
+    // Without this additional render, updates get batched and subsequent CSS transitions may not work properly. For example, when moving a thought down, it would not animate.
+    setY(_y)
   }, [y, _y])
```
raineorshine, line 110: `y` is no longer a dependency, so it can be removed.

src/components/LayoutTree.tsx
```diff
@@ -111,6 +114,31 @@ const useHideSpaceAbove = (spaceAbove: number) => {
   )
 }
 
+/** Calculate if the cursor is greater than the previous sibling at LayoutTree level. */
+const useCursorIsGreaterThanPrevSibling = (treeThoughtsPositioned: TreeThoughtPositioned[]) => {
+  // Calculate isGreaterThought at LayoutTree level
+  const cursorIsGreaterThanPrevSibling = useSelector(state => {
+    try {
```
raineorshine, line 121: It's bad practice to use try-catch for control flow. Very bad performance implications.

src/components/LayoutTree.tsx
```diff
@@ -111,6 +114,31 @@ const useHideSpaceAbove = (spaceAbove: number) => {
   )
 }
 
+/** Calculate if the cursor is greater than the previous sibling at LayoutTree level. */
+const useCursorIsGreaterThanPrevSibling = (treeThoughtsPositioned: TreeThoughtPositioned[]) => {
```
raineorshine, line 118: There are few issues with the design of `useCursorIsGreaterThanPrevSibling`:

1. `PrevSibling` is a misnomer, as `swapParent` does not deal with siblings. It swaps a child with its parent.
2. Since there is no other logic than a Redux selector, this should just be defined as a selector, not a custom hook.
3. `TreeThoughtsPositioned[]` are part of Phase II of the layout algorithm, where the linearized tree is rendered from top to bottom with x, y coordinates. This isn't a good fit for a pure Redux selector that doesn't care about thought positions. Instead you should write a pure Redux selector and access `state.cursor` and its parent.
4. Hooks/selectors should generally be defined where they are used, not higher up in the hierarchy only to be passed down through props. The whole point of `useSelector` is to avoid prop drilling. In this case, if this was done for performance reasons to avoid doing the calculation for every `TreeNode`, I would still argue that it's in the wrong place. The `LayoutTree` should not have to know about a specific condition for a specific node that is not related to layout. Instead, the `TreeNode` itself should calculate this with a simple check if it is the cursor thought. That will ensure the logic is not executed for all nodes.
5. `cursorIsGreaterThanPrevSibling` is a very verbose prop that doesn't make a lot of sense conceptually as part of the `TreeNode` interface. We have to think more about whether a prop actually makes sense given the design of the system and the distribution of responsibilities of the parts. This prop is too narrowly motivated by functional necessity without consideration for overall design.

Hope that makes sense! These are important design decisions that have a big impact on the structure of the code. I encourage you to reflect deeply on these principles when approaching React/Redux architecture.

raineorshine (changes requested):
Thanks so much! I'm happy to report that the behavior is working perfectly in nearly all cases!

# Table view in root

There is a regression when table view is activated in the root.

Note that the exact expected position is unknown, but col1 needs to be rendered in roughly the correct place, to the left of col2.

## Steps to Reproduce

```
- =view
  - Table
- a
  - b
- c
  - d
```

## Current Behavior

Col1 appears on the right edge of the screen.

Screenshot 2025-04-30 at 2 39 08 PM


## Expected Behavior

There is an issue with the spacing in `main`, but col1 is more-or-less in the correct place.

Screenshot 2025-04-30 at 2 40 03 PM

src/components/LayoutTree.tsx
zukilover: I've removed the `try..catch` block, thanks for the insight.

src/components/LayoutTree.tsx
zukilover: Thanks for the thoughtful feedback. I’ve made the changes based on your suggestions:

    - Renamed it to isCursorGreaterThanParent to better reflect the actual behavior.
    - Moved the logic into a plain Redux selector instead of using a custom hook.
    - Removed layout-specific logic involved.
    - The check now happens inside TreeNode using useSelector, so no need to pass props down.

Really appreciate the design pointers, especially around keeping responsibilities clean and avoiding tight coupling between layout and logic. I’ll keep that in mind going forward. Let me know if there’s anything I missed.

raineorshine:
I had to do a little more work to get the animation to trigger on undo/redo. There was not enough information for the Redux selector to detect this, so I added `state.lastUndoableActionType`.

See: a99fccbba3d

Issue #2768, opened 2025-01-11
Sort error after edit


raineorshine:
## Steps to Reproduce

```
- =sort
  - Alphabetical
    - Asc
- 1
- 2
```

1. Edit `1` to `3`. 
2. Toggle sort three times so that it clears the error and returns to sort/Alphabetical/Asc. 

## Current Behavior

1. The context appears to be sorted correctly, but sort icon indicates an error.

2. The sort order is incorrect:

```
- =sort
  - Alphabetical
    - Asc
- 3
- 2
``` 

## Expected Behavior

1. No error. 
2. The thoughts should be sorted in ascending order:

```
- =sort
  - Alphabetical
    - Asc
- 2
- 3
``` 

## Testing

Add an appropriate test for each step (two tests total).

PR #2811, opened 2025-02-08
fix sort error after edit

RED-ROSE515:
This PR resolves the issue : Sort error after edit #2768

This error is caused because there was an issue in storing the last none-empty value.

So, in this PR, I have fixed the thoughtNew object to reset the sortValue to undefined when newValue is non-empty.
Like this.
`...(newValue ? { sortValue: undefined } : { sortValue: oldValue || editedThought.sortValue }),`

### DIFF
--- a/src/actions/editThought.ts
+++ b/src/actions/editThought.ts
@@ -144,7 +144,7 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
     value: newValue,
     // store the last non-empty value to preserve the sort order of thoughts edited to empty
     // reset to undefined when newValue is non-empty
-    ...(newValue ? undefined : { sortValue: oldValue || editedThought.sortValue }),
+    ...(newValue ? { sortValue: undefined } : { sortValue: oldValue || editedThought.sortValue }),
     lastUpdated: timestamp(),
     updatedBy: clientId,
   }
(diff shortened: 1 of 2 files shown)


src/actions/editThought.ts
```diff
@@ -144,7 +144,7 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
     value: newValue,
     // store the last non-empty value to preserve the sort order of thoughts edited to empty
     // reset to undefined when newValue is non-empty
-    ...(newValue ? undefined : { sortValue: oldValue || editedThought.sortValue }),
+    ...(newValue ? { sortValue: undefined } : { sortValue: oldValue || editedThought.sortValue }),
```
raineorshine, line 147: This can be simplified to:

```ts
sortValue: newValue ? undefined : oldValue || editedThought.sortValue
```

src/commands/__tests__/toggleSort.ts
```diff
@@ -827,6 +827,54 @@ describe('store', () => {
   - a
   - b`)
     })
+
+    it('preserve sort order after multiple sort', async () => {
```
raineorshine, line 831: For some reason this test passes even without the fix. This means that it doesn't actually cover the test case.

raineorshine (changes requested):
Thank you! I can confirm that it is now working when I manually test it.

raineorshine (approved):
Thanks, confirmed!

Issue #2773, opened 2025-01-15
GestureDiagram: Rounded gestures


raineorshine:
- Add an opt-in `rounded?: boolean` property to the `GestureDiagram` component and the `Command` type.
- Add SVG support in the GestureDiagram for rounded gestures. **This should be a general solution**, such that rounded gestures could be enabled for any command by simply setting `rounded: true`. Each segment will need to be replaced with an appropriate arc. Do not make special cases for each command. 
  - However, it's okay if the general solution only works for swipes that continue in the same direction, like the below commands. You can ignore longer gestures or gestures that change directions.
- Enable rounded gestures for the following commands:
  - Sort - `toggleSort`
  - Previous Thought - `cursorPrev`
  - New Thought (above) - `newThoughtAbove`
  - New Subthought (next) `newUncle`
- Ensure size and positioning are correct at all font sizes.
- Add the above four gesture diagrams to [TestGestureDiagram](https://github.com/cybersemics/em/blob/f87b26310413ea5c66d7e360cbc0b1c9900d8ace/src/components/modals/TestGestureDiagram.tsx) so that the rounded gestures are under test coverage.

<table>
<thead><tr><th>Old</th><th>New</th></tr></thead>
<tbody>
<tr>
<td>
<img width="398" alt="Screenshot 2025-01-15 at 9 30 29 PM" src="https://github.com/user-attachments/assets/93c25c62-b07a-47fc-8eb9-4f19cd34d614" /></td>
<td><img width="40" alt="Pixelmator Pro 2025-01-15 21 59 35" src="https://github.com/user-attachments/assets/ea49c464-b503-42fc-98d7-673bca9f8db3" /></td>
</tr>
<tr>
<td><img width="398" alt="Screenshot 2025-01-15 at 9 29 06 PM" src="https://github.com/user-attachments/assets/8a0e8aeb-68e4-4ca6-b7d9-532cde1dd5ab" /></td>
<td><img width="39" alt="Pixelmator Pro 2025-01-15 21 59 54" src="https://github.com/user-attachments/assets/465e4dd7-2d9e-41c1-a91f-0656fb2843a7" /></td>
</tr>
<tr>
<td><img width="398" alt="Screenshot 2025-01-15 at 9 28 57 PM" src="https://github.com/user-attachments/assets/42c2c945-bb9a-435f-ab4d-ca8eb27109d9" /></td>
[...]

PR #2817, opened 2025-02-14
Rounded gestures

RED-ROSE515:
This PR resolves the issue : GestureDiagram: Rounded gestures #2773

So, in this PR, I have implemented Rounded gestures.

In this PR, I removed a special case for "rdl" and "ldr". I think I need to check more about that.
I made a `generateArcPath` function to generate the arc path segment of given direction and index.
So, I generate a quadratic curves for a one length path and generate circular arcs for others.

This is the step.
**Clockwise Check**: Determines directionality based on the first two path directions.
**Base Angle**: Starting angle derived from the initial direction (e.g., u followed by r starts at 0°).
**Angle Calculation**:
Total angle = (**number of segments** - 1) * 90°.
Each segment covers **totalAngle** / **number of segments**.
**Arc Generation**: Uses SVG's `A` command to draw arcs around a center, with segments split by startAngle and endAngle.


raineorshine (changes requested):
Thanks! This looks great, and the segment highlighting works perfectly.

One thing I'm seeing is that the arrow head should start where the arc ends, extending further than the arc. This will allow the complete arc line to be visible and help it not look scrunched at the end:

image

Overall, great organization and commenting. Thank you for your time!

src/components/GestureDiagram.tsx
```diff
@@ -253,7 +320,15 @@ const GestureDiagram = ({
                     : i === 2
                       ? 'M 54,40.5 Q 45,49.5 45,58.5'
                       : 'M 45,58.5 L 45,72'
-                : `M ${x} ${y} l ${segment.dx} ${segment.dy}`
+                : rounded
+                  ? generateArcPath(
+                      { x, y },
+                      nextPos,
+                      Array.from(path as string)[i] as Direction,
```
raineorshine, line 327: Can this be simplified to `path[i] as Direction`?

src/components/GestureDiagram.tsx
```diff
@@ -253,7 +320,15 @@ const GestureDiagram = ({
...
+                      i,
+                      Array.from(path as string) as Direction[],
```
raineorshine, line 329: Can this be simplified to `path as Direction[]`?

raineorshine (changes requested):
Looks great, thank you! Just a couple questions.

raineorshine (approved):
Thank you!

Issue #2795, opened 2025-01-28
Apply thought color to superscript


raineorshine:
## Steps to Reproduce

```
- a
- a
- hello world
- hello world
```

1. Set the color of the first `a` to blue.
2. Set the color of the first `world` to red.

## Current Behavior

The superscript of the first `a` is white.

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/bdbf864e-c1a2-474c-a551-61e1125c10e1" />

## Expected Behavior

The superscript of the first `a` should be blue because the color of the entire thought is blue.

This should not apply to the superscript after `world` because only part of `hello world` is colored. This is the same logic that is used to color the bullet.

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/6d82eb2a-bb08-44b4-a097-5b3fbeb65a2b" />

PR #2819, opened 2025-02-19
Apply thought color to superscript

RED-ROSE515:
This PR resolves the issue : Apply thought color to superscript #2795

So, in this PR, I checked if the thought is wrapped by single tag.
If yes, I set the color of superscript. If not, don't apply.

I created a util file and moved `isWrappedInSingleTag` and `getElementFill` function to that to avoid the duplication.
And used `getElementFill` function in `Bullet.ts` and `StaticSuperScript.ts`.
Get the color from `getElementFill` function and set the color of superscript with that.


src/components/StaticSuperscript.tsx
```diff
@@ -12,47 +15,53 @@ const StaticSuperscript = React.forwardRef<
...
-      cssRaw,
-    )}
-    style={style}
-  >
-    {show && (
-      <span
-        className={css({
-          fontSize: '60%',
+>(({ n, style, show = true, hideZero, absolute, cssRaw, thoughtId }, forwardRef) => {
+  const fill = useSelector(state => getElementFill(state, thoughtId!))
```
raineorshine, line 21: `!` overrides the type system and should only be used in rare cases where the truthiness of a value is guaranteed for reasons that the type system cannot capture.

Instead, check `thoughtId` manually or change `getElementFill` to take undefined.

raineorshine (commented):
Hi. I didn't test the behavior since this is still a Draft, but here are a couple advance comments on the code. Thanks!

RED-ROSE515:
Hi @raineorshine , I think current failed puppeteer test is not regression.
Please check and let me know.

raineorshine:
Thanks! Before I review, please add the appropriate tests to https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/__tests__/color.ts.

> I think current failed puppeteer test is not regression.

Yes, so sorry about that! The puppeteer tests have been very flaky on GitHub Actions recently. The current failure is not related to this PR.

src/e2e/puppeteer/__tests__/color.ts
```diff
@@ -163,3 +164,30 @@ it('Empty <span> element will be removed after setting color to default.', async
   const result = await getEditingText()
   expect(result).toBe('Golden Retriever')
 })
+
+it('Set the superScript color of the text', async () => {
```
raineorshine, line 168: This test title doesn't makes sense. Maybe something like "Set the superscript color to the color of the whole thought".

src/e2e/puppeteer/__tests__/color.ts
```diff
@@ -163,3 +164,30 @@ it('Empty <span> element will be removed after setting color to default.', async
...
+  await paste(importText)
+
+  await waitForEditable('hello world')
+  await clickThought('hello world')
+  await setSelection(6, 11)
+  await click('[data-testid="toolbar-icon"][aria-label="Text Color"]')
+  await click('[aria-label="text color swatches"] [aria-label="red"]')
+
+  const supColor2 = await getSuperScriptColor()
+  expect(supColor2).toBe(null)
```
raineorshine, line 184: This test could use a couple comments that explains the two assertions, one positive and one negative case. I'm okay having these two cases in a single test for efficiency, just needs a little explanation.

src/selectors/getElementFill.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/**
+ * Gets the fill color from a thought's value by checking style and font color attributes.
+ * @param state - The Redux state.
+ * @param thoughtId - The ID of the thought to check.
+ * @returns The fill color or undefined if not found.
+ */
+const getElementFill = (state: State, thoughtId: ThoughtId | null): string | undefined => {
+  if (!thoughtId) return undefined
```
raineorshine, line 70: I tend to prefer using `null` rather than `undefined` when dealing with an explicit return case meaning "no value". Then `undefined` retains its more intuitive semantics of meaning "not defined", such as a value missing from a map or a variable that has not been defined.

src/selectors/getElementFill.ts
```diff
@@ -0,0 +1,90 @@
...
+  return tagStack.length === 1 && (tagStack[0] === 'font' || tagStack[0] === 'span')
+}
+
+/**
+ * Gets the fill color from a thought's value by checking style and font color attributes.
+ * @param state - The Redux state.
+ * @param thoughtId - The ID of the thought to check.
+ * @returns The fill color or undefined if not found.
+ */
+const getElementFill = (state: State, thoughtId: ThoughtId | null): string | undefined => {
```
raineorshine, line 69: This function would probably be scoped a little better if it took `value: string` directly since that is what its logic is really about. That would eliminate the `State` dependency and allow it to be a more manageable pure `util` function.

src/components/ThoughtAnnotation.tsx
```diff
@@ -229,7 +229,9 @@ const ThoughtAnnotation = React.memo(
           {email && <EmailIconLink email={email} />}
           {
             // with real time context update we increase context length by 1 // with the default minContexts of 2, do not count the whole thought
-            showSuperscript ? <StaticSuperscript absolute n={numContexts} style={style} cssRaw={cssRaw} /> : null
+            showSuperscript ? (
+              <StaticSuperscript absolute n={numContexts} style={style} cssRaw={cssRaw} thoughtId={head(simplePath)} />
```
raineorshine, line 233: `ThoughtId` is unique in the thoughtspace, but unfortunately not unique on the screen since the same thought can be rendered multiple times when the context view is open.

The trick is that in the context view you need to use `parentOf(path)` rather than `path`.

## Steps to Reproduce

```
- a
  - m
    - x
- v
	- b
	  - m
	    - y
- c
  - b
```

1. Set the color of `v` to red.
2. Set the color of `v/b` to green.
3. Set the cursor on `a/m`.
4. Activate Context View.

## Current Behavior

The superscript of `a/m~/b` is red.

## Expected Behavior

The superscript of `a/m~/b` should be green like the thought.

## Testing

We should add a test for this special case.

raineorshine (changes requested):
Thank you! I have tested this manually and the behavior generally works great.

Normal view works as expected, but there is a slight change needed for it to work in context view (see below).

src/selectors/getElementFill.ts
RED-ROSE515: I tried to use null rather than undefined.
But fill color cannot be null.

raineorshine: Okay, that’s fair.

src/util/rgbHex.ts
```diff
@@ -0,0 +1,62 @@
...
+  const toHex = (value: number) => Math.min(Math.max(value, 0), 255).toString(16).padStart(2, '0')
+
+  if (a >= 1) {
+    return `#${toHex(r)}${toHex(g)}${toHex(b)}` // RGB hex only
+  } else {
+    const alphaHex = Math.round(a * 255)
+      .toString(16)
+      .padStart(2, '0')
+    return `#${toHex(r)}${toHex(g)}${toHex(b)}${alphaHex}` // RGBA hex
+  }
```
raineorshine, line 30: A slight improvement is to move the alpha condition into `alphaHex` itself. This will avoid duplicating the return statement and string interpolation.

```ts
const alphaHex = a >= 1 ? Math.round(a * 255)
      .toString(16)
      .padStart(2, '0') : ''
return `#${toHex(r)}${toHex(g)}${toHex(b)}${alphaHex}`
```

src/selectors/getThoughtFill.ts
raineorshine, line 1: This file can be removed, per the comment above.

raineorshine (changes requested):

src/components/Bullet.tsx
```diff
@@ -441,76 +440,7 @@ const Bullet = ({
...
-      if (styleMatch) return styleMatch[2]
-
-      // If no background-color, check for font color
-      const fontColorMatch = value.match(fontColorRegex)
-      if (fontColorMatch) return fontColorMatch[1]
-    }
-
-    return undefined
-  })
+  const fill = useSelector(state => getThoughtFill(state, thoughtId!))
```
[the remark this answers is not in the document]
RED-ROSE515: @raineorshine , getThoughtFill function is used in Bullet and superScript.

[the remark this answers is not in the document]
raineorshine: Sorry, I must have missed the other occurrence. Thanks!

src/e2e/puppeteer/__tests__/color.ts
```diff
@@ -186,3 +193,61 @@ it('remove all formatting from the thought', async () => {
...
+      - b
+        - m
+          - y
+    - c
+      - b
+    `
+  await paste(importText1)
+
+  // Test 1: Verify that partial text coloring doesn't affect superscript
+  await waitForEditable('hello world')
```
raineorshine, line 216: Note that `clickThought` already has `waitForEditable` built in.

raineorshine (approved):
Thank you for addressing all the requested changes. Looks good!

src/util/rgbHex.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * Converts RGBA color values to a hexadecimal color string.
+ * @returns Hexadecimal color string in the format #RRGGBBAA.
+ */
+export const rgbaToHex = (rgba: string): string => {
+  // Extract the RGBA components from the string using a regular expression
+  const rgbaRegex = /^rgba?\((\d+),\s*(\d+),\s*(\d+),\s*([01]?\.?\d*)\)$/
```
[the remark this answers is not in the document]
raineorshine: @RED-ROSE515 I found a couple more. Moved in dfd07c3fc8a.

src/e2e/puppeteer/__tests__/color.ts
RED-ROSE515: I got it.

Issue #2767, opened 2025-01-11
Cursor Up: Cursor moves to incorrect thought after New Thought Above


raineorshine:
## Steps to Reproduce

```
- a
  - b
    - =pin
    - b1
  - c
```

1. Set the cursor on `b1`
2. New Thought Above (Shift + Enter)
3. Set the cursor on `c`
4. Cursor Up (ArrowUp)

## Current Behavior

The cursor moves to the new thought.

## Expected Behavior

The cursor should move to `b1`.

Cursor Up should always move to the visible thought immediately above the cursor thought, regardless of hierarchy.

## Testing

Add a test to https://github.com/cybersemics/em/blob/128153531a41c49eb19daf7eb784a40f768a1020/src/actions/__tests__/cursorUp.ts.

PR #2820, opened 2025-02-20
Fix cursorUp

RED-ROSE515:
This PR resolves the issue : https://github.com/cybersemics/em/issues/2767

The cause of bug : When gets the last visible descendant of a thought, we used `getChildren` to get the children of thought.

So, in this PR, I have changed from `getChildren` to `getChildrenSorted` to get the children of thought.


raineorshine (approved):
Perfect, thanks!

PR #2833, opened 2025-03-04
Help Modal v1.5 Tabs

yangchristina:
Add tab component, and use in Help Modal.

Will update Command Library and Tutorial content in separate PRs.

https://github.com/user-attachments/assets/05a92927-877b-4393-ac18-f7ea2f3bd1da


src/components/Tab/Tab.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { Children, isValidElement } from 'react'
```
trevinhofmann, line 1: `Children` is new to me. Cool!

The [React docs](https://18.react.dev/reference/react/Children) do warn about using it, though:
>Using Children is uncommon and can lead to fragile code. [See common alternatives.](https://18.react.dev/reference/react/Children#alternatives)

src/components/modals/Help.tsx
```diff
@@ -382,48 +325,30 @@ const About = () => {
...
+        <TabItem value={Section.Tutorials} label='Tutorials' showDot={showDot}>
+          <Tutorials />
+        </TabItem>
+        <TabItem value={Section.Metaprogramming} label='Meta'>
+          <Metaprogramming />
+        </TabItem>
+        <TabItem value={Section.About} label='About'>
+          <About />
+        </TabItem>
+      </Tab>
```
trevinhofmann, line 351: I'm not _necessarily_ opposed to this implementation, but an alternative to using `Children` in the `<Tab>` component would be passing the tabs as a prop and rendering them there, something like:

```ts
  const tabs = [
    {
      value: Section.CommandLibrary,
      label: 'Command Library',
      children: <CommandCenter />,
    },
    {
      value: Section.Tutorials,
      label: 'Tutorials',
      showDot: showDot,
      children: <Tutorials />,
    },
    {
      value: Section.Metaprogramming,
      label: 'Meta',
      children: <Metaprogramming />,
    },
    {
      value: Section.About,
      label: 'About',
      children: <About />,
    },
  ]
  ```
  
  ```ts
  <Tab currentTab={section} onTabChange={setSection} tabs={tabs} />
  ```

Then, we could use `tabs.map` to render the `<TabItem>` components within `<Tab>`.

That would avoid needing to use `React.Chilren`, give us better type control (since we can type the `tabs` prop), and make the API of `<Tab>` clearer to anyone using it.

File: src/components/Tab/index.ts [hunk 3]
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/prefer-default-export */
+
+export { default as Tab } from './Tab'
+export { default as TabItem } from './TabItem'
```
trevinhofmann, line 4: Perhaps better names for these would be `Tabs` and `Tab`?

trevinhofmann, line 4: If we refactor `Tab` to accept a typed `tabs` prop rather than using `children`/`React.Children`, consumers of this would not need to import `TabItem` (only `Tab` would import `TabItem`), so we could simplify the export and not need to disable the ESLint `import/prefer-default-export` rule here.

trevinhofmann (changes requested):
Thanks for the PR, @yangchristina. The tabs look great on desktop and mobile! Just a few questions and notes for @yangchristina and @raineorshine.

---

## ~~Q1. Missing blue dot~~

The Figma design shows a blue dot on one of the tabs, though I'm not sure what it's meant to indicate. Is that something we should be including here?

<details>
<summary>Figma screenshot</summary>

Screenshot 2025-03-10 at 1 52 49 PM
</details>

**Edit: Ah! When reviewing the code, I see that we include the blue dot on the Tutorials tab when the tutorial has not yet been completed. This appears to be functioning correctly 👍**

## Q2. Hover indication

Should there be some indication when hovering over a tab? Usually, clickable elements will have a subtle color change, highlight, etc. to make it clear that they can be clicked. Currently, there is none for the tabs.

## Q3. Layout shift from bold tab

It's only a minor issue, but making the selected tab bold is causing the layout to shift a bit when switching between tabs. Would it be worthwhile to ensure the tabs maintain their same width when switching, so that the layout is not affected?
<details>
<summary>Shifting layout demo</summary>

https://github.com/user-attachments/assets/5247d26f-1386-421d-96bf-2f7287d39035
</details>

## Q4. Haptics

Should there be haptic feedback when clicking and/or scrolling the tabs, or would that be outside the scope of this PR?

src/components/Tab/Tab.tsx
trevinhofmann: I've suggested a possible alternative here: https://github.com/cybersemics/em/pull/2833#discussion_r1987883352

raineorshine:
2. Yes, hover color change on desktop would be great. 
3. Agreed, layout should be stable when switching tabs. I have used `-webkit-text-stroke` instead of bold, which does not change the element's width, but I'm open to other ideas. 
4. Yes, let's do a light haptic feedback on click, as it's just a single line of code. Scroll haptics are out of a scope but I will consider it in the future.

yangchristina:
> 4\. haptic feedback

@raineorshine Is haptic feedback calling `window.navigator.vibrate(500);` on tab click?

yangchristina:
@trevinhofmann 
2+3. Hover indication (decrease brightness using filter) + Fix tab bold layout issue using `WebkitTextStrokeWidth: active ? '0.05em' : 0,`

Is there another hover indicator you. had in mind @raineorshine ? Didn't go with background color change due to how I am handling the bottom border of the tab.

https://github.com/user-attachments/assets/e57cc6dc-4227-4562-b2ff-3f190590f655

3. Haptic feedback:
```
onClick={() => {
                onTabChange(value)
                window.navigator.vibrate(100)
              }}
```
Not sure if this is the appropriate amount for light haptic feedback

src/components/Tab/Tab.tsx
yangchristina: I've implemented this, thanks!

src/components/Tab/index.ts
yangchristina: Removed this import page.

trevinhofmann:
@yangchristina - We recently added a haptics utility here: https://github.com/cybersemics/em/blob/main/src/util/haptics.ts

You can use this line as a reference: https://github.com/cybersemics/em/blob/f4107595b9187803d5b02d234ee4a9d2021cec38/src/actions/dragInProgress.ts#L116

The `fastClick` utility also now has built-in support for haptics: https://github.com/cybersemics/em/blob/main/src/util/fastClick.ts

trevinhofmann (changes requested):
Thank you for the updates! Just waiting for the haptics implementation before reviewing this again.

raineorshine:
> ### 2+3. Hover indication (decrease brightness using filter) + Fix tab bold layout issue using `WebkitTextStrokeWidth: active ? '0.05em' : 0,`
> Is there another hover indicator you. had in mind @raineorshine ? Didn't go with background color change due to how I am handling the bottom border of the tab.

I think that's fine. Thanks!

yangchristina:
> @yangchristina - We recently added a haptics utility here: https://github.com/cybersemics/em/blob/main/src/util/haptics.ts
> 
> You can use this line as a reference:
> 
> https://github.com/cybersemics/em/blob/f4107595b9187803d5b02d234ee4a9d2021cec38/src/actions/dragInProgress.ts#L116
> 
> The `fastClick` utility also now has built-in support for haptics: https://github.com/cybersemics/em/blob/main/src/util/fastClick.ts

Thank you! I tried `fastClick`, but it didn't work, so I went back with onClick, and added haptic.light()

With `fastClick`, not sure if I'm doing something wrong. did {...fastClick(()=>{ ...click handler here..() })}
https://github.com/user-attachments/assets/0e3aace0-4376-4db3-b77f-376a469fb311

src/components/Tabs.tsx
```diff
@@ -0,0 +1,141 @@
...
+  children,
+  showDot = false,
+  active = false,
+  onClick,
+}: {
+  onClick?: () => void
+  active?: boolean
+} & TabDefinition<T>) => {
+  return (
+    <>
```
trevinhofmann, line 30: We should be able to remove this `<>` / `React.Fragment` wrapper since it only has one child (the `<div>`).

src/components/Tabs.tsx
```diff
@@ -0,0 +1,141 @@
...
+              value={value}
+              label={label}
+              showDot={showDot}
+              active={value === currentTab}
+              onClick={() => {
+                onTabChange(value)
+                haptics.light()
+              }}
+            >
+              {children}
```
trevinhofmann, line 131: It looks like `children` is unused in the `<Tab>` component. Can we remove it?

src/components/Tabs.tsx
```diff
@@ -0,0 +1,141 @@
...
+            </Tab>
+          ))}
+        </div>
+      </div>
+      <div>{tabs.find(tab => tab.value === currentTab)?.children}</div>
```
trevinhofmann, line 136: Also, using `children` here is okay, but I think it would be more appropriate to use a name like `content` since this is the content rendered by the tab rather than being "children" of the tab element itself. I'd also just generally prefer reserving `children` for `props.children` to avoid confusion.

trevinhofmann (changes requested):
Thank you for the updates, @yangchristina! The haptics are working for me, and the bold layout shift is resolved 👍

## 5: Hover indication not visible in Light Mode

With the Light Mode enabled, the hover indication is not visible (or is too minimal to be discerned).

https://github.com/user-attachments/assets/f5101718-1657-4b36-8c11-d0c257f598a1

yangchristina:
> Thank you for the updates, @yangchristina! The haptics are working for me, and the bold layout shift is resolved 👍
> 
> ## 5: Hover indication not visible in Light Mode
> With the Light Mode enabled, the hover indication is not visible (or is too minimal to be discerned).
> 
>  Screen.Recording.2025-03-15.at.2.11.52.PM.mov

Thank you for the review! Switched to using contrast instead of brightness for hover effects, so it works on both light and dark:

https://github.com/user-attachments/assets/493f8ff0-d8fd-4649-a416-c5ce6da11db5


https://github.com/user-attachments/assets/fe6ad49d-f7b2-41c9-9576-4be342f42c9f

src/components/Tabs.tsx
yangchristina: Yes I definitely should have renamed it after moving it into an array. Fixed now, thank you!

src/components/Tabs.tsx
yangchristina: Removed

src/components/Tabs.tsx
yangchristina: Removed, thanks!

src/components/Tabs.tsx
```diff
@@ -0,0 +1,135 @@
+import { css } from '../../styled-system/css'
+import haptics from '../util/haptics'
+
+/**
+ * Represents a single tab's configuration.
+ */
+export interface TabDefinition<T extends string> {
```
trevinhofmann, line 7: @raineorshine -- I know there's a general preference for putting types inside `src/@types`, though we also have some local types in cases where the type is unlikely to be reused (much) in other parts of the project. I'm on the fence with this one, as it does get imported into the `Help` component. Thoughts in this case, for `TabDefinition`?

raineorshine: I have a slight preference for `src/@types` so that we stick to our convention, but I agree it's debatable.

yangchristina: Moved to src/@types

src/components/Tabs.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * A container component that renders a row of scrollable tabs and their content.
+ * Tabs can be scrolled horizontally on small screens.
+ */
+const Tabs = <T extends string>({
+  currentTab,
+  onTabChange,
+  tabs,
+  className = '',
```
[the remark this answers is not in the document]
yangchristina: Yes, removed

trevinhofmann (approved):
Looks good to me! Thank you, @yangchristina!

raineorshine (approved):

PR #2834, opened 2025-03-04
Whitelist 'inherit' and 'currentColor' in PandaCSS ESLint config

yangchristina:
No code changes in this PR. Only removing eslint disable comments + updating eslint config


yangchristina:
@raineorshine there's no behavioural changes, so the failure should be due to flaky tests

raineorshine:
Thank you!

PR #2843, opened 2025-03-10
Bring back browserstack2

mstrofbass:
Moves the status bar size param into `tap`
Removes debug pauses
Update JSDOC
Remove swipe
Remove unused test file


raineorshine:
@mstrofbass Hi Blake, hope you're doing okay. Wanted to check in and see if you are still interested in working on this. Get back to me as soon as you can. Thanks!

Issue #2169, opened 2024-07-18
Drag parent hover animation continues after canDrop returns false


raineorshine:
## Steps to Reproduce

```
- a
  - b
- c
```

1. Set the cursor on `a`. 
2. Drag `b` below `c` (as a child) and hold. 
3. Drag slightly up as if trying to drop `a` into `b` (which is not possible because a thought cannot be dropped into its descendant).

## Current Behavior

The DropHover correctly disappears when the thought can no longer be dropped, i.e. canDrop returns false, but `c` is still pulsing.  

https://github.com/user-attachments/assets/07ced001-7c16-4e9e-8aa2-3cf7849f17d3

## Expected Behavior

`c` should not be pulsing. 

The parent hover pulse animation should stop as soon as the drop target changes.

## Notes

Note that the bug only occurs when a contains a child. It seems to occur in the additional padding that is inserted at the cliff. So DropCliff may be implicated.

While isChildHovering may not have access to canDrop, there is another selector used in DropEnd that is just based on the Redux state: showDropHover. We only want the parent to pulse when showDropHover is true. So that may come in handy. https://github.com/cybersemics/em/blob/f010868894f21341c49bfcbf7e5a84e8b699c3dd/src/components/DropEnd.tsx#L57

PR #2849, opened 2025-03-16
Fix: Drag parent hover animation continues after canDrop returns false

karunkop:
### **Issue** #2169 
When dragging a thought, the visual feedback for valid drop targets was inconsistent. Specifically:
- If a user dragged thought "b" below thought "c" and then attempted to drag thought "a" into "b"
- If the drop became invalid (e.g., "a" had children), the `DropHover` component correctly disappeared
- However, thought "c" continued to pulse/highlight, even though it was no longer a valid drop target

### **Root Cause**
- The `hoveringPath` in Redux wasn't cleared when a thought became an invalid drop target.
- The state only cleared on physical drag leave events but didn't respond to logical state changes (i.e., when `canDropThought` changed to `false`).
- Components responsible for pulsing did not check the latest `canDropThought` state before continuing the animation.

#### **1. Enhanced `useDragLeave.ts`**
- Tracked `prevCanDropThought` to detect state changes.
- Cleared `hoveringPath` immediately when a drop target became invalid.

#### **2. Updated `Thought.tsx`**
- Adjusted `isChildHovering` selector to factor in `canDropThought`.
- Ensured pulsing only occurs when `canDropThought` is `true`.
- Refactored existing inline condition checks with descriptive comments.

#### **3. Improved `useDragAndDropSubThought.ts`**
- Updated `dropCollect` to return `canDropThought`, ensuring dependent components receive accurate drop validity.

#### **4. Updated `DropChild.tsx` and `DropEnd.tsx`**
- Integrated `useDragLeave` to clear hover states when `canDropThought` changes.
- Ensured components respond to both physical drag events and logical state changes.

### DIFF
--- a/src/components/DropChild.tsx
+++ b/src/components/DropChild.tsx
@@ -8,6 +8,7 @@ import SimplePath from '../@types/SimplePath'
 import { isTouch } from '../browser'
 import testFlags from '../e2e/testFlags'
 import useDragAndDropSubThought from '../hooks/useDragAndDropSubThought'
+import useDragLeave from '../hooks/useDragLeave'
 import useDropHoverColor from '../hooks/useDropHoverColor'
 import useHoveringPath from '../hooks/useHoveringPath'
 import { hasChildren } from '../selectors/getChildren'
@@ -32,8 +33,9 @@ const DropChild = ({ depth, path, simplePath, isLastVisible }: DropChildProps) =
   const value = useSelector(state => getThoughtById(state, head(simplePath))?.value || '')
   const dropHoverColor = useDropHoverColor(depth || 0)
 
-  const { isHovering, dropTarget } = useDragAndDropSubThought({ path, simplePath })
+  const { isHovering, isDeepHovering, canDropThought, dropTarget } = useDragAndDropSubThought({ path, simplePath })
   useHoveringPath(path, !!isHovering, DropThoughtZone.SubthoughtsDrop)
+  useDragLeave({ isDeepHovering, canDropThought })
 
   // Calculate the height for the child thought over cliff
   const dropTargetHeight = isLastVisible ? calculateCliffDropTargetHeight({ depth }) : 0
--- a/src/components/DropChild.tsx
+++ b/src/components/DropChild.tsx
@@ -2,15 +2,12 @@ import React from 'react'
 import { shallowEqual, useSelector } from 'react-redux'
 import { css, cx } from '../../styled-system/css'
 import { dropEndRecipe, dropHoverRecipe } from '../../styled-system/recipes'
-import DropThoughtZone from '../@types/DropThoughtZone'
 import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import { isTouch } from '../browser'
 import testFlags from '../e2e/testFlags'
 import useDragAndDropSubThought from '../hooks/useDragAndDropSubThought'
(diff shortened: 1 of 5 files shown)


src/components/DropChild.tsx
```diff
@@ -32,8 +33,9 @@ const DropChild = ({ depth, path, simplePath, isLastVisible }: DropChildProps) =
   const value = useSelector(state => getThoughtById(state, head(simplePath))?.value || '')
   const dropHoverColor = useDropHoverColor(depth || 0)
 
-  const { isHovering, dropTarget } = useDragAndDropSubThought({ path, simplePath })
+  const { isHovering, isDeepHovering, canDropThought, dropTarget } = useDragAndDropSubThought({ path, simplePath })
   useHoveringPath(path, !!isHovering, DropThoughtZone.SubthoughtsDrop)
+  useDragLeave({ isDeepHovering, canDropThought })
```
raineorshine, line 38: Could `useDragLeave` be encapsulated in `useDragAndDropSubThought`? That would avoid the need to expose `canDropThought`.

src/hooks/useDragLeave.ts
```diff
@@ -40,6 +41,16 @@ const useDragLeave = ({ isDeepHovering, canDropThought }: { isDeepHovering: bool
       return
     }
 
+    // If canDrop changed from true to false, immediately clear the hovering path
+    if (!canDropThought && prevCanDropRef.current) {
+      dispatch(clearHoveringPath)
+      prevCanDropRef.current = canDropThought
+      return
+    }
```
raineorshine, line 49: This condition doesn't quite make sense to me. Isn't it strictly narrower than the condition that comes immediately after? i.e. `clearHoveringPath` will always be dispatched when `canDropThought` is false.

raineorshine (changes requested):
Thank you! I can confirm that the reported issue is fixed. Good use of the existing hooks to solve the problem without adding much complexity.

The additional calls to `useDragLeave` could end up being a performance concern, so we will have to keep an eye on that.

src/hooks/useDragLeave.ts
karunkop: Yep i can verify that the existing condition is redudant :)

src/hooks/useDragAndDropSubThought.ts
```diff
@@ -162,19 +165,26 @@ const dropCollect = (monitor: DropTargetMonitor) => ({
...
-  const [{ isHovering, isBeingHoveredOver, isDeepHovering }, dropTarget] = useDrop({
+  const [{ isHovering, isBeingHoveredOver, isDeepHovering, canDropThought }, dropTarget] = useDrop({
     accept: [DragAndDropType.Thought, NativeTypes.FILE],
     canDrop: (item, monitor) => canDrop(propsTypes, monitor),
     drop: (item, monitor) => drop(propsTypes, monitor),
     collect: dropCollect,
   })
 
+  // Encapsulate the useHoveringPath and useDragLeave hooks
+  if (props.path) {
```
raineorshine: The `Partial<DroppableSubthoughts>` definition was an odd way to define the parameters, so I fixed it in 3e1f3db06fc. That obviates the need for this condition.

FYI Hooks are not allowed in conditions. 
See: https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

raineorshine (approved):
This looks great! Solves the problem nicely. Thank you.

PR #2850, opened 2025-03-18
Feature: Add new gesture cheatsheet to mobile devices

fbmcipher:
This pull request:

* Adds a "gesture cheatsheet" – a modal where users can see all of the gestures in **em** in one place.
* The gesture cheatsheet supports search – users can search for any command to see its gesture.
* Commands can either be seen in their groups, or sorted alphabetically.
* The gesture cheatsheet can be triggered by drawing the ? (`rdld`), which replaces the existing Help command.
    * Special hardcoded cases in the codebase that mapped this gesture to `help` have been replaced.
 * A focus on polished UI design, including subtle animations.

## Other improvements
* The same searching and sorting abilities in the Gesture Cheatsheet have also been added to the CommandTable – like in Ahmad's UX designs.


src/components/CommandTable.tsx
```diff
@@ -116,108 +118,141 @@ if (commandsUngrouped.length > 0) {
...
-            // do not render groups with no commands on this platform
-            return commands.length > 0 ? (
+      {/* Smooth Fade-in Transition */}
+      <div
+        className={css({
+          textAlign: 'left',
+          opacity: isFading ? 0 : 1,
+          transition: 'opacity 0.3s ease-in-out',
+        })}
+      >
```
raineorshine, line 208: Would it be possible to use the existing [FadeTransition](https://github.com/cybersemics/em/blob/37f3e98916904810830c18b92436cad2016fac47/src/components/FadeTransition.tsx) component? This helps abstract away the timing and animation details.

Note that all of our animation durations are defined in https://github.com/cybersemics/em/blob/37f3e98916904810830c18b92436cad2016fac47/src/durations.config.ts for consistency.

src/components/SortButton.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  const dropdownStyles = css({
+    opacity: 0,
+    visibility: 'hidden',
+    transition: 'opacity 0.3s ease-in-out, visibility 0.3s ease-in-out',
+    willChange: 'opacity, visibility',
+    position: 'absolute',
+    top: '100%',
+    right: 0,
+    backgroundColor: '{colors.bg}',
```
raineorshine, line 23: This can just be `backgroundColor: 'bg'`, as PandaCSS knows the named colors from our config. 

The braces are only needed for interpolation.

src/components/SortButton.tsx
```diff
@@ -0,0 +1,151 @@
...
+        backgroundColor: '{colors.darkgray}',
+        borderRadius: '8px',
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        position: 'relative',
+      })}
+    >
+      <SortIcon size={20} fill={isLightTheme ? '{colors.light}' : '{colors.fg}'} />
+      <div className={`${dropdownStyles} ${isDropdownOpen ? dropdownVisibleStyles : ''}`}>
```
raineorshine, line 65: Use `css` or `cx` to merge Panda styles.

See: https://panda-css.com/docs/concepts/merging-styles

src/components/SortButton.tsx
```diff
@@ -0,0 +1,151 @@
...
+    setDropdownOpen(false)
+  }
+
+  return (
+    <button
+      onClick={() => setDropdownOpen(!isDropdownOpen)}
+      className={css({
+        width: '45px',
+        border: 'solid 1px {colors.gray50}',
+        backgroundColor: '{colors.darkgray}',
```
raineorshine, line 56: `backgroundColor: 'darkgray'`

src/components/CommandGrid.tsx
```diff
@@ -0,0 +1,264 @@
...
+            position: 'absolute',
+            left: '10px',
+            top: '50%',
+            transform: 'translateY(-50%)',
+            display: 'flex',
+            alignItems: 'center',
+            pointerEvents: 'none',
+          })}
+        >
+          <SearchIcon size={16} fill={isLightTheme ? '#666' : '#999'} />
```
raineorshine, line 138: You can define colors in https://github.com/cybersemics/em/blob/37f3e98916904810830c18b92436cad2016fac47/src/colors.config.ts. PandaCSS will automatically render the correct one for the theme.

src/components/CommandGrid.tsx
raineorshine, line 1: Large amount of duplication between `CommandGrid` and `CommandTable`.

src/components/CommandGridItem.tsx
raineorshine, line 1: Large amount of duplication between `CommandGridItem` and `CommandTable`.

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return () => {
+      document.body.style.overflow = 'auto'
+    }
+  }, [isOpen])
+
+  useEffect(() => {
+    if (!bottomRef.current || !isOpen) return
+
+    const observer = new IntersectionObserver(
```
raineorshine, line 53: A simpler solution is to create 100px of padding on the bottom below all the gestures. The gradient won't be visible over the empty space. 

This would create a better user experience also since the empty space signifies to the user that they are at the end.

raineorshine (changes requested):
Thanks! This looks great for a first PR. 

Please see comments below for some initial recommendations. I haven't gotten through everything, but there is enough here that I thought it would be beneficial to send it over.

The only major issue I see is the large amount of code duplication for the `CommandGrid` and related components. This might be acceptable for code that never gets touched again, but in a long-term software project we have to do the work of evaluating shared logic, parameterizing components, and determining the best abstractions to integrate new functionality. Maintaining large amounts of duplicate code is costly and invites bugs as code changes over time and diverges.

raineorshine:
image

fbmcipher:
Hey @raineorshine! Thanks so much for the detailed review and testing.

@pierogizwigilii and I have addressed and resolved all of the points above, with the exception of:

> Main body can still be scrolled on the right edge of the screen.

I'm unable to reproduce this – I've tested on MobileSafari on iOS 18.2.1, and Chrome on Android 15. Are there any steps to reproduce, or does it happen every time for you? I imagine you're testing on iPhone + MobileSafari too.

We do make sure to lock the main body from scrolling when the gesture cheatsheet is visible. Before, we were doing this by manually setting `document.body.style.overflow` to `"hidden"`, but in c3b08259e3017c62ae88d492032d42518ed705ea we now use the [`disableScroll` helper function in `src/device`](https://github.com/cybersemics/em/blob/main/src/device/disableScroll.ts) for consistency.

Perhaps if there are still issues, we should address them inside `disableScroll` rather than in the cheatsheet?

## Some notes on our resolutions:

1. Duplicate `GestureDiagramCheatsheet` has been removed – we added an extra `arrowhead` prop to it to control which kind of arrowhead to show (a filled arrowhead in most places, the cheatsheet uses an outlined arrowhead.
2. `CommandGrid` and `CommandGridItem` have also been removed – instead we use the existing `CommandTable` and `CommandRow` components with an additional `viewType` prop.
3. Factored out command groups into `constants.ts` to avoid duplication – though I 
know that this will likely need to change again for #2863.
4. `CommandRow` has been renamed to `CommandItem`, given that it is now responsible for showing both table rows and grid items.
5. Fixed all of the other smaller points you raised!

src/components/CommandTable.tsx
```diff
@@ -115,109 +28,123 @@ if (commandsUngrouped.length > 0) {
...
                 onSelect={onSelect}
+                search={search}
+                viewType={viewType}
+              />
+            )
+          } else if (previousSortOrder === 'type') {
+            return COMMAND_GROUPS.map(group => {
+              const commands = group.commands
+                .map(commandById)
+                .filter((command): command is Command => (isTouch ? !!command.gesture : !!command.keyboard))
```
raineorshine, line 115: While we're at it, could you remove `command is Command` from here and L131? I believe Typescript now infers the correct return type of `filter`.

src/components/CommandTable.tsx
```diff
@@ -115,109 +28,123 @@ if (commandsUngrouped.length > 0) {
...
+      <div
+        className={css({
+          display: 'flex',
+          flexDirection: 'row',
+          justifyContent: 'space-between',
+          gap: '5px',
+
+          /* On iOS, the sticky headers won't properly stick to the top of the scrollable container (causing a sliver
+             of content to peek through) unless we force the container into a new stacking context. */
+          willChange: 'transform',
```
raineorshine, line 80: Are you sure we need `willChange`? Usually it's needed only in rare circumstances. It does come with a (memory) performance cost.

src/components/CommandsGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+  selectedCommand,
+  commands,
+  title,
+  search,
+  viewType = 'table',
+}) => {
+  const modalClasses = modalTextRecipe()
+
+  return (
+    <div className={css({ position: 'relative', willChange: 'transform' })}>
```
raineorshine, line 31: Similarly, is `willChange` necessary here?

src/components/CommandsGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+                borderBottom: 'none',
+                position: 'sticky',
+                top: '-1.5rem',
+                background: 'linear-gradient(to bottom, {colors.bg} 85%, transparent)',
+                padding: '0.7rem 0',
+                margin: '0.5rem 0',
+                zIndex: 1,
+              }
+            : {
+                subtitle: modalClasses.subtitle,
```
raineorshine, line 47: `subtitle` appears to already be applied in `...modalClasses` above. What was the intention here?

src/components/SortButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { css, cx } from '../../styled-system/css'
+import theme from '../selectors/theme'
+import SortIcon from './icons/SortIcon'
+
+export interface SortButtonHandle {
+  closeDropdown: () => void
+}
+
+interface SortButtonProps {
+  onSortChange: (sortOrder: 'alphabetical' | 'type') => void
```
raineorshine, line 12: Let's factor out `alphabetical' | 'type' into its own type since it is shared by `CommandTable` and `SortButton`.

src/components/SortButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+        })}
+      >
+        <SortIcon size={20} fill={isLightTheme ? '{colors.lightgray}' : '{colors.fg}'} />
+      </div>
+      <div
+        className={cx(
+          css({
+            opacity: 0,
+            visibility: 'hidden',
+            transition: 'opacity 0.3s ease-in-out, visibility 0.3s ease-in-out',
```
raineorshine, line 68: You can use a predefined duration from https://github.com/cybersemics/em/blob/f8df451f3e3a4b45697d4e5d8167ead3373af52c/src/durations.config.ts.

Also, I don't believe that `visibility` can animate, or is there a behavior I'm not familiar with?

src/components/SortButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+            willChange: 'opacity, visibility',
```
raineorshine, line 69: Is `willChange` necessary?

src/components/SortButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+              display: 'inline-block',
+              position: 'relative',
+              '&:checked': {
+                borderColor: 'fg',
+              },
+            })}
+          />
+          <h3
+            className={css({
+              color: selectedSort === 'alphabetical' ? '{colors.fg}' : '{colors.gray50}',
```
raineorshine, line 127: Slightly simpler:

```
selectedSort === 'alphabetical' ? 'fg' : 'gray50'
```

src/components/SortButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+          justifyContent: 'center',
+        })}
+      >
+        <SortIcon size={20} fill={isLightTheme ? '{colors.lightgray}' : '{colors.fg}'} />
+      </div>
+      <div
+        className={cx(
+          css({
+            opacity: 0,
+            visibility: 'hidden',
```
raineorshine, line 67: Why both `opacity` and `visibility`?

src/components/SortButton.tsx
```diff
@@ -0,0 +1,176 @@
...
+            Alphabetical
+          </h3>
+        </label>
+        <label
+          className={css({
+            display: 'flex',
+            flexDirection: 'row',
+          })}
+        >
+          <input
```
raineorshine, line 142: I think there is enough duplicated code here to warrant factoring it out into a `SortOption` component. It can encapsulate the label + input + h3.

src/components/dialog/CloseButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+const CloseButton: React.FC<CloseButtonProps> = ({ onClick }) => {
+  return (
+    <button
+      onClick={onClick}
+      className={css({
+        width: '40px',
+        height: '40px',
+        borderRadius: '50%',
+        marginRight: '0.5rem',
+        backgroundColor: '{colors.darkgray}',
```
raineorshine, line 21: For color and backgroundColor, you can just specify the color name directly: `darkgray`.

src/components/dialog/CloseButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+        border: `2px solid {colors.gray50}`,
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        cursor: 'pointer',
+        transition: 'background-color 0.2s ease-in-out',
```
raineorshine, line 27: You can use a predefined duration from https://github.com/cybersemics/em/blob/f8df451f3e3a4b45697d4e5d8167ead3373af52c/src/durations.config.ts.

src/components/dialog/CloseButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+        _hover: {
+          backgroundColor: '{colors.gray15}',
```
raineorshine, line 29: For color and backgroundColor, you can just specify the color name directly: `gray15`.

src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+      className={css({
+        position: 'fixed',
+        top: 0,
+        left: 0,
+        right: 0,
+        bottom: 0,
+        display: 'flex',
+        alignItems: 'center',
+        justifyContent: 'center',
+        backgroundColor: '{colors.bgOverlay50}',
```
raineorshine, line 45: For color and backgroundColor, you can just specify the color name directly: `bgOverlay50`.

src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+        zIndex: 'modal',
+        overflow: 'hidden',
+      })}
+    >
+      <div
+        ref={dialogRef}
+        className={css({
+          backgroundColor: '{colors.bg}',
+          color: '{colors.fg}',
```
raineorshine, line 54: For color and backgroundColor, you can just specify the color name directly.

src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+            position: 'absolute',
+            bottom: 0,
+            left: 0,
+            right: 0,
+            height: '100px',
+            background: 'linear-gradient(to top, {colors.bg} 0%, transparent 100%)',
+            pointerEvents: 'none',
+            display: 'block',
+            opacity: 1,
+            transition: 'opacity 0.2s ease',
```
raineorshine, line 77: I believe `opacity` is no longer needed here.

src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,86 @@
...
+            zIndex: 2,
```
raineorshine, line 78: You can use a predefined zIndex from the Panda config. This helps ensure we have a consistent total ordering of layered elements.

https://github.com/cybersemics/em/blob/f8df451f3e3a4b45697d4e5d8167ead3373af52c/panda.config.ts#L294-L319.

src/components/dialog/DialogContent.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Content for dialog box.
+ */
+const DialogContent: React.FC<DialogContentProps> = ({ children }) => {
+  return (
+    <div
+      className={css({
+        fontSize: '1.25rem',
+        color: '{colors.fg}',
```
raineorshine, line 16: For color and backgroundColor, you can just specify the color name directly: `fg`.

src/components/dialog/DialogTitle.tsx
```diff
@@ -0,0 +1,41 @@
...
+      className={css({
+        display: 'flex',
+        justifyContent: 'space-between',
+        alignItems: 'center',
+      })}
+    >
+      <h2
+        className={css({
+          fontWeight: '700',
+          color: '{colors.fg}',
```
raineorshine, line 25: For color and backgroundColor, you can just specify the color name directly: `fg`.

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,81 @@
...
+      setTimeout(() => {
+        dispatch(toggleGestureCheatsheetActionCreator({ value: false }))
+      }, FADEOUT_DURATION)
+    }
+  }
+
+  useEffect(() => {
+    if (dialogRef.current) {
+      dialogRef.current.style.opacity = '1'
+    }
```
raineorshine, line 35: Setting the `opacity` directly on the DOM ref feels a bit low-level here. Would it be possible to do this in a more React-centric way? Or even use the [FadeTransition](https://github.com/cybersemics/em/blob/f8df451f3e3a4b45697d4e5d8167ead3373af52c/src/components/FadeTransition.tsx#L14) component if necessary?

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Handles the closure of the gesture cheatsheet.
+   */
+  const handleClose = () => {
+    if (dialogRef.current) {
+      dialogRef.current.style.opacity = '0'
+      setTimeout(() => {
+        dispatch(toggleGestureCheatsheetActionCreator({ value: false }))
+      }, FADEOUT_DURATION)
```
raineorshine, line 28: `FADEOUT_DURATION` is used by the modal, but it's of a legacy code, before we standardized animation durations and switched to PandaCSS. Could you port this over to a predefined duration from https://github.com/cybersemics/em/blob/f8df451f3e3a4b45697d4e5d8167ead3373af52c/src/durations.config.ts?

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,81 @@
...
+      dialogRef.current.style.opacity = '1'
+    }
+  }, [isOpen])
+
+  useEffect(() => {
+    if (isOpen) {
+      allowScroll(false)
+    } else {
+      allowScroll(true)
+    }
```
raineorshine, line 43: A bit simpler would be `allowScroll(!isOpen)`

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return () => {
+      allowScroll(true)
+    }
+  }, [isOpen])
+
+  /** Styles for the dialog fade in and out animation. */
+  const dialogAnimationStyles = css({
```
raineorshine, line 50: Prefer inlining for a small set of styles like this.

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,81 @@
...
+    opacity: 0,
+    transition: 'opacity 0.5s ease',
```
raineorshine, line 52: You can use a predefined duration from https://github.com/cybersemics/em/blob/f8df451f3e3a4b45697d4e5d8167ead3373af52c/src/durations.config.ts.

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,81 @@
...
+    willChange: 'opacity',
```
raineorshine, line 53: Is `willChange` necessary?

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <div ref={dialogRef} className={dialogAnimationStyles}>
+            <Dialog onClose={handleClose}>
+              <DialogTitle onClose={handleClose}>Gesture Cheatsheet</DialogTitle>
+              <DialogContent>
+                <CommandTable viewType='grid' />
+                <div
+                  style={{
+                    height: '64px',
+                  }}
+                />
```
raineorshine, line 71: This spacer could use a comment that explains why it's here, and how it's related to the linear gradient.

src/components/icons/GestureCheatsheetIcon.tsx
raineorshine, line 1: After merging `main`, you should be able to leave the `Command.svg` property empty on the `openGestureCheatsheet` command and remove this file.

raineorshine (changes requested):
Thank you for the update! It's really shaping up.

FYI There are some merge conflicts to resolve whenever you get a chance.

> > Main body can still be scrolled on the right edge of the screen.
> 
> I'm unable to reproduce this – I've tested on MobileSafari on iOS 18.2.1, and Chrome on Android 15. Are there any steps to reproduce, or does it happen every time for you? I imagine you're testing on iPhone + MobileSafari too.

I am able to reproduce this consistently on mobile Safari with the following steps:

1. Refresh the page
2. Perform '?' gesture to open cheatsheet.
3. Scroll between right edge of screen and right border of cheatsheet.

raineorshine:
- [ ] Scroll gradient covers scroll bar.

raineorshine:
- [ ] Sticky category header does not fully cover width, causing underlying content to peek through.

Brave Browser 2025-03-29 09 48 13 copy

raineorshine:
- [ ] Content is cut off on smaller devices. Instead, content should be responsive and scale with device width so there is no horizontal scrolling.

Brave Browser 2025-03-29 09 48 13

src/commands/openGestureCheatsheet.tsx
```diff
@@ -0,0 +1,30 @@
...
+  dispatch(toggleGestureCheatsheetActionCreator({ value: true }))
+  scrollTo('top')
+}
+
+const openGestureCheatsheetCommand: Command = {
+  id: 'openGestureCheatsheet',
+  label: 'Gesture Cheatsheet',
+  description: 'Opens a list of all the gestures.',
+  gesture: 'rdld',
+  multicursor: true,
```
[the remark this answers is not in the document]
fbmcipher: Got it – thanks for catching this oversight!

Resolved in 0c70b7b628c99625b04889c1334de0ecb9ab9603.

src/commands/openGestureCheatsheet.tsx
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Execute the openGestureCheatsheet command by calling the
+ * toggleGestureCheatsheetActionCreator.
+ * We scroll to the top of the page to ensure the cheatsheet is visible.
+ * Inside the cheatsheet, we lock the scroll of the page to ensure
+ * the cheatsheet is the only element that can be scrolled.
+ */
+const exec: Command['exec'] = dispatch => {
+  dispatch(toggleGestureCheatsheetActionCreator({ value: true }))
+  scrollTo('top')
```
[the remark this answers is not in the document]
fbmcipher: Agree with your rationale.

Addressed in 514083b3e8933f9db86c0a35a7f98a622b8b5fe5, and can confirm modal still displays correctly even if the body scroll is not at the top.

src/components/CommandItem.tsx
```diff
@@ -0,0 +1,180 @@
...
+      : s.endsWith('3') && n !== 13
+        ? s + 'rd'
+        : s + 'th'
+}
+
+type ViewType = 'grid' | 'table'
+
+/** Renders all of a command's details as either a grid item or table row. */
+const CommandItem = ({
+  viewType = 'table',
```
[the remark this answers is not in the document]
fbmcipher: I'm glad to hear you're happy with the refactor! 😁

src/components/GestureDiagram.tsx
```diff
@@ -30,6 +30,8 @@ interface GestureDiagramProps {
   cssRaw?: SystemStyleObject
   /** Whether to render the gesture with rounded corners. */
   rounded?: boolean
+  /** Which kind of arrowhead to draw gesture diagrams with. */
+  arrowhead?: 'filled' | 'outlined'
```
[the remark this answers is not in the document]
fbmcipher: Ah, good catch. Resolved in dd2389a6628ba261752743b44517ad2662cb34e1.

src/components/CommandTable.tsx
fbmcipher: Yes – removed in 979d220fb936dd60d265a2d600bf5ac542aaa60d.

src/components/SortButton.tsx
fbmcipher: Addressed in 62907958bc7006d949079c92edcf8c1fe4d6af11.

src/components/dialog/CloseButton.tsx
fbmcipher: Addressed in 62907958bc7006d949079c92edcf8c1fe4d6af11.

src/components/dialog/CloseButton.tsx
fbmcipher: Addressed in 62907958bc7006d949079c92edcf8c1fe4d6af11.

src/components/dialog/Dialog.tsx
fbmcipher: Addressed in 62907958bc7006d949079c92edcf8c1fe4d6af11.

src/components/dialog/Dialog.tsx
fbmcipher: Addressed in 62907958bc7006d949079c92edcf8c1fe4d6af11.

src/components/dialog/DialogContent.tsx
fbmcipher: Addressed in 62907958bc7006d949079c92edcf8c1fe4d6af11.

src/components/dialog/DialogTitle.tsx
fbmcipher: Addressed in 62907958bc7006d949079c92edcf8c1fe4d6af11.

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: That certainly is cleaner. Resolved in 4fc5edca98d98b93a57203dbe7665e9b3ca9c201.

src/components/CommandTable.tsx
```diff
@@ -115,109 +28,123 @@ if (commandsUngrouped.length > 0) {
...
+      </div>
 
-            // do not render groups with no commands on this platform
-            return commands.length > 0 ? (
+      {/* Smooth Fade-in Transition */}
+      <div
+        className={css({
+          textAlign: 'left',
+          opacity: isFading ? 0 : 1,
+          transition: 'opacity 0.3s ease-in-out',
```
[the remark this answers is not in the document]
fbmcipher: These comments seem to pertain to an earlier version of `CommandTable`. The [latest version](https://github.com/fbmcipher/em/blob/gesture-cheatsheet/src/components/CommandTable.tsx) uses `FadeTransition` instead of this old logic.

Please correct me if I'm wrong!

src/components/CommandTable.tsx
```diff
@@ -115,109 +28,123 @@ if (commandsUngrouped.length > 0) {
...
+    }
+  }, [])
+
+  useEffect(() => {
+    if (sortOrder !== previousSortOrder) {
+      setIsFading(true) // Start fading out
+      setTimeout(() => {
+        setPreviousSortOrder(sortOrder) // Update to new view after fade out
+        setIsFading(false) // Start fading in
+      }, 250) // Matches transition duration
```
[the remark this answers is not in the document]
fbmcipher: These comments seem to pertain to an earlier version of `CommandTable`. The [latest version](https://github.com/fbmcipher/em/blob/gesture-cheatsheet/src/components/CommandTable.tsx) uses `FadeTransition` instead of this old logic.

Please correct me if I'm wrong!

src/components/CommandsGroup.tsx
fbmcipher: I'm not sure. My suspicion is that this was manually included *before* `...modalClasses` or while debugging a style not appearing as expected, then accidentally committed.

I have removed the extraneous style in 697ea107600ad6101e10f512964e0e27920690bb.

src/components/SortButton.tsx
fbmcipher: Thanks – hardcoded durations have all been addressed in a60fab33978ed19db3f66ebbb21198d1c77c67f4.

That's right, `visibility` does not animate – this appears to have been an oversight. Removed in 331e8f11c1970b104fc43e1546b383232e3a87e3.

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: I've swapped out `FADEOUT_DURATION` for `durations.get('medium')` (both 400ms) in both GestureCheatsheet and ModalComponent in a06af4484d8b9ecfa3c5b7d2038388376486347a.

As these were the only two components which used `FADEOUT_DURATION`, I've also gone ahead and removed it from `constants.ts`.

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: Thanks – removed hardcoded durations in a60fab33978ed19db3f66ebbb21198d1c77c67f4.

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: Agreed – added in 83ec2255757916e6f497586a8418588467d7e587.

src/components/SortButton.tsx
fbmcipher: Good catch. Refactored in 36503289229f9bfac206d90f74f75a2fa75c572a.

src/components/dialog/Dialog.tsx
fbmcipher: Yes, you're right. Removed it in 5e46509f33b46bde15d288d5b9ede7f8cbdae8fc.

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: Resolved in dc825227f6db76b96cafff13314b637cfd63813b.

src/components/CommandsGroup.tsx
fbmcipher: It seems we *do* need `willChange` here. Without it, we get the strange sliver of text peeking through the sticky headers – and no amount of changing the z-index, position of the headers, background color etc. appeared to fix that, unfortunately.

**However**, this seems to be the only place it was needed. I overzealously applied `will-change` while trying to fix subsequent issues that came up while trying to fix the sticky header, like the sort overlay appearing underneath the sticky header.

I incorrectly assumed these were due to `will-change` creating new stacking contexts, but it must have been a simple z-index issue. My apologies for this oversight!

src/components/CommandTable.tsx
fbmcipher: Well, I was prepared to write a "yes we do" kind of comment, but I removed all of the `willChange` properties in this PR just to verify and it turns out most of them were not needed.

This is really bizarre, as we were *definitely* having recurring issues with the thoughts appearing *above* the dialog, no matter what we did with `zIndex`.

I saw that `Thought` and `Subthought` were using `willChange` to force new stacking contexts, and presumed the issue came down to this. Adding a `willChange` to the dialog fixed the issue right away, which just confirmed my suspicion.

But clearly the issue lied somewhere else! I'm not sure where, exactly, but at least we caught it! Thank you for the pointer.

We now only use `willChange` once in this PR, in the `CommandsGroup` component – where it does seem to be necessary.

src/components/SortButton.tsx
fbmcipher: Addressed above.

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: Addressed above.

src/components/CommandTable.tsx
```diff
@@ -115,109 +28,123 @@ if (commandsUngrouped.length > 0) {
...
+  const handleScroll = () => {
+    sortButtonRef.current?.closeDropdown()
+  }
+
+  useEffect(() => {
+    window.addEventListener('scroll', handleScroll, true)
+    return () => {
+      window.removeEventListener('scroll', handleScroll, true)
+    }
+  }, [])
```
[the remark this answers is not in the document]
fbmcipher: That's my mistake – this code definitely belongs in `SortButton`. 

Addressed in 7737dcf3830fa791ecd1ed6ed9a9e623859a6de9. Thanks for catching.

src/components/dialog/Dialog.tsx
fbmcipher: In this case I was able to remove the z-index entirely during my fix for the gradient overlapping scrollbar in ce98d535daab864e1f26d21b82b91cf005588dc2 – but thank you for the heads-up!

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: Thanks for raising the flag on this. We now use the FadeTransition component here.


3c744a02315bd3028f1f00d9a4107418ac579b3c

fbmcipher:
OK – I think we're getting really close! I've addressed all of your points above, and merged in main.

The last blockers are the failing Puppeteer tests:

modal-customizeToolbar-diff

**Tests for modal-customizeToolbar are failing due to the new CommandTable looking different**. 
* The test revealed some slight visual regressions with the new CommandTable, as well as an oversight in the search bar's placeholder text on desktop ("Search gestures..."), which I will address.
* The search bar looks quite different in this rendition – and we also have the Sort dropdown.
* Once I've addressed the visual regressions, am I okay to update the test so that it takes into account the new visual elements?


image
**Tests for commandPalette are failing.**
* I will remove the GestureCheatsheet from the Command Palette on desktop as it is a mobile-only command and should not be accessible on desktop.
* The commands also seem to be in a different order to the test. Will investigate, as that should not be the case.


multiline-multiline-thought-expanded-1-diff
**Tests for multiline-multiline-thought-expanded fail.**
* This one is slightly perplexing as to my awareness no code relating to multiline thoughts has been touched here. The tests pass in main, so it seems something has changed in this branch. I'll investigate this further, too.

I'll work on fixing up the code so these tests pass.

### Responses to your comments above:
* I was able to reproduce the scrolling-on-right-edge bug, and fixed in 8c1e818bacd12197878d2cf3e904c69f67421929.
    * It's difficult to disable scrolling entirely on iOS with just CSS, and most techniques I found that resolve this rely on hacky JS and setting the body to position: fixed.
    * In this solution, we simply prevent touch events from propagating on the overlay. That neatly disables the ability to scroll on it.
* Scroll bar being covered by the gradient fixed in ce98d535daab864e1f26d21b82b91cf005588dc2.
* As of 25499458af9756d0d78dc7fb03b9696817b8037a, the cheatsheet now scales with the device's width, ensuring there is no horizontal scrolling at small widths.
    * The smallest device we scale for is **320px wide**, which is the same width as the original iPhone up until the 2015 iPhone SE. I figure if we're visible *that* far back, we're doing okay!
    * I have added breakpoints in `panda.config.ts` and use PandaCSS's responsive functionality to implement this.
    * This fix also seems to have fixed the bug you pointed out where sticky category headers were not fully covering width.

raineorshine:
Great, thanks for the update. I will do another review as soon as possible.

> * Once I've addressed the visual regressions, am I okay to update the test so that it takes into account the new visual elements?

Yes

> * This one is slightly perplexing as to my awareness no code relating to multiline thoughts has been touched here. The tests pass in main, so it seems something has changed in this branch. I'll investigate this further, too.

Yes, that's odd. Does the same test fail your machine? I will re-run the CI in case it is a false positive.

src/components/CommandsGroup.tsx
raineorshine: Interesting, I hadn't used `will-change` to create a stacking context before. The usage in `Thought` and `Subthought` is for optimizing animation performance.

Do you know if using `will-change: 'transform'` without a `transform` property bypasses the performance issues that are [warned against](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change)? I think it may be safer to use one of the other methods of creating a stacking context, as I don't want to accidentally introduce performance issues with this.

Either of the following should create a stacking context without affecting the zIndex:

- `transform: 'translate(0, 0);'`
- `contain: 'layout paint'`

raineorshine:
Hi! I only had time to start the review, but a couple issues are still occurring for me.  Tested d39a982c8c7.

> I was able to reproduce the scrolling-on-right-edge bug, and fixed in [8c1e818](https://github.com/cybersemics/em/commit/8c1e818bacd12197878d2cf3e904c69f67421929).
> * It's difficult to disable scrolling entirely on iOS with just CSS, and most techniques I found that resolve this rely on hacky JS and setting the body to position: fixed.
> * In this solution, we simply prevent touch events from propagating on the overlay. That neatly disables the ability to scroll on it.

Unfortunately I'm still experiencing this issue. Sorry to ask, but are you not able to reproduce? I followed the steps in https://github.com/cybersemics/em/pull/2850#pullrequestreview-2727333729. If not, I will ask Trevin to see if he is able to reproduce it. It seems like we might be seeing different behavior.

> * Scroll bar being covered by the gradient fixed in [ce98d53](https://github.com/cybersemics/em/commit/ce98d535daab864e1f26d21b82b91cf005588dc2).

Scroll bar is still covered by the gradient when scrolling down to the bottom.

trevinhofmann:
> Unfortunately I'm still experiencing this issue. Sorry to ask, but are you not able to reproduce? I followed the steps in https://github.com/cybersemics/em/pull/2850#pullrequestreview-2727333729. If not, I will ask Trevin to see if he is able to reproduce it. It seems like we might be seeing different behavior.

I am able to reproduce this in Safari with iOS 18.3.2 using the latest commit (d39a982c8c78cdb587b9f2d93f86defa81425e25).

fbmcipher:
Thank you both – this feels very close now!

I've fixed the scroll bar hiding under the gradient issue, and made the changes I discussed in https://github.com/cybersemics/em/pull/2850#issuecomment-2788299562 to ensure we're now passing snapshot tests.

Unfortunately, try as I might I simply cannot reproduce the scrolling-on-right-edge bug. I've tested across a range of devices in Safari, all running the latest iOS 18.4.1:

* iPhone 16 (physical)
* iPhone Xr (physical)
* iPhone SE (physical)
* iPhone 15 (simulator)

I spent a good amount of time trying to break this – playing with page zoom, trying to grab the scroll bar manually, rotating the device, closing and opening the cheatsheet, having a really long set of thoughts so the main body scrolled – none of this got me any closer to reproducing the issue.

Could there perhaps be some kind of caching issue going on? Maybe some kind of browser config that's changing the behaviour? It feels unlikely, but I'm unsure what else could be going on here...

Please find attached screen recordings showing me trying to reproduce the issue following the steps in https://github.com/cybersemics/em/pull/2850#pullrequestreview-2727333729. Maybe I'm missing something...

**On simulator (easier to see the cursor)**

https://github.com/user-attachments/assets/d946d9b2-0786-4a46-b06c-6f4eb958ece1



**On physical device (a little hard to see what's going on as you can't see my finger, but same behaviour as in simulator)**

https://github.com/user-attachments/assets/81fe2d9d-b04e-4456-bb55-868c074830fe

raineorshine:
Thanks for the screen capture. From that I figured out that it only occurs when the Single Tab setting is enabled in Safari!

image

I would love to have a fix, but I am comfortable tabling it for now as long as it does not occur in the Capacitor app.

src/e2e/puppeteer/__tests__/__image_snapshots__/modal/modal-help-light.png
trevinhofmann, line 1: This snapshot diff doesn't look great to me. Each command's information (e.g. `Back`, `Escape`, `Move the cursor up a level.`) used to be grouped together well. Now the third line (`Move the cursor up a level.`) is about as close to the next command as it is to its own command.

trevinhofmann (changes requested):
Thank you for the PR, @fbmcipher! The gesture cheatsheet looks nice.

Just a few initial issues I've noticed while testing in Chrome on Android:

## 1. Gesture Cheatsheet resizes incorrectly from portrait to landscape

### Steps to reproduce:

1. Open Em in portrait
2. Open the Gesture Cheatsheet
3. Rotate to landscape

### Current behavior:

The height of the Gesture Cheatsheet is taller than the visible page.

landscape

### Expected behavior:

The Gesture Cheatsheet should be resized as if it were originally opened in landscape mode.

## 2. Box artifact around the tip of the gesture arrows

In the Gesture Cheatsheet gesture diagrams, the tip of the arrows are surrounded by a faded box artifact.

artifact

## 3. Search too fuzzy in the Gesture Cheatsheet

A bit of fuzziness is nice in search, but this implementation seems too lenient. **This might not be blocking**, but it was a strange UX during testing.

### Current Behavior

A search for "Forward" returns results such as "New Grandchild" because it matches the letters w/r/a/d.

### Expected Behavior

A search for "Forward" should probably only match "Forward" and "Jump forward".

## 4. Cancel gesture diagram is missing

In the Gesture Cheatsheet search, there is an entry for "Cancel" without a gesture diagram.

### Steps to Reproduce

1. Open the Gesture Cheatsheet.
2. Search for "Cancel"
3. Scroll down near the end to locate "Cancel" result (Odd, it should be at the top for an exact match?)

### Current Behavior

There is an entry for "Cancel", missing the gesture diagram.

### Expected Behavior

The "Cancel" gesture should probably not be included in the Gesture Cheatsheet search results.

raineorshine:
Thanks @trevinhofmann! 

Agree with all of these. I'm just going to move 2 and 3 to separate issues since they are related to existing functionality in `main`.

fbmcipher:
> **@raineorshine** Thanks for the screen capture. From that I figured out that it only occurs when the Single Tab setting is enabled in Safari!

Yes, that did it!

https://github.com/user-attachments/assets/e79f3f00-4c62-40fe-a4c5-0a8649a56cfa

There's something a little bit strange going on, though. It doesn't scroll until I hit *just* the right spot and then I can fling it around. Then it locks again, and it goes back to being difficult to reproduce. Hopefully the video demonstrates what I mean. It's really strange.

I _can_ confirm it's not an issue in Capacitor, though. Let's table for now, and keep it on our radar if it becomes a more serious issue later on.

src/e2e/puppeteer/__tests__/__image_snapshots__/modal/modal-help-light.png
fbmcipher: Thank you, you're totally right. The command table got refactored pretty heavily during this work, so it looks like a visual regression slipped in and flew past my radar.

I've tweaked the spacing in e524c49d665b46c00487b8ddc57c72f84b8121f3 and compared with the spacing in `main` to ensure it's consistent. I've also updated the snapshot tests again.

**Here's the diff between the initial snapshot, and the new snapshot as of this commit.** Do you think this looks better?

UntitledDiff

raineorshine:
> @raineorshine Do you have any thoughts about landscape mode in **em**? Is it something you've thought about or used yourself?

Since **em** is a kind of text editor and the virtual keyboard is used so often, we design primarily for portrait mode. The only real requirement for landscape mode is that the design doesn't totally break. It's okay if it is less usable though.

The only time I have used **em** in landscape mode is when I was reading a PDF with a small font size in another app, and was switching over to **em** to take notes. In that case I was in the PDF reader more often than in **em**, and it wasn't worth it to me to rotate the phone every time. I could tolerate the minuscule amount of space above the virtual keyboard as I just needed to enter a quick note.

It is possible to have an external keyboard connected to an iPad in landscape mode, but this is lower priority.

> > ## 2. Box artifact around the tip of the gesture arrows
> 
> Thanks for opening an issue on this @raineorshine. I don't have any particular insight on how to fix this right now, but I'm glad it's been tabled. Does this issue block merging this PR?

Nope! Totally separate.

> ## 4. Cancel gesture diagram is missing
> In the Gesture Cheatsheet search, there is an entry for "Cancel" without a gesture diagram.
> 
> ### Steps to Reproduce
> 1. Open the Gesture Cheatsheet.
> 2. Search for "Cancel"
> 3. Scroll down near the end to locate "Cancel" result (Odd, it should be at the top for an exact match?)
> 
> ### Current Behavior
> There is an entry for "Cancel", missing the gesture diagram.
> 
> ### Expected Behavior
> The "Cancel" gesture should probably not be included in the Gesture Cheatsheet search results.

Now that I test this, I'm actually not able to reproduce. I don't see a Cancel entry when I search "Cancel".

@fbmcipher Are you able to reproduce with @trevinhofmann's steps?

The Cancel gesture probably should be included in the cheatsheet, as that would be most helpful to the user, but let me know if you think otherwise.

> I _can_ confirm it's not an issue in Capacitor, though. Let's table for now, and keep it on our radar if it becomes a more serious issue later on.

Great

fbmcipher:
> Since em is a kind of text editor and the virtual keyboard is used so often, we design primarily for portrait mode. The only real requirement for landscape mode is that the design doesn't totally break. It's okay if it is less usable though.

OK, agreed with the rationale and focus on portrait mode.

I think we're definitely in "less usable" territory but certainly not "totally broken" territory! I think we're okay to leave this as-is for now.

> I'm wondering... would it be better to include it in the cheatsheet? If the purpose of the cheatsheet is to help teach the user about gestures, I don't see a reason to exclude the Cancel gesture.

My instinct was *not* to include it as it felt like the gesture only makes sense in context.

If we listed it along with the rest of the gestures, it might not be fully clear that the Cancel gesture can only be inputted while you're already inputting another gesture. In this sense it works quite a bit differently to the rest of the gestures on the cheatsheet. As much as we could add the description "Cancel the current gesture", I feel it's a bit of an outlier and worth excluding just for the sake of avoiding confusion.

The cancel gesture shows quite intuitively when the user might need it in the autocomplete.

That said, if you disagree, we could add it in fairly easily. As the `cancel` command doesn't have a GesturePath we'd just use the icon in place of the GestureDiagram.

raineorshine:
I hadn't thought of it as a source of confusion, but that's a fair point! I'm fine excluding it then. Thanks for your perspective.

src/components/dialog/DialogContent.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { css } from '../../../styled-system/css'
+
+interface DialogContentProps {
+  children: React.ReactNode
+}
+
+/**
+ * Content for dialog box.
+ */
+const DialogContent: React.FC<DialogContentProps> = ({ children }) => {
```
raineorshine, line 11: We can use the React-provided `PropsWithChildren`:

```ts
const DialogContent: React.FC<PropsWithChildren> = ({ children }) => {
```

src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+interface DialogProps {
+  children: React.ReactNode
+  onClose: () => void
+}
+
+/**
+ * Dialog component.
+ */
+const Dialog: React.FC<DialogProps> = ({ children, onClose }) => {
```
raineorshine, line 12: For consistency, let's remove `children` from the props and use the React-provided `PropsWithChildren` generic:

```ts
React.FC<PropsWithChildren<DialogProps>>
```

src/components/dialog/DialogTitle.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import { css } from '../../../styled-system/css'
+import CloseButton from './CloseButton'
+
+interface DialogTitleProps {
+  children: React.ReactNode
```
raineorshine, line 6: Same here

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { useEffect } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
+import { css } from '../../../styled-system/css'
+import { toggleGestureCheatsheetActionCreator } from '../../actions/toggleGestureCheatsheet'
+import allowScroll from '../../device/disableScroll'
```
raineorshine, line 5: Just noticing this, but could you rename `disableScroll.ts` to `allowScroll.ts`? I'd like to follow the convention that filenames match the name of the default export.

src/components/dialog/Dialog.tsx
```diff
@@ -0,0 +1,113 @@
...
+      >
+        {children}
+        <div
+          className={css({
+            position: 'absolute',
+            bottom: 0,
+            right: 0,
+            left: 0,
+            marginRight: 20,
+            height: '100px',
```
raineorshine, line 102: Can we make this about half as tall? I think it can be a bit subtler without losing effectiveness.

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,67 @@
...
+            {/* In Dialog's styles, we fix a gradient to the bottom of the dialog to subtly indicate that
+                the dialog is scrollable.
+                When the dialog reaches the bottom, we want to ensure the gradient is no
+                longer visible.
+                Adding a 64px margin to the bottom of the dialog content ensures that the gradient
+                is not visible, and signals to the user that they have reached the bottom of the dialog.
+            */}
+            <div
+              style={{
+                height: '64px',
```
raineorshine, line 57: The comment is helpful, but I think it would be better to create a programmatic relationship between this and the `100px` in `Dialog.tsx`. As it stands, these hard-coded numbers are functionally coupled but not connected in the code. This obscures the relationship and increases the risk that one will be changed without the other in a future refactor. 

I'm open to ideas about the best place to put the shared value. You could use a Panda recipe (as done in [toolbarPointerEventsRecipe](https://github.com/cybersemics/em/blob/3f1d91c6622195ebcebc600e6417b419591dcf5c/src/recipes/toolbarPointerEvents.ts)), or a simple constant.

src/components/dialog/GestureCheatsheet.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+            {/* In Dialog's styles, we fix a gradient to the bottom of the dialog to subtly indicate that
+                the dialog is scrollable.
+                When the dialog reaches the bottom, we want to ensure the gradient is no
+                longer visible.
+                Adding a 64px margin to the bottom of the dialog content ensures that the gradient
+                is not visible, and signals to the user that they have reached the bottom of the dialog.
+            */}
+            <div
+              style={{
```
raineorshine, line 56: I believe we can use `css(...)` since this style does not require runtime values.

raineorshine (changes requested):

src/components/dialog/DialogTitle.tsx
fbmcipher: 52d6a3e1d74c074f852f67d3fa9df3ffad08edaf

src/components/dialog/Dialog.tsx
fbmcipher: 52d6a3e1d74c074f852f67d3fa9df3ffad08edaf

src/components/dialog/DialogContent.tsx
fbmcipher: 52d6a3e1d74c074f852f67d3fa9df3ffad08edaf

src/components/SortButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+          alignItems: 'center',
+          justifyContent: 'center',
+        })}
+      >
+        <SortIcon size={20} fill={isLightTheme ? '{colors.lightgray}' : '{colors.fg}'} />
+      </div>
+      <div
+        className={cx(
+          css({
+            opacity: 0,
```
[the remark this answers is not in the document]
fbmcipher: Thank you for catching this. Best to keep things consistent, so refactored to use the `FadeTransition` component – with the `unmountOnExit` prop to make sure the dropdown is removed from the DOM when closed, preventing interaction.

da98b0bac92fac7e2ba5fde2245d3560816b8101

src/components/SortButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+}
+
+interface SortButtonProps {
+  onSortChange: (sortOrder: CommandSortType) => void
+}
+
+/**
+ * SortButton component.
+ * */
+const SortButton = forwardRef<SortButtonHandle, SortButtonProps>(({ onSortChange }, ref) => {
```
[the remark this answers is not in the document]
fbmcipher: 593ae51e6e87c5bf34e897c2170d2cca400a9a1b

src/components/SortButton.tsx
```diff
@@ -0,0 +1,117 @@
+import { forwardRef, useEffect, useState } from 'react'
+import { useSelector } from 'react-redux'
+import { css, cx } from '../../styled-system/css'
+import CommandSortType from '../@types/CommandSortType'
+import theme from '../selectors/theme'
+import SortOption from './SortOption'
+import SortIcon from './icons/SortIcon'
+
+export interface SortButtonHandle {
```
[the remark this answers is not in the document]
fbmcipher: 6f56b388484565ca90ab3d0e2796c859b745a9fb

src/components/CommandTable.tsx
```diff
@@ -124,110 +31,94 @@ if (commandsUngrouped.length > 0) {
...
-        ) : (
-          groups.map(group => {
-            const commands = group.commands
-              .map(commandById)
-              .filter((command): command is Command => (isTouch ? !!command.gesture : !!command.keyboard))
+        ) : null
+      })
+    } else {
+      const commandsWithGestures = commands.filter(command => !!command.gesture)
+      const sortedCommands = commandsWithGestures.sort((a, b) => a.label.localeCompare(b.label))
```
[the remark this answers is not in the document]
fbmcipher: I think this got lost during the `CommandTable` refactor. Initially, the cheatsheet used a duplicated `CommandTable` called `CommandGrid`. The logic must have changed there, and I must have accidentally brought that new logic in over the existing CommandTable implementation.

Very sorry for letting this slip in, I will be more mindful during refactors!

In eedb25c001dbf2c30a33383c38b886e1285de784, we
* calculate sort on runtime
* properly filter with gesture/keyboard distinction

The next step of optimisation here would probably be to define command groups on the Command interface rather than in a constant variable. CommandTable currently works off of a constant value which splits commands into groups. #2863

src/components/CommandTable.tsx
```diff
@@ -124,110 +31,94 @@ if (commandsUngrouped.length > 0) {
...
+          onSelect={onSelect}
+          search={search}
+          viewType={viewType}
+        />
+      )
+    } else if (sortOrder === 'type') {
+      return COMMAND_GROUPS.map(group => {
+        const commands = group.commands
+          .map(commandById)
+          .filter(command => (isTouch ? !!command.gesture : !!command.keyboard))
```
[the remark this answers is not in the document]
fbmcipher: Great optimisation. eedb25c001dbf2c30a33383c38b886e1285de784

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: Thanks for catching this. Fixed in 02a178477df88bff88f7a8dbf219696e47cdc37f.

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: Good point. This generally exposed a deeper opportunity for refactoring – the dialog/cheatsheet is composed of multiple elements with different styles. As of 98d06d68ca8fa36659e4585aea46ad18ebd281b3 we now use a PandaCSS slot recipe to unify all of these styles into a single file.

This then creates an opening to create the programmatic relationship between the spacer and the gradient in 7e4920dfdc9c1cd311467d5a12d8599d76f8678e.

src/components/dialog/GestureCheatsheet.tsx
fbmcipher: Addressed as part of the refactors in 98d06d68ca8fa36659e4585aea46ad18ebd281b3.

src/components/CommandsGroup.tsx
fbmcipher: In 851fc3deb62a9a8d77f365840735bdcc8526efaa we now use `contain: 'layout paint'` instead, which creates the stacking context and solves the visual artifacting nicely without the concern of `will-change`'s performance impact.

Thanks for the research!

fbmcipher:
* 851fc3d uses a more efficient method of creating a stacking context without the performance implications of `will-change`
* 5743954 minor aesthetic tweak; tightens some margins on the sticky headers to tune the visual appearance of the effect and get the stick to occur at just the right point while scrolling
* 1dbcd64 merges the latest changes from `main`, carefully resolving conflicts to preserve and accommodate the new structures Karun is introducing

raineorshine (approved):

Issue #2846, opened 2025-03-11
Drag-and-drop: Collapse expandHoverDownPaths when hoveringPath changes


raineorshine:
## Steps to Reproduce

```
- x
- a
  - b
```

1. Drag `x`
2. Hover over `a` to expand.
3. Drag `x` back over itself and hold.

## Current Behavior

`a` stays expanded via `state.expandHoverDownPaths`. It stays permanently expanded until the drag-and-drop ends.

## Expected Behavior

After one second, `a` should collapse.

Generally: When `state.hoveringPath` is no longer on a thought or its descendant, after a delay of [EXPAND_HOVER_DELAY](https://github.com/cybersemics/em/blob/d1111a36daeac9bdb0779f5e90d02a93946beca1/src/constants.ts#L16), the thought should be collapsed (i.e. the corresponding Path in `state.expandHoverDownPaths` should be removed).

## Testing

Add a puppeteer test.

raineorshine:
Ancestors should stay expanded.

## Steps to Reproduce

1. Drag `x`
2. Hover over `c` to expand.
3. Hover over `c/d` to expand.

## Current and Expected Behavior

When `c/d` is expanded, `c` should stay expanded.

PR #2851, opened 2025-03-18
Drag-and-drop: Collapse expandHoverDownPaths when hoveringPath changes.

karunkop:
PR for #2846 

## Overview

This PR adds a new feature that automatically  collapses thoughts during drag operations, providing a more intuitive user experience when reorganizing content. The implementation allows thoughts to collapse them when the drag operation moves elsewhere.

### Technical Implementation

The feature consists of two main action creators:

- expandHoverDownActionCreator: Monitors drag operations and expands thoughts when hovering over their subthoughts area
- collapseHoverDownActionCreator: Tracks hovering path changes during drag and collapses thoughts when they are exited.


src/actions/collapseHoverDown.ts
```diff
@@ -0,0 +1,197 @@
...
+    }, EXPAND_HOVER_DELAY)
+  }
+
+/**
+ * Removes a path and all its descendants from expandHoverDownPaths and expanded states.
+ * This ensures that when a parent is collapsed, all its children are collapsed as well.
+ */
+const collapseHoverDown = (state: State, { path }: { path: Path }): State => {
+  // Get all paths from both expanded and expandHoverDownPaths states
+  const expandedPaths = Object.entries(state.expanded).map(([_, path]) => path)
```
raineorshine, line 41: FYI We have [keyValueBy](https://github.com/cybersemics/em/blob/37f3e98916904810830c18b92436cad2016fac47/src/util/keyValueBy.ts) to filter objects without having to map out the entries and perform mutations on temporary data structures.

raineorshine (commented):
Hi, thanks so much for your submission.

After reviewing your approach, I'm wondering why we need a separate action-creator at all. [expandHoverDown](https://github.com/cybersemics/em/blob/37f3e98916904810830c18b92436cad2016fac47/src/actions/expandHoverDown.ts) is already on a 1s timer, and the old path should be collapsed at the same time that the new path is expanded. Would it be possible achieve the desired functionality within `expandHoverDown`?

karunkop:
> Hi, thanks so much for your submission.
> 
> After reviewing your approach, I'm wondering why we need a separate action-creator at all. [expandHoverDown](https://github.com/cybersemics/em/blob/37f3e98916904810830c18b92436cad2016fac47/src/actions/expandHoverDown.ts) is already on a 1s timer, and the old path should be collapsed at the same time that the new path is expanded. Would it be possible achieve the desired functionality within `expandHoverDown`?

While it may be possible to achieve the desired functionality within `expandHoverDown`,  we may still need to keep them separate because they work differently:

- Expanding happens when you hover over a new valid path.
- Collapsing happens when you exit any expanded path, which means we have to check past states and descendants.
If we combined them, we'd run into issues:

- Moving between paths would mess with timers—like collapsing one path while another is expanding.
- Collapsing wipes out all descendants, and shoving that into expandHoverDown would make things way more complicated.
- Some edge cases, like dragging or hovering outside, need collapsing without expanding, which is cleaner with separate logic.

Mixing them would just create timer conflicts, unnecessary state checks, and messy cleanup. Keeping them separate keeps everything clear, efficient, and working as expected.

src/actions/collapseHoverDown.ts
karunkop: Oh yep you are right! Thanks for the information

raineorshine:
> * Expanding happens when you hover over a new valid path.
> * Collapsing happens when you exit any expanded path, which means we have to check past states and descendants.
>   If we combined them, we'd run into issues:

I don't see really see the difference. Both `expandHoverDown` and `collapseHoverDown` are dispatched at exactly the same time: https://github.com/karunkop/em/blob/fd54d43701b318cbcdbf37edfb10b5d3d1b8ebc8/src/actions/dragInProgress.ts#L142C32-L143

Both `expandHoverDown` and `collapseHoverDown` modify the same state: `state.expanded` and `state.expandHoverDownPaths`.

From what I can tell, the timing is exactly the same and the state affected is exactly the same, so it's not clear why these wouldn't be handled in a unified way. i.e. **Whenever the hovering path changes, call the debounced function. When the debounced function triggers, update the state accordingly.**

> * Moving between paths would mess with timers—like collapsing one path while another is expanding.

It seems like we would only need one timer: The debounced function. The paths always collapse/expand at the same time.

> * Collapsing wipes out all descendants, and shoving that into expandHoverDown would make things way more complicated.

I think the handling of descendants may be overcomplicated. We only need a single `expandHoverDownPath`. Then you can expand each subpath, i.e. `[a]`, `[a, b]`, `[a, b, c]`, etc.

> * Some edge cases, like dragging or hovering outside, need collapsing without expanding, which is cleaner with separate logic.

You're still thinking about it procedurally (i.e. collapse A, then collapse B, then expand C, etc) rather than functionally/declaratively (when the hovering path is A → expand B). The expanded hover down path is simply a function of the hovering path, with a delay.

> Mixing them would just create timer conflicts, unnecessary state checks, and messy cleanup. Keeping them separate keeps everything clear, efficient, and working as expected.

I don't know, it seems like these supposed conflicts are only exist *because* we've separated out the functionality. As far as I can tell, it's just a complicated way to define a delayed function of the hovering path. So, I'm not convinced there would be any timer conflicts or unnecessary state checks in a unified approach.

Let me know if I'm missing something! Thanks!

karunkop:
> > * Expanding happens when you hover over a new valid path.
> > * Collapsing happens when you exit any expanded path, which means we have to check past states and descendants.
> >   If we combined them, we'd run into issues:
> 
> I don't see really see the difference. Both `expandHoverDown` and `collapseHoverDown` are dispatched at exactly the same time: https://github.com/karunkop/em/blob/fd54d43701b318cbcdbf37edfb10b5d3d1b8ebc8/src/actions/dragInProgress.ts#L142C32-L143
> 
> Both `expandHoverDown` and `collapseHoverDown` modify the same state: `state.expanded` and `state.expandHoverDownPaths`.
> 
> From what I can tell, the timing is exactly the same and the state affected is exactly the same, so it's not clear why these wouldn't be handled in a unified way. i.e. **Whenever the hovering path changes, call the debounced function. When the debounced function triggers, update the state accordingly.**
> 
> > * Moving between paths would mess with timers—like collapsing one path while another is expanding.
> 
> It seems like we would only need one timer: The debounced function. The paths always collapse/expand at the same time.
> 
> > * Collapsing wipes out all descendants, and shoving that into expandHoverDown would make things way more complicated.
> 
> I think the handling of descendants may be overcomplicated. We only need a single `expandHoverDownPath`. Then you can expand each subpath, i.e. `[a]`, `[a, b]`, `[a, b, c]`, etc.
> 
> > * Some edge cases, like dragging or hovering outside, need collapsing without expanding, which is cleaner with separate logic.
> 
> You're still thinking about it procedurally (i.e. collapse A, then collapse B, then expand C, etc) rather than functionally/declaratively (when the hovering path is A → expand B). The expanded hover down path is simply a function of the hovering path, with a delay.
> 
> > Mixing them would just create timer conflicts, unnecessary state checks, and messy cleanup. Keeping them separate keeps everything clear, efficient, and working as expected.
> 
> I don't know, it seems like these supposed conflicts only exist _because_ we've separated out the functionality. As far as I can tell, it's just a complicated way to define a delayed function of the hovering path. So, I'm not convinced there would be any timer conflicts or unnecessary state checks in a unified approach.
> 
> Thanks for the additional consideration, and definitely let me know if I'm missing something. Thanks!

Hi, I have thoroughly analyzed your proposed thought pattern and successfully implemented the unification of expansion and collapse logic while utilizing the same debounced function and timer. Though the process was somewhat tricky and presented several challenges, the changes are now stable and should function as intended.

Core Changes:
- Unified the expanding and collapsing logic within a single hoverPathDebounced function.
- Introduced the isPathInsideExpandedPath function to handle complex path relationships.
- Added a forcePathReexpansion flag to resolve the "stuck expansion" issue, where paths might visually appear expanded but were not properly tracked in state. This also addresses edge cases involving rapid hover transitions, which was one of the key challenges encountered during implementation.

Issue #2856, opened 2025-03-22
Formatted whitespace not trimmed when splitting thought


raineorshine:
## Steps to Reproduce

1. Paste as text/html: `<meta charset='utf-8'>one<i> two</i>`
3. Set the caret after `one`, before the space.
4. Hit Enter.

## Current Behavior

The space before `two` is not trimmed:

```
- one
-  two
```

## Expected Behavior

Whitespace at the beginning/end of a thought should be trimmed on change, even if it is in a formatting tag like italics, bold, underline, strikethrough, span, or font.

```
- one
- two
```

## Testing

Add a unit test.

## Notes

Make sure to find where it is correctly trimming non-formatted thoughts, and use/modify the existing logic rather than adding something completely new.

PR #2858, opened 2025-03-24
Formatted whitespace not trimmed when splitting thought

karunkop:
PR for #2856 

### Problem
- When splitting a thought, content inside formatting tags containing whitespace was not being properly trimmed. This resulted in unwanted leading spaces when the thought was split.

### Solution
- Implemented trimHtml on both the left and right values before stripping, ensuring proper whitespace removal and maintaining clean formatting.


src/actions/__tests__/splitThought.ts
```diff
@@ -132,3 +132,31 @@ it('move children to the correct sibling in a sorted context', () => {
...
+  <li>${HOME_TOKEN}${'  '}
+    <ul>
+      <li>one</li>
+      <li><i>two</i></li>
+    </ul>
+  </li>
+</ul>`)
+
+  const cursorThoughts = childIdsToThoughts(stateNew, stateNew.cursor!)
+  expect(cursorThoughts).toMatchObject([{ value: '<i>two</i>', rank: 1 }])
```
raineorshine, line 161: Here it would be slightly better to just assert the value, as rank is not relevant to this test.

raineorshine (approved):
Looks good, thanks.

Good use of the existing `trimHtml`.

Issue #2855, opened 2025-03-22
Option + ArrowLeft: Do not move caret to end of thought


raineorshine:
Tested on Chrome.

## Steps to Reproduce

1. Set the caret at the beginning of a thought.
2. Hit <kbd>Alt/Option</kbd> + <kbd>ArrowLeft</kbd>.

## Current Behavior

Caret moves to the end of the thought.

## Expected Behavior

Caret should not move. This should behave the same as <kbd>Meta</kbd> + <kbd>ArrowLeft</kbd> and bare <kbd>ArrowLeft</kbd>.

PR #2859, opened 2025-03-24
Option + ArrowLeft: Do not move caret to end of thought

karunkop:
PR for #2855 

### Problem:

- The browser's default behavior for Alt+ArrowLeft in contentEditable elements is to jump to the previous word boundary When at the beginning of a thought, there's no previous word boundary within the current contentEditable element
This caused the browser to behave unexpectedly, jumping the caret to the end of the text.

### Solution

- added a prevent default behavior for alt/option + leftarrow in global keydown handler

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -18,6 +18,7 @@ import { suppressExpansionActionCreator as suppressExpansion } from './actions/s
 import { isMac } from './browser'
 import * as commandsObject from './commands/index'
 import { AlertType, COMMAND_PALETTE_TIMEOUT, Settings } from './constants'
+import * as selection from './device/selection'
 import globals from './globals'
 import getUserSetting from './selectors/getUserSetting'
 import gestureStore from './stores/gesture'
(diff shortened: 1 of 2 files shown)


src/commands.ts
```diff
@@ -322,6 +323,12 @@ export const inputHandlers = (store: Store<State, any>) => ({
       globals.suppressExpansion = false
     }
 
+    // Handle Alt+ArrowLeft when caret is at the beginning of a thought
```
raineorshine: Try to avoid explaining functionality by saying "Handle...". The problem is that it doesn't really explain it. Handle it how? A better comment would be one that explains _why_ this is needed, what would happen if we _didn't_ do it, and any limitations on our knowledge.

```
// For some reason, when the caret is at the beginning of the thought, alt + ArrowLeft sets the caret to the end.
// Prevent this default behavior, as the caret should have nowhere to go when it is already at the beginning.
```

raineorshine (approved):
Looks good, thanks!

Issue #2848, opened 2025-03-14
Invalid cursor after indent and note


raineorshine:
## Steps to Reproduce

```
- a
- b
  - =note
    - c
```

1. Set the cursor on `b`
2. Indent
3. Activate the Note command (Meta + Option + n)

## Current Behavior

Cursor is set to invalid Path `['b']`.

## Expected Behavior

Cursor should be set to correct Path `['a', 'b']`.

## Testing

Add a unit test.

PR #2860, opened 2025-03-25
Invalid cursor after indent and note

karunkop:
PR for #2848 

### Problem

The bug was caused by a fundamental issue with how the Note component handled its path prop after structural changes in the thought hierarchy.

- Stale Path Props: When a thought was indented (moving from ['b'] to ['a', 'b']), React did not automatically update the path prop passed to the Note component. The component continued to use the original path ['b'] which was now invalid.
- Direct Path Usage: The Note component was directly using the path prop in all its event handlers (onFocus, onChange, onKeyDown) without checking if that path was still valid in the current state.
- Inconsistent State: This led to operations being performed on the wrong path. For example, after indenting thought 'b' under 'a':
   - The correct path should be ['a', 'b']
   - But the Note component was still using ['b']
   - This caused actions like toggling the note or editing content to target the wrong location, resulting in invalid paths.
   
### Solution

- Fixed the issue by implementing a getCurrentPath function that intelligently determines the most up-to-date path at the moment of each operation. Rather than assuming the path prop is always valid, check for the most current path at the moment of each operation.


raineorshine:
Thanks for the PR. 

I think we have to answer the question of why the path is stale. When the thought is indented, `state.thoughts` changes, triggering recalculation of all selectors. If the problem is stale props, you usually have to look up the component hierarchy to find the source of the problem.

karunkop:
> Thanks for the PR.
> 
> I think we have to answer the question of why the path is stale. When the thought is indented, `state.thoughts` changes, triggering recalculation of all selectors. If the problem is stale props, you usually have to look up the component hierarchy to find the source of the problem.

Upon further investigation, I identified the root cause of the stale props issue. When a thought was indented moving from the path ['b'] to ['a', 'b']. React preserved component instances based on their position in the component tree rather than their underlying data. As a result, the Note component retained its initial path prop (['b']), even though the corresponding thought had shifted within the hierarchy.

The issue was further exacerbated by the React.memo optimization applied to the Note component. When the component was first created, all event handlers and effects captured the initial path prop value within their closures (e.g., ['b']). These closures persisted for the lifetime of the component instance, preventing updates to the path value after indentation.

Additionally, React.memo performs a shallow comparison of props, meaning it does not detect changes to the contents of an array if the reference remains unchanged. Consequently, even when the parent ThoughtContainer re-rendered, the memoized Note component failed to update, as React’s prop comparison mechanism determined that "nothing had changed."

### The Fix:

The key={hashPath(path)} solution effectively resolves the stale props issue by ensuring proper component re-creation. When the key changes, React treats the component as entirely new, breaking its identity and forcing a full unmount and re-render with the correct path prop. This prevents stale closures, as the new instance captures fresh references to the updated props in its event handlers and effects. This approach ensures that when a thought’s path changes, its associated Note component is properly recreated.

raineorshine:
Thanks, that makes more sense.

It's a little odd to force a remount with `key`. Usually `key` is used to differentiate siblings that are generated with `map`. It seems like there is still a stale `path` floating around in there, we might just covering up the problem by forcing a remount.

First, let's clarify if the problem is memoization or closure. Is the Note component re-rendering on indent? If so, then memoization is not the problem.

If the problem is memoization, then we may need to update the equality operator or ensure the `path` reference is changing.

If the problem is closure, then maybe you need a `useCallback` on the event handlers that update when `path` changes.

Thanks, and let me know what you find out!

raineorshine (changes requested):
Thanks, that makes more sense.

It's a little odd to force a remount with `key`. Usually `key` is used to differentiate the DOM identities of siblings (e.g. those generated by `map`). However, this Note actually _should_ have the same identity after indent. It seems like there is still a stale `path` floating around in there, we might just covering up the problem by forcing a remount.

First, let's clarify if the problem is memoization or closure. Is the Note component re-rendering on indent? If so, then memoization is not the problem.

If the problem is memoization, then we may need to update the equality operator or ensure the `path` reference is changing.

If the problem is closure, then maybe you need a `useCallback` on the event handlers that update when `path` changes.

Thanks, and let me know what you find out!

raineorshine:
Note to self: Whatever solution we end up with, make sure #2721 does not reoccur.

karunkop:
> Thanks, that makes more sense.
> 
> It's a little odd to force a remount with `key`. Usually `key` is used to differentiate the DOM identities of siblings (e.g. those generated by `map`). However, this Note actually _should_ have the same identity after indent. It seems like there is still a stale `path` floating around in there, we might just covering up the problem by forcing a remount.
> 
> First, let's clarify if the problem is memoization or closure. Is the Note component re-rendering on indent? If so, then memoization is not the problem.
> 
> If the problem is memoization, then we may need to update the equality operator or ensure the `path` reference is changing.
> 
> If the problem is closure, then maybe you need a `useCallback` on the event handlers that update when `path` changes.
> 
> Thanks, and let me know what you find out!

Digging deeper, the Note component was experiencing  a React closure issue. (Event handlers like onFocus were capturing the path prop value from when they were defined, not when they were executed.).

### Why useCallback Didn't Solve the Issue?

Initially,I tried using useCallback with path in the dependency array.

```ts
const onFocus = useCallback(() => {
  // Function that uses path
}, [path, dispatch]);
```


- The useCallback approach `might` have failed because it only recreates the function when the component re-renders and the dependencies change. If the component doesn’t re-render when path updates due to React.memo optimization or parent component behavior, the callback remains stale. Even when it does update, the timing may be too late, as operations needing the latest path might have already executed. Additionally, since the component was wrapped with React.memo, it didn’t re-render when only the path reference changed, further preventing the callback from being updated.

The new solution is inclined towards using ref. The ref pattern solves this issue by creating a mutable reference, unlike props (which are immutable in a given render cycle), refs provide a mutable .current property that persists across renders. This approach breaks the closure's hold on the stale value, ensuring handlers always access the latest path regardless of when they were defined.

raineorshine:
Thanks for the update. However, I'm not convinced that this is the correct solution. This would mean that any time React.memo is used, each prop needs to have its own ref, which seems incredibly awkward.

Could you do some additional research into the proper way to handle this in React? Others must have encountered this situation before.

@trevinhofmann Do you have any insight into this? Why do [8520b2c](https://github.com/cybersemics/em/pull/2860/files/8520b2c32490d192256420144792ee386f19602d) and [d04c7c9](https://github.com/cybersemics/em/pull/2860/commits/d04c7c928c248c450ae01825412fbca87f5ad2b8) seem wrong? Maybe I'm missing something, but my code smell spidey-sense is going off like crazy :).

trevinhofmann:
`React.memo` is a performance utility, preventing the component from re-rendering whenever the props have not changed (even if the parent is rerendered). The problem is that it does a shallow equality check for each prop. Trying to find where the `path` prop is originally defined is a challenging exercise in excessive prop drilling and other misdirection, but it seems evident from this solution that our problem is `path` gets mutated. In other words, it seems that the true "value" inside `path` changes, but the reference does not change. So, when `React.memo` compares the old and new `path` shallowly, it believes that it hasn't changed and does not re-render.

8520b2c attempts to fix this by using `key` to force the component to remount entirely whenever the deep value of `path` changes, but this is too heavy-handed. I believe the goal is to rerender without completely remounting (and losing whatever state the component instance had).

d04c7c9 attempts to fix this with a ref. To be quite honest, I'm not sure if/how it works, as the `useEffect` would also do a shallow comparison (and in theory, we aren't rerendering the component anyway to check...). I may be misunderstanding part of the problem/solution as it doesn't make sense to me, but, as you've already pointed out, this solution is concerning because it does not scale well. We'd need to do this for every prop in a `React.memo` component that can't be compared shallowly.

---

If my understanding of the problem is correct (`React.memo` prevents the Note from rerendering when `path` is mutated), then:

1. IMO, the ideal solution is to avoid mutating values (`path`). In my own projects, I take a fairly hard stance against mutation, as it's (a) a bit of extra complexity to grok and (b) can lead to issues like these.

2. An easier solution which should be better than the other two commits might be to provide the second parameter (`arePropsEqual`) to `React.memo`. We could use `hashPath` inside `arePropsEqual` to tell `React.memo` that it should rerender the Note if and only if the `path` value changes.

raineorshine:
Thanks for the analysis, that's very helpful. 

We should definitely avoid mutating paths. Mutation can explain the stale path here, though it might also be explained by invalid memoization. I traced `path` up the component hierarchy, and it comes from `childPath` in `VirtualThought` which is memoized in [appendToPathMemo](https://github.com/cybersemics/em/blob/4d07fb51d698ee548c1d8199668d52d5bf8a029b/src/util/appendToPath.ts). There are some comments in that file that indicate some existing issues with the memoization, so this may be related. Note that the `Thought` and `VirtualThought` components use `React.memo` and depend on the stable `path` references that `appendToPathMemo` provides to avoid excessive re-renders, so we can't remove memoization completely without introducing performance issues.

My recommendation would be to try [arePropsEqual](https://react.dev/reference/react/memo#parameters) as suggested, and then create a separate issue for getting to the bottom of the `appendToPathMemo` weirdness.

karunkop:
> `React.memo` is a performance utility, preventing the component from re-rendering whenever the props have not changed (even if the parent is rerendered). The problem is that it does a shallow equality check for each prop. Trying to find where the `path` prop is originally defined is a challenging exercise in excessive prop drilling and other misdirection, but it seems evident from this solution that our problem is `path` gets mutated. In other words, it seems that the true "value" inside `path` changes, but the reference does not change. So, when `React.memo` compares the old and new `path` shallowly, it believes that it hasn't changed and does not re-render.
> 
> [8520b2c](https://github.com/cybersemics/em/commit/8520b2c32490d192256420144792ee386f19602d) attempts to fix this by using `key` to force the component to remount entirely whenever the deep value of `path` changes, but this is too heavy-handed. I believe the goal is to rerender without completely remounting (and losing whatever state the component instance had).
> 
> [d04c7c9](https://github.com/cybersemics/em/commit/d04c7c928c248c450ae01825412fbca87f5ad2b8) attempts to fix this with a ref. To be quite honest, I'm not sure if/how it works, as the `useEffect` would also do a shallow comparison (and in theory, we aren't rerendering the component anyway to check...). I may be misunderstanding part of the problem/solution as it doesn't make sense to me, but, as you've already pointed out, this solution is concerning because it does not scale well. We'd need to do this for every prop in a `React.memo` component that can't be compared shallowly.
> 
> If my understanding of the problem is correct (`React.memo` prevents the Note from rerendering when `path` is mutated), then:
> 
> 1. IMO, the ideal solution is to avoid mutating values (`path`). In my own projects, I take a fairly hard stance against mutation, as it's (a) a bit of extra complexity to grok and (b) can lead to issues like these.
> 2. An easier solution which should be better than the other two commits might be to provide the second parameter (`arePropsEqual`) to `React.memo`. We could use `hashPath` inside `arePropsEqual` to tell `React.memo` that it should rerender the Note if and only if the `path` value changes.

The custom arePropsEqual function with hasPath still doesn’t work,and here’s why. This isn’t just a theoretical explanation; I’ve actually tried this approach in practice and confirmed it doesn’t work before switching to the ref-based solution.

### Why `arePropsEqual` fails with mutated path

**The Problem:**

If you mutate path (like path[1] = 'newValue') instead of creating a new array, arePropsEqual won’t catch the change. That’s because both prevProps.path and nextProps.path still point to the same reference, even if the content changed.

**What’s Going On:**

arePropsEqual uses hashPath(prevProps.path) === hashPath(nextProps.path). If the path is mutated in place, both hashes end up being the same—even though the data changed. So React skips the re-render.

Example:

```ts
let path = ['a', 'b'];   // reference: #123
path[1] = 'c';           // still #123, just new content
```
Now both prevProps.path and nextProps.path hash to ['a', 'c'], so arePropsEqual says, “Nothing changed,” when in fact it did.

**Why It Happens:**

React relies on reference changes to know when to re-render. Mutating path keeps the reference the same, so React is blind to the update.

**Temporary Ref Fix?**

Using ref.current works because it always has the latest path since it's updated every render. Using pathRef.current in onFocus avoids stale closures and ensures it gets the current path. This keeps the handler fresh and avoids relying on outdated props. But that’s a workaround, not a fix as the root cause is indeed mutation.

Since we've all agreed that the core issue is the mutation of path, and that enforcing immutability is the right long-term solution, perhaps we should focus on moving in that direction and fixing it accordingly. Let me know if I'm overlooking anything. @trevinhofmann @raineorshine

raineorshine:
Thank you. I see now that `arePropsEqual` can't be used as a workaround for a mutated prop.

Let's get to the bottom of the mutation/memoization issue. The ref solution is an inelegant workaround.

trevinhofmann:
> The custom arePropsEqual function with hasPath still doesn’t work

Ah! Took me a minute, but I see why that's an issue now. Thank you for the explanation, @karunkop! Both the old and new props passed to `arePropsEqual` would have the same mutated reference. Apologies for that suggestion; I'm not as used to dealing with mutated values as I used to be 😅

I'm also definitely in favor of enforcing immutability of the paths upstream to fix the root of the problem. We'll need to be wary of this and other potential performance issues, though:

> Note that the `Thought` and `VirtualThought` components use `React.memo` and depend on the stable `path` references that `appendToPathMemo` provides to avoid excessive re-renders, so we can't remove memoization completely without introducing performance issues.

karunkop:
Hi @trevinhofmann , @raineorshine ,

Following up on the tricky focus behavior in our Note component (where the stale path seemed to be used, especially after toggling notes), I wanted to share the detailed investigation process and the solution I have landed on.

**Mutations Issue:**

As discussed from our previous conversation, a very reasonable suspicion was that the problem might be caused by the path prop being accidentally mutated (changed internally) somewhere upstream, instead of being replaced with a completely new array when it changed. If this happened, React.memo  wouldn't see the change because the array reference would be the same, leading the Note component not to re-render and its event handlers (like onFocus) to use old, stale path data.

To check this, I did a thorough investigation:

**Component Hierarchy:** I traced the path prop backwards from Note -> Thought -> Subthought -> VirtualThought -> TreeNode -> LayoutTree.

**State Updates**: I examined the core Redux actions responsible for updating paths and the thought structure:
  - setCursor
  - setDescendant
  - deleteAttribute
  - deleteThought
  - updateThoughts

**Utility Functions**: I checked key helper functions like appendToPath, mergeUpdates, simplifyPath, thoughtToPath, etc., to ensure they create new arrays/objects instead of modifying existing ones.

After the thorough investigation, i found out that immutability mostly holds and the core logic for handling paths and state updates generally seems to follow immutability principles correctly (using spreads ..., slice, etc., to create copies).

**The Root Cause: A Subtle Timing/Closure Issue**

Source: https://www.oscarbustos.dev/blog/react-closures-traps/

Detailed logging revealed something unexpected:

image

The Note component was re-rendering when the path value changed. A new `onFocus` handler function was being created in that re-render, correctly capturing the new path. But, when the browser actually ran that `onFocus` handler a moment later (because browser events are asynchronous), the path variable inside the handler still sometimes resolved to the value from the previous render cycle.

This strongly suggests the root cause isn't a simple mutation or a failure of React.memo. It's more likely a subtle interaction between React's render scheduling and the browser's event loop timing. There seems to be a tiny window where the handler function reference is updated, but when the browser invokes it, the closure somehow accesses a slightly stale snapshot of the path variable from the moment just before the render fully committed or settled.

**The Solution: useFreshCallback Hook**

Since directly relying on the closure wasn't consistently working due to this timing nuance, I've implemented a solution using a custom hook: useFreshCallback. We still use the reliable `onFocus` prop on the `ContentEditable` component to make sure the focus event is always caught. The function we pass to `onFocus` is created using `useFreshCallback`. It gives back a function that always looks the same to React (it's stable, like using useCallback with empty dependencies []). But, inside, it uses `useRef` method. When called, it always runs the very latest version of the logic we gave it, which it keeps updated using a `useEffect`. This effectively bypasses the closure timing issue. The `onFocus` prop gets a stable function, but the code that actually runs when focus happens is guaranteed to use the logic and data from the most recent render cycle.

This approach reliably fixes both the original stale path issue and ensures the focus event correctly using the intended path. While this fixes the Note component, the initial investigation was valuable, and we should remain observant about ensuring immutability throughout the codebase to prevent other potential issues.

raineorshine (approved):
The cause of the bug is that `react-contenteditable` does not re-attach event handlers when its props change like normal React components. It is an old library that was designed before React hooks, so it does not play nice with `useCallback`. Documented in https://github.com/lovasoa/react-contenteditable/issues/161.

The solution, as you have discovered, is to either force the component to re-render, or store the callback in a ref in order to bypass props. An essentially identical custom hook as `useFreshCallback` is described in https://github.com/lovasoa/react-contenteditable/issues/161#issuecomment-736053428.

The longer-term solution is to replace `react-contenteditable` with our own component.

Issue #2830, opened 2025-03-03
Sort dropdown


raineorshine:
## Steps to Reproduce

Click the Sort button in the toolbar.

## Current Behavior

On each click, it toggles through the three sort attributes:

- Sort Alphabetical
- Sort Descending
- None

## Expected Behavior

Replace the toggling behavior with a dropdown that contains two options:

- None
- Alphabetical

See the LetterCasePicker component for an example toolbar dropdown menu.

- [x] If current sort is None and Alphabetical is clicked, sort Alphabetical (asc).
- [x] Toggle Alphabetical between Ascending and Descending when clicked multiple times:
    - If current sort is Alphabetical (asc) and Alphabetical is clicked, sort Alphabetical (desc).
    - If current sort is Alphabetical (desc) and Alphabetical is clicked, sort Alphabetical (asc).
- [x] Render ↑ or ↓ at the beginning of the sort option depending on which is selected.
    - If current sort is ascending, render ↑.
    - If current sort is descending, render ↓.
- [ ] ~~Perform correct Lottie animation regardless of which sort option from/to.~~
  - [ ] ~~Make sure to not animate when setting sort to None (like current behavior)~~
- [x] Use dimmed Sort icon when no sort option is selected.
- [x] Match dropdown behavior of LetterCasePicker:
    - Click empty space to dismiss
    - Highlight active sort option
- [ ] Refactor to remove duplicate code between LetterCasePicker and SortPicker

## Testing

Add tests like LetterCasePicker.

PR #2868, opened 2025-03-26
Sort dropdown

RED-ROSE515:
This PR resolves the issue : Sort dropdown #2830

In this PR, I created a `SortPicker` component that users can sort by `Alphabetical` or any other types.
Currently, I created a new command to show `SortPicker` instead of using `toggleSortCommand`.

I will change `toggleSortCommand` after this PR.


raineorshine:
Looks like the yarn.lock file was changed unintentionally. Could you revert that change?

RED-ROSE515:
Okay, let me fix.

src/components/SortPicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+        setSortPreference({
+          simplePath,
+          sortPreference: { type, direction: 'Asc' },
+        }),
+      )
+    }
+  }
+
+  // Define the sort options to render
+  const sortOptions = [
```
raineorshine, line 132: This is pretty good. You have mostly eliminated the duplication and separated out the declaration from the component rendering. However, your abstractions are a bit confused. You have mixed up the **declarative* aspects of the options from the **runtime** aspects. The result is that you've cleaned up the code a bit, but you are using it like a supplementary variable or helper function rather than as a true division of labor.

1. The mixup of abstraction can be seen in the smaller bits of duplication that are peeking through: Both `isSelected` and `content` repeat the `type`. This should give you a clue that these properties are not orthogonal.
2. `content` has been defined as a wide open type, as if the option could display anything. This isn't a great representation of the actual situation, which is far more limited.

A purely declarative listing of the options should just contain the necessary information to differentiate them from each other:

- The sort type.
- If it supports a direction.

Everything else can be derived dynamically from this in the `SortOption` component, based on the runtime value of `sortPreference`. Pushing this logic *down* to `SortOption` will clean up the types, create a cleaner separation between declarative and runtime value, and hide the logic further down the component hierarchy.

See if you can refactor this according to the above guidelines. Let me know if you have any questions.

src/components/SortPicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+    e.stopPropagation()
+    e.preventDefault()
+
+    if (!cursor || isRoot(cursor) || !simplePath) return
+
+    // Close the dropdown after selection
+    dispatch(toggleSortPicker({ value: false }))
+
+    // If clicking on 'None', reset to manual sorting
+    if (type === 'None') {
```
raineorshine, line 100: All three of these branches end up calling `setSortPreference(simplePath, ...)`. Instead of separating them out like this, take advantage of inline ternary expressions to make a single call to `setSortPreference` with the appropriate values. 

The way this is written, it gives the impression that there are three different branches that do three totally different things, which is not the case. It obscures the similarity and increases the amount of code. In general you want to push conditions *down* as far as possible to consolidate shared functionality and decrease code duplication.

This structuring is indicative of thinking of code in terms of procedural steps (1, 2, 3, ...) rather than functions with a domain + range ([a1, a2] → [b1, b2, b3]).

src/components/SortPicker.tsx
```diff
@@ -0,0 +1,190 @@
...
+import TriangleDown from './TriangleDown'
+
+// Define the type for sort options
+type SortType = 'None' | 'Alphabetical'
+
+// Default sort preference object
+const DEFAULT_SORT_PREFERENCE = { type: 'None', direction: null }
+
+/** Simple selector function to get sort preference. */
+const getSortPreferenceSelector = (state: State, cursor: Path | null) => {
```
raineorshine, line 28: This selector doesn't really need to be factored out, and adds unnecessary complexity. It's small and it's only used in one place, so it would be better inlined. Plus, there's no point passing `state.cursor` as a parameter when `state` is already passed.

raineorshine (changes requested):
Thanks, this update is a big improvement!

I have some additional comments below. I think there is a great learning opportunity here, as the code in question highlights some fundamental software design principles.

raineorshine (changes requested):
Thanks, I think the overall architecture looks solid now. Just some minor revisions are needed in order to have more appropriate comments.

src/components/SortPicker.tsx
```diff
@@ -0,0 +1,180 @@
...
+    const newSortPreference = {
+      type,
+      direction:
+        type === 'None'
+          ? null
+          : sortPreference.type === type
+            ? sortPreference.direction === 'Asc'
+              ? ('Desc' as const)
+              : ('Asc' as const)
+            : ('Asc' as const),
```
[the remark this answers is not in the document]
RED-ROSE515: There is typecheck error if I do like that.
So, I fixed like this.

`    const newSortPreference = {
      type,
      direction:
        type === 'None' ? null : sortPreference.type === type && sortPreference.direction === 'Asc' ? 'Desc' : 'Asc',
    } as SortPreference`

RED-ROSE515:
@raineorshine , Thanks for your detailed feedbacks.
It was a great experience.

raineorshine (approved):

raineorshine:
Thanks for the changes. Looks great.

Now that we separated the declarative data from the computed data, we can see that `SortOptionConfig` doesn't do much besides restate some of the `SortOption` props. We can move `option.type` and `option.supportsDirection` into the props themselves and convert `sortOptions.map` to declarative JSX. This allows us to eliminate `SortOptionConfig` and co-locate the values with the JSX itself. 

So in a way this is closer to the code that you started with, just with the computed properties moved into the `SortOptions` component itself. `map` is also a reasonable approach, and we would likely turn back to it if we ever wanted to fully factor out the sort options across `SortPicker` and `toggleSort`.

PR #2875, opened 2025-03-30
Tips v1.5

yangchristina:
@raineorshine we do want a max-width for tip right? Do we also want it on alert? max-width should be 529px from the mocks
Should max-width depend on font-size or should it always be the same?

image
Screenshot 2025-03-29 at 6 49 35 PM


src/components/PopupBase.tsx
```diff
@@ -119,6 +118,7 @@ const PopupBase = React.forwardRef<HTMLDivElement, PopupBaseProps>(
           boxSizing: 'border-box',
           textAlign,
           zIndex: 'popup',
+          maxWidth: '100%',
```
yangchristina, line 121: I don't see any case where we'd want it to overflow the screen?

raineorshine:
> @raineorshine we do want a max-width for tip right? Do we also want it on alert? max-width should be 529px from the mocks
> Should max-width depend on font-size or should it always be the same?

Yes, let's have a max-width for the tip and alert that scales with font size. Let's say `30em`.

yangchristina:
@trevinhofmann Would appreciate a review!

https://github.com/user-attachments/assets/f39c1bf5-8e60-48e4-90d9-5add2b6ab2b4

trevinhofmann (changes requested):
Thank you for the PR, @yangchristina! Looking great overall!

> Also the ControlledAlert name probably isn't great, if you have a better name please let me know.

At first, I thought we might want to rename `<ControlledAlert>` to `<Alert>`, and then change the name of our current `<Alert>` component to something else. We'd also want to then rename the `alertStore` accordingly.

After thinking a bit more, though, I would be in favor of renaming `<ControlledAlert>` to `<Notification>`. That seems like an apt name, as there are multiple types of notifications we may want to give the user. Currently, that would be tips (`<Tip>`) and alerts (`<Alert>`) which each reuse the more generic notification (`<Notification>`). Personally, I like that abstraction.

---

## 1. Cursor pointer on lightbulb icon

### Current Behavior

We hovering over the lightbulb icon, the cursor changes to `pointer` which indicates that the icon is clickable. Clicking the icon does nothing, however.

### Expected Behavior

I would expect the default cursor when hovering over the lightbulb icon (unless there _is_ supposed to be some action when clicking it).

---

## 2. The Tip component is a bit too wide on mobile

### Current Behavior

<details>
<summary>Demo</summary>

tips1
</details>

### Expected Behavior

The `maxWidth` might need to be reduced slightly. As it is, the (x) close button is cut off on the right edge of the screen.

Alternatively, should we be showing the (x) close button on mobile? IIRC, we had it disabled for alerts on mobile, since the expected UX is to swipe rather than click a button in order to dismiss.

---

## 3. The Tip component is hidden with the keyboard open (Android / Chrome)

### Current Behavior

When testing on Android (Chrome), the Tip component gets hidden by the keyboard. It appears to be working as expected on iOS Safari.

<details>
<summary>Demo</summary>

https://github.com/user-attachments/assets/ec845e3b-110e-458c-9c0f-d59bb8acf925
</details>

### Expected Behavior

The Tip component should be rendered above the keyboard, as it is on iOS Safari.

yangchristina:
Thanks @trevinhofmann
1. Fixed [4d059e5](https://github.com/cybersemics/em/pull/2875/commits/4d059e50031b4737dc7d652507dfcc4fb6398f1a)
2. fixed [65e57b3](https://github.com/cybersemics/em/pull/2875/commits/65e57b368d0b5fd5502878de0e3748a04f8cd979)

yangchristina:
@trevinhofmann for (3), is this also a problem for alert? They are using the same component so should act the same way. Was this solved for Alert in the Alert PR? If so, can I see how you did it? Thanks!

src/components/Tips/Tip.tsx
```diff
@@ -1,44 +1,43 @@
...
-  )
+  const dispatch = useDispatch()
+
+  const onClose = useCallback(() => {
+    dispatch(dismissTip())
+  }, [dispatch])
+
+  if (!display) return null
+
+  return <Notification renderedIcon={icon} transitionKey={`${display}`} value={children} onClose={onClose} />
```
trevinhofmann, line 40: Since the `transitionKey` prop is only a stringified boolean instead of a unique key, this is likely the cause of (4).

src/components/Notification.tsx
```diff
@@ -0,0 +1,71 @@
...
+import FadeTransition from './FadeTransition'
+import PopupBase from './PopupBase'
+
+/** An alert component in which you can customize what is rendered. Used for Alerts + Tips. */
+const Notification: FC<
+  {
+    renderedIcon?: ReactNode
+    value: ReactNode | null
+    /* Specify a key to force the component to re-render and thus recalculate useSwipeToDismissProps when the alert changes. Otherwise the alert gets stuck off screen in the dismiss state. */
+    transitionKey: string | number
```
trevinhofmann, line 14: This solution feels a bit hacky, but should be acceptable if we at least ensure unique keys are being passed. See (4).

src/components/Notification.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { ComponentProps, FC, ReactNode, useCallback, useRef, useState } from 'react'
+import { TransitionGroup } from 'react-transition-group'
+import { css } from '../../styled-system/css'
+import { token } from '../../styled-system/tokens'
+import FadeTransition from './FadeTransition'
+import PopupBase from './PopupBase'
+
+/** An alert component in which you can customize what is rendered. Used for Alerts + Tips. */
```
trevinhofmann, line 8:
```suggestion
/** A popup component in which you can customize what is rendered. Used for Alerts + Tips. */
```

Perhaps more clear?

trevinhofmann (changes requested):

(1) & (2) have both been resolved.

For (3), this does seem to also be a problem for Alerts on the `main` branch. In that case, we can likely treat that as a separate issue and not resolve it with this PR.

---

Some questions:

* When there is both an Alert and a Tip, the Alert is displayed behind the Tip (closing the Tip reveals the Alert). I'm guessing this is acceptable, but is there anything else we want to do when both are present? cc @raineorshine
* Regarding (2), should the "x" close button be hidden on mobile as it is for alerts? cc @raineorshine
* It would be nice to have puppeteer tests to capture the appearance of these notifications, but that might be outside the scope of this PR. @yangchristina, do you know if that would be in the scope of this work? I can't find the issue to check if it mentions this.

---

## 4. On mobile, subsequent tips do not appear

Likely related to the `transitionKey` prop, no further tips appear on mobile after dismissing a tip.

### Steps to Reproduce

1. Click `New thought` button
2. Swipe down to dismiss the tip
3. Click `New thought` button

### Current Behavior

The tip does not reappear.

### Expected Behavior

Subsequent tips should continue to appear and function.

raineorshine:
> * When there is both an Alert and a Tip, the Alert is displayed behind the Tip (closing the Tip reveals the Alert). I'm guessing this is acceptable, but is there anything else we want to do when both are present? cc @raineorshine

For now, let's ensure that only a single Alert or Tip is displayed at a time. Eventually we may add the ability to stack multiple alerts. i.e. Let's follow the current behavior of Alert where a new Alert/Tip replaces the old one.

> * Regarding (2), should the "x" close button be hidden on mobile as it is for alerts? cc @raineorshine

Yes, there's no difference between the overall UI. A Tip is just an Alert with a lightbulb icon.

yangchristina:
Thank you @trevinhofmann

 4. On mobile, subsequent tips do not appear
I have run changed the transitionKey, but haven't had a chance to run it on a mobile simulator yet: [5fafccb](https://github.com/cybersemics/em/pull/2875/commits/5fafccbd5c946575f0ac29d763d989ba7a7113a0)

I don't understand why transitionKey is even needed for Tips actually, considering this line:
`if (tip !== tipId) return null` (formerly `if (!display) return null`)

Does the key actually do anything in this case?

@raineorshine 
> For now, let's ensure that only a single Alert or Tip is displayed at a time. Eventually we may add the ability to stack multiple alerts. i.e. Let's follow the current behavior of Alert where a new Alert/Tip replaces the old one.

Changed here: [74908d9](https://github.com/cybersemics/em/pull/2875/commits/74908d97759d66a628d8dfad083382b987a7a243)
Note that tips were formally allowed to stack, so this changes behaviour there.

yangchristina:
> > * Regarding (2), should the "x" close button be hidden on mobile as it is for alerts? cc @raineorshine
> 
> Yes, there's no difference between the overall UI. A Tip is just an Alert with a lightbulb icon.

implemented here: [c37ab00](https://github.com/cybersemics/em/pull/2875/commits/c37ab006c966644072f152d29236108bfc41529f)

yangchristina:
Also @trevinhofmann puppeteer tests are not in scope

src/actions/showTip.ts
```diff
@@ -1,17 +1,21 @@
...
 import Thunk from '../@types/Thunk'
 import TipId from '../@types/TipId'
+import { alertActionCreator } from './alert'
 
 /** Shows a popup with a tip. */
 const showTip = (state: State, { tip }: { tip: TipId }): State => ({
   ...state,
-  tips: state.tips.includes(tip) ? state.tips : [...state.tips, tip],
+  // for now, don't allow stacking tips
+  tips: [tip],
```
trevinhofmann, line 10: If we will not be allowing a stack of tips (and only showing the latest tip), it would be good to simplify the type from `tips: Tip[]` to `tip: Tip`.

src/actions/alert.ts
```diff
@@ -23,7 +24,7 @@ let clearAlertTimeoutId: ReturnType<typeof setTimeout> | null = null
 const alertReducer = (state: State, { alertType, showCloseLink, value, importFileId }: Options) => {
   if (value === state.alert?.value) return state
   return {
-    ...state,
+    ...(value ? dismissTip(state) : state),
```
trevinhofmann, line 27: In `src/actions/showTip.ts`, we clear alerts by dispatching an alert action in the action creator.
In `src/actions/alerts.ts`, we clear tips here by directly using `dismissTip` in the reducer.

Can we choose one (dispatching an action, or updating it in the reducer) and use it consistently across these two files? I think I'm partial to dispatching an action, but would be okay with either.

src/components/Notification.tsx
```diff
@@ -0,0 +1,73 @@
...
+  {
+    renderedIcon?: ReactNode
+    value: ReactNode | null
+    /* Specify a key to force the component to re-render and thus recalculate useSwipeToDismissProps when the alert changes. Otherwise the alert gets stuck off screen in the dismiss state. */
+    transitionKey: string | number
+    testId?: string
+  } & Pick<ComponentProps<typeof PopupBase>, 'onClose' | 'showXOnHover'>
+> = ({ renderedIcon, onClose, value, transitionKey, testId, ...props }) => {
+  const popupRef = useRef<HTMLDivElement>(null)
+  const [isDismissed, setDismiss] = useState(false)
```
trevinhofmann, line 20: For consistency, I recommend strictly following the `[value, setValue]` convention without alterations (i.e. `const [isDismissed, setIsDismissed] = useState(false)`).

However, we already have this exact naming in the CommandPallette, so this is probably okay as-is. Perhaps we could rename both of these and any others later in a separate PR): https://github.com/cybersemics/em/blob/main/src/components/CommandPalette.tsx#L506

trevinhofmann (changes requested):
Thank you for the updates, @yangchristina! Looking great overall. (4) appears to be resolved.

Regarding (2), it has been resolved by removing the (x) close button on mobile. In my opinion, the tips would still look better with a little bit of left/right margin on mobile. What are your thoughts?

Screenshot 2025-04-11 at 2 23 57 PM

src/actions/showTip.ts
yangchristina: @raineorshine thoughts on this?

raineorshine: I don't have a strong preference as we try to get this PR merged, but in general I prefer building tightly around the current functionality and then refactoring later when additional functionality is added.

src/actions/alert.ts
raineorshine: For what's worth, I like to do as much as possible in the reducer, and just use action-creators for side effects. Reducers are pure functions, composable, more easily testable, and not affected by action queueing order.

yangchristina: Sounds good, done here: [0369d3c](https://github.com/cybersemics/em/pull/2875/commits/0369d3c38fb7dc33ad46736febc82026a6c07187)

src/components/Notification.tsx
yangchristina: Done here: 9caa00e

src/actions/showTip.ts
yangchristina: Done here: [0369d3c](https://github.com/cybersemics/em/pull/2875/commits/0369d3c38fb7dc33ad46736febc82026a6c07187)

yangchristina:
Ah, I missed the part on the left and right margins. Yes, it would definitely look better with just a bit of margin, I will add some

yangchristina:
@trevinhofmann added some spacing from edge here: 223fd93, how does it look?

trevinhofmann (changes requested):
Thank you for the updates, @yangchristina. I think we're very close!

However, I am again seeing [issue (4)](https://github.com/cybersemics/em/pull/2875#pullrequestreview-2748830728) when testing on mobile.

When switching between `New thought` and `New subthought`, the tip always appears. However, when triggering the same tip twice (on mobile) after swiping the first one away, the second tip does not appear.

src/components/Tips/NewThoughtTip.tsx
```diff
@@ -1,84 +1,46 @@
...
         color={token('colors.gray66')}
+        cssRaw={css.raw({ verticalAlign: 'middle' })}
       />
     </span>
   ) : (
-    `You can add a new thought by pressing ${returnKey} on the keyboard`
+    <span>
+      You can add a new thought by pressing {returnKey} on the keyboard. You can customize the toolbar in{' '}
+      <a
+        {...fastClick(() => {
```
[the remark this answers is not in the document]
yangchristina: sounds good, thanks for letting me know

src/components/Tips/Tip.tsx
```diff
@@ -1,43 +1,45 @@
...
-          maxWidth: '20em',
-          borderRadius: '5',
-          textAlign: 'center',
-          color: 'codeBgInverse',
-        })}
-      >
-        {children}
-      </div>
-    </div>
+    <Notification renderedIcon={icon} transitionKey={tipId} value={tip === tipId ? children : null} onClose={onClose} />
```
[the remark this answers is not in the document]
yangchristina: I'm not able to reproduce this on android emulator. Also not sure why you weren't getting it before and getting it now, since I don't see anything in the changes

https://github.com/user-attachments/assets/d83c74af-abee-4b2d-a65b-6017185c4246

I haven't been able to test it on ios simulator so far, having troubling running it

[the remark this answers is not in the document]
raineorshine: > ## 4. On mobile, subsequent tips do not appear
> Likely related to the `transitionKey` prop, no further tips appear on mobile after dismissing a tip.
> 
> ### Steps to Reproduce
> 1. Click `New thought` button
> 2. Swipe down to dismiss the tip
> 3. Click `New thought` button
> 
> ### Current Behavior
> The tip does not reappear.
> 
> ### Expected Behavior
> Subsequent tips should continue to appear and function.

I am able to reproduce in Chrome Device Mode and Mobile Safari. The trick is that you have to click the New Thought button very quickly after dismissing the tip.


https://github.com/user-attachments/assets/b9af4fbe-ca35-4095-94bb-09511a2b002a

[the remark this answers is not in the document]
raineorshine: Hi @yangchristina. Were you able to reproduce this with the provided instructions?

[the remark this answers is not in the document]
yangchristina: Hi @raineorshine, I'm so sorry for finishing this so late. Yes, I have reproduced, and fixed. This was the fix: [98dd8ab](https://github.com/cybersemics/em/pull/2875/commits/98dd8ab692cf0bdfa99cc37890106c1b6a9505fc)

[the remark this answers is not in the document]
yangchristina: Basically `useSwipeToDismiss` was not resetting the `dy`, so the tip was still mounted, but could not be seen.

trevinhofmann:
@yangchristina - There was a failing puppeteer test, but it is now passing after being rerun (flaky test issue) 👍

Can you please resolve the merge conflicts in `NewSubthoughtTip.tsx` and `NewThoughtTip.tsx`? I'll be happy to take a look after that is resolved. Thank you!

yangchristina:
@trevinhofmann merge conflicts dealt with. Thank you!

src/components/Tips/NewThoughtTip.tsx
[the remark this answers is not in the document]
trevinhofmann: That change has been paused, so no worries about conflicts: https://github.com/cybersemics/em/pull/2886#pullrequestreview-2793497052

trevinhofmann (approved):
Thank you for the updates, @yangchristina! All issues seem to be resolved, and the tip component is working great when testing locally.

src/actions/showTip.ts
```diff
@@ -3,17 +3,20 @@ import Thunk from '../@types/Thunk'
...
 import { registerActionMetadata } from '../util/actionMetadata.registry'
 
-/** Shows a popup with a tip. */
+/** Shows a popup with a tip and clears any existing alert. */
 const showTip = (state: State, { tip }: { tip: TipId }): State => ({
   ...state,
-  tips: state.tips.includes(tip) ? state.tips : [...state.tips, tip],
+  tip,
+  // clear any existing alert when showing a tip
+  ...(tip && { alert: null }),
```
raineorshine, line 11: This condition is always true since `tip` is always truthy.

src/colors.config.ts
```diff
@@ -37,6 +37,7 @@ const colors = {
     breadcrumbs: 'rgba(153, 153, 153, 1)',
     activeBreadCrumb: 'rgba(144, 144, 144, 1)', // #909090
     link: 'rgba(135, 206, 235, 1)', // #87ceeb
+    lightbulbIconBg: 'rgba(135, 206, 235, 0.2)',
```
raineorshine, line 40: Let's make this slightly more general and call it `tipIconBg` or `popupIconBg`.

src/components/Notification.tsx
```diff
@@ -0,0 +1,75 @@
...
+import PopupBase from './PopupBase'
+
+/** A popup component in which you can customize what is rendered. Used for Alerts + Tips. */
+const Notification: FC<
+  {
+    renderedIcon?: ReactNode
+    value: ReactNode | null
+    /* Specify a key to force the component to re-render and thus recalculate useSwipeToDismissProps when the alert changes. Otherwise the alert gets stuck off screen in the dismiss state. */
+    transitionKey: string | number
+    testId?: string
```
raineorshine, line 16: No need to parameterize `testId` since only `'alert'` is used.

src/components/Notification.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { token } from '../../styled-system/tokens'
+import { isTouch } from '../browser'
+import FadeTransition from './FadeTransition'
+import PopupBase from './PopupBase'
+
+/** A popup component in which you can customize what is rendered. Used for Alerts + Tips. */
+const Notification: FC<
+  {
+    renderedIcon?: ReactNode
+    value: ReactNode | null
```
raineorshine, line 13: What's the difference between `value` and `children`? React components typically have one or the other, but not both. 

Can these be combined? If so, that would reduce complexity. If not, solid JSDOC comments are needed to inform other developers of the recommended usage.

src/components/Notification.tsx
```diff
@@ -0,0 +1,75 @@
...
+    /* Specify a key to force the component to re-render and thus recalculate useSwipeToDismissProps when the alert changes. Otherwise the alert gets stuck off screen in the dismiss state. */
+    transitionKey: string | number
```
raineorshine, line 15: I don't love that `transitionKey` is now part of the API of this component, as it seems like more of an implementation detail. The previous implementation at least kept it internal. However, I don't see a way to generate a serializable key when a `ReactNode` is passed instead of a string value.

src/components/Notification.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { css } from '../../styled-system/css'
+import { token } from '../../styled-system/tokens'
+import { isTouch } from '../browser'
+import FadeTransition from './FadeTransition'
+import PopupBase from './PopupBase'
+
+/** A popup component in which you can customize what is rendered. Used for Alerts + Tips. */
+const Notification: FC<
+  {
+    renderedIcon?: ReactNode
```
raineorshine, line 12: Let's call this `icon`.

src/components/PopupBase.tsx
```diff
@@ -107,6 +109,8 @@ const PopupBase = React.forwardRef<HTMLDivElement, PopupBaseProps>(
           boxSizing: 'border-box',
           textAlign,
           zIndex: 'popup',
+          // leave space so the circledCloseButton doesn't get cut off from the screen
+          maxWidth: preventFromTouchingEdge || (onClose && circledCloseButton) ? 'calc(100% - 2em)' : '100%',
```
raineorshine, line 113: Would either `showXOnHover` or `circledCloseButton` be sufficient for this condition? It would be nice to avoid adding another prop.

raineorshine (changes requested):
Thank you!

src/components/Notification.tsx
yangchristina: done

src/components/Notification.tsx
yangchristina: fixed after rebase

src/colors.config.ts
yangchristina: renamed to `popupIconBg` in [530e614](https://github.com/cybersemics/em/pull/2875/commits/530e614ad562a33f6c34a7accc6bf68e7223dd7a)

src/actions/showTip.ts
yangchristina: removed, thanks!

src/components/PopupBase.tsx
yangchristina: `onClose && circledCloseButton` is false when `isTouch` is true, since we don't show the `x` on mobile. 

One thing we can do is remove the condition altogether and just use `'calc(100% - 2em)'` if we also want this behaviour on Command palette, that way everywhere using PopupBase would have this as true.

Should command palette also leave some space from the edge on small screens?

src/components/Notification.tsx
yangchristina: Children was added to support the below code which was recently added to Alert, I should rename it to something other than children. What would be a good name `additionalPopupContent`?
```
      {alert?.importFileId && (
        <a
          onClick={() => {
            deleteResumableFile(alert.importFileId!)
            syncStatusStore.update({ importProgress: 1 })
            onClose?.()
          }}
        >
          cancel
        </a>
      )}

      {multicursor && (
        <a
          className={cx(anchorButtonRecipe(), css({ margin: '0 1em 1em' }))}
          {...fastClick(() => {
            dispatch(clearMulticursors())
            onClose?.()
          })}
        >
          Cancel
        </a>
      )}
```

src/components/Notification.tsx
yangchristina: I don't love this either, but tips do take in an jsx.element instead of something serializable. Not sure how else this can be handled

raineorshine (approved):

src/components/PopupBase.tsx
raineorshine: > `onClose && circledCloseButton` is false when `isTouch` is true, since we don't show the `x` on mobile.

I'm not quite following you. `preventFromTouchingEdge` looks to be applied in the same cases that `circledCloseButton` and `showXOnHover` is applied. So, it might make sense to combine them. I'd like avoid getting over-granular with the props when this component is only used in two places. Since the extra space at the edge is directly related to the close button, it seems acceptable to tie it to one of the existing props that controls the close button. The props are either not orthogonal, or, at the very least, we're not using all the combinations.

> One thing we can do is remove the condition altogether and just use `'calc(100% - 2em)'` if we also want this behaviour on Command palette, that way everywhere using PopupBase would have this as true.
> 
> Should command palette also leave some space from the edge on small screens?

The desktop CommandPalette needs to be rendered flush to the right edge of the screen, so we can't apply the extra space in that case.

yangchristina: Sounds good! The main reason I added it was this comment: https://github.com/cybersemics/em/pull/2875#pullrequestreview-2761338279

> In my opinion, the tips would still look better with a little bit of left/right margin on mobile. What are your thoughts?

Is this not something we want?

src/components/Notification.tsx
yangchristina: Thanks for adding comments!

src/components/PopupBase.tsx
raineorshine: The behavior is correct, it's just a question of the best implementation. In my estimation, there's no need to add a new prop when `circledCloseButton` already uniquely identifies the case where  the extra space is needed.

PR #2877, opened 2025-03-30
Refactor pickers: Create shared Popover component

yangchristina:
@raineorshine the exports design hasn't been finalized yet, but I thought I'd create the Popover shared component first.

Should I rename this to Picker instead of Popover?

Screenshot 2025-03-30 at 4 31 40 PM
Screenshot 2025-03-30 at 4 31 35 PM
Screenshot 2025-03-30 at 4 31 55 PM

### DIFF
--- a/src/components/icons/SortWithPicker.tsx
+++ b/src/components/icons/SortWithPicker.tsx
@@ -1,40 +1,16 @@
-import { useRef } from 'react'
 import { useSelector } from 'react-redux'
-import { css } from '../../../styled-system/css'
 import IconType from '../../@types/IconType'
-import FadeTransition from '../FadeTransition'
 import SortPicker from '../SortPicker'
 import SortIcon from './Sort'
 
-/** Sort Icon Component with popup SortPicker. */
+/** Sort Icon Component with popup Picker. */
 const SortWithPicker = ({ size = 18, style, cssRaw }: IconType) => {
   const showSortPicker = useSelector(state => state.showSortPicker)
-  const toolbarPopupRef = useRef<HTMLDivElement>(null)
 
   return (
     <div>
       <SortIcon size={size} style={style} cssRaw={cssRaw} animated={showSortPicker} />
-      <FadeTransition duration='fast' nodeRef={toolbarPopupRef} in={showSortPicker} exit={false} unmountOnExit>
-        <div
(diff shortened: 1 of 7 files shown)


raineorshine (approved):
Great refactor, thanks.

> Should I rename this to Picker instead of Popover?

Your call, I have no preference.

yangchristina:
> Great refactor, thanks.
> 
> > Should I rename this to Picker instead of Popover?
> 
> Your call, I have no preference.

Let's keep it as Popover then. Picker is more generic.

src/components/icons/SortWithPicker.tsx
```diff
@@ -1,40 +1,16 @@
-import { useRef } from 'react'
 import { useSelector } from 'react-redux'
-import { css } from '../../../styled-system/css'
 import IconType from '../../@types/IconType'
-import FadeTransition from '../FadeTransition'
 import SortPicker from '../SortPicker'
 import SortIcon from './Sort'
 
-/** Sort Icon Component with popup SortPicker. */
+/** Sort Icon Component with popup Picker. */
```
trevinhofmann, line 6: For the comments in these three icon files, we now have:

```
/** Letter Case Icon Component with popup Picker. */
/** Sort Icon Component with popup Picker. */
/** Text Color Icon with popup ColorPicker. */
```

The first two have switched to `Picker` while the last is still `ColorPicker`. For consistency, it should probably either be:

```
/** Letter Case Icon component with popup Picker. */
/** Sort Icon component with popup Picker. */
/** Text Color Icon component with popup Picker. */
```

Or:

```
/** Letter Case Icon component with popup LetterCasePicker. */
/** Sort Icon component with popup SortPicker. */
/** Text Color Icon component with popup ColorPicker. */
```

trevinhofmann (approved):
Thanks for the pull request, @yangchristina! I love to see more lines removed than added in a PR 😄

I have a very minor (and non-blocking) question about JSDoc comments for the icons, otherwise this looks great to me!

src/components/icons/SortWithPicker.tsx
yangchristina: Thanks! Changed to: /** Text Color Icon component with popup Picker. */

Issue #2824, opened 2025-02-21
[Android] Edit mode persists after keyboard is closed


raineorshine:
## Steps to Reproduce

Close the keyboard using the built-in button. 

## Current Behavior

The browser selection is not removed and `state.editing` is still true. 

## Expected Behavior

The browser selection should be removed and `state.editing` should be set to false.

PR #2879, opened 2025-03-31
[Android] Edit mode persists after keyboard is closed 

karunkop:
PR for issue #2824 

### Problem

In the Android Chrome browser, the editing state persisted even after the virtual keyboard was dismissed due to the absence of a native keyboard dismissal event. As a result, the application was unable to synchronize its state, keeping the editing flag active despite the keyboard no longer being visible..

### Solution

Implemented a robust keyboard visibility detection mechanism using the Visual Viewport API. By tracking the initial viewport height and monitoring changes through the resize event, we detect significant height increases (above 15%) to determine keyboard dismissal.Upon detecting keyboard dismissal, we exit editing mode, clear selection, and update the application state.


raineorshine (changes requested):

This issue was reproduced in the Android Chrome browser, not the Capacitor build, so we should find a web solution.

> ### Problem
> On mobile devices, when users close the keyboard , the text remains selected instead of clearing the browser selection. Additionally, the application's state.editing remains true rather than updating to false to indicate that editing has ended. The expected behavior is for the browser selection to be removed upon keyboard dismissal and for state.editing to accurately reflect the end of editing.

This simply restates the problem as described in the original issue, without explaining *why* it occurs, or what the root cause is.

karunkop:
> Hi, thanks for the submission.
> 
> This issue was reproduced in the Android Chrome browser, not the Capacitor build, so we should find a web solution.
> 
> > ### Problem
> > On mobile devices, when users close the keyboard , the text remains selected instead of clearing the browser selection. Additionally, the application's state.editing remains true rather than updating to false to indicate that editing has ended. The expected behavior is for the browser selection to be removed upon keyboard dismissal and for state.editing to accurately reflect the end of editing.
> 
> This simply restates the problem as described in the original issue, without explaining _why_ it occurs, or what the root cause is.

Hi, I have implemented a web based solution using visual viewport api. I have also updated the problem and solution  statement in the PR itself!

raineorshine:
Thanks for the update. 

Before I consider this solution, I would like a better explanation of the cause relative to iOS Safari. Why does it work on iOS Safari but not Android?

Otherwise, we're at risk of having two different solutions for two different platforms without really understanding why that is necessary.

karunkop:
> Thanks for the update.
> 
> Before I consider this solution, I would like a better explanation of the cause relative to iOS Safari. Why does it work on iOS Safari but not Android?
> 
> Otherwise, we're at risk of having two different solutions for two different platforms without really understanding why that is necessary.

## **Why Blur Events Work Differently on iOS Safari/Chrome vs. Android Chrome**

When a user interacts with an input field on a webpage, the expected behavior is that once they’re done typing and dismiss the keyboard, the field should exit editing mode. This typically happens when the browser fires a blur event, which signals that the input field has lost focus. However, this doesn’t always work consistently across different browsers and operating systems.

## **How iOS Safari & Chrome Handle This (The Expected Behavior)**

On iOS, when the keyboard is dismissed (either by tapping outside the input field or by pressing the "Done" button), the operating system ensures that the input field immediately loses focus. This is because iOS maintains strict synchronization between:

- **Keyboard visibility (whether the software keyboard is shown or hidden)**
- **Focus state (whether an input field is actively receiving user input)**

So, when the keyboard disappears, iOS forces the active input field to blur (lose focus), which naturally triggers the blur event in JavaScript. This seamless handling ensures that applications exit editing mode correctly without requiring extra logic.

## **Why Android Chrome Behaves Differently (The Problem)**

Android Chrome, on the other hand, does not enforce this strict synchronization. This leads to inconsistent behavior when dismissing the keyboard:

1. Lack of a Reliable Event Bridge

  - On Android, the browser does not always fire a blur event when the keyboard is dismissed.
  
  - This means the input field can stay focused even after the keyboard is hidden, making it seem like the user is still in editing mode.

2. Fragmented Implementation Across Devices & Keyboards

  - Different Android devices and third-party keyboard apps (like Gboard, Samsung Keyboard, SwiftKey) handle keyboard dismissal differently.
  
  - Some keyboards trigger blur events correctly, while others do not, making behavior inconsistent across devices.

3. Focus State Persists Even After Keyboard Dismissal

  - Unlike iOS, where keyboard dismissal forces focus loss, Android allows an input field to remain focused even when the keyboard is no longer visible.
  
  - This causes unexpected issues where a user thinks they've exited editing mode, but the application still treats the input as active.

## **Conclusion**

The key difference between iOS and Android Chrome is how they handle focus state when the keyboard is dismissed. iOS enforces a strict rule where hiding the keyboard automatically blurs the active input, ensuring smooth interactions. Android Chrome, however, does not guarantee this behavior, leading to inconsistent focus retention issues across devices and keyboards. This is why web applications that rely on blur events for exiting editing mode work well on iOS but often require additional handling on Android. So the solution that is implemented should justify the reason. If we were to be more thorough about the redundant approach, we can explicitly impose the current fix for android browsers only.

raineorshine:
Thank you, however AI output is not authoritative, and is often factually incorrect. Feel free to use it to guide research, but please provide references and something concrete that would persuade me that this is true. Is this what people are saying on StackOverflow? Is it in the context of web development rather than native development? Are there any bugs logged with browser vendors? More context is needed. Otherwise, I have no choice but to assume that the machine just spit out something that sounded true but has no basis in reality.

karunkop:
> Thank you, however AI output is not authoritative, and is often factually incorrect. Feel free to use it to guide research, but please provide references and something concrete that would persuade me that this is true. Is this what people are saying on StackOverflow? Is it in the context of web development rather than native development? Are there any bugs logged with browser vendors? More context is needed. Otherwise, I have no choice but to assume that the machine just spit out something that sounded true but has no basis in reality.

While much of the reasoning is supported by AI, I have also reviewed this discussion on Chromium's issue tracker (https://issues.chromium.org/issues/41177736), which explores blur events triggered when the virtual keyboard is hidden.

As part of this research, I am still working on gaining a deeper understanding of the underlying behavior of iOS browsers. I also came across a relevant Stack Overflow discussion (https://stackoverflow.com/questions/5652322/how-to-prevent-a-js-blur-event-in-ios-webkit-when-the-virtual-keyboard-is-dismis), where the user seeks a solution to prevent a JavaScript blur event in iOS WebKit. While the discussion primarily focuses on workarounds, it also indirectly validates the default behavior of iOS platforms in handling blur events.

Let me know if you have any insights or additional resources that could be helpful in this investigation.

karunkop:
> Empirically, there must be some point where iOS Safari and Android Chrome deviate in behavior. Where exactly in the codebase does that happen? Knowing that would help the explanation considerably, changing it from a vague "events get missed in Android" to "here is the exact line number in em's codebase where Android behaves differently."
> 
> This isn't any different than normal bug troubleshooting, where we identify where the current and expected behavior deviate. We just have the benefit of having a reference platform that works as expected (iOS).
>

The fundamental difference lies in how these platforms handle blur events when the virtual keyboard closes:

### **iOS Safari:** 
When the keyboard is closed by pressing ios keyboard built in collapse button, a proper blur event is fired on the element, which triggers the onBlur handler in the Editable component (lines 428 & 477 in Editable.tsx). This handler sets editing: false and correctly exits edit mode. See the screenshot below:
image

### **Android Chrome:** 
When the keyboard is collpased, no blur event is fired on the element. This is the key divergence point. Without the blur event, the app remains in edit mode even though the keyboard is gone. See the screenshot below:
image

However if we tap outside the editable content, proper blur event is also fired in android chrome browser. See screenshot below:
image

To answer your specific questions:

> What event gets missed on what element?
- The blur event on the editable element (the one with data-editable attribute) is missed in Android Chrome

> What is the state of window.getSelection()?
- In Android Chrome, the selection remains on the editable element after keyboard dismissal window.getSelection().focusNode still points to the text node within the editable element while in ios it points to null.

> What is the state of document.activeElement?
- document.activeElement still references the editable element in Android Chrome. In iOS Safari, it would be reset to the document body after keyboard dismissal

> Does selectionchange fire?
- No, the selectionchange event doesn't fire in Android Chrome when the keyboard is dismissed without interaction with the text. However in ios it fires
image

The core issue in Android Chrome is that dismissing the on-screen keyboard without explicitly shifting focus does not trigger a blur event. As a result, the application incorrectly maintains the input field's focused state, creating a mismatch where the app remains in edit mode even though the keyboard is no longer visible. The current workaround using keyboard height detection (measuring visualViewport height changes) is indirect but necessary because the expected blur events don't occur. A more direct solution would require finding a way to reliably detect keyboard closure in Android Chrome and trigger the appropriate state changes. The specific line where the platforms diverge is in the behavior around the blur event should be triggered on the editable element when the keyboard is dismissed.

raineorshine (changes requested):
Great, thanks for the additional information. This helps to have more concrete points about the actual behavior in **em**.

> ### **Android Chrome:**
> When the keyboard is collpased, no blur event is fired on the element. This is the key divergence point. Without the blur event, the app remains in edit mode even though the keyboard is gone. See the screenshot below:

Can you confirm that the `blur` event does not fire at all, and it's not just short circuiting before dispatching `editingAction`? I assume you put your console.log at the very beginning of th handler, but I just want to be extra sure since there are three places it shortcircuits (`isRelatedTargetEditableOrNote`, `blurring`, and `selection.isThought()`). If any of those behave differently on Android, it could explain the behavior.

If it is truly a browser issue, and there is no event that is triggered when the virtual keyboard is closed, then I'm comfortable moving forward with the proposed approach.

karunkop:
> Great, thanks for the additional information. This helps to have more concrete points about the actual behavior in **em**.
> 
> > ### **Android Chrome:**
> > When the keyboard is collpased, no blur event is fired on the element. This is the key divergence point. Without the blur event, the app remains in edit mode even though the keyboard is gone. See the screenshot below:
> 
> Can you confirm that the `blur` event does not fire at all, and it's not just short circuiting before dispatching `editingAction`? I assume you put your console.log at the very beginning of th handler, but I just want to be extra sure since there are three places it shortcircuits (`isRelatedTargetEditableOrNote`, `blurring`, and `selection.isThought()`). If any of those behave differently on Android, it could explain the behavior.

Yes i can confirm that the `blur` event in android browser doesnot fire at all and it's not just short circuiting before dispatching `editingAction`.

src/util/keyboardVisibility.ts
```diff
@@ -0,0 +1,74 @@
...
+        // Exit editing mode when keyboard is closed
+        const state = store.getState()
+        if (state.editing && state.cursor) {
+          selection.clear()
+          store.dispatch(editingAction({ value: false }))
+        }
+      }
+    }
+    // check if height is decreased significantly (keyboard opened)
+    else if (currentHeight < previousHeight) {
```
raineorshine, line 65: This can be simplified to `else`. Since `heightChangeRatio` is positive, we know that `currentHeight !== previousHeight`. Thus, `currentHeight < previousHeight` is equivalent to `currentHeight <= previousHeight`, which is equivalent to `else`.

src/util/keyboardVisibility.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const currentHeight = visualViewport.height
+  const previousHeight = lastViewportHeight || initialViewportHeight
+
+  // Calculate height change ratio
+  const heightChangeRatio = Math.abs(currentHeight - previousHeight) / previousHeight
+
+  // Update last height
+  lastViewportHeight = currentHeight
```
raineorshine, line 47: If you update `lastViewportHeight` after the threshold check, you may be able to eliminate `previousHeight`. 

Conceptually, `lastViewportHeight` and `previousHeight` appear to represent the same thing. Is that true, or am I missing something?

raineorshine (changes requested):
The new structure looks great! Thanks for the refactor.

karunkop:
Thanks for your great insight and optimal solutions. I have refactored the code to account in all the suggested changes :)

raineorshine (changes requested):
Thanks, this looks much better.

This isn't necessary to change, but I will point out that we already have a [resize handler](https://github.com/cybersemics/em/blob/3c46df70190a1c977e9da92c2b9c5d7aae0d0b91/src/util/initEvents.ts#L380) and a dedicated [viewportStore](https://github.com/cybersemics/em/blob/main/src/stores/viewport.ts) that handles similar functionality, so it would not be unusual to move all of this logic to `viewport` and have it handled in one place. That said, there is a reasonable argument that it is better to keep them decoupled, in particular because `keyboardVisibility` has a dependency on the global `store`, but `viewport` does not.

src/util/keyboardVisibility.ts
```diff
@@ -0,0 +1,63 @@
...
+import { KEYBOARD_VISIBILITY_THRESHOLD } from '../constants'
+import * as selection from '../device/selection'
+import store from '../stores/app'
+
+let lastViewportHeight: number | null = window.visualViewport?.height || null
+
+/**
+ * Reset keyboard visibility tracker state.
+ */
+const reset = (): void => {
```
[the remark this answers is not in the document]
karunkop: I think you are right, the only reason i added this at the first place is to ensure proper cleanups of variable when we remove resize event listener.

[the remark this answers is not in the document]
raineorshine: Sure, I can see that. Technically is does need to be set back to null to match the initial state. That might be relevant in some rigorous testing environments. However, here it is probably unnecessary. Leaving `lastViewportHeight` equal to the last viewport height is probably more accurate anyway, since `null` is really just a placeholder for a yet-to-be measured height.

The appeal of removing it completely is that this module can export a single function, the event handler, and not need to export a singleton object at all.

src/util/keyboardVisibility.ts
```diff
@@ -0,0 +1,63 @@
...
+ 
+ * @returns A throttled function that should be attached to viewport resize events.
+ */
+const handleKeyboardVisibility = _.throttle(() => {
+  // Only apply this for touch devices  where the issue occurs
+  if (!isTouch || !window.visualViewport) {
+    return
+  }
+
+  const visualViewport = window.visualViewport
```
raineorshine, line 30: Reassigning `window.visualViewport` is unnecessary. Object property dereference is fast, and adding an additional identifier to the scope to save 7 characters is not worth it.

src/util/keyboardVisibility.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const currentHeight = visualViewport.height
+
+  // Calculate height change ratio
+  const heightChangeRatio = Math.abs(currentHeight - lastViewportHeight) / lastViewportHeight
+
+  // check if the change is significant to be considered a keyboard change
+  if (heightChangeRatio > KEYBOARD_VISIBILITY_THRESHOLD) {
+    // check if height increased significantly (keyboard closed)
+    if (currentHeight > lastViewportHeight) {
```
raineorshine, line 46: Nested `if` statements with no `else` should be combined into a single `if` statement with `&&`.

More importantly, the sign (direction of change) matters here, so `Math.abs` is not appropriate. Just check if the height increased past the threshold.

```ts
if ((currentHeight - lastViewportHeight) / lastViewportHeight > KEYBOARD_VISIBILITY_THRESHOLD) {
...
}
```

karunkop:
> Thank you! This looks great.
> 
> There is one potential false positive I just thought of. If the user is in editing mode with the keyboard open while the device is held in landscape mode, and they rotate the device to portrait, then the visual viewport height will increase on resize. Will that incorrectly trigger the logic to disable editing mode?

You're right. This indeed highlights an important edge case. When a user rotates their device from landscape to portrait while the keyboard is open, the resulting increase in viewport height can falsely trigger our "keyboard closed" logic. This scenario illustrates the limitations of relying solely on viewport height changes to detect keyboard visibility.

That said, in my testing, the Android system typically collapses the keyboard automatically during orientation changes from landscape to portrait. This action infact should usually trigger a manual blur event, which helps mitigate the issue to some extent.

However, to ensure robustness across all devices and platforms, it would still be a good idea to implement device rotation detection. A more reliable approach would be to only infer keyboard collapse when the viewport height increases significantly without a corresponding significant change in width.

The logic could look something like this:

```ts
if (heightChangeRatio > KEYBOARD_VISIBILITY_THRESHOLD && widthChangeRatio < KEYBOARD_VISIBILITY_THRESHOLD) {
  // Keyboard likely collapsed
}
```
This refinement should help us avoid false positives caused by device orientation changes, making our keyboard visibility detection more accurate and resilient.

raineorshine (approved):
Looks great. To summarize how the solution developed:

- Identified where Android behaves differently than iOS, and confirmed that there is no event that consistently indicates when the keyboard closes on Android. This justified the need for an indirect means of determining that the keyboard has closed.
- Defined a `resize` handler that can be attached to the visual viewport. Pared down OOP logic to a simple event handler.
- Simplified logic and conditions to reduce complexity and improve intelligibility.

Issue #1189, opened 2021-05-27
Intermittent false scroll prevention


raineorshine:
Intermittently but regularly the scroll is prevented. Reproduce by scrolling in short, halting up-down swipes. 

- Is the gesture logic to prevent the scroll triggered? 
- Is it a boundary case in related to scroll/gesture thresholds?
- Related to horizontal displacement?
- UI freeze while work is being done in the background?

In this screen capture below, I am scrolling down a small amount at a regular interval (though is easier to retry by doing up-down swipes to stay at the same scroll position). I come to a complete stop each time, as the problem does not occur otherwise. You can see there are two gaps at the beginning (00:02 and 00:03) and one towards the end (00:07). That is where the scroll is failing. 

This causes not just scrolling but also attempted gestures to fail sometimes, presumably from the same mechanism. However, it never prevents an initial horizontal swipe. 

https://user-images.githubusercontent.com/750276/119823190-850bad80-bec2-11eb-8bf3-0d9dc2c9251a.mov

PR #2881, opened 2025-04-02
[Android] Screen scrolls during gesture that starts with ↑ or ↓

### LINKED ISSUE #2822: [Android] Screen scrolls during gesture that starts with ↑ or ↓
## Steps to Reproduce

Enter a gesture in the gesture zone that starts with ↑ or ↓.

## Current Behavior

The screen scrolls.

## Expected Behavior

The scroll should not scroll when touching the gesture zone.

karunkop:
Close #2822 

### Initial Approach

The initial fix was straightforward. Removing this.disableScroll = false from `onPanResponderMove` would keep scrolling disabled during gestures, preventing unwanted scrolling in the gesture zone. However, GitHub issue #1189  indicated that this line was added to address another issue, where scrolling could sometimes get stuck due to disableScroll not resetting properly.

After looking deeper, two competing issues were found. First, unwanted scrolling happened because disableScroll was being reset too early, allowing the browser’s native scrolling to kick in before the gesture was fully recognized. Second, there were cases where scrolling would get stuck. Short, interrupted swipes sometimes failed because disableScroll wasn’t properly reset, leaving the scroll disabled from a previous gesture.

### Root Cause

The root cause of the unwanted scrolling was that onPanResponderMove was always resetting disableScroll = false when a new gesture started. This conflicted with the touchstart event, which set disableScroll = true when a touch was detected in the gesture zone. Meanwhile, the intermittent scroll prevention issue happened because the gesture detection system wasn’t consistently clearing the disableScroll flag in certain cases. This left scrolling stuck in a disabled state, preventing normal interactions in the scrollable area.

### Comprehensive Solution

The solution takes a dynamic approach to handle both issues at the same time. Instead of always resetting disableScroll (which led to unwanted scrolling) or never resetting it (which caused scrolling to get stuck), it first checks the touch location. Using the existing `isInGestureZone` function, it determines whether the touch is inside the gesture zone. If it's outside, disableScroll is reset to false, allowing normal scrolling. This way, gestures remain uninterrupted while regular scrolling works as expected.


src/components/MultiGesture.tsx
```diff
@@ -166,9 +166,18 @@ class MultiGesture extends React.Component<MultiGestureProps> {
         // initialize this.currentStart on the the first trigger of the move event
         // TODO: Why doesn't onPanResponderStart work?
         if (!this.currentStart) {
-          // ensure that disableScroll is false when starting in case it wasn't reset properly
-          // may be related to https://github.com/cybersemics/em/issues/1189
-          this.disableScroll = false
+          // Check if we're in the gesture zone before deciding whether to disable scrolling
+          // This ensures we only prevent scrolling in the gesture zone, but allow it elsewhere
+          const touchLocation = e.nativeEvent.touches?.[0] || e.nativeEvent
```
trevinhofmann, line 171: The TS engine is telling me that `e.nativeEvent.touches` is always defined. Is there a reason for the `?.` optional chaining?

1. Is it possible that `e.nativeEvent.touches` is undefined, which would be contrary to the `GestureResponderEvent` type?

Or

2. Is it possible that the array is empty, in which case `e.nativeEvent.touches?.[0]` would be `undefined`, but the `?.` optional chaining doesn't affect anything.

trevinhofmann (changes requested):
@karunkop - Thank you for the PR! I appreciate the detailed PR description and helpful code comments.

I've tested on a physical Android device. On `main`, I was able to reproduce issue #2822 on `main`, and it has been resolved by this PR. I attempted to reproduce issue #1189 with the changes in this PR and could not, so this appears to avoid that regression.

I have one question about the optional chaining in `touchLocation`, otherwise this looks great to me!

src/components/MultiGesture.tsx
trevinhofmann: I appreciate the extra caution, but let's investigate if there are actual cases where it would be undefined (if the type definition is inaccurate). Otherwise, best to remove and simplify it 👍

trevinhofmann (approved):
Thank you for the update, @karunkop!

Testing locally, the issue still appears to be resolved with the latest changes. LGTM!

raineorshine (approved):

PR #2882, opened 2025-04-04
Animate Bullet cursor overlay

### LINKED ISSUE #2730: Animate cursor bullet overlay
## Steps to Reproduce

```
- a
- b
- c
```

Move the cursor from `a` to `b`. 

## Current Behavior

The cursor bullet overlay (the gray circle around the cursor bullet) immediately snaps to `b`. 

## Expected Behavior

The cursor overlay bullet should smoothly animate to its new location with veryFastDuration and ease-in-out. 

I recommend rendering the cursor bullet overlay with position: absolute in the LayoutTree next to the TreeNodes, since the LayoutTree has the thought positions.

The overlay must be perfectly centered on the bullet:

- at all font sizes
- on all platforms
- table view

When the cursor thought moves (e.g. indent, moveThoughtDown), the overlay should track with the bullet like it does currently.

cindmichelle:
Close #2730 

## Description
The main goal of this PR is to animate the bullet cursor overlay whenever the cursor moves. However, to achieve this, I also made several related changes:

- Moved `BulletCursorOverlay` to the `LayoutTree` level.

- Separated the overlay for the active cursor (`BulletCursorOverlay`) and the highlighted cursor (`BulletHighlightOverlay`) when dragging the node

- Moved variables related to bullet and cursor position (mostly padding and margin) into a hook. This ensures that any changes affecting bullet and cursor positioning are automatically synchronized across all relevant components. Affected components: `BulletCursorOverlay`, `Bullet`, `ContextBreadcrumbs`, and `ThoughtAnnotation`.

## How the Position Calculation Works
The x and y positions of the cursor overlay are determined by the bullet’s position, which is affected by left-right alignment, margin, and padding across normal view, context view, and table view. Additionally, font size changes impact positioning since most margins and paddings use the em unit.

**Calculation Breakdown:**
```
X = X of active thought + left position + margin left -  padding left
```

```
Y = Y of active thought + margin top + padding top - margin bottom - padding bottom
```

There are also a few extra factors that affect the X position:
- When the thought is in **table view** and **it's in column 1**
- When the thought is in **context view** (adds extra pixels due to rendering ContextBreadcrumb)
The calculation follows the same logic as the X position adjustments mentioned earlier.

For the Y position with contextView , slight adjustments occur when the thought is in context view and has a breadcrumb. The Y position follows this formula:

```
Breadcrumb Y position = breadcrumb margin-top + Max(fontSize, breadcrumbFontSize + breadcrumb padding-top)
```

With
- `fontSize` is `1em`
- `breadcrumbFontSize` is `0.867em`

## Notes:
- The position of the cursor overlay may not always be perfectly centered, especially when using a small font size. However, I’ve adjusted it so that the bullet always remains within the overlay, even if it’s not perfectly centered.

- When Table View and Context View are active, there is a case where the cursor overlay highlights a blank space. I’ve verified that a node should be rendered in this blank space, but it's currently missing due to a duplicate key issue in the `TreeNode` component. This issue existed before these changes but was less noticeable because the cursor overlay was previously inside the TreeNode. So when the duplicate key issue happened, the cursor overlay didnt render, because the TreeNode with the duplicate key wasnt rendered either. (will be addressed in here: #2884 )

image

## Results (Videos/Screenshots)
**Desktop Chrome (font size 18)**

https://github.com/user-attachments/assets/cba8713f-4403-4ff0-a34b-f87b64ab258f

---

**Context View active**

https://github.com/user-attachments/assets/d0125e51-7a34-4585-bd18-cdb23755f2e0

---
**When Table View and Context View is active**

https://github.com/user-attachments/assets/ceb68ba6-936f-472f-876f-5cf18f8ba9c0

---
**Font Size 32**

https://github.com/user-attachments/assets/8e0628a4-42af-4e4a-9def-6dc7b3619ff7

---
**Font Size 10**

https://github.com/user-attachments/assets/253fdc5d-d9a3-4328-b434-22d061cc42fa

---
**iOS App**

https://github.com/user-attachments/assets/8b6c5ba7-18c1-4889-ba43-5829e460af3a

---
**iOS Safari**

https://github.com/user-attachments/assets/42702f6f-cb23-49ae-a19f-ed9d0b117b15

---
**Android App**

https://github.com/user-attachments/assets/79f26c61-73d0-408f-a3cd-7bd218b14252

---
**Android Chrome**

https://github.com/user-attachments/assets/c2dfd285-fbdf-48b2-8510-b87ed3fb036d

### DIFF
--- a/src/components/BulletCursorOverlay.tsx
+++ b/src/components/BulletCursorOverlay.tsx
@@ -174,23 +174,6 @@ function PlaceholderContextBreadcrumbs({ simplePath }: { simplePath: SimplePath 
 }
 
 /**
- * PlaceholderSubThought is a component mimicking the behavior of SubThought.
- * Needed to prevent the overlay from covering the bullet.
- */
-function PlaceholderSubThought({ children }: { children?: React.ReactNode }) {
-  return (
-    <div
-      aria-label='placeholder-subthought'
-      className={css({
-        willChange: 'opacity',
-      })}
-    >
-      {children}
-    </div>
-  )
-}
-
-/**
  * PlaceholderTreeNode is a component used to mimic behavior of TreeNode.
  * Any position changes from one Thought to another will be animated within this component.
  */
@@ -361,25 +344,23 @@ export default function BulletCursorOverlay({
 
   return (
     <PlaceholderTreeNode width={width} x={x} y={y} isTableCol1={isTableCol1}>
-      <PlaceholderSubThought>
-        {showContexts && simplePath?.length > 1 && <PlaceholderContextBreadcrumbs simplePath={simplePath} />}
+      {showContexts && simplePath?.length > 1 && <PlaceholderContextBreadcrumbs simplePath={simplePath} />}
 
-        <div
-          aria-label='placeholder-thought-container'
-          className={css({
-            /* Use line-height to vertically center the text and bullet. We cannot use padding since it messes up the selection. This needs to be overwritten on multiline elements. See ".child .editable" below. */
-            /* must match value used in Editable useMultiline */
-            lineHeight: '2',
-            // ensure that ThoughtAnnotation is positioned correctly
-            position: 'relative',
-            ...(hideBullet ? { marginLeft: -12 } : null),
-          })}
-        >
-          <CursorOverlay simplePath={simplePath} path={path} leaf={leaf} isInContextView={isInContextView} />
+      <div
+        aria-label='placeholder-thought-container'
+        className={css({
+          /* Use line-height to vertically center the text and bullet. We cannot use padding since it messes up the selection. This needs to be overwritten on multiline elements. See ".child .editable" below. */
+          /* must match value used in Editable useMultiline */
+          lineHeight: '2',
+          // ensure that ThoughtAnnotation is positioned correctly
+          position: 'relative',
+          ...(hideBullet ? { marginLeft: -12 } : null),
+        })}
+      >
+        <CursorOverlay simplePath={simplePath} path={path} leaf={leaf} isInContextView={isInContextView} />
 
-          <PlaceholderThoughtAnnotation />
-        </div>
-      </PlaceholderSubThought>
+        <PlaceholderThoughtAnnotation />
+      </div>
     </PlaceholderTreeNode>
   )
 }
--- a/src/components/BulletCursorOverlay.tsx
+++ b/src/components/BulletCursorOverlay.tsx
@@ -1,0 +1,385 @@
+import { useSelector } from 'react-redux'
+import { css } from '../../styled-system/css'
+import Path from '../@types/Path'
+import SimplePath from '../@types/SimplePath'
+import Thought from '../@types/Thought'
+import ThoughtId from '../@types/ThoughtId'
+import { isSafari, isTouch, isiPhone } from '../browser'
+import useHideBullet from '../hooks/useHideBullet'
+import attributeEquals from '../selectors/attributeEquals'
+import { findAnyChild, getChildrenRanked } from '../selectors/getChildren'
+import isContextViewActive from '../selectors/isContextViewActive'
+import rootedParentOf from '../selectors/rootedParentOf'
+import equalThoughtRanked from '../util/equalThoughtRanked'
+import getBulletWidth from '../util/getBulletWidth'
+import head from '../util/head'
+import isRoot from '../util/isRoot'
+import parentOf from '../util/parentOf'
+import FauxCaret from './FauxCaret'
+
+type BulletCursorOverlayProps = {
+  x: number
+  y: number
+  simplePath: SimplePath
+  path: Path
+  isTableCol1: boolean
+  width?: number
+  value?: string
+  parentId: ThoughtId
+  showContexts?: boolean
+  leaf?: boolean
+}
+
+const isIOSSafari = isTouch && isiPhone && isSafari()
+
+/** Returns true if two lists of children are equal. Deeply compares id, value, and rank. */
+const equalChildren = (a: Thought[], b: Thought[]) =>
+  a === b ||
+  (a && b && a.length === b.length && a.every((thought, i) => equalThoughtRanked(a[i], b[i]) && a[i].id === b[i].id))
+
+/**
+ * CursorOverlay is a component that renders the cursor overlay for a thought bullet.
+ * @param props - The props for the component.
+ * @returns The rendered cursor overlay.
+ */
+function CursorOverlay({
+  simplePath,
+  path,
+  leaf,
+  isInContextView,
+}: {
+  simplePath: SimplePath
+  path: Path
+  leaf?: boolean
+  isInContextView?: boolean
+}) {
+  const bulletOverlayRadius = isIOSSafari ? 300 : 245
+
+  // Bottom margin for bullet to align with thought text
+  const glyphBottomMargin = isIOSSafari ? '-0.2em' : '-0.3em'
+
+  const showContexts = useSelector(state => isContextViewActive(state, path))
+  const fontSize = useSelector(state => state.fontSize)
+
+  const lineHeight = fontSize * 1.25
+
+  const extendClickWidth = fontSize * 1.2
+  const extendClickHeight = fontSize / 3
+
+  const isTableCol1 = useSelector(state =>
+    attributeEquals(state, head(rootedParentOf(state, simplePath)), '=view', 'Table'),
+  )
+
+  // calculate position of bullet for different font sizes
+  // Table column 1 needs more space between the bullet and thought for some reason
+  const width = getBulletWidth(fontSize) + (!isInContextView && isTableCol1 ? fontSize / 4 : 0)
+  const marginLeft = -width
+
+  return (
+    <span
+      aria-label='placeholder-bullet'
+      style={{
+        top: -extendClickHeight,
+        left: -extendClickWidth + marginLeft,
+        paddingTop: extendClickHeight,
+        paddingLeft: extendClickWidth,
+        paddingBottom: extendClickHeight + 2,
+        width,
+        position: 'absolute',
+        verticalAlign: 'top',
+        zIndex: isIOSSafari ? 4 : undefined, // fix misalignment of cursor on iOS
+      }}
+    >
+      <svg
+        viewBox='0 0 600 600'
+        style={{
+          height: lineHeight,
+          width: lineHeight,
+          marginLeft: -lineHeight,
+          // required to make the distance between bullet and thought scale properly at all font sizes.
+          left: lineHeight * 0.317,
+          marginBottom: glyphBottomMargin,
+          willChange: 'transform',
+          position: 'relative',
+          // backgroundColor: 'red',
+          top: showContexts && leaf && isIOSSafari ? '-0.05em' : undefined,
+        }}
+      >
+        <g>
+          <ellipse
+            ry={bulletOverlayRadius}
+            rx={bulletOverlayRadius}
+            cy='300'
+            cx='300'
+            className={css({
+              stroke: 'highlight',
+              fillOpacity: 0.25,
+              fill: 'fg',
+            })}
+          />
+        </g>
+      </svg>
+    </span>
+  )
+}
+
+/**
+ * PlaceholderContextBreadcrumbs is a component that renders invisible breadcrumbs for context view.
+ * Used to maintain layout consistency when context breadcrumbs are present.
+ */
+function PlaceholderContextBreadcrumbs({ simplePath }: { simplePath: SimplePath; path: Path }) {
+  const ancestors = parentOf(simplePath)
+
+  const isHaveMultipleAncestors = ancestors.length > 1
+
+  return (
+    <div
+      aria-label='placeholder-ctx-breadcrumbs-outer-container'
+      className={css({
+        marginLeft: 'calc(1.3em - 14.5px)',
+        minHeight: '1em',
+        visibility: 'hidden',
+        marginBottom: '-0.25em', // Tighten up the space between the context-breadcrumbs and the thought (similar to the space above a note).
+        paddingTop: '0.5em', // Use padding-top instead of margin-top to ensure this gets included in the dynamic height of each thought. Otherwise the accumulated y value will not be correct.
+      })}
+      style={{
+        fontSize: '0.867em',
+        marginTop: '0.533em',
+      }}
+    >
+      {isRoot(simplePath) ? null : (
+        <span
+          style={{
+            wordBreak: 'break-word',
+            textDecoration: 'none',
+            WebkitTextStrokeWidth: '0.05em',
+
+            height: '1em',
+            margin: '-0.5em',
+            padding: '0.5em',
+            ...(isHaveMultipleAncestors
+              ? {
+                  lineHeight: '16px',
+                }
+              : {}),
+          }}
+          dangerouslySetInnerHTML={{ __html: '&ZeroWidthSpace;' }}
+        ></span>
+      )}
+    </div>
+  )
+}
+
+/**
+ * PlaceholderSubThought is a component that renders a sub-thought in the thought tree.
+ * @param param0 - The props for the component.
+ * @returns The rendered component.
+ */
+function PlaceholderSubThought({ children }: { children?: React.ReactNode }) {
+  return (
+    <div
+      aria-label='placeholder-subthought'
+      className={css({
+        willChange: 'opacity',
+      })}
+    >
+      {children}
+    </div>
+  )
+}
+
+/**
+ * PlaceholderTreeNode is a component used to mimic behavior of TreeNode.
+ * Any position changes from one Thought to another will be animated within this component.
+ */
+function PlaceholderTreeNode({
+  children,
+  isTableCol1 = false,
+  x,
+  y,
+  width,
+}: {
+  children?: React.ReactNode
+  isTableCol1?: boolean
+  x: number
+  y: number
+  width: number
+}) {
+  const outerDivStyle = {
+    left: x,
+    top: y,
+    // Table col1 uses its exact width since cannot extend to the right edge of the screen.
+    // All other thoughts extend to the right edge of the screen. We cannot use width auto as it causes the text to wrap continuously during the counter-indentation animation, which is jarring. Instead, use a fixed width of the available space so that it changes in a stepped fashion as depth changes and the word wrap will not be animated. Use x instead of depth in order to accommodate ancestor tables.
+    // 1em + 10px is an eyeball measurement at font sizes 14 and 18
+    // (Maybe the 10px is from .content padding-left?)
+    width: isTableCol1 ? width : 'calc(100% - ${x}px + 1em + 10px)',
+    textAlign: isTableCol1 ? ('right' as const) : undefined,
+  }
+
+  return (
+    <div
+      aria-label='placeholder-tree-node'
+      className={css({
+        transition: 'left {durations.layoutNodeAnimation} linear,top {durations.layoutNodeAnimation} ease-in-out',
+        ...(isTableCol1
+          ? {
+              position: 'relative',
+              width: 'auto',
+            }
+          : {
+              position: 'absolute',
+              width: '100%',
+            }),
+      })}
+      style={{
+        ...outerDivStyle,
+      }}
+    >
+      {children}
+    </div>
+  )
+}
+
+/**
+ * A placeholder component for the ThoughtAnnotation component.
+ * Used to maintain consistent positioning of the cursor and bullet in normal view on iOS.
+ */
+function PlaceholderThoughtAnnotation() {
+  return (
+    <div
+      aria-label='placeholder-thought-annotation-outer-container'
+      className={css({
+        opacity: 0,
+        position: 'absolute',
+        boxSizing: 'border-box',
+        width: '100%',
+        marginTop: '0',
+        display: 'inline-block',
+        verticalAlign: 'top',
+        whiteSpace: 'pre-wrap',
+        maxWidth: '100%',
+        '@media (max-width: 500px)': {
+          marginTop: { _android: '-2.1px' },
+          marginLeft: { _android: '0.5em' },
+        },
+        '@media (min-width: 560px) and (max-width: 1024px)': {
+          marginTop: { _android: '-0.1px' },
+          marginLeft: { _android: '0.5em' },
+        },
+      })}
+    >
+      <div
+        aria-label='placeholder-thought-annotation-inner-container'
+        className={css({
+          display: 'inline-block',
+          maxWidth: '100%',
+          boxSizing: 'border-box',
+          paddingRight: '0.333em',
+        })}
+        style={{
+          padding: '0 0.333em',
+          margin: '-0.5px 0 0 calc(1em - 18px)',
+        }}
+      >
+        <span
+          className={css({
+            fontSize: '1.25em',
+            margin: '-0.375em 0 0 -0.05em',
+            position: 'absolute',
+          })}
+        >
+          <FauxCaret caretType='thoughtStart' />
+        </span>
+        <span
+          className={css({
+            visibility: 'hidden',
+            position: 'relative',
+            clipPath: 'inset(0.001px 0 0.1em 0)',
+            wordBreak: 'break-word',
+          })}
+          dangerouslySetInnerHTML={{ __html: '&ZeroWidthSpace;' }}
+        />
+      </div>
+    </div>
+  )
+}
+
+/**
+ * BulletCursorOverlay is a component used to animate the cursor overlay from the bullet.
+ * This component also contains placeholders for other components to maintain consistency of cursor overlay position.
+ **/
+export default function BulletCursorOverlay({
+  x,
+  y,
+  simplePath,
+  path,
+  isTableCol1,
+  width = 0,
+  value,
+  parentId,
+  showContexts,
+  leaf,
+}: BulletCursorOverlayProps) {
+  const childrenAttributeId = useSelector(
+    state => (value !== '=children' && findAnyChild(state, parentId, child => child.value === '=children')?.id) || null,
+  )
+  const grandparentId = simplePath[simplePath.length - 3]
+
+  const grandchildrenAttributeId = useSelector(
+    state =>
+      (value !== '=style' && findAnyChild(state, grandparentId, child => child.value === '=grandchildren')?.id) || null,
+  )
+
+  const hideBulletProp = useSelector(state => {
+    const hideBulletsChildren = attributeEquals(state, childrenAttributeId, '=bullet', 'None')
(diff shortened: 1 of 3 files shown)


src/components/BulletCursorOverlay.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+/** A larger circle that surrounds the bullet of the cursor thought. */
+const BulletCursorOverlay = ({
+  isCursorActive,
+  x = 0,
+  y = 0,
+  simplePath,
+  path,
+}: {
+  isCursorActive?: boolean
```
raineorshine, line 21: Is there a reason to pass `isCursorActive` as a prop if it is just used for a conditional render?

If there is additional complexity in the condition I sometimes push it down to the child component, otherwise I usually control rendering in the parent component.

src/components/BulletCursorOverlay.tsx
```diff
@@ -0,0 +1,135 @@
...
+  x?: number
+  y?: number
+  simplePath?: SimplePath
+  path?: Path
```
raineorshine, line 25: Similarly, `simplePath` is derivable from `path` (via `simplifyPath(state, path)`), so you should remove it from the hook arguments and calculate it internally in a `useSelector`.

src/components/LayoutTree.tsx
```diff
@@ -231,6 +233,11 @@ const LayoutTree = () => {
     },
   )
 
+  // compare between state.cursor and the position of the thought
+  const activeThought = useSelector(state =>
```
raineorshine, line 237: Let's call this `cursorThoughtPositioned`. I'd like to avoid adding new terminology like "active."

src/components/LayoutTree.tsx
```diff
@@ -275,6 +282,13 @@ const LayoutTree = () => {
         }}
       >
         <TransitionGroup>
+          <BulletCursorOverlay
```
raineorshine, line 285: It does no harm, but it's probably better to move `BulletCursorOverlay` outside `TransitionGroup` for clarity. Only `Transition` and `CSSTransition` components have an effect within `TransitionGroup`. (`TreeNode` returns a `CSSTransition` component.)

src/components/ThoughtAnnotation.tsx
```diff
@@ -107,8 +108,10 @@ const ThoughtAnnotation = React.memo(
     styleAnnotation,
     url,
     placeholder,
+    path,
     value,
   }: {
+    path: Path
```
raineorshine, line 114: If we're going to add `path` to `ThoughtAnnotation`, we should replace `simplePath` to avoid the overlap.

src/components/ThoughtAnnotation.tsx
```diff
@@ -197,7 +208,11 @@ const ThoughtAnnotation = React.memo(
             //   border-bottom: solid 1px;
             // }
           }
-          style={styleAnnotation}
+          style={{
+            padding: `0 ${thoughtAnnotationHorizontalPadding}px`,
```
raineorshine, line 212: I'm concerned that `paddingRight` (above) will be overridden now that `padding` has been moved to `style`. It might be a good idea to move `paddingRight` with it.

raineorshine (changes requested):
Hi! Thanks for your submission.

The overall approach of `useBulletPosition` is quite interesting. I appreciate that it avoids duplication, keeps everything synchronized, and is React-centric. I underestimated the different margin/padding used in different components that contain the Bullet, and this handles that nicely.

The downside is having the `BulletCursorOverlay` so tightly coupled to the layout details of table view, context view, etc, making it fragile if these change in the future. While some special casing may be necessary, the margin/padding special casing here is extensive. However, I can't think of a reliable alternative. Measuring the offset of the Bullet in the DOM would allow us to bypass case-specific margin/padding, but I worry that directly measuring from the DOM will be prone to other issues. One of the strengths of your approach is that it is pure React.

Anyway, time will tell if this approach proves difficult to maintain, but it looks good as an initial implementation. It is clearly well thought out. Thanks!

---

I'm very happy with how it behaves. The animation really helps direct attention when the cursor changes. It gives navigation a more free-flowing feeling, where the cursor is actively moving from thought-to-thought rather than being statically attached. I think the appearance of the caret will need to be delayed to match the cursor, but that is a separate concern.

The one thing that is not sufficient for production is the bullet centering. Due to the bull's-eye positioning of the bullet within the cursor overlay, any off-centeredness is extremely visible. Let's investigate this and try to better understand why it is off center at different font sizes. Verify yourself, but my initial impression is that just the vertical centering is off. Here, the red outline is on the bullet's svg element. The bullet is positioned perfectly in the center, while the cursor overlay sits too high. I would look into how that box is scaling, and think about how the `y` value needs to be adjusted. I think we can assume that it scales linearly with font size, but I'm not sure at what rate, i.e. what is the scaling coefficient, e.g. fontSize, fontSize*0.5, fontSize*2, etc.

Brave Browser 2025-04-10 08 56 56

src/components/BulletCursorOverlay.tsx
cindmichelle: Yep, makes sense. I'm gonna move the conditional logic up to the LayoutTree level instead

cindmichelle:
**Arc Desktop**

https://github.com/user-attachments/assets/342b8d35-7674-4888-a3da-84faac81a8e7

----
**Edge Browser**

https://github.com/user-attachments/assets/f94a2b4e-6a1d-4813-9064-28e45bee25ec

----

cindmichelle:
Hi @raineorshine , sorry for the long delay to update this PR! I’ve finally resolved the bullet centering issue by tweaking the approach a bit. I ended up combining my earlier method with the static overlay approach that the current bullet implementation uses.

It looks like the first approach I tried wasn’t consistently accurate, especially when the overlay position had fractional values. That’s likely because the way CSS calculates positions during DOM rendering differs from how it handles transforms, and they occur at different points in the render lifecycle.

As you probably remember (from our chat on discord), I also experimented with a different method using `getBoundingClientRect()` to get the bullet’s exact position. But that approach had quite a few issues. for example, I had to add a delay before reading the position, since the thought component renders in two phases. This meant the initial position wasn't always reliable, and overall it felt too risky to move forward with due to the number of uncertain factors (potential tech debt).

To make the combined approach (my first method + the current static overlay from main) work, I’ve now split the animation into two components: **AnimateOverlay** (from my earlier PR), and **StaticOverlay** (from the current main implementation). Here’s how the animation behaves:
**a. When the cursor is on Thought A:**
- AnimateOverlay is hidden (opacity: 0)
- StaticOverlay for Bullet A is shown (opacity: 1)

**b. When the cursor moves from Thought A → Thought B:**
- StaticOverlay for Bullet A is hidden (opacity: 0)
- AnimateOverlay is shown (opacity: 1), and animates the bullet from A to B
- Once it reaches Thought B, AnimateOverlay is hidden again
- StaticOverlay for Bullet B is then shown (opacity: 1)

This new animation setup uses PandaCSS and CSSTransition for smooth transitions.

src/components/Bullet/BulletAnimateOverlay.tsx
```diff
@@ -0,0 +1,205 @@
...
+        y: translateY,
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  return (
+    <TransitionGroup>
+      <CSSTransition
+        key={`${translateX}-${translateY}`}
```
raineorshine, line 147: It seems strange to build a React key out of the `x`, `y` position. They `key` is typically a unique identifier for the element that does not change. Can you explain?

raineorshine (changes requested):
Hi, thanks for all your effort on this.

The hybrid approach comes with some additional complexity, but the end result looks very good. With some additional tweaking, I'm hopeful we can make it work.

I have started with a behavioral review, so I don't have comments on the code itself yet. However I am seeing several errors in the JS console (desktop Chrome) which will need to be fixed.

I will post the behavioral issues I'm seeing below in separate comments for easy linkability. All are reproducible in desktop Chrome. Before considering potential solutions, let's identify the causes and assess the risk and effort involved in fixing them. The biggest priority will be to determine if any of them are fundamental issues with the current approach, or superficial issues that can be remedied on top of the existing strategy. I look forward to your response.

(FYI I would suggest skipping screenshots and videos of succeeding cases to save a little time. They have to go through QA anyway, so there's no real need to post a screenshot or video. They are more useful for pointing out failing cases than succeeding cases.)

raineorshine:
## Steps to Reproduce

```
- a
  - b
- c
```

1. Set the cursor on `a`.
2. Activate `cursorNext` by hitting <kbd>Meta</kbd> + <kbd>ArrowDown</kbd>

## Current Behavior

The cursor overlay bullet starts animating from `a` to `c`, then appears to jump back up to `a` in the middle of the animation:

https://github.com/user-attachments/assets/bca48058-8b53-48b7-840c-1b3c13ccafe8

## Expected Behavior

The cursor overlay bullet should smoothly transition from `a` to `c`.

raineorshine:
## Steps to Reproduce

```
- a
- b
  - c
```

1. Set the cursor on `a`.
2. *Quickly* hit <kbd>ArrowDown</kbd> twice.

## Current Behavior

The cursor overlay bullet appears to jump back up to `a` in the middle of the animation.

This also happens in reverse.

## Expected Behavior

The cursor overlay bullet should smoothly transition from `a` to `b` to `c`.

raineorshine:
# Outdent

There are two cases that demonstrate this issue.

## Case 1

### Steps to Reproduce

```
- a
  - b
``` 

Outdent `b`.

### Current Behavior

The cursor overlay does not stay centered on the cursor bullet during the animation.

https://github.com/user-attachments/assets/c7bea22d-9a31-47d3-9cb1-0976e4dfa8be


### Expected Behavior

The cursor overlay should stay perfectly centered on the cursor bullet when moving.

## Case 2

### Steps to Reproduce

```
- a
  - d
  - b
    - c
``` 

Outdent `b`.

### Current Behavior

The cursor overlay does not stay centered on the cursor bullet during the animation.

https://github.com/user-attachments/assets/8a0ab52f-7f83-445d-821e-0a461c49b2ca

### Expected Behavior

The cursor overlay should stay perfectly centered on the cursor bullet when moving.

# Notes

To match the bullet movement, the cursor overlay would need to animate at [`layoutNodeAnimation: 150`](https://github.com/cybersemics/em/blob/6e2705cfe561bf900da2fea83654d1732c08d1cc/src/durations.config.ts#L27-L28). However, this would feel far too slow when navigating the cursor normally.

Therefore, I think the most reasonable approach would be to disable the animated bullet cursor overlay completely when the cursor does not change, and just use the static overlay in these cases. Luckily it's very easy to to tell if the cursor has changed: you would just compare `head(state.cursor)`. The challenge is integrating that into the existing logic.

I would put this at a lower priority than the other issues, since it just qualifies the existing functionality, and does not change anything fundamental. I would even consider addressing this in a separate PR after the other issues have been fixed and this PR has been merged.

src/components/Bullet/BulletAnimateOverlay.tsx
cindmichelle: Good catch,  I used `key={${translateX}-${translateY}}` to trigger the translation from thought A to B (which also triggers remounting when target position changed). The reason I used this approach is that previously, using `in={animationActive}` didnt work when `animationActive` was either a state or a memoized value that depended on whether the current position matched the target

cindmichelle: Good point, let me investigate why the `in` props doesnt works previously

cindmichelle: seems that the `in` prop isnt working because I wrapped `CSSTransition` with `TransitionGroup`. will remove TransitionGroup since we only need a single transition. This also seems to be the cause of the cursorNext and rapid cursor issues

raineorshine:
Hi! I did some deeper investigation into the feasibility of absolutely positioning the cursor overlay bullet. 

Here is the code I've been using as a proof of concept. It renders a box centered on the svg of the last bullet in the document. It is dynamically sized based on the font size to make it easier to visualize the alignment. It re-renders every second as a simple way to update with font size changes. I suggest we both use this so we have a standardized approach.

```
typeof t != 'undefined' && clearInterval(t)
t = setInterval(() => {
    let svg = [...document.querySelectorAll('.bullet svg')].at(-1)
    let size = parseFloat(window.getComputedStyle(svg).fontSize) / 3
    let { left, top, width, height } = svg.getBoundingClientRect()
    box = document.getElementById('box') ?? document.createElement('div')
    box.style.top = `${top + window.scrollY + height/2 - size/2}px`
    box.style.left = `${left + window.scrollX + width/2 - size/2}px`
    box.style.width = `${size}px`
    box.style.height = `${size}px`
    box.style.background = 'rgba(255, 0, 0, 0.5)'
    box.style.position = 'absolute'
    box.style.zIndex = 9999999
    box.id = 'box'
    document.body.appendChild(box)
}, 1000)
```

Initial observations:

- `cx='297'` and `cy='298'` on the `BulletLeaf` is almost certainly wrong, though it apparently was used to address a positioning issue in the past. It was added in #1505 where you can see a long history of discussion about bullet positioning.
- Box is almost perfectly consistent across font sizes in Chrome. Possibly some room for slight adjustments.
- Box is less consistent across font sizes in Safari.
- `window.scrollX` and `window.scrollY` are in whole pixels and mess up the positioning in Safari.

Next steps:

- Make sure there are no other styles that could be affecting these results. We need total confidence to move forward.
- Determine if there is a pattern to the discrepancies across font sizes. If they systematically veer left, for example, we could compensate for it.
- Test across more platforms.

Let me know your thoughts about any of the above, as well as the suggested next steps. My sense is that handling timing and animation will go much easier if we have a rock solid absolute positioning algorithm figured out first.

cindmichelle:
Hi @raineorshine , I’ve tried implementing the nudge logic for overlay positioning. So far, I’ve only tested it using the default font size in Chrome on desktop, and it looks like the approach you suggested works well, with a few adjustments I made.

Here's a quick rundown of what I’ve done:
### 1. Storing bullet refs in ministore:
With this approach, we need to store a ref for each bullet in the ministore to get more accurate position data. This allows the nudge logic to run without hitting race conditions. Based on my testing, storing these refs doesn’t negatively impact performance.

### 2. Added a delay for specific cursor movements:
When the cursor moves from a parent or child thought to an uncle thought that’s ranked above the previous one, I added a delay to the animation. The full animation takes 150ms, some are straightforward movements, while others have a 70ms delay followed by an 80ms movement. Timing for nudge logic now varies depending on the movement:
- **Same level** or **parent → child**: nudge after 80ms.
- **To uncle thought**: nudge is delayed by 150ms to sync with the animation.

### 3. Handling layout shifts:
If the cursor moves while a layout shift is happening (e.g. due to an indent change), we delay the nudge by 750ms to ensure the bullet’s position is accurate.This causes a slight visual jump in the cursor, but seems necessary for correctness. (Would love your thoughts on this.)

### 4. Fast cursor movement:
The nudge logic is now debounced using `max(unprocessedDuration, currentCursorMoveDuration)`.
Without this, I noticed that fast indent changes caused incorrect bullet positioning, due to the nudge happened while the layout was still adjusting.

So far, combining animation based on calculated position with a nudge to the actual position feels like the smoothest and least awkward approach compared to the two previous ones.

Let me know what you think of this implementation!

https://github.com/user-attachments/assets/da2a1ed4-55ea-47f7-8a17-5c5afdfdfd5f

cindmichelle:
Hi @raineorshine , here’s the implementation using the placeholder approach. In this approach, we need to copy several components as placeholders to accurately mimic the cursor bullet’s position. The components currently copied are: TreeNode, Subthought, Thought, ContextBreadcrumb, ThoughtAnnotation and Bullet.

This is how the implementation looks like. I’ve tested it on desktop, iOS, and Android across multiple browsers, and it looks pretty solid now.

https://github.com/user-attachments/assets/55435c4b-89ac-43bc-923d-6dad7702fb46

----

However, I’ve noticed a few remaining issues that I’m leaving as-is for now:
**1. Context view**: When moving from a thought with a single/no breadcrumb to one with multiple breadcrumbs, the cursor movement feels jarring, especially in small font size (<16px). This is due to the height difference of the thought, multiple breadcrumbs increase the ContextBreadcrumb height because of the delimiter (16px), whereas a single breadcrumb’s height is only 1em.

https://github.com/user-attachments/assets/1d6e8111-33bd-49fd-b557-2dc098624ff9

---

**2. Table view:** Moving from col1 to col 2 isnt smooth due to a difference in textAlign styling: col1 uses `textAlign: right;`, while column 2 and non-table views fall back to the default value. I’m leaving this as-is for now since it seems like a broader table view issue. In col1, the bullet position overlaps with the thought content, so ideally fixing the bullet positioning there should also resolve this problem.

https://github.com/user-attachments/assets/51d2beb4-015e-44ad-bbcd-64505fb6e318


Let me know your thoughts on this implementation.

raineorshine:
Thank you for the update!

> **1. Context view**: When moving from a thought with a single/no breadcrumb to one with multiple breadcrumbs, the cursor movement feels jarring, especially in small font size (<16px). This is due to the height difference of the thought, multiple breadcrumbs increase the ContextBreadcrumb height because of the delimiter (16px), whereas a single breadcrumb’s height is only 1em.

Is this the reason the cursor overlay appears to overshoot at the beginning of the animation?

[static screenshot]
image

> **2. Table view:** Moving from col1 to col 2 isnt smooth due to a difference in textAlign styling: col1 uses `textAlign: right;`, while column 2 and non-table views fall back to the default value. I’m leaving this as-is for now since it seems like a broader table view issue. In col1, the bullet position overlaps with the thought content, so ideally fixing the bullet positioning there should also resolve this problem.

Yeah, that's a tricky case due to the changing text align. I think it's wise to ignore that.

cindmichelle:
> Is this the reason the cursor overlay appears to overshoot at the beginning of the animation?
> 
> [static screenshot] image
> 
@raineorshine Yep, thats correct

src/components/LayoutTree.tsx
```diff
@@ -215,6 +217,11 @@ const LayoutTree = () => {
     },
   )
 
+  // compare between state.cursor and the position of the thought
+  const cursorThoughtPositioned = useSelector(state =>
+    treeThoughtsPositioned.find(thought => equalPath(state.cursor, thought.path)),
+  )
```
raineorshine: Can you use `treeThoughtsPositioned.find(thought => thought.isCursor)` to avoid the `useSelector`?

src/components/LayoutTree.tsx
```diff
@@ -258,6 +265,18 @@ const LayoutTree = () => {
...
+        {cursorThoughtPositioned && (
+          <BulletCursorOverlay
+            isTableCol1={cursorThoughtPositioned.isTableCol1}
+            path={cursorThoughtPositioned.path}
+            simplePath={cursorThoughtPositioned.simplePath}
+            x={cursorThoughtPositioned.x}
+            y={cursorThoughtPositioned.y}
+            showContexts={cursorThoughtPositioned.showContexts}
+            width={cursorThoughtPositioned.width}
+            parentId={cursorThoughtPositioned.path[cursorThoughtPositioned.path.length - 2] as ThoughtId}
```
raineorshine: Let's use the path helper functions instead of interacting directly with `Path` as an array:

```tsx
parentId={parentOf(cursorThoughtPositioned.path)}
```

src/components/BulletCursorOverlay.tsx
```diff
@@ -0,0 +1,385 @@
...
+  width: number
+}) {
+  const outerDivStyle = {
+    left: x,
+    top: y,
+    // Table col1 uses its exact width since cannot extend to the right edge of the screen.
+    // All other thoughts extend to the right edge of the screen. We cannot use width auto as it causes the text to wrap continuously during the counter-indentation animation, which is jarring. Instead, use a fixed width of the available space so that it changes in a stepped fashion as depth changes and the word wrap will not be animated. Use x instead of depth in order to accommodate ancestor tables.
+    // 1em + 10px is an eyeball measurement at font sizes 14 and 18
+    // (Maybe the 10px is from .content padding-left?)
+    width: isTableCol1 ? width : 'calc(100% - ${x}px + 1em + 10px)',
```
raineorshine: Do you mean to use a template literal?

```diff
diff --git a/src/components/BulletCursorOverlay.tsx b/src/components/BulletCursorOverlay.tsx
index 7dfaeae05b..a028dacf82 100644
--- a/src/components/BulletCursorOverlay.tsx
+++ b/src/components/BulletCursorOverlay.tsx
@@ -212,7 +212,7 @@ function PlaceholderTreeNode({
     // All other thoughts extend to the right edge of the screen. We cannot use width auto as it causes the text to wrap continuously during the counter-indentation animation, which is jarring. Instead, use a fixed width of the available space so that it changes in a stepped fashion as depth changes and the word wrap will not be animated. Use x instead of depth in order to accommodate ancestor tables.
     // 1em + 10px is an eyeball measurement at font sizes 14 and 18
     // (Maybe the 10px is from .content padding-left?)
-    width: isTableCol1 ? width : 'calc(100% - ${x}px + 1em + 10px)',
+    width: isTableCol1 ? width : `calc(100% - ${x}px + 1em + 10px)`,
     textAlign: isTableCol1 ? ('right' as const) : undefined,
   }
```

src/components/BulletCursorOverlay.tsx
```diff
@@ -0,0 +1,385 @@
...
+          dangerouslySetInnerHTML={{ __html: '&ZeroWidthSpace;' }}
+        ></span>
+      )}
+    </div>
+  )
+}
+
+/**
+ * PlaceholderSubThought is a component that renders a sub-thought in the thought tree.
+ * @param param0 - The props for the component.
```
raineorshine: Comment

src/components/BulletCursorOverlay.tsx
```diff
@@ -0,0 +1,385 @@
...
+ * @returns The rendered component.
+ */
+function PlaceholderSubThought({ children }: { children?: React.ReactNode }) {
+  return (
+    <div
+      aria-label='placeholder-subthought'
+      className={css({
+        willChange: 'opacity',
```
raineorshine: Is `willChange: 'opacity'` needed?

src/components/BulletCursorOverlay.tsx
```diff
@@ -0,0 +1,385 @@
...
+      </svg>
+    </span>
+  )
+}
+
+/**
+ * PlaceholderContextBreadcrumbs is a component that renders invisible breadcrumbs for context view.
+ * Used to maintain layout consistency when context breadcrumbs are present.
+ */
+function PlaceholderContextBreadcrumbs({ simplePath }: { simplePath: SimplePath; path: Path }) {
```
raineorshine: `path` prop not used?

src/components/BulletCursorOverlay.tsx
```diff
@@ -0,0 +1,385 @@
...
+  const ancestors = parentOf(simplePath)
+
+  const isHaveMultipleAncestors = ancestors.length > 1
```
raineorshine: Defining an extra variable is unnecessary since `ancestors.length > 1` is short and trivial. You can inline it instead.

Also, you can skip the `parentOf` and just calculate `ancestors.length > 2`.

raineorshine (changes requested):
Great work on this latest iteration. Aside from the few edge cases that you noted, this is working really well. I tested on Desktop Chrome and Mobile Safari and the alignment looks perfect, as would be expected from reproducing the DOM hierarchy.

The only major issue, as you know, is the maintenance burden imposed by all the duplication. We've discussed a couple possibilities, namely eliminating duplication through abstraction vs cloning the DOM. I'm happy to discuss the merits of these approaches separately. This solution is working effectively that I would like to get it merged and consider refactoring in a separate PR. Likewise, the context view and table view edge cases can be handled in separate PR's. 

It also really helps that everything is contained within `BulletCursorOverlay`. It would be a lot harder to understand if it was spread out over multiple files.

Minor change requests below.

src/components/BulletCursorOverlay.tsx
cindmichelle: yep turns out we're no longer need that. will remove the prop

src/components/BulletCursorOverlay.tsx
cindmichelle: yes we still need this css property. it prevents the cursor overlay from appearing on top of the bullet, which would make the bullet not visible

src/components/BulletCursorOverlay.tsx
cindmichelle: Good catch, it should use template literal based on the original component. but let me double check again why the current implementation works and do we still need this styling

cindmichelle: I’ve double-checked, changing this to a template literal doesnt break anything, so I’ve updated it accordingly

src/components/BulletCursorOverlay.tsx
raineorshine: That sounds like a stacking context problem more than a performance issue.

We faced a similar question here: https://github.com/cybersemics/em/pull/2850#discussion_r2040001400

Could you try one of the alternatives approaches to avoid potential performance issues?

cindmichelle:
> Thanks for the updates!
> 
> It looks like the Puppeteer test is broken. Could you look into that?
> 
> If you need help determining whether it's a flaky test or not, you can ask @karunkop for help.

ive verified that this isnt a flaky test. turns out when the cursor overlay still appears when a divider exist. 

https://github.com/user-attachments/assets/8bea5e30-d480-4ea6-8311-2ae2f6214a93



but im also curious whether there are alternative ways to know if its flaky, other than in the app itself? @karunkop

src/components/BulletCursorOverlay.tsx
cindmichelle: Hi @raineorshine , I just realized that even without this style, we can still render the cursor and overlay properly. The reason it seemed problematic before was because I had added a background color for debugging during development, and I noticed that the overlay could cover the bullet appearance without `will-change: opacity`. But when I double-checked the appearance without the background color, it still looked perfect. 
(red=overlay, green=bullet)
https://github.com/user-attachments/assets/a9890a59-b550-499b-b1f2-0f61c35b0355


----

The only difference is that when we inspect the element when rendering without `will-change: opacity;` style, it selects the placeholder instead of the cursor bullet. 

with `will-change: opacity;` , it selects cursor bullet :


https://github.com/user-attachments/assets/8d1dc785-6682-4e77-a53c-d949fad4cd38


vs 

when the style or PlaceholderSubthought component being removed, it selects Placeholder cursor:


https://github.com/user-attachments/assets/8401bfaf-46d0-4c34-b10b-81583ef3ec51



Do you think it’s okay if i remove the usage of PlaceholderSubthought entirely?

raineorshine: Yes, that should be okay. Thanks!

raineorshine (changes requested):

raineorshine (approved):

Issue #2832, opened 2025-03-03
Sort: Created + Updated


raineorshine:
## Expected Behavior

Add two new sort options to the SortPicker:

- Created
- Updated

When sorting by Created, sort by `thought.created`.
When sorting by Updated, sort by `thought.lastUpdated`.

Additional Requirements:

- [x] If current sort is not Created, change sort to Created (asc).
- [x] Toggle sort by Created between Ascending and Descending when clicked multiple times:
    - If current sort is Created (asc) and Created is clicked, sort by Created (desc).
    - If current sort is Created (desc) and Created is clicked, sort by Created (asc).
- [ ] Perform correct Lottie animation regardless of which sort option from/to.

## Testing

Add SortPicker tests for sorting by Created or Updated.

PR #2883, opened 2025-04-05
Add new created and updated sort option

RED-ROSE515:
This PR resolves the issue : Sort: Created + Updated #2832

In this Pr, I added a new Sort Options created and updated.
And I updated compareThought functions to match created and updated sort options.

### DIFF
--- a/src/actions/setSortPreference.ts
+++ b/src/actions/setSortPreference.ts
@@ -99,10 +99,15 @@ const setSortPreference = (
           // Set the =sort attribute with the new type
           sortPreference.type !== currentSortPreference.type || sortPreference.type === globalSortPreference.type
             ? reducerFlow([
+                // toggleAttribute only overrides the existing attribute if the value is different. If the value is the same, it will remove the entire attribute.
+                // So we delete the attribute first to avoid this behavior.
+                deleteAttribute({
+                  path: simplePath,
+                  value: '=sort',
+                }),
                 toggleAttribute({
                   path: simplePath,
                   values: ['=sort', sortPreference.type],
-                  removeExisting: true,
                 }),
               ])
             : null,
--- a/src/actions/setSortPreference.ts
+++ b/src/actions/setSortPreference.ts
@@ -98,10 +98,16 @@ const setSortPreference = (
 
           // Set the =sort attribute with the new type
           sortPreference.type !== currentSortPreference.type || sortPreference.type === globalSortPreference.type
-            ? toggleAttribute({
-                path: simplePath,
-                values: ['=sort', sortPreference.type],
-              })
+            ? reducerFlow([
+                deleteAttribute({
+                  path: simplePath,
+                  value: '=sort',
+                }),
+                toggleAttribute({
+                  path: simplePath,
+                  values: ['=sort', sortPreference.type],
+                }),
+              ])
             : null,
 
           // Handle direction changes
--- a/src/actions/setSortPreference.ts
+++ b/src/actions/setSortPreference.ts
@@ -99,13 +99,10 @@ const setSortPreference = (
           // Set the =sort attribute with the new type
           sortPreference.type !== currentSortPreference.type || sortPreference.type === globalSortPreference.type
             ? reducerFlow([
-                deleteAttribute({
-                  path: simplePath,
-                  value: '=sort',
-                }),
                 toggleAttribute({
                   path: simplePath,
                   values: ['=sort', sortPreference.type],
+                  removeExisting: true,
                 }),
               ])
             : null,
--- a/src/components/__tests__/SortPicker.ts
+++ b/src/components/__tests__/SortPicker.ts
@@ -325,3 +325,58 @@ describe('Updated', () => {
     expect(thoughtValues).toMatchObject(['d', 'c', 'a'])
   })
 })
+
+describe('Created to Updated', () => {
+  it('Ascending', async () => {
+    // Import initial thought 'a'
+    act(() => {
+      store.dispatch([
+        importText({
+          text: `
+          - a
+        `,
+        }),
+        setCursor(['a']),
+      ])
+    })
+
+    await act(vi.runOnlyPendingTimersAsync)
+
+    // Create thought 'b'
+    act(() => {
+      store.dispatch([newThought({ value: 'b' })])
+    })
+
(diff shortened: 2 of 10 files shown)


src/components/__tests__/SortPicker.ts
```diff
@@ -124,3 +124,31 @@ it('subthoughts: Desc', async () => {
 
   expect(subthoughtsOfA.map((child: HTMLElement) => child.textContent)).toMatchObject(['3', '2', '1'])
 })
+
+// it('move cursor to previous sibling', () => {
```
raineorshine, line 128: We should add unit tests for each of the new sort options.

You will need to come up with a good way to mock the `updated` and `created` timestamps, since in the automated test they probably all be the same by default.

src/util/compareThought.ts
raineorshine, line 1: The new functions look good. I think they could be improved by taking advantage of the nice abstractions that have been created for comparing and sorting, namely [makeCompareByProp](https://github.com/cybersemics/em/blob/4d07fb51d698ee548c1d8199668d52d5bf8a029b/src/util/makeCompareByProp.ts) and [reverse](https://github.com/cybersemics/em/blob/4d07fb51d698ee548c1d8199668d52d5bf8a029b/src/util/compareThought.ts#L177). See if you can eliminate some of the repetition by abstracting low-level comparison operations.

src/selectors/getChildren.ts
```diff
@@ -130,6 +130,56 @@ const getChildrenSortedAlphabetical = moize(
...
+        ? b.created > a.created
+          ? 1
+          : b.created < a.created
+            ? -1
+            : 0
+        : a.created > b.created
+          ? 1
+          : a.created < b.created
+            ? -1
+            : 0
```
raineorshine, line 148: Did you mean to import `compareThoughtByCreated`? The low level comparison logic should probably go there.

raineorshine (changes requested):
Hi! Thanks for the submission. Tested and works well.

src/actions/setSortPreference.ts
```diff
@@ -98,10 +98,16 @@ const setSortPreference = (
...
           sortPreference.type !== currentSortPreference.type || sortPreference.type === globalSortPreference.type
-            ? toggleAttribute({
-                path: simplePath,
-                values: ['=sort', sortPreference.type],
-              })
+            ? reducerFlow([
+                deleteAttribute({
+                  path: simplePath,
+                  value: '=sort',
+                }),
```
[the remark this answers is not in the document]
RED-ROSE515: The problem was that when changing sort options, current code wasn't properly removing the existing sort attribute before setting the new one.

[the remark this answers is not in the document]
RED-ROSE515: I updated a toggleAttribute action creator to remove the existing attribute.

raineorshine (changes requested):
See comment

src/actions/setSortPreference.ts
```diff
@@ -98,10 +98,18 @@ const setSortPreference = (
...
-                path: simplePath,
-                values: ['=sort', sortPreference.type],
-              })
+            ? reducerFlow([
+                // toggleAttribute only overrides the existing attribute if the value is different. If the value is the same, it will remove the entire attribute.
+                // So we delete the attribute first to avoid this behavior.
+                deleteAttribute({
+                  path: simplePath,
+                  value: '=sort',
+                }),
```
raineorshine, line 107: Thanks, the comment looks good.

I'm noticing that the tests still pass even if `deleteAttribute` is removed. This suggests that either it is not necessary, or that there is incomplete test coverage. Can you describe the case where `deleteAttribute` is necessary?

src/util/compareThought.ts
```diff
@@ -195,3 +196,66 @@ export const compareThought: ComparatorFunction<Thought> = (a: Thought, b: Thoug
 /** A comparator that sorts in descending order, with formatted text prioritized first. */
 export const compareThoughtDescending: ComparatorFunction<Thought> = (a: Thought, b: Thought) =>
   compareReasonableDescending(a.sortValue || a.value, b.sortValue || b.value)
+
+/** Compare two thoughts by their created timestamp in ascending order (oldest first). */
+export const compareThoughtByCreated: ComparatorFunction<Thought> = (a: Thought, b: Thought) => {
+  // If both thoughts have the same ID, they're the same thought (probably an edit)
+  if (a.id === b.id) return 0
```
raineorshine: If the thoughts are the same, then this function will return `0` anyway, so no need for a special case from what I can tell.

Removed in 72d4b7bacf9.

src/util/compareThought.ts
```diff
@@ -195,3 +196,66 @@ export const compareThought: ComparatorFunction<Thought> = (a: Thought, b: Thoug
...
+
+  // If timestamps exist, compare them
+  if (a.created && b.created) {
+    return compare(a.created, b.created)
```
raineorshine: Typescript guarantees that `created` is defined, so all the other branches besides this one are unreachable.

I've removed the extraneous logic in 72d4b7bacf9.

raineorshine (commented):
I've done some refactoring and pushed 41313685451.

src/actions/setSortPreference.ts
RED-ROSE515: When we change the sortOption, it goes to None unexpectidely. 
Because of toggleAttribute function.
So, we need to use deleteAttribute.

src/util/compareThought.ts
RED-ROSE515: Thanks, I got it.

src/util/compareThought.ts
RED-ROSE515: Got it.

src/actions/setSortPreference.ts
raineorshine: Can you provide the steps to reproduce? I don't see how this can happen with the `sortPreference.type !== currentSortPreference.type` check.

RED-ROSE515: https://github.com/user-attachments/assets/e5bab7b9-adb3-4405-a391-8d8eebcea940

You can check in this video.

raineorshine (approved):

PR #2886, opened 2025-04-07
Replace fastClick with onClick and role=button

### LINKED ISSUE #2776: Replace fastClick with role='button'
Based on the discovery in https://github.com/cybersemics/em/pull/2752#issuecomment-2594168616, it looks like `role='button'` completely removes the artificial delay added to the mobile `onClick` event as a hover fallback. Thus we should be able to remove `fastClick` completely.

Test click, touch, drag-and-drop thoroughly to see if this is a safe change.

ethan-james:
Fixes #2776 

`fastClick` can be replaced by an `onClick` handler so long as `role=button` is set on the element to disable mobile WebKit's hover detection behavior (https://github.com/cybersemics/em/pull/2752#issuecomment-2594970165).

Here is a list of every place where `fastClick` is currently used:

- [x] all action buttons
- [x] auth
  - [x] close
  - [x] log in
  - [x] sign in with google
  - [x] forget password
- [x] bullet
- [x] checkbox
- [x] chevron
- [x] welcome
- [x] close button
- [x] color picker
- [x] content
- [x] context breadcrumbs
- [x] customize toolbar
  - [x] back to settings
  - [x] close
  - [x] reset toolbar
- [x] devices
  - [x] add
  - [x] cancel
  - [x] permissions list
  - [x] add a device
  - [x] pencil
  - [x] copy
  - [x] back
  - [x] device management
- [x] divider
- [x] dropdown menu
- [x] error fallback "refresh"
- [x] expand content thought ellipses
- [x] export buttons
- [x] favorites options
- [x] feedback buttons (send & cancel)
- [x] footer?
- [x] hamburger menu
- [x] help button & breadcrumbs
- [x] home button
- [x] invites button
- [x] letter case picker
- [x] link in thought
- [x] modal close button
- [x] multithought
- [x] new subthought tip
- [x] new thought tip
- [x] new thought
- [x] search subthought More... link
- [x] settings buttons
- [x] sidebar section tabs
- [x] signup modal buttons
- [x] sort picker dropdown
- [x] start/skip/restart tutorial buttons
- [x] thought annotation url link
- [x] toolbar buttons
- [x] tutorial
  - [x] help
  - [x] hint
  - [x] navigation (step link & component)


src/components/ColorPicker.tsx
```diff
@@ -110,10 +109,10 @@ const ColorSwatch: FC<{
   return (
     <span
       aria-label={label || color || backgroundColor}
-      {...fastClick(e => {
+      onTouchEnd={e => {
         // stop click empty space
         e.stopPropagation()
-      })}
+      }}
```
trevinhofmann, line 115: Is there a reason that this PR avoids adding `role='button'` to some elements (such as this one), or is it unintentional?

src/components/ErrorBoundaryContainer.tsx
```diff
@@ -53,7 +53,7 @@ const ErrorFallback = ({ error, componentStack }: { error?: Error; componentStac
         </div>
       )}
       <pre className={css({ whiteSpace: 'normal' })}>{componentStack}</pre>
-      <a {...fastClick(() => window.location.reload())} className={cx(anchorButtonRecipe(), css({ minWidth: 0 }))}>
+      <a onClick={() => window.location.reload()} className={cx(anchorButtonRecipe(), css({ minWidth: 0 }))}>
```
trevinhofmann, line 56: Another example of an element without `role='button'`

trevinhofmann (changes requested):
Thank you for the PR, @ethan-james! An exciting change! The first two things I've noticed are:

1. Some elements are missing `role='button'`
2. In `fastClick`, `enableHaptics` was `true` by default, so most of these elements will need haptics to be added

src/components/ColorPicker.tsx
ethan-james: Unintentional. Thanks for catching them!

ethan-james: Actually, this one doesn't use a click handler so I don't think it needs `role=button`. The error boundary below is a mistake.

ethan-james:
@trevinhofmann For components like `CloseButton` that have customized haptics `onTouchStart`, do they also need the default haptics `onTouchEnd`?

trevinhofmann:
@ethan-james, can you please have a look at the failing puppeteer tests before I review the PR? Thank you very much!

ethan-james:
> @ethan-james, can you please have a look at the failing puppeteer tests before I review the PR? Thank you very much!

No problem, I was hoping they were all just flaky since they pass on my machine. It turns out I needed to add `onMouseUp` to `ColorPicker`, and now the only one that's failing genuinely is flaky.

trevinhofmann:
Thank you, @ethan-james. I re-ran the failing test, and it is now passing. I'll review this PR again in the morning.

src/components/Note.tsx
```diff
@@ -173,6 +173,7 @@ const Note = React.memo(
           }}
           onBlur={onBlur}
           onFocus={onFocus}
+          role='button'
```
trevinhofmann, line 176: Was this added because it applies to some events other than `onClick`?

trevinhofmann (changes requested):
Thank you for the updates, @ethan-james!

There are still 2 references to `fastClick` in comments which should be updated accordingly:
1. `src/components/HamburgerMenu.tsx`
2. `src/e2e/puppeteer/setup.ts`

---

Thinking about this more, keeping `fastClick` or similar function may have been a nice design choice, just simplifying the implementation of it to return onClick, haptics, and role='button' -- As opposed to consistently applying this same pattern in so many places (and making it easier to forget an attribute at times):

```
onClick={...}
onTouchEnd={haptics.light}
role='button'
```

However, this approach is perfectly fine and reasonable as-is.

src/components/Checkbox.tsx
```diff
@@ -73,10 +73,12 @@ const Checkbox: FC<
         cursor: disabled ? 'default' : 'pointer',
       })}
       // prevent the default click behavior, which autoscrolls in chrome if there is a checkbox within the label
-      {...fastClick(e => {
+      onClick={e => {
         e.preventDefault()
         onChange(e)
-      })}
+      }}
+      onTouchEnd={haptics.light}
```
raineorshine, line 80: Is there a reason for placing the `haptics` in a separate `onTouchEnd` handler? It seems more straightforward to put it in the `onClick` that already exists.

src/components/HamburgerMenu.tsx
```diff
@@ -91,8 +91,8 @@ const HamburgerMenu = () => {
           // Therefore, position the HamburgerMenu at top: 1px so that the sidebar is not accidentally opened on tab change.
```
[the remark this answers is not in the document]
ethan-james: Hmm, do you think it would be better to remove the reference to `fastClick` and keep the workaround, or test to see if the workaround is still necessary?

ethan-james:
> Thank you for the updates, @ethan-james!
> 
> There are still 2 references to `fastClick` in comments which should be updated accordingly:
> 
> 1. `src/components/HamburgerMenu.tsx`
> 2. `src/e2e/puppeteer/setup.ts`
> 
> Thinking about this more, keeping `fastClick` or similar function may have been a nice design choice, just simplifying the implementation of it to return onClick, haptics, and role='button' -- As opposed to consistently applying this same pattern in so many places (and making it easier to forget an attribute at times):
> 
> ```
> onClick={...}
> onTouchEnd={haptics.light}
> role='button'
> ```
> 
> However, this approach is perfectly fine and reasonable as-is.



> Thank you for the updates, @ethan-james!
> 
> There are still 2 references to `fastClick` in comments which should be updated accordingly:
> 
> 1. `src/components/HamburgerMenu.tsx`
> 2. `src/e2e/puppeteer/setup.ts`
> 
> Thinking about this more, keeping `fastClick` or similar function may have been a nice design choice, just simplifying the implementation of it to return onClick, haptics, and role='button' -- As opposed to consistently applying this same pattern in so many places (and making it easier to forget an attribute at times):
> 
> ```
> onClick={...}
> onTouchEnd={haptics.light}
> role='button'
> ```
> 
> However, this approach is perfectly fine and reasonable as-is.

I did think about that as I was going along. One wrinkle is that in some cases, a `fastClick`-style helper would conflict with existing event handlers. I suppose it could work like `fastClick` where it could accept handlers as arguments.

src/components/Checkbox.tsx
ethan-james: I think that should be fine, if it's OK to trigger it on mouse events as well.

raineorshine: Yes, I think it's a bit cleaner to use the existing `click` handler. The haptics are a no-op if the platform doesn't support it.

src/components/HamburgerMenu.tsx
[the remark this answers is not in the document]
ethan-james: It looks like the test setup workaround can be safely removed, but some sidebar tests fail if I change `HamburgerMenu`. It may be OK to update the snapshots. It's hard to tell, I wasn't able to reproduce the original issue quickly.

trevinhofmann (approved):
Thanks, @ethan-james! The latest changes look good to me.

I would still like to remove the reference to `fastClick` in the comment of `HamburgerMenu.tsx`, but a comment is probably not worth holding up the PR further -- especially when this touches so many components and conflicts with any other work that attempts to use `fastClick`.

raineorshine (changes requested):
After reviewing the PR I've had a change of heart. I can't shake the feeling that unwrapping all the `fastClick` instances is actually moving in the wrong direction by un-abstracting the logic and creating duplication. @trevinhofmann I think your intuition was right in https://github.com/cybersemics/em/pull/2886#pullrequestreview-2779876526.

What happened in my judgment process is that I originally believed `fastClick` would simply be replaced with `onClick`, and end up being a simple substitution. When it was brought up that we would need to preserve haptics everywhere, I underestimated the repetition. When I imagine new developers on the project, I much prefer telling them to use `fastClick` rather than having to enforce `onClick` + `role` + `haptics` everywhere.

I think the right move is to cancel this change and preserve the abstraction. We would then just need to replace the guts of `fastClick` with `onClick` + `role` + `haptics`, and make sure there is no existing `onClick` that gets overridden with the spread operator.

Please let me know if this makes sense and you're able to make the appropriate changes It is up to you if you wish to do a hard reset and force push to this PR, or start a new PR from `main`.

My apologies for the reversal, and thanks for your understanding!

PR #2887, opened 2025-04-09
Update indentDepth to 2 for thoughts in column 1 of a table

### LINKED ISSUE #2842: [iOS] Caret transform bug when moving cursor from col2 child
Tested on iOS Safari.

## Steps to Reproduce

```
- a
  - =view
    - Table
  - Boston
    - b
      - b1
  - Chicago
    - c
      - c1
```

1. Set the caret on `b1`.
2. Tap to set the caret on `Boston`.

## Current Behavior

The faux caret is not activated. The native caret is shown which does not move with the thought as autofocus shifts.

https://github.com/user-attachments/assets/cb58b24c-71d1-4d83-a8cc-c40e29626961

## Expected Behavior

The native caret should be hidden and the faux caret should be rendered at the correct selection offset while the autofocus is animating.

ethan-james:
Fixes #2842 

Table depth, a value that nudges indent depth based on whether a thought is column 1 or column 2 of a table, helps account for animations that play when navigating from one table column to another. This value used to be set to 1 for thoughts in column 1 so that the hide caret animation will play correctly when navigating from outside of a table to inside column 1 of a table.

It turns out that 1 was not the right value, since it will match the indent depth of a child thought inside column 2. It is not possible to navigate to column 1 from a grandchild of a thought inside column 2 because column 1 disappears if there's a grandchild inside column 2. So the correct table depth value for a thought in column 1 should be "2".


raineorshine (changes requested):
While it's not possible to navigate to col1 from a grandchild of a thought inside col2, the reverse *is* possible when the col2 thought is pinned. You can see this by setting the caret on `Boston` and then tapping `b1` in the example below.

```
- a
  - =view
    - Table
  - Boston
    - b
      - =pin
      - b1
        - b2
          - b3
            - b4
              - b5
                - b6
  - Chicago
    - c
      - c1
```

Let's look for a more general solution and test out the various indentation jumps that are possible to ensure it works in all cases.

src/components/LayoutTree.tsx
```diff
@@ -213,10 +212,16 @@ const LayoutTree = () => {
...
     if (state.cursor) {
-      // If the current thought is column 2 of a table, offset the indentDepth by 3
-      if (attributeEquals(state, head(rootedGrandparentOf(state, state.cursor)), '=view', 'Table')) return 3
-      // If the current thought is column 1 of a table, offset the indentDepth by 1
-      if (attributeEquals(state, head(rootedParentOf(state, state.cursor)), '=view', 'Table')) return 1
+      // Since column 2 thoughts can be nested, this must traverse the thought tree to identify a parent
+      // that is in column 2
+      for (
+        let path = rootedParentOf(state, state.cursor);
+        head(path) !== head(state.rootContext);
```
ethan-james, line 219: Is this a good condition? I saw something in `rootedParentOf` about `ABSOLUTE_PATH` but I didn't know if that was necessary here.

raineorshine: `ABSOLUTE_PATH` is used for "non-rooted" thoughts created with the Quick Thought feature, but it's not fully implemented yet, so it can be ignored for now.

Since `state.cursor` is an array, there's no need to traverse up to find the root child. You can index it directly. I believe this has the same effect:

```
state.cursor && state.cursor.length > 1 && attributeEquals(state, state.cursor[0], '=view', 'Table') ? 1 : 0
```

raineorshine (changes requested):

src/components/LayoutTree.tsx
ethan-james: The reason why it needs to iterate is because, if I understand correctly, any thought can be the root of the table, not just the thought at `state.cursor[0]`.

raineorshine: As written, the `for` loop does not check for `=table`. It simply traverses to `state.cursor[0]`. Did you intend to find the closest ancestor that is in table view?

Note that tables can also be nested, but I'm not sure what `tableDepth` should be in that case.

ethan-james: Right, it's actually the `if` statement inside the `for` loop that returns: `if (attributeEquals(state, head(path), '=view', 'Table')) return 1`. I debated whether a `for` loop or a `while` loop would be cleaner and maybe I made it too confusing.

As for nested tables, they seem to work OK. I did find an issue with non-nested tables, but I'll add it below because it takes this conversation on a tangent.

ethan-james:
Here is the issue with non-nested tables that I referenced above.

I tried a case like this:

1. Add a new thought `a`
2. Add a new thought `Boston` at the same un-indented depth
3. Add a child thought `b1` to `Boston`
4. Get the cursor back to `Boston`
5. Make `Boston` a table

I believe the tree would look like this, although I can't actually make it spit out the attributes:

```
- a
- =view
  - Table
- Boston
  - b1
```

It fails to hide the caret when moving between `Boston` and `b1`, because it doesn't seem to be able to find the attribute that tells it to use `tableDepth`, meaning that `attributeEquals(state, head(path), '=view', 'Table')` is always false. I also see some layout issues there too, so maybe this is a separate issue?

Screenshot 2025-04-15 115046

src/components/LayoutTree.tsx
raineorshine: Ah, of course! I was really not reading that right 🤦‍♀️

Maybe `.some()` a little more straightforward? It avoids the nonstandard `for` loop and mutation:

```ts
// return 1 if any ancestor is in table view
state.cursor && state.cursor.length > 1 && state.cursor.some(id =>
  attributeEquals(state, id, '=view', 'Table') 
) ? 1 : 0
```

raineorshine:
> It fails to hide the caret when moving between `Boston` and `b1`, because it doesn't seem to be able to find the attribute that tells it to use `tableDepth`, meaning that `attributeEquals(state, head(path), '=view', 'Table')` is always false.

I don't think it gets to `attributeEquals`, since the `for` loop condition is false from the beginning in that example.

> I also see some layout issues there too, so maybe this is a separate issue?

That's a separate issue. Table view works in the root, but it looks like the position of col2 is off.

ethan-james:
> I don't think it gets to `attributeEquals`, since the `for` loop condition is false from the beginning in that example.

The new `some` method doesn't start from a parent, but it still seems to have the same issue (`indentDepth` & `tableDepth` are both `0`). Is it possible that `attributeEquals` has trouble when the thought and its attribute are both at root level?

ethan-james:
> But the point is that node traversal is more appropriate for a linked list or tree, rather than an array. It's technically possible to use traversal/recursion for array operations, but that's more common in Lisp/Scheme than in JS in my experience.

Yes, that makes sense.

> An attribute always applies to its parent.

Ah, got it. It actually does need to be `parentOf(state.cursor).some`, otherwise `tableDepth` will be set to `1` for `['a1']`, erroneously triggering the `hideCaret` animation between `a` and `a1` in the following case:

```
- a
  - a1
    - =view
      - Table
    - col1
      - col2
```

> Sometimes handling children in the root requires a special case since `__ROOT__` is omitted from all `Paths` other than the root itself.

OK, so would something like `attributeEquals(state, '__ROOT__', '=view', 'Table')` be effective?

raineorshine (changes requested):
> OK, so would something like `attributeEquals(state, '__ROOT__', '=view', 'Table')` be effective?

I can confirm that `attributeEquals` works for any ThoughtId, including `__ROOT__`.

As for whether it is effective or not, I'm not sure I can answer that in a general sense. I think it depends on what case you are testing and what you are trying to achieve.

From my testing, the latest commit of f423cdc046f is effective for fixing the original issue, but the root table is still not working.

raineorshine (changes requested):
Thanks for the update!

This seems to be working in all the cases I have tested. I just have some concerns about the solution design.

- It's not obvious what `Math.min` is doing and why it's needed.
- It's not obvious why `state.cursor.length` is used in one case, but `state.cursor.length - 1` is used in another.
- The two different branches are not orthogonal. That is, it's possible for the root to be in table view AND for cursor ancestors to be in table view.

Overall, my sense is that the solution is opaque and hard to reason about. This could give us trouble in the future if we find any additional cases that are not handled. While I am content to accept the solution and move on since behaviorally it is working as expected, curiosity got the better of me and I played with a couple alternatives. I would love your opinion on these.

# 1. Use strict cursor depth instead of `indentDepth`

So far we have created the faux caret depth by adding `indentDepth + tableDepth`. However, I notice that `indentDepth` does not differentiate between root children and grandchildren, and it has some complexity related to leaf nodes. It might be simpler to use a strict cursor depth (`state.cursor.length`). FauxCaret just needs different integer increments for each level. It does no harm if the faux caret depth changes when there is no change to the actual indentation, and it might simplify things. Thus `state.cursor.length` can be used as-is for the faux caret depth.

The only thing that doesn't handle is the way the depth changes when changing from normal view to table view. In that case, the indentation level changes, but the cursor does not move, so the cursor depth does not change. If we simply add the number of ancestor tables to the strict cursor depth, we get unique integers at each depth and a bump when switching between views.

While this solution still has some complexity, I find the logic of `depth + numAncestorTables` to be more intuitive than the `Math.min` logic.

```diff
diff --git a/src/components/LayoutTree.tsx b/src/components/LayoutTree.tsx
index 6f22d461d35..bd2c0d77abf 100644
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -18,7 +18,6 @@ import reactMinistore from '../stores/react-ministore'
 import scrollTopStore from '../stores/scrollTop'
 import viewportStore from '../stores/viewport'
 import head from '../util/head'
-import parentOf from '../util/parentOf'
 import HoverArrow from './HoverArrow'
 import TreeNode from './TreeNode'
 
@@ -209,18 +208,6 @@ const LayoutTree = () => {
   const cliffPadding = fontSize / 4
   const cliffPaddingStyle = useMemo(() => ({ paddingBottom: cliffPadding }), [cliffPadding])
 
-  // When the cursor is in a table, all thoughts beneath the table are hidden,
-  // so there is no concern about animation name conflicts with subsequent (deeper) thoughts.
-  const tableDepth = useSelector(state => {
-    if (!state.cursor) return 0
-    // return 1 if any ancestor is in table view
-    // rootedParentOf only includes HOME_TOKEN for top-level thoughts, so we can't use it
-    if (attributeEquals(state, HOME_TOKEN, '=view', 'Table')) return Math.min(3, state.cursor.length)
-    return state.cursor.length && parentOf(state.cursor).some(id => attributeEquals(state, id, '=view', 'Table'))
-      ? Math.min(3, state.cursor.length - 1)
-      : 0
-  })
-
   const { indentCursorAncestorTables, treeThoughtsPositioned, hoverArrowVisibility } = usePositionedThoughts(
     treeThoughts,
     {
@@ -230,6 +217,16 @@ const LayoutTree = () => {
     },
   )
 
+  /** Calculate the faux caret depth as the depth of the cursor + number of ancestor tables. This ensures that the correct FauxCaret animation is used when navigating down the hierarchy and when switching from normal view to table view, which shifts the indentation without the cursor depth changing.  */
+  const fauxCaretDepth = useSelector(state => {
+    if (!state.cursor) return 0
+    const depth = state.cursor.length
+    const numAncestorTables = [HOME_TOKEN, ...state.cursor].filter(id =>
+      attributeEquals(state, id, '=view', 'Table'),
+    ).length
+    return depth + numAncestorTables
+  })
+
   // The indentDepth multipicand (0.9) causes the horizontal counter-indentation to fall short of the actual indentation, causing a progressive shifting right as the user navigates deeper. This provides an additional cue for the user's depth, which is helpful when autofocus obscures the actual depth, but it must stay small otherwise the thought width becomes too small.
   // The indentCursorAncestorTables multipicand (0.5) is smaller, since animating over by the entire width of column 1 is too abrupt.
   // (The same multiplicand is applied to the vertical translation that crops hidden thoughts above the cursor.)
@@ -247,7 +244,7 @@ const LayoutTree = () => {
         css({
           marginTop: '0.501em',
         }),
-        fauxCaretTreeProvider(indentDepth + tableDepth),
+        fauxCaretTreeProvider(fauxCaretDepth),
       )}
       style={{
         // add a full viewport height's space above to ensure that there is room to scroll by the same amount as spaceAbove
```

# 2. Use the same indent that `transform` uses.

Another approach is to reason backwards from the final indentation that is used by `transform`. The final indentation is `indent`. We can convert it to an integer step function with `Math.ceil`, thus directly mapping the final indentation level to the faux caret depth. This removes all the complexity related to calculating different indentation edge cases.

```diff
diff --git a/src/components/LayoutTree.tsx b/src/components/LayoutTree.tsx
index 6f22d461d35..37bf90ecb68 100644
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -10,7 +10,6 @@ import testFlags from '../e2e/testFlags'
 import usePositionedThoughts from '../hooks/usePositionedThoughts'
 import useSizeTracking from '../hooks/useSizeTracking'
 import fauxCaretTreeProvider from '../recipes/fauxCaretTreeProvider'
-import attributeEquals from '../selectors/attributeEquals'
 import { hasChildren } from '../selectors/getChildren'
 import linearizeTree from '../selectors/linearizeTree'
 import nextSibling from '../selectors/nextSibling'
@@ -18,13 +17,11 @@ import reactMinistore from '../stores/react-ministore'
 import scrollTopStore from '../stores/scrollTop'
 import viewportStore from '../stores/viewport'
 import head from '../util/head'
-import parentOf from '../util/parentOf'
 import HoverArrow from './HoverArrow'
 import TreeNode from './TreeNode'
 
 /** The padding-bottom of the .content element. Make sure it matches the CSS. */
 const CONTENT_PADDING_BOTTOM = 153
-const HOME_TOKEN = '__ROOT__' as ThoughtId
 
 /** A computed store that tracks the bottom of the viewport. Used for list virtualization. Does not include overscroll, i.e. if the user scrolls past the top of the document viewportBottom will not change. */
 const viewportBottomStore = reactMinistore.compose(
@@ -209,18 +206,6 @@ const LayoutTree = () => {
   const cliffPadding = fontSize / 4
   const cliffPaddingStyle = useMemo(() => ({ paddingBottom: cliffPadding }), [cliffPadding])
 
-  // When the cursor is in a table, all thoughts beneath the table are hidden,
-  // so there is no concern about animation name conflicts with subsequent (deeper) thoughts.
-  const tableDepth = useSelector(state => {
-    if (!state.cursor) return 0
-    // return 1 if any ancestor is in table view
-    // rootedParentOf only includes HOME_TOKEN for top-level thoughts, so we can't use it
-    if (attributeEquals(state, HOME_TOKEN, '=view', 'Table')) return Math.min(3, state.cursor.length)
-    return state.cursor.length && parentOf(state.cursor).some(id => attributeEquals(state, id, '=view', 'Table'))
-      ? Math.min(3, state.cursor.length - 1)
-      : 0
-  })
-
   const { indentCursorAncestorTables, treeThoughtsPositioned, hoverArrowVisibility } = usePositionedThoughts(
     treeThoughts,
     {
@@ -247,7 +232,7 @@ const LayoutTree = () => {
         css({
           marginTop: '0.501em',
         }),
-        fauxCaretTreeProvider(indentDepth + tableDepth),
+        fauxCaretTreeProvider(Math.ceil(indent)),
       )}
       style={{
         // add a full viewport height's space above to ensure that there is room to scroll by the same amount as spaceAbove
```

I would love to get your opinion on these approaches, starting with the latter since it is the simplest. They could use some additional testing to ensure I am not missing anything. If it is inadequate, we can fall back to the next simplest approach that has the correct behavior.

ethan-james:
> 2. Use the same indent that transform uses.

That was my dream from the beginning, although the issue with `Math.ceil` is that `indent` increases by < 1 per step, so I think the first collision occurs at the jump between 2.1 and 2.8. I'll have to test it out a bit more, probably in a codepen combined with below.

My ideal solution would be to use the decimal value as part of the class name (`hideCaret2.1`) and match the animation using `[class^=hideCaret]`, but there are 2 things standing in my way:

1. I don't know how to do that in PandaCSS
2. I would need to test that the animation actually re-runs, given that the selector does not change

This would also be nice as we could get rid of the `hideCaretX` animation definitions.

> - It's not obvious what Math.min is doing and why it's needed.

I don't like that either, but it proved necessary because:

a. `indentDepth` doesn't change for the first 2-3 children
b. adding `state.cursor.length` to `indentDepth` would increase total depth by 2 and would cause collisions (after the depth wrapped at 16) between column 1 and deeper-nested children in column 2

> - It's not obvious why state.cursor.length is used in one case, but state.cursor.length - 1 is used in another.

I know that I had to use `state.cursor.length` or else top-level thoughts in tables would get `tableDepth: 0`. I forget if `state.cursor.length - 1` solves any problems other than being cleaner for non-root-level thoughts.

> 1. Use strict cursor depth instead of indentDepth

This should be a straightforward solution as long as you're OK with the faux caret showing more often than necessary. During my testing I occasionally had difficulty telling whether the animation was playing or not, because a lot of cases are very close. But definitely not all cases.

I think you have it right that cursor depth by itself is sufficient except when toggling into and out of table view.

ethan-james:
> Yes, I wondered about that. Maybe a slight variation on `indent` would work? `indentDepth + Math.ceil(indentCursorAncestorTables / fontSize)`
> 
> FYI I think the first collision occurs at 10:
> 
> ```ts
> indentDepth = 1
> while (true) {
>     if (Math.ceil(indentDepth * 0.9) === Math.ceil((indentDepth - 1) * 0.9)) {
>         console.log('collision', indentDepth)
>         break;
>     }
>     indentDepth++
> }
> ```

Yes, that sounds right, although I think the table ancestor depth offsets it by a different amount on top of indent depth.

> I'm leaning towards using your solution if it is the more accurate one (fewer/no false positives), and keeping the other solution(s) in mind in case we run into regressions in the future.

Sounds good to me. I'm also happy to look into the feasibility of `[class^=hideCaret]` now or in the future.

raineorshine (approved):

PR #2888, opened 2025-04-10
2779 bug clear thought remove note focus

### LINKED ISSUE #2779: Clear Thought on note should set selection to thought
## Steps to Reproduce

1. Create a thought. 
2. Create a note. 
3. Activate Clear Thought. 

## Current Behavior

The thought enters clearThought mode, but the caret stays on the note. 

## Expected Behavior

When clearThought is activated, note editing should be deactivated and the caret should move to the cleared thought.

ethan-james:
Fixes #2779

When executing `clearThought` while `noteFocus` is true, call `setCursor` with `noteFocus: false` before calling `clearThought` so that the cursor is on the main thought when it is cleared.

### DIFF
--- a/src/actions/cursorCleared.ts
+++ b/src/actions/cursorCleared.ts
@@ -15,6 +15,8 @@ const cursorCleared = (state: State, { value }: { value: boolean }): State => ({
   editableNonce: state.editableNonce + 1,
   // manually set edit mode to true since the cursor is not changing and normally setCursor handles this
   editing: value,
+  // clearing the cursor should always move the caret into the empty thought
+  noteFocus: false,
 })
 
 /** Temporary clear the cursor in response to the clearThought. NOOP if value is no different than current state. See actions/cursorCleared. */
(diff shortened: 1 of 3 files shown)


raineorshine (commented):
Thanks, works well!

Would it be possible to move this into the `clearThought` reducer? And can we just set `noteFocus: false` there, or do we need to call `setCursor`? 

Generally we want to push as much logic as possible down to the reducer, and only use action-creators for side effects.

ethan-james:
> Thanks, works well!
> 
> Would it be possible to move this into the `clearThought` reducer? And can we just set `noteFocus: false` there, or do we need to call `setCursor`?
> 
> Generally we want to push as much logic as possible down to the reducer, and only use action-creators for side effects.

Looks like it worked! I assumed that it would need a separate action to set the selection, but `useEditMode` is already watching `noteFocus`.

raineorshine (approved):

Issue #2841, opened 2025-03-07
[iOS] Caret does not move on second tap after note


raineorshine:
Tested on iOS Safari.

Unable to reproduce in Chrome Device Emulator.

Related: #2840 (but note the different platforms)

## Steps to Reproduce

```
- a
  - b
    - =note
      - Hello world
```

1. Set the caret on the note.
2. Set the caret at the end of `a`.
3. Swipe ← to activate Cursor Forward.

## Current Behavior

The caret remains on `a`.

## Expected Behavior

The caret should move to the cursor thought, `b`.

raineorshine:
Fixed in #2917.

PR #2889, opened 2025-04-11
Move caret to parent thought on cursorBack

### LINKED ISSUE #2840: [Mobile] Caret is not moved from note to thought on swipe-to-navigate
Tested on iOS Safari and Chrome Device Emulator.

Related: #2841 (but note the different platforms)

## Steps to Reproduce

```
- a
  - b
    - =note
      - Hello world
    - c
```

1. Set the cursor on `b`.
2. Tap `b` to enter edit mode.
3. Set the caret on the note.
4. Swipe → to activate Cursor Back.
   - Or swipe ← to activate Cursor Forward.

## Current Behavior

The caret remains on the note.

## Expected Behavior

The caret should move to the cursor thought.

e.g.

- Cursor Back: The cursor will move to `a`, and the caret should be set on `a`.
- Cursor Forward: The cursor will move to `c`, and the caret will be set on `c`.

## Testing

Add a [caret](https://github.com/cybersemics/em/blob/0764c033e346ec3dd6c10fd2b5c457a79cafc645/src/e2e/puppeteer/__tests__/caret.ts) test.

ethan-james:
Fixes #2840 

When `noteFocus` is true, invoking `cursorBack` should move the caret up to the cursor thought, and `cursorForward` should move the caret down to the next thought.

In `useEditMode`:

```
      // if there is no browser selection, do not manually call selection.set as it does not preserve the cursor offset. Instead allow the default focus event.
      const cursorWithoutSelection = editingCursorOffset !== null || !selection.isActive()
````

will be `false` if `cursorOffset` is `null`. The `cursorBack` and `cursorForward` actions currently move the caret to the beginning of the new cursor thought, so `offset: 0` preserves the existing behavior.

### CODE REVIEW

**Comment by ethan-james** [comment]
I'm not quite sure that I was able to reproduce what you were seeing. You can look at the tests and check my acceptance criteria (`[a, b]` after `cursorBack` and `[a, b, c]` after `cursorForward`) to make sure that the behavior is correct, and in line with what you're seeing in the app itself.

**Review by raineorshine — CHANGES_REQUESTED** [review]
That's not quite right. The caret should always follow the cursor thought. In other words, Cursor Forward/Backward should behave exactly the same regardless of whether the selection is on the thought proper or the thought's note.

In the original issue, thus:

- Cursor Back: The cursor will move to `a`, and the caret should be set on `a`.
- Cursor Forward: The cursor will move to `c`, and the caret will be set on `c`.

I have updated the expected behavior in the issue. Hope that helps!

> You can look at the tests and check my acceptance criteria (`[a, b]` after `cursorBack` and `[a, b, c]` after `cursorForward`) to make sure that the behavior is correct, and in line with what you're seeing in the app itself.

This is a bug with the browser selection, not the Redux state, so unfortunately the test you wrote won't cover the behavior. The browser selection is only polyfilled in JSDOM. We have a [caret](https://github.com/cybersemics/em/blob/0764c033e346ec3dd6c10fd2b5c457a79cafc645/src/e2e/puppeteer/__tests__/caret.ts) puppeteer test that might be a good place for this.

**Comment by ethan-james** [comment]
Thank you! I added puppeteer tests that pass without any changes. You can check them and make sure that I have it right this time. I think I did see the bug you report where the caret did not leave the note, but I haven't been able to reproduce it effectively. I'll take another crack at it tomorrow.

**Comment by raineorshine** [comment]
The puppeteer test looks good overall, but it is not following the exact steps in the issue. Specifically, the starting position of the cursor and the timing of entering edit mode are different. These ends up being important for reproducing the bug.

**Comment by ethan-james** [comment]
> The puppeteer test looks good overall, but it is not following the exact steps in the issue. Specifically, the starting position of the cursor and the timing of entering edit mode are different. These ends up being important for reproducing the bug.

How do they look now?

**Comment by raineorshine** [comment]
Yes, that is correct as far as the steps are concerned. Unfortunately, it looks like the tests are passing in puppeteer, even though the bug is consistently reproducible in both Chrome and Safari. I'm not sure if it is a timing difference, import difference (since paste is emulated), or something else.

**Comment by raineorshine** [comment]
Oops, we are missing something rather obvious in hindsight: The puppeteer test is desktop, but the bug is only reproducible on mobile.

We can either try to find a way to simulate mobile (which should be possible given that I am able to reproduce the bug in Chrome Device Emulator) or forego the puppeteer test and just test manually. I'm okay with either.

**Comment by ethan-james** [comment]
> Oops, we are missing something rather obvious in hindsight: The puppeteer test is desktop, but the bug is only reproducible on mobile.
> 
> We can either try to find a way to simulate mobile (which should be possible given that I am able to reproduce the bug in Chrome Device Emulator) or forego the puppeteer test and just test manually. I'm okay with either.

The new `describe` is actually nested inside `describe('mobile only')` because I wanted to do a "when -> it" style of test naming. I tried explicitly adding

```
    beforeEach(async () => {
      await emulate(KnownDevices['Pixel 5'])
    }, 5000)
```

inside my `describe`, and it seems to still pass.

I'm going to switch over to #2841 because I can reproduce it more easily, and hopefully I'll learn something.

raineorshine:
How strange! Unfortunately puppeteer is kind of a black box within docker. I'm not sure what's going on there.

raineorshine (changes requested):
Thanks, works as expected! Clean and direct solution.

Unfortunately the tests pass even when I revert the fix, so they don't actually cover the behavior. Are you also seeing that? 

For the sake of time, I think it would be best to remove all the test-related code rather than try to wrangle puppeteer for this test case.

ethan-james:
Yes, that was part of the trouble I had reproducing it. I was never able to get the tests to fail. I'll remove them.

raineorshine (approved):

Issue #2831, opened 2025-03-03
Sort: Note


raineorshine:
## Expected Behavior

Add a new sort option to the SortPicker:

- Note

When sorting by Note, sort by the value of the first child of the thought's `=note` attribute:

Before:

```
- a
  - =note
    - 2
- b
  - =note
    - 3
- c
  - =note
    - 1
```

After:

```
- c
  - =note
    - 1
- a
  - =note
    - 2
- b
  - =note
    - 3
```

Additional Requirements:

- [x] If current sort is not Note, change sort to Note (asc).
- [x] Toggle sort by Note between Ascending and Descending when clicked multiple times:
    - If current sort is Note (asc) and Note is clicked, sort by Note (desc).
    - If current sort is Note (desc) and Note is clicked, sort by Note (asc).
- [x] Perform correct Lottie animation regardless of which sort option from/to.

## Testing

Add SortPicker tests for sorting Note.

PR #2890, opened 2025-04-15
Add sort note option

RED-ROSE515:
This PR resolves the issue : Sort: Note https://github.com/cybersemics/em/issues/2831

In this Pr, I added a new Sort Option Note.
And updated compareThought and editThought functions for Note Sort, same as previous sort options.


raineorshine:
Thanks! I will wait for #2883 to be merged since this is dependent and will need to be merged.

I am noticing similar redundancy issues in this PR. Thought comparison logic is aggressively abstracted, so you should not be doing low level operations like `return 1`. I encourage you to learn from the existing code and get this cleaned up once #2883 is merged.

PR #2891, opened 2025-04-16
Multiselect: Undo/redo executeCommandWithMulticursor all at once

### LINKED ISSUE #2854: Multiselect: Undo/redo executeCommandWithMulticursor all at once
## Steps to Reproduce

```
- a
- b
- c
- d
```

1. Select `a`, `b`, `c`
2. Move Thought Down
3. Undo

## Current Behavior

Only `a` is moved back up.

```
- a
- d
- b
- c
```

## Expected Behavior

`a`, `b`, `c` should move back up as a group.

A single undo/redo command should reverse all commands executed at once with multiselect.

```
- a
- d
- b
- c
```

## Testing

Add unit tests.

## Notes

Check out the logic for navigational actions, which are grouped together in the undo history.

https://github.com/cybersemics/em/blob/20f6446613e31b0642f87d2e3db54071d87201c8/src/redux-enhancers/undoRedoEnhancer.ts#L349-L355

karunkop:
Close #2854 

## Root Cause

In `executeCommand.ts, `the multicursor implementation was clearing `state.multicursor `at the beginning of each execution loop and restoring it at the end. However, the undoRedoEnhancer had no way to know these separate operations were part of a single multicursor command, so it treated each iteration as a separate undoable action.

## Solution

- Added a new state variable `setIsMulticursorExecuting` that keeps track of when the multicursor operation starts/ends.
-  `isMulticursorExecuting` flag tracks when these operations are active, enabling related actions to be grouped into a single undo patch.
- The `isFirstMulticursorAction` flag ensures a clean break from unrelated actions by forcing the creation of a new patch at the start of a multicursor sequence, so only the relevant thoughts are affected. 
-  Actions like `toggleMulticursor` and `addMulticursor` were added to NAVIGATION_ACTIONS and made undoable, allowing cursor selections to be grouped with other navigation actions.


src/actions/setIsMulticursorExecuting.ts
```diff
@@ -0,0 +1,30 @@
+import ActionType from '../@types/ActionType'
+import CommandId from '../@types/CommandId'
+import State from '../@types/State'
+import Thunk from '../@types/Thunk'
+
+interface SetIsMulticursorExecutingPayload {
+  value: boolean
+  commandType?: ActionType | CommandId
```
raineorshine, line 8: `ActionType` and `CommandType` are very different things. I don't think we should ever union that together.

Please try to find a different approach that does not combine these types.

raineorshine (changes requested):
Thanks for your submission!

I've tested it out and the behavior works well. Great job. The tests also look very good.

Was there a reason that you did not use the navigation grouping logic, as recommended in the original issue? While suggested approaches are not strictly required, I do ask that you investigate and provide justifications for/against them.

The additional mutable state (`multicursorStartState`) and the use of `cloneDeep` is a bit of a red flag. `immer` is intentionally used for efficient cloning, so I don't think we want to resort to heavy deep clones. The `undoRedoEnhancer` is doing a lot of the work of diffing and generating patches. I wonder if there might be a way for you to utilize the existing logic.

karunkop:
> Thanks for your submission!
> 
> I've tested it out and the behavior works well. Great job. The tests also look very good.
> 
> Was there a reason that you did not use the navigation grouping logic, as recommended in the original issue? While suggested approaches are not strictly required, I do ask that you investigate and provide justifications for/against them.

Yeah sure here is my explanation. The existing navigation grouping logic in undoRedoEnhancer can't support multicursor operations due to key limitations. It groups only consecutive actions of the same type, while multicursor operations involve mixed actions like moveThoughtDown and setCursor. The execution includes intermediate steps (e.g., cursor updates) that shouldn’t be part of undo history. Since each step may affect the next, patch-based grouping is unreliable. A snapshot-based approach ensures atomicity and consistency, treating the entire multicursor operation as a single undoable unit.

karunkop:
> The additional mutable state (`multicursorStartState`) and the use of `cloneDeep` is a bit of a red flag. `immer` is intentionally used for efficient cloning, so I don't think we want to resort to heavy deep clones. The `undoRedoEnhancer` is doing a lot of the work of diffing and generating patches. I wonder if there might be a way for you to utilize the existing logic.
> 
> Let me know what you think. Thanks!

While the concern is valid, the use of multicursorStartState is a necessary trade-off to ensure consistent undo /redo behavior. It captures a complete before-and-after state, preventing inconsistencies caused by intermediate steps during multicursor operations. Importantly, this state is confined to the enhancer closure and does not pollute the Redux state, keeping the overall complexity low. The existing navigation grouping logic is fundamentally incompatible with multicursor operations, which involve interleaved and interdependent actions across multiple thoughts, as previously explained. Therefore, this solution is technically essential to guarantee atomic updates, manage state dependencies, accurately label operations, and improve performance by producing a single diff at the end rather than merging multiple patches.

This is my perspective on the concern, but I’d genuinely appreciate your feedback and thoughts on it.

raineorshine:
Hi, thanks for your response.

> The existing navigation grouping logic in undoRedoEnhancer can't support multicursor operations due to key limitations. It groups only consecutive actions of the same type, while multicursor operations involve mixed actions like moveThoughtDown and setCursor.

That's not true, it groups consecutive actions of a [variety of types](https://github.com/cybersemics/em/blob/038859e29e3fe1d9cec3b8cefe5b290c05cef03d/src/redux-enhancers/undoRedoEnhancer.ts#L354-L358): `editThought`, `closeAlert`, `cursorBack`, `cursorBeforeSearch`, `cursorDown`, `cursorForward`, etc.

> The execution includes intermediate steps (e.g., cursor updates) that shouldn’t be part of undo history.

Cursor updates are, and should be, be part of the undo history. They are [merged together](https://github.com/cybersemics/em/blob/038859e29e3fe1d9cec3b8cefe5b290c05cef03d/src/redux-enhancers/undoRedoEnhancer.ts#L375) into a single patch. Then undo/redo will pop [two patches](https://github.com/cybersemics/em/blob/038859e29e3fe1d9cec3b8cefe5b290c05cef03d/src/redux-enhancers/undoRedoEnhancer.ts#L267) off the undo history to ensure that the last edit + navigation are undone together.

> Since each step may affect the next, patch-based grouping is unreliable. A snapshot-based approach ensures atomicity and consistency, treating the entire multicursor operation as a single undoable unit.

As far as I can tell, this is just a question of [associativity](https://en.wikipedia.org/wiki/Associative_property).

Your approach:

A
...
A + B + C + D

My suggestion is to use an inductive approach:

A + B
A + B + C
A + B + C + D

i.e. `(1 + 2 + 3 + 4) === (((1+2) + 3) + 4)`

I'm speculating, but it seems that the inductive approach would avoid the need to store the state at the beginning. I'm skeptical  of your assertion that combining patches one at a time is somehow inconsistent.

Your approach is *relatively* low complexity, so it's not a huge problem. But it seems unnecessary to use two different methods of merging patches when the associative property provides an elegant solution to both.

raineorshine:
Note that `state.multicursors` should be included in the undo, just like `state.cursor`:

## Steps to Reproduce

1. Create thoughts `a`, `b`, and `c`.
2. Select `b` and `c`.
3. Indent
4. Undo [Move Cursor Forward aka Indent]
5. Undo [Edit Thought]
6. Undo [New Thought]

## Current Behavior

Multicursor still has two thoughts selected, even though we have activated `undo` to a point in history before multicursor was activated.

## Expected Behavior

On Step 5, it should return to the state before Edit Thought, with no multicursors:

```
- a
- b
- [cursor on empty thought]
```

src/actions/setIsMulticursorExecuting.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/** Action-creator for setIsMulticursorExecuting. */
+export const setIsMulticursorExecutingActionCreator =
+  (payload: SetIsMulticursorExecutingPayload): Thunk =>
+  dispatch => {
+    dispatch({
+      type: 'setIsMulticursorExecuting',
+      value: payload.value,
+      operationLabel: payload.operationLabel,
```
raineorshine, line 25: You can use the spread operator instead of enumerating each payload property, as we do in other action-creators.

https://github.com/cybersemics/em/blob/ba2d8843010ccb39adfcb643b98b2f61e2503e47/src/actions/join.ts#L85-L89

src/actions/setIsMulticursorExecuting.ts
```diff
@@ -0,0 +1,29 @@
+import State from '../@types/State'
+import Thunk from '../@types/Thunk'
+
+interface SetIsMulticursorExecutingPayload {
+  value: boolean
+  /** Label for the undo operation (typically the command name). */
+  operationLabel?: string
```
raineorshine, line 7: Let's call this `undoLabel` since it is only used that in that specific context.

src/commands/__tests__/undo-redo.ts
```diff
@@ -191,6 +196,130 @@ describe('undo', () => {
...
+  - c
+  - e`
+    expect(exported).toEqual(expectedOutput)
+
+    // Now perform one undo operation
+    store.dispatch(undo())
+
+    // On main branch, this would only move thought 'c' back up,
+    // leaving 'a' and 'b' in their moved positions
+    // With our fix, all three thoughts should be restored
```
raineorshine, line 235: We usually want to avoid referencing branches or old versions of `main`, as this comment will be invalidated as soon as this PR is merged.

The title of the test already describes this well: "undo should restore all thoughts after a multicursor moveThoughtDown operation"

src/commands/__tests__/undo-redo.ts
```diff
@@ -395,3 +524,66 @@ describe('grouping', () => {
     expect(exported).toEqual(expectedOutput)
   })
 })
+
+/******************************************************************
+ * MULTICURSOR - Undo/Redo.
+ ******************************************************************/
+
+describe('multicursor undo/redo', () => {
```
raineorshine, line 532: You can remove `undo/redo` since this is a given since we are in the `undo-redo.ts` test file.

src/redux-enhancers/undoRedoEnhancer.ts
```diff
@@ -305,6 +320,9 @@ const undoRedoReducerEnhancer: StoreEnhancer<any> =
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   <A extends Action<any>>(reducer: (state: any, action: A) => any, initialState: any): Store<State, A> => {
     let lastActionType: ActionType
+    let isMulticursorExecuting = false
+    let multicursorOperationLabel: string | null = null
+    let isFirstMulticursorAction = true // Flag to track if this is the first action in a multicursor sequence
```
raineorshine, line 325: These global state variables all need to be seriously reconsidered. As you know, adding additional state is bad for complexity, and should only be done if absolutely necessary.

For example, we already have `state.isMulticursorEditing`, and `undoRedoEnhancer` has access to the current and previous state. The need for a parallel `isMulticursorExecuting` variable that is highly questionable.

`multicursorOperationLabel` can possibly be stuffed into the first patch to avoid tracking it separately.

`isFirstMulticursorAction` appears to be capturing a part of the state machine in a variable to control program flow. It's usually better to compute these things on the fly rather than adding state.

src/redux-enhancers/undoRedoEnhancer.ts
```diff
@@ -354,9 +388,36 @@ const undoRedoReducerEnhancer: StoreEnhancer<any> =
...
         (isNavigation(actionType) && isNavigation(lastActionType)) ||
         (lastActionType === 'editThought' && actionType === 'editThought') ||
-        actionType === 'closeAlert'
+        actionType === 'closeAlert' ||
+        // For multicursor operations, group all undoable actions
+        (isMulticursorExecuting && isUndoable(actionType))
       ) {
-        lastActionType = actionType
+        // Handle multicursor operations
+        if (isMulticursorExecuting) {
```
raineorshine, line 396: You've created a separate branch of code to handle the multicursor, with its own return statement, even though it is doing the largely the same things as the navigation grouping: calculating a diff and then replacing the last undo patch. 

It would be better consolidate this logic, preferably after the additional state variables have been cleaned up.

raineorshine (changes requested):
Thanks, the behavior works really well from my testing.

I think we're on the right track with the inductive approach.

src/redux-enhancers/undoRedoEnhancer.ts
```diff
@@ -354,9 +388,36 @@ const undoRedoReducerEnhancer: StoreEnhancer<any> =
...
+            isFirstMulticursorAction = false
+
+            // Create a new undo patch for the first multicursor action
+            const undoPatch = diffState(newState as Index, state)
+
+            return undoPatch.length
+              ? {
+                  ...newState,
+                  redoPatches: [],
+                  undoPatches: [...newState.undoPatches, addActionsToPatch(undoPatch, [lastActionType])],
```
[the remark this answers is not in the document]
raineorshine: I see, it's just for the first multicursor action. The subsequent multicursor actions do get merged.

Still, I wonder if this could be more consolidated. Pushing the conditionals further "down" may allow you to eliminate the duplication.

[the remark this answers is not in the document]
karunkop: Refactor has been done :)

src/redux-enhancers/undoRedoEnhancer.ts
karunkop: Yeah, the concern makes sense, but the `multicursorOperationLabel` which now renamed to `multicursorUndoLabel`, is actually pretty important. It specifically captures the exact action right after a multicursor operation, which we then mark as the lastActionType.

image

If you check the logs in the screenshot, you’ll see how `multicursorUndoLabel` holds the correct action, while the general `undo label ` is just storing whatever's happening on the fly. That’s a problem because sometimes other actions sneak in between and end up in the undo alert when they shouldn’t.

Also, the `isFirstMulticursorAction` flag is key here. It helps us draw a clean line and create a proper undo patch for the first multicursor action (this has already been discussed in another comment thread and the PR description too.) As for why these are global variables: it’s just to make sure we can flip the flags at the right times. For eg: We set `isFirstMulticursorAction` to true at the start of a multicursor op and then reset it to false once we’re done processing. This gives us the clean separation we need and avoids things getting mistakenly merged into earlier actions.

raineorshine: > Also, the `isFirstMulticursorAction` flag is key here. It helps us draw a clean line and create a proper undo patch for the first multicursor action (this has already been discussed in another comment thread and the PR description too.) As for why these are global variables: it’s just to make sure we can flip the flags at the right times. For eg: We set `isFirstMulticursorAction` to true at the start of a multicursor op and then reset it to false once we’re done processing. This gives us the clean separation we need and avoids things getting mistakenly merged into earlier actions.

I'm not convinced. You have direct access to the current and previous state. Why would you need additional state? You're changing a simple, declarative problem into a complex, procedural problem.

Please try again and see what you can come up with. If you're really stuck, let me know and I'll offer more help.

karunkop: While we do have the direct access to the current and previous state and we might think that detecting `!state.isMulticursorExecuting && newState.isMulticursorExecuting would` work, however, this transition happens with the `setIsMulticursorExecuting` action, which itself is not undoable. The first undoable action that follows needs to know it's first in the sequence. Please see the screenshot attached:

image

In the sequence, this is followed by `setCursor`, `moveThoughtDown`, etc.

The transition happens at the `setIsMulticursorExecuting` point, but we need to mark the first undoable action that comes after this.

For a concrete example from the screenshot:

1.`setIsMulticursorExecuting` occurs at +30:00.95, setting the flag to true.
2.`setCursor` at +30:00.00 - This is the first undoable action while multicursor is true.
3.Multiple `moveThoughtDown` actions follow.

If we used `!state.isMulticursorExecuting && newState.isMulticursorExecuting`, we'd only detect this at the first action (the non-undoable setIsMulticursorExecuting), but not at the first undoable action that needs to create the undo patch. **`Also the flag to track first action in a sequence needs to persist across multiple reducer calls.`**

### Small Refactor

I agree that `isFirstMulticursorAction` naming can be quite confusing and the need of resetting/initialising it won't be ideal. So i have refactored the code to make use of `lastProcessedAction` flag that is used to track the last processed action for sequence detection and it is also a global variable because it needs to persist across multiple reducer calls. Redux handles each action on its own, so to track changes like when we enter multicursor mode, we need to remember some info across different actions. Essentially this flag help to identify when we're at the beginning of a multicursor operation sequence to create proper undo boundaries.

Use case from the above screenshot:

- `setIsMulticursorExecuting` action is dispatched (enabling multicursor mode)
- Next, a series of actions like setCursor, moveThoughtDown, etc. follow
- By tracking `lastProcessedAction`, we detect the first undoable action after multicursor mode begins.

Similar explanation can be made for `isMulticursorUndoLabel` as well. The label needs to persist across multiple dispatched actions that form a single logical operation. Essentially it groups multiple actions that happen during multicursor execution under a single semantic label in the undo history.

Let me know if this explanation aligns with your high level analyzation. If not i would hugely appreciate some help and would like to know what i am missing here! :)

raineorshine: If the additional state is needed because `isFirstMulticursorAction` is non-undoable, maybe it should be undoable? It seems like the current and previous state combined with the right accumulator would provide everything needed to merge each of the multicursor commands.

In my opinion, this is a lot of justification for a code smell that still smells. I'll take a look at it and see if I can come up with a more elegant solution.

raineorshine (approved):
If multicursor actions are marked as undoable navigation actions, then it looks like we just need to add `|| state.isMulticursorExecuting` to the merge condition to get the desired behavior. All multicursor actions are merged into the `setIsMulticursorExecuting`  patch and are treated as a single undo/redo step.

The only other change is to use `action.undoLabel` as the patch's `ActionType` so that the alert is correct.

The tests pass and it seems to work as expected when testing manually, but let me know if you see any issues with this approach.

PR #2893, opened 2025-04-17
ThoughtAnnotation: fallback to zero-width space instead of empty string

### LINKED ISSUE #2892: Faux Caret incorrect position when Training Mode is disabled
## Steps to Reproduce

1. Go into Settings and disable Training Mode. 
2. Create deeply nested subthoughts to activate the Faux Caret. 

image

## Current Behavior

The Faux Caret is rendered too low. 

https://github.com/user-attachments/assets/8fd12e0d-f253-40f6-bbe0-d4e0f4949f2b

## Expected Behavior

The Faux Caret should be rendered in the correct position regardless of Training Mode. 

## Notes

Training Mode affects a few different behaviors, but the only relevant one is likely the blank placeholder for new thoughts.

ethan-james:
Fixes #2892

The empty span in the `ThoughtAnnotation` does not leave any room in the `inline-block` layout for the `thoughtEnd` faux caret to occupy. Adding a `&ZeroWidthSpace` entity to the span makes it participate in the layout.

### DIFF
--- a/src/components/ThoughtAnnotation.tsx
+++ b/src/components/ThoughtAnnotation.tsx
@@ -232,7 +232,7 @@ const ThoughtAnnotation = React.memo(
               cssRaw,
             )}
             style={style}
-            dangerouslySetInnerHTML={{ __html: textMarkup || placeholder || '' }}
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Beautiful, thanks!

PR #2894, opened 2025-04-17
Note: check e.relatedTarget before setting editing: false

### LINKED ISSUE #2841: [iOS] Caret does not move on second tap after note

ethan-james:
Fixes #2841

The `editing({ value: false })` action was being dispatched when focus was moved from the note to the thought. I think it must be possible for `selection.isActive()` to be false `onBlur` before `onFocus` runs on the thought that is getting selected.

More testing is required to verify this assertion a little better, and also to ensure that `e.relatedTarget` is an acceptable substitute for `selection.isActive()`. There are some events that may fire with a value in `e.relatedTarget` such as opening the hamburger menu, but basic testing indicates that this works OK.


ethan-james:
@raineorshine Can you please test this branch to see if it resolves #2840 as well as #2841? It's not too terribly likely that the two are directly related, but I have difficulty reproducing #2840 and it would be good to know.

raineorshine:
Let me know if you still need me to test this against both issues, since we clarified the [steps to reproduce](https://github.com/cybersemics/em/issues/2840#issuecomment-2814047024).

ethan-james:
> Let me know if you still need me to test this against both issues, since we clarified the [steps to reproduce](https://github.com/cybersemics/em/issues/2840#issuecomment-2814047024).

That's OK, I verified that this does not fix #2840.

ethan-james:
I copied over the `relatedTarget` logic from `Editable`, which covers the same cases and should be pretty well battle-tested.

src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -272,4 +274,22 @@ describe('mobile only', () => {
     const focusNode = await getSelection().focusNode
     expect(focusNode).toBeUndefined()
   })
+
+  describe('when caret moves from inside a note', () => {
+    it('cursorForward should move the cursor to the next thought', async () => {
```
raineorshine, line 279: Test confirmed ✓

raineorshine (approved):
Beautiful, thanks!

raineorshine:
I forget to merge this after I approved it three weeks ago, but it appears to have been fixed by #2917.

Issue #2837, opened 2025-03-06
[iOS] Double tap thought opens keyboard without caret


raineorshine:
Tested on iOS Safari. 

## Steps to Reproduce

1. Create a thought `a`
2. Set the cursor to null. 
3. Double tap `a`. 

## Current Behavior

The keyboard opens but there is no caret. Typing does nothing. 

## Expected Behavior

The caret should be placed in the thought for editing. 

A double tap should do the same thing as two separate taps.

ethan-james:
I'm looking into this issue, informed by https://github.com/cybersemics/em/pull/2904#pullrequestreview-2790277755.

I like the idea of using `cursorOffset: null` to disable programmatic selection. I ran into this in `setCursor.ts`:

```
  // non-pure
  const updatedOffset = offset ?? (editingValueStore.getState() !== null ? 0 : null)
```

I tested, and this forces the offset to be 0 in cases where it ought to remain null. Do you know what types of actions I would need to test for regressions if I tried to get rid of this to allow `cursorOffset: null` to propagate to a component?

raineorshine:
I'm not sure off the top of my head, as the selection logic is quite complex and has evolved over time. In general, I would assume it is necessary if it's there.

Your best bet to determine the purpose is to do a `git blame` and see if more context can be inferred from the originating commit. 

Unfortunately there isn't complete test coverage of selection behavior, as it can be difficult to write tests for every case, and some are mobile only. So we have to be very careful to avoid regressions.

ethan-james:
I'm not getting much traction with this one, unfortunately. It seems like it should be related to #2753 because they both involve the caret not being placed properly when the keyboard opens, but in #2753 the issue is that `selection.set` does not trigger focus on the editable, while this one avoids `selection.set` and lets the native browser behavior fail to set the focus. I've compiled a list of things that don't seem to affect it.

1. `asyncFocus` isn't being called at all, which makes sense since that trick is for programmatic selections

2. `initEvents` is not triggering on iPad, which makes sense because there is no `blur` event firing, and no difference in behavior between iPad and iPhone

3. `role='button'` could be a candidate for introducing a dead zone during interactions, but it's already been applied to editables and I don't see any timing-related weirdness in the event handlers

4. ```
          // When the the cursor is first set on a thought, prevent the default browser behavior to avoid activating edit mode.
          // Do not reset until the long tap is definitely over.
          disableTapRef.current = true
          setTimeout(() => {
            disableTapRef.current = false
[...]

PR #2904, opened 2025-04-21
Note: only set selection if empty

### LINKED ISSUE #2839: [iOS] Caret placed at end of note instead of under tap
## Steps to Reproduce

```
- a
  - b
    - =note
      - Hello world
```

1. Set the cursor on `b`.
2. Tap `b` to enter edit mode.
3. Tap the note between "Hello" and "world".

## Current Behavior

The caret is placed at the end of the note.

## Expected Behavior

The caret should be placed under the tap, between "hello" and "world".

ethan-james:
Fixes #2839 

Third time's the charm. I added a `noteOffset` property to state. If it is `null`, `Note` will not attempt to programmatically set the selection. If it is a `number`, it will be used to set the selection. Currently, the only number that is used is `value.length` which sets the caret at the end, replacing the previous `selection.set(noteRef.current!, { end: true })`.


ethan-james:
I guess a self-contained solution would look something like the following:

```
    const exists = useSelector(state => !!findDescendant(state, thoughtId, '=note'))
    const hasExisted = usePrevious(exists)
    const isNew = exists && !hasExisted

    useEffect(() => {
      if (hasFocus && isNew) {
        selection.set(noteRef.current!, { end: true })
        // deleting a note, then closing the keyboard, then creating a new note could result in lack of focus,
        // perhaps related to iOS Safari's internal management of selection ranges and focus
        if (isTouch && isSafari()) noteRef.current?.focus()
      }
    }, [hasFocus, isNew])
```

raineorshine:
Hi, thanks for getting started on this. 

I believe we programmatically set the selection on a non-empty note during the `note` and `swapNote` commands.

ethan-james:
> I believe we programmatically set the selection on a non-empty note during the `note` and `swapNote` commands.

I see this in `swapNote.ts`:

```
setCursor({ path: parentNoteChildId ? appendToPath(parentOf(cursor), parentNoteChildId) : parentOf(cursor) })
```

which does not set `noteFocus`, and I can confirm that it sets the caret on the thought in main.

raineorshine:
Yes, I see that now. 

The `note` command sets the selection on the existing note though.

ethan-james:
> The `note` command sets the selection on the existing note though.

Ah, existing note, not only when a new note is created. Two possible solutions occur to me:

1. Re-create `allowDefaultSelection` from `useEditMode` to bypass the state-driven `hasFocus` effect handler on tap
2. add a flag to `setCursor` to indicate that the action is generated by a tap rather than programmatic

Option 2 means introducing something into the state that maybe doesn't belong there, but I've been looking into #2837 and I'm pretty sure that the mechanics of `allowDefaultSelection` are interfering with the double tap. So I'd be interested in hearing your thoughts on which approach would be preferable.

raineorshine:
Option 1 is pretty ugly, so I would say Option 2 is the one to pursue.

Thanks!

ethan-james:
> Option 1 is pretty ugly, so I would say Option 2 is the one to pursue.
> 
> Thanks!

I updated the PR description to outline my new approach.

raineorshine (changes requested):
Thanks for the update!

While this is an interesting approach, my overall sense is that the middleware + action extension + ministore is a bit bulky for what we need here. If I'm understanding correctly, we need a back door to block the side effect in the Note components. I would suggest either committing to fully using the Redux state or committing to using a back door/side channel, but not mixing the two.

- **Redux state** - We could add `state.noteOffset: number | null` (analogous to [`state.cursorOffset`](https://github.com/cybersemics/em/blob/e6cb97736709093a331546da0673a9ce80ccf415/src/%40types/State.ts#L44-L49)) and handle it entirely in Redux state.
  - Note the similar semantics: `A value of null means that the caret is not forcefully set on re-render, allowing the device to set it, e.g. on click.`
  - You say "there's no danger of ephemeral data lingering in the state", which sounds beneficial, but I'm not sure there is actual danger is in this case. If `noteOffset` is only used in the `useEffect`, then it is effectively inert after the initial render.
- **Back door** - Alternatively, if the Redux state is insufficient, we can commit to a completely independent back door: either a Ref if it's contained within a component or a ministore if it is across modules. Mixing this with middleware or extending the Action type is probably adding unnecessary complexity and coupling between different state management systems.

> It might be a bit over-engineered, but my hope is that I can immediately use this same ministore for #2837 to cut down on some of the edge cases in `Editable` selections. I made a cursory attempt to strongly-type the metadata, but more work is needed there. I didn't want to get ahead of myself.

I'm hesitant to build a general system in anticipation of future specifics unless those future specifics are very known/well-defined. I usually prefer to solve the specifics, then generalize afterwards in a refactor if we notice repetition/redundancy.

ethan-james:
While finishing this version, I discovered a new bug and reproduced it on main:

```
- A
  - =note
    - Note A
- B
  - =note
    - Note B
```

1. Tap Note A
2. Tap Note B
3. Tap A

Expected behavior: Cursor & caret selection both move to A
Observed behavior: Cursor moves to A, caret stays on note

raineorshine:
Good catch! Could you create a new issue for it?

src/@types/State.ts
```diff
@@ -123,6 +123,7 @@ interface State {
    * Passed to Yjs and cleared on every action.
    * See: /redux-enhancers/pushQueue.ts.
    */
+  noteOffset: number | null
```
raineorshine, line 126: Let's add a descriptive JSDOC comment here.

This accidentally picked up the comment for `pushQueue`.

src/components/Note.tsx
```diff
@@ -57,12 +59,14 @@ const Note = React.memo(
     useEffect(() => {
       // cursor must be true if note is focused
       if (hasFocus) {
-        selection.set(noteRef.current!, { end: true })
+        if (noteOffset === null) return
```
raineorshine, line 62: Let's combine this with the outer `if` statement.

raineorshine (changes requested):

raineorshine (approved):
Thank you!

PR #2907, opened 2025-04-23
Multiselect: Moving different groups of thoughts down and up breaks rank

### LINKED ISSUE #2853: Multiselect: Moving different groups of thoughts down and up breaks rank
## Steps to Reproduce

```
- x
  - a
  - b
  - c
  - d
  - e
  - f
  - g
```

1. Select `b`, `c`, `d`, `e`
2. Move Thoughts Down
3. Cursor Down
4. Select `d`, `e`, `g`
5. Move Thoughts Up 

## Current Behavior

`g` is not moved up along with `d` and `e`.

```
- x
  - a
  - f
  - b
  - d
  - e
  - c
  - g
```

https://github.com/user-attachments/assets/392ed34d-eb76-4124-9e5f-b76880368414

## Expected Behavior

`d`, `e`, `g` should be moved up as a group.

```
- x
  - a
  - f
  - b
  - d
  - e
  - g
  - c
```

## Testing

Add a unit test.

karunkop:
Close #2853 

## Root cause

The issue is noticeably seen with 7+ thoughts  that relates to how ranks accumulate and interact with the proportional nudge factor. Here's what's happening:

- Each thought has a rank value that determines its position.
- When we move thoughts, new ranks are calculated using `(prevRank + nextRank)/2 + nextRank*0.001
`.
- The nudge factor `(nextRank*0.001)` creates progressively larger gaps as ranks increase.

With 6 or fewer thoughts, the rank differences remain small enough that the proportional nudges don't cause noticeable issues. But once we have 7+ thoughts, especially after multiple move operations:

- The rank values spread out more.
- When we select non-adjacent thoughts (d, e, g), each gets a different proportional nudge.
- Thoughts with higher ranks (like g) get larger nudges.
- This causes g to drift more than d and e, breaking the expected grouping.

## Solution

The use of a fractional component of `prevSubthought.rank`originally added to maintain a nearby order of magnitude and prevent the halving function from jumping an order has been completely removed. This change is safe because the sorting algorithm has been updated from render-time to creation-time, meaning thought.rank is now directly modified during sorting.

 Thorough manual testing was performed, including scenarios with empty thoughts in sorted contexts. No regressions were found, confirming that the removal is safe.

### DIFF
--- a/src/selectors/__tests__/getDescendantThoughtIds.ts
+++ b/src/selectors/__tests__/getDescendantThoughtIds.ts
@@ -60,7 +60,7 @@ it('get descendants ordered by rank', () => {
     { value: 'a', rank: 0 },
     { value: 'b', rank: 1 },
     { value: 'c', rank: 2 },
-    { value: 'x', rank: 1.502 }, // rank is nudged towards the next sibling (see getRankBefore)
+    { value: 'x', rank: 1.5 },
   ])
 
   // ordered
@@ -69,7 +69,7 @@ it('get descendants ordered by rank', () => {
   expect(descendantsOrdered).toMatchObject([
     { value: 'a', rank: 0 },
     { value: 'b', rank: 1 },
-    { value: 'x', rank: 1.502 }, // rank is nudged towards the next sibling (see getRankBefore)
+    { value: 'x', rank: 1.5 },
     { value: 'c', rank: 2 },
   ])
 })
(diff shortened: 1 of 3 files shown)


raineorshine (approved):
Thanks! Good identification of the root cause.

I do see slightly different sort behavior in some edge cases. However, none are jumping out to me as terribly wrong, and all the major cases are covered by the existing unit tests. Better test coverage is necessary before anything if any regressions are found.

raineorshine:
e6cb9773670: Fully reverted the original implementation of [resortEmptyInPlace](https://github.com/cybersemics/em/commit/276690282a516d2e3af129b112bd31b3b2241983).

PR #2908, opened 2025-04-24
Invalid cursor after uncategorize + categorize + cursorForward

### LINKED ISSUE #2750: Invalid cursor after uncategorize + categorize + cursorForward
## Steps to Reproduce

1. Create thought `a`
2. Categorize
3. Cursor Forward
4. Cursor Back
5. Uncategorize
6. Categorize
7. Cursor Forward

## Current Behavior

Breadcrumbs show that the cursor in on `a`, but the cursor is not visible. 

https://github.com/user-attachments/assets/1452a931-68c0-4465-9288-9866f4126ef1

## Expected Behavior

Cursor should be on `a`.

karunkop:
Close #2750 

Root Cause:

The `cursorForward` action used an outdated path from cursorHistory. Its check was too basic as it only looked to see if the head thought in the path was still a child of the current thought, without checking the full structure of the path. After the thought was uncategorized and re-categorized, the structure changed and new thought IDs were created with different relationships. This led the cursor to follow a path that looked valid at the start but was no longer correct overall.

Solution:

If `validChild` exists and `cursorFromHistory` is not undefined, we now append the head thought from `cursorFromHistory` to the current cursor, rather than directly setting `cursorNew` to `cursorFromHistory`. This approach ensures that navigation after categorize/uncategorize actions uses only valid and up-to-date paths.

### DIFF
--- a/src/actions/__tests__/cursorForward.ts
+++ b/src/actions/__tests__/cursorForward.ts
@@ -39,19 +39,12 @@ describe('normal view', () => {
   it('should use a valid path after uncategorize and categorize operations', () => {
     const steps = [
       newThought('a'),
-
       setCursor(['a']),
-
       categorize,
-
       cursorForward,
-
       cursorBack,
-
       uncategorize({}),
-
       categorize,
-
       cursorForward,
     ]
 
--- a/src/actions/cursorForward.ts
+++ b/src/actions/cursorForward.ts
@@ -13,8 +13,6 @@ import head from '../util/head'
 import headValue from '../util/headValue'
 import unroot from '../util/unroot'
 
-// import validatePath from '../util/validatePath'
-
 /** Moves the cursor forward in the cursorHistory. */
 const cursorForward = (state: State) => {
   const cursorFromHistory = last(state.cursorHistory)


raineorshine (changes requested):
Thanks for a good explanation of the root cause.

Note that a true fix would be to reconstruct the Paths within `state.cursorHistory` whenever a thought is moved. However, that has its own challenges so I think you're right to look for a mitigation first.

The overall approach is reasonable, however it won't work in nested context views. i.e. The integrity of parent-child relationships that you assume is only true for `SimplePath`, not `Path`, which can cross context view boundaries. The "context view" case in `cursorForward` only handles contexts immediately within a context view. `validatePath` will incorrectly return false for deeply nested Paths within a context view, e.g. `a/m~/x/x1/x2`.

Since `cursorForward` moves the `cursor` by a single level only, I suspect that there is a solution that can append a single ThoughtId to `cursor`, and thus avoid having to reconstruct the context chain of nested context views.

**UPDATE**: Could you see if this works?

```diff
diff --git a/src/actions/cursorForward.ts b/src/actions/cursorForward.ts
index 26a74e0be12..b5cfd9608fc 100644
--- a/src/actions/cursorForward.ts
+++ b/src/actions/cursorForward.ts
@@ -33,7 +33,12 @@ const cursorForward = (state: State) => {
     const simplePath = simplifyPath(state, cursor)
     const firstChild = firstVisibleChild(state, head(simplePath))
     isValidChild = cursorFromHistory && !!getThoughtById(state, head(cursor))?.childrenMap[head(cursorFromHistory)]
-    cursorNew = isValidChild ? cursorFromHistory : firstChild ? unroot([...cursor, firstChild.id]) : cursor
+    cursorNew =
+      isValidChild && cursorFromHistory
+        ? appendToPath(cursor, head(cursorFromHistory))
+        : firstChild
+          ? unroot([...cursor, firstChild.id])
+          : cursor
   }
 
   return cursorNew
```

Also, please add a failing unit test for the original test case? Thanks!

karunkop:
> **UPDATE**: Could you see if this works?
> 
> ```diff
> diff --git a/src/actions/cursorForward.ts b/src/actions/cursorForward.ts
> index 26a74e0be12..b5cfd9608fc 100644
> --- a/src/actions/cursorForward.ts
> +++ b/src/actions/cursorForward.ts
> @@ -33,7 +33,12 @@ const cursorForward = (state: State) => {
>      const simplePath = simplifyPath(state, cursor)
>      const firstChild = firstVisibleChild(state, head(simplePath))
>      isValidChild = cursorFromHistory && !!getThoughtById(state, head(cursor))?.childrenMap[head(cursorFromHistory)]
> -    cursorNew = isValidChild ? cursorFromHistory : firstChild ? unroot([...cursor, firstChild.id]) : cursor
> +    cursorNew =
> +      isValidChild && cursorFromHistory
> +        ? appendToPath(cursor, head(cursorFromHistory))
> +        : firstChild
> +          ? unroot([...cursor, firstChild.id])
> +          : cursor
>    }
>  
>    return cursorNew
> ```

Yep this works! This also eliminates the need of `validatePath` function as it simply appends the head thought from cursorHistory to current cursor

raineorshine (approved):

PR #2909, opened 2025-04-25
Move focusNode.focus() into selection.set

### LINKED ISSUE #2753: selection.set does not always open keyboard

ethan-james:
Fixes #2753

The question of why `selection.set` doesn't work for the same node on the 2nd `clearThought` remains unanswered. The `focusNode.focus()` fix does work inside of `selection.ts`, which restores the invariant that selection behavior occurs in there.

Most search results on this topic deal with the fact that you can't select certain types of nodes like links, which is neither here nor there. Generally, the advice that I've found is of the "yeah, you need to call focus() sometimes to get the selection to show" variety, which is essentially what we're doing here.

I did learn more about `document.activeElement`, a global pointer to the selected node. This is already used in `selection.clear`:

```
  // On mobile safari it is possible that the keyboard stays up even when there is no selection.
  // Blur the active document element to close the keyboard.
  if (document.activeElement instanceof HTMLElement) {
    document.activeElement.blur()
  }
```

so I updated my solution to use that instead of always calling `focus()` on mobile Safari:

```
if (focusNode instanceof HTMLElement && document.activeElement !== focusNode) focusNode.focus()
```

https://github.com/microsoft/playwright/issues/23019 contains steps to reproduce an issue that seems the most similar, but has some key differences:

1. fails on the 1st selection rather than the 2nd
2. I think it's running on Mac desktop

but that's the closest result that I was able to find to our current issue.

Finally, I tracked down an extra call to `selection.clear` that takes place in `initEvents.ts`:

```
    // If the app is backgounded while in edit mode, then the keyboard will not open when switching back to the app, even when focusNode, document.activeElement, and visualViewport.height all indicate that the keyboard is open. Strangely, the app enters the 'passive' state after hidden -> passive -> active completes. The invalid state can be detected with document.hasFocus(). Since there is no way to force the app to be active when it is passive, then we do not bother trying to re-open the keyboard, and instead disable edit mode so that at least it matches what the user sees. Use a timeout to ensure that this is called only when the device stays in the passive state, not when it is moving from hidden -> passive -> active.
    // https://github.com/cybersemics/em/issues/1468
    else if (
      isTouch &&
      isSafari() &&
      oldState === 'active' &&
      newState === 'passive' &&
      document.activeElement &&
      !document.hasFocus()
    ) {
      passiveTimeout = setTimeout(selection.clear, 10) as unknown as number
    }
```

It is definitely running, both when closing the keyboard after `clearThought` number 1, and again when performing `clearThought` number 2. Unfortunately, it doesn't appear to have any effect on the bug. Removing it breaks the keyboard-closing behavior after `clearThought` number 1, and it doesn't seem to fix keyboard-opening during `clearThought` number 2.

I'd be happy to keep investigating, especially since I'm still holding out hope that this is related to #2837 in some way, but I wanted to make sure to fill you in before I keep going down the rabbit hole.


raineorshine:
Hi, thanks for the report!

Can you remind me of the issue that involved a "1st" and "2nd" clearThought? I can't seem to find it in the issues linked from #2753.

> so I updated my solution to use that instead of always calling `focus()` on mobile Safari:
> 
> ```
> if (focusNode instanceof HTMLElement && document.activeElement !== focusNode) focusNode.focus()
> ```

Great, it's always good to narrow down the conditions of special cases when possible.

> It is definitely running, both when closing the keyboard after `clearThought` number 1, and again when performing `clearThought` number 2. Unfortunately, it doesn't appear to have any effect on the bug. Removing it breaks the keyboard-closing behavior after `clearThought` number 1, and it doesn't seem to fix keyboard-opening during `clearThought` number 2.

Why would the lifecycle state change from `active` to `passive` in the `clearThought` case? `statechange` is only triggered when switching apps/windows.

ethan-james:
> Can you remind me of the issue that involved a 1st and 2nd `clearThought`? I can't seem to find it in the issues linked from #2753.

The `clearThought` is from https://github.com/cybersemics/em/pull/2772#discussion_r1917308772. That's how I've been reproducing this bug.

ethan-james:
> Why would the lifecycle state change from `active` to `passive` in the `clearThought` case? `statechange` is only triggered when switching apps/windows.

I added this line to the top of `onStateChange`:

```
console.info(`${oldState} -> ${newState}`)
```

and I am always seeing it toggle to `active` when the keyboard opens, and back to `passive` when it closes.

ethan-james:
> Which platform?
> 
> I tested on mobile Safari and `onStateChange` does not trigger when the keyboard changes.

Huh, yeah, the iPad does it but the iPhone does not. The iPhone also does not appear to require the `focus()` workaround for 2x `clearThought`. I already verified that `onStateChange` wasn't causing the bug on the iPad, but maybe I should go back and look harder.

raineorshine:
Wow! That's the first iPad/iPhone behavioral difference I've ever encountered.

> I already verified that `onStateChange` wasn't causing the bug on the iPad, but maybe I should go back and look harder.

I will leave it up to you how to best proceed. Ideally we can bracket the iPad/iPhone distinction and deal with it separately, but if you discover that it is critical to this functionality then we'll have to accommodate it.

ethan-james:
> Wow! That's the first iPad/iPhone behavioral difference I've ever encountered.
> 
> > I already verified that `onStateChange` wasn't causing the bug on the iPad, but maybe I should go back and look harder.
> 
> I will leave it up to you how to best proceed. Ideally we can bracket the iPad/iPhone distinction and deal with it separately, but if you discover that it is critical to this functionality then we'll have to accommodate it.

I updated the iPhone from 17 to 18 to match the iPad, and that did not appear to make a difference. I did notice a small new bug that only seems to affect iPhone, not iPad:

1. Open a fresh tab for the app (must have cursor on a thought)
2. Execute `clearThought` quickly enough to avoid command palette
3. Keyboard does not open

Do you think it's worth creating a new issue?

raineorshine:
Yes, thanks!

ethan-james:
Today I learned that `asyncFocus` is pushing the iPad from `passive -> active` and then back from `active -> passive`. In #2693 I disable `hiddenInput` after focusing on it, which would most likely trigger the transition back to `passive`. If I comment out

```
hiddenInput.disabled = true
```

then 2x `clearThought` will show the selection appropriately on the second invocation of `clearThought`. However, subsequently closing the keyboard for a second time doesn't transition the thought back to "normal" mode. It stays grayed out, so a new bug or perhaps a regression to behavior that I've seen in the past. Also, the other method for reproducing this (delete a note, close keyboard, toggle note gesture) is not affected by commenting out that line. So this doesn't seem like a solution, on top of the regressions that it might cause.

ethan-james:
The way to fix the failing tests was to put `isTouch && isSafari()` back into the conditional. This obviously doesn't fix any issues, it just stops `focus()` from being called in the test runner.

I focused on one test in `src/commands/__tests__/deleteThought.ts:

```
  it('after deleteEmptyThought, caret should move to end of previous thought', async () => {
    await act(async () => {
      store.dispatch([{ type: 'newThought', value: 'apple' }, { type: 'newThought' }, { type: 'deleteEmptyThought' }])
    })

    // Selection.focusOffset a number representing the offset of the selection's anchor within the focusNode. If focusNode is a text node, this is the number of characters within focusNode preceding the focus. If focusNode is an element, this is the number of chi,ld nodes of the focusNode preceding the focus.
    // In this case, the selection is at the end of the apple element.
    expect(window.getSelection()?.focusNode?.nodeType).toBe(Node.ELEMENT_NODE)
    expect(window.getSelection()?.focusNode?.textContent).toBe('apple')
    expect(window.getSelection()?.focusOffset).toBe(1)
  })
```

It fails because `focusOffset` ends up being 0, not 1. I tried it out in Chrome, and didn't see a problem. I wasn't able to emulate the first action, `{ type: 'newThought', value: 'apple' }`, but overall I saw the correct behavior:

Screenshot 2025-05-01 101752

Do you think it's worth trying to dig deeper, or is this type of band-aid OK?

ethan-james:
> Thanks for the update. It's strange that `asyncFocus` affects `onStateChange` at all, but I guess that's what we have to work with.
> 
> If it's possible to distinguish the `onStateChange` triggered by `asyncFocus` from the expected trigger (when switching apps), then we could add a condition to exclude the false positive.

I can inspect the `originalEvent` property on the event passed into `onStateChange`.

I tried adding `originalEvent.type !== 'focus' && originalEvent.type !== 'blur'` to the condition in `onStateChange` in an attempt to filter out all events that wouldn't trigger a state change on an iPhone. That caused worse problems (cleared thought didn't return to normal after closing keyboard).

I also tried `!(originalEvent.target instanceof HTMLInputElement)` to try to filter out events triggered by the `asyncFocus` hidden input, but that didn't appear to have any effect on the bug.

Any critiques of my approach, or any other things you think I should try?

raineorshine:
> Do you think it's worth trying to dig deeper, or is this type of band-aid OK?

It's okay. Browser selection in JSDOM is not very accurate, so the usefulness of this test is debatable anyway. If we wanted to test actual selection behavior, we would use a puppeteer test.

> I tried adding `originalEvent.type !== 'focus' && originalEvent.type !== 'blur'` to the condition in `onStateChange` in an attempt to filter out all events that wouldn't trigger a state change on an iPhone.

Do you mean iPad? I believe we established that `onStateChange` only fires on keyboard change on iPad, not iPhone, but correct me if I'm wrong.

> Any critiques of my approach, or any other things you think I should try?

The first step, which it sounds like you have already done, is to log the `statechange` event and simply compare the output between one triggered by a keyboard change and one triggered but switching apps. Any difference between the two could be used to distinguish them.

If they are identical and there is no way to distinguish them based on the event object, I would try to exploit the timing between `onStateChange` and `asyncFocus`. `asyncFocus` is probably not initially active when switching apps. We also know that `hiddenInput` does not stay focused for more than a few ms, as the point of `asyncFocus` is to allow programmatic selection of a contenteditable from a cold start when the keyboard is down. Either of this could be used as the basis to distinguish between the two sources of events.

raineorshine (approved):
Let me know if I should merge now, or if you plan on addressing the iPad issue within this PR.

ethan-james:
> Do you mean iPad? I believe we established that `onStateChange` only fires on keyboard change on iPad, not iPhone, but correct me if I'm wrong.

Right, that's what I was trying to say. I was trying to filter out all events that would trigger a state change on an iPad but not on an iPhone, which appear to be `focus` and `blur` events. Maybe they bubble on iPad in a way that they don't on iPhone? Hard to tell exactly.

As stated above, it didn't really work.

> The first step, which it sounds like you have already done, is to log the `statechange` event and simply compare the output between one triggered by a keyboard change and one triggered but switching apps. Any difference between the two could be used to distinguish them.

Directly comparing the events triggered by a keyboard change vs switching apps sounds like something that I should quickly try.

> Let me know if I should merge now, or if you plan on addressing the iPad issue within this PR.

I think I have a final report almost ready. I'll test out a couple of things and then put them in a new comment instead of this one as I originally intended.

ethan-james:
At the end of this process, I don't have much except for a way more detailed description of the problem. Maybe that will assist with future efforts.

When loading the app on iPad, it presumably starts in `active` mode. Executing `clearThought` calls `onFocus` in the editable and everything looks OK with the caret.

Closing the keyboard on iPad never triggers `onBlur` on the editable. This is very weird because `onStateChange` receives a `blur` event with the editable in its `target` property, and that's what triggers the state change from `active` to `passive`.

Screenshot 2025-05-02 at 11 37 58 AM

But the event never reaches the editable itself on iPad, and it is only `passiveTimeout = setTimeout(selection.clear, 10) as unknown as number` in `onStateChange` that eventually causes the editable to blur in the last line of the screenshot above. Without that line, it stays in gray placeholder mode.

On the iPhone, `onBlur` is triggered on the editable correctly and the event never reaches the top level where it can be captured by `page-lifecycle`.

Then, when `clearThought` is executed for the second time, the iPad fails to call `onFocus` on the editable. I see it transitioning from `passive` to `active`, calling `focus` on the hidden input, then transitioning back to `passive` and calling `blur` on the hidden input. Then it transitions one last time back to `active`, but the event `target` is `window`, not the editable.

Screenshot 2025-05-02 at 11 48 09 AM

So it's not calling `focus` on the editable, which is why we need to do it manually, even though I see it adding the range in `selection.set`.

Screenshot 2025-05-02 at 12 29 58 PM

So the unanswered questions for me are:

1. Why doesn't `onBlur` fire when the keyboard closes?
2. Why does `selection.set` trigger a `focus` event targeting the `window` instead of the editable?

raineorshine:
Hi! Thanks for deeper the investigation.

From what I understand, there are several invariants that are broken on iPad:

- `onStateChange` should fire when switching apps (and only when switching apps)
- `onFocus` should fire on the Editable on focus.
- `onBlur` should fire on the Editable on blur.

Given the amount of functionality this touches, I would say it is outside the scope of this issue, and probably needs to be addressed in multiple issues with concrete bugs and steps to reproduce.

Regarding the narrower refactoring goal of moving `focusNode.focus()` into `selection.set`, could you clarify what the current status is?

Are there any non-iPad regressions or issues that are blocking this PR from being merged?

Apologies if I'm losing the thread here, it's just a lot to keep track of!

ethan-james:
> From what I understand, there are several invariants that are broken on iPad:
> 
> * `onStateChange` should fire when switching apps (and only when switching apps)
> * `onFocus` should fire on the Editable on focus.
> * `onBlur` should fire on the Editable on blur.
> 
> Given the amount of functionality this touches, I would say it is outside the scope of this issue, and probably needs to be addressed in multiple issues with concrete bugs and steps to reproduce.

Right, I think that each invariant could be addressed separately.

I can easily short-circuit `onStateChange` and return early if the event is focus/blur, which would then create a concrete bug:

1. `clearThought` on iPad
2. close keyboard
3. thought remains in "cleared" mode with placeholder text

That might be a good place to start. By the end of the process, hopefully we would know why the events are erroneously firing on the window rather than the element on iPad, in which case the short-circuit could be removed.

> Regarding the narrower refactoring goal of moving `focusNode.focus()` into `selection.set`, could you clarify what the current status is?

`focus()` is moved into `selection.set` and seems to be working fine. I'm irked that I had to put `isTouch && isSafari()` back on it to fix the tests, because I'd prefer to target the behavior rather than the platform, but all of these issues only seem to affect iOS so it's probably fine.

> Are there any non-iPad regressions or issues that are blocking this PR from being merged?

Not that I know of. This seems like a good time to merge.

raineorshine (approved):

raineorshine:
Noticed this puppeteer caret test failed in `main` as soon as this was merged:

https://github.com/cybersemics/em/actions/runs/14825858092/attempts/1

Let's keep an eye on this, as I haven't seen this test fail before. We may have introduced some flakiness.

PR #2913, opened 2025-04-30
Distribute NAVIGATION_ACTIONS and UNDOABLE_ACTIONS

### LINKED ISSUE #2866: Distribute NAVIGATION_ACTIONS and UNDOABLE_ACTIONS
Instead of defining [NAVIGATION_ACTIONS](https://github.com/cybersemics/em/blob/177c2a73dd318f92f6462f6c9159e3fc40d93b01/src/redux-enhancers/undoRedoEnhancer.ts#L24C7-L24C25) and [UNDOABLE_ACTIONS](https://github.com/cybersemics/em/blob/177c2a73dd318f92f6462f6c9159e3fc40d93b01/src/redux-enhancers/undoRedoEnhancer.ts#L42) in one place, define them in a distributed fashion co-located with the reducers/action-creators themselves. This will make it easier to add new actions in the future.

karunkop:
Close #2866 

### Refactor implemented

- [X] Created an `ActionMetadata` registry that allows each action to register its own undo/redo-related metadata.
- [X] Updated  `undoRedoEnhancer` to use the distributed metadata registry.
- [X] Prevented double registration.
- [X] Ensured all actions are registered.


src/util/validateActionRegistrations.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  if (unregisteredActions.length > 0) {
+    console.error(
+      `The following actions are not registered in the metadata registry: ${unregisteredActions.join(', ')}`,
+    )
+    if (process.env.NODE_ENV === 'development') {
+      throw new Error(`Some actions are not registered in the metadata registry. See console for details.`)
+    }
+  } else {
+    console.info('All actions successfully registered in the metadata registry')
```
raineorshine, line 20: Let's omit extraneous chatter in the console unless there is an error. 

The rest looks good though!

src/@types/ActionMetadata.ts
```diff
@@ -0,0 +1,41 @@
+import ActionType from './ActionType'
+
+/**
+ * Metadata for actions to determine their behavior in the undo/redo system.
+ */
+export interface ActionMetadata {
```
raineorshine, line 6: Internal only

src/@types/ActionMetadata.ts
raineorshine, line 1: Let's move this to `/util`

src/util/validateActionRegistrations.ts
```diff
@@ -0,0 +1,22 @@
+import { actionMetadataRegistry } from '../@types/ActionMetadata'
+import ActionType from '../@types/ActionType'
+import * as actions from '../actions'
+
+/**
+ * Validates that all actions have been registered in the metadata registry.
+ */
+export default function validateActionRegistrations(): void {
```
raineorshine, line 8: Let's move this into the `util` module so that `actionMetadataRegistry` can be encapsulated.

raineorshine (changes requested):
Looks great!

Probably the main thing I want you to think about conceptually with this is **encapsulation**. What complexity are you trying to hide? By exporting `actionMetadataRegistry`, you've exposed the implementation details! What's the point of hiding it with `isUndoable` and `isNavigation` if you're just going to expose it somewhere else?

Every export is an API, so you should be thinking about what you're trying to hide and how you want to design each module's API, not just exporting things in an ad hoc fashion. That just gives you separation across files, without any management of complexity. Complexity is managed by bundling it away within a module, and limiting what external modules can access.

PR #2920, opened 2025-05-04
Sort: add sort by note option

RED-ROSE515:
This PR resolves the issue : Sort: Note https://github.com/cybersemics/em/issues/2831

In this Pr, I added a new Sort Options note.

### DIFF
--- a/src/actions/editThought.ts
+++ b/src/actions/editThought.ts
@@ -142,7 +142,8 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
     ...editedThought,
     generating: false,
     rank:
-      newValue !== '' && (sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated')
+      newValue !== '' &&
+      (sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated' || sortType === 'Note')
         ? getSortedRank(state, editedThought.parentId, newValue)
         : editedThought.rank,
     value: newValue,
--- a/src/components/__tests__/SortPicker.ts
+++ b/src/components/__tests__/SortPicker.ts
@@ -380,3 +380,127 @@ describe('Created to Updated', () => {
     expect(thoughtValues).toMatchObject(['a', 'c', 'd'])
   })
 })
+
+it('home: Note Asc', async () => {
+  act(() => {
+    store.dispatch([
+      importText({
+        text: `
+          - a
+            - =note
+              - 2
+          - b
+            - =note
+              - 3
+          - c
+            - =note
+              - 1
+        `,
+      }),
+      setCursor(['a']),
+    ])
+  })
+
+  await act(vi.runOnlyPendingTimersAsync)
+
+  await click('[data-testid="toolbar-icon"][aria-label="SortPicker"]')
+  await click('[aria-label="sort options"] [aria-label="Note"]')
+  await act(() => vi.runAllTimersAsync())
+
+  const thoughts = screen.getAllByTestId(/thought/)
+  expect(thoughts.map((child: HTMLElement) => child.textContent)).toMatchObject(['c1', 'a2', 'b3'])
+})
+
+it('home: Note Desc', async () => {
+  act(() => {
+    store.dispatch([
+      importText({
+        text: `
+          - =sort
+            - Note
+          - a
+            - =note
+              - 2
+          - b
+            - =note
+              - 3
+          - c
+            - =note
+              - 1
+        `,
+      }),
+      setCursor(['a']),
+    ])
+  })
+
+  await act(vi.runOnlyPendingTimersAsync)
+
+  await click('[data-testid="toolbar-icon"][aria-label="SortPicker"]')
+  await click('[aria-label="sort options"] [aria-label="Note"]')
+  await act(() => vi.runAllTimersAsync())
+
+  const thoughts = screen.getAllByTestId(/thought/)
+  expect(thoughts.map((child: HTMLElement) => child.textContent)).toMatchObject(['b3', 'a2', 'c1'])
+})
+
+it('home: Note Asc', async () => {
+  act(() => {
+    store.dispatch([
+      importText({
+        text: `
+          - a
+            - =note
+              - 2
+          - b
+            - =note
+              - 3
+          - c
+            - =note
+              - 1
+        `,
+      }),
+      setCursor(['a']),
+    ])
+  })
+
+  await act(vi.runOnlyPendingTimersAsync)
+
+  await click('[data-testid="toolbar-icon"][aria-label="SortPicker"]')
+  await click('[aria-label="sort options"] [aria-label="Note"]')
+  await act(() => vi.runAllTimersAsync())
+
+  const thoughts = screen.getAllByTestId(/thought/)
+  expect(thoughts.map((child: HTMLElement) => child.textContent)).toMatchObject(['c1', 'a2', 'b3'])
+})
+
+it('home: Note Desc', async () => {
+  act(() => {
+    store.dispatch([
+      importText({
+        text: `
+          - =sort
+            - Note
+          - a
+            - =note
+              - 2
+          - b
+            - =note
+              - 3
+          - c
+            - =note
+              - 1
+        `,
+      }),
+      setCursor(['a']),
+    ])
+  })
+
+  await act(vi.runOnlyPendingTimersAsync)
+
+  await click('[data-testid="toolbar-icon"][aria-label="SortPicker"]')
+  await click('[aria-label="sort options"] [aria-label="Note"]')
+  await act(() => vi.runAllTimersAsync())
+
+  const thoughts = screen.getAllByTestId(/thought/)
+  expect(thoughts.map((child: HTMLElement) => child.textContent)).toMatchObject(['b3', 'a2', 'c1'])
+})
--- a/src/selectors/getChildren.ts
+++ b/src/selectors/getChildren.ts
@@ -12,6 +12,8 @@ import compareByRank from '../util/compareByRank'
 import {
   compareThought,
   compareThoughtByCreated,
+  compareThoughtByNote,
+  compareThoughtByNoteDescending,
   compareThoughtByUpdated,
   compareThoughtDescending,
 } from '../util/compareThought'
@@ -185,6 +187,10 @@ export const getAllChildrenSorted = (state: State, id: ThoughtId): Thought[] => 
     return getChildrenSortedCreated(state, id)
   } else if (sortPreference.type === 'Updated') {
     return getChildrenSortedUpdated(state, id)
+  } else if (sortPreference.type === 'Note') {
+    return getChildrenRanked(state, id).sort(
+      sortPreference.direction === 'Desc' ? compareThoughtByNoteDescending(state) : compareThoughtByNote(state),
+    )
   } else {
     return getChildrenRanked(state, id)
   }
--- a/src/util/compareThought.ts
+++ b/src/util/compareThought.ts
@@ -2,10 +2,12 @@
 import _ from 'lodash'
 import ComparatorFunction from '../@types/ComparatorFunction'
 import ComparatorValue from '../@types/ComparatorValue'
+import State from '../@types/State'
 import Thought from '../@types/Thought'
 import { ALLOWED_FORMATTING_TAGS, EMOJI_REGEX, REGEX_EMOJI_GLOBAL } from '../constants'
 import isAttribute from './isAttribute'
 import lower from './lower'
+import noteValue from './noteValue'
 
 const STARTS_WITH_EMOJI_REGEX = new RegExp(`^${EMOJI_REGEX.source}`)
 const IGNORED_PREFIXES = ['the ']
@@ -221,3 +223,23 @@ export const compareThoughtByUpdated: ComparatorFunction<Thought> = (a: Thought,
 /** Compare two thoughts by their lastUpdated timestamp in descending order (newest first). Fall back to compareReasonable if created at the same time. */
 export const compareThoughtByUpdatedDescending: ComparatorFunction<Thought> = (a: Thought, b: Thought) =>
   compare(b.lastUpdated, a.lastUpdated) || compareReasonable(a.value, b.value)
+
+/** Compare two thoughts by their note value in ascending order, falling back to their value if notes are absent or equal. */
+export const compareThoughtByNote =
+  (state: State): ComparatorFunction<Thought> =>
+  (a: Thought, b: Thought) => {
+    const noteA = noteValue(state, a.id)
+    const noteB = noteValue(state, b.id)
+
+    return compareReasonable(noteA || a.value, noteB || b.value)
+  }
+
+/** Compare two thoughts by their note value in descending order, falling back to their value if notes are absent or equal. */
+export const compareThoughtByNoteDescending =
+  (state: State): ComparatorFunction<Thought> =>
+  (a: Thought, b: Thought) => {
+    const noteA = noteValue(state, a.id)
+    const noteB = noteValue(state, b.id)
+
+    return compareReasonableDescending(noteA || a.value, noteB || b.value)
+  }
--- a/src/commands/toggleSortPicker.ts
+++ b/src/commands/toggleSortPicker.ts
@@ -29,7 +29,7 @@ const toggleSortCommand: Command = {
 
     const path = simplifyPath(state, rootedParentOf(state, state.cursor))
     const sortPreference = getSortPreference(state, head(path))
-    return ['Alphabetical', 'Created', 'Updated'].includes(sortPreference.type)
+    return ['Alphabetical', 'Created', 'Updated', 'Note'].includes(sortPreference.type)
   },
   isDropdownOpen: state => !!state.showSortPicker,
   // Show an error if the ranks do not match the sort condition.
--- a/src/components/SortPicker.tsx
+++ b/src/components/SortPicker.tsx
@@ -15,7 +15,7 @@ import isRoot from '../util/isRoot'
 import Popover from './Popover'
 
 // Define the type for sort options
-type SortType = 'None' | 'Alphabetical' | 'Created' | 'Updated'
+type SortType = 'None' | 'Alphabetical' | 'Created' | 'Updated' | 'Note'
 
 /** Sort Option component for individual sort options. */
 interface SortOptionProps {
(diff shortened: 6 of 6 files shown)


src/components/__tests__/SortPicker.ts
```diff
@@ -380,3 +380,127 @@ describe('Created to Updated', () => {
...
+
+  await click('[data-testid="toolbar-icon"][aria-label="SortPicker"]')
+  await click('[aria-label="sort options"] [aria-label="Note"]')
+  await act(() => vi.runAllTimersAsync())
+
+  const thoughts = screen.getAllByTestId(/thought/)
+  expect(thoughts.map((child: HTMLElement) => child.textContent)).toMatchObject(['b3', 'a2', 'c1'])
+})
+
+it('home: Note Asc', async () => {
```
raineorshine, line 446: This test looks identical to the one above. Did you have something else in mind?

src/selectors/getChildren.ts
```diff
@@ -185,6 +187,10 @@ export const getAllChildrenSorted = (state: State, id: ThoughtId): Thought[] =>
     return getChildrenSortedCreated(state, id)
   } else if (sortPreference.type === 'Updated') {
     return getChildrenSortedUpdated(state, id)
+  } else if (sortPreference.type === 'Note') {
+    return getChildrenRanked(state, id).sort(
+      sortPreference.direction === 'Desc' ? compareThoughtByNoteDescending(state) : compareThoughtByNote(state),
+    )
```
raineorshine, line 193: I'm not sure why this breaks from the pattern and does something different with the other sort types.

src/util/compareThought.ts
```diff
@@ -221,3 +223,23 @@ export const compareThoughtByUpdated: ComparatorFunction<Thought> = (a: Thought,
...
   compare(b.lastUpdated, a.lastUpdated) || compareReasonable(a.value, b.value)
+
+/** Compare two thoughts by their note value in ascending order, falling back to their value if notes are absent or equal. */
+export const compareThoughtByNote =
+  (state: State): ComparatorFunction<Thought> =>
+  (a: Thought, b: Thought) => {
+    const noteA = noteValue(state, a.id)
+    const noteB = noteValue(state, b.id)
+
+    return compareReasonable(noteA || a.value, noteB || b.value)
```
raineorshine, line 234: I didn't think of thoughts that are missing notes when I create the original issue.

For mixed lists, could you sort all notes to the top, then sort by rank?

Please add a test for this case as well.

Thanks!

raineorshine (changes requested):
Thanks, looks good!

One thing I didn't mention explicitly, but the list must maintain its sort order even while notes are edited. This parallels all other sort types. There should be existing logic you can use.

Please add a test for this case.

## Steps to Reproduce

```
- =sort
  - Note
- c
  - =note
    - 1
- a
  - =note
    - 2
- b
  - =note
    - 3
```

Edit the value of note `2` to `4`.

## Current Behavior

```
- =sort
  - Note
- c
  - =note
    - 1
- a
  - =note
    - 4
- b
  - =note
    - 3
```

## Expected Behavior

```
- =sort
  - Note
- c
  - =note
    - 1
- b
  - =note
    - 3
- a
  - =note
    - 4
```

src/components/__tests__/SortPicker.ts
```diff
@@ -380,3 +380,100 @@ describe('Created to Updated', () => {
...
+      setCursor(['a']),
+    ])
+  })
+
+  await act(vi.runOnlyPendingTimersAsync)
+
+  // Edit the note value of 'a' from 2 to 4
+  act(() => {
+    store.dispatch([editThoughtByContext(['a', '=note', '2'], '4')])
+  })
```
raineorshine, line 473: It would be better to use a more user-centric way of interacting with the note rather than dispatching to the global store directly.

See: https://github.com/karunkop/em/blob/d82e40db37283cae9a9dad871c9a2991e6026fc0/src/components/__tests__/Note.ts#L378-L385

We currently dispatch `importText` directly to the store at the beginning of tests for simplicity, but [soon this will be refactored](https://github.com/cybersemics/em/issues/2980) to use `paste`, which is more user-centric and hides the global store access.

src/components/__tests__/SortPicker.ts
raineorshine, line 1: The tests look good!

Let's add a test for mixed lists that contain some thoughts with notes and some thoughts without.

src/util/compareThought.ts
```diff
@@ -221,3 +223,21 @@ export const compareThoughtByUpdated: ComparatorFunction<Thought> = (a: Thought,
...
+
+/** Get note value or fallback for a Thought or string input. */
+const getNoteOrFallback = (state: State, input: Thought | string): string => {
+  if (typeof input === 'string') return input
+  return noteValue(state, input.id) || input.rank.toString()
+}
+
+/** Compare two thoughts by their note value in ascending order, falling back to their rank if notes are absent or equal. */
+export const compareThoughtByNote =
+  (state: State): ComparatorFunction<Thought | string> =>
```
raineorshine, line 235: `Thought | string` doesn't really make sense as a union type. Either it's a thought or a string, but never both. Perhaps this allows you to reuse certain functions by making them polymorphic, but I don't think it's the right kind of abstraction.

It would be better to extract the note value and then compare it using a string-only comparator.

src/util/compareThought.ts
```diff
@@ -221,3 +223,21 @@ export const compareThoughtByUpdated: ComparatorFunction<Thought> = (a: Thought,
 /** Compare two thoughts by their lastUpdated timestamp in descending order (newest first). Fall back to compareReasonable if created at the same time. */
 export const compareThoughtByUpdatedDescending: ComparatorFunction<Thought> = (a: Thought, b: Thought) =>
   compare(b.lastUpdated, a.lastUpdated) || compareReasonable(a.value, b.value)
+
+/** Get note value or fallback for a Thought or string input. */
+const getNoteOrFallback = (state: State, input: Thought | string): string => {
+  if (typeof input === 'string') return input
+  return noteValue(state, input.id) || input.rank.toString()
```
raineorshine, line 230: You can't compare numbers lexicographically, because `'2' > '10'`.

Take advantage of [makeOrderedComparator](https://github.com/cybersemics/em/blob/92d6d2d02373b1a20ec7ae3003b8e0dc1deb2b97/src/util/compareThought.ts#L137) or follow the [compareDateAndOther](https://github.com/cybersemics/em/blob/92d6d2d02373b1a20ec7ae3003b8e0dc1deb2b97/src/util/compareThought.ts#L130) pattern instead.

src/selectors/getSortedRank.ts
```diff
@@ -1,25 +1,38 @@
...
+  // Handle Note sorting
+  if (sortPreference.type === 'Note') {
+    const compareFn = isDescending ? compareThoughtByNoteDescending(state) : compareThoughtByNote(state)
+    const index = thoughts.findIndex(thought => compareFn(value, thought) === -1)
+    return index === -1
+      ? (thoughts[thoughts.length - 1]?.rank || 0) + 1
+      : index === 0
+        ? thoughts[0].rank - 1
+        : (thoughts[index - 1].rank + thoughts[index].rank) / 2
+  }
```
raineorshine, line 35: This duplicates most of the logic in the final return statement. When adding new code, we always have to consolidate and remove duplication, looking for shared functionality and thinking about the right abstractions. Copying and pasting code may be faster, but it will result in the code being less readable and harder to maintain the long run.

Please refactor this with an eye for readability, maintainability, and eliminating code duplication.

raineorshine (changes requested):

src/actions/editThought.ts
```diff
@@ -169,6 +169,20 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
     [editedThought.id]: thoughtNew,
   }
 
+  // If we're editing a note, update the parent thought's rank
+  if (isNote) {
+    const parentThought = getThoughtById(state, parentOfEditedThought.parentId)
+    if (parentThought) {
+      const newParentRank = getSortedRank(state, parentThought.parentId, newValue)
```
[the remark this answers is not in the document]
raineorshine: @RED-ROSE515 This doesn't appear to be resolved.

You should check the sort preference and only change the rank in lists sorted by note.

src/components/__tests__/SortPicker.ts
```diff
@@ -380,3 +381,172 @@ describe('Created to Updated', () => {
...
+
+  await click('[data-testid="toolbar-icon"][aria-label="SortPicker"]')
+  await click('[aria-label="sort options"] [aria-label="Note"]')
+  await act(() => vi.runAllTimersAsync())
+
+  const thoughts = screen.getAllByTestId(/thought/)
+  expect(thoughts.map((child: HTMLElement) => child.textContent)).toMatchObject(['b3', 'a2', 'c1'])
+})
+
+it('home: Note Desc with edit', async () => {
```
raineorshine: "home: Note Asc with edit"

src/components/__tests__/SortPicker.ts
```diff
@@ -380,3 +381,172 @@ describe('Created to Updated', () => {
...
+  })
+
+  await act(vi.runOnlyPendingTimersAsync)
+
+  await click('[data-testid="toolbar-icon"][aria-label="SortPicker"]')
+  await click('[aria-label="sort options"] [aria-label="Note"]')
+  await act(() => vi.runAllTimersAsync())
+
+  const thoughts = screen.getAllByTestId(/thought/)
+  expect(thoughts.map((child: HTMLElement) => child.textContent)).toMatchObject(['b', 'd3', 'a2', 'c1'])
```
raineorshine: The desired output is `d3, a2, c1, b`. Thoughts with notes should always be sorted above thoughts without notes, even in descending order.

src/util/compareThought.ts
```diff
@@ -221,3 +223,20 @@ export const compareThoughtByUpdated: ComparatorFunction<Thought> = (a: Thought,
 /** Compare two thoughts by their lastUpdated timestamp in descending order (newest first). Fall back to compareReasonable if created at the same time. */
 export const compareThoughtByUpdatedDescending: ComparatorFunction<Thought> = (a: Thought, b: Thought) =>
   compare(b.lastUpdated, a.lastUpdated) || compareReasonable(a.value, b.value)
+
+/** Get note value or fallback for a Thought value. */
+const getNoteOrFallback = (state: State, input: Thought): string => {
+  return noteValue(state, [input.id]) ?? input.value
```
raineorshine: A `Path` must be a contiguous chain of ancestors from the root. `[input.id]` would only work if the input thought was a child of the root. It won't work anywhere else in the tree. You need to pass the `Path` of the thought to determine its `noteValue` thanks to the new `=path` feature in #2951.

As mentioned before, we want to fall back to rank, not thought.value.

src/selectors/getSortedRank.ts
```diff
@@ -1,40 +1,57 @@
...
 
+/** Calculates the rank for a given index in a sorted array of thoughts. */
+const calculateRank = (thoughts: { rank: number }[], index: number): number => {
+  if (index === -1) {
+    return (thoughts[thoughts.length - 1]?.rank || 0) + 1
+  }
+  if (index === 0) {
+    return thoughts[0].rank - 1
+  }
+  return (thoughts[index - 1].rank + thoughts[index].rank) / 2
```
raineorshine: Please preserve the helpful comments from the old code.

src/selectors/getSortedRank.ts
```diff
@@ -1,40 +1,57 @@
...
   }
 
-  // For alphabetical sorting, use the existing logic
-  // find the first child with value greater than or equal to the new value
+  // Handle Note sorting
+  if (sortPreference.type === 'Note') {
+    const noteId = findDescendant(state, state.cursor![state.cursor!.length - 1], '=note')
+    const noteThought = firstVisibleChild(state, noteId!)!
+    const compareFn = isDescending ? compareThoughtByNoteDescending(state) : compareThoughtByNote(state)
+    const index = thoughts.findIndex(thought => compareFn({ ...noteThought, value }, thought) === -1)
```
raineorshine: `{ ...noteThought, value }` is not ideal, as it's trying to jam something into an existing structure to make it work. 

Even though it's a bit more complex, I think it would be better to do some lower level comparisons here than creating an invalid `Thought`.

src/selectors/getSortedRank.ts
```diff
@@ -1,40 +1,57 @@
...
-    // For "Updated and Created" sorting, always place at the beginning or end since this thought was just updated
     return isDescending ? thoughts[0].rank - 1 : (thoughts[thoughts.length - 1]?.rank || 0) + 1
   }
 
-  // For alphabetical sorting, use the existing logic
-  // find the first child with value greater than or equal to the new value
+  // Handle Note sorting
+  if (sortPreference.type === 'Note') {
+    const noteId = findDescendant(state, state.cursor![state.cursor!.length - 1], '=note')
+    const noteThought = firstVisibleChild(state, noteId!)!
```
raineorshine: Avoid bypassing the type system with the `!` operator. In rare cases it may be appropriate, and should be paired with a comment. I don't think it's safe here, since `firstVisibleChild` can return undefined.

src/components/__tests__/SortPicker.ts
raineorshine: Fixed in 10fe61096b

src/util/compareThought.ts
raineorshine: Fixed in 10fe61096b

src/selectors/getSortedRank.ts
raineorshine: Fixed in 10fe61096b

src/selectors/getSortedRank.ts
raineorshine: Fixed in 10fe61096b

src/actions/editThought.ts
```diff
@@ -134,10 +134,12 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
     [oldKey]: newOldLexeme,
     [newKey]: lexemeNew,
   }
-
-  const sortPreference = getSortPreference(state, editedThought.parentId)
+  const isNote = parentOfEditedThought.value === '=note'
+  const parentThoughtId = isNote
+    ? (getThoughtById(state, parentOfEditedThought.parentId)?.parentId ?? editedThought.parentId)
+    : editedThought.parentId
+  const sortPreference = getSortPreference(state, parentThoughtId)
```
raineorshine: Reusing the existing `sortPreference` in the outer scope will incorrectly cause the rank of `thoughtNew` to change when editing a note value within an alphabetically sorted list.

```
- a
- =sort
  - Alphabetical
    - Asc
  - b
    - =note
      - value
```

This may not have any adverse effect, as note values are expected to be the only child of `=note`, so the rank doesn't matter. However, this modification certainly isn't the intended behavior.

Instead, access the sort preference of the note's parent independently using a local variable within the `isNote` block.

src/actions/editThought.ts
```diff
@@ -134,10 +134,12 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
     [oldKey]: newOldLexeme,
     [newKey]: lexemeNew,
   }
-
-  const sortPreference = getSortPreference(state, editedThought.parentId)
+  const isNote = parentOfEditedThought.value === '=note'
+  const parentThoughtId = isNote
+    ? (getThoughtById(state, parentOfEditedThought.parentId)?.parentId ?? editedThought.parentId)
```
raineorshine: Generally I prefer `rootedParentOf(parentOf(path))` to access grandparents rather than using `getThoughtById`.

raineorshine (changes requested):

src/actions/editThought.ts
```diff
@@ -134,16 +134,20 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
...
     generating: false,
     rank:
-      newValue !== '' && (sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated')
-        ? getSortedRank(state, editedThought.parentId, newValue)
+      newValue !== '' && !isNote
+        ? (() => {
+            return sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated'
+              ? getSortedRank(state, editedThought.parentId, newValue)
+              : editedThought.rank
+          })()
```
raineorshine: There is no need for an IIFE here... 🤔

```diff
diff --git a/src/actions/editThought.ts b/src/actions/editThought.ts
index e4ece24247..8e7ec3634c 100644
--- a/src/actions/editThought.ts
+++ b/src/actions/editThought.ts
@@ -142,12 +142,8 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
     ...editedThought,
     generating: false,
     rank:
-      newValue !== '' && !isNote
-        ? (() => {
-            return sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated'
-              ? getSortedRank(state, editedThought.parentId, newValue)
-              : editedThought.rank
-          })()
+      newValue !== '' && !isNote && (sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated')
+        ? getSortedRank(state, editedThought.parentId, newValue)
         : editedThought.rank,
     value: newValue,
     lastUpdated: timestamp(),
```

src/actions/editThought.ts
```diff
@@ -134,16 +134,20 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
...
   const sortPreference = getSortPreference(state, editedThought.parentId)
   const sortType = sortPreference.type
 
   const thoughtNew: Thought = {
     ...editedThought,
     generating: false,
     rank:
-      newValue !== '' && (sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated')
-        ? getSortedRank(state, editedThought.parentId, newValue)
+      newValue !== '' && !isNote
```
raineorshine: Why `!isNote`?

raineorshine (changes requested):

src/actions/editThought.ts
RED-ROSE515: What do you mean IIFE?

src/actions/editThought.ts
RED-ROSE515: Ah, I think there is no need to check !isNote.

src/actions/editThought.ts
raineorshine: https://en.wikipedia.org/wiki/Immediately_invoked_function_expression

PR #2925, opened 2025-05-05
Allow multiple keyboard shortcuts per command 

### LINKED ISSUE #2867: Allow multiple keyboard shortcuts per command
## Current Behavior

Commands define aliases with `hideFromHelp: true` to allow additional keyboard shortcuts.

For example:

https://github.com/cybersemics/em/blob/177c2a73dd318f92f6462f6c9159e3fc40d93b01/src/commands/cursorDown.tsx#L134-L139

## Expected Behavior

- [ ] Expand [`Command.keyboard`](https://github.com/cybersemics/em/blob/177c2a73dd318f92f6462f6c9159e3fc40d93b01/src/%40types/Command.ts#L65-L66) to allow `Key[]`.
- [ ] Show the first `Key` in the array in Command Palette, Help modal, etc.
- [ ] The [commandKeyIndex](https://github.com/cybersemics/em/blob/177c2a73dd318f92f6462f6c9159e3fc40d93b01/src/commands.ts#L121) will need to be adapted to support multiple keyboard shortcuts but still allow O(1) lookup.
  - Consider adding one entry per keyboard shortcut (instead of one entry per command) so that `hashKeyDown` can still be used.
- [ ] Remove command aliases that are no longer needed.

karunkop:
Close #2867 

### Refactor:

- [X] Expand Command.keyboard to allow Key[].
- [X] Show the first Key in the array in Command Palette, Help modal, etc.
- [X] The commandKeyIndex will need to be adapted to support multiple keyboard shortcuts but still allow O(1) lookup. Consider adding one entry per keyboard shortcut (instead of one entry per command) so that hashKeyDown can still be used.
- [X] Remove command aliases that are no longer needed.


src/commands.ts
```diff
@@ -54,15 +54,19 @@ const digits = keyValueBy(Array(58 - 48).fill(0), (n, i) => ({
...
-export const hashCommand = (command: Command): string => {
-  const keyboard = typeof command.keyboard === 'string' ? { key: command.keyboard } : command.keyboard || ({} as Key)
-  return (
-    (keyboard.meta ? 'META_' : '') +
-    (keyboard.alt ? 'ALT_' : '') +
-    (keyboard.shift ? 'SHIFT_' : '') +
-    keyboard.key?.toUpperCase()
-  )
+/** Hash all the properties of a command into an array of strings that can be compared with the result of hashKeyDown. */
+export const hashCommand = (command: Command): string[] => {
```
raineorshine, line 58: One thing to keep in mind is that when you hash something you turn it into a string. So the names are a bit muddled now with `hashCommand` incongruously returning an array of strings. Furthermore, the comments are now confusing, as they say the result of `hashCommand` and `hashKeyDown` can be directly compared, but that is no longer the case. 

Instead, I would suggest preserving `hashCommand` as it is, and adding the array-handling logic to `index()`. `hashCommand` and `hashKeyDown` are a logical pair and we want to preserve that symmetry. Unfortunately Typescript cannot capture the fact that they both return the same subset of `string`.

raineorshine (changes requested):
Thanks!

Good job preserving O(1) `keyDown` lookups.

src/components/CommandRow.tsx
```diff
@@ -116,8 +116,12 @@ const CommandRow = ({
         {isTouch && command.gesture ? (
           <td className={css({ minWidth: { base: '10rem', _mobile: 'auto' }, textAlign: { _mobile: 'center' } })}>
             {isTouch && command.gesture ? (
-              // GesturePath[]
-              <GestureDiagram path={command.gesture as GesturePath} size={48} arrowSize={12} />
+              // Use first gesture for display
+              <GestureDiagram
+                path={Array.isArray(command.gesture) ? command.gesture[0] : (command.gesture as GesturePath)}
```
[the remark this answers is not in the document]
karunkop: `gestureString` is definitely applicable here. Sorry my bad!

src/commands.ts
```diff
@@ -56,7 +56,14 @@ const digits = keyValueBy(Array(58 - 48).fill(0), (n, i) => ({
 
 /** Hash all the properties of a command into a string that can be compared with the result of hashKeyDown. */
 export const hashCommand = (command: Command): string => {
+  if (!command.keyboard) return ''
+
+  // If command.keyboard is an array, return empty string as we handle arrays in index()
+  if (Array.isArray(command.keyboard)) return ''
```
raineorshine, line 62: This is a bit awkward. When you find yourself adding strange edge cases for things that are not directly related to the function, you should take a look at the overall design of the function and consider if it needs to change.

Here, `hashCommand` does not depend on anything other than `command.keyboard`. It was written to take `command: Command` out of convenience. But now that it's not convenient anymore, it's not a great approach. Passing the entire `Command` also makes handling keyboard arrays more awkward in `index()`.

Instead, change `hashCommand` to accept `Command['keyboard']` and the function will become simpler.

raineorshine (changes requested):

src/commands.ts
karunkop: Even when we just accept `Command['keyboard'] `type, don't we still need to handle `Key[]` type explicitly either by adding edge case like above or by using something like `Exclude<Command['keyboard'], Key[]>`?

src/components/CommandPalette.tsx
```diff
@@ -59,7 +59,7 @@ const CommandSearch: FC<{
       if (
         e.key === 'Escape' ||
         // manually check if the commandPalette command is entered since global commands are disabled while the command palette is open
-        hashKeyDown(e) === hashCommand(commandPaletteCommand)
+        commandMatchesKeyDown(commandPaletteCommand, e)
```
raineorshine, line 62: I have my doubts about this approach.

1. `commandMatchesKeyDown` is only used in one place. Typically we only factor out functions for shared functionality, or at least keep them in the same module.
2. `commandPaletteCommand` doesn't even use multiple keyboard shortcuts.
3. Passing the event object to another module is an odd separation of concerns.

Might there be a far simpler solution here?

src/commands.ts
```diff
@@ -54,15 +54,16 @@ const digits = keyValueBy(Array(58 - 48).fill(0), (n, i) => ({
...
-    (keyboard.alt ? 'ALT_' : '') +
-    (keyboard.shift ? 'SHIFT_' : '') +
-    keyboard.key?.toUpperCase()
-  )
+/**
+ * Hash a keyboard shortcut into a string that can be compared with the result of hashKeyDown.
+ * This function only handles a single keyboard shortcut, not arrays.
+ */
+export const hashCommand = (keyboard: Exclude<Command['keyboard'], Key[]>): string => {
+  if (!keyboard) return ''
```
raineorshine, line 62: According to the type definition, `keyboard` cannot possibly be falsey.

raineorshine (changes requested):

src/components/CommandPalette.tsx
karunkop: The simplest solution I can think of is to explicitly check that `commandPaletteCommand.keyboard` is not an array (i.e., not of type Key[]). That would look something like this:

```ts
commandPaletteCommand.keyboard && !Array.isArray(commandPaletteCommand.keyboard) &&
hashCommand(commandPaletteCommand.keyboard) === hashKeyDown(e)
```

TypeScript will enforce type safety here by preventing direct comparisons with mismatched types, since we're explicitly excluding the `Key[]` type from the `hashCommand` function.

This approach also aligns with statement 2, which assumes that the command doesn't support multiple keyboard shortcuts. The only reason I added the commandMatchesKeyDown function was to handle cases where a command may have multiple shortcuts and we need to compare all of them against the keydown event.

raineorshine: > The simplest solution I can think of is to explicitly check that `commandPaletteCommand.keyboard` is not an array (i.e., not of type Key[]).

We already know that `commandPaletteCommand.keyboard` is not an array, so we don't need to check it at runtime. Just type cast it, which tells the compiler that we know it's safe.

> The only reason I added the commandMatchesKeyDown function was to handle cases where a command may have multiple shortcuts and we need to compare all of them against the keydown event.

There is only one case that uses this, and it doesn't use an array. There's no need to design for hypothetical cases that do not exist. If we need it in the future, then we will refactor and build it then.

karunkop: Understood!

PR #2926, opened 2025-05-05
Use indent value to determine faux caret animation

### LINKED ISSUE #2842: [iOS] Caret transform bug when moving cursor from col2 child

ethan-james:
Fixes #2842

`LayoutTree` uses `indent` to determine how far to transform the editable area when navigating between thoughts at different depths. If `indent` doesn't change, then there is no animation that would necessitate `hideCaret`, so this is better than using `indentDepth + tableDepth` as a proxy.

This methods takes `indentDepth`, which is a decimal, and multiplies it by `10` to get a whole number. This number gets passed into `fauxCaretTreeProvider`, where it has modulo 16 applied to it in this line: `depth % hideCaretAnimationNames.length`. The depth number will change by 9 for each child thought, and 5 when it's in a table. These numbers don't collide very often, but apparently it is still possible.

```
- A
  - B
    - C
      - D
        - E
          - =view
            - Table
          - F
            - G
```

Tapping between "D" and "G" will still cause a problem. I'll need to revisit this and see if I can iron out any collisions.


raineorshine:
Thanks for the PR!

The original benefit of the modulo is that collisions were spaced out, e.g. 1 and 17, which was unlikely to happen with real-world user navigation. Multiplying by 10 breaks the distribution, causing collisions to sometimes be closer together.

Did we try `indentDepth + indentCursorAncestorTables`? That would be the closest integer calculation. If that doesn't work, I would try a smaller multiplier, like 1.1, to preserve the distance between collisions.

ethan-james:
I assume that you mean `Math.floor(indentCursorAncestorTables / fontSize)`, which will be a small integer compared to `indentCursorAncestorTables`, which by itself is 18 or 54 or some other large number.

Unfortunately, it has roughly the same problem as the `tableDepth` variable that I created, where it is `3` for the root thought in column 2, then `1` for that thought's children. Which leads to a situation where the grandchild of column 2 conflicts with that root thought.

I'll test out using 1.1 or 1.5 as a multiplier.

raineorshine:
Ah yes, you're right. 

Okay, keep at it!

ethan-james:
I propose multiplying by 2, which seems like the only solution to the "column 2 grandchild" problem, and then increasing the modulo to 24 or 32 to compensate for the fact that the value increases faster.

src/components/LayoutTree.tsx
```diff
@@ -247,7 +232,7 @@ const LayoutTree = () => {
         css({
           marginTop: '0.501em',
         }),
-        fauxCaretTreeProvider(indentDepth + tableDepth),
+        fauxCaretTreeProvider(Math.ceil(indent * 2)),
```
raineorshine: I've moved this into `fauxCaretTreeProvider`, since this seems more like an implementation detail of the faux caret than something the `LayoutTree` needs to know about.

Also added a comment explaining the number 2, since it is not at all obvious to other developers.

See: 76f0b2e7397

raineorshine (approved):
Thanks! I tested the two cases and poked around, and could not find any breaking cases.

PR #2928, opened 2025-05-08
Command palette v1.5

yangchristina:
- [x] [Avoid autoscroll on mouseover](https://github.com/cybersemics/em/pull/2928#issuecomment-2927006190)
- [x] [Include margin in autoscroll](https://github.com/cybersemics/em/pull/2928#issuecomment-2927013886)
- [x] [Search bar doesn't scale with font size](https://github.com/cybersemics/em/pull/2928#issuecomment-2934207886)
- [x] [Unexecutable commands not differentiated](https://github.com/cybersemics/em/pull/2928#issuecomment-2934226349)
- [x] [Intermediate frames rendered on edit](https://github.com/cybersemics/em/pull/2928#issuecomment-2934361496)
- [x] ~~[Gesture Diagram is too thin](https://github.com/cybersemics/em/pull/2928#issuecomment-2934430027)~~ - Out of scope
- [x] [No margin on narrow windows](https://github.com/cybersemics/em/pull/2928#issuecomment-2934696245)

# Mobile
Screenshot 2025-05-07 at 8 09 13 PM

https://github.com/user-attachments/assets/ad3896c0-f047-46fa-b267-ebb76f97a403

# Web
Screenshot 2025-05-07 at 8 10 34 PM
Screenshot 2025-05-07 at 8 10 44 PM
Screenshot 2025-05-07 at 8 11 13 PM

# Additional changes
- Removed Popup component, using PopupBase directly
- Removed indexInToolbar

Note: After this PR, there will be 1 more small command palette PR to animate the lotties


yangchristina:
@raineorshine sorry those pictures aren't accurate, I've added a video now. I was actually testing on web, with some variables changed, because it was easier to test styling there compared to in an emulator.

yangchristina:
@trevinhofmann could I please get a review? Thanks!

yangchristina:
Flaky test? The only thing I did was remove a comment.

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,222 @@
...
+                ? 'gray'
+                : isTouch
+                  ? selected
+                    ? 'blueHighlight'
+                    : gestureInProgress === command.gesture
+                      ? 'vividHighlight'
+                      : 'gray75'
+                  : selected
+                    ? 'vividHighlight'
+                    : 'gray75',
```
trevinhofmann, line 162: Also challenging for me to parse.

"If it is... not disabled, and not touch... but it is selected... If the `gestureInProgress` equals the `command.gesture`... then `'vividHighlight'`." And that's just one of 6 cases 😅

I would typically either create a helper function or compute this as a `labelColor` variable in the component, somewhere before the `return` statement.

trevinhofmann (changes requested):
Thanks for the pull request, @yangchristina! Can you please resolve the merge conflicts?

In the meantime, the UI is looking excellent! Just a few initial comments on the code below.

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,222 @@
...
+                        )
+                      })()
+                    : // For other commands, use normal highlighting
+                      command.id === 'cancel'
+                      ? selected
+                        ? 1
+                        : undefined
+                      : gestureInProgress.length
+                  : undefined
+              }
```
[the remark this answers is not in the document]
raineorshine: The nested ternaries themselves are fine, but the nested IIFE is making this hard to read. I would probably restructure this into a helper function. 

Note that this code is just [moved from CommandPalette.tsx](https://github.com/cybersemics/em/blob/c1b79a2700a0a0efb1be12a1d4d4aa9f2e66293e/src/components/CommandPalette.tsx#L201-L220) in `main` so it's out of scope in this PR.

src/components/CommandRowOnly.tsx
raineorshine: This logic is inherited from `CommandPalette`, but has been expanded to accommodate new conditions. This expansion in complexity probably justifies it being factored out.

Note that the original code already contains a redundant else branch:

```ts
              color: disabled
                ? 'gray'
                : isTouch
                  ? selected || (gestureInProgress as string) === gestureString(command)
                    ? 'vividHighlight'
                    : 'fg'
                  : 'fg',
```

Which could have been simplified to:

```ts
              color: disabled
                ? 'gray'
                : isTouch && (selected || (gestureInProgress as string) === gestureString(command))
                  ? 'vividHighlight'
                  : 'fg',
```

This redundancy gets preserved in the new code, but is worsened a bit because it's not just `fg` that gets repeated. We also may be able to use the fact that `gestureInProgress` is always an empty string on desktop to avoid extra `isTouch` checks.

The new code is great in terms of specificity, but could use some explanatory comments. The old code had some nesting which may make it slightly harder to read, but it only had three possible values, which were pretty intuitive: `gray` for disabled, `vividHighlight` for selected, and `fg` for default. In the new code, the differences between `gray` and `gray75`, or `blueHighlight` and `vividHighlight`, require some explanation since they are non-obvious. If there is no way to simplify the ternaries, then a helper function would be reasonable.

yangchristina: Is this too much?
```ts
color: disabled
                ? 'gray'
                : viewType === 'grid'
                  ? 'fg'
                  : isSelectedStyle || (isTouch && (gestureInProgress as string) === gestureString(command))
                    ? 'blueHighlight'
                    : 'gray75',
```

src/commands.ts
```diff
@@ -75,16 +75,20 @@ export const hashKeyDown = (e: KeyboardEvent): string =>
   // use e.keyCode if available instead
   (letters[e.keyCode] || digits[e.keyCode] || e.key || '').toUpperCase()
 
+const ARROW_KEYS_TO_CHARACTER = {
```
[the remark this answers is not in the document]
yangchristina: I've created an ArrowKey.ts file

src/components/CommandRowOnly.tsx
raineorshine: Looks fine to me!

yangchristina:
@trevinhofmann merge conflicts resolved, this is ready to review. Thanks!

src/e2e/puppeteer/__tests__/__image_snapshots__/commandPalette/commandPalette.png
trevinhofmann, line 1: This might just be me, but the text keys for the commands (e.g. `[ctrl] [alt] a` are difficult to read. I'm not sure if any other text in the application is this small.

trevinhofmann (changes requested):
Thank you for the updates, @yangchristina! I have some questions, comments, and suggestions below.

## 1. Hovering over Command Library items shows the pointer cursor

### Steps to Reproduce

1. Open the Help menu by clicking `?` at the bottom
2. Navigate to the Command Library tab (should be open by default)
3. Hover over a command item

### Current Behavior


https://github.com/user-attachments/assets/245cc2c1-fe9e-4bcb-aeca-b9d8c16ee3b0

### Expected Behavior

Because the items are not clickable, it should probably only show the default cursor and not the pointer (which indicates the element can be clicked).

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,268 @@
...
+                />
+              </div>
+            ) : (
+              <GestureDiagram
+                color={disabled ? token('colors.gray') : undefined}
+                styleCancelAsRegularGesture
+                highlight={
+                  !disabled && gestureInProgress !== undefined
+                    ? command.id === 'openGestureCheatsheet'
+                      ? // For gesture cheatsheet command, find the longest matching end portion
```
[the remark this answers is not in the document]
yangchristina: I believe this part (including the comment) was copied over from previous code, and just moved here.

src/components/CommandKeyboardShortcut.tsx
```diff
@@ -0,0 +1,69 @@
...
+      {keyboard.meta && (isMac ? <Kbd>⌘</Kbd> : <Kbd isText>ctrl</Kbd>)}
+      {keyboard.alt && (isMac ? <Kbd>⌥</Kbd> : <Kbd isText>alt</Kbd>)}
+      {keyboard.control && <Kbd isText>ctrl</Kbd>}
+      {keyboard.shift && <Kbd>⇧</Kbd>}
+      {keyboard.key === 'Backspace' ? (
+        <Kbd>⌫</Kbd>
+      ) : isArrowKey(regularKey) ? (
+        <Kbd>{arrowTextToArrowCharacter(regularKey)}</Kbd>
+      ) : (
+        <kbd className={css({ fontSize: '0.56em' })}>{regularKey}</kbd>
```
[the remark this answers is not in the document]
yangchristina: Kbd component adds a border around it, and in the mocks, these don't have a border. Maybe Kbd should be renamed to something like KbdWithBorder?

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,268 @@
...
+import { CommandViewType } from '../@types/CommandViewType'
+import GesturePath from '../@types/GesturePath'
+import { isTouch } from '../browser'
+import { gestureString } from '../commands'
+import CommandKeyboardShortcut from './CommandKeyboardShortcut'
+import GestureDiagram from './GestureDiagram'
+import HighlightedText from './HighlightedText'
+
+/** Renders a GestureDiagram and its label as a hint during a MultiGesture. */
+const CommandRowOnly = forwardRef<
```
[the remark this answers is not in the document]
yangchristina: Now that `CommandRow` has been renamed to `CommandItem`, this indeed doesn't sound most descriptive. `CommandItemOnly` may be better, but I agree, it's also not great.

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,268 @@
...
+                  borderRadius: '8px',
+                  textAlign: { _mobile: 'center' },
+                })}
+              >
+                <GestureDiagram
+                  cssRaw={css.raw({
+                    width: { sm: '80px', md: '130px' },
+                    height: { sm: '80px', md: '130px' },
+                  })}
+                  path={gestureString(command) as GesturePath}
```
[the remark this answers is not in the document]
yangchristina: Done thanks!

src/commands.ts
```diff
@@ -75,16 +76,20 @@ export const hashKeyDown = (e: KeyboardEvent): string =>
...
-const arrowTextToArrowCharacter = (s: string) =>
-  (
-    ({
-      ArrowLeft: '←',
-      ArrowRight: '→',
-      ArrowUp: '↑',
-      ArrowDown: '↓',
-    }) as Index
-  )[s] || s
+export const arrowTextToArrowCharacter = (s: string) => (ARROW_KEYS_TO_CHARACTER as Index)[s] || s
```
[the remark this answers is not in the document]
yangchristina: Done

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,268 @@
...
+    disabled?: boolean
+    isActive?: boolean
+    isTable?: boolean
+    /**
+     * Controls when the command description is displayed.
+     * - If true: Always show the description.
+     * - If false: On non-touch devices, always show the description. On touch devices, show only when the command is selected.
+     */
+    alwaysShowDescription?: boolean
+    cssRaw?: SystemStyleObject
```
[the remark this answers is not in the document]
yangchristina: This is used for:

```
      cssRaw={css.raw({
        ...(customize
          ? {
              '&:active, [data-drop-to-remove-from-toolbar-hovering] &': {
                WebkitTextStrokeWidth: '0.05em',
              },
            }
          : null),
      })}
```

Since "Customize toolbar" will be removed eventually, it can be removed later on. 
Otherwise, we can add a different prop now, something like `isDragging` instead of this `cssRaw`?

src/e2e/puppeteer/__tests__/__image_snapshots__/commandPalette/commandPalette.png
yangchristina: I've made it bigger, is this better? Or still too small?

trevinhofmann:
Thank you for the updates, @yangchristina! Can you please confirm if this PR is ready for review again? I'll be happy to take another look right away.

src/components/CommandPalette.tsx
```diff
@@ -420,23 +276,29 @@ const CommandPalette: FC<{
...
     <div
       className={css({
-        ...(gestureInProgress || !isTouch ? { paddingLeft: '4em', paddingRight: '1.8em' } : null),
         marginBottom: isTouch ? 0 : fontSize,
         textAlign: 'left',
+        border: '1px solid {colors.gray15}',
+        borderRadius: '12px',
+        backgroundColor: 'gray09',
+        maxWidth: '100%',
+        ...(isTouch ? { width: '100%' } : { width: '826px' }),
```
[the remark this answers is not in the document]
yangchristina: Screenshot 2025-05-28 at 8 53 12 AM

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+    const fontSize = useSelector(state => state.fontSize)
+
+    const paddingSize = fontSize * (isTouch && !isTable ? 0.4 : 0.67)
+
+    const isSelectedStyle = selected || isDragging
+
+    return (
+      <Container
+        // @ts-expect-error Container can be 'tr' or 'div'
```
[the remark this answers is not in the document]
yangchristina: Couldn't figure out how to do this typewise. Maybe we can move this into CommandRowOnly?

Since this is the only usage of ref. Then instead of ref, we'd have to add two more props `onHover` and `scrollSelectedIntoView`

```
  useEffect(() => {
    if (selected) {
      ref.current?.scrollIntoView({ block: 'nearest' })
    }
  })

  useEffect(() => {
    /** Hover handler. */
    const onHoverCommand = (e: MouseEvent) => onHover(e, command)

    // mouseover and mouseenter cause the command under the cursor to get selected on render, so we use mousemove to ensure that it only gets selected on an actual hover
    ref.current?.addEventListener('mousemove', onHoverCommand)

    return () => {
      // eslint-disable-next-line react-hooks/exhaustive-deps
      ref.current?.removeEventListener('mousemove', onHoverCommand)
    }
  }, [onHover, command])
```

yangchristina:
> Thank you for the updates, @yangchristina! Can you please confirm if this PR is ready for review again? I'll be happy to take another look right away.

@trevinhofmann this is ready for re-review thanks! A lot of what I left was questions though.

src/components/CommandKeyboardShortcut.tsx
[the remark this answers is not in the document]
trevinhofmann: > Kbd component adds a border around it, and in the mocks, these don't have a border.

Thank you for the explanation; that makes sense!

> Maybe Kbd should be renamed to something like KbdWithBorder?

_Maybe_, but I would personally be okay with leaving it as-is. There are other styles applied than the border, and naming it something like `StyledKbd` would be redundant as most components are styled :)

src/components/CommandRowOnly.tsx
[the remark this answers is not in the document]
trevinhofmann: I don't think the naming of this is a blocking issue, so I'll leave this one to @raineorshine to see if there is any preference.

src/components/CommandRowOnly.tsx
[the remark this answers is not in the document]
trevinhofmann: We do have the `isDragging` value in `CommandItem.tsx`, although we're currently activating the styles based on `customize`. Is that because the `&:active` CSS selector is equivalent to `isDragging`?

On one hand, this might not be an important issue (especially if we plan to remove the component soon anyway).

On the other, I'd be happy to push a commit to use another prop to CommandRowOnly which applies these styles since it's a simple change. My only questions would be:

* `CommandRowOnly` already has an `isDragging` prop (which no other components are currently passing to it, I'll make a separate thread for that).
* Should we pass `customize`, `isDragging`, or `customize && isDragging` as the value for that prop?

src/components/CommandRowOnly.tsx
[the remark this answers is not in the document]
trevinhofmann: I just spent a couple hours of my own time trying to get the typing of `ref` correct, but polymorphic components in TS are **_hard_**. The only solutions I came up with were too complex to be worth doing, e.g.:
1. Type abstractions for PolymorphicProps and PolymorphicRef to address this with generics
2. Change the ref type to use `HTMLDivElement | HTMLTableRowElement`
3. Split the component into two separate components for the div/tr displays using `HTMLDivElement` for the ref of one and `HTMLTableRowElement` for the other

(1) has an untenable amount of abstraction and also had side effects that interfered with the types of non-ref props

(2) does not work, as the TS engine is not able to infer the ref type correctly based on `isTable`

(3) duplicates all the logic of `CommandRowOnly`, which is obviously unacceptable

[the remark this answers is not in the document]
trevinhofmann: Just for reference, (1) looked something like:

```ts
import { ComponentPropsWithoutRef, ElementType } from 'react'

type AsProp<C extends ElementType> = { as?: C }

type PropsToOmit<C extends ElementType, P> = keyof (AsProp<C> & P)

type PolymorphicProps<C extends ElementType, Props = object> = Props &
  Omit<ComponentPropsWithoutRef<C>, PropsToOmit<C, Props>> &
  AsProp<C>

export default PolymorphicProps
```

```ts
import { ComponentPropsWithRef, ElementType } from 'react'

type PolymorphicRef<C extends ElementType> = ComponentPropsWithRef<C>['ref']

export default PolymorphicRef
```

src/e2e/puppeteer/__tests__/__image_snapshots__/commandPalette/commandPalette.png
trevinhofmann: That looks much more readable to me. I wouldn't usually make design recommendations, but it was a genuine usability concern. Thank you!

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,267 @@
...
+      selected,
+      command,
+      gestureInProgress,
+      style,
+      disabled,
+      isActive,
+      isTable,
+      alwaysShowDescription,
+      cssRaw,
+      isDragging,
```
trevinhofmann, line 52: There are two components that use `<CommandRowOnly>`, and neither of them pass `isDragging`. Is this prop needed?

trevinhofmann (changes requested):

We just have these discussions remaining, and a comment about the `isDragging` prop:

1. https://github.com/cybersemics/em/pull/2928#discussion_r2106069852
2. https://github.com/cybersemics/em/pull/2928#discussion_r2106071771
3. https://github.com/cybersemics/em/pull/2928#discussion_r2106073441
4. https://github.com/cybersemics/em/pull/2928#discussion_r2106074141

src/components/CommandPalette.tsx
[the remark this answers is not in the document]
trevinhofmann: I believe the second screenshot would be achieved with the suggested change of `width: '826px'` 👍

src/components/CommandRowOnly.tsx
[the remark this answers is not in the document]
trevinhofmann: @raineorshine confirmed that (B) would be an improvement over the `@ts-expect-error`. Thank you, @yangchristina!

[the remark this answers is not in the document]
yangchristina: I've updated to B, moving both `useDrag`, `onHover` and `scroll selected into view` into `CommandItem` (formerly named `CommandRowOnly`

src/components/CommandRowOnly.tsx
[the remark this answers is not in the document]
yangchristina: After applying changes for moving drag handlers into `CommandRowOnly`, `CommandItem` didn't really do anything expect pass props to `CommandRowOnly`, so I removed it. I them renamed `CommandRowOnly` to `CommandItem` since I thought that would make more sense.

src/components/CommandRowOnly.tsx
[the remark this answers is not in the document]
yangchristina: I moved `customize` into `CommandRowOnly` (now `CommandItem`) since I needed it for drag handlers, so I removed `cssRaw`, and moved this directly in `CommandRowOnly`

src/components/CommandRowOnly.tsx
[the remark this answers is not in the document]
yangchristina: Turns out we no longer need `useDrag` and `customize`, so that can be removed altogether in a separate PR.

src/components/CommandRowOnly.tsx
yangchristina: Note this prop will be removed in a separate PR, functionality no longer needed

src/components/CommandRowOnly.tsx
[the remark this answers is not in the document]
yangchristina: `customize` will be removed altogether in separate PR

yangchristina:
@trevinhofmann this is ready for re-review, thanks! Note that Customize Toolbar has defects, but since it's getting removed, we are ignoring those.

raineorshine:
> @trevinhofmann this is ready for re-review, thanks! Note that Customize Toolbar has defects, but since it's getting removed, we are ignoring those.

I can confirm: The Toolbar is going to be removed completely in the next design! Replaced by the new Command Menu which has an early version in `main`.

trevinhofmann (changes requested):
Thank you, @yangchristina. The UI looks great to me!

There are some issues with the customize toolbar (e.g. drag and drop not working), but I'll ignore them since many are already present in `main` and we'll be removing it anyway 👍

I also observed that search in the Command Library/Palette is suboptimal, but I believe we intend to improve/standardize search with a later PR.

There are just a few minor code suggestions, otherwise I believe this looks good!

src/components/CommandTableOnly.tsx
```diff
@@ -44,21 +31,30 @@ const CommandTableOnly = ({
...
-              indexInToolbar={indexInToolbar !== -1 && applyIndexInToolbar ? indexInToolbar + 1 : null}
-              onSelect={onSelect}
-              selected={selectedCommand && command?.id === selectedCommand.id}
-              command={command}
-              search={search}
-            />
-          )
-        })}
+        {commands
+          .filter((command): command is Command => command !== null)
```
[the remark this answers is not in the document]
yangchristina: Done thanks!

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,313 @@
...
+  const Icon = command.svg
+  const ref = React.useRef<Pick<HTMLDivElement, 'scrollIntoView' | 'addEventListener' | 'removeEventListener'> | null>(
+    null,
+  )
+
+  const [{ isDragging }, dragSource] = useDrag({
+    type: DragAndDropType.ToolbarButton,
+    item: (): DragToolbarItem => {
+      store.dispatch(dragCommand(command?.id || null))
+      return { command: command!, zone: DragCommandZone.Remove }
```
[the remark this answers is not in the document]
yangchristina: Removed thanks! this code was moved over, and didn't check carefully enough to see if anything else had changed.

src/components/CommandRowOnly.tsx
```diff
@@ -0,0 +1,313 @@
...
+  style,
+  disabled,
+  isActive,
+  isTable,
+  alwaysShowDescription,
+  onHover,
+  customize,
+  shouldScrollSelectionIntoView,
+}) => {
+  const label = command.labelInverse && isActive ? command.labelInverse! : command.label
```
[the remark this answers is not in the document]
yangchristina: removed thanks!

trevinhofmann (approved):

yangchristina:
@trevinhofmann made one more change (375f483) to extract ternary into util function from discussion: https://github.com/cybersemics/em/pull/2928#discussion_r2088031998

raineorshine:
# Avoid autoscroll on mouseover

Desktop

## Steps to Reproduce

Mouse over commands at the top or bottom of the Command Palette.

## Current Behavior

The Command Palette autoscrolls to fit the newly selected command.

https://github.com/user-attachments/assets/c64471f1-10ea-4c25-a400-4309ffde7c50


## Expected Behavior

The Command Palette should not autoscroll on mouseover, as it is slightly jarring. It also implies that the list will continue scrolling if the mouse is held at the edge of the panel, which is not true.

Instead, only autoscroll when the selection changes via the keyboard.

raineorshine:
# Include margin in autoscroll

## Steps to Reproduce

Arrow Down until one of the commands below-the-fold is selected.
Arrow Up until the first command is selected.

## Current Behavior

There is no space between the bottom of the highlight and the bottom of the Command Palette.

Screenshot 2025-06-01 at 11 42 04 AM

Similarly when the top command is selected.

Screenshot 2025-06-01 at 11 43 01 AM

## Expected Behavior

There should be a margin between the bottom of the highlight and the bottom of the Command Palette when the bottom command is selected.

Screenshot 2025-06-01 at 11 42 28 AM

There should be a margin between the top of the highlight and the edge of the Search Bar when the top command is selected.

Additionally, if the first command in the list is selected, the Command Palette should be scrolled all the way to the top.

Screenshot 2025-06-01 at 11 43 07 AM

raineorshine:
# Holding the arrow keys does not smoothly iterate through the list

I'm not sure if this is just a performance issue, if there is active throttling, or if it is an implication of the select/autoscroll mechanism. 

Consider using [throttleAnimationFrame](https://github.com/cybersemics/em/blob/main/src/util/throttleByAnimationFrame.ts) which is [used by cursorUp/Down](https://github.com/cybersemics/em/blob/84d5b8b0a5c2936c246983a0b0aaf5bac48328e9/src/commands/cursorDown.ts#L48) for the same purpose.

## Steps to Reproduce

Hold ArrowDown to iterate down through the commands.
Hold ArrowUp to iterate up through the commands.

## Current Behavior

The selection change is slow and choppy.

https://github.com/user-attachments/assets/0ca81d23-4ac0-4768-b2b3-9d7e7987147e


## Expected Behavior

As in `main`, the selection change should be fast as the arrow keys are held down. The selection should only move up/down by one command per animation frame.

https://github.com/user-attachments/assets/e17d1a5b-0a76-416c-91f4-7c956df5af73

raineorshine:
Review in progress

raineorshine:
# Gesture diagram is too thin

Mobile

## Steps to Reproduce

Swipe →↓ and hold.

## Current Behavior

Gesture diagram is rendered too thin, and arrow heads are hard to see:

Screenshot 2025-06-03 at 10 48 30 AM

## Expected Behavior

This should be unchanged from `main`, where the GestureDiagram is rendered with a thicker stroke and arrow head:

Screenshot 2025-06-03 at 10 48 18 AM

src/components/CommandItem.tsx
```diff
@@ -1,57 +1,71 @@
...
-}
-
-type ViewType = 'grid' | 'table'
-
-/** Renders all of a command's details as either a grid item or table row. */
-const CommandItem = ({
+/** Renders a GestureDiagram and its label as a hint during a MultiGesture. */
+const CommandItem: FC<{
+  viewType?: CommandViewType
+  search: string | undefined
```
raineorshine: Prefer `search?: string`

src/components/CommandItem.tsx
```diff
@@ -1,57 +1,71 @@
...
+  /** Click handler for the command. Applies pointer styles when defined. */
+  onClick?: (e: React.MouseEvent, command: Command) => void
+  selected: boolean | undefined
```
raineorshine: Prefer `selected?: boolean`

src/components/CommandItem.tsx
```diff
@@ -1,57 +1,71 @@
...
+  command: Command
+  gestureInProgress?: string
+  style?: React.CSSProperties
```
raineorshine: Try to avoid wide open types such as `style` or `cssRaw`, as we lose a lot of type safety with this. Which variants do we need to support, how might this be encapsulated, and what public API should the component expose?

src/components/CommandItem.tsx
```diff
@@ -1,57 +1,71 @@
...
+  disabled?: boolean
+  isActive?: boolean
+  isTable?: boolean
```
raineorshine: What's the difference between `isTable` and `viewType="table"`?

src/components/CommandItem.tsx
```diff
@@ -1,57 +1,71 @@
...
+  viewType?: CommandViewType
+  search: string | undefined
+  /** Click handler for the command. Applies pointer styles when defined. */
+  onClick?: (e: React.MouseEvent, command: Command) => void
+  selected: boolean | undefined
+  command: Command
+  gestureInProgress?: string
+  style?: React.CSSProperties
+  disabled?: boolean
+  isActive?: boolean
```
raineorshine: `isActive` should probably not be a prop, since it is derivable from the Redux state.

Generally, props should be orthogonal and complexity should be pushed down the component hierarchy.

src/components/CommandItem.tsx
```diff
@@ -65,29 +79,81 @@ const CommandItem = ({
...
+  const isSelectedStyle = selected || isDragging
+
+  useEffect(() => {
+    if (selected && shouldScrollSelectionIntoView) {
+      ref.current?.scrollIntoView({ block: 'nearest' })
+    }
   })
 
-  if (!command) return null
+  useEffect(() => {
```
raineorshine: I realize it's like this in `main`, but is there any reason not to use the typical `onMouseMove` instead of manually attaching the event listener? If we can, let's switch it over while we're here.

src/components/CommandItem.tsx
```diff
@@ -65,29 +79,81 @@ const CommandItem = ({
...
+    // mouseover and mouseenter cause the command under the cursor to get selected on render, so we use mousemove to ensure that it only gets selected on an actual hover
+    ref.current?.addEventListener('mousemove', onHoverCommand)
+
+    return () => {
+      ref.current?.removeEventListener('mousemove', onHoverCommand)
+    }
+  }, [onHover, command])
+
+  /** Returns the first n segments of the gesture diagram to highlight. */
+  const getGestureHighlight = () => {
```
raineorshine: This might make more sense as a `useMemo`.

src/components/CommandPalette.tsx
```diff
@@ -109,12 +108,13 @@ const CommandSearch: FC<{
...
         type='text'
-        placeholder='Search commands by name...'
+        placeholder='Search for a command'
         ref={inputRef}
         onInput={(e: React.ChangeEvent<HTMLInputElement>) => {
           onInput?.(e.target.value)
         }}
-        className={css({ marginLeft: 0, marginBottom: 0, border: 'none' })}
+        className={css({ padding: 0, border: 'none', background: 'transparent' })}
+        style={{ marginBlock, marginInline: marginBlock * 2 }}
```
raineorshine: To specify styles that scale with font size, you can use the `em` css unit. Since the font size of all content is [set to state.fontSize in AppComponent](https://github.com/cybersemics/em/blob/2326b3ba1f32e2ad599748bc033113521751608d/src/components/AppComponent.tsx#L197), `1em` will effectively be the same as `state.fontSize`. 

Using the css unit avoids the `useSelector` dependency and is PandaCSS compatible, i.e. it can be specified in `className={css(...)}` rather than `style=...`.

**UPDATE:** We ought to set the fontSize on the root (html) element so we can use `rem` as intended, but that will require a larger refactor. Tracked in #3018.

src/components/CommandPalette.tsx
```diff
@@ -420,23 +257,29 @@ const CommandPalette: FC<{
...
+        borderRadius: '12px',
+        backgroundColor: 'gray09',
+        maxWidth: '100%',
+        width: 826,
+        maxHeight: isTouch ? undefined : '500px',
+        cursor: 'default',
+        display: 'flex',
+        flexDirection: 'column',
       })}
+      onClick={e => e.stopPropagation()}
```
raineorshine: Could you document the purpose of `e.stopPropagation`? Stopping propagation can have an impact on ancestors all the way up the component hierarchy, so it's good to document in in case there is a problem in the future.

src/components/CommandPalette.tsx
```diff
@@ -445,18 +288,15 @@ const CommandPalette: FC<{
...
           <div
             className={css({
-              marginLeft: !isTouch ? '-2.4em' : '-1.2em',
-              // offset the negative marginTop on CommandRow
-              paddingTop: 5,
-              ...(!isTouch ? { maxHeight: 'calc(100vh - 8em)', overflow: 'auto' } : null),
+              ...(!isTouch ? { maxHeight: 'calc(100vh - 10em)' } : null),
+              overflow: 'auto',
             })}
+            style={{ padding: fontSize * (2 / 3) }}
```
raineorshine: This would be better as `padding: 0.66em` as described in https://github.com/cybersemics/em/pull/2928#discussion_r2123447495.

src/components/CommandItem.tsx
```diff
@@ -96,90 +162,152 @@ const CommandItem = ({
...
       })}
-      ref={dragSource}
-      onClick={onSelect ? () => onSelect(selected ? null : command) : undefined}
+      onClick={e => {
+        if (!disabled) {
+          onClick?.(e, command)
+        }
+      }}
+      style={{
+        fontSize,
```
raineorshine: This does not seem to do anything, since this is the default font size.

src/components/CommandItem.tsx
```diff
@@ -65,29 +79,81 @@ const CommandItem = ({
...
+      ? descriptionStringOrFunction(state)
+      : descriptionStringOrFunction
+  })
+
+  const Container = isTable ? 'tr' : 'div'
+  const Cell = isTable ? 'td' : 'div'
+
+  const fontSize = useSelector(state => state.fontSize)
+
+  const paddingSize = fontSize * (isTouch && !isTable ? 0.4 : 0.67)
```
raineorshine: This would be better specified in native css `em` units, as described in https://github.com/cybersemics/em/pull/2928#discussion_r2123447495.

src/components/GestureDiagram.tsx
```diff
@@ -30,6 +30,7 @@ interface GestureDiagramProps {
   cssRaw?: SystemStyleObject
   /** Whether to render the gesture with rounded corners. */
   rounded?: boolean
+  styleCancelAsRegularGesture?: boolean
```
raineorshine, line 34: JSDOC needed

raineorshine (changes requested):
Great work! Well organized, good integration with existing structure, and I didn't find any major architectural issues. 

Visual and behavioral cases document above.

Please let me know if you have any questions!

src/components/CommandItem.tsx
yangchristina: `isTable` determines whether we use `<tr> <td>` elements vs `div`. Only command palette uses `div`.
What is better, changing command palette to also use table elements, or removing `isTable` prop and doing `const isHTMLTable = !!viewType` (CommandPalette doesn't pass in `viewType`)

src/components/CommandItem.tsx
yangchristina: becbca7

src/components/CommandItem.tsx
yangchristina: becbca7

src/components/CommandItem.tsx
yangchristina: style prop removed in becbca7 for `noInlinePadding`

src/components/CommandItem.tsx
yangchristina: removed

src/components/CommandItem.tsx
yangchristina: [56fd0db](https://github.com/cybersemics/em/pull/2928/commits/56fd0db45ab1e2f8113acd2e6f5f8a9b8525d74d)

src/components/CommandItem.tsx
yangchristina: done, variable called `gestureHighlight`

src/components/CommandItem.tsx
yangchristina: This does affect styling due to the difference between `em` and `rem` as you said [here](https://github.com/cybersemics/em/pull/2928#discussion_r2123447495).
This can be removed once we switch to using `rem`.

For example here is removing the style on "Back"
Screenshot 2025-06-07 at 11 31 03 AM

src/components/CommandItem.tsx
yangchristina: in order to use `em`, I've moved the anchor `style={{ fontSize }}` into the parent of `CommandItem` (`CommandTableOnly`, so that the container element can also use `em` units.

src/components/CommandItem.tsx
yangchristina: Moved this to parent due to below: https://github.com/cybersemics/em/pull/2928#discussion_r2134079891

yangchristina: [523b249](https://github.com/cybersemics/em/pull/2928/commits/523b2490a452c91e46e303bcf1a080a04f039ebe)

src/components/CommandItem.tsx
yangchristina: [523b249](https://github.com/cybersemics/em/pull/2928/commits/523b2490a452c91e46e303bcf1a080a04f039ebe)

src/components/GestureDiagram.tsx
yangchristina: [0cea41c](https://github.com/cybersemics/em/pull/2928/commits/0cea41c841860161cf52c3324c7331a5133ce5d5)

yangchristina:
> # Unexecutable commands not differentiated
> ## Steps to Reproduce
> 1. Set the cursor to null.
> 2. Open the Command Palette and scroll down to where the disabled commands start ("Add to Favorites").
> 
> ## Current Behavior
> Commands that cannot be executed from the current state are shown with the same style as enabled commands.
> 
> Screenshot 2025-06-03 at 9 55 30 AM
> ## Expected Behavior
> As in `main`, commands that cannot be executed from the current state should be grayed out so they are clearly distinguished.
> 
> Screenshot 2025-06-03 at 9 55 55 AM
> ## Background
> Each command has a `canExecute` method which determines if it can be executed from the current Redux state. The most common case of a command not being executable is if the cursor is null, but there are other conditions (e.g. `swapParent` requires that the cursor have a parent, etc.)
> 
> https://github.com/cybersemics/em/blob/2326b3ba1f32e2ad599748bc033113521751608d/src/%40types/Command.ts#L73-L74

@raineorshine I couldn't find any mocks for this. How should these look?

raineorshine:
You can just dim the unexecutable commands enough to make them visually distinct.

yangchristina:
> # Gesture diagram is too thin
> Mobile
> 
> ## Steps to Reproduce
> Swipe →↓ and hold.
> 
> ## Current Behavior
> Gesture diagram is rendered too thin, and arrow heads are hard to see:
> 
> Screenshot 2025-06-03 at 10 48 30 AM
> ## Expected Behavior
> This should be unchanged from `main`, where the GestureDiagram is rendered with a thicker stroke and arrow head:
> 
> Screenshot 2025-06-03 at 10 48 18 AM

I've been trying a lot of things, but if the gesture is large and thick, the arrow disappears. The thinner it is, the larger the arrow. I cannot get both the arrow and the line to be large.
Screenshot 2025-06-07 at 12 59 39 PM

Can't really get better than this (although this might be worse than before I'm not sure)
Screenshot 2025-06-07 at 1 04 29 PM

yangchristina:
> You can just dim the unexecutable commands enough to make them visually distinct.

Thanks! Added

Screenshot 2025-06-07 at 9 24 59 PM

yangchristina:
> # Search bar doesn't scale with font size
> Update: I see this is actually broken on `main`. Might as well fix it while we're here! Thanks!
> 
> ## Steps to Reproduce
> Increase Font Size.
> 
> ## Current Behavior
> Search bar and placeholder does not scale with font size.
> 
> Screenshot 2025-06-03 at 9 52 45 AM
> ## Expected Behavior
> Search bar and placeholder should scale with font size.

Done: [615470c](https://github.com/cybersemics/em/pull/2928/commits/615470c3877d7018f76c2375f2cf009905519436)

yangchristina:
Working on next:
- Immediate frames: https://github.com/cybersemics/em/pull/2928#issuecomment-2934361496

- Height move: https://github.com/cybersemics/em/pull/2928#issuecomment-2934196427

Open issue:
- Gesture diagram: https://github.com/cybersemics/em/pull/2928#issuecomment-2934430027 (https://github.com/cybersemics/em/pull/2928#issuecomment-2952958184)

src/components/CommandPalette.tsx
yangchristina: [a754059](https://github.com/cybersemics/em/pull/2928/commits/a7540595a9aafb15becae6dcf5d5a6ea91484c28)

src/components/CommandItem.tsx
raineorshine: `viewType` can only be `'table'` or `'grid'`, and it defaults to `'table'`. So I think you want `viewType === 'table'`?

src/components/CommandItem.tsx
raineorshine: Got it, thanks for confirming!

raineorshine:
> > You can just dim the unexecutable commands enough to make them visually distinct.
> 
> Thanks! Added
> 
> Screenshot 2025-06-07 at 9 24 59 PM

It's still a bit hard to tell them apart, I think because of the title. If you could dim the title down to the same color as the description I think that would fix it!

trevinhofmann:
>> I've been trying a lot of things, but if the gesture is large and thick, the arrow disappears. The thinner it is, the larger the arrow. I cannot get both the arrow and the line to be large.
>
> Perhaps @trevinhofmann can help on this one if you're stuck.

@yangchristina -- Please share any other progress you might have made on this, otherwise I'll plan to take a look into the issue tomorrow (Tuesday). Thank you very much!

src/components/CommandItem.tsx
yangchristina: On main, Gesture Cheatsheet uses table elements, and Gesture Cheatsheet is what uses `viewType==='grid'`
I guess it doesn't make much sense for a grid to use table elements though. Should I change it to use div then?

yangchristina:
> > @raineorshine I've been trying a lot of things, but if the gesture is large and thick, the arrow disappears. The thinner it is, the larger the arrow. I cannot get both the arrow and the line to be large.
> 
> [#2928 (comment)](https://github.com/cybersemics/em/pull/2928#issuecomment-2952958184)
> 
> Yeah, the GestureDiagram is a bit finicky in that way! The proportions are right in `main`, so I'm wondering what changed?
> 
> Perhaps @trevinhofmann can help on this one if you're stuck.

I think it's because of the size change? The icons are now supposed to be bigger?

yangchristina:
> > > You can just dim the unexecutable commands enough to make them visually distinct.
> > 
> > 
> > Thanks! Added
> > Screenshot 2025-06-07 at 9 24 59 PM
> 
> It's still a bit hard to tell them apart, I think because of the title. If you could dim the title down to the same color as the description I think that would fix it!

Is this good?

Screenshot 2025-06-10 at 8 06 41 PM

yangchristina:
> # Holding the arrow keys does not smoothly iterate through the list
> ~I'm not sure if this is just a performance issue, if there is active throttling, or if it is an implication of the select/autoscroll mechanism.~ The performance improves as the list is filtered down, so it's likely a performance issue.
> 
> Consider using [throttleAnimationFrame](https://github.com/cybersemics/em/blob/main/src/util/throttleByAnimationFrame.ts) which is [used by cursorUp/Down](https://github.com/cybersemics/em/blob/84d5b8b0a5c2936c246983a0b0aaf5bac48328e9/src/commands/cursorDown.ts#L48) for the same purpose.
> 
> ## Steps to Reproduce
> Hold ArrowDown to iterate down through the commands. Hold ArrowUp to iterate up through the commands.
> 
> ## Current Behavior
> The selection change is slow and choppy.
> 
>  Screen.Recording.2025-06-01.at.11.57.47.AM.mov 
> ## Expected Behavior
> As in `main`, the selection change should be fast as the arrow keys are held down. The selection should only move up/down by one command per animation frame.
> 
>  Screen.Recording.2025-06-01.at.11.56.42.AM.mov


336e235

https://github.com/user-attachments/assets/5e174427-709c-427d-b920-4060d9e4ec99

src/components/CommandPalette.tsx
```diff
@@ -66,9 +56,6 @@ const CommandSearch: FC<{
         e.preventDefault()
         e.stopPropagation()
         dispatch(commandPalette())
-
-        // remove keydown listener now since unmount is not triggered until the animation ends
-        window.removeEventListener('keydown', onKeyDown)
```
yangchristina, line 71: Let me know if I'm wrong but I don't think this is necessary anymore due to `throttleByAnimationFrame`

yangchristina:
> # Include margin in autoscroll
> ## Steps to Reproduce
> Arrow Down until one of the commands below-the-fold is selected. Arrow Up until the first command is selected.
> 
> ## Current Behavior
> There is no space between the bottom of the highlight and the bottom of the Command Palette: Screenshot 2025-06-01 at 11 42 04 AM
> 
> Similarly when the top command is selected: Screenshot 2025-06-01 at 11 43 01 AM
> 
> ## Expected Behavior
> There should be a margin between the bottom of the highlight and the bottom of the Command Palette when the bottom command is selected: Screenshot 2025-06-01 at 11 42 28 AM
> 
> There should be a margin between the top of the highlight and the edge of the Search Bar when the top command is selected.
> 
> You may need special cases for the first and last commands to ensure that the scroll position is at the very top and very bottom, respectively. We don't want to leave a few pixels of additional scroll available in those cases.
> 
> Screenshot 2025-06-01 at 11 43 07 AM Screenshot 2025-06-01 at 11 50 51 AM

Is this a good amount for margin?

https://github.com/user-attachments/assets/3c727dcb-1618-42c7-8ab4-000ee61dc8ae

raineorshine:
Thanks for bringing this up. The short answer is no, there's no reason for the brittleness. Probably the result of early design decisions plus tweaks that got bolted on when applying the GestureDiagram to new contexts. 

Given the complexity, I suggest we leave the GestureDiagram as-is in this PR, and open a separate issue to refactor it.

yangchristina:
> # Avoid autoscroll on mouseover
> Desktop
> 
> ## Steps to Reproduce
> Mouse over commands at the top or bottom of the Command Palette.
> 
> ## Current Behavior
> The Command Palette autoscrolls to fit the newly selected command.
> 
>  Screen.Recording.2025-06-01.at.11.37.01.AM.mov 
> ## Expected Behavior
> The Command Palette should not autoscroll on mouseover, as it is slightly jarring. It also implies that the list will continue scrolling if the mouse is held at the edge of the panel, which is not true.
> 
> Instead, only autoscroll when the selection changes via the keyboard.

@raineorshine I don't see a simple way to support this (hard to see a way without using `useEffect`)

Removing `useEffect` is necessary to differentiate between whether it is selected from mouse or hover. Preferably you'd be able to scroll to a command when `onKeyDown` is called:

This requires you to run `ref.current?.scrollIntoView({ block: 'nearest' })` for the newly selected run.

ex. 
```
onSelectUp = () => { 
    const nextSelected = commands[i - 1];
    const refOfNextSelected = somehowGetRefOfNextSelected();
    refOfNextSelected.scrollTo()
}
```
The only way the above would work is if we stored an array of refs and move ref outside of `CommandItem` which seems too complex of a solution.

Another way would be to use a selector. Ex. give all command items an `id`, and use `querySelector('#idhere').scrollIntoView(...)`. I think this would probably be the most simple solution. I can add a ref to a parent container, and query from there.

The third method would be to have an extra state that you set on hover and onKeyDown, `isSelectedFromHoverLast`, so `if (!isSelectedFromHoverLast) scrollHere()`. Don't like this since it feels hacky.

@trevinhofmann @raineorshine Thoughts on this?

yangchristina:
@raineorshine this thread is getting too long, so I'll put a reference to my question here: https://github.com/cybersemics/em/pull/2928#discussion_r2139100532

I believe I've addressed all your comments, but it's possible I've missed some. I will do a final check again after the last issue (auto scroll) above is fixed.

raineorshine:
Thank you!

I am going to let Trevin support you on the remaining requests. I'd like to review once at the end.

raineorshine:
Thanks for the keen breakdown of the available options for scrolling the selected command item into view.

> The only way the above would work is if we stored an array of refs and move ref outside of `CommandItem` which seems too complex of a solution.

I sense that there is a way to abstract away the details of managing a map of refs, but I don't have one readily available and I agree it feels a bit complex.

> Another way would be to use a selector. Ex. give all command items an `id`, and use `querySelector('#idhere').scrollIntoView(...)`. I think this would probably be the most simple solution. I can add a ref to a parent container, and query from there.

This is relatively straightforward, so I'm not opposed to it. That said, you lose type safety by putting the source of truth in the DOM and reading it at runtime. 

> The third method would be to have an extra state that you set on hover and onKeyDown, `isSelectedFromHoverLast`, so `if (!isSelectedFromHoverLast) scrollHere()`. Don't like this since it feels hacky.

While this is a little awkward, it is the most React-oriented and is less fragile compared to using the DOM. How about we add a prop `selectedSource?: 'mouse' | 'keyboard' | 'search'`. It can be embedded in the `selectedCommand` to ensure that it is always updated when the selected command changes:

```ts
const [selectedCommand, setSelectedCommand] = useState<{ command: Command, source: 'mouse' | 'keyboard' | 'search' }>({ command: commands[0], source: 'search' })
```

Just make sure to only set `selectedSource` when the command is actually selected, otherwise non-selected commands will re-render unnecessarily.

```ts
selectedSource={selected ? selectedCommand.source} : undefined}
```

Or, if you want to more tightly couple the `selected` prop and the `selectedSource` prop, you could combine them. That would also ensure that one is not sent without the other.

src/components/CommandItem.tsx
raineorshine: I've reviewed all the components that use `CommandItem` and have a better sense of the architecture.

`CommandItem` originally began as a row in a table, but when the `GestureCheatsheet` was added it served double duty as a grid. Now it is used in the `CommandPalette`, `Help`, and `GestureCheatsheet`. `viewType: 'grid' | 'table'` determines the layout, while `isTable` determines the HTML structure, as you pointed out, with very little effect on the actual layout. As an example of the blurring of meaning between table and grid view, we even [override a `<tbody>` element](https://github.com/Wanderoooo/em/blob/00ec9feb33f9ff42e98789c48b84bf0191d060f7/src/components/CommandTableOnly.tsx#L40) to use `display: grid` at some point.

Given how `CommandItem` has been generalized for multiple contexts, I think it makes sense to remove `isTable` completely and use `<div>` instead of all the table elements. From my testing, the only visual difference I could see was that the padding is calculated slightly differently when the container is a `table`, but it is insignificant. Eliminating `isTable` and only having the `viewType` prop will makes the component's API much more intuitive.

The source of the architectural confusion clearly didn't originate with this PR. Thanks for taking the time to clean this up while we're here!

src/components/CommandItem.tsx
```diff
@@ -1,57 +1,73 @@
...
+  isTable?: boolean
+  /**
+   * Controls when the command description is displayed.
+   * - If true: Always show the description.
+   * - If false: On non-touch devices, always show the description. On touch devices, show only when the command is selected.
+   */
+  alwaysShowDescription?: boolean
+  onHover?: (command: Command) => void
+  customize?: boolean
+  shouldScrollSelectionIntoView?: boolean
```
raineorshine: Let's call this `shouldScrollSelectedIntoView` to match `selected`.

src/components/CommandPalette.tsx
```diff
@@ -515,12 +338,14 @@ const CommandPaletteWithTransition: FC = () => {
   const showCommandPalette = useSelector(state => state.showCommandPalette)
 
   // Commands need to be calculated even if the command palette is not shown because useFilteredCommands is responsible for updating gestureStore's possibleCommands which is needed to prevent haptics when there are no more possible commands. Otherwise, either haptics would continue to fire when there are no more possible commands, or would falsely fire when the current sequence is not a valid gesture but there are possible commands with additional swipes.
-  const [recentCommands, setRecentCommands] = useState(storageModel.get('recentCommands'))
   const [search, setSearch] = useState('')
-  const commands = useFilteredCommands(search, {
-    recentCommands,
-    sortActiveCommandsFirst: true,
-  })
```
raineorshine: I believe `useFilteredCommands` needs to stay here, per the comment above.

https://github.com/cybersemics/em/blob/00ec9feb33f9ff42e98789c48b84bf0191d060f7/src/components/CommandPalette.tsx#L517

src/hooks/useFilteredCommands.ts
```diff
@@ -33,108 +33,112 @@ const useFilteredCommands = (
...
+  const possibleCommandsSorted = useMemo(() => getPossibleCommandsSorted(search), [getPossibleCommandsSorted, search])
   // persist possible commands to gestureStore so that they can be accessed by the onGestureSegment in commands.ts
   useLayoutEffect(() => {
     if (isTouch) {
       gestureStore.update({ possibleCommands: possibleCommandsSorted })
     }
   }, [possibleCommandsSorted])
 
-  return possibleCommandsSorted
+  return { commands: possibleCommandsSorted, getCommandsSorted: getPossibleCommandsSorted }
```
raineorshine: I wonder if there is a better way to do this. The addition of `getCommandsSorted` to the return type in order to be injected into the `CommandSearch`'s `onInput` handler below is a bit complex and hard to follow. I worry about the increase in complexity.

Could you try `useLayoutEffect` to select the first command instead of `useEffect`? That may allow the component to re-render before the first paint.

yangchristina:
> Thanks for the keen breakdown of the available options for scrolling the selected command item into view.
> 
> > The only way the above would work is if we stored an array of refs and move ref outside of `CommandItem` which seems too complex of a solution.
> 
> I sense that there is a way to abstract away the details of managing a map of refs, but I don't have one readily available and I agree it feels a bit complex.
> 
> > Another way would be to use a selector. Ex. give all command items an `id`, and use `querySelector('#idhere').scrollIntoView(...)`. I think this would probably be the most simple solution. I can add a ref to a parent container, and query from there.
> 
> This is relatively straightforward, so I'm not opposed to it. That said, you lose type safety by putting the source of truth in the DOM and reading it at runtime.
> 
> > The third method would be to have an extra state that you set on hover and onKeyDown, `isSelectedFromHoverLast`, so `if (!isSelectedFromHoverLast) scrollHere()`. Don't like this since it feels hacky.
> 
> While this is a little awkward, it is the most React-oriented and is less fragile compared to using the DOM. Let me know if you find a counterexample, but I can't think of any reason this wouldn't work. How about we add a prop `selectedSource?: 'mouse' | 'keyboard' | 'search'`. It can be embedded in the `selectedCommand` to ensure that it is always updated when the selected command changes:
> 
> ```ts
> const [selectedCommand, setSelectedCommand] = useState<{ 
>   command: Command, 
>   source: 'mouse' | 'keyboard' | 'search' 
>   }>({ command: commands[0], source: 'search' })
> ```
> 
> Just make sure to only set `selectedSource` when the command is actually selected, otherwise non-selected commands will re-render unnecessarily.
> 
> ```ts
> selectedSource={selected ? selectedCommand.source} : undefined}
> ```
> 
> Or, if you want to more tightly couple the `selected` prop and the `selectedSource` prop, you could combine them. That would also ensure that one is not sent without the other.
> 
> Looking at this further, I see you already added `shouldScrollSelectionIntoView`. Do you think it's possible to reuse that instead of having a separate `selectedSource` prop?

Implemented method 3 here [f934e71](https://github.com/cybersemics/em/pull/2928/commits/f934e71d7d4501d707ff25a4316ccacabb0cdfd6)

You were right, after implementing it it actually works really nicely!

src/hooks/useFilteredCommands.ts
yangchristina: Done! This worked great! Really wish I thought of this before, glad I learned this from you, thanks!
[af2737e](https://github.com/cybersemics/em/pull/2928/commits/af2737e40dd9b8f820395f47d112d25543da9221)

yangchristina:
> > [21d9c16](https://github.com/cybersemics/em/pull/2928/commits/21d9c168ca544b42cec88b646f8eac136202bcaa)
> > Would it be better if I moved all autoscroll behaviour outside of CommandItem and passed a ref in?
> 
> The behavior looks good. I _think_ doing it that way is fine, though perhaps a closer look will reveal something I'm not considering. What is your sense?

Sounds good! It's just that only CommandPalette uses the scroll behaviour, and moving it out would reduce the number of props in CommandPalette. I think this way is fine too, since then you'd also have to pass a ref down from the parent in order to move this outside, which would add some complexity in that sense. I'm going to leave the implementation as is, let me know if you want me to change it.

yangchristina:
> Thanks so much and sorry for the delay! Only a few things remaining 🙏
> 
> One thing I noticed is that the background mask no longer extends to the edge of the screen:
> 
> Brave Browser 2025-06-21 15 17 39 Brave Browser 2025-06-21 15 17 52

Done thanks for catching! [7185af0](https://github.com/cybersemics/em/pull/2928/commits/7185af0dea27b45a6a14c591159a27f2fa5c7ab0)

src/components/CommandPalette.tsx
yangchristina: Undid these changes

yangchristina:
@raineorshine this is ready for re-review now!

src/components/CommandItem.tsx
raineorshine: Okay, I see now. It is possible to pull that logic up to the `CommandPalette` which has the benefit of removing the `parentElement` coupling eliminating the `isFirstCommand` and `isLastCommand` props. On the other hand, that splits the autoscroll logic into two places, and arguably there is still an implicit dependency on `parentElement` for `scrollIntoView({ block: 'nearest' })` to work.

I've opted to leave it as is.

See: dd6beede8f (unused)

raineorshine (approved):
Thank you!!!

When I was testing I found I still needed 560f523bdd to clear the search on close.

FYI I had to revert 42b4cc276f56788574571d173c265debbbd8111f because it created DOM validation errors in the JS console. See fac8168c31.

PR #2934, opened 2025-05-11
Divider is not highlighted when it has the cursor

### LINKED ISSUE #2898: Divider is not highlighted when it has the cursor
This is a regression. The first step should be to identify the bad commit. 

## Steps to Reproduce

Create a divider by typing `---`. 

## Current Behavior

The divider does not change with the cursor. 

## Expected Behavior

The divider should be highlighted when the cursor is on it. 

## Testing

Add a divider to one of the snapshot tests. Avoid creating a new test to control test running time.

karunkop:
Close #2898 

### Regression:

image

1. In the old version (519d1aab68), there was a class in App.css that highlighted the divider when editing

```ts
.editing > .thought-container > .thought > .divider-container > .divider {
  border-color: rgb(204, 204, 204);
}
.divider {
  border: solid 1px gray;
}
```
2. In the new version (6b1c0d8a7e80fe36304d8070c303cd99f5674b6b) or the current main branch, we have migrated it from traditional CSS classes to a styled-system approach with the css function.

## Fix:

Previously, the `borderColor` was controlled via CSS. With the transition to a styled-based approach, we now need to explicitly manage the `borderColor` dynamically based on the cursor's position. Specifically whether it is on the divider or not.


raineorshine (approved):
Hi! Thanks for identifying the bad commit. Good explanation.

Solution confirmed ✓

FYI There's no need to run the divider test at multiple font sizes. I moved it to the default font size only in 3225e6792b1.

PR #2941, opened 2025-05-12
[Mobile] Cancel gesture selected instead of New Thought 

### LINKED ISSUE #2938: [Mobile] Cancel gesture selected instead of New Thought
Mobile only. Reproducible in Chrome Device Emulator.

Regression: #2925

## Steps to Reproduce

Swipe →↓ and hold to open the Command Palette.

## Current Behavior

Image

## Expected Behavior

New Thought should be highlighted.

Cancel should only be highlighted if no command gestures can be constructed out of the current gesture in progress.

karunkop:
Close #2938

### Regression

The issue was introduced as a regression in commit 320829c13f1f87f301c61189505bc41474999887. In pull request #2925, we removed all command aliases and consolidated them into their primary command definitions. Previously, the `command.gesture `property was treated as a string, and gesture comparisons relied on string equality checks. However, following the change, `command.gesture` may now be an array representing multiple possible gesture directions.

This led to an invalid comparison logic, causing the `hasMatchingCommand` flag to always evaluate to false. As a result, the fallback condition for `isCancelMatch `was triggered unconditionally, leading to the cancel command being selected regardless of the actual gesture input.

### Fix

Implemented proper handling of array-based gesture comparisons using `gestureString`

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -162,7 +162,7 @@ const { commandKeyIndex, commandIdIndex, commandGestureIndex } = index()
 
 /** Gets the canonical gesture of the command as a string, ignoring aliases. Returns an empty string if the command does not have a gesture. */
 export const gestureString = (command: Command): string =>
-  (typeof command.gesture === 'string' ? command.gesture : command.gesture?.[0] || '') as string
+  (typeof command.gesture === 'string' ? command.gesture : command.gesture?.[0] || ' ') as string
 
 /** Get a command by its id. */
 export const commandById = (id: CommandId): Command => commandIdIndex[id]
--- a/src/components/CommandPalette.tsx
+++ b/src/components/CommandPalette.tsx
@@ -249,7 +249,7 @@ const CommandRow: FC<{
               color: disabled
                 ? 'gray'
                 : isTouch
-                  ? selected || gestureInProgress === command.gesture
+                  ? selected || (gestureInProgress as string).startsWith(gestureString(command))
                     ? 'vividHighlight'
                     : 'fg'
                   : 'fg',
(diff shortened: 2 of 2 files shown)


src/commands.ts
```diff
@@ -162,7 +162,7 @@ const { commandKeyIndex, commandIdIndex, commandGestureIndex } = index()
 
 /** Gets the canonical gesture of the command as a string, ignoring aliases. Returns an empty string if the command does not have a gesture. */
 export const gestureString = (command: Command): string =>
-  (typeof command.gesture === 'string' ? command.gesture : command.gesture?.[0] || '') as string
+  (typeof command.gesture === 'string' ? command.gesture : command.gesture?.[0] || ' ') as string
```
karunkop, line 165: The reason for changing `''` to `' '` is because js evaluates expressions like `'rd'.startsWith('')` as true. Since cancel command's gesture is undefined `gestureString` will evaluate it as a empty string `''`. Due to this the comparison in `src/components/CommandPalette.tsx` will evaluate true for both cancel and the matched command gesture at the same time and they are selected simultaneously

raineorshine: Thanks for the great explanation :). 

Let's either add this as a comment, or shift the responsibility to the CommandPalette. Otherwise this may not be obvious to other developers that look at this code.

raineorshine (changes requested):
This would seem to change the semantics from exact match to startsWith in multiple places. Are you sure this is right?

karunkop:
> This would seem to change the semantics from exact match to startsWith in multiple places. Are you sure this is right?

Yep you are right. I have changed it back to the exact match comparison while still utilising `gestureString` command.

raineorshine (approved):

Issue #2529, opened 2024-11-02
Animation: Toggle Context View


raineorshine:
## Steps to Reproduce

```
- a
  - m
    - x
- b
  - m
    - y
```

Toggle Context View on `a/m`.

## Current Behavior

No animation:

https://github.com/user-attachments/assets/b548213e-051a-49f8-a2fb-06d2166e1a9d

## Expected Behavior

- [ ] Animate each child as [disappearing text](https://codepen.io/raine/pen/ZEgRMwR?editors=1100).
- [ ] Normal view fades out from lower left.
- [ ] Context view fades in from upper right.
- [ ] When closing the context view, the animation is reversed.
- [ ] Smoothly switch directions when context view is rapidly toggled.

Demonstration:

https://github.com/user-attachments/assets/8d75f68e-01e2-4590-a642-88e992294737

zukilover:
@raineorshine it seems there's a bug in Safari when using the `skew` transition. It's also evident in the codepen example that will appear reverted in Safari. We could fix it by replacing `skew(-100deg)` to `skew(100deg)` but the result is expected, other browser will show reverted animation. Trying to replace it with `matrix` transform, but it seems that pandaCSS does not support transitioning the property. I'll find another way to fix it.

raineorshine:
Thanks! Feel free to use `isSafari` to flip the skew if that’s all it take to fix it.

zukilover:
PR is ready to review: https://github.com/cybersemics/em/pull/3034

PR #2944, opened 2025-05-13
Implement animations for context view

zukilover:
Animation: Toggle Context View #2529


zukilover:
@raineorshine I'm still working on getting the transition as close to the example as possible. I had to tweak values like the text-shadow blur and the translateY transition. I'll see if I can finish it in the next couple of days.

raineorshine:
Thanks for the update! Sounds good.

zukilover:
I've added `disappearingUpperRight` and `disappearingLowerLeft` transition styles to create directional text fade animations for the context view transitions.

Each animation pair follows a consistent design pattern:
The enter animation is a reverse preview of the exit animation.
It starts with a skewed, offset position and animates toward a neutral position `(skew(0), translateX(0), translateY(0))`.
This gives a subtle hint of where the element "came from" before settling into view.

zukilover:
<details>
<summary>Here's how it works</summary>

https://github.com/user-attachments/assets/c7318e2e-1f25-4022-95f7-e832aaf04cc6




</details>

zukilover:
<details>
<summary>This is how it works with longer text</summary>

https://github.com/user-attachments/assets/c2c7fc6e-a074-4f8b-8744-19fc010e344c




</details>

raineorshine:
Initial impression: Looks amazing!!!

I'll leave the initial review to @trevinhofmann. Thanks!

src/recipes/fadeTransition.ts
```diff
@@ -1,5 +1,18 @@
...
+    | 'fast'
+    | 'slow'
+    | 'medium'
+    | 'distractionFreeTyping'
+    | 'nodeFadeIn'
+    | 'nodeFadeOut'
+    | 'nodeDissolve'
+    | 'disappearingUpperRight'
+    | 'disappearingLowerLeft'
+}
```
trevinhofmann, line 14: This type declaration is confusing, because `FadeTransitionRecipeVariant` (also `SlideTransitionRecipeVariant`) is already imported and used prior to this PR:

https://github.com/cybersemics/em/blob/aa6016e018a758bf219ca152f7fbca9643611107/src/components/FadeTransition.tsx#L3

Is there a reason that this type needs to be explicitly defined and exported here? It seems that our Panda configuration already magically generates it. This is also the only explicitly exported type in `src/recipes`, so it doesn't seem to be following the project's usual patterns.

trevinhofmann (changes requested):
Hi, @zukilover! Thank you very much for the pull request, and my sincere apologies for the review delay.

The animation looks excellent, and I haven't found any regressions:

<details>
<summary>1x Speed</summary>

https://github.com/user-attachments/assets/222ec899-2c22-4b80-bcb2-2e64b4337ae8
</details>

<details>
<summary>0.1x Speed</summary>

https://github.com/user-attachments/assets/167bd017-4192-4727-9220-6d174858cf38
</details>

In the 0.1x speed, I noticed that the thought text does not animate in until well after the bullet animates in. The animation (especially at normal speed) looks fine to me, but I'll let @raineorshine decide if the thought text itself should appear at the same time as the bullet (which is probably ideal).

Otherwise, the only issue I noticed in the code review is the (redundant?) declaration of `FadeTransitionRecipeVariant` with a comment below.

raineorshine:
Yes, the thought + bullet + superscript should all animate in/out at the same time.

src/components/TreeNode.tsx
```diff
@@ -155,12 +165,36 @@ const TreeNode = ({
...
-      duration={isEmpty ? 'nodeFadeIn' : isLastActionDelete ? 'nodeDissolve' : 'nodeFadeOut'}
+      // Context view children get special disappearing text animations
+      duration={
+        isLastActionDelete
+          ? 'nodeDissolve'
+          : isEmpty
+            ? 'nodeFadeIn'
+            : isLastActionContextView
+              ? contextAnimation
+              : 'nodeFadeOut'
```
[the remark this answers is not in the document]
zukilover: I believe it works because `isEmpty` and `isLastActionDelete` are mutually exclusive. In any case, I've reordered them to maintain consistency.

src/recipes/fadeTransition.ts
zukilover: This was due to a temporary build error, and I thought explicitly exporting the type might help. But you're right, it was unnecessary. I've removed it now.

trevinhofmann (changes requested):
Thank you for the updates, @zukilover! The animation now looks correct:

<details>
<summary>Normal speed</summary>

https://github.com/user-attachments/assets/0103399a-6adf-4cf8-ba10-faa22014ac5f
</details>

<details>
<summary>0.1x speed</summary>

https://github.com/user-attachments/assets/827b0970-a4e0-4e72-81a1-f0b064dd313e
</details>

It looks like we just have some failing puppeteer tests to address.

I tested it locally, and these failures were somehow introduced by moving `isContextViewChild` and `isInContextView` in 0c93a4d287a7975cf85a5de7d05b0acadb58ad20:

```
src/e2e/puppeteer/__tests__/virtualization.ts > virtualizaton > virtualize thoughts that are not in the viewport
AssertionError: expected 0 to be greater than 0
```

zukilover:
Thanks, @trevinhofmann.
The previous change broke the test because it called the `useSelector` hooks **after** the conditional return here: https://github.com/cybersemics/em/blob/7f413c260d609fe9f847e53003348286dedde157/src/components/TreeNode.tsx#L150-L151

I've fixed that and also moved the `contextAnimation` logic closer to `isContextViewChild` and `isInContextView` to make it easier to follow.

trevinhofmann (approved):
Thanks, @zukilover! The animation is still looking great. I'll pass this over for @raineorshine's review.

raineorshine:
# Normal view descendants disappear from the wrong side

This is going to present a bit of a challenge, as we probably don't want `TreeNode` to subscribe to every ancestor.

## Steps to Reproduce

```
- a
  - b
    - c
    - d
- x
  - a
```

Toggle the Context View on `a`.

## Current Behavior

`c` and `d` incorrectly disappear to the right.


https://github.com/user-attachments/assets/38a54377-877c-4eea-97c1-c2465ed9abd3

## Expected Behavior

Thoughts in normal view should disappear to the left.
Thoughts in context view should disappear to the right.

raineorshine (commented):
# In/out animations are different

One thing I noticed is that the in/out animations do not look the same. I would expect them to be exactly reversed from each other.

## Steps to Reproduce

```
- a
  - m
    - x
- b
  - m
    - y
```

1. Toggle the Context View on `a/m`.
2. Toggle off.

https://github.com/user-attachments/assets/0103399a-6adf-4cf8-ba10-faa22014ac5f

## Current Behavior

When the contexts `a` and `b` animate in, they don't animate in from the full distance, but fade in from a closer distance with less skew:

Brave Browser 2025-05-31 11 27 42

Similarly, when `x` animates out, it doesn't animate to the full distance, but fades out from a closer distance and with less skew:

image

## Expected Behavior

Disappear in/out animations should be identical, but in reverse. This gives the impression that the thoughts are returning to where they came from. If the animations are not exact mirrors, then it looks as if they come in from one place, and then go somewhere else.

raineorshine (changes requested):
Thank you! This looks great.

Two cases to look at above.

zukilover:
In/out animations should now be in their exact animation sequences as well. 
This has been updated in 5601a531c7bef046059d35c5bdaf3c9acf475f22.

src/recipes/fadeTransition.ts
```diff
@@ -63,6 +63,48 @@ const fadeTransitionRecipe = defineSlotRecipe({
           transition: `opacity {durations.nodeDissolve} ease-out, transform {durations.nodeDissolve} ease-out, filter {durations.nodeDissolve} ease-out`,
         },
       },
+      // Context view fades in from upper right.
+      disappearingUpperRight: {
+        enter: {
+          transform: 'skew(-25deg) translateX(-5%) translateY(50%)',
+          textShadow: '0px 0px 2em white',
```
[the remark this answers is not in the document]
zukilover: Updated. Thanks for for pointing this out.

raineorshine:
I was imagining it would be used in combination with `isLastActionContextView`. If the last action is `toggleContextView`, then we know that all descendants of the cursor are either entering or exiting a context view, I think.

zukilover:
I simply tried this:
```
const isContextViewChild = useSelector((state: State): boolean => {
    return isDescendantPath(path, state.cursor)
})
```

but it resulted in incorrect animation:

<details>
<summary>result:</summary>

https://github.com/user-attachments/assets/de343a33-e551-4e1d-870f-a7f733800172



</details>


Here's what needs to be implemented to get the correct animation:

```
const isContextViewChild = useSelector((state: State): boolean => {
    if (!state.cursor) return false
    if (!isDescendantPath(path, state.cursor)) return false

    // Check if any ancestor (including self) has context view active
    for (let i = path.length; i > 0; i--) {
      const ancestor = path.slice(0, i) as Path
      if (isContextViewActive(state, ancestor)) return true
    }
    return false
  })
```

raineorshine:
It looks like the right animation, it just needs to vary its direction based on the context view state of the cursor. You don't need to check every ancestor for the context view. 

Please keep at it and let me know if you're able to figure it out. You're very close. Thanks!

zukilover:
Checking whether the thought is a descendant of the cursor turned out to be trickier, as it always returns `true` for both normal and context views. In other words, `isContextViewChild` is always `true` for all descendants of a thought, regardless of whether it's in context view.

To distinguish between an original thought and a contextual one, I compared the `path` and `simplePath` props. If they differ, the thought is being displayed in a contextual position.

I also fixed the `fadeTransition`, so `disappearingLowerLeft` and `disappearingUpperRight` now apply symmetrical animations.

Everything should work as expected now.

zukilover:
Thanks @raineorshine for the details. I've been working on refining the context view animation logic. I hope this gets us much closer to what you've explained.

> I think you're confusing the two conditions that need to be checked:
> When toggling the context view on, all old descendants of the cursor should disappear left. All new descendants of the cursor should appear right. In each case, it is _all_ descendants of the cursor that need to be animated.

I've updated this so that in TreeNode.tsx, the logic tries to address by:

Checking `isLastActionContextView` and `isLastContextViewOn` (derived from the new `contextViewToggledOn` state) to determine when these special context view animations should occur and what the direction of that toggle was.
https://github.com/cybersemics/em/blob/b83340fc2299d0cc18e11af2efa6fb520f19a78f/src/components/TreeNode.tsx#L96-L101

Using `isDescendantOfCursor` to see if the current thought node should participate in the descendant-specific animation.
https://github.com/cybersemics/em/blob/b83340fc2299d0cc18e11af2efa6fb520f19a78f/src/components/TreeNode.tsx#L104-L107

Then, for these descendants, if the context view was turned ON (`isLastContextViewOn === true`), it applies 'disappearingUpperRight` for new children appearing and 'disappearingLowerLeft` for old children disappearing. 


> This only determines if _any_ ancestor is in the context view. It will break if a nested context view is toggled. This needs to be relative to the `cursor`, the origin of the `toggleContextView` action.

This makes total sense. The new approach now completely drops that equalPath check for the animation logic when a context view toggle is the last action. Instead, it focuses on whether the thought is a descendant of the actual cursor that initiated the toggle (isDescendantOfCursor) and the outcome of that specific toggle (isLastContextViewOn).

I added the `isAppearing` flag (from transitionGroupsProps.in). When the context view is toggled ON at the cursor, for example, a thought that is a isDescendantOfCursor might be an old original child that needs to disappear (so `isAppearing` would be false), or it might be a new contextual child that's just appearing (`isAppearing` would be true). 
https://github.com/cybersemics/em/blob/b83340fc2299d0cc18e11af2efa6fb520f19a78f/src/components/TreeNode.tsx#L113-L114

I added `contextViewToggledOn` in the action/state. This one is necessary because each node needs to know this direction of the toggle to pick the correct animation rule. By having the `toggleContextView` action itself record this outcome in the state, it provides a clear and direct signal to all affected TreeNode components. I tried to keep the change in toggleContextView.ts as minimal as possible, mainly just adding that one piece of information.
https://github.com/cybersemics/em/blob/b83340fc2299d0cc18e11af2efa6fb520f19a78f/src/actions/toggleContextView.ts#L35-L39

Please let me know what you think, and if there are any other adjustments or a different approach you'd prefer.
Thanks.

zukilover:
> The use of `isDescendantPath` looks good. As for the direction of the animation, unless I am mistaken, you can just check `isLastActionContextView && isContextViewActive(state, state.cursor)`.

Thanks for the suggestion, you’re right. This is a cleaner way to handle the animation direction.
I originally added `contextViewToggledOn` to make the outcome clear. But I think I overcomplicated it, your way makes a lot more sense.

> After that, please review your solution with performance in mind. Several calculations are done more often than needed, and breaking logic up into separate `useSelector` calls is inefficient. For example, avoid calculating `isDependentPath` on every render, as calling it on all TreeNodes will result in O(n * depth) running time.

By `isDependentPath` I assumed it's `isDescendantPath`. I decided to use it in TreeNode to avoid prop drilling. 
As per your suggestion, I moved the expensive calculation into the LayoutTree. 
In TreeNode, I've combined the useSelector calls into a single hook with isEqual to make it more efficient.

Let me know what you think.

src/components/LayoutTree.tsx
```diff
@@ -263,27 +265,33 @@ const LayoutTree = () => {
...
-                setSize,
-                cliffPaddingStyle,
-                dragInProgress,
-                autofocusDepth,
-              }}
-            />
-          ))}
+          {treeThoughtsPositioned.map((thought, index) => {
+            // Calculate isCursorDescendant once here in the parent component loop
+            const isCursorDescendant = cursor ? isDescendantPath(thought.path, cursor) : false
```
raineorshine: Calculating `isCursorDescendant` here only really avoids calculating the `cursor` in each `TreeNode`, which is not significant.

The main problem is that `isCursorDescendant` is only ever used when `isLastActionContextView` is true. Therefore, why calculate it for every node? It's wasted work. Moving this into `TreeNode` will allow you to combine it with other selectors and only call it when necessary.

See https://github.com/cybersemics/em/pull/2944/files#diff-536e9718bba011a22eb6b77e300da89088b43d8b8898617f3a4a811e9add18da for a single `useSelector` that encapsulates all of the context view animation logic. It is more efficient because it only triggers a re-render when the final value changes. There are no intermediate `useSelectors` that trigger re-renders without getting used.

As a minor refactor, I also changed the type from a boolean to `'disappearingLowerLeft' | 'disappearingUpperRight' | null` so that I could completely contain `isLastActionContextView`. Notice how this is even more encapsulated. The rest of the `TreeNode` component doesn't really need to know the details of how the context view animation is determined, and now that this is in a single selector that could easily be factored out of the component if needed. It's almost like a custom hook at this point.

raineorshine (changes requested):
Thank you!

The only problem I'm seeing still is [in/out animations are different](https://github.com/cybersemics/em/pull/2944#pullrequestreview-2883856470). It seems to be better than before, but there is still a noticeable difference. i.e. the animation when `transitionGroupsProps.in` is true should be the exact reverse of when `transitionGroupProps` is false.

src/components/LayoutTree.tsx
zukilover: Thank you for the explanation and updated code. I still need to keep up with the expected standards in the future. I appreciate you taking the time to walk me through it.

zukilover:
> It's not perfect, but it's better. I will review and make a decision about whether it is acceptable. Do you have any explanation of why it's so hard to get an exact reverse animation?

I've tried using the same transform and easing but it didn't really produce a true reverse.  Easing curves aren't really symmetrical (I used `ease` for the in animation, tried to mirror the control points `(1 - x2), (1 - y2), (1 - x1), (1 - y1)`, didn't work). Also transforms apply in order, so even mirrored values don’t look right. I had to tweak a custom bezier to get the motion to actually feel reversed.

raineorshine (approved):
> I’ve just implemented the exact reversed easing using [this tool](https://codepen.io/michellebarker/pen/jQpwKq/), which mirrors the control points as mentioned above, to show you how it works.

Nice! The easing looks good now.

The end result is fairly different looking than the demo animation in the original issue, but will accept it.

Unfortunately the performance seems particularly bad with more thoughts, so this may not make it into production. Perhaps we can use this to consider more optimal alternatives in the future. Thank you for all your work!

PR #2948, opened 2025-05-14
Disable touch events on Editable

ethan-james:
References #2837 

For this proof-of-concept, I wanted to add `onTouchEnd={e => e.preventDefault()}` in order to block all editable selection. This works in https://codesandbox.io/p/sandbox/hj2lk7, but not in em.

I didn't trust the result, so I added `e.stopPropagation()` to make sure that the event wasn't being handled by an upstream component. That broke multiselect and caused it to be difficult to verify the behavior, so I added

```
      onTouchStart={e => {
        e.stopPropagation()
        e.preventDefault()
      }}
```

as well, disabling touch events entirely. As you can see in this branch, it behaves exactly the same as before. Blocking `touchend` entirely in em does not appear to stop the selection (or change the aberrant double tap behavior), even though it does in the CodeSandbox.

I took it a little farther and commented out `onMouseDown` and `onClick`, which causes it to call `selection.set` inside of `useEditMode`, which makes sense. If I comment that out, I see some annoying autoscroll behavior and selection becomes impossible.

I commented out the new `onTouchStart` and `onTouchEnd` handlers, and selection is still impossible.

I commented out `onFocus` and now single tap will select an editable, so we're back to default browser behavior I think.

I put back

```
      onTouchEnd={e => {
        e.stopPropagation()
        e.preventDefault()
      }}
      onTouchStart={e => {
        e.stopPropagation()
        e.preventDefault()
      }}
```

and it still does not block single tap selection, even with everything else commented out.

The main issue that I see is that touch events aren't preventing default the way we expect, but there's also something weird going on in the `onFocus` handler apparently. I don't guess that the two are related, so uncovering what's wrong with touch events is probably more important.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -609,6 +609,14 @@ const Editable = ({
       // must call onMouseDown on mobile since onTap cannot preventDefault
       // otherwise gestures and scrolling can trigger cursorBack (#1054)
       onMouseDown={onTap}
+      onTouchEnd={e => {
+        e.stopPropagation()
+        e.preventDefault()
+      }}
+      onTouchStart={e => {
+        e.stopPropagation()
+        e.preventDefault()
+      }}
       onFocus={onFocus}
       onBlur={onBlur}
       onChange={onChangeHandler}


raineorshine:
Thanks for the update! 

That’s too bad it's not blocking selection as expected. I agree we need to demonstrate the double tap feasibility in a stripped down **em** before attempting to integrate with the rest of the touch/focus/selection handling. 

There's `Editable`, `ContentEditable`, `useEditMode`, and `MultiGesture` that all interact with touch events or selection, off the top of my head. You can also use Chrome Developer Tools to see which event handlers are attached to the `ContentEditable` and its ancestors.

ethan-james:
It does in fact work if I disable `MultiGesture`! I haven't made it work with `onTouchEnd` yet because, in `useEditMode`,

```
    setTimeout(() => {
      disabledRef.current = false
    })
```

is firing unexpectedly early and allowing `shouldSetSelection` to be true. Hopefully I can work that out fairly quickly tomorrow.

ethan-james:
Here it is working `onTouchEnd`. There are still a number of issues to resolve:

1. re-enable `MultiGesture` without blocking `touchend` functionality
2. fix `disableTapRef` to restore long-press
3. fix `onTouchStart` to restore drag-and-drop
4. maybe move `allowDefaultSelection` into `onMouseDown` to remove 10ms delay in `setTimeout`

raineorshine:
`MultiGesture` is supposed to allow touch events to pass through, but `PanResponder` is a bit of a black box to me.

ethan-james:
> `MultiGesture` is supposed to allow touch events to pass through, but `PanResponder` is a bit of a black box to me.

Good news, then! I can eliminate `PanResponder` entirely and `MultiGesture` still stops `touchend` from preventing `click`. `touchend` occurs, but it can't block `click` right now, and it's not caused by `PanResponder`.

ethan-james:
OK, I think I restored all existing functionality. Let me know if you find anything broken.

I created #2949 and #2950 to track what I assume are regressions. I figured that I broke one or both in this branch, but I tested in `main` and saw the same behavior.

raineorshine (changes requested):
Thank you! This looks great.

I tested navigation mode, edit mode, long tap, and drop-drop. Nearly everything seems to be working as expected, although I did find one drag-and-drop regression.

Reproducible in Mobile Safari and Chrome Device Mode.

## Steps to Reproduce

```
- a
- b
- c
- d
- e
- f
- g
- h
```

1. Set the caret on `a`.
2. Scroll down by dragging in the scroll zone.

## Current Behavior

Drag-and-drop is incorrectly activated.

https://github.com/user-attachments/assets/f83b2bfa-47c7-4f47-89d2-fa27ba601d6b

## Expected Behavior

Drag-and-drop should not be activated when scrolling.

src/components/Editable.tsx
```diff
@@ -604,11 +576,14 @@ const Editable = ({
                 : value
       }
       placeholder={placeholder}
+      // We need to check if the user clicked the thought to not set the caret programmatically, because the caret will is set to the exact position of the tap by browser. See: #981.
+      onMouseDown={e => {
+        e.stopPropagation()
```
ethan-james, line 581: This is functionality that existed in `onTap` while it was a `mousedown` handler that was not maintained in the new `touchend` handler. It fixes the drag-while-scrolling issue. I couldn't see which `mousedown` behavior it's preventing. The bug doesn't appear to be caused by the `clickOnEmptySpace` behavior in `Content`. Maybe stopping propagation of `mousedown` is stopping events further along in the event chain. If I can figure out the other edit mode issue with time left over, maybe I can look more closely at this.

ethan-james:
For the edit mode bug, you can use https://xspx58.csb.app/ (source: https://codesandbox.io/p/sandbox/xspx58). You can see that tapping normally within the border or the box will produce `touchend`, `mousedown` and `focus`. If you tap on the very edge (outside the edge, really) of the black border, you can see that only `mousedown` and `focus` are triggered.

When `mousedown` was used to block `focus`, it would work anywhere because `mousedown` always fires. Since it's possible to avoid `touchend` by tapping in very specific outer areas, `focus` will fire unimpeded.

I'm not quite sure what to do about this, but I'll give it some more thought.

raineorshine:
How bizarre! Good work tracking it down.

I suppose we can detect this case by checking if `touchend` fired just before `mousedown`.

Thankfully we have a bit more leeway on the behavior for the left/right edge click. We can set the selection programmatically if we have to, since we don't have to worry about the offset. And preserving the double tap behavior is less important at the edges, so we could sacrifice that if we had to. However, we'll have to put some thought into how to turn this into a solution. It feels doable. Thanks for your work, and let me know what you come up with!

ethan-james:
Since a successful `touchend` event will call `setCursorOnThought`, I figure that we can discard `mousedown` events when the cursor is not on the thought. If `touchend` fires successfully, `mousedown` will not fire on the first tap and `editingOrOnCursor` will be true on the second tap. This will reduce the selectable area for a thought slightly, but it can always be expanded by increasing the padding. Let me know how you feel about this solution.

raineorshine:
> Since a successful `touchend` event will call `setCursorOnThought`, I figure that we can discard `mousedown` events when the cursor is not on the thought. If `touchend` fires successfully, `mousedown` will not fire on the first tap and `editingOrOnCursor` will be true on the second tap

It sounds plausible, though it's hard for me to predict the behavior given the complexity of interactions. Worth a try!

> This will reduce the selectable area for a thought slightly, but it can always be expanded by increasing the padding.

Can you explain this more? Does it just affect selecting text, or also tap? Which platforms? Increase padding on which side(s)? Thanks.

ethan-james:
> > This will reduce the selectable area for a thought slightly, but it can always be expanded by increasing the padding.
> 
> Can you explain this more? Does it just affect selecting text, or also tap? Which platforms? Increase padding on which side(s)? Thanks.

When I said selectable area, I in fact meant that it will reduce tappable area, because it will no longer attempt to process taps that are too far outside the bounds of the element to trigger `touchend`.

Selected text would be impacted by this fix in the same way as tapping. It shouldn't be possible to select any text while the cursor is on a different thought, and blocking `onMouseDown` would enforce that. Assuming that it's actually possible to begin a selection in the extremely narrow impacted area where `touchend` doesn't fire, which I haven't personally seen.

 I haven't been able to reproduce this on mobile Chrome, so my comments are restricted to iOS Safari.

Presumably you would only be interested in increasing padding on the right if you feel the tappable area has become too small. On the left, the bullet gets the tap. Mechanically, I don't think there's anything blocking this issue from occurring on the left but I've never seen it happen there. It's hard to imagine someone trying to tap between the bullet and the thought with the intention of giving the thought focus, whereas when you tap outside the thought's boundary on the right, the intention is indeed to give it focus without activating any other functionality.

raineorshine:
Got it, thanks!

The 1em of extended tap area to the right of the thought text would need to be preserved one way or another.

> Presumably you would only be interested in increasing padding on the right if you feel the tappable area has become too small. On the left, the bullet gets the tap. Mechanically, I don't think there's anything blocking this issue from occurring on the left but I've never seen it happen there. It's hard to imagine someone trying to tap between the bullet and the thought with the intention of giving the thought focus, whereas when you tap outside the thought's boundary on the right, the intention is indeed to give it focus without activating any other functionality.

As long as the double tap issue does not occur on the thought text, that will be adequate. Double tap (or selection) on the thought's padding or edge is less relevant. 

(As an aside, single tap on the thought's padding or edge is highly relevant though, and we have worked to get the precise tap/selection behavior, down to the pixel, including a combination of [rendering overlap](https://github.com/cybersemics/em/blob/1aa474a7c006ea9d51cfed06fc27af4bc8cc322d/src/hooks/useSizeTracking.ts#L51-L54) and [`clip-path`](https://github.com/cybersemics/em/blob/1aa474a7c006ea9d51cfed06fc27af4bc8cc322d/src/recipes/editable.ts#L18-L19).)

raineorshine (approved):
Tested, and everything seems to be working as expected! Thanks for tackling this tricky problem.

Not only does fix the double tap issue, but we eliminated some complexity by removing `disableTapRef`.

Issue #2902, opened 2025-04-20
=children/=note/=path


raineorshine:
The purpose of this feature is to allow the user to use a custom subthought as the note for each child.

- [ ] Both `=note` and `=children/=note` should be supported. 
- [ ] Editing a note should edit the respective `=path` thought, and vice versa.
- [ ] No note should be rendered if the `=path` though is missing or invalid. 
- [ ] Adding a missing `=path` thought should be done by the normal process of activating `toggleNote` and editing the note. 
- [ ] Deleting a note should delete the `=path` thought. 
- [ ] Cursor and caret should behave as expected.  

# 1. Defining all childre's notes with a custom path

- [ ] Depends on #2900

## Steps to Reproduce

```
- Films
  - =children
    - =note
      - =path
        - Year
  - Anora
    - Year
      - 2024
  - Oppenheimer
    - Year
      - 2023
  - Everything Everywhere All at Once
    - Year
      - 2022
```

## Expected Behavior

In each child, render the value of the thought found at `=path`. The path should be relative to the _first non-attribute ancestor of =path_, i.e. Films. 

![image](https://github.com/user-attachments/assets/cccf2d60-1cc5-4386-b4e5-a93ededbe0d0)

# 3. Adding a missing note

## Steps to Reproduce

```
- Films
  - =children
    - =note
      - =path
        - Year
  - Anora
    - Year
      - 2024
  - Oppenheimer
  - Everything Everywhere All at Once
    - Year
      - 2022
```

![image](https://github.com/user-attachments/assets/c59e2639-3276-4c48-9e62-ef21f28179cd)

1. Set the cursor on `Oppenheimer`. 
2. Activate `toggleNote`. 
3. Edit the note to `2023`. 

## Expected Behavior

```
- Films
  - =children
    - =note
      - =path
        - Year
  - Anora
    - Year
      - 2024
  - Oppenheimer
    - Year
      - 2023
  - Everything Everywhere All at Once
    - Year
      - 2022
``` 

![image](https://github.com/user-attachments/assets/cccf2d60-1cc5-4386-b4e5-a93ededbe0d0)

# 4. Deleting a note

## Steps to Reproduce

```
- Films
  - =children
    - =note
      - =path
        - Year
  - Anora
    - Year
      - 2024
  - Oppenheimer
    - Year
      - 2023
  - Everything Everywhere All at Once
    - Year
      - 2022
```

Delete `2023`.

## Expected Behavior

```
- Films
  - =children
    - =note
[...]

PR #2951, opened 2025-05-16
=note/=path

### LINKED ISSUE #2901: =note/=path
- [ ] Depends on #2800

The purpose of this feature is to allow the user to use a custom subthought as the note for each child.

Both `=note` and `=children/=note` should be supported. 

#1. =children/=note

## Steps to Reproduce

```
- Films
  - =children
    - =path
      - ../Year
  - Anora
    - Year
      - 2024
  - Oppenheimer
    - Year
      - 2023
  - Everything Everywhere All at Once
    - Year
      - 2022
```

## Expected Behavior

In each child, render the value of the thought found at `=path`. 

image

karunkop:
Close #2901 

Tasks:

- [x] Editing a note should edit the respective =path thought, and vice versa.

- [x] No note should be rendered if the =path though is missing or invalid.

- [x] Adding a missing =path thought should be done by the normal process of activating toggleNote and editing the note.

- [x] Deleting a note should delete the =path thought.

- [x] Cursor and caret should behave as expected.

Created a centralized `resolvePathReference `utility that efficiently resolves path references (=path). This PR will now allow us to use a custom subthought as a note


src/actions/archiveThought.ts
```diff
@@ -54,13 +55,34 @@ export const pathAndRankToArchive = (
 
 /** Moves the thought to =archive. If the thought is already in =archive, permanently deletes it.
  *
- * @param path     Defaults to cursor.
+ * @param path                 Defaults to cursor.
+ * @param archivePathReference If true and the path points to a note with =path, also archives the target thought.
  */
-const archiveThought = (state: State, options: { path?: Path }): State => {
+const archiveThought = (state: State, options: { path?: Path; archivePathReference?: boolean }): State => {
```
raineorshine, line 61: Adding the note logic to `archiveThought` has the same problem described above: You are making a low-level, generic function responsible for case-specific behavior. Previously, `archiveThought` did not need to know anything about notes, which was a good thing. 

Furthermore, `archivePathReference` is not orthogonal with `path`; there are preconditions, tightly coupling this to the caller in ways that are a bit difficult to infer. So it's not the best parameter design.

src/actions/toggleNote.ts
```diff
@@ -1,37 +1,102 @@
...
-    !hasNote
-      ? setDescendant({
-          path: state.cursor!,
-          values: ['=note', ''],
+  // Check if there's a path reference
+  const pathReference = resolvePathReference(state, thoughtId, hasNote)
+
+  // Calculate offset and handle target thought creation if needed
+  let offset = 0
+  let newState = state
```
raineorshine, line 27: Please do not use mutation for control flow. `reducerFlow` is a better approach to composing reducers.

src/actions/toggleNote.ts
```diff
@@ -1,37 +1,102 @@
...
+        // If target already has children, set offset to the length of the first child's value
+        offset = targetChildren[0].value.length
+      }
+    }
+  } else if (hasNote) {
+    const noteChildren = getChildren(newState, hasNote)
+    offset = noteChildren.length ? noteChildren[0].value.length : 0
+  }
+
+  const finalState = reducerFlow([
```
raineorshine, line 80: There should not be multiple state variables (`state`, `newState`, `finalState`). Take advantage of `reducerFlow` to compose chains of reducers. There are plenty of examples of this, including nested reducers, throughout the codebase.

raineorshine (changes requested):

karunkop:
@raineorshine I have made all the requested changes. Let me know what you think about this. Thanks!

src/actions/archivePathReferencedThought.ts
```diff
@@ -0,0 +1,43 @@
...
+import Path from '../@types/Path'
+import State from '../@types/State'
+import Thunk from '../@types/Thunk'
+import { registerActionMetadata } from '../util/actionMetadata.registry'
+import reducerFlow from '../util/reducerFlow'
+import { getPathReferencePath } from './../selectors/resolvePathReference'
+import archiveThought from './archiveThought'
+
+/** Archives a thought that is referenced by a =path. */
+const archivePathReferenceThought = (state: State, options: { path?: Path; pathNote?: Path }): State => {
```
raineorshine, line 11: I'm afraid my initial intuition about this was correct: Deleting the `=path` along with the Note is a bad idea. This is creating a lot of extra complexity for no real benefit. We need to keep in mind that the end goal is `=children/=path` functionality. The purpose of splitting this into separate issues was to provide an incremental on-ramp to the full functionality. Building exceptions which will need further exception in the next iteration is not going to serve us well.

I only have limited foresight architecting from afar. The real insight occurs when you're in the code and seeing how the structure is emerging. To this end, I would have expected you to notice this red flag and bring it up with me before going too deep into the implementation.

src/selectors/resolvePathReference.ts
raineorshine, line 1: We no longer have the clunky composite return value of a single function, which is good. However, I do not find the current aggregate of helper functions to be particularly elegant. In my view, it only adds to the surface area of the internal API without hiding any real complexity.

src/actions/toggleNote.ts
```diff
@@ -14,23 +20,84 @@ import reducerFlow from '../util/reducerFlow'
...
+              value: pathValue,
+              rank: pathRank,
+            })
+          }
+          return state
+        }
+      : null,
+
+    // Create an empty child for target if needed
+    hasPathReference
```
raineorshine, line 48: Which approach is better?

```js
if (A) {
  f()
}
if (A) {
  g()
}
```

or:

```js
if (A) {
  f()
  g()
}
```

You're doing the same thing with `hasPathReference`.

raineorshine:
Thank you for your work. I've decided to pursue a new implementation rather than continue to iterate on this one.

src/commands/note.ts
```diff
@@ -33,7 +33,7 @@ const noteCommand: Command = {
   isActive: state => {
     const { cursor } = state
     const path = cursor ? simplifyPath(state, cursor) : HOME_PATH
-    return attribute(state, head(path), '=note') !== null
+    return hasNotePath(state, head(path))
```
raineorshine, line 36: The note command should be active when the thought has an actual note. Just having a note path is not enough. So it should be:

```ts
return noteValue(state, head(path)) !== null
```

src/commands/archive.ts
```diff
@@ -20,6 +25,18 @@ import isRoot from '../util/isRoot'
 
 let undoArchiveTimer: number
 
+/** Archives a note path and its optional target path if =note/=path structure exists. */
+const archiveNoteAndTargetThought =
```
raineorshine, line 29: No need for this to be a standalone function since it is so small and is only used in one place.

Factoring it out into a function actually gets in the way of eliminating some duplication. Once you inline it, look for opportunities to simplify the conditional logic.

src/actions/toggleNote.ts
raineorshine, line 1: This is okay, but I think the organization could be improved. The `=path` is detected in the outer conditional while the toggling logic is handled in the inner conditional. 

In pseudocode:

```
if =path
  if empty
  if non-empty
else
  if empty
  if non-empty
```

However, since the purpose of this action is to toggle the note, I suspect it will read better to structure it inversely: With the toggling logic in the outer conditional, relegating the `=path` specifics to the inner conditional:

```
if empty
  if =path
  else
else
  if =path
  else
```

Conceptually, toggling a note consists of detecting whether the note exists, if it's empty, and then determining from there if the `=note` or target path needs to be created. Deleting `=note` or deleting the target path should seen as the same basic action and grouped together. Creating `=note` if it doesn't exist or creating the target path if it doesn't exist should be seen as the same basic action and grouped together.

src/actions/toggleNote.ts
```diff
@@ -1,35 +1,57 @@
...
+                thoughtId: targetThought.id,
+              })
+          : // create a missing target thought if it doesn't exist via =path
+            state => setDescendant(state, { path: state.cursor!, values: [targetThought?.value, ''] })
         : null,
 
-    // toggle state.noteFocus, which will trigger the Editable and Note to re-render and set the selection appropriately
+    // if targetThought is a noteAttribute and the note content is empty and noteFocus is true, delete the note attribute
+    state.noteFocus && isNoteAttribute && isNoteContentEmpty
+      ? deleteAttribute({ path: state.cursor!, value: '=note' })
```
raineorshine, line 51: This `deleteAttribute` is far apart from `deleteThought` above, even though they represent the same delete action when toggling a note off. Structuring the logic so these are grouped together would probably make more sense.

src/components/Note.tsx
```diff
@@ -81,7 +84,20 @@ const Note = React.memo(
...
+          dispatch((dispatch, getState) => {
+            const state = getState()
+            const targetThoughtPath = resolveNotePath(state, path)
+            const targetThought = targetThoughtPath ? getThoughtById(state, head(targetThoughtPath)) : undefined
+
+            if (targetThought && targetThought.value !== '=note') {
+              dispatch(deleteThought({ pathParent: path, thoughtId: targetThought.id }))
+            } else {
+              dispatch(deleteAttribute({ path, value: '=note' }))
+            }
```
raineorshine, line 98: Look for a way to combine these cases. There shouldn't be anything fundamentally different about a regular note and a note determined by a target path. Maybe something like this?

```ts
const targetThoughtPath = resolveNotePath(state, path) ?? path
dispatch(deleteThought({ pathParent: path, thoughtId: head(targetThoughtPath) }))
```

This would be even simpler if `resoloveNotePath` didn't return `null` for the plain `=note` case.

src/util/noteValue.ts
```diff
@@ -1,16 +1,18 @@
...
 
 /** Gets the value of a thought's note. Returns null if the thought does not have a note. */
 const noteValue = (state: State, id: ThoughtId) => {
-  const noteId = findDescendant(state, id, '=note')
-  if (!noteId) return null
-  const noteThought = getThoughtById(state, noteId)
-  if (noteThought?.pending) return null
-  return firstVisibleChild(state, noteId!)?.value ?? null
+  // Try to resolve path (checks =children first, then regular =note/=path)
+  const targetPath = resolveNotePath(state, thoughtToPath(state, id))
```
raineorshine, line 11: `thoughtToPath` doesn't work with nested context views. You'll have to modify this function to take a `Path` instead of a `ThoughtId`.

src/util/noteValue.ts
```diff
@@ -1,16 +1,18 @@
...
+import resolveNotePath from '../selectors/resolveNotePath'
+import thoughtToPath from '../selectors/thoughtToPath'
+import head from './head'
 
 /** Gets the value of a thought's note. Returns null if the thought does not have a note. */
 const noteValue = (state: State, id: ThoughtId) => {
-  const noteId = findDescendant(state, id, '=note')
-  if (!noteId) return null
-  const noteThought = getThoughtById(state, noteId)
-  if (noteThought?.pending) return null
```
raineorshine, line 12: `pending` logic needs to be preserved.

src/selectors/resolveNotePath.ts
```diff
@@ -0,0 +1,99 @@
...
+  const childrenNoteId = findChildrenNote(state, thoughtId)
+  if (childrenNoteId) {
+    return resolveNoteHelper(state, thoughtId, childrenNoteId)
+  }
+
+  // 2. Check for thought's own =note
+  const noteId = findDescendant(state, thoughtId, '=note')
+  if (!noteId) return null
+
+  return resolveNoteHelper(state, thoughtId, noteId)
```
raineorshine, line 96: Try to eliminate the duplication of `resolveNoteHelper(state, thoughtId, X)`.

There's no need to factor out `resolveNoteHelper`. All of that logic should be in one function, with a simple conditional at the beginning to determine the node id.

src/selectors/resolveNotePath.ts
raineorshine, line 1: It seems that every combination of `=children` and `=path` is treated as a separate case, which is too flat. Ideally, `=path` would not have to know about `=children`. Look for a hierarchy of abstractions that allow you to encapsulate one level of logic from another.

raineorshine (changes requested):

src/components/__tests__/Note.ts
```diff
@@ -108,3 +109,308 @@ it('render note when subthought is edited from non-note attribute', async () =>
...
+      })
+    })
+    await act(vi.runOnlyPendingTimersAsync)
+
+    // verify the archived content exists inside =archive
+    expect(screen.getAllByText('a')).toHaveLength(2)
+  })
+})
+
+describe('Children Notes', () => {
```
raineorshine, line 306: These tests look great, thanks!

src/components/__tests__/Note.ts
```diff
@@ -108,3 +109,308 @@ it('render note when subthought is edited from non-note attribute', async () =>
...
+    await act(vi.runOnlyPendingTimersAsync)
+
+    // Verify both the note and original thought now show the updated text
+    const updatedElements = screen.getAllByText('2025')
+    expect(updatedElements).toHaveLength(2)
+
+    // Verify the original year is no longer present
+    expect(screen.queryByText('2009')).toBeNull()
+  })
+})
```
raineorshine, line 416: Let's add three more tests that cover `=children/=path/=note` functionality:

1. Render only the existing notes (i.e. an empty note is not rendered).

```ts
      importText({
        text: `
        - x
          - =children
            - =note
              - =path
                - Year
          - a
            - Year
              - 2009
          - b
          - c
           - year
              - 2011`,
      }),
```
2. Adding a missing note ("2. Adding a missing note" in #2902)
3. Deleting a note ("3. Deleting a note" in #2902).

src/components/__tests__/Note.ts
raineorshine, line 1: Tests look good overall!

Keep in mind the RTL functions designed to select by aria label: `findByLabelText`, `findAllByLabelText`, `queryByLabelText`, `queryAllByLabelText`. We have `aria-label='note'` that might be better suited than `findAllByText` for the tests that assert whether a note has been rendered or not.

raineorshine (changes requested):

src/selectors/resolveNotePath.ts
```diff
@@ -0,0 +1,56 @@
...
+  pathValue: string,
+): ThoughtId | null => {
+  // Find child matching path value
+  const targetThought = findAnyChild(state, thoughtId, child => child.value === pathValue)
+  if (targetThought) return targetThought.id
+
+  // If no matching child, use first child of =path attribute
+  const pathChildren = anyChild(state, findDescendant(state, noteId, '=path'))
+  return pathChildren?.id || null
+}
```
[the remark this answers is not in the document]
karunkop: Your explanation about the lack of meaningful encapsulation in `resolvePathThought` makes complete sense. The logic has now been simplified and moved directly into the caller, which improves clarity.

 Previously, using the `attribute` selector to locate the descendant for `=path` led to redundant calls to `anyChild` to resolve its id. This change removes that overlap and better separates responsibilities.

The updated comments also clarify intent, improving the overall readability and maintainability of the code.

src/selectors/resolveNotePath.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (noteThought?.pending) return null
+
+  const pathId = findDescendant(state, noteId, '=path')
+  const pathChild = pathId && findAnyChild(state, pathId, isVisible(state))
+  // Resolves to a thought that is a child of `thoughtId` if the `=path` contains a descendant thought with the same value
+  const targetThought = pathChild && findAnyChild(state, thoughtId, child => child.value === pathChild.value)
+
+  // Determines the appropriate target thought to use:
+  // - Uses `targetThought` if available
+  // - Falls back to `pathChild` for resolving or creating a missing thought via toggleNote()
```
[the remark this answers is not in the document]
raineorshine: I created an experimental branch in which I normalized `resolveNotePath` so that it no longer creates invalid paths. It's a nontrivial change. Unfortunately this created some duplication that I didn't have time to factor out. Feel free to use this for inspiration: https://github.com/cybersemics/em/commit/da369e7c99847e66816c47b92782993db6a3fc5f

src/components/__tests__/Note.ts
```diff
@@ -1,110 +1,530 @@
...
+    // Verify note content
+    const element = screen.getByText('foo')
+    expect(element)
+  })
 
-it('re-render note when =note subthought value changes', async () => {
-  await dispatch([
-    importText({
-      text: `
+  test('re-render note when =note subthought value changes', async () => {
```
raineorshine, line 38: This test should be removed, per https://github.com/cybersemics/em/pull/2951#discussion_r2106224289.

src/components/__tests__/Note.ts
```diff
@@ -1,110 +1,530 @@
...
+      }),
+      setCursor(['a', 'note']),
+      (dispatch, getState) =>
+        dispatch(
+          editThought({
+            oldValue: 'note',
+            newValue: '=note',
+            path: getState().cursor as SimplePath,
+          }),
+        ),
```
raineorshine, line 81: We should avoid dispatching to the store directly, and instead use RTL to interact with the screen like the user. This is the main philosophy behind RTL. We try not to tightly couple tests to underlying infrastructure or implementation details.

Maybe we can use `focus` and `fireEvent` for editing?

FYI We currently dispatch `importText` directly to the store at the beginning of tests for simplicity, but [soon this will be refactored](https://github.com/cybersemics/em/issues/2980) to use `paste`, which is more user-centric and hides the global store access.

src/selectors/resolveNotePath.ts
```diff
@@ -0,0 +1,45 @@
...
+import { anyChild } from './getChildren'
+import getThoughtById from './getThoughtById'
+
+interface NoteKeyResult {
+  noteKey: string
+  noteId?: ThoughtId
+}
+
+/** Resolves note key and note id by checking for note thoughts. */
+export const resolveNoteKey = (state: State, thoughtId: ThoughtId): NoteKeyResult => {
```
raineorshine: While the co-location is nice, I want to stick to one default export per file as module as possible since that is a convention on the project.

See: d7fc8936044

src/selectors/resolveNotePath.ts
raineorshine, line 1: Thanks! I think this is workaround now. The composite return type is still a bit awkward, but the overall complexity is lower and the functions are consistent and meaningful.

src/selectors/resolveNotePath.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const noteThought = getThoughtById(state, noteId)
+
+  if (noteThought?.pending) return { noteKey: '=note' }
+
+  const notePathId = findDescendant(state, noteId, '=path')
+  const noteKey = anyChild(state, notePathId)?.value ?? '=note'
+  return {
+    noteKey,
+    ...(notePathId ? {} : { noteId }),
```
raineorshine: This special case is regrettable, and doesn't totally fit the semantics of the function, but I can't think of an alternative at the moment. We can keep it for the sake of getting this merged 👍

raineorshine (changes requested):

Issue #1915, opened 2024-05-17
Bullets shift position at other font sizes


raineorshine:
## Steps to Reproduce

```
- A
  - B
    - C
      - D
```

1. Set the font size to 13. 
2. Move the cursor between `B` and `C`. 

## Current Behavior

Bullets shift position during the translateX animation. 

https://github.com/cybersemics/em/assets/750276/f23a9bee-a9fb-4c70-810b-ca4d50fa4157

## Expected Behavior

Bullets should not shift position.

## Notes

Occurs at any font size other than the default 18. Appears to be caused by some combination of translateX, the `Scale` component, and the bullet SVGs. 

Bullets were converted to SVG in order to solve cross-platform positioning issues (#1175). A robust proposal should be presented that takes account of the lessons learned in #1402 and #1505 and suggests a fix that does not cause a regression.

trevinhofmann:
This appears to affect Safari (both Mac and iOS), but not Firefox or Chrome. Still looking into it.

PR #2960, opened 2025-05-19
Disable Editables while dragInProgress

### LINKED ISSUE #2949: [iOS] Drag-and-drop highlights text

## Steps to Reproduce

```
- Apple
- Birthday
```

1. Long press Apple.
2. Drag down below Birthday. 

## Current Behavior

During drag, Apple is highlighted.

## Expected Behavior

ethan-james:
Fixes #2949

While a drag is in progress, the underlying editables don't need to be selectable, so it's best to just disable them until the drag ends.


raineorshine (changes requested):
Thanks! I can confirm it fixes the issue.

I've re-run the puppeteer tests twice and the same test is failing, so it looks like it's not a flaky test. Could you investigate? Thanks!

ethan-james:
I see it failing locally as well. I expected it to be highlight-related, but it appears that the text in the snapshot is a pixel or two higher than the new screenshot. Assuming that I am correct and the old is on the left, and the new is on the right. I think it looks better on the right, although I don't have an answer for why it would have changed. I tested drag-and-drop in this branch and in `main` and don't see any noticeable difference in alignment.

drag-should-show-alert-and-quick-drop-panels-when-long-pressing-after-drag-operation-1-diff

raineorshine:
Thanks! It seems that either `StaticThought` or `ContentEditable` shift slightly when they re-render when `dragInProgress` changes. 

The important thing is that this shift isn't noticeable when drag-and-drop begins. If you could record a video that we can view frame by frame that would be helpful. The static snapshot diff isn't quite enough to confirm.

ethan-james:
Here is a video in the branch:

https://github.com/user-attachments/assets/b0ccb79c-d208-40d1-93b2-2571cbe11a72

and here's one of `main`:

https://github.com/user-attachments/assets/845eb378-853d-4059-befc-50918e738033

I don't see any difference in positioning. They both look pretty much the same as the right-hand image in the snapshot diff, which is to say, pretty good to my eye.

raineorshine (changes requested):
Thanks!

I was able to reproduce this manually. Given that the shift is noticeable, I think we'll have to find a way to resolve it before merging.

Interesting, the thoughts only shift on the first drag-and-drop. Refreshing the app is a necessary step to reproduce. The second drag-and-drop looks fine.

## Steps to Reproduce

```
- a
- b
- c
- d
- e
- f
- g
```

1. Refresh the app.
2. Initiate drag-and-drop on any thought.

## Current Behavior

Thoughts shift position when drag-and-drop starts.

https://github.com/user-attachments/assets/1b02d3b5-13b4-4e94-a80d-634f57d2bbad

## Expected Behavior

Thoughts should not move when drag-and-drop starts.

ethan-james:
A rare Chrome bug. I'll see what I can do.

raineorshine:
Reminds me a bit of #1915, but that was Safari-only, and this is Chrome only.

Thanks for investigating!

ethan-james:
I went through to see if I could fix it properly, but the only CSS properties that change when you switch `contenteditable` off are `line-break: after-white-space` and `overflow-wrap: break-word`. Neither of those are set manually, and I don't think that trying to modify them would pay off.

I tested with `pointer-events: none` instead of disabling `contenteditable` and that fixed the layout shift, but then seems to allow you to highlight parts other parts of the app such as the scroll area in iOS Safari. Not quite what I expected.

Probably the simplest fix is the `isTouch && isSafari()` trick to target only mobile iOS when disabling `contenteditable`. Since the layout looks OK on iPhone, that should prevent it from messing up Chrome.

raineorshine:
> I went through to see if I could fix it properly, but the only CSS properties that change when you switch `contenteditable` off are `line-break: after-white-space` and `overflow-wrap: break-word`. Neither of those are set manually, and I don't think that trying to modify them would pay off.

Yeah, I don't suspect so either. 

> I tested with `pointer-events: none` instead of disabling `contenteditable` and that fixed the layout shift, but then seems to allow you to highlight parts other parts of the app such as the scroll area in iOS Safari. Not quite what I expected.

This could be a solution if selection of other parts of the screen can be prevented. We may need to fix it anyway since it can be triggered by initiating drag-and-drop further to the right of the thought text. 

> Probably the simplest fix is the `isTouch && isSafari()` trick to target only mobile iOS when disabling `contenteditable`. Since the layout looks OK on iPhone, that should prevent it from messing up Chrome.

Yes, possibly. We would need to verify that the original issue does not occur on Android.

ethan-james:
> > I tested with `pointer-events: none` instead of disabling `contenteditable` and that fixed the layout shift, but then seems to allow you to highlight parts other parts of the app such as the scroll area in iOS Safari. Not quite what I expected.
> 
> This could be a solution if selection of other parts of the screen can be prevented. We may need to fix it anyway since it can be triggered by initiating drag-and-drop further to the right of the thought text.

I'll double check this, probably tomorrow.

> > Probably the simplest fix is the `isTouch && isSafari()` trick to target only mobile iOS when disabling `contenteditable`. Since the layout looks OK on iPhone, that should prevent it from messing up Chrome.
> 
> Yes, possibly. We would need to verify that the original issue does not occur on Android.

I tested on Android and did not notice the issue. I'll put together a more authoritative video, also probably tomorrow.

ethan-james:
Here it is on Android, looking good as far as I can tell.

https://github.com/user-attachments/assets/5dff4511-c44b-4c98-9ac7-77f8d7f2365e

> This could be a solution if selection of other parts of the screen can be prevented. We may need to fix it anyway since it can be triggered by initiating drag-and-drop further to the right of the thought text.

Oh, I see what you mean. Here I am dragging from around the scroll bar area:

https://github.com/user-attachments/assets/4d4dc184-1952-4462-ad04-846bd221c228

And then I created https://github.com/cybersemics/em/issues/2961 when I discovered the same thing while long-pressing on the bullet.

src/components/StaticThought.tsx
```diff
@@ -136,6 +137,7 @@ const StaticThought = ({
     _.isEqual,
   )
 
+  const dragInProgress = useSelector(state => isTouch && isSafari() && state.dragInProgress)
```
raineorshine: There are other places in the codebase where we define `dragInProgress` as simply `useSelector(state => state.dragInProgress)`, so we should probably call this one `dragInProgressSafari` to avoid confusion.

raineorshine (approved):
> Here it is on Android, looking good as far as I can tell.

Thank you! Looks good.

> And then I created #2961 when I discovered the same thing while long-pressing on the bullet.

Thanks so much 🙏

---

I also added an inline comment so the unique circumstance is well documented in the code.

PR #2962, opened 2025-05-21
LayoutTree: userSelect: none, reset userSelect in editables

### LINKED ISSUE #2961: [iOS] Drag-and-drop on bullet highlights bullet, scroll area, etc.

This is similar to what I was seeing in https://github.com/cybersemics/em/pull/2960 while testing `pointer-events: none` on editables. This, however, relates to the bullet rather than the editable, and it happens in `main`.

## Steps to Reproduce

```
- Apple
- Birthday
```

1. Long press bullet left of Apple.
2. Drag down below Birthday. 

## Current Behavior

During drag, especially if you drag far down the screen, various background elements get highlighted.

https://github.com/user-attachments/assets/1ba11f6c-1c87-4992-a8b7-ac1e77df277e

## Expected Behavior

ethan-james:
Fixes #2961

It turns out that setting `user-select: none` on `LayoutTree` (and then re-enabling selection with `user-select: text` on the editables themselves) was enough to stop all of the highlighting. The `ScrollZone`, `QuickDropPanel` etc. don't get highlighted anymore either.

The only difference in functionality that I've noticed is that I no longer get haptic feedback (or the accompanying lens popover) when long-pressing on the bullet or the empty space to the right of the thought. I still get haptic feedback when long-pressing the thought itself. If the feedback was desirable, we will need to add it as part of `useDragHold`, probably.

### DIFF
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -231,6 +231,7 @@ const LayoutTree = () => {
       className={cx(
         css({
           marginTop: '0.501em',
+          userSelect: 'none',
         }),
         fauxCaretTreeProvider(indent),
       )}
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -231,6 +231,9 @@ const LayoutTree = () => {
       className={cx(
         css({
           marginTop: '0.501em',
+          // Disallow text selection on the entire tree to prevent selection of scroll zone and other background elements when dragging.
+          // This is overriden by the editable recipe to enable text selection on editable thoughts.
+          // https://github.com/cybersemics/em/pull/2962
           userSelect: 'none',
         }),
         fauxCaretTreeProvider(indent),
--- a/src/recipes/editable.ts
+++ b/src/recipes/editable.ts
@@ -23,6 +23,7 @@ const editableRecipe = defineRecipe({
       _mobile: '-2px',
     },
     paddingBottom: { _mobile: '0' },
+    userSelect: 'text',
   },
 })
 
--- a/src/recipes/editable.ts
+++ b/src/recipes/editable.ts
@@ -23,6 +23,8 @@ const editableRecipe = defineRecipe({
       _mobile: '-2px',
     },
     paddingBottom: { _mobile: '0' },
+    // Re-enable text selection that was disabled on the LayoutTree.
+    // https://github.com/cybersemics/em/pull/2962
     userSelect: 'text',
   },
 })


trevinhofmann (changes requested):
Thank you for the pull request, @ethan-james!

I was able to reproduce the issue on `main`.

Testing locally, this change does resolve the highlighting issue when dragging the bullet or the area to the right of thoughts. However, I still see the highlighting issue when dragging a thought itself. While technically a separate cause than #2961, I think it may be worth investigating and solving it here as well.

I'll let @raineorshine decide if haptic feedback is desirable when long-pressing on the bullet or the area to the right of the thoughts.

## 1. Dragging a thought causes areas to be highlighted

### Steps to Reproduce

```
- Bar
- Foo
```

1. Long press on the `Bar` thought
2. Drag down below `Foo`

### Current Behavior

When dragging, various background elements get highlighted.

<details>
<summary>Demo</summary>

https://github.com/user-attachments/assets/1cc54b4a-1d5d-4c45-bb98-244732ecf1f0
</details>

## Expected Behavior

Long press on a thought should not cause various highlighting.

ethan-james:
That's weird, I don't see anything like that in the #2961 branch....

https://github.com/user-attachments/assets/6ab49337-a8fa-404e-b46d-d6ddc4fc09cb

raineorshine:
> The only difference in functionality that I've noticed is that I no longer get haptic feedback (or the accompanying lens popover) when long-pressing on the bullet or the empty space to the right of the thought.

We have better control over haptics through the Capacitor plugin, so it's totally okay to lose the default haptics on long tap.

trevinhofmann:
@ethan-james, thank you. I'm retesting (1) now. The highlighting issue appears to happen sometimes, but not always. I'm trying to narrow down what the exact steps to reproduce it consistently are.

trevinhofmann:
It's still a bit inconsistent for me, but here are my findings for the "initial" state before dragging the `Bar` thought:

* The issue does not occur if the keyboard is open and editing `Bar`
* The issue sometimes occurs if the keyboard is closed, and `Bar` is selected
* The issue usually occurs if the keyboard is closed, and `Foo` is selected

Please test again a few times and let me know if you are still unable to reproduce it. Also, try dragging it to the lower-right side of the screen. Just a bit below the `Foo` thought is not far enough to cause the highlighting.

ethan-james:
> It's still a bit inconsistent for me, but here are my findings for the "initial" state before dragging the `Bar` thought:
> 
> * The issue does not occur if the keyboard is open and editing `Bar`
> * The issue sometimes occurs if the keyboard is closed, and `Bar` is selected
> * The issue usually occurs if the keyboard is closed, and `Foo` is selected
> 
> Please test again a few times and let me know if you are still unable to reproduce it. Also, try dragging it to the lower-right side of the screen. Just a bit below the `Foo` thought is not far enough to cause the highlighting.

Thanks for doing that. Nothing so far for me. Are you getting the haptic feedback when the drag begins? I would assume that you are since you're on the thought itself.

Another inconsistency that I just noticed is that if you initiate drag on the thought, and leave the thought in place (complete the drag without moving the thought), the cursor switches to the thought that you just were dragging. Whereas if you initiate drag off (left or right) of the thought, and leave the thought in place, then the cursor does not change. Not sure if that's a bug or not.

trevinhofmann (approved):
We seem to have 3 causes of the highlighting:

A) Dragging the bullet
B) Dragging to the right of the thought
C) Dragging the thought itself (inconsistently)

Since #2961 only identified (A) and (B), this change appears to fix both, and you are not able to reproduce (C), I will mark this as approved so that #2961 can be resolved. I've create a separate issue to track (C): #2964

> Are you getting the haptic feedback when the drag begins?

Not getting any haptics on iOS Safari (should they be working?), but I do get haptic feedback for the drag with the Capacitor app, yes.

Issue #2946, opened 2025-05-13
Do not dismiss Alert while hovering


raineorshine:
## Steps to Reproduce

1. Delete a non-empty thought.
2. Hover over the Alert.

## Current Behavior

The alert is dismissed after a delay,

## Expected Behavior

The Alert should not be dismissed as long as it is being hovered.

On mouseout, the timer to dismiss should restart.

PR #2970, opened 2025-05-27
Prevent alert from being dismissed while hovering

Bheart7:
The PR solves this issue https://github.com/cybersemics/em/issues/2946


raineorshine (approved):
Looks great! Thanks so much!

PR #3007, opened 2025-06-02
Disable pointer-events on Editable while long-pressing

### LINKED ISSUE #2953: [iOS] Disable magnifying glass (loupe) on long press
## Steps to Reproduce

2. Close the keyboard. 
3. Long tap `world`.  

## Current Behavior

The iOS loupe appears.

## Expected Behavior

ethan-james:
Fixes #2953

Setting `pointer-events: none` on editables while a long press is in progress does a pretty good job of preventing iOS Safari from triggering editing-related behavior. It still needs more work in order to be reliable.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -595,7 +595,11 @@ const Editable = ({
       innerRef={contentRef}
       aria-label={'editable-' + head(path)}
       data-editable
-      className={cx(multiline ? multilineRecipe() : null, editableRecipe(), className)}
+      className={cx(
+        multiline ? multilineRecipe() : null,
+        editableRecipe({ editing: disabled ? 'disabled' : undefined }),
+        className,
+      )}
       html={
         value === EM_TOKEN
           ? '<b>em</b>'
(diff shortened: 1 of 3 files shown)


ethan-james:
This latest iteration works surprisingly well (surprising because the specific timeout values are very important and counter-intuitive to me) but it's not perfect and unfortunately I think this approach is a dead-end.

Only `pointer-events: none` will reliably block selection during a long press, but it needs to be applied well before the threshold for a long press (400ms) and only works 100% of the time when the element has it applied before the touch begins. This interferes with default selection on short single or double taps. The magic combination of two 100ms timeouts is the best that I could find.

The next avenue of research that I found is to try to block touches from propagating from `Thought` down to the editable during the capture phase. For some reason it's ineffective if I disable `react-dnd`, but can block touches while drag-and-drop is enabled. In that case, it shows the default drag preview that is normally suppressed.

https://github.com/user-attachments/assets/9b915300-8c06-474b-810c-092484945622

This is probably going to be another weird rabbit-hole, so I figured I should solicit feedback first.

ethan-james:
Actually, I had pretty good luck blocking a `contenteditable` div with a non-editable div during the tap: https://codesandbox.io/p/sandbox/4rqfwf

It might be feasible to use the existing `ThoughtAnnotation` to intercept long presses in the same way.

Issue #3004, opened 2025-06-01
Favorite, Pin, and Pin All icons should be filled when active


raineorshine:
## Steps to Reproduce

Activate `Favorite`, `Pin`, or `Pin All` from the Toolbar.

## Current Behavior

✓ Inactive
✓ Animating
✗ Active

Icons are rendered with a `fg` stroke when active, after the Lottie animation completes.

https://github.com/user-attachments/assets/23a399cc-00ab-477a-adda-58609a38ff5f

## Expected Behavior

When in the active state, `Favorite`, `Pin`, and `Pin All` should be rendered with a `fg` fill instead of just a stroke. This is more consistent with the end of the animation, and will make it clearer when these commands are active.

PR #3010, opened 2025-06-04
Filled Favorite, Pin, and Pin All icons on active state

Bheart7:
The PR solves this issue https://github.com/cybersemics/em/issues/3004

### DIFF
--- a/src/@types/Command.ts
+++ b/src/@types/Command.ts
@@ -116,7 +116,7 @@ interface Command {
   preventDefault?: boolean
 
   /** An icon that represents the command in the Toolbar. */
-  svg?: (icon: IconType) => React.ReactNode
+  svg?: (icon: IconType & { active?: boolean}) => React.ReactNode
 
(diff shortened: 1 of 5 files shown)


raineorshine (changes requested):
Thanks for your submission!

This works great for the Favorites button, but the Pin and Pin All don't fill in all the way.

As for the implementation, it needs to be a bit better integrated into the existing implementation which uses `style.fill` to set the fill. The PR currently overrides `style.fill` with a separate `fill`. However, `style.fill` is [already based on `isButtonActive`](https://github.com/cybersemics/em/blob/7b1fa69c550722045cdc87646061eff0a5a8c080/src/components/ToolbarButton.tsx#L181). So there is some duplication and overlap that's going to be confusing for future developers that look at this. 

I'd like to find a more unified approach to accommodate the fact that some icons display the active state differently. Maybe `IconType` itself needs to define `active` and the `Icon` should handle the fill rather than it being passed through generically on the `style` param.

Issue #2964, opened 2025-05-22
Dragging a thought causes areas to be highlighted


trevinhofmann:
## Dragging a thought causes areas to be highlighted

This is related to #2961.

### Steps to Reproduce

```
- Bar
- Foo
```

1. Long press on the `Bar` thought
2. Drag down far below `Foo` (and to the right)

### Current Behavior

When dragging, various background elements get highlighted.

<details>
<summary>Demo</summary>

https://github.com/user-attachments/assets/1cc54b4a-1d5d-4c45-bb98-244732ecf1f0
</details>

## Expected Behavior

Long press on a thought should not cause various highlighting.

Issue #2931, opened 2025-05-09
[iOS] Writing Tools native menu appears on long tap thought


raineorshine:
## Steps to Reproduce

1. Create thought `Hello world`.
2. Close the keyboard. 
3. Long tap the thought to enter multiselect mode.
4. Long tap the thought again to exit multiselect mode.

## Current Behavior

A native context menu that says "Writing Tools" appears at the top of the screen. 

https://github.com/user-attachments/assets/43a61b50-b224-48d5-8bc2-270bbaea763d

## Expected Behavior

The native context menu should not appear on long tap.

PR #3011, opened 2025-06-05
Use HTML5 backend for drag-and-drop on iOS Safari

### LINKED ISSUE #2953: [iOS] Disable magnifying glass (loupe) on long press

ethan-james:
Fixes #2953, #2931 and #2964

While trying to block `touchstart` events from reaching the editables, I ran into some very weird behavior where `onTouchStartCapture={e => e.stopPropagation()}` seemed to stop `touchstart` events in parents as well as children. Now that I'm typing this, I realize that stopping an event on the way down (capture) probably also stops it from bubbling back up and being caught by the `TouchBackend`.

In any case, I realized that `TouchBackend` was being disabled and drag-and-drop was falling back to `HTML5Backend` which uses native drag events and was working a lot better in iOS Safari. It doesn't trigger any of the long press behavior that we want to avoid, presumably because it understands that a native drag is underway. According to https://caniuse.com/?search=drag, native drag-and-drop has been supported in iOS Safari since September 2021, so we may need to add an additional check if we want to support devices that haven't updated since then.

I tested on Android Chrome, and `HTML5Backend` had a lot more problems, so I kept `TouchBackend` for them.

I tested with drag-and-drop favorites and toolbar buttons and it seems to work without issue. Drag-and-drop thoughts work well, but don't drag-and-drop at all for thoughts other than the cursor thought. The cursor thought seems to work flawlessly, so I'm optimistic that non-cursor thoughts are solvable.

Let me know if  you think it's feasible to get rid of `TouchBackend` on iOS Safari, because this feels like a pretty good solution to me.

### DIFF
--- a/src/components/DragAndDropContext.tsx
+++ b/src/components/DragAndDropContext.tsx
@@ -1,10 +1,11 @@
 import { FC, PropsWithChildren } from 'react'
 import { HTML5Backend } from 'react-dnd-html5-backend'
-import { DndProvider, PointerTransition, TouchTransition } from 'react-dnd-multi-backend'
+import { DndProvider, MultiBackendOptions, PointerTransition, TouchTransition } from 'react-dnd-multi-backend'
 import { TouchBackend } from 'react-dnd-touch-backend'
+import { isSafari, isTouch } from '../browser'
 import { TIMEOUT_LONG_PRESS_THOUGHT } from '../constants'
 
-const options = {
+const options: MultiBackendOptions = {
   backends: [
     // HTML5 backend
     // https://react-dnd.github.io/react-dnd/docs/backends/html5
@@ -13,6 +14,11 @@ const options = {
       backend: HTML5Backend,
       transition: PointerTransition,
     },
+  ],
+}
+
+if (isTouch && !isSafari()) {
+  options.backends.push(
     // Touch backend
     // https://react-dnd.github.io/react-dnd/docs/backends/touch
     {
@@ -22,7 +28,7 @@ const options = {
       preview: true,
       transition: TouchTransition,
     },
-  ],
+  )
 }
 
 /** Drag and Drop Provider HOC. */
--- a/src/hooks/useDragAndDropThought.tsx
+++ b/src/hooks/useDragAndDropThought.tsx
@@ -232,7 +232,7 @@ const useDragAndDropThought = (props: Partial<ThoughtContainerProps>) => {
   })
 
   useEffect(() => {
-    dragPreview(getEmptyImage(), { captureDraggingState: true })
+    dragPreview(getEmptyImage())
   }, [dragPreview])
 
   const [{ isHovering, isBeingHoveredOver, isDeepHovering, canDropThought }, dropTarget] = useDrop({
--- a/src/hooks/useDragAndDropThought.tsx
+++ b/src/hooks/useDragAndDropThought.tsx
@@ -1,6 +1,7 @@
 import moize from 'moize'
+import { useEffect } from 'react'
 import { DragSourceMonitor, DropTargetMonitor, useDrag, useDrop } from 'react-dnd'
-import { NativeTypes } from 'react-dnd-html5-backend'
+import { NativeTypes, getEmptyImage } from 'react-dnd-html5-backend'
 import DragAndDropType from '../@types/DragAndDropType'
 import DragThoughtItem from '../@types/DragThoughtItem'
 import DragThoughtOrFiles from '../@types/DragThoughtOrFiles'
@@ -229,6 +230,10 @@ const useDragAndDropThought = (props: Partial<ThoughtContainerProps>) => {
     end: () => endDrag(),
     collect: dragCollect,
   })
+
+  useEffect(() => {
+    dragPreview(getEmptyImage(), { captureDraggingState: true })
+  }, [dragPreview])
 
   const [{ isHovering, isBeingHoveredOver, isDeepHovering, canDropThought }, dropTarget] = useDrop({
     accept: [DragAndDropType.Thought, NativeTypes.FILE],
--- a/src/hooks/useDragAndDropToolbarButton.ts
+++ b/src/hooks/useDragAndDropToolbarButton.ts
@@ -99,7 +99,7 @@ const useDragAndDropToolbarButton = ({ commandId, customize }: { commandId: Comm
   })
 
   useEffect(() => {
-    dragPreview(getEmptyImage(), { captureDraggingState: true })
+    dragPreview(getEmptyImage())
   }, [dragPreview])
 
   const [{ isHovering }, dropTarget] = useDrop({
--- a/src/hooks/useDragAndDropToolbarButton.ts
+++ b/src/hooks/useDragAndDropToolbarButton.ts
@@ -1,5 +1,6 @@
+import { useEffect } from 'react'
 import { DropTargetMonitor, useDrag, useDrop } from 'react-dnd'
-import { NativeTypes } from 'react-dnd-html5-backend'
+import { NativeTypes, getEmptyImage } from 'react-dnd-html5-backend'
 import Command from '../@types/Command'
 import CommandId from '../@types/CommandId'
 import DragAndDropType from '../@types/DragAndDropType'
@@ -96,6 +97,10 @@ const useDragAndDropToolbarButton = ({ commandId, customize }: { commandId: Comm
       isDragging: monitor.isDragging(),
     }),
   })
+
+  useEffect(() => {
+    dragPreview(getEmptyImage(), { captureDraggingState: true })
+  }, [dragPreview])
 
   const [{ isHovering }, dropTarget] = useDrop({
     accept: [DragAndDropType.ToolbarButton, NativeTypes.FILE],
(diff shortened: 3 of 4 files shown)


src/hooks/useLongPress.ts
```diff
@@ -66,24 +66,24 @@ const useLongPress = (
       // Delay setPressed(false) to ensure that onLongPressEnd is not called until bubbled events complete.
       // This gives other components a chance to short circuit.
       // We can't stop propagation here without messing up other components like Bullet.
-      setTimeout(() => {
-        clearTimeout(timerIdRef.current)
-        timerIdRef.current = 0
-        longPressStore.unlock()
+      // setTimeout(() => {
```
ethan-james: Not sure if this has ramifications, but it's necessary to make multiselect work reliably with the new drag events. It looks like it works very well with my drag-and-drop testing.

src/components/DragAndDropContext.tsx
```diff
@@ -13,6 +14,11 @@ const options = {
       backend: HTML5Backend,
       transition: PointerTransition,
     },
+  ],
+}
+
+if (isTouch && !isSafari()) {
+  options.backends.push(
```
raineorshine: Please use inline `...isTouch && !isSafari ? [...] : []` syntax to avoid mutations.

src/hooks/useLongPress.ts
raineorshine: Thanks! A few due diligence steps are needed to delete existing code with an unknown purpose from the codebase:

1. Use `git blame` to identify the commit that added the `setTimeout`.
2. Identify the PR or issue associated with the commit if provided.
3. If the commit addressed a specific issue, re-test the steps to reproduce to ensure that removing the code does not re-expose the original bug.

src/components/Editable.tsx
```diff
@@ -127,6 +127,7 @@ const Editable = ({
   const nullRef = useRef<HTMLInputElement>(null)
   const contentRef = editableRef || nullRef
   const editingOrOnCursor = useSelector(state => state.editing || equalPath(path, state.cursor))
+  const isPressingRef = useRef(false)
```
raineorshine, line 133: I'm sure you'll get to this, but let's make sure to add a detailed comment here explaining the purpose of `isPressingRef`.

src/components/Editable.tsx
```diff
@@ -535,7 +536,7 @@ const Editable = ({
       // Steps to Reproduce: https://github.com/cybersemics/em/pull/2948#issuecomment-2887186117
       // Explanation and demo: https://github.com/cybersemics/em/pull/2948#issuecomment-2887803425
       else {
-        e.preventDefault()
+        if (!isPressingRef.current) e.preventDefault()
```
raineorshine, line 542: Wouldn't `isPressingRef.current` always be true in `onMouseDown`, since it comes after `onTouchStart`?

raineorshine (commented):
Great idea! I wouldn't have thought of this. Yes, if Safari support for HTML5 drag-and-drop is solid, this seems feasible.

> I tested with drag-and-drop favorites and toolbar buttons and it seems to work without issue. Drag-and-drop thoughts work well now that I've fixed a couple of bugs related to `focus` and `mousedown` events, and one related to the timing of multi-select.

You can ignore toolbar button drag-and-drop since the toolbar is going away in a re-design soon.

src/components/Editable.tsx
ethan-james: It turns out that long press behavior is different, with `mousedown` firing after `touchstart`, probably after a delay to detect long presses. `focus` also fires before `touchend`, presumably once Safari converts the touch into a long press internally.

src/hooks/useLongPress.ts
ethan-james: The commit message says "Prevent Bullet click on dragHold." and I can confirm that `dragHold` doesn't affect the bullet if the long press is initiated on the bullet, with or without dragging.

src/components/Editable.tsx
raineorshine: I'm not quite understanding. When would `isPressingRef.current` by false on `mousedown`? i.e. When would `mousedown` *not* be preceded by `touchstart`?

ethan-james: On a short tap, `touchend` occurs before `mousedown` and sets `isPressingRef.current` back to false. I always thought that was the only way it would work, i.e. `touchstart` -> `touchend` -> `mousedown`. But for natively-supported long presses, `mousedown` will still occur even if `touchend` is delayed indefinitely.

raineorshine: I see, thanks! I was just thinking about `touchstart` and `mousedown`, and I neglected `touchend`. It's the early `touchend` that sets `isPressingRef.current` to false and thus allows the default browser behavior in `mousedown`.

ethan-james: Right, I considered the possibility of adding it to the component's event handler like

```
onTouchEnd={e => {
  isPressingRef.current = false
  onTap(e)
}}
```

to make it easier to see. Do you think I should do that?

raineorshine: I would keep it in `onTap` unless you specifically don't want it called on `click`.

ethan-james:
1. I found a very niche bug (#3012) that can probably be solved by simply adding `pointer-events: none` to the alert notification.
2. The dragging thought turns gray while it's being dragged, rather than remaining white as expected, and I haven't been able to easily figure out why.

Other than that, I think this is ready for some rigorous testing.

src/components/DragAndDropContext.tsx
```diff
@@ -13,15 +14,19 @@ const options = {
...
-    // Touch backend
-    // https://react-dnd.github.io/react-dnd/docs/backends/touch
-    {
-      id: 'touch',
-      backend: TouchBackend,
-      options: { delayTouchStart: TIMEOUT_LONG_PRESS_THOUGHT },
-      preview: true,
-      transition: TouchTransition,
-    },
+    ...(isTouch && !isSafari()
```
raineorshine, line 17: When we disable the `TouchBackend` on Safari mobile, we lose `delayTouchStart: TIMEOUT_LONG_PRESS_THOUGHT` (400ms).

I tried simply adding it to the `HTML5Backend`, but it has no effect. It's specific to `TouchBackend`.

## Steps to Reproduce

Long press a thought for about 600ms (200ms longer than `TIMEOUT_LONG_PRESS_THOUGHT`) and drag down.

## Current Behavior

The bullet highlight and drag-and-drop alert appear, but drag-and-drop is not activated.

The thought needs to be held longer than usual for drag-and-drop to work.

https://github.com/user-attachments/assets/bd8c7d3a-3c07-4b31-b8eb-56fb4e06f2ae

## Expected Behavior

Drag-and-drop should be activated after a thought is pressed for the requisite amount of time.

The timing should coincide with the bullet highlight and drag-and-drop alert, and the user should be able to immediately start dragging once it is activated. 

https://github.com/user-attachments/assets/50ff1c55-fb42-4b67-a159-f5bde5c94a13

src/components/DragAndDropContext.tsx
raineorshine, line 1: ## Steps to Reproduce

```
- a
- b
- c
- Hello world
```

1. Close the keyboard
2. Tap and hold `Hello world` for a good amount of time to activate drag-and-drop.
3. Drag slightly up.

## Current Behavior

Command menu appears.

https://github.com/user-attachments/assets/baf40768-04e0-406f-9232-a0180341d7cd

## Expected Behavior

Command menu should not during drag-and-drop, only long press (i.e. release without movement).

https://github.com/user-attachments/assets/274e365b-3759-42a1-aada-a57965ba566c

raineorshine (changes requested):
Thank you! It fixes all the specified issues quite nicely.

I found a couple bugs. I would recommend starting with the first one (https://github.com/cybersemics/em/pull/3011#discussion_r2136359993) since it is a potential dealbreaker.

> 2\. The dragging thought turns gray while it's being dragged, rather than remaining white as expected, and I haven't been able to easily figure out why.

Thanks, I see that, too. If the approach proves viable, this will also need to be addressed.

src/hooks/useLongPress.ts
raineorshine: Thanks!

Normally, please post the commit hash as well so we have a good record (not necessary now).

src/components/DragAndDropContext.tsx
ethan-james: First of all, I assume it's not acceptable simply to update `TIMEOUT_LONG_PRESS_THOUGHT` to match the iOS Safari drag timeout? I was already proceeding from the assumption that drag needs to begin at 400ms, but if it's OK to trigger long press after 500 or 600ms, please let me know because that's way easier.

Second, it doesn't seem like it's possible to configure the Safari drag timeout, so it looks like we need to return to using `touchstart` to begin a drag. I've been trying to isolate which part of the `HTML5Backend` disables the native long press behavior, and I came across something that needs clarification.

https://github.com/user-attachments/assets/4809f777-a3a2-4d39-89b9-dfedb0a658c0

When a thought is in editing mode, a long press will still initiate `dragHold` behavior. Is it important that native long press be re-enabled when the user is editing a thought, so that the loupe appears and selecting text via long press is possible?

PR #3016, opened 2025-06-08
Drag-and-drop multiple thoughts with multiselect

### LINKED ISSUE #2933: Drag-and-drop multiple thoughts with multiselect
## Steps to Reproduce

1. Select multiple thoughts. 
2. Drag any of the selected thoughts to another thought. 

## Current Behavior

Drag-and-drop is not allowed during multiselect. 

## Expected Behavior

- [ ] All selected thoughts should be dragged to the destination. 
- [ ] Order should be preserved. 
- [ ] Drop should be disallowed if any of the drops are illegal. 
- [ ] Dragging a thought that is not selected should add it to multiselect.

karunkop:
Close #2933 

## Tasks

- [x] All selected thoughts should be dragged to the destination.

- [x] Order should be preserved.

- [x] Drop should be disallowed if any of the drops are illegal.

- [x] Dragging a thought that is not selected should add it to multiselect.

To enable dragging and dropping of multi-cursor thoughts, the key change was updating the draggingThought state value from a SimplePath to a SimplePath[]. This allowed the application to track multiple selected thoughts simultaneously, making it possible to support multi-selection drag-and-drop functionality.


raineorshine:
Thanks! It looks like the new puppeteer test is failing. Could you investigate?

raineorshine:
The test failed in 2/3 runs, so we'll need to make it more reliable before merging it into `main`. We don't want to add any other flaky tests. Indeed, the flaky tests will be your first priority once you shift over into the test engineer role, so this will be good practice.

raineorshine:
The snapshot tests need to catch visual regressions that are as small as a single pixel. For example, if a contributor makes a change that causes the bullet to be rendered 1px too low, that’s a definite regression. The problem with failure thresholds is that we're going to get false positives for those small regressions, arguably the most important ones to catch. 

If there are general rendering inconsistencies, why don't they show up everywhere? Most snapshots are pixel perfect every time. We run the Puppeteer tests in a docker image so that we have a perfectly identical environment. 

So, I'm afraid failure thresholds are not the solution. For this and other flaky tests, we have to identify the root cause of the inconsistency. The bullet highlighting renders exactly the same every time, but maybe there is a timing issue or an animation that hasn't been disabled in the tests.

src/@types/DragThoughtOrFiles.ts
```diff
@@ -1,6 +1,6 @@
 import DragThoughtItem from './DragThoughtItem'
 
 /** Represents the currently dragged thought or file. Returned by monitor.getItem() in the drop handler, or set to { files } when dragging a native file. */
```
raineorshine: Update comment from "thought" to "thoughts". 

It might be a good idea to go through related comments to make sure none are still in the singular.

src/@types/State.ts
```diff
@@ -53,7 +53,8 @@ interface State {
   /** Set to true when dragging a native file. */
   draggingFile?: boolean
   /** Set to the dragging thought during dragInProgress. */
-  draggingThought?: SimplePath
+  draggingThought?: SimplePath[]
```
raineorshine: Update to `draggingThoughts` (plural) and comment.

src/e2e/puppeteer/helpers/multiselectThoughts.ts
```diff
@@ -0,0 +1,52 @@
+import { isMac } from '../../../browser'
+import sleep from '../../../util/sleep'
+import { page } from '../setup'
+import clickThought from './clickThought'
+
+/** Multiselect thoughts by holding Cmd/Ctrl and clicking them.*/
+const multiselectThoughts = async (
```
raineorshine, line 7: Hmm, I'm not sure why the modifier logic is here. **em** doesn't use Cmd/Ctrl + click for multiselect. It uses long press :).

src/@types/State.ts
raineorshine: It would probably be better to default to `[]` and disallow `undefined`. Otherwise it will force you to have special handling everywhere to treat `undefined` and `[]` as the same.

src/components/Thought.tsx
```diff
@@ -228,21 +228,21 @@ const ThoughtContainer = ({
...
-    // Do not highlight parent of dragging thought (i.e. when simply reordering but not moving to a new parent).
-    // The only exception is table view col1, which needs highlighting as the first thought of row1 abuts the first thought in row2.
-    if (
-      equalPath(rootedParentOf(state, state.draggingThought), simplePath) &&
-      !attributeEquals(state, head(rootedParentOf(state, simplePath)), '=view', 'Table')
-    ) {
+    const wouldCauseHighlightingIssues = state.draggingThought.some(
+      draggingThought =>
+        // Make sure we're not hovering over a thought that cannot be dropped on
+        isDescendantPath(state.hoveringPath!, draggingThought) ||
```
raineorshine: Avoid overriding the type system with `!`. Usually it's preferable to add a runtime guard.

In the rare instances where it is needed, add an explanatory comment.

src/components/Thought.tsx
```diff
@@ -228,21 +228,21 @@ const ThoughtContainer = ({
...
-      return false
-    }
-
-    // Do not highlight parent of dragging thought (i.e. when simply reordering but not moving to a new parent).
-    // The only exception is table view col1, which needs highlighting as the first thought of row1 abuts the first thought in row2.
-    if (
-      equalPath(rootedParentOf(state, state.draggingThought), simplePath) &&
-      !attributeEquals(state, head(rootedParentOf(state, simplePath)), '=view', 'Table')
-    ) {
+    const wouldCauseHighlightingIssues = state.draggingThought.some(
```
raineorshine: Highlighting "issues" is a bit too vague. I also prefer to stick with the `is` prefix for booleans rather than getting in the habit of using different grammatical tenses like `are`, `would`, `will`, `should`, etc.

Maybe `isNotHighlightable`?

src/components/Thought.tsx
```diff
@@ -228,21 +228,21 @@ const ThoughtContainer = ({
...
+      draggingThought =>
+        // Make sure we're not hovering over a thought that cannot be dropped on
+        isDescendantPath(state.hoveringPath!, draggingThought) ||
+        // Do not highlight parent of any dragging thought (i.e. when simply reordering but not moving to a new parent).
+        // The only exception is table view col1, which needs highlighting as the first thought of row1 abuts the first thought in row2.
+        (equalPath(rootedParentOf(state, draggingThought), simplePath) &&
+          !attributeEquals(state, head(rootedParentOf(state, simplePath)), '=view', 'Table')),
```
raineorshine: `attributeEquals(state, head(rootedParentOf(state, simplePath)), '=view', 'Table')` can be pulled out of the `.some` for better efficiency.

src/e2e/puppeteer/__tests__/drag-and-drop.ts
```diff
@@ -500,3 +501,86 @@ describe('hover expansion', () => {
...
+          timeout: 5000,
+        })
+      }
+    } catch (error) {
+      // Ignore errors from cleanup
+      console.warn('Mouse cleanup error:', (error as Error).message)
+    }
+
+    // take the final snapshot after cleanup
+    expect(await screenshot()).toMatchImageSnapshot()
```
raineorshine: I would suggest using a normal puppeteer test, not a snapshot test. Snapshots are more appropriate for visual regression tests. 

Here weren't not trying to assert the appearance, but that the move occurred as expected. Notice also how the test cases don't give any indication about the expected state. One has to open the snapshot image rather than having it neatly described in one place. Doing this as a snapshot test will be slower, harder to maintain, and tightly couple the tests to unrelated visual aspects (e.g. thought appearance, bullet size, positioning, etc).

I might suggest exporting the thoughts for a simple string comparison like we do in many of the other tests. I don't know if we have a puppeteer helper for this yet. While this goes against the normal imperative of only doing things in integration tests that the user can do, I think the benefits in terms of efficiency and test readability make up for it. It would be quite a pain to individually select all the thoughts from the DOM and recreate the structure when we can just export them directly. It also seems unnecessary to open the Export modal as the user would. Let me know what you think.

src/hooks/useDragAndDropSubThought.ts
```diff
@@ -101,66 +101,152 @@ const drop = (props: DroppableSubthoughts, monitor: DropTargetMonitor) => {
...
-      const alertFrom = '"' + ellipsize(thoughtFrom.value) + '"'
-      const alertTo = parentIdTo === HOME_TOKEN ? 'home' : '"' + ellipsize(thoughtTo.value) + '"'
-      const inContext = props.showContexts
-        ? ` in the context of ${ellipsize(headValue(state, props.simplePath ?? props.path) ?? 'MISSING_CONTEXT')}`
-        : ''
-
-      store.dispatch(
-        alert(`${alertFrom} moved to${dropTop ? ' top of' : ''} ${alertTo}${inContext}.`, {
-          clearDelay: 5000,
+      return { parentId, dropTop }
```
raineorshine: Try to avoid functions that return groups of values as an object. It's an ugly pattern. Remember that functions are meaningful units of work, not arbitrary bundles of functionality. There is no singular concept that corresponds to `{ parentId, dropTop }`.

I think it would be okay to return `dropTop` directly. `parentId` is easily accessed from the `Path` with `head(rootedParentOf(state, pathTo))` and does not need to be abstracted.

src/hooks/useDragAndDropThought.tsx
```diff
@@ -62,18 +73,39 @@ const canDrag = (props: ThoughtContainerProps) => {
...
+  const isCurrentThoughtSelected = isMulticursorPath(state, path)
+
+  // Add the current thought to the multicursor if it is not already selected
+  if (isMulticursorActive && !isCurrentThoughtSelected) {
+    store.dispatch(addMulticursor({ path }))
+  }
+
+  const activeMulticursors = isMulticursorActive ? Object.values(state.multicursors) : []
+  const multicursorPaths =
+    isMulticursorActive && !isCurrentThoughtSelected ? [...activeMulticursors, path] : activeMulticursors
```
raineorshine: You can simplify things down a bit when you notice that a special case is unnecessary. Sometimes the base case or the null case are handled naturally by the existing logic.

For example, there's no need to check `isMulticursorActive` since `Object.values(state.multicursors)` is equal to `[]` anyway.

And `draggingThoughts` ends up equal to just the path anyway when the multicursor is inactive, so there's no need to branch there. Sorting an array of `[path]` is harmless and you can let it fall out naturally from the multicursor logic. The end result is less code that does the same thing.

```diff
diff --git a/src/hooks/useDragAndDropThought.tsx b/src/hooks/useDragAndDropThought.tsx
index a42b7c3bc3..e8fdb89891 100644
--- a/src/hooks/useDragAndDropThought.tsx
+++ b/src/hooks/useDragAndDropThought.tsx
@@ -84,17 +84,13 @@ const beginDrag = ({ path, simplePath }: ThoughtContainerProps): DragThoughtItem
     store.dispatch(addMulticursor({ path }))
   }
 
-  const activeMulticursors = isMulticursorActive ? Object.values(state.multicursors) : []
-  const multicursorPaths =
-    isMulticursorActive && !isCurrentThoughtSelected ? [...activeMulticursors, path] : activeMulticursors
-
-  const draggingThoughts = isMulticursorActive
-    ? documentSort(state, multicursorPaths).map(path => ({
-        path,
-        simplePath: simplifyPath(state, path),
-        zone: DragThoughtZone.Thoughts,
-      }))
-    : [{ path, simplePath, zone: DragThoughtZone.Thoughts }]
+  const multicursorPaths = [...Object.values(state.multicursors), ...(!isCurrentThoughtSelected ? [path] : [])]
+
+  const draggingThoughts = documentSort(state, multicursorPaths).map(path => ({
+    path,
+    simplePath: simplifyPath(state, path),
+    zone: DragThoughtZone.Thoughts,
+  }))
 
   store.dispatch(
     dragInProgress({
```

src/hooks/useDragAndDropThought.tsx
```diff
@@ -134,74 +166,126 @@ const drop = (props: ThoughtContainerProps, monitor: DropTargetMonitor) => {
...
+  store.dispatch((dispatch, getState) => {
+    // set multicursor executing to true if there are multiple thoughts being dragged
+    if (draggedItems.length > 1) {
+      dispatch(setIsMulticursorExecuting({ value: true, undoLabel: 'Dragging Thoughts' }))
+    }
+
+    // move each dragged item to the destination path
+    draggedItems.forEach(item => {
+      const state = getState()
+      const parent = unroot(rootedParentOf(state, props.simplePath))
```
raineorshine: `unroot(rootedParentOf(state, props.simplePath))` → `parentOf(props.simplePath)`

raineorshine (changes requested):
Hi, thanks for your submission and apologies for the delayed code review!

Overall this looks good. It works well from my testing.

---

Wherever you change `draggingThought` to `draggingThought[0]`, it's probably wrong. There's no reason that the first drag dragging thought would be representative of all dragging thoughts. For example, if we used the value of the dragging thought before, we probably now need to use the value of all dragging thoughts, or a summary of their values. There's nothing special about the value of the first dragging thought. We need to think about the semantics of each circumstance and come up with behavior that makes sense to the user. I think you do a great job of that when setting `alertFrom`: https://github.com/karunkop/em/blob/d35db67b7cd12f92dfe1bfc16765ae2ecfeebb91/src/hooks/useDragAndDropSubThought.ts#L233-L234

src/e2e/puppeteer/helpers/multiselectThoughts.ts
karunkop: Hi I can see multicursors being toggled with modifier keys here https://github.com/cybersemics/em/blob/main/src/components/Thought.tsx#L501-L510

raineorshine: You're right, my apologies!

It seems I discovered a bug, as Ctrl + click does not work on the bullet.

karunkop:
https://github.com/cybersemics/em/issues/3100 is occurring frequently in almost all new PRs not just in CI but also locally

src/e2e/puppeteer/helpers/multiselectThoughts.ts
raineorshine: Now tracking in #3066

src/actions/freeThoughts.ts
```diff
@@ -52,7 +52,6 @@ const freeThoughts = (state: State): State => {
     HOME_TOKEN,
     // drag-and-drop related thoughts
     ...(state.draggedSimplePath ?? []),
-    ...(state.draggingThought ?? []),
```
raineorshine, line 55: The dragging thoughts need to be preserved, otherwise they could be removed from memory during a drag.

Please flatten and add them all to the preserve set.

src/components/DropChild.tsx
```diff
@@ -89,11 +89,16 @@ const DropChild = ({ depth, path, simplePath, isLastVisible }: DropChildProps) =
 /** Render the DropChild component if the thought is collapsed, and does not match the dragging thought. This component is an optimization to avoid calculating DropChild hooks when unnecessary. */
 const DropChildIfCollapsed = ({ depth, last, path, simplePath, isLastVisible }: DropChildProps) => {
   const isExpanded = useSelector(state => hasChildren(state, head(simplePath)) && !!state.expanded[hashPath(path)])
-  const draggingThought = useSelector(state => state.draggingThought, shallowEqual)
 
-  // Do not render DropChild on expanded thoughts or on the dragging thought.
+  // Check if this thought is any of the dragging thoughts (single or multiple)
+  const isDraggingThisThought = useSelector(state => {
+    return state.draggingThoughts.some(draggingThought => equalPath(draggingThought, simplePath)) || false
```
raineorshine: You can remove `|| false`.

src/components/DropEnd.tsx
```diff
@@ -64,10 +64,15 @@ const DropEnd = ({
     if (!isParentSorted) return false
 
     // only render drop-hover during drag-and-drop
-    const draggingThoughtValue = state.draggingThought
-      ? getThoughtById(state, head(state.draggingThought))?.value
-      : null
-    if (!draggingThoughtValue) return false
+    // For multiple dragging thoughts, use the alphabetically last one to determine if they should go at the end
```
raineorshine: This would be better solved by just converting the single `compareReasonable` to a `.some`. If any of the dragging values are greater than `lastChildValue`, then the `DropEnd` should be rendered. It's simpler and more efficient (`O(n)`) than sorting the dragging values.

src/components/DropHover.tsx
```diff
@@ -70,19 +70,26 @@ const DropHoverIfVisible = ({
 }) => {
   // true if a thought is being dragged over this drop hover
   const showDropHover = useSelector(state => {
+    // Don't show drop hover between selected thoughts during multiselect
+    const isThisThoughtBeingDragged =
+      state.draggingThoughts.slice(1).some(draggingPath => equalPath(draggingPath, simplePath)) || false
```
raineorshine: I think `draggingThoughts.slice(1)` is wrong for the same reason that `draggingThoughts[0]` is wrong. `draggingThoughts` are unordered so there's nothing special about the first dragging thought.

What would make sense here for multiple dragging thoughts?

src/components/DropHover.tsx
```diff
@@ -70,19 +70,26 @@ const DropHoverIfVisible = ({
...
+
+    if (isThisThoughtBeingDragged) return false // Don't show hover lines on thoughts that are being dragged
```
raineorshine: Small tip: I recommend generally placing comments above the line, instead of on the same line. This results in a cleaner git diff, making it clearer when the code or the comment has changed independently.

src/components/DropHover.tsx
```diff
@@ -70,19 +70,26 @@ const DropHoverIfVisible = ({
...
     if (!isParentSorted || sameContext) return testFlags.simulateDrag || isHovering
     else if (!state.dragInProgress) return false
 
-    const draggingThoughtValue = state.draggingThought
-      ? getThoughtById(state, head(state.draggingThought))?.value
-      : null
+    const draggingThoughtValues =
+      state.draggingThoughts.map(draggingPath => getThoughtById(state, head(draggingPath))?.value).filter(Boolean) || []
+
+    if (draggingThoughtValues.length === 0) return false
```
raineorshine: Is a special case needed here, or will this be handled naturally when the below loop has no items to iterate through?

src/components/DropHover.tsx
```diff
@@ -101,14 +108,16 @@ const DropHoverIfVisible = ({
 
     const value = getThoughtById(state, head(simplePath))?.value
     const prevChild = prevChildId && getThoughtById(state, prevChildId)
-    return (
-      value !== undefined &&
-      (isThoughtHovering || isSubthoughtsHovering) &&
-      draggingThoughtValue &&
-      // check if it's alphabetically previous to current thought
```
raineorshine, line 108: I thin these comments were accidentally removed.

src/components/Thought.tsx
```diff
@@ -360,21 +360,23 @@ const ThoughtContainer = ({
   // TODO: It would be nice if we could reuse canDrop.
   const isChildHovering = useSelector(state => {
     // Early return if essential conditions are not met
-    if (!isVisible || !state.hoveringPath || !state.draggingThought) {
+    if (!isVisible || !state.hoveringPath || !state.draggingThoughts || !state.draggingThoughts.length) {
```
raineorshine: The type system now ensures that `state.draggingThoughts` is not falsey, so the extra check is no longer necessary.

src/e2e/puppeteer/__tests__/drag-and-drop-multiselect.ts
```diff
@@ -0,0 +1,118 @@
...
+      - y
+      - z
+      - a
+      `)
+
+    await multiselectThoughts(['y', 'z', 'a'])
+    await dragAndDropThought('a', 'x', { position: 'before', mouseUp: true })
+
+    const exported = await exportThoughts()
+    expect(exported).toBe(`- __ROOT__
```
raineorshine: Let's use `HOME_TOKEN` instead of hardcoding `__ROOT__`:

```ts
  - y
  - z
  - a
  - x`)
```

src/e2e/puppeteer/__tests__/drag-and-drop-multiselect.ts
raineorshine, line 1: Tests look good, thanks!

src/components/DropHover.tsx
karunkop: The `.some()` method naturally handles empty arrays by returning false when there are no items to iterate through.  So no special case is needed as such.

src/components/DropHover.tsx
karunkop: `draggingThoughts` should always be in a document order as per https://github.com/cybersemics/em/pull/3016/files#diff-dca37ee3ca1df64e90e533e8dcfa457b088c224fc736e2590632a42e7a1bb8e9R87-R91. The reason for using slice to remove the first draggingThought is so that `dropHover` is rendered both at the top and bottom of the selected thoughts, only between selected thoughts is where the `dropHover` is not shown

raineorshine: Thanks! I forgot `draggingThoughts` was maintained in sorted order. (We should add a comment to `State`!)

It makes a little more sense now, but I think it doesn't account for noncontiguous series of selected thoughts:

image

The behavior should probably disallow dropping in **between two contiguous selected thoughts**. You are welcome to work on it in this PR, or we can create a separate issue for it in the interest of getting this merged. If the latter, let's leave an informative comment.

src/components/Thought.tsx
raineorshine: @karunkop It should still test `state.draggingThoughts.length` and short circuit for efficiency since this is a heavily used function.

src/components/DropHover.tsx
karunkop: I've addressed the issue for noncontiguous series of selected thoughts. Let me know once you've completed a behavioral test of the feature.

src/components/DropHover.tsx
```diff
@@ -70,19 +71,32 @@ const DropHoverIfVisible = ({
 }) => {
   // true if a thought is being dragged over this drop hover
   const showDropHover = useSelector(state => {
+    // Don't show drop hover on thoughts that are being dragged
```
raineorshine: Please try to keep performance in mind when writing loops. There is a more efficient way to do this at no cost in complexity:

- combine the loops
- calculate `appendToPath(parentOf(simplePath), prevChildId)` outside the loop.

src/components/DropHover.tsx
```diff
@@ -70,19 +71,32 @@ const DropHoverIfVisible = ({
...
+    )
     const isParentSorted = parentId && getSortPreference(state, parentId).type === 'Alphabetical'
     if (!isParentSorted || sameContext) return testFlags.simulateDrag || isHovering
     else if (!state.dragInProgress) return false
 
-    const draggingThoughtValue = state.draggingThought
-      ? getThoughtById(state, head(state.draggingThought))?.value
-      : null
+    const draggingThoughtValues =
+      state.draggingThoughts.map(draggingPath => getThoughtById(state, head(draggingPath))?.value).filter(Boolean) || []
```
raineorshine: I don't believe `|| []` can ever be reached.

raineorshine: `.filter(Boolean)` shouldn't be used on `string[]`, otherwise it will incorrectly filter out empty strings. 

Use `.filter(nonNull)` instead.

src/components/DropHover.tsx
```diff
@@ -101,14 +115,18 @@ const DropHoverIfVisible = ({
...
-      (isThoughtHovering || isSubthoughtsHovering) &&
-      draggingThoughtValue &&
-      // check if it's alphabetically previous to current thought
-      compareReasonable(draggingThoughtValue, value) <= 0 &&
-      // check if it's alphabetically next to previous thought if it exists
-      (!prevChild || compareReasonable(draggingThoughtValue, prevChild.value) === 1)
+
+    if (value === undefined || !(isThoughtHovering || isSubthoughtsHovering)) {
+      return false
+    }
```
raineorshine, line 91: Let's move this condition up to the top of the function so it can short circuit early when neither the thought nor subthought are hovering.

In general you want to short circuit functions before running expensive operations like O(n) loops.

src/components/DropHover.tsx
karunkop: Thank you for the suggestion. I’ve consolidated the loops as recommended and moved the calculation of prevThoughtPath outside the loop.

raineorshine (approved):
Thank you 🙏

PR #3022, opened 2025-06-09
Add interactable flag to Alert

### LINKED ISSUE #3012: [Mobile] Drag-and-drop interrupted by alert
## Steps to Reproduce

```
- A
- B
- C
- D
- E
- F
  - G
    - H
      - I
        - J
          - K
            - L
```

1. Create a long hierarchy of nested thoughts.
2. Scroll so that  some of the thoughts are at the bottom center of the viewport where the alert message will appear.
3. Initiate drag on a thought in a position where the alert message will appear.

## Current Behavior

iOS: The drag start selecting text in the alert and highlighting elements on the screen.

https://github.com/user-attachments/assets/3d248025-18a3-4281-975b-5ea6e5a5f014

Android: The appearance of the alert message ends the long press, activating multiselect.

## Expected Behavior

The drag should be unaffected by alert text.

ethan-james:
Fixes #3012

The multiselect alert (which may have been superseded by the new multiselect drawer) and the import file alert both have "Cancel" buttons, so it may be necessary to allow interaction with some alerts.

The drag-and-drop alert, however, will only appear during drag interactions, and must be prevented from interfering with the pointer. The new `interactable` flag makes the alert ignore pointer events.

### DIFF
--- a/src/components/Notification.tsx
+++ b/src/components/Notification.tsx
@@ -10,6 +10,8 @@ import PopupBase from './PopupBase'
 const Notification: FC<
   {
     icon?: ReactNode
+    /** A flag specifying whether the notification has interactable elements, or whether it should be ignored by the pointer or mouse. */
+    interactable?: boolean
     /** The content rendered with padding in the center of the notification. */
     value: ReactNode | null
     /* Specify a key to force the component to re-render and thus recalculate useSwipeToDismissProps when the alert changes. Otherwise the alert gets stuck off screen in the dismiss state. */
--- a/src/components/PopupBase.tsx
+++ b/src/components/PopupBase.tsx
@@ -24,6 +24,8 @@ export type PopupBaseProps = PropsWithChildren<
     circledCloseButton?: boolean
     /** If true, the popup will take up the full width of the screen. */
     fullWidth?: boolean
+    /** If false, the popup will not interact with the pointer or mouse. */
+    interactable?: boolean
     /** If defined, will show a small x in the upper right corner. */
     onClose?: () => void
     padding?: string
@@ -44,6 +46,7 @@ const PopupBase = React.forwardRef<HTMLDivElement, PopupBaseProps>(
       children,
       circledCloseButton,
       fullWidth = false,
+      interactable = true,
       onClose,
       padding,
       showXOnHover,
@@ -103,6 +106,8 @@ const PopupBase = React.forwardRef<HTMLDivElement, PopupBaseProps>(
         }
       : {}
 
+    const pointerEventsStyles = interactable ? {} : { pointerEvents: 'none' }
+
     return (
       <div
         className={css({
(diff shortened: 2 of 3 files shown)


src/components/Notification.tsx
```diff
@@ -10,6 +10,8 @@ import PopupBase from './PopupBase'
 const Notification: FC<
   {
     icon?: ReactNode
+    /** A flag specifying whether the notification has interactable elements, or whether it should be ignored by the pointer or mouse. */
+    interactable?: boolean
```
ethan-james: This may be prop drilling, but it seems to me that props for `Notification` and `PopupBase` are more presentational, and that checking `AlertType` belongs in the top-level `Alert` component.

trevinhofmann: I don't think two levels of props is excessive drilling. Would it make sense to add it to the picked props below, though, so that we don't need to redeclare it here?

```
  } & Pick<ComponentProps<typeof PopupBase>, 'onClose' | 'textAlign' | 'onMouseOver' | 'onMouseLeave'  | 'interactable'>
```

src/components/PopupBase.tsx
```diff
@@ -24,6 +24,8 @@ export type PopupBaseProps = PropsWithChildren<
     circledCloseButton?: boolean
     /** If true, the popup will take up the full width of the screen. */
     fullWidth?: boolean
+    /** If false, the popup will not interact with the pointer or mouse. */
+    interactable?: boolean
```
trevinhofmann: I don't love the name `interactable` since it's such an uncommon word ([and missing from many dictionaries](https://www.merriam-webster.com/dictionary/interactable)). Did you have any other ideas? Perhaps `interactive`? We could also use consider something less abstract such as `pointerEventsEnabled`.

raineorshine:
Just to confirm, does this still allow [useSwipeToDismiss](https://github.com/cybersemics/em/blob/c86768c18517c1f2702c096ca8da72687d0d02d8/src/hooks/useSwipeToDismiss.ts)? All alerts should be dismissable.

ethan-james:
> Just to confirm, does this still allow [useSwipeToDismiss](https://github.com/cybersemics/em/blob/c86768c18517c1f2702c096ca8da72687d0d02d8/src/hooks/useSwipeToDismiss.ts)? All alerts should be dismissable.

This alert only exists during the drag and fades out as soon as drag is released. Here I am trying to dismiss it while dragging on `main`:

https://github.com/user-attachments/assets/027ac337-1df2-4e12-a51a-85f54b317c2d

It doesn't seem to respond in any direction. To clarify, this is during the drag so the `move` handler in `useSwipeToDismiss` isn't firing. I see that `move` also isn't firing in the branch for #3011.

src/components/PopupBase.tsx
ethan-james: Yes, I think you're right, `pointerEventsEnabled` is better than trying to describe the functionality more abstractly. This fix has a very specific purpose that may not end up being applicable anywhere else.

ethan-james:
I left this open in case we're not done with it yet, but #3024 uses the new and improved CSS method, and should be tested/reviewed instead of this one.

PR #3023, opened 2025-06-11
Set draggable to true and disable HTML5Backend on iOS Safari

### LINKED ISSUE #2953: [iOS] Disable magnifying glass (loupe) on long press

ethan-james:
Fixes #2953 , #2931 and #2964

Third time's the charm, I hope. Having `draggable={true}` on an element is what disables native long press behavior in iOS Safari. However, the presence of `react-dnd`'s `HTML5Backend` was setting `draggable` to false during drag interactions. Now it will use only `TouchBackend` on touch devices.


ethan-james:
This should be ready for rigorous testing. One thing I noticed is that the cursor moves to the drag-and-dropped thought after drag-and-drop is complete. I don't remember that being the case previously, but I tested in `main` and saw that it's happening there as well.

src/components/DragAndDropContext.tsx
```diff
@@ -2,26 +2,30 @@ import { FC, PropsWithChildren } from 'react'
...
-    // https://react-dnd.github.io/react-dnd/docs/backends/touch
-    {
-      id: 'touch',
-      backend: TouchBackend,
-      options: { delayTouchStart: TIMEOUT_LONG_PRESS_THOUGHT },
-      preview: true,
-      transition: TouchTransition,
-    },
+    // If HTML5Backend is active, it will override a thought's draggable attribute and set it to false.
+    // On iOS Safari, this enables undesirable native long press behavior (#2953, #2931, #2964)
```
raineorshine: [react-dnd-multi-backend](https://github.com/LouisBrunner/dnd-multi-backend/tree/main/packages/react-dnd-multi-backend) is *supposed* to gracefully change backends once touch is detected. They have a whole pipeline concept that you can see in their docs. If what you're saying is true, it would imply that `react-dnd-multi-backend` is not working as intended. Do you think that’s true?

If you could look at `react-dnd-multi-backend`'s issues to see if this has been reported before, that might give us more information. If you don't find anything, please report an issue and link back to here. It's worth letting the maintainer know, and perhaps they can enlighten us if we're missing something. 

I think this deserves some additional due diligence since it raises questions about `react-dnd-multi-backend` and there is some risk that we are not understanding how to use it correctly. If the suggested change is correct and we are simply using `isTouch` to determine the backend, then there is no need for `react-dnd-multi-backend` at all.

ethan-james: > [react-dnd-multi-backend](https://github.com/LouisBrunner/dnd-multi-backend/tree/main/packages/react-dnd-multi-backend) is _supposed_ to gracefully change backends once touch is detected. They have a whole pipeline concept that you can see in their docs. If what you're saying is true, it would imply that `react-dnd-multi-backend` is not working as intended. Do you think that’s true?
> 
> If you could look at `react-dnd-multi-backend`'s issues to see if this has been reported before, that might give us more information. If you don't find anything, please report an issue and link back to here. It's worth letting the maintainer know, and perhaps they can enlighten us if we're missing something.
> 
> I think this deserves some additional due diligence since it raises questions about `react-dnd-multi-backend` and there is some risk that we are not understanding how to use it correctly. If the suggested change is correct and we are simply using `isTouch` to determine the backend, then there is no need for `react-dnd-multi-backend` at all.

I see this: https://github.com/LouisBrunner/dnd-multi-backend/issues/9#issuecomment-355001660 and then a little further down I see:

> I was able to get around that issue simply by listing the touch-backend first and the html5 backend second in the backends array.

I'm a little ambivalent about whether that is a suitable fix or a workaround for buggy behavior. Let me know if you have an opinion, and I'll revisit it tomorrow morning.

src/components/Editable.tsx
```diff
@@ -524,7 +524,7 @@ const Editable = ({
       // If editing or the cursor is on the thought, allow the default browser selection so the offset is correct.
       // Otherwise useEditMode will programmatically set the selection to the beginning of the thought.
       // See: #981
-      if (editingOrOnCursor) {
+      if (editingOrOnCursor && !hasMulticursor) {
```
ethan-james, line 530: This seems to fix the multicursor double tap issue where it opens the keyboard.

src/components/DragAndDropContext.tsx
ethan-james: I followed their advice about putting `TouchBackend` before `HTML5Backend`, and it seems to work fine. However, it has caused a few Puppeteer tests to fail in ways that appear consistent rather than flaky. They're all very small layout differences, and I don't immediately see the cause because the editables seem to have the same dimensions in this branch as they do on `main`. I can take a closer look tomorrow, unless you'd rather have me revert the change and potentially eliminate `multi-backend` entirely.

raineorshine: Hmmm I see the issue, but I don't know off the top of my head the best way to handle the test regression. I can confirm that the puppeteer tests are now failing locally.

raineorshine (changes requested):
Confirmed that this fixes all related issues 👍

ethan-james:
I verified that simply commenting out `TouchBackend` from `DragAndDropContext` fixes the test, while removing `draggable` from `Editable` does not. I also went through `TouchBackendImpl.ts` and `MultiBackendImpl.ts` to see if they're doing anything that would affect the layout, but unlike `HTML5BackendImpl.ts`, they are not updating any attributes. They are merely adding/removing event listeners, storing references to things, and hooking into `react-dnd` internals that I don't understand.

 I see about 3 paths forward:

1. Update the test snapshots to adjust for the tiny discrepancy and move on
2. Remove `MultiBackend` and switch between backends ourselves
3. Try to develop a minimum reproducible test case, either to submit to `dnd-multi-backend` as an issue, or simply to figure out for ourselves why this is happening

I think I prefer option #2, but I'm open to any of them. For what it's worth, there are only 2 tests failing in CI, but there are 4 failing locally because 2 of them have already been skipped using `testIfNotCI`. All 4 of them are quite similar in that the final thought (collapsed in all 4 cases, I think) is shifted by a pixel.

raineorshine:
> I verified that simply commenting out `TouchBackend` from `DragAndDropContext` fixes the test, while removing `draggable` from `Editable` does not.

It's crazy to me that `TouchBackend` affects a puppeteer test that has nothing to do with touch devices!

> I also went through `TouchBackendImpl.ts` and `MultiBackendImpl.ts` to see if they're doing anything that would affect the layout, but unlike `HTML5BackendImpl.ts`, they are not updating any attributes. They are merely adding/removing event listeners, storing references to things, and hooking into `react-dnd` internals that I don't understand.

Okay, thanks for checking.

> 2\. Remove `MultiBackend` and switch between backends ourselves

Would this be as simple as switching on `isTouch`? We already use `isTouch` extensively across the codebase, so I think it's fine to rely on. I'm starting to think the multi backend library is a liability.

I'm also fine with updating the snapshots since it appears to be an imperceptible change. We'll just need to confirm by following the steps of the failing tests manually to make sure.

ethan-james:
> It's crazy to me that `TouchBackend` affects a puppeteer test that has nothing to do with touch devices!

I know! I guess there are a lot of different moving parts, but it's still pretty unexpected.

> Would this be as simple as switching on `isTouch`? We already use `isTouch` extensively across the codebase, so I think it's fine to rely on. I'm starting to think the multi backend library is a liability.

It should be that simple, although the issue I was reading (https://github.com/LouisBrunner/dnd-multi-backend/issues/9#issuecomment-464918494) mentions supporting devices that have a mouse and a touchscreen, the way certain laptops do. I don't think that was being supported by our previous configuration, but I don't have a touchscreen laptop to use to test.

> I'm also fine with updating the snapshots since it appears to be an imperceptible change. We'll just need to confirm by following the steps of the failing tests manually to make sure. I usually open an incognito window with the staging server in the same position a window with localhost and then toggle between them to ensure there is no visual change. It's also good to ensure there isn't any shift that happens to the animation, as that can be another source of rendering inconsistency.
> 
> Let me know if you'd like me to do this testing, I'd be happy to.

I can give it a try, although I imagine you'll want to check and make sure it looks the way you want. Is there an updated URL for staging?

raineorshine:
> > Would this be as simple as switching on `isTouch`? We already use `isTouch` extensively across the codebase, so I think it's fine to rely on. I'm starting to think the multi backend library is a liability.
> 
> It should be that simple, although the issue I was reading ([LouisBrunner/dnd-multi-backend#9 (comment)](https://github.com/LouisBrunner/dnd-multi-backend/issues/9#issuecomment-464918494)) mentions supporting devices that have a mouse and a touchscreen, the way certain laptops do. I don't think that was being supported by our previous configuration, but I don't have a touchscreen laptop to use to test.

Yes, I belatedly thought of that. As it stands, `isTouch` is deeply embedded in **em**, so dual touch + mouse support is out of scope.

> > I'm also fine with updating the snapshots since it appears to be an imperceptible change. We'll just need to confirm by following the steps of the failing tests manually to make sure. I usually open an incognito window with the staging server in the same position a window with localhost and then toggle between them to ensure there is no visual change. It's also good to ensure there isn't any shift that happens to the animation, as that can be another source of rendering inconsistency.
> > Let me know if you'd like me to do this testing, I'd be happy to.
> 
> I can give it a try, although I imagine you'll want to check and make sure it looks the way you want. Is there an updated URL for staging?

https://em-proto--staging2-cxls71a0.web.app

ethan-james:
Here is a video demonstrating the bug that I just fixed. It's a little hard to describe.


https://github.com/user-attachments/assets/f1a750ae-017e-43e5-aa37-de9d052eb068

raineorshine (approved):
Thank you!!!

PR #3025, opened 2025-06-12
Move userSelect: none into Content

### LINKED ISSUE #2895: [iOS] Selection activated on tap and hold in scroll zone
## Current Behavior

With no thoughts in the thoughtspace, tap and hold on the scroll zone. 

image

image

With one or more thoughts in the thoughtspace, tap and hold on the scroll zone. 

image

## Expected Behavior

The selection should not be activated. 

- Try a CSS only solution is possible, such as disabling `user-select` or `pointer-events` on the appropriate element. 
- Avoid scroll zone detection or tap coordinates. 
- Make sure drag-and-drop behavior still works.

### LINKED ISSUE #2961: [iOS] Drag-and-drop on bullet highlights bullet, scroll area, etc.

ethan-james:
Fixes #2895

The div in `LayoutTree` that had `userSelect: none` applied to it to fix #2961 doesn't cover the entire viewport, and is sized differently when there are thoughts in the thoughtspace vs when it is empty. Applying `userSelect: none` instead to the `Content` wrapper solves this and doesn't appear to impact anything else that should be selectable.

### DIFF
--- a/src/components/Content.tsx
+++ b/src/components/Content.tsx
@@ -76,6 +76,10 @@ const Content: FC = () => {
           maxWidth: '60em',
           margin: '0 auto',
           minHeight: '100vh',
+          // Disallow text selection on the entire tree to prevent selection of scroll zone and other background elements when dragging.
+          // This is overriden by the editable recipe to enable text selection on editable thoughts.
+          // https://github.com/cybersemics/em/pull/2962
+          userSelect: 'none',
           zIndex: 'content',
           '@media (max-width: 960px)': {
             maxWidth: '80%',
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Works like a charm

PR #3026, opened 2025-06-12
Set cursor to previous thought after deleteEmptyThought

### LINKED ISSUE #3009: Selection is lost after deleting a favorite
## Steps to Reproduce

```
- a
- b
```

1. Set the cursor on `b`
2. Add to Favorites
3. Set the caret at the end of `b`.
4. Hit backspace twice to delete `b`.

## Current Behavior

Browser selection is cleared.

## Expected Behavior

Browser selection should move to the end of `a`, as when `b` is not a favorite.

ethan-james:
Fixes #3009 

One of the branches in `deleteEmptyThought` had a note: `// @MIGRATION-TODO: Set proper cursor here`.

I called `setCursor` on the previous thought. It's possible that this was left unfinished due to additional complexities in `deleteEmptyThought` that I don't understand, but it seems to fix this bug just fine.

When deleting a favorited thought, it is deleting the `=favorite` attribute that has a child containing the visible thought `b`. The presence of the child thought causes it to skip `deleteThoughtWithCursor` and use the `reducerFlow` beneath that in `deleteEmptyThought.ts`. This branch of the logic was not fully migrated.


src/actions/deleteEmptyThought.ts
```diff
@@ -81,10 +83,16 @@ const deleteEmptyThought = (state: State): State => {
         pathParent: parentOf(cursor),
         thoughtId: head(cursor),
       }),
-      // @MIGRATION-TODO: Set proper cursor here
```
raineorshine, line 84: We likely want to use https://github.com/cybersemics/em/blob/b89549e52b09ea6a9001773492e73a75fac3319e/src/actions/updateCursorAfterDelete.ts. The cursor logic after a delete is a bit more complex than `prev` as it depends on whether we are deleting the first thought, last thought, etc.

`updateCursorAfterDelete` probably didn't exist when the migration comment was added, but it has become a robust way to update the cursor properly after any thought is deleted.

ethan-james:
> For my understanding, can you make explicit why the bug occurs when deleting a favorite, but not when deleting a normal empty thought? Apparently it there is a different code pathway if the empty thought is a leaf, but it would be helpful to point that out in the initial explanation.

I added the explanation to the issue description.

PR #3034, opened 2025-06-15
Reverse context view animation for Safari

### LINKED ISSUE #2529: Animation: Toggle Context View

zukilover:
Fix #2529

### DIFF
--- a/src/recipes/fadeTransition.ts
+++ b/src/recipes/fadeTransition.ts
@@ -1,5 +1,9 @@
 import { defineSlotRecipe } from '@pandacss/dev'
 
+/*
+  Safari skew: Safari requires inverted skew values (positive instead of negative) when exceeding 90deg to have the same behavior as other browsers.
+*/
+
 const fadeTransitionRecipe = defineSlotRecipe({
   className: 'fade',
   slots: ['enter', 'exit', 'exitActive', 'enterActive', 'enterDone', 'exitDone'],
@@ -64,12 +68,11 @@ const fadeTransitionRecipe = defineSlotRecipe({
         },
       },
       // Context view fades in from upper right.
-      // Safari requires inverted skew values (positive instead of negative) when exceeding 90deg
-      // Uses _safari condition to target Safari-specific styles via [data-browser=safari] selector
       disappearingUpperRight: {
         enter: {
           transform: 'skew(-100deg) translateX(10%) translateY(-100%)',
           textShadow: '0px 0px 2em {colors.fg}',
+          // See: Safari skew note above
           _safari: {
             transform: 'skew(100deg) translateX(10%) translateY(-100%)',
           },
@@ -78,6 +81,7 @@ const fadeTransitionRecipe = defineSlotRecipe({
           transform: 'skew(0) translateX(0) translateY(0)',
           textShadow: '0px 0px 0px {colors.fg}',
           transition: `opacity {durations.disappearingUpperRight} {easings.easeInSmooth}, transform {durations.disappearingUpperRight} {easings.easeInSmooth}, text-shadow {durations.disappearingUpperRight} {easings.easeInSmooth}`,
+          // See: Safari skew note above
           _safari: {
             transform: 'skew(0) translateX(0) translateY(0)',
           },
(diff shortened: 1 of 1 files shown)


zukilover:
I've fixed the Safari issue by adding new variant to the animation.

<details>
<summary>Safari</summary>


https://github.com/user-attachments/assets/f2cbf789-4f4f-40be-9c2e-654b1c6fd5e0





</details>

src/durations.config.ts
```diff
@@ -37,6 +37,8 @@ const durationsConfig = {
   /* Duration for context view disappearing text animations */
   disappearingUpperRight: 500,
   disappearingLowerLeft: 500,
+  disappearingLowerLeftSafari: 500,
+  disappearingUpperRightSafari: 500,
```
raineorshine: Unused?

src/recipes/fadeTransition.ts
```diff
@@ -68,11 +68,17 @@ const fadeTransitionRecipe = defineSlotRecipe({
         enter: {
           transform: 'skew(-100deg) translateX(10%) translateY(-100%)',
           textShadow: '0px 0px 2em {colors.fg}',
+          _safari: {
+            transform: 'skew(100deg) translateX(10%) translateY(-100%)',
+          },
```
raineorshine, line 78: When code is non-obvious, please leave a comment for future developers to make sense of it without having to do a `git blame` and retrace the history.

raineorshine (changes requested):
Confirmed that this works as expected in Safari now!

src/recipes/fadeTransition.ts
```diff
@@ -64,15 +64,23 @@ const fadeTransitionRecipe = defineSlotRecipe({
         },
       },
       // Context view fades in from upper right.
+      // Safari requires inverted skew values (positive instead of negative) when exceeding 90deg
```
raineorshine: I slightly prefer placing the comment above `_safari` since it only applies to that section, not all of `disappearingUpperRight`.

src/recipes/fadeTransition.ts
```diff
@@ -64,15 +64,23 @@ const fadeTransitionRecipe = defineSlotRecipe({
...
+      // Uses _safari condition to target Safari-specific styles via [data-browser=safari] selector
```
raineorshine: The second line of the comment is okay, but it's not strictly necessary, since:

1) `_safari` does exactly what it appears to. It's obvious and intuitive, not needing explanation.
2) The comment merely explains how PandaCSS works.

There's no harm in it, but it's a good example of the contrast between the first line of the comment (non-obvious behavior) and the second line of the comment (obvious behavior).

src/recipes/fadeTransition.ts
```diff
@@ -64,15 +64,23 @@ const fadeTransitionRecipe = defineSlotRecipe({
...
           textShadow: '0px 0px 2em {colors.fg}',
+          _safari: {
+            transform: 'skew(100deg) translateX(10%) translateY(-100%)',
+          },
         },
         enterActive: {
           transform: 'skew(0) translateX(0) translateY(0)',
           textShadow: '0px 0px 0px {colors.fg}',
           transition: `opacity {durations.disappearingUpperRight} {easings.easeInSmooth}, transform {durations.disappearingUpperRight} {easings.easeInSmooth}, text-shadow {durations.disappearingUpperRight} {easings.easeInSmooth}`,
+          _safari: {
```
raineorshine, line 85: Here we have a tradeoff between duplicating the comment, and risking that future developers will look at this code, not understand what it does, and fail to see the other comments in the file.

Let's make it easy for future developers! We can put the comment right next to the relevant code.

If the comment is very large and you don't want to duplicate it, add it to the top of the file, and then refer back to it with a comment like "See explanation of 'Safari skew' at top of file."

PR #3039, opened 2025-06-15
Table: row height

### LINKED ISSUE #1792: Table row height is incorrect

ohrytskov:
Prevents unnecessary vertical offset from col1 ancestors without visible children, resolving layout issues when switching between Tree and Table views.

Closes #1792.

### DIFF
--- a/src/hooks/usePositionedThoughts.ts
+++ b/src/hooks/usePositionedThoughts.ts
@@ -188,8 +188,8 @@ const usePositionedThoughts = (
 
       // if the current thought is in table col1, push its y and depth onto the stack so that the next node after it can be positioned below it instead of overlapping it
       // See: ycol1Ancestors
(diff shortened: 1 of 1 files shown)


ohrytskov:
The buggy behavior came from pushing yaccum + height even for empty rows, which caused too much vertical offset to be accumulated for following rows. The fix makes sure to push the Y position of col1 only when it has visible children, which prevents this unnecessary spacing.

Issue #2939, opened 2025-05-12
Multiline thought height is wrong on categorize/uncategorize


raineorshine:
This can be reproduced on desktop either using Chrome Device Toolbar or by resizing the window so it is narrow.

Related:

- https://github.com/cybersemics/em/blob/1136c42debccb4414b03a38c3020c1278c3bd7cf/src/components/Editable/useMultiline.ts

## Steps to Reproduce

```
- Is Co-Text Enough?
  - ✗
    - *The fact that that LLM-based chatbots do not initiate third-position repair… **is further indication that they do not model intentions or other mental states.*
      - It’s not really evidence either way
      - They can clearly model intentions or other mental states to the extent that those are part of the co-text
      - It is more a reflection of the assumed understanding of progressivity and the probably of confusion or registered (acknowledged) ambiguity within the training pairs
      - It’s not clear how much clarification questions have been trained into or out of models
      - Most ChatGPT conversations are “fresh” conversations (not very deep) that have a higher need for ambiguity resolution than the content it was trained on
```

1. Set the cursor on the first thought. 
2. Activate `categorize` (Ctrl/⌘ + O)

If you move the cursor away, then the thoughts will re-render and the spacing is fixed.

3. Activate `uncategorize` (Ctrl/⌘ + C) on the empty thought and the height will be wrong again, but this time leaving too much space. 

## Current Behavior

The third and fourth thoughts overlap:

https://github.com/user-attachments/assets/1fac303b-7d7e-466e-9125-48210e8626f0

## Expected Behavior

Thought heights should be updated and re-rendered whenever the thought structure changes.

## Notes

Make sure the solution does not have a negative performance impact, i.e. careful not to introduce additonal React renders.

PR #3047, opened 2025-06-17
Fix issue of multiline thought height on categorize/uncategorize

codbuild-er:
**Fix: Multiline thought height is wrong on categorize/uncategorize #2939**

It was the delay of rendering of element.
So I used `requestAnimationFrame` function to wait for next paint.
It is for detect the difference between initial rendering and post-rendering.
If there is difference between first render and later render, it will update the state again to calculate the position correctly.


raineorshine (changes requested):

I'm wary about this solution for a few reasons.

1. It's not clear why an extra measurement is required in this case, when `useLayoutEffect` works as expected for cursorDepth, cursorLeaf, fontSize, isVisible, leaf, note, simplePath, style, isContextViewActive, editingValue, and updateSize.
2. The purpose of `useLayoutEffect` is to ensure that the height is correct on the first paint. The dimensions of elements are generally available at this time.
3. Calling `getBoundingClientRect()` twice on every effect is going to clobber performance. This is an extremely performance sensitive app, so we can't afford to do this.

Please take these into account and let me know what you can come up with. Thanks!

raineorshine (changes requested):
Thank you for the update. This solves the issue nicely based on initial testing. 

Unfortunately, I found a regression on mobile Safari only. (Tested and unable to reproduce on desktop Safari or with Chrome Device Toolbar.)

This has happened to me in the past when working on this problem, where a fix in one case breaks another case. Challenging!

## Steps to Reproduce

1. Create though `a`. 
2. Activate New Thought Above by swiping →↑←.
3. Activate Cursor Back by swiping ← to set the cursor to null. 

## Current Behavior

8669ca2a7c

The new thought's height is too too tall (Step 2) and is fixed on the next action that triggers a recalculation (Step 3). 

https://github.com/user-attachments/assets/c8f892ec-a0d8-4c01-a9d0-7765f5b48f6f

## Expected Behavior

`main` or [staging](https://em-proto--staging2-cxls71a0.web.app)

Thoughts should be rendered in the correct position with the correct height, and not shift on the next action. 

https://github.com/user-attachments/assets/be322a2b-ab7e-4b0d-a7af-98027aa373a8

raineorshine:
> Actually, the height changes instantly due to the width being set manually based on the depth. Otherwise we would see the text reflow gradually over the course of the animation.

FYI here is the code I am referring to (may not be relevant any more): 

https://github.com/cybersemics/em/blob/f06dad9901ddcff1e0a9c91933e350a89f3a811a/src/components/LayoutTree.tsx#L254-L256

src/components/VirtualThought.tsx
```diff
@@ -42,6 +43,8 @@ const selectShowContexts = (path: SimplePath) => (state: State) => isContextView
 /** Selects the cursor from the state. */
 const selectCursor = (state: State) => state.cursor
 
+const isIOSSafari = isTouch && isiPhone && isSafari()
```
raineorshine: I don't believe we need to check `isiPhone`, as it is assumed with `isTouch && isSafari()`.

Also might yield a false negative for iPads.

src/components/VirtualThought.tsx
```diff
@@ -158,7 +178,9 @@ const VirtualThought = ({
 
   // Recalculate height when anything changes that could indirectly affect the height of the thought. (Height observers are slow.)
   // Autofocus changes when the cursor changes depth or moves between a leaf and non-leaf. This changes the left margin and can cause thoughts to wrap or unwrap.
-  useLayoutEffect(updateSize, [
```
raineorshine, line 161: Let's add a comment here explaining the interaction between `useLayoutEffect` and `requestAnimationFrame`. There are a couple things that are nonobvious:

1. `useLayoutEffect` is typically used to perform a calculation synchronously, so it's unusual to combine it with an async delay.
2. `useEffect` can be delayed multiple frames and cause a flicker.
3. `useLayoutEffect` + `requestAnimationFrame` recalculates the size in the first frame (or second frame on iOS Safari) after a UI update, which is long enough to ensure the correct height is captured, but not long enough to introduce a filter.

src/components/VirtualThought.tsx
```diff
@@ -129,7 +132,24 @@ const VirtualThought = ({
   //   childPath,
   // })
 
-  const updateSize = useCallback(() => {
+  const updateSize = useCallback(async () => {
+    // Check dimensions again after a delay to catch any post-render changes
+    // Use requestAnimationFrame to wait for next paint, then check again after a frame to catch any post-render changes
+    // Wait for next frame
+    await new Promise(resolve => requestAnimationFrame(resolve))
```
raineorshine: `updateSize` is called in a `useLayoutEffect`, `useEffect`, `viewportStore.useSelectorEffect`, and all the way down to [another `useLayoutEffect` in StaticThought](https://github.com/cybersemics/em/blob/00ec9feb33f9ff42e98789c48b84bf0191d060f7/src/components/StaticThought.tsx#L125-L127) through prop drilling. Are you sure that we want to delay each of these with `requestAnimationFrame`? It seems unlikely that the `useEffect`, for example, but need another delay.

My preference would be to add the `requestAnimationFrame` only where needed (i.e. the two `useLayoutEffect`) in order to minimize the impact of this change. If we find any false negatives, we can always expand it to other call sites or hoist it back up to `updateSize`, but better to start narrow. What do you think?

raineorshine (changes requested):
Thanks so much for trusting your instincts and sticking it out. We're back to a single call which is great for performance, and as far as I can tell the additional delay on iOS is not creating any noticeable lag since the calculation is only one frame slower.

---

I am seeing a bunch of failures in the puppeteer tests. Could you investigate? Even 1px differences *can* potentially be visual regressions, so they deserve a closer look.

One of several possibilities may be the case:

1. The test snapshots *may* need an additional `requestAnimationFrame` to match the one that was added.
2. The layout *may* have shifted in a way that is imperceptible to the user and the snapshots need to be updated (this requires extreme caution)
3. There *may* be an actual visual regression that needs to be fixed in the code, and the tests themselves are fine.

`yarn test:puppeteer` (docker required)

src/components/VirtualThought.tsx
codbuild-er: I updated it as `const isIOSSafari = isTouch && isSafari()`.
Just removed `isiPhone` variable.

src/components/VirtualThought.tsx
codbuild-er: I made `updateSizeAfterLayout` function to combine with useLayoutEffect, and for other hooks, it uses `updateSize` function as before.

raineorshine (approved):
Thank you! Everything looks good on my end. Tested on Chrome and mobile Safari.

The large number of snapshot upgrades are unexplained, yet I have tested thoroughly and it does not appear to impact the app during real-world usage. The tests pass consistently with the new snapshots on localhost and in the CI. If there is a 1px drift, it's deterministic and occurs within the animation where it is not noticeable. The final position of the thoughts are stable and consistent, with nothing shifting on navigation or after the animation completes.

Interestingly enough, I *was* able to reproduce the inconsistency between `main` and this PR. This tells us that it is not *only* a test issue. However, it is only noticeable when comparing them when perfectly aligned. The position is entirely consistent within a single app instance.

(The incognito window is `main`, and the normal window is this PR.)

https://github.com/user-attachments/assets/1dba7f50-10cc-4874-bd00-1f709a7d470f

No noticeable performance issues... at least none significant enough to block merge and not deal with in a separate optimization step later.

Issue #2845, opened 2025-03-11
Multiline thought vertical alignment is wrong


raineorshine:
## Steps to Reproduce

```
- Ryoan-ji (龍安寺), a Zen Buddhist temple in Kyoto, was founded in 1450 and is renown
- a
```

1. Resize the window so that the first thought just barely fits on one line.
2. Add `e` to the end of the thought so it wraps onto two lines.

## Current Behavior

The first line of text shifts up.

## Expected Behavior

The first line of text should not shift.

## Testing

Snapshot test.

raineorshine:
Hold for #3598

BayuAri:
@raineorshine 
Multiline issue is still reproducible on bdcf7108b5990e0cd2fec5bde7c8bebc142c31c6
The line shifts when the text wraps.
This happens on the first 4 lines of the thought. 
The line does shift on 5th lines of the thought and so on.
This is only reproducible on iOS Capacitor, it is not reproducible on Android, Mobile Chrome, desktop and Safari

https://github.com/user-attachments/assets/306c8156-dc37-4508-8884-99b1719cae97

raineorshine:
@BayuAri Thanks. Can you confirm whether this is a scroll shift or a rendered layout shift? If it's only a scroll shift, then it's a different issue.

Scroll back to the top after each shift and only compare when scrollY is at 0. You will probably need to compare static screenshots if there is any scroll involved, since a 1px shift will not easily be detectable with intervening scrolling.

BayuAri:
@raineorshine 

With the help of copilot, here the result: 
Tested on 881586e6549b901c102d9a12b8029bac9c67998e

>Confirmed it's a rendered layout shift, not only a scroll shift. Measuring the first line's scroll-independent document position ([getBoundingClientRect().top + scrollY](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)) in the Capacitor WKWebView: when the thought wraps to 2 lines, that absolute position moves ~1.77px with scrollY and visualViewport both unchanged — so the line genuinely re-renders higher (this is #2845). Separately, I also see repeated 1px scroll shifts (scrollY/visualViewport +1px while the absolute layout position stays fixed) — those look like a distinct keyboard/autoscroll issue, matching your note that a pure scroll shift would be a different issue.

ethan-james:
[...]

PR #3048, opened 2025-06-17
Move preventAutoscrollEnd in onFocus into setTimeout

### LINKED ISSUE #2969: [iOS] Caret is invisible or obscured on multiline thought
Tested on Mobile Safari. 

## Steps to Reproduce

```
- A
  - B
    - C
      - This is a longer thought that keeps going and going
```

1. Set the cursor to null and refresh the app. 
2. Set the cursor on the multiline thought. 
3. Tap the multiline thought near `and`. 
4. Close the keyboard. 
5. Tap the multiline thought near `and`. 

## Current Behavior

The first time the keyboard is opened, the caret is invisible. (It becomes visible again if you type or scroll.)

The second time the keyboard is opened, the caret is partially cut off. 

image

## Expected Behavior

The caret should be completely visible when the keyboard is open.

ethan-james:
Fixes #2969

It looks like if I remove `preventAutoscrollEnd` from `onFocus` in `Editable.tsx`, then the bug goes away. `preventAutoscrollEnd` runs after a 10ms timeout initiated in `preventAutoscroll`, there is the "noticeable blink" described in the comment, and the cursor shows correctly.

If I can figure out how to fix the blink, then this approach could work well.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -493,7 +493,7 @@ const Editable = ({
    */
   const onFocus = useCallback(
     () => {
-      preventAutoscrollEnd(contentRef.current)
+      setTimeout(() => preventAutoscrollEnd(contentRef.current))
       if (suppressFocusStore.getState()) return
       // do not allow blur to setEditingValue when it is followed immediately by a focus
       blurring = false


ethan-james:
I know that there are going to be some questions about why `paddingTop` on `Editable`, as defined in `multilineRecipe`, was entirely responsible for hiding the caret. Here is as much information as I could figure out.

Clearly, `paddingTop` is overridden in `preventAutoscroll`, so it makes sense that it has some association with the `paddingTop` defined in `multilineRecipe`. However, removing or modifying ``el.style.paddingTop = `${-yCenter * 2}px``` in `preventAutoscroll` does not help.

Removing `!important` from the `base` and `_safari` styles in `multilineRecipe` will also fix it, even without removing `!important` from the `_mobile: { _safari }` directive. So if it looks like this, and is applied to `Editable`, then it works:

```
    paddingTop: {
      base: '0.33em',
      _safari: '0.3em',
      _mobile: {
        /* Hacky, nonlinear multiline padding-top that works across all font sizes on mobile Safari. Otherwise the thought goes out of alignment with the bullet when it becomes multiline. */
        /* 0.2em works at font size 13 */
        /* 0.24em works at font size 18 */
        /* works without max at font sizes 18–24 */
        _safari: 'calc(max(0.5em, 7.5px) - 5px) !important',
      },
    },
```

These changes also cause some tests to fail, so additional tweaks would need to be made in order to keep `multilineRecipe` on `Editable` rather than moving it to `StaticThought` as I have done.

I tried Googling now that I have some information about what to Google for, but I don't see anything relevant.

raineorshine:
Thanks for the explanation. Truly bizarre behavior! I'm so glad that we've honed in on `paddingTop` as the culprit.

Normally we would approach this extremely conservatively, making every effort to preserve the mobile Safari overrides and even the `!important`, as hacky as it is. The Safari override goes all the way back to 2022 (de38a12d77ab9feaad9831c4f9d3cb3bbe1a456a), and `!important` was added in the conversion to PandaCSS (2a9b60280f8b3456d1c38c9611bb286e60cb23ad).

However, the purpose of the override as described in the comment `Hacky, nonlinear multiline padding-top that works across all font sizes on mobile Safari. Otherwise the thought goes out of alignment with the bullet when it becomes multiline.`... doesn't even work! That is, the multiline bullet **is out of alignment** in `main`. This is tracked in #2845.

Therefore, you are welcome to rip out the `!important` and the override as needed. It's already broken and needs to be addressed in a separate PR from scratch.

raineorshine:
I'm not sure if this is what the failed puppeteer tests are picking up on, but I did find one visual regression.

## Current Behavior

The bottom of the thought text ("g") is clipped.

## Expected Behavior

The thought text should not be clipped.

ethan-james:
I think it looked OK on iPhone, and was only really messed up on Chrome. I added `_chrome: '0.33em'` to add specificity to the selector and allow it to take precedence over the top padding set in `editableRecipe`.

raineorshine (changes requested):
Thanks!

I noticed in the new snapshots that the superscript and url icon of multiline thoughts are out of alignment. They are both contained within the `ThoughtAnnotation` which typically overlaps the `Thought` component.

PR #3065, opened 2025-06-24
Don't dispatch setCursorOnThought if dragInProgress

### LINKED ISSUE #3051: Cursor moves to drag-and-dropped thought
## Steps to Reproduce

```
- a
```

1. Clear the cursor
2. Long press `a`
3. Move the pointer enough to trigger drag-and-drop
4. Release the pointer, with or without re-ordering `a`

## Current Behavior

Cursor moves to `a`.

## Expected Behavior

Cursor should remain cleared, or on another thought.

ethan-james:
Fixes #3051

I'm not quite sure why the transition from calling `onTap` on `mousedown` to calling it on `touchend` altered the flow, but now `setCursorOnThought` was being called at the end of a long press. Checking `dragInProgress` before the dispatch (which is delayed until after `useDragHold` has set `state.dragInProgress` to false) allows us to avoid calling `setCursorOnThought` as the result of a long press.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -127,7 +127,6 @@ const Editable = ({
   const nullRef = useRef<HTMLInputElement>(null)
   const contentRef = editableRef || nullRef
   const editingOrOnCursor = useSelector(state => state.isKeyboardOpen || equalPath(path, state.cursor))
-  const dragInProgress = useSelector(state => state.dragInProgress)
 
   // console.info('<Editable> ' + prettyPath(store.getState(), simplePath))
   // useWhyDidYouUpdate('<Editable> ' + prettyPath(state, simplePath), {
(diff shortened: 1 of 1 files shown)


src/components/Editable.tsx
```diff
@@ -571,6 +571,8 @@ const Editable = ({
 
       dispatch((dispatch, getState) => {
         const state = getState()
+        if (state.dragInProgress) return
```
raineorshine, line 574: This is a bit simpler, and avoids re-rendering every Editable when a drag starts or ends.

`useSelector` is preferred for reactive values. If the value is only used in an event handler, then it often doesn't need to be reactive.

raineorshine (approved):
Thank you!

src/components/Editable.tsx
ethan-james: I commented on the code itself, but `dragInProgress` will be false inside the `dispatch`, because it waits a tick and happens after the `useDragHold` event handlers have already run.

raineorshine: It seems to still fix the issue though! 🤔

Tested on mobile Safari.

ethan-james: Yep, looks fine to me too!

PR #3081, opened 2025-06-27
Remove useDisableLongPressToSelect

### LINKED ISSUE #3079: Remove useDisableLongPressToSelect
> I think it's probably possible to remove useDisableLongPressToSelect from AppComponentnow that the draggable attribute should be preventing selection in the first place.

ethan-james:
Fixes #3079

Tested on iPhone, Android and desktop Chrome

Now that editables have `draggable=true`, the native iOS and Android selection behavior is not triggered and we don't need `useDisableLongPressToSelect` to attempt to reset the selection.

### DIFF
--- a/src/components/AppComponent.tsx
+++ b/src/components/AppComponent.tsx
@@ -1,7 +1,7 @@
 import { Capacitor } from '@capacitor/core'
 import { StatusBar, Style } from '@capacitor/status-bar'
 import _ from 'lodash'
-import React, { FC, PropsWithChildren, useCallback, useEffect, useLayoutEffect } from 'react'
+import React, { FC, PropsWithChildren, useEffect, useLayoutEffect } from 'react'
 import { useSelector } from 'react-redux'
 import { WebviewBackground } from 'webview-background'
 import { css } from '../../styled-system/css'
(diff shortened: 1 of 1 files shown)


ethan-james:
While we're refactoring, it occurs to me that these different approaches:

https://github.com/cybersemics/em/pull/2960/files
https://github.com/cybersemics/em/pull/3065/files
https://github.com/ethan-james/em/commit/de6ee0e03b6e34f6f370b470624794cf2f768f91#diff-e60c5e4ff2b203d85c2fbac345b56bdd73231a29ab28ac0426e0a7bccd37b680R130

should be combined a little bit better in order to avoid having similar overrides scattered through different files. I tried briefly to pick one or the other approach and group them together, but didn't have much success. Removing `dragInProgressSafari` causes some weird overscroll behavior:

https://github.com/user-attachments/assets/216f30fb-adf5-4046-b121-7e77aaa4536b

ethan-james:
Also, I went through some recent PRs, thinking that some might have been superseded by the `draggable` and `userSelect: none` work. One that I can't explain yet is https://github.com/cybersemics/em/issues/2753, also referenced in https://github.com/cybersemics/em/pull/2772#discussion_r1917308772. I think that I can remove the `focusNode.focus()` fix from `selection.set`.

raineorshine (approved):
Looks good, thanks!

It doesn't seem like the other comments are blocking, so I will merge this and then respond.

Issue #3080, opened 2025-06-27
Review useDragHold and useLongPress for async event handler improvements


raineorshine:
Let's take a closer look at `useDragHold` and `useLongPress` to see if I can apply the changes that were discarded in https://github.com/cybersemics/em/pull/3011/files#diff-a2cf6931de2064f99f9e1ff56cf6db4d9b85fab750a34d61ea5c956c6a688c3f. The goal is to reduce the number of surprise asynchronous event handlers that run during/after long press. This should help make long press behaviors more predictable and easier to maintain.

ethan-james:
I opened https://github.com/cybersemics/em/pull/3109 and https://github.com/cybersemics/em/pull/3110 with small changes, and commented on https://github.com/cybersemics/em/pull/3081. Plus the closed PR at https://github.com/cybersemics/em/pull/3102.

I think that covers everything I was trying to refactor here, and I think we can close this one.

PR #3086, opened 2025-06-28
Remove focusNode.focus from selection.set

ethan-james:
References #3080 

There are 3 test cases for reproducing bugs that are fixed by calling `focusNode.focus()` in `selection.set`:

https://github.com/cybersemics/em/pull/2752#discussion_r1902247165
https://github.com/cybersemics/em/issues/2705

and then this comment, which I can't find documented elsewhere:

```
  // deleting a note, then closing the keyboard, then creating a new note could result in lack of focus
```

I have verified that all 3 behave normally now without the bug fix.

### DIFF
--- a/src/device/selection.ts
+++ b/src/device/selection.ts
@@ -2,7 +2,6 @@
 
 /** Wraps the browser Selection API in a device-agnostic interface. */
 import SplitResult from '../@types/SplitResult'
-import { isSafari, isTouch } from '../browser'
 
 export type SelectionOptionsType = {
   offset?: number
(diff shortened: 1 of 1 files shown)


ethan-james:
I tried to bisect this to figure out which commit fixed it, but could not reproduce any of the 3 test cases, even when removing `focusNode.focus()` from the original commit that fixed it (c7274bfd30b7778107cafc753d9c1b8ef8211d7a). I also cleared my local storage and started with a blank state in that commit.

My iPhone is on iOS 18.5 (build 22F76) which came out May 12, 2025, a month after this fix was merged into `main`. I don't see anything in particular about WebKit in the iOS 18.5 release notes, but my guess is that the bug was fixed in that update.

raineorshine (approved):
Great, thank you for the thorough testing and due diligence!

> My iPhone is on iOS 18.5 (build 22F76) which came out May 12, 2025, a month after this fix was merged into `main`. I don't see anything in particular about WebKit in the iOS 18.5 release notes, but my guess is that the bug was fixed in that update.

It's rare, but I have had that happen before.

PR #3087, opened 2025-06-28
Remove whitespace from multline thought test

ethan-james:
References #3080 

This is not related to refactoring that we've discussed, but I noticed that the multiline thought test is failing fairly consistently in CI and locally. Many of the tests use whitespace for readability, but I verified that pasting text into desktop Chrome leads to unexpected indentation:

Screenshot 2025-06-28 143446

and I verified that removing the whitespace causes this particular test to pass.

It may be worth revisiting this and digging deeper into the `importText` action, although I will wait for you to weigh in, and I'll most likely work on other stuff first.

### DIFF
--- a/src/e2e/puppeteer/__tests__/render-thoughts.ts
+++ b/src/e2e/puppeteer/__tests__/render-thoughts.ts
@@ -140,12 +140,10 @@ describe('multiline', () => {
   beforeEach(hideHUD)
 
   it('multiline thought', async () => {
-    await paste(`
-        - a
-        - External objects (bodies) are merely appearances, hence also nothing other than a species of my representations, whose objects are something only through these representations, but are nothing separated from them.
-        - b
-        - c
-      `)
+    await paste(`- a
+- External objects (bodies) are merely appearances, hence also nothing other than a species of my representations, whose objects are something only through these representations, but are nothing separated from them.
+- b
+- c`)
 
     const image = await screenshot()
     expect(image).toMatchImageSnapshot()


raineorshine:
Thanks for bringing this up.

The whitespace shouldn't affect the import, so that is likely bug. However, I assume that even the bug would be deterministic. Are you saying that the test intermittently fails with whitespace, and reliably passes without it?

ethan-james:
> Thanks for bringing this up.
> 
> The whitespace shouldn't affect the import, so that is likely bug. However, I assume that even the bug would be deterministic. Are you saying that the test intermittently fails with whitespace, and reliably passes without it?

I think it fails pretty deterministically. I'm just confused about why it's the only test that's affected by whitespace.

raineorshine:
No, I think it must be something else. I just manually triggered the workflow and the multiline thought test failed 3/10 times.

https://github.com/cybersemics/em/actions?query=branch%3Apr%2F3087

ethan-james:
> No, I think it must be something else. I just manually triggered the workflow on [1eb63b5](https://github.com/cybersemics/em/commit/1eb63b5467210b0de0c6ea8fb4da12bc8a8cd93a) and the multiline thought test failed 3/10 times.
> 
> https://github.com/cybersemics/em/actions?query=branch%3Apr%2F3087

Weird. Locally, I deleted the whitespace in front of each line, one at a time, and let the watcher run after each save. It failed until I got rid of all leading whitespace, then it succeeded. Then I hit undo and saved, and it failed again. Then I hit redo and saved, and it succeeded again.

I'll keep an eye on the test, but it's probably just gremlins messing with me.

karunkop:
> Well... that seems like a 100% reasonable conclusion 😅. I don't know what's going on!
> 
> I temporarily disabled the test and created #3088 to track.
> 
> @karunkop Please note the above discussion. It's in your hands now :).

Sure I will look into this soon! :)

Issue #2094, opened 2024-07-01
Table: Cannot drop before first thought in second row


raineorshine:
## Steps to Reproduce

```
- x
- a
  - =view
    - Table
  - =pin
    - true
  - b
    - c
  - d
    - e
```

Drag `x` before `e`. 

## Current Behavior

Can only drag after `c`, not before `e`. 

![image](https://github.com/cybersemics/em/assets/750276/e8991f16-0081-4572-a347-f4a7ceea8c80)

## Expected Behavior

Should be able to drop before `e`. The drop targets for `c` and `e` should be roughly equal size.

raineorshine:
# DropHover of col2 child

## Steps to Reproduce

```
- x
- a
  - =pin
  - =view
    - Table
  - b
    - =archive
      - =pin
    - c
      - =pin
      - d
```

## Current Behavior

The right edge of the blue bar for dropping after `d` is to the _right_ of the right edge of the blue bar for dropping as a child of `d`.

https://github.com/user-attachments/assets/666a7cc2-8e5b-4825-8309-0b67cdf660c6

## Expected Behavior

The right edge of the blue bar for dropping after `d` should be to the *left* of the right edge of the blue bar for dropping as a child of `d`.

In general, the `x` position of the right edge should increase as depth increases. This corresponds with the user's intuition that if they drag further to the right, the thought is going deeper.

PR #3103, opened 2025-07-01
Fix 2094: `Table: Cannot drag and drop before first thought in second row`

codebuild-ro:
I fixed the issue - #2094 
 
image
Before, the subthoguth drop target panel were overlapping the behind panel, that was the reason to let user cannot move drag and drop.
I reduced the height of the subthought drop target panel.
The height of drop target panel is independent, not related to height of thought.


src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drop-hover-after-first-thought-of-column-one-1.png
raineorshine, line 1: This appears to be a regression. The drop target for `d` needs to extend below the drop target for `e`. 

This creates the most intuitive behavior when the user is dragging a thought *down* from `e` to `d` to `a`.

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drop-target-last-child-in-cliff-1.png
raineorshine, line 1: This is a regression. The appearance on `main` is correct. The bottom edge of an ancestor drop target needs to fall below the bottom edge of the descendant's.

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drop-drop-targets-drop-end-1.png
raineorshine, line 1: The height of `b` has become too small. The greater height is intentional here, giving the user more space to drop on `b`. There are no children or subsequent siblings to interfere with it.

src/e2e/puppeteer/__tests__/drag-and-drop.ts
raineorshine, line 1: Test looks good 👍

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drag-and-drop-thought-1.png
raineorshine, line 1: ## Current Behavior

f366ad9ea6

The length of the blue bar is too long.

https://github.com/user-attachments/assets/50edb963-9bd7-4a6f-9474-c2393bb83dbf


## Expected Behavior

`main`

https://github.com/user-attachments/assets/7c2f3458-3f24-4e92-8a90-21b93c72e376

codebuild-ro: Now the hover bar length is consistent

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drop-drop-targets-drop-end-1.png
codebuild-ro: The drop panel height is rollbacked for providing more space to drop on `b`

codebuild-ro:
I made the blue hover bar length as related to viewport component.
It would be more convinient and easier to control the length of the hover bar.
Previous approach was making the blue bar length as not consistent because of using (`100% - {some variables}` inside wrapper elements which has width as `100% - {some variables}`).
image

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drop-target-last-child-in-cliff-1.png
raineorshine: The bottom edge looks good now! Thanks.

If we reduce the width of `a`, `b`, `c`, and `d`, drop targets, we'll also need to reduce the width of the `x` drop target to prevent it from sticking out.

The right edge of the drop target is proportional to depth. That is, the right edge of an ancestor drop target like `x` should be to the left of the drop target for `a`, `b`, `c`, and `d`, as shown in the original snapshot.

image

src/hooks/useDropHoverLength.ts
```diff
@@ -0,0 +1,23 @@
...
+import viewportStore from '../stores/viewport'
+
+interface DropHoverLengthProps {
+  isTableCol1?: boolean
+  isTableCol2?: boolean
+}
+
+/** Returns drop hover bar style. */
+const useDropHoverLength = (props?: DropHoverLengthProps) => {
+  const { contentWidth } = viewportStore.getState()
```
raineorshine: Can you explain your choice to base this on `contentWidth`? Since this wasn't discussed in the original issue, I would expect a thorough explanation and justification. You will need to convince me that this is the appropriate solution.

src/stores/viewport.ts
```diff
@@ -40,9 +42,14 @@ export const updateSize = _.throttle(
       }
     }
 
+    // Get content element width
+    const contentElement = document.getElementById('content')
+    const contentWidth = contentElement?.getBoundingClientRect().width || 0
```
raineorshine: I don't love the dependency on the DOM. I will await your explanation of the use of `contentWidth`, but this is looking a bit fragile and not typesafe.

src/hooks/useDropHoverLength.ts
```diff
@@ -0,0 +1,23 @@
...
+  // TODO: Need to check about the use of isTableCol1 for 50vw property
+  const isTableCol1 = props?.isTableCol1
+  const isTableCol2 = props?.isTableCol2
+
+  // 60 is the sum of left and right padding of main viewport <div>...</div>.
```
raineorshine: I'd like to see stronger type safety than hardcoding `60px` and adding a comment, as this is very fragile. We have constants and Panda recipes to create programmatic dependencies between components and styles.

src/hooks/useDropHoverLength.ts
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'react'
+import viewportStore from '../stores/viewport'
+
+interface DropHoverLengthProps {
+  isTableCol1?: boolean
+  isTableCol2?: boolean
+}
```
raineorshine: More explanation is needed for how the props affect the output, and how this hook should be used. It has hidden dependencies to the `Content` component, table view, and possibly other styles that will make this fragile and hard to maintain.

raineorshine (changes requested):
Hi, thanks for addressing the regressions! The snapshots look good.

I have some concerns about the direction the implementation is going given the increased complexity, DOM and style dependencies across components, and lack of type safety. Hopefully you can find ways to tame this complexity. This PR seems to touch an awful lot to make the single cases described in #2094 work as expected.

I found one regression that was not covered by the existing tests. I added a snapshot test in 991b912e3. Once you merge `main` you should be able to see it fail.

raineorshine:
Here is a test case that does not currently have test coverage that I know of, and it's only clear when comparing two different `DropHover` positions.

## Steps to Reproduce

```
- x
- a
  - =pin
  - b
    - =archive
      - =pin
    - c
      - d
```

## Current Behavior

The right edge of the blue bar for dropping after `d` is to the *right* of the right edge of the blue bar for dropping as a child of `d`.

https://github.com/user-attachments/assets/50935e48-00d5-4355-8624-10412f0780bc

## Expected Behavior

`main`

The right edge of the blue bar for dropping after `d` should be to the *left* of the right edge of the blue bar for dropping as a child of `d`.

In general, the `x` position of the right edge should increase as depth increases. This corresponds with the user's intuition that if they drag further to the right, the thought is going deeper.

https://github.com/user-attachments/assets/a8a5c27b-45f6-428c-8223-41c80c2314c2

src/stores/viewport.ts
codebuild-ro: I updated to use ResizeObserver which reliably tracks size changes.

That `contentWidth` is just for drag-and-drop actions.
The `contentWidth` is used in only useDropHoverLength hook, which is for the getting width of drop hover bar.
BTW, user cannot do drag-drop action before the document loaded.

And also the `contentWidth` value will be surely set after loading of document.
It is for changing of viewport. The drop hover bar width should be changed according to the viewport size.

So even though it uses DOM measurement, it's reliable because:
- Width is always available
- Updates happen automatically with viewport changes
- Clean type definitions throughout

src/hooks/useDropHoverLength.ts
codebuild-ro: With previous approach, it is not possible to make the blue bar length as consistent.
The previous approach was to render blue bar with property `width: calc(100% - 2em)` or `width: 100%` inside different width of wrappers (subthoughts has width as `calc(100% + 1em - indent)`).
So the drop hover blue bar would be different each other.

That creates the issue like:
[discussion_r2190684030](https://github.com/cybersemics/em/pull/3103/#discussion_r2190684030)
[discussion_r2178377236](https://github.com/cybersemics/em/pull/3103/#discussion_r2178377236
)
So I decided to use new approach to calculate the width of blue bar according to the content component width.

src/hooks/useDropHoverLength.ts
codebuild-ro: I added comments for useDropHoverLength in more detail

codebuild-ro:
> Here is a test case that does not currently have test coverage that I know of, and it's only clear when comparing two different `DropHover` positions.
> 
> ## Steps to Reproduce
> ```
> - x
> - a
>   - =pin
>   - b
>     - =archive
>       - =pin
>     - c
>       - d
> ```
> 
> ## Current Behavior
> The right edge of the blue bar for dropping after `d` is to the _right_ of the right edge of the blue bar for dropping as a child of `d`.
> 
>  Screen.Recording.2025-07-07.at.11.05.29.AM.mov 
> ## Expected Behavior
> `main`
> 
> The right edge of the blue bar for dropping after `d` should be to the _left_ of the right edge of the blue bar for dropping as a child of `d`.
> 
> In general, the `x` position of the right edge should increase as depth increases. This corresponds with the user's intuition that if they drag further to the right, the thought is going deeper.
> 
>  Screen.Recording.2025-07-07.at.11.06.09.AM.mov

This works with my updated code :). Seems you tested with a bit older code.

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drop-hover-of-first-child-of-col-2-1.png
raineorshine, line 1: The length of the blue bar broke in this case. The old snapshot is correct.

See: https://github.com/cybersemics/em/issues/2094#issuecomment-3053451068

src/hooks/useDropHoverLength.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Just use for dragging-dropping behavior.
+ * Returns the width for a drop hover bar based on context.
+ *
+ * Dependencies:
+ * - Requires parent Content component with id="content"
+ * - Assumes CONTENT_BOX_PADDING matches Content padding
+ * - For table view, assumes 50/50 split of table width.
```
raineorshine: Unfortunately, we can't assume the columns are 50/50. Column width is content-dependent. For example, try using a longer thought in col1.

src/hooks/useDropHoverLength.ts
```diff
@@ -0,0 +1,43 @@
+import { useMemo } from 'react'
+import { CONTENT_BOX_PADDING } from '../constants'
+import viewportStore from '../stores/viewport'
+
+interface DropHoverLengthProps {
+  isTableCol1?: boolean // For dropping into place as a first column in table view
+  isTableCol2?: boolean // For dropping into place as a second column in table view
+}
```
raineorshine: It's not clear to me when `isTableCol1` and `isTableCol2` should be passed to the hook. It seems to depend on the specific `DropHover` instance (e.g. `DropEnd`, `DropCliff`, etc), but it's not clear what determines it. Can you establish some guidelines and documentation for when these props should be used?

Continuation of https://github.com/cybersemics/em/pull/3103#discussion_r2190807742

src/stores/viewport.ts
```diff
@@ -57,4 +59,14 @@ export const updateSize = _.throttle(
   { leading: true },
 )
 
+/** Throttled update of viewport width. */
+export const updateContentWidth = _.throttle(
```
raineorshine: Can you explain why the throttling was added? How often does the observer fire in your experience?

src/hooks/useDropHoverLength.ts
```diff
@@ -0,0 +1,43 @@
...
+const useDropHoverLength = (props?: DropHoverLengthProps) => {
+  const { contentWidth } = viewportStore.getState()
+  // TODO: Need to check about the use of isTableCol1 for 50vw property
+  const isTableCol1 = props?.isTableCol1
+  const isTableCol2 = props?.isTableCol2
+
+  if (contentWidth < 0) {
+    console.error('Invalid content width')
+    return '100%' // Safe fallback
+  }
```
raineorshine: We should ensure higher up that `contentWidth` is always valid. 

Validation logic is more appropriate for user inputs we can't control.

src/hooks/useDropHoverLength.ts
raineorshine: Thanks, that makes sense.

FYI You can use `viewportStore.useSelector(...)` rather than object destructuring for better performance. e.g. There's no need to re-run the hook when viewport height changes.

src/constants.ts
```diff
@@ -605,3 +605,9 @@ export const COMMAND_GROUPS: {
 
 /** The duration of the haptics vibrate on delete or archive non-empty thought. */
 export const DELETE_VIBRATE_DURATION = 80
+
+/**
+ * Ref src/components/Content.tsx:84 (padding: '80px 10px 153px 50px')
+ * 60 is the sum of left and right padding of main viewport - <div id='content'...>...</div>.
+ **/
+export const CONTENT_BOX_PADDING = 60
```
raineorshine: All you've done is factor out the variable, but the padding in `Content.tsx` remains completely decoupled.

Please ensure that these are *programmatically* coupled. Comments are not a substitute for the type system. Line numbers in comments are even worse, as they will be invalidated when the file changes.

Continuation of https://github.com/cybersemics/em/pull/3103#discussion_r2190805150

codebuild-ro: I made the interaction of content left and right padding to calculate blue bar length.
image
image

src/hooks/useDropHoverLength.ts
codebuild-ro: I removed that validation.

src/hooks/useDropHoverLength.ts
codebuild-ro: I updated contentWidth fetching approach as you mentioned.
image

src/stores/viewport.ts
codebuild-ro: I removed this method. Since, I couldn't notice remarkable performance update with it.

src/hooks/useDropHoverLength.ts
codebuild-ro: I updated doc and guidlne for `useDropHoverLength` hook

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drop-hover-of-first-child-of-col-2-1.png
codebuild-ro: I applied the new logic of length of blue bar to DropChild too.
Now the blue hover bar length is consistent for this case too.

https://github.com/user-attachments/assets/60715488-f98a-4690-b607-56ad3fdc74e1

PR #3109, opened 2025-07-02
Move dragHold into dispatch

ethan-james:
References #3080

This is about as much as I was able to accomplish re: https://github.com/cybersemics/em/pull/3081#issuecomment-3014433269

1. If we're not tracking `dragInProgress` via `useSelector`, then we shouldn't need to do it for `dragHold` either
2. It also removes a dependency from `useEffect`
3. It adds `e.preventDefault()` to the existing `dragInProgress` condition, which may be unnecessary but I don't think is harmful
4. It allows us to simplify `(!globals.touching && (!editingOrOnCursor || !isVisible))`

`dragHold` was being checked in order to fix this bug: https://github.com/cybersemics/em/pull/3023#issuecomment-3005754047 and I verified that that bug is still fixed after moving it.


raineorshine (approved):
Seems reasonable. Thanks!

PR #3111, opened 2025-07-03
Animate sort action

### LINKED ISSUE #2530: Animation: Sort
## Steps to Reproduce

```
- T
- H
- I
- N
- K
```

1. Set the cursor on a thought.
2. Toggle Sort.

## Current Behavior

- Cursor thought snaps into place.
- Other thoughts animate into place with a translate animation.

https://github.com/user-attachments/assets/1ce40f75-33e3-41c2-9c97-133a6e7804e1

## Expected Behavior

Use the same curve animation as #2528. See [layered animations](https://tobiasahlin.com/blog/curved-path-animations-in-css/).

- [x] Translate (ease-in-out)
- [ ] Thoughts increasing in rank should curve right (ease-in)
- [ ] Thoughts decreasing in rank should curve left (ease-in)

zukilover:
Fix #2530 

Sort animation is done by adding offset to x and simulating a curve motion.
Determining `isSort` by comparing a node's previous rank with the current one. I may need to revisit this since it is now clear that sort action must actually be undoable.

### DIFF
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -1,4 +1,4 @@
-import { useLayoutEffect, useRef, useState } from 'react'
+import { useLayoutEffect, useMemo, useRef, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { CSSTransitionProps } from 'react-transition-group/CSSTransition'
 import { css } from '../../styled-system/css'
@@ -138,9 +138,6 @@ const TreeNode = ({
     !isSwap ? null : isCursorGreaterThanParent(state) ? 'clockwise' : 'counterclockwise',
   )
 
-  // Hold the sort direction in a ref so it only updates when the index actually changes.
-  const sortDirectionRef = useRef<'clockwise' | 'counterclockwise' | null>(null)
-
   /** Detect when this thought's on-screen position (array index) has changed. */
   const indexChanged = previousIndex !== undefined && previousIndex !== index
 
@@ -155,21 +152,14 @@ const TreeNode = ({
    * The direction of the curved animation in sort operations.
    * - If the thought's rank increased (moved down), curve right (clockwise).
    * - If the thought's rank decreased (moved up), curve left (counterclockwise).
-   *
-   * We cache the value in a ref so that once determined for a given sort animation
-   * it remains stable across any additional renders that may occur while the
-   * animation is in progress (e.g. position updates from virtualization).
    */
-  const sortDirection: 'clockwise' | 'counterclockwise' | null = (() => {
-    // Determine direction only on the first render after the index change.
-    if (isLastActionSort && indexChanged && previousIndex !== null) {
-      const dir: 'clockwise' | 'counterclockwise' = index > previousIndex ? 'clockwise' : 'counterclockwise'
-      sortDirectionRef.current = dir
-      return dir
-    }
-    // Otherwise fall back to the previously determined direction.
-    return sortDirectionRef.current
-  })()
+  const sortDirection = useMemo(
+    () => (isLastActionSort && previousIndex !== undefined && index > previousIndex ? 'clockwise' : 'counterclockwise'),
+    // Only recalculate sortDirection when index changes and  last action is sort.
+    // Do not update when previousIndex changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [index, isLastActionSort],
+  )
 
   /**
    * Horizontal offset for the first frame of a sort animation. This allows the X transition
@@ -266,9 +256,7 @@ const TreeNode = ({
               ? 'left {durations.layoutNodeAnimation} {easings.nodeCurveXLayerClockwise}'
               : 'left {durations.layoutNodeAnimation} {easings.nodeCurveXLayer}'
             : isLastActionSort
-              ? sortDirection === 'clockwise'
-                ? 'left {durations.layoutNodeAnimation} {easings.nodeCurveSortXLayer}'
-                : 'left {durations.layoutNodeAnimation} {easings.nodeCurveSortXLayer}'
+              ? 'left {durations.layoutNodeAnimation} {easings.nodeCurveSortXLayer}'
               : contextAnimation
                 ? 'left {durations.disappearingUpperRight} ease-out,top {durations.disappearingUpperRight} ease-out'
                 : 'left {durations.layoutNodeAnimation} ease-out,top {durations.layoutNodeAnimation} ease-out',
@@ -291,9 +279,7 @@ const TreeNode = ({
                 ? 'top {durations.layoutNodeAnimation} {easings.nodeCurveYLayerClockwise}'
                 : 'top {durations.layoutNodeAnimation} {easings.nodeCurveYLayer}'
               : isLastActionSort
-                ? sortDirection === 'clockwise'
-                  ? 'top {durations.layoutNodeAnimation} {easings.nodeCurveSortYLayer}'
-                  : 'top {durations.layoutNodeAnimation} {easings.nodeCurveSortYLayer}'
+                ? 'top {durations.layoutNodeAnimation} {easings.nodeCurveSortYLayer}'
                 : undefined,
           })}
           style={innerDivStyle}
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -5,6 +5,7 @@ import { css } from '../../styled-system/css'
 import ActionType from '../@types/ActionType'
 import State from '../@types/State'
 import TreeThoughtPositioned from '../@types/TreeThoughtPositioned'
+import durations from '../durations.config'
 import testFlags from '../e2e/testFlags'
 import useFauxCaretNodeProvider from '../hooks/useFauxCaretCssVars'
 import isContextViewActive from '../selectors/isContextViewActive'
@@ -70,9 +71,12 @@ const TreeNode = ({
 } & Pick<CSSTransitionProps, 'in'>) => {
   const [y, setY] = useState(_y)
   const [x, setX] = useState(_x)
+  const [sortOffset, setSortOffset] = useState(0)
+  const [isSortAnimating, setIsSortAnimating] = useState(false)
   // Since the thoughts slide up & down, the faux caret needs to be a child of the TreeNode
   // rather than one universal caret in the parent.
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
+  const previousRankRef = useRef<number | null>(null)
 
   const fauxCaretNodeProvider = useFauxCaretNodeProvider({
     editing,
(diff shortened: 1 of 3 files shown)


zukilover:
Here's how the sort animation works:
When sort action is triggered, it will apply an offset to `x` and starts animating with a curve motion.
Before it completes the transition, it resets the position to the original `x` value:

https://github.com/cybersemics/em/blob/c1628d019483e5a80d77fc8c0ee5208d7854764b/src/components/TreeNode.tsx#L219-L222


<details>
<summary>How it works:</summary>

https://github.com/user-attachments/assets/e8d89e36-88b7-4816-802e-658958737c05


</details>

raineorshine (changes requested):

From the video I can clearly see that `H` and `I` do not curve left as required.
Additionally, `N` curves left when it shouldn't (because it's rank isn't changing).

Please make sure the animations all work as required before requesting another review. Small tweaks or missed edge cases are understandable, but this submission obviously doesn't meet the requirements. I don't want to waste time with lots of back-and-forth time like we have on previous issues. If you're not clear about the requirements, please ask questions. I'm more than happy to clarify. Thank you!

zukilover:
Hi @raineorshine.

Based on the requirements, I'm using the node's literal rank to determine whether its rank has changed:
https://github.com/cybersemics/em/blob/c1628d019483e5a80d77fc8c0ee5208d7854764b/src/components/TreeNode.tsx#L142-L143

I tested with sorting the nodes alphabetically, then by their created time, here are the results:

Before sort:
Node | Previous Rank | Current Rank | Is Rank Changed?
-- | -- | -- | --
T | 1 | 1 | false
H | 3 | 3 | false
I | 4 | 4 | false
N | 5 | 5 | false
K | 6 | 6 | false


Sort alphabetical:
Node | Previous Rank | Current Rank | Is Rank Changed?
-- | -- | -- | --
H | 3 | 3 | false
I | 4 | 4 | false
K | 6 | 5 | true
N | 5 | 6 | true
T | 1 | 7 | true


Sort created:
Node | Previous Rank | Current Rank | Is Rank Changed?
-- | -- | -- | --
T | 7 | 1 | true
H | 3 | 3 | false
I | 4 | 4 | false
K | 6 | 5 | true
N | 5 | 6 | true


Just to confirm, by "ranks," are you referring to the node's rank value or its perceived top-to-bottom position on the screen?

raineorshine:
Ah, interesting! I guess I had some faulty assumptions about how the ranks changed on sort. The important part is which thoughts move during the sort. They need to animate based on the perceived top-to-bottom position on the screen. So I believe that would be their array index rather than `Thought["rank"]`.

Apologies for jumping to conclusions. I see now that you carefully implemented a solution based on rank. I should have written the requirements more from the user's perspective.

zukilover:
No worries @raineorshine, and thank you for the clarification.

I've now determined the sort direction based on the thought's index rather than its actual rank.
I've tested it to ensure it meets the requirements:
- Thoughts increasing in rank (i.e., moving down in the list with increasing index) should curve right (ease-in)
- Thoughts decreasing in rank should curve left (ease-in)

I hope I got it right.

<details>
<summary>
Updated animation
</summary>

https://github.com/user-attachments/assets/f6bd1653-2d7c-4eb0-b2a0-46514c8aad1a


</details>

src/components/TreeNode.tsx
```diff
@@ -133,6 +140,55 @@ const TreeNode = ({
...
 
+  // The current visual index of this thought in the rendered list.
+  const currentRank = index
+
+  // Get previous index using the usePrevious pattern.
+  const previousRank = (() => {
+    const prev = previousRankRef.current
+    previousRankRef.current = currentRank
+    return prev
+  })()
```
raineorshine: You can use the [usePrevious](https://github.com/cybersemics/em/blob/c7d9cde5095e1866525ae9fd372a694242568db2/src/hooks/usePrevious.ts) hook.

src/components/TreeNode.tsx
```diff
@@ -133,6 +140,55 @@ const TreeNode = ({
...
+  })
+
+  /**
+   * The direction of the curved animation in sort operations.
+   * - If the thought's rank increased (moved down), curve right (clockwise).
+   * - If the thought's rank decreased (moved up), curve left (counterclockwise).
+   *
+   * We cache the value in a ref so that once determined for a given sort animation
+   * it remains stable across any additional renders that may occur while the
+   * animation is in progress (e.g. position updates from virtualization).
```
raineorshine: This is an interesting but unconventional caching method. Can we just use `useMemo`?

src/components/TreeNode.tsx
```diff
@@ -152,6 +208,22 @@ const TreeNode = ({
     setY(_y)
   }, [_y])
 
+  useLayoutEffect(() => {
+    // Start sort animation when the visual rank changes, keep it active for the full duration.
```
raineorshine: Maybe we should short circuit if `!isLastActionSort` to avoid false positives?

src/components/TreeNode.tsx
```diff
@@ -133,6 +140,55 @@ const TreeNode = ({
...
+
+  // Hold the sort direction in a ref so it only updates when the index actually changes.
+  const sortDirectionRef = useRef<'clockwise' | 'counterclockwise' | null>(null)
+
+  /** Detect when this thought's perceived on-screen position (array index) has changed. */
+  const rankChanged = previousRank !== null && currentRank !== null && previousRank !== currentRank
+
+  /** True if the last action is setSortPreference. */
+  const isLastActionSort = useSelector(state => {
+    const sortActions: ActionType[] = ['setSortPreference']
```
raineorshine: Can you add `toggleSort` to this list? It's a similar command that rotates through the alphabetical sorts through a keyboard shortcut. We want the same animation for that.

src/components/TreeNode.tsx
```diff
@@ -133,6 +140,55 @@ const TreeNode = ({
     !isSwap ? null : isCursorGreaterThanParent(state) ? 'clockwise' : 'counterclockwise',
   )
 
+  // The current visual index of this thought in the rendered list.
+  const currentRank = index
```
raineorshine: Try to avoid re-aliasing a variable. It adds an unnecessary level of indirection.

raineorshine (changes requested):
Thanks! This looks like exactly what we want.

In order to avoid further confusion, can you make sure we only refer to the position as "index" and not rank? `Thought.rank` is a specific property, and, as we discovered, that's not what we're working with here. All variables names and comments should consistently refer to "index" or "child index".

Thanks!

src/components/TreeNode.tsx
zukilover: `usePrevious` has been implemented.

src/components/TreeNode.tsx
zukilover: Added short circuit.

src/components/TreeNode.tsx
zukilover: Added `toggleSort`.

src/components/TreeNode.tsx
zukilover: Remove re-alias and rename variables to better suited with the index usages.

src/components/TreeNode.tsx
zukilover: We can't use `useMemo` as it runs during the render phase. It can cache expensive computations, but when its dependency array changes it recomputes. If we used it (I've tried that) the direction would recompute (or reset) whenever the dependency array changed again, which could reset the animation mid-flight.

src/components/TreeNode.tsx
```diff
@@ -204,9 +265,13 @@ const TreeNode = ({
             ? swapDirection === 'clockwise'
               ? 'left {durations.layoutNodeAnimation} {easings.nodeCurveXLayerClockwise}'
               : 'left {durations.layoutNodeAnimation} {easings.nodeCurveXLayer}'
-            : contextAnimation
-              ? 'left {durations.disappearingUpperRight} ease-out,top {durations.disappearingUpperRight} ease-out'
-              : 'left {durations.layoutNodeAnimation} ease-out,top {durations.layoutNodeAnimation} ease-out',
+            : isLastActionSort
+              ? sortDirection === 'clockwise'
+                ? 'left {durations.layoutNodeAnimation} {easings.nodeCurveSortXLayer}'
+                : 'left {durations.layoutNodeAnimation} {easings.nodeCurveSortXLayer}'
```
raineorshine: It looks like these branches are identical, so the ternary isn't necessary.

src/components/TreeNode.tsx
```diff
@@ -225,7 +290,11 @@ const TreeNode = ({
               ? swapDirection === 'clockwise'
                 ? 'top {durations.layoutNodeAnimation} {easings.nodeCurveYLayerClockwise}'
                 : 'top {durations.layoutNodeAnimation} {easings.nodeCurveYLayer}'
-              : undefined,
+              : isLastActionSort
+                ? sortDirection === 'clockwise'
+                  ? 'top {durations.layoutNodeAnimation} {easings.nodeCurveSortYLayer}'
+                  : 'top {durations.layoutNodeAnimation} {easings.nodeCurveSortYLayer}'
```
raineorshine: These branches are also identical, and the ternary can be removed.

src/components/TreeNode.tsx
raineorshine: `useMemo` is pretty flexible based on which dependencies are used. Can you try this out and let me know if I'm missing anything?

(The below diff can be applied directly to your local git repo with `git apply`.)

```diff
diff --git a/src/components/TreeNode.tsx b/src/components/TreeNode.tsx
index a0ecaac813..33bc04459e 100644
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -1,4 +1,4 @@
-import { useLayoutEffect, useRef, useState } from 'react'
+import { useLayoutEffect, useMemo, useRef, useState } from 'react'
 import { useSelector } from 'react-redux'
 import { CSSTransitionProps } from 'react-transition-group/CSSTransition'
 import { css } from '../../styled-system/css'
@@ -138,9 +138,6 @@ const TreeNode = ({
     !isSwap ? null : isCursorGreaterThanParent(state) ? 'clockwise' : 'counterclockwise',
   )
 
-  // Hold the sort direction in a ref so it only updates when the index actually changes.
-  const sortDirectionRef = useRef<'clockwise' | 'counterclockwise' | null>(null)
-
   /** Detect when this thought's on-screen position (array index) has changed. */
   const indexChanged = previousIndex !== undefined && previousIndex !== index
 
@@ -155,21 +152,14 @@ const TreeNode = ({
    * The direction of the curved animation in sort operations.
    * - If the thought's rank increased (moved down), curve right (clockwise).
    * - If the thought's rank decreased (moved up), curve left (counterclockwise).
-   *
-   * We cache the value in a ref so that once determined for a given sort animation
-   * it remains stable across any additional renders that may occur while the
-   * animation is in progress (e.g. position updates from virtualization).
    */
-  const sortDirection: 'clockwise' | 'counterclockwise' | null = (() => {
-    // Determine direction only on the first render after the index change.
-    if (isLastActionSort && indexChanged && previousIndex !== null) {
-      const dir: 'clockwise' | 'counterclockwise' = index > previousIndex ? 'clockwise' : 'counterclockwise'
-      sortDirectionRef.current = dir
-      return dir
-    }
-    // Otherwise fall back to the previously determined direction.
-    return sortDirectionRef.current
-  })()
+  const sortDirection = useMemo(
+    () => (isLastActionSort && previousIndex !== undefined && index > previousIndex ? 'clockwise' : 'counterclockwise'),
+    // Only recalculate sortDirection when index changes and  last action is sort.
+    // Do not update when previousIndex changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [index, isLastActionSort],
+  )
 
   /**
    * Horizontal offset for the first frame of a sort animation. This allows the X transition
```

src/components/TreeNode.tsx
zukilover: This is removed.

src/components/TreeNode.tsx
zukilover: This is now removed.

src/components/TreeNode.tsx
zukilover: Hey, it’s working now! Not sure what I did wrong before, appreciate the suggestion.

raineorshine: Great!

Yeah, I did have to fiddle a bit with the dependencies.

raineorshine (approved):

PR #3112, opened 2025-07-03
useEditMode: shouldSetSelection checks selection.isThought

### LINKED ISSUE #3061: [Desktop] Selection not set after Toggle Command Palette + Cursor Down
## Steps to Reproduce

1. Create thought `a` and set the cursor to null. 
2. Open the Command Palette.
3. Close the Command Palette. 
4. Cursor Down

## Current Behavior

The cursor is set to `a` but there is no browser selection. 

## Expected Behavior

The browser selection should be at the beginning of `a`.

On desktop, the browser selection should always be set after the cursor moves.

## Testing

Add a puppeteer caret test if reproducible.

ethan-james:
Fixes #3061 

Since the command palette's wrapper `div` stays attached to the DOM when the command palette is closed, it retains focus. The fact that it is still the `focusNode` causes `(cursorOffset !== null || !selection.isActive())` in `useEditMode` to be false, preventing commands from programmatically setting focus. Setting `userSelect: none` on the wrapper `div` doesn't prevent it from being focused, so the next best solution is to call `selection.clear()` when closing the command palette.


ethan-james:
It seems to me like the test that I pushed should work, but it doesn't seem like the command palette disappears after pressing Escape. If you have any insight into why that would be the case, please let me know.

ethan-james:
I took a screenshot of the test at that point, and it looks like the command palette isn't opening in the first place. I'll see if I can figure out why tomorrow.

before

raineorshine:
Before considering the failing test, I'd like to discuss the proposed solution. This attempts to solve the problem by removing the selection when the CommandPalette closes. However, it seems like `useEditMode` should work regardless.

When was `!selection.isActive()` added and why? This might help us better determine if it is the correct condition to begin with.

We also have [selection.isThought](https://github.com/cybersemics/em/blob/1fee57f22a55cc376151468b2c1a8a43fadb5c61/src/device/selection.ts#L77) we could use. It seems to me that if `selection.isThought` is false, then we would would want to set the selection. So perhaps `!selection.isActive() || !selection.isThought`. However, that depends on what purpose `!selection.isActive()` is serving.

raineorshine (approved):
Looks good to me! Ready to merge?

ethan-james:
> Looks good to me! Still working on the test?

I am. It doesn't seem to have improved when I changed the solution.

ethan-james:
It turns out that I accidentally had it in the `mobile only` section, and it didn't like that it was emulating iPhone 11. Since this is a desktop bug, it's good to not have it emulating a mobile device.

PR #3114, opened 2025-07-07
Add move up/down animation

### LINKED ISSUE #2531: Animation: Move Thought Up/Down
## Steps to Reproduce

```
- a
- b
```

1. Set the cursor on `b`.
2. Move Thought Up.

## Current Behavior

Translate animation:

https://github.com/user-attachments/assets/57b3b3b5-ecaf-4a3d-9b70-71e8cbff3347

## Expected Behavior

- Cursor thought should enlarge (scale3d).
- Non-cursor thought should shrink, blur, and fade out (so that it appears to go underneath the cursor thought on the z axis).

zukilover:
Resolves #2531

### DIFF
--- a/panda.config.ts
+++ b/panda.config.ts
@@ -164,12 +164,12 @@ const keyframes = defineKeyframes({
       transform: 'translateX(-50%) translateY(0)',
     },
   },
-  moveThoughtAction: {
+  moveThoughtOver: {
     '0%': { transform: 'scale3d(1, 1, 1)' },
     '50%': { transform: 'scale3d(1.5, 1.5, 1)' },
     '100%': { transform: 'scale3d(1, 1, 1)' },
   },
-  moveThoughtDisplaced: {
+  moveThoughtUnder: {
     '0%': { transform: 'scale3d(1, 1, 1)', opacity: 1, filter: 'blur(0)' },
     '50%': { transform: 'scale3d(0.5, 0.5, 1)', opacity: 0.5, filter: 'blur(2px)' },
     '100%': { transform: 'scale3d(1, 1, 1)', opacity: 1, filter: 'blur(0)' },
--- a/panda.config.ts
+++ b/panda.config.ts
@@ -164,6 +164,16 @@ const keyframes = defineKeyframes({
       transform: 'translateX(-50%) translateY(0)',
     },
   },
+  moveThoughtAction: {
+    '0%': { transform: 'scale3d(1, 1, 1)' },
+    '50%': { transform: 'scale3d(1.5, 1.5, 1)' },
+    '100%': { transform: 'scale3d(1, 1, 1)' },
+  },
+  moveThoughtDisplaced: {
+    '0%': { transform: 'scale3d(1, 1, 1)', opacity: 1, filter: 'blur(0)' },
+    '50%': { transform: 'scale3d(0.5, 0.5, 1)', opacity: 0.5, filter: 'blur(2px)' },
+    '100%': { transform: 'scale3d(1, 1, 1)', opacity: 1, filter: 'blur(0)' },
+  },
   // the hideCaret animation must run every time the indent changes on iOS Safari, which necessitates replacing the animation with an identical substitute with a different name
   // See: recipes/hideCaret.ts
   // TODO: FauxCaret will break if hideCaretAnimationNames is imported from hideCaret.config.ts into hideCaret.ts, and vice versa into panda.config.ts, so we are stuck with duplicate definitions in two files.
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -162,7 +162,11 @@ const TreeNode = ({
     [index, isLastActionSort],
   )
 
-  const { isMoveAnimating, moveType, moveDivStyle } = useMoveThoughtAnimation({ indexChanged, isCursor, path })
+  const { isMoveAnimating, moveType, moveDivStyle } = useMoveThoughtAnimation({
+    indexChanged,
+    previousIndex,
+    index,
+  })
 
   /**
    * Horizontal offset for the first frame of a sort animation. This allows the X transition
@@ -237,11 +241,11 @@ const TreeNode = ({
       : moveDivStyle
 
   const moveAnimation =
-    isMoveAnimating && moveType === 'moveThoughtSibling'
+    isMoveAnimating && moveType === 'moveThoughtDisplaced'
       ? `transform {durations.layoutNodeAnimation} ease-out, filter {durations.layoutNodeAnimation} ease-out, opacity {durations.layoutNodeAnimation} ease-out`
-      : isMoveAnimating && moveType === 'moveThoughtCursor'
+      : isMoveAnimating && moveType === 'moveThoughtAction'
         ? `transform {durations.layoutNodeAnimation} ease-out`
-        : moveType === 'moveThoughtSibling'
+        : moveType === 'moveThoughtDisplaced'
           ? `opacity {durations.layoutNodeAnimation} ease-out`
           : undefined
 
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -13,8 +13,8 @@ import isContextViewActive from '../selectors/isContextViewActive'
 import isCursorGreaterThanParent from '../selectors/isCursorGreaterThanParent'
 import nextSibling from '../selectors/nextSibling'
 import prevSibling from '../selectors/prevSibling'
-import head from '../util/head'
 import equalPath from '../util/equalPath'
+import head from '../util/head'
 import isDescendantPath from '../util/isDescendantPath'
 import parentOf from '../util/parentOf'
 import DropCliff from './DropCliff'
@@ -76,7 +76,7 @@ const TreeNode = ({
   const [y, setY] = useState(_y)
   const [x, setX] = useState(_x)
   const [isSortAnimating, setIsSortAnimating] = useState(false)
-  const [isMoveThoughtAnimating, setIsMoveThoughtAnimating] = useState(false)
+  const [isMoveAnimating, setIsMoveThoughtAnimating] = useState(false)
   // Since the thoughts slide up & down, the faux caret needs to be a child of the TreeNode
   // rather than one universal caret in the parent.
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
@@ -214,7 +214,7 @@ const TreeNode = ({
   })
 
   // Determine animation type for moveThought
-  const moveAnimation: 'moveThoughtCursor' | 'moveThoughtSibling' | null = lastMoveType
+  const moveType: 'moveThoughtCursor' | 'moveThoughtSibling' | null = lastMoveType
     ? isCursor
       ? 'moveThoughtCursor'
       : thoughtId === movingThoughtId
@@ -303,14 +303,14 @@ const TreeNode = ({
           top: y,
           left: 0,
         }
-      : isMoveThoughtAnimating
-        ? moveAnimation === 'moveThoughtCursor'
+      : isMoveAnimating
+        ? moveType === 'moveThoughtCursor'
           ? {
               transformOrigin: 'left',
               transform: 'scale3d(1.5, 1.5, 1)',
               transition: `transform {durations.layoutNodeAnimation} ease-out`,
             }
-          : moveAnimation === 'moveThoughtSibling'
+          : moveType === 'moveThoughtSibling'
             ? {
                 transformOrigin: 'left',
                 transform: 'scale3d(0.5, 0.5, 1)',
@@ -319,6 +319,15 @@ const TreeNode = ({
               }
             : undefined
         : undefined
+
+  const moveAnimation =
+    isMoveAnimating && moveType === 'moveThoughtSibling'
+      ? `transform {durations.layoutNodeAnimation} ease-out, filter {durations.layoutNodeAnimation} ease-out, opacity {durations.layoutNodeAnimation} ease-out`
+      : isMoveAnimating && moveType === 'moveThoughtCursor'
+        ? `transform {durations.layoutNodeAnimation} ease-out`
+        : moveType === 'moveThoughtSibling'
+          ? `opacity {durations.layoutNodeAnimation} ease-out`
+          : undefined
 
   return (
     <FadeTransition
@@ -366,11 +375,7 @@ const TreeNode = ({
                 : 'top {durations.layoutNodeAnimation} {easings.nodeCurveYLayer}'
               : isLastActionSort
                 ? 'top {durations.layoutNodeAnimation} {easings.nodeCurveSortYLayer}'
-                : isMoveThoughtAnimating
-                  ? moveAnimation === 'moveThoughtSibling'
-                    ? 'transform {durations.layoutNodeAnimation} ease-out, filter {durations.layoutNodeAnimation} ease-out, opacity {durations.layoutNodeAnimation} ease-out'
-                    : 'transform {durations.layoutNodeAnimation} ease-out'
-                  : undefined,
+                : moveAnimation,
           })}
           style={innerDivStyle}
         >
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -76,7 +76,7 @@ const TreeNode = ({
   const [y, setY] = useState(_y)
   const [x, setX] = useState(_x)
   const [isSortAnimating, setIsSortAnimating] = useState(false)
-  const [isMoveAnimating, setIsMoveThoughtAnimating] = useState(false)
+  const [isMoveAnimating, setIsMoveAnimating] = useState(false)
   // Since the thoughts slide up & down, the faux caret needs to be a child of the TreeNode
   // rather than one universal caret in the parent.
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
@@ -262,13 +262,13 @@ const TreeNode = ({
     // In moves that cross contexts (e.g. moving the only child of «A» below its next uncle «B»),
     // there is no displaced sibling, so `movingThoughtId` is null. Still animate the cursor node.
     if (lastMoveType && (isCursor || movingThoughtId)) {
-      setIsMoveThoughtAnimating(true)
+      setIsMoveAnimating(true)
     }
 
     // After the layout node animation duration:
     // reset moveThought animation
     const timer = setTimeout(() => {
-      setIsMoveThoughtAnimating(false)
+      setIsMoveAnimating(false)
     }, durations.layoutNodeAnimation)
 
     return () => clearTimeout(timer)
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -162,7 +162,7 @@ const TreeNode = ({
     [index, isLastActionSort],
   )
 
-  const { moveDivStyle } = useMoveThoughtAnimation({ index })
+  const moveDivStyle = useMoveThoughtAnimation({ index })
 
   /**
    * Horizontal offset for the first frame of a sort animation. This allows the X transition
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -162,7 +162,7 @@ const TreeNode = ({
     [index, isLastActionSort],
   )
 
-  const moveDivStyle = useMoveThoughtAnimation({ index })
+  const moveDivStyle = useMoveThoughtAnimation(index)
 
   /**
    * Horizontal offset for the first frame of a sort animation. This allows the X transition
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -162,9 +162,7 @@ const TreeNode = ({
     [index, isLastActionSort],
   )
 
-  const { isMoveAnimating, moveType, moveDivStyle } = useMoveThoughtAnimation({
-    indexChanged,
-    previousIndex,
+  const { moveAnimation, moveDivStyle } = useMoveThoughtAnimation({
     index,
   })
 
@@ -173,6 +171,8 @@ const TreeNode = ({
    * to start immediately from the correct position.
    */
   const sortOffset = sortDirection === 'clockwise' ? 30 : -30
+
+  // moveAnimation is now provided by useMoveThoughtAnimation
 
   // We split x and y transitions into separate nodes to:
   // 1. Allow independent timing curves for horizontal and vertical movement
@@ -239,15 +239,6 @@ const TreeNode = ({
           left: 0,
         }
       : moveDivStyle
-
-  const moveAnimation =
-    isMoveAnimating && moveType === 'moveThoughtDisplaced'
-      ? `transform {durations.layoutNodeAnimation} ease-out, filter {durations.layoutNodeAnimation} ease-out, opacity {durations.layoutNodeAnimation} ease-out`
-      : isMoveAnimating && moveType === 'moveThoughtAction'
-        ? `transform {durations.layoutNodeAnimation} ease-out`
-        : moveType === 'moveThoughtDisplaced'
-          ? `opacity {durations.layoutNodeAnimation} ease-out`
-          : undefined
 
   return (
     <FadeTransition
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -162,7 +162,7 @@ const TreeNode = ({
     [index, isLastActionSort],
   )
 
-  const { isMoveAnimating, moveType, moveDivStyle } = useMoveThoughtAnimation({ indexChanged, isCursor })
+  const { isMoveAnimating, moveType, moveDivStyle } = useMoveThoughtAnimation({ indexChanged, isCursor, path })
 
   /**
    * Horizontal offset for the first frame of a sort animation. This allows the X transition
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -11,6 +11,8 @@ import useFauxCaretNodeProvider from '../hooks/useFauxCaretCssVars'
 import usePrevious from '../hooks/usePrevious'
 import isContextViewActive from '../selectors/isContextViewActive'
 import isCursorGreaterThanParent from '../selectors/isCursorGreaterThanParent'
+import nextSibling from '../selectors/nextSibling'
+import prevSibling from '../selectors/prevSibling'
 import equalPath from '../util/equalPath'
 import isDescendantPath from '../util/isDescendantPath'
 import parentOf from '../util/parentOf'
@@ -167,6 +169,32 @@ const TreeNode = ({
    */
   const sortOffset = sortDirection === 'clockwise' ? 30 : -30
 
+  /** Determine if the last undo patch action was moveThoughtUp or moveThoughtDown, and its type. */
+  const lastMoveType = useSelector((state: State): 'moveThoughtUp' | 'moveThoughtDown' | null => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    const movePatch = lastPatches?.find(patch =>
+      ['moveThoughtUp', 'moveThoughtDown'].includes(patch.actions[0] as ActionType),
+    )
+    return (movePatch?.actions[0] as 'moveThoughtUp' | 'moveThoughtDown') ?? null
+  })
+
+  const isMoveThought = lastMoveType !== null
+
+  // Id of thought displaced by cursor during moveThoughtUp/Down
+  const displacedThoughtId = useSelector((state: State) => {
+    if (!isMoveThought) return null
+    if (!state.cursor) return null
+    if (lastMoveType === 'moveThoughtUp') {
+      const next = nextSibling(state, state.cursor)
+      return next?.id ?? null
+    }
+    if (lastMoveType === 'moveThoughtDown') {
+      const prev = prevSibling(state, state.cursor, {})
+      return prev?.id ?? null
+    }
+    return null
+  })
+
   // We split x and y transitions into separate nodes to:
   // 1. Allow independent timing curves for horizontal and vertical movement
   // 2. Create L-shaped animation paths by controlling when each movement starts/ends
@@ -223,6 +251,7 @@ const TreeNode = ({
     width: isTableCol1 ? width : `calc(100% - ${x}px + 1em + 10px)`,
     ...(style || {}),
     ...fauxCaretNodeProvider,
+    ...(isMoveThought && isCursor ? { zIndex: 2 } : {}),
   }
 
   const innerDivStyle =
@@ -233,13 +262,44 @@ const TreeNode = ({
         }
       : undefined
 
+  // Determine FadeTransition type for moveThought animations
+  const currentMoveFadeType: 'moveThoughtCursor' | 'moveThoughtSibling' | null = isMoveThought
+    ? isCursor
+      ? 'moveThoughtCursor'
+      : thoughtId === displacedThoughtId
+        ? 'moveThoughtSibling'
+        : null
+    : null
+
+  // Latch the moveFadeType for the lifetime of this component instance so that
+  // the animation is not overridden by subsequent non-move actions.
+  const moveFadeRef = useRef<'moveThoughtCursor' | 'moveThoughtSibling' | null>(null)
+  if (currentMoveFadeType && !moveFadeRef.current) {
+    moveFadeRef.current = currentMoveFadeType
+  }
+
+  const moveFadeType = moveFadeRef.current
+
+  // Unique id ensures CSSTransition replays the enter animation for each move action
+  const undoIndex = useSelector(state => state.undoPatches.length)
+  const fadeId = moveFadeType ? `${thoughtKey}-${undoIndex}` : thoughtKey
+
+  // Final fade transition variant
+  const fadeType = moveFadeType
+    ? moveFadeType
+    : isEmpty
+      ? 'nodeFadeIn'
+      : isLastActionDelete
+        ? 'nodeDissolve'
+        : (contextAnimation ?? 'nodeFadeOut')
+
   return (
     <FadeTransition
-      id={thoughtKey}
+      id={fadeId}
       // The FadeTransition is only responsible for fade in on new thought and fade out on unmount. See autofocusChanged for autofocus opacity transition during navigation.
       // Archive, delete, and uncategorize get a special dissolve animation.
       // Context view children get special disappearing text animations
-      type={isEmpty ? 'nodeFadeIn' : isLastActionDelete ? 'nodeDissolve' : (contextAnimation ?? 'nodeFadeOut')}
+      type={fadeType}
       nodeRef={fadeThoughtRef}
       in={transitionGroupsProps.in}
       unmountOnExit
@@ -250,7 +310,6 @@ const TreeNode = ({
         aria-label='tree-node'
         className={css({
           position: 'absolute',
-
           transition: isSwap
             ? swapDirection === 'clockwise'
               ? 'left {durations.layoutNodeAnimation} {easings.nodeCurveXLayerClockwise}'
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -162,7 +162,7 @@ const TreeNode = ({
     [index, isLastActionSort],
   )
 
-  const { isMoveAnimating, moveType, moveDivStyle } = useMoveThoughtAnimation({ thoughtId, isCursor })
+  const { isMoveAnimating, moveType, moveDivStyle } = useMoveThoughtAnimation({ indexChanged, isCursor })
 
   /**
    * Horizontal offset for the first frame of a sort animation. This allows the X transition
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -13,6 +13,7 @@ import isContextViewActive from '../selectors/isContextViewActive'
 import isCursorGreaterThanParent from '../selectors/isCursorGreaterThanParent'
 import nextSibling from '../selectors/nextSibling'
 import prevSibling from '../selectors/prevSibling'
+import head from '../util/head'
 import equalPath from '../util/equalPath'
 import isDescendantPath from '../util/isDescendantPath'
 import parentOf from '../util/parentOf'
@@ -178,19 +179,37 @@ const TreeNode = ({
       : null
   })
 
-  // Id of thought displaced by cursor during moveThoughtUp/Down
-  // This is to isolate the animation to apply only to the moving thought.
+  // Id of the thought that was displaced by the cursor during a moveThoughtUp/Down operation.
+  // Normally, this is the sibling that swaps places with the cursor.
+  // In cross-context moves (e.g., moving the only child of A into the next uncle B),
+  // there is no displaced sibling, so we animate the parent that just lost the cursor (now the cursor's new uncle).
   const movingThoughtId = useSelector((state: State) => {
-    if (!lastMoveType) return null
-    if (!state.cursor) return null
+    if (!lastMoveType || !state.cursor) return null
+
+    // Immediate siblings of the cursor at its NEW location
+    const next = nextSibling(state, state.cursor)
+    const prev = prevSibling(state, state.cursor, {})
+
     if (lastMoveType === 'moveThoughtUp') {
-      const next = nextSibling(state, state.cursor)
-      return next?.id ?? null
+      // Try displaced sibling first (next if available, otherwise prev)
+      const displaced = next ?? prev
+      if (displaced) return displaced.id
+
+      // No sibling means cross-context move. Animate the parent that just lost the cursor (now the cursor's new uncle).
+      const oldParent = nextSibling(state, parentOf(state.cursor!))
+      return oldParent?.id ?? null
     }
+
     if (lastMoveType === 'moveThoughtDown') {
-      const prev = prevSibling(state, state.cursor, {})
-      return prev?.id ?? null
+      // Try displaced sibling first (prev if available, otherwise next)
+      const displaced = prev ?? next
+      if (displaced) return displaced.id
+
+      // No sibling means cross-context move. Animate the uncle (new parent) that gained the cursor.
+      const newParentId = head(parentOf(state.cursor!))
+      return newParentId ?? null
     }
+
     return null
   })
 
@@ -240,7 +259,9 @@ const TreeNode = ({
 
   useLayoutEffect(() => {
     // Start moveThought animation only when the change originated from a moveThought action.
-    if (lastMoveType && movingThoughtId) {
+    // In moves that cross contexts (e.g. moving the only child of «A» below its next uncle «B»),
+    // there is no displaced sibling, so `movingThoughtId` is null. Still animate the cursor node.
+    if (lastMoveType && (isCursor || movingThoughtId)) {
       setIsMoveThoughtAnimating(true)
     }
 
@@ -251,7 +272,7 @@ const TreeNode = ({
     }, durations.layoutNodeAnimation)
 
     return () => clearTimeout(timer)
-  }, [lastMoveType, movingThoughtId])
+  }, [lastMoveType, movingThoughtId, isCursor])
 
   // List Virtualization
   // Do not render thoughts that are below the viewport.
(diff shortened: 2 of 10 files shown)


zukilover:
I managed to get it working by applying the animation directly in TreeNode. This is because when moving the thought up or down, <FadeTransition>'s `in` prop doesn't change, so it doesn't trigger a re-render.

zukilover:
It turns out I can also move sub thought up/down. I need to make sure animation works for this case too.

zukilover:
<details>
<summary>How it works</summary>

https://github.com/user-attachments/assets/8610109f-94cf-421f-a379-4c5db7385134


</details>

src/components/TreeNode.tsx
```diff
@@ -231,7 +303,31 @@ const TreeNode = ({
           top: y,
           left: 0,
         }
-      : undefined
+      : isMoveAnimating
+        ? moveType === 'moveThoughtCursor'
+          ? {
+              transformOrigin: 'left',
+              transform: 'scale3d(1.5, 1.5, 1)',
+              transition: `transform {durations.layoutNodeAnimation} ease-out`,
```
raineorshine: The `moveThoughtCursor` snaps a bit too quickly at the end of the animation. Maybe it's the choice of easing function, I'm not totally sure. Could you see if you can smooth it out?

src/components/TreeNode.tsx
```diff
@@ -231,7 +303,31 @@ const TreeNode = ({
...
+            }
+          : moveType === 'moveThoughtSibling'
+            ? {
+                transformOrigin: 'left',
+                transform: 'scale3d(0.5, 0.5, 1)',
+                filter: 'blur(2px)',
+                opacity: 0,
```
raineorshine: Let's not fade to `opacity: 0`, as that makes it look like the thought has completely disappeared. They have swapped places and should both always be visible.

src/components/TreeNode.tsx
```diff
@@ -167,6 +171,57 @@ const TreeNode = ({
...
+    return lastPatches?.some(patch => ['moveThoughtUp', 'moveThoughtDown'].includes(patch.actions[0]))
+      ? (lastPatches[0].actions[0] as 'moveThoughtUp' | 'moveThoughtDown')
+      : null
+  })
+
+  // Id of the thought that was displaced by the cursor during a moveThoughtUp/Down operation.
+  // Normally, this is the sibling that swaps places with the cursor.
+  // In cross-context moves (e.g., moving the only child of A into the next uncle B),
+  // there is no displaced sibling, so we animate the parent that just lost the cursor (now the cursor's new uncle).
+  const movingThoughtId = useSelector((state: State) => {
```
raineorshine: This is good in some cases, but misses others.

1. Table view
2. Pinned thoughts
3. Descendants

I won't go to the trouble of creating steps to reproduce for each case, as they are pretty easy to discover. More importantly, I think we need a more general solution rather than trying to handle individual cases.

I would consider trying to solve it in a more generic way: Just look at which `TreeNode` y positions have increased and which have decreased after `moveThoughtUp/Down`. That will tell you which animation to use. 

The only exception I can think of is table col1's animation when col2 is moved, but that will be easy to exclude. (It's an exception because the col1 thought moves as the col2 heights change, but does not overlap the cursor thought, therefore does not need to be animated to move "underneath" the cursor thought.)

raineorshine (changes requested):
Thanks so much for your submission. The animation looks good in the basic case, and I appreciate the effort to make it work in the uncle case.

I am seeing an error in the console when `moveThoughtDown` is activated quickly twice in a row:
Brave Browser 2025-07-09 11 23 49

If you could wrap all this logic in a custom hook, that would help encapsulate the logic.

Thanks, and let me know if you have any questions.

src/components/TreeNode.tsx
zukilover: @raineorshine The current logic works for:
1. table view (col 1)
2. pinned thoughts
3. descendants

I have overlooked the table view - col 2 case.
Here's how it works for each of the cases above:
<details>
<summary>table view (col 1)</summary>

https://github.com/user-attachments/assets/704d1a86-baad-49a6-bb31-3c718e730fa7


</details>

<details>
<summary>pinned thoughts</summary>

https://github.com/user-attachments/assets/d3403be5-912e-4b33-841b-2f3f03932980


</details>

<details>
<summary>descendants</summary>

https://github.com/user-attachments/assets/73cca1e1-dc10-4221-b9bc-ca1aeddce1e4


</details>

Can you confirm if those above work as you'd expect?

src/components/TreeNode.tsx
zukilover: This happened because the moving cursor did not have transition when the animation is finished. I've added `transform` transition to smoothen it out.

<details>
<summary>Smooth animation</summary>

https://github.com/user-attachments/assets/55a05c3e-3265-447b-a6d0-f2261e33ab8b

</details>

src/components/TreeNode.tsx
zukilover: This is now using `opacity: 0.5` so that it stays visible.

src/components/TreeNode.tsx
raineorshine: Much better!

src/components/TreeNode.tsx
raineorshine: I don't want to deal with this on a case-by-case basis, as it'll explode the complexity.

The descendant case is that a moved thought *and* all of its expanded descendants should use the custom animation. That's really the only behavior that makes sense from a UX perspective. Any expanded descendants would overlap the counter-moved thought, thus would need to be animated above or below to avoid the appearance of them sliding through each other. (If this were a special case I would consider it out of scope of the original task... but it should be solved automatically by the suggested solution.)

There's also a case where the second move after moving into a pinned thought does not use the correct animation. That indicates the logic has "memory" beyond a single move, which is bad news.

Trust me, trying to address these individually is asking for trouble. Please let me know if you have any concerns about the simpler and more general approach based on the direction of movement.

zukilover: Thanks. I'm still working on this. What I have found is that somehow Y is changed twice when moving up/down. The second change has 4.5px difference than the first one. I wonder where this comes from.

zukilover: I see it comes from the `cliffPadding`. Let me find a way so that the y changes will be read once by the move animation hook.

zukilover: Using `y` changes turned out to be tricky. If I understand correctly, it works by first estimating a node's height, followed by an `updateSize` call for the final height. Using `y` changes is complicated because it triggers a re-render to settle the final layout.

I'm now trying to use `indexChanged`, since moving a thought up or down changes its index. One edge case I’ve noticed is that in table view, when I move a thought in the second column for the first time, it pushes the column down. It also appears that the second row of the first column changes its index.

I'm now trying to find a way to isolate that case.

zukilover: @raineorshine please let me know if it's acceptable to use `indexChanged` instead of `y` to check if node has moved:

https://github.com/cybersemics/em/blob/67da66ceb9858ea6a605cd683d43d5cd3f583566/src/hooks/useMoveThoughtAnimation.ts#L61-L66

raineorshine: Yes, `index` seeks even better than `y` position. Good idea!

zukilover:
Hi @raineorshine, I've actually been doing the update with performance in mind. I’ve consolidated the `useSelector` to return a single shallow object with the two values needed (`lastMoveType` and `isSiblingOfCursor`) and wrapped it with shallowEqual to avoid triggering re-renders. Also short-circuiting where possible (like skipping `equalPath` if `isCursor` is true), and memoizing intermediate values like `hasMoved` so they're only recomputed when needed.

Let me know if there's something I've overlooked.

raineorshine:
Great, thanks so much for your attention to detail. 

Let me know if it's ready for review.

raineorshine (changes requested):
I will just do a behavioral review for now.

This shouldn't be limited to the cursor thought and siblings. ALL thoughts that move up during `moveThoughtUp` need to swell, and ALL thoughts that move down during `moveThoughtUp` need to shrink and fade. 

It should be very simple and general. The only special case is table col2, but I'll go over that once we get the basic functionality correct.

Table view:

https://github.com/user-attachments/assets/791e69b2-9224-447a-8855-20dd0c6f36cf

Multicursor:

https://github.com/user-attachments/assets/6cf9c27a-edfb-4bc6-8e5d-94a959bc8056

raineorshine:
Looks great! Will do a thorough review soon.

src/components/TreeNode.tsx
```diff
@@ -161,6 +162,12 @@ const TreeNode = ({
     [index, isLastActionSort],
   )
 
+  const { isMoveAnimating, moveType, moveDivStyle } = useMoveThoughtAnimation({
+    indexChanged,
+    previousIndex,
+    index,
```
raineorshine: `indexChanged` and `previousIndex` are trivial to calculate, so there's no harm in recalculating them within `useMoveThoughtAnimation`. This would be a much better design from an API perspective. Reduce complexity by hiding it within the custom hook. Then the hook is a self-contained unit, rather than just an extension of `TreeNode`.

Currently, the inputs and outputs of `useMoveThoughtAnimation` feel too arbitrary. They have no sense of singular purpose.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Return type of useMoveThoughtAnimation.
+ */
+export interface MoveThoughtAnimation {
+  isMoveAnimating: boolean
+  /**
+   * Animation - moveThoughtAction: Node moved in the same direction as the action (primary).
+   * Animation - moveThoughtDisplaced: Node moved opposite to the action direction (secondary).
+   */
```
raineorshine: Good, I like the naming here.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,127 @@
...
+  moveType: 'moveThoughtAction' | 'moveThoughtDisplaced' | null
+  moveDivStyle: React.CSSProperties | undefined
```
raineorshine: Always document at the interface level.

src/components/TreeNode.tsx
```diff
@@ -280,7 +295,7 @@ const TreeNode = ({
                 : 'top {durations.layoutNodeAnimation} {easings.nodeCurveYLayer}'
               : isLastActionSort
                 ? 'top {durations.layoutNodeAnimation} {easings.nodeCurveSortYLayer}'
-                : undefined,
+                : moveAnimation,
```
raineorshine: Why are some styles applied in `moveDivStyle` and some applied in `moveAnimation`? They are being applied to the same element from what I can tell.

src/components/TreeNode.tsx
```diff
@@ -231,7 +238,16 @@ const TreeNode = ({
...
+      : moveDivStyle
+
+  const moveAnimation =
+    isMoveAnimating && moveType === 'moveThoughtDisplaced'
+      ? `transform {durations.layoutNodeAnimation} ease-out, filter {durations.layoutNodeAnimation} ease-out, opacity {durations.layoutNodeAnimation} ease-out`
+      : isMoveAnimating && moveType === 'moveThoughtAction'
+        ? `transform {durations.layoutNodeAnimation} ease-out`
+        : moveType === 'moveThoughtDisplaced'
+          ? `opacity {durations.layoutNodeAnimation} ease-out`
+          : undefined
```
raineorshine: Could this logic be moved into the custom hook?

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,127 @@
...
+    return moveType
+  })
+
+  // Capture the previous value of indexChanged
+  const prevIndexChanged = usePrevious(indexChanged)
+
+  // Capture the proposed moveType
+  const hasMoved = useMemo(() => {
+    return !!(lastMoveType && indexChanged && !prevIndexChanged)
+  }, [lastMoveType, indexChanged, prevIndexChanged])
```
raineorshine: This is trivial and `boolean` is compared by value, so there is no need to memoize it.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    proposedMoveType = isPrimary ? 'moveThoughtAction' : 'moveThoughtDisplaced'
+  }
+
+  // Implement the move type on the next render cycle
+  useLayoutEffect(() => {
+    if (proposedMoveType) {
+      setMoveType(proposedMoveType)
+    }
+  }, [proposedMoveType])
```
raineorshine: It looks like `proposedMoveType` is a deterministic function of `index`, `previousIndex`, `lastMoveType`, and `isPrimary`. Can you explain why the additional setter and layout effect is needed?

This feels like too much complexity for what it's doing, but maybe I'm missing something.

zukilover:
> The main issue I'm seeing is with the second half of the animation. It should be a smooth transition through the entire animation, but I'm seeing an abrupt jump in opacity and scale at the end.

I'm still working on this, and it's quite tricky. I've adjusted the animation timer to be 3/4 of `durations.layoutNodeAnimation`. This is to fix the `maximum update depth exceeded`.

https://github.com/cybersemics/em/blob/2d8bb659cb8f3f892bd0a885e08f17100c3b1d4d/src/hooks/useMoveThoughtAnimation.ts#L90-L92

Let me find an alternative as not to affect the animation while still avoid the race condition.

zukilover:
I've checked more closely and it seems that the error is triggered from this line:
https://github.com/cybersemics/em/blob/2d8bb659cb8f3f892bd0a885e08f17100c3b1d4d/src/components/Editable/useMultiline.ts#L23-L25

I can see the function is not optimized so that on every render, it creates a new function object.
I propose that we wrap it in a `useCallback` hook as follows:

diff --git a/src/components/Editable/useMultiline.ts b/src/components/Editable/useMultiline.ts
index <old>..<new> 100644
--- a/src/components/Editable/useMultiline.ts
+++ b/src/components/Editable/useMultiline.ts
@@
-import { useEffect, useState } from 'react'
+import { useEffect, useState, useCallback } from 'react'
@@
-  /** Measure the contentRef to determine if it needs to be multiline. */
-  const updateMultiline = () => {
+  /** Measure the contentRef to determine if it needs to be multiline. */
+  const updateMultiline = useCallback(() => {
@@
-    setMultiline(height - paddingTop - paddingBottom > singleLineHeight * 1.2)
-  }
+    setMultiline(height - paddingTop - paddingBottom > singleLineHeight * 1.2)
+  }, [fontSize, contentRef])
@@
-  useLayoutAnimationFrameEffect(updateMultiline, [contentRef, fontSize, isEditing, simplePath, editingValue])
+  useLayoutAnimationFrameEffect(updateMultiline, [fontSize, isEditing, simplePath, editingValue])


@raineorshine let me know if this is a good idea.

src/hooks/useMoveThoughtAnimation.ts
zukilover: VirtualThought recalculate heights from ths line: https://github.com/cybersemics/em/blob/7a51ebcb9c7263a0f306016f56d08fedf56c6c0a/src/components/VirtualThought.tsx#L161-L163 

It causes the node to render twice. I need to resolve the move type once and keep it until the height calculation is finished. Using useLayoutEffect makes sure it holds the value correctly, otherwise it will incorrectly think index has not changed (from the final re-render).

raineorshine:
> > The main issue I'm seeing is with the second half of the animation. It should be a smooth transition through the entire animation, but I'm seeing an abrupt jump in opacity and scale at the end.
> 
> I'm still working on this, and it's quite tricky. I've adjusted the animation timer to be 3/4 of `durations.layoutNodeAnimation`. This is to fix the `maximum update depth exceeded`.
> 
> https://github.com/cybersemics/em/blob/2d8bb659cb8f3f892bd0a885e08f17100c3b1d4d/src/hooks/useMoveThoughtAnimation.ts#L90-L92
> 
> Let me find an alternative as not to affect the animation while still avoid the race condition.

It would be nice if it could be modeled as a single animation. Could it be done with `@keyframes` instead of `transition`?

> I've checked more closely and it seems that the error is triggered from this line:
> 
> https://github.com/cybersemics/em/blob/2d8bb659cb8f3f892bd0a885e08f17100c3b1d4d/src/components/Editable/useMultiline.ts#L23-L25
> 
> I can see the function is not optimized so that on every render, it creates a new function object. I propose that we wrap it in a `useCallback` hook as follows:

Yup, that's fine. In fact, it's already [proposed in an open PR](https://github.com/cybersemics/em/pull/3118/files#diff-a832f94f5ebc6ce543287634224a7a0295fd51cbeaa8cdb4b1b615e9b8c14b97R24).

zukilover:
@raineorshine for the table view case, I cannot find an elegant of fixing it other than modifying the reducers directly and add a skip animation flag there. This is because I need to know when moving the thought up/down whether it is doing a row-jump or swapping with a sibling.

I propose something like this:

```diff
diff --git a/src/actions/moveThoughtDown.ts b/src/actions/moveThoughtDown.ts
index 250471fd0..265edf873 100644
--- a/src/actions/moveThoughtDown.ts
+++ b/src/actions/moveThoughtDown.ts
@@ -3,6 +3,7 @@ import State from '../@types/State'
 import Thunk from '../@types/Thunk'
 import alert from '../actions/alert'
 import moveThought from '../actions/moveThought'
+import attributeEquals from '../selectors/attributeEquals'
 import * as selection from '../device/selection'
 import findDescendant from '../selectors/findDescendant'
 import getPrevRank from '../selectors/getPrevRank'
@@ -27,7 +28,7 @@ const moveThoughtDown = (state: State): State => {
   const parentId = head(pathParent)
   const nextThought = nextSibling(state, cursor)
 
-  // if the cursor is the last child, move the thought to the beginning of its next uncle
+  // Determine if this move crosses contexts (no next sibling, moving to next uncle) and occurs wit
hin a Table View.
   const nextUncleThought = pathParent.length > 0 ? nextSibling(state, pathParent) : null
   const nextUnclePath = nextUncleThought ? appendToPath(parentOf(pathParent), nextUncleThought.id) 
: null
 
@@ -54,6 +55,10 @@ const moveThoughtDown = (state: State): State => {
   // store selection offset before moveThought is dispatched
   const offset = selection.offset()
 
+  const isCrossContext = !nextThought && !!nextUnclePath
+  const isTableView = attributeEquals(state, cursor[0], '=view', 'Table')
+  const skipMoveAnimation = isCrossContext && isTableView
+
   const rankNew = nextThought
     ? // next thought
       getRankAfter(state, simplifyPath(state, pathParent).concat(nextThought.id) as SimplePath)
@@ -63,12 +68,18 @@ const moveThoughtDown = (state: State): State => {
   const newPathParent = nextThought ? pathParent : nextUnclePath!
   const newPath = appendToPath(newPathParent, head(cursor))
 
-  return moveThought(state, {
+  const movedState = moveThought(state, {
     oldPath: cursor,
     newPath,
     ...(offset != null ? { offset } : null),
     newRank: rankNew,
   })
+
+  // Inject the skipMoveThoughtAnimation flag so that UI hooks can skip animations when needed.
+  return {
+    ...movedState,
+    skipMoveThoughtAnimation: skipMoveAnimation,
+  }
 }
 
 /** Action-creator for moveThoughtDown. */
```

This may seem complicated but honestly I couldn't find a better way than this.

zukilover:
> It would be nice if it could be modeled as a single animation. Could it be done with `@keyframes` instead of `transition`?

Yes we could do that. But it's not wired up to the pandaCSS, so we need to inject the keyframes manually.
Something like this:

```javascript
const style = document.createElement('style')
  style.id = 'move-thought-animation-keyframes'
  style.textContent = `
    @keyframes moveThoughtAction {
      0% { transform: scale3d(1, 1, 1); }
      50% { transform: scale3d(1.5, 1.5, 1); }
      100% { transform: scale3d(1, 1, 1); }
    }
    @keyframes moveThoughtDisplaced {
      0% { transform: scale3d(1, 1, 1); opacity: 1; filter: blur(0); }
      100% { transform: scale3d(0.5, 0.5, 1); opacity: 0.5; filter: blur(2px); }
    }
  `
  document.head.appendChild(style)
```

Let me know if this is acceptable.

src/components/TreeNode.tsx
zukilover: I have now included `index` as the only option for the hook.

src/hooks/useMoveThoughtAnimation.ts
zukilover: I've added comment to the interface.

src/components/TreeNode.tsx
zukilover: This was because moveAnimation added to the `css()` and moveDivStyle to the inline style. I've now removed `moveAnimation` completely and consolidate the inline style at one place within the hook.

src/components/TreeNode.tsx
zukilover: OK. It's moved.

src/hooks/useMoveThoughtAnimation.ts
zukilover: Removed `useMemo`.

raineorshine (changes requested):
> @raineorshine for the table view case, I cannot find an elegant of fixing it other than modifying the reducers directly and add a skip animation flag there. This is because I need to know when moving the thought up/down whether it is doing a row-jump or swapping with a sibling.

No, State is definitely not the place for this. I shouldn't have to explain how bad this idea is 😕.

If you need to take a break and come back to it to have a fresh look, please do.

> > It would be nice if it could be modeled as a single animation. Could it be done with `@keyframes` instead of `transition`?
> 
> Yes we could do that. But it's not wired up to the pandaCSS, so we need to inject the keyframes manually.

https://panda-css.com/docs/customization/config-functions#definekeyframes

src/hooks/useMoveThoughtAnimation.ts
raineorshine: The added state and indirection is really adding to the complexity of your solution. If you want to avoid running it twice in a short period, what about a simple throttle? I'd like to avoid `setMoveType` if possible.

Overall I'm finding this PR to be a bit verbose for what it does. I'm hesitant to accept this, as it will be hard for other developers to understand and maintain, even if it is more encapsulated in the custom hook.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,132 @@
...
+  // Override animation output when skipping is requested.
+  if (skipMoveAnimation) {
+    return {
+      moveDivStyle: undefined,
+    }
+  }
+
+  return {
+    moveDivStyle,
+  }
```
raineorshine: Why is this wrapped in an object?

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }, [proposedMoveType])
+
+  // Manage the move animation flag
+  useLayoutEffect(() => {
+    // After the layout node animation duration reset the flag so that subsequent
+    // updates can re-trigger the animation.
+    const timer = setTimeout(() => {
+      setMoveType(null)
+    }, durations.layoutNodeAnimation)
```
raineorshine: What if the animation is interrupted? This seems to assume that the `layoutNodeAnimation` will always complete.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,132 @@
...
+  const prevIndexChanged = usePrevious(indexChanged)
+
+  // Capture the proposed moveType. Only compute on the render where indexChanged flips to true.
+  const hasMoved = !!(lastMoveType && indexChanged && !prevIndexChanged)
+
+  const [moveType, setMoveType] = useState<'moveThoughtAction' | 'moveThoughtDisplaced' | null>(null)
+
+  // skip move animation if the last move was a cross-context move in a table view
+  const skipMoveAnimation =
+    (lastMoveType === 'moveThoughtUp' || lastMoveType === 'moveThoughtDown') && isCrossContext && isTableView
```
raineorshine: Skipping the table case doesn't seem to be working for me. It still runs the special animation:

https://github.com/user-attachments/assets/9f0281c0-fa3b-4990-8fea-b3cbf447da2e

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const cursor = state.cursor
+    const isTableView = cursor ? attributeEquals(state, cursor[0], '=view', 'Table') : false
+    const isCrossContext = (lastPatches ?? []).some(patch => patch.path?.endsWith('/parentId'))
+
+    return {
+      lastMoveType,
+      isTableView,
+      isCrossContext,
+    }
```
raineorshine: By returning a composite object, you're missing the value of combining multiple `useSelector` into a single `useSelector`. Composite object return types are ugly and should be avoided anyway. There's no elegance in a composite object of arbitrary values.

Notice how `isTableView` and `isCrossContext` are *only ever used in conjunction* in the expression `isCrossContext && isTableView`. You do not need to return them independently. It's less encapsulated, and it's worse performance because the memo will be refreshed when *either* of them change.

*Encapsulate* to hide complexity. *Combine* logic early to improve performance and reduce complexity downstream.

raineorshine (changes requested):
Hi, thanks for the update.

src/hooks/useMoveThoughtAnimation.ts
zukilover: This was left when I refactored the hook. It now simply returns the style.

src/hooks/useMoveThoughtAnimation.ts
zukilover: With the implementation using `throttle`, this should ignore any further calls for the next `layoutNodeAnimation` ms.

src/hooks/useMoveThoughtAnimation.ts
zukilover: It works for me. Given the same structure, I put cursor on one, move it up, it skips the curve animation.

https://github.com/user-attachments/assets/6f52ea46-5547-478b-af3d-86d5d7276d37

src/hooks/useMoveThoughtAnimation.ts
zukilover: I’ve updated the hook to return a tuple `[lastMoveType, skipMoveAnimation]` instead of a composite object. I combined `isTableView` and `isCrossContext` early into skipMoveAnimation to keep things cleaner and avoid extra memo updates. Let me know if anything still feels off.

https://github.com/cybersemics/em/blob/b497f9b43f2e99639e1f6d0c4da2e80d2c42aaa0/src/hooks/useMoveThoughtAnimation.ts#L26-L44

src/hooks/useMoveThoughtAnimation.ts
raineorshine: There appears to be an unintended depth dependency. If you insert a new thought as a parent of `B`, the col2 thoughts incorrectly use the curve animation between rows.

I'm also seeing false negatives when moving `one`, `two`, `three`, `four` within the same context if you play around with it.

Let me know if you have any trouble reproducing and I can provide exact steps to reproduce!

src/hooks/useMoveThoughtAnimation.ts
raineorshine: Okay, thanks.

FYI A tuple is a composite object, but worse because the entries are unnamed.

zukilover:
I have to bring back the internal state as I've just noticed that it failed to trigger the right animations without it.
I'm now still working to find a better solution that will let us avoid internal state.

zukilover:
I tried a few stateless refactors, but still, it needs a render after the animation ends, or it won’t know the style was cleared and the next animation won’t run.

I tested a few ways:

Ref + timestamp: unreliable without a state trigger

Ref + throttle + useReducer bump: works (but not pretty)

animationend listener: still needs a setState inside

Pure refs: no render = no animation

Here's how the fix with useReducer looks like:

```diff
diff --git a/src/hooks/useMoveThoughtAnimation.ts b/src/hooks/useMoveThoughtAnimation.ts
index c783dd16c..1d3ed78d9 100644
--- a/src/hooks/useMoveThoughtAnimation.ts
+++ b/src/hooks/useMoveThoughtAnimation.ts
@@ -1,4 +1,5 @@
-import { useMemo, useRef } from 'react'
+import { useMemo, useReducer, useRef } from 'react'
+import { throttle } from 'lodash'
 import { useSelector } from 'react-redux'
 import State from '../@types/State'
 import durations from '../durations.config'
@@ -82,21 +83,28 @@ const useMoveThoughtAnimation = ({ index }: Options): MoveThoughtAnimation => {
   // update while still persisting data between renders.
 
   const moveTypeRef = useRef<'moveThoughtAction' | 'moveThoughtDisplaced' | null>(null)
-  const expiryRef = useRef<number>(0)
 
-  // If this thought moved in this render cycle, update the refs so that the
-  // style will be applied immediately (no need for a setState or effect).
+  // A throttled function that will clear moveTypeRef after the animation duration.
+  const clearMoveTypeRef = useRef<() => void>()
+
+  // Force a re-render when the throttled clear function executes.
+  const [, forceRender] = useReducer((x: number) => x + 1, 0)
+
+  if (!clearMoveTypeRef.current) {
+    clearMoveTypeRef.current = throttle(() => {
+      moveTypeRef.current = null
+      // ensure React re-renders so the style is removed immediately
+      forceRender()
+    }, durations.layoutNodeAnimation, { leading: false })
+  }
+
+  // If this thought moved in this render cycle, update the ref and schedule its clearance.
   if (hasMoved) {
     moveTypeRef.current = getMoveType(lastMoveType, previousIndex, index)
-    // Store the time (in ms since epoch) when the animation should stop
-    expiryRef.current = Date.now() + durations.layoutNodeAnimation
+    clearMoveTypeRef.current!()
   }
 
-  // Determine if the animation is still within its active window.
-  const isAnimating = Date.now() < expiryRef.current
-  const moveType: 'moveThoughtAction' | 'moveThoughtDisplaced' | null = isAnimating
-    ? moveTypeRef.current
-    : null
+  const moveType: 'moveThoughtAction' | 'moveThoughtDisplaced' | null = moveTypeRef.current
 
   const moveDivStyle = useMemo<React.CSSProperties | undefined>(() => {
     if (!moveType) return undefined
```

zukilover:
I'm exploring other approaches now. I'd appreciate ideas here.

src/hooks/useMoveThoughtAnimation.ts
zukilover: This has been fixed. 

https://github.com/cybersemics/em/blob/9b7c69e8962d5a5f1f2fd20a7bcfe61539286227/src/hooks/useMoveThoughtAnimation.ts#L60-L63

src/hooks/useMoveThoughtAnimation.ts
zukilover: Got it. I've removed the object completely. This now returns a primitive.
https://github.com/cybersemics/em/blob/9b7c69e8962d5a5f1f2fd20a7bcfe61539286227/src/hooks/useMoveThoughtAnimation.ts#L66-L68

zukilover:
I've completely removed the complicated logic with `proposedMoveType` and replaced it with a new `getMoveType` function to better encapsulate the logic. `useState` is still needed here to trigger the cleanup of the animation so it runs correctly. The depth bug has also been fixed, here’s how it works:

<details>
<summary>depth animation</summary>

https://github.com/user-attachments/assets/85ac24ea-94a9-4e36-b419-267910882698


</details>

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,126 @@
...
+import { useSelector } from 'react-redux'
+import State from '../@types/State'
+import durations from '../durations.config'
+import attributeEquals from '../selectors/attributeEquals'
+import usePrevious from './usePrevious'
+
+/**
+ * The style to apply to the move div.
+ */
+export type MoveThoughtAnimation = React.CSSProperties | undefined
```
raineorshine: Aliasing this type is just adding an extra level of indirection that's not necessary. I would remove it and use `React.CSSProperties` directly. Or even better, narrow the type to just the CSS properties that are actually used.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,126 @@
...
+
+interface Options {
+  /** The current on-screen index of the thought. */
+  index: number
+}
```
raineorshine: No need to wrap a single argument in an object.

Prefer inlining and co-location of small types.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Determines the move animation type for a thought based on the previous and current indices and the last move action.
+ * Returns:
+ * - 'moveThoughtAction' when this thought is the one explicitly moved by the user.
+ * - 'moveThoughtDisplaced' when this thought is displaced by another thought moving past it.
+ * - null when the movement does not require an animation (e.g. indices unchanged or lastMoveType is null).
+ */
+const getMoveType = (
+  lastMoveType: 'moveThoughtUp' | 'moveThoughtDown' | null,
```
raineorshine: The function `getMoveType` takes `lastMoveType` and returns `nextMoveType`, which implies that they are the same type. However, we're actually dealing with two different types: `'moveThoughtUp' | 'moveThoughtDown' | null` and `'moveThoughtAction' | 'moveThoughtDisplaced'`.

I would suggest `lastMoveAction` and `moveAnimationName` (or appropriate variations) to make it clear that these are different types.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,126 @@
...
+  previousIndex: number | undefined,
+  currentIndex: number,
+): 'moveThoughtAction' | 'moveThoughtDisplaced' | null => {
+  if (!lastMoveType || previousIndex === undefined) return null
+
+  // Determine actual movement direction of this node relative to its previous on-screen index.
+  const direction: 'up' | 'down' = currentIndex < previousIndex ? 'up' : 'down'
+
+  const isPrimary =
```
raineorshine: `isMovingWithCursor` might be clearer.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,126 @@
...
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+
+    // Determine if the last action was a moveThoughtUp or moveThoughtDown.
+    const moveType = lastPatches?.some(patch => ['moveThoughtUp', 'moveThoughtDown'].includes(patch.actions[0]))
+      ? (lastPatches[0].actions[0] as 'moveThoughtUp' | 'moveThoughtDown')
+      : null
+
+    if (!moveType) return null
+
+    // Determine if we should skip the animation (cross-context move inside a Table view).
```
raineorshine: A little more explanation in the comments here would be helpful. 

Why should the animation be skipped? 
What does "cross context" mean? 
Why does `patch.path?.endsWith('/parentId')` work?

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,126 @@
...
+    const skipMoveAnimation = isCrossContext && isTableView
+
+    return skipMoveAnimation ? null : moveType
+  })
+
+  // Determine if the index has changed.
+  const previousIndex = usePrevious<number>(index)
+  const indexChanged = previousIndex !== undefined && previousIndex !== index
+  const prevIndexChanged = usePrevious(indexChanged)
+  const hasMoved = !!(lastMoveType && indexChanged && !prevIndexChanged)
```
raineorshine: I understand why this depends on `indexChanged`, but I'm having trouble understanding the dependency on `prevIndexChanged`. Could you explain?

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const [moveType, setMoveType] = useState<'moveThoughtAction' | 'moveThoughtDisplaced' | null>(null)
+
+  // Throttle setMoveType(null)
```
raineorshine: Why throttle?

Please explain nontrivial behavior in comments.

I think you can use `useCallback` instead of `useMemo`. See https://react.dev/reference/react/useCallback#how-is-usecallback-related-to-usememo

panda.config.ts
```diff
@@ -164,6 +164,16 @@ const keyframes = defineKeyframes({
       transform: 'translateX(-50%) translateY(0)',
     },
   },
+  moveThoughtAction: {
```
raineorshine: While I like the animation names conceptually, I think it may be a bit easier to reason about them in visual terms: `moveThoughtOver` and `moveThoughtUnder`. Could we rename these? Thanks!

raineorshine (changes requested):
Thanks, this is looking great architecturally. The state is limited to tracking the previous indexes and one local state variable to ensure the animation is only triggered once (if I'm understanding correctly). Everything is nicely encapsulated within the custom hook with a single input/output.

The requested changes are mainly for comprehensibility. Regarding the suggested renames, please make sure to go through the entire diff and carefully rename every instance and related instance. It could be easy to miss some instances or mentions in comments since they are spread throughout the diff.

Thank you!

panda.config.ts
zukilover: I've renamed the variable names to `moveThoughtOver` and `moveThoughtUnder`

src/hooks/useMoveThoughtAnimation.ts
zukilover: `!prevIndexChanged` makes sure the animation runs only when indexChanged first flips to true, so each move gets one clean animation, no repeats.

Using throttle might seem like a fix, but it's time-based, not state-aware. This approach is more reliable because it watches for actual state changes.

I've added comment to this line so that it's clearer.

src/hooks/useMoveThoughtAnimation.ts
zukilover: Comment added.

src/hooks/useMoveThoughtAnimation.ts
zukilover: Variable renamed.

src/hooks/useMoveThoughtAnimation.ts
zukilover: Thanks for the suggestion, I've implemented the changes above.

src/hooks/useMoveThoughtAnimation.ts
zukilover: `index` is now inline.

src/hooks/useMoveThoughtAnimation.ts
zukilover: I've added the specific CSS properties as suggested.

src/hooks/useMoveThoughtAnimation.ts
zukilover: I intentionally replaced `useCallback` with `useMemo` here: 9cfca8b21c8e740af9324a1effe2caf431e0e89d

Here's the story:
I initially used `useCallback` correctly like this:

```js
const clearMoveType = useCallback(
  throttle(() => setMoveType(null), delay, { leading: false }),
  [setMoveType]
);
```

However, this caused a bug where moving a thought down animated correctly, but moving it up didn't animate at all. I suspect this was because the throttled function was waiting to fire, but its internal timer got wiped out due to a re-render.

As a workaround, I tried a slightly hacky version:

```js
const clearMoveType = useCallback(() => {
    const throttled = throttle(() => setMoveType(null), durations.layoutNodeAnimation, {
      leading: false,
    })
    throttled()
  }, [setMoveType])
```

This worked because the throttled function was created and called immediately on each render. But this defeats the purpose of throttling, it bypasses internal timing and just runs once per call.

More importantly, this hurts performance, since the throttled function is recreated on every invocation. I was memoizing the wrapper, not the throttled function itself.

So I switched to `useMemo`, which correctly memoizes a value, in this case, the throttled function itself:

```js
const clearMoveType = useMemo(() =>
  throttle(() => setMoveType(null), delay, { leading: false }),
  [setMoveType]
);
```

This resolves the animation issue and ensures throttling works as intended, with better performance.

zukilover: I've added the comment on throttle.

raineorshine: Interesting! Thanks for the explanation. Apparently the additional internal state of `throttle` invalidates what would otherwise be functionally identical.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,150 @@
...
+  index: number,
+):
+  | {
+      transformOrigin?: React.CSSProperties['transformOrigin']
+      animationDuration?: React.CSSProperties['animationDuration']
+      animationTimingFunction?: React.CSSProperties['animationTimingFunction']
+      animationFillMode?: React.CSSProperties['animationFillMode']
+      animationName?: React.CSSProperties['animationName']
+      zIndex?: React.CSSProperties['zIndex']
+    }
```
raineorshine: A slightly more Typescripty way to do this would be:

```diff
diff --git a/src/hooks/useMoveThoughtAnimation.ts b/src/hooks/useMoveThoughtAnimation.ts
index f8fa3767d6..0cbe566eb1 100644
--- a/src/hooks/useMoveThoughtAnimation.ts
+++ b/src/hooks/useMoveThoughtAnimation.ts
@@ -39,14 +39,17 @@ const getMoveType = (
 const useMoveThoughtAnimation = (
   index: number,
 ):
-  | {
-      transformOrigin?: React.CSSProperties['transformOrigin']
-      animationDuration?: React.CSSProperties['animationDuration']
-      animationTimingFunction?: React.CSSProperties['animationTimingFunction']
-      animationFillMode?: React.CSSProperties['animationFillMode']
-      animationName?: React.CSSProperties['animationName']
-      zIndex?: React.CSSProperties['zIndex']
-    }
+  | Partial<
+      Pick<
+        React.CSSProperties,
+        | 'transformOrigin'
+        | 'animationDuration'
+        | 'animationTimingFunction'
+        | 'animationFillMode'
+        | 'animationName'
+        | 'zIndex'
+      >
+    >
   | undefined => {
   const lastMoveAction = useSelector((state: State) => {
     const lastPatches = state.undoPatches[state.undoPatches.length - 1]
```

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,150 @@
...
+  | undefined => {
+  const lastMoveAction = useSelector((state: State) => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+
+    // Determine if the last action was a moveThoughtUp or moveThoughtDown.
+    const moveType = lastPatches?.some(patch => ['moveThoughtUp', 'moveThoughtDown'].includes(patch.actions[0]))
```
raineorshine: Consistent with our renaming, this should be called `moveAction`.

src/hooks/useMoveThoughtAnimation.ts
zukilover: Thanks for the suggestion, I've fixed the type.

src/hooks/useMoveThoughtAnimation.ts
zukilover: Variables have been renamed as suggested. Thanks.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -0,0 +1,153 @@
...
+    // This creates a patch with a path ending in '/parentId'.
+    // For example: "thoughts/thoughtIndex/abc123/parentId" indicates that thought abc123
+    // is being assigned a new parent.
+    //
+    // We skip animation in these cases because:
+    // - Cross-context moves can cause thoughts to jump between table columns
+    // - The standard move animation assumes thoughts stay within the same context
+    // - Animating across column boundaries creates jarring visual transitions
+    const cursor = state.cursor
+    const isTableView = cursor ? cursor.some(id => attributeEquals(state, id, '=view', 'Table')) : false
```
raineorshine: Just noticed this, but it checks if *any* ancestor is in table view. I think you just want to check the grandparent. Otherwise it can incorrectly skip a cross contextual move animation in a deeply nested non-table thought.

zukilover:
> @zukilover There appears to be a regression for moving the second-to-last thought in a list down (or the last thought up). The position is off, and corrects itself when the cursor moves away. Can you confirm?

I can replicate the issue on mobile device by moving the thought up/down using gesture.
It works normally on desktop.

PR #3116, opened 2025-07-08
Trigger lottie animations when command item is focused for a while

yangchristina:
The icon animates after a command is highlighted for 3 seconds, and then every 5 seconds after that.

https://github.com/user-attachments/assets/c85e5c8a-74c6-42d2-b19c-907ef077017f

### DIFF
--- a/src/hooks/useLottieIntervalAnimation.ts
+++ b/src/hooks/useLottieIntervalAnimation.ts
@@ -1,10 +1,10 @@
 import { useCallback, useEffect, useRef, useState } from 'react'
 
 /**
- * A custom hook that triggers Lottie animations when a command is focused/highlighted.
+ * A custom hook that triggers Lottie animations when enabled.
  * The animation fires after 3 seconds of focus, then repeats every 5 seconds.
  */
-const useLottieOnFocus = ({ hasAnimation }: { hasAnimation: boolean }) => {
+const useLottieIntervalAnimation = ({ enabled }: { enabled: boolean }) => {
   const [isAnimated, setIsAnimated] = useState(false)
   const initialDelayRef = useRef<NodeJS.Timeout | null>(null)
   const repeatIntervalRef = useRef<NodeJS.Timeout | null>(null)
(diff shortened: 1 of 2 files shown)


yangchristina:
@raineorshine I've put it to highlight based on whether it is selected. Let me know if you want it to be mouse hover based instead. This PR is ready for review

src/hooks/useLottieIntervalAnimation.ts
```diff
@@ -0,0 +1,60 @@
...
+      }, 3000)
+    } else {
+      // Clear timers and reset animation state when not focused
+      clearTimers()
+      setIsAnimated(false)
+    }
+
+    return () => {
+      clearTimers()
+    }
```
raineorshine, line 51: FYI You can omit the empty function wrapper and return the function reference:

```ts
return clearTimers
```

src/hooks/useLottieIntervalAnimation.ts
```diff
@@ -0,0 +1,60 @@
+import { useCallback, useEffect, useRef, useState } from 'react'
+
+/**
+ * A custom hook that triggers Lottie animations when enabled.
+ * The animation fires after 3 seconds of focus, then repeats every 5 seconds.
+ */
+const useLottieIntervalAnimation = ({ enabled }: { enabled: boolean }) => {
```
raineorshine, line 7: Arguably "focus" and "Lottie" shouldn't be mentioned in the JSDOC or the name of the hook, since you actually have written something more general than that. Those are choices of the caller.

raineorshine (approved):
Thank you! Looks good!

src/hooks/useLottieIntervalAnimation.ts
yangchristina: That's true, a better name would probably be `usePlayAnimationOnInterval`

src/hooks/useLottieIntervalAnimation.ts
yangchristina: Thanks!

Issue #3096, opened 2025-06-29
Multiline line height is incorrect after Cursor Up


raineorshine:
May be the same as #3090. 

## Steps to Reproduce

Adjust the font size or screen with such that the long thought does not wrap when the cursor is on it, and wraps just before `goes` when the cursor is null. 

```
- A
  - B
    - C
      - This is a longer thought and goes
```

1. Set the cursor on the long thought. 
2. Cursor Up
3. Cursor Up
4. Cursor Up

## Current Behavior

The line height is incorrect when the cursor in on `B`. 

https://github.com/user-attachments/assets/3502b75d-6d24-4cfc-a47f-016ca2417a47

## Expected Behavior

The line height should update as soon as the line wraps.

PR #3118, opened 2025-07-08
Fix 3096: incorrect multiline height after cursor up

codebuild-ro:
Issue: #3096 
1. As you can see in the screenshot, `editingValue` was the problem.
image
`const editingValue = editingValueStore.useSelector(state => (isEditing ? state : null))`
This only returns editing thought value when isEditing is true and returns null when moving cursor with up/down keys.
It makes to lost track of value during cursor movement. 
So the useLayoutAnimationFrameEffect should be run when cursor up/down, but as the `editingValue` keeps null, so the multine can't be updated.
 `useLayoutAnimationFrameEffect(updateMultiline, [contentRef, fontSize, isEditing, simplePath, editingValue])`
 So I changed the `editingValue` as `const editingValue = editingValueStore.useSelector(state => state)` by removing contion.
 
 2. I updated `useLayoutAnimationFrameEffect` hook to let it handles unmount event.


src/components/Editable/useMultiline.ts
```diff
@@ -18,7 +18,7 @@ const useMultiline = (contentRef: React.RefObject<HTMLElement>, simplePath: Simp
   const [multiline, setMultiline] = useState(false)
   const fontSize = useSelector(state => state.fontSize)
   // While editing, watch the current Value and trigger the layout effect
-  const editingValue = editingValueStore.useSelector(state => (isEditing ? state : null))
+  const editingValue = editingValueStore.useSelector(state => state)
```
raineorshine: This would cause every `Editable` component to re-render whenever any thought is edited. That's too broad and will clobber performance.

codebuild-ro: I am afraid it is not.
`useMultiline` hook just returns `true` or `false` which are primitive value - not returns object thing like Non-Primitive which can generate issue because of new reference of object though the content are same.
Though that hook calculates the height as the focusing thought changes, but unless the thought multiline status is changed, that hook will return same value as previous rendering.
So the `Editable` component won't detect state change - meaning that it won't rerender.

codebuild-ro: I used `indentDepth` status to detect the multiline-able thought.
When cursor up/down, the indentDepth changes.
To make it generic, I made a hook to calculate indent detph value.
It was the code of `LayoutTree.tsx`.
```
  const indentDepth = useSelector(state =>
    state.cursor && state.cursor.length > 2
      ? // when the cursor is on a leaf, the indention level should not change
        state.cursor.length - (hasChildren(state, head(state.cursor)) ? 2 : 3)
      : 0,
  )
```

raineorshine: Thanks, but notice that we already re-calculate multiline whenever the cursor changes:

https://github.com/cybersemics/em/blob/686b34ebd2f66d8b332baf6c83de9ce32e97f783/src/components/Editable/useMultiline.ts#L45-L48

Therefore, it seems that the issue is more about the timing of `useSelectorEffect` compared with `useLayoutAnimationEffect`. Could you investigate? 

Note that [makeSelectorEffect](https://github.com/cybersemics/em/blob/686b34ebd2f66d8b332baf6c83de9ce32e97f783/src/hooks/makeSelectorEffect.ts) is designed specifically to avoid re-running the component's render function the way that `useSelector` does. It would be good to know if it has a bug, while preserving the original intention.

src/hooks/useLayoutAnimationFrameEffect.ts
```diff
@@ -10,20 +10,34 @@ import { isSafari, isTouch } from '../browser'
...
-const useLayoutAnimationFrameEffect = (callback: (() => void) | undefined, dependencies: React.DependencyList) => {
+const useLayoutAnimationFrameEffect = (
+  callback: (() => void | (() => void)) | undefined,
+  dependencies: React.DependencyList,
+) => {
   if (!callback) return
 
   useLayoutEffect(
     () => {
+      const destroyCallbackRef: { current: void | (() => void) } = { current: undefined }
```
raineorshine, line 21: I'm not sure what the intention was with `destroyCallbackRef`, but it doesn't seem to do anything?

`callback` always returns `undefined` anyway.

src/hooks/makeSelectorEffect.ts
```diff
@@ -1,25 +1,57 @@
...
-            effect(current)
-          }
-          prev.current = current
-        }),
-      [effect, equalityFn, select],
-    )
+
+    // State is used to trigger the layout effect when value changes
+    // Unlike useSelector, this doesn't cause a re-render of the component
+    const [state, setState] = useState<T>(select(store.getState()))
```
raineorshine: We want to avoid `useState` for the same reason that we are avoiding `useSelector`: We're trying to avoid re-running the component's render function when the state changes, and instead purely run the side effect. It's a granular performance optimization.

Otherwise, we would just use the standard redux `useSelector` and `useLayoutAnimationFrameEffect` and there would be no need for this custom hook.

src/components/Editable/useMultiline.ts
```diff
@@ -36,26 +36,21 @@ const useMultiline = (contentRef: React.RefObject<HTMLElement>, simplePath: Simp
     // 1.5x can cause multiline to alternate in Safari for some reason. There may be a mistake in the height calculation or the inclusion of padding that is causing this. Padding was added to the calculation in commit 113c692. Further investigation is needed.
     // See: https://github.com/cybersemics/em/issues/2778#issuecomment-2605083798
     setMultiline(height - paddingTop - paddingBottom > singleLineHeight * 1.2)
-  }
+  }, [contentRef, fontSize])
 
   // Recalculate multiline on mount, when the font size changes, edit, split view resize, value changes, and when the
   // cursor changes to or from the element.
-  useLayoutAnimationFrameEffect(updateMultiline, [contentRef, fontSize, isEditing, simplePath, editingValue])
+  useLayoutAnimationFrameEffect(updateMultiline, [isEditing, simplePath, editingValue])
```
raineorshine, line 43: I think we need to leave `fontSize` in the effect dependency list, otherwise `updateMultiline` will be reconstituted but not re-invoked when the font sizes changes.

raineorshine (changes requested):
Thanks for the update. This is an improvement. Unfortunately it renders one frame with the incorrect `multiline` value before correcting itself. I'm okay addressing that in a separate PR. At least it now doesn't get stuck with the incorrect `multiline`.

src/hooks/useLayoutAnimationFrameEffect.ts
codebuild-ro: I updated this `useLayoutAnimationFrameEffect.ts` as it can handle unsubsribe action.
It is to remove the state listener and to avoid creating mutli listener.
```
useLayoutAnimationFrameEffect(() => {
  // Set up some DOM listeners
  const dismissListener = setupListeners()
  return dismissListener // This gets stored in destroyCallbackRef
}, [deps])
```

It solves **dangling event listeners** and **memory leaks**.

raineorshine: I'm sorry, I don't see it. Why would there be a memory leak?

The way you have written it, `destroyCallbackRef.current` is always undefined.

src/components/Editable/useMultiline.ts
codebuild-ro: The `useLayoutEffect` inside `useLayoutAnimationFrameEffect` has callback in its `dependency` array.
image

`updateMultiline` is being regenerated by `useCallback`, and it has `fontSize` as one of its dependencies.
image

So, when `fontSize` changes:
  - `useCallback` creates a new `updateMultiline` function
  - This new function becomes a new dependency for the `useLayoutEffect`
  - The effect reruns because its `callback` dependency changed

Therefore, we don't need to add `fontSize` to the `useLayoutAnimationFrameEffect` dependencies because:
1. Changes to `fontSize` already trigger a recalculation through the recreation of the `updateMultiline` callback
2. Adding `fontSize` to both places would actually be redundant and could potentially cause double-updates

src/hooks/useLayoutAnimationFrameEffect.ts
codebuild-ro: It is for cleanup function call for this hook. And it is for generic usage in the future.
Right now, `useLayoutAnimationFrameEffect` is being used as alternative of `useLayoutEffect`.
And `useEffect` and `useLayoutEffect` supports cleanup function to cancel or ignore outdated asynchronous operations, ensuring only the most recent request's data updates the state.

Here is description with usage example:
```
useLayoutAnimationFrameEffect(() => {
  const cleanup = store.subscribe((state) => {
    console.log(state)
  })
  
  // if cleanup is not provided, than, the `console.log(state)` will run multiple times as many as effect reruns.
  return () => cleanup() // This is stored in destroyCallbackRef
}, [dependency array])
```

`destroyCallbackRef.current` is getting cleanup function while `callback` runs.
image

And when the `useLayoutEffect` unmounts, then that block will be run to unsubscribe the listener.
image

src/hooks/makeSelectorEffect.ts
codebuild-ro: `const [state, setState] = useState<T>(select(store.getState()))`
The usage of this, is to avoid running of `effect`, though the state content are same but the reference is different.
As you know, though, the objects content are same, if the references are different, then, those variables are different.

```
currentState = { t: 1 }
previousState = { t: 1 }
console.log(currentState === previousState) // output: false
// but in this case, we don't need to run `effect` again.
```

That is the reason why we are using `equalityFn` to compare current state and previous state deeply.
So if the contents are different between current and previous, it will run `setState(current)` and `state` will be updated and then, effect will be run.

So the `state` is changed if the current state content is different from previous state content.
This can avoid unnecessary running of effect.

So the result is:

- The combination of `equalityFn` and `useState` prevents unnecessary `effect` runs
- Even if `select` returns new object references
- The `effect` only runs when the actual content changes

src/hooks/useLayoutAnimationFrameEffect.ts
raineorshine: Okay, I understand now. I'm generally not a fan of adding complexity for future general use. But in this case since it parallels `useEffect` and `useLayout`, it's not unreasonable. 

Please explain these design choices up front so I don't have to dig for answers. A simple comment like "Add support for cleanup function to parallel `useEffect` and `useLayoutEffect`" would have sufficed.

src/hooks/makeSelectorEffect.ts
raineorshine: The issue is that even when the selector changes, often the effect itself (e.g. the measured width) doesn't change. `useState` will cause everything in the the component function to re-run, ignoring what the effect does. We want to bypass any local state, and just have an entirely un-reactive side effect. Only when a setter in the effect runs should it re-run the render function.

codebuild-ro: I updated this part.
Instead of using `useState` -> trigger re-render -> `useLayoutAnimationFrameEffect`, the hook now:

- Subscribes to store changes in `useLayoutEffect`
- When state changes, calls the effect directly from the subscription callback
- Uses `requestAnimationFrame` (with iOS Safari double-frame) for proper timing
- No component re-renders - achieving the core performance goal

Now `makeSelectorEffect` truly provides a performance optimization by running side effects when state changes without triggering component re-renders, which was the original intent. Users can now avoid the standard `useSelector` + `useLayoutAnimationFrameEffect` pattern when they only need side effects, not re-renders.

src/hooks/useLayoutAnimationFrameEffect.ts
codebuild-ro: Ok, I will leave proper detailed comment in the future

src/components/Editable/useMultiline.ts
codebuild-ro: This is already fixed before.
I reverted the change and use other approach to update `makeSelectorEffect` hook.

src/hooks/makeSelectorEffect.ts
raineorshine: What's the purpose of `effectRef`? It seems like having `effect` in the `useLayoutEffect` dependency list as before would be enough to prevent it from getting stale.

i.e. This seems to work, but maybe I'm missing something.

```diff
diff --git a/src/hooks/makeSelectorEffect.ts b/src/hooks/makeSelectorEffect.ts
index e76d4f3554..9dbea89bef 100644
--- a/src/hooks/makeSelectorEffect.ts
+++ b/src/hooks/makeSelectorEffect.ts
@@ -34,10 +34,6 @@ const makeSelectorEffect = <U extends Store<any>>(store: U) => {
     // This ref persists between renders and is updated in the subscription
     const prev = useRef<T>(select(store.getState()))
 
-    // Keep a ref to the current effect to avoid stale closures
-    const effectRef = useRef(effect)
-    effectRef.current = effect
-
     useLayoutEffect(
       () =>
         // Returns unsubscribe which is called on unmount.
@@ -48,11 +44,11 @@ const makeSelectorEffect = <U extends Store<any>>(store: U) => {
           if (equalityFn ? !equalityFn(current, prev.current) : current !== prev.current) {
             // Call effect directly without setState to avoid re-renders
             // Use requestAnimationFrame to ensure DOM updates are complete
-            requestAnimationFrame(() => effectRef.current(current))
+            requestAnimationFrame(() => effect(current))
           }
           prev.current = current
         }),
-      [equalityFn, select],
+      [effect, equalityFn, select],
     ) // These functions should be memoized to prevent unnecessary re-subscriptions
   }
 }
```

src/components/Editable/useMultiline.ts
raineorshine: I understand now. Thanks for the explanation.

src/hooks/makeSelectorEffect.ts
codebuild-ro: Thank you for your advice.
That is right and ref is not needed in this case.
I removed the usage of `effectRef`, and I just used the `effect` directly.

PR #3119, opened 2025-07-08
react-dnd manager start handler

### LINKED ISSUE #3072: [Mobile] Bullet incorrectly highlighted on gesture

ethan-james:
Fixes #3072 and #3073 

There was a brief gap between the start of a touch and the firing of `react-dnd`'s `start` event, so there was a window where long press had begun (bullet highlighted) but drag-and-drop had not been initialized.

Now `TouchBackendImpl.ts` has been patched to emit a `dragStart` event when drag-and-drop initializes. `useLongPress` listens for that event on mobile devices in order to synchronize long press timing with drag-and-drop.

### DIFF
--- a/src/components/ToolbarButton.tsx
+++ b/src/components/ToolbarButton.tsx
@@ -108,7 +108,7 @@ const ToolbarButton: FC<ToolbarButtonProps> = ({
   /** Handles the onMouseUp/onTouchEnd event. Makes sure that we are actually clicking and not scrolling the toolbar. */
   const tapUp = useCallback(
     (e: React.MouseEvent | React.TouchEvent) => {
-      longPress.props[isTouch ? 'onTouchEnd' : 'onMouseUp']?.(e)
+      longPress.props[isTouch ? 'onTouchEnd' : 'onMouseUp']?.()
       const iconEl = e.target as HTMLElement
       const toolbarEl = iconEl.closest('#toolbar')!
       const scrolled = isTouch && Math.abs(lastScrollLeft.current - toolbarEl.scrollLeft) >= 5


raineorshine:
> Reducing the `react-dnd` timeout causes problems in the other direction, where you can drag-and-drop and draw a gesture at the same time, but the bullet isn't highlighted.

Oops, that's not good either.

> Allowing `longPress` transitions from `Inactive` -> `DragInProgress` (skipping `DragHold`) fixes that and makes this a viable solutiion. I was still able to reproduce the bug after reducing the timeout by 10ms, but not after reducing it by 20ms.

Sounds promising. I think the exact timing will be affected by performance unfortunately. If the app slows down, it might create more false positives.

We could also just delay the appearance of the highlight, without touching the long press timing if that's the only thing affected. Though maybe there is an alert or something else tied to the timeout.

> So because `TouchBackend` [cancels handleTopMoveStart](https://github.com/react-dnd/react-dnd/blob/1de654eb466b9563fa50df070c244d6f3ab43ca7/packages/backend-touch/src/TouchBackendImpl.ts#L416) before [touchSlop is exceeded](https://github.com/react-dnd/react-dnd/blob/1de654eb466b9563fa50df070c244d6f3ab43ca7/packages/backend-touch/src/TouchBackendImpl.ts#L457), it's going to be difficult to ensure that drag-and-drop always starts under the same circumstances as our long press.

Is this a bug in `TouchBackendImpl`?

Would it be possible to lean on `touchSlop` instead of using our own `SCROLL_THRESHOLD`?

Thanks and let me know what you recommend.

ethan-james:
> Is this a bug in `TouchBackendImpl`?

I'd say so. I'm not sure what the rationale is for cancelling `handleTopMoveStart` so early, but it doesn't match with the behavior that I would expect.

> Would it be possible to lean on `touchSlop` instead of using our own `SCROLL_THRESHOLD`?

I don't think so. The only options that I can think of are:

1. cancel our own long press immediately to match the behavior of `TouchBackend` (remove `SCROLL_THRESHOLD`)
2. patch `TouchBackend` to stop it from cancelling `handleTopMoveStart` until `touchSlop` is exceeded

The third option would be to tweak the timing as we've discussed above, however, now that I've realized that this is not quite a race condition, I've also realized that it's still possible to break it with any timing offset. Reducing the timeout so that drag-and-drop initializes sooner will fix the initial repro steps, but it's still possible to wiggle your finger juuuuuust right and break it.

https://github.com/user-attachments/assets/a0f7adce-cadf-41f1-976b-015881307512

ethan-james:
> > 2. patch `TouchBackend` to stop it from cancelling `handleTopMoveStart` until `touchSlop` is exceeded
> 
> I'm open to this. I've already had to patch `TouchBackend` once, and it seems like a genuine bug.

OK, should I follow [these instructions](https://www.ryanelainska.com/how-to-patch-an-npm-package-with-yarn-patch) and put the patch in `.yarn/patches`?

raineorshine:
Yes, thanks! Once you edit the dependency file, it should generate a patch for you. It might even put it directly in `.yarn/patches`, but I don't remember exactly.

See: https://yarnpkg.com/cli/patch

raineorshine (changes requested):
Thank you! This seems promising, but I'm getting some weird behavior with this change. Half the time the long press gets canceled, then gestures stop working. 



## Steps to Reproduce

1. Refresh the page
2. Draw and cancel a gesture 
3. Long press and drag a thought with the timing of the original issue. 
4. Attempt to input a gesture in the gesture zone. 

## Current Behavior

The long press gets canceled for some reason, then gestures stop working.

In this video, after the long press I'm swiping on the gesture zone.

https://github.com/user-attachments/assets/4d4f44fe-25d3-482a-bc26-8170808e7799

ethan-james:
I can reproduce the issue where drag-and-drop and gestures are active at the same time. As you might expect, it occurs when `SCROLL_THRESHOLD` is exceeded *exactly* 400ms after the press begins.

Screenshot 2025-07-15 at 2 36 25 PM

1. The `onStart` function in `useLongPress` is invoked
2. `handleTopMoveStart` fires, initiating drag-and-drop
3. `move` fires in `useLongPress`, cancelling the timeout in `onStart`
4. `handleTopMove` fires in `TouchBackend`, cancelling the `handleTopMoveStart` timeout, which has already fired

There should be another line, `onStart timeout`, but it is cancelled by `move`.

The easiest way to fix this is to allow `state.longPress` to transition from `Inactive` straight to `DragInProgress` without passing through `DragHold`.

raineorshine:
Thanks so much for the in-depth analysis. I'll try to offer my thoughts, although I have to maintain a bit of distance to avoid spending too much time myself. Thanks for your understanding.

> The easiest way to fix this is to allow `state.longPress` to transition from `Inactive` straight to `DragInProgress` without passing through `DragHold`.

That seems fine. Since we use the same `TIMEOUT_LONG_PRESS_THOUGHT` for both dragHold and drag-and-drop, it would make sense to go straight from `Inactive` to `DragInProgress` if the move event occurs immediately after the timeout completes.

> 1. Is it possible to absolutely prevent `handleTopMoveStart` from being canceled? My instinct says no, at least with how the timeout is currently structured.

I'm not sure. I will have to rely on your experience so that I don't get too deep on the implementation details. I do know that we want to avoid modifying `TouchBackendImpl` more than absolutely necessary.

> As for the flickering bullet shown in the video, I still haven't been able to reproduce it, but I can see how it could be the inverse of the bug above (i.e. drag-and-drop and gesture are active at the same time).
> 
> Instead of the timeout in `useLongPress` getting canceled during the same millisecond when it should have fired, if the timeout in `handleTopMoveStart` got canceled during the same millisecond when it should have fired, then `beginDrag` would never get called and drag-and-drop would never activate, but the `dragHold` action would get called by `useDragHold`.
> 
> This would [cause the bullet to highlight](https://github.com/ethan-james/em/blob/179b9786e9d1e4c9ceb83f6e1e62f5d081f41ffd/src/components/Bullet.tsx#L422) because `state.draggedSimplePath` would be set briefly, and then cleared in [move](https://github.com/ethan-james/em/blob/179b9786e9d1e4c9ceb83f6e1e62f5d081f41ffd/src/hooks/useLongPress.ts#L139), but since `onLongPressEnd` is called with `canceled: true`, then it would [not call](https://github.com/ethan-james/em/blob/179b9786e9d1e4c9ceb83f6e1e62f5d081f41ffd/src/hooks/useDragHold.ts#L55) `dispatch(longPress({ value: LongPressState.Inactive }))`. Since drag-and-drop is not running, then it would never call `endDrag`.
> 
> `stop` in `useLongPress` would get called because `react-dnd` wouldn't be intercepting its events, but it would exit early due to [globals.longpressing being false](https://github.com/ethan-james/em/blob/179b9786e9d1e4c9ceb83f6e1e62f5d081f41ffd/src/hooks/useLongPress.ts#L108). So `state.longPress` would get stuck in an active state which would keep causing [shouldCancelGesture](https://github.com/ethan-james/em/blob/179b9786e9d1e4c9ceb83f6e1e62f5d081f41ffd/src/components/AppComponent.tsx#L79) to be true and prevent gestures.

Yes, that seems like a plausible cause. Something must be leading to `shouldCancelGesture` returning true.

It seems like the long press should *always* be set to inactive on `touchend` or `touchcancel`. The stuck state is higher impact than the flickering bullet problem, so we should take extra steps to ensure it can never occur.

> 1. Assuming the macrotask queue behaves like a queue, it should be possible to restructure the code to ensure that both move functions run either before or after both timeouts, deterministically. This would prevent situations where one timeout gets canceled but not the other.

It's a bit hard for me to wrap my had around, but I'm open to it if it's not a ton of code.

> 2. If that's too fiddly, it should be possible to store the timestamp when the timeouts begin, then check that value in the move functions. If exactly `TIMEOUT_LONG_PRESS_THOUGHT` milliseconds have passed, then the move functions can choose not to cancel the timeouts, ensuring that they both run.

Seems low impact enough.

> 3. Redo the `TouchBackend` patch so that we eliminate the `TouchBackend` timeout and instead have a way to invoke `beginDrag` on demand. This is basically a retread of my earlier fork idea.

This feels too high risk to me.

--- 

One thing you could try is setting [touchSlop](https://react-dnd.github.io/react-dnd/docs/backends/touch#options) to `SCROLL_THRESHOLD` so that it corresponds to the movement we allow before the long press is cancelled. That could help avoid drag-and-drop and gestures being active at the same time. Though maybe they could only be truly coordinated if we hooked into `TouchBackendImpl`'s touchSlop logic, which could be messy.

ethan-james:
> One thing you could try is setting [touchSlop](https://react-dnd.github.io/react-dnd/docs/backends/touch#options) to `SCROLL_THRESHOLD` so that it corresponds to the movement we allow before the long press is cancelled. That could help avoid drag-and-drop and gestures being active at the same time. Though maybe they could only be truly coordinated if we hooked into `TouchBackendImpl`'s touchSlop logic, which could be messy.

That's what I did [here](https://github.com/ethan-james/em/blob/04a56f03c2e6db66585f2440194d5769993999fb/src/components/DragAndDropContext.tsx#L12) without actually using the `SCROLL_THRESHOLD` variable because it's [not exported](https://github.com/ethan-james/em/blob/04a56f03c2e6db66585f2440194d5769993999fb/src/hooks/useLongPress.ts#L13). Probably worthwhile to clean that up before merging by adding `SCROLL_THRESHOLD` to `constants.ts`.

In any case, that fixed most of the buggy behavior. What's left is due to situations where one timeout (e.g. `useLongPress`) is canceled at the exact millisecond when it should be firing, while the other timeout (e.g. `TouchBackend`) is not canceled and takes effect. So all of the handlers activate at the same time, like:

1. `move` in `useLongPress` cancels `onStart` because `SCROLL_THRESHOLD = 10` has been exceeded
2. `onStart` would run in the same millisecond, but has been canceled
3. `handleTopMoveStart` runs in `TouchBackend`
4. `handleTopMove` runs, tries to cancel `handleTopMoveStart` because `touchSlop = 10` has been exceeded, but it's already active

So I'm going to try to do option number 2 above:

> 2. If that's too fiddly, it should be possible to store the timestamp when the timeouts begin, then check that value in the move functions. If exactly TIMEOUT_LONG_PRESS_THOUGHT milliseconds have passed, then the move functions can choose not to cancel the timeouts, ensuring that they both run.

and see if it's possible to guard against out-of-order event handlers to ensure that either

1. Both `start` functions run, or
2. Both `move` functions run and cancel their `start` functions

src/actions/longPress.ts
```diff
@@ -13,10 +13,13 @@ const longPress = (state: State, { value = LongPressState.Inactive }: Payload) =
   if (
     // DragHold must be triggered by the beginning of a long press, don't go back from another state
     (value === LongPressState.DragHold && state.longPress === LongPressState.Inactive) ||
-    // Drag can only begin after long press has begun, and should not return from DragCancelled
-    (value === LongPressState.DragInProgress && state.longPress === LongPressState.DragHold) ||
+    // Drag should only begin after long press has begun, but the occasional race condition
+    // may allow react-dnd to initiate a drag before useLongPress sets this to DragHold.
+    // Drag should not re-start after DragCanceled.
```
ethan-james, line 18: This is hopefully not true anymore, but the async nature of `useLongPress` and `beginDrag` makes me nervous that removing it might re-introduce some rare bug.

PR #3122, opened 2025-07-10
Prevent clickOnEmptyThought during long press

### LINKED ISSUE #3120: [iOS] Can't multiselect more than one thought
Reproducible on iOS, tested and works on Android and desktop Chrome.

## Steps to Reproduce

```
- A
- B
```

1. Long press `A` and release
2. Long press `B` and release

## Current Behavior

Thought `B` is the only thought highlighted. Command menu says "1 thought selected".

https://github.com/user-attachments/assets/43bf0dd4-b3db-4178-a1bf-aa61abbffc1e

## Expected Behavior

Both `A` and `B` should be highlighted. Command menu should say "2 thoughts selected".

ethan-james:
Fixes #3120

`Content` was responding to `mousedown` events that bubble up during long presses. It was dispatching `toggleDropdown` to close all dropdown menus, which in turn would dispatch `clearMulticursors` during a long press.


ethan-james:
This is working and the tests should be working, but I think maybe one of the earlier tests is interfering with my new one. Can you please confirm whether left or right is what we should be seeing here?

multiselect-ios-diff

src/components/Content.tsx
```diff
@@ -44,6 +44,7 @@ const Content: FC = () => {
     return children.length
   })
   const isAbsoluteContext = useSelector(state => isAbsolute(state.rootContext))
+  const longPress = useSelector(state => state.longPress)
```
raineorshine: This will cause `Content` and thus the entire `LayoutTree` to unnecessarily re-render when `state.longPress` changes.

`useSelector` when the value needs to be reactive. If you need to access the Redux state in response to a user event, I prefer dispatching a thunk which gives you `getState()`.

raineorshine (commented):
Maybe the deeper problem is the local `isPressed`? It seems like should be a better way to check if the Content element was actually clicked as opposed to a descendant. Perhaps `e.relatedTarget` could be used instead.

ethan-james:
> TL;DR The left screenshot, but it's not super important.

Thanks! I choose to interpret "not super important" to mean that I'm better off forcing the observed behavior and re-generating the snapshot rather than trying to dig into why it behaves differently in isolation compared to as part of the whole suite. I'm happy to dig in, though, if you do want to know.

raineorshine:
That sounds fine, thanks. 

It looks like the Puppeteer test is failing in the CI unfortunately. The flaky tests in `main` have all been disabled, so I'd like to avoid reintroducing any flakiness.

Here's a script that can help you determine if the test is flaky: https://github.com/cybersemics/em/wiki/Testing#triggering-github-actions-workflows-manually

ethan-james:
> That sounds fine, thanks.
> 
> It looks like the Puppeteer test is failing in the CI unfortunately. The flaky tests in `main` have all been disabled, so I'd like to avoid reintroducing any flakiness.
> 
> Here's a script that can help you determine if the test is flaky: https://github.com/cybersemics/em/wiki/Testing#triggering-github-actions-workflows-manually

Thanks! It's not actually flaky; it consistently produces one result during a full test run and another during an incremental re-run. Presumably some portion of the Redux state is not getting completely reset between test suites. I thought I could force it to look the same in both settings by inserting unnecessary clicks, but now I think the best thing to do is to update the snapshot during a full run rather than pressing `u` to update it incrementally.

src/e2e/puppeteer/helpers/longPressThought.ts
```diff
@@ -0,0 +1,14 @@
+import { ElementHandle } from 'puppeteer'
+import { TIMEOUT_LONG_PRESS_THOUGHT } from '../../../constants'
+import sleep from '../../../util/sleep'
+import waitForEditable from './waitForEditable'
+
+/** Tap and hold a thought until a long press occurs. */
+const longPressThought = async (value: string) => {
+  const thought = (await waitForEditable(value)) as ElementHandle
+  await thought.touchStart()
+  await sleep(TIMEOUT_LONG_PRESS_THOUGHT + 100)
```
raineorshine: Rather than wait an additional `100ms`, please wait for a specific condition (such a Redux state change or a DOM state change). Arbitrary sleeps add up over time and slow down the test suite for everyone on the team.

src/e2e/puppeteer/__tests__/multiselect.ts
```diff
@@ -0,0 +1,43 @@
...
+describe('multiselect', () => {
+  it('should multiselect two thoughts at once', async () => {
+    await paste(`
+        - a
+        - b
+        `)
+
+    await multiselectThoughts(['a', 'b'])
+
+    expect(await screenshot()).toMatchImageSnapshot({ customSnapshotIdentifier: 'multiselect' })
```
raineorshine: Let's use [hideHUD](https://github.com/cybersemics/em/blob/7a51ebcb9c7263a0f306016f56d08fedf56c6c0a/src/e2e/puppeteer/helpers/hideHUD.ts) to avoid accidentally tightly coupling the test to the toolbar rendering.

src/e2e/puppeteer/helpers/longPressThought.ts
```diff
@@ -0,0 +1,14 @@
+import { ElementHandle } from 'puppeteer'
+import { TIMEOUT_LONG_PRESS_THOUGHT } from '../../../constants'
+import sleep from '../../../util/sleep'
+import waitForEditable from './waitForEditable'
+
+/** Tap and hold a thought until a long press occurs. */
+const longPressThought = async (value: string) => {
+  const thought = (await waitForEditable(value)) as ElementHandle
+  await thought.touchStart()
```
raineorshine: The other condition that is important for reproducing the issue is that it only occurs when long tapping the thought *outside* the editable, i.e. slightly to the right of the text. If you long tap the editable text itself, the issue does not occur. I don't think this test reflects that, since `touchStart()` is called on the Editable itself. Is that true?

raineorshine (changes requested):
Thanks, the solution looks good! I have some questions about the test.

> > TL;DR The left screenshot, but it's not super important.
> 
> Thanks! I choose to interpret "not super important" to mean that I'm better off forcing the observed behavior and re-generating the snapshot rather than trying to dig into why it behaves differently in isolation compared to as part of the whole suite. I'm happy to dig in, though, if you do want to know.

That sounds reasonable, but unfortunately "should multiselect two thoughts at once" is failing for me locally even when I run all the puppeteer tests.

Fixing #3125 might be necessary now, otherwise it's going to be very slow testing the entire puppeteer test suite every time.

image

raineorshine:
I've thought about this some more, and I don't think a snapshot test is the best testing approach here actually. Snapshot tests are best for catching visual regressions. While the highlighted bullets are visual, the real problem here is not how they are rendered, but how many thoughts are selected after the long tap. I think a better approach would be to query the DOM for highlighted bullets, or just test "2 thoughts selected".

src/e2e/puppeteer/helpers/longPressThought.ts
ethan-james: Yes, I think you are correct. I think that finding the parent thought wrapper would be a good way to fix the test. It might fix the specific issue of tapping outside the area of the `Editable` if it is targeting the center of the wrapper element, but I think that the event handlers attached to the element are more important than the actual physical position of the touch.

raineorshine: Generally I think it's better to write tests from the user's perspective and avoid coupling the test to implementation details or the DOM hierarchy. So actually I would lean towards simulating a tap slightly to the right of the Editable. It will be more resilient to changes to the DOM structure in the future.

src/e2e/puppeteer/helpers/longPressThought.ts
```diff
@@ -0,0 +1,42 @@
+import { ElementHandle } from 'puppeteer'
+import { JSHandle } from 'puppeteer'
+import { page } from '../setup'
+
+interface Options {
+  /** Click on the inside edge of the editable. Default: left. */
+  edge?: 'left' | 'right'
+  /** Number of pixels of x offset to add to the click coordinates of width/2. */
```
raineorshine: I believe the `x` offset is from the left or right edge now, not the middle.

src/e2e/puppeteer/__tests__/__image_snapshots__/multiselect/multiselect-ios.png
raineorshine, line 1: Can the two snapshots be removed since we switched to non-snapshot tests?

src/e2e/puppeteer/__tests__/multiselect.ts
```diff
@@ -0,0 +1,56 @@
...
+    const b = await waitForEditable('b')
+
+    await longPressThought(a, { edge: 'right', x: 100 })
+    await longPressThought(b, { edge: 'right', x: 100 })
+
+    const highlightedBullets = await page.$$('.bullet[data-highlighted=true]')
+    const commandMenuPanelTextContent = await page.$eval('[data-testid=command-menu-panel]', el => el.textContent)
+
+    expect(highlightedBullets.length).toBe(2)
+    expect(commandMenuPanelTextContent).toContain('2 thoughts selected')
```
raineorshine, line 54: Looks good, thanks!

raineorshine (changes requested):

PR #3123, opened 2025-07-12
Mobile gestures toggle in Help modal + moving gestures to the right in CommandItem

yangchristina:
This PR may be reverted later on to undo moving gestures to the right on Command Palette.

https://github.com/user-attachments/assets/9b2547ce-f799-47ab-ab7b-41d937df9d08

Changes to GestureDiagram cancel icon styling are to make that icon match other gesture icons in alignment.


src/components/CommandItem.tsx
```diff
@@ -183,43 +185,37 @@ const CommandItem: FC<{
...
           viewType === 'grid'
             ? { minWidth: { base: '10rem', _mobile: 'auto' }, width: '100%' }
-            : { display: 'flex', justifyContent: 'center', alignItems: 'center', width: 32, height: 32 },
+            : {
+                display: 'flex',
+                justifyContent: 'center',
+                alignItems: 'center',
+                width: 'max-content',
+                height: 32,
+                _desktop: { minWidth: 32, width: 32 },
```
yangchristina: this might be a mistake, checking now

yangchristina: reverted change

yangchristina:
@raineorshine this is ready for review.

src/components/CommandItem.tsx
```diff
@@ -52,6 +52,7 @@ const CommandItem: FC<{
   isLastCommand?: boolean
   /** If the CommandItem is rendered within a <table>, then it needs to use <tr> and <td> and avoid inline padding. Confusingly, this is independent from viewType which determines whether the component is layed out in a 2-column 'grid' or a 'table' of rows. Perhaps this could be refactored to avoid using <table> and <tbody> at all while preserving the desired appearance in the Help modal. */
   tableMode?: boolean
+  isMobileGestures?: boolean
```
raineorshine, line 56: As a rule, I almost always like to add a descriptive JSDOC comment on a props type. It's a great place to communicate the API for a component or function, including its recommended usage and any considerations for the caller. It also provides an opportunity to double check that the prop makes sense independently from how it is implemented within the component. Each interface is a contract that encapsulates and hides the internal logic. Writing a good JSDOC helps you think about internal API's and encapsulation boundaries rather than using props or function parameters in a more arbitrary/instrumental way.

Here's my attempt:

`/** If true, renders the gesture diagram instead of the keyboard shortcut. Defaults to isTouch. Ignored if viewType is 'grid'. */`

raineorshine (approved):
Thanks!

src/components/CommandItem.tsx
yangchristina: Sounds good, thanks for letting me know and also explaining the reasoning! Will do this in the future.

Issue #3088, opened 2025-06-29
Flaky test: render-thoughts - multiline


raineorshine:
Test failures:

- https://github.com/cybersemics/em/actions/runs/15949546278/job/44987659232

`Error: Expected image to match or be a close match to snapshot but was 0.07583333333333334% different from snapshot (364 differing pixels).
See diff for details: /home/runner/work/em/em/src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/__diff_output__/multiline-multiline-thought-1-diff.png
 ❯ src/e2e/puppeteer/__tests__/render-thoughts.ts:151:19`

https://github.com/cybersemics/em/blob/c833c0fc3bcdf78799fc9f880fa8700d9fe3af66/src/e2e/puppeteer/__tests__/render-thoughts.ts#L142-L148

raineorshine:
Fixed in #3164

Issue #2956, opened 2025-05-17
Flaky test - url multiline


raineorshine:
Test failures:

- https://github.com/cybersemics/em/actions/runs/15088412461/attempts/2
- https://github.com/cybersemics/em/actions/runs/14812798110/job/41589583778
- https://github.com/cybersemics/em/actions/runs/13817648331/job/38654935147?pr=2800

`[src/e2e/puppeteer/__tests__/url.ts > multiline > Font Size: 13: src/e2e/puppeteer/__tests__/url.ts#L72](https://github.com/cybersemics/em/pull/2952/files#annotation_34804919320)
Error: Expected image to match or be a close match to snapshot but was 0.08416666666666667% different from snapshot (404 differing pixels).
See diff for details: /home/runner/work/em/em/src/e2e/puppeteer/__tests__/__image_snapshots__/url/__diff_output__/multiline-font-size-13-1-diff.png
 ❯ multilineTest src/e2e/puppeteer/__tests__/url.ts:72:19
 ❯ src/e2e/puppeteer/__tests__/url.ts:96:5`

https://github.com/cybersemics/em/blob/93617b7e0af854ac1e62a496a086d31dde67c9bd/src/e2e/puppeteer/__tests__/url.ts#L54-L79

raineorshine:
Single line url test appears to be flaky, too.

Temporarily skipped in dde883c1a7.

karunkop:
> Single line url test appears to be flaky, too.
> 
> Temporarily skipped in [dde883c](https://github.com/cybersemics/em/commit/dde883c1a7b837134e0c2ac55295f253b0595770).

This has been already addressed in this [PR](https://github.com/cybersemics/em/pull/3126/files#diff-a5b3a5ecfe3187d053ad665ca3ffed088c1d558c2b622af0f7324dfeafe87494R36-R39)

PR #3126, opened 2025-07-16
Flaky Tests: Pixel Difference Issues with image snapshots

### LINKED ISSUE #3124: Flaky Test: caret should move to editable after closing the command palette
Test Failures:

- https://github.com/cybersemics/em/actions/runs/16249290070
- https://github.com/cybersemics/em/actions/runs/16262142585

Image

https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/__tests__/caret.ts#L207-L220

karunkop:
Close #3124 , #3088, #2956

### Cause:

#3124 : The fundamental problem was that the `ArrowDown` keypress was happening before the selection/focus restoration completed after closing the command palette (including FadeTransition animation).

Multiline layouts need more time to fully stabilize, especially in CI environments. (#3088). Similar is the case with #2956. A optional `frames` parameter is required for complex text wrapping calculations, layout reflow for the long text and font rendering stabilization.

### Fix:

Just wait for animation frames such that it provides enough time for stabilization before snapshot.

### Success Rate:

[Puppeteer checks](https://github.com/karunkop/em/actions?query=branch%3Aflaky-test)

### Additional Notes:

This PR also fixes some other intermittent issues which were encountered during the multiple execution of the puppeteer checks. Like:

image


src/e2e/puppeteer/__tests__/render-thoughts.ts
```diff
@@ -139,8 +139,6 @@ describe('Font Size: 22', () => {
 describe('multiline', () => {
   beforeEach(hideHUD)
 
-  // TODO: Flaky test
-  // https://github.com/cybersemics/em/issues/3088
   it.skip('multiline thought', async () => {
```
raineorshine: Still skipping?

karunkop:
> Thanks for the PR!
> 
> In the future, I would recommend sticking to one PR per issue, unless it is truly a single solution that closes all the issues. Some of the solutions included here depend on the update to `waitForFrames`, admittedly, but still you're trying to do too much in a single PR. If we find a regression in the future, we're going to have a revert the entire PR and then add functionality back in incrementally. Separate PR's would result in a more granular commit history that will make `git blame` easier.
> 

Noted! Will do as suggested.

src/e2e/puppeteer/__tests__/render-thoughts.ts
karunkop: I have updated this! Thanks

karunkop:
> > ### Success Rate:
> > [Puppeteer checks](https://github.com/karunkop/em/actions)
> 
> I'm seeing that 3 of the last 10 tests have failed, so I'm going to assume they are still flaky. Can you take a look?

The github checks result that you viewed yesterday were of different branch/issue I was working on. You can view them now and see the result. I have successfully verified 30 checks to be sure of the changes. All of them have passed!

karunkop:
> > ### Cause:
> > #3124 : The fundamental problem was that the `ArrowDown` keypress was happening before the selection/focus restoration completed after closing the command palette (including FadeTransition animation).
> 
> Thanks for the explanation. In this case my question is, if `ArrowDown` should wait for the selection to be restored to the correct element, why not wait for that specific condition? `requestAnimationFrame` seems like a rather indirect approach, similar to `sleep`.
> 
> The caret tests are behavioral tests, not snapshot tests. They do not depend on rendered frames, so it's strange to have them tightly coupled to animation frames rather than DOM states.
> 

Your concern makes perfect sense and I have now updated it such that now the behavioral tests for `carets` will wait for specific condition rather than `waitForFrames` as suggested.

src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -214,6 +214,11 @@ describe('all platforms', () => {
     await press('Escape')
     await press('ArrowDown')
 
+    // Wait for the caret to move to the editable
+    // because the closing of command palette is asynchronous and the caret may not be in the editable yet
+    // otherwise the test intermittently fails in CI.
+    await waitUntil(() => window.getSelection()?.focusNode?.textContent === 'a')
+
     const textContext = await getSelection().focusNode?.textContent
     expect(textContext).toBe('a')
```
raineorshine: This `expect` is no longer necessary.

src/e2e/puppeteer/__tests__/categorize.ts
```diff
@@ -29,6 +29,11 @@ describe('categorize', () => {
         - Fusce vulputate eleifend sapien. Vestibulum purus quam, scelerisque ut, mollis sed, nonummy id, metus. Nullam accumsan lorem in dui. Cras ultricies mi eu turpis hendrerit fringilla. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia Curae; In ac dui quis mi consectetuer lacinia.
     `)
 
+    // Wait for complex text layout with very long paragraphs to complete (4 frames for extensive content)
```
raineorshine: This is better, thanks. You can bit more specific: "due to multiple rounds of requestAnimationFrame in useLayoutAnimationFrameEffect".

src/e2e/puppeteer/__tests__/categorize.ts
```diff
@@ -29,6 +29,11 @@ describe('categorize', () => {
...
+    // because large text blocks require multiple render cycles for proper wrapping and positioning
+    // otherwise categorize operations may target thoughts that haven't finished rendering
```
raineorshine: 1. I'm not sure what you mean by operations "targeting thoughts". Could you explain?

2. Why is this awaited before the click/categorizations rather than before the snapshot? Does the click/categorize somehow depend on the render state? Hoping to understanding the placement better. 

   This same questions applies to other instances below where `waitForFrames` is called before a user action rather than the snapshot. Thanks for clarifying.

src/e2e/puppeteer/__tests__/render-thoughts.ts
```diff
@@ -81,12 +81,9 @@ const testSuite = () => {
...
 
-      // TODO: Test intermittently fails with small differences in 'b'.
-      // Tested manually with navigator.webdriver = true and 'b' renders at the correct opacity in the next frame, without any animation, so I do not know why this fails.
-      // Example failed test runs:
-      // - https://github.com/cybersemics/em/actions/runs/14236307211
-      // - https://github.com/cybersemics/em/actions/runs/14783509675/job/41507408875?pr=2917
-      // Waiting for requestAnimationFrame does not fix the issue.
+      // Wait for superscript calculation and opacity changes to complete
+      // because superscript rendering uses requestAnimationFrame for positioning calculations
+      // otherwise screenshot may capture intermediate rendering states causing test flakiness
```
raineorshine, line 86: 👍

src/e2e/puppeteer/helpers/waitForFrames.ts
```diff
@@ -12,12 +12,26 @@ import { page } from '../setup'
  * - Components use rAF for state updates (e.g., Superscript calculations).
  * - CI environments have slower frame rates (20-30 FPS vs local 60 FPS).
  * - Fixed timeouts (e.g., `sleep(200)`) fail to capture final render.
+ *
+ * @param frames Number of animation frames to wait (default: 2). Use higher values for complex layouts like multiline text.
```
raineorshine: It's not complex layouts that require additional frames, but specifically the use of multiple rounds of `requestAnimationFrame` in `useLayoutAnimationFrameEffect`.

raineorshine (changes requested):
> > > ### Success Rate:
> > > [Puppeteer checks](https://github.com/karunkop/em/actions)
> > 
> > 
> > I'm seeing that 3 of the last 10 tests have failed, so I'm going to assume they are still flaky. Can you take a look?
> 
> The github checks result that you viewed yesterday were of different branch/issue I was working on. You can view them now and see the result. I have successfully verified 30 checks to be sure of the changes. All of them have passed!

Hmmm as far as I know it was the same branch: `flaky-test-categorize-caret`. Unless you were reusing that branch for other work, that's where I saw the failures.

I still see at least one failure in the last 10 runs. I'm not sure where you did the 30 checks, but I'd love to see the results after you resolve the current feedback. Thank you!

src/e2e/puppeteer/__tests__/categorize.ts
karunkop: 1. I have updated the comment so it should be clear now.

2. The wait is actually needed for the click operation to work correctly, not the categorize operations. The clickThought helper needs to find the DOM element containing that very long text, and if the text is still reflowing/repositioning due to useLayoutAnimationFrameEffect, the click might target the wrong element or fail entirely. The categorize operations themselves don't depend on render state , they're just sending key events. But they need the correct thought to be selected first via the click, which is why the wait comes before the click rather than before the final screenshot. This is what I believe is happening and to ensure this I have also run the github checks by placing the `waitForFrames` before snapshot to see if the test still passes, however flakiness is still observed in this scenario. [Checks when waitForFrames is placed just before taking snapshot](https://github.com/karunkop/em/actions?query=branch%3Aflaky-test-clone)

src/e2e/puppeteer/__tests__/caret.ts
karunkop: Sorry i didn't get this! Can you explain it further?

raineorshine: It just waited for `textContent` to equal `a` on the previous line. So what exactly is being asserted here? It can never fail. 

It is effectively asserting the same thing twice, and doing it asynchronously so it's another round trip to the browser.

src/e2e/puppeteer/__tests__/categorize.ts
raineorshine: Why would layout affect `clickThought`'s ability to find the DOM element? Doesn't make sense.

Your empirical results are solid, but the explanation falls short. Maybe it's not reflow at all, but the DOM node gets unmounted after `waitForEditable` but before `click()`. If so, that would indicate a critical bug in `clickThought`.

karunkop: image

 The snapshot difference shows a massive layout shift when `waitForFrames` is placed just before taking snapshot. Let me investigate a bit more for the likely explanation. Any form of help or speculation/hint would be greatly appreciated!

src/e2e/puppeteer/__tests__/caret.ts
karunkop: Thanks got it! Will remove the assertion!

src/e2e/puppeteer/__tests__/categorize.ts
raineorshine: Looks like https://github.com/cybersemics/em/blob/06c1657076f5aa455a769760cdcb7308ddc7ac09/src/redux-middleware/scrollCursorIntoView.ts

It shifts down, but the text does not reflow or change height.

raineorshine: @karunkop Hi! Do you have an update on this?

karunkop: Hi @raineorshine,

Sorry for the delay in providing an update on this. But after some thorough experimentation and analysis of React’s asynchronous batched rendering process and the usage of `scrollCursorIntoView` redux-middleware, I’ve landed on the following conclusion:

While the `scrollCursorIntoView` middleware does introduce asynchronous scroll behavior after cursor changes which can occasionally cause screenshots to capture mid-scroll states, the more critical issue lies earlier in the test flow. The need for `waitForFrames(4)` before `clickThought` stems from React’s batched rendering of long content. Without this wait, the initial click can occur during an intermediate render cycle, especially in CI environments running at lower frame rates. This leads to interactions with unstable or partially mounted DOM elements, corrupting all subsequent categorization steps regardless of scroll synchronization. I believe that the scroll middleware doesn’t even come into play properly if the DOM isn’t fully ready at the time of the first click.

Waiting four frames before `clickThought` ensures that the DOM has stabilized, React has completed all batched updates, and components have fully mounted. I believe similar explanation is also applicable for other instances where `waitForFrames` is called before a user action rather than the snapshot.

raineorshine: > While the `scrollCursorIntoView` middleware does introduce asynchronous scroll behavior after cursor changes which can occasionally cause screenshots to capture mid-scroll states, the more critical issue lies earlier in the test flow.

Can you provide an example snapshot diff of this happening on a flaky test? I just have https://github.com/karunkop/em/actions/runs/16547038660 which is clearly the scroll timing.

> The need for `waitForFrames(4)` before `clickThought` stems from React’s batched rendering of long content.

Why would React render long content different than short content? 
Is it long thoughts that cause the flakiness, or just multiple thoughts?
Does it always fail on the first categorization or does it fail on other categorizations?

> Without this wait, the initial click can occur during an intermediate render cycle, especially in CI environments running at lower frame rates.

Why wouldn't the clickable element be available for a click after awaiting paste?
What happens if the click occurs during an intermediate render cycle?
Is paste asynchronous?

https://github.com/user-attachments/assets/be13cccd-ab77-4758-8630-fac9d35db04f

> This leads to interactions with unstable or partially mounted DOM elements, corrupting all subsequent categorization steps regardless of scroll synchronization.

How can something be partially mounted?
What do you mean by "corrupting"?

Please don't use vague terms when we're doing detail work like this. "Corrupting" a categorization step is meaningless. What exactly is happening?

> I believe that the scroll middleware doesn’t even come into play properly if the DOM isn’t fully ready at the time of the first click.

Do you have a snapshot diff that shows this?

> Waiting four frames before `clickThought` ensures that the DOM has stabilized

What do you mean by "stabilized" and why would that interfere with click?

---

We need a more definitive explanation of what is going on here, otherwise we're going to make the same mistake over and over again, and not know how to properly design our tests to avoid flakiness.

karunkop: I will try and provide a more detail explanation as soon as possible. The use of some vague terms have clearly introduced unwanted confusions and questions. I will keep you in loop about the updates. Thanks!

Issue #3105, opened 2025-07-01
Gap between last child drop target and root drop target


raineorshine:
## Steps to Reproduce

```
- a
- b
- c
```

Drag `a` after `c` and slowly move down.

Set `em.testFlags.simulateDrop = true` to turn on drop target debugging boxes.

## Current Behavior

There is a gap between the child drop targets `b` and `c`:

![Image](https://github.com/user-attachments/assets/98cfc658-c079-4aa7-ac18-661352308e3c)

Now I will describe what happens in the video below. When dragging, first the drop target is shown at the end of the root (correct).

Then if the mouse moves down slightly, there is a 1–2px gap where the drop target for the child of `c` is shown.

https://github.com/user-attachments/assets/e2acbc8a-98c2-4234-a354-d2f5327e3b00

## Expected Behavior

There should be no gap.

Moving the dragged item down 1px at a time should highlight first the drop target child of `c` and then the drop target for the root, without any gap in between.

codebuild-ro:
I noticed another gap which won't be fixed with fixing above issue.

```
- x
- a
  - =view
    - Table
  - =pin
    - true
  - b
    - c
  - d
    - e
```

<img width="1080" height="625" alt="Image" src="https://github.com/user-attachments/assets/6b9a0b44-190b-4db7-b125-5ef5798f0bc5" />

PR #3127, opened 2025-07-17
Fix 3105: `Gap between last child drop target and root drop target`

codebuild-ro:
Issue: https://github.com/cybersemics/em/issues/3105

When a thought has a negative cliff value (`cliff < 0`), its parent thought gets extra padding at the bottom (equal to `fontSize / 4`) to create space for nested lists. This padding is added in the `TreeNode` -> `VirtualThought` -> `Subthought` component, increasing the height of the thought’s container.

However, there’s an issue: the `DropChild` component is positioned relative to this padded container but doesn’t adjust for the extra padding. Because of this, there’s a visible gap between the drop target of the last thought (nth) and the one before it (n-1th), making drag-and-drop feel less smooth.

To fix this, when cliff padding is applied (`cliff < 0)`, we move the `DropChild` component up by the same amount (`fontSize / 4`) using a negative top offset. This keeps the drop target aligned with the previous thought, maintaining visual consistency.

As the result:

- The existing cliff padding behavior stays the same for visual spacing.
- The change has minimal impact on other components.
- It only affects `DropChild` when cliff padding is actually used.
- Drag-and-drop targets stay correctly positioned.

image


raineorshine (approved):
Excellent explanation, thoughtful comments, and high quality solution. Love to see it. Well done :).

Issue #3076, opened 2025-06-27
[Mobile]


raineorshine:
Related: #3075

## Steps to Reproduce

1. Create thought `a`.
2. Create an empty subthought under `a`.
3. Continue creating empty subthoughts under `a` by hitting Enter.

## Current Behavior

Initially, the new thoughts are gently scrolled into view so that it is not covered by the keyboard.

As more thoughts are created, the viewport begins to scroll sharply up and down. Eventually it scrolls so dramatically that the virtual keyboard closes.

## Expected Behavior

The cursor thought should be scrolled into view as usual, and there should be no jarring scroll up and down. The keyboard should stay open after New Thought. It should work the same whether activating New Thought on an empty or non-empty thought.

[useHideSpaceAbove](https://github.com/cybersemics/em/blob/60bab1a58979b1eb81105c259f165f87f7c4d0f7/src/components/LayoutTree.tsx#L87) is supposed to adjust the scroll position instantaneously so that there is no visible viewport shift.

## Notes

If you add a character to one of the thoughts so it is non-empty, the issue does not occur! That seems like a great clue. I would start by trying to figure out what is different about the empty thought that causes this behavior.

Issue #3075, opened 2025-06-27
[Mobile] Keyboard does not open after New Thought in long list


raineorshine:
Tested on mobile Safari. 
Unable to reproduce in desktop Safari or Chrome Device Mode.

Related: #3043 

## Steps to Reproduce

Only occurs in long lists:

```
- x
  - Alabama
  - Alaska
  - Arizona
  - Arkansas
  - California
  - Colorado
  - Connecticut
  - Delaware
  - Florida
  - Georgia
  - Hawaii
  - Idaho
  - Illinois
  - Indiana
  - Iowa
  - Kansas
  - Kentucky
  - Louisiana
  - Maine
  - Maryland
  - Massachusetts
  - Michigan
  - Minnesota
  - Mississippi
  - Missouri
  - Montana
  - Nebraska
  - Nevada
  - New Hampshire
  - New Jersey
  - New Mexico
  - New York
  - North Carolina
  - North Dakota
  - Ohio
  - Oklahoma
  - Oregon
  - Pennsylvania
  - Rhode Island
  - South Carolina
  - South Dakota
  - Tennessee
  - Texas
  - Utah
  - Vermont
  - Virginia
  - Washington
  - West Virginia
  - Wisconsin
  - Wyoming
```

1. Set the cursor on `Alabama`.
2. Activate New Thought (sometimes it takes multiple times)

## Current Behavior

The virtual keyboard closes.

## Expected Behavior

The virtual keyboard should stay open.

The virtual keyboard should always be open after New Thought.

## Notes

Normally only occurs on long lists, but can be simulated on short lists by adding a slight delay to `selection.set`:

```diff
diff --git a/src/components/Editable/useEditMode.ts b/src/components/Editable/useEditMode.ts
index 41f49d69b0..f5558655be 100644
--- a/src/components/Editable/useEditMode.ts
+++ b/src/components/Editable/useEditMode.ts
@@ -88,7 +88,7 @@ const useEditMode = ({
           if (!selection.isThought()) {
             asyncFocus()
           }
-          setTimeout(setSelectionToCursorOffset)
+          setTimeout(setSelectionToCursorOffset, 10)
         } else {
           setSelectionToCursorOffset()
         }
```

This suggests that the long list _may_ only be causal due to the slower render performance which exposes the bug. However, I wouldn't rule out a different mechanism, such as autoscroll.

raineorshine:
[...]

Issue #999, opened 2021-01-12
Mobile: Enter breaks auto-capitalization


raineorshine:
Swiping does not break auto-capitalization.

Tested in Mobile Safari.

PR #3129, opened 2025-07-17
newThought: skip asyncFocus if selection.isThought

### LINKED ISSUE #2748: [Mobile] Scroll jumps on new thought after empty thought
Tested on iOS PWA. 

This only occurs when activating New Thought on an empty thought. Step 3, below, is just there to demonstrate that the problem does not occur when activating New Thought on a non-empty thought.  

Also note that this does not occur when scrollY is 0. 

## Steps to Reproduce

``` 
- a
  - b
``` 

1. Scroll down slightly so scrollY is not at 0. 
2. Set the cursor on `b`. 
3. New Thought *(notice how it does not autoscroll)*
4. Delete the new thought and edit `b` to the empty thought. 
5. New Thought. 

## Current Behavior

The window autoscrolls to the top. 

https://github.com/user-attachments/assets/1d4b0a0b-01ea-441e-be76-79331c0e7eff

## Expected Behavior

No autoscroll.

ethan-james:
Fixes #2748, #3075 and #3076

[asyncFocus](https://github.com/ethan-james/em/blob/7a51ebcb9c7263a0f306016f56d08fedf56c6c0a/src/actions/newThought.ts#L245)  was being called in the `newThought` action, which would usually trigger [dispatch(keyboardOpenActionCreator({ value: false }))](https://github.com/ethan-james/em/blob/7a51ebcb9c7263a0f306016f56d08fedf56c6c0a/src/components/Editable/useEditMode.ts#L123) in `useEditMode`. `newThought` would immediately dispatch a `setCursor` action that would generally cause the keyboard to stay open, but not always. Hypothetically, it would have more trouble when exceeding a frame's render budget, which would potentially explain why it's more common with longer lists.

`asyncFocus` is always called, but it doesn't trigger `keyboardOpen` if the following conditions are both true:
1. The thought has text in it
2. This is the first invocation of `newThought` after the app is reloaded
 
This doesn't exactly explain why the keyboard *only* ever closes when creating a new thought from an empty thought. It is possible to get `keyboardOpen: false` to fire by doing the following, starting from #3075:

1. Double-tap `Alabama` to edit
2. Hit enter
3. No `dispatch(keyboardOpenActionCreator({ value: false }))`
4. Hit backspace
5. Hit enter
6. Yes `dispatch(keyboardOpenActionCreator({ value: false }))`

...and yet I have never seen the keyboard close when starting from a thought with text in it. In #3075 you say `(sometimes it takes multiple times)` so maybe you have seen it when creating the first new thought. Although, presumably, not the *first* new thought because `keyboardOpen: false` would not fire if you started from `Alabama` and created a single new thought.

There is also another issue when deleting all of the empty thoughts created during testing, especially now that I've fixed the closing keyboard so it's possible to generate an enormous number of empty thoughts quickly. Start with the repro steps for #3076, then rapidly hit backspace to clear out all of the test thoughts. You should see the keyboard close because it's scrolling too quickly upward.

https://github.com/user-attachments/assets/fa5b46c6-7cf1-4028-bf30-196d1eaa4be0

~This also appears to be related to `asyncFocus`, [but a different use of asyncFocus](https://github.com/ethan-james/em/blob/f9a3d765950528e326c40a8ffe3ad762ace6a24c/src/components/Editable/useEditMode.ts#L89), and one which is already checking for `selection.isThought` so I will need to dig a little deeper and try to fix that as well.~

This appears to be related to delaying [setSelectionToCursorOffset](https://github.com/ethan-james/em/blob/f9a3d765950528e326c40a8ffe3ad762ace6a24c/src/components/Editable/useEditMode.ts#L92), which intermittently allows the keyboard to close before calling `setCursor`. Replacing `setTimeout` with `requestAnimationFrame` presumably shortens that delay and fixes the problem for reasons that I do not sufficiently understand and for now will refer to as "magic".

### DIFF
--- a/src/actions/newThought.ts
+++ b/src/actions/newThought.ts
@@ -29,6 +29,7 @@ import {
 } from '../constants'
 import asyncFocus from '../device/asyncFocus'
 import getTextContentFromHTML from '../device/getTextContentFromHTML'
+import * as selection from '../device/selection'
 import { getChildrenSorted } from '../selectors/getChildren'
 import getNextRank from '../selectors/getNextRank'
 import getPrevRank from '../selectors/getPrevRank'
@@ -241,7 +242,9 @@ export const newThoughtActionCreator =
     // cancel if tutorial has just started
     if (tutorial && tutorialStep === TUTORIAL_STEP_START) return
 
-    if (!preventSetCursor && isTouch) {
+    // asyncFocus can cause disruptive autoscroll behavior, so it's important to avoid it when the selection is already on a thought.
+    // If the selection is already on a thought, then it follows that focus will not be blocked. (#2748, #3075)
+    if (!preventSetCursor && isTouch && !selection.isThought()) {
       asyncFocus()
     }
 
--- a/src/components/Editable/useEditMode.ts
+++ b/src/components/Editable/useEditMode.ts
@@ -83,12 +83,15 @@ const useEditMode = ({
 
         setTimeout fixes it, however it introduces an infinite loop when a nested empty thought is created.
         Not calling asyncFocus when the selection is already on a thought prevents the infinite loop.
+        Also, setTimeout creates a large enough delay that the keyboard will intermittently close on iOS Safari.
+        Replacing setTimeout with requestAnimationFrame closes that gap sufficiently and keeps the keyboard open
+        while rapidly deleting thoughts. (#3129)
       */
         if (isTouch && isSafari()) {
           if (!selection.isThought()) {
             asyncFocus()
           }
-          setTimeout(setSelectionToCursorOffset)
+          requestAnimationFrame(setSelectionToCursorOffset)
         } else {
           setSelectionToCursorOffset()
         }
--- a/src/components/Editable/useEditMode.ts
+++ b/src/components/Editable/useEditMode.ts
@@ -83,9 +83,9 @@ const useEditMode = ({
 
         setTimeout fixes it, however it introduces an infinite loop when a nested empty thought is created.
         Not calling asyncFocus when the selection is already on a thought prevents the infinite loop.
-        Also, setTimeout creates a large enough delay that the keyboard will intermittently close on iOS Safari.
-        Replacing setTimeout with requestAnimationFrame closes that gap sufficiently and keeps the keyboard open
-        while rapidly deleting thoughts. (#3129)
+        Also, setTimeout is frequently pushed into the next frame and the keyboard will intermittently close on iOS Safari.
+        Replacing setTimeout with requestAnimationFrame guarantees (hopefully?) that it will be processed before the next repaint,
+        keeping the keyboard open while rapidly deleting thoughts. (#3129)
       */
         if (isTouch && isSafari()) {
           if (!selection.isThought()) {


ethan-james:
I tested this to make sure #999 is still fixed, and that it doesn't re-introduce an infinite loop. Looks OK to me.

src/actions/newThought.ts
```diff
@@ -241,7 +242,9 @@ export const newThoughtActionCreator =
     // cancel if tutorial has just started
     if (tutorial && tutorialStep === TUTORIAL_STEP_START) return
 
-    if (!preventSetCursor && isTouch) {
+    // asyncFocus can cause disruptive autoscroll behavior, so it's important to avoid it when the selection is already on a thought.
```
raineorshine, line 245: Why does `asyncFocus` disrupt autoscroll behavior? Some more specificity here would help.

I also wonder if it would help if the `hiddenInput` was positioned in the center of the viewport. It would be nice to have a more resilient `asyncFocus` that worked regardless of the selection state before it is called, but maybe that's too ambitious.

src/components/Editable/useEditMode.ts
```diff
@@ -83,12 +83,15 @@ const useEditMode = ({
 
         setTimeout fixes it, however it introduces an infinite loop when a nested empty thought is created.
         Not calling asyncFocus when the selection is already on a thought prevents the infinite loop.
+        Also, setTimeout creates a large enough delay that the keyboard will intermittently close on iOS Safari.
+        Replacing setTimeout with requestAnimationFrame closes that gap sufficiently and keeps the keyboard open
+        while rapidly deleting thoughts. (#3129)
```
raineorshine: `setTimeout(..., 0)` is faster than `requestAnimationFrame`, so this explanation doesn't makes sense. e.g. 60 fps = 16ms per animation frame.

Do you have another explanation? I'd like to update this comment.

src/components/Editable/useEditMode.ts
```diff
@@ -83,12 +83,15 @@ const useEditMode = ({
...
       */
         if (isTouch && isSafari()) {
           if (!selection.isThought()) {
             asyncFocus()
           }
-          setTimeout(setSelectionToCursorOffset)
+          requestAnimationFrame(setSelectionToCursorOffset)
```
raineorshine, line 94: ~~I've tried `requestAnimationFrame` and it led to problems, but I can't recall what they were. I will test it both on a new thoughtspace and on my personal thoughtspace in case the added load makes a difference.~~

I don't see any issues, so we can go ahead and merge and keep an eye out for regressions in case we missed something.

raineorshine (commented):
I haven't tested yet, but I have a few comments/questions.

raineorshine:
> 2\. This is the first invocation of `newThought` after the app is reloaded

Why does `asyncFocus` behave differently on the first invocation?

ethan-james:
> I also wonder if it would help if the `hiddenInput` was positioned in the center of the viewport. It would be nice to have a more resilient `asyncFocus` that worked regardless of the selection state before it is called, but maybe that's too ambitious.

I thought of positioning it differently, and it's definitely worth looking into whether that's the reason why it scrolls to the top.

> **The keyboard only closes when creating a new thought from an empty thought because `window.getSelection().focusNode.nodeType` is `Node.ELEMENT_NODE` on empty thoughts and `Node.TEXT_NODE` on non-empty thoughts.** This causes [selection.isText()](https://github.com/cybersemics/em/blob/7a51ebcb9c7263a0f306016f56d08fedf56c6c0a/src/device/asyncFocus.ts#L36) in `asyncFocus` to return false on empty thoughts and thus fail to prevent the focus as happens when activating New Thought on a non-empty thought.

What I see is:

- `Node.TEXT_NODE` when opening editing mode on an existing thought
- `Node.ELEMENT_NODE` during creation of any new thought (starting from text or empty thought)

This may explain why it behaves differently on the first invocation, although I'm not quite there yet. I can dig into this a little more tomorrow and see if I can prove it.

src/components/Editable/useEditMode.ts
ethan-james: I see:

Screenshot 2025-07-20 130031

My hunch is that `requestAnimationFrame` requests the current animation frame first, which would maybe explain why double-`rAF` is a thing. I'll see if I can find any literature on the subject.

ethan-james: And I actually have lines ordered like:

```
          setTimeout(() => console.info(`setTimeout ${Date.now()}`))
          requestAnimationFrame(() => console.info(`requestAnimationFrame ${Date.now()}`))
```

so I don't think they share the same queue.

ethan-james: Actually, if I log the time when the calls are scheduled, I see:

Screenshot 2025-07-20 131040

which indicates that `requestAnimationFrame` is, in fact, [executed just before the browser's next repaint](https://www.geeksforgeeks.org/javascript/settimeout-vs-requestanimationframe/) and `setTimeout` runs after the repaint. I changed it to ``setTimeout(() => console.info(`setTimeout ${Date.now()}`), 0)`` to make sure that I wasn't triggering some sort of undefined behavior by leaving off the explicit `0`, but it looks the same. Then I removed `requestAnimationFrame` entirely just to make sure, and `setTimeout` still runs after a ~15ms delay.

ethan-james: Then I checked Chrome to see if it behaves more in line with how we'd expect, but saw similar results:

Screenshot 2025-07-20 132315

ethan-james: In vanilla JS, I see it working the way you predicted, where `setTimeout` runs pretty much immediately and `requestAnimationFrame` waits until just before the repaint:

Screenshot 2025-07-20 135606

CodeSandbox: https://codesandbox.io/p/sandbox/6q9nly
Preview: https://6q9nly.csb.app/

The conclusion that I would tentatively draw is that the increased processing required in **em** frequently pushes `setTimeout` into the next frame, while `requestAnimationFrame` is guaranteed (hopefully?) to run before the next frame begins.

raineorshine: > I see:
> 
> Screenshot 2025-07-20 130031
> My hunch is that `requestAnimationFrame` requests the current animation frame first, which would maybe explain why double-`rAF` is a thing. I'll see if I can find any literature on the subject.

Very interesting! I've only ever seen `setTimeout(..., 0)` resolve first, but I do know that the exact ms can very a lot depending on the CPU load. I know that has been a big source of flakiness in the puppeteer tests in the CI.

src/components/Editable/useEditMode.ts
```diff
@@ -83,12 +83,15 @@ const useEditMode = ({
 
         setTimeout fixes it, however it introduces an infinite loop when a nested empty thought is created.
         Not calling asyncFocus when the selection is already on a thought prevents the infinite loop.
+        Also, setTimeout is frequently pushed into the next frame and the keyboard will intermittently close on iOS Safari.
+        Replacing setTimeout with requestAnimationFrame guarantees (hopefully?) that it will be processed before the next repaint,
+        keeping the keyboard open while rapidly deleting thoughts. (#3129)
       */
         if (isTouch && isSafari()) {
           if (!selection.isThought()) {
```
raineorshine, line 91: Also notable that `!selection.isThought()` was already used here instead of `!selection.isText()`.

PR #3133, opened 2025-07-22
Mark slow thought as pending

### LINKED ISSUE #1978: Slowly loading thought not rendered as pending
## Steps to Reproduce

```
- a
  - b
    - c
```

Add `await sleep(2000)` to [replicateChildren](https://github.com/cybersemics/em/blob/c21cdc17813b6fe06f2e112cfc56693a34ed8b4a/src/data-providers/yjs/thoughtspace.ts#L739) to simulate a slow connection.

## Current Behavior

Thoughts are only briefly rendered as pending, then are rendered as normal bullets even though they are still loading.

https://github.com/cybersemics/em/assets/750276/5f93f26a-bb47-4179-8c02-3ce95eace879

## Expected Behavior

Thoughts that are still loading should be rendered with the pulsing animation of pending thoughts.

## Notes

This bug was fixed in a510821 but had to be [reverted](https://github.com/cybersemics/em//commit/8cd02e7bac22a604bffffc3cbce77e1fd357480e) due to a regression. Make sure to test setting the cursor on a pending thought and ensure that it does not stay permanently pending.

> This reverts commit a510821d76ed449c050e1895cd560df3af4d2738. The commit caused a regression in which setting the cursor on a thought while it was still pending would cause it to stay permanently pending. (This could be simulated by adding an artificial delay to replicateThought or fetchDescendants.)
> 
> Reverting this commit does restore some undesirable old behavior: Non-root thoughts are not initially marked as pending in fetchDescendants, so the bullet is not rendered with the pulsing animation as it should. A solution that does not cause the aforementioned regression should be re-attempted.

zukilover:
Fixes #1978 

Re-applied logic from a510821: In `fetchDescendants.ts`, all non-root parents are marked with pending: true. In `updateThoughts.ts`, pending is cleared only after all children have loaded.

A new guard in `mergeUpdates.ts` keeps pending: true unless the incoming update explicitly sets pending or `overwritePending` is passed. This prevents regressions where partial updates unintentionally cleared the flag.

Partial updates now leave pending untouched. When the first child loads, `updateThoughts` explicitly sets pending: false, ending the pulsing animation at the right time.


zukilover:
This PR fulfills the following requirements:

- [x] Thoughts that are still loading should be rendered with the pulsing animation of pending thoughts.
- [x] Setting the cursor on a pending thought and ensure that it does not stay permanently pending.


<details>
<summary>This is how it currently works:</summary>

https://github.com/user-attachments/assets/e1e7bebe-e14f-40fe-86db-6aedf825b464


</details>

zukilover:
I've fixed it so that thoughts are marked as pending when their children are loading.

<details>
<summary>pending thoughts</summary>

https://github.com/user-attachments/assets/40c31876-c124-4d31-9eaf-d688fbf4d2ce


</details>

src/actions/updateThoughts.ts
```diff
@@ -186,8 +188,34 @@ const updateThoughts = (
...
+  const mergedUpdatesMap = keyValueBy(thoughtIndexUpdates, (id, thought) => {
+    if (thought) affectedParentIds.add(thought.parentId)
+    return { [id]: thought }
+  })
+
+  let thoughtIndex = mergeUpdates(thoughtIndexOld, mergedUpdatesMap, { overwritePending })
+
+  // Single sweep: clear pending from affected parents whose children are all loaded
+  if (!overwritePending) {
+    affectedParentIds.forEach(parentId => {
```
raineorshine: We probably don't need `affectedParentIds` or `mergedUpdatesMap`. You can just loop through `thoughtIndexUpdates`. By splitting this simple loop + guard into multiple steps and variables, it is making the code appear more complicated than it actually is.

src/actions/updateThoughts.ts
```diff
@@ -186,8 +188,34 @@ const updateThoughts = (
   const lexemeIndexOld = { ...state.thoughts.lexemeIndex }
   const lexemeIndexUpdatesOld = keyValueBy(lexemeIndexUpdates, key => ({ [key]: lexemeIndexOld[key] }))
 
-  // TODO: Can we use { overwritePending: !local } and get rid of the overwritePending option to updateThoughts? i.e. Are there any false positives when local is false?
-  const thoughtIndex = mergeUpdates(thoughtIndexOld, thoughtIndexUpdates, { overwritePending })
```
raineorshine: Notice how the previous code had a very neat two calls to `mergeUpdates`, one after another. We've lost that nice symmetry in the new code. Let's factor the added logic into a helper function to avoid making `updateThoughts` messier.

raineorshine (commented):
Thanks! I'm still wrapping my head around the solution, but I've done an initial scan.

It's unfortunate that the pending logic is so spread out. If you can think of any architectural improvements that would be great. This is going to be challenging to reason about in the future.

src/actions/updateThoughts.ts
raineorshine: Even better, could this logic be moved into `mergeUpdates`? It has access to `thoughtIndexOld`. That would be better encapsulated, if that's possible.

src/util/mergeUpdates.ts
```diff
@@ -20,15 +20,36 @@ const mergeUpdates = <T>(
...
+      // remain pending indefinitely after navigating to them.
+
+      // Scenarios:
+      // 1. overwritePending is true: always allow overwrite (original behaviour).
+      // 2. currentPending === true: allow overwrite so that pending can be cleared.
+      // 3. currentPending === false && incomingPending === true: block overwrite to keep non-pending state.
+      // 4. currentPending === false && incomingPending === false/undefined: allow overwrite (other fields may update).
+
+      // Determine if the incoming object has its own pending property.
+      const hasPendingProp = Object.prototype.hasOwnProperty.call(value as object, 'pending')
```
raineorshine: This would differentiate `{ pending: undefined }` from `{}`, which is not a good design. It's confusing enough to distinguish `{ pending: undefined }` from `{ pending: false }`.

src/data-providers/data-helpers/fetchDescendants.ts
```diff
@@ -209,7 +210,21 @@ async function* fetchDescendants(
...
-          return thought
+          return {
+            ...thought,
+            // Always mark parents as pending.
+            // Mark parents as pending ONLY if at least one direct child is not yet loaded.
+            // This prevents propagating the pending state up the entire ancestor chain.
+            // updateThoughts will still clear pending once the missing direct children are loaded.
+            pending:
+              !isRoot([thought.id]) &&
+              hasChildren &&
```
raineorshine: Is `hasChildren` superfluous? It's truthy if and only if `childrenIds.some(...)` is truthy.

src/data-providers/data-helpers/fetchDescendants.ts
```diff
@@ -209,7 +210,21 @@ async function* fetchDescendants(
...
+              // If every child already exists in state (regardless of its own pending status),
+              // the parent itself should not be considered pending.
+              // This ensures that pending is applied only to the direct ancestor of the node(s)
+              // currently being fetched rather than all ancestors.
+              childrenIds.some(childId => !getThoughtById(updatedState, childId)),
```
raineorshine, line 225: I might not be understanding this part of the code fully, but if `updateThoughts` clears the pending state anyway, why is it sometimes set to false here? 

The overlap in responsibility is difficult to reason about.

raineorshine (changes requested):
The change seems to work well, but I am admittedly concerned about the added complexity and the lack of encapsulation. With so many different pending states and overwrites, I can't imagine us being able to maintain this properly. This is especially true given that there is no test coverage.

This is a good starting point. If you can integrate this into a more unified vision of the pending state I think it would be a more robust solution. Thanks so much!

src/data-providers/data-helpers/fetchDescendants.ts
zukilover: `fetchDescendants` sets the initial `pending` state for thoughts it loads. `updateThoughts` then checks and clears `pending` for any parents that still have it once all their children are loaded. They both handle `pending` by design, whichever finishes first clears the flag. This keeps the UI fast and accurate.

If we had to pick just one place to handle `pending`, here’s what it would look like:

Only in `updateThoughts`: This keeps all the logic in one place, making it easier to manage. But while deep trees are loading, ancestors stay marked as pending, making the UI unexpectedly busy. It also has to scan parent thoughts every time, which can slow things down.

Only in `fetchDescendants`: The flag turns off immediately, so the UI feels faster and more responsive. There's less work per update since the reducer doesn't need to scan parents. But it runs with partial data, so it can set the wrong flags. To be accurate, it would need to re-check children for every new batch, which is expensive. It also makes consistency harder when data arrives out of order or gets cancelled.

The current approach combines both. It keeps the speed and responsiveness of `fetchDescendants` with the accuracy and reliability of `updateThoughts`, but of course at the cost of some duplicated logic.

src/data-providers/data-helpers/fetchDescendants.ts
zukilover: I agree. This was to make the intent obvious (more readable). I'll remove.

src/util/mergeUpdates.ts
zukilover: I've updated it to: 
`const hasIncomingPending = incomingPending !== undefined`

I've slightly refactored the logic for clearing the parent's pending flag: 1287ce4c56e77408c5e8f32246c613b31eb06e1d. The predicate (clearPending) remains in `updateThoughts`, but it's now passed down to the generic helper and invoked within `mergeUpdates`.

This keeps responsibilities clean, since I see `mergeUpdates` as type-agnostic (it only understands pending flag).

I know this may not fully align with your suggestion to move the logic entirely into `mergeUpdates`. We could do that, but only by giving mergeUpdates awareness of the thought-specific “children loaded” rule. In other words, it would no longer be type-agnostic and would become thought-aware, if that makes sense.

src/data-providers/data-helpers/fetchDescendants.ts
zukilover: Removed.

src/data-providers/data-helpers/fetchDescendants.ts
raineorshine: Thanks for your reply. I have some clarifying questions.

> whichever finishes first clears the flag

I'm not sure what you mean, as they're not racing. `fetchDescendants` streams results to `updateThoughts`. Every chunk of thoughts yielded by `fetchDescendants` will lead to a call to `updateThoughts`.

https://github.com/cybersemics/em/blob/3b8576b4bda9bc4447917dfc40c9198ee4a664bc/src/actions/pull.ts#L83-L101

Originally, `fetchDescendants` was supposed to be part of the db driver and decoupled from the Redux state, but unfortunately this line was blurred in order to allow it to handle cursor and expansion states which affect which thoughts are buffered (pending). Nevertheless, we still try to keep `fetchDescendants` responsible for data retrieval and buffering, and `updateThoughts` responsible for merging fetched thoughts into the Redux store.

> Only in `updateThoughts`: This keeps all the logic in one place, making it easier to manage. But while deep trees are loading, ancestors stay marked as pending, making the UI unexpectedly busy. It also has to scan parent thoughts every time, which can slow things down.

Can you explain what you mean by "scan parent thoughts"? A synchronous parent check doesn't seem expensive.

> Only in `fetchDescendants`: The flag turns off immediately, so the UI feels faster and more responsive. 

No, that's not true. `fetchDescendants` can disable pending, but this still needs to get passed to the `updateThoughts` reducer before it can affect the UI.

> There's less work per update since the reducer doesn't need to scan parents. But it runs with partial data, so it can set the wrong flags. To be accurate, it would need to re-check children for every new batch, which is expensive. 

Can you expand on this? Synchronously checking children doesn't seem expensive, but I might not understand the full extent of what you're describing.

> The current approach combines both. It keeps the speed and responsiveness of `fetchDescendants` with the accuracy and reliability of `updateThoughts`, but of course at the cost of some duplicated logic.

`fetchDescendants` handles thought buffering, so it's the natural place to put pending logic. 

It's going to be very difficult to maintain pending logic spread across multiple files.

Thanks for taking a closer look at the architecture and prioritizing encapsulation and clean design.

src/actions/updateThoughts.ts
zukilover: I've brought back the symmetry although not as nice as it was because we now need to pass `clearPending` here.

src/actions/updateThoughts.ts
zukilover: I've removed `affectedParentIds` and `mergedUpdatesMap`

src/data-providers/data-helpers/fetchDescendants.ts
zukilover: > I'm not sure what you mean, as they're not racing. `fetchDescendants` streams results to `updateThoughts`. Every chunk of thoughts yielded by `fetchDescendants` will lead to a call to `updateThoughts`.

You're right. What I meant was that the first chunk in which a parent is seen with all of its children will clear the flag, whether that happens inside `fetchDescendants` (when the chunk is built) or inside `updateThoughts` (when the chunk is merged). Let me know if there's more I should be aware of.

zukilover: > Can you explain what you mean by "scan parent thoughts"? A synchronous parent check doesn't seem expensive.

On a second thought I think that's true. It only loops over the handful of parents that appear in the current chunk, not the whole tree, so performance is effectively a non-issue.

zukilover: > No, that's not true. fetchDescendants can disable pending, but this still needs to get passed to the updateThoughts reducer before it can affect the UI.

Here's what I see: the flag cleared in `fetchDescendants` still has to travel through `updateThoughts` in the very next dispatch before React sees it, so there is no extra render frame gained, thus the user won’t notice a speed difference here.

zukilover: > Can you expand on this? Synchronously checking children doesn't seem expensive, but I might not understand the full extent of what you're describing.

Because `fetchDescendants` sees the tree one batch at a time, it would have to keep re-checking parents it already sent whenever later batches add more children, so the same rule gets scattered across many passes, making the code harder to follow even if each individual check is quick.

zukilover: I may be inaccurate here, I’m still exploring different design options right now.

raineorshine: > > I'm not sure what you mean, as they're not racing. `fetchDescendants` streams results to `updateThoughts`. Every chunk of thoughts yielded by `fetchDescendants` will lead to a call to `updateThoughts`.
> 
> You're right. What I meant was that the first chunk in which a parent is seen with all of its children will clear the flag, whether that happens inside `fetchDescendants` (when the chunk is built) or inside `updateThoughts` (when the chunk is merged). Let me know if there's more I should be aware of.

Right, that's the part I don't like. It's not a clean separation of responsibility, and it's not cleanly encapsulated.

> > Can you expand on this? Synchronously checking children doesn't seem expensive, but I might not understand the full extent of what you're describing.
> 
> Because `fetchDescendants` sees the tree one batch at a time, it would have to keep re-checking parents it already sent whenever later batches add more children, so the same rule gets scattered across many passes, making the code harder to follow even if each individual check is quick.

`fetchDescendants` sees the tree one batch at a time, and `updateThoughts` gets called one batch a time. So wouldn't the parent check occur the same number of times in either case? The difference is just whether it occurs in a db helper or in a reducer. Keeping all of the buffering logic in `fetchDescendants` would be preferable from a software design perspective.

Thanks for your reply. I appreciate the clear understanding you have developed of the logic.

src/util/mergeUpdates.ts
raineorshine: Thank you, that's a good explanation. `hasIncomingPending` is better.

raineorshine: On second thought (assuming if we don't find a way to handle it in `fetchDescendants`), I think we should move the `clearPending` logic into `mergeUpdates` and make it thought-specific. The pending flag is already thought-specific in practice, despite the ostensibly generic way the function is currently written. The `pending` flag is unique to thoughts, so I think it's okay to build it into `mergeUpdates`.

Injecting `clearPending` is generic, but it's a pretty ugly interface. In this case I don't think it's worth the cost.

src/actions/updateThoughts.ts
raineorshine: I'm not sure that injecting `clearPending` is a good tradeoff here. Some structural symmetry would help make it more readable, but I don't want to complicate the API.

Also see: https://github.com/cybersemics/em/pull/3133#discussion_r2261764185

raineorshine (changes requested):

src/data-providers/data-helpers/fetchDescendants.ts
zukilover: I tried my best putting the pending check entirely in `fetchDescendants`, but it never quite worked. The infinite pending when setting cursor on a thought reoccurred with this approach.  

Fixing that would require extra logic to remember every pending parent and re-evaluate them each batch, which I think is inefficient. I refactored the clear pending logic to `mergeUpdates` instead, I hope this is sufficient.

src/util/mergeUpdates.ts
zukilover: This has been updated as per your suggestion. We retain the structural symmetric in `updateThought` by moving the clear pending logic to `mergeUpdates`.

raineorshine (changes requested):
Thanks, this looks better. While the logic is admittedly complicated, it's not spread across multiple files. Complexity is much easier to manage and less prone to breakage if it's contained.

The only thing I'm wondering about is the heavy use of loops and mutations. As you know, this project uses a functional programming style, so we try to avoid procedural style coding when possible. Is there a reason that you have to determine the pending state in multiple passes? It would be easier to reason about if it was set once.

Maybe you can take a look at that section and consider if a more FP structure is possible (without making a poor tradeoff and inadvertently adding too much complexity for different reasons).

Thanks :)

src/util/mergeUpdates.ts
```diff
@@ -16,26 +16,79 @@ const mergeUpdates = <T>(
...
-      ) {
-        mergeResult[key] = value
+  /** Determine if we should allow overwrite based on pending logic. */
+  const shouldOverwrite = (currentPending: boolean | undefined, incomingPending: boolean | undefined) => {
+    const hasIncomingPending = incomingPending !== undefined
+    // Scenarios:
+    // 1. overwritePending is true: always allow overwrite (original behaviour).
+    // 2. currentPending === true: allow overwrite so that pending can be cleared.
+    // 3. currentPending === false && incomingPending === true: block overwrite to keep non-pending state.
+    // 4. currentPending === false && incomingPending === false/undefined: allow overwrite (other fields may update).
```
raineorshine: We've briefly discussed this before, but what is the difference between `false` and `undefined`? I feel like this needs to be explained more clearly here, because it goes against Javascript conventions where they are treated as the same. The current scenarios only mention true/false combinations.

src/util/mergeUpdates.ts
```diff
@@ -16,26 +16,79 @@ const mergeUpdates = <T>(
...
+    return overwritePending || currentPending !== false || (hasIncomingPending && incomingPending !== true)
```
raineorshine: `hasIncomingPending && incomingPending !== true` is equivalent to `incomingPending !== undefined && incomingPending !== true` which is equivalent to `incomingPending === false`. Am I missing something?

src/util/mergeUpdates.ts
```diff
@@ -16,26 +16,79 @@ const mergeUpdates = <T>(
...
+      return { ...state, [key]: nextValue }
+    },
+    { ...mergeInto } as Index<T | null>,
+  )
+
+  // Check only tracked parents for pending clearing
+  maybePending.forEach(key => {
+    const val = merged[key] as MaybePending | null
+    if (shouldClearPending(val, merged as Index<MaybePending | null>)) {
+      merged[key] = { ...(val as object), pending: false } as T
```
raineorshine: Can you do `merged[key].pending = false`? That would be simpler, and `merged` as a whole is already a new object reference.

src/util/mergeUpdates.ts
```diff
@@ -16,26 +16,79 @@ const mergeUpdates = <T>(
...
+
+      return { ...state, [key]: nextValue }
+    },
+    { ...mergeInto } as Index<T | null>,
+  )
+
+  // Check only tracked parents for pending clearing
+  maybePending.forEach(key => {
+    const val = merged[key] as MaybePending | null
+    if (shouldClearPending(val, merged as Index<MaybePending | null>)) {
```
raineorshine: Presumably `shouldClearPending` needs to look at the final merged value, and that's why this needs to happen in a second pass. Can you confirm?

src/util/mergeUpdates.ts
```diff
@@ -16,26 +16,79 @@ const mergeUpdates = <T>(
...
+  const stripPending = (val: MaybePending): T => {
+    const { pending: _discard, ...rest } = val
+    return rest as T
+  }
+
+  /** True if object has a childrenMap and is pending. */
+  const isPending = (val: T | null) =>
+    !!(val as MaybePending)?.pending && typeof val === 'object' && 'childrenMap' in (val as object)
+
+  // Track parents that might be eligible for pending clearing
```
raineorshine: Let's just call these "thoughts" rather than "parents". Usually we reserve parents to refer to `thought.parentId`.

src/util/mergeUpdates.ts
```diff
@@ -16,26 +16,79 @@ const mergeUpdates = <T>(
...
+  const maybePending = new Set<string>()
```
raineorshine: Perhaps we could call this `pendingKeys` or `pendingThoughtKeys`. From my understanding, it's not that they are maybe pending, but they are actually pending and maybe should be cleared.

src/util/mergeUpdates.ts
zukilover: `false` means it used to be pending but is now fully loaded.  
`undefined` means it was never pending at all, just a normal loaded thought (e.g. when a thought is pulled in because its id appears in a path, which bypasses the pending flag, or when you create a new thought, which is always fully materialised).

src/util/mergeUpdates.ts
zukilover: That's true. I've refactored this block and updated the comments for readability.

src/util/mergeUpdates.ts
zukilover: This line is no longer relevant, I've refactored into a more functional approach.

src/util/mergeUpdates.ts
zukilover: This line is no longer relevant, I've refactored into a more functional approach.

src/util/mergeUpdates.ts
zukilover: This line is no longer relevant, I've refactored into a more functional approach.

src/util/mergeUpdates.ts
zukilover: Good idea. And based on this, I've refactored the line which I thought was still procedural. Kindly check.

raineorshine (approved):
Thanks for the latest updates.

The ternary pending state has me worried, as I think it's going to be very hard to maintain and adapt to the new sync engine. There are too many combinations of states between current pending and incoming pending to make it easy to reason about. However, a full refactor is clearly outside the scope of this issue. Thank you for putting in the extra effort to wrangle the complexity.

Issue #3008, opened 2025-06-03
Add tests for useFilteredCommands


raineorshine:
The logic for [useFilteredCommands](https://github.com/cybersemics/em/blob/2326b3ba1f32e2ad599748bc033113521751608d/src/hooks/useFilteredCommands.ts) is quite complex. Let's add tests to prevent future regressions.

- Alphabetize
- Alphabetize using inverse label
- Filter by search
- Case insensitive search
- Fuzzy search
- Sort exact match to top
- Sort startsWith above contains
- Sort contiguous matches above noncontiguous matches
- Param: `platformCommandsOnly` 
- Mobile: Always show cancel
- Mobile: Always show gesture cheatsheet
- Sort enabled commands above disabled commands. Enabled commands are those for which `canExecute()` returns true.
- Sort recentCommands to top
- Sort selected command to top
- Mobile: Sort cancel to bottom
- Mobile: Sort gesture cheatsheet second-to-bottom
- Mobile: Filter commands that match gesture in progress

raineorshine:
Sure! Thanks for asking.

Some commands have a `labelInverse` property that changes the label in the menu when the command is active:

https://github.com/cybersemics/em/blob/7a51ebcb9c7263a0f306016f56d08fedf56c6c0a/src/%40types/Command.ts#L103-L104

Therefore active commands are sorted by inverse label when the command is active:

https://github.com/cybersemics/em/blob/7a51ebcb9c7263a0f306016f56d08fedf56c6c0a/src/hooks/useFilteredCommands.ts#L81

raineorshine:
For `boolean | undefined` values in Javascript, it's idiomatic to treat `undefined` as `false`.

PR #3139, opened 2025-07-23
Fix 3008: Add tests for `useFilteredCommands`

codebuild-ro:
https://github.com/cybersemics/em/issues/3008

I added tests for `useFilteredCmmands` hook for cases which are mentioned in the issue ticket.

Now there are 3 failing cases. - the hook should be updated for those cases

- **Alphabetize commands by label**
In the mock commands for testing, `Show Hidden Thoughts` has inverse label `Hide Hidden Thoughts` and it is active.
Though, `sortActiveCommandsFirst` is not set - meaning that it is false.
So sorting should reference regular label `Show Hidden Thoughts`, but the hook uses inverse label.

- **Ignore labelInverse when sortActiveCommandsFirst is false**
Though, `sortActiveCommandsFirst` is false, sorting sorting references inverse label for sorting which is not correct.

- **Support case insensitive search**
Searching with `thOUGHT` should return `New Thought` and `New Subthought`, but it returns empty now.


raineorshine (approved):
Great work. I appreciate the way you've asserted exactly what is needed, no more and no less, without accidental dependencies on contingent relationships. Good use of mocks and well documented. Thank you!

PR #3143, opened 2025-07-24
Note: use preventAutoscroll

### LINKED ISSUE #3128: [iOS] Tapping a note will autoscroll to top
## Steps to Reproduce

```
- a
  - b
    - =note
      - this is a note long enough to tap
```

1. Scroll down slightly
2. Tap the note

## Current Behavior

The window autoscrolls to the top.

https://github.com/user-attachments/assets/008534d8-56fb-4dd0-8fa2-e520d0b49b5c

## Expected Behavior

The autoscroll should be prevented by `preventAutoscroll` similar to a thought.

ethan-james:
Fixes #3128 

Notes should use `preventAutoscroll` in `onMouseDown` and `preventAutoscrollEnd` in `onFocus` to replicate the same behavior as editables.

I think this is working well. If the note is very close to the top of the viewport then I see it scroll up to the top, but I believe a different, intentional mechanism is causing that. Let me know.

### DIFF
--- a/src/components/Note.tsx
+++ b/src/components/Note.tsx
@@ -12,6 +12,7 @@ import { setDescendantActionCreator as setDescendant } from '../actions/setDesce
 import { setNoteFocusActionCreator as setNoteFocus } from '../actions/setNoteFocus'
 import { toggleNoteActionCreator as toggleNote } from '../actions/toggleNote'
 import { isTouch } from '../browser'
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Tested and works! Thanks!

> I think this is working well. If the note is very close to the top of the viewport then I see it scroll up to the top, but I believe a different, intentional mechanism is causing that. Let me know.

I was not able to reproduce this.

PR #3148, opened 2025-07-28
Disable scroll if long press has begun, and vice versa

### LINKED ISSUE #3141: [Mobile] Scroll is possible after drag-and-drop activates during a long press in the scroll zone

## Steps to Reproduce

```
- x
```

## Current Behavior

## Expected Behavior

ethan-james:
Fixes #3141

There are two constraints on this problem:

1. It is not possible to stop iOS Safari from scrolling once it has begun
2. It is not possible to cancel a drag in `react-dnd` once it has begun

As such, I present this prospective solution. I welcome any guidance toward mitigating the scope of its impact.

`state.isScrolling` is a new piece of Redux state similar to `keyboardOpen`. I opted to use Redux state because [canDrag](https://github.com/ethan-james/em/blob/8ca9a85e6f3b3882697b421608e6fefb2bf3123c/src/hooks/useDragAndDropThought.tsx#L59) is already accessing it. Any other reactive solution seemed more unwieldy.

[handleScrollDetection](https://github.com/ethan-james/em/blob/8ca9a85e6f3b3882697b421608e6fefb2bf3123c/src/util/handleScrollDetection.ts#L1) attaches event handlers for `scroll`, `touchstart` and `touchend` events to dispatch `scrolling` actions when scrolling begins and ends. I structured it the way I did because it didn't seem possible to throttle it, since scrolling could begin at a time arbitrarily close to the activation of a long press. Since scroll events can have a huge impact on performance, I wanted to be able to set a simple `isScrolling` variable and short-circuit the event handler right away once scrolling has already been detected.

[disableScroll](https://github.com/ethan-james/em/blob/8ca9a85e6f3b3882697b421608e6fefb2bf3123c/src/device/disableScroll.ts#L1) is a new counterpart to `allowScroll`. The `overflow: hidden` approach that `allowScroll` takes unfortunately doesn't work to block this type of scrolling in iOS Safari. I needed this to be in its own file because it needs to be called from [useDragHold](https://github.com/ethan-james/em/blob/8ca9a85e6f3b3882697b421608e6fefb2bf3123c/src/hooks/useDragHold.ts#L39) to register the event handler, and [useDragAndDropThought](https://github.com/ethan-james/em/blob/8ca9a85e6f3b3882697b421608e6fefb2bf3123c/src/hooks/useDragAndDropThought.tsx#L114) to unregister it. So the reference to the `preventDefault` function needs to be stable across those two files. ~~Maybe this could be rolled into `handleScrollDetection.ts` now that I've created that file.~~

All in all, this fix has been working well for me. It's still possible to sometimes get a bullet to highlight while scrolling, but it disappears after `touchend` without triggering any long press or drag-and-drop functionality. This is an improvement on pre-#3119 behavior, where the command menu will open if pressed and scrolled with similar timing.


src/util/handleScrollDetection.ts
```diff
@@ -0,0 +1,55 @@
...
+import { scrollingActionCreator as scrolling } from '../actions/scrolling'
+import { LongPressState } from '../constants'
+import store from '../stores/app'
+
+let isScrolling = false
+let isTouching = false
+
+/** Track scroll behavior in the Redux state in order to prevent drag-and-drop while scrolling. (#3141)
+ * Discard events that come in after the touch ends, and short-circuit the handler if scrolling has already been detected.
+ * Throttling onScroll is dicey because the scroll could begin at a time arbitrarily close to the start of a long press.
```
ethan-james, line 13: I guess this isn't really true because throttling should still execute the function as soon as the first event comes in. I was having some trouble and assumed that throttling was behind it, but I'd better test it again.

ethan-james:
After looking at this again, maybe it would be preferable to patch `react-dnd-touch-backend` just a little bit more?

```
diff --git a/src/TouchBackendImpl.ts b/src/TouchBackendImpl.ts
index 6650cf98fdb08bca5fa450853cb7907c380fe13e..008d71dc8441585e25f92daf4a22de9d8059a7bf 100644
--- a/src/TouchBackendImpl.ts
+++ b/src/TouchBackendImpl.ts
@@ -116,6 +116,10 @@ export class TouchBackendImpl implements Backend {
 		return this.options.document
 	}
 
+	public handleScroll = _.throttle(() => {
+		if (this.timeout) clearTimeout(this.timeout)
+	}, 100)
+
 	public setup(): void {
 		const root = this.options.rootElement
 		if (!root) {
@@ -128,6 +132,7 @@ export class TouchBackendImpl implements Backend {
 		)
 		TouchBackendImpl.isSetUp = true
 
+		document.body.addEventListener('scroll', this.handleScroll, { passive: false })
 		this.addEventListener(root, 'start', this.getTopMoveStartHandler() as any)
 		this.addEventListener(
 			root,
@@ -177,6 +182,7 @@ export class TouchBackendImpl implements Backend {
 			this.handleTopMoveStartCapture as any,
 			true,
 		)
+		document.body.removeEventListener('scroll', this.handleScroll)
 		this.removeEventListener(root, 'start', this.handleTopMoveStart as any)
 		this.removeEventListener(root, 'move', this.handleTopMoveCapture, true)
 		this.removeEventListener(root, 'move', this.handleTopMove as any)
```

In order to throttle it, we would need to get lodash in there, or else copy the throttle function into the `react-dnd-touch-backend` package.

ethan-james:
> What do you want to throttle? `delayTouchStart` is just an initial delay, so I'm confused.

I wanted to throttle the `onScroll` function so that it wasn't running constantly, but the solution contained in `handleKeyboardVisibility` is pretty similar to using lodash's throttle function and I do need both the `isScrolling` and `isTouching` variables to track scrolling across the different event handlers, so there's little gain in throttling that one.

If we went the `react-dnd` patch route, then throttling it would be important in order to prevent `clearTimeout` from running every time it receives a scroll event. I'm not sure if `delayTouchStart` is exactly the right answer there because I haven't tested it, but it seems to me that the function only needs to run once in a given 400-ms interval in order to cancel the timeout that's pending.

> Can you also explain a bit more why this is a problem after #3119 but not before? I think that would help me understand the justification for the additional complexity.

I'm pretty sure that what's going on is, pre-#3119, any movement would cancel the drag. Now that it's using `touchSlop`, it's possible to get the scrollbar moving shortly before the drag commences, which then makes further scrolling impossible to prevent.

raineorshine:
> > What do you want to throttle? `delayTouchStart` is just an initial delay, so I'm confused.
> 
> I wanted to throttle the `onScroll` function so that it wasn't running constantly, but the solution contained in `handleKeyboardVisibility` is pretty similar to using lodash's throttle function and I do need both the `isScrolling` and `isTouching` variables to track scrolling across the different event handlers, so there's little gain in throttling that one.
> 
> If we went the `react-dnd` patch route, then throttling would be important in order to prevent `clearTimeout` from running every time it receives a scroll event. I'm not sure if `delayTouchStart` is exactly the right answer there because I haven't tested it, but it seems to me that the function only needs to run once in a given 400-ms interval in order to cancel the timeout that's pending.

Internally a throttled function is doing a small comparison on every invocation To see if the function can be run again. So from a performance perspective, there's no real difference between the internal conditional of throttle or the `if (this.timeout)` check. So I would omit the throttle.

> > Can you also explain a bit more why this is a problem after #3119 but not before? I think that would help me understand the justification for the additional complexity.
> 
> I'm pretty sure that what's going on is, pre-#3119, any movement would cancel the drag. Now that it's using `touchSlop`, it's possible to get the scrollbar moving shortly before the drag commences, which then makes further scrolling impossible to prevent.

Ah, I see. Actually, I have seen this happen before, albeit rarely! It would be great to eliminate this edge case.

ethan-james:
> Internally a throttled function is doing a small comparison on every invocation To see if the function can be run again. So from a performance perspective, there's no real difference between the internal conditional of throttle or the `if (this.timeout)` check. So I would omit the throttle.

Yeah, I think you're right. Does this mean that you're preferring the `react-dnd-touch-backend` patch? It's definitely simpler, although I'm not entirely confident that the patch I provided above is complete. It might be possible for scroll events to buffer a bit and come in after the timeout has already been invoked. That still seems manageable to me since it can call `this.actions.endDrag()`, but would get a little more complex.

raineorshine:
I'm not clear what the `react-dnd-touch-backend` patch is intended to replace, so I'm not sure off the top of my head which one is less complex. If it replaces this entire PR, then it very well could be the simpler approach, even if you need to call `endDrag()`. What do you think?

---

As an aside, we'll soon have 3–4 independent patches to `react-dnd-touch-backend`, so it might be worth some separate documentation outside the comments that clearly state each of the patches and link to relevant PR's so that we can easily maintain it when we inevitably have to upgrade the dependency in the future.

raineorshine:
> > I'm not clear what exactly the new `react-dnd-touch-backend` patch is intended to replace, so I'm not sure off the top of my head which approach is less complex. If it replaces this entire PR, then it very well could be the simpler approach, even if you need to call `endDrag()`.
> > What do you think?
> 
> It would remove the `handleScrollDetection` logic but leave the `allowTouchToScroll` logic, so the 10 currently-modified files would be reduced to 3. 

Great, thanks. That sounds good.

> As far as what I think, I actually like having `isScrolling` in the Redux state although I'm not quite sure where else to use it right now. 

I tend to prefer Redux state for reactive values that need to be subscribed to. Since `isScrolling` is only used by `canDrag` and `useLongPress`. I also don't want to kick off execution of all selectors when the scroll starts. We don't want to risk any performance impact.

Therefore, I would just use [globals](https://github.com/cybersemics/em/blob/0d343d57ac5aed324cfa318a3107089587699f11/src/globals.ts). We can switch to a ministore in the future if we need basic reactivity, and we can switch to Redux in theh future if we need more complex selectors over 

> However, the current PR is still treating symptoms rather than the root cause, since it's trying to prevent [canDrag](https://github.com/ethan-james/em/blob/4eab5dc4d0ca1213a2405152100424e2e205669d/src/hooks/useDragAndDropThought.tsx#L68) in time to stop a drag, then attempting to cancel all drag-related actions just to be certain that a drag didn't slip through anyway. Patching seems like a more promising solution for actually squashing all of the edge cases since we can control the drag directly.

Yes, that sounds good.

PR #3151, opened 2025-07-30
Hide positioned faux caret by default

### LINKED ISSUE #3095: FauxCaret appears intermittently on refresh
Tested on mobile Safari. 

## Steps to Reproduce

```
- A
  - B
    - C
```

Refresh the app multiple times. 

## Current Behavior

Sometimes the FauxCaret appears, even though the keyboard is down.

https://github.com/user-attachments/assets/afeea706-89f0-4295-93e1-09604da5011a 

It appears to be slightly easier to reproduce in Table View, but I don't think it's specific to Table View. Probably just occurs more often when there are enough thoughts to slow down the render. 

https://github.com/user-attachments/assets/f1826445-e05c-4bae-b393-499c5d33b4b8

## Expected Behavior

The FauxCaret should only appear when navigating between cursors that have different indentation levels, when the keyboard is up.

ethan-james:
Fixes #3095 

The faux caret, specifically the positioned faux caret, had an empty `styles` object by default. The props would be analyzed and the styles would be set inside of a `useEffect`. Since `useEffect` runs asynchronously, presumably it gets pushed into a later frame under heavier load. Now the positioned faux caret is hidden when it is mounted, and the styles are set in the `useEffect` as before.

I did a bit of ad-hoc testing and can see that yes, in a thoughtspace with a single thought, render and `useEffect` are executed within a few milliseconds of each other, while in a much larger thoughtspace, renders come first and most `useEffect` calls happen more like 20-40ms later.

### DIFF
--- a/src/components/FauxCaret.tsx
+++ b/src/components/FauxCaret.tsx
@@ -43,7 +43,7 @@ const FauxCaret = ({
     fontSize?: Property.FontSize
     top?: Property.Top
     left?: Property.Left
-  }>({})
+  }>({ display: caretType === 'positioned' ? 'none' : undefined })
 
   const isEditingCursor = useSelector(state => state.isKeyboardOpen && equalPath(path, state.cursor))
   const isTableCol1 = useSelector(state => path && attributeEquals(state, head(path), '=view', 'Table'))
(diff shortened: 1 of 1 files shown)


ethan-james:
This is the [weirdest test failure](https://github.com/cybersemics/em/actions/runs/16635731058/job/47076003147?pr=3151#step:10:248) yet. I think I've done everything I can to prevent it from making any changes except to iOS Safari, but some `render-thoughts` tests still fail pretty consistently.

PR #3156, opened 2025-07-31
Fix 3090: `Multiline line height is initially incorrect after unellipsizing url`

### LINKED ISSUE #3090: Multiline line height is initially incorrect after unellipsizing url
## Steps to Reproduce

```
- Malleable Software
  - Peter van Hardenberg
    - https://www.inkandswitch.com/essay/malleable-software/?utm_source=inkandswitch&utm_medium=email&utm_campaign=new-essay-malleable-software-restoring-user
```

1. Set the cursor on `Peter van Hardenberg`. 
2. Set the cursor on the url. 

## Current Behavior

The line-height of the url is incorrect on the first paint. 

https://github.com/user-attachments/assets/96bf9378-1186-4770-a1a5-9a6cf536e933

## Expected Behavior

The line-height should be correct.

codebuild-ro:
Fixes #3090

I created and used `useMultiline` hook.
I updated the logic to detect the multiline state in time.
The main point here are:
- the combination of `useLayoutEffect` and `useLayoutAnimationFrameEffect`.
- detecting focusing thought changes

### DIFF
--- a/src/components/Editable/useMultiline.ts
+++ b/src/components/Editable/useMultiline.ts
@@ -1,6 +1,7 @@
 import { RefObject, useCallback, useLayoutEffect, useState } from 'react'
 import { useSelector } from 'react-redux'
 import useLayoutAnimationFrameEffect from '../../hooks/useLayoutAnimationFrameEffect'
+import cursorStore from '../../stores/cursor'
 import editingValueStore from '../../stores/editingValue'
 import viewportStore, { ViewportState } from '../../stores/viewport'
 
@@ -8,7 +9,7 @@ import viewportStore, { ViewportState } from '../../stores/viewport'
  * Detects if a thought is multiline using height measurement.
  *
  * Threshold: fontSize * 2 + 4px buffer
- * Performance optimized to avoid re-renders during typing.
+ * Performance optimized to avoid re-renders during typing and cursor changes.
  *
  * @param editableRef - Reference to the editable element.
  * @param isEditing - Whether the thought is currently being edited, affects rendering.
@@ -17,10 +18,6 @@ import viewportStore, { ViewportState } from '../../stores/viewport'
 const useMultiline = (editableRef: RefObject<HTMLElement>, isEditing?: boolean) => {
   const [multiline, setMultiline] = useState(false)
   const fontSize = useSelector(state => state.fontSize)
-  // While editing, watch the current Value and trigger the layout effect
-  const editingValue = editingValueStore.useSelector(state => (isEditing ? state : null))
-  // To Detect new thought is focused
-  const editingCursor = useSelector(state => state.cursor)
 
   /**
    * Calculate if content is multiline by comparing element height to threshold.
@@ -35,8 +32,21 @@ const useMultiline = (editableRef: RefObject<HTMLElement>, isEditing?: boolean) 
   }, [editableRef, fontSize])
 
   // Two effects for immediate and fallback detection
-  useLayoutEffect(updateMultiline, [updateMultiline, editingValue, editingCursor])
-  useLayoutAnimationFrameEffect(updateMultiline, [updateMultiline, editingValue, editingCursor])
+  useLayoutEffect(updateMultiline, [updateMultiline])
+  useLayoutAnimationFrameEffect(updateMultiline, [updateMultiline])
+
+  const updateEditingMultiline = useCallback(() => {
+    if (isEditing) updateMultiline()
+  }, [updateMultiline, isEditing])
+
+  // Detect cursor changes without causing re-renders
+  cursorStore.useLayoutEffect(updateMultiline)
+
+  // While editing, watch the current Value and trigger the layout effect
+  editingValueStore.useLayoutEffect(updateEditingMultiline)
+  // This is for table view mode - when column2 is merged into column1, we need to re-measure
+  // Reference Video - https://github.com/user-attachments/assets/de49ae80-9829-47e9-a890-f3d760b818b9
+  editingValueStore.useLayoutAnimationFrameEffect(updateEditingMultiline)
 
   const selectInnerWidth = useCallback((state: ViewportState) => state.innerWidth, [])
   // re-measure when the screen is resized
--- a/src/redux-middleware/syncCursorStore.ts
+++ b/src/redux-middleware/syncCursorStore.ts
@@ -1,0 +1,26 @@
+import { ThunkMiddleware } from 'redux-thunk'
+import State from '../@types/State'
+import cursorStore from '../stores/cursor'
+import hashPath from '../util/hashPath'
+
+/**
+ * Middleware that syncs the Redux cursor state with the cursor store.
+ * This allows components to subscribe to cursor changes without causing re-renders.
+ */
+const syncCursorStoreMiddleware: ThunkMiddleware<State> = ({ getState }) => {
+  return next => action => {
+    next(action)
+
+    const state = getState()
+    const currentCursor = state.cursor
+    const currentCursorString = currentCursor ? hashPath(currentCursor) : null
+    const oldCursorString = cursorStore.getState()
+
+    // Update cursor store if it's different from Redux state (string comparison)
+    if (oldCursorString !== currentCursorString) {
+      cursorStore.update(currentCursorString)
+    }
+  }
+}
+
+export default syncCursorStoreMiddleware
--- a/src/stores/app.ts
+++ b/src/stores/app.ts
@@ -18,6 +18,7 @@ import multi from '../redux-middleware/multi'
 import multicursorAlertMiddleware from '../redux-middleware/multicursorAlertMiddleware'
 import pullQueue from '../redux-middleware/pullQueue'
 import scrollCursorIntoView from '../redux-middleware/scrollCursorIntoView'
+import syncCursorStore from '../redux-middleware/syncCursorStore'
 import updateEditingValue from '../redux-middleware/updateEditingValue'
 import updateUrlHistory from '../redux-middleware/updateUrlHistory'
 import validateActionRegistrations from '../util/actionMetadata.registry'
@@ -42,6 +43,7 @@ const middlewareEnhancer = applyMiddleware(
   scrollCursorIntoView,
   clearSelection,
   updateEditingValue,
+  syncCursorStore,
   updateUrlHistory,
   freeThoughts,
   multicursorAlertMiddleware,
--- a/src/stores/cursor.ts
+++ b/src/stores/cursor.ts
@@ -1,0 +1,6 @@
+import reactMinistore from './react-ministore'
+
+/** A store that tracks the current cursor path as a string to avoid ministore array handling issues - it saves path as hash. */
+const cursorStore = reactMinistore<string | null>(null)
+
+export default cursorStore
--- a/src/stores/react-ministore.ts
+++ b/src/stores/react-ministore.ts
@@ -1,6 +1,7 @@
-import { useEffect, useSyncExternalStore } from 'react'
+import { useEffect, useLayoutEffect, useSyncExternalStore } from 'react'
 import Store from '../@types/Store'
 import makeSelectorEffect from '../hooks/makeSelectorEffect'
+import useLayoutAnimationFrameEffect from '../hooks/useLayoutAnimationFrameEffect'
 import ministore, { Ministore } from './ministore'
 
 /** Enhances a generic store with React hooks. */
@@ -18,6 +19,18 @@ const makeReactStore = <U extends Store<any>>(store: U) => {
       [cb],
     )
 
+  /** UseLayoutEffect hook that invokes a callback with current state synchronously after DOM mutations. */
+  const useChangeLayoutEffect = (cb: (state: T) => void) => {
+    const value = useSyncExternalStore(store.subscribe, () => store.getState())
+    useLayoutEffect(() => cb(value), [cb, value])
+  }
+
+  /** UseLayoutAnimationFrameEffect hook that invokes a callback with current state on next animation frame. */
+  const useChangeLayoutAnimationFrameEffect = (cb: (state: T) => void) => {
+    const value = useSyncExternalStore(store.subscribe, () => store.getState())
+    useLayoutAnimationFrameEffect(() => cb(value), [cb, value])
+  }
+
   function useSelector<U>(selector: (state: T) => U): U
   function useSelector(): T
   /** A hook that subscribes to a slice of the state. If no selector is given, subscribes to the whole state. */
@@ -33,6 +46,8 @@ const makeReactStore = <U extends Store<any>>(store: U) => {
   return {
     ...store,
     useEffect: useChangeEffect,
+    useLayoutEffect: useChangeLayoutEffect,
+    useLayoutAnimationFrameEffect: useChangeLayoutAnimationFrameEffect,
     useSelector: useSelector as <U>(selector: (state: T) => U) => U,
     useSelectorEffect: makeSelectorEffect(store),
     useState: useSelector as () => T,


raineorshine (changes requested):
Here's what happens when you use padding instead of line-height:

## Current Behavior

b5440755b3

Clicking the space between thoughts sets the caret to the beginning or end of the thought:

https://github.com/user-attachments/assets/f1f021f2-3b98-45d1-84df-984f24dae321

## Expected Behavior

`main`

Clicking the space between thoughts sets the caret near the tap point:

https://github.com/user-attachments/assets/00b2c044-ebf3-48ad-b20b-289a35ef374b

src/hooks/useThoughtMultiline.ts
```diff
@@ -0,0 +1,72 @@
...
+ * - Excludes ellipsized URLs from detection.
+ */
+const useThoughtMultiline = (
+  editableRef: RefObject<HTMLElement>,
+  thoughtWrapperRef: RefObject<HTMLElement>,
+  value: string,
+  ellipsizedUrl: boolean,
+) => {
+  // State for delayed DOM measurement results
+  const [multilineDelayed, setMultilineDelayed] = useState(false)
```
raineorshine: Can you initialize `multilineDelayed` to `calculateMultiline()` instead of `false`? That might allow you to get rid of `multilineImmediate`.

src/components/Thought.tsx
```diff
@@ -323,6 +327,15 @@ const ThoughtContainer = ({
   const styleContainer = useThoughtStyleContainer({ children, env, styleContainerProp, thoughtId, path })
   const value = useSelector(state => getThoughtById(state, thoughtId)?.value)
 
+  // Determine if the thought contains a URL that should be ellipsized
+  // URLs should not be treated as multiline to prevent unwanted line breaks
+  const ellipsizedUrl = !isEditing && containsURL(value || '')
+
+  // Detect if the thought content spans multiple lines
+  // This hook handles timing issues and provides both immediate and delayed calculations
+  // to prevent layout flickering while ensuring accurate measurements
```
raineorshine: These are implementation details that the Thought component does not need to know about. For encapsulation purposes, it's better to keep those details contained in the custom hook unless the parent component needs to know about them.

src/hooks/useThoughtMultiline.ts
```diff
@@ -0,0 +1,72 @@
...
+   * - Collapsed thoughts being expanded
+   * - Dynamic content loading
+   * - DOM elements not yet created.
+   *
+   * The layout effect ensures we get accurate measurements after the DOM
+   * has been updated but before the browser paints.
+   */
+  useLayoutAnimationFrameEffect(() => {
+    setMultilineDelayed(calculateMultiline())
+  }, [value, ellipsizedUrl, calculateMultiline])
```
raineorshine: Make sure to include `editingValue` and `fontSize` as dependencies, since they require recalculating the multiline value.

src/hooks/useThoughtMultiline.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Calculates whether the thought content is multiline based on DOM measurements.
+   *
+   * Detection Logic:
+   * 1. Early return if URL should be ellipsized (prevents multiline for URLs)
+   * 2. Early return if DOM elements don't exist yet
+   * 3. Check if content overflows by comparing:
+   * - scrollWidth vs offsetWidth (horizontal overflow)
+   * - clientWidth vs wrapper clientWidth (container overflow).
```
raineorshine: Can you explain these two types of overflow? I'd like to better understand which circumstances causes one to be true but not the other, and why the combination of both is correct.

raineorshine (changes requested):
Thank you! This is an interesting and novel approach to detecting multiline. It seems like potentially a big improvement over the version that uses `getComputedStyle`.

What is the fate of the original `useMultiline`? Is this meant to replace it? Please advise and ensure the entire codebase is maintained rather than leaving dead modules.

src/components/StaticThought.tsx
```diff
@@ -115,8 +118,6 @@ const StaticThought = ({
   const homeContext = isRoot(simplePath) && !isContextPending
   const value = useSelector(state => getThoughtById(state, head(simplePath))?.value) ?? ''
   // store ContentEditable ref to update DOM without re-rendering the Editable during editing
-  const editableRef = React.useRef<HTMLInputElement>(null)
```
raineorshine: This line has been deleted, but the comment that was with it remains.

raineorshine (changes requested):
> ## Issue
> **Flickering** (changing lineHeight from `2` to `1.25`): Visual jumps when switching between single/multiline states because of changing of line-height from `2` to `1.25` The line height was 2 initially and then as multiline style applying, line height will be changed to 1.25.
> 
> ## Approach
> Before, line height of thought is single line thought was `2` and multiline thought was `1.25`. But now the thought line height is always `1.25`. I compensate rest 0.75 with padding-top: 0.4em and padding-bottom: 0.35em. So 1.25 + 0.4 + 0.35 = 2.

Please update this to match the new approach. Thanks!

raineorshine:
`[src/e2e/puppeteer/__tests__/table-view.ts > Table View > col2 narrow + no wrap width: src/e2e/puppeteer/__tests__/table-view.ts#L55](https://github.com/cybersemics/em/pull/3156/files#annotation_37355224849)` failed twice in a row, so you may need to investigate that. I'll run it a third time just in case.

src/hooks/useThoughtMultiline.ts
codebuild-ro: We need `multilineImmediate` to get the fresh calculation on every render.
`multilineImmediate` is Needed.

**The Problem:**
- `useState(calculateMultiline())` only calls `calculateMultiline()` **once** during initialization
- On subsequent renders, `multilineDelayed` holds the **stale** value from the previous render
- This causes flickering because the state doesn't update immediately with new calculations

*The Solution:*
```typescript
// ✅ This gives us the delayed measurement when DOM is ready
const [multilineDelayed, setMultilineDelayed] = useState(false)

// ✅ This gives us fresh calculation on EVERY render
const multilineImmediate = calculateMultiline()

// ✅ Combined result prevents flickering
return multilineDelayed || multilineImmediate
```

*How It Prevents Flickering:*
1. **First render**: `multilineImmediate` provides immediate feedback
2. **Subsequent renders**: `multilineImmediate` continues to provide fresh calculations
3. **When DOM is ready**: `multilineDelayed` gets updated with accurate measurement
4. **Result**: Always have a value, preventing layout shifts

*Without `multilineImmediate`:*
- Only `multilineDelayed` would be used
- On re-renders, it would show stale values
- This would cause flickering as the state updates after the layout effect

The `multilineImmediate` calculation is crucial for preventing flickering by ensuring we always have the most current calculation available.

src/hooks/useThoughtMultiline.ts
codebuild-ro: ### **1. `scrollWidth > offsetWidth` (Content Overflow)**
**Detects:** Content that's wider than its container
**When true:** Long words, non-breaking text, horizontal overflow
**Example:** A very long word that doesn't wrap

**I removed this case since it is for ellipsized text which is single line**

### **2. `clientWidth >= wrapper clientWidth` (Container Overflow)**

**Detects:** Content that fills or exceeds container boundaries
**When true:** Wrapped text that still overflows, layout constraints
**Example:** A paragraph that wraps but still doesn't fit

src/hooks/useThoughtMultiline.ts
codebuild-ro: I added them to dependencies array.

src/components/Thought.tsx
codebuild-ro: I moved that comment to custom hook.

src/components/StaticThought.tsx
codebuild-ro: I removed that comment.

codebuild-ro:
> Thank you! This is an interesting and novel approach to detecting multiline. It seems like potentially a big improvement over the version that uses `getComputedStyle`.
> 
> What is the fate of the original `useMultiline`? Is this meant to replace it? Please advise and ensure the entire codebase is maintained rather than leaving dead modules.

`useMultiline` is no longer needed, `useThoughtMultiline` can replace `useMultiline`'s functionality totally. I will remove it.

src/components/Thought.tsx
```diff
@@ -308,6 +312,8 @@ const ThoughtContainer = ({
   )
   const isInContextView = useSelector(state => isContextViewActive(state, parentOf(path)))
 
+  const fontSize = useSelector(state => state.fontSize)
```
raineorshine: Push selectors *down* the hierarchy for better encapsulation and less prop drilling.

src/components/Thought.tsx
```diff
@@ -606,13 +615,17 @@ const ThoughtContainer = ({
...
             allowSingleContext={allowSingleContext}
+            // Pass multiline state to StaticThought for proper styling
+            // This determines whether to apply single-line or multiline CSS classes
+            multiline={isMultiline}
+            editableRef={editableRef}
             env={env}
             isContextPending={isContextPending}
             isEditing={isEditing}
-            ellipsizedUrl={!isEditing && containsURL(value)}
+            ellipsizedUrl={!isEditing && containsURL(value || '')}
```
raineorshine: `value` is already typed as non-null above:

https://github.com/cybersemics/em/blob/9889f4a15288d82c787e2ea5b4e3ef976f8fe4f3/src/components/Thought.tsx#L521-L522

src/e2e/puppeteer/__tests__/table-view.ts
```diff
@@ -30,6 +31,8 @@ describe('Table View', () => {
           - Cincinnati, Ohio
     `)
 
+    await waitForFrames()
```
raineorshine, line 36: Let's add a comment for each of the `waitForFrames` explaining why they are needed.

src/hooks/useThoughtMultiline.ts
```diff
@@ -0,0 +1,82 @@
...
+   */
+  const calculateMultiline = useCallback(() => {
+    if (!editableRef.current || !thoughtWrapperRef.current) {
+      return false
+    }
+
+    // Check if content overflows the available width
+    return editableRef.current.clientWidth >= thoughtWrapperRef.current.clientWidth
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
raineorshine: Please do not disable the linter.

src/hooks/useThoughtMultiline.ts
```diff
@@ -0,0 +1,82 @@
...
+  }, [editableRef, thoughtWrapperRef, contentWidth, fontSize, editingValue])
+
+  // Immediate calculation for current render (prevents flickering)
+  const multilineImmediate = calculateMultiline()
```
raineorshine: Maybe you want to do this in `useLayoutEffect`? It's not really appropriate to access the DOM directly in a React render function.

src/hooks/useThoughtMultiline.ts
```diff
@@ -0,0 +1,82 @@
...
+    setMultilineDelayed(calculateMultiline())
+  }, [calculateMultiline, isEditing])
+
+  /**
+   * Return the combined result: either immediate calculation or delayed measurement.
+   *
+   * This ensures we have a value for the current render (preventing flicker)
+   * while also getting accurate measurements when the DOM is ready.
+   */
+  return multilineDelayed || multilineImmediate
```
raineorshine: I'm not totally satisfied with the handoff between `multilineDelayed` and `multilineImmediate`. It seems like you should be able to find a way to get this into a single variable.

src/components/Thought.tsx
codebuild-ro: I fixed this

src/components/Thought.tsx
codebuild-ro: This is fixed.

src/hooks/useThoughtMultiline.ts
codebuild-ro: `calculateMultiline` function should be recreated for changes of contentWidth, fontSize and editingValue

src/e2e/puppeteer/__tests__/table-view.ts
codebuild-ro: I added comments.

src/hooks/useThoughtMultiline.ts
codebuild-ro: `calculateMultiline` function should be recreated for changes of contentWidth, fontSize and editingValue

src/hooks/useThoughtMultiline.ts
codebuild-ro: I removed DOM access during render - now uses useMemo for immediate estimate
I simplified state management - single isMultiline state instead of two variables
I followed React patterns for DOM measurements - Returns estimate until DOM is measured, then returns accurate result

src/hooks/useThoughtMultiline.ts
codebuild-ro: I used one variable instead of 2.

src/hooks/useThoughtMultiline.ts
codebuild-ro: I removed this part.

src/components/Editable/useMultiline.ts
raineorshine, line 1: For better commit history continuity, reuse `useMultiline.ts` rather than creating `useThoughtMultiline.ts`.

src/hooks/useThoughtMultiline.ts
```diff
@@ -0,0 +1,96 @@
...
+ * - Prevents layout flickering by updating before browser paints
+ * - Uses dual effects for maximum reliability and edge case handling
+ * - Encapsulates fontSize selector to reduce prop drilling
+ * - Follows React best practices (no DOM access during render).
+ *
+ * @param editableRef - Reference to the editable element containing the thought text.
+ * @returns Boolean indicating if the thought content spans multiple lines.
+ */
+const useThoughtMultiline = (editableRef: RefObject<HTMLElement>) => {
+  // Get fontSize from Redux store to avoid prop drilling
```
raineorshine: "to avoid prop drilling" is not something that needs to be called out in comments. That's just a best practice.

src/hooks/useThoughtMultiline.ts
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * 1. UseLayoutEffect: Immediate synchronous update to prevent flickering
+   * - Runs synchronously after DOM mutations but before browser paints
+   * - Ensures line-height changes happen immediately without visual glitches
+   * - Handles most common cases: initial render, content changes, window resize.
+   *
+   * 2. UseLayoutAnimationFrameEffect: Asynchronous safety net for edge cases
+   * - Runs on next animation frame to catch any missed changes
+   * - Handles edge cases where DOM updates are delayed or batched
+   * - Acts as a "belt and suspenders" approach for maximum reliability.
```
raineorshine: I don't know what "belt and suspenders" means. Please remove the fluff and try to describe things as they are without marketing language.

src/hooks/useThoughtMultiline.ts
raineorshine: Thank you for adding the comment explaining the reasoning.

src/hooks/useThoughtMultiline.ts
raineorshine, line 1: There are calculation dependencies like `splitThoughtValue` that were in `useMultiline` that are now missing.

Please include all calculation dependencies, or we're going to have regressions. If you're sure that a dependency is no longer needed, or is covered by another case, please provide evidence, which may include git blame, testing the original issue, etc. 

As usual, we don't delete code lightly.

src/components/Editable/useMultiline.ts
codebuild-ro: I removed `useThoughtMultiline` and reverted `useMultiline` back and updated the content.

src/hooks/useThoughtMultiline.ts
codebuild-ro: Removed that part

src/hooks/useThoughtMultiline.ts
codebuild-ro: I updated comments for many parts.

src/hooks/useThoughtMultiline.ts
codebuild-ro: `splitThoughtValue` was considered because `useLayoutEffect` doesn't dectect changes properly.
It was because some delay of DOM update, so `useLayoutEffect` couldn't dectect.
To resolve that I combined the usage of `useLayoutEffect` and `useLayoutAnimationFrameEffect` not to missing any DOM changes.
So considering `splitting thought and merging thought` can be removed by the updated approach.

src/components/Editable/useMultiline.ts
```diff
@@ -1,71 +1,94 @@
...
-    // 2x the single line height would indicate that the thought was multiline if it weren't for the change in line-height and padding.
-    // 1.2x is used for a more forgiving condition.
-    // 1.5x can cause multiline to alternate in Safari for some reason. There may be a mistake in the height calculation or the inclusion of padding that is causing this. Padding was added to the calculation in commit 113c692. Further investigation is needed.
-    // See: https://github.com/cybersemics/em/issues/2778#issuecomment-2605083798
-    setMultiline(height - paddingTop - paddingBottom > singleLineHeight * 1.2)
-  }, [contentRef, fontSize])
+  // To Detect immediate editing value change
+  const editingValue = editingValueStore.useSelector(state => state)
+  // To Detect device width change
+  const contentWidth = viewportStore.useSelector(state => state.contentWidth)
```
raineorshine: These will force the render function to re-run whenever `editingValue` or `contentWidth` changes, even multiline is unchanged. The previous implementation carefully avoided this by using the specialized hook `useSelectorEffect` to skip straight to the effect.

Since the multiline update is only a side effect, please make sure there are no intermediate values that trigger the render function. The Editable component is extremely performance sensitive, and `editingValueStore` updates in real-time as the user types.

src/hooks/useThoughtMultiline.ts
raineorshine: Thank you for explaining and justifying the removal. That sounds reasonable.

raineorshine (changes requested):

src/components/Editable/useMultiline.ts
raineorshine: Update: The existing implementation uses `useSelector`, but narrows it to `isEditing`, i.e. the cursor thought.

codebuild-ro: By detecting only `isEditing`, the problem can't be solved.
The multiline status shuold be calculated for each thought, but if we use only `isEditing`, multiline status will be calculated newly for the only focusing and blurring thoughts - only 2 thoughts.
And the other thoughst won't get any new multiline calculation - this is necessary for moving up and down of cursor (when cursor is moving up and down, the indent sizes will be changed when the thought tree is big).
So as the result, using only `isEditing` cannot be generic solution.

codebuild-ro: I updated the `useMultiline.ts`
```
/**
 * Update multiline state only when it actually changes to prevent calling of setState.
 * Uses ref to track previous state and avoid redundant setState calls.
 */
const updateMultiline = useCallback(() => {
  const multilineState = calculateMultiline()
  if (isMultilineRef.current !== multilineState) {
    isMultilineRef.current = multilineState
    setIsMultiline(multilineState)
  }
}, [calculateMultiline])

useLayoutEffect(updateMultiline, [updateMultiline])
useLayoutAnimationFrameEffect(updateMultiline, [updateMultiline])

editingValueStore.useLayoutEffect(updateMultiline)
editingValueStore.useLayoutAnimationFrameEffect(updateMultiline)

viewportStore.useLayoutEffect(updateMultiline)
viewportStore.useLayoutAnimationFrameEffect(updateMultiline)
```

codebuild-ro: I added `useChangeLayoutEffect` and `useChangeLayoutAnimationFrameEffect` into react-ministore.ts.
I didn't follow the pattern of `useChangeEffect`:
```
useEffect(
  () =>
    store.subscribe(() => {
      cb(store.getState())
    }),
  [cb],
)
```
With using that pattern, it doesn't have real necessity of using `useLayoutEffect` and `useLayoutAnimationFrameEffect`.
If I use that pattern, the `useLayoutEffect` and `useLayoutAnimationFrameEffect` will be called when just callback (`cb`) is changed, but not when state changes.
The callback will be called for immediate change of store state.
But the purpose of creating `useChangeLayoutEffect` and `useChangeLayoutAnimationFrameEffect` is to detect changes of store changes with property of `useLayoutEffect`(after DOM changes) and `useLayoutAnimationFrameEffect`(after next animation).

codebuild-ro: With this approach, I reached to generic solution.

src/components/Editable/useMultiline.ts
```diff
@@ -1,71 +1,63 @@
...
+ *
+ * Threshold: fontSize * 2 + 4px buffer
+ * Performance optimized to avoid re-renders during typing.
+ *
+ * @param editableRef - Reference to the editable element.
+ * @returns Boolean indicating if content spans multiple lines.
+ */
+const useMultiline = (editableRef: RefObject<HTMLElement>) => {
+  // Ref to prevent setting the same multiline state multiple times
+  const isMultilineRef = useRef(false)
```
raineorshine: I'm not quite seeing the reason for `isMultilineRef`. Calling `setIsMultiline` with the same value multiple times is already a no-op. Can you explain this?

codebuild-ro: I see, thank you for pointing out.
It is not necessary because of `bailing out` from React 18+.
I will update it properly, but `updateMultiline` function will still be generated by `useCallback`.
Because, `useLayoutAnimationFrameEffect` will be also run for the callback changes too.

```
const updateMultiline = useCallback(() => {
  const multilineState = calculateMultiline()
  setIsMultiline(multilineState)
}, [calculateMultiline])
```

raineorshine: Great, thanks.

Is there a reason you split out `calculateMultiline` from `updateMultiline`? It's one extra function to keep track of, and seems to just add to the complexity.

codebuild-ro: Sorry for mess, right, seems my head got a little bit hard now. :)
I removed `updateMultiline` function and now in `calculateMultiline` function, the setState is called directly.

codebuild-ro:
```
const calculateMultiline = useCallback(() => {
  if (!editableRef.current) return

  const singleLineThreshold = fontSize * 2 + 4
  // Update multiline state
  setIsMultiline(editableRef.current!.clientHeight > singleLineThreshold)
}, [editableRef, fontSize])
```

src/components/Editable/useMultiline.ts
raineorshine: > By detecting only `isEditing`, the problem can't be solved.
> The multiline status shuold be calculated for each thought, but if we use only `isEditing`, multiline status will be calculated newly for the only focusing and blurring thoughts - only 2 thoughts.
> And the other thoughst won't get any new multiline calculation - this is necessary for moving up and down of cursor (when cursor is moving up and down, the indent sizes will be changed when the thought tree is big).
> So as the result, using only `isEditing` cannot be generic solution.

Hi, sorry for the delay!

When the user is typing, only the thought being edited can change multiline status. The other thoughts can be pushed down, but their height does not change. `editingValueStore` should not be used to force recalculations for the cursor changing, as the cursor and editing are different actions. Editing can only cause one thought to reflow; the cursor moving can cause all of the thoughts to reflow.

`editingValueStore` updates in real-time as the user types, so anything that subscribes to it is extremely performance sensitive. More general solutions are great, but not at the cost of performance when we are well aware that we are in a performance sensitive part of the application.

codebuild-ro: Hi @raineorshine 
I think we need to consider this issue (https://github.com/cybersemics/em/issues/3096) at the same time.
This is not same case but similar case with this ticket.
That issue is fixed and merged, but yet it didn't resolve the issue of incorrect initial line height when thought gets multilined.
So it is occuring flickering.
To fixed that issue, we need to fix for the cases of both - **unellipsizing url** and **moving cursor**.

codebuild-ro: Thank you @raineorshine 
You are absolutely right!

src/components/Editable/useMultiline.ts
```diff
@@ -1,65 +1,42 @@
...
+ * @param isEditing - Whether the thought is currently being edited, affects rendering.
+ * @returns Boolean indicating if content spans multiple lines.
+ */
+const useMultiline = (editableRef: RefObject<HTMLElement>, isEditing?: boolean) => {
   const [multiline, setMultiline] = useState(false)
   const fontSize = useSelector(state => state.fontSize)
   // While editing, watch the current Value and trigger the layout effect
   const editingValue = editingValueStore.useSelector(state => (isEditing ? state : null))
+  // To Detect new thought is focused
+  const editingCursor = useSelector(state => state.cursor)
```
raineorshine: Once again—`useSelector` causes the render function to re-trigger whenever its return value changes. While this is fine in normal react usage, we're dealing with a very performative sensitive part of the application. `useMultiline` is carefully designed to recalculate the multiline state directly from the DOM *without* re-triggering the render function of every thought. Do not use intermediate values as effect dependencies. Use specialized functions like `useSelectorEffect` or `useLayoutEffect`. This should already be clear from the previous implementation. Please follow the existing approach to preserve the desired performance characteristics.

Let me know if there is anything I can clarify.

src/components/Editable/useMultiline.ts
```diff
@@ -1,65 +1,42 @@
...
+ *
+ * @param editableRef - Reference to the editable element.
+ * @param isEditing - Whether the thought is currently being edited, affects rendering.
+ * @returns Boolean indicating if content spans multiple lines.
+ */
+const useMultiline = (editableRef: RefObject<HTMLElement>, isEditing?: boolean) => {
   const [multiline, setMultiline] = useState(false)
   const fontSize = useSelector(state => state.fontSize)
   // While editing, watch the current Value and trigger the layout effect
   const editingValue = editingValueStore.useSelector(state => (isEditing ? state : null))
```
raineorshine: This looks good. Note that while it is an intermediate value that will cause the render function to re-trigger, it only affects the cursor thought, so it shouldn't affect performance too badly.

To bypass the intermediate value, we would need to add a `useLayoutSelectorEffect` and `useLayoutAnimationFrameSelectorEffect` to `react-ministore`. Since the existing implementation uses `useSelector`, I consider this additional optimization to be out of scope.

raineorshine (changes requested):

raineorshine (changes requested):
Hi, thanks for the update. Unfortunately, I'm not willing to add an entire new middleware and ministore just for the cursor. It's too much. If you need to add a new hook that’s fine.

codebuild-ro:
Hi @raineorshine,
If we make an custom hook, we need also hooks to detect the cursor change - it will still causes re-renders.
Any approach using `useSelector(state => state.cursor)` will cause re-renders.
There is a possible approach - `useSelectorEffect(updateMultiline, state => state.cursor, equalPath)`

But this is also need update of `makeSelectoreEffect` - because `requestAnimationFrame`.
https://github.com/cybersemics/em/blob/db56015d3eed697fb108b517761828b56a2724ed/src/hooks/makeSelectorEffect.ts#L47-L48
Because of `requestAnimationFrame`, the callback can't be run immediately for DOM update - it will be run after next frame of DOM update moment which will make delay.

So we need one of:
- To create store which saves cursor specifically
- To update `makeSelectoreEffect`

But updating `makeSelectoreEffect` would be much more.
I assume that "too much" might be about the middleware, not the store itself.
So I am going to remove middleware and add this snippet to `useMultiline.ts` to update cursor manually.

```
useEffect(() => {
  const cursor = store.getState().cursor
  const cursorHash = cursor ? hashPath(cursor) : null
  cursorStore.update(cursorHash)
}, [isEditing])
```

Will it be acceptable?

raineorshine:
> Hi @raineorshine,
> If we make an custom hook, we need also hooks to detect the cursor change - it will still causes re-renders.
> Any approach using `useSelector(state => state.cursor)` will cause re-renders.
> There is a possible approach - `useSelectorEffect(updateMultiline, state => state.cursor, equalPath)`

Yes, exactly. We want to avoid `useSelector`.

> So we need one of:
> 
> * To create store which saves cursor specifically
> * To update `makeSelectoreEffect`
> 
> But updating `makeSelectoreEffect` would be much more. I assume that "too much" might be about the middleware, not the store itself. So I am going to remove middleware and add this snippet to `useMultiline.ts` to update cursor store manually.
> 
> ```
> useEffect(() => {
>   const cursor = store.getState().cursor
>   const cursorHash = cursor ? hashPath(cursor) : null
>   cursorStore.update(cursorHash)
> }, [isEditing])
> ```
> 
> Will it be acceptable?

No, we don't want a dedicated cursor store. The cursor is stored in Redux. It wouldn't make sense to store it in two different stores.

I think we better stop here, as we've exceeded the budget on this task. Thanks so much for all your effort.

PR #3158, opened 2025-08-01
Fix: Flaky Tests > Render Thoughts

karunkop:
## Problem

Tests in [render-thoughts](https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/__tests__/render-thoughts.ts) were very flaky. Tests were failing because of very minor pixel differences as shown in the image below, to be accurate about 14 pixels difference.

image

By default, jest-image-snapshot compares screenshots using pixel-by-pixel matching (`pixelmatch`). But I believe we can switch to comparisonMethod: [`ssim`](https://github.com/americanexpress/jest-image-snapshot?tab=readme-ov-file#ssim-performance-considerations). SSIM compares overall structure and appearance, not individual pixels. It offers fewer false positives than `pixelmatch`, especially with slight font or rendering differences which should be reasonable for these tests scenario where flakiness is observed with a very minor pixel differences.

[Here](https://github.com/karunkop/em/actions?query=branch%3Aflaky-render-thoughts) are the checks conducted after implementing this comparison method.  FYI, Some of the test failures are related with other tests.


karunkop:
Updated PR : #3164

PR #3160, opened 2025-08-01
Use allowTouchToScroll to block scrolling after long press begins

### LINKED ISSUE #3141: [Mobile] Scroll is possible after drag-and-drop activates during a long press in the scroll zone

ethan-james:
Fixes #3141 

This is an improved version of #3148 

There are two constraints on this problem:

1. It is not possible to stop iOS Safari from scrolling once it has begun
2. It is not possible to cancel a drag in react-dnd once it has begun

[allowTouchToScroll](https://github.com/ethan-james/em/blob/1e86a86415cf178282a006d37055e15016dcd458/src/device/allowTouchToScroll.ts#L1) is a new counterpart to allowScroll. The overflow: hidden approach that `allowScroll` takes unfortunately doesn't work to block this type of scrolling in iOS Safari. I needed this to be in its own file because it needs to be called from [useDragHold](https://github.com/ethan-james/em/blob/1e86a86415cf178282a006d37055e15016dcd458/src/hooks/useDragHold.ts#L39) to register the event handler, and [useDragAndDropThought](https://github.com/ethan-james/em/blob/1e86a86415cf178282a006d37055e15016dcd458/src/hooks/useDragAndDropThought.tsx#L113) to unregister it. So the reference to the `preventDefault` function needs to be stable across those two files.

[handleScroll](https://github.com/ethan-james/em/blob/7f09e82887770c8f6723ad5020ff2f1df298bc92/.yarn/patches/react-dnd-touch-backend-patch-2c3a2052b6.patch#L12) watches for scroll events and cancels drag-and-drop functionality either before it has begun (`clearTimeout`) or after it has begun (`this.actions.endDrag()`). This is necessary because when we attempt to block `scroll` events by calling `preventDefault` on `touchmove` events, then the subsequent `scroll` events start coming in very slowly and we can't count on being able to respond to them quickly enough to cancel drag-and-drop before it begins.

[scrollAtEdge](https://github.com/ethan-james/em/blob/7f09e82887770c8f6723ad5020ff2f1df298bc92/src/util/initEvents.ts#L270) also produces `scroll` events that should not cancel drag-and-drop. Since the timing of scroll events is not very precise, we can't really remove `handleScroll` as an event listener at any particular time. Setting a global flag declaring that `scrollAtEdge` is active allows us to avoid canceling drag-and-drop based on `scroll` events from that source.

I created an alternate solution that does not rely on the `window['scrolling']` flag: #3161.

### DIFF
--- a/.yarn/patches/README.md
+++ b/.yarn/patches/README.md
@@ -6,12 +6,13 @@ Usage Error: Multiple candidate packages found; explicitly choose one of them (u
 
 - react-dnd-touch-backend@patch:react-dnd-touch-backend@npm%3A16.0.1#~/.yarn/patches/react-dnd-touch-backend-npm-16.0.1-2b96ba84be.patch::version=16.0.1&hash=77964d
 - react-dnd-touch-backend@patch:react-dnd-touch-backend@patch%3Areact-dnd-touch-backend@npm%253A16.0.1%23~/.yarn/patches/react-dnd-touch-backend-npm-16.0.1-2b96ba84be.patch%3A%3Aversion=16.0.1&hash=77964d#~/.yarn/patches/react-dnd-touch-backend-patch-0040823149.patch::version=16.0.1&hash=33fdd5
+- react-dnd-touch-backend@patch:react-dnd-touch-backend@patch%3Areact-dnd-touch-backend@patch%253Areact-dnd-touch-backend@npm%25253A16.0.1%2523~/.yarn/patches/react-dnd-touch-backend-npm-16.0.1-2b96ba84be.patch%253A%253Aversion=16.0.1&hash=77964d%23~/.yarn/patches/react-dnd-touch-backend-patch-0040823149.patch%3A%3Aversion=16.0.1&hash=33fdd5#~/.yarn/patches/react-dnd-touch-backend-patch-2c3a2052b6.patch::version=16.0.1&hash=a99236
 ```
 
 Specifying the candidate package involves copying the name of the last one in the list, and then pasting it in single quotes:
 
 ```
-yarn patch -u 'react-dnd-touch-backend@patch:react-dnd-touch-backend@patch%3Areact-dnd-touch-backend@npm%253A16.0.1%23~/.yarn/patches/react-dnd-touch-backend-npm-16.0.1-2b96ba84be.patch%3A%3Aversion=16.0.1&hash=77964d#~/.yarn/patches/react-dnd-touch-backend-patch-0040823149.patch::version=16.0.1&hash=33fdd5'
+yarn patch -u 'react-dnd-touch-backend@patch:react-dnd-touch-backend@patch%3Areact-dnd-touch-backend@patch%253Areact-dnd-touch-backend@npm%25253A16.0.1%2523~/.yarn/patches/react-dnd-touch-backend-npm-16.0.1-2b96ba84be.patch%253A%253Aversion=16.0.1&hash=77964d%23~/.yarn/patches/react-dnd-touch-backend-patch-0040823149.patch%3A%3Aversion=16.0.1&hash=33fdd5#~/.yarn/patches/react-dnd-touch-backend-patch-2c3a2052b6.patch::version=16.0.1&hash=a99236'
 ```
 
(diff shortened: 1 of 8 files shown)


ethan-james:
Unfortunately, I haven't been able to quite get the `handleScroll` patch to where we need it to be. I'll take another crack at it tomorrow and hopefully come up with something. The issue now is that the `scroll` events can come in later than expected, as seen here:

Screenshot 2025-08-01 112106

Maybe logging affects the exact timing, but getting a `scroll` event 200ms after `beginDrag` should activate `allowTouchToScroll(false)` and disable any further scrolling is messing me up. I'll double-check that `allowTouchToScroll` is working as expected, but my hypothesis is that it is working and the events are just coming in much more sporadically than expected.

raineorshine:
Okay, thanks for sticking with it!

ethan-james:
OK, there are 3 pieces to this new `handleScroll` function, which I will explain and then you can let me know how you'd prefer to solve this.

```
public handleScroll = (): void => {
    if (window['scrolling']) {
      window.removeEventListener('scroll', this.handleScroll)
      return
    }
    if (this.timeout) {
	clearTimeout(this.timeout)
	this.timeout = 0
    } else if (this.monitor.isDragging()) {
	this.actions.endDrag()
    }
}
```

1. The `clearTimeout` section in the middle is the base case. If a scroll event comes in during the long press delay window, cancel the timeout and proceed with the scrolling without activating drag-and-drop
2. `this.actions.endDrag()` will catch scroll events that sneak in late, i.e. after the timeout fires and drag-and-drop begins. Unfortunately a lot of them do sneak in late, and the UI is still imperfect, but any stray highlighting will disappear after less than a second when scrolling has already begun. There won't be any functional impact if drag-and-drop and scrolling race in this way.
3. `window['scrolling']` at the top. This global variable is set by `scrollAtEdge` in `initEvents.ts` to make an exception for "real" scrolling that should not result in `this.actions.endDrag()` being called. Attaching it to `window` was the quickest, hackiest way that I could think of to get this working.

So the reason why it needs to look like this is that scroll events will come in late, which means it's impossible to prevent scrolling-while-dragging behavior cleanly with `clearTimeout`. Any scroll events that occur while dragging should cancel the drag, except when `scrollAtEdge` is legitimately causing scrolling.

Trying to remove the `handleScroll` event handler without relying on signals from outside of `react-dnd-touch-backend` didn't seem like it was going to work, but hopefully you have an idea that will be better! If we do need to rely on `scrollAtEdge` sending a signal to `react-dnd-touch-backend`, then I can think of 3 possibilities:

1. global variable like I'm using here
2. custom event dispatched by `scrollAtEdge`
3. signal object passed into `DragAndDropContext` as an option

ethan-james:
And I still need to double-check `allowTouchToScroll` to make sure I'm not missing something obvious that is causing scroll events to come in late. I'll try to do that later this afternoon.

ethan-james:
A couple of updates of little consequence:

1. `scroll` events come in at a rate of one per frame under normal circumstances (https://codesandbox.io/p/sandbox/p2t7ql) but come in much slower (150-200ms) when a scroll is already in progress but `touchmove` is calling `preventDefault`. It would be nice to receive the events quickly in order to cancel the bullet highlight quickly, but I don't think there's a way to use this to prevent `beginDrag` and stop the need to track `scrollAtEdge`.
2. `handleTopMove` is calling `preventDefault` on `touchmove` events once dragging has begun. I think that this mostly explains why things worked better before #3119, and we could use this to replace the `allowTouchToScroll` function that I added. I played with it for a bit, though, and tracking the state of the long press inside `react-dnd-touch-backend` seems to be the same amount of work as tracking the state of the long press within our app code so we don't see any real advantage to doing it inside `TouchBackendImpl`.

I have one more idea involving detecting `touchSlop` that does not exceed the threshold but is subsequently followed by a `scroll` event. That might make it possible to selectively call `this.actions.endDrag` and not need to track `scrollAtEdge`, but I'm not sure yet if there's any way to make it work the way I'm imagining.

PR #3161, opened 2025-08-04
Replace window['scrolling'] with scrollHasBegun flag in TouchBackendImpl

### LINKED ISSUE #3141: [Mobile] Scroll is possible after drag-and-drop activates during a long press in the scroll zone

ethan-james:
Fixes #3141

References #3160

This approach relies on the assumption that, when a scroll is in danger of being initiated, `handleTopMove` always runs during the period when `this.timeout` is pending even if the `scroll` event comes in some time later. It appears to hold up well enough, and does not require a communication channel between `scrollAtEdge` and `react-dnd-touch-backend`.

As best I can tell through my testing, `handleTopMove` will fire before `allowTouchToScroll` starts blocking touch events, but we still seem to get the "slow scroll" behavior that I associate with `touchmove` events that are being blocked.

Screenshot 2025-08-03 192322

1. `handleTopMove` (`touchmove` handler) sets the `scrollHasBegun` flag
2. 1ms later, long press activates, calling `allowTouchToScroll(false)` and logging `disableScroll`
3. `handleScroll` scroll handler is triggered ~250ms later (slow)
4. The drag is ended (`this.actions.endDrag()`) and `intercepting scroll` is logged during the same pass of `handleScroll`

I guess it calls `preventDefault` on all `touchmove` events in subsequent frames, and we still see the slow behavior. I don't fully understand why that would be the case, but at this point I'm confident that I understand iOS Safari scroll behavior about as well as I ever will.


src/device/allowTouchToScroll.ts
```diff
@@ -0,0 +1,24 @@
+/** Prevent touchmove from allowing the page to scroll when a long press is active (#3141). */
```
raineorshine: Let's explain this difference between this and `allowScroll`.

Aside: Could this possibly replace or be combined with `allowScroll`? Separate PR probably.

.yarn/patches/react-dnd-touch-backend-patch-2c3a2052b6.patch
```diff
@@ -0,0 +1,131 @@
...
++++ b/dist/TouchBackendImpl.js
+@@ -38,6 +38,22 @@ export class TouchBackendImpl {
+     get document() {
+         return this.options.document;
+     }
++
++    // Cancel drag-and-drop if a scroll event is received so that scroll doesn't interfere with dragging.
++    // Use an arrow function to bind this event handler to the TouchBackendImpl object so that we can use `this`.
++    handleScroll = (e) => {
++        if (window['scrolling']) {
```
raineorshine: The PR title says that `window['scrolling']` has been replaced, but I still see a couple references. Was this intentional? I don't see anywhere that `window['scrolling']` is set, but maybe I'm missing something.

.yarn/patches/react-dnd-touch-backend-patch-eb6b24281e.patch
```diff
@@ -0,0 +1,58 @@
...
+-                console.info('beginDrag')
+                 this.actions.beginDrag(moveStartSourceIds, {
+                     clientOffset: this._mouseClientOffset,
+                     getSourceClientOffset: this.getSourceClientOffset,
+@@ -387,6 +386,7 @@ export class TouchBackendImpl {
+             return undefined;
+         };
+         this.handleTopMoveEndCapture = (e)=>{
++            this.scrollHasBegun = false
+             this._isScrolling = false;
```
raineorshine: What's the difference between `this.scrollHasBegun` and `this._isScrolling`?

.yarn/patches/react-dnd-touch-backend-patch-2c3a2052b6.patch
```diff
@@ -0,0 +1,131 @@
...
+                 this._isScrolling = true;
+                 return;
+             }
++
+             // If we're not dragging and we've moved a little, that counts as a drag start
+             if (!this.monitor.isDragging() && // eslint-disable-next-line no-prototype-builtins
+             this._mouseClientOffset.hasOwnProperty('x') && moveStartSourceIds && distance(this._mouseClientOffset.x || 0, this._mouseClientOffset.y || 0, clientOffset.x, clientOffset.y) > (this.options.touchSlop ? this.options.touchSlop : 0)) {
+                 this.moveStartSourceIds = undefined;
++                // window.removeEventListener('scroll', this.handleScroll);
++                console.info('beginDrag')
```
raineorshine: I assume this was for debugging and can be removed now.

.yarn/patches/react-dnd-touch-backend-patch-2c3a2052b6.patch
```diff
@@ -0,0 +1,131 @@
...
++    // Use an arrow function to bind this event handler to the TouchBackendImpl object so that we can use `this`.
++    handleScroll = (e) => {
++        if (window['scrolling']) {
++            window.removeEventListener('scroll', this.handleScroll)
++            return
++        }
++        if (this.timeout) {
++            clearTimeout(this.timeout);
++            this.timeout = 0;
++        } else if (this.monitor.isDragging()) {
```
raineorshine: I assume `this.monitor.isDragging()` is just a getter. Any performance risk here?

I know I recommended *not* throttling the scroll callback before. I think that still holds, but just want to check. 

Maybe it's possible to only check to disable dragging when the scroll starts, rather than on every tick?

src/hooks/useDragAndDropThought.tsx
```diff
@@ -109,6 +110,7 @@ const endDrag = () => {
   // Reset the lock variable to allow immediate long press after drag
   longPressStore.unlock()
   globals.longpressing = false
+  allowTouchToScroll(true)
```
raineorshine, line 117: Let's add a comment here to explain why react-dnd's default scroll prevention is not enough.

src/hooks/useDragHold.ts
raineorshine, line 1: Let's add comments to the two additions here to contextualize them within the broader solution. I'd rather have more comments than less, and a bit of redundancy is okay for a complex problem like this.

src/device/allowTouchToScroll.ts
ethan-james: I think it's possible. `allowScroll` seems geared toward preventing scrolling while a modal is open, so I didn't know if they were solutions to different problems. Definitely worth exploring in a separate PR.

.yarn/patches/react-dnd-touch-backend-patch-2c3a2052b6.patch
ethan-james: This ended up being 2 separate patches, mainly because my `/tmp` directory emptied out at some point. It's probably worth the forensic effort to backtrack and combine both patches properly.

.yarn/patches/react-dnd-touch-backend-patch-eb6b24281e.patch
ethan-james: `this._isScrolling` involves `this.options.scrollAngleRanges`, which basically prevents `react-dnd` from activating drag-and-drop when touches are moving in a certain direction. It's pretty much the opposite of what we want, although I should probably understand it a bit better. I'll take a quick look while I'm combining the patches.

.yarn/patches/react-dnd-touch-backend-patch-2c3a2052b6.patch
ethan-james: The final version ended up removing the scroll handler after it executes. I wasn't sure that was going to be possible, but when I combine the patches it will make more sense.

src/device/allowTouchToScroll.ts
raineorshine: I just did some testing of `allowScroll` and discovered it's broken anyway.

Created #3165 and #3166.

raineorshine (approved):
Looks great! Thanks!

PR #3163, opened 2025-08-05
Fix 3084: `[Mobile]: Training alert appears before gesture completes`

### LINKED ISSUE #3084: [Mobile]: Training alert appears before gesture completes
Reproduced in Chrome Device Mode.

## Steps to Reproduce

Quickly swipe →↓ and hold.

## Current Behavior

A "New Thought" alert is shown.

https://github.com/user-attachments/assets/f58959f0-c32b-46e8-9933-1c595c981830

## Expected Behavior

The`experienceMode` alert should be shown after any gesture completes (except forward/back), and then fade out after a delay.

## Testing

Add a puppeteer test that asserts that the alert is:

1. not shown when the gesture is in progress
2. shown after the gesture completes

We should arguably assert that the alert is cleared after the delay, but it is currently [hardcoded](https://github.com/cybersemics/em/blob/a7f899f25b484134dcdc8c0b73493a19432b1b38/src/commands.ts#L220) and I don't think mocking it would add much value.

codebuild-ro:
Fixes #3084

- Update logic to show command alert - now it is showing after gesture is finished and disappears after 5 seconds
- Added puppeteer testing
  - Verifies no alerts during gesture progress
  - Verifies alerts appear after gesture completion
  - Includes visual regression protection
  
### Cause
Actually, it was implemented for that behavior purposely, so it is difficult to consider it as a bug.
In one word, it was implemented for wrong behavior logic.
So I removed original code for that behavior and added similar code to other place which is for gesture ending handler.


src/e2e/puppeteer/__tests__/gesture-alert.ts
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Helper function to smoothly move touch from one point to another.
+ * Creates a series of touchMove events to simulate realistic gesture movement.
+ *
+ * @param x1 - Starting X coordinate.
+ * @param y1 - Starting Y coordinate.
+ * @param x2 - Ending X coordinate.
+ * @param y2 - Ending Y coordinate.
+ */
+const move = async (x1: number, y1: number, x2: number, y2: number): Promise<void> => {
```
raineorshine: This is a great idea, but you can use the existing [swipe](https://github.com/cybersemics/em/blob/9889f4a15288d82c787e2ea5b4e3ef976f8fe4f3/src/e2e/puppeteer/helpers/swipe.ts) helper. You'll just need to extend it to work in multiple directions.

src/e2e/puppeteer/__tests__/gesture-alert.ts
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Helper function to perform an incomplete gesture - Create New Subthought (without touchEnd).
+ * This simulates a gesture that is started but not completed, which should not trigger any alerts.
+ * The gesture follows the pattern: right → down → right (likely matches a valid command).
+ *
+ * Note: No touchEnd() is called, so the gesture remains incomplete.
+ */
+const performIncompleteGesture = async () => {
```
raineorshine: I would prefer an API like `swipe('rdr')`

src/e2e/puppeteer/__tests__/gesture-alert.ts
```diff
@@ -0,0 +1,139 @@
...
+   * This test ensures that when a user starts a gesture but doesn't complete it
+   * (no touchEnd event), no alert should be shown. This prevents alerts from
+   * appearing prematurely and interfering with the gesture interaction.
+   */
+  it('should not show alert during gesture progress', async () => {
+    // Perform an incomplete gesture (no touchEnd)
+    await performIncompleteGesture()
+
+    // Wait a bit to ensure any potential alerts would have time to appear
+    await sleep(1000)
```
raineorshine: We do not allow arbitrary `sleep`, as it slows down the test suite for everyone on the team. You can wait for specific conditions, or `waitForFrames` if just waiting for rendering.

src/e2e/puppeteer/__tests__/__image_snapshots__/gesture-alert/alert-after-gesture-completion.png
raineorshine, line 1: No snapshot tests for this. This is a behavioral bug, not a visual regression.

raineorshine (changes requested):

What was the cause of the bug? That's the most important thing when it comes to bug fixes. Please clearly describe the cause of the bug. That will help me evaluate your proposed solution. Thanks!

codebuild-ro:
> Hi, thanks for your submission.
> 
> What was the cause of the bug? That's the most important thing when it comes to bug fixes. Please clearly describe the cause of the bug. That will help me evaluate your proposed solution. Thanks!

Actually, it was implemented for that behavior purposely, so it is difficult to consider it as a bug.
In one word, it was implemented for wrong behavior logic.
So I removed original code for that behavior and added similar code to other place which is for gesture ending handler.

src/e2e/puppeteer/__tests__/gesture-alert.ts
codebuild-ro: I updated swipe function with using my `move` function.

src/e2e/puppeteer/__tests__/gesture-alert.ts
codebuild-ro: I removed `sleep` waiting function.

src/e2e/puppeteer/__tests__/__image_snapshots__/gesture-alert/alert-after-gesture-completion.png
codebuild-ro: I removed snapshot testing. Just remained behavioral test.

raineorshine:
It's still possible to describe how the previous implementation failed to result in the correct behavior. I don't think it was ever intentional.

src/commands.ts
```diff
@@ -206,30 +205,6 @@ export const inputHandlers = (store: Store<State, any>) => ({
...
-      store.dispatch(
-        // alert the command label if it is a valid gesture
-        alert(command && command?.label, {
-          alertType: AlertType.GestureHint,
-          clearDelay: 5000,
-          showCloseLink: false,
-        }),
-      )
-    }
-
```
codebuild-ro, line 232: This was written in `handleGestureSegment` function which handles ongoing gesture.

JoeDuong-dev:
@codebuild-ro Can you clarify the behavior this fix is addressing?

Update logic to show command alert - now it is showing after gesture is finished and disappears after 5 seconds

I tested on both mobile devices and the web (via DevTools), but the gesture panel does not disappear after 5 seconds as expected. For example, after swiping →↓ and holding, the panel appeared. I kept holding for longer than 5 seconds, and it remained visible.

With this branch, I do notice an improvement, the gesture panel triggers less often, but I still see the issue occasionally. This might be because I’m holding the ↓ gesture slightly too long. How long does the gesture need to be held to trigger the panel?

codebuild-ro:
Hi @JoeDuong-dev 
In this ticket, we need to show the alert after gesture finishes.
Seems you confused the alert as gesture panel.
image

src/e2e/puppeteer/helpers/swipe.ts
```diff
@@ -1,50 +1,111 @@
...
+/**
+ * Helper function to smoothly move touch from one point to another.
+ * Creates a series of touchMove events to simulate realistic gesture movement.
+ *
+ * @param x1 - Starting X coordinate.
+ * @param y1 - Starting Y coordinate.
+ * @param x2 - Ending X coordinate.
+ * @param y2 - Ending Y coordinate.
+ */
+const move = async (x1: number, y1: number, x2: number, y2: number): Promise<void> => {
```
raineorshine, line 13: Looks good

src/e2e/puppeteer/helpers/swipe.ts
```diff
@@ -1,50 +1,111 @@
...
+ * @param gesture - String of directions (e.g., "rd" for right-down).
+ * @param completeGesture - Whether to complete the gesture with touchEnd
+ * Set to false to test during-gesture behavior.
+ * Set to true to test post-gesture behavior.
+ */
+const swipe = async (gesture: string, completeGesture = false) => {
+  const directions = gesture.split('') as Direction[]
+  const screenWidth = await page.evaluate(() => window.screen.width)
+
+  // Calculate total movement in each direction
```
raineorshine: This doesn't work for gestures that wrap back on themselves, like →↓←.

Let's remove this and opt for a simpler approach with a fixed width if possible. We are using a fixed starting point anyway. There's no reason to add additional complexity if it's not needed.

src/e2e/puppeteer/__tests__/gesture-alert.ts
raineorshine, line 1: Test looks good, thanks.

src/e2e/puppeteer/helpers/swipe.ts
codebuild-ro: I simplified the `swipe` function. 
```
const swipe = async (gesture: string, completeGesture = false) => {
  const directions = gesture.split('') as Direction[]

  // Fixed step sizes for consistent gesture behavior
  const stepSize = 80

  // Starting position for the gesture
  let y = 300
  let x = 100

  // Generate points for each direction in the gesture
  const points = [{ x, y }]
  for (const direction of directions) {
    switch (direction) {
      case 'r':
        x += stepSize
        break
      case 'l':
        x -= stepSize
        break
      case 'd':
        y += stepSize
        break
      case 'u':
        y -= stepSize
        break
    }
    points.push({ x, y })
  }

  // Execute the gesture with optional completion
  await swipePoints(points, completeGesture)
}
```

raineorshine (approved):

JoeDuong-dev:
Re-tested and confirmed that this is working as intended. I tested with various gesture combinations, and the alert displayed after each gesture. The panel then dismissed automatically after 5 seconds.

Everything is functioning according to the expected behavior.

Thanks @codebuild-ro for clarifying the confusion.

PR #3164, opened 2025-08-06
Disabled AA(Anti-Aliasing) for flaky snapshots causing pixel differences.

karunkop:
## Root Cause:

After extensive testing, primarily through hit-and-trial, it was determined that visual snapshot flakiness was caused by inconsistent antialiasing across environments, especially after interactions like `ArrowUp` or clicking on a thought. The default screenshot method captured browser-rendered content with antialiasing enabled, which led to:

- Pixel differences due to subpixel rendering
- Inconsistent snapshots between local and CI environments

## Solution:

Introduced a new utility: `screenshot-with-no-antialiasing.ts`, which disables all antialiasing and smoothing effects before capturing screenshots.

**Comprehensive CSS overrides targeting all antialiasing properties:** 
- Font smoothing: -webkit-font-smoothing: none, font-smooth: never
- Text rendering: text-rendering: optimizeSpeed
- Image rendering: image-rendering: pixelated, crisp-edges
- Hardware acceleration: backface-visibility: hidden, perspective: 1000px
- Filter effects: filter: none

Along with this implementation, I have updated all those test suites that involves text heavy rendering snapshots. These changes also allowed me to remove the usage of `waitForFrames` in tests like `render-thoughts`, `url` etc. 

> Note: We can still see the use of `waitForFrames` in `categorize` test as it involves importing/pasting long multiline thoughts which may require several frame cycles to complete before any user interaction especially in CI environment.

The new implementation has significantly improved test stability. You can see the checks results [here](https://github.com/karunkop/em/actions?query=branch%3Aantialiasing)

## Additional Notes:

We can safely close out draft PRs like #3158 , #3126 . This PR resolves issues such as #3088 and #2956


src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drop-hover-after-first-thought-of-column-one-1.png
raineorshine, line 1: z-index

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drop-target-last-child-in-cliff-1.png
raineorshine, line 1: z-index

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drag-drop-target-last-visible-child-1.png
raineorshine, line 1: z-index

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drop-drop-targets-drag-and-drop-thought-and-drop-child-1.png
raineorshine, line 1: z-index

src/e2e/puppeteer/__tests__/__image_snapshots__/drag-and-drop/drop-drop-targets-drop-uncle-1.png
raineorshine, line 1: z-index

src/e2e/puppeteer/helpers/screenshot-with-no-antialiasing.ts
```diff
@@ -0,0 +1,63 @@
+import { ScreenshotOptions } from 'puppeteer'
+import { page } from '../setup'
+
+/** Takes a screenshot with antialiasing disabled. */
+const screenshot = async (options?: ScreenshotOptions) => {
```
raineorshine: Can be refactored in a separate PR, but do you think we could apply this as the default? i.e. Is there ever a reason to take an antialiased screenshot?

src/e2e/puppeteer/helpers/waitForFrames.ts
```diff
@@ -1,23 +1,28 @@
...
  * - CI environments have slower frame rates (20-30 FPS vs local 60 FPS).
  * - Fixed timeouts (e.g., `sleep(200)`) fail to capture final render.
  */
-const waitForFrames = () =>
-  page.evaluate(() => {
-    return new Promise(resolve => {
-      requestAnimationFrame(() => requestAnimationFrame(resolve))
-    })
-  })
+const waitForFrames = (count: number = 1): Promise<void> => {
```
raineorshine, line 16: The name is a bit misleading, since each invocation waits two frames. e.g. `waitForFrames(2)` actually waits 4 frames.

I think it would be better to keep it 1:1 for comprehensibility, and not do it in pairs.

src/e2e/puppeteer/helpers/screenshot-with-no-antialiasing.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Ensure antialiasing is disabled before taking screenshot
+  await page.evaluate(() => {
+    const style = document.createElement('style')
+    style.id = 'screenshot-antialiasing-disable'
```
raineorshine: You can pull out the style element ID into a constant and reuse it below.

src/e2e/puppeteer/__tests__/categorize.ts
```diff
@@ -29,6 +29,9 @@ describe('categorize', () => {
         - Fusce vulputate eleifend sapien. Vestibulum purus quam, scelerisque ut, mollis sed, nonummy id, metus. Nullam accumsan lorem in dui. Cras ultricies mi eu turpis hendrerit fringilla. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia Curae; In ac dui quis mi consectetuer lacinia.
     `)
 
+    // Wait 4 frames due to multiple rounds of requestAnimationFrame in useLayoutAnimationFrameEffect
+    // which requires several frame cycles to complete regardless of content complexity
+    await waitForFrames(4)
```
raineorshine, line 34: As written, this waits 8 frames, correct? Why?

And the old question: Why does this need to be done before `clickThought`? Clicking a thought is not dependent on layout.

The PR descriptions mentions importing as the cause, but that's not mentioned here.

raineorshine (changes requested):
I think you got it! It's interesting that even running the tests in a Docker container doesn't avoid rendering discrepancies from antialiasing. Great find.

Thank you for the good writeup of the cause, solution, and associated issues.

The only functional problem I see is the broken z-indexes in the drag-and-drop screenshots, marked below.

Otherwise this looks like the solution we've been looking for.

src/e2e/puppeteer/helpers/screenshot-with-no-antialiasing.ts
karunkop: Technically, there's no strong justification for taking antialiased snapshots. However, the CSS properties commonly used to disable antialiasing, particularly `backface-visibility` and `perspective`, have been found to interfere with `drag-and-drop` tests. These properties tend to affect `z-index`rendering, which leads to expected but inconsistent snapshot results in those scenarios.

The primary reason for implementing this snapshot utility was to handle text-heavy components, especially those tied to user interactions, which are more susceptible to pixel-level rendering differences. Given this context, I recommend keeping the current setup as is for now. That said, we can continue to explore more reliable alternatives to ensure consistency across all Puppeteer-based tests.

src/e2e/puppeteer/__tests__/categorize.ts
karunkop: > As written, this waits 8 frames, correct? Why?

I've already updated waitForFrames to reflect a 1:1 frame wait, making its behavior more intuitive. Initially, I wasn't aware that each invocation was effectively waiting for two frames, which made the naming misleading. That was an oversight on my part.

> And the old question: Why does this need to be done before `clickThought`? Clicking a thought is not dependent on layout.
> 
> The PR descriptions mentions importing as the cause, but that's not mentioned here.

The only thing I can confirm at this point is that, without `waitForFrames` immediately following a paste operation, particularly when the pasted content is long and spans multiple lines, the entire content sometimes shifts downward, as illustrated in the attached screenshot.

image

You had previously pointed out that the [scrollCursorIntoView](https://github.com/cybersemics/em/blob/06c1657076f5aa455a769760cdcb7308ddc7ac09/src/redux-middleware/scrollCursorIntoView.ts) middleware might introduce asynchronous scroll behavior. Specifically, the section between [lines 93–131](https://github.com/cybersemics/em/blob/main/src/redux-middleware/scrollCursorIntoView.ts#L93-L131) contains layout animations that can affect element positioning. These animations appear to require a few frames to complete before the DOM reaches a visually stable state.

While I can't say with full certainty that this is the root cause, adding a delay via `waitForFrames` does seem to mitigate the issue. I’ll continue investigating this more thoroughly, but for now, I don’t have a definitive explanation.

src/e2e/puppeteer/helpers/screenshot-with-no-antialiasing.ts
raineorshine: Thanks, sounds good.

Created #3168.

src/e2e/puppeteer/__tests__/categorize.ts
raineorshine: Thanks. Everything else looks good with this PR, so will merge and create a new issue for investigating the justification for `waitForFrames`.

Created #3169

raineorshine (approved):
Thank you!!!

PR #3167, opened 2025-08-07
Add touchAction: none to dialog recipe overlay

### LINKED ISSUE #3165: [Mobile Safari] Document body is scrollable when cheatsheet is open

Related: #2919

## Steps to Reproduce

1. Scroll up and down a bit.

## Current Behavior

Document body scrolls.

## Expected Behavior

ethan-james:
Fixes #3165

Tested on iOS Safari, iOS PWA and Android Chrome

Setting `overflow: hidden` on the `body` element would block many attempts at scrolling, but did not work in all cases. More assertive efforts such as `position: fixed`, or `overflow: hidden` on the `html` element, would work but would force it to scroll to the top.

Fortunately, iOS Safari seems to fully respect `touch-action: none`. Setting this on the overlay element in `dialogRecipe` prevents any unwanted scrolling while the modal is visible. I think this will be useful in all future dialogs, but in any case `dialogRecipe` is new and only used by `GestureCheatsheet`, so the impact of this change is limited.

I was able to remove `allowScroll` from `GestureCheatsheet`, and also from `CommandPalette` as described in https://github.com/cybersemics/em/issues/3165#issuecomment-3161974221. `allowScroll` is currently unused, and we could probably get rid of it entirely. I didn't go so far as to do that just yet, but please let me know if you think I should.


raineorshine (approved):
Amazing!

I'll hand this over to @JoeDuong-dev for testing.

Please remove `allowScroll` completely if it's unused.

JoeDuong-dev (approved):

ethan-james:
It seems that the body scrolls when you start your touch inside the scrollable content pane and then drift out onto the dialog overlay. It's easiest for me to get it to happen when I start in the bottom left corner. Adding `allowScroll` back doesn't seem to help, and I found the place where [the Dialog component tries to block touchmove events](https://github.com/ethan-james/em/blob/1f291e845f0a2530d5da74162530238feb5ce23b/src/components/dialog/Dialog.tsx#L49) in the manner that I was using for `allowTouchToScroll`.

This event handler is checking `e.target` to see if it should prevent `touchmove` events, which is a problem because `e.target` stays the same throughout the entire touch. If you start on one element and then move over to another, you still get the original element as the target.

I tried to fix it so that it checks the element under the current touch position:

```
    /** This event handler prevents touch events from propagating to the page. */
    const preventTouchMove = (e: TouchEvent) => {
      const hoverElement = document.elementFromPoint(e.touches[0].clientX, e.touches[0].clientY)

      if (hoverElement === overlayElement) {
        // Only prevent scrolling if NOT inside dialog content
        e.preventDefault()
      }
    }
```

which fixes the following case:

1. start the touch at the position indicated by the red X in the previous comment
2. drag your finger straight out to the right edge
3. drag your finger down

but not this similar case:

1. start the touch at the position indicated by the red X in the previous comment
2. drag your finger diagonally down and right to the edge

I think we may be looking at a variant of the "once a scroll has begun it can't be stopped" issue that was causing problems on #3141. I'm not yet sure how to fix it, but I can try some more over the weekend.

ethan-james:
I created a new PR (#3171) with a different approach that prevents the body from scrolling at all while the dialog is open.

ethan-james:
> When I was dealing with the scroll stop problem on the MultiGesture years ago, I wanted to stop a scroll that had already begun on the body, and there was no way for me to know ahead of time when that scroll would occur (because scrolling used to occur in the gesture zone). However in this case, we know _before_ the user begins scrolling that body scrolling should be disallowed. Maybe I'm not seeing the similarity, but they seem like separate problems. We're looking for a way to disable body scrolling only.

I'm not sure because the internals are opaque, but it sure seems to me like it starts scrolling in the content pane and then decides to continue that scroll in the viewport (body) without respecting the `touchmove` event handler that tries to stop it.

Issue #3069, opened 2025-06-26
Smarter sort-as-you-type for dates


raineorshine:
## Steps to Reproduce

```
- Journal
  - =sort
    - Alphabetical
      - Desc
  - 6/21
  - 6/18
  - 6/1
  - 5/10
```

1. Set the cursor on `Journal`. 
2. Activate New Subthought Above
  - Ctrl + Shift + Enter (Windows) or ⌘ + ⇧ + Enter (macOS)

Case 1: Edit the new thought to `6`
Case 2: Edit the new thought to `5`

## Current Behavior

`6` gets sorted to the bottom of the list. 

## Expected Behavior

- `6` should stay at the top of  the list. 
- `6/` should stay at the top of  the list.
- `6/2` should stay at the top of  the list **until the cursor changes**. Then it should be sorted between `6/1` and `6/18`. 
- `5` should be sorted between `5/10` and `6/1`. 

In a descending list of dates, we should assume that the new thought is going to be the newest date (e.g. `6/22`). This fits the natural user behavior of adding new entries to a reverse chronological journal. We should not sort the thought based on the intermediate values as the user types (as is done with alphabetically sorted non-dates).

`6` and `6/` should always be sorted above `6/1` in a descending sorted list of dates.

codebuild-ro:
```
- Journal
  - =sort
    - Alphabetical
      - Desc
  - 6/21
  - 6/15
```
What is the expected behavior if use types `6/1` and hold?
I mean should it be between `6/21` and `6/15`?

PR #3170, opened 2025-08-08
Fix 3069: `Smarter sort-as-you-type for dates`

codebuild-ro:
Fixes https://github.com/cybersemics/em/issues/3069

### Cause

It was because every thought change does calculate ranking of that thought.

### Solution

- During editing: Thoughts stay in their current position while being edited, preventing disruptive re-sorting
- After losing focus: Thoughts are automatically re-sorted into their correct alphabetical position
- Works for date thought in a sorted context, not for non-date patterns


src/redux-middleware/scrollCursorIntoView.ts
raineorshine, line 1: It would be better to put the new logic into its own middleware than reuse `scrollCursorIntoView`, which is unrelated.

src/redux-middleware/scrollCursorIntoView.ts
```diff
@@ -172,6 +177,43 @@ const scrollCursorIntoViewMiddleware: ThunkMiddleware<State> = ({ getState }) =>
...
+      if (previousThought && isDatePattern(previousThought.value) && wasBeingEdited) {
+        const parentId = previousThought.parentId
+        const sortPreference = getSortPreference(stateNew, parentId)
+        const simplePath = simplifyPath(stateNew, cursorOld)
+
+        if (sortPreference.type === 'Alphabetical') {
+          // Re-sort the date thought by calling editThought with forceSorting: true
+          // This triggers getSortedRank with the cursor no longer on the thought
+          // The date thought will be sorted into its correct position based on its value
+          requestAnimationFrame(() => {
```
raineorshine: Why rAF?

src/redux-middleware/scrollCursorIntoView.ts
```diff
@@ -172,6 +177,43 @@ const scrollCursorIntoViewMiddleware: ThunkMiddleware<State> = ({ getState }) =>
...
+
+    // Check if we need to re-sort date thoughts when cursor changes
+    // This handles the case where a date thought was being edited and the cursor moves away
+    // Only date patterns (like "6/21", "12/1") get re-sorted after losing focus
+    if (cursorChanged && cursorOld) {
+      const previousThoughtId = head(cursorOld)
+      const previousThought = getThoughtById(stateNew, previousThoughtId)
+
+      // Check if the previous thought was actually being edited (not just cursor on it)
+      const wasBeingEdited = stateOld.cursor && equalPath(stateOld.cursor, cursorOld)
```
raineorshine: `cursorOld` seems to always be equal to `stateOld.cursor`. Am I missing something?

src/selectors/getSortedRank.ts
```diff
@@ -30,6 +41,16 @@ const getSortedRank = (state: State, id: ThoughtId, value: string) => {
...
 
+  let editedThought: Thought | undefined = undefined
+  // Find the thought that matches the value being sorted
+  // If multiple thoughts have the same value, use the most recently updated one
+  // This handles edge cases where duplicate values might exist
+  for (const thought of thoughts) {
+    if (thought.value === value) {
+      editedThought = (editedThought ? editedThought.lastUpdated : 0) > thought.lastUpdated ? editedThought : thought
+    }
+  }
```
raineorshine: This style of coding doesn't fit the project very well. As you know, we follow a functional programming paradigm. Please avoid mutations and for loops when there are purer approaches.

src/util/isDatePattern.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Checks if a string matches a date pattern like "M/d" or "M-d"
+ * Accepts 1-2 digits for month and day, separated by "/" or "-"
+ * Examples: "6/21", "6-21", "12/1", "12-1".
+ */
+const isDatePattern = (value: string): boolean => {
```
raineorshine: Let's use the existing [parseDate](https://github.com/cybersemics/em/blob/0b18b89b7acaf33d144ce3deee4fec7549173fac/src/util/compareThought.ts#L55) function so that we don't have multiple date detection algorithms in the codebase.

src/selectors/getSortedRank.ts
```diff
@@ -7,7 +8,7 @@ import getSortPreference from './getSortPreference'
 import thoughtToPath from './thoughtToPath'
 
 /** Calculates the rank for a given index in a sorted array of thoughts. */
-const calculateRank = (thoughts: { rank: number }[], index: number): number => {
+const calculateRank = (thoughts: Thought[], index: number, id?: string): number => {
```
raineorshine: I don't love the new API of this function. That third parameter is very awkward. I think this needs some work, or needs to be addressed higher up.

src/actions/editThought.ts
```diff
@@ -137,12 +144,19 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
...
     ...(editedThought.generating ? { generating: false } : null),
     rank:
-      newValue !== '' && (sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated')
+      newValue !== '' &&
+      (sortType === 'Alphabetical' || sortType === 'Created' || sortType === 'Updated') &&
+      // For non-date thoughts: always sort normally
+      // For date thoughts: only sort when cursor is not on the thought AND forceSorting is true
+      (!isDatePattern(newValue) ||
+        (editingThoughtId !== editedThoughtId && // Only sort when cursor is not on this thought
+          forceSorting)) // Only sort when explicitly requested (e.g., after losing focus)
```
raineorshine: Good, this is the right place to bypass sort on edit.

raineorshine (changes requested):

Handling the updated sort in Redux middleware seems reasonable, though you probably don't want to use unrelated middleware for this purpose.

The desired behavior is slightly different. The thought rank should be stable as the user types, regardless of how fast they type. Your algorithm only keeps the rank stable if "6/2" is typed very fast. It should remain stable if is edited in multiple steps: "6" → "6/" → "6/2".

Instead of checking if the edited thought is a date, you should probably check if the rest of the list consists of only dates. That way you can positively identify the condition for the special sorting behavior even when the new thought doesn't look like a date yet. Just keep efficiency in mind.

src/selectors/getSortedRank.ts
codebuild-ro: I rolled back this part.

src/util/isDatePattern.ts
codebuild-ro: That function is removed.
There is new `isDatePttern` in codebase of branch for fixing 3070.
I used that.

src/selectors/getSortedRank.ts
codebuild-ro: I rolled back this part too.

src/redux-middleware/scrollCursorIntoView.ts
codebuild-ro: This part is rolled back.

src/redux-middleware/scrollCursorIntoView.ts
codebuild-ro: I rolled back the code in that file.

codebuild-ro: I rolled back the code in that file.

JoeDuong-dev:
Thanks @codebuild-ro. I’ve tested the changes, and the previous bug no longer occurs. Using the 10/30 - 10/20 - 10/10 test case, the ordering is correct and updates immediately.

However, I noticed an issue when testing with a different date format. In this scenario, I used 10/31/15 - 10/20/23 - 10/10/11. The expected order was 10/20/23 > 10/31/15 > 10/10/11, but instead 10/31/15 appeared above 10/20/23.
Screenshot 2025-08-20 at 3 42 32 PM


@raineorshine Regarding the keyboard shortcut: Yes, I meant creating a new sub-thought above using ⌘ + ⇧ + Enter (macOS). I know ⌘ + Enter creates a new sub-thought, but it seems ⌘ + ⇧ + Enter was not working as intended and was creating a regular sub-thought instead of one above. This issue isn’t related to this ticket. I’ll test again on both Windows and Mac keyboards to see if I can reproduce the bug.

src/components/Thought.tsx
```diff
@@ -518,6 +521,28 @@ const ThoughtContainer = ({
...
 
+  useEffect(() => {
+    const prevValue = prevValueRef.current || ''
+    const isFocusLostAndUpdated = !isEditing && isEditing !== prevEditingRef.current && prevValue !== value
+    if (!isEditing) prevValueRef.current = value
+
+    if (isFocusLostAndUpdated && value) {
+      // Dispatch editThought when other thought gets focus
+      dispatch(
+        editThought({
```
raineorshine: I feel like `moveThought` makes more sense than `editThought` here. `editThought` is for changing the value of a thought, while `moveThought` is specifically for changing the context or rank.

src/components/Thought.tsx
```diff
@@ -518,6 +521,28 @@ const ThoughtContainer = ({
     isTableCol1,
   })
 
+  useEffect(() => {
```
raineorshine: Let's extract this into a custom hook along with the two refs in order to avoid adding clutter to the component.

src/actions/editThought.ts
```diff
@@ -137,14 +143,19 @@ const editThought = (state: State, { cursorOffset, force, oldValue, newValue, pa
   const isNote = parentOfEditedThought.value === '=note'
   const sortPreference = getSortPreference(state, editedThought.parentId)
   const sortType = sortPreference.type
+  const isSortable =
+    newValue !== '' &&
+    // When sortType is Alphabetical, the thought list should be sorted for below cases:
+    // - the cursor is on a different thought AND list involves a date
+    // - thought list doesn't involve any date
+    ((((isDateInvolved && editingThoughtId !== editedThoughtId) || !isDateInvolved) && sortType === 'Alphabetical') ||
```
raineorshine: You can probably simplify this expression so that `isDateInvolved` is not repeated.

src/actions/editThought.ts
```diff
@@ -42,18 +43,23 @@ export interface editThoughtPayload {
 
 /** Changes the text of an existing thought. */
 const editThought = (state: State, { cursorOffset, force, oldValue, newValue, path }: editThoughtPayload) => {
-  if (oldValue === newValue || isDivider(oldValue)) return state
+  const editingThoughtId = state.cursor && head(state.cursor)
+  const editedThoughtId = head(path)
+  const editedThought = getThoughtById(state, editedThoughtId)
+  const siblings = editedThought?.parentId ? getAllChildrenSorted(state, editedThought.parentId) : []
+  const isDateInvolved = siblings.some(thought => isDatePattern(thought.value))
```
raineorshine: I'm a bit wary of calling `isDatePattern` at O(n) on every edit. Can you consider if there's a more performant approach?

We could also change `some` to `every` which would short circuit much earlier.

src/components/Thought.tsx
codebuild-ro: You are right! I used `moveThought` instead of `editThought` now.

src/actions/editThought.ts
codebuild-ro: In `editThought`, I removed the sort func for `date-involved` thought list.
The sorting of `date-involved` thought list will be handled in `moveThought`.

```
// When sortType is Alphabetical, the thought list should be sorted if it doesn't involve any date
((!isDateInvolved && sortType === 'Alphabetical') || sortType === 'Created' || sortType === 'Updated')
```

src/actions/editThought.ts
codebuild-ro: `some` is to check if thought list has at least 1 date sub-thought, though there is non-date sub-thought.
`every` seems not necessary for this checking.

src/components/Thought.tsx
codebuild-ro: I made `useThoughtFocus.ts` file.

https://github.com/codebuild-ro/em-fork/blob/8c095ed92f849cee6833851fe84ee63eaa3fef9e/src/hooks/useThoughtFocus.ts#L10-L28

src/actions/editThought.ts
raineorshine: `some` will loop through the entire list if there are no dates. It would be preferable to short circuit at the first non-date.

codebuild-ro: For now, currrent approach is checking if there is at least 1 date thought.
So do you mean, we need to apply the logic to only the thought list which is composed by only date thoughts?

raineorshine: Yes, I suggest making it short circuit after the first non-date. We need to do something to cut down the running time since this is run on every sibling on every edit.

The way it is written currently is O(n) on most lists, which is close to the worst case scenario.

codebuild-ro: Ok, I got the sense of the point.
I updated that part. Thank you!

raineorshine (commented):
Thanks for your work on this. After some additional consideration of the outcome implementation, I've decided to go a different direction on this. I think a temporary in-memory rank that overrides the real rank might be a better fit, as it won't interfere with undo/redo or be persisted to the db.

PR #3171, opened 2025-08-09
Hide overflow and offset scroll while dialog is open

### LINKED ISSUE #3165: [Mobile Safari] Document body is scrollable when cheatsheet is open

ethan-james:
Fixes #3165

Since it seems to be impossible to stop a scroll once it has begun, the only solution to stop the body from scrolling is to convince iOS Safari that there is nothing to scroll. Applying `overflow: hidden` to both `body` and `html` elements achieves this, but then the content scrolls back to the top because the browser really is convinced that it cannot have a scroll offset.

Applying a negative margin to the content helps to mimic the scrolled position while `allowScroll(false)` is active. There is a slight flicker when restoring the scroll position after the dialog is closed, but other than that it seems to be working well.

## Steps to Reproduce

Image

1. Scroll down in the thoughtspace so that the top thought is obscured (or more)
3. Start your drag from the position indicated above, or the bottom left corner of the dialog
4. Drag diagonally downward toward the edge of the screen; ideally the inner pane should start scrolling before your finger leaves it

## Current Behavior

Document body scrolls.

## Expected Behavior


ethan-james:
I think the flicker is better if I switch from `requestAnimationFrame` to `queueMicrotask`.

raineorshine:
Thank you! Could you post a video of the flicker? That will help me assess the severity.

ethan-james:
> Thank you! Could you post a video of the flicker? That will help me assess the severity.

https://github.com/user-attachments/assets/b4ef212c-3228-4818-9c88-8a98916e8488

ethan-james:
> Thank you! Could you post a video of the flicker? That will help me assess the severity.

I can't really see it at all in this video, which might be in part due to the frame rate, but the flicker is often very minor.

src/device/allowScroll.ts
```diff
@@ -1,14 +1,23 @@
...
+  document.documentElement.style.setProperty('--scroll-offset', `-${scrollY}px`)
+  document.documentElement.setAttribute('data-dialog-open', 'true')
 }
 
-/** Re-enables scrolling on the body element by removing overflow:hidden. */
+/** Re-enables scrolling on the body element by removing overflow:hidden and restoring the scroll position. */
 const enableScroll = () => {
-  document.body.style.overflow = ''
+  document.documentElement.style.removeProperty('--scroll-offset')
+  document.documentElement.removeAttribute('data-dialog-open')
```
raineorshine: I'm curious about the use of `data-dialog-open`. If we're already manually accessing the DOM and updating styles, why not set `marginTop` directly? That would allow the logic to be encapsulated.

Either an all-panda approach or a no-panda approach might make more sense than a mixed approach, unless I'm missing something.

Thanks!

ethan-james: I think this is about as close as I can get to an all-Panda approach, unless you have any suggestions. Setting the margin manually seemed pretty awkward, but now that I've got the `#root > :first-child` selector worked out, it would be straightforward I think.

ethan-james:
I realized that the thoughts are virtualized, so there's a limit to how far you can offset them when the browser believes it is at scroll position `0, 0`. In this case, the limit is "Montana", and if you scroll down farther than that, you won't see anything below the bottom edge of the gesture cheatsheet. I can't think of a way to fix this that doesn't hurt performance, so can I argue that it's a cosmetic issue?

src/device/allowScroll.ts
raineorshine: Yeah, the dynamic `marginTop` makes it hard to do in a purely Panda approach.

raineorshine:
Okay, thanks. That's fine as long as the same number of thoughts are virtualized regardless of whether the dialog is open or closed.

ethan-james:
> Okay, thanks. That's fine as long as the same number of thoughts are virtualized regardless of whether the dialog is open or closed.

Yes, I see that while scrolling, new thoughts are added and removed from the bottom but not from the top. So if they have scrolled all the way to the bottom, all thoughts are rendered when the dialog is closed or open.

raineorshine:
> > Okay, thanks. That's fine as long as the same number of thoughts are virtualized regardless of whether the dialog is open or closed.
> 
> Yes, I see that while scrolling, new thoughts are added to and removed from the bottom but not from the top. So if they have scrolled all the way to the bottom, all thoughts are rendered when the dialog is closed or open.

Thanks!

The thoughts above the screen need to be virtualized eventually, but that's another matter...

raineorshine (approved):

raineorshine:
@JoeDuong-dev This is ready for a final round of testing. The code is approved.

With this one you really have to test it thoroughly, touching as many parts of the screen as possible. See descriptions above for different edge cases. Thanks!

ethan-james:
I see some unusual new behavior when swiping left-to-right:

https://github.com/user-attachments/assets/c9f9805b-e4e8-4255-ac2e-78fb8a67d2b9

Unusual because it seems to have been introduced by the commit 4e64f83de9fbaadb5cd0a77bf8b828773bd940c1 where I merged `main` into the branch, but doesn't appear to show up in `main` itself.

In better news, though, I certainly don't see any background scrolling while the dialog is open.

raineorshine:
> I pushed a fix to contain overscroll within the content pane, which hopefully will prevent [scroll chaining](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior) from scrolling the `#content` div or the body. I have to leave for the afternoon so I can't test rigorously until later, but you can have a look if you want.

Aha! I couldn't remember what that property was called, but that seems important. I find the auto behavior very annoying.

> I see some unusual new behavior when swiping left-to-right:

Oops, that's not good.

ethan-james:
> Oops, that's not good. Let me know if you are able to get to the bottom of it.

OK! I was hoping you would say that it was some new intentional behavior, but I will look into it.

ethan-james:
> Oops, that's not good. Let me know if you are able to get to the bottom of it.

I put back the code that was blocking touch events on the overlay, and that seems to have fixed it.

JoeDuong-dev:
@ethan-james 
I re-tested across all my devices (iPhone, iPad, and DevTools) on both Chrome and Safari, and scrolling is now correctly disabled whenever the cheat sheet is open. I also tested and verified edge cases such as: Multi-finger gestures, Momentum/inertia scrolling while opening the cheat sheet, Pinch-to-zoom, Switching between apps and returning (app resume state) and Orientation changes.

The only issue I found is with Orientation changes where I was still able to scroll while the cheatsheet is opened. 

Steps to reproduce (on iPhone 15 Pro Max – Safari browser):
	1.	In vertical mode, scroll until the scrollbar appears.
	2.	Open the cheat sheet with the gesture.
	3.	Rotate the phone to landscape.
	4.	Rotate back to vertical.
	5.	Attempt to scroll while the cheat sheet is still open.

This is the only scenario where the background remains scrollable. I’ve gone through ~20 test cases overall, so we’re very close. Please let me know if you’re able to reproduce this issue on your side or if you’d like me to provide a video.

cc @raineorshine

JoeDuong-dev:
@ethan-james Here is a video of the bug, with narration. I was only able to reproduce this bug on Safari, not on Chrome. 

https://github.com/user-attachments/assets/d5caaf2b-c7d2-48f2-834b-b98d362079c5

raineorshine:
This will be closed if #3197 gets merged.

Issue #2959, opened 2025-05-18
Flaky test: After categorize, caret should be on new thought


raineorshine:
Test failures:

- https://github.com/cybersemics/em/actions/runs/14956001275
- https://github.com/cybersemics/em/actions/runs/14862444452
- https://github.com/cybersemics/em/actions/runs/14797208786
- https://github.com/cybersemics/em/actions/runs/15524784518/attempts/1?pr=2944
- https://github.com/cybersemics/em/actions/runs/15530792257/attempts/1?pr=2920

`src/e2e/puppeteer/__tests__/caret.ts > mobile only > After categorize, the caret should be on the new thought: 
AssertionError: expected 'b' to be '' // Object.is equality`

https://github.com/cybersemics/em/blob/93617b7e0af854ac1e62a496a086d31dde67c9bd/src/e2e/puppeteer/__tests__/caret.ts#L231-L252

Similar:

- https://github.com/cybersemics/em/actions/runs/15532708535/attempts/1?pr=2944

`src/e2e/puppeteer/__tests__/caret.ts > all platforms > backspace on empty thought should move caret to the end of the previous thought:
AssertionError: expected '' to be 'first' // Object.is equality`

raineorshine:
Temporarily disabled the test in ae5e48d499 because it's making our CI very unreliable.

karunkop:
Fixed in #3172

PR #3172, opened 2025-08-10
Fix for flaky caret tests

### LINKED ISSUE #2954: Flaky test: caret - backspace on empty thought
Test failure: https://github.com/cybersemics/em/actions/runs/15088412461

`
[src/e2e/puppeteer/__tests__/caret.ts > all platforms > backspace on empty thought should move caret to the end of the previous thought: src/e2e/puppeteer/__tests__/caret.ts#L206](https://github.com/cybersemics/em/pull/2952/files#annotation_34806150609)
AssertionError: expected +0 to be 1 // Object.is equality
`

https://github.com/cybersemics/em/blob/93617b7e0af854ac1e62a496a086d31dde67c9bd/src/e2e/puppeteer/__tests__/caret.ts#L186-L208

karunkop:
Close #2954, #2959. 

### Problem

#2959: I think here test appears to be causing temporary mismatches when tests read the window because selection is set asynchronously after Redux updates in complex flows (like categorize), particularly on mobile emulation and CI. Selection is deferred on mobile Safari using `requestAnimationFrame` in https://github.com/cybersemics/em/blob/main/src/components/Editable/useEditMode.ts#L88-L92, generating a timing window following Redux updates.

#2954:  Here multiple async focus flow like: `arrowUp` moves cursor to prev thought → Enter creates an empty sibling → `Backspace` merges empty thought into previous → updates Redux cursor/offset → DOM selection is set later by an effect.
The test reads `window.getSelection()` immediately after `Backspace`, often before the effect has run.

### Fix:

Implemented `waitForEditingState` which waits until for both state and DOM to be in sync eliminating the race condition typically in CI environment

### Results:

[Here](https://github.com/karunkop/em/actions?query=branch%3Aflaky-caret-tests) is the success rate after the changes


src/e2e/puppeteer/helpers/waitForEditingState.ts
```diff
@@ -0,0 +1,34 @@
...
+    (value: string, offset?: number) => {
+      if (!em) return false
+      const el = document.querySelector('[data-editing=true] [data-editable]') as HTMLElement | null
+      if (!el) return false
+
+      const state = em.testHelpers.getState()
+      const cursor = state.cursor
+      if (!cursor) return false
+
+      const thought = em.getThoughtById(em.head(cursor))
```
raineorshine: The use of `em.testHelpers.getState` is tightly coupling the test to various parts of the Redux state (implementation details), which we really want to avoid. It's important that integration tests behave like a normal user and do not have access to what is "under the hood." 

The few times we add a backdoor in existing tests are as last resorts, when there is no other way to test something. Now that you are a dedicated test engineer and we have other people working on the test suite, we need to maintain high standards and work hard to promote separation of concerns and maintainability.

karunkop: Hi, sure thing! I have gotten rid of the dependency with the Redux state and waited for specific selection API nodes and offsets. This helps ensure that the DOM state for editing content and caret offsets meet expected conditions. It waits for UI updates before assertions, reducing false negatives that are observed in CI.
As for [retry.ts](https://github.com/cybersemics/em/blob/abf129c6fb07948345274af85efb31f07bbbfd80/src/e2e/puppeteer/helpers/retry.ts) which is added newly, is essential to prevent timeout issues of the test.
image
What I believe is happening for the timeout case is that the tap on the `Categorize` button sometimes doesn’t register under mobile emulation. When that happens, the action never fires, no new empty thought is created, and `waitForSelectionNode('', 0)` times out.

While this may seem a little bit hacky, we could consider completely ditching the selector method to wait and click the toolbar button and instead implement a helper that allows gesture support for mobile emulation? If what I believe is the reason for the timeout, that implementation should technically resolve the timeout issue, which was also seen before this recent implementation of the retry mechanism. I can see that we already have a [helper](https://github.com/cybersemics/em/blob/main/src/e2e/iOS/helpers/gesture.ts) that supports gesture drawing, but it looks like it is designed specifically for iOS? Also, all tests that make use of this helper seem to be skipped, so it would be a good thing to discuss this beforehand before implementing.

raineorshine: I'm not willing to slow down the tests with retries and delays. You'll have to get to the bottom of why the categorize button does not work.

raineorshine: > While this may seem a little bit hacky, we could consider completely ditching the selector method to wait and click the toolbar button and instead implement a helper that allows gesture support for mobile emulation? If what I believe is the reason for the timeout, that implementation should technically resolve the timeout issue, which was also seen before this recent implementation of the retry mechanism. I can see that we already have a [helper](https://github.com/cybersemics/em/blob/main/src/e2e/iOS/helpers/gesture.ts?rgh-link-date=2025-08-14T08%3A47%3A39Z) that supports gesture drawing, but it looks like it is designed specifically for iOS? Also, all tests that make use of this helper seem to be skipped, so it would be a good thing to discuss this beforehand before implementing.

Either toolbar or gesture input is acceptable, since it's not critical to the test. There's also something to be said for decoupling the tests from the specific keyboard or gesture shortcuts associated with a command.

The helper for puppeteer tests is: https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/helpers/swipe.ts

karunkop: The latest updates have been completed. I've removed all retries and delays from the tests. Instead, they now wait for the appropriate conditions before proceeding with assertions, making the behavior more robust.
Additionally, I made some adjustments to the `swipe.ts` helper to support all gesture types. This was necessary because the initial implementation wasn’t working as expected for the `categorize` gesture. The changes ensure consistent behavior across different gesture interactions.

src/e2e/puppeteer/helpers/swipe.ts
```diff
@@ -34,8 +40,13 @@ const swipePoints = async (points: { x: number; y: number }[], complete: boolean
 
   await page.touchscreen.touchStart(start.x, start.y)
 
+  // Small delay to ensure touchStart is processed
+  await sleep(10)
```
raineorshine: Let's zoom in and try to understand these sleeps better. Why 10ms? What do you mean by "processed"? Is the delay resulting from the MultiGesture component, or something else?

Why does only the categorize gesture break? You may want to consider a narrower solution for the other gestures that work without the delay. I appreciate having a single general solution, but not at the cost of slowing down the test suite.

src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -179,32 +180,39 @@ describe('all platforms', () => {
...
+    await press('Backspace')
 
-    // offset at the end of the thought is value.length for TEXT_NODE and 1 for ELEMENT_NODE
-    const focusNodeType = await getSelection().focusNode?.nodeType
-    expect(offset).toBe(focusNodeType === Node.TEXT_NODE ? 'first'.length : 1)
+    // this is necessary because sometimes in CI, the caret is not immediately moved to the end of the previous thought
+    // without this, the test will intermittently fail in CI
+    await waitUntil(() => {
+      const selection = window.getSelection()
+      return selection && selection.focusOffset > 0 && selection.focusNode?.textContent === 'first'
```
raineorshine: Rather than test `focusOffset > 0`, can you assert the exact offset? 

Then you wouldn't have to take the extra step of typing 'x', which is a rather indirect way of testing the focus offset.

src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -242,26 +250,23 @@ it('clicking backspace when the caret is at the end of a thought should delete a
...
     - a
       - b`
 
     await paste(importText)
 
-    await clickThought('b')
-    await clickThought('b')
-
-    // close keyboard
-    await clickBullet('b')
```
raineorshine: The test still needs to close the keyboard to have the correct coverage. Otherwise, the keyboard stays open and it no longer tests that the selection is restored correctly by the categorize command.

You could create a helper to click on empty space.

src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -1,21 +1,22 @@
...
-import testIfNotCI from '../helpers/testIfNotCI'
+import swipe from '../helpers/swipe'
 import waitForEditable from '../helpers/waitForEditable'
 import waitForHiddenEditable from '../helpers/waitForHiddenEditable'
-import waitForSelector from '../helpers/waitForSelector'
 import waitForThoughtExistInDb from '../helpers/waitForThoughtExistInDb'
 import waitUntil from '../helpers/waitUntil'
 
-vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+vi.setConfig({ testTimeout: 60000, hookTimeout: 20000 })
```
raineorshine: Is increasing the timeout necessary, or just for debugging? I'd hope the test can complete in less than 20 seconds, otherwise our test suite is going to run very slowly.

raineorshine (changes requested):
Thanks for the update.

Once you make the requested changes, could you confirm a batch of 10 passing CS runs so ensure the new solution isn't flaky? Thanks!

src/e2e/puppeteer/__tests__/caret.ts
karunkop: Yes, correct. It was originally implemented for debugging purposes. During earlier troubleshooting efforts involving multiple solution attempts, the tests were frequently timing out in the CI environment, which led me to initially increase the timeout duration. I have reverted back to the original timeout duration.

src/e2e/puppeteer/helpers/swipe.ts
karunkop: > Let's zoom in and try to understand these sleeps better. Why 10ms? What do you mean by "processed"?
> 
I had introduced a delay under the assumption that a pause after `touchstart` might be necessary, but this proved to be incorrect. I inadvertently left the delay in and forgot to remove it in later revisions. Thank you for bringing this to my attention.

>  Is the delay resulting from the MultiGesture component, or something else? Why does only the categorize gesture break? You may want to consider a narrower solution for the other gestures that work without the delay. I appreciate having a single general solution, but not at the cost of slowing down the test suite.

The issue was not related to timing but rather stemmed from incorrect initial x,y coordinates and inaccurate calculation of coordinates during the `touchMove` event. This resulted in the `categorize` gesture not being recognized correctly. The problem also affected other gestures that rely on leftward swipes, including `outdent`, `archive` etc.

raineorshine (approved):
Looks good, thanks!

This is ready to merge, just please confirm a batch of 10 passing CI runs so ensure the new solution solves the flakiness.

Issue #3070, opened 2025-06-26
Sort non-dates after dates


raineorshine:
In a sorted list of dates, sort `6` and `6/` after `6/1`. 

i.e. Sort non-dates after dates.

If descending, sort `6` and `6/` before `6/1`.

PR #3176, opened 2025-08-12
Fix 3070: Fix `Sort non-dates after dates`

codebuild-ro:
Fixes: https://github.com/cybersemics/em/issues/3070

`isDatePattern` is created for checking if text is date formatted or not.
And I more prioritized the date formatted text than non-dates text.
Also I added test cases for updated logic.

### DIFF
--- a/src/util/compareThought.ts
+++ b/src/util/compareThought.ts
@@ -51,15 +51,26 @@ const removeIgnoredPrefixes = (s: string) => s.replace(REGEX_IGNORED_PREFIXES, '
 /** Removes emojis, spaces, and prefix 'the' to make a string comparable.  */
 const normalizeCharacters = _.flow(removeEmojisAndSpaces, removeIgnoredPrefixes, removeDiacritics)
 
-/** Parse a date string and handle M/d (e.g. "2/1") for Safari. */
-const parseDate = (s: string): number =>
-  Date.parse(
-    REGEX_SHORT_DATE_WITH_DASH.test(s)
-      ? `${s}-${CURRENT_YEAR}`
-      : REGEX_SHORT_DATE_WITH_SLASH.test(s)
-        ? `${s}/${CURRENT_YEAR}`
-        : s,
-  )
+/** Parse a date string and handle M/d (e.g. "2/1") and written formats (e.g. "March 3") for Safari. */
+const parseDate = (s: string): number => {
+  // Handle numeric short dates (M/d or M-d)
+  if (REGEX_SHORT_DATE_WITH_DASH.test(s)) {
+    return Date.parse(`${s}-${CURRENT_YEAR}`)
+  }
+  if (REGEX_SHORT_DATE_WITH_SLASH.test(s)) {
+    return Date.parse(`${s}/${CURRENT_YEAR}`)
+  }
+
+  // Handle written short dates (Month Day without year)
+  const writtenShortPattern =
+    /^(January|February|March|April|May|June|July|August|September|October|November|December)\s+\d{1,2}$/i
+  if (writtenShortPattern.test(s)) {
+    return Date.parse(`${s}, ${CURRENT_YEAR}`)
+  }
+
+  // Handle full dates (with year) - let Date.parse handle them
+  return Date.parse(s)
+}
 /** Converts a string to a number. If given a number, returns it as-is. If given a string with a prefixe such as "#" or "$", strips it and returns the actual number. If given a number range, returns the start of the range. If the input cannot be converted to a number, returns NaN. */
 const toNumber = (x: number | string): number =>
   // match hyphen, em-dash, and em-dash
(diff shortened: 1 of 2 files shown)


src/util/compareThought.ts
```diff
@@ -68,6 +79,28 @@ const toNumber = (x: number | string): number =>
...
+
+  // Match numeric patterns with consistent separators
+  // Either all slashes: "6/21" or "6/21/2025"
+  const slashPattern = /^\d{1,2}\/\d{1,2}(\/\d{4})?$/
+  // Or all dashes: "6-21" or "6-21-2025"
+  const dashPattern = /^\d{1,2}-\d{1,2}(-\d{4})?$/
+
+  // Match written month patterns: "Month Day, Year" or "Month Day"
+  const writtenPattern =
+    /^(January|February|March|April|May|June|July|August|September|October|November|December)\s+\d{1,2}(,\s*\d{4})?$/i
```
raineorshine: While it's nicely contained, regex instantiation is known to be slow, so you'll want to declare them as constants outside the function.

src/util/compareThought.ts
```diff
@@ -68,6 +79,28 @@ const toNumber = (x: number | string): number =>
...
+ * Also accepts written month names like "March 3, 2020" or "December 3, 2020"
+ * Requires consistent separators (all slashes or all dashes)
+ * Examples: "6/21", "6-21", "12/1", "12-1", "6/21/2025", "6-21-2025", "March 3, 2020", "December 3, 2020".
+ */
+export const isDatePattern = (value: string): boolean => {
+  const trimmed = value.trim()
+
+  // Match numeric patterns with consistent separators
+  // Either all slashes: "6/21" or "6/21/2025"
+  const slashPattern = /^\d{1,2}\/\d{1,2}(\/\d{4})?$/
```
raineorshine: You've defined a regex here for `slashPattern`, but we also have `REGEX_SHORT_DATE_WITH_DASH`. Please unify or put them on the same footing so that we have a more cohesive design.

src/util/compareThought.ts
```diff
@@ -51,15 +51,26 @@ const removeIgnoredPrefixes = (s: string) => s.replace(REGEX_IGNORED_PREFIXES, '
...
+
+  // Handle written short dates (Month Day without year)
+  const writtenShortPattern =
+    /^(January|February|March|April|May|June|July|August|September|October|November|December)\s+\d{1,2}$/i
+  if (writtenShortPattern.test(s)) {
+    return Date.parse(`${s}, ${CURRENT_YEAR}`)
+  }
+
+  // Handle full dates (with year) - let Date.parse handle them
+  return Date.parse(s)
```
raineorshine: Notice that `Date.parse` is duplicated four times in this function, repeated in every return statement. 

Factoring out this duplication will clarify the distinction between the branches.

src/util/compareThought.ts
```diff
@@ -68,6 +79,28 @@ const toNumber = (x: number | string): number =>
...
 const isNumber = (x: number | string): boolean => !isNaN(toNumber(x))
 
+/** Checks if a string matches a date pattern like "M/d", "M-d", "M/d/yyyy", "M-d-yyyy", or written formats.
+ * Accepts 1-2 digits for month and day, optionally followed by year
+ * Also accepts written month names like "March 3, 2020" or "December 3, 2020"
+ * Requires consistent separators (all slashes or all dashes)
+ * Examples: "6/21", "6-21", "12/1", "12-1", "6/21/2025", "6-21-2025", "March 3, 2020", "December 3, 2020".
+ */
+export const isDatePattern = (value: string): boolean => {
+  const trimmed = value.trim()
```
raineorshine: It doesn't seem like trimming should belong here.

src/util/compareThought.ts
```diff
@@ -126,14 +159,24 @@ export const compareFormatting = <T, U>(a: T, b: U): ComparatorValue => {
   return aIsHtml && !bIsHtml ? -1 : bIsHtml && !aIsHtml ? 1 : 0
 }
 
-/** A comparison function that sorts date strings. */
-export const compareDateStrings: ComparatorFunction<string> = (a: string, b: string) =>
-  compare(parseDate(a), parseDate(b))
+/** A comparison function that sorts date strings. Only handles date vs date comparisons. */
+export const compareDateStrings: ComparatorFunction<string> = (a: string, b: string) => {
+  const aIsDate = isDatePattern(a)
+  const bIsDate = isDatePattern(b)
```
raineorshine: Great, I like how you've extended this function and followed the pattern of other comparison functions.

raineorshine (changes requested):
Overall this looks very good. Thanks!

src/util/compareThought.ts
codebuild-ro: I removed the trimming codde in `isDatePattern` function.
And I added the trimming code into usage of `isDatePattern` function. - `isDatePattern(text.trim())`

src/util/compareThought.ts
codebuild-ro: I fixed this and now `Date.parse` function is called only once at the end.

src/util/compareThought.ts
codebuild-ro: I fixed and declared as constant variables.

src/util/compareThought.ts
codebuild-ro: I fixed and declared as constant variables.

src/util/compareThought.ts
```diff
@@ -51,15 +69,28 @@ const removeIgnoredPrefixes = (s: string) => s.replace(REGEX_IGNORED_PREFIXES, '
...
-      : REGEX_SHORT_DATE_WITH_SLASH.test(s)
-        ? `${s}/${CURRENT_YEAR}`
-        : s,
-  )
+/** Parse a date string and handle M/d (e.g. "2/1") and written formats (e.g. "March 3") for Safari. */
+const parseDate = (s: string): number => {
+  // Use pre-compiled regex for optimal performance
+  const shortDateMatch = s.match(REGEX_SHORT_DATE_PARSE)
+
+  let dateString = s
```
raineorshine: Note quite there yet. Notice the duplication in your template strings. Start from here and avoid `let`:

```ts
const separator = s.includes('/') ? '/' : s.includes('-') ? '-' : ', '
```

src/util/compareThought.ts
```diff
@@ -16,13 +16,31 @@ const IGNORED_PREFIXES = ['the ']
...
+
+// Combined regex for detecting all valid date formats:
+// - M/d or M/d/yyyy (slash format with optional year)
+// - M-d or M-d-yyyy (dash format with optional year)
+// - Month Day or Month Day, yyyy (written format with optional year)
+// Uses separate patterns to ensure consistent separators (no mixed slashes/dashes)
+const REGEX_DATE_COMBINED = new RegExp(
+  `^(\\d{1,2}\\/\\d{1,2}(\\/\\d{4})?|\\d{1,2}-\\d{1,2}(-\\d{4})?|(${MONTH_NAMES})\\s+\\d{1,2}(,\\s*\\d{4})?)$`,
+  'i',
+)
```
raineorshine, line 36: Thanks, this looks better now. Ultimately I think we could have a single REGEX with an optional year, but we don't have to do that here.

src/util/compareThought.ts
codebuild-ro: I updated the `parseDate` function to avoid duplication of template string.
```
const parseDate = (s: string): number => {
  const shortDateMatch = s.match(REGEX_SHORT_DATE_PARSE)

  const dateString = shortDateMatch
    ? `${s}${s.includes('/') ? '/' : s.includes('-') ? '-' : ', '}${CURRENT_YEAR}`
    : s

  return Date.parse(dateString)
}
```

raineorshine (approved):

PR #3177, opened 2025-08-12
Replace dragHold and dragInProgress actions/state with longPress

### LINKED ISSUE #3173: Replace dragHold and dragInProgress actions with longPress
The actions can be consolidated into a single pathway by using a [union type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) to vary the shape of the payload depending on the enum value.

```
interface InactivePayload {
  value: LongPressState.Inactive
}

interface DragHoldPayload {
  value: LongPressState.DragHold
  simplePath?: SimplePath
  sourceZone?: DragThoughtZone
}

interface DragInProgressPayload {
  value: LongPressState.DragInProgress
  // Sets state.draggingThoughts. Either hoveringPath or file must be set if value is true.
  draggingThoughts?: SimplePath[]
  hoveringPath?: Path
  hoverZone?: DropThoughtZone
  // Sets state.draggingFile. Either hoveringPath or file must be set if value is true.
  draggingFile?: boolean
  offset?: number
  sourceZone?: DragThoughtZone
}

interface DragCanceledPayload {
  value: LongPressState.DragCanceled
}

type Payload = InactivePayload | DragHoldPayload | DragInProgressPayload | DragCanceledPayload
```

ethan-james:
Fixes #3173

Instead of maintaining separate flags for `dragHold` and `dragInProgress`, we want to centralize and enforce exclusivity by using the `longPress` enum.


src/actions/longPress.ts
```diff
@@ -1,41 +1,164 @@
...
+        haptics.light()
+        break
+      case LongPressState.Inactive:
+        globals.touching = false
+        break
+    }
+
     dispatch({ type: 'longPress', ...payload })
 
+    if (value === LongPressState.DragInProgress) {
```
raineorshine, line 145: The original `dragInProgress` action creator executes the below logic regardless of `value`, i.e. when both enabling and disabling a drag in progress. The new code only executes when enabling. I think this will result in different behavior?

src/hooks/useDragDropFavorites.ts
```diff
@@ -45,8 +44,7 @@ const endDrag = () => {
   longPressStore.unlock()
   globals.longpressing = false
   store.dispatch([
-    dragInProgress({ value: false }),
-    dragHold({ value: false }),
+    longPress({ value: LongPressState.Inactive }),
```
raineorshine, line 45: Great, I like the way this refactor has combined many of the cases where `dragInProgress` and `dragHold` needed to be updated together.

raineorshine (changes requested):
Thanks!

src/hooks/useDragDropFavorites.ts
ethan-james: Me too!

src/actions/longPress.ts
ethan-james: Yes, thanks, good catch. I updated it to also call those when `previousValue === LongPressState.DragInProgress`, i.e. when leaving `DragInProgress` state.

I left `shaker` in its previous condition since `hoveringPath` no longer exists on `Payload` in other states, and was always getting called with an `undefined` ID whenever `dragInProgress({ value: false })` was called. Which I believe means that it wasn't doing anything when called in that way.

Probably a good idea for me to take one last look and test more thoroughly tomorrow morning.

raineorshine: Yeah, it's fine if the shaker is only activated during `DragInProgress`.

ethan-james:
> One failing drag-and-drop test. Otherwise looks good!

I haven't been able to get it to fail locally, or again on CI.

I did find the source of the 'Invalid longPress transition' errors that have been bugging me during local Puppeteer runs. Desktop doesn't wait for `dragHold` before beginning `dragInProgress`, so the desktop long press timeout was executing after a drag had also begun. I cleaned that up, although it wasn't affecting functionality, so if you think there's a performance penalty then I can revert it.

ethan-james:
Well, I spoke too soon, but at least a different drag-and-drop test failed this time.

ethan-james:
That was a bit of a journey, but it turns out that canceling the timeout led to `longPressStore.lock()` not being called. This was buggy behavior and could have caused other problems, but in this case it was preventing

```
  useEffect(() => {
    if (!isLocked) setPressing(false)
  }, [isLocked, setPressing])
```

from erroneously calling `setPressing(false)`. I set up an `onDragEnd` handler to make sure that `stop` gets called on desktop when the drag ends, and now I think everything is working.

raineorshine (approved):
Thank you!

Issue #3175, opened 2025-08-10
Re-test PR #2960


ethan-james:
In light of #3023 invoking native browser draggable behavior, it may no longer be necessary to disable the editable during a drag.

ethan-james:
https://github.com/ethan-james/em/blob/3b8576b4bda9bc4447917dfc40c9198ee4a664bc/src/components/Editable.tsx#L518

`onFocus` is still called during the drag, so I need to keep checking if a long press is in progress before calling `setCursorOnThought`.

ethan-james:
https://github.com/ethan-james/em/blob/3b8576b4bda9bc4447917dfc40c9198ee4a664bc/src/actions/dragInProgress.ts#L127 (https://github.com/cybersemics/em/issues/1704)

I was able to create https://github.com/cybersemics/em/pull/3179 to get rid of the unnecessary `selection.clear()`.

ethan-james:
4. https://github.com/cybersemics/em/pull/2962

This is still very important. It turns out that if you remove `useSelect: 'none'`, you can still highlight text during drag. I wouldn't have thought that would be possible anymore.

ethan-james:
5. https://github.com/cybersemics/em/pull/3102

I believe that I could remove the `setTimeout` from `useLongPress` if I could run the `stop` function `onClick` rather than `onMouseUp`/`onTouchEnd`. It still might be too wide-ranging for a refactor task, but I can check back in after re-testing points 6 and 7 on the list.

raineorshine:
Thanks for taking the time to do this. It's great to streamline our drag-and-drop and long press code.

FYI I think we lost the ability to keep the virtual keyboard up while dragging and dropping. This makes it annoying to go back to editing after dragging a thought. However, the advantage is that you have more visible screen space for dragging, and it is less susceptible to weird edge cases where the browser automatically changes the selection. We can open a separate issue, but thought I would mention it here. Probably need user feedback for a clearer design direction anyway.

ethan-james:
8. https://github.com/ethan-james/em/blob/80a075ffba1c268d2ae77aff7c3fa0ed14120a08/src/components/Editable.tsx#L577 (see if it will still scroll to editable during drag)

It does still scroll and move the cursor after a long press, so this needs to stay.

ethan-james:
10. Added to https://github.com/cybersemics/em/pull/3187#issuecomment-3207832649

ethan-james:
13. Addressed in #3190

[...]

Issue #1704, opened 2023-02-12
[mobile] Gestures stop working after tap-to-select


raineorshine:
## Steps to Reproduce

1. Create a thought 
2. Close the keyboard
3. Tap and hold over a word in the thought in order to activate drag-and-drop
  - Mobile Safari's tap-to-select functionality will select the word 
4. Drag the thought away and then back
  - This should visibly clear the selection 
5. End the drag 
6. Try to make a gesture 

## Current Behavior

Nothing happens.

The selection still exists even thought it is not visible, which disables MultiGesture.  

## Expected Behavior

Gestures should work. 

The selection should be cleared after the drag ends. 

## Notes

This won't be an issue if tap-to-select can be disabled during drag-and-drop. Still, it doesn't hurt to have an independent fix since it is very bad for gestures to silently fail.

PR #3179, opened 2025-08-13
Remove selection.clear when ending long press

ethan-james:
References #3175

Thanks to #3023, tap-to-select no longer activates during a long press. Now this workaround is unnecessary. I tested against #1704 to make sure. I also added this change to #3177 since I moved `dragInProgress.ts` into `longPress.ts`. So we may not end up merging this if that PR gets approved soon.


raineorshine (approved):

raineorshine:
Looks good!

PR #3184, opened 2025-08-14
Remove globals.longpressing

ethan-james:
References #3175

There is only [one place](https://github.com/ethan-james/em/blob/80a075ffba1c268d2ae77aff7c3fa0ed14120a08/src/hooks/useLongPress.ts#L89) where `globals.longpressing` was being tested, and the main reason for that was to check if the long press had been canceled by our `touchmove` handler, which is now gone.

There is now significant redundancy between [onLongPressEnd](https://github.com/ethan-james/em/blob/77f854d559f95ea8147484ada69f9e52a708d56b/src/hooks/useDragHold.ts#L48) and [endDrag](https://github.com/ethan-james/em/blob/77f854d559f95ea8147484ada69f9e52a708d56b/src/hooks/useDragAndDropThought.tsx#L108). I'm pretty confident that I can do away with `endDrag` entirely, since in my testing I always see [stop](https://github.com/ethan-james/em/blob/77f854d559f95ea8147484ada69f9e52a708d56b/src/hooks/useLongPress.ts#L75) getting called at the end of any long press, including a drag.

I figured it would be best to start small and work toward eliminating `endDrag` and even [the useEffect in useDragHold](https://github.com/ethan-james/em/blob/77f854d559f95ea8147484ada69f9e52a708d56b/src/hooks/useDragHold.ts#L70), especially due to this comment:

```
// react-dnd stops propagation so onLongPressEnd sometimes doesn't get called.
```

I don't think that `react-dnd` is still stopping propagation, but please let me know if there are drag-and-drop edge cases that I might need to test.

### DIFF
--- a/src/globals.ts
+++ b/src/globals.ts
@@ -2,10 +2,6 @@
 
 /** THE BAD PLACE where mutable globals are defined. */
 
-// globally tracks when a long press starts and ends
-// this is useful to prevent long-tap-to-select on mobile safari
-let longpressing = false
-
 // track whether the user is touchmoving so that we can distinguish touchend events from tap or drag
 // not related to react-dnd
 let touching = false
@@ -39,7 +35,6 @@ const globals = {
   freeThoughtsThreshold,
   errorTimer,
   offlineTimer,
-  longpressing,
   rendered,
   suppressExpansion,
   touching,
--- a/src/hooks/useDragAndDropThought.tsx
+++ b/src/hooks/useDragAndDropThought.tsx
@@ -23,7 +23,6 @@ import { ThoughtContainerProps } from '../components/Thought'
 import { AlertType, LongPressState } from '../constants'
 import allowTouchToScroll from '../device/allowTouchToScroll'
 import * as selection from '../device/selection'
-import globals from '../globals'
 import documentSort from '../selectors/documentSort'
 import findDescendant from '../selectors/findDescendant'
 import getNextRank from '../selectors/getNextRank'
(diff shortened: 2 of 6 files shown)


raineorshine (approved):
Thanks so much!

> There is only [one place](https://github.com/ethan-james/em/blob/80a075ffba1c268d2ae77aff7c3fa0ed14120a08/src/hooks/useLongPress.ts#L89) where `globals.longpressing` was being tested, and the main reason for that was to check if the long press had been canceled by our `touchmove` handler, which is now gone.

Yes, traced back to ce0bbeb, but not issue number.

`globals.longpressing` predates ce0bbeb, but the other uses are gone so it's simpler to evaluate the implications of removing it.

> There is now significant redundancy between [onLongPressEnd](https://github.com/ethan-james/em/blob/77f854d559f95ea8147484ada69f9e52a708d56b/src/hooks/useDragHold.ts#L48) and [endDrag](https://github.com/ethan-james/em/blob/77f854d559f95ea8147484ada69f9e52a708d56b/src/hooks/useDragAndDropThought.tsx#L108). I'm pretty confident that I can do away with `endDrag` entirely, since in my testing I always see [stop](https://github.com/ethan-james/em/blob/77f854d559f95ea8147484ada69f9e52a708d56b/src/hooks/useLongPress.ts#L75) getting called at the end of any long press, including a drag.

Great observation, that would be wonderful.

> I don't think that `react-dnd` is still stopping propagation, but please let me know if there are drag-and-drop edge cases that I might need to test.

I don't know any off the top of my head. It's possible it has changed after our TouchBackend work. I'm willing to risk it, as this seems like a good opportunity to simplify. We can always revert if we discover a regression.

PR #3186, opened 2025-08-17
Fix for drag and drop favorite throws an error

### LINKED ISSUE #3180: Drag-and-drop favorite throws an error
Reproducible in iOS Safari, Android Chrome and desktop Chrome

## Steps to Reproduce

1. Place cursor on a thought
2. Favorite the thought
3. Open the sidebar (to Favorites tab)
4. Long press on the thought itself (not the home icon above it)
5. Drag

## Current Behavior

Throws a `draggedItems.every is not a function` error.

https://github.com/user-attachments/assets/cb029ceb-2e16-4264-848c-4e47c293ad89

## Expected Behavior

No error.

---

The error is coming from https://github.com/cybersemics/em/blob/3b8576b4bda9bc4447917dfc40c9198ee4a664bc/src/hooks/useDragAndDropThought.tsx#L158, not from drag-and-drop favorites. Presumably all droppables are being checked regardless of source zone. This looks like it's related to #3016 by @karunkop, although I'd be happy to take it on.

karunkop:
Close #3180 

### Problem

Previously in #3016 , we updated how `draggingThoughts` is structured. Now it's consistently treated as an array of `SimplePath` items. This change ensures that when multiple thoughts/a single thought are dragged together, each one is represented properly in the form of an array.

However, in the case of dragging favorite thoughts, the `beginDrag` handler was still returning a single `DraggingThought` item, rather than an array. Since the `useDragAndDropThought` and `useDragAndDropSubThought` hooks expect `draggedItems` to be an array, this inconsistency caused unexpected behavior.

> **Although it is not clear to me why does these hooks run in case of favorite thoughts ?**

### Fix

I updated the `useDragDropFavorites` hook so that it now returns `draggingThoughts` as an array of `DragThoughtItem` matching the structure expected elsewhere in the drag-and-drop system. This change aligns the favorites drag-drop behavior with the rest of the hooks.

### Tests

Added e2e test cases for dragging and dropping favorites


src/e2e/puppeteer/helpers/dragAndDropFavorite.ts
```diff
@@ -0,0 +1,80 @@
...
+    await page.locator('[data-drag-in-progress="true"]').wait()
+
+    if (mouseUp) {
+      await page.mouse.up()
+      await waitUntil(() => !document.querySelector('[data-drag-in-progress="true"]'))
+    }
+  } catch (error) {
+    console.error('Drag and drop error:', error)
+    throw error
+  }
```
raineorshine: Why the try-catch?

src/components/Favorites.tsx
```diff
@@ -46,7 +46,12 @@ const DragAndDropFavorite = ({
   return (
     // Set overflow:auto so the drop target fully wraps its contents.
     // Otherwise the context-breadcrumbs margin-top will leak out and create a dead zone where the favorite cannot be dropped.
-    <div {...dragHoldResult.props} className={css({ overflow: 'auto' })} ref={node => dragSource(dropTarget(node))}>
+    <div
+      {...dragHoldResult.props}
+      className={css({ overflow: 'auto' })}
+      data-testid='favorite-item'
```
raineorshine: Let's call this `drag-and-drop-favorite` to match the component name.

src/e2e/puppeteer/__tests__/drag-and-drop-favorites.ts
```diff
@@ -0,0 +1,121 @@
...
+      const links = Array.from({ length: result.snapshotLength }, (_, i) => {
+        const link = result.snapshotItem(i) as HTMLElement
+        return link.textContent?.trim()?.toLowerCase() || ''
+      }).filter(text => text !== '')
+
+      return links
+    } catch (error) {
+      console.error('Error in XPath approach:', error)
+      return null
+    }
```
raineorshine: Probably don't want to swallow errors.

src/e2e/puppeteer/__tests__/drag-and-drop-favorites.ts
```diff
@@ -0,0 +1,121 @@
...
+import dragAndDropFavorite from '../helpers/dragAndDropFavorite'
+import paste from '../helpers/paste'
+import press from '../helpers/press'
+import waitForSelector from '../helpers/waitForSelector'
+import { page } from '../setup'
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+/** Get the order of favorites as displayed in the sidebar. */
+const getFavoritesOrder = async () => {
```
raineorshine: Let's call this `selectFavorites` or `selectFavoritesText`. It makes more sense to say that it is retrieving the favorites text *in order* rather than saying it *gets the order* itself.

src/e2e/puppeteer/__tests__/drag-and-drop-favorites.ts
```diff
@@ -0,0 +1,121 @@
...
+    try {
+      // XPath to find all thought-links that are NOT in breadcrumbs, within favorite-items
+      const xpath = `//*[@data-testid='favorite-item']//*[@data-thought-link and not(ancestor::*[@aria-label='context-breadcrumbs'])]`
+
+      const result = document.evaluate(xpath, document, null, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE, null)
+
+      if (result.snapshotLength === 0) {
+        console.warn('No favorite items found with XPath:', xpath)
+        return null
+      }
```
raineorshine: Is a special case needed? I would think 0 results would simply return an empty array, not null.

src/e2e/puppeteer/__tests__/drag-and-drop-favorites.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    await clickThought('d')
+    await click('[aria-label="Add to Favorites"]')
+
+    // Open favorites in sidebar
+    await click('[aria-label="menu"]')
+    await page.locator('[data-testid="sidebar"]').wait()
+
+    // Verify initial order
+    expect(await getFavoritesOrder()).toEqual(['a', 'b', 'c', 'd'])
```
raineorshine: The assertion is very readable, thanks.

src/e2e/puppeteer/helpers/dragAndDropFavorite.ts
```diff
@@ -0,0 +1,80 @@
...
+import waitUntil from './waitUntil'
+
+/**
+ * Helper to find favorite item by text value.
+ * Uses XPath to find the thought-link with exact text, not in breadcrumbs, within favorite-item.
+ */
+const findFavoriteItem = async (value: string) => {
+  return await page.evaluateHandle(value => {
+    // XPath to find thought-link with exact text value which is not in breadcrumbs but just within favorite-item
+    const xpath = `//*[@data-testid='favorite-item']//*[@data-thought-link and not(ancestor::*[@aria-label='context-breadcrumbs']) and normalize-space()='${value.toLowerCase()}']/ancestor::*[@data-testid='favorite-item']`
```
raineorshine: First, I don't see a reason to make this case insensitive, especially given that the JSDOC says it looks for the "exact text."

raineorshine: Second, can you explain the explicit breadcrumb exclusion? Are these for favorites from deeply nested contexts?

raineorshine: Third, I see `//*[@data-testid='favorite-item']` and the beginning and `/ancestor::*[@data-testid='favorite-item']` at the end of the selector. I'm not as familiar with XPATH, but is this necessary?

src/e2e/puppeteer/__tests__/drag-and-drop-favorites.ts
karunkop: No, the special case is not needed. This was actually added initially for the purpose of debugging and to warn that no favorite thoughts were found in the DOM. I have removed it now.

src/e2e/puppeteer/helpers/dragAndDropFavorite.ts
karunkop: The try-catch block was initially added to catch errors related to XPath while implementing it. I have removed it now

src/e2e/puppeteer/helpers/dragAndDropFavorite.ts
karunkop: The explicit exclusion of breadcrumbs is necessary because, in the DOM structure, we are specifically targeting direct <a> tag descendants. For example, in the following HTML snippet:

```html
<div class="ov_auto" data-testid="drag-and-drop-favorite" draggable="true">
  <div>
    <div aria-label="context-breadcrumbs" class="fs_0.867em c_gray66 ml_0 mt_1em min-h_1em lh_1.5">
      <div>
        <span class="">
          <a tabindex="-1" data-thought-link="true" class="..." role="button">d</a>
          <span class="..."></span>
        </span>
      </div>
    </div>
    <a tabindex="-1" data-thought-link="true" class="..." role="button">e</a>
    <span class="..."></span>
  </div>
</div>
```


Here, both the breadcrumb `d` and the actual favorite thought `e` use the `data-thought-link` attribute. However, the breadcrumb is nested within an element labeled as `context-breadcrumbs`, while the favorite thought is a direct child of the container.

Without explicitly excluding elements inside the breadcrumb container, any XPath expression that matches based solely on the `data-thought-link` attribute would incorrectly include the breadcrumb `d`. This would result in false positives when identifying favorite thoughts. By explicitly filtering out the breadcrumbs, we ensure that only valid, direct favorite thoughts like `e` are matched.

karunkop: > Third, I see `//*[@data-testid='favorite-item']` and the beginning and `/ancestor::*[@data-testid='favorite-item']` at the end of the selector. I'm not as familiar with XPATH, but is this necessary?

After a bit more googling, I was able to remove the redundancy with the pattern. In summary, this XPath expression selects all drag-and-drop-favorite elements that contain a data-thought-link attribute (not within context-breadcrumbs) with normalized text content exactly matching the specified value.

raineorshine: > The explicit exclusion of breadcrumbs is necessary because, in the DOM structure, we are specifically targeting direct tag descendants.

Okay, thanks for the explanation. I wish there was a way to select without the dependency on the DOM hierarchy, but I'll accept this for now.

PR #3187, opened 2025-08-18
useDragAndDropThought: remove redundant endDrag handler

ethan-james:
References #3175

After removing `globals.longpressing` (#3184) and adding an `onDragEnd` handler for desktop (#3177), there is no more need for `endDrag` in `useDragAndDropThought`. One way or another, `stop` in `useLongPress` will get called, and will clean up the long press functionality

### DIFF
--- a/src/hooks/useLongPress.ts
+++ b/src/hooks/useLongPress.ts
@@ -21,6 +21,7 @@ const useLongPress = (
 ) => {
   const [pressing, setPressing] = useState(false)
   // Track isLocked state from longPressStore in local state
+  const isLocked = longPressStore.useSelector(state => state.isLocked)
   const longPressState = useSelector(state => state.longPress)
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
Hi, thanks for this improvement!

> This is the only place where `longPressStore.isLocked` was being tested, so I can probably remove the store entirely. I want to circle back and look at it again as part of #3175 before I do that.

If `longPressStore` is not used anywhere else, we should remove it completely in a single PR. Could we separate that change out into its own PR? I think that would make more sense than doing it partially here, and it would give us a better division of commits for regression testing.

It would also be worth referencing the commit/PR that introduced `longPressStore`, and adding a little explanation of why the original need is no longer present. Making a separate PR will help keep that explanation separate from the refactoring task here.

Thanks!

raineorshine (approved):
Thank you!

PR #3188, opened 2025-08-20
Added test to drag & drop as thought to remove it from favorite

Valere3162:
Added test to drag & drop as thought to remove it from favorite

### DIFF
--- a/src/commands/__tests__/favoriteDragAndDrop.ts
+++ b/src/commands/__tests__/favoriteDragAndDrop.ts
@@ -1,0 +1,54 @@
+import { act } from 'react'
+import { importTextActionCreator as importText } from '../../actions/importText'
+import { HOME_TOKEN } from '../../constants'
+import exportContext from '../../selectors/exportContext'
+import store from '../../stores/app'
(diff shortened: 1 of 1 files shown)


src/commands/__tests__/favoriteDragAndDrop.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(thoughtA).toBeTruthy()
+
+    await act(async () => {
+      store.dispatch(setCursor(['A']))
+    })
+
+    // Simulate the unfavorite action by executing the favorite command
+    // as if it were triggered by the drag and drop.
+    await act(async () => {
+      executeCommand(favorite, { store })
```
raineorshine, line 45: Executing the command directly means that this test doesn't actually cover drag-and-drop at all.

Luckily, Karun just added some drag-and-drop tests. You can follow his lead, and add a test there.

https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/__tests__/drag-and-drop-favorites.ts

raineorshine (changes requested):

PR #3189, opened 2025-08-20
Remove scrollTo when drag begins at scrollY position 0

ethan-james:
References #3175

This band-aid originated in https://github.com/cybersemics/em/commit/16971ff08b3eb84b66f8f49b3dff1e657283e4a1, which I can't install due to a dependency that has disappeared from the internet. Based on my testing, though, drag appears to behave well while scroll is at the top. There is also a comment stating `It still glitches out if you scroll back to the top during a drag`, and I don't see any glitching while using `scrollAtEdge` to scroll back to the top.

The `scrollTo` behavior doesn't seem to cause any issues other than a tiny layout shift at the beginning of the drag, so of course we can keep it if we feel that removing it might still cause problems.

### DIFF
--- a/src/actions/longPress.ts
+++ b/src/actions/longPress.ts
@@ -5,7 +5,6 @@ import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
 import State from '../@types/State'
 import Thunk from '../@types/Thunk'
-import { isSafari } from '../browser'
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Okay, thanks! I'm willing to remove it. Maybe the fixes we did to `TouchBackend` fixed it.

Thank you for tracing the history, too!

PR #3190, opened 2025-08-20
useDragHold: remove isDragging effect, clearMulticursors in beginDrag instead

ethan-james:
References #3175 

The effect handler in `useDragHold` was performing 4 functions:

1. `setIsPressed(false)` when a drag began
2. `dispatch([alert(null)])` when a drag began
3. `dispatch(clearMulticursors())` when a drag began
4. `longPressStore.unlock()`, probably never (thankfully)

Now that `globals.longpressing` was removed in #3184, `onLongPressEnd` will always run and the local state `isPressed` can be cleared directly rather than relying on the `isDragging` side effect.

[dispatch(alert(null))](https://github.com/ethan-james/em/blob/14c356fa8255f72fc5a62f3339ed03583d41040a/src/hooks/useDragAndDropThought.tsx#L123) already runs in `endDrag`, which is the only place where the alert needs to be cleared. Running it in `useDragHold` was redundant.

I moved `dispatch(clearMulticursors())` into [beginDrag](https://github.com/ethan-james/em/blob/14c356fa8255f72fc5a62f3339ed03583d41040a/src/hooks/useDragAndDropThought.tsx#L100).

The condition `if (!isDragging && state.longPress === LongPressState.DragHold)` was probably never true because `state.longPress` wasn't part of the dependency array. This is good because running `longPressStore.unlock()` as soon as a drag begins seems much earlier than it would run in any other context.

### DIFF
--- a/src/hooks/useDragAndDropThought.tsx
+++ b/src/hooks/useDragAndDropThought.tsx
@@ -11,6 +11,7 @@ import SimplePath from '../@types/SimplePath'
 import State from '../@types/State'
 import { addMulticursorActionCreator as addMulticursor } from '../actions/addMulticursor'
 import { alertActionCreator as alert } from '../actions/alert'
+import { clearMulticursorsActionCreator as clearMulticursors } from '../actions/clearMulticursors'
 import { createThoughtActionCreator as createThought } from '../actions/createThought'
 import { errorActionCreator as error } from '../actions/error'
 import { importFilesActionCreator as importFiles } from '../actions/importFiles'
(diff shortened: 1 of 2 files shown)


raineorshine (approved):
Thanks! 

Can you reference the original commit/PR in which the `useEffect` was added? That will be helpful for documentation purposes.

> The condition `if (!isDragging && state.longPress === LongPressState.DragHold)` was probably never true because `state.longPress` wasn't part of the dependency array.

This is not necessarily true. The thunk is pulling a fresh state once it's dispatched, so `state.longPress` may be updated by the time it's checked. Presumably I didn't want to trigger the render function again for performance reasons, and determined that it was safe to check the updated state. Though I recognize that this approach might not be a best practice.

raineorshine (approved):
Great! Thanks so much for the explanation and additional documentation.

PR #3191, opened 2025-08-21
Remove longPressStore

ethan-james:
References #3175 

We're setting `pressing` state to determine which thought is being long-pressed, which is necessary because the new `dragStart` event fires indiscriminately and notifies all listeners, i.e. all thoughts. As a result, I can't conceive of a situation where `isLocked` changes independently of `pressing`. The touch that triggers `pressing` will be confined to the single thought in the source zone that is being long-pressed. Therefore, and because `isLocked` was causing problems by making [useEffect](https://github.com/ethan-james/em/blob/db56015d3eed697fb108b517761828b56a2724ed/src/hooks/useLongPress.ts#L30) re-run, I removed the check for `isLocked`.

This is the only place where `longPressStore.isLocked` was being tested, so I removed `longPressStore` entirely.

It was added in #2847 in order to replace a local `lock` variable in `useLongPress`. Now that we're tracking the `pressing` state of a thought, and guaranteeing that `stop` runs and cleans up `useLongPress` at the end of all long presses or drags, it is no longer needed.


raineorshine:
@JoeDuong-dev A rough test with various combinations of long press and drag-and-drop would be great! 

This is a refactor, and should result in no behavioral changes to the app. Look for any unexpected long press or drag-and-drop triggers or cancellations.

ethan-james:
> Saw those merge conflicts coming 😆
> 
> Looks great!

Thanks! Hopefully they're just changes to files that were deleted elsewhere.

PR #3197, opened 2025-08-26
Add touch-action: none and overscroll-behavior: contain to dialog recipe

### LINKED ISSUE #3165: [Mobile Safari] Document body is scrollable when cheatsheet is open

ethan-james:
Fixes #3165 

If it works, this is a much cleaner solution than #3171. Now that we learned that `overscroll-behavior: contain` seems to effectively stop iOS Safari from transferring scroll to a container, I realized that I could try to pair it my earliest solution, `touch-action: none` on the overlay. So far it looks like they actually work together to keep all scrolling within the dialog's content pane. Calling `preventDefault` on touch events is still necessary to disable Safari's [swipe to go back](https://www.reddit.com/r/iphone/comments/1d99y8s/can_i_turn_off_swipe_to_go_back_to_previous_screen/) behavior.


ethan-james:
When scrolling like a maniac, I frequently end up switching away from the app. When I return, I see the body scrollbar appear, but it doesn't seem to move at all.

ethan-james:
> Love this!
> 
> @JoeDuong-dev Ready for testing.

Great! I'm curious whether this has the same orientation bug.

JoeDuong-dev (commented):
@ethan-james I tested using the same scenarios from https://github.com/cybersemics/em/pull/3171, and unfortunately the results are similar. Most cases passed, but the same orientation bug still occurs.

This time, I was also able to reproduce the issue on Chrome as well as Safari. On Chrome, if the user scrolls until the scrollbar appears, opens the cheatsheet with the gesture, rotates the phone to landscape mode, and then scrolls on the sidebar, the page remains scrollable while in landscape mode.

ethan-james:
> @ethan-james I tested using the same scenarios from #3171, and unfortunately the results are similar. Most cases passed, but the same orientation bug still occurs.
> 
> This time, I was also able to reproduce the issue on Chrome as well as Safari. On Chrome, if the user scrolls until the scrollbar appears, opens the cheatsheet with the gesture, rotates the phone to landscape mode, and then scrolls on the sidebar, the page remains scrollable while in landscape mode.

Thanks @JoeDuong-dev, I'm looking into it. I haven't been able to reproduce it yet in either browser, but I'm studying your video again. I notice that in that, you are actually scrolling down below the footer. I imagine that might be because #3171 offsets the content as a workaround for disabling scrolling. In this branch, are you still scrolling below the footer, or does it stop scrolling when you hit the bottom of the page?

ethan-james:
> @ethan-james I have uploaded another video with narration to clear up any confusion.

I see it! Wow, good catch. I can only reproduce it on Chrome on iOS. Can you please confirm: in Safari does it occur after rotating to landscape as it does in Chrome, or only after rotating back to portrait?

raineorshine:
Thanks for all your work on this!

PR #3200, opened 2025-08-28
Update configuration of dependabot

codebuild-ro:
### Fix Dependabot peer dependency resolution issues.
image

Updates configuration to prevent peer dependency conflicts by:
- Limiting to patch updates only
- Adding ignore rules for problematic packages


raineorshine (approved):
Thank you, I'll accept this as a temporary step.

PR #3203, opened 2025-08-29
Remove toggleMulticursor and stopPropagation from Bullet click handler

### LINKED ISSUE #3066: Multiselect: Ctrl + click does not work on bullet
## Steps to Reproduce

Ctrl + click (Windows) or ⌘ + click (macOS) a thought's bullet.

## Current Behavior

Nothing happens.

## Expected Behavior

Should toggle multiselect on the thought.

Should work the same as Ctrl + click on the editable.

ethan-james:
Fixes #3066

`toggleMulticursor` was being dispatched in `Bullet` and again in `Thought`, and they were canceling each other out. Now `Thought` can handle ctrl-click actions for both `Editable` and `Bullet`.

Also, `Bullet` was still calling `stopPropagation` to prevent `Content.clickOnEmptySpace`. This is no longer necessary after #3178.

### DIFF
--- a/src/components/Bullet.tsx
+++ b/src/components/Bullet.tsx
@@ -9,7 +9,6 @@ import ThoughtId from '../@types/ThoughtId'
 import { deleteAttributeActionCreator as deleteAttribute } from '../actions/deleteAttribute'
 import { setCursorActionCreator as setCursor } from '../actions/setCursor'
 import { setDescendantActionCreator as setDescendant } from '../actions/setDescendant'
-import { toggleMulticursorActionCreator as toggleMulticursor } from '../actions/toggleMulticursor'
 import { isMac, isSafari, isTouch, isiPhone } from '../browser'
 import { AlertType, LongPressState } from '../constants'
 import attributeEquals from '../selectors/attributeEquals'
@@ -483,7 +482,6 @@ const Bullet = ({
 
       // short circuit if toggling multiselect
       if (!isTouch && (isMac ? e.metaKey : e.ctrlKey)) {
-        dispatch(toggleMulticursor({ path }))
         return
       }
 
@@ -508,9 +506,6 @@ const Bullet = ({
           setCursor({ path: shouldCollapse ? pathParent : path, preserveMulticursor: true }),
         ])
       })
-
-      e.stopPropagation()
-      // stop click event from bubbling up to Content.clickOnEmptySpace
     },
     [dispatch, dragHold, path, simplePath],
   )


JoeDuong-dev:
@ethan-james I just tested this and multi-select is working. 

I’d like to clarify one behavior: if there are multiple levels of sub-thoughts (as shown in the screenshot), how should multi-select function? Will each sub-thought be selected individually, or does selecting a parent thought also include all of its sub-thoughts?

Screenshot 2025-08-29 at 2 20 15 PM

ethan-james:
> @ethan-james I just tested this and multi-select is working.
> 
> I’d like to clarify one behavior: if there are multiple levels of sub-thoughts (as shown in the screenshot), how should multi-select function? Will each sub-thought be selected individually, or does selecting a parent thought also include all of its sub-thoughts?

I believe that's correct (select each sub-thought individually) but the real answer is that it should behave the same way that meta-click works on the thought text on `main`. Meaning that, in `main`, you can ctrl-click directly on the `1 level deeper` text rather than the bullet and it should give you the same result.

JoeDuong-dev:
@ethan-james Thanks for clarifying that behavior.

I’m noticing a difference between using Ctrl-click (or Cmd-click) on the bullet versus on the text.
	•	When I Ctrl-click the text, it gets selected, but my “test 2” collapses.
	•	When I Ctrl-click the bullet directly, that collapse does not occur.

Jam link: https://jam.dev/c/ba6fb328-9aa3-49e5-a44b-031b2f70d186

ethan-james:
> I’m noticing a difference between using Ctrl-click (or Cmd-click) on the bullet versus on the text. • When I Ctrl-click the text, it gets selected, but my “test 2” collapses. • When I Ctrl-click the bullet directly, that collapse does not occur.

Hmm, thanks for pointing that out. My guess would be that they should not collapse so that you can multiselect whatever you want, but that's not the way it has been working in `main`. Only cmd-click on the editable text used to work, therefore it would collapse. Maybe @raineorshine can clarify.

ethan-james:
I also noticed that multiselect status doesn't clear (on desktop or on mobile) when performing normal taps/clicks. I see in the code that it is [ignoring touches while multicursor is active](https://github.com/ethan-james/em/blob/66e740ae87c52ada9cd24ff44545404dd97c5ea2/src/components/Editable.tsx#L555) but it is still [moving the cursor](https://github.com/ethan-james/em/blob/66e740ae87c52ada9cd24ff44545404dd97c5ea2/src/components/Editable.tsx#L608). Is this what we want to see?

https://github.com/user-attachments/assets/a03f0fd3-cce8-4f5b-8bd7-53c4c842b7fa

raineorshine:
Yes, that’s expected. At the moment, cursor and multiselect are independent to allow navigation while thoughts are selected. I'd love to support single tap to select once multiselect is active, but haven't found a good alternative that still allows navigation.

PR #3204, opened 2025-08-30
Animate drag thought

### LINKED ISSUE #2532: Animation: Drag thought
## Steps to Reproduce

```
- a
- b
- c
- d
- e
```

Drag `a` after `e`.

## Current Behavior

`a` snaps into place and the other thoughts translate up.

Note: The snapping may be related to #2498.

https://github.com/user-attachments/assets/ec16840b-217b-424d-9d24-45a013a38289

## Expected Behavior

`a` should enlarge and translate into place (so that it looks like it is above the other thoughts on the z axis).

zukilover:
Fixes #2532 

Animate dragged thoughts with `moveThoughtOver`.
Introduce a new hook, `useDragThoughtAnimation`, to apply the animation only to the currently dragged thought(s). 
The drag animation composes in `TreeNode` and takes precedence when active.


zukilover:
<details>
<summary>Drag single thought</summary>

https://github.com/user-attachments/assets/41bf8c60-26bf-4712-b56f-33edb1eb5020


</details>

<details>
<summary>Drag multiple thoughts</summary>

https://github.com/user-attachments/assets/c85009b0-463e-46be-bb2b-8b10e61c22ce


</details>

src/components/TreeNode.tsx
```diff
@@ -163,6 +166,14 @@ const TreeNode = ({
   )
 
   const moveDivStyle = useMoveThoughtAnimation(index)
+  // True if this TreeNode represents a thought in the current drag selection
+  const isDragged = useSelector(
+    state =>
+      state.longPress === LongPressState.DragInProgress &&
+      (state.draggingThoughts || []).some(p => head(p) === thoughtId),
+  )
```
raineorshine: Can we move this into `useDragThoughtAnimation` and instead pass the `thoughtId` as an argument?

I don't see a reason to keep this out of the custom hook, which already has a dependency on the Redux state. Is there something I'm missing? 

I'm responding to the general rule to push complexity *down* the stack.

src/hooks/useDragThoughtAnimation.ts
```diff
@@ -0,0 +1,59 @@
...
+    )
+    return actionsInPatch.has('moveThought')
+  })
+
+  // Track the previous on-screen index so we only animate on the first frame after an actual position change.
+  // Without prevIndexChanged guard, quick consecutive renders could retrigger the animation.
+  const previousIndex = usePrevious<number>(index)
+  const indexChanged = previousIndex !== undefined && previousIndex !== index
+  const prevIndexChanged = usePrevious(indexChanged)
+  const hasMovedDnD = !!(isLastActionMoveThought && isDragged && indexChanged && !prevIndexChanged)
```
raineorshine: This and other logic here is very similar to `useMoveThoughtAnimation`. Are you sure you don't want to put them in the same hook? It seems like there may be some duplication we can avoid. However, please let me know if you other reasoning.

src/hooks/useDragThoughtAnimation.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+const useDragThoughtAnimation = (index: number, isDragged: boolean): React.CSSProperties | undefined => {
+  // Detect if the last patch included a drag-and-drop moveThought.
+  // Aggregate actions across the whole patch to handle merged multicursor operations.
+  const isLastActionMoveThought = useSelector((state: State): boolean => {
+    const lastPatches = state.undoPatches[state.undoPatches.length - 1]
+    const actionsInPatch = new Set(
+      (lastPatches ?? []).flatMap(op => (op as unknown as { actions?: string[] }).actions || []),
+    )
+    return actionsInPatch.has('moveThought')
```
raineorshine: Instantiating a Set is unnecessary if we're just checking if any operation in the patch contains `moveThought`. Maybe you use a nested `some`?

raineorshine (changes requested):
Thank you!

src/hooks/useDragThoughtAnimation.ts
zukilover: I initially implemented this within the same hook but I wanted to avoid cluttering it for the drag and drop animation.

src/components/TreeNode.tsx
zukilover: That makes sense. I’ve updated the hook to take `thoughtId` so it stays self-contained and keeps `TreeNode` clean.

src/hooks/useDragThoughtAnimation.ts
zukilover: I've updated it with `some`.

zukilover:
I’m consistently hitting a different e2e error locally than what shows up in CI/CD:

```
Error: Test timed out in 5000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
⎯⎯⎯⎯⎯
```

I'm looking into this.

src/hooks/useDragThoughtAnimation.ts
raineorshine: Sure, but that's a lot of duplicated logic. I'm not sure that's the best tradeoff long term.

raineorshine:
> I’m consistently hitting a different e2e error locally than what shows up in CI/CD:
> 
> ```
> Error: Test timed out in 5000ms.
> If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
> ⎯⎯⎯⎯⎯
> ```
> 
> I'm looking into this.

Thanks! Feel free to reach out to @karunkop for support if the test suite is behaving unexpectedly.

zukilover:
@karunkop, I see you fixed the e2e test I mentioned above in 86ef2b680ef434c185fdc26a549e36950f712795.
Could you also check the other failed tests in this PR? I ran them locally and they all passed for me.

zukilover:
@karunkop yeah that's what I've updated in https://github.com/cybersemics/em/pull/3204/commits/313c1300d71385126e9c8cac5ab4da9a04a03612

It seems to be affecting `dropPosition` in the test, but I'm not sure why.

zukilover:
There seems to be discrepancies between local and CI/CD. As I mentioned, the tests all passed for me locally.

karunkop:
> There seems to be discrepancies between local and CI/CD. As I mentioned, the tests all passed for me locally.

Flaky tests often pass consistently in local environments but can fail intermittently in CI. In this case, the issue seems to be more persistent in CI. I suggest investigating why the recent changes in this PR affected the cursor position after drop. Thanks!

raineorshine:
@karunkop If you could test if this branch passes for you locally, that would be helpful to confirm what @zukilover is experiencing. If so, then maybe you can provide some ideas for troubleshooting since it's hard when the issue cannot be reproduced locally.

src/hooks/useDragThoughtAnimation.ts
zukilover: Let me know if you prefer using a shared hook, or if you’d rather fully integrate the logic into `useMoveThoughtAnimation`.

Here’s how the shared hook would look:

```js
const useMoveAnimationBase = (
  index: number,
  shouldAnimate: boolean,
  getAnimationName: (previousIndex: number, currentIndex: number) => 'moveThoughtOver' | 'moveThoughtUnder' | null,
): React.CSSProperties | undefined => {
  const previousIndex = usePrevious<number>(index)
  const indexChanged = previousIndex !== undefined && previousIndex !== index
  const prevIndexChanged = usePrevious(indexChanged)
  const hasMoved = !!(shouldAnimate && indexChanged && !prevIndexChanged)

  const [moveAnimation, setMoveAnimation] = useState<'moveThoughtOver' | 'moveThoughtUnder' | null>(null)

  // Throttle clear to avoid stacking multiple clears during rapid moves
  const clearMoveAnimation = useMemo(
    () => throttle(() => setMoveAnimation(null), durations.get('layoutNodeAnimation'), { leading: false }),
    [],
  )

  useLayoutEffect(() => {
    if (!hasMoved || previousIndex === undefined) return
    const animationName = getAnimationName(previousIndex, index)
    if (animationName) {
      setMoveAnimation(animationName)
      clearMoveAnimation()
    }
  }, [hasMoved, previousIndex, index, clearMoveAnimation, getAnimationName])

  return useMemo<React.CSSProperties | undefined>(() => {
    if (!moveAnimation) return undefined

    const base: React.CSSProperties = {
      transformOrigin: 'left',
      animationDuration: `${durations.get('layoutNodeAnimation')}ms`,
      animationTimingFunction: 'ease-out',
      animationFillMode: 'none',
    }

    if (moveAnimation === 'moveThoughtOver') {
      return { ...base, animationName: 'moveThoughtOver', zIndex: 2 }
    }
    if (moveAnimation === 'moveThoughtUnder') {
      return { ...base, animationName: 'moveThoughtUnder' }
    }
    return undefined
  }, [moveAnimation])
}

export default useMoveAnimationBase

```

raineorshine: Hi, thanks for your reply. I'm pretty sure a single hook makes the most sense here. Conceptually, the animations for move and drag-and-drop both capture the same phenomenon: avoiding visual "collisions" between thoughts that are moving positions. There are slight variations to capture the different emphasis of active vs passive movements, but they are all part of the same effect. The logic of detecting the changed index is identical, so should be consolidated.

I can't see a reason to separate these, which would produce duplicate code and obscure the commonalities.

> Let me know if you prefer using a shared hook, or if you’d rather fully integrate the logic into useMoveThoughtAnimation.

If I'm understanding correctly, the idea of a shared hook would result in a base hook and two custom hooks that use it? I think this level of abstraction probably isn't justified. It adds additional complexity without much benefit. I think it would be better to put all of the functionality in `useMoveThoughtAnimation`.

zukilover:
@raineorshine, I have put all of the logics in `useMoveThoughtAnimation` hook. Kindly review.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -56,15 +57,26 @@ const useMoveThoughtAnimation = (
...
     const lastPatches = state.undoPatches[state.undoPatches.length - 1]
 
+    /** Utility function to check if a given action type appears in any op.actions in the last patch. */
+    const hasAction = (type: string) =>
+      (lastPatches ?? []).some(op => (op as unknown as { actions?: string[] }).actions?.some(a => a === type))
+
+    // For drag-and-drop moveThought, animate only dragged thoughts
+    const isMoveThought = hasAction('moveThought')
+    const isDragged = !!thoughtId && (state.draggingThoughts || []).some(p => head(p) === thoughtId)
+    if (isMoveThought && isDragged) return 'dnd' as const
```
raineorshine: `lastMoveAction` makes sense when it reflects an actual action type, i.e. `'moveThoughtUp' | 'moveThoughtDown' | null`. Adding a dummy type `'dnd'` throws off the semantics. The type `'moveThoughtUp' | 'moveThoughtDown' | 'dnd' | null` is not very coherent, as it no longer refers to an action type, but a kind of ad hoc union type. 

Does `lastMoveAction` have to know about dragging? Maybe this logic would be better off in a separate `useSelector` and we can avoid dummy types.

raineorshine (changes requested):
Thanks! This is looking good.

src/hooks/useMoveThoughtAnimation.ts
zukilover: I removed the ‘dnd’ variant. Drag detection is handled in a separate selector, so no dummy types.

src/hooks/useMoveThoughtAnimation.ts
```diff
@@ -112,15 +125,25 @@ const useMoveThoughtAnimation = (
   }, [setMoveAnimation])
 
   useLayoutEffect(() => {
```
raineorshine, line 127: The control flow within this effect is structured somewhat oddly:

1. Set the animation and return
2. Short circuit
3. Set the animation and return

It's more conventional to short circuit at the beginning of a function and then consolidate duplicate logic such as setting the animation. Is there any way to streamline this?

I'm thinking something like:

```
const moveAnimationName = hasMovedDnD 
  ? 'moveThoughtOver' 
  : getMoveAnimation(lastMoveAction, previousIndex, index)
```

raineorshine (changes requested):
Thank you!

Looks like the `durations` fix got lost in the latest revision, and that's why the Puppeteer test is failing. See: #3216.

src/hooks/useMoveThoughtAnimation.ts
zukilover: Thanks for the suggestion. I tried the one-liner, but `getMoveAnimation` was still firing in cases where I wanted to short-circuit (e.g. when hasMoved is false or the action isn’t up/down). I ended up keeping the guards but consolidating the animation logic into one place. It works well and still reads cleaner.

raineorshine: Yes, that looks better! Thank you!

PR #3208, opened 2025-09-01
Add bias to MultiGesture, Render perpendicular segments in GestureDiagram 20 degrees more acute

### LINKED ISSUE #1983: 📱 Favor perpendicular gesture over opposite gesture
## Steps to Reproduce

Swipe right and then 45 degrees down-left. 

## Current Behavior

The gesture is inconsistent or cancelled. 

The gesture oscillates between ↓ and ← along the 45 degrees line, creating gestures like →↓←↓←↓←↓… instead of →↓. 

## Expected Behavior

1. The gesture should consistently register as →↓ (New Thought) when the second swipe is within 65 degrees of straight down.

This should apply to all perpendicular gestures, i.e.:

After ↓/↑ : Favor ←/→
After ←/→: Favor ↑/↓

~~2. The GestureDiagram should be updated so perpendicular gestures are rendered 20 degrees more acute.~~

## Notes

In order to achieve this, we effectively move the dividing line where the direction is perfectly ambiguous between ← and ↓ from 45-45 to 25-65. 

This is calculated on touchmove so is very performance sensitive. There may be a more efficient heuristic that can be used over `Math.sin`.

ethan-james:
Fixes #1983

`MultiGesture`: Calculate the new 25/65 degree bias for both horizontal and vertical gesture segments. Switch between biased or unbiased mappings based on the direction of the previous segment.

`GestureDiagram`: New logic in `pathSegmentDelta` for drawing path segments that are perpendicular to the previous segment at an angle 20 degrees more acute than a right angle. Ignore perpendicular segments that come after other perpendicular segments so that we don't lose 20 degrees repeatedly and draw a weird lightning bolt. Also ignore `rddl` because it's a special shape that doesn't look right if it is skewed.

### DIFF
--- a/src/components/GestureDiagram.tsx
+++ b/src/components/GestureDiagram.tsx
@@ -133,8 +133,11 @@ const GestureDiagram = ({
     // Don't skew a line segment that comes immediately after a skewed segment.
     // In the case of multiple perpendicular segments in a row, only skew every other segment.
     // Also don't skew 'rddl'.
-    const skew = path !== 'rddl' && path !== 'lddr'
-    isPerpendicular(dir, prev) && (!isPerpendicular(prev, beforePrev) || isPerpendicular(beforePrev, pathDirs[i - 3]))
+    const skew =
+      path !== 'rddl' &&
+      path !== 'lddr' &&
+      isPerpendicular(dir, prev) &&
+      (!isPerpendicular(prev, beforePrev) || isPerpendicular(beforePrev, pathDirs[i - 3]))
 
     // If the new segment is perpendicular to the previous segment, then special cases like reversals don't apply.
     // Draw the new segment 20 degrees more acute than an otherwise perpendicular line. (#1983)
--- a/src/components/GestureDiagram.tsx
+++ b/src/components/GestureDiagram.tsx
@@ -128,14 +128,13 @@ const GestureDiagram = ({
     const next = pathDirs[i + 1]
     const afterNext = pathDirs[i + 2]
     const horizontal = dir === 'l' || dir === 'r'
+    const path = pathDirs.join('')
 
     // Don't skew a line segment that comes immediately after a skewed segment.
     // In the case of multiple perpendicular segments in a row, only skew every other segment.
     // Also don't skew 'rddl'.
-    const skew =
-      pathDirs.join('') !== 'rddl' &&
-      isPerpendicular(dir, prev) &&
-      (!isPerpendicular(prev, beforePrev) || isPerpendicular(beforePrev, pathDirs[i - 3]))
+    const skew = path !== 'rddl' && path !== 'lddr'
+    isPerpendicular(dir, prev) && (!isPerpendicular(prev, beforePrev) || isPerpendicular(beforePrev, pathDirs[i - 3]))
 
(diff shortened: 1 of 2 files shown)


src/components/GestureDiagram.tsx
```diff
@@ -123,6 +128,45 @@ const GestureDiagram = ({
...
+    // Don't skew a line segment that comes immediately after a skewed segment.
+    // In the case of multiple perpendicular segments in a row, only skew every other segment.
+    // Also don't skew 'rddl'.
+    const skew =
+      path !== 'rddl' &&
+      path !== 'lddr' &&
+      isPerpendicular(dir, prev) &&
+      (!isPerpendicular(prev, beforePrev) || isPerpendicular(beforePrev, pathDirs[i - 3]))
+
+    // If the new segment is perpendicular to the previous segment, then special cases like reversals don't apply.
```
raineorshine: Would the reversal logic need to change significantly to make the skew case work without short circuiting?

Generally I try to avoid multiple return statements in the same function, as it creates additional logical paths to have to follow. It's better if the default logic shakes out as usual.

I have some concern that `dir === 'l' ? -1 : 1` is subtly duplicating some of the reversal/flipOffset logic below, by bypassing it and making it concrete. 

Is there a general solution here? Does it involve just a few more branching cases in the below logic? Or is the skew logic different enough to justify short circuiting and branching the function into two different codepaths that must be analyzed and maintained separately?

It's hard for me to know at a glance, so I'm interested to hear your opinion. Thanks!

raineorshine (commented):
Thanks! 

Please update the GestureDiagram snapshot.

src/components/GestureDiagram.tsx
ethan-james: Sure, if we're keeping the main (adjacent arm) length the same, then it becomes much easier to fit the skew (opposite arm) into the existing calculation. I can turn

```
(reversal ? reversalOffset! : 0) * (flipOffset ? -1 : 1)
```

into something like

```
(reversal ? reversalOffset! : skew ? Math.sin(Math.PI / 9) * size : 0) * (flipOffset ? -1 : 1)
```

raineorshine: The overall diagram height of `rdr` should be the same as the height of `rdl`, but because `d` in the former is skewed, it needs to be longer than that in the latter. This was what you pointed out earlier with the length needing to be adjusted.

JoeDuong-dev:
@ethan-james Quick clarification on this ticket. While reviewing https://github.com/cybersemics/em/issues/1983, I wanted to confirm whether we also need to test scenarios where the swipe-right gestures → are within 45 degrees angle to a horizontal line.

JoeDuong-dev (approved):
@ethan-james Thank you for the clarification. Based on your explanation and the details in [issue #1983](https://github.com/cybersemics/em/issues/1983), I’ve been testing variations of the →↓ gesture (New Thought) along with other perpendicular gestures. The gestures are correctly and consistently recognized, following the rules:
	•	After ↓/↑: favor ←/→
	•	After ←/→: favor ↑/↓
	•	Along with the 45° / 65° angle rule.
	
Tested on Safari (iOS).

PR #3211, opened 2025-09-02
Refactor BulletCursorOverlay using Clone DOM Approach

### LINKED ISSUE #3162: Improvement plan for BulletCursorOverlay
### Context

cindmichelle:
Fix #3162 

### with saving cursor thought ref in ministore approach 

https://github.com/user-attachments/assets/3e43e17c-3098-4851-9208-6cf87b632ca6

---
### indicate cursor thought with `data-cursor=true` DOM attribute approach (latest)

https://github.com/user-attachments/assets/db1466e6-65ac-4dd4-aa9c-247285b73707

### DIFF
--- a/src/components/BulletCursorOverlay.tsx
+++ b/src/components/BulletCursorOverlay.tsx
@@ -9,6 +9,7 @@ import attributeEquals from '../selectors/attributeEquals'
 import { findAnyChild, getAllChildren } from '../selectors/getChildren'
 import getThoughtById from '../selectors/getThoughtById'
 import isContextViewActive from '../selectors/isContextViewActive'
+import rootedParentOf from '../selectors/rootedParentOf'
 import simplifyPath from '../selectors/simplifyPath'
 import head from '../util/head'
 import isDivider from '../util/isDivider'
@@ -18,7 +19,7 @@ const isIOSSafari = isTouch && isiPhone && isSafari()
 const bulletOverlayRadius = isIOSSafari ? 300 : 245
 
 // debounce delay for syncing the placeholder to the live cursor node
-const CURSOR_NODE_UPDATE_DEBOUNCE = 40
+const CURSOR_NODE_UPDATE_DEBOUNCE = 80
 
 /**
  * BulletCursorOverlway is a component used to animate the cursor overlay from the bullet.
@@ -28,6 +29,11 @@ export default function BulletCursorOverlay({ env }: { env?: LazyEnv }) {
   const containerRef = React.useRef<HTMLDivElement>(null)
 
   const cursor = useSelector(state => state.cursor)
+
+  // Check if this is a table view once
+  const isTableView = useSelector(state =>
+    cursor ? attributeEquals(state, head(rootedParentOf(state, cursor)), '=view', 'Table') : false,
+  )
 
   // Get required data for useHideBullet hook
   const { simplePath, thoughtId, childrenThoughts, isInContextView } = useSelector(state => {
@@ -141,7 +147,7 @@ export default function BulletCursorOverlay({ env }: { env?: LazyEnv }) {
   // trigger debounced update on dependencies change
   React.useEffect(() => {
     debouncedUpdate()
-  }, [debouncedUpdate, cursor])
+  }, [debouncedUpdate, cursor, isTableView])
 
   if (bulletIsDivider || shouldHideBullet) {
     return null


src/components/BulletCursorOverlay.tsx
```diff
@@ -1,290 +1,86 @@
...
+    // Clear existing content and copy the nested child from treeNode
+    container.innerHTML = ''
+    if (treenodeElement.firstElementChild) {
+      const nestedChild = treenodeElement.firstElementChild.cloneNode(true) as HTMLElement
+      container.appendChild(nestedChild)
+
+      // Find and hide the thought-annotation element
+      const thoughtAnnotation = nestedChild.querySelector('div[aria-label="thought-annotation"]') as HTMLElement
+      if (thoughtAnnotation) {
+        thoughtAnnotation.style.visibility = 'hidden'
```
raineorshine: I'd like to better understand why you are setting cloned descendant elements to visibility `hidden` in `BulletCursorOverlay` rather than in `TreeNode` when the node is cloned. It seems like something to do during cloning. Could you explain?

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,290 +1,86 @@
...
+      }
+
+      const staticThought = nestedChild.querySelector('div[aria-label="static-thought"]') as HTMLElement
```
raineorshine: This is one of many tight couplings to the DOM. I'd like to consider how we might be able to reduce the surface area to make this more encapsulated and less fragile.

src/stores/cursor.ts
```diff
@@ -0,0 +1,58 @@
...
+  x?: number
+  y?: number
+  treeNode?: HTMLElement | null
+  thought?: HTMLElement | null
+  thoughtAnnotation?: HTMLElement | null
+  contextBreadcrumb?: HTMLElement | null
+}) => {
+  const oldState = ministore.getState()
+  ministore.update({
+    ...oldState,
```
raineorshine: `update` should already handle partial updates, so you don't need to merge in `...oldState`.

src/components/TreeNode.tsx
```diff
@@ -210,9 +214,28 @@ const TreeNode = ({
...
+        const svgElement = bulletSpan.querySelector('svg')
+        if (svgElement) {
+          // Remove all children of the SVG (this removes the <g> element and its contents)
+          svgElement.innerHTML = ''
+        }
+      }
+
+      cursorStore.onChange({
+        treeNode: clonedOuterDiv,
+      })
```
raineorshine: What's the advantage of going through the ministore back channel rather than appending it directly?

The ministore adds complexity, and we already have a lot of direct DOM couplings, but maybe there's something else I'm not considering.

src/components/BulletCursorOverlay.tsx
cindmichelle: I didnt set `visibility: hidden` on the TreeNode level before because it would affect everything (including the cursor overlay SVG). But I just got the idea to set `visibility: visible` back on the CursorOverlay component itself, and it seems to work. Thanks for sparking the idea and helping me explore the implementation further!

src/components/BulletCursorOverlay.tsx
cindmichelle: yep we can avoiding this pattern by implementing this comment now https://github.com/cybersemics/em/pull/3211/files#r2319893043

src/stores/cursor.ts
cindmichelle: ahh ok, i think overengineered this one 😅

src/components/TreeNode.tsx
cindmichelle: We cant append it directly (from TreeNode.tsx) because the thing we want to append is still a React element, whereas `parent.appendChild(child)` only works with DOM nodes/refs. so we use `ReactDOM.createPortal()` instead

Since portals are declarative, we place `createPortal()` inside the BulletCursorOverlay component (as it is now). The portal also needs a target to mount into, so I’m storing `targetSvg` in state inside `BulletCursorOverlay`.

For passing the cloned element around, alternatives to the ministore would be React Context or prop drilling, but functionally they’re equivalent here (just different transport mechanisms).

src/components/BulletCursorOverlay.tsx
raineorshine: Let me try to clarify. I don't mean setting `visibility: hidden` on a different part of the hierarchy, such as the `TreeNode`. I'm referring to the timing of when to remove/hide the nodes. 

In this implementation, you're cloning the tree node, propagating it through the ministore, and then hiding the parts that aren't needed. However, it seems strange to me to break these actions up. Why not clone and remove before storing? That would avoid storing unnecessary nodes and remove complexity downstream.

Usually when processing data you want to filter unnecessary items/properties out as early as possible.

raineorshine: Also, can the nodes be removed completely instead of hiding them?

src/components/TreeNode.tsx
raineorshine: I think I see. Yes, it's a bit unusual to clone a chain of ancestors and append a persistent React component to it.

Portals allow you to "push" components to other parts of the component hierarchy. Maybe it would be better to "pull" the cloned ancestors into the `BulletCursorOverlay` instead? I'm not sure, but the encapsulation is not very good right now with it spread across two files and one ministore. I'd like to see this developed a bit more.

src/components/BulletCursorOverlay.tsx
cindmichelle: At first I was only going to clone the necessary elements, but then I thought if someone adds or removes something under the TreeNode, it could affect the bullet position again. So I decided to just clone the whole thing to avoid having to adjust it manually later on. Whats your take on this trade-off?

raineorshine: Won't it be re-cloned every time?

cindmichelle: Yes, it’ll be re-cloned every time the cursor changes.

I think my explanation above was a bit confusing, so let me spell out my concern about **only cloning the necessary parts**. This also depends on how we do it, either we _pick up only what’s necessary_, or we _clone everything and then remove what’s unnecessary._

If we only pick up the necessary bits, then whenever something changes in the original components, we’d have to include that new element in the clone as well.

On the flip side, if we clone everything and try to strip out the unnecessary parts, then when someone adds a new element in the original component that actually doesnt affect the bullet position, we’d still have to manually remove it from the clone again.

So it feels like the clone only what’s necessary route adds extra maintenance overhead for BulletCursorOverlay.

I’m also curious, if we clone and includes the unnecessary elements when render, is that going to hurt performance?

src/components/TreeNode.tsx
cindmichelle: Yeah the encapsulation in this approach isnt great. 

To improve encapsulation, maybe we could add `data-cursor` attribute to the TreeNode root element, set it to `true` if it's a cursor. Then in `BulletCursorOverlay`, we can just do a `.querySelector("div[data-cursor=true]")` to find the TreeNode and kick off the clone process from there. wdyt?

src/components/BulletCursorOverlay.tsx
raineorshine: > I think my explanation above was a bit confusing, so let me spell out my concern about **only cloning the necessary parts**. This also depends on how we do it, either we _pick up only what’s necessary_, or we _clone everything and then remove what’s unnecessary._

We clone everything and then remove what's unnecessary. My question is more about _when_ to remove what's unnecessary. Do we do it immediately after the clone, or wait until the clone is rendered? I'm advocating for the former if possible.

> On the flip side, if we clone everything and try to strip out the unnecessary parts, then when someone adds a new element in the original component that actually doesn't affect the bullet position, we’d still have to manually remove it from the clone again.

I don't totally understand this. Can you explain what you mean by adding a new element? Are you talking about the user or a developer, and what kind of new element?

> I’m also curious, if we clone and includes the unnecessary elements when render, is that going to hurt performance?

I don't think it'll have a big effect on performance. It's more about trying to encapsulate this functionality better.

src/components/TreeNode.tsx
raineorshine: Yes, I'm interested in that. It's direct DOM manipulation, but we already have so much of that anyway.

If it was a clean portal with just the mounting node, that would be one thing. But there are so many DOM nodes that are being referenced for cloning and removing, I think the declarative benefits of the portal are somewhat lost anyway.

src/components/BulletCursorOverlay.tsx
cindmichelle: > We clone everything and then remove what's unnecessary. My question is more about when to remove what's unnecessary. Do we do it immediately after the clone, or wait until the clone is rendered? I'm advocating for the former if possible.
yep if we go with the _remove what’s unnecessary_ approach, we’ll remove it right after the clone.

> I don't totally understand this. Can you explain what you mean by adding a new element? Are you talking about the user or a developer, and what kind of new element?

All good, let me explain in a bit more detail. Right now the TreeNode component hierarchy looks like this:
```
└─TreeNode
    └─VirtualThought
      └─Subthought
        └─Thought
          ├─ContextBreadcrumbs
          ├─Bullet
          └─StaticThought
            ├─ThoughtAnnotation
            └─Editable
 ```          

Based on the BulletCursorOverlay with the placeholder approach, the components we need to accurately mimic the bullet position are just `TreeNode`, `Thought`, `ContextBreadcrumb`, `Bullet`, and `ThoughtAnnotation`.

If we use the **remove what’s unnecessary** approach, the simplest DOM structure we copy would look like hierarchy below. (We could also remove the DOM elements for VirtualThought, Subthought, and StaticThought, but I think that would make the process more complicated.)
```
  └─TreeNode
    └─VirtualThought
      └─Subthought
        └─Thought
          ├─ContextBreadcrumbs
          ├─Bullet
          └─StaticThought
           └─ ThoughtAnnotation
```

But if we go with the approach above, then if we change the TreeNode component hierarchy in the future, we’ll need to manually eliminate those elements again. For example, here’s what the hierarchy would look like after adding a new component that doesn’t affect the bullet position
```
└─TreeNode
    └─VirtualThought
      └─Subthought
        ├─NewElementOne
        └─Thought
          ├─ContextBreadcrumbs
          ├─Bullet
          ├─NewElementTwo
          └─StaticThought
            ├─ThoughtAnnotation
           └─Editable
```

We could skip manual removals and let new unnecessary elements be cloned too. but if we’re doing that, we might as well deep-clone from the start imo

Comment (raineorshine) [reply src/components/BulletCursorOverlay.tsx]: I would suggest going with the cleanest approach, with the fewest couplings to specific DOM elements. The hierarchy is expected to change very little on the future.

**Comment by cindmichelle** [comment]
Hi @raineorshine, just updated this PR with direct manipulation approach (querying the DOM with `data-cursor=true` to get the latest cursor node). 

Right now I’m using 2× requestAnimationFrame to delay the querying DOM elements before the `data-cursor` attr switches to the new cursor thought, and also to capture the final position when in table view. 

For Android and iOS, I decided to add one extra rAF since I noticed a slight position shift when the virtual keyboard appears, which makes the position without the keyboard invalid. I also saw a similar issue when switching from normal view to table view, where the position was incorrect. So I added one more rAF to ensure DOM cloning happens after the browser knows whether the virtual keyboard has appeared, and to fix the table view transition issue.


https://github.com/user-attachments/assets/db1466e6-65ac-4dd4-aa9c-247285b73707

File: src/components/BulletCursorOverlay.tsx [hunk 5]
```diff
@@ -1,375 +1,157 @@
...
 
 const isIOSSafari = isTouch && isiPhone && isSafari()
+const bulletOverlayRadius = isIOSSafari ? 300 : 245
 
-/** Returns true if two lists of children are equal. Deeply compares id, value, and rank. */
-const equalChildren = (a: Thought[], b: Thought[]) =>
-  a === b ||
-  (a && b && a.length === b.length && a.every((thought, i) => equalThoughtRanked(a[i], b[i]) && a[i].id === b[i].id))
+// debounce delay for syncing the placeholder to the live cursor node
+const CURSOR_NODE_UPDATE_DEBOUNCE = 80
```
raineorshine, line 22: Is there impact on the UX with this debounce? Is this a necessary compromise for the DOM cloning approach compared to the React approach?

It may be worth recording a video of this compared to `main` and stepping through frame-by-frame to confirm.

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
-              fillOpacity: 0.25,
-              fill: 'fg',
-            })}
-          />
-        </g>
-      </svg>
-    </span>
+  // Check if this is a table view once
+  const isTableView = useSelector(state =>
+    cursor ? attributeEquals(state, head(rootedParentOf(state, cursor)), '=view', 'Table') : false,
```
raineorshine, line 35: Rather than having closure over a separate `cursor` variable, access `state.cursor` directly in this `useSelector`. This will be more self-contained.

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
+    }
+  }, [cursor])
+
+  // debounced caller that runs afterNextPaint
+  const debouncedUpdate = useDebouncedLatest(updateCursorNode, CURSOR_NODE_UPDATE_DEBOUNCE)
+
+  // trigger debounced update on dependencies change
+  React.useEffect(() => {
+    debouncedUpdate()
+  }, [debouncedUpdate, cursor, isTableView])
```
raineorshine, line 150: 1. Import `useEffect` and use without `React.` for consistency across the project.
2. `() => { a() }` is functionally equivalent to `a`, i.e. you can drop the wrapper and use point-free style:

```
`useEffect(debouncedUpdate, [...])`
```

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
+        svg.innerHTML = svgInnerHTML
+
+        svg.style.visibility = 'visible'
+      }
+      prevDomElementRef.current = cursorTreeNode
+    }
+  }, [cursor])
+
+  // debounced caller that runs afterNextPaint
+  const debouncedUpdate = useDebouncedLatest(updateCursorNode, CURSOR_NODE_UPDATE_DEBOUNCE)
```
raineorshine, line 145: Let's call this `updateCursorNodeDebounced`

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
-    left: x,
-    top: y,
-    // Table col1 uses its exact width since cannot extend to the right edge of the screen.
-    // All other thoughts extend to the right edge of the screen. We cannot use width auto as it causes the text to wrap continuously during the counter-indentation animation, which is jarring. Instead, use a fixed width of the available space so that it changes in a stepped fashion as depth changes and the word wrap will not be animated. Use x instead of depth in order to accommodate ancestor tables.
-    // 1em + 10px is an eyeball measurement at font sizes 14 and 18
-    // (Maybe the 10px is from .content padding-left?)
-    width: isTableCol1 ? width : `calc(100% - ${x}px + 1em + 10px)`,
-    textAlign: isTableCol1 ? ('right' as const) : undefined,
-  }
+    return { simplePath: path, thoughtId: id, childrenThoughts: children, isInContextView: contextView }
```
raineorshine, line 51: Since you are returning an object literal, this will return a new object reference every time, regardless of whether any of the values have changed, i.e. `({}) !== ({})`. This has the unfortunate effect of causing the component to re-render on every `State` change.

To do a deep equal comparison and only re-render when a value changes, pass lodash's `isEqual` as the equality function:

```
useSelector(fn, isEqual)
```

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
+        })
+      : false
+
+  const prevDomElementRef = React.useRef<HTMLElement | null>(null)
+
+  /** Updates the placeholder node to mirror the current cursor node and target its bullet svg. */
+  const updateCursorNode = useCallback(() => {
+    if (!containerRef.current || !cursor) return
+
+    // Find the cursor TreeNode using the data-cursor attribute
```
raineorshine, line 104: A bit more explanation needed here to explain why two `data-cursor` nodes would exist in the DOM at the same time.

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
+    if (cursorTreeNode.firstElementChild) {
+      const nestedChild = cursorTreeNode.firstElementChild.cloneNode(true) as HTMLElement
+
+      container.appendChild(nestedChild)
+
+      const svg = nestedChild.querySelector('span[aria-label="bullet"] svg') as SVGSVGElement | null
+      if (svg) {
+        // Replace the SVG contents
+
+        const svgInnerHTML = `<g style="visibility: visible; will-change: transform;"><ellipse ry="${bulletOverlayRadius}" rx="${bulletOverlayRadius}" cy="300" cx="300" class="stk_highlight fill-opacity_0.25 fill_fg"></ellipse></g>`
```
raineorshine, line 135: Let's assign this directly rather than creating a separate variable for it.

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
+    const cursorTreeNode =
+      cursorTreeNodes.length > 1
+        ? Array.from(cursorTreeNodes).find(node => node !== prevDomElementRef.current)
+        : cursorTreeNodes[0]
+
+    if (!cursorTreeNode) return
+
+    const container = containerRef.current
+
+    // Copy all attributes from the cursor TreeNode to the container (except data-cursor)
```
raineorshine, line 116: Can you add a comment explaining which attributes are copied and why this is necessary?

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
+    Array.from(cursorTreeNode.attributes).forEach(attr => {
+      if (attr.name !== 'data-cursor') {
+        container.setAttribute(attr.name, attr.value)
+      }
+    })
+
+    // Clear existing content and copy the nested child from cursor TreeNode
+    container.innerHTML = ''
+    container.style.visibility = 'hidden'
```
raineorshine, line 125: What is the reason for setting visibility to `hidden` on the container?

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
+      container.appendChild(nestedChild)
+
+      const svg = nestedChild.querySelector('span[aria-label="bullet"] svg') as SVGSVGElement | null
+      if (svg) {
+        // Replace the SVG contents
+
+        const svgInnerHTML = `<g style="visibility: visible; will-change: transform;"><ellipse ry="${bulletOverlayRadius}" rx="${bulletOverlayRadius}" cy="300" cx="300" class="stk_highlight fill-opacity_0.25 fill_fg"></ellipse></g>`
+        svg.innerHTML = svgInnerHTML
+
+        svg.style.visibility = 'visible'
```
raineorshine, line 138: What is the reason for setting visibility to `visible` on the svg?

src/components/BulletCursorOverlay.tsx
```diff
@@ -1,375 +1,157 @@
...
+          env,
+          hideBulletProp: hideBulletAttr,
+          isEditing: true, // keep this as true since cursor overlay only works when editing is true
+          simplePath,
+          isInContextView,
+          thoughtId,
+        })
+      : false
+
+  const prevDomElementRef = React.useRef<HTMLElement | null>(null)
```
raineorshine, line 98: Let's add a JSDOC explaining the purpose of this ref.

src/components/BulletCursorOverlay.tsx
cindmichelle: Here's the result between this branch with debounce 80ms vs. main branch

**this branch with debounce 80ms**
*cursor stops in the last thought at 0:13

https://github.com/user-attachments/assets/3b449dfc-4bc8-42b0-a88e-a862d7a3b1b2

**main branch**
* cursor stops in the last thought at 0:10

https://github.com/user-attachments/assets/6239f1bd-97ce-4f1a-a907-bc817506d2c0

Based on the video above, it seems this branch take 3s longer to move through all the available thoughts

cindmichelle: this is when i try to lower down the debounce timeout to be 50ms. looks much better i think than the 80ms one.

https://github.com/user-attachments/assets/79599456-bc4e-4b66-bbd6-034804d05c24

@raineorshine do you think this is still acceptable? fyi, the videos were recorded at 30 fps, so it should appear smoother when running locally.

PR #3212, opened 2025-09-02
Upgrade @capacitor/*

codebuild-ro:
I did major version updagrades for all `@capacitor/*` packages from v6 -> v7.
And I replaced `redux-devtools-extension` with `@redux-devtools/extension`.

### DIFF
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -2,8 +2,8 @@
 
 android {
   compileOptions {
-      sourceCompatibility JavaVersion.VERSION_17
-      targetCompatibility JavaVersion.VERSION_17
+      sourceCompatibility JavaVersion.VERSION_21
+      targetCompatibility JavaVersion.VERSION_21
   }
 }
 


raineorshine:
Thank you! Can you confirm that Capacitor still builds, and were you able to test on iPhone or Android?

https://github.com/cybersemics/em#native-app-development

@JoeDuong-dev This is going to take some more setup (XCode) to test properly.

raineorshine:
Thanks. Could you clarify, did you test the Capacitor build on Android only, or both iPhone and Android?

raineorshine:
> Ah, yes, it seems so :). Sorry to make confusion. So I mean I couldn't test on iPhone regarding capacitor. But at least, I could test on Android. So Capacitor for:
> 
> * **Android** - works
> * **iPhone** - unkown
> 
> > Also I tested on iPhone and Android enviroment, and also worked fine.
> 
> This meant the web app, not the mobile app.

Thank you, I understand now!

JoeDuong-dev:
@codebuild-ro Having trouble running `yarn cap:ios` for iOS testing. Android works fine. I’ll dig into it later today and update here.

codebuild-ro:
I updated `vite-plugin-terminal` doesn't support vite 7.
So I upgraded vite to 6 and vitest to 3.

JoeDuong-dev:
In Xcode, under App → Signing & Capabilities, there is an issue with the Team and Bundle Identifier configuration. This will likely prevent the app from running on a physical device, since Apple requires a valid provisioning profile that matches both the app’s Bundle ID and the associated developer account.

Screenshot 2025-09-05 at 3 55 26 PM

Still having issue running the project on Xcode simulator

I ran 
pod install
npx @capacitor/cli sync
npx @capacitor/cli open ios

Then: npx @capacitor/cli open ios to open and build the project in Xcode. Currently downloading iOS 18.5 to see if it will run.

raineorshine (changes requested):
> In Xcode, under App → Signing & Capabilities, there is an issue with the Team and Bundle Identifier configuration. This will likely prevent the app from running on a physical device, since Apple requires a valid provisioning profile that matches both the app’s Bundle ID and the associated developer account.

Apparently we need a different configuration, but I'm going to need some help here because it works on my machine and I'm not sure how to make it work for the whole team.




> Besides that, I'm still having issue running the project on Xcode simulator
> 
> I ran `pod install` `npx @capacitor/cli sync`
> 
> Then: `npx @capacitor/cli open ios` to open and build the project in Xcode. Currently downloading iOS 18.5 to see if it will run.

You'll want to follow the instructions and use `yarn cap:ios`, as described in the README instructions. I linked to them [above](https://github.com/cybersemics/em/pull/3212#issuecomment-3247404645).

---

It looks like this upgrade breaks the iOS Capacitor build. We will need someone with a working XCode setup to upgrade capacitor and fix any iOS issues.

@karunkop Are you able to do this?

JoeDuong-dev:
@raineorshine How far along did you get with testing the Capacitor build on iOS? I managed to build and run it after some tinkering, but the app only shows a black screen.
Screenshot 2025-09-08 at 4 00 29 AM

PR #3213, opened 2025-09-02
Scale tall gestures

### LINKED ISSUE #3210: [Mobile] Select Between gesture overruns its row
## Steps to Reproduce

Execute →↓ gesture and hold.

## Current Behavior

The command palette opens, and Select Between (`rddl`) is rendered so that it overlaps with New Grandchild.

Image

## Expected Behavior

The row should have enough vertical space so that any gesture is separated from its neighbors.

ethan-james:
Fixes #3210

The wrinkle that makes this difficult is that, depending on the height of the text in the gesture's row, it may or may not be necessary to scale tall gestures. The "Customize Toolbar" table has plenty of room, while the command palette does not.

Therefore, I introduced the `scaleTallGestures` flag to let the parent control whether or not to attempt scaling. I also introduced a static `0.75` scaling factor because that looks pretty good in this case.

It may be worthwhile to revisit the [cell](https://github.com/ethan-james/em/blob/e3d4d046b638d00dd8895ca46b0250ff239376ab/src/components/CommandItem.tsx#L301) where the command palette gesture diagram resides, since a 32px by 32px cell with an absolutely-positioned 70px diagram in it is limiting our flexibility. However, for solving a one-off problem where one or two gestures need to be scaled, this is pretty straightforward and effective.

There are potentially some alignment quirks. It looks like it right-justifies pretty well in most cases, but it looks good next to some gestures, and less good next to others like "New Grandchild".

image0 (4)
image1

### DIFF
--- a/src/components/CommandItem.tsx
+++ b/src/components/CommandItem.tsx
@@ -302,6 +302,7 @@ const CommandItem: FC<{
         >
           <GestureDiagram
             color={disabled ? token('colors.gray') : undefined}
+            scaleTallGestures
             styleCancelAsRegularGesture
             highlight={gestureHighlight}
             path={command.id === 'cancel' ? null : gestureString(command)}
--- a/src/components/GestureDiagram.tsx
+++ b/src/components/GestureDiagram.tsx
@@ -30,6 +30,9 @@ interface GestureDiagramProps {
   cssRaw?: SystemStyleObject
   /** Whether to render the gesture with rounded corners. */
   rounded?: boolean
+  /** The command palette doesn't have enough room for tall gestures at their natural height, but Customize Toolbar does.
+   * Allow the implementation to specify whether tall gestures should be scaled (#3210). */
+  scaleTallGestures?: boolean
   /** If true, the cancel gesture will have the same styling as the other gestures. Otherwise, there are additional sizing and margin styles applied. */
   styleCancelAsRegularGesture?: boolean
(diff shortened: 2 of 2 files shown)


raineorshine (changes requested):
I think I'm looking for a bit more of a general solution that scales based on the total extent of the gesture. ALL gestures should be rendered at the same size, regardless of where they are rendered.

If you could consider what a more general solution would entail and let me know what design questions come up, I'm happy to discuss!

ethan-james:
> I think I'm looking for a bit more of a general solution that scales based on the total extent of the gesture. ALL gestures should be rendered at the same size, regardless of where they are rendered.
> 
> If you could consider what a more general solution would entail and let me know what design questions come up, I'm happy to discuss!

I think a more general solution would involve a small amount of padding around the edge to prevent any gesture from fully filling a 70px (for example) square. All else being equal, that would cause all gestures to shrink a bit from how they currently appear. Would I need to switch from a square SVG to a rectangular one to accommodate top and bottom padding that ensures a buffer between table rows? I think so, although I'm not quite seeing the full picture.

raineorshine:
I would prefer to render padding/margin in css outside the svg. It will be easier to work with the diagrams in a variety of contexts that way.

The core issue seems to be that tall diagrams are not scaling the same way as square diagrams. They should be scaled based on `max(width, height)` rather than just width. Padding might prevent the diagram from overlapping another row, but it wouldn't scale it down to be the same size as the other gestures.

PR #3214, opened 2025-09-02
GestureDiagram: Scale arrowhead and stroke width

### LINKED ISSUE #3057: GestureDiagram: Scale arrowhead correctly with size
## Steps to Reproduce

Swipe →↓ and hold.

## Current Behavior

Gesture diagram is rendered too thin, and arrowheads are hard to see:

Screenshot 2025-06-03 at 10 48 30 AM

## Expected Behavior

Proportions between arrowhead and diagram should be consistent across sizes.

Image

Screenshot 2025-06-03 at 10 48 18 AM

_Originally posted by @raineorshine in https://github.com/cybersemics/em/issues/2928#issuecomment-2934430027_

---
            
> > > I've been trying a lot of things, but if the gesture is large and thick, the arrow disappears. The thinner it is, the larger the arrow. I cannot get both the arrow and the line to be large.
> > 
> > 
> > Perhaps @trevinhofmann can help on this one if you're stuck.
> 
> @yangchristina -- Please share any other progress you might have made on this, otherwise I'll plan to take a look into the issue tomorrow (Tuesday). Thank you very much!

Thanks @trevinhofmann! Controlling the icon is about the ratios between size, strokeWidth and arrowSize. ArrowSize and strokeWidth are inversely correlated, you can look at the calculations inside `GestureDiagram`. I believe increasing `size` increases the size of the icon, but also makes things thinner, but I'm not sure if I'm remembering things right. When strokeWidth is larger, the icon will have thicker strokes, but the arrows will get smaller, until you can't see them anymore. When arrowSize is larger, the icons will get really tiny, until they be come a little dot.

_Originally posted by @yangchristina in https://github.com/cybersemics/em/issues/2928#issuecomment-2959823944_

---

More specifically, the issue started with this change: https://github.com/cybersemics/em/commit/91b9644fbdee0af7d03a5c73f1e000b8dbaa74ef#diff-c9a68e908244ceefdd99802fdfec04f786fac555aa5f3b749559bf4f9d2b64e5R114-R116

> Controlling the icon is about the ratios between size, strokeWidth and arrowSize. ArrowSize and strokeWidth are inversely correlated, you can look at the calculations inside `GestureDiagram`.

Yes, I'm seeing the same. This behavior in `GestureDiagram` seems very strange and fragile.

> I believe increasing `size` increases the size of the icon, but also makes things thinner, but I'm not sure if I'm remembering things right.

Increasing `size` does make the lines thinner and the arrow relatively larger.

> When strokeWidth is larger, the icon will have thicker strokes, but the arrows will get smaller, until you can't see them anymore.

Yep, increasing `strokeWidth` increases the thickness of the line segments but relatively decreases the arrow size.

> When arrowSize is larger, the icons will get really tiny, until they be come a little dot.

Yes again. This seems like the quickest way to "fix" the relatively small arrows, except that it also decreases the size of the entire gesture diagram. And if we then also increase `size` to fix that, it makes the lines/arrows smaller yet again.

---

@raineorshine, the `GestureDiagram` dimension calculations seem very brittle. I see three options, and I'm wondering if you have a preference for proceeding:

1) Refactor `GestureDiagram` to simplify it considerably by essentially drawing the same "image" in a square SVG based only on the gesture path and then scaling the size of the image based on a single prop (e.g. `size`). Is there a particular reason we need all of `arrowSize`, `strokeWidth`, `height`, `width`, `size`, `flexibleSize`?

2) Refactor `GestureDiagram` slightly so that the `size`, `strokeWidth`, and `arrowSize` props don't all conflict with each other. Seeing the issues mentioned above by @yangchristina and confirmed by myself, it _appears_ impossible to configure it at the sizes that we want because each prop affects the result of the others.

3) Leave the `GestureDiagram` as-is, try to revert back to the original sizes we had before, and see if we can use something else to scale the entire diagram as needed. Perhaps we can just use some CSS (with a parent container?) to scale up the width/height with that.

I'm not a huge fan of (3) because it feels a bit hacky, ignores the problems with `GestureDiagram`, and I'm not entirely sure yet if it would work. (1) would be my preference, but it's probably outside the scope of this PR and also depends on the answer to "Is there a particular reason we need...?" being "No" which might not be the case.

This is a tricky one.

_Originally posted by @trevinhofmann in https://github.com/cybersemics/em/issues/2928#issuecomment-2965331658_
          
---

Thanks for bringing this up. The short answer is no, there's no reason for the brittleness. Probably the result of early design decisions plus tweaks that got bolted on when applying the GestureDiagram to new contexts. 

_Originally posted by @raineorshine in https://github.com/cybersemics/em/issues/2928#issuecomment-2965374442_

ethan-james:
Fixes #3057

I have two unanswered questions about this approach:

1. Why does scale only need to be applied to wide gestures and not to tall gestures as well?
2. Why does 4/3 scale look so much better than 2x scale?

The SVG is drawn with path segments that are each `size` in length, and then each gesture is scaled down to a square that is `${size}px` on each side. In order to make the arrow and line size the same across different gestures, a scaling factor can be applied to them based on how wide the SVG is before it is scaled down to fit into the square.

I see discussion in #3057 about scaling down the path segments first, and then applying a uniform `arrowSize` and `strokeWidth` across the entire diagram. That also seems like a good approach, possibly more consistent than scaling the `strokeWidth` of each segment, and I think the core of that approach would be to change [M ${x} ${y} l ${segment.dx} ${segment.dy}](https://github.com/ethan-james/em/blob/2257fd3f93a3710baa2af2507277f6bc9823be30/src/components/GestureDiagram.tsx#L321) to account for the total width/height of the diagram, like:

```
M ${x} ${y} l ${segment.dx * size / sumWidth} ${segment.dy * size / sumHeight}
```

### DIFF
--- a/src/components/GestureDiagram.tsx
+++ b/src/components/GestureDiagram.tsx
@@ -258,7 +258,7 @@ const GestureDiagram = ({
     return `M ${startX} ${startY} A ${radius} ${radius} 0 0 ${sweepFlag} ${endX} ${endY}`
   }
 
-  const scale = Math.max(1, Math.max(sumWidth, sumHeight) / 100)
+  const scale = Math.max(1, (sumWidth * 2) / (size * 3))
 
   return (
     <svg


raineorshine:
> I have two unanswered questions about this approach:
> 
> 1. Why does scale only need to be applied to wide gestures and not to tall gestures as well?
> 2. Why does 4/3 scale look so much better than 2x scale?

Yeah, I can tell you that I fiddled with the scale factor manually a lot when I first built this. It's not very easy to adjust, as changing one variable seems to throw off the other.

> I see discussion in #3057 about scaling down the path segments first, and then applying a uniform `arrowSize` and `strokeWidth` across the entire diagram. That also seems like a good approach, possibly more consistent than scaling the `strokeWidth` of each segment, and I think the core of that approach would be to change [M ${x} ${y} l ${segment.dx} ${segment.dy}](https://github.com/ethan-james/em/blob/2257fd3f93a3710baa2af2507277f6bc9823be30/src/components/GestureDiagram.tsx#L321) to account for the total width/height of the diagram, like:
> 
> ```
> M ${x} ${y} l ${segment.dx * size / sumWidth} ${segment.dy * size / sumHeight}
> ```

Yes, I think that was a mistake to individually scale the elements within the SVG. It's so easy to scale SVG as a whole. Better to render each of the elements at a fixed size so everything is proportional, and then scale the diagram as a whole.

PR #3215, opened 2025-09-04
Continue parent hover highlight pulse after drop

zukilover:


zukilover:
<details>
<summary>Pulse animation</summary>

https://github.com/user-attachments/assets/d4e572db-1b08-453a-9f23-bf114b8eece7


</details>

zukilover:
@raineorshine There’s another way to do it without adding a new redux state, but it would require a mix of effect, internal state, and a custom event, which is less performant. Let me know if that works for you.

src/components/Thought.tsx
```diff
@@ -425,6 +425,9 @@ const ThoughtContainer = ({
     [style?.textDecoration],
   )
 
+  // true when this thought is the parent of a recently dropped child, to continue a slower pulse briefly
+  const isRecentlyDropped = useSelector(state => !!state.droppedPath && equalPath(state.droppedPath, simplePath))
```
raineorshine: Is `!!state.droppedPath &&` necessary? It seems like `equalPath(state.droppedPath, simplePath)` would always be false in this case anyway, since `simplePath` is always truthy.

src/actions/setDroppedPath.ts
```diff
@@ -0,0 +1,33 @@
...
+  droppedPath: path,
+})
+
+/** Action-creator to set a parent to highlight, and auto-clear after slowPulse duration. */
+export const setDroppedPathActionCreator =
+  ({ path }: { path: Path }): Thunk =>
+  dispatch => {
+    dispatch({ type: 'setDroppedPath', path })
+
+    const clearDelay = durations.get('slowPulse')
```
raineorshine: Would you mind changing this to `verySlowPulse`? `slowPulse` ended up looking too quick.

src/hooks/useDragAndDropThought.tsx
```diff
@@ -229,6 +230,12 @@ const drop = (props: ThoughtContainerProps, monitor: DropTargetMonitor) => {
 
     haptics.medium()
 
+    // set post-drop highlight on destination parent so the pulse continues briefly at a slower rate
+    const destinationParent = parentOf(props.simplePath)
+    if (destinationParent) {
```
raineorshine: `destinationParent` is always truthy.

src/hooks/useDragAndDropSubThought.ts
```diff
@@ -213,6 +214,12 @@ const drop = (props: DroppableSubthoughts, monitor: DropTargetMonitor) => {
       )
     })
 
+    // set post-drop highlight on destination parent so the pulse continues briefly at a slower rate
+    const destinationParent = rootedParentOf(getState(), getPathTo(getState(), draggedItems[0].path))
+    if (destinationParent) {
```
raineorshine: `destinationParent` is always truthy.

raineorshine (changes requested):
> @raineorshine There’s another way to do it without adding a new redux state, but it would require a mix of effect, internal state, and a custom event, which is less performant. Let me know if that works for you.

This looks fine, thanks!

There could be a performance concern with triggering a Redux state change, since all the selectors get triggered again. It would be more efficient to keep the state change localized. But by the time that `droppedPath` is reset, all of the layout animations will have completed, so I think it will be okay. The reduction in complexity is beneficial.

src/components/Thought.tsx
zukilover: Yeah, the guard’s redundant. I’ve removed it. 
I was thinking if both were ever null, `equalPath` would return true, but that can’t happen here.

src/hooks/useDragAndDropSubThought.ts
zukilover: Okay, updated.

src/hooks/useDragAndDropThought.tsx
zukilover: Okay, updated.

src/actions/setDroppedPath.ts
zukilover: Okay, updated.

JoeDuong-dev (approved):
This ticket has been tested and verified. On both Chrome (desktop) and Safari (iOS), the drag-and-drop animation correctly maintains the parent highlight in a pulsing state. The pulsing animation continues for as long as the user holds the drag-and-drop action.

PR #3216, opened 2025-09-08
Animate drag thought test

karunkop:
I have added a disableAnimations helper file to remove the animation dependency in test suite. The main reason for off position in cursor was the timing inconsistency at the time of taking snapshot when mouseUp is false.

[Here](https://github.com/karunkop/em/actions?query=branch%3Aanimate-drag-thought-test) is the result of the CI checks after the changes.

> Note: this is a draft PR. So please make the changes accordingly


raineorshine:
Thanks for your submission. 

We actually already have a mechanism for disabling animations: `durations` are automatically zero'd out during puppeteer tests. The problem is that `useDragThoughtAnimation` imports https://github.com/cybersemics/em/blob/e9ec914d33af5dd150c0d5147ee9972e041bd71c/src/durations.config.ts directly instead of using https://github.com/cybersemics/em/blob/e9ec914d33af5dd150c0d5147ee9972e041bd71c/src/util/durations.ts as it should.

PR #3219, opened 2025-09-09
Thought should not fade in after split thought on Enter

### LINKED ISSUE #2554: Thought should not fade in after split thought on Enter
## Steps to Reproduce

```
- ab
```

Set the caret in between 'a' and 'b' and hit Enter.

## Current Behavior

A new thought is created and `b` fades in.

https://github.com/user-attachments/assets/beec533d-1182-4a45-9f31-8b55a575d92f

## Expected Behavior

`b` should not fade in since the new thought was created from a split. It should snap in instantly.

zukilover:
Resolves #2554 

I added a condition in `Subthought.tsx` to skip the fade-in when splitting a thought. It sets opacity directly to the target value rather than starting at 0.

### DIFF
--- a/src/components/Subthought.tsx
+++ b/src/components/Subthought.tsx
@@ -81,6 +81,9 @@ const Subthought = ({
         findAnyChild(state, grandparentId, child => child.value === '=grandchildren')?.id) ||
       null,
   )
+  const isSplitThought = useSelector(
+    state => state.lastUndoableActionType === 'splitThought' && state.cursor && head(state.cursor) === head(simplePath),
+  )
   const hideBullet = useSelector(state => {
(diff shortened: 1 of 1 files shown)


zukilover:
<details>
<summary>split thought</summary>

https://github.com/user-attachments/assets/e5db1530-6f08-46e6-8f2a-6aeac626b317


</details>

raineorshine (approved):

raineorshine:
Thanks!

> @raineorshine While testing on Safari (mobile), I noticed that the caret cannot be moved between characters as it can on desktop. Although this is outside the scope of this ticket, I wanted to document it and confirm whether this behavior is intentional.

If you can provide steps to reproduce over Discord I'd be happy to confirm.

PR #3221, opened 2025-09-10
GestureDiagram: scale SVG by width/height or size

### LINKED ISSUE #3210: [Mobile] Select Between gesture overruns its row

ethan-james:
Fixes #3210

This approach creates a flexible container whose size is controlled by `width` and `height`, or `size` if it should be square. Currently this is a relatively small change, but it opens the door for addressing #3057 by creating fixed-size (e.g. 1px by 1px) diagrams that can have a constant `strokeWidth` and `arrowSize` applied to them without causing those sizes to appear different as the differently-sized diagrams scale to fit their containers.

### DIFF
--- a/src/components/GestureDiagram.tsx
+++ b/src/components/GestureDiagram.tsx
@@ -194,9 +194,6 @@ const GestureDiagram = ({
     }
   }
 
-  const sumWidth = Math.abs(pathSegments.reduce((accum, cur) => accum + cur.dx, 0))
-  const sumHeight = Math.abs(pathSegments.reduce((accum, cur) => accum + cur.dy, 0))
-
   /** Crop the viewbox to the diagram and adjust the svg element's height when first rendered. */
   const onRef = (el: SVGGraphicsElement | null) => {
     if (!el) return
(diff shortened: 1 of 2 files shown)


src/components/GestureDiagram.tsx
```diff
@@ -259,88 +247,91 @@ const GestureDiagram = ({
...
-                : arrowhead === 'outlined'
-                  ? 'M 0 0 L 5 5 L 0 10'
-                  : undefined
-            }
-            fill={
-              arrowhead === 'outlined'
-                ? 'none'
-                : highlight != null && highlight >= path.length
+      <svg
+        className={css(inGestureContainer && { position: 'relative', top: '10px' }, { width: '100%', height: '100%' })}
```
raineorshine, line 255: If both `maxWidth` and `maxHeight` are set and the svg stretches to fit, wouldn't this change the proportions of the diagram? Just want to make sure that this is not possible.

ethan-james: By default it preserves the aspect ratio. You have to specifically override it to get it to distort: https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/preserveAspectRatio#none

raineorshine: Great, thanks for clarifying!

PR #3225, opened 2025-09-11
Dependency minor updates

codebuild-ro:
I upgraded many modules' versions.
The important upgrading is `vitest` (from 2.1.9 to 3.0.9) and `vite` (from 5.4.19 to 6.3.6).
The latest version of `vitest` is 3.2.4 but it makes the unit test cases fails randomly by causing errors like `window is not defined` or `localstrage is not defined`.
So I updated the version with safest one which can adapt to current codebase. - it was  3.0.9.
And regarding the `vite` version, the latest one is 7.1.5 - but the latest version of `vite-plugin-terminal` was not supporting vite 7.
So I limited the `vite` version to 6.3.6.

I replaced `redux-devtools-extension` with `@redux-devtools/extension`.


codebuild-ro:
`eslint` verion is upgraded to 9. And `eslint-config-react-app` is removed from the project.
- We are using a custom flat config - Our eslint.config.js already defines all the ESLint rules and configurations we need
- It was incompatible with ESLint 9.x - eslint-config-react-app only supports ESLint 8.x (^8.0.0)
- It's redundant - We already have all the necessary ESLint plugins and rules configured manually
So we don't need that module anymore.

PR #3227, opened 2025-09-12
GestureDiagram: draw segments using a gradient

### LINKED ISSUE #2727: Gesture Diagram: Gradient
## Current Behavior

Segments of the GestureDiagram are rendered in a solid foreground color (fg):

Image

## Expected Behavior

Segments should be rendered with an opacity gradient, fading in from the start (transparent) to the end (fg).

(Ignore the other differences like the color and arrow style for now.)

Image

ethan-james:
Fixes #2727 

image0 (6)

This approach creates a separate `linearGradient` SVG element for each segment of the path, and then creates a style rule defining the start and end of color of each segment. I think that the style rule is necessary in order to use the `color-mix` CSS rule to easily transition from one color to the other, although I could do some more research if you think there's a better way.

Two things to note about this approach:

1. Rather than use a continuous gradient from 0-100% opacity over the length of the segment, I faked it by starting at 25% opacity, ramping up to 50% opacity at the end of the first path segment, and then continuing at 50% until finally increasing all the way to 100% in the final path segment. I feel like this better approximates the gradient that you posted in the issue, and it's easy to tweak to get a nicer-looking result.

2. If I actually modify the alpha transparency of the path segments, the linecaps look bad because linecaps from the previous and next segments are drawn on top of each other, effectively doubling the opacity between path segments. I think that mixing the foreground color with the background color mimics a transparency change well, but let me know if there is a chance that these will be drawn on top of anything where different colors need to show through.

Curved segments look good now. Segments with diagonal lines don't look perfect, but it's hard to tell and I think maybe they're OK. The gesture cheatsheet special case is a little bit off, but it's getting kind of messy in there and I think we should nail down the overall approach before I dive in and tweak it further.


raineorshine (changes requested):
Thanks! I love the overall approach and the thoughtfulness put into recreating the right effect from the mockups.

If you could update the Puppeteer snapshot that would be great.

> 1. Rather than use a continuous gradient from 0-100% opacity over the length of the segment, I faked it by starting at 25% opacity, ramping up to 50% opacity at the end of the first path segment, and then continuing at 50% until finally increasing all the way to 100% in the final path segment. I feel like this better approximates the gradient that you posted in the issue, and it's easy to tweak to get a nicer-looking result.

Looks good to me!

> 2\. If I actually modify the alpha transparency of the path segments, the linecaps look bad because linecaps from the previous and next segments are drawn on top of each other, effectively doubling the opacity between path segments. I think that mixing the foreground color with the background color mimics a transparency change well, but let me know if there is a chance that these will be drawn on top of anything where different colors need to show through.

The GestureDiagrams are used as overlays in the tutorial unfortunately. Though I see the challenge of using actual transparency. Let me know if you can think of a way for this to be possible.

Brave Browser 2025-09-16 14 50 37

> Segments with diagonal lines don't look perfect, but it's hard to tell and I think maybe they're OK.

Can you post an example? Thanks!

ethan-james:
> > Segments with diagonal lines don't look perfect, but it's hard to tell and I think maybe they're OK.
> 
> Can you post an example? Thanks!

Screenshot 2025-09-16 155901

Since the segments are drawn one after the other, and the gradients are specific to an individual segment, it kind of looks like it's folded over.

ethan-james:
> The GestureDiagrams are used as overlays in the tutorial unfortunately. Though I see the challenge of using actual transparency. Let me know if you can think of a way for this to be possible.

Here's how it looks in the tutorial. The gradient is not transparent, so it has the same effect as the previous version where it is only transparent when the blinking animation makes it transparent.


https://github.com/user-attachments/assets/84c33f3b-9d7a-4d55-9aed-4ee0ebb77684

raineorshine:
> Since the segments are drawn one after the other, and the gradients are specific to an individual segment, it kind of looks like it's folded over.

I'll see how it feels in use, but I think that’s fine for the initial implementation. 

> > The GestureDiagrams are used as overlays in the tutorial unfortunately. Though I see the challenge of using actual transparency. Let me know if you can think of a way for this to be possible.
> 
> 
> 
> Here's how it looks in the tutorial. The gradient is not transparent, so it has the same effect as the previous version where it is only transparent when the blinking animation makes it transparent.
> 
> 
> 
> 
> 
> https://github.com/user-attachments/assets/84c33f3b-9d7a-4d55-9aed-4ee0ebb77684
> 
> 
> 
> 

That looks fine to me!

src/components/GestureDiagram.tsx
```diff
@@ -292,8 +317,82 @@ const GestureDiagram = ({
               style={{ filter: dropShadow }}
             />
           </marker>
+          {extendedPath === 'rdld' ? (
+            <>
+              <radialGradient
```
raineorshine: Let's factor out the question mark diagram (same file, different component) as the main component is getting quite long.

src/components/GestureDiagram.tsx
```diff
@@ -247,9 +248,33 @@ const GestureDiagram = ({
...
+  }
+
+  /** Generates an SVG path string for a curved segment of the gesture.*/
+  const generateArcPath = (index: number, pathDirs: Direction[]) => {
+    const { startX, startY, radius, sweepFlag, endX, endY } = generateArcCoordinates(index, pathDirs)
     return `M ${startX} ${startY} A ${radius} ${radius} 0 0 ${sweepFlag} ${endX} ${endY}`
   }
 
+  /** Generates radial gradients for curved segments of the gesture. */
+  const generateArcGradient = (index: number, extendedPath: GesturePath) => {
```
raineorshine: This might be more conventional as a JSX component,`ArcGradient`, then consuming it as JSX. What do you think?

src/components/GestureDiagram.tsx
```diff
@@ -292,8 +317,82 @@ const GestureDiagram = ({
...
+                >
+                  <stop offset='0%' className={`${extendedPath}-gradient-${i}-start`} />
+                  <stop offset='100%' className={`${extendedPath}-gradient-${i}-stop`} />
+                </linearGradient>
+              )
+            })
+          )}
         </defs>
 
+        <style>
```
raineorshine: Maybe we can factor out the `<style>` into its own component as well to reduce the size of the main component?

raineorshine (changes requested):
Thanks for the update!

> > 1. Rather than use a continuous gradient from 0-100% opacity over the length of the segment, I faked it by starting at 25% opacity, ramping up to 50% opacity at the end of the first path segment, and then continuing at 50% until finally increasing all the way to 100% in the final path segment. I feel like this better approximates the gradient that you posted in the issue, and it's easy to tweak to get a nicer-looking result.
> 
> Looks good to me!

Actually, some of the longer gestures appear to fade out too quickly. The arrowhead is bright white and then the tail is immediately gray. I would expect a little more of a gradual falloff from 100% to 75% to 50% etc. Maybe we can find some middle ground?

Brave Browser 2025-09-17 20 49 12
Brave Browser 2025-09-17 20 49 18

ethan-james:
> Actually, some of the longer gestures appear to fade out too quickly. The arrowhead is bright white and then the tail is immediately gray. I would expect a little more of a gradual falloff from 100% to 75% to 50% etc. Maybe we can find some middle ground?

I added some logic to make the second-to-last segment ramp up from 50% to 75%, and then the final segment ramps from 75% to 100%. It doesn't apply to 1-or-2-segment paths. So a 1-segment path goes straight from 25% to 100%, and a 2-segment path goes 25%-50%-100% just as it did before. I think it looks very nice.

raineorshine (approved):
Awesome! Looks great.

PR #3235, opened 2025-09-15
Quick drop panel test

### LINKED ISSUE #3145: Add QuickDropPanel Test
Add a puppeteer behavioral test to cover #3142.

karunkop:
Close #3145 

I have added e2e puppeteer behavioral test for quick drop on the invisible panel. This PR covers both desktop and mobile emulation for the same behavior. In the test I have added this specific function
```ts
/** Check if a thought is in the DOM. */
const isThoughtInDOM = async (value: string) => {
  const thoughtElement = await getEditable(value)
  const thoughtElementExists = await thoughtElement.evaluate(element => element !== null)
  return thoughtElementExists
}
```
Not sure if this is a correct way to check thoughts in the DOM. Looking forward for your feedback.

[Here](https://github.com/karunkop/em/actions?query=branch%3Aquick-drop-panel-test) is the CI result to ensure that the newly added test is not flaky.

### DIFF
--- a/src/e2e/puppeteer/__tests__/quickdrop-panel.ts
+++ b/src/e2e/puppeteer/__tests__/quickdrop-panel.ts
@@ -1,14 +1,12 @@
 import { KnownDevices } from 'puppeteer'
 import click from '../helpers/click'
 import clickThought from '../helpers/clickThought'
-import dragAndDropThought from '../helpers/dragAndDropThought'
+import dragToQuickDropPanel from '../helpers/dragToQuickDropPanel'
 import emulate from '../helpers/emulate'
 import getEditable from '../helpers/getEditable'
-import longPressThought from '../helpers/longPressThought'
 import paste from '../helpers/paste'
+import waitForAlertContent from '../helpers/waitForAlertContent'
 import waitForEditable from '../helpers/waitForEditable'
-import waitUntil from '../helpers/waitUntil'
-import { page } from '../setup'
 
 vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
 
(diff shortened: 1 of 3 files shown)


src/e2e/puppeteer/helpers/longPressThought.ts
```diff
@@ -53,6 +56,31 @@ const longPressThought = async (
     bulletElement,
   )
 
+  if (quickDrop) {
```
raineorshine: I think I'd rather have this functionality in its own helper or in `dragAndDropThought`. I understand the desire to reuse some of the existing logic, but a long press is not a drag-and-drop. It's important that we keep the test API clear and intuitive, and dragging to the `QuickDropPanel` is best understood as a type of drag-and-drop operation.

src/e2e/puppeteer/__tests__/quickdrop-panel.ts
```diff
@@ -0,0 +1,150 @@
...
+import waitForEditable from '../helpers/waitForEditable'
+import waitUntil from '../helpers/waitUntil'
+import { page } from '../setup'
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+/** Check if a thought is in the DOM. */
+const isThoughtInDOM = async (value: string) => {
+  const thoughtElement = await getEditable(value)
+  const thoughtElementExists = await thoughtElement.evaluate(element => element !== null)
```
raineorshine, line 16: If `getEditable` finds an element, then wouldn't it always exist?

Does `getEditable` return `null` in this case?

src/e2e/puppeteer/__tests__/quickdrop-panel.ts
```diff
@@ -0,0 +1,150 @@
...
+  return thoughtElementExists
+}
+
+/** Move mouse to QuickDropPanel area (right edge, 2em width ≈ 32px). */
+const dropOnQuickDropPanel = async () => {
```
raineorshine: This function says `dropOnQuickDropPanel` but it doesn't actually drop. `moveToQuickDropPanel` would more accurately reflect the behavior.

src/e2e/puppeteer/__tests__/quickdrop-panel.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(await isThoughtInDOM('a')).toBe(true)
+
+    await clickThought('a')
+    await click('[aria-label="Add to Favorites"]')
+
+    // wait until the favorite alert appears
+    await waitUntil(() => {
+      const favoriteAlertContent = document.querySelector('[data-testid="alert-content"]')
+      return favoriteAlertContent?.textContent?.includes('Added a to favorites')
+    })
```
raineorshine: This pattern of reading the alert content is used multiple times. Might benefit from a helper function?

src/e2e/puppeteer/__tests__/quickdrop-panel.ts
karunkop: Even if getEditable doesn't find an element, it return something like `CdpJSHandle {}`
image

So it was necessary to evaluate and check for null element.

raineorshine: Okay, good to know. I was hoping it would be simpler!

raineorshine (approved):
Thanks!

PR #3236, opened 2025-09-16
Jest reference error

### LINKED ISSUE #3035: Flaky test: jest is not defined
`ReferenceError: jest is not defined`

https://github.com/cybersemics/em/actions/runs/15656655010/job/44108578939

karunkop:
###  Initial Investigations and Findings

Close #3035 

### Cause

OK, so the error: `ReferenceError: jest is not defined` is occuring because of a certain dependency with jest function internally by `jest-canvas-mock` >`createImageBitmap.js` which is apparently used under the hood by `vitest-canvas-mock` , essentially used to mock canvas in unit tests. 

### Different solutions implemented and tested

1. Firstly after discovering the reason behind the issue, i.e., Jest being unable to access globally sometimes in CI, I tried to stub jest globally in `setupTests` file before importing `vitest-canvas-mock`

```ts
vi.stubGlobal('jest', vi)
// dynamically import after global access to jest 
await import('vitest-canvas-mock')
```
Unfortunately this didn't work. Maybethere was something going on when modules were loaded. This was also one of the solutions that was posted in https://github.com/wobsoriano/vitest-canvas-mock/issues/10. Tried almost all of the solutions mentioned here, but none of them seem to be proving to be successful.

2. In the same repo above, there was this https://github.com/wobsoriano/vitest-canvas-mock/issues/10#issuecomment-2517322304, where it suggested to completely get rid of the `vitest-canvas-mock` lib as internally it is still dependent on `jest-canvas-mock`. So I did that. Removed `vitest-canvas-mock`lib completely and installed https://github.com/schirrel/vi-canvas-mock. I was hopeful this would work, but it caused issues with too many unit test files, causing them to fail. However, this was without stubbing Jest globally. When I added that to the `setupTests.js` file again, everything seemed to work again. 

The interesting thing here is I am still unknown about why stubbing Jest globally is necessary even after completely replacing the previous mock lib with `vi-canvas-mock` which claims to have 0 dependencies with Jest. I can see there are some files where Jest is being referenced within the codebase, but this seems to be easily replaceable with vi itself. Tried this too, but it didn't work either.

Overall, after these changes, I am sure that the Jest reference error is fixed. For my sanity, I tested unit tests multiple times in CI as well. [Here](https://github.com/karunkop/em/actions?query=branch%3Ajest-reference-error) is the result of those checks.


raineorshine (approved):
Amazing work! Just the kind of research, thoroughness, and persistence I like to see.

raineorshine (approved):
Thank you!

Issue #2783, opened 2025-01-23
New Thought shifts up after initial render


raineorshine:
## Steps to Reproduce

1. Create thought `a`. 
2. New Thought. 

## Current Behavior

The new thought bullet and editable shift up a few pixels after its initial render. 

https://github.com/user-attachments/assets/07ee4adc-2d7a-4182-a7b9-dc21b72a810c

## Expected Behavior

The new thought should be rendered in the correct position from the start.

PR #3243, opened 2025-09-19
Store cliff in setSize so that isNewCliff can check whether it's actually replacing a previous cliff

### LINKED ISSUE #3097: New thought shifts down on initial render
## Steps to Reproduce

```
- a
  - b
```

1. Set the cursor on `a`. 
2. New Thought (→↓)

## Current Behavior

The new thought animates down a few pixels after the initial render. 

https://github.com/user-attachments/assets/ee9315ee-f1e9-4cfc-a06a-c187c34558e0

## Expected Behavior

The new thought should just fade on. It should be rendered at the correct y position immediately. 

https://github.com/user-attachments/assets/7c723792-8913-4945-b812-d2a464f7cabf

ethan-james:
Fixes #3097

As per [this comment](https://github.com/ethan-james/em/blob/6b7d5ae04f4b21e4b3e3a1d64e4dca7f755a856a/src/hooks/usePositionedThoughts.ts#L175), it is probably necessary to store the previous thought's cliff so that we can tell if it's losing it. The `isNewCliff` check was introduced in order to fix https://github.com/cybersemics/em/issues/2783, and now we have the opposite problem where it's compensating for a cliff that doesn't exist. The two cases,

```
- A
- B
```

and

```
- A
  - AA
- B
```

look pretty similar, from the standpoint of `treeThoughtsPositioned`, once `AA` collapses. So I added a `cliff` property to the `sizes` store, because that's the same store that is causing the new thought to overcorrect on its height, and it's the same store that is lagging behind `treeThoughtsPositioned` so it contains useful, consistently-outdated information at the time when `treeThoughtsPositioned` is being recalculated.

I am trying to convince myself that this is a pretty good solution, although it definitely needs more cleanup. Mostly I wanted to get it out of my brain before the end of the week so that you could let me know if you think it's viable before I finish fleshing it out.

### DIFF
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -295,7 +295,7 @@ const TreeNode = ({
               indexDescendant={indexDescendant}
               isMultiColumnTable={false}
               leaf={leaf}
-              onResize={setSize}
+              onResize={props => setSize({ ...props, cliff })}
               path={path}
               prevChildId={prevChild?.id}
               showContexts={showContexts}
--- a/src/components/VirtualThought.tsx
+++ b/src/components/VirtualThought.tsx
@@ -33,6 +33,8 @@ export type OnResize = (args: {
   id: ThoughtId
   /** Used by the LayoutTree to crop hidden thoughts below the cursor without disrupting the autofocus animation when parents fade in. */
   isVisible: boolean
+  /** Trying to store the cliff in the size object so that it can be referenced in treeThoughtsPositioned in subsequent passes. */
+  cliff?: number
   /** A key that uniquely identifies the thought across context views. */
   key: string
 }) => void
--- a/src/hooks/usePositionedThoughts.ts
+++ b/src/hooks/usePositionedThoughts.ts
@@ -21,7 +21,7 @@ const usePositionedThoughts = (
   }: {
     maxVisibleY: number
     singleLineHeight: number
-    sizes: Index<{ height: number; width?: number; isVisible: boolean }>
+    sizes: Index<{ height: number; width?: number; isVisible: boolean; cliff?: number }>
   },
 ): {
   indentCursorAncestorTables: number
@@ -100,6 +100,7 @@ const usePositionedThoughts = (
 
     const treeThoughtsPositioned = treeThoughts.map((node, i) => {
       const prev = treeThoughts[i - 1] as TreeThought | undefined
+      const prevCliff = prev ? sizes[prev.key].cliff || 0 : 0
       const next = treeThoughts[i + 1] as TreeThought | undefined
 
       // cliff is the number of levels that drop off after the last thought at a given depth. Increase in depth is ignored.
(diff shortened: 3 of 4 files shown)


raineorshine:
I'm not sure yet if we need to explicitly store the previous cliff. If `isNewCliff` was successful at fixing #2783, perhaps we could take a similar approach here?

`isNewCliff` subtracts `cliffPadding` when it detects when the cliff is going to be _removed_ from the previous thought. To fix #3097, we need a condition that detects when a cliff is going to be _added_ to the previous thought. Then we can *add* `cliffPadding` in anticipation.

Another way to think about this is that `!sizes[node.key]` tells us if this thought is being rendered for the first time. Combined with `cliff` and `prevCliff`, that should be enough to tell us we need to artificially add/remove `cliffPadding`.

---

If that doesn't work, I'm open to adding cliff to `sizes`. It's not out of place there: height is cached to make it possible to calculate thought positions, and cliff is a critical part of those calculations. That said, if we can determine how to render that first cliff in a "stateless" way, that's less complexity.

raineorshine:
Okay, wishful thinking I suppose! Thanks for investigating.

I trust your judgment about the best way to move forward on this.

ethan-james:
> Thanks for investigating the test failures. If there are imperceptible visual differences, it's acceptable to update the snapshots. I can't say why the snapshots changed if the y values are unchanged... that is odd. Let's just confirm that there are no visible changes using the image diff slider on GitHub afterward the snapshots have been updated and we should be good!
> 
> FYI Karun has actually eliminated all the flakiness in the tests, so they are expected to be stable now. It's safe to trust the CI again :).

I do actually see `x` moving up and down slightly. I'll take another quick look. Glad to hear that Karun fixed the tests!

https://github.com/user-attachments/assets/b211b808-b055-4cbf-a0c4-c11e3db925ae

raineorshine:
Thanks! The vertical alignment of the thought relative to the bullet is quite precise, so that is a case where 1px matters.

Feel free to reach out to @karunkop for test support.

ethan-james:
> Thanks! The code looks good.
> 
> What is your recommended way to move forward? Do you think the layout shift is a regression? Is there anything that we could do to increase our confidence that this isn't a regression?

I see no layout shift during the drag itself, so I think that it's safe to compare screenshots on page load, with the cursor on `c`. If that assertion sounds reasonable, I put together a little onion skin comparing what I see on `main` with what I see on the `3097-bug-setSize-stores-cliff` branch. It doesn't look like much because I don't see any differences between the two screenshots. You are welcome to replace them with your own screenshots, or swap in `test.png` if you want to verify that the slider actually works. It seems to get disconnected pretty easily, so using the refresh button in the right browser pane of CodeSandbox might help with that. https://codesandbox.io/p/sandbox/kq5x8g

I played around a little more with fast refresh and/or HMR, and the only thing I could discover is that saving changes to most files triggers a full page refresh, while changes specifically to `/src/devices/scrollCursorIntoView.ts` trigger what is probably Vite's HMR under the hood. Maybe that's because the new `scrollCursorIntoView` isn't a dependency of any React effects? I wasn't sure that this was going to get me anywhere, so I didn't spend a lot of time on it.

I think that the layout shift is not a regression, and that it's probably best to merge what we have.

raineorshine (approved):

PR #3244, opened 2025-09-20
Fix y value when moving thought up/down

zukilover:
This fixes regression in #2531 

During `moveThought` animations we scale via transform: scale3d(...): https://github.com/cybersemics/em/blob/345bca476e8573427fafb936fac52468108b7a8f/panda.config.ts#L167-L171

So measuring heights mid-animation yields distorted values which interferes with y accumulation. 

Added `data-move-animating="true"` on the inner animated container only when `moveThought` animation is active.
In `VirtualThought`, `updateSize` now checks for an ancestor with the attribute above. If present, it skips measurement and schedules a re-measure after the animation duration. This prevents capturing scaled bounding boxes.

It cleans up the scheduled timeout on unmount.

### DIFF
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -234,10 +234,7 @@ const TreeNode = ({
           top: y,
           left: 0,
         }
-      : moveDivStyle
-
-  // Flag when moveThought animation is active so children (e.g., size measurement) can adjust behavior.
-  const isMoveAnimating = !!moveDivStyle && !isSwap && !isLastActionSort
+      : {}
 
   return (
     <FadeTransition
@@ -269,7 +266,6 @@ const TreeNode = ({
         style={outerDivStyle}
       >
         <div
-          data-move-animating={isMoveAnimating ? 'true' : undefined}
           className={css({
             ...(isTableCol1
               ? {
@@ -314,6 +310,7 @@ const TreeNode = ({
               prevCliff={treeThoughtsPositioned[index - 1]?.cliff}
               isLastVisible={isLastVisible}
               autofocus={autofocus}
+              moveStyle={moveDivStyle}
             />
           </div>
           {dragInProgress &&
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -236,6 +236,9 @@ const TreeNode = ({
         }
       : moveDivStyle
 
+  // Flag when moveThought animation is active so children (e.g., size measurement) can adjust behavior.
+  const isMoveAnimating = !!moveDivStyle && !isSwap && !isLastActionSort
+
   return (
     <FadeTransition
       id={thoughtKey}
@@ -266,6 +269,7 @@ const TreeNode = ({
         style={outerDivStyle}
       >
         <div
+          data-move-animating={isMoveAnimating ? 'true' : undefined}
           className={css({
             ...(isTableCol1
               ? {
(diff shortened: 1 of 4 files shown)


zukilover:
<details>
<summary>result</summary>

https://github.com/user-attachments/assets/201e5c65-1c5f-4804-9d95-8df0bf31fff0


</details>

raineorshine (changes requested):
Hi, thanks for your submission! Good work identifying `scale3d` as the cause of the problem.

Unfortunately skipping that first measurement causes the next thought to be rendered with an incorrect y value and then correct itself on the next measurement. You can see in your video how `B` drifts up after the move, which shouldn't happen.

It is rendered too far down because `C` should have lost its `cliff` (the extra space at the end of a list) as `B` becomes the last thought, but the skipped measurement means that `usePositionedThoughts` thinks that `C` still has a cliff.

https://github.com/cybersemics/em/blob/345bca476e8573427fafb936fac52468108b7a8f/src/hooks/usePositionedThoughts.ts#L182-L183

Background: This is basically causing #2783 to recur. Ethan is actively working on the related #3243 in order to fix  #3097, though that shouldn't be too relevant to you. The point is that it's important that the height gets updated on move before the first paint. Skipping the measurement breaks the layout.)

Maybe you can scale a descendant or insert a wrapper div rather than scaling `innerDivStyle` so that the height measurement is not affected?

raineorshine:
Hi! Ready for review or still working?

zukilover:
@raineorshine I’m still testing it, but I’ll wrap it up shortly.

zukilover:
Here are the latest updates:

I removed the measurement skip logic and wrapped `VirtualThought` in a sibling `<div>` that applies `moveDivStyle` only for move animations. This way, the move animation scales just the wrapper, not the measured node.

I also tweaked the animation duration to better sync the move animation with the Y transition for smoother results.

Lastly, I noticed some discrepancies in max height between mobile and desktop. Switched `VirtualThought` height measurement to use `offsetHeight` to avoid fractional values on mobile.

zukilover:
<details>
<summary>Result</summary>

https://github.com/user-attachments/assets/a2271736-ac01-4b63-8dca-fbbfc6f6e748


</details>

raineorshine:
Thank you!

> I removed the measurement skip logic and wrapped `VirtualThought` in a sibling `<div>` that applies `moveDivStyle` only for move animations. This way, the move animation scales just the wrapper, not the measured node.

That seems reasonable, thanks.

> Lastly, I noticed some discrepancies in max height between mobile and desktop. Switched `VirtualThought` height measurement to use `offsetHeight` to avoid fractional values on mobile.

@zukilover Can you say more what problem fractional values cause? Usually subpixel calculations are handled fine by the browser, so I'm curious.

---

@JoeDuong-dev Can you test the following cases on mobile and desktop? Thanks so much!

1.  #2531 should be working.
2. The [first regression](https://github.com/cybersemics/em/pull/3114#issuecomment-3293302360) should be fixed.
3. The [second regression](https://github.com/cybersemics/em/pull/3244#pullrequestreview-3249289414) should be fixed.

Note that the cursor overlay bullet (darker gray bullet that animates) does not line up with the bullet as it animates. This is a known issue and is not in scope for this ticket.

raineorshine (approved):

ethan-james:
> Okay, noted!
> 
> @ethan-james TLDR; @zukilover [is changing](https://github.com/cybersemics/em/pull/3244/files#diff-6b09511cd1c1a624a0f81d4a2b3a43dc73d8a096c4a1c823c2daebf0a292d323R143-R149) the thought height calculation from `getBoundingClientRect().height` to `offsetHeight`, which may be related to #3243 and #2783. Hopefully it doesn't impact your work there. Let's keep a close eye out for regressions.

Thanks for the heads-up! I was hoping this might resolve some of the tiny layout shifts that I'm seeing, but when I checked out this branch all of the tests passed locally. If I instead copy the `offsetHeight` logic into my branch for #3243, I see some new test failures. So I'll need to follow up once this gets merged.

PR #3246, opened 2025-09-23
[Liminal UI] Implement overlay, falloff and smaller design changes

### LINKED ISSUE #3230: [Liminal UI] Implement overlay, falloff and smaller design changes
**Parent issue: #3229**

--- 

Image

## Tasks
- [ ] **Implement new overlay**: Replace the grey card with the background glow and blur elements, while making sure to maintain the sliding animation & swipe-to-close gesture.
- [ ] **Implement smaller design changes** like the slightly larger "thoughts selected" text, and the new "Done" button.
- [ ] **Ensure the Command Center scales gracefully across different mobile phone sizes.** The minimum phone size to consider is the iPhone SE (375px width).

## Notes

Image

Instead of a typical "card", we use two different elements together to create a visual separation between the buttons and the rest of the app:

#### Design notes
   - A translucent image that acts as the "overlay" that the buttons sit on top of.
   - In the Liminal UI design language, this is our equivalent of a “card” – it provides a visual separation between the content and the Command Center.
   - *Unlike* a card, the background glow takes up the full screen.
	   - This is a common design pattern in Liminal UI – light *spreads out* rather than being boxy and contained.
	   - Though the content of the Command Center only takes up a third of the screen, the glow spreads out to take over the entire screen.
	   - Look how the Command Center itself ends just underneath the thought "Poetry", yet the light spreads all the way up to the thought "Business".

#### Implementation notes
   -  **A PNG image for the background glow is available in the [assets folder](https://public.3.basecamp.com/p/TmFdohvobCnQpbMUqntjsw3L/vault).**
	   - An SVG is provided too, but since it makes heavy use of overlapping gaussian blurs, it's likely to incur a **hefty performance penalty** (especially when we then *animate* on top of that!)
	   - So, we should just use the precomputed PNG image instead.
   - The element should use the `screen` CSS blend mode so that fully white text from the editor pierces through the light.
   - Make sure to maintain the sliding animation and swipe-to-close gesture in the current implementation.

#### Design notes
   - Because the Command Center is translucent, we need something opaque **underneath it** so the user’s thoughts don’t show through the buttons.
   - Look at the image of the Command Center at the top. Thoughts like "Philosophy" and "Poetry" are partially obscured by opacity and blur, and thoughts underneath it are completely obscured.
   - In the designs, we achieve this with a **simple box with a linear gradient fading to black** *underneath* the background glow. This fades the content underneath the glow element to black, making it easier to see the commands.
   - We use a **progressive blur** backdrop filter to ramp up the intensity of the blur as the gradient progresses. See how "Poetry" is much blurrier than "Philosophy". This creates the impression of depth. 

#### Implementation notes
   - **You can find the exact gradient used in the design in [its Figma component](https://www.figma.com/design/oQnevfxv79fhTJnOs0fz9k/em-design?node-id=4093-58047&t=I1UConvjpb4bjayX-4).**
   - Progressive backdrop blur can be done with pure CSS, but it's not a standard feature/single property. It might require some experimentation to get this effect working.

## Considerations
- Liminal UI's heavy use of gaussian blurs and blend modes may impact mobile performance.  
- Stage 1 will be a good test case for these potential issues. Please flag if animations feel sluggish, or if certain devices or browsers struggle.

yangchristina:
Resolves #3230

https://github.com/user-attachments/assets/1938af77-f70a-46a2-a229-0aa7d7ecc8dc


yangchristina:
@fbmcipher Thanks! Yes, I'll try both and send them to you once done.

@raineorshine Is using MUI a requirement? The MUI Swipeable Drawer doesn’t seem to support the behaviour we’re aiming for, and even if I get it working, the solution would likely be quite hacky.

yangchristina:
@fbmcipher here is fade only:


https://github.com/user-attachments/assets/1049bd81-91ac-4880-8dcc-2112f3c53685

yangchristina:
@fbmcipher Here is Fade & slide. I made the fade and slide transitions match the slide transition of the MUI drawer, but we can alway change the easing for just the fade transition.

https://github.com/user-attachments/assets/878176a6-ac26-4a99-8717-2e069d3833a5

raineorshine:
It's not ideal to reference a ref value in a React render function, because it's nonreactive. It won't re-render if the height changes, and the particular value that it references on a given render is going to implicitly depend on the timing of the component lifecycle and state updates. The value will end up either being outdated sometimes, or fortuitously never outdated, which is not something you want to rely on.

I don't have an immediate alternative in mind, but maybe you and @fbmcipher can come up with a more conventional solution.

yangchristina:
> It's not ideal to reference a ref value in a React render function, because it's nonreactive. It won't re-render if the height changes, and the particular value that it references on a given render is going to implicitly depend on the timing of the component lifecycle and state updates. The value will end up either being outdated sometimes, or fortuitously never outdated, which is not something you want to rely on.
> 
> I don't have an immediate alternative in mind, but maybe you and @fbmcipher can come up with a more conventional solution.

@raineorshine I can use useState to store the height instead and a ResizeObserver, that way it will update when the height changes. I know this is still not ideal, but the progressive blur effect doesn't work when it is stored inside SwipeableDrawer.
It doesn't blur the below thoughts when it is within SwipeableDrawer, it only worked when I moved it outside. This blur must match the height of the CommandMenu, hence why I need to track the height of CommandMenu.

fbmcipher:
@yangchristina Fade and slide works! Let's keep things as simple as possible for this iteration. If the animation still feels off, we can take another look later in a separate piece of work.

I'll field some of the more technical questions tomorrow.

src/components/Footer.tsx
```diff
@@ -160,7 +160,7 @@ const Footer = () => {
         backgroundColor: 'footerBg',
         boxSizing: 'border-box',
         width: '100%',
-        zIndex: 'modal',
+        zIndex: 'footer',
```
yangchristina, line 163: prevent footer from showing over command menu

src/components/icons/CloseIconV2.tsx
```diff
@@ -0,0 +1,28 @@
+import { css, cx } from '../../../styled-system/css'
+import { iconRecipe } from '../../../styled-system/recipes'
+import { token } from '../../../styled-system/tokens'
+import IconType from '../../@types/IconType'
+
+/** Close icon. */
+const CloseIconV2 = ({ fill, width = 10, height = 11, cssRaw, style }: IconType) => {
```
yangchristina: more rounded version of old close icon

yangchristina:
> @yangchristina Fade and slide works! Let's keep things as simple as possible for this iteration. If the animation still feels off, we can take another look later in a separate piece of work.
> 
> I'll field some of the more technical questions tomorrow.

Thank you! I've implemented the fade and slide, and updated the video in the PR description.

src/durations.config.ts
```diff
@@ -37,6 +37,8 @@ const durationsConfig = {
   /* Duration for context view disappearing text animations */
   disappearingUpperRight: 500,
   disappearingLowerLeft: 500,
+  /* Duration for command menu swipable drawer animation. */
```
fbmcipher, line 42: Given the `medium` duration is also 400ms, is it possible for us to just use that duration and avoid introducing the new `commandMenuDrawer` duration here?

Typically, we want to keep animation speeds consistent across the app, and defining durations for specific features might make that tougher.

src/recipes/fadeTransition.ts
```diff
@@ -28,6 +28,11 @@ const fadeTransitionRecipe = defineSlotRecipe({
         enterActive: { transition: `opacity {durations.medium} ease 0ms` },
         exitActive: { transition: `opacity {durations.medium} ease 0ms` },
       },
+      commandMenuDrawer: {
+        // Easing follows that of Material UI SwipeableDrawer.
+        enterActive: { transition: `opacity {durations.commandMenuDrawer} cubic-bezier(0, 0, 0.2, 1) 0ms` },
+        exitActive: { transition: `opacity {durations.commandMenuDrawer} cubic-bezier(0, 0, 0.2, 1) 0ms` },
+      },
```
fbmcipher, line 35: Matching the easing curve is a nice, pragmatic solution that avoids too much hacking with MUI.

src/components/CommandMenu/CommandMenu.tsx
```diff
@@ -85,93 +110,147 @@ const CommandMenu = () => {
...
+                      padding: '8px 16px',
+                      mixBlendMode: 'soft-light',
+                      display: 'flex',
+                      alignItems: 'center',
+                      gap: '1ch',
+                      /** Button won't show without z-index. */
+                      zIndex: 'modal',
+                    })}
+                    style={{
+                      fontSize: fontSize * 0.85,
```
fbmcipher: This is an interesting bit of code. Multiplying by the editor’s font size means the text size for `MultiselectMessage` and `Done` buttons scales with the editor’s text size.

I’m not fully sure that’s the _right_ behaviour. In some places (like the toolbar) UI elements _do_ respond to the editor’s font size, and in others (like the sidebar) they _don’t_ – so the current approach we have in the app feels inconsistent.

For now, I’d lean toward keeping the editor’s font size separate from UI sizing. Instead of multiplying `0.85 * fontSize`, we could simply use `font-size: 0.85em`. This would make the sizing relative to the parent element’s font size (CSS’s normal behaviour), avoids importing a Redux selector, and keeps the UI independent from the editor’s text size.

If we later decide that UI elements should scale with the editor’s font size, we can just change the base em value to scale accordingly – without touching each component individually.

What do you think?

src/components/CommandMenu/CommandMenu.tsx
```diff
@@ -85,93 +110,147 @@ const CommandMenu = () => {
...
+                    }}
+                  >
+                    <CloseIconV2 /> Done
```
fbmcipher: Just need to strip out `<CloseIconV2 />` as the latest mock no longer has a close icon – just the word "Done".

src/components/icons/CloseIconV2.tsx
fbmcipher: Nice catch noticing the slightly more rounded close icon!

Though, we can likely delete this file as we no longer use a close icon in the latest mock.

fbmcipher (changes requested):
@yangchristina This looks great, and fade & slide looks great too! The light gradually fading in as the drawer opens feels much better than the janky animation we had before.

During testing, I did notice one detail that feels missing: the backdrop doesn't fade out _as_ you close the drawer with your finger.

https://github.com/user-attachments/assets/12261fc1-7649-4eb3-9033-ee7a8f834833

However, I accept that this probably isn't easy to do in a performant and elegant way with MUI's `SwipeableDrawer` limitations.

Between `fade` and `fade+slide`, I actually preferred `fade` – but I couldn't advocate for its implementation given the complexity and hackiness it'd take to work against MUI.

I like the idea of removing our dependency on MUI's `SwipeableDrawer` and finding a more customisable component/hook so we can customise further, but I think that work would be better done in another PR.

For now, this looks and works great IMO!

> Actually, I just found out that the culprit for the progressive blur not working isn't the portal, but actually mix-blend-mode. It was working when outside the drawer because it wasn't affected by mix-blend-mode: screen on the overlay there. I will look more into this.

The likely culprit here: when we use `mix-blend-mode`, a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Stacking_context) is created. That changes the element’s compositing order and isolation, which can break or change the expected blending with siblings and ancestors. This behaviour is worth keeping in mind as our designs use a lot of blend modes.

Just a few minor changes from me!

src/durations.config.ts
yangchristina: The reason I added this is to match MUI's easing. Because of the implementation of `FadeTransition`, it requires the variant to also exist in the durationConfig object. To remove it from durationConfig means I'd have to do something like:
```
<CSSTransition
      key={id}
      classNames={fadeClasses}
      timeout={durations.get(type === 'commandMenuDrawer' ? 'medium' : type)}
```
This doesn't seem ideal, so adding to the durationConfig seems like a better solution to me.
Please let me know if you have a better solution!

src/components/CommandMenu/CommandMenu.tsx
yangchristina: Updated! See video below for how it looks now.

src/components/CommandMenu/CommandMenu.tsx
yangchristina: Updated, thanks! I agree, using `em` is cleaner than multiple fontSize selectors. The only potential concern is that changing a font-size on a parent could cause unexpected side-effects, using `rem` would work in that case, but I'm not sure if we support `rem` changing by selector.

[ba2122d](https://github.com/cybersemics/em/pull/3246/commits/ba2122db6ac1c523bb32d81e8cd4c930436003f8)

Screenshot 2025-10-03 at 11 40 52 PM

src/components/icons/CloseIconV2.tsx
yangchristina: removed thanks! [2111e93](https://github.com/cybersemics/em/pull/3246/commits/2111e9363be6ef80908054335ea73481660de2f3)

src/components/CommandMenu/CommandMenu.tsx
yangchristina: [2111e93](https://github.com/cybersemics/em/pull/3246/commits/2111e9363be6ef80908054335ea73481660de2f3)

src/components/CommandMenu/CommandMenu.tsx
fbmcipher: Thank you! 

> The only potential concern is that changing a font-size on a parent could cause unexpected side-effects

Good point! Let's keep an eye out for that. Looking through the codebase, the majority of fontSize styles use `em`, so let's standardise on `em` for now.

If we start seeing unwanted cascading side effects, we can revisit and move to `rem` at that point.

src/durations.config.ts
fbmcipher: Thank you for explaining – that makes total sense now!

All things considered, it seems like you've arrived at the most ideal solution. Happy with this :)

fbmcipher (approved):
Thanks for the changes and explanations!

> Thanks! Turns out MUI's exit transition easing is different from the enter transition. I've updated the easing here:
https://github.com/cybersemics/em/commit/f545e661606a65e91f7d5d25ddf4a4c756030020. How is this?

The new easing on close feels great – glad we're fully consistent with MUI wherever possible.

What I actually meant, though (I could have visualised better in the video) – is that the glow/overlay doesn't fade out as the Command Center is closed **gesturally**:


https://github.com/user-attachments/assets/91f893cd-5b4e-4eae-ac09-f9be9243ab27



It looks great when it closes via the Done button or other tap interaction – the glow fades  out nicely as the panel slides down.

But if the user slides down with their finger, then the overlay _doesn't_ fade out until we reach the bottom. The idea of light sliding like this doesn't *feel* quite right – light doesn't move like that IRL.

Though I fully appreciate that any solution would likely be inelegant and potentially introduce perf issues due the limitations we discussed around MUI, so I'm happy to circle back around to this later/in a separate issue.

---

Everything else looks great to me! I'm happy to approve and pass on to @raineorshine for one last check before we merge.

As soon as it's merged, we can move onto the next issue!

I'm really happy with how true to the design this PR is!

src/components/CommandMenu/CommandMenu.tsx
yangchristina: Sounds good! I think there was talk to move to `rem` at some point, but I didn't see it ever happen.

yangchristina:
> What I actually meant, though (I could have visualised better in the video) – is that the glow/overlay doesn't fade out as the Command Center is closed **gesturally**:

I see, thanks for explaining! Yes, the fade amount currently isn't tied to the height of the drawer, and MUI doesn't change the value of `showCommandMenu` until you let go, so the fade won't begin until you let go. Unfortunately I don't see a way to track the transform value of the drawer right now. I would be happy to circle back to this on a later issue as well.

yangchristina:
> Thanks for thinking ahead. Let's do it.
> 
> 1. Just fix the uncanny gap, then merge this PR.
> 2. Migrate to [Vaul](https://vaul.emilkowal.ski/) or [react-spring-bottom-sheet](https://react-spring.bottom-sheet.dev/) as a refactor in a new PR. Ignore background glow behavior, but otherwise ensure no behavioral changes from current implementation.
> 3. Implement the fade animation and background glow behavior as desired.

@raineorshine Sounds good! I've removed the margin-top, as suggested:


3d12c82

Screenshot 2025-10-16 at 11 24 11 PM

yangchristina:
@raineorshine would it be fine to implement the buttons before migrating the drawer library? I already have the changes ready, so would be easier to get those in first.

fbmcipher:
@yangchristina @raineorshine Just a heads-up: I opened #3283 with more details on the new drawer implementation, based on the upcoming stages of Command Center development.

yangchristina:
@raineorshine is there anything else needed for this PR?

Since this hasn't been merged yet, and I need this branch as the base for my next PR, I've created the PR on my forked repository for now, since I didn't see a way to do it on the main repository. 

https://github.com/yangchristina/em/pull/270#issue-3534399705

@fbmcipher how is this for active styling? (Sorry I couldn't mention you from my forked repository, so am asking here)

src/components/CommandMenu/CommandMenu.tsx
```diff
@@ -85,93 +105,144 @@ const CommandMenu = () => {
...
-                  display: 'flex',
-                  alignItems: 'flex-end',
-                  justifyContent: 'space-between',
-                  width: '100%',
-                })}
-              >
-                <MultiselectMessage />
-                <button
-                  {...fastClick(onClose)}
+      <>
```
raineorshine: Is there a reason for the fragment wrapper?

raineorshine:
> @raineorshine is there anything else needed for this PR?

Just waiting for @JoeDuong-dev to test!

> Since this hasn't been merged yet, and I need this branch as the base for my next PR, I've created the PR on my forked repository for now, since I didn't see a way to do it on the main repository. 

You can branch off your other branch and open a PR here. The diff will show all the changes, but I can select from the commits dropdown to limit the diff to just the most recent batch of work. Then when this PR gets merged, we can update the base.

src/components/CommandMenu/CommandMenu.tsx
yangchristina: Not anymore, removed thanks! 
cd3c3af

yangchristina: Actually let me undo that commit, something seems to have gone wrong there.

yangchristina: I thought there was an issue with the commit, but turns out it's just github diff not being clear. It is just removing the fragment, even though the diff looks very messy. added back here [359a1a2](https://github.com/cybersemics/em/pull/3246/commits/359a1a21c5c420f49ea9e0d1a2d29cae4aa6ce82)

yangchristina:
@raineorshine I think this is a flaky test?
drag-should-show-alert-and-quick-drop-panels-when-long-pressing-after-drag-operation-1-diff

raineorshine:
That seems like a reasonable theory since this PR doesn't touch thought layout. We eliminated all flakiness in the tests earlier this year, so if this is indeed flaky then it's a recent regression.

@karunkop is in charge of the test suite, so he can investigate further and provide guidance on how to move forward.

yangchristina:
@karunkop The tests appear to be passing for me locally. The only change I made was to remove a react fragment, therefore it shouldn't have caused any tests to break. I just merged with main to see if that was the issue, but it looks like the test failure is consistent.

I tried running puppeteer locally, and got passing tests.

Screenshot 2025-10-26 at 6 07 41 PM

Do you have an idea of what the issue could be?

yangchristina:
@raineorshine tests are passing now after merging with main.

PR #3247, opened 2025-09-24
Scroll cursor into view in BulletCursorOverlay

### LINKED ISSUE #3083: Scroll to anticipated cursor location
This will require some careful planning. Let's discuss the overall approach and make sure we feel confident before moving forward with an implementation.

## Current Behavior

`scrollCursorIntoView` scrolls to the `[data-editing=true]` element in the DOM. This does not work very effectively if the element is still animating into place. At best, it requires triggering `scrollCursorIntoView` again after the animation ends, which results in a choppy 2-phase scroll.

## Expected Behavior

Since we know the expected `y` position of the cursor in `LayoutTree` *before* it begins to animate, we should be able to scroll to the anticipated position instead of the actual position. Thus we can avoid multiple scrolls.

The only other thing to consider is that when `state.expanded` changes and new thoughts are loaded in, the y positions may change. Thus, maybe we wait until the new expansion state, but can at least begin scrolling when the animation starts rather than having to wait until it is in its final position.

ethan-james:
Fixes #3083 

`BulletCursorOverlay` is rendered at the final position of the cursor thought, relative to the document rather than relative to the viewport. It receives a `y` position from `LayoutTree` that can easily be passed to `scrollCursorIntoView`. The only missing piece is the `height` of `BulletCursorOverlay`, which currently I'm passing in as a hard-coded value so that `scrollCursorIntoView` can orient itself relative to the top or bottom of the thought.

### DIFF
--- a/src/hooks/useScrollCursorIntoView.ts
+++ b/src/hooks/useScrollCursorIntoView.ts
@@ -17,7 +17,12 @@ const useScrollCursorIntoView = (y: number, height: number) => {
   // Scroll the cursor into view after it is edited, e.g. toggling bold in a long, sorted context.
   // The cursor typically changes rank most dramatically on the first edit, and then less as its rank stabilizes.
   editingValueStore.useEffect(() => {
-    // sizeRef doesn't seem to update quickly enough to run this in the same render cycle
+    /** The hazard here is that editingValueStore.useEffect creates a closure around the provided values.
+     * Since ministore runs synchronously, it is not possible to update dependencies before the subscribers run,
+     * and it is not possible to unsubscribe/resubscribe based on new dependencies for the same reason.
+     * Since sizeRef is an object, it is possible to mutate its properties within the existing closure after
+     * React's render cycle runs and processes the effect above. That's why setTimeout is necessary here (#3083).
+     */
     setTimeout(() => throttledScrollCursorIntoView(sizeRef.current.y, sizeRef.current.height))
   })
 


src/components/BulletCursorOverlay.tsx
```diff
@@ -351,6 +353,8 @@ export default function BulletCursorOverlay({
     thoughtId: head(simplePath),
   })
 
+  useEffect(() => scrollCursorIntoView(y, 36), [y])
```
ethan-james: I see `const lineHeight = fontSize * 1.25` on [line 74](https://github.com/ethan-james/em/blob/9f35bb70816bf17178edcd30a0213b1a3330ed63/src/components/BulletCursorOverlay.tsx#L74) that provides a calculated height, but it works out to 22.5 at the default font size, which is pretty small. In order to get an accurate measurement of the height of the placeholder thought (in order to scroll the entire cursor into view rather than scrolling the y position of the top of the thought into view) then we would need to measure it programatically, introducing a layout call. How do you feel, is 36 a good enough approximation? It's the actual height of the placeholder thought as measured in the dev tools.

src/e2e/puppeteer/__tests__/virtualization.ts
```diff
@@ -77,6 +77,7 @@ describe('virtualizaton', { retry: 3 }, () => {
     // })
 
     await paste(text)
+    await clickThought('a1')
```
ethan-james, line 80: I'm also looking at why the behavior changed here.  The `importText` action seems to behave differently from the `importData` action that is triggered when I paste text into a thought. So for me, the cursor always stays on `a1` and no scroll is triggered in the browser, but apparently the new `scrollCursorIntoView` does scroll in this situation where the old version did not.

src/components/BulletCursorOverlay.tsx
raineorshine: I use `fontSize * 2 - 2` in [useSingleLineHeight](https://github.com/ethan-james/em/blob/4ab019689e80e6640fd430e60c118466d01b1f5f/src/components/LayoutTree.tsx#L36).

src/components/LayoutTree.tsx
```diff
@@ -139,6 +153,7 @@ const LayoutTree = () => {
 
       setLayoutTop(ref.current?.getBoundingClientRect().top ?? 0)
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
raineorshine: When disabling the linter, I like to include a comment that explains what dependency is being ignored and why. This helps to discourage developers from disabling the linter unnecessarily.

src/components/LayoutTree.tsx
```diff
@@ -111,14 +111,28 @@ const useHideSpaceAbove = (spaceAbove: number) => {
   )
 }
 
+/** A hook that returns a ref to the content div and updates the viewport store's contentWidth property on resize. */
```
raineorshine: Comment needs updating

src/redux-middleware/scrollCursorIntoView.ts
```diff
@@ -154,39 +118,4 @@ scrollTopStore.subscribe(() => {
...
-        () => {
-          // indicate that the cursor has changed and we want to scroll it into view
-          // this is needed for when thoughts need to be pulled from storage prior to scrolling
-          userInteractedAfterNavigation = false
-          scrollCursorIntoView()
-        },
-        // If the cursor changed, we need to wait for most of the layout animaton to complete before scrolling.
-        // Otherwise the cursor will not be at its final position and it will nott scroll far enough.
-        // 50% seems to be enough to avoid the issue, but this can be fine-tuned as needed.
-        sortChanged ? durations.get('layoutNodeAnimation') / 2 : 0,
```
raineorshine, line 186: So happy to get rid of these animation timing dependencies... :)

raineorshine (changes requested):

src/components/LayoutTree.tsx
ethan-james: I didn't think that it should be complaining because it was a ref before and it's still a ref, but I also don't think there's any harm in passing the ref as a dependency and removing the exemption.

ethan-james:
Things are working pretty well with the new `scrollCursorIntoView`, except that unfortunately, a pretty major bug has cropped up in `useLongPress`. This work appears to be causing the [setPressing](https://github.com/ethan-james/em/blob/61d186cf8faffbf762a1b35796d068a6ff4bfe76/src/hooks/useLongPress.ts#L81) setter to become stale somehow, so that when it is called by the `onTouchEnd` handler, it doesn't actually update the value of `pressing`. This prevents long press from being canceled, so it activates after a single click.

ethan-james:
It only seems to be triggered when selecting a thought that loads off-screen, with two strange caveats:

1. Due to overscroll, the affected thoughts are part of the DOM on load
2. It doesn't matter if `scrollCursorIntoView` immediately scrolls an off-screen thought into view, the same lower-down thoughts are still affected

https://github.com/user-attachments/assets/e6d345a2-1643-4601-8028-874fbc5a3ae5

This is all happening in desktop Chrome, by the way. Due to the way the drag-and-drop [touch backend](https://github.com/ethan-james/em/blob/61d186cf8faffbf762a1b35796d068a6ff4bfe76/src/hooks/useLongPress.ts#L46) works, mobile should be unaffected.

raineorshine:
Another twist! 😅

I don't have an immediate solution in mind, so let me know what your thoughts are after some additional investigation and brainstorming.

ethan-james:
> Another twist! 😅
> 
> I don't have an immediate solution in mind, so let me know what your thoughts are after some additional investigation and brainstorming.

It was scrolling too frequently because it wasn't properly taking the scroll top into account. This meant that `LayoutTree` was re-rendering in order to virtualize new thoughts, and was causing thoughts to re-render in the middle of a click. There were a few different contributing factors here, but my understanding is that it should be impossible for a click to trigger a scroll, because the clicked thought is by definition within the viewport. Therefore, the stale `setPressing` function is avoided if it doesn't scroll in response to a click on a thought.

ethan-james:
> I just tested on `main`, and sadly it looks like `scrollCursorIntoView` is not working as expected there 😔. I didn't know this when we started this issue. It appears to have a false positive where it scrolls when the thought is well within the viewport. I'm not sure when this regression was introduced.

I know that on mobile it often scrolls as though the virtual keyboard were present when it's not, but on desktop the `virtualKeyboardHeight` should be 0 I think.

raineorshine:
Okay, I tested the broken behavior in `main` and it goes back at least one year, which doesn't make sense. I thought maybe there was a browser update that changed the behavior, but I confirmed it's broken on both Desktop Chrome device emulator and iOS Safari going back one year.

It's odd to know this worked at some point, to see code that clearly intends to scroll the cursor into view only if needed, but to not be able to identify a good commit where this behavior actually works as expected 😕. I think we have no choice but to move forward and address unexpected behaviors as new bugs. There are lots of [Autoscroll bugs](https://github.com/cybersemics/em/issues/?q=is%3Aissue%20state%3Aopen%20milestone%3A%22%F0%9F%9B%BC%20Autoscroll%22) anyway.

> > The clicked thought may only partially be in the viewport, and so it would need to be scrolled a small amount on click to be fully in view.
> > Also, we have a margin of half the thought's height from the bottom edge of the screen.
> 
> For the most part, I see that those thoughts are covered by the home menu, which perhaps erroneously is not factored into the viewport height calculation.

That's factored in with `navbarRect?.height`.

src/components/BulletCursorOverlay.tsx
```diff
@@ -351,6 +356,12 @@ export default function BulletCursorOverlay({
     thoughtId: head(simplePath),
   })
 
+  // Scroll the cursor into view after it is edited, e.g. toggling bold in a long, sorted context.
+  // The cursor typically changes rank most dramatically on the first edit, and then less as its rank stabilizes.
+  const editingValue = editingValueStore.useSelector(value => value)
```
raineorshine: `editingValueStore` updates in real-time as the user types, so we have a performance limitation to work with here (as seen in the current implementation). It's not safe to call `scrollCursorIntoView` or re-render the component at this frequency.

I would recommend throttling to reduce the volume and [store.useEffect](https://github.com/cybersemics/em/blob/ee8582f3b6325996dbe86c13238ec8ea5dca03e6/src/stores/react-ministore.ts#L35) to avoid re-rendering the React component.

src/components/LayoutTree.tsx
```diff
@@ -139,7 +153,7 @@ const LayoutTree = () => {
 
       setLayoutTop(ref.current?.getBoundingClientRect().top ?? 0)
     }
-  }, [dragInProgress])
+  }, [dragInProgress, ref])
```
raineorshine, line 156: Is `ref` stable? I thought it was only the content of a ref that persists.

raineorshine (changes requested):

src/components/LayoutTree.tsx
ethan-james: I always thought it was, but now I realize that pretty much all patterns that would benefit from referential equality are discouraged, like passing it around to other components.

In https://react.dev/reference/react/useRef#reference, I see:

```On the next renders, useRef will return the same object.```

so I think the answer is yes.

File: src/components/BulletCursorOverlay.tsx (hunk 6 above)
Comment (ethan-james) [reply src/components/BulletCursorOverlay.tsx]: I did try to throttle it last week, and had a lot of trouble getting it to use the current `y` and `height` values while calling `scrollCursorIntoView` only when the store's value changed. I pushed a commit that calls `store.useEffect` so you can see what I'm trying. Is there a pattern anywhere in the codebase where `store.useEffect` depends on additional variables that shouldn't get stale?

Comment (ethan-james) [reply src/components/BulletCursorOverlay.tsx]: I decided that it would be helpful for me to package everything that seems to be necessary to get it to update properly into a `useScrollCursorIntoView` hook. We can take a look at it and hopefully do away with some of the moving parts.

File: src/components/LayoutTree.tsx (hunk 7 above)
Comment (raineorshine) [reply src/components/LayoutTree.tsx:156]: @ethan-james Never mind, I confirmed the ref box is always stable.

https://www.epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array

File: src/hooks/useScrollCursorIntoView.ts [hunk 8]
```diff
@@ -0,0 +1,27 @@
...
+import editingValueStore from '../stores/editingValue'
+
+const throttledScrollCursorIntoView = throttle((y: number, height: number) => scrollCursorIntoView(y, height), 400)
+
+/** Call scrollCursorIntoView when the y position of its container changes, or when the editing value changes. */
+const useScrollCursorIntoView = (y: number, height: number) => {
+  const sizeRef = useRef({ y, height })
+
+  useEffect(() => {
+    // store.useEffect doesn't take a dependency array, so parameters will get stale
```
raineorshine, line 13: Could we add an optional dependencies array param to react-ministore's `useChangeEffect`? That would more closely align with `React.useEffect` and would avoid the need for `sizeRef`.

ethan-james: With a dependency array, I still see the callback getting called with a stale y value. The timeout in the callback is still in place, although it shouldn't be necessary if this effect follows the same rules as a vanilla effect. I do see [emitter.off](https://github.com/ethan-james/em/blob/6e4d146c6cf98cdb9fa90246bb8b3449d1445169/src/stores/ministore.ts#L51) running correctly, although I also see it running repeatedly whenever I scroll. It still seems that ministore's change handler is being called with a callback that formed a closure containing the old y value.

https://github.com/user-attachments/assets/7c012b03-9615-49de-b08d-cca174916654

ethan-james: Probably what's happening is that `editingStoreValue` gets updated first, so it fires the callback before `y` and `height` change in the next render.

ethan-james: And the reason why the `sizeRef` version works is because the object's `current` property changes and thus the reference to it contained in the old closure still points to fresh data.

ethan-james: And the reason why the timeout is necessary in the `sizeRef` version is because `sizeRef`'s `current` property doesn't get updated until

```
  useEffect(() => {
    // store.useEffect doesn't take a dependency array, so parameters will get stale
    sizeRef.current = { y, height }
  }, [y, height])
```

runs, which is still a new render cycle and not immediate.

ethan-james: The only way that I can think of to make this work would be to introduce an artificial delay in updating `editingValue` in order to give `useEffect` a chance to run and unsubscribe/re-subscribe with fresh dependencies. I did push the change, because I think it's pretty hard to reason about it without seeing it.

src/components/BulletCursorOverlay.tsx
ethan-james: Instead of throttling the update on `editingValueStore`, would you be open to saving the y value that it's trying to scroll to, and short-circuiting if it tries repeatedly to scroll with the same y value? This would be essentially the same as throttling, but it would have the same behavior whenever any of the dependencies (`y`, `height` or `editingValue`) changes and would allow me to put them back into a single effect in order to avoid the timing issues that I'm having with subscribing to `editingValueStore`.

ethan-james: I guess it's not quite a simple as that, since user scrolling needs to be differentiated from autoscrolling in order to detect duplicate scroll requests without performing too many calculations. There was some code related to that in the old version of `scrollCursorIntoView`.

src/hooks/useScrollCursorIntoView.ts
```diff
@@ -0,0 +1,19 @@
...
+import editingValueStore from '../stores/editingValue'
+
+const throttledScrollCursorIntoView = throttle((y: number, height: number) => scrollCursorIntoView(y, height), 400)
+
+/** Call scrollCursorIntoView when the y position of its container changes, or when the editing value changes. */
+const useScrollCursorIntoView = (y: number, height: number) => {
+  // Scroll the cursor into view after it is edited, e.g. toggling bold in a long, sorted context.
+  // The cursor typically changes rank most dramatically on the first edit, and then less as its rank stabilizes.
+  editingValueStore.useEffect(() => {
+    setTimeout(() => throttledScrollCursorIntoView(y, height), 1000)
```
raineorshine: Is the 1s delay intentional? Scrolling the cursor automatically into view should generally happen as soon as possible.

src/stores/react-ministore.ts
```diff
@@ -9,13 +9,15 @@ const makeReactStore = <U extends Store<any>>(store: U) => {
   type T = U extends Store<infer V> ? V : never
 
   /** A hook that invokes a callback when the state changes. */
-  const useChangeEffect = (cb: (state: T) => void) =>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const useChangeEffect = (cb: (state: T) => void, deps: any[] = []) =>
```
raineorshine: You can import the `DependencyList` type from `React` (which is just `readonly unknown[]` under the hood).

src/hooks/useScrollCursorIntoView.ts
ethan-james: Yes, sorry, I added a long delay to demonstrate that the data is still stale even long after it has re-rendered.

src/components/BulletCursorOverlay.tsx
raineorshine: > Is there a pattern anywhere in the codebase where `store.useEffect` depends on additional variables that shouldn't get stale?

Adding the dependencies array should allow the callback to be updated when the dependencies change.

ethan-james: > > Is there a pattern anywhere in the codebase where `store.useEffect` depends on additional variables that shouldn't get stale?
> 
> Adding the dependencies array should allow the callback to be updated when the dependencies change.

The problem is that the callback gets updated after the subscriber has already run with old values.

src/hooks/useScrollCursorIntoView.ts
raineorshine: I have some concern about how the `setTimeout` might be affecting your results. While the `useEffect` callback is updated when the dependencies change, the timeout callback is not. There is nothing that is cancelling the timeout on unsubscribe either.

raineorshine:
> > Great, thanks! That all makes sense, and provides a helpful overview of where we are at.
> > 
> > One thing I'm noticing is that react-ministore's `useSelector` uses `useSyncExternalStore` to stay synced with React, while `useChangeEffect` does not.
> > 
> > Do you think that using `useSyncExternalStore` in `useChangeEffect` would help with the timing issues stale data?
> 
> Yeah, that sounds like it could be a good way to selectively remove synchronous behavior in cases like this where we need to wait for React before we can decide to run a subscription. I pushed a commit that switched `useChangeEffect` to use `useSyncExternalStore` in the way that I imagine it should work. It's currently re-subscribing on every scroll, which doesn't make any sense to me. I can dig deeper tomorrow, but if you see an obvious error that I made, please do point it out.


Sorry, I meant to comment on the PR rather than the issue. 

I'm not super familiar with `useSyncExternalStore`, so it's hard for me to say. I'm also not sure if it forces a re-render when the store changes, which we're trying to avoid. 

I think our next best solution is to use the ref approach and come back later to try a refactor.

src/hooks/useScrollCursorIntoView.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  // Scroll the cursor into view after it is edited, e.g. toggling bold in a long, sorted context.
+  // The cursor typically changes rank most dramatically on the first edit, and then less as its rank stabilizes.
+  editingValueStore.useEffect(() => {
+    /** The hazard here is that editingValueStore.useEffect creates a closure around the provided values.
+     * Since ministore runs synchronously, it is not possible to update dependencies before the subscribers run,
+     * and it is not possible to unsubscribe/resubscribe based on new dependencies for the same reason.
+     * Since sizeRef is an object, it is possible to mutate its properties within the existing closure after
+     * React's render cycle runs and processes the effect above. That's why setTimeout is necessary here (#3083).
+     */
```
raineorshine, line 25: Thanks for adding the explanation!

raineorshine (approved):

PR #3249, opened 2025-09-24
Connect dragSource to StaticThought, not Thought

### LINKED ISSUE #3239: Disallow drag-and-drop on empty space to the right of thoughts
Replaces #3085

## Steps to Reproduce

```
- Hello
- World
```

Image

## Current Behavior

## Expected Behavior

ethan-james:
Fixes #3239

I wanted to fix this by shrinking the actual `ThoughtContainer` so that it would fit the width of the editable inside. That didn't work well with empty thoughts, since the placeholder would wrap inside a narrow editable, and `ThoughtAnnotation` is absolutely-positioned which prevents a wrapper from exploiting its width. I remember that you mentioned possibly revisiting thought annotations, so maybe this can be revisited at that time.

I ended up passing the `dragSource` parameter down from the full-width `Thought` into the restricted-width `StaticThought`. This has two differences compared to shrinking the width of the `ThoughtContainer`:

1. Long press is still possible in the empty space. My feeling was that long press and drag-and-drop behavior should mirror each other as much as possible, but of course long press doesn't cause any harm in the scroll area so maybe it's better this way.
2. The full-width `Thought` still serves as the drop target, which seems like a good thing.

TODO: fix `dragAndDropThought` test helper to keep touches contained within the new draggable area

### DIFF
--- a/src/components/StaticThought.tsx
+++ b/src/components/StaticThought.tsx
@@ -1,5 +1,6 @@
 import _ from 'lodash'
 import React from 'react'
+import { ConnectDragSource } from 'react-dnd'
 import { useSelector } from 'react-redux'
 import { css, cx } from '../../styled-system/css'
 import { thoughtRecipe } from '../../styled-system/recipes'
@@ -7,6 +8,7 @@ import { SystemStyleObject } from '../../styled-system/types'
 import LazyEnv from '../@types/LazyEnv'
 import Path from '../@types/Path'
 import SimplePath from '../@types/SimplePath'
+import { isSafari, isTouch } from '../browser'
 import { MIN_CONTENT_WIDTH_EM } from '../constants'
 import useLayoutAnimationFrameEffect from '../hooks/useLayoutAnimationFrameEffect'
 import attributeEquals from '../selectors/attributeEquals'
@@ -30,6 +32,7 @@ import HomeIcon from './icons/HomeIcon'
 export interface ThoughtProps {
   allowSingleContext?: boolean
   debugIndex?: number
+  dragSource: ConnectDragSource
   editing?: boolean | null
   env?: LazyEnv
   // When context view is activated, some contexts may be pending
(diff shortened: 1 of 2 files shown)


ethan-james:
Unfortunately, it doesn't look like this approach is going to work either. I forgot that the `dragSource` needed to encompass both the editable and the bullet. The thought annotation is still full-width, so trying to put a wrapper around both `Bullet` and `StaticThought` is proving difficult. I could do multiple drag sources, but that seems like it will have its own problems. I'll come back to it tomorrow and see what I can do.

raineorshine:
> I wanted to fix this by shrinking the actual `ThoughtContainer` so that it would fit the width of the editable inside. That didn't work well with empty thoughts, since the placeholder would wrap inside a narrow editable, and `ThoughtAnnotation` is absolutely-positioned which prevents a wrapper from exploiting its width.

Even with a min width?

> I remember that you mentioned possibly revisiting thought annotations, so maybe this can be revisited at that time.

I've thought of cloning the editable and using that as a spacer for the annotations rather than the full `ThoughtAnnotation` that reconstructs editable in React. However that's still the same basic approach of positioning the annotations in the document flow after the editable.

Other approaches would be possible if it weren't for multiline thoughts. When a thought wraps onto multiple lines, there is no way to measure the width of the last line that I know of.

> 1. Long press is still possible in the empty space. My feeling was that long press and drag-and-drop behavior should mirror each other as much as possible, but of course long press doesn't cause any harm in the scroll area so maybe it's better this way.

My preference would be to disable long press in the scroll zone for consistency, but I'm okay with either given that long press doesn't interfere with scroll as much.

> 2\. The full-width `Thought` still serves as the drop target, which seems like a good thing.

Yes, definitely.

> Unfortunately, it doesn't look like this approach is going to work either. I forgot that the `dragSource` needed to encompass both the editable and the bullet. The thought annotation is still full-width, so trying to put a wrapper around both `Bullet` and `StaticThought` is proving difficult. I could do multiple drag sources, but that seems like it will have its own problems. I'll come back to it tomorrow and see what I can do.

Ah, that's too bad!

The only problem I see with multiple drag sources is that the browser drag preview will look different depending on whether the bullet or the thought was picked up. Theoretically we can control the drag preview, but it's browser specific and I've had some trouble in the past trying to configure this in react-dnd.

image

raineorshine:
Let's try it!

ethan-james:
I added https://github.com/cybersemics/em/pull/3252 to replace this approach.

PR #3252, opened 2025-09-25
Move dragSource into thought-container, use flexbox to shrink to fit content

### LINKED ISSUE #3239: Disallow drag-and-drop on empty space to the right of thoughts

ethan-james:
Fixes #3239 

Back on the container-shrinking approach, I was able to split up the `dragSource` and `dropTarget` in order to keep the drop target on the full-width container (`[aria-label=child]`) while using `display: flex` to shrink the drag source (`[aria-label=thought-container]`) to fit its content.

TODO: Lots of failing tests to look into.

### DIFF
--- a/src/components/Thought.tsx
+++ b/src/components/Thought.tsx
@@ -528,7 +528,7 @@ const ThoughtContainer = ({
   return (
     <div
       {...dragHoldResult.props}
-      ref={node => dragSource(dropTarget(node))}
+      ref={node => dropTarget(node)}
       aria-label='child'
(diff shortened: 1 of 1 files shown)


raineorshine:
Thanks for the heads up. In that case, I think we should probably halt work and switch to higher priorities. We can come back to this in the future if it proves troublesome to users.

PR #3254, opened 2025-09-26
Add key to ArcGradient

### LINKED ISSUE #2727: Gesture Diagram: Gradient

ethan-james:
Fixes bug introduced by #3227, which fixed #2727

Taking the same key that was applied to `<linearGradient>` and adding it to `<ArcGradient>` should fix warnings about missing keys.

### DIFF
--- a/src/components/GestureDiagram.tsx
+++ b/src/components/GestureDiagram.tsx
@@ -393,7 +393,7 @@ const GestureDiagram = ({
           ) : (
             pathSegments.map((segment, i) => {
               return rounded ? (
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Thank you!

PR #3150, opened 2025-07-29
Fix 3090: `Multiline line height is initially incorrect after unellipsizing url`

codebuild-ro:
Fixes #3090

## Issue
**Flickering**: Visual jumps when switching between single/multiline states because of  changing of line-height from `2` to `1.25`

## Approach
Implemented a consistent line-height system using CSS custom properties:

### **Architecture Changes**
- **Global CSS Custom Property**: `--thought-line-height: 1.25` in `panda.config.ts`
- **Centralized Logic**: Single memoized style object in `Thought.tsx`
- **Proper Targeting**: Line-height applied to editable container (where text is rendered)
- **Bullet Alignment**: Bullet container uses `lineHeight: 2` to compensate for new structure

### **Visual Improvements**
- **Asymmetric Padding**: `0.4em` top + `0.35em` bottom for better text alignment
- **No Flickering**: Consistent line-height across all states
- **Better Typography**: Accounts for text baseline and improves vertical centering (padding-top: 0.4em and padding-bottom: 0.3em)

### **Line-height Calculation**
```typescript
// Single-line: lineHeight (1.25) + padding (0.4em + 0.35em) = 2.0
// Multiline: lineHeight (1.25) only (padding handled by multiline recipe)
```

### **CSS Custom Properties**
```css
/* Global declaration */
:root { --thought-line-height: 1.25; }

/* Usage in components */
lineHeight: 'var(--thought-line-height)'
```

## Visual Impact
- **Before**: Flickering when switching single/multiline
- **After**: Smooth transitions with consistent line-height
- **Typography**: Better text alignment with asymmetric padding
- **Maintainability**: Single source of truth for line-height values

PR #3255, opened 2025-09-28
Puppeteer test for dead zone between editables

### LINKED ISSUE #3153: Create test for dead zone between editables
Create an automated puppeteer test to avoid this regression: https://github.com/cybersemics/em/pull/3150#pullrequestreview-3073941431

Start by proposing a test design. Once approved you can proceed to implementation.

karunkop:
Close #3153 

image

In the above picture, the left side represents the current/actual behavior of how editables are placed, i.e., with slight overlapping to fix undesirable selection behavior as discussed in https://github.com/cybersemics/em/issues/3153#issuecomment-3300494989

So in the tests it is very important to consider this overlapping gap that exists. Accordingly, I have created a test that basically tests in three different scenarios.

> **clickX** -> middle of the thought horizontally.
> **overlapHeight** -> should be -2.25px for default font size of 18. If positive we can suspect there is a gap between editables.

1. Click the mouse at the coordinates (clickX, firstThoughtBottom + overlapHeight). Normally this should set the cursor always in the first thought, as overlapHeight is negative, but if the gap is positive, then it will possibly set the cursor to the second thought or may even do nothing, which means selection should be lost and focusOffset should also be possibly 0.

2. Click the mouse at the coordinates (clickX, secondThoughtTop - overlapHeight + 1). Normally this should set the cursor always in the second thought. But if the gap exists, then this will guarantee a click will occur 1px below the firstThoughtBottom, i.e., in the gap. As above, the test should fail, as we expect selection to not be lost, but clicking in the dead zone always makes the selection lost.

3. This is very similar to 2. But this time, other than in the normal case, the click will occur exactly in the middle of the gap, and, the same as above, the assertions will verify the rest.

It would be great if we could test this in #3150 but I am not sure how to do that as of now. Then we can ensure that the assertions are properly detecting the unwanted gaps between editables


src/e2e/puppeteer/__tests__/editable-gap.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  // Check if cursor is not at the very beginning (0) or end (textContent.length)
+  // This guards against padding clicks that set cursor to edges
+  return focusOffset > 0 && focusOffset < textContent.length
+}
+
+/** Helper function to check if selection is lost (dead zone detection). */
+const isSelectionLost = async () => {
+  const selection = await getSelection()
+  const focusNode = await selection.focusNode
```
raineorshine: You can use the async chaining API I designed to avoid doing this in two lines. It's unique to the Puppeteer `getSelection` helper, but it results in syntax that's close to native browser syntax, just with `await` in front:

```ts
const focusNode = await getSelection().focusNode
```

src/e2e/puppeteer/__tests__/editable-gap.ts
```diff
@@ -0,0 +1,116 @@
...
+  return !focusNode || focusNode === null
```
raineorshine: `|| focusNode === null` appears to be redundant.

src/e2e/puppeteer/__tests__/editable-gap.ts
```diff
@@ -0,0 +1,116 @@
...
+  const secondThoughtTop = rect2.y
+
+  // Note: negative = overlap
+  const overlapHeight = secondThoughtTop - firstThoughtBottom
+
+  // click at the middle of the thought
+  const clickX = rect1.x + rect1.width / 2
+
+  // Test 1: Click in top edge of the overlap zone (should be in first thought)
+  console.info('Test 1: Clicking in top edge of the overlap zone')
```
raineorshine: Additional logging in tests is usually unnecessary since a failed assertion logs its line number, telling you exactly how far it executed. Was there another reason for adding this?

src/e2e/puppeteer/__tests__/editable-gap.ts
```diff
@@ -0,0 +1,116 @@
...
+// Intentional overlap between thoughts to fix selection behavior
+// See: useSizeTracking.ts lineHeightOverlap calculation
+// const EXPECTED_OVERLAP = DEFAULT_FONT_SIZE / 8
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+/** Helper function to check if cursor is positioned in the middle of a thought. */
+async function isCursorInMiddle(expectedThought: string): Promise<boolean> {
+  const editingText = await getEditingText()
+  if (editingText !== expectedThought) return false
```
raineorshine: I probably wouldn't bundle the editing text value assertion into `isCursorInMiddle`. Were you doing that for efficiency? I think it's probably better to let the function stick to its main purpose.

raineorshine (changes requested):
Thanks! The different assertions look reasonable to me.

> **overlapHeight** -> should be -2.25px for default font size of 18. If positive we can suspect there is a gap between editables.

Does `clip-path` affect height or click area? This might be worth confirming empirically. If it affects the click area but not the height, then an `overlapHeight` of 0 would result in a gap equal to the size of the clip-path.

src/e2e/puppeteer/__tests__/editable-gap.ts
karunkop: No, since this is still work in progress, I thought it would be helpful while I am testing this. I will remove all the info logs once the PR is ready.

src/e2e/puppeteer/__tests__/editable-gap.ts
karunkop: Yeah thanks!

src/e2e/puppeteer/__tests__/editable-gap.ts
raineorshine: Okay, thanks!

If you request a review, I would just ask that you let me know which parts are still WIP. Otherwise I will assume it is all ready to go and open to feedback.

Or you can mark it as a draft and I will avoid reviewing until you say it's ready.

karunkop:
> I was referring to the `clip-path` property, not `lineHeightOverlap`:
> 
> https://github.com/cybersemics/em/blob/abbca6a258f7166ef7ce6d88651ec11983a145c2/src/recipes/editable.ts#L18-L19
> 
> It's clear that `lineHeightOverlap` affects the distance between thoughts. However, I'm wondering about the effect of the `clip-path` CSS property. Since you are measuring the height of the element in the DOM, it is worth confirming the behavior of `clip-path`. Does `clip-path` change the height of the element, or just the clickable region? If it just changes the clickable region, then your height measurement may not properly take into account `clip-path` itself.
> 
> To put it a different way: Do your tests cover `clip-path`? If `clip-path` is accidentally removed by a developer (which has happened), do the tests fail?

I can confirm that `clip-path` property only changes the clickable region but not the actual height of the element. I have updated the PR to account for `clip-path` when calculating the overlap height.

src/e2e/puppeteer/__tests__/editable-gap.ts
karunkop: Understood!

src/e2e/puppeteer/__tests__/editable-gap.ts
```diff
@@ -0,0 +1,108 @@
...
+import clickThought from '../helpers/clickThought'
+import getEditable from '../helpers/getEditable'
+import getEditingText from '../helpers/getEditingText'
+import getSelection from '../helpers/getSelection'
+import hideHUD from '../helpers/hideHUD'
+import paste from '../helpers/paste'
+import { page } from '../setup'
+
+// 0.1em of the default font size
+const clipHeight = 0.1 * DEFAULT_FONT_SIZE
```
raineorshine: This is implicitly coupled to [editable.ts](https://github.com/cybersemics/em/blob/8f9d35089f8bf2d0b8a7183d82ad9f7ca45bb486/src/recipes/editable.ts#L18-L19). For type safety, we want to make this coupling explicit. Otherwise it's easy for these to become out of sync.

Perhaps the easiest way to couple these together is to have them both programmatically rely on a constant. If that's not possible, you can probably store the clip height in a CSS variable and then fetch that in the test when calculating `firstThoughtBottom`.

karunkop: I implemented the latter approach of storing the clip bottom value as a CSS variable handled by Panda CSS. This way we can easily retrieve property value for `editableClipBottom` and `fontSize` from the required element.

src/e2e/puppeteer/__tests__/editable-gap.ts
```diff
@@ -0,0 +1,123 @@
...
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+/**
+ * Gets the clip height from the PandaCSS CSS custom property.
+ * This ensures the test stays in sync with the {spacing.editableClipBottom} token in panda.config.ts.
+ */
+async function getClipHeight(element: ElementHandle<HTMLElement>) {
+  return await element.evaluate(el => {
+    // Get the CSS custom property value
+    const clipBottomValue = window.getComputedStyle(el).getPropertyValue('--spacing-editable-clip-bottom')
```
raineorshine: Thanks! This is good as far as programmatically coupling the value. However, we can do a little better on type safety. 

The best choice would be to use Panda style token access which I think is `token('spacing.editableClipBottom')` imported from `'../../styled-system/tokens'`. 

If that doesn't work in the test for some reason, just add an assertion that `clipBottomValue` is defined. It won't fail at compile time, but at least the test will fail if the CSS variable changes.

karunkop: Hey yeah, thanks for this suggestion. This works just fine in the test as well. However, I have a question. So for fontSize, should I just use the `DEFAULT_FONT_SIZE` directly or evaluate it as

```ts
const fontSize = parseFloat(window.getComputedStyle(el).fontSize)
```

raineorshine: I would be okay with either

raineorshine:
Great work, thanks

PR #3260, opened 2025-10-01
TreeNode: isMounted flag to track visibility of thought

### LINKED ISSUE #3089: Thought animated from position in previous view
## Steps to Reproduce

```
- a
  - =view
    - Table
  - b
    - c
  - d
    - e
```

1. Set the cursor on `b`. 
2. Toggle Table View off. 
3. Set the cursor on `d`. 

## Current Behavior

When `d` is expanded, its child `e` transitions in from its old position within the table view. 

https://github.com/user-attachments/assets/68fb055d-441d-46f4-8255-c03e9852718c

## Expected Behavior

`e` should fade in, in place, without a transition animation. The animation should be exactly the same whether or not the table view was previously activated.

When the Table View toggled off and `d` is collapsed, `e` is no longer visible and its position should not be cached. 

https://github.com/user-attachments/assets/8a34ac58-9c12-4603-808c-287f52b49872

ethan-james:
Fixes #3089 

Since components that are part of the `TransitionGroup` don't really unmount when they leave the DOM, the y positions of the tree nodes are preserved until they re-appear. Combined with #2564, which delays re-calculation of the component's y value until the next render, that means that a component will be rendered at its original position before it moves to its correct new position.

I added an `isMounted` flag that tracks whether the `CSSTransition` has called `onExit`. In the case where the component has been "unmounted", then we use the first incoming y value instead of the cached value so that the component appears in the correct spot and only has a fade transition, not a move transition on top of that.

I'd like to take a deeper look at `CSSTransition` to figure out if the `unmountOnExit` prop is supposed to actually unmount the component, and if it's being prevented from doing so by some sort of syntactical issue in the transition definitions. If it would really unmount when it leaves the DOM, then it wouldn't have a problematic cached y value in the first place.

### DIFF
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -77,6 +77,7 @@ const TreeNode = ({
   // Since the thoughts slide up & down, the faux caret needs to be a child of the TreeNode
   // rather than one universal caret in the parent.
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
+  const [isMounted, setIsMounted] = useState(false)
   // Store the on-screen index from the previous render to determine movement direction.
   const previousIndex = usePrevious<number>(index)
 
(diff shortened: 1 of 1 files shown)


raineorshine:
Great, thanks! This looks reasonable.

> I'd like to take a deeper look at `CSSTransition` to figure out if the `unmountOnExit` prop is supposed to actually unmount the component, and if it's being prevented from doing so by some sort of syntactical issue in the transition definitions. If it would really unmount when it leaves the DOM, then it wouldn't have a problematic cached y value in the first place.

`unmountOnExit` does actually unmount the component on exit, after the animation completes and the component has faded out completely. I'm not sure it interacts with cached values though. We use `unmountOnExit` depending on the scenario, though I try to use it when possible in order to get the memory savings of a full unmount.

Some more information from `FadeTransition` that might be relevant:

https://github.com/cybersemics/em/blob/4d4c5cf86552a65c4225d868dfd48bf7b20d167d/src/components/FadeTransition.tsx#L35-L36

PR #3262, opened 2025-10-02
Update page-lifecycle and y-indexeddb with tarball links

codebuild-ro:
I assumed the reason is because of upgraded yarn version.
So I replcaed git dependency url to tarball link.

1. `page-lifecycle` - from `git+https://github.com/magic-akari/page-lifecycle#feat/add-types` to `https://codeload.github.com/magic-akari/page-lifecycle/tar.gz/50b50421bdeab3d211a57e81a277f699638373b0`
2. `y-indexeddb` - from `https://github.com/raineorshine/y-indexeddb#y-indexeddb-multiplex` to `https://codeload.github.com/raineorshine/y-indexeddb/tar.gz/60b960009085b1a988b5064ee35703229231531f`

This will work because
- Yarn downloads a pre-packed .tar.gz directly from GitHub’s codeload service instead of trying to git clone + pack.
- No race conditions with tmp directories.
- Dependabot’s environment handles tarball URLs consistently.

And regarding other updates, I reverted all of them.

Also I added `codeload.github.com` to allowed host list.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
       "workspace:"
     ],
     "allowed-hosts": [
+      "codeload.github.com",
       "github.com",
       "npm",
       "yarn"
@@ -71,7 +72,9 @@
     "resetMocks": false
   },
   "resolutions": {
-    "@pandacss/node@npm:0.47.0": "patch:@pandacss/node@npm%3A0.47.0#~/.yarn/patches/@pandacss-node-npm-0.47.0.patch"
+    "@pandacss/node@npm:0.47.0": "patch:@pandacss/node@npm%3A0.47.0#~/.yarn/patches/@pandacss-node-npm-0.47.0.patch",
+    "page-lifecycle": "https://codeload.github.com/magic-akari/page-lifecycle/tar.gz/50b50421bdeab3d211a57e81a277f699638373b0",
+    "y-indexeddb": "https://codeload.github.com/raineorshine/y-indexeddb/tar.gz/60b960009085b1a988b5064ee35703229231531f"
   },
   "workspaces": [
     "packages/*"
@@ -110,7 +113,7 @@
     "murmurhash3js": "^3.0.1",
     "nanoid": "^5.1.5",
     "openai": "^5.20.3",
-    "page-lifecycle": "git+https://github.com/magic-akari/page-lifecycle#feat/add-types",
+    "page-lifecycle": "^0.1.2",
     "pluralize": "^8.0.0",
     "qrcode.react": "^4.2.0",
     "rc-slider": "^11.1.9",
@@ -144,7 +147,7 @@
     "workbox-strategies": "^7.3.0",
     "workbox-window": "^7.3.0",
     "xhtml-purifier": "^0.4.1",
-    "y-indexeddb": "https://github.com/raineorshine/y-indexeddb#y-indexeddb-multiplex",
+    "y-indexeddb": "^9.0.11-multiplex.0",
     "y-protocols": "^1.0.6",
     "yallist": "^5.0.0",
     "yjs": "^13.6.27"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,8 +10,3 @@ updates:
     schedule:
       interval: 'weekly'
     open-pull-requests-limit: 10
-    ignore:
-      - dependency-name: 'page-lifecycle'
-        update-types: ['version-update:semver-major', 'version-update:semver-minor', 'version-update:semver-patch']
-      - dependency-name: 'y-indexeddb'
-        update-types: ['version-update:semver-major', 'version-update:semver-minor', 'version-update:semver-patch']
--- a/.ncurc.js
+++ b/.ncurc.js
@@ -27,10 +27,5 @@ module.exports = {
     // Error running image diff: Unknown Error
     // https://github.com/vitest-dev/vitest/releases/tag/v2.0.0
     'puppeteer',
-
-    // Git-based dependencies that cause Dependabot issues
-    // These packages use Git URLs instead of npm registry, causing CI/CD failures
-    'page-lifecycle',
-    'y-indexeddb',
   ],
 }
--- a/.ncurc.js
+++ b/.ncurc.js
@@ -27,5 +27,10 @@ module.exports = {
     // Error running image diff: Unknown Error
     // https://github.com/vitest-dev/vitest/releases/tag/v2.0.0
     'puppeteer',
+
+    // Git-based dependencies that cause Dependabot issues
+    // These packages use Git URLs instead of npm registry, causing CI/CD failures
+    'page-lifecycle',
+    'y-indexeddb',
   ],
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -8887,7 +8887,7 @@ __metadata:
     nanoid: "npm:^5.1.5"
     npm-run-all2: "npm:^8.0.4"
     openai: "npm:^5.20.3"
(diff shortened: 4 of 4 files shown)


raineorshine (approved):
Great, thank you!

raineorshine (changes requested):
I ran the CI a few more times, and it looks like Puppeteer is failing about 50% of the time:

Image

Puppeteer is [passing consistently in main](https://github.com/cybersemics/em/actions/workflows/puppeteer.yml?query=branch%3Amain).

package.json
```diff
@@ -110,7 +113,7 @@
     "murmurhash3js": "^3.0.1",
     "nanoid": "^5.1.5",
     "openai": "^5.20.3",
-    "page-lifecycle": "git+https://github.com/magic-akari/page-lifecycle#feat/add-types",
+    "page-lifecycle": "^0.1.2",
```
raineorshine: We actually need the `#feat/add-types` branch from https://github.com/GoogleChromeLabs/page-lifecycle/pull/12. `v0.1.2` is missing type definitions.

package.json
```diff
@@ -144,7 +147,7 @@
     "workbox-strategies": "^7.3.0",
     "workbox-window": "^7.3.0",
     "xhtml-purifier": "^0.4.1",
-    "y-indexeddb": "https://github.com/raineorshine/y-indexeddb#y-indexeddb-multiplex",
+    "y-indexeddb": "^9.0.11-multiplex.0",
```
raineorshine: Can you clarify this change? I'm not sure what `^9.0.11-multiplex.0` is. I don't remember https://github.com/raineorshine/y-indexeddb/tree/y-indexeddb-multiplex getting merged into the official repo.

package.json
codebuild-ro: Would you check this part please?
I just replace with the version from the code.
https://github.com/magic-akari/page-lifecycle/blob/50b50421bdeab3d211a57e81a277f699638373b0/package.json#L3
And real reference is declared in `resolution` object.
This means the modules will be fetch from tarball link.

```
  "resolutions": {
    "@pandacss/node@npm:0.47.0": "patch:@pandacss/node@npm%3A0.47.0#~/.yarn/patches/@pandacss-node-npm-0.47.0.patch",
    "page-lifecycle": "https://codeload.github.com/codebuild-ro/page-lifecycle-fork/tar.gz/d407c14451b366fa156f53053bb42ab05d60719a",
    "y-indexeddb": "https://codeload.github.com/raineorshine/y-indexeddb/tar.gz/60b960009085b1a988b5064ee35703229231531f"
  },
```

package.json
codebuild-ro: This is also same reason with `page-lifecycle` for version and tarball link.
https://github.com/raineorshine/y-indexeddb/blob/60b960009085b1a988b5064ee35703229231531f/package.json#L3

package.json
raineorshine: Ahh, I forgot about resolutions! Do you know of any way to make this clearer for other developers? They might make the same mistake as me and not realize that it doesn't use the version listed in the package.json.

If resolutions overrides the package source, can we put any value there?

codebuild-ro: > Do you know of any way to make this clearer for other developers?

What do you think about add special explanation in readme?

> If resolutions overrides the package source, can we put any value there?

Regardless of what we put in dependencies, the resolutions value will be used.

raineorshine: The README is too far away for it to be useful for this kind of thing. It needs to be co-located. 

If the dependency value is ignored, I would recommend putting a dummy value that points to resolutions. I'm not sure if this is very conventional, but I can't think of a better way. It would be misleading otherwise.

codebuild-ro: What about removing values from resolution and use direct value into dependency values?
https://github.com/cybersemics/em/pull/3262/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R114
https://github.com/cybersemics/em/pull/3262/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R148

raineorshine: That could work, though that just obscures it in the other direction, making it hard to know where the actual source code for the dependencies lives.

.github/workflows/puppeteer.yml
```diff
@@ -51,7 +51,16 @@ jobs:
         run: yarn build
 
       - name: Serve
-        run: yarn servebuild &
+        run: |
+          yarn servebuild &
+          SERVER_PID=$!
+          echo "Server PID: $SERVER_PID"
+          sleep 10
```
raineorshine: We generally don't allow hardcoded sleep values on the project because they slow down the CI for everyone on the team. You need to wait for a specific event or use retries.

raineorshine (changes requested):
Thanks!

This looks good if it solves the original issue. We just need a way to document and link developers to the source of the dependencies, as the tarball url currently obscures where it comes from. 

See: https://github.com/cybersemics/em/pull/3262#discussion_r2436597294

.github/workflows/puppeteer.yml
```diff
@@ -51,7 +51,39 @@ jobs:
...
+          echo "Waiting for server to be ready..."
+          for i in {1..30}; do
+            if ! kill -0 $SERVER_PID 2>/dev/null; then
+              echo "Server process died!"
+              exit 1
+            fi
+
+            # Test the full request path and check for 502 errors
+            HTTP_STATUS=$(curl -s -o /dev/null -w "%{http_code}" -k https://localhost:2552 2>/dev/null || echo "000")
+
```
codebuild-ro, line 68: image

When start `Test` step, the log said that the server responds but with 502 (Bad gateway) status.
So I added script to wait until the server returns with 200 status code.

package.json
```diff
@@ -110,7 +113,7 @@
     "murmurhash3js": "^3.0.1",
     "nanoid": "^5.1.5",
     "openai": "^5.20.3",
-    "page-lifecycle": "git+https://github.com/magic-akari/page-lifecycle#feat/add-types",
+    "page-lifecycle": "⚠️ OVERRIDDEN BY 'resolutions' - Source: https://github.com/magic-akari/page-lifecycle#feat/add-types | Tarball: https://codeload.github.com/magic-akari/page-lifecycle/tar.gz/50b50421bdeab3d211a57e81a277f699638373b0 | Configured in: package.json#resolutions",
```
codebuild-ro: I added descriptive text instead of version or just source URL.

README.md
```diff
@@ -164,3 +164,27 @@ This project uses [PandaCSS](https://panda-css.com/) for styling. Styles are aut
...
+- Commit hash: `50b50421bdeab3d211a57e81a277f699638373b0`
+- Tarball URL: `https://codeload.github.com/magic-akari/page-lifecycle/tar.gz/50b50421bdeab3d211a57e81a277f699638373b0`
+
+### Updating Dependencies
+
+To update these custom dependencies:
+1. Check the source repository for new commits
+2. Get the new commit hash
+3. Update the tarball URL in the `resolutions` section of `package.json`
+4. Test thoroughly as these are custom forks
```
codebuild-ro: Added description about tarball link usage and how to make tarball link from source.

raineorshine:
@karunkop Could you confirm that this change doesn't introduce any flakiness by running the CI 10 times? Thanks!

karunkop:
> @karunkop Could you confirm that this change doesn't introduce any flakiness by running the CI 10 times? Thanks!

This has been tested in CI sequentially for 30 times, and I can confirm that there is no flakiness seen in this PR. [Here](https://github.com/karunkop/em/actions?query=branch%3Atest-flakiness-dependabot) is the result of the checks.

PR #3264, opened 2025-10-04
Animaton: Drop into collapsed thought #2533

### LINKED ISSUE #2533: Animaton: Drop into collapsed thought
## Steps to Reproduce

```
- a
- b
- c
- d
- e
```

Drag and drop `a` into `d`.

## Current Behavior

`a` disappears and the rest of the thoughts translate into place:

https://github.com/user-attachments/assets/cce01bb1-0608-4f19-8722-c90b98c47a64

## Expected Behavior

`a` should translate to the destination thought and then disappear.

This should only occur if the destination is collapsed and thus the dragged thought is unmounted. If the destination thought is expanded (or will be expanded after the drop, such as when the cursor is on the dragged thought), then it should translate into place as normal (current behavior).

zukilover:
Resolves #2533

### DIFF
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -264,6 +264,7 @@ const TreeNode = ({
                 : 'left {durations.layoutNodeAnimation} ease-out,top {durations.layoutNodeAnimation} ease-out',
         })}
         style={outerDivStyle}
+        data-thought-id={thoughtId}
       >
         <div
           className={css({
--- a/src/hooks/useDragAndDropSubThought.ts
+++ b/src/hooks/useDragAndDropSubThought.ts
@@ -26,6 +26,7 @@ import store from '../stores/app'
 import appendToPath from '../util/appendToPath'
 import ellipsize from '../util/ellipsize'
 import equalPath from '../util/equalPath'
+import fauxAnimation from '../util/fauxAnimation'
 import haptics from '../util/haptics'
 import hashPath from '../util/hashPath'
 import head from '../util/head'
@@ -174,6 +175,29 @@ const drop = (props: DroppableSubthoughts, monitor: DropTargetMonitor) => {
     const pathTo = getPathTo(state, item.path)
     return head(rootedParentOf(state, item.path)) !== head(rootedParentOf(state, pathTo))
   })
+
+  // If the destination is collapsed, animate a faux clone of the dragged thought to the destination thought's position.
+  // Capture DOM positions BEFORE dispatching the move to avoid layout shifts during animation.
+  try {
+    // only animate a single thought for now
+    if (draggedItems.length === 1) {
+      const isDestinationExpanded = isPathExpanded(state, props.path)
+      if (!isDestinationExpanded) {
+        const destinationThoughtId = head(props.path)
+        const destinationEl = document.querySelector(
+          `[aria-label="tree-node"][data-thought-id="${destinationThoughtId}"]`,
+        ) as HTMLElement | null
+        if (destinationEl) {
+          const toRect = destinationEl.getBoundingClientRect()
+          const fromThoughtId = head(draggedItems[0].path)
+          // kick off ghost animation before DOM updates
+          fauxAnimation({ fromThoughtId, toRect })
+        }
+      }
+    }
+  } catch {
+    // non-fatal: if any DOM operation fails, just skip the faux animation
+  }
 
   store.dispatch((dispatch, getState) => {
     /** Returns true if the thought should be dropped at the top of a collapsed parent. */
--- a/src/util/fauxAnimation.ts
+++ b/src/util/fauxAnimation.ts
@@ -1,0 +1,66 @@
+import ThoughtId from '../@types/ThoughtId'
+
+/** Options for the ghost drop animation. */
+type DropFauxOptions = {
+  /** The thought id of the dragged item to clone. */
+  fromThoughtId: ThoughtId
+  /** Destination rectangle (in viewport coordinates) to animate to. */
+  toRect: DOMRect
+  /** Duration in ms. */
+  duration?: number
+  /** Optional z-index for the ghost. */
+  zIndex?: number
+}
+
+/**
+ * Clones the dragged thought DOM node and animates it to the destination rect.
+ * Used when dropping into a collapsed destination where the original node unmounts immediately.
+ *
+ * Best-effort; silently aborts if DOM nodes cannot be found.
+ */
+const startDropFauxAnimation = ({ fromThoughtId, toRect, duration = 200, zIndex = 3 }: DropFauxOptions) => {
+  // find the source element rendered by TreeNode
+  const source = document.querySelector<HTMLElement>(`[aria-label="tree-node"][data-thought-id="${fromThoughtId}"]`)
+  if (!source) return
+
+  // clone the rendered element for a visual-only animation
+  const clone = source.cloneNode(true) as HTMLElement
+
+  // measure source rect after cloning to avoid reflow on source
+  const fromRect = source.getBoundingClientRect()
+
+  // absolute position in the page overlay
+  clone.style.position = 'fixed'
+  clone.style.pointerEvents = 'none'
+  clone.style.margin = '0'
+  clone.style.left = `${fromRect.left}px`
+  clone.style.top = `${fromRect.top}px`
+  clone.style.width = `${fromRect.width}px`
+  clone.style.height = `${fromRect.height}px`
+  clone.style.zIndex = String(zIndex)
+  clone.style.transition = `left ${duration}ms ease-out, top ${duration}ms ease-out, width ${duration}ms ease-out`
+
+  // place in a top-level overlay container to avoid clipping
+  const container = document.body
+  container.appendChild(clone)
+
+  // next frame to ensure initial styles are applied
+  requestAnimationFrame(() => {
+    clone.style.left = `${toRect.left}px`
+    clone.style.top = `${toRect.top}px`
+    // maintain width; destination may be narrower, but adjusting width slightly improves perceived continuity
+    clone.style.width = `${toRect.width}px`
+
+    /** Remove the clone after the animation completes. */
+    const remove = () => {
+      clone.removeEventListener('transitionend', remove)
+      if (clone.parentNode) clone.parentNode.removeChild(clone)
+    }
+    clone.addEventListener('transitionend', remove)
+
+    // fallback cleanup in case transitionend doesn't fire
+    window.setTimeout(remove, duration + 50)
+  })
+}
+
+export default startDropFauxAnimation


src/util/fauxAnimation.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    /** Remove the clone after the animation completes. */
+    const remove = () => {
+      clone.removeEventListener('animationend', remove)
+      if (clone.parentNode) clone.parentNode.removeChild(clone)
+    }
+    clone.addEventListener('animationend', remove)
+
+    // fallback cleanup in case animationend doesn't fire
+    window.setTimeout(remove, duration + 50)
```
raineorshine: I think you can subscribe to `animationcancel` and avoid the fallback.

src/components/TreeNode.tsx
```diff
@@ -264,6 +264,7 @@ const TreeNode = ({
                 : 'left {durations.layoutNodeAnimation} ease-out,top {durations.layoutNodeAnimation} ease-out',
         })}
         style={outerDivStyle}
+        data-thought-id={thoughtId}
```
raineorshine, line 268: Note that `thoughtId` is not unique for rendered thoughts: If the context view is open, the same thought may be rendered multiple times on screen at once. 

Instead, you can serialize `Path`, which is always unique.

src/util/fauxAnimation.ts
```diff
@@ -0,0 +1,73 @@
...
+  })
+}
+
+export default startDropFauxAnimation
```
raineorshine: On this project we follow a convention where the filename matches the default export name. This makes it easier to know where every function lives.

src/util/fauxAnimation.ts
```diff
@@ -0,0 +1,73 @@
...
+/** Options for the faux drop animation. */
+type DropFauxOptions = {
+  /** The thought id of the dragged item to clone. */
+  fromThoughtId: ThoughtId
+  /** Destination rectangle (in viewport coordinates) to animate to. */
+  toRect: DOMRect
+  /** Duration in ms. */
+  duration?: number
+  /** Optional z-index for the faux clone. */
+  zIndex?: number
```
raineorshine: Let's avoid adding optional parameters that are never used. We can always parameterize them in the future if needed, but better to avoid the additional complexity and creating the wrong expectations for the current code. When I saw these parameters, I assumed different values were being passed in different invocations, which turned out to be untrue. The existence of code implies its usage.

src/util/fauxAnimation.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * Clones the dragged thought DOM node and animates it to the destination rect.
+ * Used when dropping into a collapsed destination where the original node unmounts immediately.
+ *
+ * Best-effort; silently aborts if DOM nodes cannot be found.
+ */
+const startDropFauxAnimation = ({ fromThoughtId, toRect, duration = 200, zIndex = 3 }: DropFauxOptions) => {
```
raineorshine: Let's use the `durations` config and global z-index definition in the panda config rather than hardcoding values.

src/hooks/useDragAndDropSubThought.ts
```diff
@@ -175,6 +176,36 @@ const drop = (props: DroppableSubthoughts, monitor: DropTargetMonitor) => {
...
 
+  // If the destination is collapsed, animate a faux clone of the dragged thought to the destination thought's position.
+  // Capture DOM positions before dispatching the move to avoid layout shifts during animation.
+  if (draggedItems.length === 1) {
+    const isDestinationExpanded = isPathExpanded(state, props.path)
+    if (!isDestinationExpanded) {
+      const destinationThoughtId = head(props.path)
+      if (destinationThoughtId) {
+        // select by aria-label, then filter by data attribute to avoid selector injection and errors
+        const nodes = document.querySelectorAll('[aria-label="tree-node"]') as NodeListOf<HTMLElement>
```
raineorshine: I'm not sure what "selector injection and errors" means. Why can't a normal selector be used to select the thought?

src/hooks/useDragAndDropSubThought.ts
```diff
@@ -175,6 +176,36 @@ const drop = (props: DroppableSubthoughts, monitor: DropTargetMonitor) => {
...
+        }
+
+        if (destinationEl) {
+          const toRect = destinationEl.getBoundingClientRect()
+          const fromThoughtId = head(draggedItems[0].path)
+          if (fromThoughtId) {
+            // kick off faux animation before DOM updates
+            fauxAnimation({ fromThoughtId, toRect })
+          }
+        }
```
raineorshine: I would probably move all this logic to `fauxAnimation`. Notice that `useDragAndDropSubThought` does not currently have any DOM access. Putting all the direct DOM access in `fauxAnimation` keeps it more encapsulated.

src/util/fauxAnimation.ts
raineorshine: How about we call this `animateDroppedThought` or `animateToCollapsedParent`?

src/util/fauxAnimation.ts
```diff
@@ -0,0 +1,73 @@
...
+  clone.style.left = `${fromRect.left}px`
+  clone.style.top = `${fromRect.top}px`
+  clone.style.width = `${fromRect.width}px`
+  clone.style.height = `${fromRect.height}px`
+  clone.style.zIndex = String(zIndex)
+  // drive translate distances with CSS variables to use keyframes
+  const dx = toRect.left - fromRect.left
+  const dy = toRect.top - fromRect.top
+  clone.style.setProperty('--faux-dx', `${dx}px`)
+  clone.style.setProperty('--faux-dy', `${dy}px`)
```
raineorshine: Let's call this `--clone-dx` and avoid the "faux" terminology (here and elsewhere). "clone" is a bit more specific.

src/util/fauxAnimation.ts
```diff
@@ -0,0 +1,73 @@
...
+  clone.style.position = 'fixed'
+  clone.style.pointerEvents = 'none'
+  clone.style.margin = '0'
+  clone.style.left = `${fromRect.left}px`
+  clone.style.top = `${fromRect.top}px`
+  clone.style.width = `${fromRect.width}px`
+  clone.style.height = `${fromRect.height}px`
+  clone.style.zIndex = String(zIndex)
+  // drive translate distances with CSS variables to use keyframes
+  const dx = toRect.left - fromRect.left
```
raineorshine: Let's add `1em` to `dx` so that it looks like the thought is animating into a subthought position.

src/util/fauxAnimation.ts
zukilover: This has been updated.

src/components/TreeNode.tsx
zukilover: This now uses `hashPath` instead of `thoughtId`

src/util/fauxAnimation.ts
zukilover: I've renamed to `animateDroppedThought`

src/util/fauxAnimation.ts
zukilover: It now accepts 2 params only:

```js
{
  /** Hashed path of the dragged item to clone (from hashPath). */
  fromPath: string
  /** Hashed path of the destination to animate to and track during animation (for handling layout shifts). */
  toPath: string
}
```

src/util/fauxAnimation.ts
zukilover: Updated to use panda config:

```js
const duration = durations.get('fast')
const zIndex = token('zIndex.cloneDroppedThought')
```

src/hooks/useDragAndDropSubThought.ts
zukilover: It was an overly cautious approach really. I've updated it so that it uses simpler selector:

```js
document.querySelector<HTMLElement>(`[data-path="${toPath}"]`)
```

src/hooks/useDragAndDropSubThought.ts
zukilover: I've moved it to `animateDroppedThought` util.

src/util/fauxAnimation.ts
zukilover: Renamed to `animateDroppedThought`

src/util/fauxAnimation.ts
zukilover: Renamed all faux terms to clone.

src/util/fauxAnimation.ts
zukilover: I added indent offset by using the `fontSize`. I see this is how indentation is calculated in `usePositionedThoughts`. Let me know if this looks OK.

src/components/TreeNode.tsx
raineorshine: Thanks.

`data-thought-id` can be removed if it's no longer used.

src/util/animateDroppedThought.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // read destination position from CSS properties
+    const destLeft = parseFloat(destinationEl.style.left) || 0
+    const destTop = parseFloat(destinationEl.style.top) || 0
+
+    // calculate translation distances
+    // add 1em indent to make it clear the thought is becoming a subthought
+    const computedStyle = window.getComputedStyle(source)
+    const fontSize = parseFloat(computedStyle.fontSize)
+    const indentOffset = fontSize // 1em
```
raineorshine: Rather than use `getComputedStyle`, you can do the calculation in CSS: `calc(${dx}px + 1em)`.

Or add a `1em` margin-left or translate if it doesn't conflict with existing style property.

`getComputedStyle` [forces reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#calling-getcomputedstyle) which may impact performance.

raineorshine (approved):
Looks good!

src/util/animateDroppedThought.ts
zukilover: I've updated so that it uses `calc` for the indentation.

JoeDuong-dev (approved):
I’ve been testing this ticket and was able to reproduce the original issue before re-testing it on this branch. The drag-and-drop behavior is now working as expected. When the cursor is not on the dragged thought, the dragged thought correctly disappears into the parent and becomes its subthought. The user can expand the parent thought to view it, and the animation works as intended.

I also tested various edge cases with different text types, cursor positions, and drag-and-drop directions. The behavior remained consistent across all scenarios.

karunkop:
> @karunkop Can you investigate the failed test run here: https://github.com/cybersemics/em/actions/runs/18637290170/attempts/1?pr=3264
> 
> Thanks!

Hi @raineorshine, I was not able to replicate this issue in the CI. [Here](https://github.com/karunkop/em/actions?query=branch%3Atest-flakiness-faux-animation) is the result for the checks that I did.

PR #3265, opened 2025-10-04
Add packages/webview/.build to .gitignore and eslint ignore

fbmcipher:
After running `yarn build` (or a script that calls it, like `yarn cap:ios:static`), the `webview` package builds and stores artifacts in `packages/webview/.build`.

Currently these artifacts:

- get picked up as _unstaged changes_ in git, making it harder to review actual changes in PRs
- trigger ESLint errors (because they're minified files), which blocks commits without `--force`

This PR adds the `.build` directory to both git and ESLint ignore lists so build artifacts are no longer committed or linted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,3 @@ __diff_output__
 ## PandaCSS
 styled-system
 styled-system-studio
-
-# build files from webview package
-packages/webview/.build
--- a/packages/webview/.gitignore
+++ b/packages/webview/.gitignore
@@ -12,3 +12,6 @@ DerivedData
 swiftpm/configuration/registries.json
 .swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
 .netrc
+
+# build artifacts
+.build


.gitignore
```diff
@@ -31,4 +31,7 @@ __diff_output__
 
 ## PandaCSS
 styled-system
-styled-system-studio
\ No newline at end of file
+styled-system-studio
+
+# build files from webview package
+packages/webview/.build
```
raineorshine: We actually have a [nested `.gitignore`](https://github.com/cybersemics/em/blob/main/packages/webview/.gitignore) in the `packages/webview` directory, so let's add `.build` there.

raineorshine (changes requested):
Thanks for this!

Interestingly, my machine does not produce a `.build` directory after running `yarn build` with yarn v4.5.1. Can't hurt to ignore it though!

.gitignore
raineorshine: @fbmcipher

fbmcipher: done in 76a6e49 – sorry for the delay!

PR #3269, opened 2025-10-08
Test for scrollCursorIntoView after thought replication

### LINKED ISSUE #3261: Add test for `scrollCursorIntoView` after thought replication
The cursor is expected to be scrolled into view when thoughts are loaded from the db. 

I think it would be a good idea to get this case under test coverage. It will help prevent a regression when Ethan refactors `scrollCursorIntoView` for #3083.

To test this properly, it requires injecting `sleep(n)` into `replicateChildren` or `fetchDescendants`, otherwise the thoughts load so fast that the DOM is already updated by the time the initial `scrollCursorIntoView` resolves.

We could do a snapshot test, but I think reading `window.scrollY` would be cleaner.

---

## Setup

`git apply < pbpaste`

```diff
diff --git a/src/data-providers/yjs/thoughtspace.ts b/src/data-providers/yjs/thoughtspace.ts
index 6d6a4a5d58..82cc43e73a 100644
--- a/src/data-providers/yjs/thoughtspace.ts
+++ b/src/data-providers/yjs/thoughtspace.ts
@@ -18,6 +18,7 @@ import groupObjectBy from '../../util/groupObjectBy'
 import hashThought from '../../util/hashThought'
 import mergeBatch from '../../util/mergeBatch'
 import nonNull from '../../util/nonNull'
+import sleep from '../../util/sleep'
 import taskQueue, { TaskQueue } from '../../util/taskQueue'
 import throttleConcat from '../../util/throttleConcat'
 import { DataProvider } from '../DataProvider'
@@ -591,6 +592,7 @@ export const replicateChildren = async (
     remote?: boolean
   } = {},
 ): Promise<Thought[] | undefined> => {
+  await sleep(100)
   // Only await the config promise once. Otherwise the initial call to replicateChildren for a given docKey will not set thoughtDocs synchronously, and we will lose memoization of concurrent calls.
   if (!configCache) {
     await config
diff --git a/src/redux-middleware/scrollCursorIntoView.ts b/src/redux-middleware/scrollCursorIntoView.ts
index 74f356aad8..7d788e821f 100644
--- a/src/redux-middleware/scrollCursorIntoView.ts
+++ b/src/redux-middleware/scrollCursorIntoView.ts
@@ -137,7 +137,7 @@ let prevIsPulling: boolean | null = null
  */
 syncStatusStore.subscribe(state => {
   if (!state.isPulling && prevIsPulling && !userInteractedAfterNavigation) {
-    scrollCursorIntoView()
+    // scrollCursorIntoView()
   }
 
   prevIsPulling = state.isPulling
```

## Steps to Reproduce

```
- a
  - =pin
  - b
  - c
  - d
  - e
  - f
  - g
  - h
  - i
  - j
  - k
  - l
  - m
  - n
  - o
  - p
  - q
  - r
  - s
- t
```

1. Set the cursor on `t`
2. Refresh the page.

## Current Behavior

In `main`, `scrollCursorIntoView` is called on `syncStatusStore.subscribe` and the cursor is scrolled into view after all thoughts are loaded.

## Expected Behavior

With the diff applied, the cursor is pushed out of the viewport once all the thoughts are loaded, and `scrollCursorIntoView` is not activated. 

 _Originally posted by @raineorshine in [#3083](https://github.com/cybersemics/em/issues/3083#issuecomment-3357763402)_

karunkop:
Close #3261 

This test looks for the cursor thought and checks whether the scroll has occurred and is inside the viewport. If, because of some regression, `scrollCursorIntoView` is not activated, then the test will timeout as the cursor thought is not inside the viewport and the scrollY will remain 0 as well.

### DIFF
--- a/src/e2e/puppeteer/__tests__/scroll-cursor-into-view.ts
+++ b/src/e2e/puppeteer/__tests__/scroll-cursor-into-view.ts
@@ -1,0 +1,77 @@
+import clickThought from '../helpers/clickThought'
+import getEditingText from '../helpers/getEditingText'
+import paste from '../helpers/paste'
+import refresh from '../helpers/refresh'
+import waitForEditable from '../helpers/waitForEditable'
+import waitForThoughtExistInDb from '../helpers/waitForThoughtExistInDb'
+import waitUntil from '../helpers/waitUntil'
+import { page } from '../setup'
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+describe('scrollCursorIntoView', () => {
+  it('should scroll cursor into view after page refresh when cursor thought is outside viewport', async () => {
+    const importText = `
+- a
+  - =pin
+  - b
+  - c
+  - d
+  - e
+  - f
+  - g
+  - h
+  - i
+  - j
+  - k
+  - l
+  - m
+  - n
+  - o
+  - p
+  - q
+  - r
+  - s
+  - u
+  - v
+- t
+    `
+    await paste(importText)
+
+    // Verify the initial scroll position is 0
+    const initialScrollY = await page.evaluate(() => window.scrollY)
+    expect(initialScrollY).toBe(0)
+
+    await clickThought('t')
+
+    await waitForThoughtExistInDb('t')
+    await refresh()
+
+    // Wait for the cursor to be restored to thought 't'
+    await waitForEditable('t')
+
+    // Verify the editing thought is still 't'
+    const editingText = await getEditingText()
+    expect(editingText).toBe('t')
+
+    // Verify the cursor was scrolled into view after refresh
+    await waitUntil(() => {
+      const el = document.querySelector('[data-editing=true]')
+      if (!el) return false
+
+      const rect = el.getBoundingClientRect()
+      const toolbarRect = document.getElementById('toolbar')?.getBoundingClientRect()
+      const toolbarBottom = toolbarRect ? toolbarRect.bottom : 0
+
+      const viewport = {
+        top: toolbarBottom,
+        bottom: window.innerHeight,
+      }
+
+      const isInViewport = rect.top >= viewport.top && rect.bottom <= viewport.bottom
+
+      // Ensure the cursor is scrolled into view
+      return isInViewport && window.scrollY > 0
+    })
+  })
+})


src/e2e/puppeteer/__tests__/scroll-cursor-into-view.ts
```diff
@@ -0,0 +1,77 @@
...
+    const editingText = await getEditingText()
+    expect(editingText).toBe('t')
+
+    // Verify the cursor was scrolled into view after refresh
+    await waitUntil(() => {
+      const el = document.querySelector('[data-editing=true]')
+      if (!el) return false
+
+      const rect = el.getBoundingClientRect()
+      const toolbarRect = document.getElementById('toolbar')?.getBoundingClientRect()
```
raineorshine: Only reference DOM elements in tests using `data-testid` or `aria-label`, not `class` or `id`.

karunkop: I overlooked this, despite knowing it! I have amended the fix.

karunkop:
> Thanks, I like the direction this is taking.
> 
> I might suggest using a different name for `testFlags.idbDelay`. `IndexedDB` is a specific technology i.e. an implementation decision. `testFlags` is a test API. Do the tests need to know what technology the database is using? What happens when we switch sync engines? Try to design API's between modules on a "need to know" basis, and describe the purpose/function rather than implementation or technology details.
> 
> Hope that makes sense!

Thank you; this actually makes a lot of sense. I have updated the PR and renamed the testFlag API name as suggested and also added a clear purpose behind the addition of the test flag. Also, @raineorshine, do you believe this test is sufficient enough to cover the regression mentioned?

raineorshine:
Thanks!

Since #3247 hasn't been merged yet, I would expect this test to fail if it properly covers the test case, since the behavior is broken in `main`.

raineorshine (approved):

raineorshine:
Note:

> **UPDATE:** Applying the diff above does not appear to expose the bug anymore, and I can't explain why. Tested on [1286688](https://github.com/cybersemics/em/commit/1286688d97786f504494ec70c841d6e8971f9809) and [799c7fa](https://github.com/cybersemics/em/commit/799c7fab0ec28c1270ce95366e869f7d48fac890).

https://github.com/cybersemics/em/issues/3261#issuecomment-3419888303

Merging this PR anyway, since it adds good test coverage. Leaving `testFlags.replicationDelay` since it may be useful in the future, although it probably makes no difference to the outcome of this test.

Issue #3273, opened 2025-10-10
[Desktop Chrome] Drag-and-drop is cancelled too easily


raineorshine:
There are a couple ways that a drag-and-drop may be cancelled, such as the [shaker](https://github.com/cybersemics/em/blob/ee8582f3b6325996dbe86c13238ec8ea5dca03e6/src/actions/longPress.ts#L149-L151). First identify why this drag is being cancelled, and then we can discuss solutions.

## Steps to Reproduce

```
- a
- b
- c
- d
- e
- f
- g
```

Drag `a` in slowish circles over the other thoughts.

## Current Behavior

Drag is cancelled without rapid movement.

https://github.com/user-attachments/assets/53e4917a-fad3-4923-962a-46b72bc19ddb

## Expected Behavior

Drag-and-drop should only be cancelled with a rapid scribble movement.

raineorshine:
Thanks!

I'll mark this issue as resolved and open a new issue to replace the shaker.

PR #3275, opened 2025-10-13
Replace useHoveringPath hook with hover handler in useDrop

### LINKED ISSUE #3274: Replace `useHoveringPath` with `hover`
> > Aside: From an overall design perspective, it's not ideal that we are dispatching a state change (`longPress`) based on another reactive state change (`isHovering`). However, `isBeingHoveredOver` comes directly from react-dnd's `useDrop`, not our own action creator, so perhaps we don't have an opening to dispatch `longPress` further upstream.
> 
> It looks like there's a [hover](https://react-dnd.github.io/react-dnd/docs/api/use-drop) handler that, at least superficially, seems to operate on the same basis as `isBeingHoveredOver`.

Yes, I think you're right. `isBeingHoveredOver` must have been co-opted at some point to update the hovering path. But a callback like `hover` is more direct than using a reactive value + `useEffect`.

So, what do you think about replacing `useHoveringPath` with `hover` in a separate refactoring task?

_Originally posted by @raineorshine in https://github.com/cybersemics/em/issues/3272#issuecomment-3391087460_

ethan-james:
Fixes #3274 and #3278 and hopefully #3273

Move the `useHoveringPath` hook into `react-dnd`'s `hover` handler in `useDrop`. This places the `hoveringPath` functionality closer to the drag-and-drop functionality and removes an unnecessary detour into reactivity. It gives us more control over dispatching the `longPress` action creator exactly once for a given `hoveringPath` + `hoverZone`.

### DIFF
--- a/src/actions/longPress.ts
+++ b/src/actions/longPress.ts
@@ -1,5 +1,4 @@
 import _ from 'lodash'
-import { XYCoord } from 'react-dnd'
 import DragThoughtZone from '../@types/DragThoughtZone'
 import DropThoughtZone from '../@types/DropThoughtZone'
 import Path from '../@types/Path'
@@ -32,8 +31,6 @@ interface DragInProgressPayload {
   value: LongPressState.DragInProgress
   // Sets state.draggingThoughts. Either hoveringPath or file must be set if value is true.
   draggingThoughts?: SimplePath[]
-  // Get the pointer position so that subsequent dispatches can be ignored unless the pointer has moved.
-  hoverPosition?: XYCoord | null
   hoveringPath?: Path
   hoverZone?: DropThoughtZone
   // Sets state.draggingFile. Either hoveringPath or file must be set if value is true.
@@ -56,19 +53,6 @@ const longPressInactiveState = {
   hoverZone: undefined,
   simplePath: undefined,
 }
-
-/** Store the position where the last expandHoverDown action occurred in order to prevent it from firing again
- * until the pointer position changes (#3278).
- */
-let lastHoverDownPosition: XYCoord | null = null
-
-/** Compare the new hover position against the global lastHoverDownPosition to determine if the pointer has moved. */
-const hoverPositionHasMoved = (newHoverPosition: { x: number; y: number } | null) =>
-  !newHoverPosition ||
-  !lastHoverDownPosition ||
-  // There does seem to be a little bit of slop in the coordinates returned by monitor.getClientOffset()
-  Math.abs(newHoverPosition.x - lastHoverDownPosition.x) > 5 ||
-  Math.abs(newHoverPosition.y - lastHoverDownPosition.y) > 5
 
 /** Reducer for managing the state of a long press. */
 const longPress = (state: State, payload: Payload) => {
@@ -150,11 +134,6 @@ export const longPressActionCreator =
 
     switch (value) {
       case LongPressState.DragInProgress:
-        // Ignore a hover if the pointer position has not moved since the last time longPress was dispatched
-        if (!hoverPositionHasMoved(payload.hoverPosition || null)) return
-
-        if (payload.hoverPosition !== undefined) lastHoverDownPosition = payload.hoverPosition
-
         haptics.light()
         break
       case LongPressState.Inactive:
--- a/src/hooks/useDragAndDropSubThought.ts
+++ b/src/hooks/useDragAndDropSubThought.ts
@@ -39,6 +39,7 @@ import isDivider from '../util/isDivider'
 import isDraggedFile from '../util/isDraggedFile'
 import isEM from '../util/isEM'
 import isRoot from '../util/isRoot'
+import throttleByMousePosition from '../util/throttleByMousePosition'
 import { DropValidationResult } from './useDragAndDropThought'
 import useDragLeave from './useDragLeave'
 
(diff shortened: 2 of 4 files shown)


raineorshine:
You can work with @karunkop to get to the bottom of the test failure. Thanks!

raineorshine:
If `globals.lastHoverDownPosition` is no longer needed in `expandHoverUpDebounced`, we might be able to encapsulate all related functionality at the module level in the `longPress` action creator itself. That would put it in a single file, avoid the global, and avoid an extra util function. 

It would just require `newHoverPosition` be passed into `longPress`. It's a tradeoff, but pushing the complexity *down* and reducing the footprint of this escape hatch for one additional param seems worth it to me.

karunkop:
> You can work with @karunkop to get to the bottom of the test failure. Thanks!

I can confirm that the test is failing consistently in local as well, so the changes in this PR have affected the test one way or another. Working on this!

raineorshine (commented):
Thanks @ethan-james and @karunkop!

@JoeDuong-dev Ready for testing! Please check that this fixes #3278 and #3273. It would also be good to play around and test hover behavior in general to make sure nothing broke. Collapsed thoughts and hidden ancestors should expand after being hovered over for about 1 second.

> Big thanks to @karunkop for discovering that removing [this line](https://discordapp.com/channels/@me/1399814494808703119/1433402516120535060) causes the test to stop failing. I'm still not quite sure why that would be, but I have tested and verified that it doesn't seem to break subthought hover behavior. It seems that `DropChild` is only rendered on [collapsed thoughts](https://github.com/ethan-james/em/blob/479a85bcc5fbe4c2f72bf60fefe4ebed68552511/src/components/DropChild.tsx#L120) and `DropCliff` is only rendered for [cliff thoughts](https://github.com/ethan-james/em/blob/479a85bcc5fbe4c2f72bf60fefe4ebed68552511/src/components/TreeNode.tsx#L329) so I don't think we will run into any situations where hover behavior is triggered via `useDragAndDropSubThought` on any drop target that has a child that is being shown. It seems that would be the only situation where the result of `monitor.isOver({ shallow: true })` would be important.

It's totally possible that the shallow/deep distinction is no longer relevant, but I'm not 100% sure. Either way, I think we've done our due diligence, and I'll keep my eye out if anything behaves strangely.

raineorshine:
It looks like `drag` fires continuously with the same x,y coordinates, so we know the user hasn't moved. We would just need to re-enable hover once a move is detected. Is that possible?

ethan-james:
> It looks like `drag` fires continuously with the same x,y coordinates, so we know the user hasn't moved. We would just need to re-enable hover once a move is detected. Is that possible?

That's what the current version attempts to accomplish. The main issue is that the first `drag` event to trigger the [longPress](https://github.com/ethan-james/em/blob/479a85bcc5fbe4c2f72bf60fefe4ebed68552511/src/hooks/useDragAndDropThought.tsx#L311) may not represent the final mouse position once it stops moving. That's why I was saying above that we may be able to capture the mouse position at the time when [expandHoverDown](https://github.com/ethan-james/em/blob/479a85bcc5fbe4c2f72bf60fefe4ebed68552511/src/actions/longPress.ts#L177) actually [fires](https://github.com/ethan-james/em/blob/479a85bcc5fbe4c2f72bf60fefe4ebed68552511/src/actions/expandHoverDown.ts#L28). I think that would eliminate many false positives, but it requires that we continue to update the saved mouse position while the `setTimeout` is running, and then store that final mouse position once the expansion occurs.

raineorshine:
Can you remind me why #3278 is combined with the refactoring? I usually try to keep refactoring in its own PR for clean separation and better regression tracing.

ethan-james:
> Can you remind me why #3278 is combined with the refactoring? I usually try to keep refactoring in its own PR for clean separation and better regression tracing.

It's starting to seem like it wasn't the best idea, but basically it was completely dependent on the refactoring. If we were still using `isHoveringPath`, then the mouse position that we're storing would need to be sourced from somewhere completely different than the new `hover` handler.

ethan-james:
> At some point the mouse position will stabilize if the user does not move their mouse, right? So even if we miss one drag, it seems like the ping-pong issue would be prevented if we disable hoveringPath when the mouse position is unchanged. You're more familiar with the problem than I am, so maybe I'm missing something.

Oooh, so maybe we should be tracking the mouse position earlier in the process so that new hovers don't make it all the way to the `longPress` dispatch if the mouse position hasn't changed. It's a different moving part that warrants further investigation. [Earlier in the process](https://github.com/cybersemics/em/blob/86a25e4cfa195529258cedcc8ce0b7022434603f/src/hooks/useDragAndDropThought.tsx#L312) I had the mouse position check in `useDragAndDropThought` but it came after the state-based `hoveringPath` short-circuit. Maybe it could instead go first, and short-circuit based on mouse position before it considers `hoveringPath`.

> My overall feeling is that it seems like there should be a simpler solution to this problem. But I could be wrong.
> 
> I would say, take a step back, think if there is any possible way to do this more simply or statelessly, and if not proceed with your best idea. We already have slop, timing, and state, which is a lot considering the surface area of the behavior. I will give it some thought as well and possibly experiment a bit. It would be nice to avoid adding too much complexity to the hoveringPath behavior since this issue is such an edge case. Thanks!

I agree, and I got nervous about moving the state deeper into the action creator to account for the timing. Simple is good. My thought above is not too simple, but I'll let you know if I can think of anything better.

raineorshine:
Thanks @JoeDuong-dev!

> Even slight mouse movement can sometimes trigger the ping-pong effect again.

Could you share a video? I'd like to better understand this edge case. 

If the mouse is moved even a tiny bit, I would expect the hover state to update based on where the mouse is, so I'm wondering what appears unexpected to you.

If the hover state is ping-ponging when the mouse is only moved horizontally, then we could potentially change `throttleByMousePosition` to only throttle horizontal changes if that helps.

JoeDuong-dev:
@raineorshine Linking the video comment from issue 3278 here so we can track progress on this PR:

https://github.com/cybersemics/em/issues/3278#issuecomment-3536162197

raineorshine:
Yes, after seeing the video, I do think it's too sensitive.

I'm going to go ahead and merge this as it is solves several issues and the mouse sensitivity is a lower priority issue. We can return to this in the future to improve the UX.

Thank you both!

PR #3282, opened 2025-10-15
Continue to show drop child when thought is expanded

### LINKED ISSUE #3278: Drag-and-drop child drop zone causes thought to ping-pong between expanded and collapsed
In trying to fix the failing test for #3275, I ran across some inconsistencies between nested thought drag-and-drop behavior in certain browsers. All of the inconsistencies are related to small differences in the drop zones of thoughts vs subthoughts.

## Steps to Reproduce

```
- A
  - A1
    - A1-1
    - A1-2
  - B
- C
```

1. Clear the cursor
2. Drag C over A (`{ position: 'child' }`)

## Current Behavior

If the pointer is a little bit on the high side (at the very upper tip of the drop zone) then it will ping-pong between expanding and collapsing the thought's children.

https://github.com/user-attachments/assets/b502b45f-57f0-4470-8a0e-cf230e3e52ea

## Expected Behavior

When a thought's children expand, the drop zone should remain stable so that the `hoveringPath` does not change.

---

On iOS, I need to wiggle my finger a bit, but I can create the same effect:

https://github.com/user-attachments/assets/26316cdd-43a8-4843-95cf-f23814ca5e73

On Android, it doesn't really do the same thing and I need to mess with it more, but can still produce inconsistent results:

https://github.com/user-attachments/assets/682d60c7-2b6f-429d-a119-7506e67652a9

ethan-james:
Fixes #3278

Currently, this just removes the `isExpanded` check from `DropChildIfCollapsed` so that it continues to draw the drop child when the thought is expanded. From the testing I've done so far, this appears to do the trick.


ethan-james:
For reference, here is the diff of the snapshots:

drag-drop-hover-of-first-child-of-col-2-1-diff

raineorshine (commented):
> Although I do notice that the hover line under `c` changes when it expands, in a way that seems consistent with cliff padding being adjusted. So maybe it would be helpful to get a quick primer on what's most important here, if you have time. One option is to update the snapshot and accept the current positioning, if we can conclude that replacing `DropHover` with `DropChild` is OK for table column 2.
> 
>  Recording.2025-10-17.161854.mp4

This looks good to me 👍

If you update the snapshots I'll do a final review. This will also tell me if the existing tests are adequate to indirectly cover the case in #3278, or if we need a new test.

PR #3288, opened 2025-10-21
Refactor BulletCursorOverlay through abstraction

### LINKED ISSUE #3162: Improvement plan for BulletCursorOverlay

cindmichelle:
Fix #3162 with first approach.
---

Refactor BulletCursorOverlay by abstracting the shared pieces needed to compute the bullet position, eliminating the duplication that existed before.

This refactor touches
- TreeNode
- ThoughtAnnotation
-  Thought
- ContextBreadcrumb
- BulletCursorOverlay

Each abstracted component now accepts a new prop `asPlaceholder` to indicate whether it should render in hidden/placeholder mode.

### MacOS Safari
https://github.com/user-attachments/assets/d529cb03-65fc-4703-9339-01c621bc9e09

### iOS Safari

https://github.com/user-attachments/assets/3961c7b1-bdf4-4ca6-8823-45d9d94bf48e

### DIFF
--- a/src/components/ContextBreadcrumbs.tsx
+++ b/src/components/ContextBreadcrumbs.tsx
@@ -160,6 +160,8 @@ const ContextBreadcrumbs = ({
   staticText,
   thoughtsLimit,
   linkCssRaw,
+  asPlaceholder = false,
+  ariaLabel,
 }: {
   charLimit?: number
   cssRaw?: SystemStyleObject
@@ -176,6 +178,8 @@ const ContextBreadcrumbs = ({
   staticText?: boolean
   thoughtsLimit?: number
   linkCssRaw?: SystemStyleObject
+  asPlaceholder?: boolean
+  ariaLabel?: string
 }) => {
   const [disabled, setDisabled] = React.useState(false)
   const simplePath = useSelector(state => simplifyPath(state, path), shallowEqual)
@@ -202,17 +206,19 @@ const ContextBreadcrumbs = ({
 
   const homeIconStyle: React.CSSProperties = { position: 'relative', left: -1, top: 2 }
 
+  const label = hidden ? undefined : (ariaLabel ?? 'context-breadcrumbs')
+
   return (
     <div
-      aria-label={hidden ? undefined : 'context-breadcrumbs'}
+      aria-label={label}
       className={css(
         {
           fontSize: '0.867em',
-          color: 'gray66',
+          ...(!asPlaceholder ? { color: 'gray66' } : {}),
           marginLeft: 'calc(1.3em - 14.5px)',
           marginTop: '0.533em',
           minHeight: '1em',
-          visibility: hidden ? 'hidden' : undefined,
+          visibility: hidden || asPlaceholder ? 'hidden' : undefined,
         },
         cssRaw,
       )}
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -18,6 +18,7 @@ import parentOf from '../util/parentOf'
 import DropCliff from './DropCliff'
 import FadeTransition from './FadeTransition'
 import FauxCaret from './FauxCaret'
+import TreeNodeWrapper from './TreeNodeWrapper'
 import VirtualThought, { OnResize, SetSize } from './VirtualThought'
 
 /** Renders a thought component for mapped treeThoughtsPositioned. */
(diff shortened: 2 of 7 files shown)


src/components/ContextBreadcrumbs.tsx
```diff
@@ -160,6 +160,8 @@ const ContextBreadcrumbs = ({
   staticText,
   thoughtsLimit,
   linkCssRaw,
+  asPlaceholder = false,
```
raineorshine, line 163: For optional props, it's more idiomatic to leave it undefined and treat `undefined` and `false` the same, rather than explicitly set it to `false`.

src/components/ContextBreadcrumbs.tsx
```diff
@@ -160,6 +160,8 @@ const ContextBreadcrumbs = ({
...
+  ariaLabel,
```
raineorshine, line 164: Let's avoid parameterizing `ariaLabel`. It can be controlled by `asPlaceholder`, if at all.

src/components/ContextBreadcrumbs.tsx
```diff
@@ -202,17 +206,19 @@ const ContextBreadcrumbs = ({
...
+
   return (
     <div
-      aria-label={hidden ? undefined : 'context-breadcrumbs'}
+      aria-label={label}
       className={css(
         {
           fontSize: '0.867em',
-          color: 'gray66',
+          ...(!asPlaceholder ? { color: 'gray66' } : {}),
```
raineorshine, line 217: If `ContextBreadcrumbs` are hidden in the cursor overlay, then it shouldn't matter what the color is, or am I missing something?

src/components/ContextBreadcrumbs.tsx
```diff
@@ -202,17 +206,19 @@ const ContextBreadcrumbs = ({
 
   const homeIconStyle: React.CSSProperties = { position: 'relative', left: -1, top: 2 }
 
+  const label = hidden ? undefined : (ariaLabel ?? 'context-breadcrumbs')
```
raineorshine, line 209: Is `aria-label` used in the tests to differentiate the `ContextBreadcrumbs` in the cursor overlay bullet? If not, it would be nice to simplify this and not have to worry about the distinction.

src/components/TreeNode.tsx
```diff
@@ -240,6 +241,23 @@ const TreeNode = ({
...
   const type = isEmpty ? 'nodeFadeIn' : isLastActionDelete ? 'nodeDissolve' : (contextAnimation ?? 'nodeFadeOut')
+  const outerTransition = isSwap
+    ? swapDirection === 'clockwise'
+      ? 'left {durations.layoutNodeAnimation} {easings.nodeCurveXLayerClockwise}'
+      : 'left {durations.layoutNodeAnimation} {easings.nodeCurveXLayer}'
+    : isLastActionSort
+      ? 'left {durations.layoutNodeAnimation} {easings.nodeCurveSortXLayer}'
+      : contextAnimation
+        ? 'left {durations.disappearingUpperRight} ease-out,top {durations.disappearingUpperRight} ease-out'
+        : 'left {durations.layoutNodeAnimation} ease-out,top {durations.layoutNodeAnimation} ease-out'
```
raineorshine, line 252: By dis-locating props and styles from the point of use, you force the developer to, at best, scan their eyes up and down to understand the effect of the code, and at worst scroll up and down. It's better to inline it so it's co-located. If the component becomes too dense, cleave along component boundaries rather than separating props and styles.

This principle of co-location is used across the project, and is a big reason we went with PandaCSS, since it supports inline styles without losing the benefits of CSS precompiler optimizations.

src/components/ContextBreadcrumbs.tsx
cindmichelle: not yet, but when we add unit tests for BulletCursorOverlay, we’ll need the aria-label to tell the real component apart from the placeholder.

src/components/ContextBreadcrumbs.tsx
cindmichelle: youre right. this logic doesnt matter since we already set visibility hidden for the whole component

src/components/TreeNodeWrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+}: TreeNodeWrapperProps) {
+  const baseOuterStyle = {
+    left: x,
+    top: y,
+    // Table col1 uses its exact width since cannot extend to the right edge of the screen.
+    // All other thoughts extend to the right edge of the screen. We cannot use width auto as it causes the text to wrap continuously during the counter-indentation animation, which is jarring. Instead, use a fixed width of the available space so that it changes in a stepped fashion as depth changes and the word wrap will not be animated. Use x instead of depth in order to accommodate ancestor tables.
+    // 1em + 10px is an eyeball measurement at font sizes 14 and 18
+    // (Maybe the 10px is from .content padding-left?)
+    width: isTableCol1 ? (width ?? 0) : (`calc(100% - ${x}px + 1em + 10px)` as const),
+    ...(asPlaceholder && isTableCol1 ? { textAlign: 'right' as const } : {}),
```
cindmichelle, line 54: @raineorshine This is used to fix the overlay position when it is in tableCol1. Tbh i dont quite remember how I figured out that `text-align: right` could fix it 😬

https://github.com/user-attachments/assets/ea84a7ec-fa26-4d65-af7e-95407799ba81

PR #3308, opened 2025-10-24
Set shimHiddenThought if autofocus is hide-parent

### LINKED ISSUE #3256: [Chrome] Nested cursor thought shifts position when changing cursor
Tested in desktop Chrome, potentially related to #3097 

## Steps to Reproduce

```
- a
  - b
    - c
      - x
    - d
  - e
```

1. Set the cursor on `x`
2. Reload
3. Set the cursor on `c`
4. Set the cursor back on `x`
5. Set the cursor back on `c`

## Current Behavior

The layout of `x` changes slightly during the initial renders, then animates down 2 pixels when the cursor changes for the first time, and then remains stable when it is moved again. 

https://github.com/user-attachments/assets/fc670137-c78a-493a-8c48-13cd574b364b

## Expected Behavior

`x` should react the same each time the cursor is set on it and then changed. Preferably there would be no layout shift.

---

I was working on #3097, and PR #3243, in order to try to figure out why there is a layout shift in a couple of the tests for a nested thought with a cliff. I have been very curious about the [sub-pixel negative margin](https://github.com/ethan-james/em/blob/4ab019689e80e6640fd430e60c118466d01b1f5f/src/recipes/editable.ts#L15) and whether that plays a role in the failing tests, since I can't see any difference in the pixel placement of any of the virtualized thoughts, or any difference in the styles that are applied to them.

This seems like a fresh angle to follow up on, probably not related to the sub-pixel negative margin mentioned above. I also wanted to gauge whether this was a valuable thing to keep spending time on.

ethan-james:
Fixes #3256

Changing the cursor thought resulted in a hidden thought (`a` in this case) going from `autofocus=hide` to [autofocus=hide-parent](https://github.com/ethan-james/em/blob/827dbc69f38300a234baa38cb21b9535a176817f/src/selectors/calculateAutofocus.ts#L74). As a result, [shimHiddenThought](https://github.com/ethan-james/em/blob/827dbc69f38300a234baa38cb21b9535a176817f/src/components/VirtualThought.tsx#L122) would become false, and this would [remove the height](https://github.com/ethan-james/em/blob/827dbc69f38300a234baa38cb21b9535a176817f/src/components/VirtualThought.tsx#L235) from that hidden thought. Removing the height changed it from its designed value of 34px to its default value of 36px.

I assumed that it was an oversight that `hide-parent` was not being treated the same as `hide`. Let me know if that's not the case!

### DIFF
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -43,7 +43,7 @@ const useSingleLineHeight = (sizes: Index<{ height: number; width?: number; isVi
   const singleLineHeight = useMemo(() => {
     // The estimatedHeight calculation is ostensibly related to the font size, line height, and padding, though the process of determination was guess-and-check. This formula appears to work across font sizes.
     // If estimatedHeight is off, then totalHeight will fluctuate as actual sizes are saved (due to estimatedHeight differing from the actual single-line height).
-    const estimatedHeight = fontSize * 2 - 2
+    const estimatedHeight = fontSize * 2
 
     const singleLineHeightMeasured = Object.values(sizes).find(
       // TODO: This does not differentiate between leaves, non-leaves, cliff thoughts, which all have different sizes.
--- a/src/components/VirtualThought.tsx
+++ b/src/components/VirtualThought.tsx
@@ -121,7 +121,7 @@ const VirtualThought = ({
   const isVisible = zoomCursor || autofocus === 'show' || autofocus === 'dim'
   const shimHiddenThought = useDelayedAutofocus(autofocus, {
     delay: durations.get('layoutSlowShift'),
-    selector: autofocusNew => autofocus.startsWith('hide') && autofocusNew.startsWith('hide') && !!height,
+    selector: autofocusNew => autofocus === 'hide' && autofocusNew === 'hide' && !!height,
   })
 
   // console.info('<VirtualThought>', prettyPath(childPath))


ethan-james:
> Whoops, I see this comment now: https://github.com/ethan-james/em/blob/827dbc69f38300a234baa38cb21b9535a176817f/src/components/VirtualThought.tsx#L120
> 
> ```
> // Only shim 'hide', not 'hide-parent', thoughts, otherwise hidden parents snap in instead of fading in when moving up the tree.
> ```

I think that it looks OK with `hide-parent` also being shimmed. I took some videos comparing `main` with this branch, but they don't show the fade-in very well. I eyeballed the two branches and I think they look similar with the following repro:

## Steps to Reproduce

```
- a
  - b
    - c
      - x
    - d
  - e
```

1. Set the cursor on `x`
2. Reload
3. Up arrow to place the cursor on `c`
4. Up arrow to place the cursor on `b`

---

If you know of another case that I should test based on 3a729d6acf3b8f90a4291f1e181cf1761a6962bb, please let me know.

raineorshine:
**Update:** Maybe `autofocusChanged` is not behaving as expected, as that seems directly implicated in choosing the animation duration.

https://github.com/cybersemics/em/blob/0652c0b6a3537519dda75f83325ea885b22510b4/src/components/Subthought.tsx#L137-L142

We'll have to be very careful here, as the opacity and transition logic is sensitive.

https://github.com/cybersemics/em/blob/0652c0b6a3537519dda75f83325ea885b22510b4/src/components/Subthought.tsx#L111-L116

raineorshine:
> Removing the height changed it from its designed value of 34px to its default value of 36px.

I would expect these values to be identical. Can you investigate what is causing the difference?

raineorshine:
I don't know another test case off the top of my head and I don't have the bandwidth to do more in depth testing to confirm the behavior. I would encourage you to keep working on this to find a way to move forward, and if you are completely stuck we can put it on hold.

Karun recently added test coverage for the gap in #3255, and I can say that `clipPath` was worked on fairly recently and is essential for the text selection behavior as far as I know. You can work with Karun on this if you both have the availability, just be cognizant of budget.

ethan-james:
> I don't know another test case off the top of my head and I don't have the bandwidth to do more in depth testing to confirm the behavior. I would encourage you to keep working on this to find a way to move forward, and if you are completely stuck we can put it on hold.
> 
> Karun recently added test coverage for the gap in #3255, and I can say that `clipPath` was worked on fairly recently and is essential for the text selection behavior as far as I know. You can work with Karun on this if you both have the availability, just be cognizant of budget.

Sounds good, I will take a look at #3255 and I will try to eliminate the discrepancy between shimmed and un-shimmed height in hidden thoughts.

PR #3311, opened 2025-10-26
Swap parent animation is broken when keyboard is up

### LINKED ISSUE #3306: Swap parent animation is broken when keyboard is up
Tested on iOS Safari. 

## Steps to Reproduce

```
- A
  - B
```

1. Set the cursor on `B`.
2. Tap `B` to open the keyboard. 
2. Activate Swap Parent from the toolbar. 

## Current Behavior

The animation is choppy and cut off. 

https://github.com/user-attachments/assets/f6e4891b-d99d-4ca3-b9d0-20f23e848055

## Expected Behavior

There should be a smooth animation of `A` and `B` swapping places. 

https://github.com/user-attachments/assets/c48d9aa9-5d0b-46c2-8e1f-c3717357b1df

zukilover:
Fixes #3306

### DIFF
--- a/src/components/TreeNodePositioner.tsx
+++ b/src/components/TreeNodePositioner.tsx
@@ -159,6 +159,7 @@ const TreeNodePositioner = ({
       aria-label={cursorOverlay ? 'cursor-overlay-tree-node' : 'tree-node'}
       className={css({
         position: 'absolute',
+        willChange: isSwap ? 'left' : undefined,
         transition: cursorOverlay
           ? 'left {durations.layoutNodeAnimation} linear,top {durations.layoutNodeAnimation} ease-in-out'
           : isSwap
@@ -186,6 +187,7 @@ const TreeNodePositioner = ({
                 position: 'absolute',
                 width: '100%',
               }),
+          willChange: isSwap ? 'top' : undefined,
           transition: isSwap
             ? swapDirection === 'clockwise'
               ? 'top {durations.layoutNodeAnimation} {easings.nodeCurveYLayerClockwise}'


zukilover:
The broken animation is caused by caret anchoring and selection updates interfering with position-based transitions (left/top) while a contenteditable has focus.

I notice there is an extra programmatic selection work on Safari, which tries to keep the caret visible/stable:
https://github.com/cybersemics/em/blob/66a6493cc252ca8addab512b0834ea4b602be483/src/components/Editable/useEditMode.ts#L89-L94

It suppresses position transitions to preserve caret/viewport invariants.

I think we can schedule `setSelectionToCursorOffset` after `durations.get('layoutNodeAnimation')`.
The problem I'm facing now, after swap animation is complete, it does not restore the caret.

raineorshine:
> The issue seems tied to TouchBackend, which causes the animation to be suppressed on Safari when the keyboard is open, likely due to viewport or compositing changes during touch events. I’m still exploring a cleaner solution.

Thanks. That's unfortunate. I wouldn't have thought TouchBackend would affect browser selection or animations.

Does the animation break when Swap Parent is activated via gesture, or only via the toolbar? This might give us a clue about whether the tap is implicated somehow. (You would have to add a temporary gesture to the `swapParent` command to test.

> @raineorshine can you confirm if restoring the caret is required here? That’s the only piece missing from the fix above.

If the keyboard is up before Swap Parent is activated then it should stay up, and the caret will need to be restored after the render. `useEditMode` determines all the necessary conditions to restore the caret so that it happens automatically in a `useEffect` after the thought re-renders.

zukilover:
@raineorshine I've managed to restore the focus after swap parent with `focus()` method. Let me know if this works for you.

<details>
<summary>iOS</summary>

https://github.com/user-attachments/assets/7ebfa7ed-ca3b-4370-b37c-88a192384f9b


</details>

<details>
<summary>desktop</summary>

https://github.com/user-attachments/assets/b26c22d2-7154-4bb3-8af7-9bbd131dcc5b




</details>

zukilover:
I've tried with custom gesture (added `urd`), it works as expected for me:

<details>
<summary>gesture</summary>

https://github.com/user-attachments/assets/05f44789-c89e-4394-91e7-da89af722e53


</details>

raineorshine (changes requested):
Thanks for the update!

Based on my testing, this does seem to alleviate the problem, though it's not altogether gone for me. The animation seems to skip a few frames at the beginning, so it's not as smooth as the animation when there is no browser selection. However, maybe that's unavoidable with everything going on during the first hundred milliseconds after swapParent is activated. I'd be curious what @JoeDuong-dev observes.

Regarding the implementation, `setSelectionToCursorOffset` provides a unified effect that restores the browser selection in the appropriate conditions. All of the conditions are defined in one place. So, while it wouldn't be appropriate to bypass it, perhaps you can use the information gained to update the conditions.

It would be good to narrow in more specifically on why this solution works. Is it because of the move to synchronously setting the browser selection? If so, does synchronously calling `setSelectionToCursorOffset` work?

`asyncFocus` is only needed when entering edit mode, and Auto-Capitalization is not relevant here, so we do have some additional leeway in the `swapParent` case.

raineorshine (approved):

PR #3321, opened 2025-10-28
BulletCursorOverlay: Refactor Thought wrapper

### LINKED ISSUE #3162: Improvement plan for BulletCursorOverlay

cindmichelle:
Fix #3162 

Pulled an element out of the Thought component so it can be reused by both Thought and BulletCursorOverlay.
We need this element to mimic the thought’s position, which is then used to determine the cursor overlay position.

https://github.com/user-attachments/assets/f93abd3a-5660-4142-ae17-dd1003dcd10c

### DIFF
--- a/src/components/Thought.tsx
+++ b/src/components/Thought.tsx
@@ -41,7 +41,6 @@ import durations from '../util/durations'
 import equalPath from '../util/equalPath'
 import equalThoughtRanked from '../util/equalThoughtRanked'
 import getBulletWidth from '../util/getBulletWidth'
-import hashPath from '../util/hashPath'
 import head from '../util/head'
 import isAttribute from '../util/isAttribute'
 import isDescendantPath from '../util/isDescendantPath'
@@ -56,6 +55,7 @@ import ContextBreadcrumbs from './ContextBreadcrumbs'
 import DropHover from './DropHover'
 import Note from './Note'
 import StaticThought from './StaticThought'
+import ThoughtWrapper from './ThoughtWrapper'
 
 /**********************************************************************
  * Redux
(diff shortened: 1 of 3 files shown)


src/components/__tests__/SortPicker.ts
```diff
@@ -39,7 +39,7 @@ describe('Alphabetical', () => {
       const thoughtC = getThoughtByContext(['c'])
       expect(thoughtC).toBeTruthy()
 
-      const thoughts = screen.getAllByTestId(/thought/)
+      const thoughts = screen.getAllByTestId(/^thought/)
```
cindmichelle: Changed the test ID query to `^thought` so it only selects elements with `thought` prefix.
Actually, the failing tests could be fixed by setting the testId for the component with `cursorOverlay` prop to `undefined` for now, but I think this might still be useful later when we add tests for the bullet cursor overlay

raineorshine: These should probably select `aria-label='thought-container'` using `getAllByLabelText`. `data-testid` is means for looking up specific thoughts by path. Let's take the opportunity to fix these statements.

Using complex testid queries creates complex dependencies between the tests and the DOM. We should aim for simpler and more direct means of coupling the tests to the DOM.

src/components/Thought.tsx
```diff
@@ -240,6 +240,40 @@ const useCol1Alignment = ({ path, value, isTableCol1 }: UseCol1AlignParams) => {
  * Components
  **********************************************************************/
 
+/**
+ * A wrapper used inside the Thought component and BulletCursorOverlay.
+ * In BulletCursorOverlay, it’s used to maintain the height of the thought,
+ * which helps determine the cursor overlay’s position.
+ */
+export const ThoughtWrapper = ({
```
raineorshine: As a convention, we do not allow named exports in this project. All modules should export a single, default export that matches the filename. 

This allows developers to quickly know where a function or component is located without having to Find in Files or look up by Reference.

src/components/Thought.tsx
```diff
@@ -240,6 +240,40 @@ const useCol1Alignment = ({ path, value, isTableCol1 }: UseCol1AlignParams) => {
  * Components
  **********************************************************************/
 
+/**
+ * A wrapper used inside the Thought component and BulletCursorOverlay.
+ * In BulletCursorOverlay, it’s used to maintain the height of the thought,
+ * which helps determine the cursor overlay’s position.
+ */
```
raineorshine: Good comment, thanks.

PR #3329, opened 2025-11-03
Upgrade jest related dependencies

codebuild-ro:
Upgrade jest related node modules:

- @types/node                  ^22.18.0  →  ^24.9.2
- @types/sinonjs__fake-timers    ^8.1.5  →  ^15.0.1
- happy-dom                     ^19.0.2  →  ^20.0.10
- jest-environment-node         ^29.7.0  →  ^30.2.0
- jsdom                         ^26.1.0  - ^27.1.0
- vitest           ^4.0.5 → ^4.0.6
- jest-extended ^6.0.0 → ^7.0.0

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5,6 +5,13 @@ __metadata:
   version: 8
   cacheKey: 10c0
 
+"@acemir/cssom@npm:^0.9.19":
+  version: 0.9.23
+  resolution: "@acemir/cssom@npm:0.9.23"
+  checksum: 10c0/a71a7b09ca7bff4b783221bc1eaaea57e57dbf40af219c1f55a764bceed1ba90ed6e7d37e95ff14a61c95a6276b665fe8926c7a0c698dacb9a972b1e60653aa1
+  languageName: node
+  linkType: hard
+
 "@adobe/css-tools@npm:^4.4.0":
   version: 4.4.0
   resolution: "@adobe/css-tools@npm:4.4.0"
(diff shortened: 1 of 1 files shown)


raineorshine (changes requested):
> The latest `vitest` module (version 4.0.6) still uses `jsdom` v26.1.0.
> The testing doesn't work with latest version (27.1.0) of `jsdom` unfortunately.
> [vitest-dev/vitest@`main`/packages/vitest/package.json#L217](https://github.com/vitest-dev/vitest/blob/main/packages/vitest/package.json?rgh-link-date=2025-11-03T07%3A51%3A12Z#L217)
> So I think it is better to keep current version of `jsdom` until `vitest` is updated for `jsdom` v27.x.

Okay, thanks for the explanation. 

For dependencies that are intentionally not being upgraded, please update the [.ncurc.js](https://github.com/cybersemics/em/blob/main/.ncurc.js) with comments and [.depcheckrc.json](https://github.com/cybersemics/em/blob/main/.depcheckrc.json) accordingly to ensure that updates to these libraries are excluded by npm-check-updates and dependabot. We can revisit once `vitest` supports `jsdom` v27.

@karunkop Please review this when you have a moment and let me know if you have any concerns! It would be a good idea to make sure this doesn't introduce any additional peerDependency warnings either.

codebuild-ro:
> @codebuild-ro It looks like only a single test fails when jsdom is upgraded, as seen in #3332. So it seems feasible to upgrade jsdom even without vitest's internal dependency changing. What do you think?

I got it.
I will try to resolve that testing failure to upgrade jsdom.

src/components/__tests__/Superscript.ts
```diff
@@ -75,7 +75,7 @@ it('Superscript should not render on thoughts that match EM descendants', async
   await act(vi.runOnlyPendingTimersAsync)
 
   expect(() => screen.getByRole('superscript')).toThrow('Unable to find an accessible element')
-})
+}, 10000 /* increase timeout for failure in CI flow */)
```
codebuild-ro: I increased timeout value for giving more time to resolve failure in CI flow.

raineorshine: Sorry, I can't accept arbitrary timeouts in the codebase. Identify the specific condition that needs to be waited for.

raineorshine (changes requested):

karunkop:
> @karunkop I tried as you did, but the testings work fine on the local.

May I know which node and yarn version are you using locally? @codebuild-ro

codebuild-ro:
> May I know which node and yarn version are you using locally? @codebuild-ro

Node: 20.19.5
Yarn: 4.5.1

karunkop (approved):
The changes looks good to me and there are no such issues regarding test suites as well. However I have recently been seeing this warning in the console:

`The Vitest environment puppeteer-environment defines the "transformMode". This options was deprecated in Vitest 4 and will be removed in the next major version. Please, use "viteEnvironment" instead.`

I guess we can just update as vitest suggest before it is removed in the next major version. @codebuild-ro

codebuild-ro:
> `The Vitest environment puppeteer-environment defines the "transformMode". This options was deprecated in Vitest 4 and will be removed in the next major version. Please, use "viteEnvironment" instead.`

Hi @karunkop,
I updated as the vitest suggests.

raineorshine:
Thank you both!

@karunkop Did the test error go away? Are there certain node versions that don't work? Thanks!

raineorshine (approved):
Thanks!

PR #3339, opened 2025-11-03
BulletCursorOverlay: Refactor ThoughtAnnotation wrapper

cindmichelle:

---

Extract some elements from ThoughtAnnotation into a wrapper component so they can be reused by both ThoughtAnnotation and BulletCursorOverlay.

In BulletCursorOverlay, this wrapper helps fix a misalignment issue that occurs only in certain inconsistent cases, caused by one of ThoughtAnnotation’s child elements leaking its height into the container component, which then affects the positioning of adjacent components.

--- 

Please review start from this [commit](https://github.com/cybersemics/em/commit/71c1e8b050622343ffdd5aba9cec83ac705842e1)

### DIFF
--- a/src/components/BulletCursorOverlay.tsx
+++ b/src/components/BulletCursorOverlay.tsx
@@ -276,9 +276,8 @@ export default function BulletCursorOverlay({
       )}
 
       <ThoughtWrapper path={path} hideBullet={hideBullet} cursorOverlay>
-        {!hideBullet && (
-          <CursorOverlay simplePath={simplePath} path={path} leaf={leaf} isInContextView={isInContextView} />
-        )}
+        <CursorOverlay simplePath={simplePath} path={path} leaf={leaf} isInContextView={isInContextView} />
+
         <ThoughtAnnotationWrapper cursorOverlay />
       </ThoughtWrapper>
     </PlaceholderTreeNode>


src/components/ThoughtAnnotationWrapper.tsx
```diff
@@ -0,0 +1,147 @@
...
+        },
+        '@media (min-width: 560px) and (max-width: 1024px)': {
+          marginTop: { _android: '-0.1px' },
+          marginLeft: { _android: '0.5em' },
+        },
+      })}
+      style={{
+        ...(cursorOverlay
+          ? {
+              visibility: 'hidden',
```
raineorshine: Using `visibility: hidden` still renders all the React components, which can hurt performance. React rendering might be fast, but some components have additional hooks and effects that will run, e.g. `FauxCaret`. It would be better to selectively show/hide the necessary descendants at the React level.

In a normal app this might not matter, but **em** is highly performance sensitive so we have to only render what is necessary.

src/components/BulletCursorOverlay.tsx
```diff
@@ -362,9 +301,10 @@ export default function BulletCursorOverlay({
       {showContexts && simplePath?.length > 1 && <PlaceholderContextBreadcrumbs simplePath={simplePath} />}
 
       <ThoughtWrapper path={path} hideBullet={hideBullet} cursorOverlay>
-        <CursorOverlay simplePath={simplePath} path={path} leaf={leaf} isInContextView={isInContextView} />
-
-        <PlaceholderThoughtAnnotation />
+        {!hideBullet && (
+          <CursorOverlay simplePath={simplePath} path={path} leaf={leaf} isInContextView={isInContextView} />
+        )}
+        <ThoughtAnnotationWrapper isTableCol1={isTableCol1} cursorOverlay value={value} />
```
raineorshine: Is it possible to remove `isTableCol1` and `value`? They are not used in the old `PlaceholderThoughtAnnotation`, so I assume they have no effect in the `BulletCursorOverlay`.

src/components/ThoughtAnnotationWrapper.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+type ThoughtAnnotationWrapperProps = {
+  cursorOverlay?: boolean
+  ellipsizedUrl?: boolean
+  multiline?: boolean
+  value: string
+  styleAnnotation?: React.CSSProperties
+  cssRaw?: SystemStyleObject
+  style?: React.CSSProperties
+  children?: React.ReactNode
```
raineorshine: Minor point, but rather than explicitly define `children`, it's more conventional to extend `FC` and `PropsWithChildren` from `React`, e.g.:

https://github.com/cybersemics/em/blob/45204b5f0f0297eeaec4faf68076e625cabfd022/src/components/Checkbox.tsx#L51-L60

src/components/ThoughtAnnotationWrapper.tsx
```diff
@@ -0,0 +1,147 @@
+import { css, cx } from '../../styled-system/css'
+import { multilineRecipe } from '../../styled-system/recipes'
+import { SystemStyleObject } from '../../styled-system/types'
+import { MIN_CONTENT_WIDTH_EM } from '../constants'
+import isAttribute from '../util/isAttribute'
+import FauxCaret from './FauxCaret'
+
+type ThoughtAnnotationWrapperProps = {
```
raineorshine: Normally props would need to be documented better with JSDOM comments, but you can leave this for now since we intend on dissolving this component once we fix the height leakage problem.

src/components/BulletCursorOverlay.tsx
cindmichelle: yep can do

src/components/BulletCursorOverlay.tsx
```diff
@@ -362,9 +301,10 @@ export default function BulletCursorOverlay({
       {showContexts && simplePath?.length > 1 && <PlaceholderContextBreadcrumbs simplePath={simplePath} />}
 
       <ThoughtWrapper path={path} hideBullet={hideBullet} cursorOverlay>
-        <CursorOverlay simplePath={simplePath} path={path} leaf={leaf} isInContextView={isInContextView} />
-
-        <PlaceholderThoughtAnnotation />
+        {!hideBullet && (
```
raineorshine: Same question as: https://github.com/cybersemics/em/pull/3342#discussion_r2511590154

PR #3340, opened 2025-11-03
BulletCursorOverlay: Render actual ContextBreadcrumb inside BulletCursorOverlay

cindmichelle:

---

Add `cursorOverlay` to indicating that the component being used inside BulletCursorOverlay. For now, the additional prop is used for controlling the visibility of the context breadcrumb inside BulletCursorOverlay. ContextBreadcrumb is needed to be import into BulletCursorOverlay to handle the case when context view is active

---
## Results

**When context view is active**

https://github.com/user-attachments/assets/71b43b02-c9ca-474c-a4e6-76dfea019464

### DIFF
--- a/src/components/ContextBreadcrumbs.tsx
+++ b/src/components/ContextBreadcrumbs.tsx
@@ -160,7 +160,6 @@ const ContextBreadcrumbs = ({
   staticText,
   thoughtsLimit,
   linkCssRaw,
-  cursorOverlay,
 }: {
   charLimit?: number
   cssRaw?: SystemStyleObject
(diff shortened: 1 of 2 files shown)


src/components/ContextBreadcrumbs.tsx
```diff
@@ -213,7 +216,7 @@ const ContextBreadcrumbs = ({
           marginLeft: 'calc(1.3em - 14.5px)',
           marginTop: '0.533em',
           minHeight: '1em',
-          visibility: hidden ? 'hidden' : undefined,
+          visibility: hidden || cursorOverlay ? 'hidden' : undefined,
```
raineorshine: Perhaps you can reuse `hidden` and avoid increasing the surface area of the API?

PR #3342, opened 2025-11-05
BulletCursorOverlay: Refactor TreeNode wrapper

cindmichelle:

---
Extract shared div containers from TreeNode; reuse in BulletCursor to support x/y position transitions

### DIFF
--- a/src/components/BulletCursorOverlay.tsx
+++ b/src/components/BulletCursorOverlay.tsx
@@ -143,58 +143,6 @@ function CursorOverlay({
 }
 
 /**
- * PlaceholderTreeNode is a component used to mimic behavior of TreeNode.
- * Any position changes from one Thought to another will be animated within this component.
- */
-function PlaceholderTreeNode({
-  children,
-  isTableCol1 = false,
-  x,
-  y,
-  width,
-}: {
-  children?: React.ReactNode
-  isTableCol1?: boolean
-  x: number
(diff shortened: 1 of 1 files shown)


src/components/BulletCursorOverlay.tsx
```diff
@@ -358,14 +309,27 @@ export default function BulletCursorOverlay({
...
+      width={width}
+      path={path}
+      index={index}
+      isMounted
+    >
       {showContexts && simplePath?.length > 1 && <PlaceholderContextBreadcrumbs simplePath={simplePath} />}
 
       <ThoughtWrapper path={path} hideBullet={hideBullet} cursorOverlay>
-        <CursorOverlay simplePath={simplePath} path={path} leaf={leaf} isInContextView={isInContextView} />
+        {!hideBullet && (
```
raineorshine: Was this condition added by mistake? I believe `hideBullet` should only be used to hide the bullet, not the cursor overlay. The user still needs to see where the cursor is in tables or other places where the bullets aren't visible.

cindmichelle: Is it okay for it to look like this when the bullet is hidden?
image

raineorshine: Yes. We can play with the style, but there always needs to be some cursor indicator visible.

PR #3347, opened 2025-11-06
Command center buttons

### LINKED ISSUE #3229: [Liminal UI] Command Center
> **Begin the Liminal UI design implementation with a new design for the Command Center – an interactive overlay using complex background glow, blurring and blending techniques to achieve its luminous effect.**

# Background

Image

The Command Center is an overlay that provides quick access to common commands in **em**.

When the user selects a thought via tap-and-hold, the Command Center slides onto the screen, along with a flourish of colour – which creates visual separation between the thoughtspace/editor and the commands.

**Initially**, the Command Center shows the user a **standard set of commands**. If the user scrolls, or taps the ⌃ arrow, the Command Center transitions into an **expanded view** which provides access to all commands in the app. This state shift is accompanied by a subtle animation.

The Command Center is only available on mobile, where it will ultimately replace the existing toolbar.

> **I recommend trying the [Figma prototype](https://www.figma.com/proto/oQnevfxv79fhTJnOs0fz9k/em-design?page-id=3320%3A11662&node-id=3331-13691&viewport=-185%2C-878%2C0.4&t=CzPMjkxksO8O4btV-1&scaling=contain&content-scaling=fixed&starting-point-node-id=3331%3A13691&show-proto-sidebar=1) for an interactive representation of the design!** This will help you visualise how the Command Center should look, feel and animate on the screen.
> 
> Click and hold a thought in the prototype to open the Command Center.

# Development Stages

We'll start by focussing on just implementing the **first stage of the new Command Center** – the **initial view** – which will give us a good sense of the basics.

* **Stage 1: Initial View**
	- #3230 
	- #3231 

After **Stage 1**, we'll tackle increasingly complex elements with more interactivity:

- **Stage 2: Expanded View**
- **Stage 3: Search**

And lastly, once all commands are accessible in the Command Center, we'll replace the toolbar with a more mobile-friendly navigation bar.

- **Stage 4: Replace Toolbar with simplified nav (Undo/Redo/Search)**

yangchristina:
Recreating https://github.com/yangchristina/em/pull/270 to point to correct base

resolves #3229

Screenshot 2025-10-20 at 5 28 38 PM

### DIFF
--- a/src/components/CommandMenu/PanelCommand.tsx
+++ b/src/components/CommandMenu/PanelCommand.tsx
@@ -42,7 +42,7 @@ const ActiveButtonGlowImage: FC<ActiveButtonGlowImageProps> = ({ isActive, type 
           backgroundGradient: 'activeGlow',
           borderRadius: '0px',
           pointerEvents: 'none',
-          ...(type === 'luminosity' ? { mixBlendMode: 'luminosity' } : { mixBlendMode: 'saturation' }),
+          mixBlendMode: type === 'luminosity' ? 'luminosity' : 'saturation',
           filter: 'blur(23px)',
           padding: 40 /* Expands the Paint Rect so it includes all of the blur. */,
           margin: -40 /* Pulls it back so layout doesn't break. */,
@@ -103,7 +103,7 @@ const PanelCommand: FC<PanelCommandProps> = ({ command, size }) => {
       >
         {SVG && (
           <SVG
-            style={{ justifySelf: size === 'small' ? 'center' : 'center' }}
+            style={{ justifySelf: 'center' }}
             size={size === 'small' ? 24 : size === 'medium' ? 22 : 24}
             animated={isAnimated}
             animationComplete={() => setIsAnimated(false)}


yangchristina:
@fbmcipher I've moved the PR here, so we can continue the discussion here. Previously on https://github.com/yangchristina/em/pull/270

src/components/CommandMenu/CommandMenu.tsx
```diff
@@ -222,7 +232,7 @@ const CommandMenu = () => {
                 gridTemplateColumns: 'repeat(4, 1fr)',
                 gridTemplateRows: 'auto',
                 gridAutoFlow: 'row',
-                gap: '0.7rem',
+                gap: '0.75rem',
```
fbmcipher: Replace with:

```
gap: 0.7rem
gridRowGap: 1rem
```

src/components/CommandMenu/CommandMenu.tsx
```diff
@@ -175,11 +173,12 @@ const CommandMenu = () => {
...
             })}
           />
+          <Overlay />
+
           <div
             className={css({
               position: 'relative',
-              zIndex: 1,
-              margin: '0 1.2rem calc(1.2rem + env(safe-area-inset-bottom)) 1.2rem',
+              margin: '0 1.9rem calc(1.9rem + env(safe-area-inset-bottom)) 1.9rem',
```
fbmcipher: Replace with:

```
margin: 0 1.5rem calc(1.5rem + env(safe-area-inset-bottom)) 1.5rem
```

(looks better across all phones, both small and large)

src/recipes/panelCommandRecipe.ts
```diff
@@ -3,77 +3,107 @@ import { defineRecipe } from '@pandacss/dev'
...
+    boxSizing: 'border-box',
+    padding: '0.5rem 1rem',
+    borderRadius: 48,
+    overflow: 'hidden',
+    mixBlendMode: 'color-dodge',
+
     cursor: 'pointer',
     transition: 'opacity {durations.medium} ease, background-color {durations.medium} ease',
     opacity: 1,
+    height: 58,
```
fbmcipher: Replace with:

```
height: 60
```

fbmcipher (changes requested):
Thanks @yangchristina – moving the conversation here.

> @fbmcipher I see, thanks for explaining this! I think the mock has another 0.5rem for the gap, which I've added [52a56c6](https://github.com/yangchristina/em/pull/270/commits/52a56c60c482b4a0c68ab2e6926677a48e93587c)

Thank you, this is better – but I think it still needs a little bit of tweaking. The *gap between rows* is intentionally larger than the *gap between columns*.

To speed us along here, I've suggested changes to the code that will get things in the right place.

In general, that might help speed things along – when the general implementation is spot-on and we're just pixel-peeping, it might be easier for me to suggest those changes to you directly, rather than us going back and forth.

How does that sound?

fbmcipher:
Lastly (and unfortunately) there was one small bug I spotted while testing on iOS, reproducible on both simulators and physical devices running iOS 18 and iOS 26.

---

## Steps to Reproduce

1. Tap and hold a thought to select it and open the Command Center.
2. Press the Favorite button.

## Current Behavior
The selection state for the Favorite command gets cut off by an invisible line, and abruptly appears and disappears without an animation.

CleanShot 2025-11-12 at 23 11 34@2x

## Expected Behavior
The selection state for the command fades in gracefully, and doesn't get cut off.

## Notes
- I think this might have something to do with the falloff.
- Animation-wise, we can probably use a simple fade using one of the predefined constants in `durations`.

src/components/CommandMenu/CommandMenu.tsx
yangchristina: Thanks for providing these! 
397a380

src/components/CommandMenu/CommandMenu.tsx
yangchristina: https://github.com/cybersemics/em/commit/397a380f89edbaf8c19e8b59296e06d6413f79c0

src/recipes/panelCommandRecipe.ts
yangchristina: https://github.com/cybersemics/em/commit/397a380f89edbaf8c19e8b59296e06d6413f79c0

yangchristina:
@fbmcipher that's weird, I'm getting a different issue

https://github.com/user-attachments/assets/d6a1c7ab-a67f-4cfe-b64f-9d0c7b2aae4b


https://github.com/user-attachments/assets/06d9713a-2581-4b30-aa9e-5aad1e1ecbec

yangchristina:
> The selection state for the Favorite command gets cut off by an invisible line, and abruptly appears and disappears without an animation.

@fbmcipher I'm not sure why, but adding `backdrop-filter: blur(0.1px)` appears to fix it. I don't really know why, more of an accidental discovery. f9b2a72

Is this still happening for you?

yangchristina:
> 3. The command active glow should fade in and out using a duration

@fbmcipher by this you mean when toggling favourites right?

https://github.com/user-attachments/assets/a520ab6c-34a4-49d2-994f-4f31042d579e

I couldn't get it to work with `FadeTransition`, since there were issues with mix-blend-mode when transitioning opacity, if the opacity transition wasn't on the element with the blur. Is this acceptable from a performance perspective? Because the active glow will always be rendered, with just it's opacity changing depending on `isActive`. using `FadeTransition` would have been preferred, because then we can unmount on exit, but I don't see a non-complex way to do this without`FadeTransition` that still allows for unmounting on exit.

[078eb05](https://github.com/cybersemics/em/pull/3347/commits/078eb05f53b401e9bd92a2ead77c3670f41e1cd2)
[ebf4fab](https://github.com/cybersemics/em/pull/3347/commits/ebf4fab590f22aaba24069aea6b1686145b7573e)

yangchristina:
> Sadly, `will-change` is super memory intensive, so I would ask you to look for a different solution.

@raineorshine I see, sounds good. 
The three options I see right now are:

- `transform: 'translateZ(0)',`
- `'-webkit-backdrop-filter': 'blur(0.1px)'`
- `will-change: 'opacity'

I think `will-change: opacity` is actually supposed to be the best option memory-wise? But I haven't really found good sources for this.

yangchristina:
> Conclusion: I think will-change is okay as long as we use it correctly

@fbmcipher thank you for this! Yes I agree, I believe `will-change` was added as a way to stop people from using hacks like `transform: translateZ(0)`, so it should be better than `transform: translateZ(0)`.

I am using `willChange` in `ActiveButtonGlowImage` for command menu. It is only applied to safari, and we have `unmountOnExit`, so I think it should be fine performance wise? `willChange` will only be applied for active commands when the command men is open.
```
_safari: {
            willChange: 'opacity',
},
```
Actually I'll see if I can switch willChange off

raineorshine:
Thanks for doing all of the research into `will-change`, compositor layers, and performance implications. Very helpful.

> By comparison, `will-change` (when set to either `opacity` or `transform`) _just_ creates a compositor layer. [(ref)](https://web.dev/articles/simplify-paint-complexity-and-reduce-paint-areas)

That's good to know the additional memory usage is limited to the creation of a new compositor layer. That should be more predictable for us to design around.

> Worse still, the layers are kept in memory – **even if the Command Center is closed.**

How is this possible of the Command Center has unmounted? Since the element is no longer in the DOM, we couldn't remove `will-change` programmatically if we wanted.

> In other words, in an animation we should:
> 
> * start with `will-change` **disabled**
> * turn it on just before the animation plays
> * leave it on for the duration of the animation
> * then turn it **off**
> 
> That way, the compositing layer is created only when needed, and cleaned up when not needed. That'll help us minimise our memory footprint.

That seems reasonable.

> ### For context, we are still being memory efficient
> As a reference, here are some popular mobile web applications and their GPU memory usage/layer count. Numbers according to Chrome DevTools connected to an Android phone on **initial load**. Take these numbers with a pinch of salt, as memory usage in DevTools _is_ an estimate, but it's useful context.
> 
> * Twitter: 82 layers, 326MB GPU memory
> * Reddit: 18 layers, 169MB GPU memory
> * NY Times: 9 layers, 424MB GPU memory
> 
> And for reference, **em** with the CC open (prior to `backdrop-filter` fix):
> 
> * **em**: 90 layers, 20MB GPU memory

How many visible thoughts are on screen in this test? Real-world usage can have 2,000+ (including hidden parents/uncles that are ready to animate in).

Blend modes might be fine for the Command Center since it's a fixed cost, but if we apply any blend mode to thoughts then it's going to be O(n).

---

Note that **em** currently experiences performance issues whenever the tree is re-rendered, including on navigation or a command. If CPU is the bottleneck, then we should have some room to use `will-change` and blend modes. If GPU is the bottleneck, then it will amplify the current performance issues.

fbmcipher:
> @yangchristina I am using willChange in ActiveButtonGlowImage for command menu. It is only applied to safari...

Sounds good to me!

Now we know what to look out for, let's be sure to keep an eye out for rendering quirks between Safari and Chrome in the future.

Only using `will-change` on Safari here is a fine way to optimise – just as long as the result ends up looking the same across both platforms. We don't want to accidentally introduce visual inconsistencies between platforms.

fbmcipher:
> @yangchristina `willChange` will only be applied for active commands when the command men is open.

This doesn't seem to be true. When the Command Center closes, it doesn't get unmounted – it stays off-screen, and therefore button states still change when the focussed thought changes.

e.g. if the user has a favorited thought selected, even if the Command Center is closed the Favorite button glows – thus creating those compositor layers and doing extra undesired work in the background.

I think let's look at this in the _next PR_, since we're changing the Command Center's drawer implementation anyway. I'll add it to the list of requirements there.

fbmcipher:
> @yangchristina Actually I'll see if I can switch willChange off

That would be great! It's a small optimisation, but it would be great to be able toggle willChange off when animations complete. Then we're fully following best practices and being as efficient as possible. It would also mitigate the issue of compositor layers sticking around even after the Command Center is closed.

That seems to be our last task until this PR is ready for review and merge!

src/components/CommandMenu/PanelCommand.tsx
```diff
@@ -1,19 +1,56 @@
...
+      <div
+        ref={nodeRef}
+        className={css({
+          gridArea: 'command',
+          objectFit: 'contain',
+          objectPosition: 'center',
+          backgroundGradient: 'activeGlow',
+          borderRadius: '0px',
+          pointerEvents: 'none',
+          ...(type === 'luminosity' ? { mixBlendMode: 'luminosity' } : { mixBlendMode: 'saturation' }),
```
raineorshine: I'm assuming that `mixBlendMode: type` doesn't work here because it's not statically analyzable, but we can still simplify slightly:

```ts
mixBlendMode: type === 'luminosity' ? 'luminosity' : 'saturation'
```

src/components/CommandMenu/PanelCommand.tsx
```diff
@@ -47,33 +83,45 @@ const PanelCommand: FC<PanelCommandProps> = ({ command, size }) => {
...
+          panelCommandRecipe({
+            isButtonExecutable,
+          }),
+          css({ gridArea: 'command' }),
+        )}
+        {...fastClick(handleTap)}
+      >
+        {SVG && (
+          <SVG
+            style={{ justifySelf: size === 'small' ? 'center' : 'center' }}
```
raineorshine: Looks like this is always `'center'` and can be simplified.

src/durations.config.ts
```diff
@@ -41,6 +41,8 @@ const durationsConfig = {
   disappearingLowerLeft: 500,
   /* Duration for command menu swipable drawer animation. */
   commandMenuDrawer: 400,
+  activeButtonGlowLuminosity: 400,
+  activeButtonGlowSaturation: 400,
```
raineorshine, line 45: Let's stick with the standard duration `medium` since the active button does not use a special value or require animation coordination between different elements as far as I can tell.

yangchristina: This behaviour comes from how `FadeTransition` works. The `timeout` prop **must match the CSS transition duration**, so we currently do:

```tsx
timeout={durations.get(type)}
```

Because of that, any new `type` added to `fadeTransitionRecipe` also needs an entry in `durationConfig` so that `durations.get(type)` returns a value.

If we wanted to avoid updating `durationConfig` every time, we could fall back to a default:

```tsx
timeout={type in durationsConfig ? durations.get(type) : durations.get('medium')}
```

But I think it's still better to be explicit. Requiring the duration to be added to the config prevents mistakes where a new transition type is introduced without a matching timeout, which would cause the timeout and CSS duration to get out of sync.

src/components/CommandMenu/PanelCommand.tsx
yangchristina: 7de9fa0

src/components/CommandMenu/PanelCommand.tsx
yangchristina: 7de9fa0

src/durations.config.ts
raineorshine: Ah, I see. Thanks for the reminder.

In some cases I wish the animation type could bet a bit more independent from the animation. But in other cases a particular animation should always be executed with a specific duration. So it's hard to know how tightly to couple them in the code 🤔

JoeDuong-dev (approved):

raineorshine (approved):

Issue #3085, opened 2025-06-28
Do not allow drag to be initiated in scroll zone


raineorshine:
## Steps to Reproduce

```
- This is a long thought that keeps going and going until it wraps onto two lines and becomes a multiline thought.
```

Tap and hold the text of the thought in the far right scroll zone, then begin dragging.

## Current Behavior

Drag-and-drop is allowed.

## Expected Behavior

Drag-and-drop should be disabled in the scroll zone. 

This will help prevent false negatives when the user is trying to scroll. The scroll zone is quite small now, so there's plenty of space on the rest of the screen for the user to drag.

ethan-james:
Am I correct to assume that this also applies to long presses without drag-and-drop?

ethan-james:
I guess it would make sense to allow a long press without moving your finger, but a long press begins before their intent is clear....

raineorshine:
You raise a good point. Long press should work as usual, since it doesn't interfere with scrolling. I do expect long pressing the right edge of a long thought to still open the Command Center. However, this feels inconsistent. Normally long press can always lead into a drag-and-drop. It's strange to long press, highlight the bullet, and then not allow drag-and-drop.

I have an idea for a more conservative approach that will reduce some of the false positive drags without removing functionality. See #3239.

Sorry to invalidate your work. Your understanding of the problem will no doubt help with the new task!

PR #3355, opened 2025-11-07
Split dragSource into dragSourceBullet and dragSourceEditable

### LINKED ISSUE #3239: Disallow drag-and-drop on empty space to the right of thoughts

ethan-james:
Fixes #3239 

Splitting into two drag sources works pretty well. I also moved the `draggable` attribute down into `Bullet` and `StaticThought` separately, and only apply the `dragSource` to `StaticThought` on mobile.

We are also back to the situation described in https://github.com/cybersemics/em/issues/3085#issuecomment-3310017154 where long press is still triggered by the thought container. Do you want me to try to move that into `Bullet` and `StaticThought` as well?

### DIFF
--- a/src/e2e/puppeteer/__tests__/multiselect.ts
+++ b/src/e2e/puppeteer/__tests__/multiselect.ts
@@ -40,8 +40,8 @@ describe('mobile only', () => {
     const a = await waitForEditable('a')
     const b = await waitForEditable('b')
 
-    await longPressThought(a, { edge: 'right', x: 100 })
-    await longPressThought(b, { edge: 'right', x: 100 })
+    await longPressThought(a, { edge: 'right' })
+    await longPressThought(b, { edge: 'right' })
 
     // In CI, sometimes the count of highlighted bullets are incorrect. The selector query runs immediately after both long presses, but react might not have finished re-rendering all bullet components.
     // Wait for the command menu panel to show "2 thoughts selected" before we query for highlighted bullets.


JoeDuong-dev (approved):
The fix for https://github.com/cybersemics/em/issues/3239 has been tested and verified. The highlighted green area in the example scenario no longer triggers a drag-and-drop action when the user taps and holds. This behavior was tested with various text lengths and layouts to ensure the functionality works correctly across different scenarios.

src/components/BulletPositioner.tsx
```diff
@@ -294,6 +298,7 @@ const BulletPositioner = forwardRef<SVGSVGElement, PropsWithChildren<BulletPosit
         data-testid={cursorOverlay ? undefined : 'bullet-' + hashPath(path)}
         aria-label={cursorOverlay ? undefined : 'bullet'}
         data-highlighted={cursorOverlay ? undefined : isHighlighted}
+        ref={dragSource ? dndRef(ref => dragSource(ref)) : undefined}
```
raineorshine, line 304:
```suggestion
        ref={dragSource ? dndRef(dragSource) : undefined}
```

src/components/Thought.tsx
```diff
@@ -537,14 +546,10 @@ const ThoughtContainer = ({
...
       {...dragHoldResult.props}
-      ref={dndRef(node => dragSource(dropTarget(node)))}
+      ref={dndRef(node => dragPreview(dropTarget(node)))}
       aria-label='child'
       data-divider={isDivider(value)}
       data-editing={isEditing}
-      // HTML5Backend will override this to be "true" on platforms that use it.
-      // iOS Safari needs it to be true to disable native long press behavior. (#2953, #2931, #2964)
-      // Android works better if draggable is false.
-      draggable={isTouch && isSafari()}
```
raineorshine, line 547: Did something change on iOS Safari where it is no longer needed?

raineorshine (changes requested):
Thank you! There are a few other things that need to be brought into alignment in the UX when long pressing the empty space for it to be a true dead zone:

- No bullet highlight
- No Alert
- No Command Center

Sorry for not mentioning these explicitly in the original issue!

ethan-james:
> Thank you! There are a few other things that need to be brought into alignment in the UX when long pressing the empty space for it to be a true dead zone:
> 
> * No bullet highlight
> * No Alert
> * No Command Center
> 
> Sorry for not mentioning these explicitly in the original issue!

That's OK, I asked earlier if long press should be given the same treatment, and it sounds like the answer is yes.

src/components/Thought.tsx
ethan-james: I moved it into `StaticThought` where the `dragSource` is now. I neglected to add `draggable` to `Bullet` as well, but I just tested it and it seems OK without on iOS Safari. Probably because the native editable long press behavior that we were over-riding doesn't exist on the bullet.

raineorshine: I see now! Thanks!

ethan-james:
I moved `dragHoldResult.props`, which controls long press behavior, down into `StaticThought` and `BulletPositioner`. Since long press behavior isn't as sensitive as drag behavior, I wasn't sure whether it would be better to call `useDragHold` in each of those components rather than passing the props down into them. Let me know if you have a preference.

PR #3357, opened 2025-11-08
ThoughtAnnotation: calculate position based on selection range

### LINKED ISSUE #3352: ThoughtAnnotation: Calculate position instead of using invisible placeholder
## Background

`ThoughtAnnotation` is an absolutely positioned placeholder that is used to position the superscript and url link immediately to the right of the thought text. It is a separate element in order to avoid interfering with the contenteditable (e.g. which has additional padding-right for usability).

## Problem

This design is expensive in both complexity and performance. Every thought with a superscript has to maintain an invisible placeholder which updates in real-time during editing.

Note that a simple `right` position of the bounding box will not work because the last character of a multiline thought may end well before the right edge:
Image

## Solution

However, it looks like a character range can be used to calculate the exact x,y position of the last character in multiline text. Could this be used to position the superscript instead of the invisible placeholder?

e.g.

```js
function getLastCharacterXPosition(el) {
  const range = document.createRange();
  const textNode = el.lastChild;
  if (!textNode || textNode.nodeType !== Node.TEXT_NODE) return null;

  // Select the last character
  range.setStart(textNode, textNode.length - 1);
  range.setEnd(textNode, textNode.length);

  // Get bounding box
  const rect = range.getBoundingClientRect();

  // rect.right gives you the x position (relative to viewport)
  return rect.right;
}
```

ethan-james:
Fixes #3352 

As discussed, getting the bounding rect of a [selection range](https://github.com/ethan-james/em/blob/1700e10d50f6805897c77fc5f9af733ca5ab5bcc/src/components/ThoughtAnnotation.tsx#L342) seems to work well for positioning the thought annotation without duplicating the thought in an invisible container.

TODO:

- ~~get the positioning right for superscripts, urls, email~~
- ~~get the positioning right for the end-of-thought faux caret~~
- ~~create a separate selection to position the start-of-thought faux caret correctly~~
- ~~get editable's bounding rect for empty thoughts with annotations~~

---

I remember that we had an issue when we were building the faux caret where the end-of-thought position was being reported inaccurately by iOS Safari. This may also impact the positioning of superscript, which would mean we need to update some snapshots. It most likely won't create a visual discrepancy in the same way that replacing the real caret with the faux caret often does.

### DIFF
--- a/src/hooks/usePositionedAnnotation.ts
+++ b/src/hooks/usePositionedAnnotation.ts
@@ -1,4 +1,4 @@
-import { RefObject, useCallback, useEffect, useMemo, useState } from 'react'
+import { RefObject, useCallback, useEffect, useLayoutEffect, useMemo, useState } from 'react'
 import { useSelector } from 'react-redux'
 import Path from '../@types/Path'
 import isDescendantPath from '../util/isDescendantPath'
@@ -83,12 +83,10 @@ const usePositionedAnnotation = (
 
   // useSelector would be a cleaner way to get the editableRef's new position
   // but, on load, the refs are null until setTimeout runs
-  useEffect(() => {
+  useLayoutEffect(() => {
     setOpacity('0')
-    if (!inTransition) {
-      positionAnnotation()
-      setOpacity('1')
-    }
+    positionAnnotation()
+    setOpacity('1')
   }, [
     descendant,
     editableRef,


src/components/BulletCursorOverlay.tsx
```diff
@@ -195,7 +195,7 @@ export default function BulletCursorOverlay({
             isTableCol1={isTableCol1}
           />
         )}
-        <ThoughtAnnotationWrapper cursorOverlay />
```
raineorshine, line 193: @ethan-james Can you explain this change? The `cursorOverlay` prop is used to set the positioner to overlay only. Since the cursor overlay is not used to render an actual thought, it can skip a lot of internal processing and effects. 

If this prop is removed here in `BulletCursorOverlay`, I'm not sure where else it is used. 

Please discuss with @cindmichelle who is responsible for the cursor overlay implementation. Once you come to a consensus, I can check in. Thanks!

src/components/ThoughtAnnotation.tsx
```diff
@@ -223,6 +214,11 @@ const ThoughtAnnotationContainer = React.memo(
       isEditing ? editingValue : null,
     )
 
+    // We're trying to get rid of contentWidth as part of #3369, but currently it's the easiest reactive proxy for a viewport resize event.
+    const contentWidth = viewportStore.useSelector(state => state.contentWidth)
+    const cursor = useSelector(state => state.cursor)
```
raineorshine: I want to check in about the potential performance implications of this. This would cause every visible `ThoughtAnnotation` instance to re-render whenever the cursor changes. Does this differe from `main`? Can this be narrowed down?

src/components/BulletCursorOverlay.tsx
ethan-james: It seemed to me that `cursorOverlay` was only being used to decide whether or not to render the `thoughtStart` faux caret inside of `ThoughtAnnotationWrapper`. @cindmichelle , does that sound right to you? I moved that faux caret outside of `ThoughtAnnotationWrapper` because it needed to be positioned relative to a parent instead, so that it could align with the left side of that parent.

raineorshine: My apologies, I didn't notice that `cursorOverlay` was successfully removed from `ThoughtAnnotationWrapper`, too. I thought it was left unused for some reason, but I was mistaken. I think we're good here!

src/components/ThoughtAnnotation.tsx
```diff
@@ -121,65 +119,56 @@ const ThoughtAnnotation = React.memo(
...
-     * So now subscribing to get context so that StaticSuperscript is not re-rendered for all lexemeIndex change.
-     * It will re-render only when respective Lexeme is changed.
-     * Changed as part of fix for issue 1419 (https://github.com/cybersemics/em/issues/1419).
-     */
-
-    const textMarkup = useSelector(state => {
-      const labelId = findDescendant(state, head(simplePath), '=label')
-      const labelChild = anyChild(state, labelId || undefined)
-      return isEditing ? (liveValueIfEditing ?? value) : labelChild ? labelChild.value : value
-    })
```
raineorshine, line 148: Was this functionality removed intentionally?

ethan-james: It showed as unused after I deleted the container where the text was being rendered. I didn't look too closely at what it did, so maybe there was useful stuff in there. I'll take another look.

raineorshine: It's used to force a re-render of the superscript. If #1419 doesn't regress, then it's safe to remove.

ethan-james: #1419 looks OK.

src/components/ThoughtAnnotation.tsx
ethan-james: Yes, thanks, I think I can re-use `isEditing` to re-render the annotations when a URL ellipsizes instead of doing it every time the cursor changes.

src/components/ThoughtAnnotationWrapper.tsx
raineorshine: I believe the goal would be to remove the hidden `<span>` entirely since it is no longer used to position the annotations. Is this possible, or are there other things that depend on it?

If it is still required, maybe we can put together a rough plan of what it would take to remove it.

I'm also open to doing this in another PR if having an extra step in the commit history after merge would be helpful.

ethan-james: If I remove the span, then it messes up the superscripts:

color-theme-superscript-on-light-theme-1-diff

I thought it might be as simple as adding an explicit height to the container, such as `height: multiline ? '1.25em' : '2em'`. That did fix the superscripts, but it caused other layout problems:

color-theme-superscript-on-light-theme-1-diff

I went ahead and removed the `textMarkup` and `placeholder` props so that it will only ever render `'&ZeroWidthSpace;'`. In terms of removing the span entirely, I think that there are two moving parts:

1. Setting the height of the container correctly so that it leaves room for annotations without pushing lower thoughts out of place
2. Figuring out if anything breaks now that we are relying on typographic CSS such as [verticalAlign: 'text-top'](https://github.com/ethan-james/em/blob/cf05215cf7deaf6674dcbb168b4391358177edd7/src/components/ThoughtAnnotation.tsx#L47) while no longer including any text in the annotation.

I wouldn't mind doing this in a separate PR so that we can get this one merged, but it's up to you.

raineorshine:
Thanks. I'm bringing the discussion here as there seems to be more at stake than originally anticipated. 

> If I remove the span, then it messes up the superscripts:
> 
> 
> 
> color-theme-superscript-on-light-theme-1-diff
> 
> 
> 
> I thought it might be as simple as adding an explicit height to the container, such as `height: multiline ? '1.25em' : '2em'`. That did fix the superscripts, but it caused other layout problems:
> 
> 
> 
> color-theme-superscript-on-light-theme-1-diff
> 
> 
> 
> I went ahead and removed the `textMarkup` and `placeholder` props so that it will only ever render `'&ZeroWidthSpace;'`. In terms of removing the span entirely, I think that there are two moving parts:
> 
> 
> 
> 1. Setting the height of the container correctly so that it leaves room for annotations without pushing lower thoughts out of place
> 
> 2. Figuring out if anything breaks now that we are relying on typographic CSS such as [verticalAlign: 'text-top'](https://github.com/ethan-james/em/blob/cf05215cf7deaf6674dcbb168b4391358177edd7/src/components/ThoughtAnnotation.tsx#L47) while no longer including any text in the annotation.
> 
> 


This is giving me some doubts about the current solution and whether I want to merge. The main point of this task was to remove the dependency on the the invisible placeholder element and reduce complexity. Now we have a hybrid solution which relies partly on DOM measurement and partly on the placeholder, which seems to me less good than either pure approach. 

We began with the theory that the superscript can be rendered a fixed number of em up from the bottom-right corner of the editable, without any placeholders. Is this not true? Maybe better understanding the path to true decoupling from the placeholder will help us get to the right solution.

ethan-james:
> > 1. Setting the height of the container correctly so that it leaves room for annotations without pushing lower thoughts out of place
> > 2. Figuring out if anything breaks now that we are relying on typographic CSS such as [verticalAlign: 'text-top'](https://github.com/ethan-james/em/blob/cf05215cf7deaf6674dcbb168b4391358177edd7/src/components/ThoughtAnnotation.tsx#L47) while no longer including any text in the annotation.
> 
> This is giving me some doubts about the current solution and whether I want to merge. The main point of this task was to remove the dependency on the the invisible placeholder element and reduce complexity. Now we have a hybrid solution which relies partly on DOM measurement and partly on the placeholder, which seems to me less good than either pure approach.
> 
> We began with the theory that the superscript can be rendered a fixed number of em up from the bottom-right corner of the editable, without any placeholders. Is this not true? Maybe better understanding the path to true decoupling from the placeholder will help us get to the right solution.

Maybe I've been treating `vertical-align` as more magical than it truly is. I removed to span to check things out, and nudging the superscript went pretty well. There's one table view test that's failing. I can look into that more tomorrow to see if it's related to `vertical-align`, or I can update the snapshot if you'd rather move forward with it as it is.

raineorshine:
The updated snapshot is okay with me!

raineorshine (changes requested):
Thanks!

There are a couple regressions related to Table View and Context View, which makes sense given the different layouts used there. Hopefully we can come up with a fairly general solution. I'll let you investigate and determine the effort involved.

I've added snapshot tests in ab158c293d to cover the cases that were missing, so you should be able to see the tests fail if you merge `main` now.

ethan-james:
> Thanks!
> 
> There are a couple regressions related to Table View and Context View, which makes sense given the different layouts used there. Hopefully we can come up with a fairly general solution. I'll let you investigate and determine the effort involved.
> 
> I've added snapshot tests in [ab158c2](https://github.com/cybersemics/em/commit/ab158c293dcbda1557e80e2000d3bdef740d21db) to cover the cases that were missing, so you should be able to see the tests fail if you merge `main` now.

Thanks for doing that, it was helpful for troubleshooting. I fixed the table view test completely, and I fixed the major issue with the context view where it was calculating the position of the thought annotation before the `disappearingUpperRight` animation was complete. That issue isn't reflected in the test, which instead shows small discrepancies that look similar to the other snapshots that we've updated.

I didn't see anything obviously different about the positioning of a context view thought compared to a regular thought, and indeed the superscript for `a` is off even thought it's not part of the active context view. I tried making small adjustments to the positioning [here](https://github.com/ethan-james/em/blob/5bc4502007543b14279fe9e936fd3b0c9a5b843e/src/components/ThoughtAnnotationWrapper.tsx#L31) or [here](https://github.com/ethan-james/em/blob/5bc4502007543b14279fe9e936fd3b0c9a5b843e/src/components/StaticSuperscript.tsx#L56), and ended up in a familiar situation where other tests start failing before the superscript nudges up by a pixel on `a`.

Anyway, I would recommend updating the snapshot for the context view test. Let me know if you want me to dig deeper instead.

src/components/ThoughtAnnotation.tsx
```diff
@@ -279,13 +282,69 @@ const ThoughtAnnotationContainer = React.memo(
...
+              if (!isAtEdge) right = rect.right - offset.left + (isTableCol1 ? 12 : 0)
+            }
+
+            // rect.right gives you the x position (relative to viewport)
+            annotationRef.current.style.left = `${right}px`
+            annotationRef.current.style.top = `${top}px`
+            if (isInContextView && display !== undefined) annotationRef.current.style.display = display
+          }
+        },
+        isInContextView ? durations.get('disappearingUpperRight') : 0,
```
raineorshine: I found a couple cases where this delay doesn't work. 

One is if the context view is rapidly toggled:

https://github.com/user-attachments/assets/cf61b970-6701-4304-b0d4-285a6c5c1c7a

Another is if the cursor is quickly moved before the animation completes:

https://github.com/user-attachments/assets/5eb9cf27-206e-481e-9da1-37d85bf29646

src/components/ThoughtAnnotationWrapper.tsx
```diff
@@ -13,41 +10,26 @@ import FauxCaret from './FauxCaret'
...
       className={css({
         position: 'absolute',
         pointerEvents: 'none',
         userSelect: 'none',
         boxSizing: 'border-box',
-        width: '100%',
-        // maxWidth: '100%',
-        marginTop: '0',
+        lineHeight: multiline ? 1.25 : undefined,
+        marginTop: multiline ? 'calc(-0.12em - 0.5px) !important' : value ? '-0.425em' : undefined,
```
raineorshine: Let's avoid the use of `!important`. There should be other ways to set up the styles or create Panda recipes to allow style overriding in a less brute force way.

raineorshine (changes requested):
Thanks for the updat!

> Anyway, I would recommend updating the snapshot for the context view test. Let me know if you want me to dig deeper instead.

If the discrepancies are small and the position looks correct when using the app, I'm okay with updating the snapshots.

raineorshine:
Thanks! To thoroughly test the thought annotation, I would test superscripts, urls, code, and colors within normal view, table view, and context view.

Navigate around, resize the window, etc. This is not mobile specific so that should save you some time.

Note that there are some layout bugs related to table view in `main` so you'll need to compare the behavior with `main` to confirm if it's an actual regression or not.

Some possible test data: 
*(This can be copied and pasted directly into em)*
```
- 1
  - a
    - b
      - c
        - =view
          - Table
        - d
          - d1
        - e
          - e1
            - f
        - stackoverflow.com
          - stackoverflow.com
- 2
  - a
    - b
      - c
        - =view
          - Table
        - d
          - d1
        - e
          - e1
            - f
- 3
  - url
    - stackoverflow.com
  - code
```

src/hooks/useContextAnimation.ts
```diff
@@ -0,0 +1,41 @@
+import { useSelector } from 'react-redux'
+import Path from '../@types/Path'
+import isContextViewActive from '../selectors/isContextViewActive'
+import isDescendantPath from '../util/isDescendantPath'
+
+/** Return an animation name for a thought that is appearing or disappearing as a result of toggling a context view.
+ * This is used to display the animation in TreeNode, and also to determine whether thought annotations should delay
+ * their position calculation until after the animation has completed (#3352).
+ */
+const useContextAnimation = (path: Path, isAppearing?: boolean) => {
```
raineorshine: This hook is too thin to justify the abstraction, as it is just a wrapper around `useSelector`. It would be better to factor out the selector and use `useSelector(contextAnimation(path, isAppearing))` as needed.

src/hooks/useContextAnimation.ts
```diff
@@ -0,0 +1,41 @@
...
+    )
+    if (!isLastActionContextView) return null
+
+    // Determine the animation direction for disappearing text
+    let animation: 'disappearingLowerLeft' | 'disappearingUpperRight' = 'disappearingLowerLeft'
+
+    if (isDescendantPath(path, state.cursor)) {
+      const isCursorInContextView = isLastActionContextView
+        ? !!state.cursor && isContextViewActive(state, state.cursor)
+        : false
```
raineorshine: Literal `true` and `false` can always be reduced out of ternary expressions. This is equivalent to:

```ts
isLastActionContextView && !!state.cursor && isContextViewActive(state, state.cursor)
```

src/components/ThoughtAnnotationWrapper.tsx
```diff
@@ -13,41 +10,26 @@ import FauxCaret from './FauxCaret'
...
       className={css({
         position: 'absolute',
         pointerEvents: 'none',
         userSelect: 'none',
         boxSizing: 'border-box',
-        width: '100%',
-        // maxWidth: '100%',
-        marginTop: '0',
+        lineHeight: multiline ? 1.25 : undefined,
+        marginTop: multiline ? 'calc(-0.12em - 0.5px)' : value ? '-0.425em' : undefined,
```
raineorshine, line 34: Could you add some comments explaining the `marginTop` and `marginLeft` values? It's not clear if they are arbitrary and independent, or tightly coupled to other styles such as `multilineRecipe`.

src/components/ThoughtAnnotationWrapper.tsx
```diff
@@ -99,42 +71,6 @@ const ThoughtAnnotationWrapper: FC<
           minWidth: `${MIN_CONTENT_WIDTH_EM - 0.333 - 0.333}em`, // min width of thought (3em) - 0.333em left padding - 0.333em right padding
         }}
       >
-        <span
```
raineorshine, line 102: Yesssss delete it all 😈

src/components/ThoughtAnnotation.tsx
```diff
@@ -279,13 +284,78 @@ const ThoughtAnnotationContainer = React.memo(
...
+
+          top = rect.top - offset.top
+          // offset annotation container to account for -12px left margin in ThoughtPositioner #3352
+          if (!isAtEdge) right = rect.right - offset.left + (isTableCol1 ? 12 : 0)
+        }
+
+        // rect.right gives you the x position (relative to viewport)
+        annotationRef.current.style.left = `${right}px`
+        annotationRef.current.style.top = `${top}px`
+        annotationRef.current.style.opacity = '1'
```
raineorshine: Is there an advantage to setting the styles directly on the annotation ref vs a more React-centric approach of using local state?

src/components/ThoughtAnnotation.tsx
```diff
@@ -279,13 +284,78 @@ const ThoughtAnnotationContainer = React.memo(
       setCalculateContexts(true)
     }, [])
 
+    const positionAnnotation = useCallback(() => {
+      if (editableRef.current && annotationRef.current) {
```
raineorshine: You can invert this condition and `return` to short circuit the function and make this a little cleaner by avoiding the extra indentation level.

src/components/ThoughtAnnotation.tsx
```diff
@@ -279,13 +284,78 @@ const ThoughtAnnotationContainer = React.memo(
       setCalculateContexts(true)
     }, [])
 
+    const positionAnnotation = useCallback(() => {
```
raineorshine: This could use a header comment summarizing the purpose of this function.

It may also be worth it to pull `positionAnnotation` and the `useEffect` out into a hook in order to hide the implementation details from the rest of the component.

src/components/ThoughtAnnotation.tsx
ethan-james: `annotationRef` gets passed from `ThoughtAnnotationContainer` into `ThoughtAnnotation`, but there's probably no reason why left/top/opacity props couldn't get passed in as well. The `setTimeout` may complicate things a bit but I can try it out.

src/components/ThoughtAnnotation.tsx
ethan-james: I think the messiness of making a hook out of this comes from the large dependency array:

```
[
      contentWidth,
      contextAnimation,
      descendant,
      editableRef,
      email,
      fontSize,
      isEditing,
      isInContextView,
      isTableCol1,
      numContexts,
      positionAnnotation,
      showSuperscript,
      styleAnnotation,
      url,
    ]
```

`contextWidth` is going away soon and will need to be replaced with a resize handler, which is a positive case for encapsulating this logic I think. `email`, `isEditing`, `numContexts`, `showSuperscript`, `styleAnnotation` and `url` are all of the ones that could change from outside the component.

src/components/ThoughtAnnotation.tsx
raineorshine: Worth a try. We want to avoid direct DOM manipulation when possible.

src/components/ThoughtAnnotation.tsx
raineorshine: You could make a hook that returns the effect callback, and leave the `useEffect` in the component. It's a slightly odd abstraction boundary, but it does save you from having to pass all the arbitrary dependencies into the hook.

I'll leave it up to you to decide.

src/components/ThoughtAnnotationWrapper.tsx
ethan-james: For multiline, I consolidated the top margins from the children in ae2661a61c687b70ea4f55fa7a8a4c4d049f1ef9.

ethan-james: For `marginLeft`, I also consolidated `calc(1em - 18px)` and `-0.333em` into a single style, which is how it ended up as `calc(0.666em - 18px)`.

src/components/ThoughtAnnotation.tsx
ethan-james: I moved the hook invocation from `ThoughtAnnotationContainer` into `ThoughtAnnotation`, which allows me to get rid of the dependencies that [determine whether the annotation shows at all](https://github.com/ethan-james/em/blob/865b9ae880d4d15772174cbc1ed10562bdeca5a4/src/components/ThoughtAnnotation.tsx#L284).

raineorshine:
Are you testing the sample data above? I'm seeing a few positioning issues:

image

image

src/selectors/getContextAnimationName.ts
raineorshine, line 1: Let's put this under `/src/selectors` and remove the currying so it follows the convention of the other selectors.

ethan-james:
I think I fixed all of the table and double-annotation issues. The only thing left that I know of is the `code` issue. It turns out that [const textNode = editableRef.current.lastChild](https://github.com/ethan-james/em/blob/5732a9c844f9bbad77f4468b8698a821e9e63130/src/hooks/usePositionedAnnotation.ts#L36) stops working when a style is applied and a `font` tag is inserted. It works OK if the style is only applied to the first part of the thought, but a thought with a style applied to all of it no longer has a `textNode` to reference as a direct child of `editableRef.current`. Do you want me to keep going and make that logic more robust in order to solve what is hopefully the last positioning issue, or do you think we should re-evaluate?

raineorshine:
Thanks for the update!

Yes, we will need to handle nested HTML. It may also occur with `<b>`, `<i>`, etc as well.

`editableRef.current.lastChild` seems a bit fragile since it is tightly coupled not just to the main DOM node of a different component, but to the structure of its descendants. I would be nice if there was a way to decouple that, but we can focus on behavioral integrity before we consider further refactoring.

src/components/StaticSuperscript.tsx
```diff
@@ -52,7 +56,7 @@ const StaticSuperscript = React.forwardRef<
               className={css({
                 position: 'relative',
                 zIndex: 'stack',
-                top: '-1px',
+                top: type === 'multiline' ? '-2em' : type === 'singleline' ? '-2.745em' : undefined,
```
raineorshine: A couple of thoughts about the `top` value here:

1. Would it be possible set these on an ancestor? It seems like a breach of the encapsulation boundary that `StaticSuperscript` should have to know about whether it is single line or multiline. `StaticSuperscript` is a simple component that shouldn't have to know much about outside context or where it is being used.
2. I assume these hardcoded values are tightly coupled to the line-height? If so, they should be programmatically controlled and commented.

raineorshine (changes requested):
Thanks for giving some thought to the coupling with `editableRef`. Looking at the new helper function `getFinalTextNode`, I think I'm leanings towards keeping the existing direct reference rather than trying to decouple it. None of the refactoring options are super attractive. Importantly, `getFinalTextNode` is pretty general. While there is a tight coupling with the descendant text node, it's not like it relies on a specific DOM structure. It just assumes that there is a text node *somewhere* within `editableRef`. I think this is okay. Regrettably, there are a lot of components that depend on `editableRef` directly so we don't have great separation of concerns between the different thought wrappers, but that is what it is.

Detailed testing coming soon.

src/components/StaticSuperscript.tsx
ethan-james: I tried to create a wrapper in `ThoughtAnnotation` that could have positioning applied directly to it since none of the other invocations of `StaticSuperscript` use its `absolute` prop, so I could remove both `type` and `absolute` at the same time. I tweaked the positioning to account for changes in font size, etc., and got all of the tests except one to pass with the following:

`top: url || email ? '-0.425em' : multiline ? '-1.25em' : 'calc(-2em + 3.75px)'`

Then I checked it out in the browser, and saw:

Screenshot 2025-12-16 142951

I updated the positioning to look like `top: '-0.425em'`, and everything looks better in the browser:

Screenshot 2025-12-16 142241

This seems like a great solution, except that all of the tests are failing. I'll give a bit more thought to why there would be such a large discrepancy, and please let me know if you can think of anything. I think that my assessment for now is that these elements are still a bit too brittle to survive much refactoring.

ethan-james: I'm actually going to push my changes so that we can make sure it looks the same when the CI test suite runs.

raineorshine: I'm sorry, I'm not able troubleshoot this or get too into the details myself. I look forward to seeing what you come up with.

ethan-james: I was able to fix the tests by re-introducing some typography (a zero-width space) back into the annotation element. It's hard to say exactly why this is necessary because it looked fine everywhere except inside Puppeteer. I tried replacing the character with `height: multiline ? '1.25em' : '2em'` to make the annotation wrapper occupy the required amount of space, but that didn't fix the problem and it did break the layout in other tests. If you'd like, I can take another look tomorrow to see why typography is still so important to the layout of this element, or I can wrap the zero-width space so that it only shows up in CI.

There was one 13px superscript test that was slightly off after moving the positioning into a wrapper around `StaticSuperscript`, but I went ahead and updated that.

raineorshine: Interesting! I wonder if the zero-width space is necessary for the same reason that the ThoughtAnnotation is necessary for layout in BulletCursorOverlay (#2882). Cindy and I were surprised to discover that the cloned ThoughtAnnotation was affecting layout despite it being position absolute.

Discord 2025-12-22 05 01 33

https://github.com/user-attachments/assets/2ed9330d-f18c-4489-84b7-39c282788c91

> If you'd like, I can take another look tomorrow to see why typography is still so important to the layout of this element, or I can wrap the zero-width space so that it only shows up in CI.

Let's revisit that another time. Could you link to this comment in the code? It might be helpful to have a reference more specific than git blame if some time passes before we look at this again.

> There was one 13px superscript test that was slightly off after moving the positioning into a wrapper around `StaticSuperscript`, but I went ahead and updated that.

👍

ethan-james:
I think I've managed to fix the incorrect positioning, but unfortunately the fix creates a bit of a flicker in some situations. A better solution would be to remove `setTimeout` from normal (non-animated) annotations, and that almost works but it seems to cause problems with multiline thoughts, especially when a multiline URL becomes the cursor thought and goes from ellipsized to un-ellipsized.

I thought maybe I could watch for changes to `multiline` and treat that as a special case, but I'm not sure it would solve the flicker, and it didn't work anyway. When a URL goes from ellipsized to un-ellipsized (multiline) it is measured [asynchronously](https://github.com/ethan-james/em/blob/56ccf65ada7c01193f08cba8f9c3b9e48222f793/src/components/Editable/useMultiline.ts#L43) and is not updated until it's too late to update positioning. I'll let you know if I can think of a better solution.

ethan-james:
Actually, I realized that I can use `multiline` as a dependency to trigger an async re-positioning for multiline thoughts, eliminating the need for a timeout if the thought isn't multiline. Seems to work OK so far!

ethan-james:
> I can confirm that [Edited thought superscript position incorrect on first paint](https://github.com/cybersemics/em/pull/3357?notification_referrer_id=NT_kwDOAAtyxLIyMDI3ODI5ODgxNzo3NTAyNzY#pullrequestreview-3602545278) is fixed.
> 
> I'm still seeing the broken behavior after toggle context view: [Deactivated context view superscript position incorrect on first paint](https://github.com/cybersemics/em/pull/3357?notification_referrer_id=NT_kwDOAAtyxLIyMDI3ODI5ODgxNzo3NTAyNzY#issuecomment-3680330877).
> 
> The test case from [#3357 (comment)](https://github.com/cybersemics/em/pull/3357#issuecomment-3620946333) still looks solid.

It seems like when virtualized thoughts re-enter the viewport after the context view thoughts leave, they come back with the annotation in the wrong position. I'm not sure why that would be the case, but hiding the annotation until positioning occurs seems to reliably fix the issue.

fbmcipher:
@raineorshine Sure thing – I'll run a couple quick tests and post my results here tomorrow.

ethan-james:
> 1. It looks like some superscripts are being hidden that don't need to be hidden. It should only affect children that are being animated in/out, not the cursor thought, siblings, or ancestors.

It seems like the context animation is applied to all thoughts, not just ones that are entering or leaving the context view. Probably because it's tough to determine which thoughts are animating out and leaving the DOM. Do you think it's worthwhile to try to narrow that down so that the animation is not applied to thoughts that are higher up than the cursor thought?

ethan-james:
> > 1. It looks like some superscripts are being hidden that don't need to be hidden. It should only affect children that are being animated in/out, not the cursor thought, siblings, or ancestors.
> 
> It seems like the context animation is applied to all thoughts, not just ones that are entering or leaving the context view. Probably because it's tough to determine which thoughts are animating out and leaving the DOM. Do you think it's worthwhile to try to narrow that down so that the animation is not applied to thoughts that are higher up than the cursor thought?

It seems like using [belowCursor](https://github.com/ethan-james/em/blob/1740231c9f67c71fd2157df650ea5fb232c1d97a/src/selectors/linearizeTree.ts#L120) works pretty well to narrow down which annotations are recalculated, apart from some unpleasant prop drilling. However, there are still thoughts down below that shouldn't recalculate, as you can see if you start from the repro in https://github.com/cybersemics/em/pull/3357#issuecomment-3680330877 and toggle the context view on `c` instead of `b`.

https://github.com/user-attachments/assets/ecfbfe37-ce30-405d-a386-cafdf13fe4b8

raineorshine (commented):
After playing with this a bit, I unfortunately have come to the conclusion that rendering the superscripts after a delay is too disruptive to the UX. This is true whether the superscripts snap in or fade in smoothly. 

Theoretically it's possible to measure the bounding box synchronously in `useLayoutEffect` and render the superscript before the first paint, though it will have a performance impact. I'd love to get your input on how viable this might be before we spend additional time.

I did another review below, but they are moot points if we're unable to efficiently render the superscript synchronously.

Thanks so much for your work.

---

I'm still seeing the issue described at the end of this comment. Are you able to reproduce? https://github.com/cybersemics/em/pull/3357#pullrequestreview-3619256831

> **Note 1:** This was my interpretation of the test based on the instructions given and the fact that the `positionAnnotation` callback executes after the Context View animation completes. Let me know if I got it wrong.

I don't think this changes the results necessarily, but I meant a simple cursor navigation in normal view that expands a context. Activating the Context View has its own performance issues, so I would avoid that and test superscript rendering without the context view animation. Sorry for the confusion!

> * Measure the **time from tapping the Context View icon** to **the point where layout fully settles**, using frame-by-frame timings and snapshots taken by the Chrome Performance Profiler.

I'm not sure that measuring the time in between the start and finish really measures performance. That's just measuring animation duration, which should generally be constant. If performance drops, the browser will drop frames without changing the animation duration. Only if the CPU is thrashing will the animation actually slow down. I think we need to be measuring either the CPU usage in ms or the average frame rate, ideally without the confounding variable of the Context View animation.

raineorshine:
Oops, I meant selection range, not bounding box! The question is whether it's possible or not to use the same technique that we are using here, but before the first paint.

PR #3367, opened 2025-11-11
Replace ResizeObserver with event listener

### LINKED ISSUE #3356: ResizeObserver error when some thoughts wrap
Tested on desktop Chrome

## Steps to Reproduce

```
- a
- This is a long thought that wraps to a new line at small widths
```

Shrink the browser viewport until the 2nd thought wraps.

## Current Behavior

`"ResizeObserver loop completed with undelivered notifications."`

Image

## Expected Behavior

No error.

ethan-james:
Fixes #3356 

Here is a straightforward replacement of the `ResizeObserver` with a `resize` event listener. It seems to work fine, but it doesn't necessarily "solve" anything except that it no longer shows an error across the top of the viewport.

I'm definitely interested in pursuing a CSS-based solution for the entire `useDropHoverWidth` use case, so I'll look a little deeper into that.

### DIFF
--- a/src/components/Content.tsx
+++ b/src/components/Content.tsx
@@ -51,7 +51,6 @@ const useContentWidth = () => {
       if (!(contentRef.current?.firstChild instanceof HTMLElement)) return
       const rect = contentRef.current.firstChild.getBoundingClientRect()
       if (!rect) return
-      console.log(rect.width)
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Thanks! I will merge this and respond in a new issue to consider the possibility of a CSS only drop hover width calculation.

PR #3370, opened 2025-11-12
TreeNode: set max-width

### LINKED ISSUE #3353: Thought width exceeds viewport width below around 500px
Tested on desktop Chrome

## Steps to Reproduce

```
- This is a long thought that wraps to a new line at small widths
```

Reduce the browser viewport width to 500px.

## Current Behavior

The thought is cut off before it wraps.

Image

## Expected Behavior

The thought should wrap at the edge of the browser and all thought text should be visible.

ethan-james:
Fixes #3353 

I added a `max-width` to `TreeNode` in order to stop it from overrunning the viewport width when the `transform` applied to indented thoughts exceeds the right padding. In order to do this, I needed to updated [bulletWidth in LayoutTree](https://github.com/ethan-james/em/blob/749e16c39f5888e21028d202b83429d03bc146e9/src/components/LayoutTree.tsx#L144) in order to calculate the width of the bullet in the same way that it is calculated in [Bullet](https://github.com/ethan-james/em/blob/749e16c39f5888e21028d202b83429d03bc146e9/src/components/Bullet.tsx#L467). This might have unforeseen consequences, and I could duplicate that calculation inside `TreeNode` instead.

### DIFF
--- a/src/components/LayoutTree.tsx
+++ b/src/components/LayoutTree.tsx
@@ -260,6 +260,20 @@ const LayoutTree = () => {
 
   useLayoutTreeTop(ref, autocrop)
 
+  const treeThoughtsMemoized = useMemo(
+    () =>
+      treeThoughtsPositioned.map(thought => ({
+        ...thought,
+        style: {
+          ...thought.style,
+          // Ensure that transforming the thought's position by its indent level cannot push it off-screen.
+          // The extra 17px is to make sure it doesn't get cut off under the scrollbar
+          maxWidth: `calc(${window.innerWidth > 560 ? '90' : '100'}vw - ${CONTENT_BOX_PADDING_LEFT + thought.x}px - ${1.5 - indent}em)`,
+        },
+      })),
+    [indent, treeThoughtsPositioned],
+  )
+
   return (
     <div
       className={cx(
@@ -304,15 +318,9 @@ const LayoutTree = () => {
           />
         )}
         <TransitionGroup>
-          {treeThoughtsPositioned.map((thought, index) => (
+          {treeThoughtsMemoized.map((thought, index) => (
             <TreeNode
               {...thought}
-              style={{
-                ...thought.style,
-                // Ensure that transforming the thought's position by its indent level cannot push it off-screen.
-                // The extra 17px is to make sure it doesn't get cut off under the scrollbar
-                maxWidth: `calc(${window.innerWidth > 560 ? '90' : '100'}vw - ${CONTENT_BOX_PADDING_LEFT + thought.x}px - ${1.5 - indent}em)`,
-              }}
               index={index}
               // Pass unique key for the component
               key={thought.key}


ethan-james:
I also see that, at larger font sizes at widths exceeding the 560px breakpoint that I was targeting, we can also get some cutoff. I can try to fix that as part of this issue if you want.

Screenshot 2025-11-11 190803

ethan-james:
> Okay, sounds good.
> 
> At the end we'll need to reduce this down to the fewest tests that are needed cover the behavior, since snapshot tests are expensive and we don't want to slow down the test suite unnecessarily.

I figured out that the margin on `editableRecipe` was different from the margin being applied to `ThoughtAnnotation`, so I corrected that. I removed the tests for font size 23, because they weren't demonstrating anything new compared to font size 28. I also removed tests for all widths except for 560px and 575px to test the behavior on either side of the [560px breakpoint](https://github.com/ethan-james/em/blob/e533c7f8f3b563376aaa27ffa945e7581c0d837d/src/components/Content.tsx#L118).

src/components/LayoutTree.tsx
```diff
@@ -260,6 +260,9 @@ const LayoutTree = () => {
 
   useLayoutTreeTop(ref, autocrop)
 
+  // Add a negative marginRight equal to translateX to ensure the thought takes up the full width.
+  const marginRight = `${-indent + (isTouch ? 2 : -1)}em`
```
raineorshine: We actually prefer inlining this since it is only used in one place and it is easier to understand when co-located with the element that uses it. If the code becomes too verbose, it's better to split into smaller components than split the CSS away from the DOM.

src/components/LayoutTree.tsx
```diff
@@ -306,6 +309,12 @@ const LayoutTree = () => {
           {treeThoughtsPositioned.map((thought, index) => (
             <TreeNode
               {...thought}
+              style={{
```
raineorshine: An object literal will be a new reference every time, even if `maxWidth` hasn't changed. While `TreeNode` is not currently memoized, I think it's generally a good practice to preserve referential integrity of props. Therefore, I recommend using `useMemo` to create a stable style object that can be passed to `TreeNode`.

ethan-james: This is taking place inside of `treeThoughtsPositioned.map`, so my options for memoizing it are to:

1. create a new component
2. memoize the entire `treeThoughtsPositioned` object

I think option 2 makes sense to me, but I figured I should check in case that introduces too much risk.

raineorshine: Oof, that's a lot of restructuring to get this memoized. Yes, I think option 2 makes the most sense.

An alternative is to pass `indent` as a prop and allow `TreeNode` to calculate and memoize `maxWidth`. That would be a bit of a lighter touch.

ethan-james: Memoizing `treeThoughtsPositioned` in a straightforward way seems to work pretty well. I'll give it some more thought and see if I can come up with any possible issues with this approach. It occurs to me that `window.innerWidth` isn't reactive, so the `maxWidth` calculation won't update when it crosses the 560px breakpoint. I don't see any issues with how it's working now, though. Maybe the right thing to do is pick either 90vw or 100vw and use that for all viewport widths.

PR #3373, opened 2025-11-15
Drop hover width based off of 100%, subtract depth

### LINKED ISSUE #3369: CSS-based drop hover width
This issue was created to see if a CSS-based solution is possible to calculate the drop hover width. `useDropHoverWidth` is not the most elegant, as it requires a back channel to set the content width which gets used in the drop hover calculations.

_Responding to @ethan-james in https://github.com/cybersemics/em/issues/3367#issuecomment-3518944345_

> Yeah, unfortunately, the content width doesn't have much of a connection to the width of thoughts, which are absolutely-positioned and some variant of 100% width.

I think that's okay, because the drop hover width is relative to the screen width, breakpoints, and depth of the hovering path, not the editable width.

> Especially when we consider the [960px breakpoint](https://github.com/ethan-james/em/blob/461ad3ad9dde43af4be5eccefc0f4dff2cf052f6/src/components/Content.tsx#L119) where `margin: 0 auto` starts to leave variable-sized gaps on either side of the `Content` component.

That should be pretty easy to calculate in CSS as either a percentage or in `vw`. The drop hover width can also have different breakpoints if different calculations are needed to correspond to different content widths at different screen sizes.

> If the goal is to keep the drop targets the same width as the content pane, then the only CSS-based solution that I can think of involves storing a `--content-width` CSS variable on resize, which is similar in terms of complexity to the existing solution.

The goal is for the drop hover with the deepest path to extend to the right edge of the screen, and then for shallower paths to be shortened by the difference in indentation. For example, if thoughts with depths 2–5 are visible, the drop hover widths might be `100% - padding - depth * indentSize`.

That's off the top of my head. Would need some battle testing to see how this could be implemented in CSS, and if there are other edge cases I'm missing.

ethan-james:
Fixes #3369

I updated [useDropHoverWidth](https://github.com/ethan-james/em/blob/f09d12827ffbd9fe7fb510b01328127a6c3c9fd3/src/hooks/useDropHoverWidth.ts#L70) to start from the basic calculation that should match `contentWidth`, `calc(100% - 2em - ${fontSize}px + 10px)`, which is the inverse of [how TreeNode width is calculated](https://github.com/ethan-james/em/blob/f09d12827ffbd9fe7fb510b01328127a6c3c9fd3/src/components/TreeNodeWrapper.tsx#L142) with the `marginLeft` from [dropEndRecipe](https://github.com/ethan-james/em/blob/f09d12827ffbd9fe7fb510b01328127a6c3c9fd3/src/recipes/dropEnd.ts#L10) added in. Then I started trying to offset the hover targets for thoughts that are not the max visible depth. I needed to calculate [maxDepth](https://github.com/ethan-james/em/blob/f09d12827ffbd9fe7fb510b01328127a6c3c9fd3/src/hooks/usePositionedThoughts.ts#L37). You pointed me in the right direction with `100% - padding - depth * indentSize`, but it's actually the inverse of depth because the deepest depth (`maxDepth`) gets 0px subtracted from it, and then every shallower child gets `1em * (maxDepth - depth)` subtracted.

That went pretty well, but then I started to run into edge cases with [DropEnd](https://github.com/ethan-james/em/blob/f09d12827ffbd9fe7fb510b01328127a6c3c9fd3/src/components/DropEnd.tsx#L115) and [DropHover](https://github.com/ethan-james/em/blob/f09d12827ffbd9fe7fb510b01328127a6c3c9fd3/src/components/DropHover.tsx#L56) and all the rest. It will still take some digging to match all of the snapshots, although hopefully I'll be able to simplify my implementation in the process. Let me know if I should be trying to accomplish something other than matching snapshots, as I'm not sure they always follow the rule that you outlined, "The goal is for the drop hover with the deepest path to extend to the right edge of the screen, and then for shallower paths to be shortened by the difference in indentation." I figure that once the browser passes above 1200px, the thoughts don't ever actually stretch all the way to the right, and that may be OK. Historically, all drop targets are `contentWidth` pixels wide, and haven't been adjusted to match their various margins.

### DIFF
--- a/src/components/DropEnd.tsx
+++ b/src/components/DropEnd.tsx
@@ -106,7 +106,7 @@ const DropEnd = ({
         dropEndRecipe(),
         css({
           display: 'list-item',
-          // PandCSS doesn't like interpolated values like '${marginLeft}em'
+          // PandaCSS doesn't like interpolated values like '${marginLeft}em'
           marginLeft: isRootPath ? '-4em' : last ? '-2em' : 0,
           // offset marginLeft, minus 1em for bullet
           // otherwise drop-hover will be too far left


ethan-james:
I'm made some progress on this, but there are still 8-10 failing tests. I reviewed them to see if there's a common thread, but it's not looking great. There are a variety of issues including simple things like a couple drop hovers that are still too long or short, but there are also drop hovers that are  showing up with the wrong indentation, and drop targets that seem to be missing entirely or else are covered by another drop target and are not selectable. Given the limited scope of these changes, I'm not even sure how a drop target would go missing, nor am I quite sure how much longer it would take to track down the sources of all of these issues, but my guess is quite a while.

raineorshine:
> there are also drop hovers that are showing up with the wrong indentation, and drop targets that seem to be missing entirely or else are covered by another drop target and are not selectable

I think it would be a good idea to assess these complications and determine if the overall approach is still viable and does not introduce additional complexity. If there are too many edge cases to address, we may inadvertently make the system fragile in different ways.

In theory, it doesn't seem like a depth subtraction-based width is that complex. So, there must be some hidden dependencies that we haven't yet clearly articulated. What else does the drop hover width depend on, or vice versa? The drop-hover element itself is position absolute, so it shouldn't be affecting the layout. Is it tied to the drop target in a different way that prevents it from being adjusted independently?

Perhaps take a step back and consider some of these more general questions about the drop hover and how it interacts with the layout engine and drag-and-drop functionality. Hopefully that helps give you a clearer vision of what limitations we are working with and what the best way is to move forward.

Thanks!

Issue #3354, opened 2025-11-07
Split "-" into child


raineorshine:
Extend the `splitSentence` command to handle this extra use case. If we find they need to be orthogonal we can separate it out into its own command.

# Case 1: Split into child

## Steps to Reproduce

```
- one - 1
```

Activate Split Sentences on the thought.

## Current Behavior

Cannot split sentences: thought is an empty thought or has only one sentence.

## Expected Behavior

Split on `-`, `–`, or `—` (dash, en-dash, or em-dash) and insert the right half as a child of the thought.

```
- one
  - 1
```

# Case 2: Split sentence

If a thought has both dash and multiple sentences, the current Split Sentence behavior should take precedence.

## Steps to Reproduce

```
- one - 1. two. three.
```

Activate Split Sentences on the thought.

## Current Behavior

```
- one - 1.
- two.
- three.
```

## Expected Behavior

Same.

PR #3388, opened 2025-11-19
Update `split sentence` functionality

codebuild-ro:
Fixes: #3354 (Split "-" into child)
- Update the `split sentence` function
- Added unit test cases for action and command

### DIFF
--- a/src/commands/__tests__/splitSentences.ts
+++ b/src/commands/__tests__/splitSentences.ts
@@ -108,7 +108,7 @@ describe('splitSentences', () => {
   - me`)
   })
 
-  it('splits thought with dash into main thought and child (Case 1)', () => {
+  it('splits thought with dash into main thought and child', () => {
     store.dispatch([
       importText({
         text: `
@@ -126,7 +126,7 @@ describe('splitSentences', () => {
     - 1`)
   })
 
-  it('splits by sentences when both dash and multiple sentences are present (Case 2)', () => {
+  it('splits by sentences when both dash and multiple sentences are present', () => {
     store.dispatch([
       importText({
         text: `


src/actions/__tests__/splitSentences.ts
```diff
@@ -675,3 +675,95 @@ describe('parenthetical content', () => {
   - This (has parentheses) in the middle`)
   })
 })
+
+describe('dash splitting', () => {
+  it('splits thought with dash into main thought and subthought (Case 1)', () => {
```
raineorshine: "Case 1" and "Case 2" are not meaningful here and should be removed. The test cases are self-explanatory.

src/actions/__tests__/splitSentences.ts
```diff
@@ -675,3 +675,95 @@ describe('parenthetical content', () => {
   - This (has parentheses) in the middle`)
   })
 })
+
+describe('dash splitting', () => {
```
raineorshine, line 679: Tests look good, thanks!

src/actions/__tests__/splitSentences.ts
codebuild-ro: I removed those cases.

raineorshine: Sorry for the confusion. It's not the test cases that were the problem, but the arbitrary numbering. Another developer would not know what "Case 1" refers to, and it's not necessary to understand the test case. It was only meaningful in the context of the GitHub issue.

codebuild-ro: Ah, I got it, I recovered the testing cases and just removed those words in description of tests.

raineorshine (approved):

Issue #3245, opened 2025-09-22
Url link hidden by autofocus is still clickable


raineorshine:
## Steps to Reproduce

```
- a
  - stackoverflow.com
  - b
    - c
      - d
```

1. Set the cursor on `c`. 
2. Tap the hidden url icon. 

## Current Behavior

The url opens. 

## Expected Behavior

Nothing should happen. 

Hidden url icons should not be clickable.

PR #3407, opened 2025-11-26
Fix URL link hidden by autofocus is still clickable

codebuild-ro:
Fix: https://github.com/cybersemics/em/issues/3245

Remove `pointerEvent: "none"` from `ThoughtAnnotationWrapper`.
This way is fine, because of the rendering order.
So first `ThoughtAnnotationWrapper` is rendered and after that, `ThoughtAnnotation` is rendered. So `ThoughtAnnotation` overlaps the `ThoughtAnnotationWrapper`.
Before, I had mistake to understand the rendering order.

So  there is no need to add property `pointerEvent: "all"` to icon components.


raineorshine (changes requested):

This solves the problem, but I'm not sure yet if this is the best solution. 

First, there is a bit of forensics that is needed to understand the current use of `pointerEvents: all`. What purpose does it serve? When was it added and why? Is it overriding an inherited value? The default value is `auto` so it's not obvious why it would need to be set to `all`. We always make sure to understand the cause of the problem first before attempting a fix. Git blame will help here.

Second, the assignment of responsibility seems wrong here. Why should `UrlIconLink` have to know whether it is hidden or not? If an ancestor controls its visibility, it seems like the ancestor should be responsible for this.

src/components/icons/UrlIcon.tsx
```diff
@@ -13,7 +13,6 @@ const UrlIcon = () => {
         // Make sure the icon doesn't take up extra space.
         padding: '0.05em 0.167em 0.167em 0',
         cursor: 'pointer',
-        pointerEvents: 'all',
```
codebuild-ro: I removed this to reduce the depth of prop drilling of `isVisible` property.
Instead, I used that `pointerEvents: 'all'` in [`UrlIconLink`](https://github.com/cybersemics/em/pull/3407/files#diff-c9cf3925bce23c2f617f716ccd3eeeb33cfd62477a81721d836ec197a4a3fec0R66).

※※※※※※※※※※※※※※※※※※※※※※※※※※※※※※※
https://github.com/cybersemics/em/blob/ab158c293dcbda1557e80e2000d3bdef740d21db/src/components/ThoughtAnnotationWrapper.tsx#L45
Please notice that `Subthought` has `pointerEvents: 'none'`

So the children inside `ThoughtAnnotationWrapper` cannot be clickable. To solve that, the previous developer added this property - `pointerEvents: 'all'` (to override inherited value).
https://github.com/cybersemics/em/blob/ab158c293dcbda1557e80e2000d3bdef740d21db/src/components/ThoughtAnnotation.tsx#L162-L165

src/components/ThoughtAnnotation.tsx
```diff
@@ -56,14 +56,14 @@ const addMissingProtocol = (url: string) =>
...
-const UrlIconLink = React.memo(({ url }: { url: string }) => {
+const UrlIconLink = React.memo(({ isVisible, url }: { isVisible?: boolean; url: string }) => {
   const dispatch = useDispatch()
   return (
     <a
       href={addMissingProtocol(url)}
       rel='noopener noreferrer'
       target='_blank'
-      className={urlLinkStyle}
+      className={cx(urlLinkStyle, css({ pointerEvents: !isVisible ? 'none' : 'all' }))}
```
codebuild-ro: `UrlIconLink` has [`UrlIcon`](https://github.com/cybersemics/em/pull/3407/files#diff-77809095470ad3b5dc6fe23e8239ca2e15e67d06f6782486a9fc626646ea11cd) as a child.
In [`UrlIcon`](https://github.com/cybersemics/em/pull/3407/files#diff-77809095470ad3b5dc6fe23e8239ca2e15e67d06f6782486a9fc626646ea11cd) component, there is a style `pointerEvents: 'all'`.
I extracted out `pointerEvents` property from that component to reduce the depth of prop drilling of `isVisible`.

This property is calculated here. And it is drilled down to children deeply (in original approach).
https://github.com/cybersemics/em/blob/ab158c293dcbda1557e80e2000d3bdef740d21db/src/components/Subthought.tsx#L67

raineorshine (commented):
Hi, thanks for your submission. Is there any chance there is a simpler way to do this? The PandaCSS recipe does provide some type safety, but now we have a double override. Plus, `& > *` is a child selector, which creates a type unsafe connection to the DOM hierarchy. We typically try to avoid selectors like this.

I would like to think that a more elegant solution exists. Let me know if you have any other ideas!

It might also be worth investigating the original reason why `pointerEvents: none` was set on the ThoughtAnnotation. Maybe there is a different way to achieve that, and avoid the need for awkward overrides.

raineorshine:
@codebuild-ro

codebuild-ro:
>When I remove `pointerEvents: 'none'`, everything still seems to work. Do you know what I'm missing?

Did try to remove `pointerEvents: 'none'` from `ThoughtAnnotationWrapper`?

codebuild-ro:
> Issue is still reproducible. Hidden url is still clickable.
> 
> ## Step to reproduce
> 1. Create some thought
> 2. Add a url as a subthought in one of them
> 3. Click a thought that has no url
> 4. Click url icon on the thought that has url as subthought from step 2
> 
> ## Current behavior
> url is clickable and new browser is opened See below video:
> 
>  Hidden.URL.icon.is.stll.clickable.MOV 
> ## Expected behavior
> Nothing should happen because the url is hidden under

Hi @BayuAri ,
Were you able to check the issue description please?
Your testing scenario is not directly related to the issue.

codebuild-ro:
> Issue is still reproducible. Hidden url is still clickable.
> 
> ## Step to reproduce
> 1. Create some thought
> 2. Add a url as a subthought in one of them
> 3. Click a thought that has no url
> 4. Click url icon on the thought that has url as subthought from step 2
> 
> ## Current behavior
> url is clickable and new browser is opened See below video:
> 
>  Hidden.URL.icon.is.stll.clickable.MOV 
> ## Expected behavior
> Nothing should happen because the url is hidden under

Hi @BayuAri ,
Were you able to check the issue description please?
Your testing scenario is not directly related to the issue.

What do you think about this? @raineorshine

PR #3422, opened 2025-12-01
Set drop hover width to 50vw

### LINKED ISSUE #3369: CSS-based drop hover width

ethan-james:
Fixes #3369

In order to simplify drop hover logic, `useDropHoverWidth` now unconditionally returns `50vw` so that we can see if that looks good.

### DIFF
--- a/src/recipes/dropHover.ts
+++ b/src/recipes/dropHover.ts
@@ -7,7 +7,7 @@ const dropHoverRecipe = defineRecipe({
     position: 'absolute',
     height: '3px',
     borderRadius: '99px',
-    width: `token(sizes.dropHover)`,
+    width: 'dropHover',
     zIndex: 'stack',
   },
   variants: {
--- a/src/components/DropEnd.tsx
+++ b/src/components/DropEnd.tsx
@@ -105,7 +105,7 @@ const DropEnd = ({
         css({
           display: 'list-item',
           // If dropping target is table column 1, do not set width (but use width property of dropEndRecipe)
-          width: isParentTableCol1 ? undefined : 'token(sizes.dropHover)',
+          width: isParentTableCol1 ? undefined : 'dropHover',
           marginLeft: isRootPath ? '-4em' : last ? '-2em' : undefined,
           // offset marginLeft, minus 1em for bullet
           // otherwise drop-hover will be too far left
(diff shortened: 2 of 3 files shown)


raineorshine (changes requested):
Just noticing this, but since the value is now constant at compile-time, we should be able to move it into a Panda `css` or `cx` block. It would be good to prefer the Panda approach when possible and reserve `style` for dynamic runtime values. Is that a change you could make before we merge?

src/recipes/dropHover.ts
```diff
@@ -7,6 +7,7 @@ const dropHoverRecipe = defineRecipe({
     position: 'absolute',
     height: '3px',
     borderRadius: '99px',
+    width: `token(sizes.dropHover)`,
```
raineorshine: Simpler:

```
  width: 'dropHover'
```

ethan-james: It works! Thanks for that.

Issue #3408, opened 2025-11-26
[Liminal UI] Refactor the Gesture Menu into its own component


fbmcipher:
> **Before implementing a new design for the Gesture Menu, separate it from the Command Palette.**

Currently, the **Gesture Menu** is drawn by the `CommandPalette` component.

The `CommandPalette` component is responsible for *both* the **Gesture Menu on mobile** *and* the **Command Palette on desktop**.

<img width="1625" height="895" alt="Image" src="https://github.com/user-attachments/assets/6a831ab3-715d-41c1-b323-744a70e88ee4" />

---

These features currently share a lot of core functionality and a similar interface. However, the visual and functional aspects of the **Command Palette** and the **Gesture Menu** have changed a lot during the Liminal UI design process, and they are now quite different from one another.

For that reason, I think we should **separate the Gesture Menu into its own component** before we begin redesigning it.

This will enable us to **modify the Gesture Menu as much as we need to** without the risk of **breaking the Command Palette on desktop**.

### Tasks

- [ ] Identify which parts of `CommandPalette` are responsible for the **Gesture Menu**, and which parts of it are responsible for the **Command Palette**
- [ ] Identify any shared functionality that is used between both the **Gesture Menu** and the **Command Palette**, and abstract them into hooks or reusable functions
- [ ] Separate the **Gesture Menu** into its own component
- [ ] Test to ensure no breaking changes with either the **Gesture Menu** or the **Command Palette**

Some level of duplication is okay, though please use your judgment to work out where duplication is and isn't acceptable.

### Expected Outcome

- The **Command Palette** and **Gesture Menu** look and function identically to before.
- `CommandPalette` is responsible for *only* the Command Palette.
- `GestureMenu` is responsible for *only* the Gesture Menu.
- Any substantially complex functionality that is shared between the two has been abstracted into separate import.
- Tests should continue to pass.

Thank you very much!

Issue #3018, opened 2025-06-09
Migrate `em` to `rem` css units


raineorshine:
1 `rem` equals the font size of the root element (`<html>`).
1 `em` equals the element's inherited font size (thus is sensitive to all ancestor's `fontSize`).

Currently, `em` is used to scale elements across the app according to `state.fontSize`, which is a user font size setting. This includes the size of buttons, panels, toolbar, icons, etc, not just font size. User controlled font size is critical for both user preference and low vision users, and mobile users do not have convenience of <kbd>Command/Ctrl</kbd> + <kbd>+</kbd>. 

The use of `em` is made possible by [setting fontSize on the Content container](https://github.com/cybersemics/em/blob/7a1af9934377f3c5bd5bd68f73dc5d7947a80b4e/src/components/AppComponent.tsx#L197). However, `em` units are affected by the font size of all ancestors, which is usually not the intention. This really should have used `rem` all along. Regrettably, there are places where the desired font size *depends* on the inheritance feature, so this is not a trivial change.

Refactor `em` to `rem` and fix any visual regressions.

cindmichelle:
Hi @raineorshine , I noticed that some parts of the codebase already use `rem`. After setting `1rem` to follow `state.fontSize`, a few components change visually, since we previously assumed `1rem = 16px` in the [panda.config.ts](https://github.com/cybersemics/em/blob/68271624e6454795b04ca9e3c310414dd26904a2/panda.config.ts#L249).

<img width="305" height="444" alt="Image" src="https://github.com/user-attachments/assets/32943489-1eae-4346-b089-5b534ece527e" />

Should we normalize the existing `rem` values to keep the same px sizes, or just leave them as-is?

PR #3436, opened 2025-12-10
Refactoring CommandPalette and GestureMenu

cindmichelle:
Close https://github.com/cybersemics/em/issues/3408

## Description
- Separating CommandPalette and GestureMenu into two independent components.
- Split the state that controls their visibility into two separate states.
    - `showCommandPalette` now controls the `<CommandPalette/>`.
    - `showGestureMenu` now controls the `<GestureMenu/>`.
- Separate the action creators used for GestureMenu and CommandPalette.

## Results
**CommandPalette**

https://github.com/user-attachments/assets/476fb3d9-c5e7-4c97-89f3-7fb61821ad7d

----
**GestureMenu**

https://github.com/user-attachments/assets/512dcce1-9d06-4760-a6d5-ddcc5a89ab1d

### DIFF
--- a/src/components/CommandPalette.tsx
+++ b/src/components/CommandPalette.tsx
@@ -268,21 +268,19 @@ const CommandPalette: FC<{
           })}
         >
           {commands.length > 0 ? (
-            <>
-              {commands.map((command, index) => (
-                <CommandItem
-                  search={search}
-                  key={command.id}
-                  onClick={onExecute}
-                  onHover={onHover}
-                  selected={command === selectedCommand.command}
-                  command={command}
-                  shouldScrollSelectedIntoView={selectedCommand.source === 'keyboard'}
-                  isFirstCommand={index === 0}
-                  isLastCommand={index === commands.length - 1}
-                />
-              ))}
-            </>
+            commands.map((command, index) => (
+              <CommandItem
+                search={search}
+                key={command.id}
+                onClick={onExecute}
+                onHover={onHover}
+                selected={command === selectedCommand.command}
+                command={command}
+                shouldScrollSelectedIntoView={selectedCommand.source === 'keyboard'}
+                isFirstCommand={index === 0}
+                isLastCommand={index === commands.length - 1}
+              />
+            ))
           ) : (
             <span className={css({ marginLeft: '1em' })}>No matching commands</span>
           )}
--- a/src/components/GestureMenu.tsx
+++ b/src/components/GestureMenu.tsx
@@ -26,8 +26,6 @@ const GestureMenu: FC<{
 
   const hasMatchingCommand = commands.some(cmd => (gestureInProgress as string) === gestureString(cmd))
 
-  /* Select the last command in the list. */
-
   return (
     <div
       className={css({
@@ -52,33 +50,31 @@ const GestureMenu: FC<{
         })}
         style={{ fontSize }}
       >
-        {gestureInProgress && commands.length > 0 ? (
-          <div
-            className={css({
-              padding: '0.85em 0.66em',
-            })}
-          >
-            {commands.map((command, index) => {
-              // Check if the current gesture sequence ends with help gesture
-              const cheatsheetInProgress = gestureInProgress
-                ?.toString()
-                .endsWith(gestureString(openGestureCheatsheetCommand))
-              const isCheatsheetMatch = command.id === 'openGestureCheatsheet' && cheatsheetInProgress
-              const isCancelMatch = command.id === 'cancel' && !hasMatchingCommand && !cheatsheetInProgress
+        <div
+          className={css({
+            padding: '0.85em 0.66em',
+          })}
+        >
+          {commands.map((command, index) => {
+            // Check if the current gesture sequence ends with help gesture
+            const cheatsheetInProgress = gestureInProgress
+              ?.toString()
+              .endsWith(gestureString(openGestureCheatsheetCommand))
+            const isCheatsheetMatch = command.id === 'openGestureCheatsheet' && cheatsheetInProgress
+            const isCancelMatch = command.id === 'cancel' && !hasMatchingCommand && !cheatsheetInProgress
 
-              return (
-                <CommandItem
-                  gestureInProgress={gestureInProgress as string}
-                  key={command.id}
-                  selected={isCheatsheetMatch || gestureInProgress === gestureString(command) || isCancelMatch}
-                  command={command}
-                  isFirstCommand={index === 0}
-                  isLastCommand={index === commands.length - 1}
-                />
-              )
-            })}
-          </div>
-        ) : null}
+            return (
+              <CommandItem
+                gestureInProgress={gestureInProgress as string}
+                key={command.id}
+                selected={isCheatsheetMatch || gestureInProgress === gestureString(command) || isCancelMatch}
+                command={command}
+                isFirstCommand={index === 0}
+                isLastCommand={index === commands.length - 1}
+              />
+            )
+          })}
+        </div>
       </div>
     </div>
   )


src/components/CommandPalette.tsx
```diff
@@ -247,76 +243,56 @@ const CommandPalette: FC<{
...
-          </>
-        ) : null}
+                  })
+                })()}
+              </>
+            ) : (
+              <span className={css({ marginLeft: '1em' })}>No matching commands</span>
+            )}
+          </div>
+        </>
```
fbmcipher: Nitpick: We no longer need to wrap this code with the fragments on lines 248 & 295 – these can be removed.

src/commands.ts
```diff
@@ -206,14 +206,14 @@ export const inputHandlers = (store: Store<State, any>) => ({
 
     // command palette
```
fbmcipher: The comment `// command palette` should probably change to `// gesture menu` instead.

I see you caught and changed a lot of these cases already – thanks! – just looks like a couple slipped through the cracks.

src/commands.ts
```diff
@@ -162,7 +162,7 @@ const index = (): {
   }
 }
 
-let commandPaletteGesture: number | undefined
+let gestureMenuPalette: number | undefined
```
fbmcipher: Nitpick: The word "palette" refers specifically to the Command Palette. Since this variable name corresponds to a timeout, maybe `gestureMenuTimeout` would be a more descriptive and appropriate name.

fbmcipher: We should also probably update the comment on lines 176-192 (shown below) with the new variable name and explain the responsibility of `gestureMenuTimeout`. It still refers to `CommandPaletteGesture`, the name of the variable *before* the refactor.

https://github.com/cybersemics/em/blob/67981a4524849feb93c8a6ed132ee9a32a1be4e3/src/commands.ts#L176-L191

src/components/GestureMenu.tsx
```diff
@@ -0,0 +1,144 @@
...
+              )}
+            </div>
+          </>
+        ) : null}
+      </div>
+    </div>
+  )
+}
+
+/** A CommandPalette component that fades in and out based on state.showGestureMenu. */
```
fbmcipher: Another comment that needs text replacing from `CommandPalette` to `GestureMenu`. Probably worth cmd+f to look across each of these files and replace with new naming.

src/components/GestureMenu.tsx
```diff
@@ -0,0 +1,144 @@
...
+import CommandItem from './CommandItem'
+import FadeTransition from './FadeTransition'
+import PopupBase from './PopupBase'
+
+/**********************************************************************
+ * Components
+ **********************************************************************/
+
+/** Render a command palette with gesture autocomplete. */
+const GestureMenu: FC = () => {
```
fbmcipher: OK, let's talk about this block of code.

There are two similarly-named components in both `CommandPalette.tsx` and `GestureMenu.tsx`:

- `CommandPalette` / `GestureMenu`: The command palette (or gesture menu) itself.
- `CommandPaletteWithTransition` / `GestureMenuWithTransition`: A higher-order component which animates and conditionally renders `CommandPalette` or `GestureMenu` based on Redux state. This is the component that's actually exported from each file, and used in `AppComponent`.

Before your refactor, this block of code was inside `CommandPaletteWithTransition` – and the state/hooks here were used in `CommandPaletteWithTransition` and passed to `CommandPalette` as a prop. That was done intentionally in https://github.com/cybersemics/em/commit/c47d56afd18d3ef95ad20a45675336c81cf6a7fd, based on the comment:

```
// Commands need to be calculated even if the command palette is not shown because useFilteredCommands is responsible for updating gestureStore's possibleCommands which is needed to prevent haptics when there are no more possible commands. Otherwise, either haptics would continue to fire when there are no more possible commands, or would falsely fire when the current sequence is not a valid gesture but there are possible commands with additional swipes.
```

In other words, we need to handle this state *inside* `GestureMenuWithTransition`, **not** `GestureMenu` – as otherwise, when `GestureMenu` is unmounted it will no longer update `possibleCommands`.

What are your thoughts? Did you make the change intentionally to fix something, or is this just a oversight?

src/components/CommandPalette.tsx
```diff
@@ -113,17 +109,18 @@ const CommandSearch: FC<{
...
+/** Render a command palette with keyboard. */
+const CommandPalette: FC = ({}) => {
+  // Commands need to be calculated even if the command palette is not shown because useFilteredCommands is responsible for updating gestureStore's possibleCommands which is needed to prevent haptics when there are no more possible commands. Otherwise, either haptics would continue to fire when there are no more possible commands, or would falsely fire when the current sequence is not a valid gesture but there are possible commands with additional swipes.
+  const [recentCommands, setRecentCommands] = useState(storageModel.get('recentCommands'))
+  const [search, setSearch] = useState('')
+  const commands = useFilteredCommands(search, {
+    recentCommands,
+    sortActiveCommandsFirst: true,
+  })
+
```
fbmcipher: Refer to my comment on `GestureMenu.tsx`. I think the same points apply here – unless there was a reason to move this code and state/hooks out of `CommandPaletteWithTransition`, it should probably be moved back there.

src/components/AppComponent.tsx
```diff
@@ -165,7 +166,8 @@ const AppComponent: FC = () => {
     >
       <Alert />
       <Tips />
-      <CommandPalette />
+      {!isTouch && <CommandPalette />}
```
fbmcipher, line 169: This is brilliant – having a separate component for these two features makes sense and sets us up for development ahead.

src/components/GestureMenu.tsx
```diff
@@ -0,0 +1,144 @@
...
+                        />
+                      )
+                    })
+                  })()}
+                </>
+              ) : (
+                <span className={css({ marginLeft: '1em' })}>No matching commands</span>
+              )}
+            </div>
+          </>
```
fbmcipher: Nitpick: We no longer need to wrap this code with the fragments on lines 57 & 95 – these can be removed.

src/components/GestureMenu.tsx
cindmichelle: This was an oversight on my side.
Initially I thought it was safe to move the logic into `GestureMenu` and `CommandPalette` since there’s no dependency from the component wrapper to GestureMenu.

I missed the comment though, and I only realised now that the difference becomes more obvious in the `CommandPalette` case: when there should be no commands rendered anymore because the search command invalid, the current logic still keeps them visible.

src/components/CommandPalette.tsx
```diff
@@ -247,76 +242,54 @@ const CommandPalette: FC<{
          * See `onClick` on parent.
          * */
         onClick={e => e.stopPropagation()}
-        style={{ fontSize }}
+        style={{ fontSize, marginBottom: fontSize }}
```
fbmcipher: During the refactor, it looks like `marginBottom` was moved from the styles passed into `css()` (lines 227-233) into the inline `style` prop.

In general, we want styles to live inside `css()`, since PandaCSS gives us build-time optimisation that we lose when using inline styles.

Was this change intentional? If not, I think we should move `marginBottom` back into `css()`.

The existing `fontSize` in `style`  was already present before the refactor, so that can stay. There are a few intentional exceptions like this in the codebase, but they're not the rule.

cindmichelle: @fbmcipher yes this change was intentional. the reason is because im getting eslint error when using the `css({marginBottom: fontSize})`

image

cindmichelle: I just realized that when I pass `marginBottom` into `css()`, neither of these two values triggers an ESLint error.

```
marginBottom: `${fontSize}`
```

or
```
marginBottom: `${fontSize}px`
```

do you have preference to use which one?

fbmcipher: Thanks @cindmichelle, you're totally right – I didn't spot the linter error.

Interestingly, despite the linter throwing a warning, the code appears to work in both dev and prod builds!

Let's go with 

    marginBottom: `${fontSize}`  

That should work and resolve the linter error.

src/components/CommandPalette.tsx
```diff
@@ -249,74 +245,52 @@ const CommandPalette: FC<{
...
-        ) : null}
+        <div
+          className={css({
+            overflow: 'auto',
+            padding: '0.85em 0.66em',
+          })}
+        >
+          {commands.length > 0 ? (
+            <>
+              {(() => {
```
raineorshine: No need for the IIFE now that here is no additional nested logic.

src/components/CommandPalette.tsx
```diff
@@ -249,74 +245,52 @@ const CommandPalette: FC<{
...
+                return commands.map((command, index) => {
+                  return (
```
raineorshine: Prefer arrow function without braces or `return`.

src/components/CommandPalette.tsx
```diff
@@ -224,22 +220,22 @@ const CommandPalette: FC<{
...
-          borderRadius: isTouch ? undefined : '12px',
+          borderRadius: '12px',
           backgroundColor: 'gray09',
           maxWidth: '100%',
           maxHeight: '100%',
           width: 826,
           cursor: 'default',
           display: 'flex',
           flexDirection: 'column',
+          marginBottom: `${fontSize}`,
```
raineorshine: Can you confirm that this `marginBottom` works as expected here? Unless I'm mistaken, `fontSize` is not a statically analyzable value.

This could also be replaced with a simpler `1rem`.

src/components/GestureMenu.tsx
```diff
@@ -0,0 +1,148 @@
...
+        style={{ fontSize }}
+      >
+        {gestureInProgress && commands.length > 0 ? (
+          <>
+            <div
+              className={css({
+                padding: '0.85em 0.66em',
+              })}
+            >
+              {commands.length > 0 ? (
```
raineorshine: `commands.length > 0` was already checked above.

src/components/CommandPalette.tsx
cindmichelle: You’re right, after rechecking in DevTools, marginBottom isnt applied with this approach. My bad, I missed this after the change.

I tried `marginBottom: 1rem` , but it seems inaccurate since the `1rem` value currently still a static value from `panda.config.ts`, not `state.fontSize`.
image

For now, it feels more reasonable to keep it as 1em and convert it to `1rem` in #3018. Thoughts?

src/components/GestureMenu.tsx
```diff
@@ -24,6 +24,8 @@ const GestureMenu: FC<{
   const gestureInProgress = gestureStore.useSelector(state => state.gesture)
   const fontSize = useSelector(state => state.fontSize)
 
+  const hasMatchingCommand = commands.some(cmd => (gestureInProgress as string) === gestureString(cmd))
```
cindmichelle, line 27: i decided to moved `hasMatchingCommand` out of the IIFE after taking another look. now we can get rid the IIFE from GestureMenu too

src/components/GestureMenu.tsx
```diff
@@ -0,0 +1,136 @@
...
+/** Render a gesture menu with gesture autocomplete. */
+const GestureMenu: FC<{
+  commands: Command[]
+}> = ({ commands }) => {
+  const gestureInProgress = gestureStore.useSelector(state => state.gesture)
+  const fontSize = useSelector(state => state.fontSize)
+
+  const hasMatchingCommand = commands.some(cmd => (gestureInProgress as string) === gestureString(cmd))
+
+  /* Select the last command in the list. */
```
raineorshine: This comment seems out of place?

src/components/CommandPalette.tsx
```diff
@@ -249,74 +245,48 @@ const CommandPalette: FC<{
...
-          </>
-        ) : null}
+        <div
+          className={css({
+            overflow: 'auto',
+            padding: '0.85em 0.66em',
+          })}
+        >
+          {commands.length > 0 ? (
+            <>
```
raineorshine: I think you can remove the fragment and just do `commands.length > 0 ? commands.map(...)`

src/components/GestureMenu.tsx
```diff
@@ -0,0 +1,136 @@
...
+          maxWidth: '100%',
+          maxHeight: '100%',
+          width: 826,
+          cursor: 'default',
+          display: 'flex',
+          flexDirection: 'column',
+        })}
+        style={{ fontSize }}
+      >
+        {gestureInProgress && commands.length > 0 ? (
```
raineorshine: Is it possible to remove `gestureInProgress` from this condition since `GestureMenu` is only shown when there is a gesture in progress anyway? It may have only been needed to differentiate it from the `CommandPalette`.

raineorshine: `commands.length > 0` is always true on mobile since the Cancel and Gesture Cheatsheet commands are always shown.

cindmichelle: just tested again, and yes looks like no regression if we remove this ternary logic

cindmichelle: turns out after checking them again, this is causing some regression where all the commands shows for a second before the gestureMenu close (#3493). So we bring this logic back at #3524

PR #3457, opened 2025-12-17
Grouping dependencies in dependabot

karunkop:
@raineorshine , I have scanned the `package.json` file and have compiled the grouping accordingly. These grouping are purely on the basis of packages being grouped with their related dependencies/ type dependencies. Let me know if I have missed any important packages that needs to be grouped as well.

Also, https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates, the suggested example in the docs show grouping version-updates according to update-types, like minor or patch upgrades grouped to one and major to a single PR. Shall we implement the version upgrades accordingly ?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,73 @@ updates:
     schedule:
       interval: 'weekly'
     open-pull-requests-limit: 10
+    # Group related deps into single PR
+    groups:
+      react:
+        patterns:
+          - "react"
+          - "react-dom"
+          - "@types/react"
+          - "@types/react-dom"
+      capacitor:
+        patterns:
+          - "@capacitor/*"
+      mui-emotion:
+        patterns:
+          - "@mui/*"
+          - "@emotion/*"
+      react-dnd:
+        patterns:
+          - "react-dnd*"
+          - "dnd-core"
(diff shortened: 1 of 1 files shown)


karunkop:
> Thanks!
> 
> > Also, [docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates), the suggested example in the docs show grouping version-updates according to update-types, like minor or patch upgrades grouped to one and major to a single PR. Shall we implement the version upgrades accordingly ?
> 
> We could group patch updates together into one PR, minor updates together into one PR, and major updates separately (unless part of a group). What do you think?

That would be ideal but upon my testing what I found is that there could be sone PRs which could clash based upon semantic versioning groups and specific package groups. Like for eg: `lodash` could end of being updated as a patch/minor version upgrade and could also be included in other PR with specific group update (i.e with it's type dependency). So may be for now I think it would be a good idea to stick with specific group updates as included in `dependabot.yml` to avoid confusions. We can always see and change the configuration accordingly once the PR starts being created.

PR #3468, opened 2025-12-21
Migrate from MUI Swipeable Drawer to react-modal-sheet

### LINKED ISSUE #3283: [Liminal UI] Replace `SwipeableDrawer` with a more flexible drawer implementation
> **Migrate to a new drawer implementation**, which will act as a foundation for further work on the Command Center.

> Ignore background glow behavior, but otherwise ensure no behavioral changes from current implementation.

Image

MUI’s `SwipeableDrawer` has several limitations that make further Command Center development difficult without resorting to hacky workarounds.

Let’s migrate to a more flexible drawer implementation that will grow with our needs.

## Features the new drawer should ideally support

**Here's a list of features the new drawer implementation will need to support.**

We shouldn't implement any new features in this PR, but the chosen component/library should be capable of them so we can proceed confidently knowing the solution can scale to the tasks ahead.

- **Gesture support**
	- The ability to swipe the drawer to close it, as with the current implementation.
- **Performant observability of drawer height**
	- We need to be able to **track the drawer as it opens and closes** – whether the drawer is being open/closed **programatically**, or **gradually via the swipe gesture**.
- **Snap points**
	- When we come around to implementing the "Expanded View", we will need the drawer to **expand further** once open to a second stage. (See image 3.)
	- These are called "snap points" in most drawer implementations.
	- The user should be able to **swipe up** (between image 2 to 3) to move to the next snap point, **or** tap the arrow.
	- So we will need both **gesture** and **programatic** ability to switch between snap points.
	- We will also need to be able to **switch out the contents of the drawer between the snap points**, so we'll need some way to **detect the current snap point**.
- **Behaves gracefully with the keyboard**
	- When the mobile keyboard opens, the drawer should respond predictably.
	- This will be especially relevant when implementing the Command Center’s search view. (See image 4.)
- **Drawer can fully unmount when closed**
	- The current MUI drawer stays mounted in the DOM even when it's closed. For more memory and performance optimisation, it would be great to identify a library where we can choose to unmount the drawer when it's closed.

**Also, some nice-to-haves:**

- **Supports horizontal (sidebar/hamburger menu) orientation**
	- At some point we'll need to implement the new design for the Sidebar – could be cool to find a solution that would work for this, too

## Potential candidates

I only did a cursory review of these, but they seem like good starting points for exploration:

- [Vaul](https://vaul.emilkowal.ski/)
- [react-spring-bottom-sheet](https://react-spring.bottom-sheet.dev/)

yangchristina:
Resolves #3283 

IOS:

https://github.com/user-attachments/assets/16473cad-9b29-41b7-9188-9eccb1e00655

Chrome:

https://github.com/user-attachments/assets/920662a7-a1a0-4bc4-9fcd-3fc0ff564701


raineorshine:
> @fbmcipher what is the reason we have a z-index on navbar? Is it possible to remove it? I can't add z-index to the drawer, since adding z-index creates a new stacking context, which breaks blending (ex. either progressive blur breaks, or sheet content blending with backdrop breaks).
> 
> With all the blending in the new mocks, I think we are going to have to start using z-index sparingly, since it creates a new stacking context.

I don't think there is a way to avoid z-index, as we need to control the z order of absolutely positioned elements. We have a nice, centralized, [global z-index schedule](https://github.com/cybersemics/em/blob/f154b6cd98da33a77b59b128533d30ad90d6faca/panda.config.ts#L422-L454).

The only other way I know to control z order is by source order (literally the order of elements in the DOM), which offers less control and can sometimes create impossible situations if there are other source ordering dependencies.

raineorshine:
Sure, that seems acceptable.

yangchristina:
@fbmcipher Right now, if you see the video inside the PR description, the active button glow fades in after the modal slides open. This is because it doesn't blend properly with the backdrop if it starts blending with the backdrop before the backdrop has even faded in. Is this acceptable?

fbmcipher:
@yangchristina 

> For the will-change on the button element, I can't see any difference it makes:

On Chrome, there's no difference – it works with or without `will-change`.

But on Safari, without `will-change`, the mix-blend-mode isn't applied properly due to the compositing layer rendering bug we discussed.

Screenshot 2026-01-02 at 20 24 00

---

> Does the mixBlendMode: 'soft-light', even do anything? I can't see any difference with or without this mix-blend-mode.

Toggling it _does_ change the appearance for me. Here's what I see:

https://github.com/user-attachments/assets/52ba1f0a-8fd4-4563-93be-39a02dda1248

Would be curious to know if you're toggling in dev tools (and if so which browser) or in code w/ live server? Maybe the changes you're making aren't taking effect for some reason.

---

In any case, your comment shows the correct appearance of the button in all of the screenshots!

fbmcipher:
@yangchristina Having tested this a little bit, I also wanted to say – the touch gestures feel super smooth, and the way the background fades in and out with the gesture looks great!

src/recipes/fadeTransition.ts
```diff
@@ -29,11 +29,6 @@ const fadeTransitionRecipe = defineSlotRecipe({
         enterActive: { transition: `opacity {durations.medium} ease 0ms` },
         exitActive: { transition: `opacity {durations.medium} ease 0ms` },
       },
-      commandCenterDrawer: {
```
fbmcipher, line 32: I find myself missing the easing in the new drawer implementation. Without the easing, the animation feels a bit abrupt. 

The easing added a subtle _smoothness_ to the way the light and panel animate.

Is there a way for us to use the old easing curve with the new drawer?

src/components/CommandCenter/PanelCommand.tsx
```diff
@@ -15,6 +15,8 @@ interface PanelCommandProps {
   command: Command
   /** The size of the button. */
   size?: 'small' | 'medium'
+  /** Whether the glow effect is ready to be displayed. */
```
fbmcipher: As discussed, we might be able to remove this if the `opacity` bug hypothesis holds true.

fbmcipher (commented):
This generally looks great, and I'm really happy with the new drawer! I especially like the way the background glow is tied to the gesture, it feels really smooth.

Just a couple small bits from me.

fbmcipher:
@yangchristina Got you. Yeah, I don't see any difference with the `mix-blend-mode:lighten` on `button` either – feel free to remove it.

src/components/CommandCenter/PanelCommand.tsx
yangchristina: Removed [797ea6d](https://github.com/cybersemics/em/pull/3468/commits/797ea6d50534a59a4a0009b3cdd0ce0288bedfe8)

src/recipes/fadeTransition.ts
yangchristina: @fbmcipher is it okay to have the same easing for ease in and ease out? Here's with MUI's ease in, for both easing in and out.


https://github.com/user-attachments/assets/ea1251a1-a386-4452-879f-6d501735cc0c

src/components/CommandCenter/PanelCommand.tsx
fbmcipher: tyvm!

src/recipes/fadeTransition.ts
fbmcipher: I'm okay with this!

src/e2e/puppeteer/__tests__/ui.ts
```diff
@@ -62,5 +64,8 @@ it('CommandCenter', async () => {
   // wait for the command center panel to appear before taking screenshot
   await waitForSelector('[data-testid=command-center-panel]')
 
+  // wait for the command center panel to slide in fully before taking screenshot
+  await sleep(durationsConfig.medium)
```
raineorshine: Since animations cannot easily be covered by automated tests, we automatically reduce all animation durations to 0 in the e2e tests. This has the benefit of speeding up the tests without changing the animation start or end states. 

This is handled in https://github.com/cybersemics/em/blob/6efcce7bd22c01fdcb9006aa138b12c8cf79a829/src/util/durations.ts and is ensured by always using `durations.get(...)` instead of importing `durations.config` directly. If you do this correctly then you shouldn't need any `sleep` in the tests, but instead can wait one frame for the render or wait for a specific `data-testid`.

yangchristina: Fixed, thanks! [db3cf62](https://github.com/cybersemics/em/pull/3468/commits/db3cf62b6c61aa24a75af05498bef79b03acf955)

fbmcipher:
Thank you @yangchristina! I'll review tomorrow, when I have access to my physical development devices again.

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -102,17 +77,43 @@ const HiddenOverlay = () => {
...
+const CommandCenter = ({ mountPoint }: Pick<SheetProps, 'mountPoint'>) => {
   const dispatch = useDispatch()
   const showCommandCenter = useSelector(state => state.showCommandCenter)
   const isTutorialOn = useSelector(isTutorial)
+  const ref = useRef<SheetRef>(null)
 
-  const onOpen = useCallback(() => {
-    dispatch(toggleDropdown({ dropDownType: 'commandCenter', value: true }))
-  }, [dispatch])
+  const height = useTransform(() => {
```
fbmcipher, line 88: 1. These `useTransform` blocks are currently quite difficult to read. It's hard to know at a glance what exactly they're responsible for, what the values represent, and how they relate to one another. Some explanatory comments for each `useTransform` hook would be helpful!

2. I suspect that animating the `bottom` and `height` style properties is causing some of the performance regressions we're seeing in this PR. [Animating those properties causes a layout reflow](https://stackoverflow.com/a/53892597) each time the tweened value updates, which causes stuttering during animation. Instead, we should be animating using `transform` wherever possible. That way, we avoid the computational cost of layout reflows. (Animating `opacity`, on the other hand, is absolutely fine!)

fbmcipher (changes requested):
Thanks @yangchristina! We're getting closer, but there are two major issues we should address before merging.

## 1. Performance regression compared to main

If you compare the frame rate of the animation in your screen recording at the top of this PR to the existing implementation on `main`, there’s a noticeable regression.

- On `main`, the Command Center opens and closes at a smooth ~60fps.
- In this branch, there’s visible stuttering during both the open and close animations.

This could be related to the specific properties you're animating – animating `bottom` and `height` style properties could be causing layout reflows on every frame. I've left more thoughts on this in the code review.

## 2. Scroll zone conflicts with swipe-down-to-close gesture

There’s an interaction issue with the new swipe-to-close gesture:

- Swiping down on the **left side** of the Command Center closes it as expected.
- Swiping down on the **right side** of the Command Center (which coincides with the scroll zone) scrolls the content instead of closing the Command Center.

### Steps to Reproduce

1. Open the Command Center.
2. Swipe down the Command Center on the **right side of the screen**.

### Expected Result

- The Command Center slides down, tracking the finger.

### Actual Result

- The viewport scrolls instead.

https://github.com/user-attachments/assets/f87382dd-9457-492a-a2cb-5be78b5657b1

@raineorshine: I want to check this in with you too, because it's an edge case between interactions that hasn't come up yet.

My take: if the Command Center is open, and the user swipes down on it, that gesture should probably close the Command Center – even if they are swiping in the scroll zone. Scrolling should only take precedence when the swipe originates in the editor, even if the Command Center is visible.

Curious to know what you both think!

src/components/CommandCenter/CommandCenter.tsx
yangchristina: The reason I'm not using transform, is it is causing issues with mix-blend-mode. So far I haven't found a way to get this to work with `transform`. I will look more into this.

https://github.com/user-attachments/assets/ad9b868d-488d-4dda-881d-c68ddf7c781b

> Changing left or margin will trigger a redraw in layout layer (which, in turn, will trigger redraws in the other two layers) for the animated element and for subsequent elements in DOM.
> 
> Changing transform will trigger a redraw in compositor layer only for the animated element (subsequent elements in DOM will not be redrawn).

I think we do need to redraw the layout layer in order for the blending to work properly? So I'm skeptical it would work otherwise.

yangchristina:
@fbmcipher On Chrome, performance for this PR seems better on high resolution devices, but worse for low-tier mobile devices.

`main`:

https://github.com/user-attachments/assets/db6febd3-1d1e-4a6c-b0ac-7ab51132aaee

This PR:

https://github.com/user-attachments/assets/494837b2-245a-4085-816a-885e0e2e8bc5


## For low-tier mobile devices:

`main`:

https://github.com/user-attachments/assets/659e8acc-6ab9-486d-a45d-cfde3c2858b1



This PR

https://github.com/user-attachments/assets/23eeb762-b400-46c8-a529-e917e688f288

raineorshine:
No, the scroll zone is still fully functional above the Command Center.

The Command Center is not a modal... it's a virtual keyboard that exists side-by-side with the editor. They are both interactive, just in their respective areas.

src/components/CommandCenter/CommandCenter.tsx
fbmcipher: > The reason I'm not using transform, is it is causing issues with mix-blend-mode. So far I haven't found a way to get this to work with transform. I will look more into this.

Thank you for looking into it more deeply, @yangchristina. I know it's been challenging, but I'm really hoping if we figure out a performant solution here, it'll guide us along and smooth out further design implementations down the line. Your work on the progressive blur, for example, has helped Cindy implement the new Gesture Menu design.

> I think we do need to redraw the layout layer in order for the blending to work properly? So I'm skeptical it would work otherwise.

OK, truthfully I'm not 100% on the deeper details here, so I'm happy to trust your judgment on how this should best be implemented!

yangchristina:
@fbmcipher here is a picture of performance profiling for main (left) vs this pr (right). I'm not seeing any significant difference.

Screenshot 2026-01-27 at 8 37 56 PM

src/components/CommandCenter/CommandCenter.tsx
yangchristina: I'm not really seeing evidence right now that animating `bottom` is causing performance issues. The reason we could get around doing this on main, was the the backdrop went within the `transform`, making it slide up and down with the rest of the command center. Since we don't want this behaviour, the same approach won't work here.

fbmcipher:
OK! This is great – I'm glad we've tracked down the performance issue.

> One of the built in ones? Like "easeIn" | "easeOut" | "easeInOut"?

Yeah, I think that's okay. The main concern was consistency with the Sidebar, which at the time of this PR starting still used `SwipeableDrawer`. But I actually [recently submitted a refactor](https://github.com/cybersemics/em/pull/3654) that switches the drawer there over to a different library, so now we no longer *have* to use MUI's curve.

Let's go for `easeOut`!

---

Last but not least, I ran through another test and just wanted to raise a minor but noticeable difference between `main` and this PR. If you see the videos below, maybe you'll see what I mean:

### Steps to Reproduce

1. Open the Command Center on both `main` and this PR.

### Current Behavior

The glow image fades in abruptly. This is the behavior seen in this PR.

https://github.com/user-attachments/assets/2022672d-997b-463a-92f1-22f7e6d2b524

### Expected Behavior

The glow image fades in gently. This is the behavior seen in `main`.

https://github.com/user-attachments/assets/7b3a11cb-699b-4bc7-ab3f-19e7eb3be354



### Notes

At first glance it looks like the glow image does not fade *at all* in this PR – it looks like it's just flashing in.

But if you slow the video down, you _can_ actually see the glow image's opacity changing – it's just happening very quickly, and not over the lifecycle of the animation like it does in `main`.

This could be a few things – easing? animation duration? – but thankfully doesn't seem to be super serious, and is the last discrepency I can see at this time.

yangchristina:
@BayuAri Thank you! Fixed here c78c9b9

https://github.com/user-attachments/assets/744da3fb-ac6f-4d9a-8d23-0c9935141e5c

cc: @fbmcipher

yangchristina:
@fbmcipher puppeteer test failed due to timeout, would you mind rerunning it?
```
 FAIL   puppeteer-e2e  src/e2e/puppeteer/__tests__/drag-and-drop-favorites.ts > favorites drag and drop > should reorder favorites by dragging to the end or top of the list
Error: Test timed out in 20000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
```

BayuAri:
@yangchristina 

## Issue A: Alert is displayed behind the Command Center is fixed. 
Alert is displayed in front of the Command Center accordingly.


image

## Issue B: Command Center is blurry.  
## Step to reproduce
1. Create a thought
2. Swipe up to trigger Command Center


## Current behavior
Command Center is blurry

image

## Expected behavior
Command Center should be solid

yangchristina:
@BayuAri fixed thanks! [00abc88](https://github.com/cybersemics/em/pull/3468/commits/00abc885cc6138dc680069bdfd8245b771d21008)

https://github.com/user-attachments/assets/5929b910-7f25-49c5-aee0-c6b5ad8fa340

yangchristina:
@BayuAri fixed, thanks!

https://github.com/user-attachments/assets/e9eec7ee-046a-45f5-a2aa-a0dee188d77f

@fbmcipher is this calculation fine? I'm making it so the blur height starts at 0, and linearly increases, until 110px above the sheet when fully open.

```
  const blurHeight = useTransform(height, height => {
    // Start at 0, then smoothly grow with progress
    return height + 110 * sheetProgress.get()
  })
```

BayuAri:
@yangchristina 
List of issues:
- [x] [Issue A: Alert is displayed behind the Command Center](https://github.com/cybersemics/em/pull/3468#issuecomment-3842170973) >> **Fixed**
- [x] [Issue B: Command Center is blurry](https://github.com/cybersemics/em/pull/3468#issuecomment-3871412219) >> **Fixed**
- [x] [Issue C : [Mobile] Blurry area at the bottom of the screen](https://github.com/cybersemics/em/pull/3468#issuecomment-3909179925) >> **Fixed**
- [ ] Issue D: [Mobile] Command Center is in front of the Sidebar menu

## Issue D: [Mobile] Command Center is in front of the Sidebar menu
## Step to reproduce
1. Create a thought
2. Swipe up to open Command Center
3. Tap Hamburger button to open Sidebar Menu


## Current behavior
Command Center is  in front of the Sidebar menu


image

## Expected behavior
Command Center should be behind of the Sidebar menu

BayuAri:
@yangchristina 
List of issues:
- [x] [Issue A: Alert is displayed behind the Command Center](https://github.com/cybersemics/em/pull/3468#issuecomment-3842170973) >> **Fixed**
- [x] [Issue B: Command Center is blurry](https://github.com/cybersemics/em/pull/3468#issuecomment-3871412219) >> **Fixed**
- [x] [Issue C : [Mobile] Blurry area at the bottom of the screen](https://github.com/cybersemics/em/pull/3468#issuecomment-3909179925) >> **Fixed**
- [x] [Issue D: [Mobile] Command Center is in front of the Sidebar menu](https://github.com/cybersemics/em/pull/3468#issuecomment-3935008186) >> **Fixed**

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -121,155 +123,153 @@ const CommandCenter = () => {
...
-            disableRestoreFocus: true,
-            style: {
-              pointerEvents: 'none',
-              zIndex: token('zIndex.modal'),
-              backgroundColor: 'transparent',
-            },
-          }}
+          detent='content'
+          unstyled
+          mountPoint={mountPoint}
```
raineorshine: Can `mountPoint` be removed? It appears to default to `document.body` and work without explicitly setting it, but maybe I'm missing something.

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -121,155 +123,153 @@ const CommandCenter = () => {
...
+          tweenConfig={tweenConfig}
+          style={{ zIndex: token('zIndex.commandCenter') }}
```
raineorshine: Is there a reason that `style` is used instead of `className`?

```jsx
          className={css({
            zIndex: 'commandCenter',
          })}
```

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -102,17 +77,44 @@ const HiddenOverlay = () => {
   )
 }
 
+const tweenConfig: SheetTweenConfig = {
+  duration: durations.get('medium') / 1000,
+  ease: 'easeOut',
+}
```
raineorshine: Can we inline this? Better to have it co-located with the component.

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -102,17 +77,44 @@ const HiddenOverlay = () => {
...
+
 /**
  * A panel that displays the Command Center.
  */
-const CommandCenter = () => {
+const CommandCenter = ({ mountPoint }: Pick<SheetProps, 'mountPoint'>) => {
   const dispatch = useDispatch()
   const showCommandCenter = useSelector(state => state.showCommandCenter)
   const isTutorialOn = useSelector(isTutorial)
+  const ref = useRef<SheetRef>(null)
```
raineorshine: Let's call this `sheetRef` since it doesn't refer to the root node. There are different things going on in this component and I think the specificity would help here.

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -121,155 +123,153 @@ const CommandCenter = () => {
...
+              width: '100%',
+              bottom: 0,
             })}
+            style={{ opacity }}
+          />
+          <Sheet.Container
+            data-testid='command-menu-panel'
+            style={{
+              backgroundColor: 'transparent',
+              // Make sure it overrides any inline styles
```
raineorshine: Are these styles meant to override the default styles of `Sheet.Container`? I'm assuming there are stome styles that are still applied by default even though the Sheet is marked as `unstyled`.

If so, this might be a little clearer: `// override default Sheet.Container styles`

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -121,155 +123,153 @@ const CommandCenter = () => {
...
+              pointerEvents: 'auto',
+              boxShadow: 'none',
+              zIndex: 'auto',
+            }}
+            onScroll={e => {
+              /**
+               * Prevent scroll events in command center from bubbling up
+               * to window scroll listeners.
+               */
+              e.stopPropagation()
```
raineorshine: Otherwise what will happen?

It's good to explain *why* something is done, not just *what* is done. The current comment is basically a definition of `e.stopPropagation`.

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -121,155 +123,153 @@ const CommandCenter = () => {
...
+              height: '100vh',
+              width: '100%',
+              bottom: 0,
             })}
+            style={{ opacity }}
+          />
+          <Sheet.Container
+            data-testid='command-menu-panel'
+            style={{
+              backgroundColor: 'transparent',
```
raineorshine, line 201: Prefer `className(css(...))` for static styles that panda can optimize.

src/components/CommandCenter/CommandCenter.tsx
```diff
@@ -121,155 +123,153 @@ const CommandCenter = () => {
...
-              className={css({
-                display: 'flex',
-                flexDirection: 'column',
-                margin: '0 1.333rem calc(1.333rem + env(safe-area-inset-bottom)) 1.333rem',
-                gap: '0.889rem',
-              })}
+            <Sheet.Content
+              style={{
+                overflow: 'visible',
+              }}
```
raineorshine: Prefer `className(css(...))` for static styles that panda can optimize.

src/components/CommandCenter/PanelCommand.tsx
```diff
@@ -17,21 +17,22 @@ interface PanelCommandProps {
...
   const nodeRef = useRef<HTMLDivElement>(null)
   return (
     <FadeTransition
-      type={type === 'luminosity' ? 'activeButtonGlowLuminosity' : 'activeButtonGlowSaturation'}
+      type='activeButtonGlow'
       in={isActive}
       unmountOnExit
       nodeRef={nodeRef}
+      /** When `in={true}` on mount, ensures the initial opacity of the element is correct. */
+      appear
```
raineorshine: Let's clarify what "correct" means exactly. Is it that the opacity should be 0 or 1, or something else?

raineorshine (changes requested):
Thanks, looking good.

Overall structure looks good. Just some minor comments about comments.

The only thing I'm noticing with the behavior is that the animation is slower than `main` and so it's feeling sluggish to me. I know you worked through some [performance issues related to easing](https://github.com/cybersemics/em/pull/3468#issuecomment-3809364773), but if you could get the duration closer to `main` that would be great. I think the easing is okay... the animation should be slower at the end (i.e. some variation of ease-out).

src/components/CommandCenter/CommandCenter.tsx
yangchristina: Added comment: `/** Override default Sheet zIndex. */`

src/components/CommandCenter/CommandCenter.tsx
yangchristina: done, moved overflow and boxShadow in

src/components/CommandCenter/CommandCenter.tsx
yangchristina: done, updated comment and moved some styles to className, and removed ones that Sheet.Container already had by default

src/components/CommandCenter/CommandCenter.tsx
yangchristina: done

src/components/CommandCenter/CommandCenter.tsx
yangchristina: I recall I added it for this issue, where if you try to close the bottom sheet inside the scrollzone, something interfered, and it would cancel that. It doesn't seem to be working now though. Is this an issue?

https://github.com/user-attachments/assets/b969a58d-3bb4-46c8-b5ff-39981bb10a8a

src/components/CommandCenter/CommandCenter.tsx
yangchristina: I believe mountPoint was related to mix-blend mode, and having things blend together, if I recall correctly. Now that we aren't using mix-blend-mode, we no longer need mountPoint={mountPoint}

src/components/CommandCenter/PanelCommand.tsx
yangchristina: updated comment: [9f527c5](https://github.com/cybersemics/em/pull/3468/commits/9f527c59cd78e7b0998a036194a2ce87f7dc0c40)

src/components/CommandCenter/CommandCenter.tsx
yangchristina: renamed

raineorshine:
Thanks for the update! Please request re-request a review whenever you're ready.

src/components/CommandCenter/CommandCenter.tsx
yangchristina: @raineorshine @fbmcipher thoughts on this or what can be causing it?

raineorshine: @yangchristina I'm not quite following you. You added `e.stopPropagation()`, so we just need better justification for it in the comments. If you're not sure why it's there, you will need to go back and do some testing so that you are fully confident about it.

Or, perhaps you're raising a current problem? If it helps, you can describe the issue in terms of Steps to Reproduce, Current Behavior, and Expected Behavior, with the commit branch or commit hash where the behavior is observed. As it stands, I'm a bit fuzzy about what you're reporting/asking.

yangchristina: My bad, I’m describing the current problem.

`e.stopPropagation()` was originally added to fix this issue, but the fix doesn’t seem to work anymore and I’m not sure why.

**Steps to reproduce:**

1. Open the command center.
2. Place your finger on the **right edge of the screen** (within the app’s scroll zone), and inside the command center
3. Drag the command center downward with your finger to reduce its height.

**Current behavior:**
The sheet releases and stops following the finger.

**Expected behavior:**
The sheet should continue following the finger as it’s dragged.

This behavior occurs on the current commit.

If you repeat this but outside the scrollzone on the right side of the app, you will notice that the behaviour is different (works as expected)

fbmcipher: @yangchristina Given the interference is localized to the scroll zone, the most likely culprit is one of the body-level touch events in `MultiGesture` which enable/disable scrolling (or somewhere else in this file). This is probably the event you tried to stop propagation to:

https://github.com/cybersemics/em/blob/440ff7ec0cce0c77defeabd2f8f85c043c3b2663/src/components/MultiGesture.tsx#L135-L150

raineorshine: Thanks, that's clear!

`stopPropagation` can be sensitive to event ordering. Sometimes `e.relatedTarget` can be used to filter out events from unrelated components without having to resort to `stopPropagation`. Hopefully we can keep the Command Center decoupled from MultiGesture.

yangchristina: [f9c0cfd](https://github.com/cybersemics/em/pull/3468/commits/f9c0cfd1f5f2fde87c9dcc37af1918dec99c3a18)

@fbmcipher is this solution okay?

fbmcipher: @yangchristina I think so! It's a clean, general solution to the problem that avoids coupling the `MultiGesture` to other components, and it doesn't seem to interfere with the expected operation of `MultiGesture`.

src/components/CommandCenter/CommandCenter.tsx
raineorshine: Could you inline this since it's only used in one place?

package.json
```diff
@@ -117,6 +117,7 @@
     "lottie-react": "^2.4.1",
     "marked": "^17.0.4",
     "moize": "^6.1.7",
+    "motion": "^12.23.26",
```
raineorshine, line 120: Are `motion` and `framer-motion` both needed, or can we just use one? My understanding is that they are the same library.

src/components/MultiGesture.tsx
```diff
@@ -135,6 +136,14 @@ class MultiGesture extends React.Component<MultiGestureProps> {
     // enable/disable scrolling based on where the user clicks
     // TODO: Could this be moved to onMoveShouldSetResponder?
     document.body.addEventListener('touchstart', e => {
+      // Ignore touches outside MultiGesture's DOM tree (e.g. portaled overlays).
+      // Set disableScroll to prevent the browser from capturing the touch for native scrolling,
+      // which would break the overlay's own drag handling (e.g. sheet drag).
+      if (this.rootRef.current instanceof HTMLElement && !this.rootRef.current.contains(e.target as Node)) {
+        this.disableScroll = true
```
raineorshine: This doesn't makes sense to me. If we're abandoning the gesture, why would we need to `disableScroll`? The `MultiGesture` should only be disabling scroll (i.e. preventing `touchmove`) when there is a gesture in progress.

src/components/MultiGesture.tsx
```diff
@@ -135,6 +136,14 @@ class MultiGesture extends React.Component<MultiGestureProps> {
     // enable/disable scrolling based on where the user clicks
     // TODO: Could this be moved to onMoveShouldSetResponder?
     document.body.addEventListener('touchstart', e => {
+      // Ignore touches outside MultiGesture's DOM tree (e.g. portaled overlays).
+      // Set disableScroll to prevent the browser from capturing the touch for native scrolling,
+      // which would break the overlay's own drag handling (e.g. sheet drag).
+      if (this.rootRef.current instanceof HTMLElement && !this.rootRef.current.contains(e.target as Node)) {
```
raineorshine: Can this be done in `shouldCancelGesture`? `shouldCancelGesture` provides a way for callers to cancel gestures without `MultiGesture` having to know about outside state.

src/components/CommandCenter/CommandCenter.tsx
raineorshine: Going back to the original comment, this line of code still needs to be updated or justified. `stopPropagation` should generally be avoided in favor of `e.relatedTarget` checks. Modifying propagation tends to have more unintended consequences.

raineorshine (changes requested):
Hi, thanks for the update

> The only thing I'm noticing with the behavior is that the animation is slower than `main` and so it's feeling sluggish to me. I know you worked through some [performance issues related to easing](https://github.com/cybersemics/em/pull/3468#issuecomment-3809364773), but if you could get the duration closer to `main` that would be great. I think the easing is okay... the animation should be slower at the end (i.e. some variation of ease-out).

Did you get a chance to look into the animation? The goal is to match the behavior in `main`.

src/components/MultiGesture.tsx
yangchristina: disableScroll is necessary for this to work. I've tried other things, but only this works, but I haven't been able to find the reason so far.

yangchristina: Actually nevermind, using a preventDefault on touchmove inside command center also works

src/components/MultiGesture.tsx
yangchristina: That doesn't work, since `shouldCancelGesture` makes `abandon=true`, but we need `disableScroll=true,`, since the fix is doing `e.preventDefault()` inside of `touchmove` listener.

yangchristina: There are 2 options, either we put a `e.preventDefault` inside CommandCenter:
```
  const preventTouchMoveRef = useCallback((el: HTMLDivElement | null) => {
    if (!el) return
    /** Prevent native page scroll on touchmove. */
    const handler: EventListenerOrEventListenerObject = e => e.preventDefault()
    el.addEventListener('touchmove', handler, { passive: false })
  }, [])
```

or we make CommandCenter be `disableScroll=true` in `MultiGesture`. Ex. use a data attribute like `data-disable-scroll`, if a element has this data attribute, then always `disableScroll`.

Thoughts on which solution is better?

yangchristina: cc: @fbmcipher

package.json
yangchristina: Switcing all imports to just use `motion`. They are the same thing, it's just `react-modal-sheet` requires motion.

src/components/CommandCenter/CommandCenter.tsx
yangchristina: removed line

src/components/CommandCenter/CommandCenter.tsx
yangchristina: done

src/components/MultiGesture.tsx
raineorshine: Let's have the CommandCenter handle it. Thanks!

yangchristina:
@raineorshine is this ok easing?

https://github.com/user-attachments/assets/7bb2bd7e-d8c5-428e-b949-bb6dc7185a01


This is on main:

https://github.com/user-attachments/assets/1c96a700-73e6-455f-b5dc-7207113c4137

raineorshine:
That's better. It's still a bit too linear, lagging behind at the beginning of the animation compared to the stronger ease-out of the animation in `main`.

Here is what they look like side-by-side (`main` on the left and ec4350550ff on the right) 5 frames after the Command Center first becomes visible:
image

After 10 frames it has mostly caught up:
image

yangchristina:
@raineorshine Probably because we switched to using transform, it doesn't seem like easing is affecting performance at all right now. Added it back. [8dffe16](https://github.com/cybersemics/em/pull/3468/commits/8dffe164a250f162d6a874c933ba5ae63a9a6171)

using same ease as main:
Screenshot 2026-05-04 at 10 54 37 PM

easeOut
Screenshot 2026-05-04 at 10 53 31 PM


https://github.com/user-attachments/assets/94d57774-78d7-4251-9559-402d91d459af

raineorshine:
Thanks for the update. I'm still seeing some significant differences from `main`. There are differences in easing in both the opening and closing animations, and `main` appears to be smoother.

Here's the first frame of the closing animation. `main` on the left; 8dffe164a25 on the right. Recorded on iPhone 15 Pro. The gap down in the first frame is too large.

<p float="left">
  image
  
</p>

raineorshine:
Hi! Checking in. Please provide a status update at your convenience.

yangchristina:
> Hi! Checking in. Please provide a status update at your convenience.

@raineorshine I was missing the different easing on close before, fixed that. I'm not really sure where you saw a different opening animation, since they should have the same easing.

https://github.com/user-attachments/assets/c0015edd-8a8d-4b61-bf0b-bd2331a14b63



This is on main:

https://github.com/user-attachments/assets/1c96a700-73e6-455f-b5dc-7207113c4137

yangchristina:
Here is an easier video to see. Left is this pr. right is main.

https://github.com/user-attachments/assets/dfbc9d56-a25a-4c31-bae3-a2d25798c09f

raineorshine:
Can you include a specific question, proposed solution, or confidence level when sharing videos? Otherwise it’s hard for me to tell what feedback or action you’re looking for. 

My main job (and Faiz's) is to give you the tools you need to submit a solution you are confident about. It's less efficient for us to provide feedback on intermediate steps.

yangchristina:
> Can you include a specific question, proposed solution, or confidence level when sharing videos? Otherwise it’s hard for me to tell what feedback or action you’re looking for.
> 
> My main job (and Faiz's) is to give you the tools you need to submit a solution you are confident about. It's less efficient for us to provide feedback on intermediate steps.

@raineorshine You mentioned before that the animation for opening and closing the command center here was different from main. I still don’t notice a difference in the opening animation, but I updated the closing animation so it now matches main. If there’s still a mismatch in the opening animation, could you point out what you’re seeing?

fbmcipher:
@yangchristina Let me see if I can track down the mismatch @raineorshine's pointing out – I will update later today!

raineorshine:
@yangchristina Okay, we'll help you out since I know your time is limited.

fbmcipher:
@raineorshine @yangchristina My first move whenever I'm tracking down an animation disparity like this is to record both screens, align them in an NLE like Premiere/After Effects and then play them side by side – it makes frame variances way easier to spot, since you can scrub through frame by frame and see where they begin to diverge.

On the left is the animation as of `main` (the MUI implementation); on the right is the version at the head of this PR. I've slowed it down 3x so we can see what's going on more clearly.


https://github.com/user-attachments/assets/f9803f81-2e39-46b9-a152-fdc359a32670



## Notes

- The first open animation is noticeably different between the two. The MUI implementation seems to have a first-run animation cost — there's a visible hitch right at the start. 

- After that first open, the two are much more aligned on open/close, settle at around the same time and do appear to follow the same easing curve, but there's slight frame-by-frame differences between each.

- If you slow it down far enough, you can catch frames where this PR's animation is a frame ahead or behind the MUI one — most visible between 0:13 and 0:16.
  - My take: the animation generally feels smoother on the right: when you scrub one frame at a time, the amount the sheet transforms each frame is more consistent on the right than on the left. I don't have a code-level explanation for that yet, just noting what I see.

---

My read is that MUI's inconsistent first-open cost is probably what threw off the easing comparison between the two implementations; that the easing curves themselves are equivalent; and the animation in this PR performs better and more consistently than the outgoing implementation (no first-open animation stutter, and more consistent frame-by-frame movement).

Here's the real-time comparison as well, in case you want to scrub through and draw your own conclusions:

https://github.com/user-attachments/assets/44f5ef48-890b-4cfc-8b8f-cafac71918f1

Please do let me know what you both think!

raineorshine:
Hi, thanks for the detailed comparison.

Confusingly, I'm not exactly seeing the same thing as you, and I'm not sure if it's because of the physical device or what. I am running staging and 49bd1e0c0a side-by-side. In yours, they appear very close together. If anything, 49bd1e0c0a leads a hair. On my device, 49bd1e0c0a noticeably lags behind `main` in the first half of the opening animation, similar to what I reported in https://github.com/cybersemics/em/pull/3468#issuecomment-4328552478.

This certainly makes is challenging to be on the same page!

If you're seeing something different than me, my feedback is not going to be very helpful. I'll tweak the easing/duration myself until I'm satisfied with the result. In the future we can test on different devices and perhaps gain more insight into the differences.

PR #3537, opened 2025-12-30
Enable Browserstack iOS Tests

### LINKED ISSUE #2127: Restore BrowserStack tests
The WebdriverIO tests that run on [BrowserStack](https://www.browserstack.com/) have been broken for a while. They need to be re-enabled and fixed as they provide crucial test coverage of iOS Safari behavior.

|                        |                                                                                |
|------------------------|--------------------------------------------------------------------------------|
| Tests                  | https://github.com/cybersemics/em/tree/main/src/e2e/iOS                        |
| JSDOM Environment      | https://github.com/cybersemics/em/blob/main/src/e2e/webdriverio-environment.js |
| Config                 | https://github.com/cybersemics/em/tree/main/src/e2e/config                     |
| GitHub Action workflow | https://github.com/cybersemics/em/blob/main/.github/workflows/ios.yml          |

A BrowserStack account is required. To run WebdriverIO tests, add `BROWSERSTACK_USERNAME=your_username` and `BROWSERSTACK_ACCESS_KEY=your_access_key` to `.env.test.local` in the project root and run `npm run test:e2e:ios`.

You can fix the jest tests first and then convert to vitest, or convert to vitest from the beginning. Up to you.

karunkop:
Close #2127 

## Changes Made:

I have completely replaced the exisiting webdriverio infrastructure paired up with jest and `browserstack-local` lib with `Webdriverio` own test runner `@wdio/cli` which does all the configuration, framework and service such as `browserstack` management. Since `wdio` handles everything, i.e starting tunnel before tests, automatic stops after completion, we do not require manual tunnel setup like before. Also` wdio` is built specifically for browser/app automation (unlike vitest/jest which is for unit testing). `wdio` executes test suites with native `WebDriver` support via `@wdio/mocha-framework` and ` @wdio/browserstack-service` which is responsible for automatic tunnel, session, and credential management. `wdio` also provides lifecycle hooks that are very helpful for initiating a session more effeciently.

The configuration files are now divided as `wdio.base.conf.ts`  which contains common settings for ios safari testing, `wdio.browserstack.conf.ts` using `@wdio/browserstack-service` that creates a automatic tunnel with cloud browserstack service on the go, and `wdio.local.conf.ts` for local Appium testing which is only used incase we need to run tests on local IOS simulator.

Here is the relevant links to wdio documentation:

- https://v6.webdriver.io/docs/clioptions.html
- https://v6.webdriver.io/docs/configurationfile.html
- https://v6.webdriver.io/docs/browserstack-service.html

### Local Development

```
Run on BrowserStack
yarn run test:ios:browserstack
```
```
Run on local Appium (requires local setup)
yarn run test:ios:local
```

### Prerequisites

Environment Variables:
```
.env.test.local
BROWSERSTACK_USERNAME=your_username
BROWSERSTACK_ACCESS_KEY=your_access_key
```

### Start the Dev Server:

```
yarn start  # Runs on port 3000
```

### tap.ts and gesture.ts:  

Now updated to use `performActions` directly to avoid the automatic `releaseActions` call. Safari/XCUITest doesn't support the DELETE /actions endpoint (releaseActions) now which WebDriverIO's action().perform() calls automatically after performing. This was one of the main issues which was causing tests to fail.  Minor adjustments/updates to some other helpers are also done along the way as per wdio browser API.

> Note: There are still 5 tests that are skipped in `caret.ts`. These tests are needed to be studied in much more detail and debug accordingly.

As of now, running tests in CI is pretty straightforward. Although we need to do it manually for time-being as it is temporarily disabled expect when manually triggered. We can always check for builds in https://automation.browserstack.com/projects/em/builds. Local build and CI build are clearly identified because of their identifier. We can also walthrough the entire automation via https://automation.browserstack.com/projects/em/builds/CI/2?tab=tests&testListView=spec&details=2566272669 which was super helpful for debugging purpose :)

> Note: I also did manage to restore ios tests in our old setup with vitest as well. https://github.com/cybersemics/em/compare/main...karunkop:em:restore-browserstack-ios-tests. But I think it would be a good idea to upgrade the architecture with native webdriverio cli as it is very convenient and up to date.

### DIFF
--- a/src/e2e/iOS/helpers/paste.ts
+++ b/src/e2e/iOS/helpers/paste.ts
@@ -15,17 +15,15 @@ async function paste(pathUnranked: string | string[], text?: string): Promise<vo
     (_pathUnranked: string[], _text: string) => {
       const em = window.em as WindowEm
       em.testHelpers.importToContext(_pathUnranked, _text)
+
+      // Wait for React to render the DOM changes
+      return new Promise<void>(resolve => {
+        requestAnimationFrame(() => resolve())
+      })
     },
     _pathUnranked,
     _text,
   )
-
-  // wait until react has rendered the DOM changes.
-  await browser.execute(() => {
-    return new Promise<void>(resolve => {
-      requestAnimationFrame(() => resolve())
-    })
-  })
 }
 
 export default paste


src/e2e/iOS/helpers/waitForElement.ts
```diff
@@ -1,16 +1,15 @@
...
-  selector: string,
-  { timeout = 5000 }: WaitForElementOptions = {},
-): Promise<Element> => {
-  await browser.waitUntil(async () => !!(await browser.$(selector)).elementId, { timeout: timeout })
+/**
+ * Wait for element to exist.
+ * Uses the global browser object from WDIO.
+ */
+const waitForElement = async (selector: string, { timeout = 5000 }: WaitForElementOptions = {}): Promise<Element> => {
+  await browser.waitUntil(async () => !!(await browser.$(selector)).elementId, { timeout })
```
raineorshine, line 12: This might be a bit slow since it has to cross the API boundary twice. Does the global `browser` have a native way to wait for a selector? Otherwise we should probably do the polling in the inner await.

We can come back to this later as an optimization but I wanted to point it out.

src/e2e/iOS/helpers/editThought.ts
```diff
@@ -1,14 +1,19 @@
...
-const editThought = async (browser: Browser, value: string) => {
-  await browser.sendKeys([value])
-  return await waitForEditable(browser, value)
+ * Uses the global browser object from WDIO.
+ */
+const editThought = async (value: string) => {
+  // Use browser.keys() which works better on iOS Safari than sendKeys
+  // Type each character individually for reliability
+  for (const char of value) {
+    await browser.keys(char)
```
raineorshine, line 14: It might not be necessary to send the keys individually. After all, the native `input` event can even send multiple characters in a single event when typing quickly.

src/e2e/iOS/helpers/index.ts
raineorshine, line 1: We can migrate this to directly imported helpers in another PR as we did with Puppeteer.

src/e2e/iOS/helpers/index.ts
```diff
@@ -45,31 +62,14 @@ const helpers = {
   getSelection,
   hideKeyboardByTappingDone,
   isKeyboardShown,
+  keyboard,
```
raineorshine, line 65: Great, thanks for updating this to match our Puppeteer helper.

src/e2e/iOS/helpers/paste.ts
```diff
@@ -1,28 +1,26 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const em = (window as any).em
+      em.testHelpers.importToContext(_pathUnranked, _text)
     },
     _pathUnranked,
     _text,
   )
 
-  browser.pause(10000)
+  await browser.pause(10000)
```
raineorshine: TODO: Change to wait until

raineorshine (approved):
Overall looks good! Just need your help getting them to pass on my end.

Please merge `main` and resolve merge conflicts when you get a chance.

raineorshine:
I'm also seeing a few log files which maybe need to be added to the `gitignore` to exclude them from source control:

- browserstack.log
- log/
- logs/

raineorshine:
Passed!

https://automate.browserstack.com/projects/em/builds/Local+rainerevere/4

raineorshine:
Yes, I believe so. It looks like flakiness warnings are configured in the BrowserStack settings at https://automate.browserstack.com/projects/em/settings/smart_tags.

image

src/e2e/iOS/helpers/paste.ts
```diff
@@ -1,28 +1,54 @@
-import { Browser } from 'webdriverio'
-import { HOME_TOKEN } from '../../../constants'
-import { WindowEm } from '../../../initialize'
+import { HOME_TOKEN } from '../../../constants.js'
+import equalPath from '../../../util/equalPath'
 
-const em = window.em as WindowEm
```
raineorshine, line 5: Why replace typesafe `WindowEm` with unsafe `any`? Maybe you had a typing issue, but `any` is not an acceptable solution.

src/e2e/iOS/helpers/paste.ts
```diff
@@ -1,28 +1,54 @@
...
+async function paste(pathUnranked: string | string[], text?: string): Promise<void> {
   const _pathUnranked = typeof pathUnranked === 'string' ? [HOME_TOKEN] : (pathUnranked as string[])
   const _text = typeof pathUnranked === 'string' ? pathUnranked : text!
 
-  // Note: This helper is exposed because copy paste seemed impossible in headless mode. With headless false copy paste with ctrl + v seems to work. ??
+  // Capture the cursor before import to detect when it changes
+  const cursorBefore = await browser.execute(() => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const em = (window as any).em
+    const state = em?.testHelpers?.getState?.()
```
raineorshine: Are you sure you want to tightly couple the `paste` helper to `em.testHelpers.getState`? That function is not relied on anywhere else in the codebase and is strictly there for debugging.

src/e2e/iOS/helpers/paste.ts
```diff
@@ -1,28 +1,54 @@
...
     },
     _pathUnranked,
     _text,
   )
 
-  browser.pause(10000)
+  // Wait for the import to complete by checking if the cursor has changed or been set.
+  // importText sets the cursor to the last imported thought, so we wait for the cursor to change
+  // (if it was already set) or to be set to a non-null value (if it was null).
+  await browser.waitUntil(
```
raineorshine: `importToContext` is synchronous, so why would we need an async wait? If you're just waiting for a single frame to render the updated thoughts, maybe you could await `requestAnimationFrame` and avoid the complex cursor dependencies.

karunkop: Yes this makes sense. Since `importToContext` dispatches redux action immediately, it makes sense just to wait for react to render the DOM changes by awaiting a `requestAnimationFrame`

src/e2e/iOS/helpers/editThought.ts
karunkop: When keys are not sent individually, this webdriver error occurs:

`WebDriverError: An unknown server-side error occurred while processing the command. Original error: Error Domain=com.facebook.WebDriverAgent Code=1 "Key Down action 'o' must have a closing Key Up successor in '{
    type = keyDown;
    value = o;
}'" UserInfo={NSLocalizedDescription=Key Down action 'o' must have a closing Key Up successor in '{
    type = keyDown;
    value = o;
}'} when running "actions" with method "POST"
unknown error: WebDriverError: An unknown server-side error occurred while processing the command. Original error: Error Domain=com.facebook.WebDriverAgent Code=1 "Key Down action 'o' must have a closing Key Up successor in '{
    type = keyDown;
    value = o;
}'" UserInfo={NSLocalizedDescription=Key Down action 'o' must have a closing Key Up successor in '{
    type = keyDown;
    value = o;
}'} when running "actions" with method "POST"
    at async editThought (/Users/karunkop/Workspace/em/src/e2e/iOS/helpers/editThought.ts:12:3)
    at async newThought (/Users/karunkop/Workspace/em/src/e2e/iOS/helpers/newThought.ts:18:5)
    at async Context.<anonymous> (/Users/karunkop/Workspace/em/src/e2e/iOS/__tests__/caret.ts:31:5)`

package.json
```diff
@@ -234,5 +243,5 @@
     "vitest-localstorage-mock": "^0.1.2",
     "webdriverio": "^9.19.2"
   },
-  "packageManager": "yarn@4.12.0"
+  "packageManager": "yarn@4.5.1"
```
raineorshine: I think there was a merge mistake as this is reverting the yarn upgrade.

src/e2e/iOS/helpers/paste.ts
```diff
@@ -1,28 +1,31 @@
...
     _pathUnranked,
     _text,
   )
 
-  browser.pause(10000)
+  // wait until react has rendered the DOM changes.
+  await browser.execute(() => {
+    return new Promise<void>(resolve => {
+      requestAnimationFrame(() => resolve())
+    })
```
raineorshine: If you can move this rAF to the above `browser.execute` that would be even better.

raineorshine (changes requested):

src/e2e/iOS/config/wdio.browserstack.conf.ts
```diff
@@ -0,0 +1,112 @@
...
+  ],
+
+  // Hooks
+  // Navigate once at the start of the session
+  before: async function () {
+    await browser.url('http://bs-local.com:3000')
+    await browser.waitUntil(
+      async () => {
+        const body = await browser.$('body')
+        return await body.isExisting()
```
raineorshine, line 83: `return await` can always be simplified to `return` since waiting for the promise to resolve and then returning within an async function is equivalent to returning a promise.

src/e2e/iOS/helpers/editThought.ts
raineorshine: Strange, but will keep that in mind. Maybe losing some time on the round trip but I don't want to prematurely optimize.

src/e2e/iOS/helpers/gesture.ts
```diff
@@ -21,30 +28,44 @@ const gesture = async (
     yStart = yStart ?? windowSize!.height / 2
   }
 
-  const actions = browser.action('pointer').move({ x: xStart, y: yStart }).down().pause(waitMs)
+  // Build actions array for performActions
+  // Safari/XCUITest doesn't support the DELETE /actions endpoint (releaseActions)
+  // which WebDriverIO's action().perform() calls automatically after performing
+  const pointerActions: PointerAction[] = [
+    { type: 'pointerMove', duration: 0, x: Math.round(xStart), y: Math.round(yStart), origin: 'viewport' },
```
raineorshine, line 35: Curious why `Math.round` is needed?

Issue #2973, opened 2025-05-31
Inline PanResponder and remove react-native-web


raineorshine:
Initially our mobile strategy was to build a React Native app with a shared codebase for web using `react-native-web`. However, the web version was optimized so heavily for PWA on iOS during the prototype stage, that a React Native build is no longer necessary. Despite the many browser quirks and workarounds, we have achieved native-like performance with a PWA wrapped in Capacitor. A single platform codebase will be much easier to maintain in the long run.

Currently, the only component that uses `react-native-web` is the MultiGesture component, which uses `PanResponder`.

https://github.com/cybersemics/em/blob/aa6016e018a758bf219ca152f7fbca9643611107/src/components/MultiGesture.tsx#L2

The task is to extract `PanResponder` and its dependencies so that we can permanently remove `react-native` and `react-native-web` as project dependencies.

PanResponder source: https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/vendor/react-native/PanResponder/index.js

- [ ] Move `PanResponder` and dependencies to `/src/components/PanResponder`.
- [ ] Export `GestureResponderEvent` from `PanResponder`.
- [ ] Convert Flow to Typescript.
- [ ] Remove `react-native` and `react-native-web` from package.json.

PR #3608, opened 2026-01-12
Create inline PanResponder and remove react-native-web

codebuild-ro:
Resolves: https://github.com/cybersemics/em/issues/2973

- [x] Move `PanResponder` and dependencies to `/src/components/PanResponder`.
- [x] Export `GestureResponderEvent` from `PanResponder`.
- [x] Convert Flow to Typescript.
- [x] Remove `react-native` and `react-native-web` from package.json.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -229,7 +229,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/code-frame@npm:^7.0.0, @babel/code-frame@npm:^7.10.4, @babel/code-frame@npm:^7.12.13, @babel/code-frame@npm:^7.24.7, @babel/code-frame@npm:^7.25.9, @babel/code-frame@npm:^7.26.0":
+"@babel/code-frame@npm:^7.0.0, @babel/code-frame@npm:^7.10.4, @babel/code-frame@npm:^7.12.13, @babel/code-frame@npm:^7.25.9, @babel/code-frame@npm:^7.26.0":
   version: 7.26.2
   resolution: "@babel/code-frame@npm:7.26.2"
   dependencies:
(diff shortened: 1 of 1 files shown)


raineorshine (approved):

raineorshine:
@BayuAri Can you test that gestures still work on mobile? This is a pure refactor. Thanks!

raineorshine:
If it was ported over from an imported library to inline code without any modifications, why would there be a behavioral change? I wouldn't expect any changes to the PanResponder logic to be necessary if it was copied exactly. 

It may be fine, but there is some risk if we don't understand what caused the deviation in behavior.

BayuAri (commented):
@codebuild-ro 

The gesture drawing issue is now fixed. Animation is drawn accordingly. 
However, the scroll gesture is not working.
User can not scroll up and down in the scroll zone

Here is video comparison with `main`

https://github.com/user-attachments/assets/1f024c98-68ec-4b20-b96f-ba35392ff98f

BayuAri (commented):
@codebuild-ro 
Confirmed fixed now.
Gesture, drawing,  and scroll features are working as expected

src/components/MultiGesture.tsx
```diff
@@ -270,7 +270,7 @@ class MultiGesture extends React.Component<MultiGestureProps> {
   render() {
     const ref = React.createRef<HTMLDivElement>()
     return (
-      <View {...this.panResponder.panHandlers}>
+      <View panHandlers={this.panResponder.panHandlers}>
```
raineorshine: There should be no need to change the component's props if the API is preserved.

raineorshine:
Sorry for the unapprove, I was looking at old code.

Looks good now, thanks!

codebuild-ro:
@raineorshine 
I found the reason.
On iOS, it doesn't work because of this [update](https://github.com/cybersemics/em/pull/3641/changes#diff-c0d1c142ad5998f55e6583442bc0e707e395ef0f3e4b7a34488741aa1b2af138L78-R78).

PR #3619, opened 2026-01-13
ToolbarButton: remove fastClick

### LINKED ISSUE #3286: [iOS] Virtual keyboard closes on toolbar tap
## Steps to Reproduce

1. Create a thought
2. Keep the keyboard open
3. Tap the Bold button or Pin Thought button in the toolbar. 

## Current Behavior

The keyboard closes. 

## Expected Behavior

The keyboard should stay open.

ethan-james:
Fixes #3286

The [tapUp](https://github.com/ethan-james/em/blob/bf0930f567e8b2c3e8429f80a70b18b2aa475fef/src/components/ToolbarButton.tsx#L135) handler calls `e.preventDefault()` on iOS, which kept the keyboard open. When we switched `fastClick` to use an `onClick` handler, that handler activated too late to stop the keyboard from closing. I took the event handlers attached in `fastClick` and re-applied them directly to `ToolbarButton`, replacing the click handler with `onTouchEnd` on mobile.

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -104,7 +104,7 @@ const ColorSwatch: FC<{
     dispatch(formatSelection('backColor', selected ? 'bg' : (backgroundColor ?? 'bg')))
   }
 
-  /** Toggles the text color onTouchEnd or onClick on desktop */
+  /** Toggles the text color onTouchEnd or onClick on desktop. */
   const tapUp = (e: React.MouseEvent | React.TouchEvent) => {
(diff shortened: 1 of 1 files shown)


ethan-james:
Thanks @BayuAri! I actually see that same behavior (keyboard closes when tapping a disabled toolbar button) even as far back as bb8b2c7c07c1ff11de502de9456d863af73a49d0, before we changed the way `fastClick` works and introduced this regression. @raineorshine Do you have a preference for how disabled buttons should affect the keyboard?

raineorshine (approved):

PR #3640, opened 2026-01-15
Track mounted status in sizes cache

### LINKED ISSUE #3310: Subthought shifts up after expanding parent within 1 second of collapsing
Tested on iOS Safari. 

Related: #3097, #2783

## Steps to Reproduce

```
- a
- b
  - c
```

1. Set the cursor on `b`. 
2. Set the cursor to null. 
3. Within 1 second, set the cursor back to `b'. 
   - If you wait 1 second, the issue does not occur. 
   - This corresponds to the 1 second delay of the cached thought height in the size ref. 

## Current Behavior

`c` animated up as it fades in. 

https://github.com/user-attachments/assets/191ebb83-ab50-4cf3-b6d3-3f4b18650324

## Expected Behavior

`c` should only fade in, with no change in y position.

ethan-james:
Fixes #3310 

In `usePositionedThought`, the presence of a new cliff is detected by checking that a node does not exist in the `sizes` cache. Since there's a 1-second debounce when removing a node from `sizes`, it will erroneously set `isNewCliff = false` when that debounce is still in progress. I added a `mounted` property to the `sizes` objects that can be immediately set to false in order to calculate `isNewCliff` correctly.

It would be nice to revisit this debounce and figure out if it's still necessary. I see that it was merged as part of #1713, but that doesn't have much information about the individual commits. Do you happen to remember a test case for reproducing the original issue? I tried setting `SIZE_REMOVAL_DEBOUNCE` to 0 and didn't notice anything immediately breaking.

### DIFF
--- a/src/hooks/usePositionedThoughts.ts
+++ b/src/hooks/usePositionedThoughts.ts
@@ -21,7 +21,7 @@ const usePositionedThoughts = (
   }: {
     maxVisibleY: number
     singleLineHeight: number
-    sizes: Index<{ height: number; width?: number; isVisible: boolean; cliff?: number }>
+    sizes: Index<{ height: number; width?: number; isVisible: boolean; mounted: boolean; cliff?: number }>
   },
 ): {
   indentCursorAncestorTables: number
@@ -179,7 +179,12 @@ const usePositionedThoughts = (
           - [empty]
       */
       const isNewCliff =
-        !sizes[node.key] && cliff < 0 && prev && node.depth >= prev.depth && (prevCliff === undefined || prevCliff < 0)
+        // the node does not exist in the sizes cache, or it has been removed but is still cached (#3310)
+        !sizes[node.key]?.mounted &&
+        cliff < 0 &&
+        prev &&
+        node.depth >= prev.depth &&
+        (prevCliff === undefined || prevCliff < 0)
 
       // Capture the y position of the current thought before it is incremented by its own height for the next thought.
       const y = yaccum - (isNewCliff ? cliffPadding : 0)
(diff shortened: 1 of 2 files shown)


raineorshine:
Hi! Thanks for identifying this. I believe there was an old assumption that delaying the removal of an unmounted size would have no functional difference (only perhaps a difference in performance). It was assumed that once the thought was unmounted, it wouldn't matter which size was cached. However, this it ended up having an impact on `isNewCliff`. We added that construct later in order to address a different positioning issue, and it's understandable that we missed this very indirect dependency. Unless we track the cliff independently from height, we must assume that **unmounted sizes do have an immediate impact on layout.**

I trace the size removal debounce back to a55c915ddea90747f5f2d1a669d971881188777b. While it does not provide exact steps to reproduce, we do get an explanation of the purpose of the delay:

`
// Removing a height immediately on unmount can cause an infinite mount-unmount loop as the VirtualThought re-render triggers a new height calculation (iOS Safari only).`
`// Debouncing height removal mitigates the issue.`
`// Use throttleConcat to accumulate all keys to be removed during the interval.`

(The referenced `throttleConcat` has already been removed. From what I recall it may have introduced some timing issues. However, the removal of `throttleConcat` means that all the timeouts are firing at different rates, which could be having a negative performance impact as the layout gets re-rendered multiple times. Clearly this is a fragile piece of code that hasn't quite lived up to its original goal of efficiently garbage collecting cached sizes.)

**I would be content testing on Mobile Safari and confirming that the mount-unmount loop can no longer be reproduced in this PR.** We can always revert if it shows up again.

I have a thought about `mount` as a solution though. If you're calling `setSizes({ ..., mounted: false })` immediately in `removeSize`, it presumably defeats the purpose of the timeout/debounce. `setSizes` will trigger a state change and force a re-render. We might as well delete the size instead of adding an intermediary `mount` state. Making it an immediate change fits the above premise that unmounted sizes can have an immediate impact on layout via `isNewCliff` anyway.

So yes, the debounce is (hopefully) unnecessary, and it may be better to remove it now as I don't see the `mount` solution offering much as an incremental solution. We just need to 1) test on mobile Safari and make sure the mount-unmount issue is no longer a problem, and 2) make sure there isn't a negative performance impact.

Thanks!

ethan-james:
If #3645 works out, this can go away.

Issue #3607, opened 2026-01-11
Eliminate pause.ts if it's no longer used


raineorshine:
https://github.com/cybersemics/em/blob/6efcce7bd22c01fdcb9006aa138b12c8cf79a829/src/e2e/iOS/helpers/pause.ts

Issue #3606, opened 2026-01-11
Reduce/eliminate swipe pause in iOS test helper


raineorshine:
5 seconds is a huge amount of time to sleep. Let's reduce and ideally eliminate this artificial delay in order to speed up the tests.

https://github.com/cybersemics/em/blob/6efcce7bd22c01fdcb9006aa138b12c8cf79a829/src/e2e/iOS/helpers/swipe.ts

PR #3642, opened 2026-01-16
Eliminate duplication between browserstack configs

### LINKED ISSUE #3604: Eliminate duplication between browserstack configs
Eliminate duplication between browserstack configs following the DRY principle. This will reduce the amount of code, enforce consistency, and prevent identical implementations from drifting apart over time.

`wdio.browserstack.conf.ts‎`
`wdio.local.conf.ts`

karunkop:
Close #3604 , #3606, #3607

- Refactored such that the wdio config files share common configs and capabilities.
- Removed unused helper files.

### DIFF
--- a/src/e2e/iOS/config/wdio.base.conf.ts
+++ b/src/e2e/iOS/config/wdio.base.conf.ts
@@ -64,6 +64,21 @@ const baseConfig = {
   },
 
   // Hooks
+  // Check if app is running before starting any workers (only in local development, not CI)
+  // Reason: when app is not running, tests will run but fail with a timeout unnecessarily.
+  // wdio.local.conf.ts always runs locally, so this check always runs for it.
+  // wdio.browserstack.conf.ts can run locally or in CI, so we check !process.env.CI.
+  onPrepare: async function () {
+    if (!process.env.CI) {
+      try {
+        await checkAppRunning()
+      } catch (error) {
+        console.error(error instanceof Error ? error.message : 'App is not running on http://localhost:3000')
+        process.exit(1)
+      }
+    }
+  },
+
   // Navigate once at the start of the session
   before: async function () {
     await browser.url('http://bs-local.com:3000')
--- a/src/e2e/iOS/config/wdio.browserstack.conf.ts
+++ b/src/e2e/iOS/config/wdio.browserstack.conf.ts
@@ -1,6 +1,6 @@
 import dotenv from 'dotenv'
 import path from 'path'
-import baseConfig, { checkAppRunning } from './wdio.base.conf.js'
+import baseConfig from './wdio.base.conf.js'
 
 // Load .env.test.local before checking env vars since this file is imported
 // at module load time, before vitest's automatic env loading kicks in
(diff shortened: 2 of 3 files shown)


src/e2e/iOS/config/createIOSCapability.ts
```diff
@@ -0,0 +1,20 @@
...
+  browserName: 'Safari' as const,
+  'appium:automationName': 'XCUITest' as const,
+}
+
+/**
+ * Creates a base iOS Safari capability with the specified device name.
+ * @param deviceName - The device name (default: 'iPhone 15 Plus').
+ * @returns Base capability object.
+ */
+function createIOSCapability(deviceName: string = 'iPhone 15 Plus') {
```
raineorshine: I would say that the thinness of this function doesn't justify the abstraction. The only thing that is abstracted away is `appium:deviceName`. Worse, the caller already has to know about the appium prefixed keys because it uses `appium:platformVersion`. So you are getting no encapsulation value.

Add to this the fact that `createIOSCapability` breaks co-location, and we're looking at I think more downsides than upsides.

Storing the base capabilities on `baseConfig` and merging/overriding them with `[...baseConfig.capabilities, ...]` would eliminate the duplication without adding abstraction overhead.

src/e2e/iOS/config/wdio.base.conf.ts
```diff
@@ -1,5 +1,27 @@
+import http from 'http'
 import path from 'path'
 
+/**
+ * Checks if the app is running on locally.
+ * @throws Error if the app is not running.
+ */
+export const checkAppRunning = (): Promise<void> => {
```
raineorshine, line 8: I'm not sure I understand how this is different than `await browser.url('http://bs-local.com:3000')`. Could you explain?

karunkop: So this is done basically to ensure no browserstack sessions are created unnecessarily when the app has not started locally, so when `browser.url()` tries to navigate, it fails with a timeout error (in 30+ seconds). With `checkAppRunning()` function in `onPrepare()` hook, it runs and verifies before any browser sessions are created. If it fails/throws error, the process is immediately exited before creating any sesssions. Basically we don't need to wait for the whole timeout period to know that the tests were failed because of this particular error.

raineorshine: Doesn't that add a round trip to the browser though? Slightly slowing down all the passing tests in order to get the failing test to fail faster? Let me know if I'm missing something.

karunkop: This setup is just for local development. Also this isn't a browser round trip, it's a Node.js HTTP request from the test runner to `localhost:3000` before any browser sessions are created. However yeah this still adds slight overhead to test run.

raineorshine: Thanks. Could you clarify why the app would not be started locally? Is this to catch if the developer forgot to start the dev server before running the webdriver tests?

raineorshine: I see that `onPrepare` has the CI check in `wdio.browserstack.conf.ts` but not in `wdio.local.conf.ts`. Can you explain the difference here?

If `wdio.local.conf.ts` is always run locally, then `process.env.CI` will always be false and it seems like both configs can use the same `onPrepare`, unless I'm missing something.

karunkop: > Thanks. Could you clarify why the app would not be started locally? Is this to catch if the developer forgot to start the dev server before running the webdriver tests?

Yes exactly!

karunkop: > I see that `onPrepare` has the CI check in `wdio.browserstack.conf.ts` but not in `wdio.local.conf.ts`. Can you explain the difference here?
> 
> If `wdio.local.conf.ts` is always run locally, then `process.env.CI` will always be false and it seems like both configs can use the same `onPrepare`, unless I'm missing something.

Yep, wdio.local.conf.ts (appium testing) is only run locally. I have moved the onPrepare hook to the base config file.

PR #3643, opened 2026-01-16
Enable Browserstack iOS tests in GitHub Actions

### LINKED ISSUE #3603: Enable Browserstack iOS tests in GitHub Actions
Add `test:ios:browserstack` to the CI so we have automated test coverage on PR's just like vitest and puppeteer tests.

karunkop:
Close #3603 

- enabled browserstacks ios automated tests while opening PRs

For the ios test to work in CI, we need to add in `BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY` as secrets in this repository. @raineorshine

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -15,13 +15,7 @@ on:
 
 env:
   CI: true
-  COPILOT_AGENT_FIREWALL_ALLOW_LIST_ADDITIONS:
-    googlechromelabs.github.io,storage.googleapis.com
-    # BrowserStack credentials - WDIO's browserstack-service handles the tunnel
-  BROWSERSTACK_USERNAME: ${{secrets.BROWSERSTACK_USERNAME}}
-  BROWSERSTACK_ACCESS_KEY: ${{secrets.BROWSERSTACK_ACCESS_KEY}}
-  BROWSERSTACK_BUILD_NAME: 'CI - ${{github.run_id}}'
-  BROWSERSTACK_PROJECT_NAME: 'em'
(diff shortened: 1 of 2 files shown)


raineorshine (changes requested):
Thanks!

`BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY` are already stored as secrets in the repo. I think there is a problem with the configuration. 

image

If you look at the CI error, it is trying to inject `.env.test.local` which I don't think makes sense. Local configs should either not be loaded in the CI, or repo secrets should take precedence. I'm not sure. If you could look into our `.env` setup and see what makes the most sense that would be great.

karunkop:
> Thanks!
> 
> `BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY` are already stored as secrets in the repo. I think there is a problem with the configuration.
> 
> image
> If you look at the CI error, it is trying to inject `.env.test.local` which I don't think makes sense. Local configs should either not be loaded in the CI, or repo secrets should take precedence. I'm not sure. If you could look into our `.env` setup and see what makes the most sense that would be great.

https://github.com/orgs/community/discussions/140642#discussioncomment-10848724

I have tried by limiting the local secrets injection in CI and it still didn't work. Repos secret seems to be not accessible in PRs from the forked repo as stated here: 
`However, note that workflows triggered by Pull Requests from forks to the upstream repository still won't have access to secrets for security reasons.`
image

Looks like this will only work after the PR gets merged? @raineorshine

PR #3644, opened 2026-01-16
Do not split decimal numbers 

### LINKED ISSUE #3437: Split sentence should ignore decimal numbers
An unanticipated edge case for #3354.

## Steps to Reproduce

```
A - B 1.1
```

1. Activate Split Sentence. 
2. Activate Split Sentence again. 

## Current Behavior

Do nothing and alert the user that it is not possible. 

## Expected Behavior

The first activation should split in "-". 

The second activation should do nothing and alert the user that it is not possible.

Note that this should only be handled this way for decimal numbers, not 'a.a', which should split as usual. 

## Test

Add an appropriate test case.

codebuild-ro:
Fixes: https://github.com/cybersemics/em/issues/3437

- Add decimal handling section
- Add unit test cases for new logic

### DIFF
--- a/src/commands/__tests__/splitSentences.ts
+++ b/src/commands/__tests__/splitSentences.ts
@@ -145,6 +145,61 @@ describe('splitSentences', () => {
   - three.`)
   })
 
+  it('splits on dash when decimal number is present (ignores decimal period)', () => {
+    store.dispatch([
+      importText({
+        text: `
+          - A - B 1.1
+        `,
+      }),
+      setCursor(['A - B 1.1']),
+    ])
+
+    executeCommand(splitSentencesCommand, { store })
+
+    const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
+    expect(exported).toBe(`- __ROOT__
+  - A
+    - B 1.1`)
+  })
+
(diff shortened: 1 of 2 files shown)


src/commands/__tests__/splitSentences.ts
```diff
@@ -145,6 +145,61 @@ describe('splitSentences', () => {
...
+
+    executeCommand(splitSentencesCommand, { store })
+
+    const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
+    expect(exported).toBe(`- __ROOT__
+  - A
+    - B 1.1`)
+  })
+
+  it('does not split on dash when decimal number is present and already split', () => {
```
raineorshine: The test title is misleading, as none of the thoughts contain a dash. Maybe "does not split decimal numbers" would be more accurate.

raineorshine (changes requested):
It seems like this could be accomplished more simply by excluding decimals that are immediately followed by digits in the `mainSplitRegex`:

```diff
-  const mainSplitRegex = /[.;!?]+/g
+  const mainSplitRegex = /[;!?]+|\.(?!\d)/g
```

raineorshine:
Hi! Any update on this?

PR #3645, opened 2026-01-16
Remove debounce from removeSize

### LINKED ISSUE #3310: Subthought shifts up after expanding parent within 1 second of collapsing

ethan-james:
Fixes #3310, hopefully supersedes #3640

I ripped out [setTimeout](https://github.com/ethan-james/em/blob/f954fce7c614f5cfa2bc5ed273930368ac6901b6/src/hooks/useSizeTracking.ts#L26) entirely, rather than calling `setTimeout` with a zero delay, which is another option in case this approach fails.

I tested it with the repro steps from #3310, and will continue to test it in other ways.

### DIFF
--- a/src/hooks/useSizeTracking.ts
+++ b/src/hooks/useSizeTracking.ts
@@ -10,10 +10,6 @@ const useSizeTracking = () => {
   const fontSize = useSelector(state => state.fontSize)
   const unmounted = useRef(false)
 
-  // Removing a size immediately on unmount can cause an infinite mount-unmount loop as the VirtualThought re-render triggers a new height calculation (iOS Safari only).
-  // Debouncing size removal mitigates the issue.
-  // Use throttleConcat to accumulate all keys to be removed during the interval.
-  // TODO: Is a root cause of the mount-unmount loop.
   const removeSize = useCallback((key: string) => {
     if (unmounted.current) return
     setSizes(sizesOld => {


ethan-james:
I do see that if there are 2 thoughts being shown/hidden, the upper one is still treated as a new cliff even thought it has a child that will also be shown.

https://github.com/user-attachments/assets/6807c52b-fb7b-4e5e-8834-d14e2c5a5d6f

It also exists in #3640 as well as `main`.

src/hooks/useSizeTracking.ts
```diff
@@ -3,34 +3,23 @@ import { useSelector } from 'react-redux'
...
   // Track dynamic thought sizes from inner refs via VirtualThought. These are used to set the absolute y position which enables animation between any two states. isVisible is used to crop hidden thoughts.
   const [sizes, setSizes] = useState<Index<{ height: number; width?: number; isVisible: boolean }>>({})
   const fontSize = useSelector(state => state.fontSize)
   const unmounted = useRef(false)
 
-  // Track debounced height removals
-  // See: removeSize
-  const sizeRemovalTimeouts = useRef(new Map<string, number>())
-
   // Removing a size immediately on unmount can cause an infinite mount-unmount loop as the VirtualThought re-render triggers a new height calculation (iOS Safari only).
```
raineorshine: You can remove the comment as well.

raineorshine (approved):

raineorshine:
> I do see that if there are 2 thoughts being shown/hidden, the upper one is still treated as a new cliff even thought it has a child that will also be shown.

Thanks. If it's not currently being tracked, could you create a new issue for this?

raineorshine (approved):

BayuAri (commented):
Confirmed that the bounce animation has been removed when displaying subthought from hidden state.


https://github.com/user-attachments/assets/1b237772-b8cc-48be-b4c1-e285cdd684e9

Issue #3623, opened 2026-01-14
Command Center background not loaded in offline PWA


raineorshine:
This is a variant of #3492 that isn't covered by the solution in #3550.

## Setup

This is reproducible when em is added to the home screen using Safari's Add to Home Screen feature. This PWA behaves very similarly to the Capacitor app when the device is offline, though I can't say if the caching policy is any different.

The simplest way to reproduce this is to navigate to [staging](https://em-proto--staging2-cxls71a0.web.app/) in mobile Safari and add it your home screen and follow the instructions below.

To reproduce on localhost is a bit trickier, as a PWA app served with the live reload dev server **cannot be added to the Home Screen in fullscreen mode** (it will merely open the app as a new tab in mobile Safari). However, there is a workaround. You can `yarn serve:build` to serve the production build, which **can** be added as a full screen app to the home screen. Once it's added, kill the `serve:build` process and start the dev server. Since the IP and port are the same, you should be able to now access your localhost dev server through the PWA that was previously added to your home screen. If you kill it and reopen it a second time then it will clear the cache and now serve the live reload server.

## Steps to Reproduce

1. Open em as PWA with the internet on.
2. Enter Airplane mode and turn off WiFi.
3. Kill em from the app switcher.
4. Open em as PWA while offline.

## Current Behavior

The [service worker](https://github.com/cybersemics/em/blob/main/src/serviceWorkerRegistration.ts) will serve the entire app from the cache while offline, but the Command Center background image is not cached. When the Command Center is opened, the background image is missing.

https://github.com/user-attachments/assets/ff1940be-11c6-4c13-af78-463ab069eb11

## Expected Behavior

The Command Center background image should be cached and work while offline.

PR #3646, opened 2026-01-16
[Liminal UI] implement new background for GestureMenu

### LINKED ISSUE #3425: [Liminal UI] Begin implementing a new design for Gesture Menu
> **Begin implementing a new design for Gesture Menu, starting with the background glow and overlay.**

Image  
  
> **The new design for the Gesture Menu.**
  

The Gesture Menu appears while the user is inputting a gesture in **em**. It shows the user the command they are about to activate, and also shows a list of other commands that the user can activate at that moment. 

---

The new design for the Gesture Menu is made up of a few elements put together, shown below:

Image

This issue will focus on implementing the **first two elements** – the **overlay** and the **glow** – which are used together to create a **backdrop** which separates the Gesture Menu from the contents of the app underneath. 

**Importantly,** we **won't** implement the new design for the commands themselves in this PR. We'll do that in the next PR. This should help us prevent merge conflicts and long PR rounds.

So by the end of the PR, the expectation is a version of the Gesture Menu _in-between_ the old and the new design. That'll probably look something like this:

Image

#### Design notes

- This layer separates the **contents of the Gesture Menu** from **the contents of the rest of the app.**
- **The overlay is made up of:**
	- A linear gradient, which transitions from **black** at the top to **fully transparent** at the bottom.
	- A progressive blur backdrop filter which blurs the app's contents underneath, starting from **really blurry** at the top to **not blurry at all** at the bottom. This creates the impression of depth.
		- See `#3` in the top row of mocks below – "Philosophy" is much less blurry than the thoughts above it, which can't be read at all.
- **The overlay should change height depending on the size of the list of commands.**
	- Look at the image below for reference.
	- Look at the top three mocks below. See how "Poetry", "Trips", "Books" and "Films" are clearly visible in `#3` – while in `#2`, the overlay obscures them. See in `#1`, how they are even more obscured.
	- Look at the bottom three mocks. The blue rectangle shows the position and size of the overflow layer. Note that the overlay is consistently larger than the list of commands by a fixed amount, highlighted in red. Perhaps this could be implemented as a `padding-bottom`?

Image

#### Implementation notes
- **You can find the exact gradient used in the design in [its Figma component](https://www.figma.com/design/oQnevfxv79fhTJnOs0fz9k/em-design?node-id=5987-109170&t=1gGQGTbhbyoP48wC-11)**.
- Progressive backdrop blur can be done with pure CSS, but it's not a standard feature/single property. It might require some experimentation to get this effect working.
	- [Christina implemented progressive backdrop blur in the Command Center using masks](https://github.com/cybersemics/em/blob/ed28af075968265ebfe2e4da48457f782aae9da5/src/components/CommandMenu/CommandMenu.tsx#L146-L157) – that might be a good example to follow.

#### Design notes
- This is a simple translucent image that displays **on top** of the overlay. This brings our "liminal" design style onto the interface.
- Take a look at the image below. The first mock *just* shows the overlay layer. The second image brings in the glow layer, overlaid on top – which completes the interface.

Image

#### Implementation notes
- **A WEBP image for the background glow is available in the [assets folder](https://public.3.basecamp.com/p/TmFdohvobCnQpbMUqntjsw3L/vault/9357491211)**.
- The glow needs to scale smoothly and responsively depending on device width. Take a look at the image below for reference, which shows how this should roughly look on some different device sizes. Roughly speaking, the image should stay aligned **to the centre of the screen**.

Image

[Link in Figma](https://www.figma.com/design/oQnevfxv79fhTJnOs0fz9k/em-design?node-id=5987-111010&t=1gGQGTbhbyoP48wC-11)

**Note:** For now, please ignore the contents of the Gesture Menu and how it breaks into multiple columns – we'll tackle that in an upcoming PR. Just look at how the background glow and blur scales gracefully across different device sizes. That's the behaviour we'd like to see in this PR.

---

Please feel free to ask any questions, or if there's anything I can help clarify!

cindmichelle:
Close #3425 

## Description
implementing new background design for GestureMenu. The background is now split into three components:
- **ProgressiveBlur:** Handles the gradual blur effect for the gesture menu background.
- **Overlay:** Renders a gradient overlay for the gesture menu.
- **Glow:** Renders the glow effect for the gesture menu.

https://github.com/user-attachments/assets/09922ed3-dcd9-4b46-82c8-cb4be773529f


src/components/GestureMenu.tsx
```diff
@@ -39,8 +39,6 @@ const GestureMenu: FC<{
         className={css({
           marginBottom: 0,
           textAlign: 'left',
-          border: '1px solid {colors.gray15}',
-          backgroundColor: 'gray09',
```
cindmichelle, line 43: @raineorshine @BayuAri you might seeing the  regression tests failing, caused by the removal of line above.

This change was intentional to match the updated design, which no longer includes a border.

Since the border was very thin, the difference is quite subtle and not immediately noticeable in the [visual regression outputs](https://github.com/cybersemics/em/actions/runs/21080321671/artifacts/5159549768), but it’s enough to trigger snapshot mismatches

raineorshine: Thanks! If that’s the case, please update the snapshots. The PR can only be merged when all the tests pass and there is no flakiness.

src/components/GestureMenu.tsx
```diff
@@ -82,6 +81,58 @@ const GestureMenu: FC<{
...
+        width: '100%',
+        top: 0,
+        height: '100%',
+      })}
+    />
+  )
+}
+
+/** Renders the glow effect for the gesture menu. */
+function Glow() {
```
fbmcipher: Do you think we should preload the glow image to prevent it flashing in late, like you did in #3550?

fbmcipher (changes requested):
Thank you so much @cindmichelle! This is a great initial PR for this feature. I really like the split out components for `Glow`, `Overlay` and `ProgressiveBlur` – very clean and easy to read!

The code seems good to me – just a couple design discrepancies I noticed. Let me know if I can help clarify any of these.

---

## 1. The `Glow` image should not resize with the list of visible commands

### Steps to Reproduce
1. Input a gesture.
2. Observe the `Glow` image as the list of commands shortens.

### Expected Behavior
- The `Glow` image stays fixed as a full-screen overlay, and does not resize.

### Actual Behavior
- The `Glow` image resizes on each shortening of the command list, creating a visually disruptive effect.

### Notes
Screenshot 2026-01-17 at 01 06 52

- ※ See the screenshot above for the intended behavior. In the first row, the Glow image remains fixed while the command list changes. (The command list is removed in these designs for clarity.)

- The `ProgressiveBlur` and `Overlay` *should* resize. This is implemented correctly in your PR (thank you!) It's just the `Glow` image that should stay fixed in place.


---
## 2. The wrong portion of the `Glow` image is shown

- While the `Glow` should stay fixed on the same screen size, it ***should* resize between different device sizes.** This is detailed in the **"Implementation notes"** section of the **Glow** component spec in #3425. 
	- We actually use a **subset** of the provided `Glow` image on mobile. You might need to use some CSS cropping to get the right portion. Try to get the same portion of the `Glow` image. See illustrated below: you can see how much of the image is on-screen and how much is off-screen. For example, we only use about ~40% of the width of the image on mobile.
	- **The general rule**: expand the visible area of the image as the browser resizes (cropping less of the image), and **only once the full image is visible** should scaling occur.
		- **Example**: on mobile, ~40% of the image width is visible; on tablet, more is revealed; on large desktop, the image begins scaling.

Screenshot 2026-01-17 at 03 33 30

BayuAri (commented):
Background image is implemented, I have nothing to add, already covered by @fbmcipher.

The resized glow is more visible on landscape view.  

image

src/components/GestureMenu.tsx
```diff
@@ -69,6 +67,7 @@ const GestureMenu: FC<{
                   gestureInProgress={gestureInProgress as string}
                   key={command.id}
                   selected={isCheatsheetMatch || gestureInProgress === gestureString(command) || isCancelMatch}
+                  // selected={false}
```
raineorshine: Looks like this was just for testing.

raineorshine (changes requested):

src/components/GestureMenu.tsx
cindmichelle: You’re right, we’ll need to add the preload like we did for CommandCenter as well. I’m thinking of holding off on that for now until this issue #3623 resolved, since applying it with the approach #3550 would likely still hit issue in PWA

BayuAri (commented):
## Glow image resize when the list of gesture menu updated.
## Step to reproduce
1. Open the app on iPhone
2. Tilt the device to see the landscape mode
3. Swipe left and hold
4. While holding swipe down, don't lift your finger
5. Still holding, swipe right


## Current behavior
Glow image is moving on step 3 to 4.
Glow image is not moving on step 4 to 5

Note: The glow image movement is not obvious on iPhone portrait mode but maybe this is due to the narrow width screen size, it is quite visible on landscape mode.

**Video:**


https://github.com/user-attachments/assets/d69c182e-adcd-42e0-84f2-2d634bbf5312



## Expected behavior
Glow image should not move when user adds more gesture

cindmichelle:
Hi @BayuAri, apologies I think I accidentally pushed the changes to a different branch earlier, not this one. That’s probably why the background felt a bit clunky when you tested it.

Would you mind testing again using the latest commit I just pushed? Looks like the sudden background shift is gone now.

Let me know if you spot anything else I should fix 🙏

fbmcipher:
Thanks @cindmichelle for this great work, and @BayuAri for spotting the glow image resize bug!

(FWIW, it looks like that's been fixed after the changes you pushed, @cindmichelle!)

---

This is amazing work @cindmichelle! It's really shaping up. There's just a couple small things that I noticed during review.

## 1. Gesture Menu background should go under the status bar in Capacitor

@BayuAri already opened this issue in #3727, so I won't repeat here! I left a comment there with some extra detail on this issue. It would be great if you could take a look at this issue after this PR is merged. It should only be a small fix.

If it is not too time-consuming, it could be a good idea to test PRs in Capacitor as well. Though **em** is mostly used in Mobile Safari today, ultimately **em** will be released as a native app using Capacitor, and the designs have been made with Capacitor in mind. So it's worth us testing with Capacitor to make sure we implement everything correctly!

## 2. Blur "snaps in" after `FadeTransition` in Mobile Safari

### Steps to Reproduce

1. Start inputting a gesture.
2. Watch the editor and the gesture trail as the Gesture Menu fades in.

### Current Behavior

The Gesture Menu fades in, but the blur is not applied until the animation ends. This looks like the blur "snapping" in, which feels unpolished:

https://github.com/user-attachments/assets/a8e4d312-a6d1-4c89-806e-61e5f9a44f0b

### Expected Behavior

The Gesture Menu fades in with the blur already applied – like in Chrome:

https://github.com/user-attachments/assets/b4f6ce48-42b6-43f9-bc44-fe83212d2d11

### Notes

This is a [known Safari bug](https://stackoverflow.com/questions/74085350/backdrop-filter-doesnt-apply-if-parents-opacity-is-less-than-1). `backdrop-filter` doesn't properly apply if an ancestor's opacity is less than `1`. This bug doesn't appear in Chrome.

The solution is to control the `backdrop-filter`'s opacity _directly_, rather than through an ancestor. So to solve this problem, we'd probably need a separate `FadeTransition` that fades in `ProgressiveBlur` **separately and at exactly the same time** as the glow image.

When using `backdrop-filter` or progressive blurs, it is worth testing on Safari as well as Chrome, and keeping an eye out for subtle issues like this. These are really subtle issues, though, so I totally get that they might slip under the radar.

It's unfortunate that we need to hack around so many things like this in Safari, but from my experience it seems to just be a matter of reality when building complex web applications that target iOS.

I've encountered some of these issues with Christina, and through working through them together I'm beginning to understand some of the common issues with `backdrop-filter` on iOS and how to solve them. So if you run into issues in the future, please don't hesitate to ask!

BayuAri:
@cindmichelle 
Aside from what @fbmcipher has been pointed out, this looks good on me.
Blur background image is static on iOS and Mobile Safari on landscape and portrait

cindmichelle:
@fbmcipher thanks again for the feedback! i'll try to looking into these again tomorrow!

However since the component that handling gesture trace now lives in a different component from GestureMenu, moving it above the blur layer would require extra work and testing. It also feels slightly out of scope for this ticket (which is about implementing the new background). Are we happy if I handle that into a separate ticket?

cindmichelle:
Hi @fbmcipher and @BayuAri , I’ve just pushed the updates based on Faiz’s latest feedback!

Just to be clear, I'll only cover feedback 1 and 2 for now.

Regarding feedback 1, I decided to move the `overlay50` background-style one level deeper so it can be included within the FadeTransition,due to putting `PopupBase` inside `FadeTransition` would reintroduce the previous issue with the `backdrop-filter` style again. 

And based on my understanding, this issue is mainly a styling concern, and it’s still possible to resolve it by adjusting the layer structure. I also feel that `PopupBase` itself not necessarily be wrapped inside FadeTransition, since it also can act purely as a positioning container to ensure the content appears at the top layer when rendered.




---

For feedback 2, initially I planned to simply add a fade-out animation by introducing an extra local state in `GestureMenuWithTransition` (https://github.com/cybersemics/em/pull/3646/changes/354c95b200b80cabcb27d0922067cfff11cfd54c) . However, I noticed that this approach would still be problematic. When the GestureMenu fades out, its content would snap out immediately due to the current gesture state behavior, which wipes the state as soon as the gesture trace is no longer in progress.

https://github.com/user-attachments/assets/06b30e56-bf95-4437-988e-6211dad52aad


Because of that, I decided to take a different approach by introducing a lifecycle state for the GestureMenu. The new state in the gesture mini-store now has the following variants:
- `hidden`: GestureMenu is not rendered and not animated
- `entering`: GestureMenu container is rendered but not yet animated
- `visible`: GestureMenu container is rendered and the fade-in animation starts
- `exiting` : GestureMenu container is rendered and the fade-out animation starts

With this lifecycle state introduced, it also impacts the MultiGesture component. Previously, it would reset the gesture mini-store state on touchEnd. Now, it will only reset when the lifecycle state is hidden.

### Result:
https://github.com/user-attachments/assets/4f87a9b8-361f-4de3-a3cd-7db56410c569


Let me know if anything is still unclear!

cindmichelle:
@fbmcipher @BayuAri please dont review my latest changes for now. i saw a bit of regression on my latest commit where the glow are no longer full screen. im fixing them now

src/stores/gesture.ts
```diff
@@ -4,12 +4,52 @@ import cancelShortcut from '../commands/cancel'
 import openHelpCommand from '../commands/help'
 import reactMinistore from './react-ministore'
 
+/** Animation states for the gesture menu. */
+type GestureMenuAnimationState = 'hidden' | 'entering' | 'visible' | 'exiting'
```
fbmcipher, line 7: This solves the problem, but I think there's a simpler approach that touches less code and maintains less state.

Do you think a single `gestureMenuActive` boolean could solve the issue? We could set it to true when the menu opens, and false when the exit animation completes via `onExited`. 

Then, `clearGesture` just checks that flag. That way you keep `fadeAnimation` and `shouldRender` as local state and avoid the 4-state machine entirely.

fbmcipher (changes requested):
Thanks – this is closer but I'm still seeing elements pop in and out when the gesture menu shows and hides. There shouldn't be anything that pops in and out – elements need to fade smoothly. See the video attached below:

https://github.com/user-attachments/assets/66c33c07-dc53-4836-b23b-4f6ed134e78c

I'd definitely encourage you to screen-cap and slow the footage down to get a closer look at the animations. It might feel a bit contrived, but users will notice subtleties like this, and it'll take away from the sense of elegance and smoothness that we're aiming for.

I also left a comment on the general approach used – I think there's a simpler solution we could use.

src/stores/gesture.ts
cindmichelle: Hi @fbmcipher, just to clarify, are you suggesting that instead of introducing `gestureMenuAnimationState`, we use `gestureMenuActive` stored in the gesture ministore?

My concern with that approach is that we would then need to coordinate four different boolean states to drive the modal animation lifecycle (`showGestureMenu`, `gestureMenuActive`, `shouldRender`, `fadeAnimation`). That makes the flow harder to reason about, since we’d need to ensure all of them stay in sync to avoid invalid UI states. It also increases cognitive load, as the lifecycle becomes distributed across multiple flags.

In contrast, I think using `showGestureMenu` + a single `gestureMenuAnimationState` (e.g. 'entering' | 'entered' | 'exiting' | 'idle') keeps the lifecycle more explicit and reduces the chance of conflicting states, since the animation phases are modeled directly rather than coordinated across multiple booleans.

Let me know what you thought on this!

cindmichelle:
> @cindmichelle Sorry to jumping into this, I saw below issue with tracing.
> 
> ## Tracing is displayed behind the Command Center
> ## Step to reproduce
> 1. Create a thought
> 2. Swipe up to trigger Command Center
> 3. Draw a tracing that would crossing the Command Center
> 
> ## Current behavior
> Tracing is shown the Command Center
> 
> image
> **Video:**
> 
>  Tracing.is.displayed.behind.the.Command.Center.MP4 
> ## Expected behavior
> Refer to `main`, tracing should be in front of the Command Center

Hi @BayuAri thanks for the feedback earlier! but since the latest feedback is more related to gesture trace, would you like to add into #3791 instead? This PR is intended to revamp the gesture menu background.

I can take a look at it once I start working on that issue, thanks :)

cindmichelle:
@fbmcipher I found a way to prevent the blur from disappearing immediately during the fade-out animation by adjusting the animation timing. For the fade-in, I’m using ease-out so the blur appears immediately and feels more natural. But for the fade-out, I switched to `easeInSlow`. (i initially tried `cubic-bezier(0.9, 0, 1, 1)` , but then I tested the existing `easeInSlow` config and didn’t notice any meaningful difference between the two.)

https://github.com/user-attachments/assets/4f3b5d9e-cbf7-4aaa-8527-ae96369a305d

src/stores/gesture.ts
raineorshine: I think the issue here is that application state (gesture) is being mixed up with UI state (animation). 

What's the source of truth? The MultiGesture is the source of truth, and it propagates that truth to the rest of the application through the `gestureStore`. It's a bit strange to embed the animation state of a specific component into a global store. 

A better approach would be to have the `GestureMenu` keep its own `delayedGesture`/`displayGesture` state. This is similar to other delayed local state. 

The only problem I can see with this is if there are other components that depend on the delayed gesture state. In that case we really will need to make it global. Otherwise we can encapsulate the animation state in the `GestureMenu` component and keep a cleaner separation between application state and UI state.

raineorshine (changes requested):
# Gesture Trace flashes as Gesture Menu closes

## Steps to Reproduce

1. Enter an invalid gesture at the bottom of the screen and hold.
2. Release after the Gesture Menu appears.

## Current Behavior

57cbd6f01a1

Gesture Trace becomes bright before fading out.

https://github.com/user-attachments/assets/02938ed8-ca51-460d-b415-d56e281189c0

## Expected Behavior

`main`

Gesture Trace should fade out gracefully.

https://github.com/user-attachments/assets/e5863049-fcb2-4580-a4f2-36b853ea968e

raineorshine (changes requested):
## Steps to Reproduce

1. Enter an invalid gesture at the bottom of the gesture zone and hold until the Gesture Menu appears.
2. Release.

## Current Behavior

9ce2f1cd1d7

There is a slight delay and then the Gesture Trace fades out.

## Expected Behavior

`main`

The exact timing of the Gesture Trace animation should be preserved. The start, end, easing, and duration have all been carefully calibrated through real-world usage. The animation timing in this PR should exactly match `main`.

fbmcipher:
@cindmichelle Here's what I see – this video slows down the Gesture Trace and Gesture Menu fadeout animation so we can compare the behavior in this PR to the correct behavior in `main`. (left: `main`, right: this PR)

https://github.com/user-attachments/assets/02541654-30a4-4937-82fd-0f5ea1ee5faf

In `main`, the Gesture Trace begins to fade out out as soon as the finger is released from the screen. In this PR, it looks like the Gesture Trace does not begin to fade out until _after_ the Gesture Menu fades out.

I often sync screen recordings and slow them down like this – I find it's very useful to compare and analyse animations when they're slowed way down. I would definitely recommend giving this a try too!

---

@raineorshine Given the timing is sensitive in this PR, there's another divergence I wanted to flag up – though this one was intentional, and I wrote it in the spec. At 0:01 in this video, when the user releases their finger, the Gesture Menu in `main` disappears without any animation. On the right, in this PR, the Gesture Menu fades out.

Is this change acceptable? Maybe the Gesture Menu can fade out with the same timing as the Gesture Trace.

raineorshine:
Yes, I'm good with that.

cindmichelle:
Hi @fbmcipher @raineorshine  I've updated this PR with the fixes for the 2 feedback items above. 

About the autofocus jerk, this happens because the size of command list exceeds the gesture menu's container height (100vh). This behavior actually existed in the old design, but it's only noticeable now due to the fadeout animation on the gesture menu, which causes the layout tree shift happen in parallel while the gesture menu is still rendering.

Currently I only applied a quick fix to prevent the command list from overflowing. As for limiting the number of commands to render at a time, I think it's better to handle that in https://github.com/cybersemics/em/issues/3801 so this PR doesn't get too large in scope. Here's a video of the correct behavior that I've slowed down

https://github.com/user-attachments/assets/ee24edc3-1615-4887-8e38-dbdb6140e778

Following that, I've also fixed the gesture menu and trace gesture to start fade out at the same timing while preserving the TraceGesture fadeout duration (GestureMenu 200ms, TraceGesture 400ms)

https://github.com/user-attachments/assets/acbd7983-e0dd-42ce-a888-d09f128f11a9

BayuAri (commented):
- [x] [[iOS Safari] Autofocus jerks after New Subthought](https://github.com/cybersemics/em/pull/3646#issuecomment-4194995564) >>  **fixed**
- [x] [There is a slight delay and then the Gesture Trace fades out](https://github.com/cybersemics/em/pull/3646#pullrequestreview-4064613887) >> **Fixed**

raineorshine:
Thank you! I think we're close to merging.

@cindmichelle Puppeteer test is still failing.
@fbmcipher If I could get one more thorough review from you on this that would be great.

fbmcipher:
@raineorshine @cindmichelle I think this is ready to go. There are a couple of small issues – but I feel like they should not hold up this PR as they feel like they're more general issues to think about and fix elsewhere.

### There's still some noticeable banding on the Gesture Menu

The situation is much better, but still kind of noticeable on my iPhone 12 mini. Cindy and I had a bit of back-and-forth about this on Discord to work out the best process to generate banding-free images. The image itself is high-quality enough – it looks great in Photoshop on my Mac – but it seems like as soon as you're in a browser, different rules apply.

I'm not sure what the solution to this problem is yet. Maybe it's as simple as trying another different file format or playing with the CSS – or maybe we've run into an intractable problem with the current way we're rendering these glows and need to figure out something different (SVG? overlapping CSS gradients?) Or maybe there's something I haven't considered.

For now I think it's okay to merge, just wanted to flag it.

### Slight delay when first loading image

I don't want this to hold up the merge for this pull request because it doesn't seem to happen all the time. But this is something I've noticed in this PR, and in some of my other liminal implementations too. The very first time that glow image gets painted on the screen, there's a slight stutter. I think this is some kind of first-time image decoding cost.

Just wanted to note this here. If I continue to notice it in testing, I'll open a separate issue for it and we can tackle it elsewhere. If we find a general fix, this logic could be shared between all glow images via `useGlowImage()`.

### Need to merge into `main` again

This is just a small thing, but to get the dev server working I had to manually merge `main` in my local branch. @cindmichelle I think you just need to merge in `main` and push one more time before this PR can be merged!

cindmichelle:
@fbmcipher @raineorshine ive run `git merge` into main, this PR should be ready for merge

src/components/GestureMenu.tsx
```diff
@@ -99,38 +181,79 @@ const GestureMenuWithTransition: FC = () => {
     sortActiveCommandsFirst: true,
   })
 
-  // if dismissed, set timeout to 0 to remove alert component immediately. Otherwise it will block toolbar interactions until the timeout completes.
+  // Sync Redux showGestureMenu to gestureStore animation state
+  React.useEffect(() => {
```
raineorshine: Prefer `useEffect`

raineorshine (approved):
Thanks everyone for all your work.

@fbmcipher Good points. If you could open new issues for those that would be great.

Issue #3555, opened 2026-01-02
[iOS Capacitor][Android] Toolbar of the Customize Toolbar page is overlapping with device's status bar


BayuAri:
## Step to reproduce
1. Tap Setting button
2. Tap Customize Toolbar button
3. Scroll up the screen

## Current behavior
Toolbar is sticky at the top of the screen but overlapping with device's status bar

<img width="782" height="890" alt="Image" src="https://github.com/user-attachments/assets/8a5563a0-49fb-4faa-999e-0c1b1541eb09" />

## Expected behavior
Toolbar should be sticky under the device's status bar.
Refer to the behavior on Mobile Safari

Issue #3652, opened 2026-01-18
[Android] Find a way to reliably stick an element to the mobile keyboard


fbmcipher:
## Background

As you can see from the mockup below, the Formatting Bar is designed to stay positioned **directly above** the keyboard at all times.

This should even extend to the keyboard's open/close animation: as the keyboard closes, the Formatting Bar should slide down along with it.

It should also stay stuck to the keyboard even as the user scrolls.

<img width="643" height="675" alt="Image" src="https://github.com/user-attachments/assets/4a093fb9-6812-47ce-b15c-2b524fdb89fe" />

This is non-trivial on iOS.

> **Note:** On iOS, Safari adds its own accessory bar to the keyboard, which prevents us from displaying our own in Safari. Instead, test this issue using Capacitor. There are instructions on how to use Capacitor in the README.md file.

## Task

**Develop a technique on Android that allows an element to stick to the bottom of the screen, or the keyboard if it's open.** 

As Android supports the [VirtualKeyboard API](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API), getting the keyboard's height should be more straightforward than on iOS. My recommendation would be to extend and generalize the `safariKeyboardStore` to work on Android – abstracting platform-specific differences behind a single interface.

Due to variance between different Android devices, a general solution to matching the animation of the element with the keyboard on Android would be needed. Maybe there's a Capacitor-related solution?

## Resources

Some helpful links and resources:

- [`react-bottom-fixed` on GitHub](https://github.com/almond-bongbong/react-bottom-fixed)
    - An open-source React component which solves a similar problem. While it tracks keyboard open/close pretty well, it doesn't handle scrolling. 
- [Development of Alert 1.5 on em](https://github.com/cybersemics/em/pull/2746#pullrequestreview-2569273109)
    - A discussion from about a year ago which resulted in the implementation of the current `safariKeyboardStore`. There's a lot of useful trial-and-error insight here worth reading through.
- https://ishadeed.com/article/virtual-keyboard-api/
    - Great introductory article about the VirtualKeyboard API, which might be useful for tracking keyboard changes on Android. Unfortunately, this API is unsupported on iOS.
- https://gist.github.com/ryanlintott/b57b975c8d2e3cad0353a6b6e9d49928
    - Swift library which does something similar – might be useful if targeting a Capacitor native solution

PR #3653, opened 2026-01-18
[iOS Capacitor] Find a way to reliably stick an element to the mobile keyboard

### LINKED ISSUE #3596: [iOS] Find a way to reliably stick an element to the mobile keyboard
## Background

As you can see from the mockup below, the Formatting Bar is designed to stay positioned **directly above** the keyboard at all times.

This should even extend to the keyboard's open/close animation: as the keyboard closes, the Formatting Bar should slide down along with it.

It should also stay stuck to the keyboard even as the user scrolls.

Image

This is non-trivial on iOS.

> **Note:** On iOS, Safari adds its own accessory bar to the keyboard, which prevents us from displaying our own in Safari. Instead, test this issue using Capacitor. There are instructions on how to use Capacitor in the README.md file.

## Task

**Develop a technique on iOS that allows an element to stick to the bottom of the screen, or the keyboard if it's open.** 

On iOS, the keyboard **overlays** the viewport, so we need to use absolute or fixed positioning to manually position an element above the keyboard. To position that element directly above the keyboard, we need to track the keyboard's height.

There's some existing code that tracks the keyboard's height in [safariKeyboardStore.ts](https://github.com/cybersemics/em/blob/main/src/stores/safariKeyboardStore.ts). You can use that store to observe when the keyboard is open/closed and animate the element's position accordingly.

Sliding the element up and down in sync with the keyboard will be more tricky. As far as I can see, there isn't a straightforward way to observe the keyboard's height as it changes.
- One solution could be to use the same animation curve and timing that iOS uses when it hides/shows the keyboard. I extracted the animation timings in [this gist](https://gist.github.com/fbmcipher/2012a1c0be77c09d3c39b200ee5d4fe4) to save you some time. 
- Failing that, maybe Capacitor provides some native hooks we can use.

You can use keyboard/input blur DOM events to determine when the keyboard is closing.

## Resources

Some helpful links and resources:

- [`react-bottom-fixed` on GitHub](https://github.com/almond-bongbong/react-bottom-fixed)
    - An open-source React component which solves a similar problem. While it tracks keyboard open/close pretty well, it doesn't handle scrolling. 
- [Development of Alert 1.5 on em](https://github.com/cybersemics/em/pull/2746#pullrequestreview-2569273109)
    - A discussion from about a year ago which resulted in the implementation of the current `safariKeyboardStore`. There's a lot of useful trial-and-error insight here worth reading through.
- https://ishadeed.com/article/virtual-keyboard-api/
    - Great introductory article about the VirtualKeyboard API, which might be useful for tracking keyboard changes on Android. Unfortunately, this API is unsupported on iOS.
- https://gist.github.com/ryanlintott/b57b975c8d2e3cad0353a6b6e9d49928
    - Swift library which does something similar – might be useful if targeting a Capacitor native solution

fbmcipher:
**Closes #3596.** Leverage Capacitor's access to system-level updates on the virtual keyboard to develop a more reliable solution to keyboard height tracking on iOS.

---

### Details

This PR:

- **Generalizes and refactors the existing `safariKeyboardStore`** 
    - `safariKeyboardStore` and `viewportStore` contained some complex logic responsible for estimating the keyboard's height and determining open/close state
    - **Separate logic & state**
        - Now, a new `virtualKeyboardStore` is **purely responsible for storing state** about the virtual keyboard (open state & height)
        - **"Handlers"** implement platform-specific virtual keyboard logic.
            - An `iOSSafariHandler` contains the same estimation-based logic as before, unchanged.

- **A new `iOSCapacitorHandler`** uses Capacitor's native `@capacitor/keyboard` library to get instant system-level updates on the keyboard's open/close state and height. 
 
- **Use an approximation of the easing curve iOS uses to animate the on-screen keyboard** to synchronize our animations with the system
    - This is important to give the impression that the Formatting Bar (and other elements) are moving in relation with the system keyboard. 
    - This curve is obfuscated on iOS, so it took some trial and error to find a curve that was as close as possible to iOS. 
    - It should be noted that the curve used in this PR is not 1:1 with iOS' – so if you take the video and do frame by frame analysis, you might be able to spot some differences. It might take a little more fine-tuning to get it just right.

- **This work also paves the way for Android-specific implementations in the future.**

### Videos

**This video shows alerts smoothly animating along with the on-screen keyboard.** 

https://github.com/user-attachments/assets/d5c6b964-1c7e-47cc-8a46-9b392c99d376

https://github.com/user-attachments/assets/88d91e79-0e83-4660-bb5a-85a2d43815e2

**In this video, I experimented the work in this PR to illustrate how the Formatting Bar could animate with the keyboard.**

https://github.com/user-attachments/assets/5b9ea80a-3e22-4ddc-800d-990f988278ee

### Notes

- The layout shift seen in the videos above does not seem to be a regression – rather, an exacerbation of issue #3555. It seems to happen whenever `usePositionFixed()` is used.
    - **It is not seen on Safari** – only in Capacitor.
    - This definitely needs fixing, but my sense is that this should be done in a separate PR – what do you think?

- After briefly evaluating pure-React solutions like `react-bottom-fixed`, I unfortunately found their performance to be inadequate on iOS.

- `viewportStore` seems to do its own measurement of the virtual keyboard's height using the `VisualViewport` API. Perhaps there is an opportunity to unify this logic with the platform-specific "handler" system developed here. I didn't tackle that in this PR, to keep this PR as modular/low-impact/pure as possible.

- We could likely use the exact same easing curve to improve the animation/sense of immersion on iOS Safari, but since there are other problems with the iOS Safari implementation, I haven't attempted that in this PR.


src/device/virtual-keyboard/handlers/iOSCapacitorHandler.ts
```diff
@@ -0,0 +1,64 @@
...
+    Keyboard.addListener('keyboardWillShow', info => {
+      const height = info.keyboardHeight || 0
+      viewportStore.update({ virtualKeyboardHeight: height })
+      virtualKeyboardStore.update({ open: true, source: 'ios-capacitor' })
+
+      // Start storing animated height values in virtualKeyboardStore.
+      animatedHeight.start({
+        to: height,
+        onChange: value => {
+          virtualKeyboardStore.update({ height: value as unknown as number })
```
raineorshine: Since `VirtualKeyboardState` types `height` as number, we want to adhere to the type system and make sure that only numbers go in. The validation can be moved up into this module.

src/stores/virtualKeyboardStore.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/** A store that tracks the state of the virtual keyboard. Its value is updated by platform-specific Handlers. */
+const virtualKeyboardStore = {
+  ...store,
+  /** Updates the store with validation to prevent flickering or disappearing values. */
+  update: (updates: Partial<VirtualKeyboardState>) => {
+    const validatedUpdates: Partial<VirtualKeyboardState> = { ...updates }
+    // Prevent NaN or null heights from being stored
+    if (updates.height !== undefined && (updates.height === null || isNaN(updates.height))) {
+      console.warn('[VirtualKeyboardStore] Invalid height attempted:', updates.height)
```
raineorshine: What would cause a `NaN` or `null` height?

Let's avoid `console.warn`. Either the error can be handled, in which case no logging is needed, or it can't be handled, in which a hard error should be thrown or handled higher up.

src/util/initEvents.ts
```diff
@@ -415,11 +405,6 @@ const initEvents = (store: Store<State, any>) => {
     window.removeEventListener('drop', drop)
     lifecycle.removeEventListener('statechange', onStateChange)
     resizeHost.removeEventListener('resize', updateSize)
-
-    // Remove Visual Viewport event listener
-    if (isTouch && isAndroidWebView() && window.visualViewport) {
-      window.visualViewport.removeEventListener('resize', handleKeyboardVisibility)
-    }
```
raineorshine, line 433: Would you like to hook up the `destroy` handler here?

raineorshine (changes requested):
Thanks! Overall looks good. I'm glad that `@capacitor/keyboard` proved useful.

> * It should be noted that the curve used in this PR is not 1:1 with iOS' – so if you take the video and do frame by frame analysis, you might be able to spot some differences. It might take a little more fine-tuning to get it just right.

That sounds fine. We can fine-tune it later.

> The layout shift seen in the videos above does not seem to be a regression – rather, an exacerbation of issue [[iOS Capacitor][Android] Toolbar is overlapping with device's status bar #3555](https://github.com/cybersemics/em/issues/3555). It seems to happen whenever `usePositionFixed()` is used.
> 
> * **It is not seen on Safari** – only in Capacitor.
> * This definitely needs fixing, but my sense is that this should be done in a separate PR – what do you think?

Sure, I'm okay with that as long as you don't think it is a regression. Once we merge this PR it will be difficult to separate the new behavior from the old.

> * `viewportStore` seems to do its own measurement of the virtual keyboard's height using the `VisualViewport` API. Perhaps there is an opportunity to unify this logic with the platform-specific "handler" system developed here. I didn't tackle that in this PR, to keep this PR as modular/low-impact/pure as possible.

Yes, that might be possible. It will take some closer review of the existing code since different components depend on the height of the viewport with or without the virtual keyboard. Though it was recently simplified in https://github.com/cybersemics/em/commit/6a29dc16aa5c2614f873d4feb34634f2cc810118, which might make this easier.

---

Let me know when the CI passes and this is ready for manual testing. Thanks!

There's also a small merge conflict due to some refactoring I did today.

src/device/virtual-keyboard/handlers/iOSCapacitorHandler.ts
fbmcipher: This admittedly clunky workaround was implemented to work around a mistyped callback function in the `react-spring` library we were using to match the iOS keyboard animation. This bug has unfortunately been unfixed for a few years. https://github.com/pmndrs/react-spring/issues/2183

Fortunately, I found that the `motion` package (which we use for animation elsewhere in Liminal UI) *also* supports spring animations and is properly typed – so I've refactored the animation logic to use `motion` instead.

Fixes the typing issue, and also lets us remove an extra dependency. 7fb19af090ecc4907edf74d090262017733c13c4

src/stores/virtualKeyboardStore.ts
fbmcipher: Written while I was implementing `react-spring`, but doesn't seem to be necessary, especially with the move to `motion`. Removed in 7fb19af.

src/util/initEvents.ts
fbmcipher: Yes, I think we should. Thank you! d525fe9 991c2b5

src/device/virtual-keyboard/handlers/iOSCapacitorHandler.ts
raineorshine: That's a win!

src/components/PopupBase.tsx
```diff
@@ -54,17 +55,21 @@ const PopupBase = React.forwardRef<HTMLDivElement, PopupBaseProps>(
...
       height,
     })
+    const scrollTop = useScrollTop({ disabled: positionFixedStyles?.position === 'fixed' })
+
+    // measure the height of the popup after it has been rendered
+    useLayoutEffect(() => {
+      if (innerRef.current && anchorFromBottom) {
+        setHeight(innerRef.current.getBoundingClientRect().height)
+      }
+    }, [anchorFromBottom, children, scrollTop]) // measure on mount and whenever children (content) or scrollTop change (which affects absolute positioning)
```
raineorshine: `scrollTop` is unthrottled, so this will call `getBoundingClientRect()` and recalculate layout **on every tick** of the scroll. Scroll handlers are known to have a big impact on performance, so this probably isn't going to work. 

`useLayoutEffect` is blocking, so this will also re-render the component and all selectors on every tick of the scroll.

src/stores/virtualKeyboardStore.ts
```diff
@@ -0,0 +1,18 @@
...
+  height: 0,
+})
+
+/** A store that tracks the state of the virtual keyboard. Its value is updated by platform-specific Handlers. */
+const virtualKeyboardStore = {
+  ...store,
+  update: (updates: Partial<VirtualKeyboardState>) => {
+    store.update(updates)
+  },
+}
```
raineorshine: Looks like the store wrapper may no longer be needed.

raineorshine (changes requested):

src/components/PopupBase.tsx
```diff
@@ -54,17 +54,20 @@ const PopupBase = React.forwardRef<HTMLDivElement, PopupBaseProps>(
...
       offset: anchorOffset,
       height,
     })
+
+    // measure the height of the popup after it has been rendered
+    useLayoutEffect(() => {
+      if (innerRef.current && anchorFromBottom) {
+        setHeight(innerRef.current.getBoundingClientRect().height)
+      }
+    }, [anchorFromBottom, children]) // measure on mount and whenever children (content) changes
```
raineorshine: Can you confirm that `children` is a stable ref? I don't have experience passing it as a dependency to a hook.

raineorshine:
Sorry, looks like a small browserslist update caused merge conflict in the lock file.

BayuAri (commented):
Alert is always prompted above the keyboard when available and alert moves to the bottom accordingly when keyboard is dismissed on iOS Capacitor.

BayuAri:
## Issue A: The is a chance that the alert is overlapping with the caret and the text being written by user.
## Step to reproduce
1. Create a thought
2. Add a long text that will reach the center of the device's screen
3. Create a new thought by gesture
4. Verify that New Thought alert is prompted
5. Type text 


## Current behavior
iOS 26.2: 
When text meets the alert, the text written text is hidden behind the alert.

https://github.com/user-attachments/assets/f8d80ceb-91e8-4776-bfc5-63dbb60ce458


iOS 18.6.2.
Alert and caret are overlapping with written text

image

https://github.com/user-attachments/assets/841410f5-042d-427b-a4b7-1f92d70a0d7f

## Expected behavior
Focus and alert should not be overlapping. 
Alert should be displayed under the focus (caret)

raineorshine:
@BayuAri Can you additionally test the following steps on the iOS Capacitor build on this PR and `main`? I'm going to use the toolbar as it demonstrates the same position fixed properties but does not depend on the height of the virtual keyboard. If the toolbar doesn't jitter on scroll, then we know the alert can be built to not jitter on scroll.

FYI This is for the iOS Capacitor build only, as we know it can't be solved on Mobile Safari.

# Scroll jitter

## Steps to Reproduce

1. Create a thought.
2. Keep the virtual keyboard open.
3. Scroll.

## Current Behavior

The toolbar jitters during scrolling.

## Expected Behavior

The toolbar should be fixed to the viewport and not move at all during scrolling.

BayuAri:
@raineorshine 
The toolbar jitters on this PR and also `main` when keyboard is on screen

raineorshine:
Hold until #3758 is solved.

src/components/PopupBase.tsx
fbmcipher: After some research I don't think that `children` is referentially stable. React creates new JSX elements on every render, so it will change each time the parent re-renders, even if the content is identical.

It's safe to remove – at this time there aren't any elements that use `usePositionFixed` that change height after their initial render, so we don't need to update `height` after the first time we set it.

Removed in ebb63170c8ae0b62c127e7a47303356d7bd00c73.

src/components/PopupBase.tsx
fbmcipher: Thanks, you're totally right. Not sure why we were re-measuring the height of this element on each scroll tick... safely removed with no effect on functionality.

package.json
```diff
@@ -117,6 +117,7 @@
     "lottie-react": "^2.4.1",
     "marked": "^17.0.5",
     "moize": "^6.1.7",
+    "motion": "^12.28.1",
```
raineorshine: Do we need both `motion` and `framer-motion`? I believe they refer to the same library.

src/device/virtual-keyboard/safeArea.ts
```diff
@@ -0,0 +1,12 @@
+/** Measures the safe-area-bottom inset in pixels. Returns 0 if unavailable. */
```
raineorshine: This seems like potentially a better way to get safe-area-bottom-inset without triggering a reflow. Would this work?

```css
:root {
  --safe-area-inset-bottom: env(safe-area-inset-bottom);
}
```

```js
const insetBottom = getComputedStyle(document.documentElement)
  .getPropertyValue('--safe-area-inset-bottom');

console.log(insetBottom); // e.g. "20px"
```

src/hooks/usePositionFixed.ts
```diff
@@ -19,22 +40,56 @@ const usePositionFixed = ({
...
+      //   scrollTop + innerHeight - virtualKeyboard.height
+      //
+      // We clamp this to document.body.scrollHeight so the element never extends past
+      // the document boundary (e.g. when the page is shorter than the viewport).
+      //
+      // Then subtract the element's own height and offset if provided by the caller, and subtract the
+      // safe-area-bottom inset so the element doesn't overlap the rounded-screen home indicator.
+      //
+      const visibleBottom = Math.min(document.body.scrollHeight, scrollTop + innerHeight - virtualKeyboard.height)
+      top = `calc(${visibleBottom - (height ?? 0) - offset}px - ${token('spacing.safeAreaBottom')})`
```
raineorshine: While we're here, let's change `${token('spacing.safeAreaBottom')}` to `env(safe-area-inset-bottom)` since we are in a `calc` expression. I think that was an oversight in the previous implementation.

src/device/virtual-keyboard/handlers/iOSCapacitorHandler.ts
```diff
@@ -0,0 +1,82 @@
...
+    let controls: AnimationPlaybackControls | null = null
+
+    Keyboard.addListener('keyboardWillShow', info => {
+      // Get the raw height of the keyboard from the event...
+      const rawHeight = info.keyboardHeight || 0
+
+      // ...then subtract the safe-area-bottom inset to get the height above the safe-area baseline.
+      // Because we always add a safe-area-bottom inset whenever we position elements, this normalized height
+      // is the value we actually need. Consider this an additional 'safe area inset' that applies only when the keyboard is open.
+      const height = rawHeight - measureSafeAreaBottom()
```
raineorshine: Let's call this `targetHeight` to parallel `iOSSafariHandler`.

src/device/virtual-keyboard/safeArea.ts
```diff
@@ -0,0 +1,12 @@
...
+  if (typeof document === 'undefined') return 0
+  const div = document.createElement('div')
+  div.style.cssText = 'position:fixed;bottom:0;height:env(safe-area-inset-bottom);visibility:hidden'
+  document.body.appendChild(div)
+  const value = div.getBoundingClientRect().height
+  document.body.removeChild(div)
+  return value
+}
+
+export default measureSafeAreaBottom
```
raineorshine: Please match the filename to the default export name. This makes it easier to quickly navigate to the source of a function by name.

src/@types/VirtualKeyboardState.ts
```diff
@@ -0,0 +1,9 @@
+/** The state of the virtual keyboard. */
+export default interface VirtualKeyboardState {
+  /** True if the virtual keyboard is open. */
+  open: boolean
+  /** The height of the virtual keyboard in pixels. */
+  height: number
+  /** The source of the virtual keyboard state. */
+  source?: 'ios-capacitor' | 'ios-safari' | 'android-capacitor' | 'android-chrome'
```
raineorshine: Is `source` needed? It doesn't look like it's being used.

This is load-time platform information that can be accessed through [browser.ts](https://github.com/cybersemics/em/blob/f13525a0dab62f083e7d40e18356494202f2de80/src/browser.ts) anyway.

raineorshine (changes requested):

src/device/virtual-keyboard/safeArea.ts
fbmcipher: Will do, sorry about that. Fixed in c4a7d6be142f39524dc8e9a0ec25948ca4bf7145

package.json
fbmcipher: You're right, we don't need both.

`motion` is the package's new name. It looks like `framer-motion` aliases to `motion`, so it looks like they are basically interchangeable right now.

That being said, since the package *has* formally changed name, I think we should standardize on using `motion`. Perhaps I'll make that change in a small follow-up PR. It will keep our dependencies up-to-date, and doing this in a separate PR avoids polluting the scope of this PR.

For now, I've removed the extraneous `motion` import and updated this submission to stick to using `framer-motion`. See 4de8ba8e6534949c92a9c7cc689bd62aa92aa591

src/@types/VirtualKeyboardState.ts
fbmcipher: It was useful during testing to make sure the right handler gets loaded for each platform, but it isn't used nor does it affect any functionality now. Removed in ba7c4563e89ea9e636107856d38e195136149983

src/device/virtual-keyboard/handlers/iOSCapacitorHandler.ts
fbmcipher: Done: 8cbe140dcca1b9351f8b5bf32822b613c21e6b6a

src/device/virtual-keyboard/safeArea.ts
fbmcipher: Nice, yes that works. I might have been too conservative about avoiding touching raw CSS and preferring to work within JS/Panda. Using this solution in 24561cac35de26334be71fd5c2497cbf685ba9a6

Though it relies on a CSS declaration outside of PandaCSS, it's the most performant way to solve the problem.

src/hooks/usePositionFixed.ts
fbmcipher: Thanks, that feels more efficient than going through JS. 5f6951d7e899ec0c0eae8757e02a00426a3ab1b2

src/index.css
```diff
@@ -3,3 +3,11 @@
   See: https://panda-css.com/docs/concepts/writing-styles
 */
 @layer reset, base, tokens, recipes, utilities;
+
+:root {
+  /* If needed, global CSS variables can be defined here as a last resort.
+     Prefer styling within PandaCSS using tokens instead. */
+
+  /* Declare CSS variables for safe-area insets, so they can be read in JS */
+  --safe-area-inset-bottom: env(safe-area-inset-bottom);
```
raineorshine: Global CSS variables go in the panda config: https://github.com/cybersemics/em/blob/6e99e806c5c0468c9497beb4eda0732d4e89c54b/panda.config.ts#L347

src/device/virtual-keyboard/measureSafeAreaBottom.ts
```diff
@@ -0,0 +1,10 @@
+/** Returns the safe-area-bottom inset as an integer by reading the CSS variable `--safe-area-inset-bottom`. */
+const measureSafeAreaBottom = (): number => {
```
raineorshine: Probably should call this `getSafeAreaBottom` since it is not measuring any more.

raineorshine (changes requested):

src/device/virtual-keyboard/measureSafeAreaBottom.ts
fbmcipher: Good point 3c8baa6be0b852b795a4c87516863cdcf10f0f4c

src/index.css
fbmcipher: Thanks! Done in 93449c2e7291a03d8f16477bb34faf2078242345

src/device/virtual-keyboard/safeArea.ts
raineorshine: `--safe-area-inset-bottom` is atypical because it is a browser-provided value that is only accessible through CSS. It's unlike other browser-provided values which are accessible through the DOM API, and it's unlike styles we set ourselves through PandaCSS.

BayuAri (commented):
List of issues:
- [x] The is a chance that the alert is overlapping with the caret and the text being written by user. >> **fixed**
- [x] Caret is hidden behind the Alert when overlapping, it is no longer blinking in front of the alert. >> **fixed**
- [x] [Android] Alert is prompted at the bottom of the screen behind the virtual keyboard

raineorshine (approved):

fbmcipher:
> @BayuAri [Android] Alert is prompted at the bottom of the screen behind the virtual keyboard

This is reproducible, but it's out of scope here because this issue only pertains to iOS. We have an issue that separately tracks behavior on Android (#3652), so I will move this issue over there instead.

fbmcipher:
@raineorshine I think this PR is ready to merge. I pushed two small improvements today:

- Sticky elements now continue to follow the keyboard if it resizes after open (like if the selected keyboard layout changes, or if the emoji panel is opened)
- I cleaned up the Capacitor handler to remove some redundant code.

Tested and working as you'd expect on both Mobile Safari and Capacitor.

https://github.com/user-attachments/assets/3bf0410d-69ba-402f-9e41-6cf838751f11

raineorshine (approved):
Thank you!

PR #3654, opened 2026-01-18
Replace the Sidebar's `SwipeableDrawer` with a more flexible drawer implementation

### LINKED ISSUE #3593: Replace `SwipeableDrawer` with a more flexible drawer implementation
> **Migrate the Sidebar to a new drawer implementation, which will provide a foundation for further work on the Sidebar.**

Image

MUI's `SwipeableDrawer`, which is currently used in the sidebar, has several limitations that will make further improvements to the Sidebar difficult without resorting to hacky workarounds.

**Let’s identify and migrate to a more flexible drawer implementation that will grow with our needs.**

By the end of this PR, the Sidebar should **behave identically to today**, but be backed by a more flexible drawer implementation.
## Features the new drawer should ideally support

**Here's a list of features the new drawer implementation will need to support.**

We shouldn't implement any new features in this PR, but the chosen component/library should be capable of them so we can proceed confidently knowing the solution can scale to the tasks ahead.

- **Actively maintained, with support for React 19**
	- This will help us avoid compatibility issues later down the line.
- **Gesture support**
	- The ability to swipe the Sidebar closed using a gesture, as with the current implementation.
- **Ability to programatically observe of drawer width**
	- We need to be able to **track the position of the drawer as it opens and closes from left to right** – whether the drawer is being open/closed **programatically**, or **gradually via the swipe gesture**.
- **Ability to pin the drawer open**
	- In "pinned" mode, the drawer transitions from a modal state which overlays on top of the document to a pinned state which resizes the rest of the document and sits along side it.
	- This either needs to be a feature of the drawer, or something we can easily implement *with* it.
- **Drawer can fully unmount when closed**
	- The current MUI drawer stays mounted in the DOM even when it's closed. For more memory and performance optimisation, it would be great to identify a library where we can choose to unmount the drawer when it's closed.

- **Ruled out:**
	- **Vaul**: Looks great, but unfortunately unmaintained.
	- **react-modal-sheet**: Works great in the Command Center, but only supports sliding from bottom to top, **not** left to right.

## Developing our own sidebar with `framer-motion`
In the absence of any libraries that support everything we want, we can implement our own, but this should be a last resort.

fbmcipher:
**Closes #3593.** Replaces MUI's `SwipeableDrawer` with a more modern drawer implementation (based on Radix UI `Dialog`), which provides the flexibility needed for future Liminal UI design implementation.

---

### Details

This PR is a **pure refactor** which replaces MUI's drawer with a drawer implementation using Radix UI's `Dialog` component and Framer Motion.

This provides us greater control of gestures and animations. Radix UI's component is lower-level, less-batteries-included than MUI's, so we have to write a little more code ourselves. Radix still provides enough of the good stuff though (accessibility, screen reader support.) Framer Motion gives us the animation and touch handling logic.

The result is a drawer implementation which meets **all of our criteria laid out in #3593**, looks and feels identical to MUI's, and yet still gives us enough control and freedom to do whatever we want with it.

Most importantly, we can now observe the drawer's x-position without resorting to performance-endangering hacks, which is necessary to implement the new Sidebar design.

Care has been taken to ensure the new sidebar interacts identically to the outgoing sidebar, on both mobile & desktop. 

### Notes

- **All sidebar-related tests continue to pass.**
- These tests **fail**, but they also fail in `main` - so this does not seem like a regression caused by this PR.
    - ` FAIL   puppeteer-e2e  src/e2e/puppeteer/__tests__/url.ts > collapsed thought with url child`
    - ` FAIL   puppeteer-e2e  src/e2e/puppeteer/__tests__/thought-wrap.ts > Font Size: 18 (default) > thoughts wrapping at 575px`
    - ` FAIL   puppeteer-e2e  src/e2e/puppeteer/__tests__/caret.ts > all platforms > caret should move to editable after closing the command palette, then executing a cursor down command`
- For now, the Sidebar **remains mounted in the DOM when closed.**
    - It was actually one of my specified requirements in #3593 to identify a library where the sidebar could be unloaded when closed. It's possible with Radix UI, but it just broke too much (especially tests) that it just wasn't worth it right now.


src/components/Sidebar.tsx
```diff
@@ -55,151 +49,220 @@ const SidebarLink = ({
...
+      dispatch([toggleSidebarActionCreator({ value })])
+    },
+    [dispatch],
+  )
+
+  /** Dynamically determine the width of the sidebar. */
+  const width = typeof window !== 'undefined' && window.innerWidth < 768 ? '90%' : '400px'
+
+  /** Get the width of the sidebar in pixels, which is used for progress-based animations. */
+  const widthPx = typeof window !== 'undefined' && window.innerWidth < 768 ? window.innerWidth * 0.9 : 400
```
raineorshine: Let's use [viewport.innerWidth](https://github.com/cybersemics/em/blob/17c0e2b99b9c5f3c37803ace9a1d88c19d2af633/src/stores/viewport.ts#L22) which decouples the component from the global `window` object and reacts to resize events.

src/components/Sidebar.tsx
```diff
@@ -55,151 +49,220 @@ const SidebarLink = ({
...
+    }
+    return () => {
+      document.body.style.overflow = ''
+    }
+  }, [showSidebar])
+
+  useEffect(() => {
+    /** Watch for esc key. Handling this manually makes sure the current selection in the editor is kept when esc is hit. */
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') {
```
raineorshine, line 106: You may need to short circuit this handler if `!showSidebar`, otherwise it may end up interfering with other components.

src/components/Sidebar.tsx
```diff
@@ -55,151 +49,220 @@ const SidebarLink = ({
...
+              animate={{ x: showSidebar ? 0 : -widthPx }}
+              transition={transition}
+              className={css({
+                position: 'fixed',
+                top: 'safeAreaTop',
+                left: 0,
+                bottom: 0,
+                width,
+                backgroundColor: 'sidebarBg',
+                zIndex: 'sidebar !important',
```
raineorshine: You might not need `!important` any more since that was specifically to override an MUI style.

src/components/Sidebar.tsx
fbmcipher: Thanks, you're right. I make the event listener conditional on the sidebar's open state with an effect in 2fe06b13a3402fe92cc3fac3abe419992625f20c.

src/components/Sidebar.tsx
fbmcipher: Yep, it was no longer needed. Removed in 6f3e7ba8597c9c086b90dae0aae8346119375cb3

src/components/Sidebar.tsx
fbmcipher: Fixed in 495683e – the sidebar's width now resizes with the window, as expected.

src/components/Sidebar.tsx
```diff
@@ -55,151 +50,228 @@ const SidebarLink = ({
...
-              width: '0px', // Remove scrollbar space
-              background: 'transparent', // Optional: just make scrollbar invisible
-              display: 'none',
-            },
+            position: 'fixed',
+            inset: 0,
+            zIndex: 'sidebar',
+            pointerEvents: 'none',
             userSelect: 'none',
-            // must be position:relative to ensure drop hovers are positioned correctly when sidebar is scrolled
```
raineorshine, line 157: This comment still seems relevant.

src/components/Sidebar.tsx
```diff
@@ -55,151 +50,228 @@ const SidebarLink = ({
...
-            position: 'relative',
-            padding: '0 1em',
           })}
-          data-scroll-at-edge
         >
-          <FadeTransition type='fast' in={showSidebar}>
-            <div
-              style={{
-                // match HamburgerMenu width + padding
```
raineorshine, line 166: This comment still seems relevant.

src/components/Sidebar.tsx
```diff
@@ -55,151 +50,228 @@ const SidebarLink = ({
...
-            if (!drawer) return
-            const translateX = parseInt(drawer.style.transform.slice(10))
-            // set isSwiping if translateX is not parseable (NaN) or explicitly set to 0
-            if (translateX) {
-              setIsSwiping(true)
-              dispatch(longPress({ value: LongPressState.Inactive }))
-            }
-          },
-          10,
-          // no need to check on the first touchmove trigger since translateX is probably not yet set
```
raineorshine, line 131: This comment still seems relevant.

src/components/Sidebar.tsx
```diff
@@ -55,151 +50,228 @@ const SidebarLink = ({
...
-        },
-      }}
-      anchor='left'
-      onOpen={() => toggleSidebar(true)}
-      onClose={() => toggleSidebar(false)}
-      open={showSidebar}
-    >
-      {/* <RecentlyEdited /> */}
-      <div
-        // We need to disable favorites drag-and-drop when the Sidebar is being slid close.
```
raineorshine, line 113: This comment still seems relevant, minus the MUI-related stuff.

src/components/Sidebar.tsx
fbmcipher: bf64eab71995bfdb65d55cf94cf4a5ec25fd62c8

src/components/Sidebar.tsx
fbmcipher: bf64eab71995bfdb65d55cf94cf4a5ec25fd62c8

src/components/Sidebar.tsx
fbmcipher: bf64eab71995bfdb65d55cf94cf4a5ec25fd62c8

src/components/Sidebar.tsx
fbmcipher: bf64eab71995bfdb65d55cf94cf4a5ec25fd62c8

Issue #3666, opened 2026-01-19
[iOS] Faux caret intermittently shows for previous cursor


ethan-james:
Tested on iOS Safari

## Steps to Reproduce

```
- A
  - B
    - C
```

1. Set the cursor on `C`. 
2. Set the caret at the start of `C`. 
3. Draw →↓→ to create a new subthought

## Current Behavior

Rarely, the faux caret shows on line `C` rather than the new line.

https://github.com/user-attachments/assets/18b2da69-e404-4615-9593-5eec1ace3b06

The bug appears only once at the very end.

## Expected Behavior

The faux caret should always appear on the new subthought.

PR #3667, opened 2026-01-19
Replace setTimeout with requestAnimationFrame in useFauxCaretCssVars

### LINKED ISSUE #3574: [iOS] Faux Caret appears at offset 0 after quickly deleting new subthought
## Steps to Reproduce

```
- A
  - B
    - C
```

1. Set the cursor on `C`. 
2. Set the caret at the end of `C`. 
3. Swipe →↓→ to activate New Subthought.
4. Before the indentation animation finishes, hit backspace on the virtual keyboard. 

## Current Behavior

The Faux Caret is rendered at offset 0, at the beginning of `C`.

https://github.com/user-attachments/assets/119d8fbf-455a-42fb-80f4-fd10316305f2 

## Expected Behavior

The Faux Caret should be rendered where the hidden caret actually is, as the end of `C`.

ethan-james:
Fixes #3574

Flipping the cursor between different indent levels causes [isStartOfElementNode](https://github.com/ethan-james/em/blob/f5874eda23a1228ed88dce41c475f6c836e903ce/src/device/selection.ts#L158) to intermittently return the previous selection rather than the new selection. Replacing `setTimeout` with [requestAnimationFrame](https://github.com/ethan-james/em/blob/f5874eda23a1228ed88dce41c475f6c836e903ce/src/hooks/useFauxCaretCssVars.ts#L43) seems to reliably fix this. It's counter-intuitive because usually `requestAnimationFrame` seems to run with less of a delay than `setTimeout`, but I did mess around and get it to show me `setTimeout` coming in first (following the repro steps for #3574).

Screenshot 2026-01-19 at 1 46 45 PM

The `3` and `4` are logging `path.length`, showing that it is running for `C` and `D` respectively.

I went back through the old faux caret PRs and tested this against as many test cases as I could find, and it seems solid. #3666 still affects this solution, and looks a bit worse because both the previous cursor and the new cursor show a faux caret simultaneously.

### DIFF
--- a/src/hooks/useFauxCaretCssVars.ts
+++ b/src/hooks/useFauxCaretCssVars.ts
@@ -1,10 +1,11 @@
-import { useEffect, useState } from 'react'
+import { useState } from 'react'
 import { useSelector } from 'react-redux'
 import FauxCaretType from '../@types/FauxCaretType'
 import Path from '../@types/Path'
 import { isSafari, isTouch } from '../browser'
 import { isEndOfElementNode, isStartOfElementNode } from '../device/selection'
 import editingValueStore from '../stores/editingValue'
+import useLayoutAnimationFrameEffect from './useLayoutAnimationFrameEffect'
 
 /** Returns CSS variables that will suppress faux carets at the start or end of thoughts or notes.
  * */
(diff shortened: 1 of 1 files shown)


raineorshine (commented):
> Replacing `setTimeout` with [requestAnimationFrame](https://github.com/ethan-james/em/blob/f5874eda23a1228ed88dce41c475f6c836e903ce/src/hooks/useFauxCaretCssVars.ts#L43) seems to reliably fix this. It's counter-intuitive because usually `requestAnimationFrame` seems to run with less of a delay than `setTimeout`, but I did mess around and get it to show me `setTimeout` coming in first (following the repro steps for #3574).

That makes sense to me. `setTimeout(..., 0)` will run fastest since it only waits for the next event loop, while `requestAnimationFrame` will wait an entire animation frame (e.g. 16.666ms at 60fps). (Interestingly enough, the only thing faster than `setTimeout(..., 0)` is `await Promise.resolve()`.)

Note that this is already in a `useEffect` which is async itself, so `useEffect` + `requestAnimationFrame` could even end up taking longer than one frame.

What about https://github.com/cybersemics/em/blob/6671cd3f203f8bf672dff1b64665f0ff09bb7948/src/hooks/useLayoutAnimationFrameEffect.ts instead? I think that might be a better fit.

> both the previous cursor and the new cursor show a faux caret simultaneously

If needed, we can add some global state that enforces one faux caret at a time. Though maybe it won't be necessary after #3666 is fixed.

raineorshine:
@BayuAri Please hold off on testing this until I approve the approach.

ethan-james:
> Note that this is already in a `useEffect` which is async itself, so `useEffect` + `requestAnimationFrame` could even end up taking longer than one frame.
> 
> What about https://github.com/cybersemics/em/blob/6671cd3f203f8bf672dff1b64665f0ff09bb7948/src/hooks/useLayoutAnimationFrameEffect.ts instead? I think that might be a better fit.

Yes, `useLayoutAnimationFrameEffect` seems to work perfectly and it simplifies the code. Thanks for pointing that out!

BayuAri (commented):
Confirmed fixed.
Caret is landed back onto the end of the thought

**Video:**


https://github.com/user-attachments/assets/f3eb248f-dc82-4539-b89b-bde35c1f1c3a

**Tested on** 
- Safari 
- Capacitor app. 

**Devices:**
- iPhone 12 PM iOS 18.6.2
- iPhone 16e iOS 18.6.2

PR #3680, opened 2026-01-21
fix: Home icon in navbar breadcrumb shift

### LINKED ISSUE #3649: [iOS] Home icon shifts when cursor moves to first thought
Tested on iOS Safari. Not reproducible in desktop Chrome or desktop Safari. 

## Steps to Reproduce

1. Create thought `A`. 
2. Move the cursor back and forth between `A` and null.

## Current Behavior

The home icon in the footer shifts down when the cursor moves to `A`. 

https://github.com/user-attachments/assets/22a97286-f488-4cdc-acfd-542b81073aa5

## Expected Behavior

The home icon should not move.

cindmichelle:
Fix #3649

----
~Revert `rem` units back to `em` for the ContextBreadcrumb component, to fix the Home icon in the Navbar breadcrumb shifting when there is no Thought vs. when a Thought is highlighted.~

~I noticed that the ContextBreadcrumb container in Thought and Navbar uses different font sizes, which makes it not feasible to use `rem` for this component. This was an oversight on my part, as I previously only compared the rem result for ContextBreadcrumb in Thought.~

## Summary

Fixes the clunky Home icon issue in the navbar breadcrumb caused by font-size changes after the rem conversion.

## Root Cause

ContextBreadcrumb actually has two font-size variants, but previously we only handled the variant used in `Thought`. When the same logic was applied to Navbar, it introduced a regression because the navbar originally used a different font size.

## Changes

- Made ContextBreadcrumb rem conversion logic more flexible
- Introduced two variants:
   - default → used in BulletPositioner and Thought
   - small → used in Navbar and ThoughtLink
- Applied scaling styles in ContextBreadcrumb's root via inline styles in order to make the spacing more dynamics
- Added an extra wrapper `<div>` to handle component-specific spacing, instead of applying spacing directly to ContextBreadcrumb via cssRaw. This changes are applied on BulletPositioner, Thought, ThoughtLink.

### DIFF
--- a/src/components/ContextBreadcrumbs.tsx
+++ b/src/components/ContextBreadcrumbs.tsx
@@ -4,7 +4,7 @@ import { shallowEqual, useSelector } from 'react-redux'
 import { TransitionGroup } from 'react-transition-group'
 import { css } from '../../styled-system/css'
 import { extendTapRecipe } from '../../styled-system/recipes'
-import { token } from '../../styled-system/tokens'
+import { ColorToken, token } from '../../styled-system/tokens'
 import { SystemStyleObject } from '../../styled-system/types'
 import Path from '../@types/Path'
 import ThoughtId from '../@types/ThoughtId'
@@ -179,7 +179,7 @@ const ContextBreadcrumbs = ({
   thoughtsLimit?: number
   linkCssRaw?: SystemStyleObject
   variant?: ContextBreadcrumbsVariant
-  color?: string
+  color?: ColorToken
 }) => {
   const [disabled, setDisabled] = React.useState(false)
   const simplePath = useSelector(state => simplifyPath(state, path), shallowEqual)
--- a/src/components/NavBar.tsx
+++ b/src/components/NavBar.tsx
@@ -1,7 +1,6 @@
 import { useRef } from 'react'
 import { shallowEqual, useSelector } from 'react-redux'
 import { css } from '../../styled-system/css'
-import { token } from '../../styled-system/tokens/index.mjs'
 import Path from '../@types/Path'
 import { isTouch } from '../browser'
 import { BASE_FONT_SIZE } from '../constants'
@@ -28,7 +27,7 @@ const CursorBreadcrumbs = ({ position }: { position: string }) => {
     >
       <ContextBreadcrumbs
         variant='small'
-        color={token('colors.breadcrumbs')}
+        color='breadcrumbs'
         linkCssRaw={css.raw({
           '&:hover': {
             color: 'fg',


raineorshine (changes requested):
Hi! The goal is to transition from em to rem, so this seems like backward progress to me. If rem is not feasible, what changes to the codebase need to be made to make it feasible? Maybe there is a programmatic condition that needs to be added so the font size can be adjusted to different contexts.

src/components/ThoughtLink.tsx
cindmichelle: @raineorshine I’m planning to leave this line as-is, since converting it to rem using the scaler makes the calculation more complex than needed

raineorshine:
Hi, thanks for the update.

> Hi @raineorshine, for this implementation I made a small adjustment to the scaler sizing naming.
> 
> Initially I agreed with using `small | medium | large`, but during implementation I decided to use the consuming component name as the naming for ContextBreadcrumb instead.

This goes against our principles of encapsulation unfortunately. Lower level components should definitely not be aware of how they are being consumed, much less bake that into their public interface.

> The reason is that there’s a slight difference between the font sizes used by ThoughtLink and the Navbar. The Navbar previously used a static `font-size: 14px`, while ThoughtLink uses 0.867em with a 16px container's font-size (16 × 0.867 = 13.872px).

Can we combine these into one? The difference seems insignificant.

src/components/ThoughtLink.tsx
raineorshine: `ThoughtLink` is only used by Favorites, Recently Edited, and Recently Deleted, which all appear in the Sidebar. Given how similar the context is and how similar their appearance must be, can we not find a `rem`-based solution?

`rem` is ancestor independent, while `em` is ancestor dependent. In general, converting `em` to `rem` will simply make ancestor dependencies explicit rather than implicit. As far as I know we do not rely heavily rely on relative ancestor sizing, but please let me know if I'm missing something!

src/components/ContextBreadcrumbs.tsx
```diff
@@ -206,13 +239,21 @@ const ContextBreadcrumbs = ({
   return (
     <div
       aria-label={hidden ? undefined : 'context-breadcrumbs'}
+      style={
+        {
+          '--breadcrumb-font-size': `${scaler * 0.867}rem`,
```
raineorshine: Why set a CSS variable instead of setting the styles in the style attribute directly?

src/components/ThoughtLink.tsx
cindmichelle: Here are the computed styles for this line before the rem conversion.
You can see that the computed font size is _13.782px_, which was coming from the **html root font size** (previously set to a static _16px_, regardless of user font size changes).

image

Converting this line to a rem-based calculation would introduce unnecessary complexity without adding practical benefit. The current em usage already preserves the intended effective value (~13.782px) in a simple and predictable way.

If we were to switch this to rem while retaining the same behaviour, we would need to introduce an explicit calculation such as:
```
scaler * 13.782 / userFontSize
```

with the scaler derived from `13.782px / 0.867rem`. (0.867rem coming from the original fontsize of ContextBreadcrumb's root div)

At that point, the dependency becomes more complex and harder to reason about than the existing em-based approach. Given that this value is intentionally not meant to scale independently, keeping it as em is the clearer and more maintainable choice

src/components/ContextBreadcrumbs.tsx
cindmichelle: This is mainly because PandaCSS cant handle dynamic values in its generated styles.

We could technically use inline styles to pass dynamic values per instance, but that introduces another problem. If a parent component provides custom `cssRaw` targeting the same CSSprop, those overrides won’t apply anymore, since inline styles always win.

That would break our ability to customise styles from the parent, which we still need. So i think by using a custom CSS variable that injected from inline styles and being consume by pandacss would be the suitable workaround that solves both issues.

src/components/ThoughtLink.tsx
raineorshine: Seems like the problem is that the original font size did not scale with user font size.

You're right that we want to avoid complex calculations, but the solution is to consider the underlying intention and simplify rather than avoid `rem`.

src/components/ContextBreadcrumbs.tsx
raineorshine: > We could technically use inline styles to pass dynamic values per instance, ...

Yes, that's what I had in mind.

> ...but that introduces another problem. If a parent component provides custom `cssRaw` targeting the same CSSprop, those overrides won’t apply anymore, since inline styles always win.

I am not in factor of the `cssRaw` pattern in general. The surface area of the API is way too open. Parents should not be able to customize the CSS of child components. We are aiming for meaningful, type-safe interfaces between components.

`cssRaw` was used in some places to facilitate the PandaCSS refactor, but it should be considered a temporary stopgap only.

> That would break our ability to customise styles from the parent, which we still need. So i think by using a custom CSS variable that injected from inline styles and being consume by pandacss would be the suitable workaround that solves both issues.

The CSS variables solve the problem, but it results in verbose syntax with an added level of indirection. I think we need to find a cleaner architecture and try to avoid overly broad style overriding from the parent.

src/components/ThoughtLink.tsx
cindmichelle: That makes sense, thanks for calling that out.

Just to confirm my understanding, the expectation is that the font-size in this context should respect user font scaling, and once that’s corrected, the related spacing can be simplified accordingly.

And for this PR, I'll kept this line as-is to minimise scope, but I’m happy to follow up with a refactor on the sidebar’s base font-size so we’re not compensating downstream.

raineorshine: Yes, that’s correct. It would be odd if the text did not scale with user font size as it would diverge from the rest of the UI. That seems to be an oversight in the original implementation. Fixing the design, simplifying the implementation, and converting to `rem` all work in our favor here. This would be a harder situation if they did not align and we had to consider tradeoffs.

Sure, a separate PR to fix that sounds fine. Thanks!

src/components/ContextBreadcrumbs.tsx
cindmichelle: Just to confirm my understanding again, once we finish turning the contextBreadcrumb placement into a variant (small | default) and remove the `cssraw` pattern from this component, the usage would look something like this:
```
<ContextBreadcrumb variant="small" style={{ marginTop: '10px'}} />
```
Does this align with what you had in mind? In your view, would passing a style prop just for marginTop still be considered overly broad style overriding?

If so, the alternative would be to introduce another variant (for example, smallWithMargin and smallWithoutMargin). Personally, that feels less ideal to me compared to passing margin via the variant usage as in the example above, but keen to hear your thoughts

raineorshine: These aren't ideal. `style` is way too wide an interface, and it's strange for the parent to be controlling the margin of its child by passing props. 

If the margin depends on the context, why is `ContextBreadcrumbs` handling it instead of the parent?

cindmichelle: Ahh, I think I’m starting to get it now.

Instead of introducing a custom style prop that can override too much, it feels more ideal to handle this at the parent level by adding a wrapper div in parent-level for the additional layout style.

I agree this makes a lot more sense than allowing the component to be controlled via a broad style prop.

raineorshine: Yes, if possible that would be preferable. Wrapper divs are a good solution when the margin/styles are on the "outside" of the element. Context-dependent styles on descendants are trickier to architect cleanly, but less common.

cindmichelle: Hi @raineorshine , after looking at this again , I noticed that wrapping the styles with a div doesn’t fully solve the cssRaw workaround.

The reason is that not all CSS styles can be inherited from a wrapper div to its child (margin, padding, vertical-align, etc). Also, previously the styles passed via the cssRaw prop were used to override existing styles on ContextBreadcrumb, so when putting the styles in a one-level above, it wont always override existing styles in child-level.

For margin and padding, which can’t be overridden when applied on the wrapper div, we could instead apply inverse spacing on the wrapper to counteract the child’s existing values (e.g, applying `margin: -7px` on the wrapper to offset an existing `margin: 10px` on the ContextBreadcrumb, in order to mimic `margin:3px` from cssraw).

However, for non-numerical styles (e.g. color, vertical-align, etc), we’d likely need to introduce dedicated props so they can be explicitly consumed by ContextBreadcrumb.

Are you happy with the proposed solution above?

raineorshine: I would suggest making the minimal change necessary to fix the position regression, and we can do more refactoring in a separate PR. So applying an inverse margin in a wrapper div and leaving the other styles would be fine.

Why `px` instead of `rem`?

cindmichelle: > I would suggest making the minimal change necessary to fix the position regression, and we can do more refactoring in a separate PR. So applying an inverse margin in a wrapper div and leaving the other styles would be fine.

Sounds good to me

> Why px instead of rem?
Ahh it was just an example

raineorshine:
Sounds great!

src/components/NavBar.tsx
```diff
@@ -22,23 +22,20 @@ const CursorBreadcrumbs = ({ position }: { position: string }) => {
   )
 
   return (
-    <ContextBreadcrumbs
-      cssRaw={css.raw({
-        width: '100%',
-        color: 'breadcrumbs',
-        paddingLeft: '15px',
-        fontSize: '14px',
-        verticalAlign: 'bottom',
```
cindmichelle, line 31: Removed this part because after inspecting it, the `vertical-align: bottom` has no effect here since the element is currently using `display: block` by default.

Based on the [documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/vertical-align), vertical-align only applies to inline, inline-block, and table-cell elements, so it wasn’t doing anything in this case.

PR #3703, opened 2026-01-22
Allow dragging of selected text on desktop within a single thought (no note dragging)

### LINKED ISSUE #3530: Text selector drag and drop is not working
## Step to reproduce
1. Create a thought with long title
2. Double tap on a word on the title
3. Verify that the word is wrapped in iOS text selector 
4. Tap and hold the text selector then move

## Current behavior
Tap and hold gesture activates em's Thought drag and move function.
Moving around the finger while tapping the screen will drag the whole thought to a new place/thought
Video:

https://github.com/user-attachments/assets/816d1468-f389-4377-99e7-46f907bffd77

## Expected behavior
When iOS text selector is active, tap and hold should activate the text drag and drop function from iOS, not the drag and drop thought function.
The selected word should be lifted off and can be dragged and dropped anywhere refer to below video.
I think this is related to #3483, I see this during testing

https://github.com/user-attachments/assets/4ea9f057-0da3-4668-812b-f5e0b194399b

ethan-james:
Fixes #3530 on desktop

Multiple people have reported this behavior as a bug in `react-dnd`, and someone suggested a workaround in app code: https://github.com/react-dnd/react-dnd/issues/3157#issuecomment-986654550.

Unfortunately, that wasn't quite enough to fix text selection, as `react-dnd-html5-backend` is also [blocking](https://github.com/react-dnd/react-dnd/blob/1de654eb466b9563fa50df070c244d6f3ab43ca7/packages/backend-html5/src/HTML5BackendImpl.ts#L700) drop events at the top level during the capture phase.

* `react-dnd-html5-backend` was patched to stop it from interfering with drag-and-drop events
* a [workaround](https://github.com/react-dnd/react-dnd/issues/3157) was added to `ContentEditable` to re-enable native `dragover` handling, where before it was being intercepted and block by `react-dnd-html5-backend`
* the workaround is limited to the active cursor thought in order to prevent dragging text between thought (too many style- or action-related edge cases)
* the workaround is disabled when a thought drag is in progress in order to preserve drag-and-drop functionality
* dragging highlighted text within a note should work with the same workaround, but the drop handlers in `react-contenteditable` are getting stale for some reason that I haven't diagnosed yet

I also discovered a small logic bug where

```
1. click on a note
2. click on its associated thought
```

was not setting `noteFocus` to false. This prevented me from disabling text-dragging between a note and a thought, so I added a dispatch [onBlur](https://github.com/ethan-james/em/blob/bd04749d988f975e0e7bef8d163cff4dffba8167/src/components/Note.tsx#L138) to correctly set `noteFocus`.

### DIFF
--- a/src/components/ContentEditable.tsx
+++ b/src/components/ContentEditable.tsx
@@ -5,15 +5,15 @@ interface ContentEditableProps extends React.HTMLProps<HTMLDivElement> {
   style?: React.CSSProperties
   html: string
   disabled?: boolean
+  hasFocus?: boolean
   innerRef?: React.RefObject<HTMLDivElement | null>
-  isCursor?: boolean
   onChange: (originalEvt: ContentEditableEvent) => void
 }
 
 /**
  * Content Editable Component.
  */
-const ContentEditable = React.memo(({ style, html, disabled, innerRef, isCursor, ...props }: ContentEditableProps) => {
+const ContentEditable = React.memo(({ style, html, disabled, hasFocus, innerRef, ...props }: ContentEditableProps) => {
   const newContentRef = useRef<HTMLDivElement>(null)
   const contentRef = innerRef || newContentRef
   const prevHtmlRef = useRef<string>(html)
@@ -95,7 +95,7 @@ const ContentEditable = React.memo(({ style, html, disabled, innerRef, isCursor,
       }}
       // Allow dragging a text selection within an editable (#3530)
       // https://github.com/react-dnd/react-dnd/issues/3157
-      onDragOver={disabled || !isCursor ? undefined : e => e.stopPropagation()}
+      onDragOver={disabled || !hasFocus ? undefined : e => e.stopPropagation()}
       onInput={handleInput}
       onKeyDown={(e: React.KeyboardEvent<HTMLDivElement>) => {
         if (props.onKeyDown) props.onKeyDown(e)
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -128,6 +128,7 @@ const Editable = ({
   const contentRef = editableRef || nullRef
   const isCursor = useSelector(state => equalPath(path, state.cursor))
   const editingOrOnCursor = useSelector(state => isCursor || state.isKeyboardOpen)
+  const hasFocus = useSelector(state => isCursor && !state.noteFocus)
 
   // Disable contenteditable during drag-and-drop, otherwise thought text will become selected on mobile Safari.
   // On desktop Chrome, disabled is used to allow dragover events to avoid disrupting drag-and-drop behavior.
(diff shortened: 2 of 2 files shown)


raineorshine:

It looks like the puppeteer tests are failing.

BayuAri:
Dragging text between Thought is disabled.
Draging text from Thought to Note is disabled

## Issue: Dragging from Note to Thought is enabled. 
## Step to reproduce
1. Create a thought with a long text
2. Add Note with long text
3. Select any word or words on Note
4. Drag it to the text on the thought above


## Current behavior
Selected text from Note is dropped onto the thought text along with the text size and formatting
image

**Video:**

https://github.com/user-attachments/assets/af79eaca-4de8-4438-87e2-e25e5bd12431



## Expected behavior
Dragging from Note to Thought should be disabled

src/components/Note.tsx
```diff
@@ -132,13 +132,16 @@ const Note = React.memo(
     /** Set editing to false onBlur, if keyboard is closed. */
     const onBlur = useCallback(
       (e: React.FocusEvent) => {
+        const relatedTarget = e.relatedTarget as Element
```
raineorshine, line 135: This cast is unsafe, as `relatedTarget` could be null. While there are guards in place below to prevent runtime errors, it would be better to type this correctly and ensure we have the type system backing us up.

src/components/ContentEditable.tsx
```diff
@@ -5,14 +5,15 @@ interface ContentEditableProps extends React.HTMLProps<HTMLDivElement> {
   style?: React.CSSProperties
   html: string
   disabled?: boolean
+  hasFocus?: boolean
```
raineorshine, line 8: From an encapsulation perspective, I think it would be better to narrow the semantics of this parameter so that `ContentEditable` does not need to know about dependencies elsewhere in the app. As far as `ContentEditable` is concerned, this is just a parameter to control whether the `dragover` bubbles up or not. So I would call it something like `stopDragOver` and leave its usage up to the parent.

If `ContentEditable` were trying to hide implementation details, then `hasFocus` would be appropriate. But there are no implementation details to hide, so better to call it what it is without making assumptions about how it will be used.

raineorshine (changes requested):

raineorshine:
Hi! I've gotten a little confused about what we're doing here. Could you help me by making some clarifications?

- Explain what behavioral changes on what platforms we can expect from this PR
- Update PR title to reflect changes
- Suggest ways to add test coverage

Thank you!

src/components/ContentEditable.tsx
ethan-james: OK, I'm going to make the changes on #3706 because this impacts both desktop and mobile, but feel free to keep reviewing this one if it's easier.

PR #3706, opened 2026-01-23
Allow dragging of selected text on mobile

ethan-james:
Fixes https://github.com/cybersemics/em/issues/3530 on mobile

The `react-dnd-touch-backend` doesn't interfere with dragging text in the same way that `react-dnd-html5-backend` does, but drag-and-drop and long press behavior is available for editables, so it needs to be disabled while text is selected. I added a `selectionRangeStore` that will remove that functionality when `onselectionchange` is triggered with a selection range (`!selection.isCollapsed()`).

### DIFF
--- a/src/components/ContentEditable.tsx
+++ b/src/components/ContentEditable.tsx
@@ -8,6 +8,9 @@ interface ContentEditableProps extends React.HTMLProps<HTMLDivElement> {
   disabled?: boolean
   innerRef?: React.RefObject<HTMLDivElement | null>
   onChange: (originalEvt: ContentEditableEvent) => void
+  /** Stops the dragover event from propagating to ancestors on desktop. By default, dragover events are allowed and bubble up to ancestors where
+   * they may affect other drag-and-drop behavior (e.g. react-dnd). Stopping dragover events may be useful to allow native browser behavior such
+   * as text selection drag-and-drop if react-dnd would otherwise interfere. Setting this to true will also prevent the default drop events on mobile. */
   stopDragOver?: boolean
 }
 
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -128,6 +128,8 @@ const Editable = ({
   const contentRef = editableRef || nullRef
   const isCursor = useSelector(state => equalPath(path, state.cursor))
   const editingOrOnCursor = useSelector(state => isCursor || state.isKeyboardOpen)
+  // Stop dragover events from propagating up on non-cursor thoughts or notes, otherwise text selection drag-and-drop will be canceled by
+  // react-dnd on desktop.
   const stopDragOver = useSelector(state => !isCursor || state.noteFocus)
 
   // Disable contenteditable during drag-and-drop, otherwise thought text will become selected on mobile Safari.
--- a/src/components/Note.tsx
+++ b/src/components/Note.tsx
@@ -196,6 +196,9 @@ const Note = React.memo(
           // For some reason, pointerEvents: 'none' on ContentEditable or its parent does prevent onFocus.
           // This is strange, as it seems to prevent onFocus in Subthought.tsx.
           disabled={disabled}
+          // Prevent drag-and-drop of text selection between thoughts and notes. This also disables dragging
+          // text within the note, which was previously possible on mobile but not desktop. This may be addressed
+          // on both platforms by https://github.com/cybersemics/em/issues/3739.
           onDrop={isTouch ? (e: React.DragEvent) => e.preventDefault() : undefined}
           onKeyDown={onKeyDown}
           onChange={onChange}
--- a/src/components/Thought.tsx
+++ b/src/components/Thought.tsx
@@ -607,6 +607,9 @@ const ThoughtContainer = ({
           <StaticThought
             allowSingleContext={allowSingleContext}
             dragSource={dragSourceEditable}
+            // Disallow long press on mobile when there is a selection range, otherwise it will interfere with native selection controls like
+            // adjusting the selection or selection drag-and-drop. This is part of matching logic across StaticThought, Thought, and
+            // useDragAndDropThought that disables long press when there is a selection range.
             longPressProps={isTouch && !hasSelectionRange ? dragHoldResult.props : undefined}
             env={env}
             isContextPending={isContextPending}
--- a/src/stores/selectionRangeStore.ts
+++ b/src/stores/selectionRangeStore.ts
@@ -1,7 +1,10 @@
 import _ from 'lodash'
 import reactMinistore from './react-ministore'
 
-/** A reactive store that can block long press when a non-collapsed selection range is active. */
+/** A reactive store that can block long press when a non-collapsed selection range is active.
+ * This will always be false on desktop, or if the selection range is active but collapsed.
+ * It is invoked in a selectionchange event handler, and throttled by SELECTION_CHANGE_THROTTLE.
+ */
 const selectionRangeStore = reactMinistore<boolean>(false)
 
 export default selectionRangeStore


src/components/ContentEditable.tsx
```diff
@@ -51,12 +51,27 @@ const ContentEditable = React.memo(({ style, html, disabled, innerRef, ...props
...
-  const handleInput = (originalEvent: React.SyntheticEvent<HTMLInputElement>) => {
-    const innerHTML = contentRef!.current!.innerHTML
+  const handleInput = async (originalEvent: React.SyntheticEvent<HTMLInputElement>) => {
+    let innerHTML = contentRef!.current!.innerHTML
+    const nativeEvent = originalEvent.nativeEvent as InputEvent
 
     // prevent innerHTML update when editing
     allowInnerHTMLChange.current = false
 
+    if (nativeEvent?.dataTransfer?.items) {
```
raineorshine: Can you confirm that `nativeEvent?.dataTransfer?.items` is only defined when dragging selected text? We don't want to accidentally include other cases.

src/components/ContentEditable.tsx
```diff
@@ -51,12 +51,27 @@ const ContentEditable = React.memo(({ style, html, disabled, innerRef, ...props
...
+      const htmlItem = Array.from(nativeEvent.dataTransfer.items).find(item => item.type === 'text/html')
+      const plainItem = Array.from(nativeEvent.dataTransfer.items).find(item => item.type === 'text/plain')
+
+      const htmlString = htmlItem ? ((await new Promise(resolve => htmlItem.getAsString(resolve))) as string) : null
+      const plainString = plainItem ? ((await new Promise(resolve => plainItem.getAsString(resolve))) as string) : null
+
+      if (htmlString && plainString && htmlString !== plainString) {
+        const htmlRegex = new RegExp(htmlString.replace(/style="[^"]*"/, 'style="[^"]*"'))
+        innerHTML = innerHTML.replace(htmlRegex, plainString)
```
raineorshine: Can you uniquely identify the styles that need replacing? The thought could have many other valid formatting properties applied through a style attribute, such as bold, italic, color, etc.

raineorshine: Also, we have some fairly complex HTML parsing logic in the import pipeline (https://github.com/cybersemics/em/blob/feb215fcc7d8f90fb75e98daf057deb21d3fde28/src/actions/importData.ts and https://github.com/cybersemics/em/blob/feb215fcc7d8f90fb75e98daf057deb21d3fde28/src/actions/importData.ts). I would be a bit worried about the distribution of HTML parsing responsibility being too spread out if we add this logic here. The `ContentEditable` component is very low level and arguably should not have to know about its contents.

src/components/ContentEditable.tsx
ethan-james: That's why I added the `htmlString !== plainString` condition, to ignore non-actionable transfers. It could use some more thorough testing, but it got me to this point.

src/components/ContentEditable.tsx
ethan-james: I'll check on uniquely-identifying them. The issue was that the `htmlString` contained many more styles than the span that the browser had "pasted" into `innerHTML`. I'll also take a look at the import pipeline to see if I can possibly move this into there.

src/components/Note.tsx
```diff
@@ -192,6 +195,7 @@ const Note = React.memo(
           // For some reason, pointerEvents: 'none' on ContentEditable or its parent does prevent onFocus.
           // This is strange, as it seems to prevent onFocus in Subthought.tsx.
           disabled={disabled}
+          onDrop={isTouch ? (e: React.DragEvent) => e.preventDefault() : undefined}
```
ethan-james, line 202: Before this, it actually was possible to drag text within a note, in a way that was not possible on desktop. However, it was also possible to drag from a thought into a note, so I added this and now there is parity across desktop and mobile. In theory, it should be pretty easy to allow dragging within a note on mobile using `hasFocus`. In practice, I'm running into some weird stale event handler shenanigans.

raineorshine:
Thanks for reporting these issues. Since we are enabling selection-drag for the first time, some of these usability issues should probably be logged as separate issues if/when this PR gets merged. They are not regressions, strictly speaking, since the selection-drag behavior doesn't exist at all on `main` currently. Each of these issues will need to be evaluated individually in terms of design, usability, expected native behavior, differences between platforms, risk, and engineering effort. I think that's a bit too much to handle in this one PR.

@ethan-james @BayuAri Please work together to differentiate regressions, break out individual issues, and tighten the scope so that this body of work is more approachable. Thank you!

src/components/Note.tsx
```diff
@@ -132,13 +132,17 @@ const Note = React.memo(
     /** Set editing to false onBlur, if keyboard is closed. */
     const onBlur = useCallback(
       (e: React.FocusEvent) => {
+        if (!(e.relatedTarget instanceof Element)) return
```
ethan-james, line 135: > This cast is unsafe, as `relatedTarget` could be null. While there are guards in place below to prevent runtime errors, it would be better to type this correctly and ensure we have the type system backing us up.

@raineorshine Does this do the job?

BayuAri:
Separating issues as reference:

##  Issue H: Text selector tool can not be dragged after cancelling a drag
## Step to reproduce
1. Create a thought
2. Add a long text on it
3. Select any text with >6 characters(letter) word
4. Tap and hold
5. Verify that animation is shown; the word is floating
6. Lift off the finger from the screen
7. Verify that the blue  text selector reappears
8. Drag the bule circle to add more selection

## Current behavior
Nothing happens.
The text selector is not responding to user's dragging on the blue dot.
Note: this also happens after reproducing Issue H 

**Video:**

https://github.com/user-attachments/assets/33cd723d-5357-4bb9-a112-dba3f3d257f0


## Expected behavior
The blue text selector should be draggable to add more selection

raineorshine:
Lastly, if we do start addressing all the edge cases related to this behavior, we would benefit from automated test coverage to prevent regressions. I'm thinking that a combination of Browserstack test for iOS and Puppeteer tests for desktop will work. What do you think?

(It would be nice to use mobile emulation in Puppeteer for mobile testing but it won't emulate the Safari specific platform behavior unfortunately.)

**Update:** As I re-read this, I think the selection drag-and-drop feature is probably too low priority to build out test coverage now. Probably will merge this, hope for the best, and tackle test coverage if/when we get around to the related issues.

ethan-james:
> > > Thanks for reporting these issues. Since we are enabling selection-drag for the first time, some of these usability issues should probably be logged as separate issues if/when this PR gets merged. They are not regressions, strictly speaking, since the selection-drag behavior doesn't exist at all on `main` currently. Each of these issues will need to be evaluated individually in terms of design, usability, expected native behavior, differences between platforms, risk, and engineering effort. I think that's a bit too much to handle in this one PR.
> > > @ethan-james @BayuAri Please work together to differentiate regressions, break out individual issues, and tighten the scope so that this body of work is more approachable. Thank you!
> > 
> > 
> > Sounds good! I can create new issues and leave a brief comment on what I think is involved, especially whether I think it's native iOS behavior or behavior that's specific to our editables.
> 
> Great, thanks!
> 
> If you could do the following, that would be great:
> 
> 1. If a newly created issue is reporting behavior that is observable _after_ this PR gets merged (as opposed to current `main`), please add "Depends on [Allow dragging of selected text on mobile #3706](https://github.com/cybersemics/em/pull/3706)" to the top of the issue.
> 2. Update [Bayu's list of issues](https://github.com/cybersemics/em/pull/3706#issuecomment-3817587819) so it is in sync.
> 3. Recommended next steps

Looks like @BayuAri already updated the list of issues to point to the issues that I created. Thanks for that!

All of the new issues depend on #3706. The one notable deviation is #3739, because dragging within a note (and between notes) works pretty normally on `main`. Dragging within a note became disabled when I blocked dragging between thoughts and notes, as detailed in #3739. So the next step would be to review the complexity of the PR that you outlined in the next comment, and then figure out if we can get it over the finish line.

src/components/ContentEditable.tsx
```diff
@@ -1,104 +1,112 @@
...
 import { useSelector } from 'react-redux'
+import { isTouch } from '../browser'
 
 interface ContentEditableProps extends React.HTMLProps<HTMLDivElement> {
   style?: React.CSSProperties
   html: string
   disabled?: boolean
   innerRef?: React.RefObject<HTMLDivElement | null>
   onChange: (originalEvt: ContentEditableEvent) => void
+  stopDragOver?: boolean
```
raineorshine, line 14: Let's add a JSDOC comment here that explains the purpose of `stopDragOver`. The idea is to provide a bit more context about its usefulness while keeping it general enough that it is not referencing the thought structure or other things that `ContentEditable` should not know about.

How about this?

*Stops the `dragover` event from propagating to ancestors. By default, `dragover` events are allowed and bubble up to ancestors where they may affect other drag-and-drop behavior (e.g. react-dnd). Stopping `dragover` events may be useful to allow native browser behavior such as selection drag-and-drop if react-dnd would otherwise interfere. Setting this to true will also prevent the default `drop` events.*

src/components/Editable.tsx
```diff
@@ -126,14 +126,14 @@ const Editable = ({
   const oldValueRef = useRef(value)
   const nullRef = useRef<HTMLInputElement>(null)
   const contentRef = editableRef || nullRef
-  const editingOrOnCursor = useSelector(state => state.isKeyboardOpen || equalPath(path, state.cursor))
+  const isCursor = useSelector(state => equalPath(path, state.cursor))
+  const editingOrOnCursor = useSelector(state => isCursor || state.isKeyboardOpen)
+  const stopDragOver = useSelector(state => !isCursor || state.noteFocus)
```
raineorshine, line 133: Let's add a comment to `stopDragOver` to explain the reasoning behind it.

Maybe you can help me with the wording here, and any other implications that may be relevant:

*Stop dragover events from propagating up on non-cursor thoughts or notes, otherwise selection drag-and-drop will...*

src/components/Note.tsx
raineorshine: Let's add a comment that explains this in the code, as that will be easier to reference in the future than doing a git blame. 

*This line is intended to disable functionality in order to create parity between desktop and mobile. etc*

src/components/Thought.tsx
```diff
@@ -526,6 +527,8 @@ const ThoughtContainer = ({
     isTableCol1,
   })
 
+  const hasSelectionRange = selectionRangeStore.useSelector(value => value)
```
raineorshine: `selectionRangeStore.useState()`

src/components/Thought.tsx
```diff
@@ -604,7 +607,7 @@ const ThoughtContainer = ({
           <StaticThought
             allowSingleContext={allowSingleContext}
             dragSource={dragSourceEditable}
-            longPressProps={isTouch ? dragHoldResult.props : undefined}
+            longPressProps={isTouch && !hasSelectionRange ? dragHoldResult.props : undefined}
```
raineorshine, line 613: Let's add an explanatory comment here.

How about this?

*Disallow long press on mobile when there is a selection range, otherwise it will interfere with native selection controls like adjusting the selection or selection drag-and-drop. This is part of matching logic across StaticThought, Thought, and useDragAndDropThought that disables long press when there is a selection range.*

src/stores/selectionRangeStore.ts
```diff
@@ -0,0 +1,7 @@
+import _ from 'lodash'
+import reactMinistore from './react-ministore'
+
+/** A reactive store that can block long press when a non-collapsed selection range is active. */
```
raineorshine: While the update logic is technically external to the store definition, it is critical to usage so I would write the key points in the JSDOC to reflect when/how the store is updated.

- False with when there is no selection OR when there is a collapsed selection.
- Only operational on mobile. Always false on desktop.
- Throttled with SELECTION_CHANGE_THROTTLE.

raineorshine (changes requested):
Thanks. After reviewing all the changes once more, I think the complexity is acceptable, and I don't see a good way to centralize it. 

There is actually a reduction in complexity in Note.tsx since we can remove the special handling for Chrome. The rest of the logic is divided between long press disabling, dragover propagation, and drop disabling, which can't easily be contained in a single abstraction.

Given the diversity of places this touches, I think the best approach is to avoid awkward abstractions which trade one kind of complexity for another, and settle for thoroughly commenting the code. The PR is a bit undercommented anyway, but I'm going to give you some suggestions that go further in the comments to elucidate the behavior and hopefully make our lives easier in the future. I'm open to any feedback or additions you may have. Thanks!

PR #3746, opened 2026-01-30
Position FauxCaret after useLayoutAnimationFrameEffect

### LINKED ISSUE #3665: [iOS] Faux caret mis-positioned when toggling table view
Tested on iOS Safari

## Steps to Reproduce

```
- A
  - B
    - C
```

1. Set the cursor on `B`. 
2. Set the caret at the end of `B`. 
3. Tap table view toolbar button.

## Current Behavior

The faux caret appears in the wrong place.

https://github.com/user-attachments/assets/db730586-6b8f-4e9f-b912-4a52a398126e

## Expected Behavior

The faux caret should appear in the same place where the real caret will appear.

---

It seems like this issue only occurs on load. If you perform another action such as creating and deleting a new thought, then the faux caret appears in the correct place after following the steps above.

ethan-james:
Fixes #3665

Swapping `useEffect` + `setTimeout` out for `useLayoutAnimationFrameEffect` causes the faux caret to sweep into its proper place during the first part of the table-toggling animation. I consider this to be an upgrade, although there's definitely still room for improvement to see if I can get it to appear in the correct place for the entire duration of the animation.

https://github.com/user-attachments/assets/4589caf4-33a4-4129-a047-b329856760c5

### DIFF
--- a/src/components/FauxCaret.tsx
+++ b/src/components/FauxCaret.tsx
@@ -1,5 +1,5 @@
 import { head } from 'lodash'
-import { useEffect, useState } from 'react'
+import { useState } from 'react'
 import { useSelector } from 'react-redux'
 import { css } from '../../styled-system/css'
 import { Property } from '../../styled-system/types/csstype'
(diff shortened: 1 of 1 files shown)


raineorshine:
This looks pretty good to me, since it tracks the thought animation fairly closely.

ethan-james:
> This looks pretty good to me, since it tracks the thought animation fairly closely.

Nice! I was speculating that it might be a mismatch between tweening for different animations, but if you think it looks good then we can move forward.

raineorshine (approved):

PR #3757, opened 2026-02-03
Hide start/end faux caret on tap

### LINKED ISSUE #3666: [iOS] Faux caret intermittently shows for previous cursor

ethan-james:
Fixes #3666 

The original issue was fixed in #3746, but there were still some double-faux-caret issues. Since [fauxCaretType](https://github.com/ethan-james/em/blob/d1995e6d404412272fc381ff6bdd07ab992f8617/src/hooks/useFauxCaretCssVars.ts#L26) for start/end faux carets is controlled independently from [the positioned faux caret](https://github.com/ethan-james/em/blob/d1995e6d404412272fc381ff6bdd07ab992f8617/src/components/FauxCaret.tsx#L61), it's possible for both of them to be active at the same time.

The straightforward solution that I tried here is to add a `fauxCaretTapHandler` that will hide the start/end faux carets when a thought is tapped. The only time that I can foresee where a faux caret would change from start/end to positioned is after a tap. It would also be possible to try and create a centralized enum that permits only one discrete type of faux caret to exist at a time.

### DIFF
--- a/src/components/TreeNode.tsx
+++ b/src/components/TreeNode.tsx
@@ -70,7 +70,7 @@ const TreeNode = ({
   const fadeThoughtRef = useRef<HTMLDivElement>(null)
   const [isMounted, setIsMounted] = useState(true)
 
-  const { fauxCaretNodeProviderStyles, fauxCaretTapHandler } = useFauxCaretNodeProvider({
+  const { styles: fauxCaretNodeProviderStyles, hide: hideFauxCaret } = useFauxCaretNodeProvider({
     editing,
     fadeThoughtElement: fadeThoughtRef.current,
     isCursor,
@@ -156,7 +156,12 @@ const TreeNode = ({
           ...fauxCaretNodeProviderStyles,
         }}
       >
-        <div ref={fadeThoughtRef} onClick={fauxCaretTapHandler}>
+        <div
+          ref={fadeThoughtRef}
+          // It's possible for both the positioned faux caret and the start/end faux caret to be active at once, so hide the start/end
+          // caret after a tap activates the positioned faux caret. https://github.com/cybersemics/em/pull/3757
+          onClick={hideFauxCaret}
+        >
           <VirtualThought
             debugIndex={testFlags.simulateDrop ? indexChild : undefined}
             depth={depth}
--- a/src/hooks/useFauxCaretCssVars.ts
+++ b/src/hooks/useFauxCaretCssVars.ts
@@ -61,13 +61,13 @@ const useFauxCaretNodeProvider = ({
   )
 
   return {
-    fauxCaretNodeProviderStyles: {
+    styles: {
       '--faux-caret-line-start-opacity': fauxCaretType === 'thoughtStart' ? undefined : 0,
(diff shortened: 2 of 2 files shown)


src/hooks/useFauxCaretCssVars.ts
```diff
@@ -60,10 +61,13 @@ const useFauxCaretNodeProvider = ({
...
-    '--faux-caret-line-end-opacity': fauxCaretType === 'thoughtEnd' ? undefined : 0,
-    '--faux-caret-note-line-start-opacity': fauxCaretType === 'noteStart' ? undefined : 0,
-    '--faux-caret-note-line-end-opacity': fauxCaretType === 'noteEnd' ? undefined : 0,
+    fauxCaretNodeProviderStyles: {
+      '--faux-caret-line-start-opacity': fauxCaretType === 'thoughtStart' ? undefined : 0,
+      '--faux-caret-line-end-opacity': fauxCaretType === 'thoughtEnd' ? undefined : 0,
+      '--faux-caret-note-line-start-opacity': fauxCaretType === 'noteStart' ? undefined : 0,
+      '--faux-caret-note-line-end-opacity': fauxCaretType === 'noteEnd' ? undefined : 0,
+    },
+    fauxCaretTapHandler: isTouch && isSafari() ? () => setFauxCaretType('none') : undefined,
```
raineorshine: "Handler" is a name that only makes sense in the context of the `TreeNode`, where it used as the `onClick` handler. `useFauxCaretCssVars` doesn't know about event handlers or click events. It just exposes a way to hide the faux caret. Therefore, I would call this `hide`.

Similarly, `fauxCaretNodeProviderStyles` seems to namespace the variable name in anticipation of the caller. From the perspective of `useFauxCaretCssVars`, it's obvious that these styles are related to the faux caret, so that doesn't need to be part of the name. We can just call it `styles`. The caller can always rename it at declaration time if needed.

Lastly, I'm noticing that the handler is identical to the `editingValueStore` effect above. We could remove the duplication and define a single handler for both.

src/components/TreeNode.tsx
```diff
@@ -153,10 +153,10 @@ const TreeNode = ({
         isMounted={isMounted}
         style={{
           ...style,
-          ...fauxCaretNodeProvider,
+          ...fauxCaretNodeProviderStyles,
         }}
       >
-        <div ref={fadeThoughtRef}>
+        <div ref={fadeThoughtRef} onClick={fauxCaretTapHandler}>
```
raineorshine: As usual, we should add inline comments for any logic that is not obvious in order to help future developers (including our future selves ☺️).

Why do we need to hide the faux caret on click? What would happen otherwise? What issue/PR is relevant?

PR #3779, opened 2026-02-05
Fix HTML single-line regex to allow importText action

### LINKED ISSUE #3744: Copied text from Note become subthought when pasted onto a thought
## Step to reproduce
1.  Create a thought
2. Create a note
3. Double tap a word on the note text
4. Copy the selection
5. Tap text on the thought to place the caret
6. Tap the caret to trigger text tools
7.  Select Paste

## Current behavior
Copied text from Note is pasted as a new subthought. 
Bullet and the font color of the subthought are grey

Image

**Video:**

https://github.com/user-attachments/assets/90c77383-4103-488b-b2f5-a76916a03f4b

## Expected behavior
Pasted text should be a part of the thought text.
It should not create a new subthought.

ethan-james:
Fixes #3744 

[REGEX_HTML_SINGLE_LINE](https://github.com/ethan-james/em/blob/aa6866d934b35c95a83bbe128429b6f3b69248e8/src/actions/importData.ts#L22) wasn't matching anything, so the only way to invoke [importText](https://github.com/ethan-james/em/blob/aa6866d934b35c95a83bbe128429b6f3b69248e8/src/actions/importData.ts#L90) was by pasting as plain text. Now it will match single-line text pasted normally, although there are some side effects that I can document below.

### DIFF
--- a/src/device/selection.ts
+++ b/src/device/selection.ts
@@ -470,10 +470,12 @@ export const html = () => {
         const parentElement = node.parentElement
         const clonedElement = parentElement.cloneNode(true) as HTMLElement
         removeEmptyElementsRecursively(clonedElement!, range.startContainer.textContent!)
-        // iOS Safari inserts non-breaking spaces into text replacements which don't match regular spaces (#3779)
         containerHtml = clonedElement ? clonedElement.innerHTML : null
       }
     }
+
+    // iOS Safari converts non-breaking spaces into UTF-8 characters when accessing range textContent.
+    // Convert them back into HTML character entities to ensure that REGEX_HTML_SINGLE_LINE matches (#3779).
     return containerHtml?.replace(range.startContainer.textContent!.replace(/\u00A0/g, '&nbsp;'), selection.toString())
   }
 


ethan-james:
Pasting note text into a thought produces a similar style to what was happening with text drag-and-drop:

Screenshot 2026-02-05 152736

ethan-james:
Also, here's what happens when you paste markdown (without pasting as plain text) in the middle of the thought.

## Steps to reproduce

```
- This is the thought
```

1. Copy some markdown, such as
```
- a
  - b
    - c
```
2. Click in between `is` and `the`
3. Paste the markdown
4. Position the caret at the end of `- c`
5. Backspace

## Current behavior

The backspace key takes you out of the styled portion without deleting characters

https://github.com/user-attachments/assets/12b18c9e-6269-430e-8059-afd83f8c8e23

## Expected behavior

Text should be editable within the styled portion

---

The styles don't seem so bad compared to the equivalent behavior on `main`:

Screenshot 2026-02-05 155233

except for what happens when you try to hit backspace. Maybe that is a separate issue related to styled spans.

raineorshine:
Looks good to me.

That regex seems to be intended just for copying single lines from  pdfs, but since the change is so small and none of the tests fail I'm not going to spend the time to question it. If any of the existing tests start failing please pause and consult me before doing additional work, as there is a lot of complexity to imports. 

You may add a unit test at your convenience.

> Pasting note text into a thought produces a similar style to what was happening with text drag-and-drop:
> 
> 
> 
> Screenshot 2026-02-05 152736
> 
> 

Related to #2467 but probably needs its own issue. 

> Also, here's what happens when you paste markdown (without pasting as plain text) in the middle of the thought.

Bad either way, but too involved to address here. I believe there are some import tests that cover this that are being skipped currently due to known issues.

raineorshine:
## Solution Discussion

> > Related to #2467 but probably needs its own issue.
> 
> Cool, thanks, that is a good summation of the desired functionality. I can take #2467 if you think it's a good issue for me.

Let me think about that. It's a very big undertaking with a lot of risk.

> Knowing that the pasted text can contain a lot of markup and that can be delicate when considering the multitude of different sources for pasted text, I tested an approach based on [onCopy](https://github.com/ethan-james/em/blob/6ba5af3d4dbdd36b356e48370232c4e4b40a615e/src/components/Note.tsx#L198). For the specific case of copying text from a note and pasting it into a thought, this could be a good way to preserve styles within the note without bringing along the note's default styles.

As long as [text/em](https://github.com/cybersemics/em/blob/24a52944f6b1941a7ad71065efa47ac66dd13119/src/components/Editable/useOnCopy.ts#L27) is not set when copying from a note, then it will be treated as an external source and may be resolved automatically by the fix for #2467.

Otherwise, yes, I think overriding `onCopy` and `onCut` could work.

> Let me know if you think this is worth pursuing. Unfortunately it requires calling `preventDefault`, which would make it a little more difficult to override `onCut` since the highlighted text also needs to be manually removed from the editable.

That should be fine, because we override `onCut` in the Editable in https://github.com/cybersemics/em/blob/24a52944f6b1941a7ad71065efa47ac66dd13119/src/components/Editable/useOnCut.ts and manually remove the selected text with [selection.removeCurrentSelection()](https://github.com/cybersemics/em/blob/24a52944f6b1941a7ad71065efa47ac66dd13119/src/device/selection.ts#L429-L433)

## Testing

> I added a test for copying text from a note and pasting it into a thought. You mentioned a "unit test" and I thought about adding something to the non-Puppeteer test suite to test an `importData` action. My first attempt passed on `main`, though, and I decided that it didn't deal with the complexity of styled note text, so I switched to Puppeteer.

Yes, I think to test this realistically in the DOM we do need Puppeteer, because by the time it gets to `importData` we've lost the thought/note distinction. Artificially crafting the `importData` payload would not cover a regression in the UI.

> There were a couple of new pieces here:
> 
> 1. selecting a note rather than a thought editable
> 2. double-clicking
> 
> and so I put that logic directly into the test. I initially started by copying `waitForEditable` into `waitForNote`, but returning `ElementHandle | JSHandle` didn't work well for fetching the bounding box that I needed to double-click. I could refactor the double-click functionality and the `waitForNote` functionality if you would prefer, but I thought I'd put up the straightforward version before I starting anticipating future needs.

That part looks great to me. It's fairly low level, but it's user-oriented. You've decoupled the test from the DOM hierarchy and the implementation details of the Note which is nice.

> I also couldn't decide whether testing the text (`expect(thoughtValue).toBe('This is a Thisthought')`) would be better than a screenshot test, but I thought that since part of the issue involves the default styling of the note, a screenshot would be better.

This is subtle, but I would prefer to avoid a snapshot test here. Snapshot tests are best for covering visual regressions with the layout, positioning, or appearance of elements. While the pasted note is "visual" in a general sense, this is really a behavioral test where we want to test the thought value after copying and pasting a note. We want to test the HTML content of the thought after paste, not how it looks when is rendered.

The behavioral test will be faster and more targeted. A snapshot test is tightly coupled to everything that shows up in the snapshot, so it will break if other aspects of the UI change, such as thought spacing, font size, and color.

FYI This will probably entail updating the [exportThoughts](https://github.com/cybersemics/em/blob/feb215fcc7d8f90fb75e98daf057deb21d3fde28/src/e2e/puppeteer/helpers/exportThoughts.ts) helper to take an optional `type` parameter so that we can export as "text/html".

src/e2e/puppeteer/helpers/exportThoughts.ts
```diff
@@ -9,8 +10,12 @@ const em = window.em as WindowEm
  * Export the current state of thoughts as plain text.
  * This allows puppeteer tests to verify thought structure without using snapshots.
  */
-const exportThoughts = async (): Promise<string> => {
-  const exported = await page.evaluate(HOME_TOKEN => em.exportContext([HOME_TOKEN], 'text/plain'), HOME_TOKEN)
+const exportThoughts = async (mimeType: MimeType = 'text/plain'): Promise<string> => {
```
raineorshine: Would you mind making a small change? I prefer optional arguments to be specified in an options object rather than as positional arguments. The options object is self-documenting and easier to extend.

```ts
const exportThoughts = async ({ mimeType }: { mimeType?: MimeType = 'text/plain' } = {}): Promise<string> => {
```

src/e2e/puppeteer/__tests__/__image_snapshots__/render-thoughts/superscripts-paste-text-from-a-note-into-a-thought-1.png
raineorshine, line 1: Okay to delete snapshot

src/e2e/puppeteer/__tests__/render-thoughts.ts
```diff
@@ -245,4 +247,37 @@ describe('Superscripts', () => {
...
+          - This is a note
+    `
+
+    await paste(importText)
+
+    const note = await page.$('[aria-label=note-editable]')
+    const boundingBox = await note?.boundingBox()
+
+    if (boundingBox) {
+      // Calculate the edge of the note, inside the first word ('This')
```
raineorshine: I think a slight improvement would be to comment at the top of this section about the overall purpose:

```
// Double click inside the left edge to select the first word
```

Then the rest of the calculations make sense without having to explain them at a low level.

src/e2e/puppeteer/__tests__/render-thoughts.ts
```diff
@@ -245,4 +247,37 @@ describe('Superscripts', () => {
...
+        - =note
+          - This is a note
+    `
+
+    await paste(importText)
+
+    const note = await page.$('[aria-label=note-editable]')
+    const boundingBox = await note?.boundingBox()
+
+    if (boundingBox) {
```
raineorshine: Tests differ from application code in that we want them to fail hard if the test does not pass, so I would generally avoid truthy checks in tests. Instead, you can either:

- Throw an error when `boundingBox` is not defined. **← better error messaging**
- Tack on `!` to the end and let it fail at runtime. **← more terse**

BayuAri (commented):
Original issue is fixed on Mobile but not on Desktop.

Here is the list issue on this PR:
- [ ] Issue A: [Desktop] Copied text from Note is pasted as a new subthought
- [ ] Issue B: Copied a partial text from a text replacement is pasted as the whole text

ethan-james:
Thanks @BayuAri. I pushed a fix for Issue A, which was happening because Mac Chrome uses different markup for single-line HTML compared to Windows, where I had been testing. Issue B is happening because [selection.html()](https://github.com/ethan-james/em/blob/7f6bcfbabb28d16dfbf945e45f6d6989907a4814/src/device/selection.ts#L479) is returning the entire note text instead of just the selected portion. I think it's happening because there's a `&nbsp;` character in one string and a regular space in the other, and that's causing `containerHtml?.replace(range.startContainer.textContent!, selection.toString())` not to match on `range.startContainer.textContent`, but it's being very stubborn and I'll need to finish it up tomorrow.

BayuAri (commented):
Testing on  [3517fde](https://github.com/cybersemics/em/pull/3779/commits/3517fdeefec170be28714874c9ec0729f19b52ec)

- [x] Issue A: [Desktop] Copied text from Note is pasted as a new subthought  >> **Fixed**
- [x] Issue B: [Mobile] Copied a partial text from a text replacement is pasted as the whole text >> **Fixed**

src/device/selection.ts
```diff
@@ -470,7 +470,8 @@ export const html = () => {
         const parentElement = node.parentElement
         const clonedElement = parentElement.cloneNode(true) as HTMLElement
         removeEmptyElementsRecursively(clonedElement!, range.startContainer.textContent!)
-        containerHtml = clonedElement ? clonedElement.innerHTML : null
+        // iOS Safari inserts non-breaking spaces into text replacements which don't match regular spaces (#3779)
```
raineorshine: Converting character entities to UTF-8 characters seems out of scope for this function. Character entities are valid HTML. The question I ask is why `selection.html()` should be responsible for a shortcoming of `importData`.

ethan-james: I went back and double-checked the `setDescendant` action since we're dealing with notes here, and verified that the `&nbsp;` is being inserted into the note text in a way that seems kosher. The real problem seems to be that `clonedElement.innerHTML` contains the character entity (`&nbsp;`) while `range.startContainer.textContent` contains the UTF-8 character (`\u00A0`). The two of those are supposed to be referencing the same text in order to perform the replacement:

```
containerHtml?.replace(range.startContainer.textContent!, selection.toString())
```

although I suppose that it makes sense that `innerHTML` and `textContent` contain different representations of entities.

The conversion could also happen the other way, i.e.

```
range.startContainer.textContent!.replace(/\u00A0/g, '&nbsp;')
```

if that strikes you as any better.

raineorshine: Great, thanks. Yes, let's convert `textContent` to `&nbsp;`. That way we are treating the HTML as the source of truth.

raineorshine (changes requested):

src/device/selection.ts
ethan-james: OK, done.

PR #3788, opened 2026-02-07
Move preventDefault from keydown to beforeinput

### LINKED ISSUE #3707: [mobile] Keyboard doesn't open after Select All + 2x New Thought
I recommend thoroughly identifying the cause and planning a solution before starting the implementation on this one, as editing mode is quite sensitive.

## Steps to Reproduce

1. New Thought (→↓).  
2. Select All (←↓→). (**UPDATED: This is the correct gesture now.**) 
3. Close keyboard. 
4. New Thought. 
5. Close keyboard.
6. New Thought.

## Current Behavior

A new thought is created but the keyboard stays down. 

## Expected Behavior

The keyboard should always come up when a new thought is created.

## Notes

Restoring the browser selection on the cursor thought occurs automatically on render based on the conditions in `useEditMode`. Avoid setting the selection through a different mechanism. 

Select All is necessary to reproduce, but I suspect that the problem is not theoretically limited to that one command. It just happens to be setting up ab invalid selection state. Ideally we find a solution that prevents the invalid selection state to begin with, which will guard against other bug paths. I have noticed other cases where the keyboard doesn't come up, but have not been able to reproduce.

ethan-james:
Fixes #3707 

I didn't notice any actions being dispatched out of order, or any event handlers firing unusually, although it's not easy to get total transparency into event handling. I looked at `selection.set` and didn't see any difference in behavior between the thought where the keyboard opens and the thought where the keyboard doesn't open. I tried the old `document.activeElement.focus()` trick to see if Safari was treating the selection range as "unfocused", but no luck there either. I also tried replacing [selection.set](https://github.com/ethan-james/em/blob/2f079a84b6a8eb33907ca38b475aa50cac5f0e8c/src/components/Editable/useEditMode.ts#L59) with `contentRef.current.focus()` to see if Safari didn't like something about the selection range, but that didn't help.

The issue can be resolved by removing [requestAnimationFrame](https://github.com/ethan-james/em/blob/2f079a84b6a8eb33907ca38b475aa50cac5f0e8c/src/components/Editable/useEditMode.ts#L95) from `setSelectionToCursorOffset`. I tested #999 and #3129 after removing `requestAnimationFrame`, and #3129 seems fine but #999 is still a problem because auto-capitalization is not working.

I further discovered that auto-capitalization can potentially be fixed by no longer calling [preventDefault](https://github.com/ethan-james/em/blob/2f079a84b6a8eb33907ca38b475aa50cac5f0e8c/src/commands.ts#L609) on `keyDown`, but of course that allows `Enter` to be propagated into the new thought. This behavior can be blocked by calling `preventDefault` on `beforeinput` instead of `keydown`.

### DIFF
--- a/src/components/Editable/useEditMode.ts
+++ b/src/components/Editable/useEditMode.ts
@@ -88,14 +88,11 @@ const useEditMode = ({
 
         If the last action is swapParent, set the selection synchronously to keep the focus stable after the swap.
       */
-        if (isTouch && isSafari() && lastUndoableActionType !== 'swapParent') {
-          if (!selection.isThought()) {
-            asyncFocus()
-          }
-          setSelectionToCursorOffset()
-        } else {
-          setSelectionToCursorOffset()
+        if (isTouch && isSafari() && lastUndoableActionType !== 'swapParent' && !selection.isThought()) {
+          asyncFocus()
         }
+
+        setSelectionToCursorOffset()
       }
     },
     // React Hook useEffect has missing dependencies: 'contentRef', 'editMode', and 'style?.visibility'.


src/components/Editable/useEditMode.ts
```diff
@@ -92,7 +92,7 @@ const useEditMode = ({
           if (!selection.isThought()) {
             asyncFocus()
           }
-          requestAnimationFrame(setSelectionToCursorOffset)
+          setSelectionToCursorOffset()
         } else {
           setSelectionToCursorOffset()
         }
```
raineorshine: Please restructure this to avoid duplication.

PR #3811, opened 2026-02-17
iOS gesture cleanup

karunkop:
https://github.com/cybersemics/em/pull/3724


raineorshine:
Thanks, but this won't work because it breaks colocation of gestures and commands. We want the gesture to be defined within the command module.

Issue #3168, opened 2025-08-08
Disable antialiasing on all snapshots


raineorshine:
It would be preferable to have a single snapshot function so that developers do not have to decide whether to use the antialiased version (default) or the non-antialiased version.

We will need to fix the stacking context which is altered by `backface-visibility` and `perspective` to avoid breaking drop target z-index.

> Technically, there's no strong justification for taking antialiased snapshots. However, the CSS properties commonly used to disable antialiasing, particularly `backface-visibility` and `perspective`, have been found to interfere with `drag-and-drop` tests. These properties tend to affect `z-index`rendering, which leads to expected but inconsistent snapshot results in those scenarios.
> 
> The primary reason for implementing this snapshot utility was to handle text-heavy components, especially those tied to user interactions, which are more susceptible to pixel-level rendering differences. Given this context, I recommend keeping the current setup as is for now. That said, we can continue to explore more reliable alternatives to ensure consistency across all Puppeteer-based tests.

https://github.com/cybersemics/em/pull/3164#discussion_r2259194532

PR #3830, opened 2026-02-20
Consolidate screenshot antialiasing logic

karunkop:
#3168 

- Ensured consistency by keeping screenshot utility that disables anti-aliasing.

While testing for flakiness, there were some puppeteer tests that were failing intermittently. Here are those tests that were failing time to time:
- https://github.com/karunkop/em/actions/runs/22214713593
- https://github.com/karunkop/em/actions/runs/22174601938
- https://github.com/karunkop/em/actions/runs/22214497205
- https://github.com/karunkop/em/actions/runs/22174846897

The above issues found have been fixed in separate PR, #3831


src/e2e/puppeteer/__tests__/__image_snapshots__/modal/modal-help.png
raineorshine, line 1: It looks like the search icon is missing from the snapshot. 

Ideally we want the search icon to be included so we have visual regression coverage over it.

src/e2e/puppeteer/__tests__/ui.ts
```diff
@@ -19,15 +19,18 @@ expect.extend({
 
 vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
 
+/** Takes a screenshot with hardware acceleration disabled. */
+const takeScreenshot = () => screenshot({ hardwareAcceleration: false })
```
raineorshine: `takeScreenshot` is a poor name and impossible to meaningfully distinguish from `screenshot`.

Regardless, there isn't enough duplication here to warrant the abstraction. There is no complexity to hide, so better to inline it.

src/e2e/puppeteer/__tests__/sidebar.ts
```diff
@@ -14,6 +14,8 @@ expect.extend({
 
 vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
 
+/** Takes a screenshot with hardware acceleration disabled. */
+const takeScreenshot = () => screenshot({ hardwareAcceleration: false })
```
raineorshine, line 18: No need for a separate function.

raineorshine (changes requested):

There's still a bit of the core problem left in that the developer needs to somehow know whether to disable hardware acceleration or not. I'd like to merge this PR anyway, because it is an improvement. I will create a new issue for either eliminating or defining more predictable criteria for when to use `hardwareAcceleration: false`.

PR #3831, opened 2026-02-20
Puppeteer ci errors

karunkop:
As found out in #3830, this PR contains the fixes for those flaky tests. However there is still one test related with drag and drop favorites which keeps on timing out:

- https://github.com/karunkop/em/actions/runs/22214713593

All other appeared flaky tests have been fixed!


src/e2e/puppeteer/helpers/waitForColorSwatches.ts
```diff
@@ -0,0 +1,10 @@
+import waitForSelector from './waitForSelector'
+
+/** Waits for the text or background color swatch to be visible and returns the selector. */
+const waitForColorSwatches = async (type: 'text' | 'background', color: string) => {
+  const selector = `[aria-label="${type} color swatches"] [aria-label="${color}"]`
+  await waitForSelector(selector, { visible: true, timeout: 5000 })
```
raineorshine: This raises some questions for me.

The `click` handler [already waits for the selector](https://github.com/cybersemics/em/blob/5e343bc556459fe339208cb532a93fe495e3134c/src/e2e/puppeteer/helpers/click.ts#L38), so why do we need a special case for color swatches?

What is `visible: true`, and why is it needed here when it's not needed in other places that `waitForSelector` is used?

src/test-helpers/importToContext.ts
```diff
@@ -1,27 +1,33 @@
...
-        importText({
-          path,
-          text: _text,
-        }),
-      )
-    )
+    const isRootImport = path?.length === 1 && path[0] === HOME_TOKEN
+    const result = path && dispatch(importText({ path, text: _text }))
+    // if it is a root import, explicitly set the cursor to the last root-level thought
+    // sometimes in puppeteer tests, the cursor is not set to the last root-level thought after import, so we need to set it explicitly.
```
raineorshine: Why does Puppeteer set the cursor sometimes but not otherwise?

We're always aiming to understand the root cause of issues. Not only does that allow us to design the correct solution, but it expands our understanding of the app's behavior and often underlying technical details of the browser or platform.

Computers are deterministic and "sometimes" just means that there is an undiscovered timing issue. It is exceedingly rare that we would proceed with a workaround without an explanation, and that only after significant research, many attempts, and thorough documentation about the process.

I don't remember where, but we have encountered this issue before and I think we has some insight about the cause. I suggest searching old issues.

src/e2e/puppeteer/helpers/waitForColorSwatches.ts
karunkop: Yes you are correct about the click handler already waiting for the selector. That was an oversight on my part. Puppeteer was timing out as it was not able to find the `color swatches` selector, so i thought may be we will need to waitForSelector before click. `waitForSelector` helper already uses `visible: true` which basically waits until an element with that selector is visible in the DOM. (https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/helpers/waitForSelector.ts#L5)

This issue is very rare in CI so it's hard to reproduce. May be we can just increase the timeout for color swatches selector but that would be arbitrary

src/test-helpers/importToContext.ts
karunkop: Yes we have encountered the cursor being null after paste issue before. 
https://github.com/cybersemics/em/blob/main/src/e2e/puppeteer/__tests__/ui.ts#L64-L65
Here we opted for a hacky solution using `clickThought` so cursor is guarenteed to be set after paste. I have thoroughly gone through the logic for `importText` (https://github.com/cybersemics/em/blob/main/src/actions/importText.ts#L209-L217) which seems fine and it should be always setting the cursor on the last thought after paste. So the issue has to be related with some sort of race conditions in CI between arrowUp press event and paste. I don't know for sure just yet as the issue seems very rare in CI too.

raineorshine: Yes, exactly. This was my response back in December:

> Thanks for the quick response! However, we need to get to the bottom of the problem and understand the cause.
> 
> The real question is, why is the cursor null after paste in the CI? It should be set to the last pasted top level thought.

I merged that solution at the time, but made it clear that we needed to find the root cause. Please do not implement workarounds.

src/e2e/puppeteer/helpers/waitForColorSwatches.ts
raineorshine: Let's keep it open until you identify the timing issue and root cause.

Issue #3615, opened 2026-01-13
Text Formatting is not carried over when splitting sentence


BayuAri:
## Step to reproduce
1. Create a new thought
2. Add a multi sentences copy as the title
3. Tap Bold button
4. Verify that tittle are bold text now
5. Tap Split Sentence button

## Current behavior
The first split thought is bold but the other split-thoughts are not bold. 
The formatting only applies onto the first split-thought.
This also happens on Underline, Italic and Strikethrough formatting

https://github.com/user-attachments/assets/8087ff92-bfcc-4e5b-909f-191e1c4c3545

## Expected behavior
The formatting should be carried over to all of the new split-thought.

ethan-james:
I verified that this breaks up the sentences without any regard for styles, such that

```
**This is sentence one. This is sentence two. This is sentence three.**
```

becomes

```
  - **This is sentence one.
  - This is sentence two.
  - This is sentence three.**
```

So we need a method that will allow us to figure out what tags are active at a given point so that we can break them up in a comprehensive way. I thought that I might find something like that in `formatSelection` but didn't realize that was using native browser behavior to handle styles. Let me know if there's somewhere in the code where I could find an existing style parser, and if not then we can brainstorm what it would need to handle robustly.

PR #3832, opened 2026-02-20
Use selection.splitNode() to preserve tags while splitting sentences

### LINKED ISSUE #3455: [Mobile][Web] Split thought with color splits the html hardcode
**Step to repro**

1. Create a new thought
2. Add a any collor to the thought
3. Tap or click Split button on tthe toolbar
4. Observe the splitted thought

**Current behavior:**
Split thought is showing html hardcode 

**Expected:**
Splitting should not be implemented on the hardcode.

Image

Image

ethan-james:
Fixes #3455 and #3615

#3455 has a bit of a different root cause than #3615, but I needed to alter the comma-splitting logic in different ways for both, so it would cause merge conflicts to separate them.

#3455 had two issues: semicolons in style tags, and commas in style tags. Semicolon-splitting could be solved by adding an `isStyle` function that performs a similar check to the existing `isUrl` function. Comma-splitting logic happened in a later step, and the logic wasn't as robust. So I added a `separateByComma` function that does the following:

1. replaces commas within style blocks with a comma token
2. replaces commas outside of style blocks with the separator token
3. replaces the comma token with a `,`

For #3615, the solution was to place the markup into a div and use `selection.splitNode()` to preserve tags when splitting into sentences. It was necessary to traverse nodes using `offsetFromClosestParent`, but the end result is that the text is extracted from the div with all tags correctly closed using native browser behavior.

### DIFF
--- a/src/commands/__tests__/splitSentences.ts
+++ b/src/commands/__tests__/splitSentences.ts
@@ -32,7 +32,7 @@ describe('splitSentences', () => {
   - This is sentence three.`)
   })
 
-  it('splits a thought with multiple sentences and styles', () => {
+  it('splits a thought with multiple sentences all wrapped in a single bold tag', () => {
     store.dispatch([
       newThought({ value: '' }),
       setCursor(['']),
@@ -48,7 +48,7 @@ describe('splitSentences', () => {
   - **This is sentence three.**`)
   })
 
-  it('splits a thought with a single sentence plus font tags', () => {
+  it('does not split a thought with a single sentence and a font tag', () => {
     store.dispatch([
       newThought({ value: '' }),
       setCursor(['']),
@@ -67,7 +67,7 @@ describe('splitSentences', () => {
 </ul>`)
   })
 
-  it('splits a thought with a multiple comma-delimited sentences plus font tags', () => {
+  it('splits a thought with a multiple comma-delimited sentences plus a font tag', () => {
     store.dispatch([
       newThought({ value: '' }),
       setCursor(['']),


src/commands/__tests__/splitSentences.ts
```diff
@@ -30,6 +32,64 @@ describe('splitSentences', () => {
...
+    executeCommand(splitSentencesCommand, { store })
+
+    const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
+    expect(exported).toBe(`- __ROOT__
+  - **This is sentence one.**
+  - **This is sentence two.**
+  - **This is sentence three.**`)
+  })
+
+  it('splits a thought with a single sentence plus font tags', () => {
```
raineorshine: There is no split in this case, so we can say: `does not split a thought with a single sentence and font tag`

src/commands/__tests__/splitSentences.ts
```diff
@@ -30,6 +32,64 @@ describe('splitSentences', () => {
   - This is sentence three.`)
   })
 
+  it('splits a thought with multiple sentences and styles', () => {
```
raineorshine: "and styles" is too general, Let's say "multiple sentences in a bold tag".

We want to keep the tests and test descriptions as concrete as possible and not make assumptions about other styles, which may or may not be handled differently.

src/commands/__tests__/splitSentences.ts
```diff
@@ -30,6 +32,64 @@ describe('splitSentences', () => {
...
+    expect(exported).toBe(`<ul>
+  <li>__ROOT__  
+    <ul>
+      <li><font color="#000000" style="background-color: rgb(0, 214, 136);">font</font></li>
+    </ul>
+  </li>
+</ul>`)
+  })
+
+  it('splits a thought with a multiple comma-delimited sentences plus font tags', () => {
```
raineorshine: This says font "tags" but the test case only contains one font tag.

ethan-james:
> Thanks for the submission and the clear PR description. This looks fine to me.
> 
> Just a few recommendations to improve the test titles.
> 
> Since `splitNode` is now HTML aware, it may have made sense to migrate to proper HTML parsing and split nodes at the AST level before serializing it back into an HTML string. The risk is that we encounter additional edge cases that this algorithm can't handle, or that the complexity of the regex approach continues to grow, in which case we'll have to throw away a lot of this work. However, since we are already herIe I'm okay to keep what we have and proceed tentatively.

I did consider that, and then I decided it might be a case of the cure being worse than the disease. We may need to go that way in the future, but I identified that:

1.  there are only 5 markup tags that we're using (that I know of)
2.  font tag is the only one with an attribute, and
3. the style attribute is the only attribute

It's also possible for `.` or `!` to appear in style rules, but I don't think it's possible for us to generate such rules. Let me know if there's something else I need to test.

raineorshine (approved):
Good, I'm glad you've considered it. We'll see how this holds up. If we don't encounter more edge cases then I'm happy to call it a win and move on.

> I did consider that, and then I decided it might be a case of the cure being worse than the disease. We may need to go that way in the future, but I identified that:
> 
> 1. there are only 5 markup tags that we're using (that I know of)
> 2. font tag is the only one with an attribute, and
> 3. the style attribute is the only attribute

Yes, I think that's about right. My worry is with multiple overlapping styles with different ranges, but I haven't stress tested it.

@BayuAri Feel free to do another round of UX stress testing of split sentences on `main` to uncover any faults! (I merged this prematurely.)

PR #3834, opened 2026-02-23
Fix flaky color swatch test by waiting for element visibility

### LINKED ISSUE #3827: Flaky test: color swatch element not found
https://github.com/cybersemics/em/actions/runs/22185136413/attempts/1?pr=3823

wulfmeister:
## Summary

Fixes the intermittent failure in `color.ts > remove all formatting from the thought` where the test fails with:

```
No element found for selector: [aria-label="text color swatches"] [aria-label="blue"]
```

## Root Cause

The `click` test helper has two code paths:

1. **Simple selector** (no offset/x/y) — calls `page.click(selector)` directly
2. **With offset** — calls `page.waitForSelector(selector, { timeout: 1000 })` first, then clicks

The color tests use the simple path. The problem is that after clicking the "Text Color" toolbar button, the `ColorPicker` renders inside a `Popover` that uses `FadeTransition` with `unmountOnExit`. Even though animation durations are set to 0 during e2e tests (via `navigator.webdriver` in `src/util/durations.ts`), the React render cycle still takes a non-zero amount of time to mount the popover children into the DOM. When `page.click()` fires immediately after, the swatch elements may not exist yet.

## Fix

Add `page.waitForSelector(selector, { visible: true })` to the simple selector path in the `click` helper, making it consistently wait for element visibility before clicking — matching the pattern already used elsewhere in the test suite (e.g., `ui.ts` uses `waitForSelector` before interacting with command palette, gesture menu, etc.).

Fixes #3827

## Test plan

- [x] All 138 unit test files pass (1292 tests)
- [ ] CI e2e tests pass (requires browserless, cannot run locally)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/e2e/puppeteer/helpers/click.ts
+++ b/src/e2e/puppeteer/helpers/click.ts
@@ -1,4 +1,4 @@
-import { JSHandle } from 'puppeteer'
+import { ElementHandle, JSHandle } from 'puppeteer'
 import { page } from '../setup'
 
 interface Options {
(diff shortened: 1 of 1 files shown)


wulfmeister:
Hi! I wanted to try out Opus 4.6 using the new updates to the Claude Code desktop app (layer on top of the CLI) which has built in browser testing/iterating. I chose 3 random Issues with varying difficulty. Please feel free to reject/ignore!

src/e2e/puppeteer/helpers/click.ts
raineorshine: Fix looks good, but it would be better to restructure the code so that `page.waitForSelector` is not duplicated in two places.

wulfmeister:
Tried again, hopefully that is more elegant

PR #3854, opened 2026-02-26
Removed hardware acceleration css property when disabling antialiasing

### LINKED ISSUE #3847: Unify `hardwareAcceleration` snapshot option
> It would be preferable to have a single snapshot function so that developers do not have to decide whether to use the antialiased version (default) or the non-antialiased version.
> 
> We will need to fix the stacking context which is altered by `backface-visibility` and `perspective` to avoid breaking drop target z-index.
> 
> > Technically, there's no strong justification for taking antialiased snapshots. However, the CSS properties commonly used to disable antialiasing, particularly `backface-visibility` and `perspective`, have been found to interfere with `drag-and-drop` tests. These properties tend to affect `z-index`rendering, which leads to expected but inconsistent snapshot results in those scenarios.
> > The primary reason for implementing this snapshot utility was to handle text-heavy components, especially those tied to user interactions, which are more susceptible to pixel-level rendering differences. Given this context, I recommend keeping the current setup as is for now. That said, we can continue to explore more reliable alternatives to ensure consistency across all Puppeteer-based tests.
> 
> [#3164 (comment)](https://github.com/cybersemics/em/pull/3164#discussion_r2259194532)

---

There's still a bit of the core problem left in that the developer needs to somehow know whether to disable hardware acceleration or not.

We need a configuration that works with on all snapshots, i.e. always antialiased or never antialiased.

_Originally posted by @raineorshine in https://github.com/cybersemics/em/pull/3830#pullrequestreview-3834821748_

karunkop:
Close #3847 

I managed to remove `perspective` css property which caused issue with z-indexes for modals and popovers. Also I conducted multiple flakiness test and the tests seem to be stable.

[Here](https://github.com/karunkop/em/actions?query=branch%3Asnapshot-antialiasing) is the result for the CI checks.


raineorshine (approved):
Beautiful! Well done.

PR #3871, opened 2026-02-27
Convert isEditAddition to editThoughtDirection enum

### LINKED ISSUE #3551: Undo color highlighting makes the thought invisible
## Step to reproduce
1. Create some thoughts
2. Select any thought
3. Tap Text Color button on the toolbar
4. Select Highlight
5. Verify that the selected thought is highlighted with correct color
6. Tap Undo button

## Current behavior
The selected thought is invisible.
When the highlight is implemented, the text's color is auto-changed from white to black, so when user taps redo, the highlight is removed but the text's color is not reverted back to black
Video: 

https://github.com/user-attachments/assets/d47ffc6c-374d-4ec5-9abd-b0e5a9ca71e0

## Expected behavior
The text's color should be reverted to white so it is visible.

ethan-james:
Fixes #3551, although we may still wish to combine the two [formatSelection](https://github.com/ethan-james/em/blob/65d9e9acd061ec2a3db4f3424c0bd4874a7f6716/src/components/ColorPicker.tsx#L97) dispatches

`isEditAddition` was always false for the first `editThought` action because it was checking `lastAction?.type === 'editThought'`. So when the second `editThought` action came in, it would never be merged even though `isEditAddition` became true. This bug was covered up somewhat because [thoughtChangeHandler](https://github.com/ethan-james/em/blob/65d9e9acd061ec2a3db4f3424c0bd4874a7f6716/src/components/Editable.tsx#L302) was throttled, so if you typed faster than 1 character every 500ms, multiple edits would come in as a single dispatch and create a single undo/redo step.

Now it's tracking [EditThoughtDirection](https://github.com/ethan-james/em/blob/65d9e9acd061ec2a3db4f3424c0bd4874a7f6716/src/redux-enhancers/undoRedoEnhancer.ts#L21) in an enum to make it more explicit what this functionality is designed to do, as well as to decouple the merging from the last action type. Rather than a true/false boolean, we have three states so that it can be set to `None` before an `editThought` action comes in.

---

This fixes #3551 in an incidental way, since adding the font tag and then adding the background color both increase the length of the thought. Not sure if you're going to love that since it's conceptually different from typing more text, so we could still combine the two dispatches as discussed in the original issue. It didn't seem like passing an array of actions or a thunk made any difference, so we would probably be looking at option 3, New `formatColor` action creator.

### DIFF
--- a/src/commands/__tests__/redo.ts
+++ b/src/commands/__tests__/redo.ts
@@ -106,7 +106,7 @@ it('redo contiguous changes', () => {
   const exportedBeforeRedo = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
 
   const expectedOutputBeforeRedo = `- ${HOME_TOKEN}
-  - Atlantic
+  - A
   - B`
 
   expect(exportedBeforeRedo).toEqual(expectedOutputBeforeRedo)


ethan-james:
There seems to be a bit more complexity in what actions are supposed to be batched together, so I'll see what I can do to preserve the existing functionality more closely.

raineorshine (commented):
Yes, correct!

raineorshine (changes requested):
Thank you, looks good.

Please add a test at your convenience.

raineorshine (approved):
Good catch. Can you create a new issue and assign it to Karun? We should create a standard way of clearing the middleware state as part of the test lifecycle.

Will merge this in the mean time. Thanks!

PR #3899, opened 2026-03-06
After the cursor is cleared, store a style wrapper that can be re-applied after typing

### LINKED ISSUE #3673: Clear Thought should not remove color
## Steps to Reproduce

1. Create a thought. 
2. Set the color of the thought. 
3. Activate Clear Thought (→←).
4. Edit the thought. 

## Current Behavior

The edited value is reverted to the default color. 

https://github.com/user-attachments/assets/6b671740-2a9a-42f6-9f26-6c7b94f3a1c4

image

## Expected Behavior

The edited value should retain its color.

image

ethan-james:
Fixes #3673 

If a thought had a style tag that wrapped the entire thought, store that wrapper when the cursor is cleared
After the cursor is cleared, store any style wrapper that the old thought had if it wrapped the entire thought. If the user types in the cleared thought, re-apply the wrapper and immediately propagate the new value to the editable using `editThought`'s `force` property to update the editable nonce.

The only thing that I've uncovered that might still be an issue is the following:

## Steps to Reproduce

1. Create a thought.
2. Set the color of the thought.
3. Activate Clear Thought (→←).
4. Set the color of the thought back to white or another color.
5. Edit the thought.

## Current Behavior

The new text receives the color from step 2.

## Expected Behavior

Maybe it should be set to the color from step 4? Let me know what you think.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -381,7 +381,6 @@ const Editable = ({
         // That style should be re-applied once they type something. (#3673)
 
         const wrappedValue = state.cursorCleared ? applyOuterTag(e.target.value, oldValue) : e.target.value
-        console.log(wrappedValue)
         const newValue = stripEmptyFormattingTags(addEmojiSpace(trimHtml(wrappedValue)))
 
         /* The realtime editingValue must always be updated (and not short-circuited) since oldValueRef is throttled. Otherwise, editingValueStore becomes stale and heights are not recalculated in VirtualThought.


src/components/Editable.tsx
```diff
@@ -121,6 +122,23 @@ const Editable = ({
   const rank = useSelector(state => getThoughtById(state, head(simplePath))?.rank || 0)
   const fontSize = useSelector(state => state.fontSize)
   const isCursorCleared = useSelector(state => !!isEditing && state.cursorCleared)
+
+  // When the cursor is cleared, there may be an existing style that wraps the entire thought.
+  // That style should be re-applied once they type something. (#3673)
+  const cursorClearedWrapper = useSelector(state => {
```
raineorshine: Are you sure this needs to be reactive? It's only used in the change handler from what I can tell.

You can reorganize `newValue` into the thunk to get access so the Redux state.

src/util/wrapHtml.ts
raineorshine, line 1: I prefer to factor out helper functions only when they are used by multiple files, or if they are very large. Otherwise define it within the module where it is used, keeping it as close to the call site as possible. 

Co-location by default, abstraction if necessary.

raineorshine (changes requested):
Thanks, looks good overall.

> The only thing that I've uncovered that might still be an issue is the following:
> 
> ## Steps to Reproduce
> 1. Create a thought.
> 2. Set the color of the thought.
> 3. Activate Clear Thought (→←).
> 4. Set the color of the thought back to white or another color.
> 5. Edit the thought.
> 
> ## Current Behavior
> The new text receives the color from step 2.
> 
> ## Expected Behavior
> Maybe it should be set to the color from step 4? Let me know what you think.

Eventually that seems like the correct behavior, but it's probably out of scope for this since the same behavior is also broken in `main` when trying to set the color of an empty thought. i.e. It's not currently possible to set the color before starting to type into an empty thought.

ethan-james:
> Thanks and let me know if this makes sense.

Yes, it does make sense, thanks. I do want to keep it as 2 steps so that I can compare the wrapped value to `e.target.value` to see whether it should flush.

raineorshine (approved):

PR #3885, opened 2026-03-04
fix: Gesturemenu background should be overlay with device's status bar in Capacitor app

cindmichelle:
Close #3727

Add a fix where GestureMenu background should be underneath of device's status bar when being opened from Capacitor

### Notes
please start the code review changes from [63795b3](https://github.com/cybersemics/em/pull/3885/commits/63795b3f22b053e34e12101da262c2c1280264b7) , since the earlier commit are more related to #3646 

---
## Results

**Android**

https://github.com/user-attachments/assets/dc039ac6-096e-40f8-8ce7-cae92a21a4b3

**iOS**

https://github.com/user-attachments/assets/437764c2-9984-443e-8ca0-767f25e5ff03

PR #3909, opened 2026-03-07
[Liminal UI] New design for Tip

### LINKED ISSUE #3872: New design for Tip
> **Implement a new design for the `Tip` component in line with the Liminal UI redesign, including responsive design support.**

Image

---

## Breakdown

The new design for the Tip is made up of a few elements put together:

Image

[View in Figma](https://www.figma.com/design/oQnevfxv79fhTJnOs0fz9k/em-design?node-id=3501-28023&t=4XKQVnCEfliqbx7n-4)

#### Design notes

- This layer separates the **contents of the Tip** from **the contents of the rest of the app**.
- **The overlay is made up of:**
    - A simple linear gradient which transitions from **fully transparent** at the top to **black** at the bottom.
    - A progressive blur backdrop filter which blurs the app's contents underneath, starting from 24px blur at the bottom to 0px blur at the top.

#### Design notes

- This is a simple translucent image that displays on top of the overlay. This brings our "liminal" design style onto the interface.
- Look at the first image. On the desktop and landscape mobile displays, the white part of the glow is directly 
- On mobile (portrait) devices, the glow image is **flipped horizontally**.

#### Implementation notes

- **A WEBP image for the background glow is available in the [assets folder](https://public.3.basecamp.com/p/TmFdohvobCnQpbMUqntjsw3L/vault).**
- Be sure to use the `mixBlendMode: 'screen'` style on the image, so it blends with the contents underneath rather than showing as an opaque image.

### 3. Clear button

- A simple simple text button that, when clicked, will hide the tip.
- Use pre-existing icon from the project for the "close" icon – it's okay if the icon used isn't a 100% match, so long as it's a close icon.

### 4. Animation

Let's try a simple opacity fade-in/fade-out for now. Use one of the preset durations in `durations.config.ts` and `FadeTransition` to implement this.

## Responsive design

The Tip should scale smoothly and responsively across different device and tip sizes. There are a few responsive rules that need to be implemented:

### 1. Breakpoints

- The primary break is between the mobile (portrait) design and the mobile (landscape) and wider designs.
- Use `md` as the maximum breakpoint for mobile (portrait).

### 2. Tip differences between mobile (portrait) devices and other breakpoints

- Look at the provided glow image. The image is a gradient that transitions from purple (left) to white (right).
- Now, look at the mobile (portrait) design in the image at the top of this issue. The glow transitions from white (left) to purple (right, but off-screen)
    - In other words, on the portrait mobile breakpoint, the glow image is _flipped_, so it will need to be flipped via CSS on mobile (portrait) breakpoints.
    - Note that the tip is fixed to the bottom-left corner of the screen on mobile, and that the tip's contents flow left-to-right.
- Now, look at the mobile (landscape) / desktop designs. The glow transitions from purple (left) to white (right), just like the image – no need to flip it.
    - Note that on these breakpoints, the tip is fixed to the bottom-right corner, and the tip's contents flow from right-to-left.

### 3. Glow image positioning rules

- The glow image is large, and designed to span off-screen in most cases. This is especially visible on mobile screens. See below for a visualization of how the glow image should be positioned on all of our breakpoints.

Image

This diagram shows what percent of the glow image is left under the viewport, and what is visible.

### 4. Accommodate different tip heights

- The longer the tip's text is, the taller the tip's content will be. We need to accommodate this variance by **repositioning the glow image** and **resizing the overlay**.
- See below for a visualization of the desired behavior. Red lines added for clarity. Note that the more lines there are in the tip, the **higher up the glow image is moved**, and the **taller the overlay is (indicated by the top red line)** – which blurs out more thoughts.
- **Note:** The glow image **does not** resize, it simply moves upward to cover more of the screen. The overlay **does** resize.

Image

Exact values can be tuned visually – no strict pixel-perfect requirement.

fbmcipher:
Closes #3872 

## Summary

Redesigns the Tip component in a new Liminal UI visual style, fully responsive across all device sizes. Adds swipe-to-dismiss on touch devices with velocity-aware gesture recognition.

CleanShot 2026-03-07 at 02 11 13@2x

CleanShot 2026-03-07 at 02 13 23@2x

### Tip redesign (`Tip.tsx`)
- **Three-layer visual design**: gradient overlay with progressive blur with dims underlying app content, a pre-rendered WebP glow image, and the content layer (label, message, clear button). This is similar the structure of the Command Center.
- **Swipe-to-dismiss**: direction-agnostic gesture that aggregates swipe distance and smooths velocity. A combined score of distance + velocity determines dismissal, so a quick flick works as well as a slow drag. This is _similar_ to the logic used in the sidebar, with the exception of the direction-agnostic logic which allows for "scribble-to-dismiss"-like interactions.  
- **Snap-back animation**: if the swipe doesn't meet the threshold, opacity animates back to 1.
- **Safe area insets**: bottom padding accounts for notch/home indicator via `env(safe-area-inset-bottom)`
- **Updated `ProgressiveBlur.tsx`** with a new `mask` prop, which allows us to mask the progressive blur effect however we'd like. Used on desktop to feather the left edge of the blur.
- **Fully responsive.** The tip is left-aligned on mobile (portrait), and right-aligned on everything else including mobile (landscape). The background image scales accordingly. The tip takes full 100% width, clamped to a maximum of 1500px.

### Swipe-to-dismiss duplication

There's some cross-over between the swipe gesture logic in the sidebar and in the tip. After follow-up PRs apply swipe-to-dismiss to Alert & Confirm, a pure refactor will extract this logic into a shared hook to eliminate the duplication. 

### Triggering a tip in testing

**There are currently no tips that trigger on mobile in normal usage.**

The easiest way I have found to manually trigger a tip is to apply this patch which adds a "Test Tip" button to the toolbar:

<details>
<summary>Testing patch</summary>

Apply with `git apply patch.diff`:

```diff
--- a/src/components/Toolbar.tsx
+++ b/src/components/Toolbar.tsx
@@ -290,6 +290,25 @@
             data-scroll-at-edge={customize}
             onScroll={onScroll}
           >
+            <button
+              onClick={() => dispatch(showTip({ tip: TipId.NewThought }))}
+              className={css({
+                fontSize: '10px',
+                padding: '6px 8px',
+                marginTop: '12px',
+                marginRight: '4px',
+                background: 'rgba(255,255,255,0.1)',
+                color: 'white',
+                border: '1px solid rgba(255,255,255,0.2)',
+                borderRadius: '4px',
+                cursor: 'pointer',
+                whiteSpace: 'nowrap',
+                zIndex: 'toolbar',
+                flexShrink: 0,
+              })}
+            >
+              Test Tip
+            </button>
             {commandIds.map(id => {
```

</details>

### DIFF
--- a/src/components/Tips/useSwipeToClear.ts
+++ b/src/components/Tips/useSwipeToClear.ts
@@ -49,6 +49,8 @@ const useSwipeToClear = ({
         onComplete: () => {
           dismissing.current = false
           onDismissed()
+          // Reset so the next showing of the tip starts with completion = 0.
+          swipeDistance.set(0)
         },
       })
     },
@@ -120,6 +122,8 @@ const useSwipeToClear = ({
         if (dist >= threshold) {
           // Swipe distance already reached threshold — completion is already 1, dismiss immediately.
           onDismissed()
+          // Reset completion back to 0.
+          swipeDistance.set(0)
         } else {
           // Velocity-boosted dismiss — animate the remaining distance to threshold.
           const remainingPx = threshold - dist


src/components/Tips/Tip.tsx
```diff
@@ -1,94 +1,386 @@
...
+      {/* ── Layer 2: Glow image ──────────────────────────────────────────── */}
+      {/* A pre-rendered WebP light-leak effect positioned behind the text content.
+          On mobile it appears from the left (flipped via scaleX(-1)); on desktop from the right. */}
+      <div
+        className={css({
+          animation: 'fadein {durations.medium} ease',
+          position: 'absolute',
+          pointerEvents: 'none',
+          opacity: 1,
+          backgroundImage: 'url(/img/tip/tip-glow-alpha.webp)',
```
raineorshine: I believe we need to pre-load the image so there isn't a delay when the tip is shown for the first time.

e.g. https://github.com/cybersemics/em/pull/3885/changes#diff-e2c4cb66ea36c7c639caff40eeb98d0d538180de000796dad0af45cb707127e6R201-R206

src/components/Tips/Tip.tsx
```diff
@@ -1,94 +1,386 @@
...
+        pointerEvents: 'none',
+        // Allow dragging through the tip overlay so drag-and-drop still works.
+        _dragHold: { pointerEvents: 'none' },
+        display: 'flex',
+        userSelect: 'none',
+      })}
+      style={{
+        ...positionFixedStyles,
+        // Override usePositionFixed bottom offset to sit flush at the bottom of the viewport.
+        bottom: 0,
```
raineorshine: Overriding `bottom` would seem to defeat the purpose of `usePositionFixed`. What is the intention here?

What is the expected behavior of the Tip when the virtual keyboard is up? Letting the keyboard hide the tip is easier and less buggy. It would preclude us ever displaying a tip while the user is typing, so we have to be willing to accept that.

src/components/Tips/Tip.tsx
```diff
@@ -1,94 +1,386 @@
...
+      }}
+    >
+      {/* ── Layer 1: Gradient overlay + progressive blur ─────────────────── */}
```
raineorshine: Let's factor this out into a `TipBlur` component. You can keep it in the Tip module, but separating it into a separate component will keep the main `Tip` component cleaner.

src/components/Tips/Tip.tsx
```diff
@@ -1,94 +1,386 @@
...
+            animation: 'fadein {durations.medium} ease',
+            position: 'absolute',
+            inset: 0,
+            background: 'linear-gradient(180deg, {colors.bgTransparent} 0%, {colors.bg} 100%)',
+          })}
+          style={{ opacity: isDismissing ? 0 : swipeOpacity, transition: fadeOut }}
+        />
+      </div>
+
+      {/* ── Layer 2: Glow image ──────────────────────────────────────────── */}
```
raineorshine: Let's also separate the glow into its own subcomponent.

src/components/Tips/Tip.tsx
```diff
@@ -1,94 +1,386 @@
...
+    [dispatch, swipeDistance, swipeOpacity],
   )
+
+  // ── Blur opacity (MotionValue) ──────────────────────────────────────────
+  // The blur layer uses a framer-motion MotionValue instead of inline CSS opacity.
+  // This avoids a Safari bug where animating opacity on a parent of backdrop-filter
+  // children breaks the blur rendering.
+  const blurOpacity = useMotionValue(0)
+  const isVisible = tip === tipId
+  const wasVisible = useRef(false)
```
raineorshine: Maybe you can use [usePrevious](https://github.com/cybersemics/em/blob/8e794ad2ead51c83d0e8a5b27a513aa2fd6f9939/src/hooks/usePrevious.ts#L4)?

raineorshine:
> When footer is visible, tip should be display in front of the footer (overlapping)
> @raineorshine is this expected behavior correct?

Yes, that's correct for now. The footer is going to be moved into the Sidebar soon.

> This issue is new on the PR but is this included in the scope for this PR?
> 
> ## [Android] Keyboard does not appear when creating a new thought or subthought after receiving a tip

Yes, if any unexpected behavior is changed from `main` that would be considered a regression. The appearance of the Tip and the touch interactions are new, but I would expect the browser selection to be unaffected by this PR.

> ## Tip + Sidebar Menu
> Tip and Sidebar Menu is overlapping, is this ok? I would say that the Tip should be hidden when user opens Sidebar Menu and it should be shown when user closes the Sidebar Menu,. The tip should only be dismissed manually by user to ensure that user has seen it. @fbmcipher @raineorshine

I'll follow @fbmcipher's guidance on this. Same question with whether to dismiss the Tip or not when the Command Center or a modal is opened.

src/components/Tips/Tip.tsx
fbmcipher: Done in a624cbbd59f10d808c9035e8e66db8f2b9ddf683. I created a new hook `usePrefetchImages` to centralize this logic, as we'll need to use the same technique across all components that use glow images.

src/components/Tips/Tip.tsx
fbmcipher: Thanks – resolved in cb198bb048071ae538e856d9b7f1863c8da9717d

src/components/Tips/Tip.tsx
fbmcipher: cb198bb048071ae538e856d9b7f1863c8da9717d

src/components/Tips/Tip.tsx
fbmcipher: Yes – thanks for spotting! 529bb060fa01a243ec5b2ad4858c8e3dcbb25482

src/components/Tips/Tip.tsx
fbmcipher: - `usePositionFixed` adds a slight offset (i.e. it wasn't positioning elements *right at the bottom* of the screen). `bottom: 0` was added to mitigate that, though yes, you're right that it does defeat the purpose of `usePositionFixed`.
- I think there are legitimate reasons we'd want to show a tip with the keyboard up, but right now `usePositionFixed` is pretty broken, and we're in the process of fixing it in other PRs. For now, the simpler decision is to let the keyboard hide tips – which I've done in 4367947bbf2adff36788ab38d1b621990b80fb84.
- If we come across a legitimate need to show a tip with the keyboard open, we can change that behavior.

src/components/Tips/Tip.tsx
raineorshine: Thanks, this helps.

The duplication of the URL is not ideal as it requires manually keeping it in sync with usage in the component. I have some ideas of how to improve this, but will address it in a separate issue: #3926

src/components/Tips/Tip.tsx
raineorshine: Thanks, that all sounds reasonable.

src/components/Tips/Tip.tsx
```diff
@@ -1,94 +1,329 @@
...
+import usePrevious from '../../hooks/usePrevious'
+import durations from '../../util/durations'
+import fastClick from '../../util/fastClick'
+import ProgressiveBlur from '../ProgressiveBlur'
+import CloseIcon from '../icons/CloseIcon'
+import useSwipeToClear from './useSwipeToClear'
 
-const icon = (
+/** Cumulative swipe distance (in px) at which a swipe fully fades and dismisses the tip. */
+const SWIPE_DISMISS_THRESHOLD = 200
```
raineorshine: Let's move this into the hook and treat it as the default.

src/components/Tips/Tip.tsx
```diff
@@ -1,94 +1,329 @@
...
 
-  const handleClose = useCallback(() => {
-    setIsDismissed(true)
-    onClose()
-  }, [onClose])
+  /** Prefetch the glow image used in the tip, so that the user doesn't see a loading delay when the tip first becomes visible. */
+  usePrefetchImages(['/img/tip/tip-glow-alpha.webp'])
+
+  /** True while the fade-out CSS transition is running (after the user taps Clear or completes a swipe). */
+  const [isDismissing, setIsDismissing] = useState(false)
```
raineorshine: I'm trying to wrap my head around why this is a reactive value. My understanding is that `isDismissing` can be set to true by `onSwipeDismiss` or `handleClose`, which then triggers the opacity animation through an effect. Is there a reason this has to be routed through local state and an effect rather than just calling a function directly that animates the opacity?

Either there is some kind of syncing with the React lifecycle that is needed or maybe I don't understand how `framer-motion` works.

---

Normally I wouldn't just copy-paste Claude output, but it *tentatively* affirms this, and provides the code example to better demonstrate what I have in mind:

<i>
Your intuition is correct — `isDismissing` doesn't *need* to be state. The only reason to route through state + effect would be if you needed React to re-render in response (e.g., to change `pointerEvents` or conditionally render something). But looking at the code, neither `isDismissing` nor `isAnimating` affects the render output at all.

The `animate()` call can be made directly from `handleClose` and `onSwipeDismiss` without involving state or an effect:

````tsx
// ...existing code...

  /** Animates opacity to 0 and dispatches dismissTip on completion. */
  const animateDismiss = useCallback(() => {
    const duration = durations.get('medium') / 1000
    animate(opacity, 0, {
      duration,
      ease: 'easeOut',
      onComplete: () => dispatch(dismissTip()),
    })
  }, [opacity, dispatch])

  const onSwipeDismiss = useCallback(
    (immediate: boolean) => {
      if (immediate) {
        dispatch(dismissTip())
      } else {
        animateDismiss()
      }
    },
    [dispatch, animateDismiss],
  )

  const { completion, touchHandlers } = useSwipeToClear({
    threshold: SWIPE_DISMISS_THRESHOLD,
    onDismiss: onSwipeDismiss,
  })

  /** Opacity derived from swipe completion. Linearly decreases from 1 -> 0. */
  const swipeOpacity = 1 - completion

  /** Begins the fade-out transition when the user taps the Clear button. */
  const handleClose = useCallback(() => {
    animateDismiss()
  }, [animateDismiss])

// ...existing code...

  useEffect(() => {
    if (isVisible && !wasVisible) {
      // Fade in — tip becoming visible (initial or reappearing after temporary hide).
      const duration = durations.get(wasVisible === false ? 'fast' : 'medium') / 1000
      opacity.set(0)
      const controls = animate(opacity, 1, { duration, ease: 'easeOut' })
      return () => controls.stop()
    }
    if (!isVisible && wasVisible) {
      // Fade out — temporarily hidden by keyboard/command center/sidebar/alert.
      const duration = durations.get('fast') / 1000
      const controls = animate(opacity, 0, { duration, ease: 'easeOut' })
      return () => controls.stop()
    }
    if (!isVisible) return
    // During an active swipe, sync opacity directly to swipe progress.
    opacity.set(swipeOpacity)
  }, [isVisible, wasVisible, swipeOpacity, opacity])

// ...existing code...
````

The one subtle thing to be aware of: since `opacity` is a `MotionValue` (not React state), calling `animate(opacity, ...)` from an event handler is perfectly safe — framer-motion manages it outside React's render cycle, so no lifecycle synchronization is needed.
</i>

File: src/actions/showTip.ts [hunk 8]
```diff
@@ -2,19 +2,29 @@ import State from '../@types/State'
...
+    const state = getState()
+    const hasOpenDropdown =
+      state.showCommandCenter ||
+      state.showColorPicker ||
+      state.showLetterCase ||
+      state.showSortPicker ||
+      state.showUndoSlider
+    if (hasOpenDropdown) {
+      dispatch(toggleDropdown())
+    }
```
raineorshine: Are you sure you need to check `hasOpenDropdown` first? Just calling `toggleDropdown` should close any open dropdowns.

Note that this would be a suboptimal way to do it anyway, as it requires the list of dropdowns be maintained separately from `toggleDropdown`. The main purpose of `toggleDropdown` is to centralize dropdown handling and avoid duplication and accidental decoupling.

src/components/Tips/Tip.tsx
```diff
@@ -1,94 +1,329 @@
...
-  const [isDismissed, setIsDismissed] = useState(false)
 
-  const onClose = useCallback(() => {
-    dispatch(dismissTip())
-  }, [dispatch])
+  // Selectors for UI state that should temporarily hide the tip when active to avoid visual conflicts.
+  const isKeyboardOpen = useSelector(state => state.isKeyboardOpen)
+  const showCommandCenter = useSelector(state => state.showCommandCenter)
+  const showSidebar = useSelector(state => state.showSidebar)
+  const hasAlert = useSelector(state => !!state.alert)
```
raineorshine: You can combine these into a single selector for better encapsulation (and in many cases better performance, though that's not the case here).

In general you want to aim for thick selectors to reduce React re-renders.

src/hooks/usePrefetchImages.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/** Prefetches images on mount so they are decoded and cached by the time they are rendered. */
+const usePrefetchImages = (srcs: string[]) => {
+  useEffect(() => {
+    srcs.forEach(src => {
+      const img = new Image()
+      img.src = src
+      img.decode()
+    })
+    // Prefetch only on mount since srcs are static and a new array reference is created each render.
```
raineorshine: My understanding is that the array reference does not effect the hook recalculation. It just looks at the elements in the array.

Same behavior, but it might be simpler to provide pass `srcs` as a dependency and not have to disable the linter if that's the case.

BayuAri (commented):
List of issues:

- [ ] [Tip slides down when keyboard is closed](https://github.com/cybersemics/em/pull/3909#issuecomment-4019832402) is **still reproducible on iOS Capacitor.**
This is not reproducible on Android nor Mobile Safari

## Step to Reproduce
1. Tap Test Tip button
2. Create Thought One
3. Tap the Scroll Zone or other blank area to dismiss the keyboard 

## Current behavior
After 2-3 secs, the Tip slides down and back up


https://github.com/user-attachments/assets/d5dd4c21-02ff-4364-9f07-9f2b5ceeef24


## Expected behavior
Nothing should happen to the Tip



- [x] [Tip + Command Center](https://github.com/cybersemics/em/pull/3909#issuecomment-4019835922) is **fixed**
Command Center and Tip are not overlapping. The Tip is shown when Command Center is dismissed. The Tip is persistent until user manually clears it.

 
https://github.com/user-attachments/assets/aca606e2-2b27-422e-a700-7029f97b77c0


- [x] [Tip rendered below footer](https://github.com/cybersemics/em/pull/3909#issuecomment-4019836927) is **fixed**

 
- [x] [[Mobile - Safari] The first tip is too high](https://github.com/cybersemics/em/pull/3909#issuecomment-4021105579) is **fixed**

- [ ] [[Android] Keyboard does not appear when creating a new thought or subthought after receiving a tip](https://github.com/cybersemics/em/pull/3909#pullrequestreview-3913476373) is still reproducible.
There is a new issue here with the same STR
## Tip is not shown after tapping Test Tip when keyboard is onscreen
## Step to Reproduce
1. Create a thought
2. When keyboard is onscreen, tap Test Tip button
3. Swipe →↓ to create a new thought


## Current behavior
Keyboard is not shown.
Keyboard is dismissed but Tip is not shown on step 2, this is a new issue. 
Home button at the bottom left hand corner is not visible on screen, user needs to scroll far down to make it visible.  


## Expected behavior
Tips should be shown accordingly

BayuAri:
## Tip is gone when user closes the app and reopen
## Step to Reproduce
1. Tap Test Tip
2. When the Tip is onscreen, close the app
3. Reopen the app
## Current behavior




## Expected behavior

BayuAri:
## Tip and Command Center are overlapping when user Select All or Select Between
## Step to Reproduce
1. Create a thought
2. Dismiss keyboard
3. Tap Test Tip
4. Swipe →↓← to Select All or ←↓→ to Select Between



## Current behavior
Command Center slides in behind the Tip

https://github.com/user-attachments/assets/114fc2c2-65f8-4d33-b99c-9354e61a4a06



## Expected behavior
The Tip should be dismissed when Command Center slides in.
Refer to Swipe up to open Command Center behavior

BayuAri:
## [iOS] [Mobile Safari]Tip is not visible when user cursor is on Note
## Step to Reproduce A: New Note
1. Create Thought One
2. Dismiss the keyboard
3. Tap Test Tip
4. Tap Note button on the toolbar
5. Tap Scroller Zone to dismiss the keyboard 


## Current behavior
The Tip disappears and Home buttons are not visible on screen. 
The Tip reappears when user places the cursor back on the thought. 


https://github.com/user-attachments/assets/5a412715-8834-4cab-a11f-be1477f0aa90




## Expected behavior
The Tip should be persistent on screen when cursor is on Note

src/components/Tips/Tip.tsx
fbmcipher: You (and Claude) were right – I had overcomplicated the logic here, partially as a result of the `wasVisible`/`isVisible` distinction and associated logic (which I've entirely removed).

We can just use a `MotionValue`, and we can completely sidestep state without changing the appearance at all: 885d8b07144e022d43b0078b49ca1645b26b6596

src/hooks/usePrefetchImages.ts
fbmcipher: Ah, TIL. I thought that `useEffect` always compared the identity of deps, not their elements/contents.

I compared the behavior of `usePrefetechedImages` with/without `srcs` as a dependency and couldn't see a difference in the network/console, so I removed the linter override and added `srcs` as a dep. 79386d1d84e018b23dd54be3c9cd1815ffbaeebc

src/actions/showTip.ts
fbmcipher: You're right, I totally missed that `toggleDropdown` closes all open dropdowns. Sorry for missing that. 9842ccc30b36b7d5c241ce5c677aca56ff1d515f

src/components/Tips/Tip.tsx
fbmcipher: Good idea: d8df08ad6666a2b47774adf0a853bad35d497975

src/components/Tips/Tip.tsx
fbmcipher: Thank you – I'll bear that in mind, about thick selectors. Now, we calculate `isHidden` within the selector's callback: 3fa04021855927aae586446191c8aea3b3081221

src/hooks/usePrefetchImages.ts
raineorshine: Yes, the outer array is ignored. It iterates through and checks the identify of elements of the dependency array. Elements within the array are compared by reference though.

BayuAri:
## Tip is shown on Setting and Customize Toolbar screens
## Step to Reproduce
1. Create Thought One
2. Tap Test Tip button
3. Tap Setting button on the toolbar
4. Tap Customize Toolbar


## Current behavior
Tip is shown on Setting and Customize Toolbar screens

image


## Expected behavior
Tip should be temporary hidden when user opens Setting and Customize Toolbar screens.
Refer to the behavior on Sidebar menu

src/hooks/useSafeArea.ts
```diff
@@ -0,0 +1,44 @@
...
+  const style = getComputedStyle(document.documentElement)
+  return {
+    top: parseInt(style.getPropertyValue(vars.top)) || 0,
+    right: parseInt(style.getPropertyValue(vars.right)) || 0,
+    bottom: parseInt(style.getPropertyValue(vars.bottom)) || 0,
+    left: parseInt(style.getPropertyValue(vars.left)) || 0,
+  }
+}
+
+/** Provides cached mobile device safe area insets (top, right, bottom, left), as integer values. Updates on resize. */
```
raineorshine: We should note in the JSDOC that the returned value is not reactive. Something else in the component has to trigger a render to pick up the new value.

I'm fine with it, but if we run into stale value issues we'll have to switch to local reactive state (possibly throttled).

src/actions/showTip.ts
raineorshine: I should probably add a dedicated `closeDropdowns` action-creator. `toggleDropdown` is a bit too polymorphic. Totally not obvious.

src/components/Tips/Tip.tsx
```diff
@@ -1,93 +1,315 @@
...
+    onDismiss: onSwipeDismiss,
+  })
+
+  /** Opacity derived from swipe completion. Linearly decreases from 1 -> 0. */
+  const swipeOpacity = 1 - completion
+
+  // ── Handlers ────────────────────────────────────────────────────────────
+
+  /** Begins the fade-out transition when the user taps the Clear button. */
   const handleClose = useCallback(() => {
```
raineorshine: `handleClose` is nothing more than`animateDismiss`, so you can just use `animateDismiss`.

src/components/Tips/Tip.tsx
```diff
@@ -1,93 +1,315 @@
...
+    },
+    [dispatch, animateDismiss],
+  )
+
+  const { completion, touchHandlers } = useSwipeToClear({
+    onDismiss: onSwipeDismiss,
+  })
+
+  /** Opacity derived from swipe completion. Linearly decreases from 1 -> 0. */
+  const swipeOpacity = 1 - completion
```
raineorshine: Might as well move this into the `useEffect` since that's the only place it is used.

src/components/Tips/Tip.tsx
```diff
@@ -1,93 +1,315 @@
...
+        opacity={opacity}
+        mask='linear-gradient(to right, transparent, black 60%)'
+      />
+    </div>
+
+    {/* Semi-transparent gradient from bgTransparent → bg, providing the darkening effect.
+        Uses a CSS keyframe animation for fade-in (must be a static string for Panda CSS
+        build-time extraction — dynamic template literals won't generate CSS rules). */}
+    <motion.div
+      style={{ opacity: opacity }}
```
raineorshine: Prefer shorthand:

```ts
  style={{ opacity }}
```

src/components/Tips/Tip.tsx
```diff
@@ -1,93 +1,315 @@
...
+            display: 'flex',
+            userSelect: 'none',
+            position: 'fixed',
+          })}
+        >
+          {/* Blur layer is outside the opacity wrapper because ProgressiveBlur requires
+          its own opacity to work around a Safari backdrop-filter bug. */}
+          <TipBlur opacity={opacity} />
+          {/* Glow and content layers are wrapped in a motion.div driven by opacity. */}
+          <motion.div style={{ opacity: opacity, display: 'flex', width: '100%' }}>
```
raineorshine: Can `display` and `width` be defined in a `className(css(...))` block?

src/components/Tips/Tip.tsx
```diff
@@ -1,93 +1,315 @@
...
+                alignItems: { base: 'flex-start', lg: 'flex-end' },
+                textAlign: { base: 'left', lg: 'right' },
+                padding: '1rem 1.5rem',
+                // Extra top padding creates visual breathing room above the text.
+                paddingTop: '4.5rem',
+                // On devices with safe area insets (notch/home indicator), add the inset to the bottom padding.
+                // On devices without, fall back to 1.5rem. Desktop always uses 1.5rem.
+                paddingBottom: { base: 'max(1.5rem, calc(0.5rem + env(safe-area-inset-bottom)))', lg: '1.5rem' },
               })}
+              style={{ touchAction: 'none' }}
```
raineorshine: Can this be defined in a `className(css(...))` block?

I suspect that Claude may not be aware of our PandaCSS conventions 😄

src/components/Tips/Tip.tsx
```diff
@@ -1,93 +1,315 @@
...
+              {...touchHandlers}
             >
-              {icon}
-              {value}
+              {/* TIP label — uses plus-lighter blend mode for a subtle luminous effect against the gradient. */}
+              <span
+                className={css({
+                  fontSize: '0.85em',
```
raineorshine: For TIP, content, and clear button, is there a reason that fontSize is specified in `em` instead of `rem`?

We're working on porting `em` to `rem` in #3018 so I just want to avoid giving Cindy extra work there. But if there is a hierarchical font size dependency then `em` units may be justified.

src/components/Tips/Tip.tsx
fbmcipher: Oops. Good candidate for the kind of change we'd want the style guide to catch. 67bf0cca8e19789bb3a4e9e56d1f97344d7d1b31

src/components/Tips/Tip.tsx
fbmcipher: 67bf0cca8e19789bb3a4e9e56d1f97344d7d1b31

src/components/Tips/Tip.tsx
fbmcipher: Also noted for the style guide. f9c2dcd35bc9651abb28b143200e542ae9e3dfd4

src/components/Tips/Tip.tsx
fbmcipher: Good catch f9a6d48cb5eb3c7c97fac190024b469ba604e9fe

src/components/Tips/Tip.tsx
fbmcipher: `rem` works too. I think just force of habit on my part. I'll be sure to stick with `rem` units from here on out. 7318adbbc1ee98153e05a8779d9cd172defcaf6c c39383ffed29bad001b342d101e9b0b85b9b7309

src/hooks/useSafeArea.ts
fbmcipher: I missed that. I've noted this all in b4e733695c2c9624fd069fe2dcc8d26bef42c611 for reference if it becomes any issue.

BayuAri:
## [Android] Home button not visible and Tip disappears after user opens then closes keyboard
This might be related to 3918 with a different STR but Tip always disappear when keyboard is onscreen.  
## Step to Reproduce: Newly created thought
1. Create a new thought or tap on any existing thought to trigger keyboard opening
2. Tap on a blank area to dismiss keyboard (not down arrow on virtual button that will cause #3918)

## Step to Reproduce:  with Tip on screen
1. Tap Test Tip button to trigger Tip
2. Create a new thought or tap on any existing thought to trigger keyboard opening
3. Tap on a blank area to dismiss keyboard 

## Current behavior
Home button is not visible on screen.
Home button is moved far down offscreen, user needs to manually scroll down to see it. 
Tip also disappears
When this happens, Test Tip button will not work in trigger the Tip.
User needs to kill the app and reopen to make the Home button visible onscreen without scrolling down.
Note: This does not happen on `main`.
User can simply tap the blank / scroll area to retrigger the Home button appearance when reproducing #3918

**Video:**


https://github.com/user-attachments/assets/b5ab536c-8cad-46a2-97e6-5239f4b82d9d



## Expected behavior
Home button should remain on the button left hand corner.
Tip should reappear when keyboard is dismissed.

BayuAri (commented):
**List of issues:**

- [x] [Tip slides down when keyboard is closed](https://github.com/cybersemics/em/pull/3909#issuecomment-4019832402) is **fixed**
- [x] [Tip + Command Center](https://github.com/cybersemics/em/pull/3909#issuecomment-4019835922) is **fixed**
- [x] [Tip rendered below footer](https://github.com/cybersemics/em/pull/3909#issuecomment-4019836927) is **fixed**
- [x] [[Mobile - Safari] The first tip is too high](https://github.com/cybersemics/em/pull/3909#issuecomment-4021105579) is **fixed**
- [x] [Tip + Sidebar Menu](https://github.com/cybersemics/em/pull/3909#pullrequestreview-3913476373) is **fixed**
- [x] [[iOS] Tip is gone after opening appswitcher and back](https://github.com/cybersemics/em/pull/3909#issuecomment-4059273447)
- [x] [Tip is overlapping with New Thought alert](https://github.com/cybersemics/em/pull/3909#issuecomment-4059356708) is **fixed**
- [x] [Tip and Command Center are overlapping when user Select All or Select Between](https://github.com/cybersemics/em/pull/3909#issuecomment-4059404430) is **fixed**
- [x] [[Desktop] Test Tip button is not working on Chrome and Safari](https://github.com/cybersemics/em/pull/3909#issuecomment-4059577204) is **fixed**
- [x] [Tip is shown on Share page] (https://github.com/cybersemics/em/pull/3909#issuecomment-4066051558)  is **fixed**
- [x] [Tip is shown on Device Management screen](https://github.com/cybersemics/em/pull/3909#issuecomment-4066067516)   is **fixed**
- [x] [Tip is shown on Setting screen](https://github.com/cybersemics/em/pull/3909#issuecomment-4066088999)   is **fixed**
- [x] [The Tip temporary disappears when alert is shown](https://github.com/cybersemics/em/pull/3909#issuecomment-4067548246)   is **fixed** with temporary overlapping Alert and Tip, will be handled on the next PR
- [ ] [Normal and flick swipe do not dismiss the Tip easily](https://github.com/cybersemics/em/pull/3909#issuecomment-4067445179) >> **partially fixed.** 
Fast flick closes the Tip as expected but Normal Swipe does not the Tip easily especially downward. User needs to swipe longer than the height of the Tip's banner wherever the start of the gesture is. If the length of the gesture is shorter then the height, the Tip will reappear. 
Downward siping is harder especially on iOS that will trigger Reachability feature but still not dismissing the Tip. Please see video below, recorded on Android to spot the swiping gesture easier
**Video:**

https://github.com/user-attachments/assets/6cc002c0-a4bf-45a9-9774-c65d224c4eb8


**New Issue:** 

- [ ] [Android] Home button not visible and Tip disappears after user opens then closes keyboard

fbmcipher:
> [iOS] Home button is out of view when user taps Test Tip button when keyboard is onscreen

Thank you both for confirming! I am now able to reproduce this.

The issue is that when "Test Tip" is pressed, `keyboard` remains set to `true`, despite the virtual keyboard closing.

This does not happen when any region of the screen _other_ than "Test Tip" is pressed (empty space, hamburger, or toolbar buttons.)

For this reason, though there is a potential issue here if we choose add additional pressable buttons to the app, it feels out of scope for this PR and doesn't strictly relate to the tip.

fbmcipher:
> Normal swipe is working as expected at the middle of the tips and bottom of the Tip body, however, user needs a longer swiping gesture when starting the swipe from around the edge of the gradient (blur) area of the tip. If not long swipe, the Tip will reappear, can we do something on this ?

Try giving this another go @BayuAri. I think there was a bug that caused the tip to stay open when it should have been closed, which made it harder than expected to close the tip using the swipe gesture. I fixed this in 098fcd06263661da6fbff15962ae9c5cf12f9933.

BayuAri:
## [iOS Capacitor] Tip is auto-dismissed when user closes the app or opens app switcher
## Step to Reproduce - Close the app
1. Tap Test Tip button to trigger the Tip
2. Close the app
3. Reopen the app

## Step to Reproduce - App Switcher 
1. Tap Test Tip button to trigger the Tip
2. Swipe up from the bottom of the screen to open App Switcher
3. Select em app


## Current behavior
Tip is dismissed


https://github.com/user-attachments/assets/f9b29982-c2a4-4ef9-9a12-faefd07f3e6e


## Expected behavior
Tip should remain until user manually dismisses it.
Note: Not reproducible on Android nor Mobile Safari

src/components/Tips/Tip.tsx
```diff
@@ -1,93 +1,339 @@
...
+
+        dismissing.current = true
+        animate(opacity, 0, {
+          duration,
+          ease: 'easeOut',
+          onComplete: () => {
+            dismissing.current = false
+            dispatch(dismissTip())
+          },
+        })
```
raineorshine: This block appears to be identical to `animateDismiss()`.

fbmcipher: Thanks for catching – in f5d70d02fd, I've deduplicated this dismiss animation by adding a `duration` prop to `animateDismiss()`.

fbmcipher:
> I've read the reasoning behind https://github.com/cybersemics/em/commit/098fcd06263661da6fbff15962ae9c5cf12f9933, but I have to say it is very unusual to maintain two state variables with exactly the same value, one reactive and one non-reactive. I've never had to do this in React before. Is it possible that there is a cleaner solution to this problem? Does the value need to be reactive? Thanks.

After a fresh pair of eyes, there is a cleaner way – to use a `MotionValue` for swipe distance instead of state/refs. f151fe63370aa3061c95141d3e648e5ab86a036b

The benefits of this approach:

- We can have one single source of truth for swipe distance. 
- By using a `MotionValue`, the value stays in sync with the animation's state without having to use `onUpdate` to manually keep state/ref in sync with the animation.
- `MotionValue` updates outside of React's render cycle, so we don't trigger accidental Tip re-renders
- We can still use `useMotionValueEvent` to respond to changes to `MotionValue`

fbmcipher:
@BayuAri Thank you as always for your thorough review.

> [iOS Capacitor] Tip is auto-dismissed when user closes the app or opens app switcher

This is fixed by slightly increasing the dead zone at the bottom of the tip on iOS devices with the gesture bar.

> [Android] Home button not visible and Tip disappears after user opens then closes keyboard

This issue is currently being tracked and pending a fix in #3918, so I think it is out of scope for this PR.

(My rationale: The virtual keyboard has some separate issues with its state not properly syncing with Redux's `isKeyboardOpen`. Because the Tip uses `isKeyboardOpen` to determine its visibility, it is affected by these bugs. However, the Tip itself is not the source of these bugs, so they should be fixed elsewhere.)

BayuAri:
- [ ] [[iOS Capacitor] Tip is auto-dismissed when user closes the app or opens app switcher](https://github.com/cybersemics/em/pull/3909#issuecomment-4190952788)
Issue is partially fixed.
With normal speed swipe up, the Tip is persisted when user reopens the app, however, with quick flick swipe up, the Tip is prone to disappear.  
Reproducible on iPhone 14PM iOS 26.2.1 and 16e iOS 18.7.7


**iPhone 14 PM**

https://github.com/user-attachments/assets/08fa8944-7d2b-436b-b875-e414d7264ff3

**iPhone 16e**
It is easier on 16e even with normal swipe up.

https://github.com/user-attachments/assets/f9db8b12-7492-413d-9cc5-7e3b0d94b72f

**Swipe up but stop in the middle of swiping**
Only reproducible on 16e but not on 14PM



https://github.com/user-attachments/assets/f7576d93-d08e-48de-ada1-a844afe1d419

raineorshine:
> > I've read the reasoning behind [098fcd0](https://github.com/cybersemics/em/commit/098fcd06263661da6fbff15962ae9c5cf12f9933), but I have to say it is very unusual to maintain two state variables with exactly the same value, one reactive and one non-reactive. I've never had to do this in React before. Is it possible that there is a cleaner solution to this problem? Does the value need to be reactive? Thanks.
> 
> After a fresh pair of eyes, there is a cleaner way – to use a `MotionValue` for swipe distance instead of state/refs. [f151fe6](https://github.com/cybersemics/em/commit/f151fe63370aa3061c95141d3e648e5ab86a036b)
> 
> The benefits of this approach:
> 
> * We can have one single source of truth for swipe distance.
> * By using a `MotionValue`, the value stays in sync with the animation's state without having to use `onUpdate` to manually keep state/ref in sync with the animation.
> * `MotionValue` updates outside of React's render cycle, so we don't trigger accidental Tip re-renders
> * We can still use `useMotionValueEvent` to respond to changes to `MotionValue`

I don't have as much experience with motion, but if that approach reduces the complexity I'm all for it.

fbmcipher:
@BayuAri Would you mind testing this issue again? 

> [iOS Capacitor] Tip is auto-dismissed when user closes the app or opens app switcher

It looks like this issue occurs on some devices more often than others – I couldn't reproduce on my iPhone 12 mini, or in the simulator – and from your results it looks like it happens more often on the 16e than the 14PM.

That said, I used Copilot to find the most likely cause of the issue and its diagnosis sounds pretty on-point – so I think it should be solved now:

> **The bug**: On iOS Capacitor, the tip was auto-dismissed when users swiped up to go home or open the app switcher.

> **Root cause**: When iOS intercepts a bottom-edge swipe for system navigation, the browser fires a `touchcancel` event — not `touchend`. The hook's `onTouchCancel` was aliased directly to `onTouchEnd`, so iOS system swipes ran the full dismiss scoring logic. The accumulated distance and velocity from the upward swipe easily exceeded the 175px threshold, dismissing the tip even though the user was navigating the OS, not interacting with the app.

> **The fix**: Give `onTouchCancel` its own handler that resets gesture state and springs back — never evaluates the dismiss score. We confirmed with debug logs that iOS does fire `touchcancel` during home/app-switcher swipes.

BayuAri:
@fbmcipher 

- [x] [[iOS Capacitor] Tip is auto-dismissed when user closes the app or opens app switcher](https://github.com/cybersemics/em/pull/3909#issuecomment-4190952788) >>> **Issue is fixed**

- [ ] Tip is not shown on the second attempt after being dismissed
## Step to Reproduce
1. Tap Test Tip button
2. Tap Clear button or swipe up to dismiss the Tip
3. Tap Test Tip button again

## Current behavior
Nothing happens.
Workaround: 
- iOS and Android: Tapping the bottom area of the screen makes Tip shown
- Desktop: Tip is invisible but it is there, it can be shown by the pointer that shows the finger pointer on where the Setting button is located

https://github.com/user-attachments/assets/7d5edc03-d950-4cdd-8833-3f46e7494924


https://github.com/user-attachments/assets/649f3a6b-e032-42a7-b86b-c8333ebf552b


## Expected behavior
Tip should be shown

fbmcipher:
Thanks for catching @BayuAri – fixed now.

BayuAri:
@fbmcipher 

**List of issues:**

- [x] [Tip slides down when keyboard is closed](https://github.com/cybersemics/em/pull/3909#issuecomment-4019832402) is **fixed**
- [x] [Tip + Command Center](https://github.com/cybersemics/em/pull/3909#issuecomment-4019835922) is **fixed**
- [x] [Tip rendered below footer](https://github.com/cybersemics/em/pull/3909#issuecomment-4019836927) is **fixed**
- [x] [[Mobile - Safari] The first tip is too high](https://github.com/cybersemics/em/pull/3909#issuecomment-4021105579) is **fixed**
- [x] [Tip + Sidebar Menu](https://github.com/cybersemics/em/pull/3909#pullrequestreview-3913476373) is **fixed**
- [x] [[iOS] Tip is gone after opening appswitcher and back](https://github.com/cybersemics/em/pull/3909#issuecomment-4059273447)
- [x] [Tip is overlapping with New Thought alert](https://github.com/cybersemics/em/pull/3909#issuecomment-4059356708) is **fixed**
- [x] [Tip and Command Center are overlapping when user Select All or Select Between](https://github.com/cybersemics/em/pull/3909#issuecomment-4059404430) is **fixed**
- [x] [[Desktop] Test Tip button is not working on Chrome and Safari](https://github.com/cybersemics/em/pull/3909#issuecomment-4059577204) is **fixed**
- [x] [Tip is shown on Share page](https://github.com/cybersemics/em/pull/3909#issuecomment-4066051558)  is **fixed**
- [x] [Tip is shown on Device Management screen](https://github.com/cybersemics/em/pull/3909#issuecomment-4066067516)   is **fixed**
- [x] [Tip is shown on Setting screen](https://github.com/cybersemics/em/pull/3909#issuecomment-4066088999)   is **fixed**
- [x] [The Tip temporary disappears when alert is shown](https://github.com/cybersemics/em/pull/3909#issuecomment-4067548246)   is **fixed** with temporary overlapping Alert and Tip, will be handled on the next PR
- [x] [Normal and flick swipe do not dismiss the Tip easily](https://github.com/cybersemics/em/pull/3909#issuecomment-4067445179) >> is **fixed.** 
- [x] [Tip can not be dismissed after receiving another Tip when keyboard is onscreen](https://github.com/cybersemics/em/pull/3909#issuecomment-4059335362) is **fixed**
- [x] [[iOS] Home button is out of view when user taps Test Tip button when keyboard is onscreen](https://github.com/cybersemics/em/pull/3909#issuecomment-4152391957) is **fixed**
- [x] [Variation of swiping on Blur area and down-ward swiping direction](https://github.com/cybersemics/em/pull/3909#issuecomment-4152280398) is **fixed**
- [x] [[iOS Capacitor] Tip is auto-dismissed when user closes the app or opens app switcher](https://github.com/cybersemics/em/pull/3909#issuecomment-4190952788) is **fixed**
- [x] [Tip is not shown on the second attempt after being dismissed](https://github.com/cybersemics/em/pull/3909#issuecomment-4276057931) is **fixed**

PR #3911, opened 2026-03-09
Check parent element in selection.isEditable

### LINKED ISSUE #3805: ColorPicker closes after caret moves to colored word
## Steps to Reproduce

1. Create thought `one two three`.
2. Select the word "two".
3. Open the ColorPicker and choose green.
4. Tap in between "t" and "w".

## Current Behavior

The ColorPicker closes.

## Expected Behavior

The ColorPicker should stay open during edit mode.

ethan-james:
Fixes #3805 

The `Content` wrapper listens for clicks and tests the event target to see if an editable attribute is present on the element in order to short-circuit [clickOnEmptySpace](https://github.com/ethan-james/em/blob/d946620ae9d879bc9ba0caac596e2d82aec1e512/src/components/Content.tsx#L72). When the thought has formatting, the clicked element can be a child of the editable, and so it doesn't have the correct attributes. Since formatting tags can be nested, `selection.isEditable` needs to traverse the parent elements to try to find an editable.

This traversal could also be moved out of `selection.isEditable` into the `fastClick` handler for `Content` if you think that this is an edge case that won't affect other users of `selection.isEditable`.

Does this also need a Puppeteer test?

### DIFF
--- a/src/e2e/puppeteer/__tests__/color.ts
+++ b/src/e2e/puppeteer/__tests__/color.ts
@@ -9,6 +9,7 @@ import getSuperscriptColor from '../helpers/getSuperScriptColor'
 import paste from '../helpers/paste'
 import press from '../helpers/press'
 import setSelection from '../helpers/setSelection'
+import { page } from '../setup'
 
 vi.setConfig({ testTimeout: 60000, hookTimeout: 60000 })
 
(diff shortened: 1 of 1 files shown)


raineorshine:
> The `Content` wrapper listens for clicks and tests the event target to see if an editable attribute is present on the element in order to short-circuit [clickOnEmptySpace](https://github.com/ethan-james/em/blob/d946620ae9d879bc9ba0caac596e2d82aec1e512/src/components/Content.tsx#L72). When the thought has formatting, the clicked element can be a child of the editable, and so it doesn't have the correct attributes. Since formatting tags can be nested, `selection.isEditable` needs to traverse the parent elements to try to find an editable.

Good catch!

> This traversal could also be moved out of `selection.isEditable` into the `fastClick` handler for `Content` if you think that this is an edge case that won't affect other users of `selection.isEditable`.

I think `isEditable` should return true regardless of whether the thought is formatted or not.

> Does this also need a Puppeteer test?

Yes, thanks! We're aiming for as much test coverage as possible of formatting and caret behavior, since that tends to be fragile.

raineorshine (approved):
Thank you! That sounds good to me.

I've created #3915 for further refactoring.

PR #3919, opened 2026-03-11
Consolidate `isThought`, `isOnThought`, `isEditable`, and `isNote`

### LINKED ISSUE #3915: Consolidate `isThought`, `isOnThought`, `isEditable`, and `isNote`
> It looks like `isOnThought` already has ancestor traversal. Could you look at how `isEditable`, `isThought`, and `isOnThought` are used and do some restructuring to ensure that we don't duplicate that logic?
> 
> Or you could craft a tight solution that solves the original issue and then refactor in a separate PR.

I'm fine with trying it here. I think it should be possible to combine `isThought` and `isOnThought` into a single function that calls `isEditable` on the selection `focusNode`. This represents a slight change in behavior because `isThought` was only checking the current `focusNode` and its immediate parent rather than traversing all of the node's parents. My assumption is that this was a buggy implementation that hadn't yet displayed any symptoms, but would have potentially failed on some example like `<b><i>nested</i></b>`.

One thing that I've noticed is that back in bbe278e I updated `isEditable` to return true for notes as well as the main thought. This seems to have gone OK, although there's a separate `isNote` function that is distinct from `isThought`/`isOnThought`. One way of looking at it is that a note is part of a thought, so the selection is still on the thought even if it's in the note. Let me know if you see any reason for concern.

_Originally posted by @ethan-james in https://github.com/cybersemics/em/issues/3911#issuecomment-4034928351_

ethan-james:
Fixes #3915 

One of the caret tests was failing because `isEditable` wasn't handling `TEXT_NODE` type nodes correctly. Now I have replaced `isOnThought` with `isThought`, which passes the `focusNode` straight into `isEditable`.

Now `isEditable` will return true while the caret is on a thought or a note, and `isNote` will only return true if it's on a note.

### DIFF
--- a/src/device/selection.ts
+++ b/src/device/selection.ts
@@ -79,9 +79,10 @@ const isContentEditable = (node: Element | null): boolean => node?.hasAttribute(
 const isNoteEditable = (node: Element | null): boolean => node?.ariaLabel === 'note-editable'
 
 /** Returns true if the node is part of a note. Defaults to using the active selection. */
-export const isNote = (node?: EventTarget | null): boolean =>
-  node === undefined ? isNoteEditable(document.activeElement) : isNoteEditable(getEditableCandidate(node))
-
+export const isNote = (node?: EventTarget | null): boolean => {
+  const editable = node === undefined ? document.activeElement : getEditableCandidate(node)
+  return isNoteEditable(editable)
+}
 /** Returns true if the node is part of a thought. Defaults to using the active selection. */
 // We should see if it is possible to just use state.isKeyboardOpen and selection.isActive()
 export const isThought = (node?: EventTarget | null): boolean => {


ethan-james:
> Okay, I see now. I was misreading that.
> 
> In that case, a little sleuthing is probably in order to ascertain the original intention.
> 
> We can take turns having our brains broken 😂

I'm experiencing a heat wave this week, so it might be my turn for a while 😂

ethan-james:
I updated `cursorDown` to use `selection.isThought()`, and all of the prose view tests are passing. I looked through the code to figure out if there was a way that I could test the new functionality, but I didn't see that prose mode is currently affecting how thoughts are rendered. Does that sound correct?

ethan-james:
> Prose view is currently broken unfortunately, so there is not much we can test against. The `commands/__tests__/proseView` tests only cover the metaprogramming attributes.

That's what I saw, so I guess we can assume that it's OK the way it is for now.

> > I don't see much context for what was intended except for the comment
> > // isActive is not enough on its own, because there is a case where there is a selection object but no focusNode and we want to still execute the command
> > and the fact that they're replacing window.getSelection() with something more targeted. Since they have no impact on the conditions that surround them, it might be worth a try just to remove them.
> 
> It's unfortunate that the comment doesn't point to a specific issue with Steps to Reproduce. Five years have passed and I'm not sure what would produce the case described.
> 
> I would be willing to remove it, but not as part of this PR. It is too risky and doesn't give us fine-grained commit history. If we encounter a regression we will be forced to revert the entire PR. Instead, I suggest preserving the behavior of this one case as-is, perhaps with direct `window.getSelection()` access, and then removing it in a separate PR. That will help mitigate a regression. In the future we can add the necessary puppeteer caret tests.

Sounds like a plan.

src/device/selection.ts
```diff
@@ -56,31 +57,31 @@ export const isCollapsed = (): boolean => !!window.getSelection()?.isCollapsed
...
+export const isNote = (node: EventTarget | null = document.activeElement): boolean =>
+  getEditableCandidate(node)?.ariaLabel === 'note-editable'
 
-/** Returns true if the active element is an editable thought or note. */
-export const isThought = (): boolean =>
-  !!document.activeElement?.hasAttribute('data-editable') || document.activeElement?.ariaLabel === 'note-editable'
+/** Returns true if the node is part of a thought. Defaults to using the active selection. */
+// We should see if it is possible to just use state.isKeyboardOpen and selection.isActive()
+export const isThought = (node: EventTarget | null = document.activeElement): boolean => {
+  const editable = getEditableCandidate(node)
```
raineorshine: `document.activeElement` only points at contenteditable, so there is no need to traverse the ancestors in this case. While it does no harm and the performance impact is probably negligible, I think it would be cleaner to avoid the traversal by default. It avoids mistakenly suggesting that `document.activeElement` can point at a formatting tag.

ethan-james: So useful!

raineorshine (changes requested):
There are a lot of small repetitions in this code that you should be noticing on your own. I see `?.hasAttribute('data-editable')` and `.ariaLabel === 'note-editable'` repeated in multiple places, as well as multiple return statements where a single return statement will do.

Please review the code and only submit after you are confident that it is clean and simplified. That will save us time going back and forth. Thanks so much for your continued attention to detail.

PR #3927, opened 2026-03-12
Don't update thought rank when edited and sorted by Created

### LINKED ISSUE #3782: Sort Picker icon color turns red when Created is selected and user changes font color
## Step to reproduce
1. Create some thoughts
2. Tap Sort Picker button on Toolbar
3. Select Created
4. Tap Font Color button
5. Select any color/highlight

## Current behavior
Sort Picker icon color turns red

Image

**Video:**

https://github.com/user-attachments/assets/9ef0a8d0-1c70-4bd2-983c-e18e952c69ec

## Expected behavior
Sort Picker icon color should remain white indicating the Sorting is active

ethan-james:
Fixes #3782 

When Created and Updated sorts were added in 0676598, the following comment was added to [getStoredRank](https://github.com/cybersemics/em/blob/06765984a740bf317a845a2a97c311ba096f9498/src/selectors/getSortedRank.ts#L19):

```
// For "Updated and Created" sorting, always place at the beginning or end since this thought was just updated
```

This seems like it's correct for Updated, but incorrect for Created because it will move a thought that was created earlier down below a thought that was created later. So I changed it to remove Created from the condition in `getSortedRank`, and in `editThought`. Now Created thoughts don't get their rank updated unless they are re-sorted. I added a component test for editing thoughts with Created sort active.

I also added logic to the Alphabetical sort section of [getSortedRank](https://github.com/ethan-james/em/blob/9caa2de5f48369868307f65a59809d302cadd91e/src/selectors/getSortedRank.ts#L57) that will strip font tags out of the values before comparing them. This prevents color changes from affecting alpha sort in the same way that bold/italic/underline/strikethrough tags will affect it. You can see that in the new `Ascending + color formatting should ignore font tag when re-ranking` test.

### DIFF
--- a/src/actions/newThought.ts
+++ b/src/actions/newThought.ts
@@ -135,12 +135,16 @@ const newThought = (state: State, payload: NewThoughtPayload | string) => {
     return lastChild ? appendToPath(simplePath, lastChild.id) : null
   })
 
+  // if the sort preference is Created, then the current timestamp will be used to sort the new thought into place (#3782)
+  const created = Date.now()
+  const sortPreference = getSortPreference(state, insertId)
+
   // if meta key is pressed, add a child instead of a sibling of the current thought
   // if shift key is pressed, insert the child before the current thought
   const newRank = insertContext
     ? getNextRank(state, ABSOLUTE_TOKEN)
-    : value !== '' && getSortPreference(state, insertId).type === 'Alphabetical'
-      ? getSortedRank(state, insertId, value)
+    : sortPreference.type === 'Created' || (value !== '' && sortPreference.type === 'Alphabetical')
+      ? getSortedRank(state, insertId, value, created)
       : insertBefore
         ? insertNewSubthought || !simplePath || isRoot(simplePath)
           ? getPrevRank(state, insertId, { aboveMeta })
@@ -148,7 +152,7 @@ const newThought = (state: State, payload: NewThoughtPayload | string) => {
         : insertNewSubthought || !simplePath
           ? // if inserting an empty thought into a sorted context via insertNewSubthought, get the rank after the last sorted child rather than incrementing the highest rank
             // otherwise the empty thought will not be correctly sorted by resortEmptyInPlace
-            value === '' && getSortPreference(state, insertId).type === 'Alphabetical' && getLastSortedChildPath()
+            value === '' && sortPreference.type === 'Alphabetical' && getLastSortedChildPath()
             ? getRankAfter(state, getLastSortedChildPath()!)
             : getNextRank(state, insertId)
           : getRankAfter(state, simplePath)


raineorshine:
Maybe it shouldn't be re-ranked, or maybe it's just being re-ranked incorrectly, depending on how it's architected. 

Either way, your test is correct and should pass under the correct behavior.

raineorshine (approved):

BayuAri (commented):
Created Sorting Rank is confirmed not updated when user edits the thought context with text color and copy changes. 
However, there are regression as follow

- [ ] [Issue A: Red Sort icon when user creates a new thought with text during Created Sorting is active](https://github.com/cybersemics/em/pull/3927#issuecomment-4087808842)
- [ ] [Issue B: Text formatting does not affect the Created Sorting Rules](https://github.com/cybersemics/em/pull/3927#issuecomment-4087940130)

src/selectors/getSortedRank.ts
```diff
@@ -21,7 +22,7 @@ const calculateRank = (thoughts: { rank: number }[], index: number): number => {
 }
 
 /** Gets the new rank of a value to be inserted into a sorted context. */
-const getSortedRank = (state: State, id: ThoughtId, value: string) => {
+const getSortedRank = (state: State, id: ThoughtId, value: string, created?: number) => {
   const children = id ? getAllChildrenSorted(state, id) : []
```
raineorshine, line 25: The addition of `created` seems reasonable to me, because without it `getSortedRank` does not have enough information to sort the new value into the correct position.

Please provide a JSDOC and update the function description to explain the usage of `created`.

src/selectors/getSortedRank.ts
```diff
@@ -47,11 +56,14 @@ const getSortedRank = (state: State, id: ThoughtId, value: string) => {
...
+  // Ignore font tags when alphabetically sorting thoughts (#3927)
+  const cleanedValue = value.replaceAll(FONT_TAG_REGEX, '')
+
   // For alphabetical sorting
   const index = children.findIndex(child =>
     isDescending
-      ? compareReasonableDescending(child.value, value) !== -1
-      : compareReasonable(child.value, value) !== -1,
+      ? compareReasonableDescending(child.value.replaceAll(FONT_TAG_REGEX, ''), cleanedValue) !== -1
+      : compareReasonable(child.value.replaceAll(FONT_TAG_REGEX, ''), cleanedValue) !== -1,
```
raineorshine: This appears to duplicate some of the functionality in `compareReasonable` and `compareReasonableDescending`. 

I expect low-level comparison logic to be pushed down to the compare thought util. `getSortedRank` should ideally just identify the sort preference and then delegate to the appropriate comparison function.

src/components/__tests__/SortPicker.ts
```diff
@@ -178,6 +178,152 @@ describe('Created', () => {
...
+      store.dispatch([newThought({ value: '' }), editThoughtByContext([''], 'c')])
+    })
+
+    await act(vi.runOnlyPendingTimersAsync)
+
+    // Get all thoughts in order
+    const thoughts = screen.getAllByLabelText('thought-container')
+    const thoughtValues = thoughts.map(t => t.textContent)
+
+    // Verify thoughts are sorted by creation time ascending (aa, b)
```
raineorshine: I believe this comment should be updated to reflect the new test case.

src/selectors/getSortedRank.ts
ethan-james: Yes, sorry, that should have been removed when I had to add it to `compareReasonable`.

raineorshine (approved):
Looks good!

PR #3792, opened 2026-02-11
New background for the Sidebar with progressive blur and gradient

fbmcipher:
**Resolves #3594**

## Summary

This PR replaces the sidebar’s solid background with a layered progressive blur & gradient, forming the foundation of the new liminal sidebar design.

Instead of a card-style surface, sidebar content now floats over a blurred view of the underlying app with a dark gradient overlay, producing a gentler effect.

This builds on the refactor away from MUI in #3654, and unblocks further development of the sidebar in #3710.

https://github.com/user-attachments/assets/cf43a120-5408-405a-b8e3-6d3aa3925e8c

### Progressive blur component (`ProgressiveBlur.tsx`)
- We use progressive blur throughout the Liminal UI redesign in Figma, but there isn't a simple CSS property for progressive blur.
- Previous implementations used a gradient mask over a single backdrop blur. This approximates progressive blur, but does not *vary* *blur radius* and produced visible flickering artifacts as the sidebar moved.

<details>
<summary>before</summary>

https://github.com/user-attachments/assets/cafd8537-6711-4b43-81c0-07ab45e90b7b

</details>

<details>
<summary>after</summary>

https://github.com/user-attachments/assets/7274ad9d-1401-43cc-a460-a21d79cd233f

</details>

- **Solution**: Developed a new reusable component for progressive blur effects that uses a different technique.
	- Our component is adapted from [this open-source solution](https://github.com/AndrewPrifer/progressive-blur). The general trick: stack multiple `backdrop-filter: blur()` layers, each covering a horizontal slice with a different blur radius.
	- This is a popular workaround to implement progressive blur on the web, used by various other [libraries](https://npmjs.com/package/react-progressive-blur) and [blogs](https://www.devslovecoffee.com/blog/making-apple-progressive-blur-on-web).
	- We make our *own* component, rather than just using the library, to add **extra control** for opacity animations, layer count and removing GPU hacks like `will-change` implemented by the library (which were unnecessary in this case)
- The `layers` prop controls how many blur layers are rendered. Fewer layers = better performance, more layers = more faithful progressive blur. This can be tuned as needed.
- Currently, the component supports progressive blurs in 8 directions (`to right`, `to left`, `to top`, `to bottom`, and diagonals). In the future, it can be extended to support arbitrary directions, which would directly match Figma's progressive blur capabilities.

### Sidebar background
- Replaced the solid grey sidebar background with a layered approach
	- a full-screen dimming overlay
[...]

PR #3936, opened 2026-03-14
[Liminal UI] Sidebar header with animations

### LINKED ISSUE #3710: Sidebar header with animations
> **Replace the tab selector at the top of the sidebar with an animated collapsible switcher.**

Image

## Assets

Image

**Two new assets are used in this design:**

- `overlay-layer-1` (left): takes up 100% of the height and width of the sidebar, using `background-size: cover` and `background-position: top center`.
	- Adds middle tones to the colour blend of the sidebar.
- `overlay-layer-2` (right): takes up 100% of the width of the sidebar, but stays localized to the top portion of the sidebar – positioned underneath the new **collapsible switcher**.

These are available in the [assets folder on Basecamp](https://public.3.basecamp.com/p/TmFdohvobCnQpbMUqntjsw3L/vault/9503618879).

## Collapsible switcher
- This is the primary new feature in this issue.
- Rather than showing **all pages** at once, we show just the currently selected page with a dropdown arrow.
- When tapped, an animation plays – see below. This is from a slightly older design, but the general principles hold true.
- Note:
	- The `overlay-layer-2` image expands in height.
	- The header increases in height smoothly, pushing down the remaining content of the sidebar.
	- The list of pages fades into the increased header space.
	- The dropdown arrow rotates 180 degrees.

https://github.com/user-attachments/assets/d26d4e34-9675-4bf4-ba1c-3a315da20d58

fbmcipher:
**Closes #3710**

## Summary

- Replaces the sidebar's tab selector with an animated collapsible section switcher.
- Adds two new glow overlay layers that shift color per section.
- Refines the sidebar's visual design to match the liminal UI direction.

This builds on the progressive blur & gradient work in #3792, and the sidebar refactor away from MUI in #3654. This mostly completes the implementation of the redesigned sidebar; with just small tweaks to the design of items inside the sidebar remaining.

Care has been taken to ensure interactions remain consistent with the existing sidebar.

This implementation looks best on Capacitor (so that the region underneath the status bar can be utilized), but still works in Mobile Safari & PWA.

CleanShot 2026-03-14 at 09 12 47@2x

### Collapsible section switcher (`SidebarHeader`)
- **Dropdown with animated expand/collapse**: tapping the header toggles a dropdown showing the other two sections that smoothly pushes down and de-emphasizes the content underneath it.
- **Backdrop dismiss**: a full-screen invisible backdrop behind the dropdown closes it on tap.
- **Content obscuring**: while the dropdown is open, the scrollable content below is blurred (on Safari) or dimmed (on Chromium/Firefox) to draw focus to the section picker.

### Glow overlays (`SidebarOverlay1`, `SidebarOverlay2`)
- **Two new overlay layers** sit behind the sidebar content, each using a pre-rendered AVIF image to create a colored glow effect.
- **Overlay 1**: responds to the dropdown's expanded state – the glow intensifies when the dropdown opens.
- **Overlay 2**: adds middle tones to the color blend.
- **Image preloading**: both overlay images are decoded on mount so they appear instantly the first time the sidebar opens.

### Hue-rotate per section (`useSectionHue`)
- Each sidebar section has a defined `hue` and `saturate` value. When the user switches sections, a CSS `hue-rotate()` and `saturate()` filter animates across all overlay layers and the sidebar gradient, shifting the entire color palette.
- Uses **shortest-path calculation** around the 360° color wheel to prevent the animation from going "the long way around" (e.g., 350° to 10° animates +20°, not -340°).

### Responsive breakpoint hook (`useBreakpoint`)
- New reusable hook that returns `true` when the viewport width is at or above a given Panda CSS breakpoint token.
- Uses `viewportStore` (reactive to resize).

### Browser-specific rendering

There are two key browser-specific behaviors:

1. **Blur toggle** (`BLUR_ENABLED`): progressive blur and content blur effects are disabled on Chromium, where they cause performance issues, especially on some Android devices. On Chromium, content dimming (opacity) is used instead when the dropdown is open.
2. **Gradient/blur layer order**: On Safari, the gradient renders _above_ the blur to avoid patchy artifacts. On Chromium, the blur renders _above_ the gradient to avoid visible banding. Both produce visually equivalent results.

### Performance

- Focussed heavily on performance during development. Testing showed consistent 60fps on all of my test devices:
    - iPhone 12 mini (2020)
    - Samsung Z Fold 6 (2024)
    - Chrome: Mac laptop
    - Chrome: Windows tablet

## Tests

- Three sidebar snapshot tests have been updated to reflect the new visual design:
  - `sidebar-empty-light`
  - `sidebar-recently-edited`
  - `sidebar-recently-edited-light`


src/components/Sidebar.tsx
```diff
@@ -10,104 +35,526 @@ import { longPressActionCreator as longPress } from '../actions/longPress'
...
+ * - The scrollable content area below pushes down to make room for the dropdown.
+ *
+ * @param sections - All available sidebar sections.
+ * @param sectionId - The currently active section.
+ * @param onSectionChange - Callback when user selects a different section.
+ * @param isOpen - State determining whether the dropdown is currently expanded.
+ * @param setIsOpen - State setter to toggle the dropdown open/closed.
+ * @param onDropdownHeight - The parent needs to know when the dropdown height changes
+ * so it can adjust the scrollable content area's paddingTop to "push" it down. This
+ * callback reports the dropdown's pixel height whenever it changes.
```
raineorshine: Instead of old style `@param` markings, specify parameter comments inline with the type. They have better Intellisense support, e.g. displaying when hovering over a prop in a component.

```ts
interface SidebarHeaderProps {
  /** All available sidebar sections. */
  sections: SidebarSection[]
  /** The currently active section. */
  sectionId: SidebarSectionId
  ...
```

src/components/Sidebar.tsx
```diff
@@ -10,104 +35,526 @@ import { longPressActionCreator as longPress } from '../actions/longPress'
...
+   * When the dropdown opens or the active section changes, measure the
+   * dropdown's natural height and report it to the parent. The parent uses
+   * this to animate the scrollable content area downward by the same amount,
+   * creating a "push" effect rather than an overlay.
+   */
+  useEffect(() => {
+    if (isOpen && dropdownRef.current) {
+      onDropdownHeight(dropdownRef.current.scrollHeight)
+    }
+  }, [isOpen, sectionId, onDropdownHeight])
```
raineorshine: I'm not a big fan of the push down effect. Too much visual movement.

Sorry that I didn't catch this during our design discussions!

Would it be possible to use an overlay?

src/components/Sidebar.tsx
```diff
@@ -10,104 +35,526 @@ import { longPressActionCreator as longPress } from '../actions/longPress'
...
+             * Fades in/out with opacity. Positioned absolutely below the header
+             * row so it doesn't affect the header's layout.
+             */}
+            <motion.div
+              ref={dropdownRef}
+              initial={{ opacity: 0 }}
+              animate={{ opacity: 1 }}
+              exit={{ opacity: 0 }}
+              transition={{ duration: durations.get('medium') / 1000, ease: EASE_OUT }}
+              style={{
```
raineorshine: Is it possible to use `className(css(...))` here?

src/components/Sidebar.tsx
```diff
@@ -10,104 +35,526 @@ import { longPressActionCreator as longPress } from '../actions/longPress'
...
+        height: '100vh',
+        width: '100vw',
+        backgroundImage: 'url(/img/sidebar/overlay-layer-1.avif)',
+        backgroundPositionX: '-150px', // negative offset to crop the left edge of the image
+        backgroundRepeat: 'no-repeat',
+        pointerEvents: 'none',
+        zIndex: 'sidebar',
+        // fade out the bottom edge of the overlay on portrait mobile for a smoother transition to the background
+        lgDown: {
+          maskImage: 'linear-gradient(to top, transparent 200px, black 80%)',
```
raineorshine: Maybe `bg` instead of `black`?

src/components/Sidebar.tsx
```diff
@@ -10,104 +35,526 @@ import { longPressActionCreator as longPress } from '../actions/longPress'
...
+        backgroundImage: 'url(/img/sidebar/overlay-layer-2.avif)',
+        backgroundSize: '100% 800px',
+        backgroundPosition: 'top left',
+        backgroundRepeat: 'no-repeat',
+        pointerEvents: 'none',
+        zIndex: 'sidebar',
+
+        // on lg+ (landscape mobile and larger) screens, fade off the last 10% vertically to prevent a hard line at the bottom edge
+        lg: {
+          maskImage: 'linear-gradient(to right, black 95%, transparent 100%)',
```
raineorshine: Maybe `bg` instead of `black`?

src/components/Sidebar.tsx
```diff
@@ -450,11 +1050,28 @@ const Sidebar = () => {
...
       <Dialog.Root open={showSidebar} onOpenChange={toggleSidebar} modal={false}>
-        {/* forceMount prop keeps the sidebar mounted when closed.
-        this is temporarily added to match the behavior of the outgoing MUI drawer
-        it can be removed in a later PR to optimize performance */}
+        {/*
+         * forceMount keeps the sidebar mounted in the DOM even when closed.
+         * This is temporarily added to match the behavior of the outgoing MUI drawer.
+         * It can be removed in a later PR to optimize performance, but would require
+         * ensuring that any state within the sidebar subtree is properly preserved.
+         */}
```
raineorshine: 👍

src/components/Sidebar.tsx
```diff
@@ -514,88 +1162,155 @@ const Sidebar = () => {
...
+                     * The 3.75rem top margin provides spacing from the top of the viewport
+                     * (below the safe area inset).
+                     */}
                     <FadeTransition type='fast' in={showSidebar}>
                       <div
                         style={{
-                          // match HamburgerMenu width + padding
-                          marginLeft: fontSize * 1.3 + 30,
+                          marginTop: '3.75rem',
+                          padding: '0 1em',
```
raineorshine: This can be moved to `className(css(...))`

src/hooks/useBreakpoint.ts
```diff
@@ -0,0 +1,10 @@
+import { BreakpointToken, token } from '../../styled-system/tokens'
+import viewportStore from '../stores/viewport'
+
+/** Returns true when the viewport width is at or above the given Panda CSS breakpoint. */
+const useBreakpoint = (breakpoint: BreakpointToken): boolean => {
+  const innerWidth = viewportStore.useSelector(state => state.innerWidth)
+  return innerWidth >= parseInt(token(`breakpoints.${breakpoint}`))
```
raineorshine: You'll want to move this calculation inside the `useSelector`.

Generally we want to aim for thick selectors. That way the component only re-renders when the breakpoint changes, instead of whenever the width changes.

src/util/safeY.ts
```diff
@@ -0,0 +1,4 @@
+/** Helper function that calculates y values that account for the safe area inset. */
```
raineorshine: The comment is overly general and fails to explain what the function really does. A JSDOC can elide implementation details, but shouldn't avoid describing the purpose in precise terms. It should give a developer unfamiliar with the function not just a high level idea of what it is for, but a sense of when they should use it and what the impact will be.

The phrase "account for" is the main culprit. Similar to the word "handle", it doesn't have any explanatory power. I would only use those words as an introduction to a more pointed explanation, if at all.

I would say something like this:

```ts
/** Adds safe-area-inset-top in a CSS calculation to ensure that the y value does not overlap the safe area. */
```

This clarifies that it is a CSS calculation, as opposed to a JS calculation, which is not just an implementation as it affects how the result can be used. It also specifies that it is for top only.

It's a tiny function so it's not the end of the world if the developer has to Go To Definition, but I think the principle is worth pointing out here.

raineorshine (changes requested):
Looks great! Love the color rotation effect.

Please check the Puppeteer tests in the CI when you get a chance.

Here is some initial feedback:

1. There is a dead zone in between each menu option where tapping does nothing. The active link area should be continuous.
2. Add about 10px of extra click space around the menu option links to make the click zone more forgiving.
3. Animate the active menu option highlight color so it is less abrupt.
4. The intensification glow behind the open menu should animate in a little slower, ideally lagging slightly behind the menu options. The glow is a gaseous entity, and gases diffuse slowly, so we want to avoid making any glow movement too snappy.
5. There is an aria warning in the console in Chrome Device Emulator when tapping in the right gutter to close the Sidebar:
   Image

BayuAri (commented):
Great work.
Question please:
Option under the Favorite header still have the grey background and not experiencing the gradient background, is this expected? Or do we have another plan for this?

Image

src/components/Sidebar.tsx
fbmcipher: Generally, I'd agree if we were using `black` as a visible color, but here we're using it as part of a mask. If we use `bg` in the mask, it would change the behavior of the mask between dark mode and light mode (when `bg` became white, it would mark those regions as transparent rather than opaque.)

src/components/Sidebar.tsx
fbmcipher: Generally, I'd agree if we were using `black` as a visible color, but here we're using it as part of a mask. If we use `bg` in the mask, it would change the behavior of the mask between dark mode and light mode (when `bg` became white, it would mark those regions as transparent rather than opaque.)

src/components/Sidebar.tsx
fbmcipher: Done in 5b0e040142a7c97e3333aaf1f47142d110376f89. Some components use inline prop types – in those cases I've written the parameter comments like this – is this OK?

```typescript
const SidebarOverlay1 = ({
  opacity,
  expanded,
  hue,
  sat,
}: {
  /** Shared motion value providing opacity derived from the sidebar's x position. */
  opacity: MotionValue<number>
  /** Whether the dropdown is currently expanded. */
  expanded: boolean
  /** Shared motion value driving CSS hue-rotate on the overlay. */
  hue: MotionValue<number>
  /** Shared motion value driving CSS saturate on the overlay. */
  sat: MotionValue<number>
}) => {
...
```

src/components/Sidebar.tsx
fbmcipher: It's okay, I anticipate some design changes during implementation, especially with interactive elements where it's harder to get a feel of them.

How does this feel? The dropdown fades in rather than pushes down. The gaseous glow fades in at its own pace.

(please note the extreme banding here is only visible in screen recordings!)

https://github.com/user-attachments/assets/2cca89ab-3aa0-40b7-bcb1-411117ee5b87





e05d4c72d93ea7c01ad2ea14c4fe4ca8ee14bfd6

src/components/Sidebar.tsx
fbmcipher: Yes, we can basically use PandaCSS in most cases, including this one.

I've made changes to use the PandaCSS pattern wherever possible, using PandaCSS' variants/recipes in some cases where boolean/dynamic styles are needed. 1ef7ad74d0ad6155c15747b55ed3d7bb361c9a61 20da6064cf67c4000e045b1c675882d0e30a7dc5

src/components/Sidebar.tsx
fbmcipher: 1ef7ad74d0ad6155c15747b55ed3d7bb361c9a61 20da6064cf67c4000e045b1c675882d0e30a7dc5

src/hooks/useBreakpoint.ts
fbmcipher: Thanks, will keep this in mind as I use selectors. 985bfb3c119b5419684ddc298eb6489700130cee

src/util/safeY.ts
fbmcipher: Thank you, this is a really helpful review that I'll keep in mind whenever I write inline documentation. How's d62bf620fb2c0561884f3bff7cfd20815c47f6dd?

src/components/Sidebar.tsx
raineorshine: My bad! Thanks for the correction.

src/components/Sidebar.tsx
```diff
@@ -514,88 +1119,136 @@ const Sidebar = () => {
...
+                      animate={{
+                        ...(BLUR_ENABLED
+                          ? { filter: dropdownOpen ? 'blur(8px)' : 'blur(0px)' }
+                          : { opacity: dropdownOpen ? 0.3 : 1 }),
+                      }}
+                      transition={{ duration: durations.get('medium') / 1000, ease: EASE_OUT }}
+                      data-scroll-at-edge
+                      onScroll={e => {
+                        const scrolled = e.currentTarget.scrollTop > 0
+                        if (scrolled !== isScrolled) setIsScrolled(scrolled)
```
raineorshine: As far as I know, React setters already have this condition built in as a no-op.

src/components/Sidebar.tsx
```diff
@@ -145,32 +563,117 @@ const SidebarBackground = ({
...
+  const sat = useMotionValue(1)
+
+  useEffect(() => {
+    const section = SECTIONS.find(s => s.id === sectionId)!
+    const currentHue = hue.get()
+    // Normalize current hue to [0, 360) then find shortest-path diff in [-180, 180]
+    let diff = section.hue - (((currentHue % 360) + 360) % 360)
+    if (diff > 180) diff -= 360
+    if (diff < -180) diff += 360
+    const t = { duration: durations.get('slow') / 1000, ease: EASE_OUT }
```
raineorshine: You can drop the easing on the hue/sat animation. Easing is meant to approximate non-linear kinematics for moving bodies, which doesn't apply to color. 

(As a gaseous entity, it would technically follow pressure gradients, but I don't think we want to build a fluid dynamic physics engine into our color system 😂).

raineorshine (changes requested):
Thank you for the update!

Here's what I found:

1. Let's vertically center the dropdown caret in the middle of the lowercase glyphs. It looks too high in its current position.
2. The dim/blur mask applied to sidebar content when the sections dropdown is open feels too strong to me. I would go for a much lighter effect.
3. One usability issue I'm noticing is that the order of the menu items are different depending on which section is active. I realize this is how it was designed with the inactive items following the active item, but I think this is a bit unconventional. Dropdowns usually display their items in a consistent order.

fbmcipher:
> The dim/blur mask applied to sidebar content when the sections dropdown is open feels too strong to me. I would go for a much lighter effect.

> One usability issue I'm noticing is that the order of the menu items are different depending on which section is active. I realize this is how it was designed with the inactive items following the active item, but I think this is a bit unconventional. Dropdowns usually display their items in a consistent order.

It sounds like this needs a little bit of a design rethink. I have some ideas to show you at our next design meeting – but until we have a design we're both confident with, I'm going to put this on pause.

fbmcipher:
@raineorshine Could I ask you for your thoughts on these two animations?

These changes were made in response to [this comment](https://github.com/cybersemics/em/pull/3936#pullrequestreview-3970113512).

**Option A:** Two-stage animation. The header slides into its place in the dropdown, and then the remaining options fade in.

https://github.com/user-attachments/assets/b3259f33-95c8-4326-b78d-0b06b41b7aab

**Option B:** Single-stage animation. The header slides into its place in the dropdown at the same time as the other options fade in.

https://github.com/user-attachments/assets/3c29588a-6a5b-4101-8ba1-7c70a5c540c5


I've pushed both into this branch, so you can feel it out directly if you like. You can toggle between both in the "Options" of the sidebar. This is a temporary option and will be removed once we figure out which of these (if either) works better.

raineorshine:
Let's do Option B! Thanks!

If you can fade in the dropdown arrow so it's less abrupt that would be great, too.

src/recipes/sidebarContentMask.ts
raineorshine, line 1: sidebarContentMaskRecipe has no variants and is only used in one place in Sidebar.tsx, so it should be inlined in the `css` call.

Generally we prefer to inline styles so they are co-located with the component. Recipes are a last resort when we have a lot of variants that we need to cleanly define.

src/components/Sidebar.tsx
```diff
@@ -1,113 +1,601 @@
...
+   * imperatively (instead of via animate prop + keyframes) so the selected item's "snap to
+   * 1, hold, snap to 0 at the end of the close" can be expressed as exact zero-duration
+   * `set()` calls. The declarative keyframe form has a non-zero interp window between
+   * adjacent times — even when shrunk to <1ms, an unlucky rAF tick can sample inside it
+   * and produce a sub-frame partial-opacity flicker right as the static header swaps in.
+   * Imperative `set()` has no such window. */
+  const itemOpacities = useRef<Record<string, MotionValue<number>>>({})
+  sections.forEach(s => {
+    if (!itemOpacities.current[s.id]) {
+      itemOpacities.current[s.id] = motionValue(0)
```
raineorshine: Mutating refs in the component body is probably not correct. It looks like `itemOpacities` is never assigned to again? If it is just initializing the ref, then it should get passed to `useRef` as the default value. If it's intended to be a side effect, it should go in `useEffect` or `useLayoutEffect`.

Refs are non-reactive. So using `itemOpacities.current[s.id]` in a `style` attribute below means that it will use the last known value, but will not re-render if the value changes. This kind of dependency is unusual and hard to reason about.

I can kind of follow the JSDOC, but I'm not entirely clear why a ref is needed. It mentions something about a flicker on first render, which usually suggests `useLayoutEffect`. I wonder there may be a more idiomatic React solution that is easier to reason about.

src/components/Sidebar.tsx
```diff
@@ -1,113 +1,601 @@
...
+    if (sectionRow && motionDiv) {
+      const includesMotionDivOffset = sectionRow.offsetParent !== motionDiv
+      return sectionRow.offsetTop + (includesMotionDivOffset ? 0 : motionDiv.offsetTop)
+    }
+    // If the elements aren't mounted yet, estimate based on index * row height
+    const idx = sections.findIndex(s => s.id === sectionId)
+    return Math.max(0, idx) * 54
+  }, [sectionId, sections])
+
+  const selectedOffset = getSelectedOffset()
```
raineorshine: Side effects in the component body are usually wrong. Did you mean to put this in `useLayoufEffect` or `useEffect`?

src/components/Sidebar.tsx
```diff
@@ -1,113 +1,601 @@
...
+          left: 0,
+          zIndex: 1,
+          display: 'flex',
+          flexDirection: 'column',
+          margin: 0,
+          width: '100%',
+          pointerEvents: isOpen ? 'auto' : 'none',
+        })}
+      >
+        {sections.map((s, i) => {
```
raineorshine: This is a lot of logic within a `map`, and the component is already pretty big. I would suggest breaking this out into a subcomponent called `SidebarDropdownItem`.

Factoring this out may also create the opportunity to contain `getSelectedOffset` better, but I'm not sure.

src/components/Sidebar.tsx
```diff
@@ -1,113 +1,601 @@
...
+  // row vs. the selected dropdown item) at exactly the start or end of the window without
+  // an interpolated fade in between.
+  const slide = { duration: STAGE_DURATION, ease: EASE_OUT, delay: 0 }
+  const atStart = { duration: 0, delay: 0 }
+  const atEnd = { duration: 0, delay: STAGE_DURATION }
+
+  /** Refs to each dropdown item's inner div, keyed by section id. Used by getSelectedOffset
+   * below to measure where the selected row currently sits, which determines how far the
+   * selected item must translate up to land on the header row. */
+  const itemEls = useRef<Record<string, HTMLDivElement | null>>({})
```
raineorshine: There is a complex relationship between `itemEls`, `getSelectedOffset`, and the section's motion.div which I'm having some trouble wrapping my head around. Are you sure that `SidebarHeader` needs to individually track item divs? `getSelectedOffset` does not appear to use an accumulator, i.e. logic that depends on all sections at runtime. It seems to only affect the selected item, though maybe I'm missing something.

It would be nice if `selectedOffset` could be push down to the dropdown item level.

src/components/Sidebar.tsx
```diff
@@ -145,32 +632,160 @@ const SidebarBackground = ({
...
+
+  useEffect(() => {
+    // Never interrupt the dropdown animation: while the dropdown is open or mid-close, the
+    // dropdown effect owns mask timing.
+    if (dropdownOpen || dropdownCloseInProgress.current) return
+    animate(scrollHintMaskY, isScrolled ? 0 : SCROLL_HINT_MASK_OFFSET, {
+      duration: STAGE_DURATION,
+      ease: EASE_OUT,
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
raineorshine: Do not disable the linter. If it's absolutely necessary, an explanatory comment is required.

raineorshine (changes requested):
Tested on Chrome and Mobile Safari. The behavior looks great and feels silky smooth.

Just a few questions about implementation.

There is a small merge conflict from merging #4280 first.

src/components/Sidebar.tsx
fbmcipher: You're right, the logic here is pretty hard to look at. I've broken out `SidebarDropdownItem` into its own component, moved its specific logic there, and cleaned up the code substantially.

src/components/Sidebar.tsx
fbmcipher: I was able to substantially refactor here:

- SidebarDropdownItem now measures itself – no more need for itemEls array or getSelectedOffset.
- MotionValues are used to co-ordinate complex animation logic wherever possible – these are passed as props to share animations state between synchronized components.

I hope this is much more readable now.

src/components/Sidebar.tsx
fbmcipher: It should have been in a `useLayoutEffect`, but I fortunately I was able to remove this complex code entirely.

src/components/Sidebar.tsx
fbmcipher: Thank you for this deep review – it was really helpful in finding a better solution to this problem. Admittedly I used AI here to fix the flickering issue mentioned in the JSDoc – it did, but at the expense of some spaghetti and unidiomatic code.

Each dropdown item now owns its opacity via its own `useMotionValue`, and the animation runs in a `useLayoutEffect`. The unidiomatic ref map is eliminated entirely.

raineorshine (approved):
This looks good, thanks.

@BayuAri Will you do another pass of manual testing before we merge? Thanks!

BayuAri:
@fbmcipher 


## Issue A: [Android Capacitor] Flickering glow when changing the dropdown if opening the Sidebar menu while Command Center is active

## Step to Reproduce
1. Create a thought on Android Capacitor app
2. Swipe up to open Command Center
3. Tap Hamburger button on top left hand corner
4. Tap the dropdown next to Favorites
5. Select other menu on the dropdown


## Current behavior
The glow background is flickering when opening the dropdown menu and changing the menu.
Note: this is not reproducible on Mobile Chrome, Safari and iOS Capacitor


https://github.com/user-attachments/assets/ed3cca5c-1230-483a-beca-6ce2b3f3ba04


## Expected behavior
Nothing should happen to the glow background in effect of opening and selecting other menu from the dropdown

fbmcipher:
@BayuAri Good catch, that looks pretty serious. Unfortunately, I can't reproduce this on any of these Android devices. I used BrowserStack to test on older Android versions – please excuse the lag on those recordings, which is due to streaming the real device's screen from BrowserStack's cloud.

### **Google Pixel 6**, Android 16 (physical)
https://github.com/user-attachments/assets/cc454be6-5c10-4a6b-afe9-d459473602ff


### **Samsung Galaxy A35**, Android 14 (BrowserStack)
https://github.com/user-attachments/assets/3b01b39d-cea7-4a04-ab39-0bde3822083b

### **Samsung Galaxy S21 Ultra**, Android 11 (BrowserStack)
https://github.com/user-attachments/assets/b5665dfe-28cc-46c8-bf1c-9d1891bfe3b5

Could you share a little bit more about what Android device and OS version you're using? In addition, what version of Android WebView are you using? I suspect this could have something to do with a WebView version disparity between your test device and what my devices are using.

I'm on Android WebView 148.0.7778.217:

CleanShot 2026-06-17 at 16 24 02@2x

BayuAri:
@fbmcipher 
I am on Samsung A56 OS 16 
Android System Webview Version 149.0.7827.91

image

BayuAri:
@fbmcipher 
Maybe try to add more than 12 favorites thought.

BayuAri:
@fbmcipher 

## Issue B:[Capacitor] Blinking and missing glow layer when the app is sent to background and reopen
## Step to Reproduce
1. Create at least 12 thoughts
2. Add all of them to favorites
3. Swipe up to open Command Center
4. Tap Hamburger button to open Sidebar Menu
5. Close the app
6. Reopen the app


## Current behavior
The glow is blinking. 
Some part of the glow are gone.
Sometime, the Header text and glow layer are not visible.
Note: The blinking glow also happens on iOS but on a much lower scale that only happens when reopen the app after more than 1 min at the background.
 
**Android**

image

**Android**

https://github.com/user-attachments/assets/46ae4c95-68e7-4ad1-8fa1-2da61a3ba2ef


**iOS**

https://github.com/user-attachments/assets/6e0c2e9a-b854-4ff0-bba4-59ce85a09f71


Device:
- Samsung A56 OS 16
- iPhone 14 PM iOS 26.4.2

## Expected behavior
The glow layer and header text should always be visible and not blinking (flickering) when user reopens app from background.

fbmcipher:
Thank you so much @BayuAri, the repro steps were really useful as I was struggling to reproduce beforehand. I've pushed a new version which should fix the flickering – would you mind taking another look?

<details><summary><h2>Changes</h2></summary>

Some of my changes touch code outside of the sidebar, due to the cascading nature of these GPU-based performance issues.

- **37ad954339 – fix Android GPU corruption in active-button glow**
  Activating a Command Center button (e.g. toggling Favorite) corrupted the screen on Android with a smeared black trace following the command glow. The glow was implemented using _two stacked blend layers_, both of which used `filter: blur()`. This was obviously too much for the GPU. Replaced with a single `plus-lighter` layer on _a pre-baked, pre-blurred glow asset_, so no blur at runtime. Visually identical.

- **df99252186 – perf(sidebar): reduce Android GPU load; fix dropdown flicker & Android detection**
  Four independent fixes:

  1. Android now renders 1 progressive-blur (`backdrop-filter`) layer instead of 4 (iOS).
  2. The sidebar subtree no longer stays mounted while closed (Radix `forceMount` removed). It mounts on open, and unmounts after the close slide-out completes, so a closed sidebar costs nothing.
  3. The section-picker dropdown flickered for one frame during open/close because the static header row and the selected dropdown item – two copies of the same row that swap via instant opacity snaps – were scheduled by different mechanisms (declarative `animate` prop vs. imperative animation) and drifted a frame apart; both snaps now run from the same layout effect so they land on the same frame.
  4. Also fixes `isAndroid`, which tested `navigator.platform === 'Linux armv7l'` (32-bit ARM only) and was silently `false` on all modern 64-bit Android, disabling `data-platform=android` and every `_android:` style condition.

- **83a3e2c18 – perf(sidebar): animate Overlay1 dropdown resize via transform, not background-size**
  When the section-picker dropdown opens, the primary glow overlay grows and shifts. This was animated via `background-size`/`background-position`, which are paint-bound properties — the full-viewport (100vw×100vh) layer repainted on every frame of the animation. Now, an inner layer is sized to the collapsed crop and animated with a GPU-composited `scale`/`translate` from the top-left origin, which is mathematically equivalent to the old background-size tween (size = base × scale, so the linear tweens match frame-for-frame). No visual change; the outer layer keeps the crop, mask, and opacity.

- **56f4b30830 – perf(sidebar): cache overlay filters on static layers; animate only opacity/transform**
  The main fix, measured ~40 → ~67fps on a Pixel 6. The two glow overlays and the edge gradient each combined an animated opacity (driven by the drawer's slide position) with a `hue-rotate`/`saturate` filter on the same element. Animating any property on a filtered element forces the filter to re-run every frame, so the slide repainted three near-full-viewport layers per frame instead of compositing them. Each is now split in two: an outer layer that animates opacity only (a plain layer the compositor can fade), and an inner layer holding the static filter + image/gradient/mask, rasterized once and cached — the filter only re-renders on a section change, when the hue actually animates. Also documents in-code why the scroll-area mask keeps its paint-bound `mask-position` animation: the composited alternative (static mask on a translated carrier with counter-translated content) breaks touch scrolling — transforming an actively-scrolled element cancels the gesture — and desyncs under load, flashing the list.

</details> 

## Technical notes

We see flickering artifacts whenever we exceed the GPU's budget for a given frame. We see it more easily on Android than on iOS because there's a much wider range of hardware profiles on Android. Pretty much any iPhone released in the last 6-7 years has a very capable GPU. On the other hand, many mid-range or older high-end Android phones have weaker GPUs on average than iPhones.

This explains why the lagging gets *worse* when the Command Center is open and the Favorite button is glowing. More work for the GPU in the background, so more chance that it runs over-budget and starts tearing/glitching.

Comparing two devices released around the same time – my Google Pixel 6 (2021) and iPhone 12 (2020), the iPhone 12 runs the new sidebar at a smooth 60fps, with no visible glitching. The Pixel 6, on the other hand, lags at around ~30fps and shows the same artifacting issues Bayu reports above. It especially gets worse the more we stack effects on top of each other – blurs on top of blend modes on top of masks. Curiously, the effect I thought would be the worst – `backdrop-filter` – has little or zero impact on either the framerate or the amount of visible glitching.

In contrast my Samsung Z Fold 7 (2025) runs without a hitch and no glitching. So the most at-risk group of devices here seems to be older, or less powerful, Android phones. It's obviously still important to optimize for these, but it's worth knowing what the problematic devices really are.

Going forward, I'll be exercising more restraint on these kinds of styles, and/or gating specific visual effects to iOS. This PR is the most ambitious visual implementation we've taken on so far, so it's been a good lesson to know how much we can get away with and where we need to optimize.

The lag issues on Android devices seem to be a problem *outside* the sidebar, too (general interaction on Pixel 6 is kind of sluggish), so I'll open up a separate ticket to track those issues.

BayuAri:
@fbmcipher 

Flickering issue is still reproducible on my Samsung A56 OS 16.

(https://github.com/cybersemics/em/pull/3936/commits/808b18bd001377d3a1f4f4bb5f9a3791fbaa6d7c)

https://github.com/user-attachments/assets/38f0a874-0e32-4c37-870e-e547ca5ed842


https://github.com/user-attachments/assets/19aaf1c6-33e2-49a5-be67-8462fc6eefb9

The flickering is no longer reproducible on iOS now.  
I am using iPhone 14PM iOS 26.4.2 and iPhone 16e iOS 18.7.8

Tested on [808b18b]


A side note, this is not related to the scope of this PR, but the cursor is offset on the thought's bullet.
Just noting that this regression will not be brought onto the `main`. 
This is only on Android

image

raineorshine:
> A side note, this is not related to the scope of this PR, but the cursor is offset on the thought's bullet.
> Just noting that this regression will not be brought onto the `main`.

For the greatest clarity, please report regressions confidently and clearly with a minimum test case and steps to reproduce. I see other screenshots where the cursor overlay bullet is centered correctly, so you should explain exactly what conditions result in a misaligned bullet. If the problem already exists in `main`, do not mention it here at all, and instead report it in a separate issue. Thanks.

BayuAri:
@fbmcipher 

Updated STR on flickering issue

## Step to Reproduce
1. Create more than 20 thoughts
2. Select All
3. Copy All from Command Center
4. Favorite All from Command Center
5. Tap Hamburger button
6. Change the side bar menu from Favorite to Recently Edited to Recently Deleted


## Current behavior
Android: the glow is flickering.
Tested on Samsung A56 OS 16.

https://github.com/user-attachments/assets/496f6b90-6c15-4666-97ce-87af38c94d52


iOS : The glow crashes eventually after swicthing between the dropdown several times
Reproducible on iPhone 14PM iOS 26.4.2 but not reproducible on iPhone 16e iOS 18.7.8

https://github.com/user-attachments/assets/4bf1b324-d5e8-4f08-915c-ccaccc6e88b3


## Expected behavior
Glow should be stable disregard the dropdown menu changes.

BayuAri:
## Issue C: [Android] Offset cursor overlay on the bullet 
## Step to Reproduce
1. Create a thought on Android 
2. Place the cursor on any thought

## Current behavior
The cursor overlay is not centered on the bullet
Note: this is not reproducible on `main` nor previous commit [7789698](https://github.com/cybersemics/em/pull/3936/commits/778969874e0e49d573c070eb73cf8386f86f57ea)

image

## Expected behavior
The cursor overlay should be centered

fbmcipher:
> [Android] Offset cursor overlay on the bullet

- **Resolved in d599f9f.**
  - **Cause**: Fixed `isAndroid` detection logic in 80818b caused 2019-era Android-only styles to apply to bullet and cursor. These fixes seem to have been implemented to position the bullet in the correct position on Android (back in 2019, a text bullet was used rather than an SVG, so the bullet looked different between iOS/Android and needed different styles per-platform).
  - **Resolution**: This code is no longer necessary (and is actively causing regressions/divergence from `main`), so removed it. Bullet now renders correctly as in `main` and looks identical to iOS, so the 2019 fix no longer seems necessary.


> the glow is flickering
> The glow crashes eventually after swicthing between the dropdown several times

Thanks for updated STR – repro'd. Working on a fix.

raineorshine:
Any update on this one? Thanks!

fbmcipher:
Yes, and thank you both for your patience. Given the severity of some of the performance issues seen here (including out of memory crashes on iOS and tiling/glitching issues on Android) I decided to take an extended break from this submission, before coming back with a fresh perspective to approach the issues from.

**TL;DR**: Performance rises from 20-30fps [in my last comment](https://github.com/cybersemics/em/pull/3936#issuecomment-4862127726) to **90fps on my Pixel 6** test device with no aesthetic compromise.

My **iPhone 12 mini** test device no longer crashes or OOMs.

The main performance issues have been resolved and this is ready to test again @BayuAri. In particular, I'm very interested to see how it performs on the Samsung phone you used.

Please confirm that the performance issues are resolved. Once they are, let me know – I'll do a quick code quality check, deal with conflicts, and then hand back to @raineorshine for (hopefully) the final review before merge. Just want to make sure the performance issues are fully resolved first!

### Technical summary of changes (AI generated)

#### 1. Close the Command Center when the sidebar is open

The glitching was most severe when the Command Center was open at the same time as the Sidebar. Because the Sidebar has a blur backdrop-filter, it has to recompute everything underneath it whenever the browser decides to repaint the Sidebar. Normally this is actually pretty cheap on the GPU, but because the Command Center itself is quite visually complex, this causes lots of repaint work.

The quick-n-dirty solution is **just to hide the Command Center when the Sidebar is open**. This matches the behaviour of the keyboard, which also closes when the sidebar opens.

#### 2. Prebake hue shift in image rather than using `filter: hue-rotate`

To change the colour of the sidebar on each section we used a `hue-rotate()`/`saturate()` filter on full-viewport overlay layers. WebKit allocates a fresh GPU texture buffer **every time a filtered layer (re)renders and releases them lazily** – so switching sections repeatedly grew the GPU process by hundreds of MB until OOM hits and the app crashes.

The tints are now **baked into separate images** that we fade between, which is much lighter computationally and eliminates the issue entirely.

#### 3. Use `transform` on mask instead of `mask-position`

The list content under the open dropdown is hidden by a CSS mask gradient that used to slide via `mask-position`. `mask-position` isn't compositor-animatable in *any* engine – animating it re-rasterized the entire list every frame (~30fps on the Pixel 6, and the source of a lot of the tiling artifacts).

The mask gradient is now **completely static** and the slide is reproduced with a transform pair: the mask-bearing wrapper translates while the scroller inside counter-translates, so the content and its scroll clip stay pixel-stationary while the gradient glides across them — visually identical to the original animation, but both transforms are plain CSS transitions running on the compositor. Zero per-frame repaints. The scroll-hint fade at the top of the list rides the same mechanism, which eliminated the last remaining mask repaint.

#### 4. Refactor dropdown – one single 'title' element instead of swapping between a static/moving pair

The dropdown previously swapped between a static header element and a moving dropdown row on open/close/section-switch. Each swap changed the layer tree mid-animation — elements mounting/unmounting, `will-change` flipping on and off, the icon resizing — and every one of those forces a re-raster that can miss the frame deadline and blank the row for a frame (the flicker on fast section switches).

Now the selected row **is** the persistent title: it slides to the top to become the header when the dropdown closes, and is just the first list item when open. Every row keeps a constant compositing setup — same `will-change`, an always-present transform, one transition, constant 28px icon — so switching sections only changes compositor-side values (transform/opacity) and never restructures layers.

#### 5. `will-change` on Android

Several sidebar layers (the glow overlays, dropdown rows, the blur stack, even the hamburger icon) were being rasterized on demand rather than composited, which is where the remaining one-frame blanks and the low blur framerate came from (measured 51→74fps on the blur alone after promotion). `will-change` pre-promotes these layers so their textures are cached and animations just composite.

> **Note**: Using the new performance metric tooling I built for this PR, **no increased memory usage was observed when adding `will-change`** on Android. I've added it pretty liberally there because it was the only way to force compositor promotion for some layers, and therefore the only route to optimizing performance for layers that were otherwise being rendered on the CPU.

> The memory impact on iOS was the opposite story: promoted layers are precisely what fed the WebKit GPU-process growth behind the jetsam crashes. So every `will-change` in the sidebar is gated to Android (`isAndroid ? … : undefined`) and iOS runs with zero forced promotions — WebKit's own heuristics do the right thing there, and the crossfades composite smoothly without it (measured: no FPS change).

BayuAri:
##  Issue D: Blank area at the bottom of the Side Menu bar when any option is selected.
## Step to Reproduce
1. Create 20 thoughts
2. Favorite them all
3. Tap Hamburger button
4. Tap Favorite dropdown


## Current behavior
A blank layer at the bottom is moving down displaying all of the Favorite list to the bottom


https://github.com/user-attachments/assets/f3675865-135e-48c2-ac1b-6271c9bc14fa

image

## Expected behavior
@fbmcipher 
I am not sure whether this is intended by design or not, however, if this is by design, I would say that it is too high. It supposes to match the height of bottom bar on the Homescreen.

fbmcipher:
@BayuAri Good catch. Should be an easy fix.

BayuAri:
List of issues:
> **Fixed**
- [x] [Issue A: [Android Capacitor] Flickering glow when changing the dropdown if opening the Sidebar menu while Command Center is active](https://github.com/cybersemics/em/pull/3936#issuecomment-4706096544)

- [x] [Issue B: [Capacitor] Blinking and missing glow layer when the app is sent to background and reopen](https://github.com/cybersemics/em/pull/3936#issuecomment-4736898034)

- [x] [Issue C: [Android] Offset cursor overlay on the bullet](https://github.com/cybersemics/em/pull/3936#issuecomment-4880700874)


> **New Issue**
- [ ] [Issue D: Blank area at the bottom of the Side Menu bar when any option is selected.](https://github.com/cybersemics/em/pull/3936#issuecomment-5003361098)

raineorshine:
> **TL;DR**: Performance rises from 20-30fps [in my last comment](https://github.com/cybersemics/em/pull/3936#issuecomment-4862127726) to **90fps on my Pixel 6** test device with no aesthetic compromise.

Great!

> The quick-n-dirty solution is **just to hide the Command Center when the Sidebar is open**. This matches the behaviour of the keyboard, which also closes when the sidebar opens.

That sounds fine to me.

> To change the colour of the sidebar on each section we used a `hue-rotate()`/`saturate()` filter on full-viewport overlay layers. WebKit allocates a fresh GPU texture buffer **every time a filtered layer (re)renders and releases them lazily** – so switching sections repeatedly grew the GPU process by hundreds of MB until OOM hits and the app crashes.

!!!

Perhaps we can gather some of these learnings into a Liminal UI technical guidelines doc so avoid getting bit in the future?

> #### 3. Use `transform` on mask instead of `mask-position`
> The list content under the open dropdown is hidden by a CSS mask gradient that used to slide via `mask-position`. `mask-position` isn't compositor-animatable in _any_ engine – animating it re-rasterized the entire list every frame (~30fps on the Pixel 6, and the source of a lot of the tiling artifacts).

Is the guidance here to never use `mask-position`, never animate `mask-position`, or something more nuanced?

fbmcipher:
> **@BayuAri** Issue D: Blank area at the bottom of the Side Menu bar when any option is selected.

Fixed in 9b4ba56.

> **@raineorshine** Perhaps we can gather some of these learnings into a Liminal UI technical guidelines doc so avoid getting bit in the future?

Good idea, something like that would be useful information for both humans and agents. I'll use agents to put a doc together and run it through it carefully, and submit a PR when it's ready. 

> Is the guidance here to never use `mask-position,` never animate `mask-position`, or something more nuanced?

Something more nuanced: it's fine to use `mask-position` statically, but we should never animate it – because that causes repaints/mask recalculations on each frame.

Instead, keep the *mask* static while animating the position of a _wrapper_ with `transform`. This eliminates the repainting issue while having exactly the same visual result.

This allows the browser to reuse the rasterized layers and update their transform _in the compositor_ (fast and easy GPU translation) rather than repainting/recalculating the mask throughout the animation.

---

I also made a few smaller changes, including:

- replace baked glow images for sidebar with AVIF images (support 12-bit colour depth to reduce chroma banding)
- vastly simplified word-salad AI comments while still remaining detailed and readable
- move "close Command Center when Sidebar opens" behaviour into the CommandCenter rather than in `multicursorAlertMiddleware`
- updated ContextBreadcrumbs unit test to open the sidebar before querying favorites, because the sidebar subtree is now unmounted when closed
- fixed an unspotted bug where the last 10% of the screen horizontally couldn't be used for scrolling in the sidebar (important because the user will likely want to use the scroll zone in the sidebar out of force of habit)

This is ready for one last test @BayuAri – and if that passes, I think it's close to (if not ready for) merging, @raineorshine.

src/components/BulletPositioner.tsx
raineorshine, line 1: Big win if we can delete all this!

@BayuAri Since this has dedicated rules for Android, please test the cursor overlay bullet at each of these screen sizes and make sure it is perfectly aligned with the thought bullet.

- phone: < 500px or 560px (both breakpoints appear to be used)
- tablet: 560–1024px)
- desktop: > 1024px)

src/components/Note.tsx
```diff
@@ -181,13 +181,6 @@ const Note = React.memo(
             position: 'relative',
             marginBottom: '2px',
             padding: '0 0 4px 0',
-            '@media (max-width: 1024px)': {
```
raineorshine, line 189: @BayuAri Please confirm that note positioning on Android phone is unchanged (or improved).

src/components/ThoughtAnnotationWrapper.tsx
```diff
@@ -52,14 +52,6 @@ const ThoughtAnnotationWrapper: FC<
         whiteSpace: 'pre-wrap',
         /* override editable-annotation's single line to have same width with .editable. 100% - 1em since .editable has padding-right 1em */
         maxWidth: ellipsizedUrl ? 'calc(100% - 2em)' : '100%',
-        '@media (max-width: 500px)': {
```
raineorshine, line 55: @BayuAri Could you check that the layout of thought annoations haven't changed on Android phone?

- Url icon position
- Superscript position

src/components/sidebarSectionTints.json
raineorshine, line 1: We don't really have a convention for storing configurations in arbitrary JSON files. Any reason this isn't inlined or in the Panda config?

src/recipes/sidebarMaskGeometry.ts
raineorshine, line 1: I don't think we typically use dedicated constants files in the recipes folder. Inline this if it's only used in one place, or add to `constants.ts` otherwise.

src/recipes/sidebarContentMask.ts
raineorshine, line 1: Prefer inline CSS if possible. From what I can tell this is a single use recipe with only base styles, but correct me if I'm wrong.

raineorshine (changes requested):
Hi, thanks for the update. I did another round of code review.

General observation: The agent seems to be prone to unnecessarily splitting off configs and constants into separate files (defeating co-location), but has the opposite tendency with components, where it allows components to get really huge. The Sidebar component is over 1300 lines of code now. I'm not sure where a reasonable cutoff is, but this would be fairly daunting for another developer to jump into.

src/components/ThoughtAnnotationWrapper.tsx
BayuAri: @raineorshine 
The position of url and email icons as well as superscript are identical with `main`
There is a slight too low url icon that is logged on #4684


image

src/components/BulletPositioner.tsx
BayuAri: @raineorshine 
Bullet is place in the center of the cursor overlay

image

src/components/Note.tsx
BayuAri: @raineorshine 
Note positioning is still intact 

image

BayuAri:
@fbmcipher 
Looks good  on my end.

List of issues:
> **Fixed**
- [x] [Issue A: [Android Capacitor] Flickering glow when changing the dropdown if opening the Sidebar menu while Command Center is active](https://github.com/cybersemics/em/pull/3936#issuecomment-4706096544)

- [x] [Issue B: [Capacitor] Blinking and missing glow layer when the app is sent to background and reopen](https://github.com/cybersemics/em/pull/3936#issuecomment-4736898034)

- [x] [Issue C: [Android] Offset cursor overlay on the bullet](https://github.com/cybersemics/em/pull/3936#issuecomment-4880700874)

- [x] [Issue D: Blank area at the bottom of the Side Menu bar when any option is selected.](https://github.com/cybersemics/em/pull/3936#issuecomment-5003361098)

fbmcipher:
Agreed, the file is pretty long. How would you feel about a pattern like this? This'd mirror a similar structure to other complex components like `CommandCenter` and `Tutorial`:

```
  src/components/
  └── Sidebar/
      |__ Sidebar.tsx              # Redux, dialog lifecycle, composition
      ├── SidebarHeader.tsx        # picker, rows, dropdown animation
      ├── SidebarGlow.tsx          # primary and secondary overlays
      ├── SidebarBackground.tsx    # dimmer, blur, gradient
      ├── sidebarSections.ts       # section types/data and tint helpers
      └── useSidebarSwipe.ts       # touch state machine
```

src/recipes/sidebarMaskGeometry.ts
fbmcipher: Got it, I'll move these over to constants.

src/components/sidebarSectionTints.json
fbmcipher: This is actually an (accidentally committed) artifact of a script I used to "bake" the hue-filter calculations per-section into avif images. I'll remove it in my next commit.

raineorshine:
I like it, thanks. For large components like this, I think it's good for us to establish a convention.

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,526 @@
...
+  // ============================
+  // Refs
+  // ============================
+
+  /** Ref to the drawer element, used to detect if touches are inside the drawer. */
+  const drawerRef = useRef<HTMLDivElement>(null)
+
+  /** Mirror longPressState into a ref so document-level touch handlers always see the current value without re-registering. */
+  const longPressRef = useRef(longPressState)
+  longPressRef.current = longPressState
```
raineorshine: Apologies to give feedback on something that is already in `main`, but because this PR extends the pattern to `isBlockedRef` and `onSwipeEndRef`, I think we have to consider them all together. I think this slipped through in #3792.

---

I'm skeptical of the approach of mirroring Redux state into refs. We have two existing patterns that cover the need for fresh state values:

- If a component's render depends on a fresh value from the Redux state, it should use `useSelector` to get a reactive value. 
- If an event handler in a component depends on a fresh value, it should dispatch a thunk in which it has access to `getState()`.

Mirroring part of the Redux state into a mutable ref creates excess variables and introduces the possibility of values becoming stale. Often it relies on indirect timing like when the render function gets called due to other state changes.

If there is a real need here, or this pattern offers advantages that I'm not aware of, please let me know. Otherwise, I suggest using `useSelector` or `getState`.

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,526 @@
...
+ * Sidebar (root)
+ * ├── SidebarBackground (dimming overlay + progressive blur + gradient)
+ * ├── SidebarGlow (primary glow effect + secondary glow effect)
+ * └── Dialog.Content (the actual drawer panel)
+ * ├── SidebarHeader (section picker with animated dropdown)
+ * │ └── SidebarSectionRow (icon + label)
+ * └── Scrollable content area
+ * ├── Favorites
+ * ├── RecentlyEdited
+ * └── RecentlyDeleted.
```
raineorshine, line 23: This is a good component hierarchy. Each one makes sense and is self-contained.

I think `Sidebar` is still too big. Let me offer my reasoning.

Look how deep it gets, with each level having many parameters and css applied:

```
<Dialog.Root>
  <Dialog.Portal>
    <div data-testid='sidebar'>
      <Dialog.Content>
        ...
          <div aria-label='sidebar'>
            ...
```

This may have worked in `main`, but with the new additions, this structure becomes untenable. By the time we get to the mask gradient, we're so many levels deep that it's hard to read and is not clear how much coupling there is between ancestors and descendants.

More important than reducing file size is reducing complexity. You reduce complexity by defining encapsulation boundaries and hiding implementation details. For example, `maskGradient`, `maskSlideY`, `DROPDOWN_MASK_OFFSET`, `SCROLL_HINT_MASK_OFFSET`, `DROPDOWN_MASK_BAND`, and `MASK_OVERSIZE` are all tightly related and belong together, but is it really necessary for them to live in the top-level `Sidebar` component? Why does the `Sidebar` need to know the exact details of how the mask works? Why does the `Sidebar` need dedicated local state for `isScrolled` when it's only relevant to the mask? 

Push complexity down and keep components on a "need to know" basis. This clarifies coupling, reduces internal API surface area, and creates simplicity out of complexity by hiding implementation details. And readability and a reduction in average file size are a byproduct.

This highlights how adding code is not merely additive; it's multiplicative. At each step up in volume of code there is a step up in complexity that demands a reevaluation of encapsulation boundaries. The deep nesting in Sidebar was already present in `main`, but it didn't become a problem until the Sidebar grew to accommodate the new header animation.

src/components/Sidebar/SidebarBackground.tsx
raineorshine, line 1: This is a good sized component and uses encapsulation effectively to hide the browser-specific component gradient/blur ordering.

fbmcipher:
Thanks — I've split the file up, so this is ready for another review.

`Sidebar.tsx` is down to ~530 lines from ~1340. Still large, but its purpose reads much more clearly now that the visual layers, section data, and gesture handling each live in their own module.

Notably: I deleted the `child.ts` recipe. It existed to position the BulletCursorOverlay bullet on Android, but those adaptations now create misalignments so I removed them earlier in this branch — which left the recipe with an empty `base`. Rather than leave the shell behind, I removed the recipe too.

Let me know if there are any other issues holding this back from merging and I'll do my best to fix them up quickly!

src/constants.ts
```diff
@@ -3,11 +3,56 @@
...
+
+/** Height (px) of the mask's fade-to-black ramp. Doubles as the scroll-hint top fade. */
+export const SCROLL_HINT_FADE = 48
+
+/** Extra carrier extent (px) beyond the scroll viewport — keeps every mask slide position covered. */
+export const MASK_OVERSIZE = DROPDOWN_MASK_BAND + SCROLL_HINT_FADE
+
+/** Offsets that position the mask geometry relative to the scroll area's top edge. */
+export const DROPDOWN_MASK_OFFSET = -DROPDOWN_MASK_BAND
+export const SCROLL_HINT_MASK_OFFSET = -SCROLL_HINT_FADE
```
raineorshine: Maybe there is a compromise here. I didn't like the constants in the recipes folder because I wanted to keep files there strictly limited to recipes. However, it also doesn't really make sense to put them in the global constants file if they only pertain to the Sidebar, or part of the sidebar.

I would suggest the following priority for placing constants:

1. Component module
2. Named component folder (for large components with many subcomponents)
3. Global constants

This is a new precedent since named component folders are a new pattern. I think before Liminal UI they were only used for Editable and Modals. With more complex front-end components, it makes sense to establish a middle layer of organization.

What do you think?

raineorshine (changes requested):
Thank you for the update!

src/constants.ts
fbmcipher: I like it. I think, if I'm reading you right, it's something like this:

- If the constant is just used within a single component: it can be a `const` at the top of that component module.
- If the constant is used within multiple subcomponents in a larger component group: `constants.ts` file _inside_ named component folder, i.e. `src/components/Sidebar/constants.ts`
- If the constant is used throughout the application: global `constants.ts`

The second one is the new one – to go along with the new named component folder pattern. Implemented in [833736d](https://github.com/cybersemics/em/pull/3936/commits/833736d18cc4d9eee07eb793210ad0e92b048afb).

(`constants.ts` currently uses an eslint-disable because it uses named exports. I've exempted files named `constants.ts` from this rule in 129d95e which seemed to me to be cleaner than using eslint-disable each time.)

src/components/Sidebar/Sidebar.tsx
fbmcipher: I think I understand. There were multiple (5+) independent systems in Sidebar.tsx. Some share values, some don't. I guess it's not really clear at a glance what each of those systems is responsible for, or the relationship between them. This really hurts comprehensibility – someone new to the codebase would have a hard time understanding the boundaries between each of these systems and what they were responsible for.

I've done some more cleaning up. Here's the file structure now:

CleanShot 2026-08-05 at 03 24 13@2x

The new stuff:

- `useDrawerPosition`: Owns the `x` value of the drawer, a value which is used by _many_ other components in the Sidebar subtree for animation purposes, plus the open/close `transition` (duration + easing) that drives it.
- `useDrawerWidth`: Owns the `width` value of the drawer — both the raw pixel value used for animation math and the CSS string used for rendering. Another value that's used by many other components in the subtree.
- `SidebarDrawer`: The sliding panel itself (`Dialog.Content` + the animated `motion.div`). Owns its own mount lifecycle and exposes the gestures that can close it (edge-tap, swipe) as plain callbacks, rather than leaking DOM mechanics like event-capture phases into its props.
- `SidebarContent`: The masked, scrollable content area. Owns the mask/scroll-hint geometry and switches between Favorites, Recently Edited, and Recently Deleted.

I paid particular attention to the boundaries between these pieces. Hopefully I haven't _over_-modularized, but at the very least it's much easier to see the different complex systems of the sidebar and how they come together.

The net of all of this is that the return statement of Sidebar is now much more readable:

CleanShot 2026-08-05 at 03 28 43@2x

How does this look now? Do you see any more opportunities for improvement?

src/components/Sidebar/Sidebar.tsx
fbmcipher: Thanks for catching – it's a weird pattern for sure and kind of strange to mirror store state into a ref. In this case reading from the store works just fine and doesn't come with any downsides as far as I can see.

There's a similar "ref-mirroring" approach used in `useSidebarSwipe.ts`, but this time mirroring a _callback_ rather than a store value. The idea is to keep the document-level touch listeners set up once, without tearing them down and re-adding them every time the callback changes, while still always calling the latest version of that callback when a swipe ends. Interestingly React 19.2 has a new primitive for exactly this: `useEffectEvent`, which does the exact same thing. So we can use that instead of ref-mirroring.

As you can imagine this part was vibe-coded. I'm not 100% sure why the agent chose this "ref-mirroring" approach, but I'll keep an eye out for it in the future. I swept the rest of the sidebar feature for the same shape and didn't find any other instances.

src/components/Sidebar/Sidebar.tsx
raineorshine: > I think I understand. There were multiple (5+) independent systems in Sidebar.tsx. Some share values, some don't. I guess it's not really clear at a glance what each of those systems is responsible for, or the relationship between them. This really hurts comprehensibility – someone new to the codebase would have a hard time understanding the boundaries between each of these systems and what they were responsible for.

That's true, but encapsulation is about more than comprehensibility. Functions, hooks, components, and modules all provide lexical scoping which ensures that separate concerns are genuinely inaccessible from other parts of the codebase. When code is hidden behind an abstraction boundary, it forces the interface to be made clear, and guarantees that changes elsewhere cannot directly affect its implementation. When unrelated code is intermixed, you can't assume one way or another; it could be dependent, or it could not be independent.

> How does this look now? Do you see any more opportunities for improvement?

Looks reasonable at first glance. I'll do a more thorough review once it's ready.

src/constants.ts
raineorshine: > * If the constant is used within multiple subcomponents in a larger component group: `constants.ts` file _inside_ named component folder, i.e. `src/components/Sidebar/constants.ts`

Yes, or named exports from the primary component.

fbmcipher:
I think that this is ready for another review @raineorshine. Let me know if you still see issues, particularly surrounding the issue of composition boundaries in the larger Sidebar component.

Feels like valuable work to get this right to establish a clear precedent for the rest of the Liminal UI components ahead.

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,335 @@
...
+  useEffect(() => {
+    if (showSidebar) {
+      document.body.style.overflow = 'hidden'
+    } else {
+      document.body.style.overflow = ''
+    }
+    return () => {
+      document.body.style.overflow = ''
+    }
+  }, [showSidebar])
```
raineorshine: Let's factor this out into a global hook.

src/components/Sidebar/useSidebarSwipe.ts
```diff
@@ -0,0 +1,203 @@
...
+ */
+const useSidebarSwipe = ({ enabled, drawerRef, width, x, onSwipeEnd }: UseSidebarSwipeOptions) => {
+  /** Whether a horizontal swipe is currently in progress. */
+  const [isSwiping, setIsSwiping] = useState(false)
+
+  /** Stable-identity wrapper that always calls the latest onSwipeEnd — lets the touch listeners
+   * below stay registered across renders without depending on the caller's callback identity. */
+  const handleSwipeEnd = useEffectEvent((offset: number, velocity: number) => {
+    onSwipeEnd(offset, velocity)
+  })
```
raineorshine: Prefer point-free style

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+  /** Handle escape key to close sidebar. */
```
raineorshine: If the dropdown is open, the first Escape should close it, and the second Escape should close the modal.

raineorshine (changes requested):
I think this is pretty much good to go. A lot of complexity here, but the top-level Sidebar component is about as thin as it can get. The subcomponents and hooks encapsulate what they can. This had added difficulty because there are so many interdependencies, such as multiple components animating based on the live sliding `x` value, and various events that need to fire or reset at the beginning, end, and threshold points of the drawer state. I think it's a solid implementation given everything that is here.

raineorshine:
> Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
> Co-Authored-By: Claude Opus 5 (unknown context) <noreply@anthropic.com>
> Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
> Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
> Co-authored-by: Raine Revere <raine@cybersemics.org>
 
It was a team effort 😂

PR #3937, opened 2026-03-14
Remove setTimeout from onBlur and rely on relatedTarget

### LINKED ISSUE #3934: [iOS] Keyboard incorrectly opens after Command Center opens and closes
## Steps to Reproduce

1. New Thought
2. Open the Command Center with ↑  gesture
3. Close the Command Center by hitting done
   - Note: The issue cannot be reproduced by closing the keyboard with ↓  gesture
5. New Thought
6. Close the keyboard
7. Tap the first thought
## Current Behavior

The cursor moves to the first thought and the keyboard opens.

## Expected Behavior

The cursor should move to the first thought but the keyboard should not open.

Following the standard behavior, the keyboard should only open when tapping a thought that already has the cursor.

raineorshine:
Fixes #3934

### Cause

In `onBlur`, keyboard dismissal was gated behind a `selection.isThought()` check inside a `setTimeout`:

```ts
setTimeout(() => {
  if (!blurring) return
  if (!selection.isThought()) {
    dispatch(keyboardOpenActionCreator({ value: false }))
  }
})
```

This check was unreliable because `window.getSelection().focusNode` is independent of `document.activeElement`. When a contenteditable loses focus, the browser clears `activeElement` but **does not clear the selection**. The `focusNode` persists pointing at the last caret position inside the editable, making `selection.isThought()` return `true` even after the element has blurred — incorrectly preventing `keyboardOpen(false)` from being dispatched.

This was specifically triggered by the Command Center flow: every command button calls `e.preventDefault()` on touch to keep focus on the editable while the Command Center is open. When Done is tapped, the editable finally blurs — but the stale `focusNode` from the preserved selection caused `selection.isThought()` to incorrectly block the `keyboardOpen(false)` dispatch.

### Why `setTimeout` and `blurring` Existed

The `blurring` flag and `setTimeout` were introduced in 3c89489 to defer the keyboard dismissal, allowing `onFocus` firing on a new editable to cancel the blur handling — distinguishing "focus moved to another editable" from "focus left all editables." This predated the use of `e.relatedTarget`.

`e.relatedTarget` was later introduced in 952767e and already handles the "focus moved to another editable or note" case via the early return:

```ts
if (isRelatedTargetEditableOrNote) return
```

Since `e.relatedTarget` is set synchronously and reliably at blur time, the `setTimeout`, `blurring` flag, and `selection.isThought()` check are all redundant. Removing them and dispatching `keyboardOpen(false)` unconditionally (when `isRelatedTargetEditableOrNote` is false) is both simpler and correct.

### DIFF
--- a/src/e2e/puppeteer/__tests__/caret.ts
+++ b/src/e2e/puppeteer/__tests__/caret.ts
@@ -2,7 +2,6 @@ import { KnownDevices } from 'puppeteer'
 import categorizeCommand from '../../../commands/categorize'
 import newThoughtCommand from '../../../commands/newThought'
 import openCommandCenterCommand from '../../../commands/openCommandCenter'
-import { WindowEm } from '../../../initialize'
 import click from '../helpers/click'
 import clickBullet from '../helpers/clickBullet'
 import clickThought from '../helpers/clickThought'
@@ -369,9 +368,7 @@ describe('mobile only', () => {
     // Step 6: tap the first thought — keyboard should NOT open
     await clickThought('a')
 
-    // isKeyboardOpen should remain false; if it becomes true the test will time out
-    // TODO: Find a less invasive way to confirm the keyboard is closed
-    const em = window.em as WindowEm
-    await waitUntil(() => !em.testHelpers.getState().isKeyboardOpen)
+    // keyboard should not open, so the active element should be the body or null
+    await waitUntil(() => !document.activeElement || document.activeElement === document.body)
   })
 })


src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -340,4 +346,32 @@ describe('mobile only', () => {
...
+    await gesture(newThoughtCommand)
+
+    // Step 5: close the keyboard via the native Done button (blur the active element)
+    await closeKeyboard()
+
+    // Step 6: tap the first thought — keyboard should NOT open
+    await clickThought('a')
+
+    // isKeyboardOpen should remain false; if it becomes true the test will time out
+    // TODO: Find a less invasive way to confirm the keyboard is closed
```
ethan-james: Could possibly add an attribute to [the body](https://github.com/cybersemics/em/blob/67357db1ab012edee1045f955e960d2e45a6d647/src/components/AppComponent.tsx#L138-L139)

raineorshine: Yes, though it would be good to detect it directly from the browser and keep the Redux state a black box.

Looks like `await waitUntil(() => !document.activeElement || document.activeElement === document.body)` also fails in `main` and passes in this branch so I will use that.

Issue #3938, opened 2026-03-14
[iOS] activeElement and focusNode become out of sync after Command Center and keyboard close


raineorshine:
The only known bug related to this right now is #3934 (fixed in #3937), but the general assumption has been that `activeElement` and `focusNode` should be kept in sync.

Either we need to keep them in sync, or abandon the invariant and transition to using `document.activeElement` as the source of truth.

## Steps to Reproduce

1. New Thought
2. Open the Command Center with ↑  gesture
3. Close the Command Center by hitting done
   - Note: The issue cannot be reproduced by closing the keyboard with ↓  gesture
5. New Thought
6. Close the keyboard
7. Log `document.activeElement` and `window.getSelection()?.focusNode` to the console.

## Current Behavior

activeElement: `<body>`
focusNode: `<div class="editable">`

## Expected Behavior

activeElement: `<body>`
focusNode: null

Issue #3933, opened 2026-03-14
Bottom alert is not visible after second New Thought


raineorshine:
Related to [usePositionFixed](https://github.com/cybersemics/em/blob/a1b77143675bf854ea13228bd0b2fe4a98c4e9f0/src/hooks/usePositionFixed.ts)

## Steps to Reproduce

1. New Thought
2. Close the keyboard
3. New Thought

## Current Behavior

The New Thought alert is not visible.

## Expected Behavior

The alert should be positioned above the virtual keyboard, just like after Step 1.

PR #3940, opened 2026-03-14
isThought: Use activeElement instead of focusNode

### LINKED ISSUE #3934: [iOS] Keyboard incorrectly opens after Command Center opens and closes

raineorshine:
Makes #3938 irrelevant
Fixes #3934

### Problem

`selection.isThought()` and `selection.isOnThought()` used `window.getSelection().focusNode` to determine whether an editable thought was focused. However, browsers do not clear `focusNode` when an element is blurred — it persists pointing to the last focused node until a new selection is made elsewhere. This caused `isThought()` to return `true` even after an editable had lost focus (e.g. after closing the Command Center), leading to bugs where `keyboardOpen(false)` was incorrectly suppressed.

`document.activeElement`, by contrast, is authoritative: it is cleared immediately when an element is blurred and reflects the true current focus state.

### Changes

**selection.ts**
- Rewrote `isThought()` to use `document.activeElement` instead of `focusNode`:
  ```ts
  export const isThought = (): boolean =>
    !!document.activeElement?.hasAttribute('data-editable') ||
    (document.activeElement as HTMLElement)?.ariaLabel === 'note-editable'
  ```
- Removed `isOnThought()` — it was a more complex traversal that ultimately checked the same thing; `isThought()` now covers both thoughts and notes.

**commandStateStore.ts**, **ColorPicker.tsx**
- Updated the two call sites of `isOnThought()` to use `isThought()`.

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -62,7 +62,7 @@ const ColorSwatch: FC<{
       (commandStateColor === '#ccccccff' && commandStateBackgroundColor === '#333333ff') ||
       (commandStateColor === addAlphaToHex(rgbToHex(themeColor.fg)) &&
         commandStateBackgroundColor === addAlphaToHex(rgbToHex(themeColor.bg)) &&
-        !selection.isOnThought())
+        !selection.isThought())
     ) {
       const colorMatches = currentThoughtValue.match(colorRegex) || []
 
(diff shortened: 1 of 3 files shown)


raineorshine:
@ethan-james I ran into this when working on #3933 and #3938. What do you think? Partially supercedes #3919.

PR #3410, opened 2025-11-28
Fix for ios specific behavior of caret jumping to beginning/end on tap top/bottom edge

karunkop:
Close #3374 

### Cause:

On iOS devices/ safari browser, when we click in the "void" area around text (e.g., padding, empty space after character), the browser's default behavior places the caret to the end of text. See the image below
image

Initially, this specific issue was believed to be solved in https://github.com/cybersemics/em/commit/1550e4f5cc9d5a52fe3dacf97b732170a8e11573 with clip-path. The commit was dated back to 2022, hence I am not sure whether this actually ever fixed the issue in safari/ios devices as it was not mentioned expliclity. The strong reasoning that this is a ios specific behavior comes from test in other note taking apps like notion where the behavior is found to be the same. i.e caret jumping to the end when clicked outside text vicinity. So even though with clip-path we avoid the dead-zone/gap, the issue still persists because of this reason.

### Possible solution:

As a fix, I have implemented a solution that detects the space the user clicked and mark it as valid or invalid region depending on where the click occured. The function takes the click coordinates and calculates the caret position from it which is determined by `Range` object. `getCharRect` helper  gets a character's visual bounding box. This creates a `Range` spanning exactly one character (from targetOffset to targetOffset + 1), then gets its screen rectangle. This tells us the exact pixel coordinates of any character.
Example: For text "Hello", getCharRect(0) returns the rect for "H", getCharRect(1) returns rect for "e", etc.

Now `isInsideCharRect` checks if click is inside a character's box. If the click lies anywhere outside character's box, it is marked as invalid and we prevent the default behavior of the device/browser via `e.preventDefault()`. If it lies within the boundary of character, default selection is allowed and no behavior is prevented.

PR #3949, opened 2026-03-16
Manual registration of touch and mouse events

karunkop:
As discussed here in https://github.com/cybersemics/em/pull/3410#discussion_r2937235727

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -569,12 +569,11 @@ const Editable = ({
    */
   const handleTapBehavior = useCallback(
     (e: MouseEvent | TouchEvent) => {
-      // When MultiGesture is below the gesture threshold it is possible that onTap and onTouchEnd
+      // When MultiGesture is below the gesture threshold it is possible that onClick and onTouchEnd
       // both trigger. Prevent handleTapBehavior from running a second time via touchend in that case.
       // https://github.com/cybersemics/em/issues/1268
       if (e.type === 'touchend' && globals.touching && e.cancelable) {
         e.preventDefault()
-        return
       }
 
       dispatch((dispatch, getState) => {


src/components/Editable.tsx
```diff
@@ -592,64 +645,41 @@ const Editable = ({
...
-          // dragInProgress: not sure if this can happen, but I observed some glitchy behavior with the cursor moving when a drag and drop is completed so check dragInProgress to be safe
-          (!globals.touching && (!editingOrOnCursor || !isVisible))
-        ) {
-          e.preventDefault()
+    editable.addEventListener('mousedown', onMouseDown)
+    editable.addEventListener('click', onClick)
 
-          if (!isVisible) {
-            selection.clear()
+    if (isTouchSafari) {
```
raineorshine: I think we want to attach the `touchend` handler on all platforms, not just iOS Safari?

raineorshine (changes requested):

src/components/Editable.tsx
```diff
@@ -563,8 +563,60 @@ const Editable = ({
...
+   * Checks long-press, multicursor, disabled, and visibility to decide whether to set the cursor.
+   */
+  const handleTapBehavior = useCallback(
+    (e: MouseEvent | TouchEvent) => {
+      // When MultiGesture is below the gesture threshold it is possible that onTap and onTouchEnd
+      // both trigger. Prevent handleTapBehavior from running a second time via touchend in that case.
+      // https://github.com/cybersemics/em/issues/1268
+      if (e.type === 'touchend' && globals.touching && e.cancelable) {
+        e.preventDefault()
+        return
```
raineorshine: This appears to be different than the same case in `main`, which doesn't have a `return` here. Was this intentional?

src/components/Editable.tsx
```diff
@@ -592,64 +644,36 @@ const Editable = ({
...
       }
-    },
-    [contentRef, editingOrOnCursor, fontSize, allowDefaultSelection, hasMulticursor],
-  )
-
-  /** Sets the cursor on the thought on touchend or click. Handles hidden elements, drags, and editing mode. */
-  const onTap = useCallback(
-    (e: React.MouseEvent | React.TouchEvent) => {
-      // Avoid triggering haptics twice since this handler is used for both onClick and onTouchEnd.
-      if (e.type !== 'touchend') {
```
karunkop, line 603: @raineorshine, based on the current logic, haptics.light() is triggered only when a click event fires. Is that understanding correct?

My concern is that on iOS, when a user performs a tap-and-hold (long press), the synthetic mouse event gets cancelled, meaning events like mousedown and click doesn't fire at all. In that case, haptics.light() would not be triggered.

To account for this, I moved the logic to touchend. However, if the original assumption about click reliably firing is correct, I’ll need to revisit and adjust this part in the refactor.

src/components/Editable.tsx
karunkop: No this was not intentional. I will remove it from the logic.

src/components/Editable.tsx
raineorshine: That makes sense to me. I'm not 100% sure how it behaves in Capacitor but I don't want to block this PR with additional testing across platforms. It's an easy change if we need to tweak it later. Thanks!

Issue #3902, opened 2026-03-06
Note's color reverts into white when user reapply the same background highlight color on Note that is under a thought with background highlight


BayuAri:
## Step to reproduce
1. Create a thought with text
2. Apply any background highlight color
3. Add a note with text
4. Tap Text Color button
5. Select Green Highlight Background button (this could be any color)
6. Verify that the Note's background is green and Note's font color is black
7. Tap Green Highlight Background button again

## Current behavior
Green highlight background reverts to default as black on the Note
Note text font color turns to white.
White Text Color button is marked as active with white border

<img width="684" height="415" alt="Image" src="https://github.com/user-attachments/assets/e0dff4d8-c420-440a-b6ba-470f8a450af0" />

**Video:**

https://github.com/user-attachments/assets/60796ad6-bdf4-42bf-89b8-a37d65896ff0

## Expected behavior
Note background color should revert to default of grey on black background.
No Text Color button is marked as active with white border, not even the Grey Text Color button because the grey color is different with the default

PR #4001, opened 2026-03-28
Remove custom background from Note selection

### LINKED ISSUE #3901: Unable to remove background highlight on Note
## Step to reproduce
1. Create thought "Excepteur Sint Occaecat Cupidatat Non Proident"
2. Add a note "Ut Enim Ad Minim Veniam"
3. Tap Text Color button
5. Verify that the Note's background is green and Note's font color is black
6. Tap Green Highlight Background button again

## Current behavior
Green highlight background remains on the Note and the Note text font color turns to white.
White Text Color and Green Background buttons are marked as active with white border.
Selecting Green Text Color button will make the note invisible due to the same color applied on background and font. 

**Mobile**
Image

**Video:**

https://github.com/user-attachments/assets/0d328b65-9d0a-48a7-b9d0-1aba42d4df45

## Expected behavior
Note background color should revert to default of grey on black background.

Text color should only apply 1 each of the time; text and background colors should not be applied on the same time.

ethan-james:
Fixes #3901 and #3902 (see below)

---

#3901 

`formatSelection` would use the actual DOM element for a note rather than a thought, which is why basic formatting worked, but all of the state that it was checking was for the cursor thought, ignoring notes. Now it checks `state.noteFocus` and uses that to determine whether it needs to fetch and update a note instead of a thought.

I tried to write a component test for this, but ran into the `document.execCommand is not implemented in JSDOM` warning so I'm trying a Puppeteer test.

---

#3902 (merged from #4004)

I added a new `fgNote` color to `colors.config.ts` so that we can reset a note to its default semi-transparent grey color rather than the default white color for thoughts. `ColorPicker` checks `state.noteFocus` to determine which color to send to `formatSelection`. This could be handled inside of `formatSelection` if you would prefer. That could cause merge conflicts with #4001, but it turns out that this depends on that PR anyway.

I would have preferred to remove the color directly, but it seems that doing something like `document.execCommand('removeFormat')` removes all formatting, and there's no option to only remove specific formatting. It turns out that correctly setting the color to its default does cause `document.execCommand('foreColor', ...)` to remove the font tag from the thought or note, which is nice.

I also could have tried a post-processing approach similar to [backColor](https://github.com/ethan-james/em/blob/eb7d601b69ee2bb74a01fe2686cb578bb58b16e7/src/actions/formatSelection.ts#L71). It would be more verbose, but I think that I could make a copy of [shouldRemoveStyle](https://github.com/ethan-james/em/blob/eb7d601b69ee2bb74a01fe2686cb578bb58b16e7/src/actions/formatSelection.ts#L74) that does something like

```
const fgColor = state.noteFocus ? colors.fgNote : colors.fg
const isSameColor = elementColor && rgbToHex(elementColor) === rgbToHex(fgColor)
```

but that still requires adding to `colors.config.ts`.

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -91,17 +91,18 @@ const ColorSwatch: FC<{
     )
   })
 
-  // Note is semi-transparent by default and its color must be reset to that rather than white, which is the fg color for thoughts. (#3902)
-  const fgColor = useSelector(state => (state.noteFocus ? 'fgNote' : 'fg'))
-
   /** Toggles the text color to the clicked swatch. If the swatch is already selected, sets text color and background color back to default. */
   const toggleTextColor = () => {
-    dispatch(
-      formatSelection(
-        'foreColor',
-        selected ? fgColor : color || (backgroundColor && backgroundColor !== 'fg' ? 'black' : 'bg'),
-      ),
-    )
+    dispatch((dispatch, getState) => {
+      // Note is semi-transparent by default and its color must be reset to that rather than white, which is the fg color for thoughts. (#3902)
+      const fgColor = getState().noteFocus ? 'fgNote' : 'fg'
+      dispatch(
+        formatSelection(
+          'foreColor',
+          selected ? fgColor : color || (backgroundColor && backgroundColor !== 'fg' ? 'black' : 'bg'),
+        ),
+      )
+    })
 
     // Apply background color to the selection
     dispatch(formatSelection('backColor', selected ? 'bg' : (backgroundColor ?? 'bg')))


src/components/ColorPicker.tsx
```diff
@@ -91,12 +91,15 @@ const ColorSwatch: FC<{
     )
   })
 
+  // Note is semi-transparent by default and its color must be reset to that rather than white, which is the fg color for thoughts. (#3902)
+  const fgColor = useSelector(state => (state.noteFocus ? 'fgNote' : 'fg'))
```
raineorshine: A reactive value is not needed here. You can access Redux state directly from dispatch's `getState`.

BayuAri (commented):
Test cases:

- [x] Original issue 
- [x] [Reapplied background highlight is clearing the existing background but the color of the Note is white](https://github.com/cybersemics/em/pull/4001#pullrequestreview-4030600065)
- [ ] [Unable to apply the same text / background color of a thought to its note](https://github.com/cybersemics/em/pull/4001#issuecomment-4159427636) 
@ethan-james  @raineorshine 
Should this be handle on an independent issue?

Issue #1957, opened 2024-05-26
Format as code


raineorshine:
Currently `<code>TEST</code>` is correctly formatted as code when rendered in a thought. However, there is no command to format text as code. 

Add a toolbar button to format selected text or the entire thought as code. Follow the same semantics as bold, italic, underline, strikethrough.

PR #4034, opened 2026-04-03
ColorPicker: Allow commandStateStore to handle background-color

### LINKED ISSUE #3903: Note's font color becomes invisible after undoing a background highlight selection
## Step to reproduce
1. Create thought "Lorem Ipsum Dolor Sit Amet, Consectetur Adipiscing Elit"
2. Add note "Ut Enim Ad Minim Veniam"
3. Tap Text Color button
5. Tap Undo button

## Current behavior
Note becomes invisible.
Note's font color is black on black background

Image

**Video:**

https://github.com/user-attachments/assets/4d54f27c-3d0a-4414-bb10-0a41a429b3dc

## Expected behavior
Text's font color should be reverted to default grey

### LINKED ISSUE #3904: Applying Text Color onto the whole thought text is not responded when a partial of the text already has the same text color
## Step to reproduce
1. Create thought "Lorem Ipsum Dolor Sit Amet, Consectetur Adipiscing Elit"
2. Select "Consectetur" 
3. Tap Text Color button
4. Select Green Background Highlight button
5. Verify that the "Consectetur" background is green and font is white
6. Place the caret on the thought text
7. Tap Text Color button
8. Select Green Background Highlight button again

## Current behavior
 "Consectetur" background is reverted to default; black and the text font is white

Note: This also valid for Text Color selection for both Thought and Note text

**Video:**

https://github.com/user-attachments/assets/121d6983-c36e-412e-9c33-1598dd9adb8d

## Expected behavior
The whole thought text should change color into green as user's selection

### LINKED ISSUE #3905: Undoing text background highlight makes the font invisible when it has a partial text with color
## Step to reproduce
1. Create a thought "Lorem Ipsum Dolor Sit Amet, Consectetur Adipiscing Elit, Sed Do Eiusmod Tempor Incididunt Ut Labore Et Dolore Magna Aliqua. Ut Enim Ad Minim Veniam, Quis Nostrud Exercitation Ullamco Laboris Nisi Ut Aliquip Ex Ea Commodo Consequat. Duis Aute Irure Dolor In Reprehenderit In Voluptate Velit Esse Cillum Dolore Eu Fugiat Nulla Pariatur. Excepteur Sint Occaecat Cupidatat Non Proident, Sunt In Culpa Qui Officia Deserunt Mollit Anim Id Est Laborum."   for easy spotting.
2. Select "Deserunt"
3. Tap Text Color button
4. Select Green Background Highlight
5. Select "Dolor" 
6. Select Green Background Highlight again
7. Verify that nothing happens #3904 
8. Select Red Background Highlight (other than green)
9. Verify that the thought text is highlight with red background
10. Tap Undo

## Current behavior
The thought text is invisible except the word from step 2 (Deserunt)  with green background highlight

Image

**Video:**

https://github.com/user-attachments/assets/fe67ab08-72fb-468c-b8b5-8301ae945991

## Expected behavior
The thought text should be white as default except the word from step 2 (Deserunt) with green background highlight

ethan-james:
Fixes #3903, #3904 and #3905

It seems that removing [the background-color regex](https://github.com/ethan-james/em/blob/794a962cd85d1079a6dcd7452cbedbc3e627aa22/src/components/ColorPicker.tsx#L61) fixes this issue pretty effectively. Allowing `commandStateStore` to track the state of formatting at the current caret position means that `selected` will be true while the caret is inside of a formatting tag, and false while it is outside. This may lead to some unexpected behavior and I will follow up with examples next week, but I feel that this behavior is better addressed in `commandStateStore` or possibly in [formatSelection](https://github.com/ethan-james/em/blob/794a962cd85d1079a6dcd7452cbedbc3e627aa22/src/actions/formatSelection.ts#L71) rather than through use of this additional regex.

I tried to remove the `textHexColor` regex as well, but I started to see issues with `selected` being incorrect, so I figure that we can come back to that if this approach seems to be working.

It seemed to work fine in the browser, but the Puppeteer test was showing different behavior until I made a change to the `getCommandState` util. It was matching on any `font` or `span` tag within the text that it was analyzing, and I don't think it makes sense to match `foreColor` or `backColor` unless it wraps the entire thought. It is also possible to have the thought wrapped in other formatting tags (`<b>`, `<i>`, etc.) so I updated the regex to allow any of those tags at the beginning of the matched text, but to reject any text other than a formatting tag in front of the `font` or `span`.

I did test this against #1957 and #2267 to make sure I wasn't creating any regressions. Let me know if there are other issues associated with this functionality that I should check out.


ethan-james:
Here is an example of unexpected behavior that was not possible before fixing this issue, because it wasn't possible to highlight two different sections of a thought using the same color. I recommend that we create a new issue for this.

## Steps to Reproduce

1. Create a thought "hello there world"
2. Double click "hello"
3. Set background color
4. Double click "world"
5. Set same background color
6. Single click "hello"
7. Toggle background color (click same background color again)

### Expected behavior
Either it adds the background color to the entire thought (ignores caret position within formatted text) or else it removes the formatting tag that immediately surrounds the caret (caret-aware command state)

### Current behavior
Removes background color from all selections that have that same color applied

https://github.com/user-attachments/assets/8cc37c1c-e1a3-473b-b362-87d79d20dd0b

ethan-james:
I'm going to add `/skip-tdd` to the unit tests for the reason described above (changed them from a valid-but-possibly-outdated format to a valid-and-standard format) but I noticed a couple of things:

1. TDD Puppeteer tests seem to be missing `action.yml` https://github.com/cybersemics/em/actions/runs/24164197763/job/70522175765?pr=4034
2. Adding `/skip-tdd` as a comment in `__tests__/getCommandState.ts` seems to skip all TDD tests, while I was just trying to skip the unit tests specifically

ethan-james:
This fixes #3905 as well, so I added a test to cover the case of undo restoring previous background colors.

src/e2e/puppeteer/__tests__/color.ts
```diff
@@ -341,3 +341,20 @@ it('Can change the background color of a note to match its thought', async () =>
   const note = await getFirstNoteText()
   expect(note).toBe('<font color="#000000" style="background-color: rgb(255, 87, 61);">Note</font>')
 })
+
+it('Can change the background color of a thought that already has the same background color applied to part of its text', async () => {
+  const text =
+    'a long enough thought where a tap will fall outside of the <font color="#000000" style="background-color: rgb(255, 87, 61);">formatting</font>'
```
raineorshine: This text is a bit confusing, since the thought is not tapped in this test case, and the length of the thought does not appear to be relevant. Maybe the text was from a different example?

raineorshine: Simpler:

```
await paste(`...`)
```

src/util/__tests__/getCommandState.ts
```diff
@@ -135,34 +135,51 @@ it.skip('text and background color on span tag', () => {
...
+    foreColor: undefined,
+    backColor: undefined,
+  })
+})
+
+// /skip-tdd
 it('fully styled thought', () => {
   expect(
     getCommandState(
-      '<b><i><u><strike><code><font color="rgb(255, 0, 0)"><span style="background-color: rgb(0, 0, 255)">text</span></font></code></strike></u></i></b>',
```
raineorshine, line 141: > because it's not possible to create separate tags for foreground and background color using any method that I know.

It is possible to paste valid HTML with separate tags for foreground and background color. Our import logic does not currently have the ability to merge or normalize pasted tags. Also, different platforms may yield different HTML when formatting selected text, so I can't be 100% sure this is safe across iOS and Android. So, it would be best if we could support separate tags.

raineorshine (changes requested):

raineorshine:
It seems I released the new TDD workflow prematurely. Thanks for your patience as I roll out this new part of the CI!

> * TDD Puppeteer tests seem to be missing `action.yml` [cybersemics/em/actions/runs/24164197763/job/70522175765?pr=4034](https://github.com/cybersemics/em/actions/runs/24164197763/job/70522175765?pr=4034)

I will look into this. I'm not sure what it means, since `action.yml` exists.

> * Adding `/skip-tdd` as a comment in `__tests__/getCommandState.ts` seems to skip all TDD tests, while I was just trying to skip the unit tests specifically

These were supposed to be PR comments, not in-code comments, so that was another big oversight. Test code should never be permanently exempt from TDD.

I'll add `/skip-tdd-puppeteer` commands for more granular control.

raineorshine:
> > * TDD Puppeteer tests seem to be missing `action.yml` [cybersemics/em/actions/runs/24164197763/job/70522175765?pr=4034](https://github.com/cybersemics/em/actions/runs/24164197763/job/70522175765?pr=4034)
> 
> I will look into this. I'm not sure what it means, since `action.yml` exists.

Hopefully fixed between #4062 and #4072.

> > * Adding `/skip-tdd` as a comment in `__tests__/getCommandState.ts` seems to skip all TDD tests, while I was just trying to skip the unit tests specifically
> 
> These were supposed to be PR comments, not in-code comments, so that was another big oversight. Test code should never be permanently exempt from TDD.
> 
> I'll add `/skip-tdd-puppeteer` commands for more granular control.

Let's just use labels (#4069).

Added `skip-tdd-unit` and `skip-tdd-puppeteer` in #4075.

src/util/__tests__/getCommandState.ts
ethan-james: I think that if the regex is getting any more complex, then we're better off replacing it with `DOMParser`. I tried it out on desktop Chrome and iOS Safari and it seems to work the same, and it works OK in the unit tests.

raineorshine (approved):
Great!

BayuAri:
Issue #3904 is partially fixed.
Confirmed fixed on Background Color implementation on Text and Note, however, It is still reproducible on Text Font Color and Note Font Color
I will log this as separate comment for test case coverage. 

## Issue C: Color is reset to default when trying to apply the same Font Color onto text with Partial Font Color

```
- Lorem ipsum dolor sit amet
```

## Step to Reproduce
1. Select any word
2. Apply Green Font Color
3. Place the caret onto the thought text
4. Apply Green Font Color for all the text


## Current behavior
Text  Font Color is reset onto its default white font on black background

https://github.com/user-attachments/assets/43fca50d-6806-44e6-8b6a-f8e5c58d40de


## Expected behavior
Font color should be green , following the selection on Step 4.
Refer to the behavior of implementing Background Color on as the main issue of #3904

BayuAri:
## Issue D: Color is reset to default when trying to apply the same Font Color onto Note with Partial Font Color
Similar with Issue C but for Note

```
- Lorem ipsum dolor sit amet
```

## Step to Reproduce
1. Add a Note to the thought 
2. Select any word in the Note
3. Apply Green Font Color
4. Place the caret onto the thought text
5. Apply Green Font Color for all the text


## Current behavior
Text  Font Color is reset onto its default white font on black background


https://github.com/user-attachments/assets/43ab2aec-da9b-43e4-803a-53fc104c5d60


## Expected behavior
Font color should be green , following the selection on Step 5.
Refer to the behavior of implementing Background Color on as the main issue of #3904

BayuAri:
## Issue E: [iOS] Text Color is reset to default when apply any Font Color onto a text with Background Color

```
- Lorem ipsum dolor sit amet
```

## Step to Reproduce
1. Select any word
2. Apply Green Background Color
3. Place the caret anywhere on the thought text
4. Apply Green Background Color
5. Apply any Font Color, for example Red


## Current behavior
Text Color is reverted to the default of white font on black background
Note: Only reproducible on iOS Capacitor and Mobile Safari


https://github.com/user-attachments/assets/3bf2df1a-0ad8-4738-b8e0-fb5f1c1b7ce5


## Expected behavior
Text Font Color should be red or any user selection on Step 5

BayuAri:
## Issue F: [iOS] Note Text Color is reset to default when apply any Font Color onto a text with Background Color
Similar with Issue E but for Note

```
- Lorem ipsum dolor sit amet
```

## Step to Reproduce
1. Add a Note
2. Select any word on the Note
3. Apply Green Background Color
4. Place the caret anywhere on the Note text
5. Apply Green Background Color
6. Apply any Font Color, for example Red


## Current behavior
Note Text Color is reverted to the default of white font on black background.
User needs to tap the Font Color selection 2x to apply it.
PS: Only reproducible on iOS Capacitor and Mobile Safari


https://github.com/user-attachments/assets/8fac9056-2871-494a-b2e0-77b3de108cd1




## Expected behavior
Text Font Color should be red or any user selection on Step 5

BayuAri:
Issue #3905 is partially fixed.
Original STR is confirmed fixed.

## Issue G: [Desktop] Undoing deletes the whole thought when applying the whole  Background Color after applying a partial Background Color 

## Step to Reproduce
1. Start with no existing thought or delete all thought prior starting if any
2. Create a new thought "The world is a great place for adventure"
3. Select any word, for example: "world"
4. Apply Green Background Color
5. Put the caret anywhere on the thought text
6. Apply Green Background Color to all the text
7. Click Undo


## Current behavior
The thought is deleted.


https://github.com/user-attachments/assets/a8e3faa9-e4e2-461d-80d6-2fa9d0fdaa46



## Expected behavior
It should undo the step 6; removing Background Color implementation.

BayuAri:
Original issue of #3905 is reproducible on iOS Capacitor and Mobile Safari with a tweak on the STR

## Issue H: [iOS] Undoing make the text invisible when applying the whole  Background Color after applying a partial Background Color 

## Step to Reproduce
1. Start with no existing thought or delete all thought prior starting if any
2. Create a new thought "The world is a great place for adventure"
3. Select any word, for example: "world"
4. Apply Green Background Color
5. Put the caret anywhere on the thought text
6. Apply Green Background Color to all the text
7. Click Undo


## Current behavior
The text is invisible ;black font on black background, except the selected word on step 3 "world" 


https://github.com/user-attachments/assets/2fe82e04-6c69-4d42-b968-fd0ef659cb2d



## Expected behavior
It should undo the step 6; removing Background Color implementation.

raineorshine:
@ethan-james Looks like there are a variety of related issues here. I will leave it up to you to sort out which ones are genuine regressions, which ones should be new issues, and how you want to divide up the PR.

BayuAri (commented):
List of issues:

- [x] [Applying Text Color onto the whole thought text is not responded when a partial of the text already has the same text color #3904](https://github.com/cybersemics/em/issues/3904)
- [x] [Undoing text background highlight makes the font invisible when it has a partial text with color #3905](https://github.com/cybersemics/em/issues/3905)
- [ ] [Issue C: Color is reset to default when trying to apply the same Font Color onto text with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227618547)
- [ ] [Issue D: Color is reset to default when trying to apply the same Font Color onto Note with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227630914)
- [ ] [Issue E: [iOS] Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227704166)
- [ ] [Issue F: [iOS] Note Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227748288)
- [ ] [Issue G: [Desktop] Undoing deletes the whole thought when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228031188)
- [ ] [Issue H: [iOS] Undoing make the text invisible when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228065337)

ethan-james:
> These were supposed to be PR comments, not in-code comments, so that was another big oversight. Test code should never be permanently exempt from TDD.
> 
> I'll add `/skip-tdd-puppeteer` commands for more granular control.

I see now that the tests were not in fact skipped by the comment that I added to code, but rather by my PR comment mentioning that I was adding a comment to the code.

ethan-james:
> @ethan-james Looks like there are a variety of related issues here. I will leave it up to you to sort out which ones are genuine regressions, which ones should be new issues, and how you want to divide up the PR.

> List of issues:
> 
> * [ ]  [Issue C: Color is reset to default when trying to apply the same Font Color onto text with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227618547)
> * [ ]  [Issue D: Color is reset to default when trying to apply the same Font Color onto Note with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227630914)

These are related to text color rather than background color, which @BayuAri mentioned at the bottom of the initial issue. I stated my intention to come back to them after making sure background color worked. Probably should have done that before taking the PR out of draft. I pushed a commit today that should fix these.

> * [ ]  [Issue E: [iOS] Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227704166)
> * [ ]  [Issue F: [iOS] Note Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227748288)

These are new issues that weren't possible before fixing this issue. They are related to [removeTags](https://github.com/cybersemics/em/blob/fa11f23b469a3656b4198f503058633065f5c6d6/src/actions/formatSelection.ts#L96). I think I'm on track to fix them as part of this PR.

> * [ ]  [Issue G: [Desktop] Undoing deletes the whole thought when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228031188)

This is more closely related to #3551. I may break it out or I may be able to sneak it into this PR.

> * [ ]  [Issue H: [iOS] Undoing make the text invisible when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228065337)

This is slightly different from Issue G but I'll probably keep them together unless I discover that they have radically different causes.

ethan-james:
> List of issues:
> 
> * [x]  [Applying Text Color onto the whole thought text is not responded when a partial of the text already has the same text color #3904](https://github.com/cybersemics/em/issues/3904)
> * [x]  [Undoing text background highlight makes the font invisible when it has a partial text with color #3905](https://github.com/cybersemics/em/issues/3905)
> * [x]  [Issue C: Color is reset to default when trying to apply the same Font Color onto text with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227618547)
> * [x]  [Issue D: Color is reset to default when trying to apply the same Font Color onto Note with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227630914)
> * [x]  [Issue E: [iOS] Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227704166)
> * [x]  [Issue F: [iOS] Note Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227748288)
> * [ ]  [Issue G: [Desktop] Undoing deletes the whole thought when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228031188)
> * [ ]  [Issue H: [iOS] Undoing make the text invisible when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228065337)

@raineorshine Do you want me to try to add Browserstack tests for the iOS issues E and F?

raineorshine:
Yes, that would be great!

ethan-james:
> Issue #3905 is partially fixed. Original STR is confirmed fixed.
> 
> ## Issue G: [Desktop] Undoing deletes the whole thought when applying the whole Background Color after applying a partial Background Color
> ## Step to Reproduce
> 1. Start with no existing thought or delete all thought prior starting if any
> 2. Create a new thought "The world is a great place for adventure"
> 3. Select any word, for example: "world"
> 4. Apply Green Background Color
> 5. Put the caret anywhere on the thought text
> 6. Apply Green Background Color to all the text
> 7. Click Undo
> 
> ## Current behavior
> The thought is deleted.
> 
>  Issue.G.Desktop.Undoing.deletes.the.whole.thought.when.applying.the.whole.Background.Color.after.applying.a.partial.Background.Color.mov 
> ## Expected behavior
> It should undo the step 6; removing Background Color implementation.

OK, after taking a whirlwind tour, we're back where we started. These steps are no longer exactly reproducible after I fixed Issues E and F, but here's what I have:

Screenshot 2026-04-15 at 2 42 40 PM


Line 1: [foreColor](https://github.com/ethan-james/em/blob/fa11f23b469a3656b4198f503058633065f5c6d6/src/components/ColorPicker.tsx#L104) applied to a single word - thought gets longer, `editThought` merged in [undoRedoEnhancer](https://github.com/ethan-james/em/blob/fa11f23b469a3656b4198f503058633065f5c6d6/src/redux-enhancers/undoRedoEnhancer.ts#L257)

Line 2: [backColor](https://github.com/ethan-james/em/blob/fa11f23b469a3656b4198f503058633065f5c6d6/src/components/ColorPicker.tsx#L112) applied to a single word - thought gets longer, `editThought` merged in `undoRedoEnhancer`

Line 3: `foreColor` moved to a new tag to apply to entire thought - thought gets longer, `editThought` merged in `undoRedoEnhancer`

Line 4: `backColor` moved from inner tag to outer tag - thought stays the same length, `undoRedoEnhancer` could be tweaked to merge `editThought` action, but currently doesn't merge

Line 5: Empty `font` tag is removed - thought gets shorter, `editThought` definitely won't be merged by `undoRedoEnhancer`

So my fix for #3551 is too brittle to survive all of these formatting changes. Especially since @BayuAri stated:

> ## Expected behavior
> It should undo the step 6; removing Background Color implementation.

Assuming that @raineorshine agrees that each color formatting step should be treated as its own undo step, then I need to do a couple of things:

1. combine `foreColor` and `backColor` changes into a single `editThought` action (first discussed in https://github.com/cybersemics/em/issues/3551#issuecomment-3968803511)
2. separate `editThought` actions that change colors from `editThought` actions that change text in order to avoid merging them

Does this need to apply to all formatting (bold, italic, etc.) or would it be better to try and merge those updates with other unstyled text updates?

raineorshine:
Thanks for the update!

Color formatting and b, i, u, etc should be separate undoable steps.

> 1. combine `foreColor` and `backColor` changes into a single `editThought` action (first discussed in [Undo color highlighting makes the thought invisible #3551 (comment)](https://github.com/cybersemics/em/issues/3551#issuecomment-3968803511))

Let's discuss the pros and cons of combining actions vs merging patches. I'm not yet sure which approach is going to be better.

To the extent that `formatSelection` relies on `document.execCommand` to do the hard work of determining where to insert the opening and closing tags, I'm not sure it's so easy to reproduce manually. Are you planning on still using `document.execCommand` or replacing it entirely?

Note that the `editThought` action is throttled, so there is potentially an opportunity to skip the first `onThoughtChangeHandler` trigger or somehow handle the double `input` event at the `Editable` level.

> 2\. separate `editThought` actions that change colors from `editThought` actions that change text in order to avoid merging them

`editThought` actions that change colors may be distinguishable by the fact that the plain text value of the thought does not change. I can't think of a case where we would want to merge edits when the plain text value has not changed. Merging edits is meant for contiguous typing or deleting.

ethan-james:
> Let's discuss the pros and cons of combining actions vs merging patches. I'm not yet sure which approach is going to be better.

My thinking is `foreColor` + `backColor` would be the only case where formatting steps are merged. In all other situations, they will be separated so that they can form individual undo steps. It didn't seem like we would want to add a special case in `undoRedoEnhancer` to handle it, although I haven't planned out the step 2 update (separate formatting changes from text changes) so it's possible that the new logic would make it easy to merge `foreColor` + `backColor`.

> To the extent that `formatSelection` relies on `document.execCommand` to do the hard work of determining where to insert the opening and closing tags, I'm not sure it's so easy to reproduce manually. Are you planning on still using `document.execCommand` or replacing it entirely?

I wasn't planning on doing anything to change `execCommand`. Now that I'm using `DOMParser` in various places, it's possible that we're approaching a place where `execCommand` could potentially be replaced, but I haven't given it a lot of thought yet.

> `editThought` actions that change colors may be distinguishable by the fact that the plain text value of the thought does not change. I can't think of a case where we would want to merge edits when the plain text value has not changed. Merging edits is meant for contiguous typing or deleting.

Yeah, that make sense.

raineorshine:
> > Let's discuss the pros and cons of combining actions vs merging patches. I'm not yet sure which approach is going to be better.
> 
> My thinking is `foreColor` + `backColor` would be the only case where formatting steps are merged. In all other situations, they will be separated so that they can form individual undo steps.

Can you clarify how you intend to merge `foreColor` and `backColor` in one step? `document.execCommand` can only apply one color at a time.

ethan-james:
> > > Let's discuss the pros and cons of combining actions vs merging patches. I'm not yet sure which approach is going to be better.
> > 
> > 
> > My thinking is `foreColor` + `backColor` would be the only case where formatting steps are merged. In all other situations, they will be separated so that they can form individual undo steps.
> 
> Can you clarify how you intend to merge `foreColor` and `backColor` in one step? `document.execCommand` can only apply one color at a time.

Oh, right, I guess I can't unless I go around `document.execCommand` and dispatch an `editThought` action directly.

raineorshine:
Yes, exactly. I don't want you to go down that road unless you have a clear idea of how to safely replace `execCommand`.

It's a tradeoff with `execCommand`. We give up applying `foreColor` + `backColor` in a single edit in exchange for native selection and DOM handling.

You still have the possibility of intercepting and merging the edits before they get to the store via `onThoughtChangeHandler`. It just depends on whether you think that or undo patch merging is the cleaner solution.

In summary, our options are:

1. Replace `execCommand` with DOM parsing
2. Merge edits in `Editable` before they get to the store
3. Merge undo patches in `undoRedoEnhancer`

ethan-james:
> Yes, exactly. I don't want you to go down that road unless you have a clear idea of how to safely replace `execCommand`.
> 
> It's a tradeoff with `execCommand`. We give up applying `foreColor` + `backColor` in a single edit in exchange for native selection and DOM handling.
> 
> You still have the possibility of intercepting and merging the edits before they get to the store via `onThoughtChangeHandler`. It just depends on whether you think that or undo patch merging is the cleaner solution.
> 
> In summary, our options are:
> 
> 1. Replace `execCommand` with DOM parsing
> 2. Merge edits in `Editable` before they get to the store
> 3. Merge undo patches in `undoRedoEnhancer`

OK, thanks for the warning. I'll start with the second part, "separate `editThought` actions that change colors from `editThought` actions that change text in order to avoid merging them" and let you know what I end up with before deciding on how to combine the colors.

ethan-james:
> OK, thanks for the warning. I'll start with the second part, "separate `editThought` actions that change colors from `editThought` actions that change text in order to avoid merging them" and let you know what I end up with before deciding on how to combine the colors.

I identified a new bug (#4133) that interferes with forced `editThought` actions like the empty-tag-stripping action in [formatSelection](https://github.com/ethan-james/em/blob/ab8d3595601b90045e786107182ab9dc24b4fdcc/src/actions/formatSelection.ts#L143), but other than than I think I have a handle on separating formatting edits out from text edits. I added a new `mergeWithLast` property to the empty-tag-stripped `editThought` action. Hopefully I can figure out a more elegant solution by the time we're done with this, but it is similar to the existing `mergeUndo` property. `mergeWithLast` flags a new action that should be merged with a previous action, in contrast to `mergeUndo` which looks like it flags an action that should be merged with the **next** action. I think that's necessary because there isn't currently a way to determine whether the empty-tag-stripping action will be dispatched or not.

For the remaining piece,

```
combine foreColor and backColor changes into a single editThought action (first discussed in https://github.com/cybersemics/em/issues/3551#issuecomment-3968803511)
```
I'm thinking through the possibility of detecting `foreColor` + `backColor` in `undoRedoEnhancer`. It might look something like:

1. compare text content of `oldValue` and `newValue` as we now do in [getEditThoughtDirection](https://github.com/ethan-james/em/blob/37d79ed0a9172a6de0292bc2f945fcf5abe40002/src/redux-enhancers/undoRedoEnhancer.ts#L46)
2. if the text content is the same, then query `font`/`span` tags within [oldElement](https://github.com/ethan-james/em/blob/37d79ed0a9172a6de0292bc2f945fcf5abe40002/src/redux-enhancers/undoRedoEnhancer.ts#L49) and [newElement](https://github.com/ethan-james/em/blob/37d79ed0a9172a6de0292bc2f945fcf5abe40002/src/redux-enhancers/undoRedoEnhancer.ts#L50)
3. iterate through the tags to find any where `textContent` and `color` attribute match, but `newElement` has a new style tag
4. flag for merging if there's a match

The other case that I can think of is when `foreColor` has just been called and the thought is in an intermediate state like:

```
<font color=\"#000000\">hello <font style=\"background-color: rgb(0, 214, 136);\">world</font></font>
```

That needs to be merged with the next action (apply background-color to outer tag, remove from inner tag) rather than the previous action where both color and background-color were on the inner tag.

An alternative approach would be to introduce the idea of a "transaction", possibly to `commandStateStore`. `Editable` could watch to see if a transaction is in progress (begun but not committed) and tag the `editThought` actions that it dispatches with some sort of transaction ID that makes them easy to merge. Sounds like a nice idea to me, but we don't need to devote any time to it unless you think it's great. I'll default to the plan of merging in `undoRedoEnhancer`.

raineorshine:
> I added a new `mergeWithLast` property to the empty-tag-stripped `editThought` action. Hopefully I can figure out a more elegant solution by the time we're done with this, but it is similar to the existing `mergeUndo` property. `mergeWithLast` flags a new action that should be merged with a previous action, in contrast to `mergeUndo` which looks like it flags an action that should be merged with the **next** action. I think that's necessary because there isn't currently a way to determine whether the empty-tag-stripping action will be dispatched or not.

Sounds good. Since we already have `mergeUndo`, I think it makes sense to have a way to merge in the other direction. 

Let's call them `mergeNext` and `mergePrev` to create some semantic symmetry. (git's `squash` and `fixup` also comes to mind, but I always get those confused, so I'm not prepared to invest in that naming.)

> I'm thinking through the possibility of detecting `foreColor` + `backColor` in `undoRedoEnhancer`. It might look something like:
> 
> 1. compare text content of `oldValue` and `newValue` as we now do in [getEditThoughtDirection](https://github.com/ethan-james/em/blob/37d79ed0a9172a6de0292bc2f945fcf5abe40002/src/redux-enhancers/undoRedoEnhancer.ts#L46)
> 2. if the text content is the same, then query `font`/`span` tags within [oldElement](https://github.com/ethan-james/em/blob/37d79ed0a9172a6de0292bc2f945fcf5abe40002/src/redux-enhancers/undoRedoEnhancer.ts#L49) and [newElement](https://github.com/ethan-james/em/blob/37d79ed0a9172a6de0292bc2f945fcf5abe40002/src/redux-enhancers/undoRedoEnhancer.ts#L50)
> 3. iterate through the tags to find any where `textContent` and `color` attribute match, but `newElement` has a new style tag
> 4. flag for merging if there's a match

This would work, but I think we should think about division of responsibilities. `undoRedoEnhancer` is designed to undo/redo any action based on the state diff payload. I don't think it should be involved in knowing the details of how different actions work. Arguably, `editThoughtDirection` is already a violation of this principle, and `mergeNext`/`mergePrev` is the proper generalization of this concept that allows action creators to handle the action-specific details.

Second, `formatSelection` does all the hard work of determining whether to apply foreColor, backColor, or both. It would be odd to add detection logic further downstream when `formatSelection` already knows everything that is going to be applied. The question is is just how to pass that information down to the `undoRedoEnhancer`.

The real problem is that we are using `document.execCommand` which is not composable. I could be wrong, but the benefits still seem to outweigh the risk of implementing our own selection handling at this time. Unfortunately the flow between `document.execCommand` → input event → `onChangeHandler` means that we can't pass the merge intention directly. We need a back channel... which means either a global variable or a ministore (like `commandStateStore` as you suggest).

> An alternative approach would be to introduce the idea of a "transaction", possibly to `commandStateStore`. `Editable` could watch to see if a transaction is in progress (begun but not committed) and tag the `editThought` actions that it dispatches with some sort of transaction ID that makes them easy to merge. Sounds like a nice idea to me, but we don't need to devote any time to it unless you think it's great. I'll default to the plan of merging in `undoRedoEnhancer`.

This would be a nice general solution for complex/heterogeneous batches of actions, however I don't think our situation demands that level of complexity at this point. The separate "begin" and "commit" steps that are characteristic of transaction systems would need additional handling. Rather, we just have a series of homogeneous, contiguous `editThought` actions that need a small cue to be able to set `mergePrev` or `mergeNext` appropriately.

Hopefully that helps conceptualize the problem and the encapsulation concerns. Thanks and me know if you have any questions!

ethan-james:
> This would work, but I think we should think about division of responsibilities. `undoRedoEnhancer` is designed to undo/redo any action based on the state diff payload. I don't think it should be involved in knowing the details of how different actions work. Arguably, `editThoughtDirection` is already a violation of this principle, and `mergeNext`/`mergePrev` is the proper generalization of this concept that allows action creators to handle the action-specific details.

It sounds like it might be worthwhile to try to move `editThoughtDirection` out of `undoRedoEnhancer`, possibly into a hook that's consumed by `Editable` and determines whether to add `mergePrev` to the actions that are being dispatched in `thoughtChangeHandler`. I could create a new issue for this, or else try to roll it into #3903.

> We need a back channel... which means either a global variable or a ministore (like `commandStateStore` as you suggest).

I added a `batchEditing` ministore that functions in a similar way to the existing `suppressFocus` ministore. Let me know what you think of that solution.

raineorshine:
> > This would work, but I think we should think about division of responsibilities. `undoRedoEnhancer` is designed to undo/redo any action based on the state diff payload. I don't think it should be involved in knowing the details of how different actions work. Arguably, `editThoughtDirection` is already a violation of this principle, and `mergeNext`/`mergePrev` is the proper generalization of this concept that allows action creators to handle the action-specific details.
> 
> It sounds like it might be worthwhile to try to move `editThoughtDirection` out of `undoRedoEnhancer`, possibly into a hook that's consumed by `Editable` and determines whether to add `mergePrev` to the actions that are being dispatched in `thoughtChangeHandler`. I could create a new issue for this, or else try to roll it into #3903.

Yes, I agree. Let's refactor `editThoughtDirection` in a separate PR once this is merged.

> > We need a back channel... which means either a global variable or a ministore (like `commandStateStore` as you suggest).
> 
> I added a `batchEditing` ministore that functions in a similar way to the existing `suppressFocus` ministore. Let me know what you think of that solution.

Looks good to me!

ethan-james:
> Yes, that would be great!

I'm having a little difficulty with the Browserstack tests. It doesn't seem to be able to open the color picker menu, and I've run out of free minutes on my Browserstack account. Please let me know if you have any tips, thanks!

raineorshine:
@karunkop Can you help Ethan with the browserstack tests? Thanks!

ethan-james:
> Yes, I agree. Let's refactor `editThoughtDirection` in a separate PR once this is merged.

Sounds good. The work that I was attempting in https://github.com/cybersemics/em/pull/4060 would also be something that would be refactored, so I can do that in a separate PR as well. Since we have already developed `mergePrev` here, I think it makes sense to add #3903 to this PR as well.

I was trying to solve a related issue in #4060, where longer/shorter edits in notes are not merged like they are in thoughts. So I can tackle that separately and just fix #3903 here.

raineorshine:
Sounds good. Thanks for keeping track of all the moving parts!

karunkop:
> > Yes, that would be great!
> 
> I'm having a little difficulty with the Browserstack tests. It doesn't seem to be able to open the color picker menu, and I've run out of free minutes on my Browserstack account. Please let me know if you have any tips, thanks!

Quick update on this issue: it appears that the toolbar buttons are not being triggered by the current implementation in tap.ts (iOS helper).

At first, I suspected this might be due to the “Text Color” icon not being visible on mobile, requiring a scroll action before interaction. However, further testing showed that even toolbar icons already visible on screen are not responding to the tap actions. I also explored a few alternative APIs from WebdriverIO, but none seem to work.

As a workaround, I implemented a helper function that directly interacts with the DOM. It waits for the target element using the appropriate selector and then dispatches mouse events programmatically:

```ts
node.dispatchEvent(new MouseEvent('mousedown', { button: 0 }))
node.dispatchEvent(new MouseEvent('mouseup', { button: 0 }))
node.click()
```

This approach resolves the issue functionally. That said, it bypasses the standard user interaction flow by triggering internal DOM events rather than simulating an actual user tap, so it doesn’t feel like the most robust or ideal long-term solution.

I would love to hear some ideas on this! @raineorshine @ethan-james

ethan-james:
> Quick update on this issue: it appears that the toolbar buttons are not being triggered by the current implementation in tap.ts (iOS helper).
> 
> At first, I suspected this might be due to the “Text Color” icon not being visible on mobile, requiring a scroll action before interaction. However, further testing showed that even toolbar icons already visible on screen are not responding to the tap actions. I also explored a few alternative APIs from WebdriverIO, but none seem to work.
> 
> As a workaround, I implemented a helper function that directly interacts with the DOM. It waits for the target element using the appropriate selector and then dispatches mouse events programmatically:
> 
> ```ts
> node.dispatchEvent(new MouseEvent('mousedown', { button: 0 }))
> node.dispatchEvent(new MouseEvent('mouseup', { button: 0 }))
> node.click()
> ```
> 
> This approach resolves the issue functionally. That said, it bypasses the standard user interaction flow by triggering internal DOM events rather than simulating an actual user tap, so it doesn’t feel like the most robust or ideal long-term solution.
> 
> I would love to hear some ideas on this! @raineorshine @ethan-james

Thanks @karunkop! Did you get it to work locally, or in another branch? I tried my hand at it and ended up with

```
2026-05-05T21:27:10.5930819Z unknown error: WebDriverError: An unknown server-side error occurred while processing the command. Original error: el2.dispatchEvent is not a function. (In 'el2.dispatchEvent(new MouseEvent("mousedown",{button:0}))', 'el2.dispatchEvent' is undefined) when running "execute/sync" with method "POST"
```

Maybe I'm not doing it right.

As for the architecture, I agree that it's not ideal to trigger internal DOM events. The only possible cause that comes to mind is that maybe the event handlers in `ToolbarButton.tsx` are too selective for webdriver:

```
      onMouseDown={isTouch ? undefined : tapDown}
      onClick={isTouch ? undefined : tapUp}
      onTouchStart={isTouch ? tapDown : undefined}
      onTouchEnd={isTouch ? tapUp : undefined}
```

raineorshine:
Bypassing webdriver is not a good option unfortunately. Webdriver controls a full browser instance, so we should be able to do pretty much anything the user can do. Something else must be preventing this from working as expected.

ethan-james:
> The only possible cause that comes to mind is that maybe the event handlers in `ToolbarButton.tsx` are too selective for webdriver:
> 
> ```
>       onMouseDown={isTouch ? undefined : tapDown}
>       onClick={isTouch ? undefined : tapUp}
>       onTouchStart={isTouch ? tapDown : undefined}
>       onTouchEnd={isTouch ? tapUp : undefined}
> ```

I tried updating `ToolbarButton` to call `tapUp` `onClick` to see if that would be better for webdriver, but it didn't appear to have any effect.

ethan-james:
@karunkop managed to get the tests passing in a branch, #4249, thanks for that! However, it was accomplished by making [tapUp](https://github.com/ethan-james/em/blob/6db5541a2c24cc94d77956386f432c54b27cd87b/src/components/ToolbarButton.tsx#L112) run `onClick` in addition to `onTouchEnd`, a change that only affects the iOS tests. On an actual device, `onTouchStart` calls `preventDefault` in order to prevent an active editable from blurring, which also cancels `onClick`. It worked in the tests because webdriverio doesn't call touch events at all, so the `preventDefault` never occurred.

I created another test branch, #4250, to continue testing. I added logging statements to `tapDown` and `tapUp`, and verified that they aren't called at all. I switched to a passing test that uses the [tap](https://github.com/ethan-james/em/blob/6db5541a2c24cc94d77956386f432c54b27cd87b/src/e2e/iOS/helpers/tap.ts#L22) helper, and added logging statements `onTouchStart` and `onTouchEnd`. Those also didn't get called, and the `tap` helper only works there because the behavior under test runs on focus or click.

Any chance that we could revisit these tests at a later time? We could be wrong about something here, but it doesn't seem like webdriverio has functional support for touch events. Possibly this is only the case for web apps. I found https://github.com/webdriverio/webdriverio/discussions/7997, which seems to be a related discussion, but not our exact situation.

raineorshine:
Sure. If webdriverio has issues with touchevents, that warrants deeper research outside of this PR to figure out how to get proper mobile coverage. Touch events are critical for a lot of functionality.

raineorshine (approved):

BayuAri:
## Issue I: Needs 3x Undo action to undo a partial Font Color on a note

## Step to Reproduce
1. Create new thought
2. Add a note on the thought
3. Select a word on the note
4. Apply any Font Color
5. Click Undo


## Current behavior
Nothing happens.
Needs 3 Undo action to undo a partial Font Color on a note

https://github.com/user-attachments/assets/aa3efded-631c-44db-9451-3228ad43080d


## Expected behavior
Font Color reverts to previous color, in this case it is note's default grey color


Update:
Logged this as seperate ticket as this is also reproducible on `main` #4273

BayuAri (commented):
**Fixed of issues:**

- [x] [Applying Text Color onto the whole thought text is not responded when a partial of the text already has the same text color #3904](https://github.com/cybersemics/em/issues/3904)
- [x] [Undoing text background highlight makes the font invisible when it has a partial text with color #3905](https://github.com/cybersemics/em/issues/3905)
- [x] [Issue C: Color is reset to default when trying to apply the same Font Color onto text with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227618547)
- [x] [Issue D: Color is reset to default when trying to apply the same Font Color onto Note with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227630914)
- [x] [Issue E: [iOS] Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227704166)
- [x] [Issue F: [iOS] Note Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227748288)
- [x] [Issue G: [Desktop] Undoing deletes the whole thought when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228031188)
- [X] [Issue H: [iOS] Undoing make the text invisible when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228065337)

**Outstanding issue:**

- [ ] [Note's font color becomes invisible after undoing a background highlight selection  #3903](https://github.com/cybersemics/em/issues/3903#issue-4031973385)

ethan-james:
> **Outstanding issue:**
> 
> * [ ]  [Note's font color becomes invisible after undoing a background highlight selection  #3903](https://github.com/cybersemics/em/issues/3903#issue-4031973385)

🤦 I got so caught up with the iOS tests that I forgot to finish #3903. Sorry about that @BayuAri, it should be fixed now.

BayuAri:
@ethan-james 
Outstanding issue is fixed. 
This also fixes #4274 and #4273

List of test cases
**Fixed of issues:**

- [x] [Applying Text Color onto the whole thought text is not responded when a partial of the text already has the same text color #3904](https://github.com/cybersemics/em/issues/3904)
- [x] [Undoing text background highlight makes the font invisible when it has a partial text with color #3905](https://github.com/cybersemics/em/issues/3905)
- [x] [Issue C: Color is reset to default when trying to apply the same Font Color onto text with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227618547)
- [x] [Issue D: Color is reset to default when trying to apply the same Font Color onto Note with Partial Font Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227630914)
- [x] [Issue E: [iOS] Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227704166)
- [x] [Issue F: [iOS] Note Text Color is reset to default when apply any Font Color onto a text with Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4227748288)
- [x] [Issue G: [Desktop] Undoing deletes the whole thought when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228031188)
- [X] [Issue H: [iOS] Undoing make the text invisible when applying the whole Background Color after applying a partial Background Color](https://github.com/cybersemics/em/pull/4034#issuecomment-4228065337)
- [x] [Note's font color becomes invisible after undoing a background highlight selection  #3903](https://github.com/cybersemics/em/issues/3903#issue-4031973385)

raineorshine (approved):
Thank you, nice work

Issue #3886, opened 2026-03-04
[Mobile] Tracing is cut off when user adds tracing after swipe down or up


BayuAri:
## Step to reproduce
1. Pulldown the screen to refresh em app
2. Draw a trace
3. Swipe up or down a bit but do not refresh
4. Draw a trace

## Current behavior
After several attempt, the tracing might cut of and the highlight is also gone

https://github.com/user-attachments/assets/c2707d57-6390-4e89-9a02-54aa2983297f

https://github.com/user-attachments/assets/c182f581-ca50-4984-8d20-1c80b5e10eea

## Expected behavior
Tracing should not experience a cut off

raineorshine:
This can be reproduced consistently if you swipe quickly after an overscroll bounce. 

https://github.com/user-attachments/assets/217a6769-7ae2-4cd3-9022-07ba02e239b6

PR #4038, opened 2026-04-05
Fix Gesture Menu getting stuck open by preventing PanResponder termination during scrolling

### LINKED ISSUE #3887: [Mobile] Transparent layer is shown when user dismisses a stuck Gesture Menu with swipes up
```
- One
- Two
```

## Step to Reproduce A: Pull down the gesture area after pull down on scroll area
1. Put the cursor on any thought
2. Pull down the screen on the scroll area
3. While the bouncing effect is ongoing, do a quick short flick in the opposite way; downward

https://github.com/user-attachments/assets/52d68b52-dbe0-4e7c-8b56-7530dd79d0e8

## Step to Reproduce B: Swipe up after the bouncing animation when Command Center is on screen
1. Create at least 10 thoughts to trigger long scrolling experience
2. Set the cursor on the bottom thought
3. Swipe up to open Command Center
4. Scroll up to the first thought on the scroll area
5. When reaching the top, bouncing animation will be played
6. When the bouncing animation is going up, quickly swipe up on the gesture area

https://github.com/user-attachments/assets/8b6e91db-fa7c-42a7-ad6d-bd7e6528ec4a

## Current behavior
There is a high chance that a transparent layer is activate after the short flick.
The layer will block user to interact with the app, it blocks the screen.  
Workaround: open iOS app switcher and select em app to clear the transparent layer

## Expected behavior
Nothing should happen when user swipe up during the bouncing animation is going up.

fbmcipher:
**Fixes #3887**

## Summary

Fixes the gesture menu getting permanently stuck in a visible state when rapidly swiping in the scroll zone then performing a gesture.

## Problem

When a user scrolls in the scroll zone and then immediately performs a gesture, iOS momentum scrolling continues generating scroll events during the gesture.

Logging shows this causes the `PanResponder` to repeatedly terminate and re-grant on each frame – `onPanResponderTerminate` and `onPanResponderGrant` cycle continuously as long as scroll events are firing.

This happens because React Native Web's responder system (`ResponderSystem.js`) terminates the current responder when a scroll event fires on a parent of the responder node.

Because the PanResponder is in a **terminated** state at `touchend` (once the user's finger leaves the screen after inputting a gesture), `onPanResponderRelease` never fires. This means `handleGestureEnd` never runs and the 400ms `gestureMenuTimeout` that opens the gesture menu is never cleared – leaving showGestureMenu permanently stuck as true.

## Fix

`onPanResponderTerminationRequest` now returns `false` when disableScroll is true (i.e., the user is in the gesture zone), preventing the scroll events from stealing the responder during an active gesture.

## Video

left: this PR, right: `main`

https://github.com/user-attachments/assets/cd4134ab-5270-4055-9f65-4cf201bc11ce


BayuAri (commented):
STR A and B are confirmed fixed, however, the issue is still reproducible with below steps.

##  Transparent layer is visible when user draws a slow and continuous gesture when the app is loading
## Step to Reproduce
1. Delete all of the existing thoughts
2. Kill the app
3. Reopen the app
4. When the app is loading, draw a slow and continuous  gesture until seeing the gesture menu


## Current behavior
Gesture trace will be cut off ##3886
Gesture menu will likely stuck on screen when user has let go the finger off the screen.
The transparent layer is visible.
This is reproducible on iOS and Android Capacitor.

https://github.com/user-attachments/assets/51d77c92-1d29-439c-af00-48b813b962aa


## Expected behavior
The app should be able to handle this use case without causing the transparent layer.

fbmcipher:
Thank you @BayuAri!

I was able to reproduce the issue, but only occasionally. That said, after looking through the failure logs, I think I was able to fix it. Would you mind testing again?

BayuAri:
Transparent layer is no longer reproducible behind the stuck gesture menu.
But the stuck gesture menu is still reproducible easily on iOS and Android Capacitor as well as  on Mobile Safari and Chrome. 
@fbmcipher 
Does this fix #3886 as well?

https://github.com/user-attachments/assets/a2c4b065-2222-417d-8453-43de5fed4db3

Tested on [4430724](https://github.com/cybersemics/em/pull/4038/commits/4430724f2057194adc1f29e12543ecf84461078d)

raineorshine:
@fbmcipher Hi! Do you have an update on this?

fbmcipher:
Yes, sorry! I spent the last few weeks focussing on other issues, but kept this in the back of my head and managed to find a simple fix last Tuesday, which I've just verified and pushed.

@BayuAri, would you mind retesting [with these steps](https://github.com/cybersemics/em/pull/4038#pullrequestreview-4060692021)?

# Explanation

The issue @BayuAri spotted happens when you input a gesture while the app is still loading **and there are no thoughts yet**.

With 0 thoughts, `Content` mounts `EmptyThoughtspace` instead of `LayoutTree`.

Inside `EmptyThoughtspace` there's a three-way swap driven by `offlineStatusStore` and `state.isLoading`:

- `preconnecting` → renders `null`
- `connecting` / `isLoading` → renders `<LoadingEllipsis>`
- otherwise → renders the default "Swipe to add a new thought." hint (or the tutorial message)

The store starts at `preconnecting` and advances through the websocket connection lifecycle and times out (`src/stores/offlineStatusStore.ts`) – **this is still active, even though there's nothing for it to connect to yet.** This might be a separate issue worth looking into.

## The bug

The gesture handler (`MultiGesture` / react-native-web `PanResponder`) wraps `Content` high up in `AppComponent`.

The element the gesture touch event physically lands on, though, lives *inside* `EmptyThoughtspace`, the `LoadingEllipsis` overlay, or the swipe-hint span.

Touch events' `target` gets **pinned** at `touchstart` to whatever element the finger first landed on. Every subsequent `touchmove`/`touchend` for that touch is dispatched to that same element **even after it's been removed from the DOM**. (Notably, this isn't the case for **pointer** events, just the older **touch** events that `PanResponder` uses.)

So when loading resolves mid-gesture, **the render swap takes place and the touch target unmounts**. The remaining `touchmove`/`touchend` still fire, but never reach `PanResponder`. This causes the gesture is lost and the gesture menu freezes on screen, as seen in Bayu's issue.

## The fix

Simple: disable pointer events on the unstable subtree in the `EmptyThoughtspace` root div.

Now, all gesture touch events will pass through `EmptyThoughtspace` and get fired on `#content` instead – a stable ancestor that never gets unloaded.

## Before (`main`)

https://github.com/user-attachments/assets/26b0b50a-7ae5-43d2-ae2a-ca532cc1bed0

## After (this PR)

https://github.com/user-attachments/assets/b2380244-3d5c-4481-b80a-b689c76eece6

raineorshine (approved):
Thanks, code looks good to me. Please resolve the merge conflict at your convenience.

It's not ideal that we have to modify `EmptyThoughtspace` styles in order to get `MultiGesture` to work, but I guess if the deeply nested `target` depends on what is rendered at the time the gesture is initiated, then it is necessary. Hopefully the `pointerEvents: 'none'` fix does not extend to other components.

BayuAri (commented):
Confirm fixed.
Issue is no longer reproducible.

**Test Case**
- [x] [Step to Reproduce A: Pull down the gesture area after pull down on scroll area](https://github.com/cybersemics/em/issues/3887#issue-4022733029)
- [x] [Step to Reproduce B: Swipe up after the bouncing animation when Command Center is on screen](https://github.com/cybersemics/em/issues/3887#issue-4022733029)
- [x] [Step to Reproduce C: Transparent layer is visible when user draws a slow and continuous gesture when the app is loading](https://github.com/cybersemics/em/pull/4038#pullrequestreview-4060692021)

PR #4060, opened 2026-04-09
Merge setDescendant actions in undoRedoEnhancer

### LINKED ISSUE #3903: Note's font color becomes invisible after undoing a background highlight selection

ethan-james:
Fixes #3903

As noted, this is the same as #3551 but for notes. We can track `setDescendant` actions instead of `editThought` actions in `undoRedoEnhancer` and merge actions that have the same [EditThoughtDirection](https://github.com/ethan-james/em/blob/9e75f26b413cafc8162b76727420929b4d4fe891/src/redux-enhancers/undoRedoEnhancer.ts#L25). I had Copilot write the solution so I'm going to take some time tomorrow to clean it up and try to combine it into the existing `lastEditThoughtDirection` logic before I commit it.

I noticed a couple of things not directly related to this issue while I was figuring this out:

1. `setDescendant` actions seem to fire twice for each edit, so I needed to modify the logic to accommodate the duplicates:

```
        return newValue.length === oldValue.length
          ? lastSetDescendantNoteDirection
          : newValue.length > oldValue.length
            ? EditThoughtDirection.Longer
            : EditThoughtDirection.Shorter
```

Ideally I would be able to fix that at the source, as part of this issue or a new issue, so that it's not dispatching duplicate actions.

2. When editing a note in desktop Chrome, trailing(?) whitespace gets converted to an HTML entity, so the progressive values in `setDescendant` actions look like:

```
  - note
  - note&nbsp;
  - note t
```

which messes up the direction sense and breaks merging. That doesn't impact formatting tags, but ideally I can hunt that down and normalize those values as well.

### DIFF
--- a/src/commands/__tests__/undo-redo.ts
+++ b/src/commands/__tests__/undo-redo.ts
@@ -6,12 +6,14 @@ import { importTextActionCreator as importText } from '../../actions/importText'
 import { indentActionCreator as indent } from '../../actions/indent'
 import { moveThoughtDownActionCreator as moveThoughtDown } from '../../actions/moveThoughtDown'
 import { newThoughtActionCreator as newThought } from '../../actions/newThought'
+import { setDescendantActionCreator as setDescendant } from '../../actions/setDescendant'
 import { undoActionCreator as undo } from '../../actions/undo'
 import { executeCommandWithMulticursor } from '../../commands'
 import moveThoughtDownCommand from '../../commands/moveThoughtDown'
 import { HOME_TOKEN } from '../../constants'
 import { initialize } from '../../initialize'
 import childIdsToThoughts from '../../selectors/childIdsToThoughts'
+import contextToPath from '../../selectors/contextToPath'
 import exportContext from '../../selectors/exportContext'
 import { getLexeme } from '../../selectors/getLexeme'
 import store from '../../stores/app'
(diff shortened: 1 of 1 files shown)


raineorshine:
> As noted, this is the same as #3551 but for notes. We can track `setDescendant` actions instead of `editThought` actions in `undoRedoEnhancer` and merge actions that have the same [EditThoughtDirection](https://github.com/ethan-james/em/blob/9e75f26b413cafc8162b76727420929b4d4fe891/src/redux-enhancers/undoRedoEnhancer.ts#L25).

Sounds good.

> Ideally I would be able to fix that at the source, as part of this issue or a new issue, so that it's not dispatching duplicate actions.

Yes, that's worth looking into. I don't know offhand why `setDescendant` fires twice with the same value.



> 3\. When editing a note in desktop Chrome, trailing(?) whitespace gets converted to an HTML entity, so the progressive values in `setDescendant` actions look like:
> 
> ```
>   - note
>   - note&nbsp;
>   - note t
> ```
> 
> which messes up the direction sense and breaks merging. That doesn't impact formatting tags, but ideally I can hunt that down and normalize those values as well.

Good catch. There is some logic in https://github.com/cybersemics/em/blob/8f597f81de6611dd90ec57e49c842366ae52ae1a/src/components/Note.tsx#L117-L119 that handles Mobile Safari-specific `<br>` insertions. That seems like a similar mitigation, so maybe we can strip the trailing whitespace there.

Maybe we should factor out the `newValue` transformation that handles trailing spaces from the Editable and use it for both the Editable and Note?

https://github.com/cybersemics/em/blob/8f597f81de6611dd90ec57e49c842366ae52ae1a/src/components/Editable.tsx#L378

ethan-james:
> Yes, that's worth looking into. I don't know offhand why `setDescendant` fires twice with the same value.

It looks like the first commit where `setDescendant` fires twice was 00edbf1f4e7ab89f733b5420b158ae7e7cff8985, which unfortunately is not the easy answer I was hoping to find. Hopefully I can get to the bottom of it pretty quickly next week.

raineorshine:
Interesting. `debuggingEnhancer` (now `validateStateEnhancer`) is called twice in the reducer chain, but it just passes through the next state, so I don't see how `setDescendant` would be called twice. I would think `setDescendant` would be called once via the `appReducer`.

PR #4066, opened 2026-04-09
TDD Fail

raineorshine:
A dummy PR to test the tdd workflow.

Adds a test that succeeds on both `main` and this PR, thus failing the TDD check.


raineorshine:
I don't know why TDD thinks this has the `skip-tdd` label.

Issue #3511, opened 2025-12-26
Upgrade react-signature-pad-wrapper without breaking drag text selection


raineorshine:
The [react-signature-pad-wrapper](https://github.com/michaeldzjap/react-signature-pad-wrapper) dependency was originally upgraded from 4.2.0 to 4.3.2 in #3430 which unknowingly broke the native drag text selection behavior (#3483).

Goal: Upgrade `react-signature-pad-wrapper` without reintroducing #3483.

There is only `4.3.0` and `4.3.1` intervening versions. It may be worth identifying exactly which version upgrade broke and investigating why.

We definitely monkey with `SignaturePad` internals in https://github.com/cybersemics/em/blob/761bf8fc7de9aa479aab5ed327bae3a3208e71cb/src/components/TraceGesture.tsx, including overriding `touchstart` and `touchmove`, so it's especially fragile to upgrades.

Issue #3483, opened 2025-12-22
[iOS] Text Selector feature is not working properly on Swipe Zone


BayuAri:
## Step to reproduce
1. Open em on Safari mobile or iOS app
2. Create a thought with some words
3. Double tap on a word that lies on Swipe Zone on the above thought
4. Verify the word is wrapped in the selector tool
5. Tap on the text selector and drag

## Current behavior
Most of the time, nothing happens.
If user tries to tap on the selector too long, drag and drop feature will be active.
Note: This does not happen when the selected word is located on Scroll Zone

**Video:** 

https://github.com/user-attachments/assets/d014b29c-ec20-457d-898c-0e9a05b26e0a

**Screenshot:**

<img width="800" height="1029" alt="Image" src="https://github.com/user-attachments/assets/0d80bff7-cadc-47e5-a144-79c769f577e0" />

## Expected behavior
Text selector should be draggable, refer to user's drag direction on Swipe Zones, refer to the behavior of the Scroll Zone

BayuAri:
@BayuAri 
Ooops 
My bad, I have move this to the correct ticket and trying to work on your answer
Thanks

PR #4141, opened 2026-04-19
Puppeteer Test Warnings

### LINKED ISSUE #3406: Unable to preventDefault inside passive event listener due to target being treated as passive
Puppeteer test warning

Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/feature/5093566007214080

karunkop:
Close #3406 

### Cause:

Gesture drawing prevents scrolling using `preventDefault()` on touch events, but since signature-pad attaches handlers to window where touch handlers are passive by default in Chrome, calling `preventDefault()` is not allowed and triggers a warning.

### Solution

Newer version of this library solved the issue and no puppeteer warnings were shown in the console.


raineorshine (changes requested):
Thanks!

Upgrading signature causes an unusual regression documented in #3483. A solution to that issue will need to be found before this can be merged.

karunkop:
> Thanks!
> 
> Upgrading signature causes an unusual regression documented in #3483. A solution to that issue will need to be found before this can be merged.

Is there an issue opened for that? Or should I open a new one and start working on that?

raineorshine:
There's no open issue because it's not an issue in `main`. It's a regression in this PR. The task is to fix #3406 without causing a regression, and hours should be logged against #3406.

karunkop:
> There's no open issue because it's not an issue in `main`. It's a regression in this PR. The task is to fix #3406 without causing a regression, and hours should be logged against #3406.

I found this https://github.com/cybersemics/em/issues/3511, so we can close off this once the review is done!

raineorshine (approved):
@BayuAri Can you test that this PR does not bring back #3483? Thanks!

BayuAri (commented):
Text Selector is working as expected and #3483 is not reproducible on this branch.

PR #4143, opened 2026-04-19
[fixed] Use `https` dev server instead of `http` to enable access to new web APIs in development environment

fbmcipher:
**This is a resubmission of PR #4039. That PR showed green/passing CI tests, but failed when merged.**

## Why was CI green?

- `pull_request_target` runs the workflow file from the **base** branch, not the PR branch.
- At the time, main's `ios.yml` didn't set `TUNNEL_TOKEN`, so the gate plugin saw no token, returned `undefined`, and never ran.
- After merge, the workflow file *did* set the token, the gate activated, and a bug in the PR was exposed.

## Why did it break?

- We need to use a Cloudflare tunnel instead of the default BrowserStack tunnel, as that's the most straightforward path to get full HTTPS coverage in the BrowserStack iOS e2e tests. Other open-source projects have struggled with this issue.
- To protect what would otherwise be an _unprotected Cloudflare tunnel_ (which – though isn't the biggest security issue – is a little unhygenic), we protect the tunnel with a token that only the CI knows.
- This token checking mechanism only checked for the token in the URL, not in follow-up requests - causing HTTP 403 errors when the browser tried to load JS/CSS/font subresources.
- Fixed in commit 3ac35a9f

## Manual testing

Since we can't rely on the PR's green tick, I manually scheduled a run on the broken commit that was merged `8277b24a`, and the new fixed commit `3ac35a9f`:

- Broken commit `8277b24a` → ❌ [run 24617280414](https://github.com/cybersemics/em/actions/runs/24617280414) — same `NoSuchElementError: #skip-tutorial` as your failed run
- Fix commit `3ac35a9f` → ✅ [run 24617280774](https://github.com/cybersemics/em/actions/runs/24617280774) — all 3 specs pass in 4m32s

## One more test, or merge now?

To be 100% sure, I wanted to merge into my fork `fbmcipher/em`, but I can't run this workflow from there as it requires `BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY` secrets. GitHub secrets are encrypted, so I couldn't just copy them from `cybersemics/em`.

So we have two options:

1. Share the `BROWSERSTACK_USERNAME` / `BROWSERSTACK_ACCESS_KEY` with my fork `fbmcipher/em` and I can run the workflow on a test-merge branch myself – gives full guarantee before merging.
2. Or just merge the fix branch as-is; if it breaks again, revert and let me know.

Apologies again – without the secrets I can't fully simulate a merge to main on my own fork, which is what made this slip through the first time.


raineorshine:
[passed on main](https://github.com/cybersemics/em/actions/runs/24681610655/job/72180058957) :)

raineorshine:
Sadly, the BrowserStack workflow is failing in new PRs now. Sorry this one is proving difficult! I'll give you the secrets so you can troubleshoot directly.

image

https://github.com/cybersemics/em/actions/runs/24686106621/job/72198565341?pr=4125

Reverted in 7909e9a3990

raineorshine:
@fbmcipher Hi! Do you have an update on the https feature?

It looks like the storage methods of the new sync engine require https, so this is more critical now.

PR #4180, opened 2026-04-25
Express Worker Bee as skills and custom instructions

fbmcipher:
CleanShot 2026-04-25 at 02 02 32@2x

## Overview

In some contexts, it's not possible to assign custom agents on GitHub Copilot – for example, when mentioning Copilot from a PR comment. This limits the effectiveness of the custom agent approach.

Fortunately, we can get around this with a few changes:

- `copilot-instructions.md` now contains the high-level agent "temperament" (the confident, methodical engineer persona) and the general methodology from Worker Bee (create branch, open PR, wait for CI, iterate until passing, communication style)
- **Skills** handle CI monitoring and Puppeteer test diagnosis – lazily loaded only when needed.
- Separate `.instructions.md` files break out what was previously crammed into one file (code standards, testing guidelines) into focused, structured documents.

This is actually better than the custom agent approach for a few reasons. It works everywhere Copilot runs, but it also gives us a stable foundation to add many more custom instructions in a structured way.

## Skills

Skills are lazily loaded. Each `SKILL.md` file includes a short description that acts as a hint to the agent. If the agent determines that a skill is relevant, it loads the full content into its context window. If not, it's never loaded.

This is a recent innovation. In a talk I watched from Anthropic last week, they actually recommend skills over custom agents now – primarily because of their lazy-loading ability. 

Rather than dumping the entire instruction set into context upfront (which risks context overload causing the LLM to skip or miss important instructions), skills are loaded only when the agent deems them to be relevant.

This gives us the room to expand to potentially hundreds of skills without degrading agent performance.

## Multiple custom instruction files

The existing `copilot-instructions.md` had a lot going on – code standards, testing guidelines, methodology, references to documentation – all in one file. The agent wasn't following these consistently, and I think part of the issue was the LLM getting overwhelmed by unstructured information.

Splitting rules into their own focused files seems to have improved consistency. The agent now follows the new Worker Bee methodology in `copilot-instructions.md` reliably. Splitting custom instructions into multiple files also gives us a more stable architecture to build on.

The root `copilot-instructions.md` is now solely responsible for high-level agent temperament and methodology. We should write separate `instructions.md` file for anything else.

Interestingly, these custom instruction files can be loaded based on a glob path. So you could conditionally load a special custom instruction file for any `*e2e*` or `*test*`.

## Documentation / single source of truth

Early on in our AI exploration together, I noticed that GitHub Copilot doesn't follow direct links in custom instructions – so the URLs to **em**'s wiki that were in the old instructions file weren't being read. You have mentioned wanting a single source of truth for documentation that serves both humans and agents.

I think the skill system makes this possible. Looking ahead, I can imagine a lazy-loading documentation architecture. After some research today, I've seen teams doing this with vector databases and MCP servers that support semantic search. The agent would dynamically fetch documentation and add it to its context window only when relevant, rather than having everything loaded upfront.

This could address some of the issues we've been seeing – the agent would consult documentation before executing, which could improve its architectural decisions. Not implemented yet, but the skill system gives us the foundation.

## A/B testing

- **Before**: #4113: Even after being told explicitly to use the Worker Bee agent, Copilot couldn't observe CI results, and didn't fix Puppeteer tests after a couple of tries.
- **After**: https://github.com/fbmcipher/em/pull/16: After a single prompt, the agent is able to fix 12 of the 13 failing Puppeeter tests. It worked for 52 minutes unattended.

The 13th Puppeteer test failed due to a marginal snapshot difference (~14px):

command-center-1-diff

It reasoned that it needed to update the snapshot, but couldn't set up the Docker SSL proxy required for Puppeteer snapshot updates. Since we're about to merge a fix that moves the dev server over to HTTPS anyway, I'll hold off on a fix for this until we know more about what's going on.

But this is still a drastic improvement!

### DIFF
--- a/.github/agents/worker-bee.agent.md
+++ b/.github/agents/worker-bee.agent.md
@@ -8,7 +8,6 @@ You are a confident, reliable, diligent engineer who proactively manages code qu
 You will autonomously manage lifecycle of code changes: create a new branch, commit and push changes, open a draft pull request, and ensure all CI checks pass before considering the work complete. Please see the methodology listed below. NEVER skip a step.
 
 ## Methodology
-
 - Always begin by creating a new branch for the work. If a previous agent working on the same task already created a branch and a PR, use that one.
 - When opening a PR, include the bare issue number at the top of the description (e.g. "#1234").
 - Make all commits in this branch. Push after each meaningful change. Never commit directly to main or protected branches.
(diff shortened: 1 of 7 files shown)


raineorshine (commented):
Thanks for the thorough analysis and testing! Great work.

My only question is, are all of the `instructions.md` files loaded into context automatically? I understand that skills are lazy loaded, but I'm not as clear about how instruction files are loaded. I only see documentation for [path-specific instructions](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions#creating-path-specific-custom-instructions) that use `applyTo` to target certain paths and file types.

fbmcipher:
Great question!

[I asked a Copilot agent on my fork to summarize all custom instructions](https://github.com/fbmcipher/em/tasks/8342ff1b-1d4a-4906-8ba6-c77db65e7755?author=fbmcipher) to find out.

The agent correctly summarized all the custom instructions, including the `instructions.md` files, without using a `read` tool to cat their contents into its context window:

CleanShot 2026-04-27 at 18 41 02@2x

So I can say with confidence that all `instructions.md` files are loaded into context automatically when `applyTo` is not provided.

The docs mention an explicit `**` glob can be passed to `applyTo`, so I'd guess that's the implicit default when it's omitted.

raineorshine:
Okay, interesting. Strange that it's not documented officially!

PR #4181, opened 2026-04-25
Flaky test: Sidebar - Node is either not clickable or not an Element

### LINKED ISSUE #4148: Flaky test: Sidebar - Node is either not clickable or not an Element
```
 ❯ CdpElementHandle.clickablePoint node_modules/puppeteer-core/lib/esm/puppeteer/api/ElementHandle.js:692:23
 ❯ CdpElementHandle.<anonymous> node_modules/puppeteer-core/lib/esm/puppeteer/api/ElementHandle.js:118:20
 ❯ CdpElementHandle.click node_modules/puppeteer-core/lib/esm/puppeteer/api/ElementHandle.js:722:30
 ❯ CdpElementHandle.<anonymous> node_modules/puppeteer-core/lib/esm/puppeteer/api/ElementHandle.js:129:24
 ❯ CdpFrame.click node_modules/puppeteer-core/lib/esm/puppeteer/api/Frame.js:749:17
 ❯ src/e2e/puppeteer/__tests__/sidebar.ts:47:5


Error: Node is either not clickable or not an Element
```

https://github.com/cybersemics/em/actions/runs/24735442185/attempts/1?pr=4140

karunkop:
Close #4148 

### Cause

The test was waiting for the outer sidebar wrapper, which is always mounted and visible, instead of the actual sliding drawer inside it. So the test tried to click a link before the drawer had fully appeared on screen. Because the element was technically “visible” but still off-screen due to animation, puppeteer couldn’t find a clickable area and threw an error.

### Fix

Updated `openSidebar()` to wait until the sidebar is actually visible and on-screen by checking aria-hidden="false" and ensuring a sidebar link has a valid position (rect.left >= 0 and width > 0), so clicks only happen after the drawer finishes sliding in.

### DIFF
--- a/src/e2e/puppeteer/__tests__/sidebar.ts
+++ b/src/e2e/puppeteer/__tests__/sidebar.ts
@@ -17,7 +17,15 @@ vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
 /** Open sidebar and wait for it to slide all the way open. */
 const openSidebar = async () => {
   await click('[aria-label=menu]')
-  await page.locator('[data-testid="sidebar"]').wait()
+  // Wait for aria-hidden="false" and the first link to be on-screen (rect.left >= 0), since the outer sidebar is always mounted and doesn’t reflect the drawer’s slide-in animation.
+  await page.waitForFunction(() => {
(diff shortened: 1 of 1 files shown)


raineorshine (approved):
Looks good, thanks!

PR #4182, opened 2026-04-25
Layout shift tests

### LINKED ISSUE #3672: Test that thoughts do not shift y position after first render
There are a few bugs that involve a thought's initial y position changing and causing it to drift slightly up or down as it fades in. The test cases vary depending on whether the thought is the last subthought or not, triggering the additional cliff margin, and the timing of measuring the DOM. Given the sensitivity of this logic, it would be great to get test coverage of this behavior to help avoid future regressions.

Test cases: 

- #2783
- #3097
- #3310
- #3647
- #3671

What would be a good way to test this? I was thinking the y position of each thought could be read after its initial render and at the end of its animation to make sure it is unchanged. This might be complicated by the fact that animations are disabled in Puppeteer. Let me know what you think would work. 

Ideally we find a way to test this without re-enabling the full animations and thus slowing down the tests.

karunkop:
Close #3672 

As discussed in the issue, here is the implementation for the test coverage using mutation observer.

Since two of the tests cover the existing bug, the test fails like this:

https://github.com/cybersemics/em/actions/runs/24926853611/job/72998137615?pr=4182


src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,186 @@
...
+        let observer: MutationObserver | null = null
+
+        /** Parse the `top` CSS property from a style string. */
+        const parseTop = (styleStr: string | null): number | null => {
+          const m = /top:\s*([-\d.]+)px/.exec(styleStr ?? '')
+          return m ? parseFloat(m[1]) : null
+        }
+
+        /** Schedule to resolve and disconnect the observer after `settleMs` milliseconds. */
+        const scheduleSettle = () => {
```
raineorshine: Let's call this `resolveDebounced`. Debouncing is the more common terminology for this behavior.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,186 @@
...
+const measureYShift = (thoughtText: string, { settleMs = 500 }: { settleMs?: number } = {}): Promise<YShiftResult> =>
+  page.evaluate(
+    (thoughtText: string, settleMs: number, overallTimeoutMs: number) =>
+      new Promise<YShiftResult>((resolve, reject) => {
+        let target: HTMLElement | null = null
+        let before: number | null = null
+        let after: number | null = null
+        let sawStyleMutation = false
+        let settleTimer: ReturnType<typeof setTimeout> | null = null
+        let observer: MutationObserver | null = null
```
raineorshine, line 40: Add JSDOCs at variable declarations so that developers don't need to infer their use from the code.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,186 @@
...
+          const m = /top:\s*([-\d.]+)px/.exec(styleStr ?? '')
+          return m ? parseFloat(m[1]) : null
+        }
+
+        /** Schedule to resolve and disconnect the observer after `settleMs` milliseconds. */
+        const scheduleSettle = () => {
+          if (settleTimer) clearTimeout(settleTimer)
+          settleTimer = setTimeout(() => {
+            observer?.disconnect()
+            resolve({ thoughtText, before, after, delta: (after ?? 0) - (before ?? 0) })
```
raineorshine: Just return `before` and `after` and let the caller calculate the delta. 

Composite outputs of a function should always be orthogonal.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,186 @@
...
+import press from '../helpers/press'
+import waitForEditable from '../helpers/waitForEditable'
+import { page } from '../setup'
+
+vi.setConfig({ testTimeout: 20000, hookTimeout: 20000 })
+
+const Y_TOLERANCE = 0.5
+
+type YShiftResult = {
+  thoughtText: string
```
raineorshine: Let's call this `value` to follow existing conventions. While it is very general, `value` is used ubiquitously throughout the codebase to refer to thought values. It's better to stay consistent than make up new terminology. You can use JSDOCs to clarify that it refers to the thought value.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,186 @@
...
+
+        observer = new MutationObserver(mutations => {
+          for (const m of mutations) {
+            // Determine the target from the first insertion containing the thought, using its text node and parent positioner; the initial position is read from the DOM but may be corrected using the first mutation’s oldValue.
+            if (!target && m.type === 'childList') {
+              for (const n of Array.from(m.addedNodes)) {
+                if (!(n instanceof Element)) continue
+                const host = Array.from(n.querySelectorAll('[data-editable]'))
+                  .find(el => (el.textContent ?? '').trim() === thoughtText)
+                  ?.closest('[aria-label="tree-node"]') as HTMLElement | null
```
raineorshine: You can skip the `[data-editable]` and just query `[aria-label="tree-node"]`:

```typescript
const host = Array.from(n.querySelectorAll('[aria-label="tree-node"]'))
  .find(el => (el.textContent ?? '').trim() === thoughtText) as HTMLElement | null
```

Then the code depends only on tree-node and not on the DOM relationship between tree-node and data-editable.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,186 @@
...
+          }, settleMs)
+        }
+
+        observer = new MutationObserver(mutations => {
+          for (const m of mutations) {
+            // Determine the target from the first insertion containing the thought, using its text node and parent positioner; the initial position is read from the DOM but may be corrected using the first mutation’s oldValue.
+            if (!target && m.type === 'childList') {
+              for (const n of Array.from(m.addedNodes)) {
+                if (!(n instanceof Element)) continue
+                const host = Array.from(n.querySelectorAll('[data-editable]'))
```
raineorshine: Let's call this `treeNode`.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,186 @@
...
+                  before = after = parseTop(target.getAttribute('style'))
+                  scheduleSettle()
+                  break
+                }
+              }
+            } else if (target && m.type === 'attributes' && m.target === target) {
+              const oldTop = parseTop(m.oldValue)
+              const newTop = parseTop(target.getAttribute('style'))
+              if (oldTop === newTop) continue
+              if (!sawStyleMutation) {
```
raineorshine: The purpose of `sawStyleMutation` is a bit unclear to me. Why do we need to overwrite the initial `before` value?

If `before !== null` or `after === before` could possibly cover this case, that would be a bit simpler.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,186 @@
...
+        }, overallTimeoutMs)
+      }),
+    thoughtText,
+    settleMs,
+    8000,
+  )
+
+/** Expect the y position of the thought to be stable within the tolerance. */
+const expectStableY = ({ thoughtText, before, after, delta }: YShiftResult) => {
+  if (Math.abs(delta) > Y_TOLERANCE) {
```
raineorshine: Use the appropriate `expect` function rather than throwing:

```ts
expect(Math.abs(delta)).toBeGreaterThan(Y_TOLERANCE)
```

With a custom error message:

```ts
expect(Math.abs(delta), `Thought "${thoughtText}" y shifted ${Math.abs(delta).toFixed(2)}px ${delta < 0 ? 'up' : 'down'} (before=${before}, after=${after}).`).toBeLessThanOrEqual(Y_TOLERANCE)
```

raineorshine (changes requested):
Thanks. The tests look very clean and readable.

raineorshine:
FYI I fixed the TDD workflow so you should be able to add `.skip` to the intentionally breaking tests. Just rebase on `main`.

The Puppeteer test will PASS in this PR and `.skip` will be stripped so they FAIL in `main`, allowing the TDD workflow to pass.

src/e2e/puppeteer/__tests__/layout-shift.ts
karunkop: > You can skip the `[data-editable]` and just query `[aria-label="tree-node"]`:
> 
> ```ts
> const host = Array.from(n.querySelectorAll('[aria-label="tree-node"]'))
>   .find(el => (el.textContent ?? '').trim() === thoughtText) as HTMLElement | null
> ```
> 
> Then the code depends only on tree-node and not on the DOM relationship between tree-node and data-editable.

Target tree node was not found when I just did querySelectorAll for `[aria-label="tree-node"]`.

raineorshine: Please don't shoot this back to me so quickly. If my exact code snippet didn't work, try a variation that does. You need to to deeply understand what the code is doing and why. If my suggestion is faulty, explain to me why.

karunkop: `querySelectorAll` searches only descendants, never the element itself. When a single thought mounts, the `MutationObserver` delivers n as the `[aria-label="tree-node"]` element directly, so `n.querySelectorAll('[aria-label="tree-node"]')` returns an empty NodeList, find never executes, the observer never resolves, and the test times out.

In contrast, `[data-editable]` is always a descendant of the tree-node, so `querySelectorAll` finds it reliably. `.closest('[aria-label="tree-node"]')` then walks upward and includes the element itself, correctly returning n when n is already the tree-node.

raineorshine: Thanks for the explanation.

If `n` is the tree-node itself, why not just check it directly? Why the traversal?

karunkop: Are you referring to Array.from(n.querySelectorAll(...)) traversal?

I think my earlier explanation caused some confusion. n is not guaranteed to be the tree-node itself. It comes from m.addedNodes, which can contain any DOM node that gets inserted, often a wrapper element (likely a span from FadeTransition/CSSTransition) rather than the actual thought node.
Because of this, directly checking n is not reliable. Instead, we first find the matching [data-editable] element inside the inserted subtree, then use .closest('[aria-label="tree-node"]') to reliably resolve the correct tree node, regardless of any nested wrapper structure.

raineorshine: Okay, thanks. Now I fully understand the need for the traversal. The subtree mutation observer can return wrapped nodes or ancestors that contain the tree-node. 

I wonder if observing the tree-node wrapper div directly rather than all of document.body. However, at this point I don't see a strong advantage to that given that this is test code and it is covering what we need.

raineorshine (approved):

raineorshine:
Unfortunately this caused the Puppeteer tests to fail in `main`, so I had to revert.

PR #4183, opened 2026-04-26
Glassy modal for mobile Command Universe

### LINKED ISSUE #3943: Liminal glass overhaul for `Dialog` component
> **Implement a new liminal glass treatment for the `Dialog` component.**

Image

---

## Component-by-component breakdown
This design uses many subcomponents that work together to create the end result – all listed below, grouped by their responsibilities.

All needed assets are available in [this zip archive](https://github.com/user-attachments/files/26004534/assets.zip).

Image

#### Backdrop/overlay

1. **Background blur**
	- A completely transparent, full screen layer with `backdrop-filter(24px)` – acts as part of the modal's backdrop/overlay. No progressive blur needed.

2. **Background glow**
	- Uses `dialog-background-glow.avif` asset.
	- A simple, subtle glow image that acts as a part of the modal's backdrop/overlay. Adds subtle color.
	- Displayed at 30% opacity.
	- Position it at `top: 0: left: 0`, and make it a full-screen element.
	- Use only the bottom `50%` of the image, set the background-position-x to center, and set the background-size to **twice the width of the screen** and **twice the width of the height.**

#### Glass sheet

3. **Glass sheet fill**
	- A semi-translucent panel with rounded corners (`32px`) designed to resemble a sheet of glass. 
	- Composed of a **simple radial gradient fill**: `background: radial-gradient(107.42% 123.59% at 50% 100%, rgba(191, 143, 254, 0.14) 0%, rgba(75, 138, 149, 0.01) 100%);

4. **Glass sheet stroke**
	- A separate layer which adds a border to the sheet of glass.
	- Uses a **linear gradient** to approximate the effect of light refracting and falling off through the border of the glass.
	- *CSS does not directly support using a linear gradient as a border background*, so some kind of workaround will be required (`inset`, pseudo-element, masked `div`, etc.)
	- *This layer should be masked using the luminance mask.*

#### Highlights

5. **Highlight (unclipped)**
	  - Uses `dialog-highlight.avif` asset.
	  - This image simulates an additional highlight at the top of the modal's glass.
	  - Position this image at the top of the modal, to create the impression of light hitting and reacting with the top of the glass.
	  - Its width should match that of the modal.
	  - Set its opacity to 10%.

6. **Highlight (clipped)**
	- Identical in placement, size and image to the above layer, but clipped to the modal's boundaries. This "doubling-up" creates the impression of some light remaining trapped in the modal's glass.
	- *Use `overflow: none` or some other technique to ensure this image doesn't spill out of the modal.*

7. **Rainbow (unclipped)**
	- Uses `dialog-highlight-rainbow.avif` asset.
	- This image simulates rainbow refractions through mist and glass.
	- Use only the bottom `30%` of the image, and position it at `20%` of the modal glass's `y` position. Width should be `200vw`.
	- Use the `plus-lighter` blend mode with 10% opacity.

8. **Rainbow (clipped)**
	- Identical in placement, size and image to the above layer, but clipped to the modal's boundaries. This "doubling-up" creates the impression of some light remaining trapped in the modal's glass.
	- *Use `overflow: none` or some other technique to ensure this image doesn't spill out of the modal.*
	- Use the `plus-lighter` blend mode with 10% opacity.

--- 

## Mask
In this design, a **luminance mask** is used to control the direction and amplitude at light across a field.

Use this mask on the specified elements to create the impression of light falling off over 

distance:

```css
background: radial-gradient(144.9% 90.61% at 46.96% 13.01%, #FFF 0%, rgba(255, 255, 255, 0.00) 87.02%);
```

---

## Notes

- **In this issue, we only need to implement for phones in portrait and landscape.**
	- The only dialog at present is the **Gesture Cheatsheet**, which is currently only supported on phones.
	- In a future PR, the **Gesture Cheatsheet** will be expanded to support desktops as well.
	- At that point, we'll make incremental improvements to the code developed here to support larger screen sizes.

- **Note that the content of the Gesture Cheatsheet itself, including the dialog's header and buttons, have not changed at all.**
	- These will be changed in follow-up PRs.

fbmcipher:
Closes #3943

CleanShot 2026-04-26 at 02 13 29@2x

---

- Implements the liminal UI redesign for the modal — adds all layers from #3943 (blurred background, rainbow refractions, and glass sheet effect)
- Responsive across mobile sizes, portrait/landscape, iPhone + Android
- Styles consolidated into a single `dialogRecipe` slot recipe. This allows coupled values to live next to each other in the recipe


raineorshine (approved):
Looks good! Code is approved. I did a brief test on desktop and will let @BayuAri do more thorough mobile testing.

@BayuAri This is only a visual change, so just look out for visual glitches or responsive design issues.

raineorshine:
I think we lost the bottom scroll gradient.

Tested in Chrome Device Emulator.

## Current

f00044e69ef

image

## Expected

`main`

image

BayuAri:
@raineorshine @fbmcipher 

Question please
Search field, Sort and X button have black background on the current built but it is different with the [mock up](https://github.com/cybersemics/em/issues/3943#issue-4078357186), is this acceptable?

I would prefer to match with mock up to match the blur background. 


image

Command Universe fits all screens small, mid and large (iOS) without any cut off as it is on `main`

No other issue from my end.  

Tested on 
- iPhone 13 Mini iOS 18.7.6
- iPhone 16e iOS 18.7.8
- iPhone 14 PM iOS 26.2.1
- Samsung A56 OS 16

fbmcipher:
Thank you for the speedy review @BayuAri!

The buttons and search bar will be redesigned in the next issue #4185, so I made the decision to leave them until that task.

Do you think that's acceptable @BayuAri @raineorshine?

@raineorshine: I'll push a commit tonight to reinstate the missing scroll gradient.

fbmcipher:
The code was still there, but for whatever reason the gradient didn't show up. Replaced the `div` gradient approach with a mask, which has the same effect:

CleanShot 2026-04-28 at 23 19 51@2x

Also @raineorshine: is it expected behaviour for unavailable commands to show up dimmed? In this screenshot, the titles for each command are dimmed because they're not available in the current context (`isExecutable` is `false`).

raineorshine:
> The code was still there, but for whatever reason the gradient didn't show up. Replaced the `div` gradient approach with a mask, which has the same effect:

It's still not working for me on Mobile Safari. It cuts off sharply. Are you seeing something different? Tested 73cd6903c1a.

> Also @raineorshine: is it expected behaviour for unavailable commands to show up dimmed? In this screenshot, the titles for each command are dimmed because they're not available in the current context (`isExecutable` is `false`).

The commands should be dimmed on desktop only. I'll create an issue for that!

fbmcipher:
CleanShot 2026-05-03 at 03 27 14@2x

Fixed now @raineorshine – it was working in Safari desktop, but it wasn't showing on Mobile Safari due to a `dvh`/`vh` viewport sizing disparity. Fixed fc1ecae (left)

I also took the opportunity to add a corresponding top scroll mask in a277e78 (right), which uses the new [animation-timeline](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/animation-timeline) API to hide/show the mask as the user scrolls with zero JS.

raineorshine:
I can see a bit of a mask on the bottom, but it's still cutting off too sharply. Below I have positioned the bottom edge so that it cuts through Jump Back/Jump Forward. Instead of fading all the way to black like in `main`, it is only partially masked before being cut off by the bottom edge of the modal.

< `main` | a277e78b455 >

<p style="float: left;">
image
image
</p>

Aside: I think we will want to lighten the bottom edge of the modal, as it is disappearing into the background a little too much.

fbmcipher:
Sorry for the back-and-forth! Could you eyeball the image on the far right (which represents latest commit 4c68180) and let me know if that's what you're looking for?

In that image, I've slightly brightened the dialog, and made sure the mask properly fades to zero like in `main`.

CleanShot 2026-05-22 at 00 20 36@2x

raineorshine:
That looks good. Now we just need to hide the mask when we are at the bottom of the document. Since it's a scroll hint, it should only be shown there is more content to scroll to.

fbmcipher:
@raineorshine Done 52131a7. Sorry for missing that before. It uses the same `animation-timeline`-based mechanism as the top mask, so no JS needed.

raineorshine (approved):
Perfect, thanks

PR #4237, opened 2026-05-07
Reduce exact rank assertions in ordering tests

### LINKED ISSUE #4236: Reduce exact rank assertions in ordering tests ahead of TreeCRDT integration
Soon [TreeCRDT](https://github.com/cybersemics/treecrdt) will be the backend for the thought tree! This will lead to some changes in `em`.

One important difference is that TreeCRDT represents sibling order differently from the current numeric `rank` field in `em`.

`rank` is still the ordering mechanism today, and this issue is not about changing that. The goal is just to make the tests less coupled to exact rank numbers when the behavior being tested is really sibling order.

Some ordering tests currently assert values like `rank === 0.5` or `rank === 2`. In many higher-level tests, the important behavior is simply that thoughts appear in the expected order after actions like move, move up/down, indent/outdent, sort, split, or import.

## Scope

- Update higher-level ordering tests to assert sibling order by id or value where possible.
- Keep exact rank assertions where the test is specifically about rank calculation or rank normalization.

## Out of scope

- Any behavioral changes. This should just be a test update.

The later rank/ordering rework should happen in the upcoming TreeCRDT integration PR. This issue is only about making existing tests describe the visible ordering behavior more directly.

marcus-pousette:
Closes #4236.

This is a test-only pass that makes higher-level ordering tests assert ordered thought values instead of exact numeric `rank` values where the rank number is not the behavior under test.

Changes:
- Add `expectThoughtValues` for readable ordered thought assertions.
- Replace exact `rank` expectations in cursor/path/order behavior tests.
- Leave rank-specific expectations in tests that are directly checking rank calculation, rank preservation, or rank normalization.

### DIFF
--- a/src/selectors/__tests__/getDescendantThoughtIds.ts
+++ b/src/selectors/__tests__/getDescendantThoughtIds.ts
@@ -10,6 +10,13 @@ import childIdsToThoughts from '../childIdsToThoughts'
 import contextToPath from '../contextToPath'
 import getDescendantThoughtIds from '../getDescendantThoughtIds'
 
+/** Tests thought values without asserting order. */
+const expectThoughtsUnordered = (thoughts: { value: string }[], values: string[]) => {
+  const thoughtValues = thoughts.map(thought => thought.value)
+  expect(thoughtValues).toHaveLength(values.length)
+  expect(thoughtValues).toEqual(expect.arrayContaining(values))
+}
+
 it('get descendants', () => {
   const text = `
     - a
@@ -23,12 +30,12 @@ it('get descendants', () => {
   const descendantThoughtIds = getDescendantThoughtIds(state, HOME_TOKEN)
   const descendantsAllThoughts = childIdsToThoughts(state, descendantThoughtIds)
 
-  expect(descendantsAllThoughts.map(thought => thought.value)).toEqual(['a', 'b', 'c', 'd', 'e', 'f'])
+  expectThoughtsUnordered(descendantsAllThoughts, ['a', 'b', 'c', 'd', 'e', 'f'])
 
   const descendantsIdsOfA = getDescendantThoughtIds(state, head(contextToPath(state, ['a']) as SimplePath))
   const descendantsAThoughts = childIdsToThoughts(state, descendantsIdsOfA)
 
-  expect(descendantsAThoughts.map(thought => thought.value)).toEqual(['b', 'c', 'd'])
+  expectThoughtsUnordered(descendantsAThoughts, ['b', 'c', 'd'])
 })
 
 it('get descendants ordered by rank', () => {
@@ -45,7 +52,7 @@ it('get descendants ordered by rank', () => {
   // unordered
   const descendantsUnordered = childIdsToThoughts(state, getDescendantThoughtIds(state, HOME_TOKEN))
 
-  expect(descendantsUnordered.map(thought => thought.value)).toEqual(['a', 'b', 'c', 'x'])
+  expectThoughtsUnordered(descendantsUnordered, ['a', 'b', 'c', 'x'])
 
   // ordered
   const descendantsOrdered = childIdsToThoughts(state, getDescendantThoughtIds(state, HOME_TOKEN, { ordered: true }))
@@ -79,7 +86,7 @@ it('filter descendants', () => {
     }),
   )
 
-  expect(descendantsAll.map(thought => thought.value)).toEqual(['a', 'b', 'c', 'd', 'e'])
+  expectThoughtsUnordered(descendantsAll, ['a', 'b', 'c', 'd', 'e'])
 
   // short circuit
   // [e, f, g, h] should never be touched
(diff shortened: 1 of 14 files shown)


src/test-helpers/expectThoughtValuesInOrder.ts
```diff
@@ -0,0 +1,6 @@
+/** Test thoughts in a readable way by comparing their values in order. */
+const expectThoughtValuesInOrder = (thoughts: { value: string }[], values: string[]) => {
```
raineorshine: The most common way to assert thoughts is ordered, so let's simply call this `expectThoughts`. We can make `expectThoughtsUnordered` for the exception.

src/test-helpers/expectThoughtValuesInOrder.ts
```diff
@@ -0,0 +1,6 @@
...
+  expect(thoughts.map(thought => thought.value)).toEqual(values)
```
raineorshine: sort by rank

src/selectors/__tests__/getDescendantThoughtIds.ts
raineorshine, line 1: It's not clear from the JSDOC what order `getDescendantThoughtIds` is expected to return descendants. Each of these need to be evaluated on a case-by-case basis.

Regardless, `expectThoughtValuesInOrder` is only meant to handle siblings.

raineorshine:
Hi, thanks for creating this.

A few initial comments above.

marcus-pousette:
@raineorshine Thanks for the review. See my latest commit [a751028](https://github.com/cybersemics/em/pull/4237/commits/a7510280466bf4ff74df1cc75ea256a0969d6cd3). I hope this aligns with what you had in mind.

src/selectors/__tests__/getDescendantThoughtIds.ts
```diff
@@ -56,22 +45,12 @@ it('get descendants ordered by rank', () => {
   // unordered
   const descendantsUnordered = childIdsToThoughts(state, getDescendantThoughtIds(state, HOME_TOKEN))
 
-  expect(descendantsUnordered).toMatchObject([
-    { value: 'a', rank: 0 },
-    { value: 'b', rank: 1 },
-    { value: 'c', rank: 2 },
-    { value: 'x', rank: 1.5 },
-  ])
+  expect(descendantsUnordered.map(thought => thought.value)).toEqual(['a', 'b', 'c', 'x'])
```
raineorshine: Since the comments specifically describe this as unordered, we probably shouldn't assert the order of siblings here. Maybe `expect.arrayContaining` would be a better fit.

Ideally we probably want to  assert the order of thoughts at different depths, while ignoring the order of siblings in this case. The documentation for `getDescendantThoughtIds` leaves out this distinction. I'll leave it up to you whether you want to assert depth order while ignoring sibling order, or just do an unordered containing assertion on all descendants.

src/selectors/__tests__/getDescendantThoughtIds.ts
```diff
@@ -100,13 +79,7 @@ it('filter descendants', () => {
...
   )
 
-  expect(descendantsAll).toMatchObject([
-    { value: 'a', rank: 0 },
-    { value: 'b', rank: 0 },
-    { value: 'c', rank: 1 },
-    { value: 'd', rank: 1 },
-    { value: 'e', rank: 0 },
-  ])
+  expect(descendantsAll.map(thought => thought.value)).toEqual(['a', 'b', 'c', 'd', 'e'])
```
raineorshine: Since this is a filter test, I recommend ignoring order completely.

src/selectors/__tests__/getDescendantThoughtIds.ts
```diff
@@ -135,10 +108,5 @@ it('filter and continue traversing', () => {
     }),
   )
 
-  expect(descendantsAll).toMatchObject([
-    { value: 'b', rank: 0 },
-    { value: 'c', rank: 1 },
-    { value: 'd', rank: 1 },
-    { value: 'e', rank: 0 },
-  ])
+  expect(descendantsAll.map(thought => thought.value)).toEqual(['b', 'c', 'd', 'e'])
```
raineorshine: Similarly unordered

raineorshine (commented):
Thanks! This looks better. The existing `expectPathToEqual` is a good idea for cursor assertions.

src/selectors/__tests__/getDescendantThoughtIds.ts
marcus-pousette: Resolved https://github.com/cybersemics/em/pull/4237/commits/7d9f2a1a3de6b87792775b58c52674396d18030b

src/selectors/__tests__/getDescendantThoughtIds.ts
marcus-pousette: Resolved https://github.com/cybersemics/em/commit/7d9f2a1a3de6b87792775b58c52674396d18030b

src/selectors/__tests__/getDescendantThoughtIds.ts
marcus-pousette: Good point. I went with exact unordered membership here using expectThoughtsUnordered, so this no longer asserts sibling order in the default case. I also clarified the JSDoc to say sibling order is only guaranteed with ordered: true, and kept the exact order assertion only for that branch.

raineorshine (approved):
Thanks!

PR #4252, opened 2026-05-12
Create iOS tests for thought/note formatting

ethan-james:
These are some tests that were originally planned for https://github.com/cybersemics/em/pull/4034 but stalled out due to difficulties that we were having using BrowserStack to drive toolbar taps. I think they still add value and should be merged.


src/e2e/iOS/helpers/tapToolbar.ts
```diff
@@ -5,10 +5,12 @@ import waitForElement from './waitForElement.js'
...
  * ToolbarButton binds onTouchStart/onTouchEnd when isTouch and onMouseDown/onClick otherwise, so the toolbar is only
- * reachable with a touch pointer on a device; tap's 'mouse' default never fires the command. The y offset is the
- * Safari chrome offset used throughout this suite, since tap reads page coordinates but taps in screen coordinates.
+ * reachable with a touch pointer on a device; tap's 'mouse' default never fires the command. A touch pointer also keeps
+ * the caret where it is, since ToolbarButton preventDefaults touchend to suppress the blur (a mouse tap blurs the
+ * editable, which for a note clears `noteFocus` before the command runs). The y offset is the Safari chrome offset used
+ * throughout this suite, since tap reads page coordinates but taps in screen coordinates.
  */
-const tapOptions = { y: 60, pointerType: 'touch' } as const
+export const toolbarTapOptions = { y: 60, pointerType: 'touch' } as const
```
trevinhofmann, line 13: The `code-standards.instructions.md` says:

> - Only a single, default export is allowed. Named exports are not allowed.
>  - Exception: action-creators are co-located with reducers in `src/actions` and exported as named exports.
>  - Filenames should exactly match the default export name.

But there seem to be hundreds of instances of this already in the repo, so not sure if that's being followed anyway 🤔

trevinhofmann (approved):
One comment below, but LGTM overall.

src/e2e/iOS/helpers/tapToolbar.ts
raineorshine: The principle has been consistently eroded by agents, so the "rule" has become rather ad hoc. It might be best to abandon it at this point.

Beyond the aesthetic preference (which I can get over) the main value is in limiting the surface area of the internal API. So far, agents are actually pretty bad at being parsimonious in this regard, and tend to explode the number of subroutines. However, that should probably be expressed as a high-level architecture rule rather than low-level export restriction.

Open to your thoughts.

ethan-james: We could create a `constants.ts` file if you think it would be added to in the future.

raineorshine: One has an additional export and one has an additional file, so I'm not seeing a significance difference between the two.

raineorshine (approved):

src/e2e/iOS/helpers/tapToolbar.ts
ethan-james: I did actually ask Copilot to look for additional variables that could be extracted into `constants.ts`, and it didn't come up with anything that I thought was worthwhile.

Issue #4230, opened 2026-05-05
Tapping between a single-line thought and a multiline thought causes layout shift


ethan-james:
Tested on iOS Safari

## Steps to Reproduce

```
- A wrinkle in time
- This is a multiline thought with no formatting
```

1. Set the caret on `A`
2. Set the caret on `This is a multiline thought with no formatting`

## Expected behavior

No layout shift.

## Current behavior

There is a perceptible layout shift.

https://github.com/user-attachments/assets/11b1fc63-5df1-4276-90c2-4ac5498c849f

---

I'm guessing that this would be related to `preventAutoscroll`, but it's notable that it can be reproduced with only 2 thoughts.

ethan-james:
It doesn't seem to be related `preventAutoscroll`. I can remove it completely and see the same result. It looks like the entire layout is shifting up and down a bit, although I can't see any CSS changing. Even the `asyncFocus` editable shifts up and down.

https://github.com/user-attachments/assets/465195f0-8cdd-4fc1-bbc0-7070b978d8e3

Maybe in will go away in the course of working on #3598.

BayuAri:
@raineorshine 
Issue  is no longer reproducible on https://github.com/cybersemics/em/commit/bdcf7108b5990e0cd2fec5bde7c8bebc142c31c6 on iOS Safari but it is reproducible on iOS Capacitor. 

When caret is on active on iOS Capacitor, the layout shifts along with toolbar.
However, the toolbar does not shift when only cursor is active. 
I think this will be fixed on #4435

Issue #4014, opened 2026-04-01
[Android Capacitor] Text jumps up and down when user types on the second line of a thought


BayuAri:
## Step to Reproduce
1. Create a single line thought
2. Continue typing text on above thought on the second line of the text thought

## Current behavior
Text line jumps up and down
This might me related to #3837 

https://github.com/user-attachments/assets/cea36026-246c-4f2d-9fda-50f6713e7898

## Expected behavior
Text line should not be moving while user types

BayuAri:
Issue is fixed with https://github.com/cybersemics/em/pull/4253#issuecomment-4598242750

Issue #3589, opened 2026-01-08
Expensive callbacks, like updateSize, are executed multiple times for each thought


fbmcipher:
When `autofocus` changes, logging shows that some callbacks are being executed multiple times per thought.

As these are _expensive_ callbacks which perform layout measurements and write state, they should only be executed **once per thought**.

It's also generally a bad idea for us to have identical side effects running multiple times, as it makes it harder for us to keep track of the lifecycle for each thought and could introduce unintended bugs.

**In my testing, I found that the updateSize and updateMultiline callbacks had the most significant impact to performance**. There may well be more, but these two would be a good place to start.

Here's where each of the affected callbacks are currently called. Each hook has a different set of dependencies, but the fact we're seeing multiple executions in a single layout change indicates there must be some kind of overlap between dependencies.

- **`updateSize`**: 
    - Called in [`StaticThought.tsx:130`](https://github.com/cybersemics/em/blob/main/src/components/StaticThought.tsx#L130) by `uLAFE`
    - Called in [`VirtualThought.tsx:181`](https://github.com/cybersemics/em/blob/main/src/components/VirtualThought.tsx#L181) by `uLAFE`
    - Called in [`VirtualThought.tsx:196`](https://github.com/cybersemics/em/blob/main/src/components/VirtualThought.tsx#L196) by `useSelectorEffect`
    - Called in [`VirtualThought.tsx:201`](https://github.com/cybersemics/em/blob/main/src/components/VirtualThought.tsx#L201) by `useSelectorEffect`
    - Called in [`VirtualThought.tsx:215`](https://github.com/cybersemics/em/blob/main/src/components/VirtualThought.tsx#L215) by `useEffect`
- **`updateMultiline`**:
    - Called in [`useMultiline.ts:44`](https://github.com/cybersemics/em/blob/main/src/components/Editable/useMultiline.ts#L44) by `uLAFE`
    - Called in [`useMultiline.ts:61`](https://github.com/cybersemics/em/blob/main/src/components/Editable/useMultiline.ts#L61) by `useSelectorEffect`
    - Called in [`useMultiline.ts:78`](https://github.com/cybersemics/em/blob/main/src/components/Editable/useMultiline.ts#L78) by `useSelectorEffect`
    - Called in [`useMultiline.ts:99`](https://github.com/cybersemics/em/blob/main/src/components/Editable/useMultiline.ts#L99) by `useEffect`
[...]

Issue #3481, opened 2025-12-22
Email Icon on thought is truncated


BayuAri:
## Step to reproduce
1. Create a thought
2. Create a subthought with email address
3. Navigate focus onto the thought on step 1

## Current behavior
Email icon is truncated

<img width="543" height="256" alt="Image" src="https://github.com/user-attachments/assets/cd4f9854-f1e5-4c87-9164-399e6f9165d4" />

## Expected behavior
Email icon should not be truncated.
Refer to URL icon

BayuAri:
Issue is no longer reproducible, this is fixed with #4253 

<img width="365" height="122" alt="Image" src="https://github.com/user-attachments/assets/de8f6231-46b2-4286-86f5-7905ece0303b" />

PR #4253, opened 2026-05-12
Remove useMultiline

### LINKED ISSUE #3598: Remove useMultiline
- [ ] Depends on #3410

The [useMultiline](https://github.com/cybersemics/em/blob/61a7aaae947bea6aeaab6be08942c25f55e6db9c/src/components/Editable/useMultiline.ts) hook is a stateful representation of whether a thought wraps onto multiple lines, and is used to set the line-height. Single line thoughts have a greater line-height than multiline thoughts.

Previously we needed to use line-height rather than margin or padding to get the correct caret placement behavior. If #3410 is successful, then we will be able to do this programmatically without different line heights. We can preserve the visual spacing by simply adjusting the y position of the absolutely positioned TreeNodes.

Dissolving the line-height distinction will allow us to remove `useMultiline` and eliminate a bunch of complexity (e.g. [multiline recipe](https://github.com/cybersemics/em/blob/61a7aaae947bea6aeaab6be08942c25f55e6db9c/src/recipes/multiline.ts)) and timing issues (e.g. #3279), and ease performance issues (#3589).

ethan-james:
Fixes #3598

* Replaced `line-height: 2` with `1.25` on single-line thoughts and added padding to compensate
* Removed `useMultiline` hook
* Removed `multilineRecipe`
* Standardized superscript position for single- and multi-line thoughts
* Re-designed `Divider` with more even spacing
* Removed `lineHeightOverlap` from `useSizeTracking`
* Replaced `offsetHeight` with `getBoundingClientRect()` on touch devices


raineorshine:
@BayuAri Could you do a first pass of testing on this? This is a nearly-pure refactor, so no change of behavior is expected. This PR changes the algorithm used to calculate thought spacing, which may result in some small differences.

Test cases are here: https://github.com/cybersemics/em/issues/3598#issuecomment-4467459902

In the first pass, let's focus on any thought or bullet spacing that looks off by more than 2–3 pixels and test across 1) desktop Chrome, 2) mobile Safari, and 3) Android Chrome. In the second pass, we'll hone in on very specific pixel differences and what is acceptable (or even preferable) drift from `main`.

So the process will roughly be like this:

1.  @BayuAri 1st pass + @ethan-james Iterate
2. @raineorshine 1st pass + @ethan-james Iterate
3.  @BayuAri 2nd pass + @ethan-james Iterate
4. @raineorshine 2nd pass + @ethan-james Iterate

Thanks!

raineorshine:
@ethan-james I took a quick look through all the snapshot diffs.

The shifted 2nd thought, described in https://github.com/cybersemics/em/issues/3598#issuecomment-4513580907, is odd, but as long as it is not accumulating down the page or shifting between renders it should be fine.

I've added an additional render-thoughts test (at multiple font sizes) in 941f10bd27 for long lists of identical siblings. This will give us test coverage of the relative position of the thought to its bullet as `y` accumulates down the page. Once you merge `main` we can analyze the snapshot diff.

The only snapshot test I saw that is for sure broken is `/render-thoughts/superscripts-table1-diff.png`:

superscripts-table-1-diff

BayuAri:
@raineorshine @ethan-james 
This looks good on my end. 

image

Below issues are no longer reproducible on this branch
- [Email Icon on thought is truncated #3481](https://github.com/cybersemics/em/issues/3481) 
- [[Android Capacitor] Text jumps up and down when user types on the second line of a thought #4014](https://github.com/cybersemics/em/issues/4014)

raineorshine:
@BayuAri What platform + browser is that on? The bullet is too high relative to the thought.

Compare with `main` on Desktop Chrome:
image

BayuAri:
@raineorshine 
You are right, the bullet is a bit too high.
Sorry for missing this. 
It is too high on Mobile Safari and iOS Capacitor

image

ethan-james:
> @raineorshine You are right, the bullet is a bit too high. Sorry for missing this. It is too high on Mobile Safari and iOS Capacitor
> 
> image

Interesting, that must be what [marginTop](https://github.com/ethan-james/em/blob/6db5541a2c24cc94d77956386f432c54b27cd87b/src/recipes/editable.ts#L19) was doing in `main`. I don't really see it on mine, what version of iPhone are you using for the mobile app, @BayuAri?

image0 (11)

BayuAri:
@ethan-james 
The screenshot is from iPhone 14PM iOS 26.4.2 and Samsung A56 OS 16

BayuAri:
@ethan-james 
The difference of the bullet position is more visible on Desktop Chrome. 

image

ethan-james:
> @ethan-james The difference of the bullet position is more visible on Desktop Chrome.
> 
> image

Aha, I do see that the positioning is a bit different on Desktop Chrome on Mac compared to Windows, where I've been doing my testing. It looks to me like there are actually 3 different things going on here.

1. Bullet position for multiline thoughts has changed significantly, with approval from @raineorshine 
2. Bullet position for all thoughts is slightly too high on desktop Mac
3. Bullet position for all thoughts is dramatically too high on iPhone 14PM iOS 26.4.2

As I mentioned, item 3 is probably related to the `marginTop` that used to exist in the editable recipe. Item 2 is trickier. I tried to restore the line-height-based bullet position as well as the -0.5px negative top margin that existed in the editable recipe on `main`. That seemed to move things in the wrong direction, making it look closer to how it looks on iPhone 14PM, even though it should have been pushing the bullet down and pulling the editable up. There's also the issue that any non-line-height-based changes that I make affect all platforms equally, which pushes things out of position on Windows Chrome and my iPhone 12.

It's a tricky balance and I'll need to keep playing with it.

ethan-james:
> @ethan-james I took a quick look through all the snapshot diffs.
> 
> The shifted 2nd thought, described in [#3598 (comment)](https://github.com/cybersemics/em/issues/3598#issuecomment-4513580907), is odd, but as long as it is not accumulating down the page or shifting between renders it should be fine.
> 
> I've added an additional render-thoughts test (at multiple font sizes) in [941f10b](https://github.com/cybersemics/em/commit/941f10bd278d5b328e324fb51577dac7c186e484) for long lists of identical siblings. This will give us test coverage of the relative position of the thought to its bullet as `y` accumulates down the page. Once you merge `main` we can analyze the snapshot diff.
> 
> The only snapshot test I saw that is for sure broken is `/render-thoughts/superscripts-table1-diff.png`:

OK, I fixed the table test and tweaked editable and bullet positioning slightly in a way that will hopefully look better on Mac and iOS. @BayuAri can be the judge of that, especially if the bullets are still farther off on iPhone 14PM. I didn't notice anything like that during my testing.

The positioning in the snapshot tests changed slightly, but I would describe it as equivalent or slightly better than the way it was failing previously.

BayuAri (commented):
Compare to `main` ,  the bullet positioning is lower on the `PR`, however, the position is more centered than `main`
Should we go with this new one?
@raineorshine 
Please advice

Image

ethan-james:
> Compare to `main` , the bullet positioning is lower on the `PR`, however, the position is more centered than `main` Should we go with this new one? @raineorshine Please adviceIf 
> 
> Image

If it's consistent across different Chrome browsers, then I should be able to target those and bump it up a bit.

raineorshine:
I've documented the discrepancies below. "too high" means the text is too high relative to its bullet.

I draw a horizontal line straight through the center of the bullet and see where it falls on the text. On the glyph "C", the line should fall 1px below the vertical center.

I can't tell you the exact pixel discrepancies because the images are scaled down in this composite screenshot.

image


> Compare to `main` , the bullet positioning is lower on the `PR`, however, the position is more centered than `main` Should we go with this new one? @raineorshine Please advice

You need to be more specific about which platform you are talking about please.

ethan-james:
> I've documented the discrepancies below. "too high" means the text is too high relative to its bullet.
> 
> I draw a horizontal line straight through the center of the bullet and see where it falls on the text. On the glyph "C", the line should fall 1px below the vertical center.
> 
> I can't tell you the exact pixel discrepancies because the images are scaled down in this composite screenshot.

Great, thank you for that. I'll see what I can do about desktop Chrome, and then see what I can do about ~~mobile Chrome~~ Android without breaking desktop.

ethan-james:
> > I've documented the discrepancies below. "too high" means the text is too high relative to its bullet.
> > I draw a horizontal line straight through the center of the bullet and see where it falls on the text. On the glyph "C", the line should fall 1px below the vertical center.
> > I can't tell you the exact pixel discrepancies because the images are scaled down in this composite screenshot.
> 
> Great, thank you for that. I'll see what I can do about desktop Chrome, and then see what I can do about ~mobile Chrome~ Android without breaking desktop.

I did a bunch of measuring on Windows, Mac, iPhone and Android, and I think that targeting `_iphone` and setting the negative margin (that I brought over from `main`) only on that platform is the way to go. @BayuAri Let me know if it worked, as I'm still not 100% sure that I'm seeing the exact same measurements as you.

@raineorshine I wasn't able to easily get an Android build running. I got as far as:

```
Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.
```

I can keep trying, or if you'd prefer, you can give me a prescription for fixing the positioning on Android. I don't think the `_android` Panda target will be targeted enough, but I may be able to use `isCapacitor()` in conjunction with that.

raineorshine:
If you'd like to punt on the Android position, that's fine. It doesn't change the core solution, and should be as easy as changing a few pixel values. Just make a note of where in the code the Android positioning can be tweaked.

ethan-james:
> If you'd like to punt on the Android position, that's fine. It doesn't change the core solution, and should be as easy as changing a few pixel values. Just make a note of where in the code the Android positioning can be tweaked.

I'd say so, yes. It doesn't look like it's worse off than it was before. Hopefully in the future I'll figure out how to do an Android build without needing to sink a bunch of time into it right now. I can add a comment in the editable recipe.

BayuAri:
### iOS: Pass

image

###  Mobile Safari: Pass
image

### Android: Pass
image

###  Mobile Chrome: Too low on PR/4253
image


### Desktop Chrome:  Too low on PR/4253
image


Bullet position on PR/4253 are now centered and balanced on iOS, Mobile Safari and Android.
It is a little too low on Mobile and Desktop Chrome.

ethan-james:
> Bullet position on PR/4253 are now centered and balanced on iOS, Mobile Safari and Android.
> It is a little too low on Mobile and Desktop Chrome.

@BayuAri Maybe I was wrong about only applying the negative margin on iPhone. How does it look now? Keep in mind that the positioning of multiline thoughts has intentionally changed between `main` and this branch, so the positioning of multiline thoughts on `main` is not always a good baseline. In your mobile Chrome screenshot, I see that the multiline thought is too high on `main`, although it looks pretty good in the desktop Chrome screenshot.

BayuAri:
@ethan-james 
Here is the comparison on the PR/4253


iOS Capacitor: Pass
Android Capacitor:  Text is a little too high against the bullet.


image





Mobile Safari: Pass
Mobile Chrome: Pass
Desktop Chrome: Pass
image

src/components/BulletPositioner.tsx
```diff
@@ -328,9 +328,9 @@ const BulletPositioner = forwardRef<SVGSVGElement, PropsWithChildren<BulletPosit
         style={{
           top: -extendClickHeight,
           left: -extendClickWidth + marginLeft,
-          paddingTop: extendClickHeight,
+          paddingTop: '0.708em',
           paddingLeft: extendClickWidth,
-          paddingBottom: extendClickHeight + 2,
+          paddingBottom: 'calc(0.583em + 2px)',
```
raineorshine: Extending the click height of an element relies on exactly offsetting padding with negative top/left. This ensures that the net positional change is 0,0, but there is more area to click.

It doesn't make sense to adjust padding independently from top/left. These are intentionally tightly coupled and have to be adjusted together.

src/recipes/editable.ts
```diff
@@ -4,23 +4,17 @@ const editableRecipe = defineRecipe({
   className: 'editable',
   description: 'editable styles',
   base: {
+    minHeight: '1.25em',
```
raineorshine: Why specify a `minHeight`? I would expect the natural height and padding to determine the height.

raineorshine (changes requested):
Looking great!

I reviewed all the snapshots and the differences seem inconsequential. Feel free to update them. Also tested manually on iPhone at a couple different font sizes.

I can already tell a couple [multiline bugs](https://github.com/cybersemics/em/issues/?q=is%3Aissue%20state%3Aopen%20multiline%20label%3Abug) will be fixed by this.

src/recipes/editable.ts
ethan-james: Since iOS Safari rounds down the line-height, a single-line thought will have an overall height of 33.25px at 18px font size, rather than 33.75px. Giving it the `minHeight` helps keep the thoughts the same distance apart. However, it wasn't working as intended due to `boxSizing: border-box`, so it would probably be best if I removed it.

src/components/BulletPositioner.tsx
```diff
@@ -328,9 +328,9 @@ const BulletPositioner = forwardRef<SVGSVGElement, PropsWithChildren<BulletPosit
         style={{
           top: -extendClickHeight,
           left: -extendClickWidth + marginLeft,
-          paddingTop: extendClickHeight,
+          paddingTop: `calc(0.375em + ${extendClickHeight}px)`,
           paddingLeft: extendClickWidth,
-          paddingBottom: extendClickHeight + 2,
+          paddingBottom: `calc(0.25em + ${extendClickHeight + 2}px)`,
```
raineorshine: That's better, thanks.

If 0.375em and 0.25em correspond to the padding in the editable recipe, we need to tie them together programmatically: either using PandaCSS tokens or JS constants. Hardcoding identical values gives the false impression that they are independent and is prone to drift.

src/components/Thought.tsx
```diff
@@ -167,7 +167,7 @@ const useCol1Alignment = ({ path, value, isTableCol1 }: UseCol1AlignParams) => {
     return cursorParentId ? getChildren(state, cursorParentId).map(t => t.value) : []
   }, shallowEqual)
 
-  interface TransitionStyle {
+  interface TransitionStyle extends CSSProperties {
```
raineorshine: This is fixed in `main` now.

raineorshine (changes requested):

raineorshine (approved):
Looks good! Is this ready to merge?

ethan-james:
> Looks good! Is this ready to merge?

I think so! I updated the snapshots, and we were just going back and forth on bullet positioning.

raineorshine:
Deleting code is the best 🤓

@BayuAri Could you re-test each of the open multiline issues on `main`? Comment "Still reproducible in {COMMIT_HASH}" or "No longer reproducible in {COMMIT_HASH}" and close the issue as resolved.

BayuAri:
@raineorshine 

#3279 is no longer reproducible on https://github.com/cybersemics/em/commit/bdcf7108b5990e0cd2fec5bde7c8bebc142c31c6

#2845 is **still** reproducible on https://github.com/cybersemics/em/commit/bdcf7108b5990e0cd2fec5bde7c8bebc142c31c6

#4230 is no longer reproducible on https://github.com/cybersemics/em/commit/bdcf7108b5990e0cd2fec5bde7c8bebc142c31c6

#3589  is no longer reproducible on https://github.com/cybersemics/em/commit/bdcf7108b5990e0cd2fec5bde7c8bebc142c31c6

raineorshine:
Unfortunately I had to revert due to a regression.

Tested on Mobile Safari.

## Steps to Reproduce

```
- A
- B
- C
- D
- E
- F
  - G
    - H
```

1. Set the cursor on `G` and enter edit mode.
2. Swipe ← to activate Cursor Forward.
3. Swipe → to activate Cursor Back.

## Current Behavior

69fc54820234a817bd3819cd13aac3bb924df284

The cursor thought jumps up for one frame.

## Expected Behavior

`69fc548^`

The cursor thought does not jump on Cursor Back/Forward.

ethan-james:
> Unfortunately I had to revert due to a regression.
> 
> Tested on Mobile Safari.
> 
> ## Steps to Reproduce
> ```
> - A
> - B
> - C
> - D
> - E
> - F
>   - G
>     - H
> ```
> 
> 1. Set the cursor on `G` and enter edit mode.
> 2. Swipe ← to activate Cursor Forward.
> 3. Swipe → to activate Cursor Back.
> 
> ## Current Behavior
> [69fc548](https://github.com/cybersemics/em/commit/69fc54820234a817bd3819cd13aac3bb924df284)
> 
> The cursor thought jumps up for one frame.
> 
> ## Expected Behavior
> `69fc548^`
> 
> The cursor thought does not jump on Cursor Back/Forward.

OK, I think I figured this out in #4440.

raineorshine:
Re-merged in #4440.

PR #4254, opened 2026-05-13
Layout shift tests

### LINKED ISSUE #3672: Test that thoughts do not shift y position after first render

karunkop:
Close #3672 

New PR resolving timeout issue for layout shift test in CI!

Previous PR: https://github.com/cybersemics/em/pull/4182


raineorshine (changes requested):
Hi, thanks for identifying the race condition. I have a some feedback about the design.

1. `yieldForPendingEvaluate` is arguably an implementation detail of `measureYShift`. It's not clear why it should be independent or exposed to the test cases.
2. Some tests call `yieldForPendingEvaluate` and some do not, without a clear explanation of why. For developers looking to use `measureYShift`, this is a bad situation, as they won't know whether to call `yieldForPendingEvaluate` or not. It should ideally be handled automatically or deterministically.
3. This is the first time we've used a no-op `page.evaluate` as a form of sleep. Ideally we poll and wait for the exact condition we want. Can we explicitly check that the observer is registered? If not, then we resort to a timing-based wait. We can use `queueMicrotask`, `setTimeout`, or `requestAnimationFrame` (in that order). If none of those works, we can use a `sleep`, that as you know that is a worst case scenario since it's arbitrary, prone to race conditions under load, and slows down the tests for everyone. Given all of these existing wait methods, why are we introducing a new one with `yieldForPendingEvaluate`? Introducing a wait method that crosses the `page` boundary is also unusual. Why can't we wait inside the page?

karunkop:
> Hi, thanks for identifying the race condition. I have a some feedback about the design.
> 
> 1. `yieldForPendingEvaluate` is arguably an implementation detail of `measureYShift`. It's not clear why it should be independent or exposed to the test cases.
> 2. Some tests call `yieldForPendingEvaluate` and some do not, without a clear explanation of why. For developers looking to use `measureYShift`, this is a bad situation, as they won't know whether to call `yieldForPendingEvaluate` or not. It should ideally be handled automatically or deterministically.
> 3. This is the first time we've used a no-op `page.evaluate` as a form of sleep. Ideally we poll and wait for the exact condition we want. Can we explicitly check that the observer is registered? If not, then we resort to a timing-based wait. We can use `queueMicrotask`, `setTimeout`, or `requestAnimationFrame` (in that order). If none of those works, we can use a `sleep`, that as you know that is a worst case scenario since it's arbitrary, prone to race conditions under load, and slows down the tests for everyone. Given all of these existing wait methods, why are we introducing a new one with `yieldForPendingEvaluate`? Introducing a wait method that crosses the `page` boundary is also unusual. Why can't we wait inside the page?

You are right, the previous solution of using `yieldForPendingEvaluate` will cause confusions down the line. I have implied a hack that has effectively solved the timeout issue with the test. Basically introduced a testFlag in window.em.testFlags that will keep track of whether the mutation observer for layout shift is ready to start observing changes. The test in `layout-shift.ts` now waits until mutation observer is observing the document.

karunkop:
As for TDD checks failing, the TDD workflow currently checks out the base commit and only cherry-picks the updated test files (`src/e2e/puppeteer/__tests__/*.ts`) from the pr head. Because supporting e2e files like `src/e2e/testFlags.ts` are not included, the base ends up using stale type definitions, which causes the typeScript build to fail.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,209 @@
...
+
+describe('thought y position stability', { retry: 3 }, () => {
+  it('new thought should not shift y position after first render', async () => {
+    await paste(`
+        - a
+      `)
+    await waitForEditable('a')
+
+    const measurePromise = measureYShift('')
+    await waitForLayoutShiftObserver()
```
raineorshine: `waitForLayoutShiftObserver` is called if and only if `measureYShift` is called, so they should be combined. Separating them avoids the awkwardness of a two-step await, but at the expense of exposing `layoutShiftObserverReady`, which is more of an implementation detail of `measureYShift`.

A two-step await can be achieved by returning an object with the second promise:

```
// waits for layoutShiftObserverReady
const { measurePromise } = await measureYShift('')
...
await measurePromise
```

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,209 @@
...
+                }
+              }
+            } else if (target && m.type === 'attributes' && m.target === target) {
+              const oldTop = parseTop(m.oldValue)
+              const newTop = parseTop(target.getAttribute('style'))
+              if (oldTop === newTop) continue
+              // On first style change, use oldValue as the true initial `top`; later changes skip this, and if values match again, the update has no effect.
+              if (after === before) before = oldTop
+              after = newTop
+              resolveDebounced()
```
raineorshine: Can you remind me the purpose of the debounce logic and waiting for the y value to "settle"? Shouldn't the test fail as soon as the `y` value changes?

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,209 @@
...
+        })
+
+        observer.observe(document.body, {
+          childList: true,
+          subtree: true,
+          attributes: true,
+          attributeFilter: ['style'],
+          attributeOldValue: true,
+        })
+        em.testFlags.layoutShiftObserverReady = true
```
raineorshine: I'm confused by the timing of setting `layoutShiftObserverReady`. If it is only set to true synchronously as soon as `measureYShift` is called, what's the point of `waitForLayoutShiftObserver`?

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,209 @@
...
+          return m ? parseFloat(m[1]) : null
+        }
+
+        /** Schedule to resolve and disconnect the observer after `settleMs` milliseconds. */
+        const resolveDebounced = () => {
+          if (resolveDebounceTimer) clearTimeout(resolveDebounceTimer)
+          resolveDebounceTimer = setTimeout(() => {
+            observer?.disconnect()
+            em.testFlags.layoutShiftObserverReady = false
+            resolve({ thoughtValue: thoughtValue, before, after })
```
raineorshine: Can `before` be determined before the user action? The intuitive sequence would be:

1. measure (before)
2. install observer
3. action
4. measure (after)

I'm not sure if separating measure (before) from installing the observer makes sense, but having the final promise resolve the `before` measurement seems strange.

raineorshine (changes requested):

raineorshine:
> As for TDD checks failing, the TDD workflow currently checks out the base commit and only cherry-picks the updated test files (`src/e2e/puppeteer/__tests__/*.ts`) from the pr head. Because supporting e2e files like `src/e2e/testFlags.ts` are not included, the base ends up using stale type definitions, which causes the typeScript build to fail.

Thanks, this is a flaw in the TDD workflow. I'll have to think about whether we should patch test helpers changes as well, or just log a warning that says we can't cleanly run the test changes on the base.

src/e2e/puppeteer/__tests__/layout-shift.ts
karunkop: The issue  is that a thought's tree node is first placed at one position, then after layout runs again, it moves to a different position. So we're interested in the node's initial position versus its final position, not whether it changed at any point during the process. 

`resolveDebounced` waits until the tree node stops changing for 500ms before resolving. This ensures we capture the node's final position after all layout updates have finished, rather than reacting to the first position React assigns. As a result: before = the initial top value assigned to the node and after = the final top value once all layout adjustments are complete. Without the debounce, we'd likely resolve on the first mutation and miss any later position updates, making it impossible to detect the actual visual jump the user sees.

src/e2e/puppeteer/__tests__/layout-shift.ts
karunkop: `waitForLayoutShiftObserver` exists because `page.evaluate()` is asynchronous across two runtimes.

When the test calls `page.evaluate()`, node immediately gets back a promise, but the browser may not have executed the callback yet. Inside that callback, the `MutationObserver` is created and `layoutShiftObserverReady` is set to true.

`waitForLayoutShiftObserver` simply waits until the browser has finished that setup, ensuring the observer is ready before the test continues.

src/e2e/puppeteer/__tests__/layout-shift.ts
karunkop: For tests that create a new thought with Enter, there is no tree node to measure before the action because the node doesn't exist yet. That's why the observer must be set up first. When Enter creates the node, the observer captures its initial top position, then waits for any layout updates to finish before capturing the final position. But i understand your point about having the final promise resolve the `before` measurement thing. As of now I can't think of a better architecture yet about this.

src/e2e/puppeteer/__tests__/layout-shift.ts
raineorshine: Let me try to ask this a different way. Is there ever a case where the node is rendered at position A, then position B, then position A again? That's the only circumstance where A != Z would differ from A != B.

src/e2e/puppeteer/__tests__/layout-shift.ts
raineorshine: With the two-step await, the observer will be set up after the first await. I don't think `layoutShiftObserverReady` itself makes a difference.

src/e2e/puppeteer/__tests__/layout-shift.ts
raineorshine: I see, thanks. I misunderstood `before` to mean before the action. It really indicates the first y value of the thought after the action is executed.

src/e2e/puppeteer/__tests__/layout-shift.ts
karunkop: I don't think so. I have updated the code to remove the debounce logic completely

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,194 @@
...
+              const newTop = parseTop(target.getAttribute('style'))
+              if (oldTop === newTop) continue
+              // On first style change, use oldValue as the true initial `top`; later changes skip this, and if values match again, the update has no effect.
+              if (after === before) before = oldTop
+              after = newTop
+            }
+          }
+          // Resolve the promise and disconnect the observer.
+          if (settled || !target) return
+          settled = true
```
raineorshine: What's the point of `settled` if we disconnect the observer and resolve the promise immediately?

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,194 @@
...
+  // but page.evaluate is async across CDP. Yield one in-page microtask so that setup
+  // completes before the test dispatches the user action.
+  await page.evaluate(() => new Promise<void>(resolve => queueMicrotask(resolve)))
+
+  return { measurePromise }
+}
+
+/** Expect the y position of the thought to be stable within the tolerance. */
+const expectStableY = ({ thoughtValue, before, after }: YShiftResult) => {
+  const delta = (after ?? 0) - (before ?? 0)
```
raineorshine: Defaulting `before` and `after` to 0 will result in a false positive (incorrect pass) if both are still null. Instead we want to fail the test.

raineorshine (changes requested):
Thanks, that looks better.

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,193 @@
...
+      new Promise<YShiftResult>((resolve, reject) => {
+        /** The target tree-node whose y position is being measured. */
+        let target: HTMLElement | null = null
+        /** The first inline `top` React ever assigned to the target tree-node. */
+        let before: number | null = null
+        /** The latest inline `top` on the target tree-node when the observer resolves. */
+        let after: number | null = null
+
+        /** The mutation observer to observe the target tree-node. */
+        let observer: MutationObserver | null = null
```
raineorshine: `const`

src/e2e/puppeteer/__tests__/layout-shift.ts
```diff
@@ -0,0 +1,193 @@
...
+              if (oldTop === newTop) continue
+              // On first style change, use oldValue as the true initial `top`; later changes skip this, and if values match again, the update has no effect.
+              if (after === before) before = oldTop
+              after = newTop
+            }
+          }
+          // Resolve the promise and disconnect the observer.
+          if (!target) return
+          observerInstance.disconnect()
+          resolve({ thoughtValue, before, after })
```
raineorshine: I'm not sure why this is placed after the mutation loop. The async behavior is very simple:

1. Record the `target` and `before` on the first treeNode mutation.
2. Wait for `top` to change.
3. When `top` changes, resolve the promise.
4. Otherwise let it time out.

You don't need to break from the loop to resolve the promise. That seems to be based on the old settlement logic.

raineorshine (changes requested):

raineorshine (approved):

PR #4255, opened 2026-05-13
Clean up BrowserStack CI logs + add iOS console proxy

fbmcipher:
*No associated issue.*

## Background

This PR tidies up the log output for BrowserStack CI runs by **turning down the log verbosity** and **outputting per-`it()` log groups with clear ticks and crosses** so humans and agents alike can more easily parse their results.

It also uses a "console proxy" to dump console output during tests into CI logs, which is necessary because BrowserStack does not support reading the console.

## Console proxy

This feels like an odd limitation, but BrowserStack specifically does not support accessing the iOS console:

CleanShot 2026-05-13 at 07 42 36@2x

To work around this, we [create a basic proxy around the console.\* functions]() which caches log output in a `sessionStorage` variable *as well as* outputting to console. This behaviour is selectively enabled using the `?__ios_console_proxy` query parameter.

After each test, we then "drain" the buffer, which outputs all of that test's logs at the end of a test.

This mechanism will also be useful for agents to iterate on issues, as it will give them the ability to, say, evaluate an expression, log it to the console, and drain the buffer to access the result of the expression.

Admittedly, integrating a mechanism like this was not my first preference, so I took care to make sure it was at least as non-intrusive as possible, and opt-in. Please let me know if you can see a more elegant way to make this work!

## Miscellaneous

- To ensure it gets executed first, we import the console proxy at the very top of `index.tsx`. This violates the order set by `@trivago/prettier-plugin-sort-imports`, so `.prettierc.json` gets a slight modification to allow us to import the console proxy first.


src/util/iOSConsoleProxy.ts
raineorshine, line 1: This is a generic console proxy that is not iOS specific, so the iOS namespacing is misleading. Generally you want to try to avoid having callees reference information about callers they don't need to know about. This keeps functions flexible and able to be used in as broad a context as possible, as well as avoid implying dependencies that do not exist.

src/e2e/iOS/config/wdio.base.conf.ts
raineorshine, line 1: Direct, type-unsafe access to `?__ios_console_proxy` and `__drainiOSConsoleLogs__` is not ideal. I realize it's a bit challenging when dealing with a different execution context, but is there a more typesafe approach that hides more of the implementation details and doesn't break if the proxy gets refactored?

raineorshine (changes requested):
Thanks!

src/e2e/iOS/config/wdio.base.conf.ts
fbmcipher: How about this:

- `VITE_BROWSER_CONSOLE_CAPTURE=1` flag passed at build-time and run-time, instead of the `?__ios_console_proxy` parameter. Cleaner, and we can leverage tree-shaking to ensure the proxy doesn't get included in builds unless it's needed.
- Full type-safety in the `src/util/consoleProxy.ts` module.
- Rather than storing logs in the global `window` context, store them at module level.
- Import and call new helper function `drainConsoleProxy` from `wdio.browserstack.conf.ts`, rather than directly accessing the window object. This makes sure all implementation details for the console proxy stay in `src/util/consoleProxy.ts`, rather than spilling over into other modules.

src/util/iOSConsoleProxy.ts
fbmcipher: Thank you, that's very true. I've removed the iOS namespacing across the board to avoid making that incorrect implication.

src/e2e/iOS/config/wdio.base.conf.ts
raineorshine: Yes, that's better!

raineorshine (approved):

raineorshine:
@karunkop Please note this change to the BrowserStack test suite.

PR #4280, opened 2026-05-19
Drag and drop favorite bug

### LINKED ISSUE #3868: Flaky test: Drag and drop favorite thoughts timing out issue in puppeteer.
- https://github.com/cybersemics/em/actions/runs/22460881168/job/65054517727

The above timing out issue has been seen multiple times lately in CI.

Image

karunkop:
Close #3868

The drag-and-drop timeout issue for favorites no longer appears to occur after adopting the same sidebar wait approach used when opening favorites in the sidebar. While the root cause of the issue has still not been fully identified, this refactor is an improvement regardless. I’ve also created a separate helper function for opening the sidebar so that other tests can reuse the same approach when needed.

The CI check results also look promising: https://github.com/karunkop/em/actions?query=branch%3Adrag-and-drop-favorite-bug

That said, we’ll continue monitoring the issue. If it persists, we can reopen the investigation.


raineorshine (approved):
Thanks!

PR #4310, opened 2026-05-26
feat(skills): give the agent iOS automation via BrowserStack App Automate

fbmcipher:
## Background

Part of our strategy to improve agents' engineering capacity has been to give AI the ability to **directly interact with a browser** so that it can reproduce issues, validate hypotheses, and test submissions without just submitting based on a *vibe* that they've solved the problem.

A large portion (majority?) of the issues on the project actually pertain to iOS. You could consider iOS one of the main platforms for **em**; @raineorshine daily drives an iPhone, and much of the work done on **em** specifically targets iOS due to its quirky web implementation.

So expanding AI's agency even further by giving it the ability to control a headless iOS device was a natural step to take.

This could:

- **Improve first-pass implementation from AI agents.** Along with the Worker Bee's agent guidelines and the `issue-repro` skill, agents are forced to verify both the failing and passing test cases in a GitHub issue. This reduces the chance of a bad first implementation, reducing code review burden.

- **Expands the complexity ceiling for issues we can assign to AI to fix.**

- **Begins to erode one of the highest friction issue sets in the project.** iOS-specific quirks are **a perpetual source of subtle bugs** that have historically required hands-on debugging on a physical device, slowing iteration and concentrating the work on a small number of contributors.

## Approach

Leverage our existing unlimited BrowserStack subscription. Run an iOS device in the cloud. Use the `wdio` MCP server to give the agent the ability to control and observe it.

The work happened in roughly this order, each phase building on the previous. A _lot_ of work got done in this PR...

### 1. Extract browser interaction into reusable skills

We started with a single, monolithic **`issue-repro`** skill that owned everything: navigate to the dev server, dispatch gestures, take screenshots, all of it. As issue reproduction matured, that became too narrow — other workflows might want to drive a browser too.

So we extracted a "router" skill **`browser-control`**, with two lazy-loaded sub-skills:

- **`browser-control-chrome`** — desktop web and Android mobile-emulated Chrome, via the `chrome-devtools` MCP.
- **`browser-control-ios`** — real-device iOS, via the `wdio` MCP on BrowserStack.

Now, any agent/prompt can take full advantage of browser control by just asking it ("test this issue in the browser") without needing to use `issue-repro` (which only works with Steps to Reproduce)

### 2. Pick BrowserStack **App Automate** (not Automate)

For the iOS path we deliberately chose App Automate over ordinary Automate. The difference matters: regular Automate only runs the page in **Mobile Safari** on the BrowserStack device — purely a web context – BrowserStack limits access to system chrome (even though Appium does technically provide it).

*App* Automate runs the **real Capacitor app** on a real iPhone / iPad, and exposes both contexts in a single session: native (XCUITest) and web (WKWebView). That gives the agent access to the keyboard, autocorrect, text-selection handles, the native edit menu, scroll physics, the status bar, and any native dialog or share sheet – all the iOS UI surfaces that simply don't exist in a web DOM.

I've uploaded a **server-mode build** of **em**'s Capacitor app to BrowserStack. On launch it loads its assets from the runner's dev server over the BrowserStack Local tunnel — so day-to-day code changes don't need an rebuild.

*(If/when we have CI for native Capacitor builds, we'll want to refresh that stored build periodically too.)*

### 3. Session lifecycle management

BrowserStack is designed more for *end-to-end testing*, not for *agentic interaction* (though it works quite well for that!)

So there were a couple issues we needed to solve:

- **Heartbeat**. BrowserStack idle-kills sessions after a few minutes of no commands. The agent often thinks for longer than that. A `heartbeat.sh` daemon pings BrowserStack every 90s while a session is running, which helps keep the session alive.

- **Session-death recovery**. If a session _does_ die mid-run, the agent is instructed to spit out logs so we can figure out what happened, then restart the session properly.

### 4. iOS console capture

BrowserStack doesn't expose iOS console logs over the wire, so we developed an console proxy to work around the limitation. We already merged this!

### 5. Native double-tap-to-select-word (and other system interactions)

In general, the MCP server's `tap_element` tool works fine for touching the screen, but we had some trouble getting the same tool to work for text selection.

With AI, I ran ~15 controlled tests against a live App Automate session, ruling out each interaction Appium gives us to tap – until finding the `/touch/perform` route (that goes through `XCUICoordinate.tap`), which sent the touch event at the appropriate level system level:

This shipped as two new sub-skills:

- `browser-control-ios-touch-fallback` — Documents the JSONWP action primitives.
- `interaction-ios-select-text` — A skill that explains to an agent *how* to select text with touch, should it need to.

Works:

CleanShot 2026-05-26 at 16 34 57@2x

## Conclusion & Next Steps

Agents now have the ability to drive iOS & Chrome instances headlessly. We should now pick out 5-10 issues to assign to AI to fix autonomously and perform more in-depth testing.

We can then either continue to make improvements to the *implementation* phase of the agent; or, if results are promising enough, start turning up the dial on how many issues we assign to AI and start solving issues at the *code review* end.


raineorshine (approved):
Amazing work! Excited to see how this performs.

There's a lot going on here. I have a few points related to determinism and agent infrastructure management which we can discuss at our next AI strategy meeting.

PR #4316, opened 2026-05-28
Refactor: Extract `CommandGrid` from `CommandTable`, and new shared hook `useCommandList`

### LINKED ISSUE #4149: Extract CommandGrid components from CommandTable
## Background

When the first Gesture Cheatsheet was implemented, `CommandTable` was the primary view in **em** that listed and grouped commands. So the decision was made to *extend* the tabular `CommandTable` with a new `grid` view option. This reduced duplication at the time.

Since then, the Liminal UI design work has significantly evolved the grid view. It is now central to the cross-platform Command Universe – which is now the primary surface for browsing commands in **em**.

Also, `CommandItem` and `CommandTable` both have extensive `viewType === 'grid'` branches for layout and spacing divergences. At that level of branching, the cleaner approach is to just split out these incompatible designs into their own components.

## Approach

Rather than extending `CommandTable`, let each command view own its specific UI/presentation. We can do this while avoiding duplication by **decoupling logic from presentation**:

- **Inline logic → refactor into shared hooks** (filtering, search state, sort state, interactions)
- **Presentation → separated, composable components** (item layout, group layout, search button, sort button)

### Extract `CommandUniverseGrid`

Break out the `grid` prop path from `CommandTable` into a new `CommandUniverseGrid` component.

- Create `CommandUniverseGrid` as a focused grid primitive. Its props are `{ commands, selectedCommand?, onSelect?, onHover? }` 
- Essentially it receives **pre-filtered commands and just renders them**. This matches how `DesktopCommandUniverse` is currently organized.
- **UI chrome like the search bar & sort button move *out* of the grid and *into* the modal.**
- In `MobileCommandUniverse`, consume `<CommandUniverseGrid />` instead of `<CommandTable viewType='grid' />`.
- Remove the `viewType` prop from `CommandTable` and all grid branches in `CommandTable`, `CommandTableContent`, `CommandTableOnly`, and `CommandsGroup`.

### Split `CommandItem` into `CommandTableItem` and `CommandUniverseGridItem`

Like `CommandTable`, `CommandItem` is responsible for both grid and table presentations. Because grid and table views have a completely different DOM structure and styles, in practice there are essentially *two different components here that are being joined into one*.

So again we should **split these into two components, while sharing genuinely reusable logic via hooks or exported functions.**

fbmcipher:
_Closes #4149_.

This is a pure refactor, so functionality is unchanged and tests continue to pass.

## Background

When the first Gesture Cheatsheet was implemented, `CommandTable` was the primary view in **em** that listed and grouped commands. So the decision was made to *extend* the tabular `CommandTable` with a new `grid` view option.

This reduced duplication at the time, but as a result `CommandItem` and `CommandTable` were left with complex `viewType === 'grid'` branches for layout and spacing divergences. This is a maintainability problem, but also makes development of the grid-based Command Universe much more difficult.

At this level of branching, the cleaner approach is to just split out these incompatible designs into their own components. Any genuinely reusable logic should be shared via hooks.

Rather than extending `CommandTable`, let each command view own its specific UI/presentation. We do this while avoiding duplication by **decoupling logic from presentation**.

## Changes

- A new `useCommandList` hook which centralizes logic for sort, search and command grouping. This can be used by any component that needs to consume components or provide that functionality.
- `CommandTable` and `CommandTableItem` go back to _just being responsible for the tabular representation of commands_, as currently seen in Help and Customize Toolbar. This means we can remove all the complex `viewType` conditional logic. Since search and sort logic is now consumed from `useCommandList`, `CommandTable` reads much more cleanly.
- New `CommandUniverseGrid` and `CommandUniverseGridItems` components, used in the Gesture Cheatsheet / Mobile Command Universe.


raineorshine (approved):
@BayuAri Ready for testing. Test that the Command Palette, Help, and Gesture Cheatsheet panels preserve presentation, filtering, and search functionality on desktop and mobile. Thanks!

Command Palette - Cmd + P (desktop only)
Help - Click the "?" icon in the footer (mobile and desktop)
Gesture Cheatsheet - "?" gesture (mobile only)

BayuAri (commented):
Everything looks good and working as expected matching with the `main`

PR #4337, opened 2026-06-01
iOS: Preserve cursor-only state after cursorBack tap

marcus-pousette:
Strengthens the existing iOS `Swipe over cursor` test without adding runtime changes.

The test now covers the real BrowserStack path observed in diagnostics:
- close the native keyboard
- perform cursorBack via swipe-right
- tap the thought through the existing iOS tap helper
- wait for browser/focus settling
- assert the thought is selected as the app cursor, but native keyboard/edit mode remains closed

Diagnostics from a separate throwaway PR showed the tap path is a trusted iOS `touchend` followed by a trusted focus event while `isKeyboardOpen === false`. This PR captures that intended contract in the existing test.

### DIFF
--- a/src/e2e/iOS/__tests__/caret.ts
+++ b/src/e2e/iOS/__tests__/caret.ts
@@ -2,6 +2,7 @@
  * IOS Safari caret positioning tests.
  * Uses WDIO test runner with Mocha framework.
  */
+import type { WindowEm } from '../../../initialize'
 import gestures from '../../../test-helpers/gestures'
 import clickThought from '../helpers/clickThought'
 import editThought from '../helpers/editThought'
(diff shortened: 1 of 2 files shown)


raineorshine:
After an arbitrary focus (which is odd without a tap event, but let's assume that this is somehow possible), I would expect the keyboard to open. This is triggered in the [Editable's onFocus handler](https://github.com/cybersemics/em/blob/49da39fe933591b04a193a5d01e7caa9ecfaba5d/src/components/Editable.tsx#L587). Do you expect something different?

(I think you mean cursorBack, not cursorUp. cursorUp is a different command that is desktop-only.)

marcus-pousette:
Ye, the test is not ideal atm. I am struggling a bit to reproduce the test correctly in how it was flaky when we swapped the backend out, but somehow it can be surfaced as a timing issue. Let me do another attempt here to make it more convincing. 

I will ping you when I got this to be more accurate

Issue #3877, opened 2026-03-02
Thought can accept Text Color and Highlighting after being splitted


BayuAri:
## Step to reproduce
1. Create a thought with text "Lorem Ipsum Dolor Sit Amet, Consectetur Adipiscing Elit, Sed Do Eiusmod Tempor Incididunt Ut Labore Et Dolore Magna Aliqua. Ut Enim Ad Minim Veniam, Quis Nostrud Exercitation Ullamco Laboris Nisi Ut Aliquip Ex Ea Commodo Consequat. Duis Aute Irure Dolor In Reprehenderit In Voluptate Velit Esse Cillum Dolore Eu Fugiat Nulla Pariatur. Excepteur Sint Occaecat Cupidatat Non Proident, Sunt In Culpa Qui Officia Deserunt Mollit Anim Id Est Laborum."
2. Tap Text Color button
3. Select any highlight color (the bottom row)
4. Tap Split Sentence button
5. Select any newly split thought
6. Tap Text Case button
7. Select Upper Case or Title Case button
8. Tap Text Button
9. Select any text color (the above row)

## Current behavior
Thought text is highlighted and text colored at the same time.
Note: select the same color for text and highlight, the text will not be visible because they are blending in

<img width="1083" height="1026" alt="Image" src="https://github.com/user-attachments/assets/cfa4ba48-b17e-4111-b2a7-afecebba5c87" />

**Video:**

https://github.com/user-attachments/assets/8a86bf2b-20c7-454e-ba8b-2421196997ce

## Expected behavior
Highlight Color should be reverted to black as default. 

Thought text only accepts either Text or Highlight Color, it should not be applied both at the same time.

ethan-james:
This looks like a duplicate of #3930.

ethan-james:
This had the same cause as #3930 and shouldn't require an additional test because the core repro steps are the same.

PR #4346, opened 2026-06-03
Case-insensitive background color regex

### LINKED ISSUE #3930: Text accept Font and Background Highlight color after changing the text onto Upper Case
## Step to Reproduce
1. Create Thought **Hello World**
2. Tap Text Color and select Purple Background Highlight
3. Tap Text Case and apply Upper Case
4. Tap Text Color and select Purple Font Color

## Current behavior
Text is invisible because Background and Font have the same color.
The color selection is just to show that Background and Font Color can be applied at the same time after the text is being Upper Case.

Image

**ideo:**

https://github.com/user-attachments/assets/c4ac8d3c-33a4-4af4-aef4-2f90f13f8237

## Expected behavior
Font should be purple and background highlight returns to its default black.
Text should only accept 1 either Font or Background color.

ethan-james:
Fixes https://github.com/cybersemics/em/issues/3930 and https://github.com/cybersemics/em/issues/3877

The uppercase text tool converts all of the HTML in the thought to uppercase, including formatting tags. The background-color regex was case-sensitive, so I changed that to be case-insensitive instead.

### DIFF
--- a/src/actions/formatSelection.ts
+++ b/src/actions/formatSelection.ts
@@ -16,6 +16,8 @@ import strip from '../util/strip'
 import { editThoughtActionCreator as editThought } from './editThought'
 import { setDescendantActionCreator as setDescendant } from './setDescendant'
 
+const BACKGROUND_COLOR_REGEX = /background-color\s*:\s*[^;]+;?/i
+
 /** Format the browser selection or cursor thought as bold, italic, strikethrough, underline. */
 export const formatSelectionActionCreator =
   (
@@ -43,7 +45,7 @@ export const formatSelectionActionCreator =
       selection.text()?.length === strip(thought.value).length
     ) {
       // Check the value of the note or thought for a custom background color (#3901)
-      const hasCustomBackgroundColor = /background-color\s*:\s*[^;]+;?/.test(
+      const hasCustomBackgroundColor = BACKGROUND_COLOR_REGEX.test(
         state.noteFocus ? (noteValue(state, state.cursor) ?? '') : thought.value,
       )
       const savedSelection = selection.save()
--- a/src/e2e/puppeteer/__tests__/format.ts
+++ b/src/e2e/puppeteer/__tests__/format.ts
@@ -1,5 +1,7 @@
 import click from '../helpers/click'
+import clickThought from '../helpers/clickThought'
 import exportThoughts from '../helpers/exportThoughts'
+import getEditingText from '../helpers/getEditingText'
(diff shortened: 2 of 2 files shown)


raineorshine (approved):

raineorshine:
@ethan-james What's the feasibility of us being able to write an automated test for this? I'm afraid we're falling behind on formatting behavior test coverage.

ethan-james:
> @ethan-james What's the feasibility of us being able to write an automated test for this? I'm afraid we're falling behind on formatting behavior test coverage.

Shouldn't be too hard. Do you want a unit test and a Puppeteer test, or just a unit test?

raineorshine:
Which do you think would be better for this case? 

I'm happy to provide confirmation, but I always like for you to start with an initial judgment when it is within your domain. Review the [Test Levels](https://github.com/cybersemics/em/wiki/Testing#test-levels) to help guide your decision. We've done lots of testing at different levels, so you should have a pretty good understanding of the advantages/disadvantages of each level. Present your suggestion along with your reasons if there is any ambiguity.

You can even feed AI the test levels and PR and ask it what level it thinks is appropriate if that helps kickstart the thinking process.

ethan-james:
> Which do you think would be better for this case?
> 
> I'm happy to provide confirmation, but I always like for you to start with an initial judgment when it is within your domain. Review the [Test Levels](https://github.com/cybersemics/em/wiki/Testing#test-levels) to help guide your decision. We've done lots of testing at different levels, so you should have a pretty good understanding of the advantages/disadvantages of each level. Present your suggestion along with your reasons if there is any ambiguity.
> 
> You can even feed AI the test levels and PR and ask it what level it thinks is appropriate if that helps kickstart the thinking process.

Well, I would prefer to avoid a Puppeteer test for functional testing, but I wasn't sure that I would be able to faithfully test the logic in a way that would catch regressions in the actions/commands because everything eventually becomes an `editThought` action which will stay the same in a test even if the logic that creates it changes.

Then I remembered that this requires JSDOM so it is a moot point and will require a Puppeteer test.

BayuAri (commented):
Confirm fixed.
Test cases:

- [x] [Text accept Font and Background Highlight color after changing the text onto Upper Case](https://github.com/cybersemics/em/issues/3930#issue-4069227079)
- [x] [Thought can accept Text Color and Highlighting after being splitted](https://github.com/cybersemics/em/issues/3877#issue-4011438564)

Issue #1523, opened 2022-02-26
BrowserStack: Session terminated


raineorshine:
BrowserStack tests intermittently fail on the iOS tests. It may be related to #1475. It is not the same test that fails every time, and sometimes they all succeed. 

In the below example, the `Tap hidden root thought` test fails.

https://github.com/cybersemics/em/runs/5344813600?check_suite_focus=true

First the session name is set to `Tap hidden root thought`:

```
2022-02-26T14:26:03.823Z INFO webdriver: COMMAND executeScript("browserstack_executor: {"action": "setSessionName", "arguments": {"name": "Tap hidden root thought"}}'
}
```

Then for some reason there is a call to `deleteSession` (part of [teardown](https://github.com/cybersemics/em/blob/61868799752a2b7b706a94952ec27e589c0ffcf2/src/e2e/webdriverio-environment.js#L64)). Could this be triggered from a test running in parallel?

```
2022-02-26T14:25:51.134Z INFO webdriver: COMMAND deleteSession()
2022-02-26T14:25:51.134Z INFO webdriver: [DELETE] hub.browserstack.com/wd/hub/session/b5fda78bb91aa37aa4454818201fc5131fcb9519
```

Then it searches for `.new-thought-instructions`, which is part of [initSession](https://github.com/cybersemics/em/blob/61868799752a2b7b706a94952ec27e589c0ffcf2/src/e2e/iOS/helpers/initSession.ts#L22).

```
2022-02-26T14:25:51.152Z INFO webdriver: COMMAND findElement("css selector", ".new-thought-instructions")
2022-02-26T14:25:51.152Z INFO webdriver: [POST] hub.browserstack.com/wd/hub/session/b5fda78bb91aa37aa4454818201fc5131fcb9519/element
2022-02-26T14:25:51.152Z INFO webdriver: DATA { using: 'css selector', value: '.new-thought-instructions' }
```

When the test proceeds, webdriver gives a `Request failed with status 404 due to Session not started or terminated` and goes through all 10 retries before giving up.

```
[...]

Issue #1475, opened 2022-01-14
Parallel BrowserStack tests


raineorshine:
An issue with parallel Just tests in BrowserStack was noted in #1470. The current solution does not scale beyond 3 test files. A better solution would be to limit the concurrency of the Jest tests. I suggest `--maxWorkers` at the end, but it has yet to be tested.

> @raineorshine I was observing that the browser stack tests failed intermittently during github checks. I also tested it from my local machine and it sometimes failed intermittently. I looked into it and came to a convincing reason for it.
> 
> So this is how browser stack tests are running.
> 
> * Jest environment is started.
> * It creates a browser stack local testing connection with a local identifier. A local identifier is used to identify the connection we make to the browser stack.
> * webdriverio is used to connect to the browser stack to run the tests.
> 
> However, the problem was that the jest ran three tests in parallel. But we for each test we created a browser stack local testing connection with the same local identifier. That meant if one of the tests completes and stops the connection with that local identifier, that connection would probably close for others too. So that caused some tests to fail due to connection errors.
> 
> So now the connection is started with the unique local identifier for every test that is running in parallel. I tested it by running the tests multiple times. And it seems to be working without failure.
> 
> However, we need to make sure we don't max out the parallel threads provided by BrowserStack. For now, we have only three tests, so it will work without problem.
> 
> Please let me know what you think about it.

> > How can we limit the number of parallel threads to 3 if we add additional iOS test files?
> 
> I am not sure about this. For now, the number of threads is equal to the number of ios test files. The limit for the basic plan is 5 parallel threads. To overcome this we can either run tests in a sequence that will increase the total test time. Or we need to find a way to instantiate a connection once before e2e tests run, share it among all the test files and close it after all the tests are executed. I have little experience on BrowserStack so there could possibly be other ways too.

> > Cool. Thanks for sharing your ideas.
> > Ideally we could limit concurrent tests, maybe at the web driver level.
> 
> I believe that should be done on jest level. Since it runs tests in parallel. If we could configure jest to use a maximum of 5 test files in parallel, it could be the best solution to this problem. I am not sure if we have that option in jest. However, we can always resort to running tests in sequence using `--runInBand` flag in jest.

> > Could you try `--maxWorkers`?
> > [jestjs.io/docs/cli#--maxworkersnumstring](https://jestjs.io/docs/cli#--maxworkersnumstring)
> 
> I'll create a separate issue for this.

PR #4352, opened 2026-06-03
Agent improvement roundup: use pre-existing test helpers with `browser-control`; `run-test` skill; TDD workflow; `tdd-write-failing-test` skill

fbmcipher:
This PR includes the following AI improvements from the last week:

### Agents can now use pre-existing test helpers
Agentic browser sessions on both Chrome and iOS (via `browser-control`) now reuse the existing ecosystem of test helpers in the **em** repository, rather than defining test helpers in `.md` skill files like we did before. This is more economical, eliminates a source of duplication, and makes test reproduction more reliable and deterministic.

Agents still have access to the full capabilities of the Chrome/WebDriver MCP servers, so can still execute arbitrary JavaScript, read the DOM, etc *alongside* using the existing suite of test helpers.
#### Details

- **Added clear directives to `browser-control`**
	- The full set of available helpers for the specified platform is listed using an `ls` call in the agent's context window, so it has full awareness of all the helpers it has at its disposal.
	- It is also told *how to execute these helpers* (see below.)
	- If it needs to perform an interaction that isn't covered by a helper, it's given permission to use the associated MCP server to achieve its task.
	- But, the agent is told to **always prefer existing helpers**. This should increase the consistency of issue reproduction.

- **`attachExisting` functions in `src/e2e/iOS` and `src/e2e/puppeteer`**
	- With E2E tests, the default assumption is that a new browser instance will be initialized to execute that test suite.
	- A new `attachExisting` function for each test suite enables us to **bring our own browser instance** rather than spinning up a new one.
	- This is vital so that we can attach the existing WebDriver/Chrome instances we spin up for `browser-control`.
	- This lets us have our cake and eat it too - we can use all of the existing test helpers *and* still enjoy the benefits of interactive debugging with the MCP servers within a single shared browser instance.

- **Ad-hoc TypeScript execution**
	- To actually execute sequences of test helpers, agents are instructed to write a file in `/tmp`, import the `attach` function as well as any needed helpers, and write their code in an IIFE.
	- Then, agents execute sequences using `npx tsx`.
	- This seems to work quite well. Agents happily leverage this new ability to execute **sequences** (rather than just executing tool calls one at a time, as they did before), which seems to have substantially improved the time it takes them to reproduce issues.
	- Notably, the usage of `npx tsx` ensure we keep type safety in this execution mode.

- **Trivial refactor to the Puppeteer test suite to make it work**
	- Before this PR, Puppeteer session state was stored at module level scope in `setup.ts`, which is also where Puppeteer E2E tests are initialized (vitest entry point). 
	- Now, we separate session state from `setup.ts` (in new file `session.ts`), and export a setter function so that we can attach our already-running Chrome browser with Puppeteer.
	- This required an import change in all Puppeteer tests (import `page` from `session` rather than `setup`, which is arguably more correct anyway.)
	- This change isn't required for WebDriver-based tests, where session state is kept at global scope (`window.browser`)

- **A couple new test helpers, like** `showEditMenu` & `resetApp`
	- Adding new test helpers now increases both testing *and* agent performance - compounding improvements.

### New `run-test` skill

Until now, our strategy has been to **not execute tests within the agent's environment**, and instead rely on the pre-existing CI testing mechanism and observe pass/fail state there. This was a neat optimization trick and arguably helped us separate concerns, especially at that early stage of agent optimization.

However, the new TDD strategy *requires* us to execute a *single* test - sometimes *repeatedly* until we get the expected result (either *green* or *red*). We can't get that level of granularity by relying on CI, so we need a way to run tests inside the agent's environment.

This new `run-test` skill just explains to the agent *how* it can run a specific test on either Puppeteer or WebDriver test suite. It's basically just an instruction file that explains to the agent how to use the pre-existing `yarn test:` scripts.

*Unlike* the other improvements we've discussed here, we *do not reuse the same browser instance* because tests should always be run in a clean environment.

**Important:** `run-test` **ignores** `.skip`. That means we can run `.skip`ped tests manually while still making sure CI shows green.

### TDD workflow: red-green-iterate

**When assigned to an issue with Steps to Reproduce, the following things will happen:**

1. **Agent determines which platform to test with.** Generally, Chrome with mobile emulation is preferred, only using the BrowserStack iOS route if the issue is specifically related to iOS or if reproduction fails using Chrome.
2. **Initial reproduce stage.** The agent follows the STR to try and reproduce the issue. The agent is encouraged to use pre-existing test helpers, but can also use all available MCP tools.
3. **Write failing test.** Once reproduced, the agent writes a failing test using a new `tdd-write-failing-test` skill. The agent is instructed to follow existing test conventions. Crucially, the test is `.skip`ped so that it's CI-safe.
4. **Run failing test using `run-test`.** **Only** the failing test is executed, to verify that it does in fact *fail.* If it *does not fail*, iterate and loop until the test does fail. `run-test` **never** respects `.skip`.
5. **Implement fix.**
6. **When the agent thinks it has fixed the issue, run the test again using `run-test`.** If passing - great! - the issue is *likely* fixed. If not, continue to iterate until it passes.
7. **Remove `.skip` before submitting the PR.**

### Note

- None of these changes apply for issues that *do not include Steps to Reproduce.* In those cases, the old CI monitoring technique developed for the Worker Bee back in April is the dominant workflow.

### Test results

Just one run for now, but shows enough promise:

- https://github.com/fbmcipher/em/pull/80#event-26269456950
	- As expected, it reproduced the issue, wrote a *failing test that encoded the expected outcome*, then iterated a fix until passing.
	- The *solution* it devised to the problem I might take exception to – forcing a remount smells like a quick hack rather than a thoughtful fix – but that's a separate issue. The test-driven iteration workflow itself worked exactly as we discussed.

Of course, the next step would be deeper testing across a wider variety of issues, but this run acts as at least basic validation of the concept.


scripts/shared-chrome.mjs
```diff
@@ -0,0 +1,39 @@
...
+  // "No usable sandbox". Disable the sandbox on Linux only: the runner is an ephemeral, single-tenant
+  // VM loading only our own dev server. This is independent of headless/Xvfb (the crash happens headed
+  // too). `--disable-dev-shm-usage` avoids separate crashes from the small /dev/shm in CI containers.
+  // macOS keeps its sandbox. Override with EM_CHROME_NO_SANDBOX=0.
+  ...(process.platform === 'linux' && process.env.EM_CHROME_NO_SANDBOX !== '0'
+    ? ['--no-sandbox', '--disable-setuid-sandbox', '--disable-dev-shm-usage']
+    : []),
+]
+
+// eslint-disable-next-line no-console
```
raineorshine:
```suggestion
```

scripts/shared-chrome.mjs
```diff
@@ -0,0 +1,39 @@
...
+console.log(`Launching shared Chrome on :${port}\n  ${executablePath} ${args.join(' ')}`)
```
raineorshine: Use `console.info` for persistent logging. `console.log` is only for temporary debugging on this project (hence the lint rule).

```suggestion
console.info(`Launching shared Chrome on :${port}\n  ${executablePath} ${args.join(' ')}`)
```

raineorshine (changes requested):

scripts/shared-chrome.mjs
fbmcipher: Thank you – fixed in 2c73842.

As you might have already guessed, this code was written by an agent (Opus 4.8). I'm thinking it might be a good idea to disallow linter/TypeScript opt-outs like this in custom instructions, unless "absolutely necessary" (i.e. there is literally no other solution to the problem), as this isn't the first time I've seen an agent do this in this project.

raineorshine (approved):

raineorshine (changes requested):
I ran the BrowserStack CI and it failed 2/3 times for different reason. Could you look into this and figure out what is causing the flakiness?

✗ Run 1: "Home click home link to set the cursor to null" test timeout
✓ Run 2: Success
✗ Run 3: ✖ Failed to create a session: [iPhone 15 Plus iOS 17 #0-2] WebDriverError: The operation was aborted due to timeout when running "https://hub-use.browserstack.com/wd/hub/session" with method "POST"

I checked `main` and the last six BrowserStack tests there passed, so I think the flakiness is here.

fbmcipher:
@raineorshine OK, I think this is good to go.

I slipped in the fixes I mentioned at our meeting yesterday:
- Fix for https://github.com/cybersemics/em/pull/4386#issuecomment-4676334476: The agent will still work, even if `issue-repro` doesn't apply.
- Fix for https://github.com/cybersemics/em/pull/4381 and https://github.com/cybersemics/em/pull/4376: The agent can now open PRs again, by pointing the agent at the new name used for its PR creation tool.

Regarding the test flakiness: there were two potential causes for this, and I think that both should now be fixed:

1. The new `resetApp` helper was a little flaky. In order to DRY, I replaced the `beforeTest` logic with the new shared `resetApp` helper. This was causing flaky test initialization.
2. Now that we use BrowserStack in both CI and agent runs, there's a higher chance we reach our session concurrency limit. This concurrency limit was documented in some older issues (#1475, #1523). If the concurrency limit is reached, tests fail. I've tried to fix this by making sessions more resilient (wait/retry until there are available BrowserStack session slots again, rather than just failing), and reduce the number of concurrent tests from 3 to 2 (more resilient at the expense of slightly slower iOS e2e test runs)

To verify, I ran 10 sequential iOS CI runs, and all of them passed!

> @fbmcipher is attempting to deploy a commit to the **Cybersemics Institute** Team on [Vercel](https://vercel.com).
> 
> A member of the Team first needs to [authorize it](https://vercel.com/git/authorize?team=Cybersemics%20Institute&slug=cybersemics&teamId=team_mICFyIiQ5jMdpqQfup3uZBt0&type=github&job=%7B%22headInfo%22%3A%7B%22sha%22%3A%228c1db8e929bd3798299e0f69a864b5f806fe565e%22%7D%2C%22id%22%3A%22QmYq9QwutkNwUUMAcggbjPG6WwRyXFMRJfArMVG8ocLT1u%22%2C%22org%22%3A%22cybersemics%22%2C%22prId%22%3A4352%2C%22repo%22%3A%22em%22%7D).

Just this Vercel CI that's failing.

src/e2e/iOS/helpers/resetApp.ts
```diff
@@ -0,0 +1,38 @@
...
+const RESET_SENTINEL = '__emResetSentinel'
+
+/** Reset to a clean, empty thoughtspace: clear storage, refresh, dismiss the tutorial. */
+const resetApp = async (): Promise<void> => {
+  await browser.execute(sentinel => {
+    localStorage.clear()
+    sessionStorage.clear()
+    // browser.refresh() can resolve before WKWebView navigates; tag the current document so the poll
+    // below can tell it apart from the reloaded one.
+    ;(window as unknown as Record<string, boolean>)[sentinel] = true
```
raineorshine, line 11: Just to confirm, you're saying that sometimes the sentinel is still set after `await browser.refresh`? I would expect everything in-memory to be cleared on refresh.

src/e2e/iOS/helpers/resetApp.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  // Use execute() not the WebDriver element protocol — it can keep missing elements after WKWebView reloads.
+  // Each iteration: skip if the sentinel is still set (stale page), click #skip-tutorial if present,
+  // then resolve once it's gone and empty-thoughtspace is visible.
+  // Can't treat empty-thoughtspace alone as ready — it also renders during the tutorial.
+  await browser.waitUntil(
+    async () =>
+      browser.execute(sentinel => {
+        if ((window as unknown as Record<string, boolean>)[sentinel]) return false
+        const skip = document.getElementById('skip-tutorial') as HTMLElement | null
```
raineorshine, line 23: Is there a reason the sentinel wait is combined with the skip tutorial logic? That seems confusing to me. In my mind it goes:

1. Wait for the sentinel to be cleared (i.e. wait for the full refresh)
2. Wait for the Skip button and then skip the tutorial.
3. Wait for the empty thoughtspace.

Notice how the old code is flatter and easier to reason about:

```ts
// Wait for the tutorial skip button and click it
const skipElement = await $('#skip-tutorial')
await skipElement.waitForExist({ timeout: 90000 })
await skipElement.waitForClickable({ timeout: 10000 })
await skipElement.click()

// Wait for the empty thoughtspace to be ready
const emptyThoughtspace = await $('[aria-label="empty-thoughtspace"]')
await emptyThoughtspace.waitForExist({ timeout: 90000 })
```

Now we have to reason about three different ways that `waitUntil` can return `false` in order to understand the sequence of steps. Maybe I'm missing something though.

raineorshine (approved):
> To verify, I ran 10 sequential iOS CI runs, and all of them passed!
> 
> >

Beautiful, that gives me confidence that this is solid now.

---

I have some followup questions about `resetApp`, but I won't hold up the merge.

PR #4369, opened 2026-06-05
Broken test: Puppeteer Caret (Command Center)

### LINKED ISSUE #4309: Broken test: puppeteer/__tests__/caret.ts - tapping a thought after opening and closing
Getting a lot of failures from this test all of a sudden. Likely a recent regression.

- https://github.com/cybersemics/em/actions/runs/26471702323/attempts/1
- https://github.com/cybersemics/em/actions/runs/26472360999/attempts/1
- https://github.com/cybersemics/em/actions/runs/26472360999/job/78136326762
- https://github.com/cybersemics/em/actions/runs/26530006217/job/78143834775
- https://github.com/cybersemics/em/actions/runs/26481631950/job/77984344714?pr=4311

`[[puppeteer-e2e] src/e2e/puppeteer/__tests__/caret.ts > mobile only > tapping a thought after opening and closing Command Center via Done should not open the keyboard: src/e2e/puppeteer/__tests__/caret.ts#L352](https://github.com/cybersemics/em/commit/70e9ee3f55a5d5c0c9dcd3c6674bfd43918b1dff#annotation_57063582633)
Error: Test timed out in 20000ms.`

karunkop:
Close #4309 

### Likely cause of the issue:

From my understanding, the test could timeout because of two reasons:

- `waitForSelector(..., { hidden: true })` may wait indefinitely if the command centre panel remains in the DOM but isn’t considered “hidden” by Puppeteer
- The test assumes `document.body` becomes the active element after closing the keyboard. If focus never changes as expected, the condition is never met, causing the test to wait until it times out.

### Solution:

Added a proper waitUntil condition that accurately checks the redux state for `showCommandCenter` and `isKeyboardOpen`


raineorshine (changes requested):
Calling `getState` in an integration test is not allowed. The point of an integration test is that you have to test from the outside. You only have access to what the user can see and interact with.

Accessing the Redux state directly is only allowed in rare circumstances where there is a known limitation, such as `paste` not working in headless mode.

karunkop:
> Calling `getState` in an integration test is not allowed. The point of an integration test is that you have to test from the outside. You only have access to what the user can see and interact with.
> 
> Accessing the Redux state directly is only allowed in rare circumstances where there is a known limitation, such as `paste` not working in headless mode.

Could you review the changes again? I have now added a wait condition to check for multicursor state via bullet atrribute highlight which should be gone when the command centre panel is closed

raineorshine:
Thanks, will do.

Please use the "Request Review" GitHub feature to request a review after updating a PR.

src/e2e/puppeteer/__tests__/caret.ts
```diff
@@ -360,7 +360,12 @@ describe('mobile only', () => {
 
     // Step 3: close the Command Center via the Done button
     await click('[data-testid="command-center-done"]')
-    await waitForSelector('[data-testid=command-center-panel]', { hidden: true })
+    // The sheet stays in the DOM when closed (transformed off-screen), so { hidden: true }
+    // on the panel never resolves. Wait for the visible multiselect highlight to clear instead.
+    await waitUntil(() => !document.querySelector('[aria-label="bullet"][data-highlighted="true"]'), {
```
raineorshine: Prefer `waitForSelector('[aria-label="bullet"][data-highlighted="true"]', { hidden: true })` instead of a custom `waitUntil`.

raineorshine (changes requested):

raineorshine (approved):

Issue #4179, opened 2026-04-25
Remove `allowDefaultSelection`


raineorshine:
- [x] Blocked by #3410
- [ ] Blocked by #4247 

Let's see if `allowDefaultSelection` is still needed in `useEditMode`.

 _Originally posted by @ethan-james in [#3410](https://github.com/cybersemics/em/pull/3410/changes#r3120661696)_

raineorshine:
Reviewing #3410, I can confirm that `allowDefaultSelection` is no longer needed and can be inlined.

ethan-james:
> Reviewing [#3410](https://github.com/cybersemics/em/pull/3410), I can confirm that `allowDefaultSelection` is no longer needed and can be inlined.

I believe that it can be eliminated entirely, but obviously that needs to be tested thoroughly.

raineorshine:
Even better

ethan-james:
I added a new instance of `allowDefaultSelection` so this is now tied to #4247.

ethan-james:
> I added a new instance of `allowDefaultSelection` so this is now tied to [#4247](https://github.com/cybersemics/em/issues/4247).

Now, as of #4371, `allowDefaultSelection` has moved to being a method for stopping [shouldSetSelection](https://github.com/ethan-james/em/blob/f1a66974ba6e1736b61b794d417d94eca64e7ccf/src/components/Editable/useEditMode.ts#L72) from setting the selection in cases where [setCaretOffset](https://github.com/ethan-james/em/blob/f1a66974ba6e1736b61b794d417d94eca64e7ccf/src/components/Editable/useEditMode.ts#L142) runs first.

In this role, it could easily be inlined, but not yet removed. I created #4470 to explore the possibility of removing the need for `allowDefaultSelection` entirely.

Issue #3374, opened 2025-11-15
Caret moves to beginning/end on tap top/bottom edge


raineorshine:
Tested on mobile Safari.

This was previously solved in 1550e4f5cc9d5a52fe3dacf97b732170a8e11573. There doesn't appear to be an original issue or PR.

## Steps to Reproduce

```
- one two three
- four five six
```
1. Set the caret on `one two three`.
2. Tap the exact top edge of `five`.

OR

1. Set the caret on `four five six`.
2. Tap the exact bottom edge of `two`.

## Current Behavior

The caret moves to the end of `one two three`.

## Expected Behavior

The caret should move to the middle of the thought that is tapped, i.e. near `five` or `two`. 

The caret should not move to the beginning or end of the thought, and there should be no dead zone between the thoughts.

## Testing

~~Should be covered by #3153, but test may need to be updated.~~ Requires WebdriverIO tests for iOS testing. 

## Notes

This is a regression, and was originally solved in 1550e4f5cc9d5a52fe3dacf97b732170a8e11573 with `clip-path`. Since tapping the top or bottom edge of an editable sets the caret to the beginning or end, we 1) use `clip-path` to clip the problematic edges, and 2) subtract the next thought's y position by the clip amount  to avoid any gap between the thoughts.

raineorshine:
Test cases for final manual testing:

- [x] Original issue
- [x] Nested HTML (bold, italic, underline, etc)
- [x] IndexSizeError ([Issue 1A](https://github.com/cybersemics/em/pull/3410#issuecomment-3665808622))
- [ ] Hard to move between words that has 2-3 characters ([Issue 1B](https://github.com/cybersemics/em/pull/3410#issuecomment-3665808622))
- [x] [Dead zone between thoughts where nothing happens](https://github.com/cybersemics/em/pull/3410#pullrequestreview-3610748955)
- [x] [Incorrect caret placement when the user taps the end of the line](https://github.com/cybersemics/em/pull/3410#issuecomment-3704343279)
- [x] [Wide blank space between thoughts](https://github.com/cybersemics/em/pull/3410#issuecomment-3723286106)
- [x] Hard to place the caret in the middle of a word ([Issue 2A](https://github.com/cybersemics/em/pull/3410#issuecomment-3784022515))
- [x] Scrolling only works on the blank area ([Issue 2B](https://github.com/cybersemics/em/pull/3410#issuecomment-3784022515))
[...]

PR #4371, opened 2026-06-05
Move setCaretOffset into onMouseDown and preventAutoscrollEnd back into onFocus

### LINKED ISSUE #4298: [iOS] Caret stays on old sibling with children
## Steps to Reproduce

```
- Hello
  - A
  - B
  - C
- World
  - E
  - F
  - G
```

1. Set the caret at the end of `Hello`. 
2. Tap the bottom-right corner of the "d" in `World`. 
3. Tap the end of `Hello`.  

## Current Behavior

On Step 2, the caret moves to the beginning of `World`. 
On Step 3, the caret moves to the end of `World`. 

## Expected Behavior

On Step 2, the caret should move to the end of `World`. 
On Step 3, the caret should move to the end of `Hello`.

ethan-james:
Fixes #4298 

## Overview

Triggering `preventAutoscrollEnd` in `onMouseUp` was problematic because `mouseup` isn't always dispatched, and when it is dispatched, it doesn't necessarily mean that the editable was focused. Moving `preventAutoscrollEnd` back into `onFocus` means that `setCaretOffset` needs to fire earlier, such as in `onMouseDown` where it spent most of its time during development of #3410.

## Complications (solved)

- [x]  Native iOS drag-and-drop text selection breaks when `setCaretOffset` is called `onMouseDown`.

**Solution:** In order to preserve native behavior, it is necessary to keep track of whether a press has ended (`touchend` has been dispatched) or whether it is ongoing in order to short-circuit `onMouseDown`.

- [x] Using `setCaretOffset` updates the Redux state in a way that triggers the `shouldSetSelection` effect, which creates duplicate invocations of `preventAutoscroll` and `selection.set()`.

**Solution:** Call `allowDefaultSelection` at the same time as `setCaretOffset` to prevent `shouldSetSelection` from being true.

---

## Follow-Up (speculative)

If we want, we can pursue the possibility of moving all `selection.set` behavior into the `shouldSetSelection` effect rather than maintaining separate `selection.set` behavior in `setCaretOffset`. The flow would look like:

1. `setCaretOffset` updates `caretOffset` in the Redux store
2. the `shouldSetSelection` effect evaluates its conditions and determines whether to act
3. `preventAutoscroll` and `selection.set` are called in one declarative place

### Confidence Level

low-to-medium pending further testing

I verified that deferring `selection.set` until the effect runs seemed to work well enough when tapping between thoughts, although triggering the `selection.set` in `setCaretOffset` is still necessary when tapping within the active thought because the `shouldSetSelection` effect isn't currently triggered when only the `state.cursorOffset` value updates.

I didn't test any of the commands that rely on `shouldSetSelection`, and I am somewhat confident that the flicker/lag can be avoided by either using `useEffect` as it is currently structured, or by changing to `useLayoutEffect`, which introduces additional uncertainty regarding all of those commands that need to be tested.

### Pros

1. consolidating all `selection.set` behavior would make `useEditMode` easier to reason about
2. a single declarative solution for selection handling within editables would be more robust
3. do away with `allowDefaultSelection` entirely (#4179)

### Cons

1. the [shouldSetSelection](https://github.com/ethan-james/em/blob/a0a03fd732b904c49566f88075c4ff880aff900a/src/components/Editable/useEditMode.ts#L72) effect is widely used, and many commands would need to be tested if it were changed
2. calling `selection.set` in `setCaretOffset` guarantees that the caret position with minimal lag/flicker, which may be more difficult to guarantee if it is offloaded onto the React render cycle

## Test Coverage (unsuccessful)

I tried to add a Browserstack iOS test, but I was never able to get it to fail on main (https://github.com/cybersemics/em/pull/4403)


raineorshine (approved):
Thank you! The code looks fine.

> ## Follow-Up (speculative)
> If we want, we can pursue the possibility of moving all `selection.set` behavior into the `shouldSetSelection` effect rather than maintaining separate `selection.set` behavior in `setCaretOffset`. The flow would look like:
> 
> 1. `setCaretOffset` updates `caretOffset` in the Redux store
> 2. the `shouldSetSelection` effect evaluates its conditions and determines whether to act
> 3. `preventAutoscroll` and `selection.set` are called in one declarative place

That makes sense to me, and more closely matches the earlier declarative design. We can try this in a separate PR unless you think otherwise.

> ## Test Coverage (unsuccessful)
> I tried to add a Browserstack iOS test, but I was never able to get it to fail on main (#4403)

If manual testing doesn't turn up anything, I'm okay merging this PR and then working on an automated test afterwards. Since browser selection behavior is critical, fragile, and high risk, we will want to add test coverage.

---

@BayuAri This is a high risk change that requires extensive testing of the browser selection across all platforms. Any of the test cases from https://github.com/cybersemics/em/issues/3374#issuecomment-4411254679 are fair game, but use your best judgment as you decide what to test. I'll let you manually test first before me. Thank you!

ethan-james:
> Thank you! The code looks fine.

🎉

> That makes sense to me, and more closely matches the earlier declarative design. We can try this in a separate PR unless you think otherwise.

Yes, I think exploring it in a separate PR makes sense.

> If manual testing doesn't turn up anything, I'm okay merging this PR and then working on an automated test afterwards. Since browser selection behavior is critical, fragile, and high risk, we will want to add test coverage.

Sounds good. It would be good to explore whether `mouseup` is dispatched on the correct element or not to see if the `tap` helper is behaving the same as a real-world tap.

BayuAri:
Original issue is fixed.
Caret is placed according on the correct thought

## Issue A: Toolbar is jumping down and up when user moves the caret between Thoughts

```
- Hello
   - A
   - B
   - C
- World
   - F
   - G
   - H
- Salsa
- Tango
```

## Step to Reproduce
1. Place caret at the end of "Hello"
2. Move the caret to the end of "World"
3. Move the caret to the end of "Salsa"
4.  Move the caret to the end of "Tango"


## Current behavior
The toolbar is jumping down and up. 
Focus is always placed under the toolbar, so the thought is auto scrolled up. 

As reference, the behavior on `main` is a bit difference, the toolbar is jittery and it would jump but not as always as the built on this PR. I am not sure if this is a regression or not, but the jumping is too frequent on the PR

image

**PR/4371**

https://github.com/user-attachments/assets/dd9d3296-88e2-4eff-8563-74537af0dfeb

**main**

https://github.com/user-attachments/assets/1bc94fd1-9552-49c1-9d33-149fc9a7e671



## Expected behavior
Toolbar should not jump when user moves caret between thoughts



**Side note**: a blank area under the Thought World on the `main` video is logged as #4408
Thought this is not reproducible on this PR, I will revisit the ticket when this PR is merged onto `main` for a double check.

image

BayuAri (commented):
I found no more issue on this PR aside from [Issue A: Toolbar is jumping down and up when user moves the caret between Thoughts](https://github.com/cybersemics/em/pull/4371#issuecomment-4740860710)

I logged 2 related issue uncovered with this PR tested as follow
#4408
#4426

raineorshine:
> I found no more issue on this PR aside from [Issue A: Toolbar is jumping down and up when user moves the caret between Thoughts](https://github.com/cybersemics/em/pull/4371#issuecomment-4740860710)

@ethan-james What's your assessment of this issue? My impression from Bayu is that it was preexisting but exacerbated in this PR.

ethan-james:
> > I found no more issue on this PR aside from [Issue A: Toolbar is jumping down and up when user moves the caret between Thoughts](https://github.com/cybersemics/em/pull/4371#issuecomment-4740860710)
> 
> @ethan-james What's your assessment of this issue? My impression from Bayu is that it was preexisting but exacerbated in this PR.

I hadn't really noticed it on the iPhone 12 Pro, although it's not too hard to reproduce. It is much more dramatic on iPhone 16 Pro Max. I zeroed in on a test case where I toggle between `F` and `G`, and you can see that it looks pretty similar between the two branches.

Main:

https://github.com/user-attachments/assets/6f618d44-c7b7-42ee-b73a-991e0e1c8425

206dcb91d6812c9b718b1b2265a702d80b865bdc:

https://github.com/user-attachments/assets/92d93435-ae62-46b1-b466-f3ef1f4202c5

I broadened my tests to see if there were more cases of the toolbar jumping in the branch than there were on main, but my assessment is that they're pretty similar, at least on the iPhone 16 Pro Max simulator and the iPhone 12 Pro device.

raineorshine:
Thanks. Looks like a preventAutoscroll issue. I'm satisfied merging this and then addressing the scroll issue separately. It might even be solved by #4435.

ethan-james:
> Thanks. Looks like a preventAutoscroll issue. I'm satisfied merging this and then addressing the scroll issue separately. It might even be solved by #4435.

Yes, the test case looks perfect on the `issue-3765` branch.

PR #4373, opened 2026-06-08
Fix stale selection offset during chained gestures

marcus-pousette:
Prevents indent/outdent from reusing a browser selection offset from a different editable.

Adds coverage for the chained gesture case that was logging `IndexSizeError`.

The test is perhaps not perfect here, but IndexSizeError can right now that can be reproduced in the current test coverage, which I think should be better handled.

This was found by workin on the TreeCRDT integration PR


raineorshine (approved):
Looks good, thanks.

Note that this probably masks a bug in chained gesture that should reset the selection offset between commands.

PR #4384, opened 2026-06-10
Initialize app events before thoughtspace startup

marcus-pousette:
Registers app input handlers **before** async thoughtspace startup so keyboard commands are active while storage initialization is still settling. This seems to fix flakes that appear in e2e testing with slower startup paths.

Adds a Puppeteer regression that blocks thoughtspace initialization with a test-controlled Promise, presses Enter before initialization is released, and verifies a new thought is created from the empty thoughtspace.

### DIFF
--- a/src/e2e/puppeteer/__tests__/startup.ts
+++ b/src/e2e/puppeteer/__tests__/startup.ts
@@ -1,11 +1,5 @@
-import type { WindowEm } from '../../../initialize'
+import getEditingText from '../helpers/getEditingText'
 import { page } from '../setup'
-
-type WindowEmWithStartupFlags = WindowEm & {
-  testFlags: WindowEm['testFlags'] & {
-    thoughtspaceInitBlocker?: Promise<void> | null
-  }
-}
 
 type StartupWindow = Window & {
   __EM_TEST_FLAGS__?: {
@@ -28,32 +22,15 @@ it('handles keyboard commands while thoughtspace initialization is delayed', asy
 
   await page.reload({ waitUntil: 'domcontentloaded' })
 
-  await page.waitForFunction(() => Boolean((window.em as WindowEmWithStartupFlags).testFlags.thoughtspaceInitBlocker))
   await page.waitForFunction(() => !document.querySelector('[aria-label=modal]'))
   await page.waitForSelector('[aria-label=empty-thoughtspace]')
+  expect(await getEditingText()).toBeUndefined()
 
   try {
     await page.keyboard.press('Enter')
 
-    await page.waitForFunction(() => {
-      const em = window.em as WindowEm
-      return em.testHelpers.getState().lastUndoableActionType === 'newThought'
-    })
-
-    const result = await page.evaluate(() => {
-      const em = window.em as WindowEm
-      const editable = document.querySelector('[data-editing=true] [data-editable]')
-
-      return {
-        editingText: editable?.innerHTML,
-        lastUndoableActionType: em.testHelpers.getState().lastUndoableActionType,
-      }
-    })
-
-    expect(result).toEqual({
-      editingText: '',
-      lastUndoableActionType: 'newThought',
-    })
+    await page.waitForSelector('[data-editing=true] [data-editable]')
+    expect(await getEditingText()).toBe('')
   } finally {
     await page.evaluate(() => {
       const startupWindow = window as StartupWindow
--- a/src/e2e/puppeteer/__tests__/startup.ts
+++ b/src/e2e/puppeteer/__tests__/startup.ts
@@ -1,0 +1,63 @@
+import type { WindowEm } from '../../../initialize'
+import { page } from '../setup'
+
+type WindowEmWithStartupFlags = WindowEm & {
+  testFlags: WindowEm['testFlags'] & {
+    thoughtspaceInitBlocker?: Promise<void> | null
+  }
+}
+
+type StartupWindow = Window & {
+  __EM_TEST_FLAGS__?: {
+    thoughtspaceInitBlocker?: Promise<void>
+  }
+  __releaseThoughtspaceInit?: () => void
+}
+
+vi.setConfig({ testTimeout: 30000, hookTimeout: 20000 })
+
+it('handles keyboard commands while thoughtspace initialization is delayed', async () => {
+  await page.evaluateOnNewDocument(() => {
+    const windowWithTestFlags = window as StartupWindow
+    windowWithTestFlags.__EM_TEST_FLAGS__ = {
+      thoughtspaceInitBlocker: new Promise<void>(resolve => {
+        windowWithTestFlags.__releaseThoughtspaceInit = resolve
+      }),
+    }
+  })
+
+  await page.reload({ waitUntil: 'domcontentloaded' })
+
+  await page.waitForFunction(() => Boolean((window.em as WindowEmWithStartupFlags).testFlags.thoughtspaceInitBlocker))
+  await page.waitForFunction(() => !document.querySelector('[aria-label=modal]'))
+  await page.waitForSelector('[aria-label=empty-thoughtspace]')
+
+  try {
+    await page.keyboard.press('Enter')
+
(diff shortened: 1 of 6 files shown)


src/e2e/puppeteer/__tests__/startup.ts
```diff
@@ -0,0 +1,63 @@
...
+  await page.waitForFunction(() => Boolean((window.em as WindowEmWithStartupFlags).testFlags.thoughtspaceInitBlocker))
+  await page.waitForFunction(() => !document.querySelector('[aria-label=modal]'))
+  await page.waitForSelector('[aria-label=empty-thoughtspace]')
+
+  try {
+    await page.keyboard.press('Enter')
+
+    await page.waitForFunction(() => {
+      const em = window.em as WindowEm
+      return em.testHelpers.getState().lastUndoableActionType === 'newThought'
```
raineorshine: Accessing the Redux store is specifically disallowed in puppeteer and iOS tests. These are integration tests. They should only be able to access the UI as the user does.

If this is the *only* way to accomplish the desired behavior, then we can discuss possible exceptions. However I should emphasize that this should only be necessary in extreme circumstances.

Adding a dependency on `lastUndoableActionType` is also poor because this is an implementation detail of the undo/redo functionality. We don't want generic setup tests to be tightly coupled to the undo/redo implementation.

src/initialize.ts
```diff
@@ -84,9 +84,14 @@ const updateThoughtsThrottled = throttleConcat<PushBatch, void>((batches: PushBa
...
 
-/** Initilaize local db and window events. */
+/** Initialize local db and window events. */
 export const initialize = async () => {
   initOfflineStatusStore(/* websocket */)
+  const eventHandlers = initEvents(store)
+
+  if (testFlags.thoughtspaceInitBlocker) {
+    await testFlags.thoughtspaceInitBlocker
+  }
```
raineorshine: Is `initThoughtspace` stubbed? I seem to recall that the data provider is stubbed out with an in-memory provider. It would be great if we could inject the delay without modifying application code.

raineorshine (changes requested):

Registering app input handlers before async thoughtspace startup seems reasonable to me. It may introduce the possibility of users being able to perform certain actions before the thoughtspace is ready, but we would need special handling for that anyway. We probably should allow user interaction from t0 and then hold transactions in the middleware until the thoughtspace is ready.

src/e2e/puppeteer/__tests__/startup.ts
marcus-pousette: Yea, I agree. I removed the Redux/lastUndoableActionType assertion. The test now only verifies the user-visible behavior: after pressing Enter during delayed thoughtspace startup, an editable empty thought appears in the DOM.

https://github.com/cybersemics/em/pull/4384/commits/8439a6e5f4a1cf24f8f00c6ec6d79b068b15e5e9

src/initialize.ts
marcus-pousette: initThoughtspace is not stubbed in Puppeteer today. This test runs against the real browser bundle and yjs provider. I removed the Redux/lastUndoableActionType dependency and moved the blocker out of initialize.ts into the yjs provider init path, similar to the existing provider-level replicationDelay test flag. A full in-memory provider swap for Puppeteer would be cleaner, but I think that is separate test infrastructure work, maybe something we should do in a separate PR? I am also happy to do it part of this PR if ncessary but the work might create more LoC change

marcus-pousette:
@raineorshine Hey! I addressed your comments, feel free to review!

src/initialize.ts
```diff
@@ -178,7 +179,7 @@ export const initialize = async () => {
 
   return {
     thoughtsLocalPromise,
```
raineorshine: Fwiw, `thoughtsLocalPromise` is not used outside this module and can be removed from the return.

raineorshine (changes requested):
Hi, thanks for the update.

This nicely simplifies the intervention in the application code and the overall LOC is small. However, the number of different variables exposed to different parts of the application at different points in its life cycle does not seem to me to be justified.

- `PreloadedTestFlags`
- `__EM_TEST_FLAGS__`
- `StartupWindow`
- `thoughtspaceInitBlocker`
- `__releaseThoughtspaceInit`

This is a large amount of conceptual overhead for a simple mutex. I'm already dreading having to explain to a future developer the difference between `__EM_TEST_FLAGS__`, `em.testFlags`, and `PreloadedTestFlags`.

---

It seems like the difficulty stems from the fact that `initialize` is called immediately on load and then launches into `initThoughtspace` without any opportunity to intervene. 

One idea:

1. Separate `initEvents` from `initialize`.
2. Create a test flag like `em.testFlags.preventInitialize = true` that blocks `initialize` entirely on load, then expose `initialize` to call as you please: `em.testFlags.initialize()`

This still creates two points of surface area, but it avoids the awkwardness of different test flags on different lifecycles. It also avoids additional global variables that start with `__`, which `em.testFlags` was specifically designed to avoid by consolidating test injections in one typesafe place.

Whether you take the above approach or find an alternative, I'd love to see a solution that entails a commensurate level of complexity to the effect desired.

src/initialize.ts
raineorshine: Thanks, I agree we can consider it in a separate PR. I'll focus on a clean way to block thoughtspace initialization in this PR.

marcus-pousette:
Totally agree with your comment! Testing something that happens when the apps initializes is not the easy hence why first commit was a bit messy initially, but your idea is really good and I did another pass to fix it https://github.com/cybersemics/em/pull/4384/commits/4ad10980f66b5e7a631bcc7915f69b6bd7430e35

src/initialize.ts
```diff
@@ -84,9 +84,27 @@ const updateThoughtsThrottled = throttleConcat<PushBatch, void>((batches: PushBa
   })
 }, UPDATE_THOUGHTS_THROTTLE)
 
-/** Initilaize local db and window events. */
+let eventHandlers: ReturnType<typeof initEvents> | null = null
+
+/** Initialize app event handlers once. */
+export const initAppEvents = () => {
+  if (eventHandlers) return eventHandlers
```
raineorshine: I think you can hide the idempotency within `initEvents` itself. `eventHandlers` is always initialized when the handlers are added and always cleaned up when the the handlers are cleaned up.

That would be better than having `initEvents` and `initAppEvents` both floating around.

src/e2e/testFlags.ts
```diff
@@ -1,24 +1,37 @@
...
+    ? null
+    : ((window.em as { testFlags?: Partial<TestFlags> } | undefined)?.testFlags ?? null)
+
+/** Test flags that are injected into window.em.testFlags. */
+const testFlags: TestFlags = {
   logActions: false,
   logMultigesture: false,
   expandHoverDelay: null,
   replicationDelay: 0,
+  preventInitialize: preloadedTestFlags?.preventInitialize ?? false,
```
raineorshine, line 33: I forgot we had to merge the preloaded test flags from `evaluateOnNewDocument`. This part is still a bit awkward, but I think it's acceptable for now. 

In the future we can maybe find a cleaner pattern for hydrating the test flags.

raineorshine (changes requested):
Looks good, thanks. Just one suggestion.

src/util/initEvents.ts
```diff
@@ -406,10 +416,14 @@ const initEvents = (store: Store<State, any>) => {
     lifecycle.removeEventListener('statechange', onStateChange)
     resizeHost.removeEventListener('resize', updateSize)
     virtualKeyboardHandler.destroy()
+    eventHandlers = null
   }
 
+  const handlers = { keyDown, keyUp, cleanup }
```
raineorshine: You can assign directly to `eventHandlers` rather than creating an additional local variable.

raineorshine (changes requested):

raineorshine (approved):
Thank you!

PR #4435, opened 2026-06-22
More robust autoscroll and autoscroll prevention

### LINKED ISSUE #3765: [iOS Capacitor] `position: fixed` elements jump when tapping a thought in a long list
Tested & reproducible in both iOS Safari & iOS Capacitor.

## Steps to Reproduce

1. Ensure the thoughtspace contains enough thoughts that the editor begins to scroll.
2. Scroll to the top of the editor.
3. Tap one of the thoughts near the top of the screen, opening the virtual keyboard.
4. Tap one of the thoughts towards the _bottom_ of the screen.

## Current Behavior

iOS' native autoscroll interferes. This causes the webview's scroll position to change. While it does, `position: fixed` elements move with the scroll position.

https://github.com/user-attachments/assets/bb8b1150-7ce5-40ed-9eba-ca758df14335

## Expected Behavior

`position: fixed` elements should stay fixed in-place, even during autoscroll transitions.

## Notes

- I found `preventAutoscroll`, which generally does a good job of mitigating iOS' autoscroll transitions – maybe it's just missing some logic for this particular edge case?
- If we're unable to adapt `preventAutoscroll` to fix, this issue might require a native solution in Capacitor.

fbmcipher:
**Fixes #3765**

https://github.com/user-attachments/assets/1ea02e18-0dbc-448d-a9b2-cdcd59c087ea

## Background

On iOS Capacitor, the existing autoscroll made `position: fixed` elements (toolbar, navbar) visibly jump, and the cursor would often end up hidden behind the virtual keyboard.

In #3765, [investigations concluded](https://github.com/cybersemics/em/issues/3765#issuecomment-3881695313) that this is an issue with iOS' native autoscroll that wasn't treatable from within JavaScript. Given that we are already manipulating iOS' autoscroll with the `preventAutoscroll.ts`, we came to the decision to [figure out a way to fully prevent native autoscroll, and take over ourselves with JavaScript](https://github.com/cybersemics/em/issues/3765#issuecomment-3899292075).

[Sandbox testing showed this could work](https://github.com/cybersemics/em/issues/3765#issuecomment-3932445927), but with the tradeoff that we'd need to fully take over caret placement logic in JavaScript, too. Luckily, this was *already* being implemented in #3410, and recently shipped – unblocking this issue for further development.

## Approach

Using the `preventScroll` option on `HTMLElement.focus()` (which is supported in iOS 15.5 or later), we now **stop iOS' native autoscroll from ever firing at all** when focussing a new editable in a new module called `focusWithAutoscroll.ts`. This is contrast to the centering logic approach used by `preventAutoscroll.ts`, which worked in Mobile Safari but broke down in Capacitor.

`scrollCursorIntoView` now solely handles autoscroll. It has been updated with smooth scrolling and a new "comfort zone" concept which makes autoscroll work with a simple rule: if the cursor is about to cross a comfort zone threshold, scroll so the cursor falls within the "comfort zone" region.

### `focusWithoutAutoscroll.ts`

This new module is used across the application to focus an editable element without autoscroll. It's used to prevent autoscroll based on touch focus changes, cursor changes, keyboard events, programmatic focus changes, etc.

### `autoscrollEdges.ts`

This module defines, in viewport co-ordinates, the editor's "**comfort zone**". These measurements create a kind of "safe area". A cursor that falls outside the comfort zone will trigger autoscroll. This visualization by Opus 4.8 may help to visualize this:

CleanShot 2026-06-23 at 00 00 23@2x

(Before, these values were computed inline in `scrollCursorIntoView`. Now, they're defined in their own module.)

A new measurement, called the "landing margin", determines how far into the comfort zone the cursor should settle once a scroll has fired – rather than landing right at the edge of the comfort zone. Here, that's `height / 2`

When the keyboard is open, a special `bottomBuffer` changes to move the autoscroll edge to the tip of the keyboard. We subscribe to the new `virtualKeyboardStore` to get accurate and up-to-date keyboard measurements. 

These values might need tweaking to align with the desired feel, but the general functionality 

### `scrollCursorIntoView.ts`

This is the module that actually handles autoscroll. It mostly gets a cleanup:

- No need to wait for `preventAutoscroll`'s timeout – much cleaner.
- Consume keyboard height from `virtualKeyboardStore` instead of `viewport.ts`. This gets autoscroll working on Android/iOS Capacitor.
- Use the `autoscrollEdges` rather than computing the thresholds inline.
- Tricky logic which tries to find the cursor's *target* position (i.e. where it *will be*), rather than where it is mid-animation. That lets us scroll as soon as we know the cursor is about to step outside the comfort zone.
- *Don't* smooth scroll to targets that are really far away. In my testing, it looked visually jarring when scrollIntoViewIfNeeded was called on a function that was more than a screenful away. In these cases, we just instantly scroll to the destination.

### `virtualKeyboardStore`

Added a `targetHeight` value (where the keyboard will be) alongside the existing spring-animated `height` (which is used for elements that want to move/animate/stick *with* the keyboard).

We can do this elegantly with Capacitor (`iOSCapacitorHandler`), which reports that value to us. On iOS (`iOSSafariHandler`), we need to predict it, and correct our prediction once we know the keyboard's height. This seemed to work okay in testing.

While we're at it, moved the virtual keyboard prediction logic into `iOSSafariHandler` rather than `viewport.ts`. 

## Notes
- `scrollCursorIntoView` uses direct DOM reads to read information about the cursor overlay. This isn't new – `main` did this too – but it does feel like an unusual approach. I've left it here for now, LMK if you feel a need to refactor this. Probably we'd just track information about the cursor in a ministore.

## Test checklist
- [ ] iOS Capacitor: tap a thought near the bottom of a long list — toolbar/navbar no longer jump; cursor lands above the keyboard.
- [ ] iOS Safari: autoscroll behaves the same, though some toolbar jumping is unavoidable in Safari.
- [ ] Hold Enter to add many thoughts rapidly — scroll keeps a steady one-line-per-Enter rhythm.
- [ ] Tap a thought already covered by an open keyboard — it scrolls into view.
- [ ] Top edge: navigating up reveals ~one thought under the toolbar.
- [ ] Desktop (no virtual keyboard): no behavior change; text selection still works on non-iOS.


src/components/Note.tsx
```diff
@@ -150,7 +149,19 @@ const Note = React.memo(
...
+      if (!note) return
+
+      // Block native focus + native caret-from-tap, then route through focusWithoutAutoscroll
+      // (the same single entry point Editable uses) — focuses with preventScroll, places the
+      // caret at the tap offset, and suppresses the selection-driven autoscroll on iOS.
+      const { offset } = getCaretOffset(note, { clientX: e.clientX, clientY: e.clientY })
+      if (offset !== null) {
+        e.preventDefault()
+        focusWithoutAutoscroll(note, { offset })
+      }
```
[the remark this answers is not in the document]
fbmcipher: Great catch.

src/device/virtual-keyboard/handlers/iOSSafariHandler.ts
```diff
@@ -39,11 +66,34 @@ const updateIOSSafariKeyboardState = () => {
...
+      // Publish the keyboard's *predicted* final height once, at open detection — emulating
+      // Capacitor's keyboardWillShow semantics (final height upfront, one targetHeight event per
+      // keyboard movement). Safari can only measure the keyboard progressively as it slides (see
+      // currentTargetHeight above); publishing every intermediate measurement would fire
+      // subscribers like useScrollCursorIntoView 5-10 times per open, restarting the autoscroll
+      // mid-flight each time. The prediction is the settled height of the previous open, seeded
+      // from measureKeyboardHeight's per-orientation estimate on the first open.
+      if (isOpening) {
+        predictedFinalHeight = Math.max(targetHeight, predictedFinalHeight ?? targetHeight)
+        virtualKeyboardStore.update({ open: true, targetHeight: predictedFinalHeight })
```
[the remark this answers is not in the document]
fbmcipher: @copilot Agreed, but we should probably use the `ScreenOrientation` API here instead of inferring the screen orientation using `window.innerHeight > window.innerWidth`.

fbmcipher:
@raineorshine Ready for review!

Note: I regenerated one snapshot test (`color-theme-colored-and-highlighted-text-1`) because the behaviour in `main` actually had a subtle scroll bug baked in which caused the `scrollY` to floor to `1` rather than `0`.

This PR fixes the bug, so the snapshot test needed to be regenerated.

src/device/focusWithoutAutoscroll.ts
raineorshine, line 1: This should probably go in the selection module.

src/device/preventAutoscroll.ts
raineorshine, line 1: With a PR as large as this, I would have considered preserving `preventAutoscroll`. Flip a flag so the next `selection.set` uses `focusWithoutAutoscroll`, and then flip it back on `preventAutoscrollEnd`. This would have preserved the existing architecture during the migration and reduced the size of the diff. Then removing `preventAutoscroll` completely would have been a pure refactoring PR.

src/components/Editable/useEditMode.ts
```diff
@@ -80,28 +67,19 @@ const useEditMode = ({
...
-        Not calling asyncFocus when the selection is already on a thought prevents the infinite loop.
-        Also, setTimeout is frequently pushed into the next frame and the keyboard will intermittently close on iOS Safari.
-        Replacing setTimeout with requestAnimationFrame guarantees (hopefully?) that it will be processed before the next repaint,
-        keeping the keyboard open while rapidly deleting thoughts. (#3129)
+      if (!shouldSetSelection) return
 
-        If the last action is swapParent, set the selection synchronously to keep the focus stable after the swap.
-      */
-        if (isTouch && isSafari() && lastUndoableActionType !== 'swapParent' && !selection.isThought()) {
-          asyncFocus()
```
raineorshine, line 155: Can you confirm that Auto-Capitalization and swapParent behavior are preserved?

src/components/Editable/useEditMode.ts
```diff
@@ -122,28 +100,11 @@ const useEditMode = ({
...
-  // Provide an escape hatch to allow the next default selection rather than setting it.
-  // This allows the user to set the selection in the middle of a non-cursor thought when keyboard is open.
-  // Otherwise the caret is moved to the beginning of the thought.
-  const allowDefaultSelection = useCallback(() => {
-    disabledRef.current = true
-    // enable on next tick, which is long enough to skip the next setSelectionToCursorOffset
-    setTimeout(() => {
-      disabledRef.current = false
-    })
-  }, [])
```
raineorshine, line 203: What changed that this is no longer required?

src/device/autoscrollEdges.ts
raineorshine, line 1: It looks like this is only used in `scrollCursorIntoView`. If so, it should probably not be separated out into a separate module. This separation is obscuring some of the complex DOM and component dependencies, including toolbar, navigation, keyboard and Layout. I would suggest putting this back into `scrollCursorView` and then thinking more deeply about encapsulation boundaries, dependencies, and the desired API.

`device` utilities are meant to be low level functions that abstract away some of the browser API details and can easily be stubbed in tests or other non-browser environments. `scrollCursorIntoView` already violates this encapsulation, so we'd like to try to find way to not make it worse.

src/stores/viewport.ts
```diff
@@ -1,66 +1,32 @@
...
 
-/** Throttled update of viewport height. Invoked on window resize. */
+/** Throttled update of viewport dimensions. Invoked on window resize. */
 export const updateSize = _.throttle(
   () => {
-    // There is a bug in iOS Safari where visualViewport.height is incorrect if the phone is rotated with the keyboard up, rotated back, and the keyboard is closed.
-    // It can be detected by ensuring the visualViewport portrait mode matches window portrait mode.
-    // If it is invalid, go back to the default
-    const isPortrait = window.innerHeight > window.innerWidth
-    const currentKeyboardHeight = window.visualViewport ? window.innerHeight - window.visualViewport.height : 0
```
raineorshine, line 40: Is this special case no longer needed because `visualViewport` is not used anymore? Just want to make sure the iOS Safari bug described here doesn't return.

raineorshine (commented):
Hi, thanks so much for the PR. This will be a huge improvement!

`focus({ preventScroll: true })` did not exist when em began, and it's about time we got a native way to prevent autoscroll.

I do have some concern about how much is done in one PR. This not only makes reviewing more challenging, but increases the likelihood of regressions. Can you explain why all of these changes must to be done together? For example, preventing native autoscroll is orthogonal to `scrollCursorIntoView` as far as I know. I would have assumed a good starting point would be to just sub out the outdated `preventAutoscroll` with native `focus({ preventScroll: true })`. If that was safe, then we could move on to other improvements. 

The concept of a comfort zone and autoscrollEdges sound promising, but I would be more inclined to consider changes to `scrollCursorIntoView` in a dedicated PR with specific test cases that we are targeting.

Given that this PR makes deep changes to `useEditMode`, it means that all of the manual test cases from #3410 and #4371 are at risk. There are delicate timing issues related to pointer events, touch events, focus events, and offset calculations. All selection behavior will need to be re-tested.

@BayuAri We can cautiously proceed with manual testing, but if you encounter more than one selection-related regression, you should bail so that we can stop and see if there is a way to break this PR into smaller pieces. There is no point documenting a large number of test cases when there are potentially too many changes to safely make at once.

@fbmcipher Please hold off on the architectural feedback and internal changes until we demonstrate the behavioral viability.

Thanks!

BayuAri (commented):
### Test Results:
- [x]  iOS Capacitor: tap a thought near the bottom of a long list — toolbar/navbar no longer jump; cursor lands above the keyboard.
- [x]  iOS Safari: autoscroll behaves the same, though some toolbar jumping is unavoidable in Safari.
- [x]  Top edge: navigating up reveals ~one thought under the toolbar.
- [x]  Desktop (no virtual keyboard): no behavior change; text selection still works on non-iOS.
- [ ] Hold Enter to add many thoughts rapidly — scroll keeps a steady one-line-per-Enter rhythm.
- [ ]  Tap a thought already covered by an open keyboard — it scrolls into view.

### For discussion about the last 2 items:

### 1. Scroll is not steady on Mobile Safari when rapidly creating  a series of thought. 
After 7-8 thoughts, the newly created is started at the middle of the screen. 

https://github.com/user-attachments/assets/e02fb47e-e3da-4549-b239-5a96416b8fea

### 2. The view is scrolled onto the last position of the cursor when user selects all and favorite
## Step to Reproduce
1. Create 20 thoughts
2. Place the cursor on Thought 20
3. Scroll up to let the Thought 20 hidden from view
4. Select all 
5. Tap Favorite on Command Center


## Current behavior
The view is scrolled based on the position of the cursor where it left off. However, the thought is hidden under the Command Center.  
I think the app is assuming the viewport without Virtual Keyboard onscreen. 
Note: this does not happen on `main`

https://github.com/user-attachments/assets/fbcefa0b-bb82-4e2a-993f-e20878f3e80f

## Expected behavior
Nothing should happen
Auto scroll should not be triggered by favoriting on Select All thoughts. 


### Vibrating thoughts
In addition, I see an oddly vibrating thoughts on Mobile Safari 1x on this branch but I can not reproduce it anymore. 

https://github.com/user-attachments/assets/4a298c12-f78f-4a88-9683-8b20ecc8b28d


@raineorshine 
Please advice

fbmcipher:
@raineorshine Okay, sounds good – thank you for intervening.

The reason for the conflation was that I was focussed on a single PR to fix the single issue #3765 – but, yes, you're right that this is too big and complex as-is.

It might be a good idea to split this PR into _two_ separate chunks so that we can discuss "native scroll prevention" and "autoscroll" as separate tickets.

From the initial tests @BayuAri just did above, it looks like the "native scroll prevention" portion of this submission is working...

> - [X] iOS Capacitor: tap a thought near the bottom of a long list — toolbar/navbar no longer jump; cursor lands above the keyboard.

...but that the new JS-based autoscroll mechanism needs more work...

> - [ ] Hold Enter to add many thoughts rapidly — scroll keeps a steady one-line-per-Enter rhythm.
> - [ ] Tap a thought already covered by an open keyboard — it scrolls into view.

I'm thinking the best course forward would be to submit a new PR that _just_ focuses test on testing the new "native scroll prevention" as its own ticket and with its own test cases before we move on to tackling JS autoscroll. What do you think?

raineorshine:
> @raineorshine Okay, sounds good – thank you for intervening.
> 
> The reason for the conflation was that I was focussed on a single PR to fix the single issue #3765 – but, yes, you're right that this is too big and complex as-is.

If you mean "do whatever it takes to fix #3765" then it does accomplish that, but without a clear migration path to assess and bound risk.

> It might be a good idea to split this PR into _two_ separate chunks so that we can discuss "native scroll prevention" and "autoscroll" as separate tickets.

Likely more than two. Autoscroll itself consists of dozens of test cases that we have not defined yet. Improving `scrollCursorIntoView` is simply out of scope for #3765.

> I'm thinking the best course forward would be to submit a new PR that _just_ focuses test on testing the new "native scroll prevention" as its own ticket and with its own test cases...

Yes, that would be great.

---

Some questions afterwards:

- What is the minimum that is needed to fix #3765?
- What regressions does this create that will need fixing? What is the cause?
- Is that a manageable number to tackle at once?
- How can we break down the task into smaller PRs with *well-defined risk*?

The challenge comes from the fact that:

1. We have to break things to get them working again.
2. The code touches the highest risk areas of the codebase with the least reliable automated tests.

The size and complexity of the PR are a factor, but the bigger issue is how many sensitive parts of the codebase are being modified here. Just as we never delete any code without fully understanding why it's safe to do so, we don't want to delete or modify any browser selection code, platform-specific workarounds, or scroll logic without justifying the replacement.

The current PR description focuses too much on the new feature and how it works instead of on what code is disrupts, what the risks are, and the migration path.

fbmcipher:
@raineorshine Following our discussion today, I will **close this PR for now**.

PR #4440, opened 2026-06-23
Compensate for existing top padding in preventAutoscroll

### LINKED ISSUE #3598: Remove useMultiline

ethan-james:
Fixes #3598
Fixes regression caused by #4253 

Switching from line-height to padding caused an issue in [preventAutoscroll](https://github.com/ethan-james/em/blob/35018112b3a421cdce2d7ecd2e7fd6fd3f4f0308/src/device/preventAutoscroll.ts#L71) because its padding is overridden without accounting for the new top padding. In the declarative `shouldSetSelection` path that calls `preventAutoscroll`, it is left to [timeout](https://github.com/ethan-james/em/blob/35018112b3a421cdce2d7ecd2e7fd6fd3f4f0308/src/device/preventAutoscroll.ts#L81) after 10ms, which makes the issue more visible than if `preventAutoscrollEnd` is called `onFocus`.


src/device/preventAutoscroll.ts
```diff
@@ -60,12 +60,15 @@ const preventAutoscroll = (
 
   // below center
   if (yCenter < 0) {
+    // Read the rendered paddingTop in pixels. getComputedStyle resolves the value even when it is applied via a CSS class rather than the inline style property (el.style.paddingTop), and always returns a px value.
+    const paddingTopComputed = parseFloat(getComputedStyle(el).paddingTop) || 0
```
raineorshine: The programmatic value is a good idea, but we have to avoid the old style read/write interleaving problem. Performing back-to-back reads and writes of styles forces multiple paints and is a common performance problem. Instead we need to find a way to do all the reads first and then all the writes second.

raineorshine (changes requested):
Thanks for the quick fix.

src/device/preventAutoscroll.ts
ethan-james: Ah, I thought I was being clever by putting it inside the relevant condition, but I see what you mean.

raineorshine (approved):

Issue #4294, opened 2026-05-22
[IOS Capacitor] Overlapping Share dialog and virtual keyboard when sharing while the keyboard opens


BayuAri:
```
- A
- B
- C
- D
- E
- F
- G
- H
- J
- K
- L
- M
```

## Step to Reproduce
1. Kill the app and reopen
2. Place the caret on Thought M; 
3. Tap the left side of the screen to dismiss the caret
4. Tap Thought A to move the cursor onto it
5. Tap Thought A again to place the cursor
6.  Tap Export button on the toolbar
7. Tap Share button
## Current behavior
Share button is showing up under the virtual keyboard.
Note: if above step is not reproducible, repeat the step on other thought.

<img width="576" height="1263" alt="Image" src="https://github.com/user-attachments/assets/2b8f04ea-ad9c-466f-8fa1-ecc9d1878a1c" />

https://github.com/user-attachments/assets/ece407f2-e27b-4bee-b287-35c8c1d98b7b

## Expected behavior
Share button should not be overlapping with virtual keyboard.

PR #4445, opened 2026-06-24
iOS: dismiss keyboard before opening Share sheet to prevent overlap (#4294)

BayuAri:
#4294

## Problem

On the iOS Capacitor app, exporting from a focused thought could open the native Share sheet while the software keyboard was still visible, causing the share surface to overlap and hide controls (#4294).

## Root cause

`onExportClick` in `src/components/modals/Export.tsx` calls `navigator.share()` while the editable from the cursor is still focused, so iOS keeps the software keyboard up and presents the share sheet over it.

## Fix

In `onExportClick`, on the iOS Capacitor app (`isIOS`), blur the focused editable (`selection.clear()`) and dismiss the keyboard (`Keyboard.hide()`) **before** invoking `navigator.share()`. This is done synchronously (no `await`) so the user-activation context required by `navigator.share()` is preserved.

## Reproduction note

This bug is specific to the native iOS Capacitor app (native software keyboard overlapping the native Share sheet) and cannot be reproduced in desktop/Chrome emulation. Automated iOS reproduction via BrowserStack App Automate was unavailable in this environment (BrowserStack credentials empty, `wdio` MCP not configured), so the fix is grounded in the documented reproduction evidence and the existing keyboard/share code paths.

## Tests

Added `src/components/__tests__/ModalExportShare.ts`, a unit regression test that mocks the iOS platform and the Capacitor `Keyboard` plugin and asserts that `Keyboard.hide()` is called **before** `navigator.share()` when the Share button is tapped. Verified the test fails on the unfixed code (for the right reason — `Keyboard.hide` not called) and passes with the fix.

## Architectural plan

See the issue-repro + plan gate summary in the PR thread. Key decisions: extend the sole `onExportClick` handler rather than introduce a new path; dismiss the keyboard synchronously to preserve `navigator.share()` user activation; gate the new branch on `isIOS` so download/web/desktop paths are untouched.

### DIFF
--- a/src/components/__tests__/ModalExportShare.ts
+++ b/src/components/__tests__/ModalExportShare.ts
@@ -1,0 +1,66 @@
+import { Keyboard } from '@capacitor/keyboard'
+import { act } from 'react'
+import { importTextActionCreator as importText } from '../../actions/importText'
+import { showModalActionCreator as showModal } from '../../actions/showModal'
+import click from '../../test-helpers/click'
+import createTestApp, { cleanupTestApp } from '../../test-helpers/createTestApp'
+import dispatch from '../../test-helpers/dispatch'
+import { setCursorFirstMatchActionCreator as setCursor } from '../../test-helpers/setCursorFirstMatch'
+
+// Emulate the native iOS Capacitor app so that the iOS-specific keyboard-dismissal path in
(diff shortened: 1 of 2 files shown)


BayuAri:
Closing in favor of #4339, which is the PR linked to issue #4294. The fix has been consolidated onto #4339's branch.

PR #4446, opened 2026-06-24
Webdriverio : touch events

### LINKED ISSUE #4251: Find out if webdriverio can trigger touch events

karunkop:
Close #4251

 Implementations as per discussion in 
 - https://github.com/cybersemics/em/issues/4251#issuecomment-4547762894
 - https://github.com/cybersemics/em/issues/4251#issuecomment-4589389967


src/e2e/iOS/__tests__/color.ts
```diff
@@ -0,0 +1,51 @@
+import getEditingText from '../helpers/getEditingText'
+import paste from '../helpers/paste'
+import tap from '../helpers/tap'
+import waitForElement from '../helpers/waitForElement'
+
+/** Tap an element by selector at its center.*/
+const tapElementBySelector = async (selector: string) => {
```
raineorshine: This functionality should be rolled into the `tap` helper. Great opportunity for polymorphism here to improve the API. `WebdriverIO.Element` and `string` have no overlap.

Notice that we keep a separation between test cases, which reproduce user actions in a procedural, readable, user-oriented way, and helpers, which provide the low-level browser interaction that is needed.

src/e2e/iOS/__tests__/color.ts
```diff
@@ -0,0 +1,51 @@
...
+  if (!isClickable) {
+    await el.scrollIntoView({ block: 'center' })
+    await el.waitForClickable({ timeout: 10000 })
+  }
+
+  return tap(el, { horizontalTapLine: 'center' })
+}
+
+/** Retrieve the innerHTML of the first note on the page. Assumes that there will be only a single note. */
+const getFirstNoteText = () => browser.execute(() => document.querySelector('[aria-label="note-editable"]')?.innerHTML)
```
raineorshine: I think `$(...)` may be a bit more standard than `browser.execute(() => document.querySelector(...))`.

src/e2e/iOS/helpers/gesture.ts
```diff
@@ -17,52 +17,86 @@ interface PointerAction {
...
-  // Build actions array for performActions
-  // Safari/XCUITest doesn't support the DELETE /actions endpoint (releaseActions)
-  // which WebDriverIO's action().perform() calls automatically after performing
-  const pointerActions: PointerAction[] = [
-    { type: 'pointerMove', duration: 0, x: Math.round(xStart), y: Math.round(yStart), origin: 'viewport' },
-    { type: 'pointerDown', button: 0 },
-    { type: 'pause', duration: waitMs },
-  ]
+  if (xStart === undefined || yStart === undefined) {
+    const viewport = await browser.execute(() => ({
```
raineorshine: Try to avoid separate round trips to the browser. Every await adds time. You can do all this in a single `browser.execute`.

src/e2e/iOS/helpers/hideKeyboardByTappingDone.ts
```diff
@@ -1,14 +1,20 @@
...
   const oldContext = ((await browser.getContext()) as string) || 'NATIVE_APP'
   await browser.switchContext('NATIVE_APP')
   const done = await browser.$('//XCUIElementTypeButton[@name="Done"]').getElement()
   await done.click()
   await done.waitForExist({ reverse: true })
   await browser.switchContext(oldContext)
+
+  await browser.execute(() => {
+    const active = document.activeElement
+    if (active instanceof HTMLElement) active.blur()
```
raineorshine: What happens on a physical iPhone when the keyboard is closed by tapping Done? Is `document.activeElement` already cleared?

We want to make sure our tests don't deviate from actual behavior.

raineorshine (changes requested):

src/e2e/iOS/helpers/hideKeyboardByTappingDone.ts
karunkop: What do you mean when you say we don't want tests to deviate from actual behavior in this case?

raineorshine: I mean that we don't want integration tests to do anything that the user can't do with their device. The user closes the keyboard by tapping the Done button and they don't have access to `blur()`. So why is the test doing something "special" that the user can't do?

We don't want to give integration tests special powers. Then we end up having test coverage of those special powers instead of test coverage over the actual behavior that the user is capable of.

src/e2e/iOS/__tests__/color.ts
```diff
@@ -1,31 +1,18 @@
...
+const getFirstNoteText = async () => (await $('[aria-label="note-editable"]')).getHTML({ includeSelectorTag: false })
 
 it('Can change the background color of a thought that already has the same background color applied to part of its text, then change the text color', async () => {
   const content = '- some <font color="#000000" style="background-color: rgb(255, 87, 61);">formatted</font> text'
   await paste(content)
 
-  await tapElementBySelector('[data-testid="toolbar-icon"][aria-label="Text Color"]')
-  await tapElementBySelector('[aria-label="background color swatches"] [aria-label="red"]')
-  await tapElementBySelector('[aria-label="text color swatches"] [aria-label="red"]')
+  await tap('[data-testid="toolbar-icon"][aria-label="Text Color"]', { horizontalTapLine: 'center' })
```
raineorshine: Why doesn't the default horizontalTapLine work?

Color tests should not have to worry about where to tap the button.

karunkop: The default value for `horizontalTapLine` is `left`. However, for toolbar icon buttons, calculating the tap coordinates from the left edge does not consistently target the button accurately. To address this, the `center` option was introduced, which calculates the tap coordinates at the horizontal center of the button, resulting in more reliable and accurate tap interactions.

> Color tests should not have to worry about where to tap the button.

You're right about this part. I was thinking if we want to keep this functionality within `tap`, we could introduce a new option that indicates the target is a toolbar icon, allowing the tap coordinates to be calculated accordingly.

raineorshine: What about making `center` the global default? This more closely matches user semantics. If you tell a user to tap on a button, they're more likely to tap somewhere near the center than tap the exact edge.

The question is whether that will break other tests. If other tests depend on the tap occurring on the left edge (for example to set the cursor at the beginning of the thought), then arguably those tests should make that dependency explicit by specifying `horizontalTapLine`.

Please try to avoid phrases like "reliable" or "accurate" without going into the details. We're dealing with exact pixels. The button is rendered in the same place every time. You should understand exactly why left edge tap breaks on toolbar buttons. Only then can we determine the best solution. If we lack understanding, then our solution may be based on false assumptions and we may fail to apply the lesson properly in future scenarios.

> I was thinking if we want to keep this functionality within `tap`, we could introduce a new option that indicates the target is a toolbar icon, allowing the tap coordinates to be calculated accordingly.

`tap` is a general function. General functions should not be concerned with specific usage. Your gut feeling should tell you that parameterizing general functions with context-dependent behavior is a bad idea. At least `horizontalTapLine` is general.

raineorshine:
@karunkop Is this ready for review or are you still working on it? Thanks!

karunkop:
> @karunkop Is this ready for review or are you still working on it? Thanks!

@raineorshine , This is now ready for review!

raineorshine:
Thank you! 

It looks like BrowserStack is failing though.

karunkop:
> Thank you!
> 
> It looks like BrowserStack is failing though.

Looks like that was the session issue. Thought it was fixed after we capped the `maxInstances` to 2 for running parallel sessions. For now I have re-run the test. Might need a detailed investigation about this particular issue.

src/e2e/iOS/helpers/tap.ts
```diff
@@ -37,82 +67,81 @@ const tap = async (
...
+          },
+          { type: 'pointerDown', button: 0 },
+          { type: 'pause', duration: releaseDelayMs },
+          { type: 'pointerUp', button: 0 },
+        ],
+      },
+    ])
+  } finally {
+    await browser.switchContext(oldContext)
+    await restoreScroll(scrollBefore)
```
raineorshine, line 143: Restoring the scroll position is 1) not the responsibility of tap (mixing concerns) and 2) will prevent us from ever testing scrollCursorIntoView independently.

> A delayed scroll from `useScrollCursorIntoView` can override the restored position

Why? How? Under what circumstances? Why in a test and not on a physical device?

Remember that device tests are supposed to match the user's behavior as closely as possible and no do anything that the user wouldn't/can't do on their own.

Feel free to discuss the tradeoffs and architectural decisions with me as you look for a better solution. Thanks.

raineorshine (changes requested):

raineorshine:
@karunkop Please respond to the last feedback when you get a chance.

In progress PRs should not sit for over a month.

src/e2e/iOS/__tests__/color.ts
raineorshine: Added as part of #5018

PR #4462, opened 2026-06-26
Restyle the Mobile Command Universe gesture diagrams

### LINKED ISSUE #4186: Restyle the gesture diagram in the Command Universe
> **Restyle the Gesture Diagram in the Command Universe – a thicker stroke, a bluer shade in the gradient, a cleaner arrowhead, and proper centering.**

Image

## Background

`GestureDiagram` (`src/components/GestureDiagram.tsx`) renders the static SVG previews shown next to each command in the Command Universe, Gesture Menu, and Help modal. In #4185 we restyled everything around the diagrams. In this spec, we focus on restyling the diagrams themselves.

The key point: only restyle the `GestureDiagram` as it appears in the Command Universe. **Other sites where `GestureDiagram` is used should appear identically.**

`TraceGesture` (the live stroke drawn as the user performs a gesture) is **out of scope** for this spec.

## Scope

- [ ] Thicker stroke on `GestureDiagram`, prop-controllable
- [ ] Two-color gradient — `#58B5D4` @ 45% opacity → `#FFFFFF` — exposed as a prop on `GestureDiagram`
- [ ] Cleaner arrowhead geometry
- [ ] Diagrams correctly centered in relation to content (the "Jump Forward" example)

#### Design notes
- Diagrams should read noticeably heavier than they do today. Match the "after" mockup by eye.

#### Implementation notes
- `GestureDiagram` already accepts `strokeWidth`. Just bump the value at the call site in `CommandItem.tsx` (currently `strokeWidth={4}` for the inline gesture cell, `strokeWidth={7.5}` for the grid view). Eyeball the value from the mockup.

#### Design notes
- Gradient runs along the direction of the stroke, in the same per-segment way the existing single-color gradient does.

#### Implementation notes
- Today, `GradientStyleBlock` interpolates a single color toward `colors.bg`:
```ts
  stop-color: color-mix(in srgb, ${stopColor} ${startPercent}%, ${token('colors.bg')})
```
- Add a new prop:
```ts
  gradient?: { from: string; to: string }
```
- When `gradient` is supplied, replace the single-color path in `GradientStyleBlock`.
- When the prop is not supplied, fall back to the current single-color behavior. Don't break existing call sites.

#### Design notes
- Replace the current choppy outlined arrowhead with the cleaner chevron in the screenshot.

#### Implementation notes
- Today the outlined marker is:
```tsx
  <path
    d="M 0 0 L 5 5 L 0 10"
    fill="none"
    stroke={...}
    strokeWidth={strokeWidth / 3}
  />
```
  inside a `viewBox='0 0 10 10'`. The `/ 3` ratio is the main cause of the spindly appearance.
- Bump the marker stroke ratio (try `/ 1.5` or `* 0.6` — eyeball against the mockup).
- Add `strokeLinejoin="round"` and `strokeLinecap="round"` on the marker path.
- If the new chevron geometry differs from `M 0 0 L 5 5 L 0 10`, replace the path data accordingly.
- The filled arrowhead variant (`'M 0 0 L 10 5 L 0 10 z'`) is unaffected by this change.

#### Design notes
- The gesture diagram should sit visually centered inside its 32×32 cell – in the current build, `Jump Forward` has noticeably uneven top/bottom space. The red lines are the same size, which illustrates that there is more empty space on one side of the gesture diagram than the other. Not sure what's causing this:

Image

## Notes

- The arrowhead change applies to the outlined variant used in the Command Universe. The filled variant should remain untouched.

fbmcipher:
**Closes #4186.**

Screenshot 2026-06-26 at 01 15 36

This PR reworks how command gesture diagrams are rendered on the Command Universe ("Commands") surface. This involved giving `GestureDiagram` additional the props and capabilities it needed to match the spec mockup (like gradient strokes and a wider chevron arrowhead sizing), wiring those into the command grid, and fixing an iOS-only centering bug along the way.

## Testing

@BayuAri The changes in this PR should be purely aesthetic – no functionality changes in the Command Universe are expected. Because we've modified `GestureDiagram` a lot in this PR, it is probably worth checking the other interfaces in which that component is used, to make sure they still look and work the same as before. Thank you :)

---

### `GestureDiagram`

The bulk of the work. All additions to `GestureDiagram` are opt-in props with defaults that preserve the existing rendering everywhere else `GestureDiagram` is used.

- **Two-color gradient (`gradient`)** – fades the _entire_ stroke between two colors along the path. Critically, when a `gradient` is supplied the whole gesture is drawn as a *single* chord-aligned `<path>` rather than per-segment paths. Without doing this, the gesture diagram showed visible seams where per-segment strokes connected. Of course, we rely on the per-segment paths in other uses of `GestureDiagram`.

- **Wide chevron arrowhead (`arrowhead='outlined-wide'`), matching the design. Its geometry is exposed via `chevronApexAngle` (default 80°) and `chevronSize` (default 2.2× stroke) so the shape, size and angle of the arrowhead is fully tunable.

- **`cornerRadius`** – rounds the bends/corners in the gesture. This only works when `gradient` is passed, because it only works with the continuous `<path>` rendering mode.

- **`tipExtension`** – adds length to just the **final segment**, so that there's breathing room between the last bend and the arrowhead. Without this prop, the `outlined-wide` arrowhead overlaps the gesture and looks messy.

- **`fillContainer`** – the diagram sizes itself to its parent and uses a centered square viewBox clamped to a common extent, so every gesture fills its cell at a uniform visual scale regardless of shape.
  
- **`glow`** – toggles the soft drop-shadow glow (defaults to `true` to preserve current behavior). Our new design does *not* have a glow around gestures.

### iOS Safari centering fix

Straight gestures now compute their `viewBox` **during render** (so React owns the attribute) instead of setting it imperatively in an `onRef` callback. The `onRef` method was causing an inconsistency where iOS Safari wasn't always repainting, leaving some gestures off-center. Computing during render solved the issue.

### Command grid integration

- Wired the upgraded `GestureDiagram` into `CommandUniverseGridItem` with the production gradient / chevron / fit-to-container settings.
- **Replaced the table markup with divs.** The grid had inherited `<table>/<tbody>/<tr>/<td>` from `CommandTable`, but it renders a 2-column grid of cards, not a table, so we use divs instead.

### To-Do
- [ ] Resolve the failing Puppeteer test.


BayuAri:
## Issue A: [Chrome]  [TypeScript] Cannot find module 'webview-background' or its corresponding type declarations error

## Step to Reproduce
1. Open em on Mobile or Desktop Chrome

## Current behavior
`[TypeScript] Cannot find module 'webview-background' or its corresponding type declarations. error` is shown

image

## Expected behavior
User should be able to open em app with no issue

fbmcipher:
@BayuAri I think this I've seen this error before. Could you try a fresh reinstall (i.e. deleting `node_modules` with `rm -r node_modules` and then reinstalling with `yarn install`)? I don't think it is related to this PR.

BayuAri:
@fbmcipher 
You are right.
It is gone after i restart the server and vs code.
Sorry for the false alarm

BayuAri (commented):
It is great gradient color and smooth clean design. 
No issue from me.

raineorshine:
> [TypeScript] Cannot find module 'webview-background' or its corresponding type declarations.

@BayuAri This means the local `webview-background` package was not built correctly. It is usually built automatically after running `yarn`, but you can also trigger it manually with `yarn build:packages`.

I would recommend not getting in the habit of deleting `node_modules`. It's very rare that `node_modules` would become corrupted.

src/components/GestureDiagram.tsx
```diff
@@ -348,25 +423,158 @@ const GestureDiagram = ({
     return `M ${startX} ${startY} A ${radius} ${radius} 0 0 ${sweepFlag} ${endX} ${endY}`
   }
 
+  // Driven by the `gradient` prop. When a two-color gradient is supplied, render the whole gesture
+  // as a single <path> with one chord-aligned linear gradient. This avoids the visible seams that
+  // per-segment paths produce at their overlapping rounded joins, and applies to straight,
+  // `rounded`, and the special-case 'rdld' help glyph. When unset, the per-segment <style> path runs
+  // instead (see GradientStyleBlock).
+  const useSingleGradient = !!gradient
```
raineorshine, line 431: Reserve `useX` naming for hooks.

This variable probably shouldn't exist anyway. Just use `gradient`/`!gradient` as needed.

src/components/GestureDiagram.tsx
```diff
@@ -348,25 +423,158 @@ const GestureDiagram = ({
...
+            }
+          })()
+        : { start: positions[0], end: positions[positions.length - 1] }
+    : null
+
+  // Gated on `arrowhead === 'outlined-wide'` (and only for straight gestures — not rdld/rounded).
+  // Renders the chevron as a separate <path> whose apex sits at the gesture's natural tip, with two
+  // legs splaying BACKWARD (≈65° apex by default). Independent of `gradient`: the chevron can use
+  // the gesture's gradient or fall back to a solid color.
+  const useChevronArrowhead = arrowhead === 'outlined-wide' && path !== 'rdld' && !rounded && positions.length >= 2
```
raineorshine, line 474: Reserve `useX` naming for hooks.

src/components/GestureDiagram.tsx
```diff
@@ -414,41 +643,137 @@ const GestureDiagram = ({
...
+        )}
+
+        {useSingleGradient ? (
+          // Single continuous gesture path with chord-aligned gradient. For straight gestures,
+          // interior vertices may be softened with quadratic curves (cornerRadius > 0) so the
+          // bends look like rounded turns rather than mitered joins. For `rounded` gestures,
+          // concatenate the per-segment arcs into one continuous path so segment joins
+          // disappear and the gradient flows as one stroke.
+          <path
+            d={(() => {
```
raineorshine, line 657: Nested IIFEs are usually a sign that you need to break this down into smaller components/hooks.

src/components/CommandUniverseGridItem.tsx
```diff
@@ -50,31 +50,46 @@ const CommandUniverseGridItem: FC<CommandUniverseGridItemProps> = ({ command, se
...
+              size={150}
+              arrowSize={1}
+              strokeWidth={12}
+              color='#ffffff'
+              arrowhead='outlined-wide'
+              chevronApexAngle={80}
+              chevronSize={2.2}
+              cornerRadius={12}
+              tipExtension={28}
+              gradient={{ from: 'rgba(88, 181, 212, 0.45)', to: 'rgba(255, 255, 255, 1)' }}
```
raineorshine, line 85: Want to use a theme color?

raineorshine (changes requested):
Thank you for the submission!

`chevronPoints` and `computedViewBox` should probably be in a `useMemo` and/or custom hook. `GestureDiagram` gets a bit too big if we stuff all the new functionality into the existing component.

Please add visual test coverage of all the new GestureDiagram props to https://github.com/cybersemics/em/blob/afc7fa022b4a25710216f40c7efa915dc756a811/src/components/modals/TestGestureDiagram.tsx and update the snapshot. Since `chevronApexAngle`, `chevronSize`, `cornerRadius`, `tipExtension`, etc seem to all be set and used together, I'm fine with a minimal number of `<GestureDiagram>` instances that combine all the new props. No need to test them orthogonally and all a lot of test cases that are not actually used.

fbmcipher:
Closing as superseded by #5320

PR #4465, opened 2026-06-26
Migrate dev server to HTTPS 

fbmcipher:
Resubmission of #4003 + #4143. This version fixes the issues from prior submissions and extends HTTPS support to the interactive agent tooling we developed last month.

> **CI for lint will be red until #4447 merges** due to an OOM error.

## Background

The dev server now runs over HTTPS so development can use secure-context-only web APIs, needed for sync engine work and #3652, and probably others. Dev was previously HTTP-only.

## What changed
- **Dev server** — `yarn start` serves HTTPS by default via `@vitejs/plugin-basic-ssl` (`HTTP=1` opts out); `servebuild` → `vite preview`. 
- **CI** — Puppeteer and iOS Safari run against HTTPS (BrowserStack via a `cloudflared` trusted-cert tunnel)
- **Legacy fallback** — for in-flight PRs, CI still supports HTTP (this is what caused tests in other PRs to fail last time – see below)
- **Capacitor apps** — trust the self-signed dev cert
- **Agent tooling** — the shared Chrome the browser-control skills attach to launches with `--ignore-certificate-errors`

## Why it's safe this time
#4143 broke BrowserStack on *other* PRs because `pull_request_target` runs the base-branch workflow against the PR's checked-out code. Pre-migration PRs serve HTTP, but the HTTPS workflow tried to reach the server at `https://`, causing tests to fail.

Now, a new **"Detect dev server mode"** step greps the checked-out `vite.config.ts` for `@vitejs/plugin-basic-ssl` and uses the matching protocol – so pre-migration PRs run on HTTP and post-migration PRs on HTTPS. 

## Testing

Tested and working:

- Desktop development + preview
- Puppeteer (local Docker + CI)
- iOS Safari E2E (BrowserStack)
- iOS Capacitor app (simulator + real device)
- Android Capacitor app
- Local Appium E2E
- Copilot agent session

All load and work over HTTPS.

## User facing changes

- `yarn start` will now run a HTTP**S** dev server on `https://localhost:3000` (or `3001`, `3002` and so on if there are multiple servers running simultaneously). Any bookmarks or web clips will need to be updated.
- The developer will *specifically* need to enter the leading `https://` in their browser when navigating to the dev server, as most browsers assume `http://` when entering a port number other than 443.
- A security warning will appear the first time the developer accesses the URL, as shown below. **This is harmless** - it's just because `vite-dev-https` uses a self-signed certificate. It's safe to dismiss. Sometimes, you need to reload the browser after accepting the certificate. From this point on, you won't be asked again.

CleanShot 2026-03-29 at 00 11 20@2x


raineorshine (approved):
Amazing, thank you!

Issue #3801, opened 2026-02-13
Visually indicate the remaining hidden gestures with a depth effect


fbmcipher:
[**Link to design on Figma**](https://www.figma.com/design/oQnevfxv79fhTJnOs0fz9k/em-design?node-id=11438-186652&t=uZqEQfNqzw1uHaiJ-4)

> [!NOTE]
> **Revised 4 Aug 2026** after design review. This issue originally specified a persistent footer row to keep "Cancel" and "Command Universe" on screen at all times. That approach has been dropped — see [this comment](https://github.com/cybersemics/em/issues/3801#issuecomment-5187223831) for more information.

---

Sometimes, it's possible that the Gesture Menu returns so many commands that the list of gestures overflows. Here's an example. This is particularly noticeable on smaller phone screens.

<details>
<summary>Example</summary>
<img width="184.5" height="377" alt="Image" src="https://github.com/user-attachments/assets/2f88694a-0079-4a6b-b82b-fc8ea2ad43a1" />
</details>

---

### 1. Cap the number of visible gestures

Rather than simply displaying all gestures and relying on `overflow` to push them off-screen, we should instead **control** the number of visible gestures on-screen. We can work out the available space we have for gestures, divide by the height of an individual gesture, and round down to work out the number of gestures we show at once.

The Gesture Menu is **not** fixed to full height — it sizes to its contents, stays anchored to the top, and shrinks as the command list shortens.

### 2. Visually indicate the remaining hidden gestures using a depth effect

Apply a gradual effect to the last four gestures that results in the gestures feel like they "descend into the fog". This subtly implies that there are more gestures waiting to emerge, and aligns with the Liminal UI design direction. For now, this should only happen when there's only a single column.

<img width="207.5" height="438" alt="Image" src="https://github.com/user-attachments/assets/545aa196-d3f5-4383-a88e-f040f21926db" />

- 1: `scale(0.875)` `opacity: 0.875` `blur(1.5px)`
- 2: `scale(0.85)` `opacity: 0.8` `blur(2px)`
- 3: `scale(0.825)` `opacity: 0.7` `blur(2.5px)`
- 4: `scale(0.75)` `opacity: 0.45` `blur(3px)`

(Note: in this image, Cancel and Command Universe are in a static, sticky header. This is an **old design.** The expected behavior is that they flow in the same list as the rest of the other commands. See [this comment](https://github.com/cybersemics/em/issues/3801#issuecomment-5187223831) for more details.)

---

[...]

PR #4466, opened 2026-06-26
fix(GestureMenu): Trace should above the GestureMenu blur

### LINKED ISSUE #3791: Position gesture trace above blur layer
Originally posted in #3646.

### Steps to Reproduce
1. Swipe → in the gesture zone at the bottom of the screen, pause/hold, and observe the gesture trace.
2. Add ↓ to the gesture input, pause/hold, and observe the gesture trace.
3. Add → to the gesture input, pause/hold, and observe the gesture trace.

### Current Behavior
The gesture trace is rendered beneath the blur layer. Because the blur layer changes height/position, the gesture trace becomes more or less blurry depending on how many commands are shown in the Gesture Menu.

Screenshot 2026-02-11 at 19 12 09

### Expected Behavior
The gesture trace should be positioned *above* the blur layer. It should darken, but not blur. See the image below:

Screenshot 2026-02-11 at 19 15 19

cindmichelle:
Fixes [#3791](https://github.com/cybersemics/em/issues/3791)

## Summary
Reorder the GestureMenu hierarchy so TraceGesture renders above the GestureMenu blur.

### Problem statements
When performing a gesture, the TraceGesture (drawn stroke) appeared blurry over the gesture menu commands instead of staying sharp. 

The root cause is that the gesturemenu's blur and TraceGesture lived in different stacking contexts, so z-index couldn't order them. react-native-web's <View> creates its own stacking context (position: relative; z-index: 0), trapping TraceGesture inside it; the blur lived at root level, and since View is z-index: 0 in root, the entire View (trace included) always painted below the blur.

### Solutions
As discussed in the comments below, we're moving forward with the plan **to move `ProgressiveBlur` one level up so it sits alongside `GestureMenu`'s `PopupBase`, and to reparent `GestureMenu` under `MultiGesture`'s `View`**.

Previously, I planned to extract `ProgressiveBlur` out of `GestureMenu` so it could be rendered under `MultiGesture`'s `View`. But Faiz pointed out that approach would become problematic if we ever decide to make `GestureMenu` dynamically sized again. (FYI based on  https://github.com/cybersemics/em/pull/4466#issuecomment-4907475672, we're making it full-height for now).

And after spending some time understanding the `MultiGesture` component, I think reparenting `GestureMenu` under `MultiGesture` is the cleaner architectural choice. `GestureMenu` is part of the overall gesture UX, where it presents the available gesture actions to the user—so conceptually it belongs alongside the rest of the gesture system.

**Proposed Hierarchy**
```
MultiGesture
├── View (z-index: auto)
│   ├── GestureMenu (mounts when menu is active)
│   │   ├── PopupBase
│   │   └── GestureContentBlur (z-index: gestureContentBlur)
│   ├── TraceGesture (fixed, z-index: gestureTrace)
│   ├── ScrollZone
│   └── div[eventNode]
│       └── Content (z-index: content)
└──
```
With this hierarchy, `ProgressiveBlur` remains part of `GestureMenu` while sitting alongside its PopupBase, avoiding future constraints if GestureMenu becomes dynamically sized again. It also gives us a clean way to control the rendering order, allowing TraceGesture to render above the blur while keeping the menu content on top.

## Changes
- **`MultiGesture.tsx`** — set `zIndex: 'auto'` on `<View>` to dissolve its stacking context; mount `<GestureMenu />` inside the `<View>`, alongside `<TraceGesture>`, `<ScrollZone>`, and the content div.
- **`AppComponent.tsx`** — removed the root-level `<GestureMenu />` mount; it now lives inside `MultiGesture`.
- **`GestureMenu.tsx`** (moved to `src/components/GestureMenu/`) — renders `GestureContentBlur` as a sibling of `PopupBase` (not a child), so its `gestureContentBlur` z-index is ordered in the shared `<View>` stacking context rather than trapped inside `PopupBase`'s higher `popup` stacking context. Blur opacity is driven off `durations.fast` so it fades in sync with the menu's `FadeTransition`. `willChange: 'opacity'` keeps the compositor layer alive so Android WebView doesn't drop the subtree for a frame at fade-out, which was flashing the blur through the menu.
- **`GestureContentBlur.tsx`** (new) — wraps the shared `ProgressiveBlur` component (replacing the menu's old dedicated blur), pinned to the full viewport (`position: fixed; inset: 0`) with a `minBlur` floor so the blur reaches the bottom of the screen instead of decaying to 0 partway down.
- **`constants.ts`** (new) — shared sizing values (padding, row gap, bottom tail) used by both `GestureMenu` and the `FadeTransition` wrapper's `paddingBottom`, so they stay in sync.
- **`GestureMenuItem.tsx`** — moved to `src/components/GestureMenu/`.
- **`panda.config.ts`** — adds a `gestureContentBlur: 23` z-index token, between `content (2)` and `gestureTrace (24)`.

## Notes
- **`<View>` no longer forms a stacking context** — `zIndex: 'auto'` on `MultiGesture`'s `<View>` was required so the blur and trace could be ordered against each other. `View` previously had `position: relative` and an explicit `z-index: 0`, which makes it a stacking context — a sealed box where z-index only ranks elements *within* it, so from outside the box the whole thing counts as a single layer at rank 0 regardless of what z-index its children use internally; `gestureContentBlur: 23` inside `View` meant nothing to anything outside it, the same class of bug as the original trace-under-blur issue with a different pair, and it surfaced during testing on Capacitor iOS and Safari where `NavBar` (root-level) was painting above the blur (trapped inside `View`) for exactly this reason, which is what led us to this fix. Setting `zIndex: 'auto'` removes that box, so `View`'s children (`Content`, `GestureMenu`/blur, `TraceGesture`, `ScrollZone`) now promote directly into the root stacking context and compete there using their own z-index tokens against everything else at root, including `NavBar` — the side effect being that the sealed box was also providing isolation, so anything added inside `View` in future no longer stays contained and now ranks directly against every other root-level z-index in the app; regression-checked the affected children (`TraceGesture`, `ScrollZone`, `Content` still layer and render correctly) and `ProgressiveBlur`'s other consumers (`Tip`, `Sidebar`, which don't live inside `View`) are unaffected.

### Trace now above the blur 
| Platform          | Screenshot                                                                                          |
| ----------------- | --------------------------------------------------------------------------------------------------- |
| Safari iOS        |  |
| Capacitor iOS     |  |
| Capacitor Android |   |


fbmcipher:
First off, my visual review – looks great! Exactly what I was after. 🙏

My concern is architectural. Now that the blur lives outside the menu, it has to calculate the menu's height itself (`commands.length` × constants) instead of just inheriting it the way it used to. My worry is if someone ever needs to change the menu's composition (header spacing, a new row type, whatever), the calc won't necessarily follow, and the blur stops lining up. This kind of coupling gives some me pause, and it starts to blur the responsibilities of individual components. The blur really feels like it should live in the GestureMenu, not somewhere else.

Here's what I'm thinking... the blur used to size itself for free – `height: 100%` of the menu. We only lost that because it had to move out to sit under the trace. What if we leave the blur where it naturally sizes itself, and move the *trace* instead? We'd get the layering we want *and* the blur goes back to inheriting its height: no calc, nothing to drift, and no `ResizeObserver` needed (which I know is off the table on perf grounds anyway).

I know the catch is untangling `TraceGesture` from `MultiGesture` – you flagged the shared ref, and that's the real work here, I'm genuinely not pretending it's trivial. But it's also the thing Raine explicitly greenlit in the issue:

> You can pull TraceGesture out of the MultiGesture component if needed. I'd rather a more decoupled hierarchy than portals.

And it delivers on the flexibility I'd asked for there too – being able to change the layer order later without surgery each time:

> It would be great to implement this in a way that allows for future flexibility – in case we later decide that the TraceGesture needs to be layered differently.

What do you think? It might be worth a chat with AI to see how feasible this is. If this feels right, I'm happy to have a think with you about how to structure the refactor with you if that's useful. If there's a reason we shouldn't take this route please also lmk!

cindmichelle:
@fbmcipher btw, after looking at #3801, I just realized that the GestureMenu background supposed to be full-screen height. However, I only able to find one design in Figma looks like this, could you confirm the Design 2 below is the final design?

| Design 1          | Design 2                                                                                          |
| ----------------- | --------------------------------------------------------------------------------------------------- |
| image       | image |


I also had a chat with Claude, and one concern about separating `MultiGesture` and `TraceGesture` is that we're splitting what is currently a cohesive component. Right now, `MultiGesture` is responsible for detecting the `PanResponder`, while `TraceGesture` is responsible for visualizing the gesture trace. If we move `TraceGesture` outside, we'd need to create `eventNodeRef` at the `AppComponent` level so both `MultiGesture` and `TraceGesture` can share the same ref.

And after spending more time understanding the `MultiGesture` component, I'm starting to think it might make more sense to re-parent `GestureMenu` under `MultiGesture`, since `GestureMenu` is should be part of the overall gesture UX. It's responsible for presenting the set of actions users can perform with gestures, so conceptually it feels closely tied to the gesture system itself.

The only downside I see is that it gives `MultiGesture` another responsibility, which could reduce its cohesion a bit. That said, I think the UX relationship between `MultiGesture` and `GestureMenu` makes this a reasonable tradeoff.

If we go with that reparenting GestureMenu approach, my plan would be to structure it like this. Based on Claude observations, this seems feasible and shouldn't introduce any regressions compared to what I've implemented so far.

```text
MultiGesture
├── View (z-index: auto)
│   ├── GestureMenu (mounts when menu is active)
│   │   ├── PopupBase
│   │   └── ProgressiveBlur
│   ├── TraceGesture (fixed, z-index: gestureTrace)
│   ├── ScrollZone
│   └── div[eventNode]
│       ├── Content (z-index: content)
│       └── GestureContentBlur (z-index: gestureContentBlur)
└──
```

I think the re-parenting approach is a better fit than refactoring `TraceGesture` out of `MultiGesture`, but I'd like to hear your thoughts as well

fbmcipher:
> I also had a chat with Claude, and one concern about separating MultiGesture and TraceGesture is that we're splitting what is currently a cohesive component. Right now, MultiGesture is responsible for detecting the PanResponder, while TraceGesture is responsible for visualizing the gesture trace. If we move TraceGesture outside, we'd need to create eventNodeRef at the AppComponent level so both MultiGesture and TraceGesture can share the same ref.

I hear you here. There's something elegant in having the gesture handler (MultiGesture) and the visual (TraceGesture) co-located.

> And after spending more time understanding the MultiGesture component, I'm starting to think it might make more sense to re-parent GestureMenu under MultiGesture, since GestureMenu is should be part of the overall gesture UX. It's responsible for presenting the set of actions users can perform with gestures, so conceptually it feels closely tied to the gesture system itself.

In principle this reparenting sounded like the right idea, and I nearly wrote a message agreeing to it. But look at the proposed hierarchy:

CleanShot 2026-07-04 at 03 33 12@2x

Each of the layers used to render the gesture menu is now in a different subtree. This makes the code hard to read and reason about and relate to the gesture menu visually. In this situation, sure we protect the coupling between MultiGesture and TraceGesture – but at the expense of the clarity of the gesture menu.

(AI's take, so take this with a pinch of salt: it also wouldn't actually fix the blur-height problem. The reparent only moves *where the menu mounts* — it leaves `GestureContentBlur` sitting in a different box from the command list. The blur can only size itself *for free* if it lives in the **same box** as the commands, so CSS hands it their height. Moving the mount point doesn't do that — so the blur still has nothing to inherit from, and still has to compute its height from the command count. Same formula, same drift. Which is exactly why I *really* don't want to keep manually calculating the blur height – refer to my comment above.)

---

Thanks for this discussion though – it's genuinely helping me pin down what we *do* want. The way I see it, our constraints are:

- ✅ **Don't manually calculate the blur's height** — let the blur *reflow* naturally: it lives in the same box as the commands and takes their height via CSS (`inset: 0`), so there's no command-count formula and nothing to drift. Bonus: it's pure CSS — **no `ResizeObserver`**, no JS at all.
- ✅ **Colocate the gesture visuals** – trace, blur and menu in one clear place. (This is your instinct, and I think it's right)
- ✅ **Keep the layer order flexible** — remember I flagged wanting to easily re-layer the trace. From a software design perspective it is important (especially while we're testing a brand-new design for the first time) that we build our code with _optionality_ so we can move the trace over/under later by changing basically one value, rather than a complete rearchitecture.

Both approaches below are really the *same architecture*: one gesture layer where trace / blur / glow / commands are _flat siblings_ ordered by z-index tokens. That's what buys us both the reflow **and** the easy re-layering.

Unfortunately I think the only way to do this is to move TraceGesture *out* of the MultiGesture. I know you had objections with this approach – please do let me know once you've had a read.

## 1. Separate TraceGesture & MultiGesture, use ministore for the ref

Instead of a ref in AppComponent, MultiGesture publishes the eventNode into a properly-scoped **gesture store**, and TraceGesture reads it from there. It's genuinely not that unusual to use a store (or React.Context) to centralize a shared ref.

```
AppComponent
├─ MultiGesture                 (recognizer — unchanged)
│  └─ View
│     └─ div[eventNode]          → writes the node into gestureStore
│        └─ Content
│
└─ GestureLayer                  position: fixed  (one stacking context)
   ├─ blur           z-token     (reflows: inset:0 of the command box)
   ├─ glow / overlay  z-token
   ├─ TraceGesture    z-token     ← reads the node from gestureStore
   └─ commands        z-token
```
*MultiGesture and GestureLayer are just siblings; the store is the (invisible) link.*

**Strengths**
- Matches the decentralized instinct — separate components, coordinated by a store.
- Flattest tree — no wrapper; the two are siblings.
- Fits em's grain — coordinating through ministores is the house pattern.
- Sidesteps the AppComponent ref you didn't want.

**Weaknesses**
- First store to hold a live *DOM node* in em — potentially defensible, but worth a comment explaining why.
- The coupling is *implicit* — nothing in the tree shows the two are linked; you'd have to know the store connects them. Would probably want to document this as a comment too.

## 2. Common ancestor

A small, purpose-built coordinator owns the eventNode ref (a normal `useRef`) and hands it to both the recognizer and the gesture layer — **not** AppComponent, a dedicated coordinator whose whole job is exactly this.

```
AppComponent
└─ GestureCoordinator            renders a fragment · owns eventNodeRef = useRef()
   ├─ MultiGesture               (recognizer)          ← eventNodeRef
   │  └─ View
   │     └─ div  ref={eventNodeRef}
   │        └─ Content
   │
   └─ GestureLayer               position: fixed       ← eventNodeRef
      ├─ blur           z-token  (reflows: inset:0 of the command box)
      ├─ glow / overlay  z-token
      ├─ TraceGesture    z-token  ← reads eventNodeRef.current
      └─ commands        z-token
```
*The coordinator owns the ref and hands it to both; it renders no DOM of its own, so it adds no wrapper element and no stacking context.*

**Strengths**
- Plain React ref — no store novelty, no DOM-in-a-store.
- The coupling is *explicit* — the coordinator visibly wires the two together, so a future dev sees the relationship in the tree.
- Textbook React ("lift the ref to the nearest common ancestor").

**Weaknesses**
- One extra level of nesting — a touch less flat.
- Slightly more centralized (both branches under one parent) — a hair against the decentralized instinct.

---

So, how do you feel about these two? I'm genuinely open here — if there's something I've missed, a cleaner third path, or a reason to lean a different way entirely, I'd really like to hear it. And feel free to run this past your agent too, so it can weigh my priorities against yours — I'd love to land somewhere that works well for both of us. 🙂

cindmichelle:
@fbmcipher Apologies, I was a bit hallucinating when I created the proposed hierarchy. Now I understand why the previous hierarchy was hard to accept, since it was still separating the blur from the `GestureMenu`, which doesn't really solve the issue around the blur and the menu 🤦‍♀️

**In the corrected proposed hierarchy**, `GestureContentBlur` would replace the original `ProgressiveBlur` under `GestureMenu`, since they both serve the same purpose. So it should look like this:

image


With this proposed hierarchy, we no longer need to split the blur and `GestureMenu`, and we also don't need to refactor `TraceGesture` out of `MultiGesture`. We should still be able to render the trace above the blur as long as the z-index of `TraceGesture > blur`. This also keeps the layer ordering flexible—we'd only need to adjust the z-index values if we ever want to change it.

And thanks for the suggestions on how we could separate `TraceGesture` from `MultiGesture` without cluttering `AppComponent ` ! That said, I'd still prefer to avoid refactoring `TraceGesture` out of `MultiGesture` if possible. Splitting the multi-gesture logic and its visual representation would make the debugging process for MultiGesture a bit harder  for someone less familiar with the codebase. Especially since we're planning to move the trace to  live under the menu, it could make that relationship harder to follow.

But I'd like to hear your thoughts on the _corrected hierarchy proposal_. Do you think it still makes sense? If not, I can go with the common ancestor approach instead.

---

By the way, regarding the blur height resizing, could you respond to my question from my earlier post? Asking this because if we decide not to make the blur height dynamic for the final design, I can remove that requirement from this PR implementation as well. The other issue  mentions that we're planning to move away from the fixed full-height behavior.

> @fbmcipher btw, after looking at https://github.com/cybersemics/em/issues/3801, I just realized that the GestureMenu background supposed to be full-screen height. However, I only able to find one design in Figma looks like this, could you confirm the Design 2 below is the final design?

image

fbmcipher:
@cindmichelle Hey, no worries about the hallucination, it's easily done. Just be sure to rigorously test your agent's assumptions. I've noticed *especially* with issues of architecture/composition, agents can still really struggle – even high-powered ones like Opus/Fable.

LLMs are often falsely confident and we need to verify their results by asking them lots of questions, challenging their assumptions with our knowledge, and reasoning through its responses for ourselves.

---

> `GestureContentBlur` would replace the original `ProgressiveBlur` under GestureMenu, since they both serve the same purpose.

To be clear, in your original post you said that `GestureContentBlur` *uses* `ProgressiveBlur`, so I'm guessing this is a wrapper for `ProgressiveBlur` that *replaces* the existing `ProgressiveBlur`, rather than a new blur implementation that replaces `ProgressiveBlur` entirely? Just want to be precise.

---

Of the two architectures, I honestly think either is fine as long as it meets our requirements. I was at first a little unsure about moving the entire GestureMenu *into* MultiGesture, but I see now that the `ScrollZone` is also already there, so agreed – it would be neat if all gesture-related UX was in one tree.

As long as we don't have to manually calculate the blur's height, and the layer order can be flexible, I'm happy with either.

---

Sorry, missed this in my last reply! We probably want to keep the Gesture Menu full height, since we changed the design in #3801.

cindmichelle:
@fbmcipher Thanks for confirming. I'll move forward with the reparenting approach for now unless we run into a showstopper. But from my current perspective, I don't see any issues with this approach, but we'll find out as I implement it.

----
  
> To be clear, in your original post you said that GestureContentBlur uses ProgressiveBlur, so I'm guessing this is a wrapper for ProgressiveBlur that replaces the existing ProgressiveBlur, rather than a new blur implementation that replaces ProgressiveBlur entirely? Just want to be precise.

Correct. `GestureContentBlur` is just a wrapper around the `ProgressiveBlur` (from `/components`) to make the blur height dynamic. But since you've confirmed that `GestureMenu` will be full-height in the final design, we should be able to remove that wrapper and use `ProgressiveBlur` directly.

You may also notice that `GestureMenu` previously had its own `ProgressiveBlur`. However, since we want the blur to apply only to components with a lower z-index (e.g. `LayoutTree`, `Navbar`, etc.), it makes more sense to use the shared `ProgressiveBlur` from `/components` going forward. The original `ProgressiveBlur` component uses `backdrop-filter: blur()`, which blurs everything that has already been rendered behind it. With the previous approach, that meant the `Trace` would always end up being blurred as well.

src/components/GestureMenu/constants.ts
fbmcipher, line 1: I think this file is a leftover from the previous approach since each constant is only used once. We can probably move these values back into the Panda `css()` expressions and delete this file.

src/components/GestureMenu/GestureMenu.tsx
```diff
@@ -61,10 +63,10 @@ const GestureMenu: FC<{
...
         {gestureInProgress && (
           <div
-            className={css({
-              padding: '2.25rem',
-              paddingTop: !isBrowser ? '0.75rem' : undefined,
-            })}
+            style={{
+              padding: `${GESTURE_MENU_PADDING_REM}rem`,
+              ...(!isBrowser ? { paddingTop: '0.75rem' } : null),
+            }}
```
fbmcipher: Related to my comment about removing `constants.ts`:

I see you had to break styles out of css({}) to get dynamic values to work.

Wherever you have done this, it would be great to move these style expressions back into `css()` so that we get back the benefits of PandaCSS, like static extraction and type-safety.

src/components/GestureMenu/GestureMenuItem.tsx
```diff
@@ -99,6 +99,8 @@ const GestureMenuItem: FC<{
                 ? '0px 0px 24px rgba(255, 255, 255, 0.8), 0px 0px 12px rgba(255, 255, 255, 0.6)'
                 : undefined,
           })}
+          // Sourced from GestureMenu/constants so the content-blur height formula stays in sync.
```
fbmcipher: Now that we've pivoted away from the content-blur height formula mechanism, we should remove any references to it.

src/components/GestureMenu/GestureContentBlur.tsx
```diff
@@ -0,0 +1,62 @@
...
+    >
+      <ProgressiveBlur
+        // Top-weighted falloff, but minBlur keeps every layer blurred so the effect reaches the
+        // bottom of the viewport instead of hitting 0 partway down.
+        direction='to bottom'
+        maxBlur={8}
+        minBlur={3}
+        opacity={blurOpacity}
+      />
+    </div>
```
fbmcipher, line 65: Not an issue in your code, just something I want to point out when using `ProgressiveBlur` that I recently learned.

`ProgressiveBlur` has a very different impact on performance between iOS (or WebKit-based browsers) and Android (or Chromium based browsers.)

`ProgressiveBlur` works by rendering *layers* of blur that gradually increase/decrease in intensity, and feathers between them. The more layers there are, the smoother that transition is – but at the cost of performance. This can be controlled with the `layers` prop.

Android has a much lower tolerance for how many concurrent layers can be rendered until incurring performance penalty. I set the `layers` prop to `1` on Android, and `4` for iOS.

I'll open an issue to set the default to a more reasonable level, but for now just specify the layer count to make sure we don't run into performance issues:

```typescript
     <ProgressiveBlur
     ...
         layers={isAndroid ? 1 : 4}
     />
```

src/components/GestureMenu/GestureMenu.tsx
```diff
@@ -147,30 +149,6 @@ const GestureMenu: FC<{
...
-        position: 'absolute',
-        backdropFilter: 'blur(5px)',
-        mask: 'linear-gradient(180deg, {colors.black} 0%, {colors.bgOverlay80} 80%, {colors.bgTransparent} 100%)',
-        width: '100%',
-        top: 0,
-        height: '100%',
-      })}
-      style={{
-        // Use ease-out on enter so the blur appears immediately, and easeInSlow on exit so it lingers before fading.
-        transition: `opacity ${token('durations.fast')} ${animationState === 'exiting' ? token('easings.easeInSlow') : 'ease-out'}`,
```
fbmcipher, line 167: Noticed that the exit transition changed from `easeInSlow` to `fast`. It looks okay to me, but just wondering if there was an intention behind the change?

src/components/GestureMenu/GestureContentBlur.tsx
```diff
@@ -0,0 +1,62 @@
...
+    // Match the menu content's FadeTransition type='fast' (durations.fast, ease-out) so the blur
+    // and menu fade in/out together. durations.get returns ms (0 in e2e tests); motion wants seconds.
+    const controls = animate(blurOpacity, animationState === 'visible' ? 1 : 0, {
+      duration: durations.get('fast') / 1000,
+      ease: 'easeOut',
+    })
+    return controls.stop
+  }, [animationState, blurOpacity])
+
+  if (animationState === 'hidden') return null
```
fbmcipher, line 39: I appreciate the work here to animate the opacity of `ProgressiveBlur` to workaround the WebKit issue. It looks great on iOS, but on Android it looks like the blur flashes in on a single frame rather than fading in:

https://github.com/user-attachments/assets/f763cf15-2a4c-4509-85bc-98d334804829

I'm not sure why this happens, but I do know from my other Liminal UI work that there are a lot of issues like this – things behave one way on iOS/WebKit and another very different way in Android/Chrome. Worth looking into.

fbmcipher (changes requested):
Thank you! I've left some detailed notes, but this is getting close!

src/components/GestureMenu/GestureContentBlur.tsx
cindmichelle: Looks like this was happening because the blur was too expensive to render with the default `layers` prop =`8`. After reducing it based on your suggestion (`4` on iOS and `1` on Android), the blur animation no longer stutters.

src/components/GestureMenu/GestureMenu.tsx
cindmichelle: It was because `easeInSlow` isn't available in `motion/react`. Although we can mimic it with a cubic Bézier curve, the current `ease-out` fade-out animation already feels smooth enough compared to the `main` implementation using PandaCSS's `easeInSlow`. 


Here's the side-by-side comparison (I also attached a version of this branch using the custom cubic-bezier curve as an additional comparison, if needed.)


https://github.com/user-attachments/assets/d15d8921-bbfd-4fcd-a1f6-4dfd2a79391c

fbmcipher: Thank you for the comparison @cindmichelle, really helpful! It's not a huge deal either which way but just felt it was something I should flag up.

Since you went through the effort of implementing it, we might as well go with the custom Bézier curve :)

We get the benefit of consistency with main, too, which is nice.

src/components/GestureMenu/GestureContentBlur.tsx
fbmcipher: Great news :)

cindmichelle:
Hi @fbmcipher , it;s ready for review again!

fbmcipher:
Thanks @cindmichelle, both design and code are looking great. There's just one last thing.

In this PR, the Gesture Menu went from changing height as the list filtered to being **full-height**. That was the right decision, and it looks much better for it.

However, there is one layer that still flows with the list rather than remaining full height. You can see it when the list shortens. It's the dark shadow that moves up as the list clears, while the glow background and blur layer remain static:

https://github.com/user-attachments/assets/8570a1c1-9e88-433e-9489-97ee389118cd

That layer is `Overlay` in `src/components/GestureMenu/GestureMenu.tsx`. It's a sibling of `Glow`, but unlike `Glow` it's set to `height: 100%` – and because it sits inside the wrapper that the command list sizes, its height ends up tracking the list.

Three small changes will sort it out. Because it's so small I might as well hand you the diff!:

```diff
-        background: 'linear-gradient(180deg, {colors.black} 0%, {colors.bgOverlay80} 70%, {colors.bgTransparent} 100%)',
+        background: 'linear-gradient(180deg, {colors.black} 0%, {colors.bgOverlay80} 60%, {colors.bgOverlay50} 100%)',
         top: 0,
         width: '100%',
-        height: '100%',
+        height: '100dvh',
```

`height: 100dvh` is the real fix – it anchors the layer to the viewport the same way `Glow` already is, so the list re-filters inside a backdrop that never moves. The change in background just tweaks the proportions of the gradient a little bit so that the gestures above the overlay are always visible.

---

Once you've made that change, go ahead and hand over to Raine for her final review before merge :) Thank you for all your work on this issue!

fbmcipher (changes requested):

raineorshine (approved):
Looks good, thanks! 🙏

PR #4467, opened 2026-06-26
Retarget focus to editable after autocomplete

### LINKED ISSUE #4222: [iOS] Gesture does not work near caret after autocorrect
## Steps to Reproduce

1. Type `Adf` (or any misspelled word)
2. Press space to accept the autocorrect to `All`
3. Start a gesture in the red box (below)

image

image

## Current Behavior

Nothing happens. 

## Expected Behavior

Gestures should work as usual. 

## Notes

Note that this is an issue both when the autocorrect menu is open (Step 1) and after it is accepted with space (Step 2). It is not an issue if the autocorrect menu is closed by tapping on the ✗.

Possibly related to `isNear` but not confirmed.

ethan-james:
Fixes #4222 

After using the native autocomplete spacebar-to-complete functionality in iOS Safari, a dead zone appears around and underneath the text as shown in the description of #4222. No touch events reach the app within this area. In order to dismiss this dead zone, it is necessary to synchronously shift focus away from the editable and then return it to the editable.

## Root Cause
**Confidence:** low

Copilot claims that:

> UIKit registers an "undo autocorrect" gesture recognizer at the native layer (above WebKit JS dispatch). Taps near the replaced word are consumed by UIKit — they never enter the DOM at all — explaining why even capture-phase document listeners received nothing.

There is definitely a dead zone created at the native layer, but I don't think that it necessarily follows that it is created by an "undo autocorrect" gesture recognizer. See point 2 in complications below.

## Complications

1. Moving focus away from the thought causes `onBlur` to flush `throttledChangeRef` with an outdated value
**Solution:** Manually update `oldValueRef.current` and call `throttledChangeRef.current.flush()` within the `onAutocompleteInput` handler

2. Native undo (three-finger swipe or shake) is overridden when manually updating value
**Solution:** The autocomplete arrives as two separate events: `insertReplacementText` followed by `insertText` consisting of a trailing space. The focus retargeting must be delayed until `insertText` is received in order to preserve native undo.
  a. I wouldn't have noticed this if Copilot hadn't mentioned the "undo autocorrect" gesture recognizer, but three-finger swipe can be performed anywhere, not just in the area surrounding the text.
  b. Copilot swears that `insertReplacementText` will not always be followed immediately by `insertText`, and so it is necessary to reset the handler after `AUTOCOMPLETE_SPACE_WINDOW_MS = 250`. I have not seen any case where only `insertReplacementText` arrives.

### DIFF
--- a/src/components/Editable.tsx
+++ b/src/components/Editable.tsx
@@ -339,35 +339,37 @@ const Editable = ({
   }, [])
 
   useEffect(() => {
-    if (!isTouch || !isSafari()) return
+    if (!isTouch || !isSafari() || !contentRef.current) return
+
+    const editable = contentRef.current
 
     /** After iOS autocomplete (insertReplacementText) accepts a word, no touch events reach the DOM
      * in a "dead zone" beneath the word until focus is retargeted. Moving focus to the asyncFocus dummy input
      * and back to the previous active element allows touch events to reach the DOM again.
      */
     const onAutocompleteInput = (e: Event) => {
-      if (!contentRef.current || !(e instanceof InputEvent) || e.inputType !== 'insertReplacementText') return
+      if (!editable || !(e instanceof InputEvent) || e.inputType !== 'insertReplacementText') return
 
       const savedCharOffset = selection.offsetThought() ?? selection.offset() ?? 0
 
       // Under normal circumstances, iOS autocomplete triggers an insertReplacementText event followed by an insertText event with a space.
       // It is possible to track both events and apply the fix only when the space is inserted, but it is simpler to add the space directly.
-      oldValueRef.current = contentRef.current.textContent + ' '
+      oldValueRef.current = editable.textContent + ' '
 
       // Queue and flush the change with the space-inclusive value to ensure it's captured before the editable blurs.
       throttledChangeRef.current(oldValueRef.current, { rank, simplePath })
       throttledChangeRef.current.flush()
 
       asyncFocus({ force: true })
-      contentRef.current.focus({ preventScroll: true })
+      editable.focus({ preventScroll: true })
 
       // Restore the selection offset to the character after the inserted word, and accounting for the space added above
-      queueMicrotask(() => selection.set(contentRef.current, { offset: savedCharOffset + 1 }))
+      queueMicrotask(() => selection.set(editable, { offset: savedCharOffset + 1 }))
     }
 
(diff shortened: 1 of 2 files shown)


src/components/Editable.tsx
```diff
@@ -337,6 +338,58 @@ const Editable = ({
...
+
+      const savedCharOffset = selection.offsetThought() ?? selection.offset() ?? 0
+
+      // Queue and flush the change with the browser-applied value to ensure it's captured before the editable blurs.
+      oldValueRef.current = editable.textContent || ''
+      throttledChangeRef.current(oldValueRef.current, { rank, simplePath })
+      throttledChangeRef.current.flush()
+
+      asyncFocus({ force: true })
+      editable.focus({ preventScroll: true })
```
raineorshine: @ethan-james Will `selection.set` trigger a focus anyway? If so, the explicit `focus` call may not be necessary.

raineorshine (approved):
Thanks, looks good.

@BayuAri Some behaviors to test. Once the original issue is confirmed solved, the main regression you are looking out for is messed up caret offset after insert or undo.

- Original issue #4222
- Mobile Safari and iOS Capacitor
- Shake to undo
- Three-finger swipe to undo/redo
- Paste into the middle of a thought
- Autofill insert text into the middle of a thought

The reason I think it would be a good idea to test Paste and Autofill is that they are other ways to insert multiple character at once.

Thanks!

Image

src/components/Editable.tsx
ethan-james: Yes, I think you're right. It needs `preventAutoscroll` to prevent a layout shift, but that's probably better because it can be removed once #3765 is completed.

BayuAri (commented):
Original issue confirmed fixed.

Test cases:

- [x] Draw gesture after applying auto correct with space with Predictive Text is ON
- [x] Draw gesture after  applying auto correct with space with Predictive Text is OF
- [x] Applying three-finger swipe to undo or redo
- [x] Applying  Shake to undo or redo
- [x] Applying insert text in the middle of thought text.
- [x] Applying Autofill to insert text into the middle of a thought

Tested on Mobile Safari and iOS Capacitor. 

Caret and focus are working as expected.

Issue related with Three Finger Swipe and Shake Device to undo/redo is logged as #4477  and #4476

raineorshine:
This bug has been driving me crazy for months in my personal use of em 😵‍💫

Thanks @ethan-james for the fix 😊

PR #4475, opened 2026-06-29
Filter out current thought when comparing new value in alpha sort

### LINKED ISSUE #3983: Red Sort icon when create new thought during ↑Alphabetically Sorting is active
## Step to Reproduce
1. Create Thought One
2. Tap Sort button
3. Select Alphabetically
4. Select ↑Alphabetically
5. Create Thought Two

## Current behavior
Sorting icon turns red

Image

**Video:**

https://github.com/user-attachments/assets/de0f4b3e-ee10-4977-a20e-551f35db4f37

## Expected behavior
Sorting icon should not turn red

ethan-james:
Fixes #3983

When `editThought` actions run, [getSortedRank](https://github.com/ethan-james/em/blob/f12839fc892598b9f86ddd9c11c322cf5c9a5610/src/actions/editThought.ts#L147) is called with the `newValue` of the thought. However, the thoughts that are compared come straight from the Redux store, and that includes the `oldValue` of the thought that is currently being edited. If it was a new thought, then it was ranked beneath existing thoughts even though `''` should have a lower rank than thoughts with values.

As a result, `getSortedRank` looks at a list of children that contains:

```
    {
        "childrenMap": {},
        "created": 1782682821881,
        "id": "pquWlPm4qXF_U",
        "lastUpdated": 1782682821881,
        "parentId": "__ROOT__",
        "rank": 0.5,
        "updatedBy": "lzVg1WYMcoAV2H8vsvOqVuhvT2rNBw+1ZJcevkPfeUY=",
        "value": ""
    }
```

instead of:

```
    {
        "childrenMap": {},
        "created": 1782682962580,
        "id": "tqbo5EgeTi6Ab",
        "lastUpdated": 1782682962580,
        "parentId": "__ROOT__",
        "rank": -1,
        "updatedBy": "lzVg1WYMcoAV2H8vsvOqVuhvT2rNBw+1ZJcevkPfeUY=",
        "value": ""
    }
```

as it would be if the thought were sorted during `newThought`. When `editThought` tries to put the edited thought [in between its neighbors](https://github.com/ethan-james/em/blob/f12839fc892598b9f86ddd9c11c322cf5c9a5610/src/selectors/getSortedRank.ts#L20), it gets assigned a rank of 0.25 instead of the correct value, -0.5.

## Solution

`getSortedRank` should not be comparing a thought against its previous value, so I added a `staleId` option so that it can filter out the previous version of the thought.

## Alternate Solution

It also would work if the thought were sorted during `newThought` and had its proper rank applied at that time. If that would be a desirable UI change, I can switch to that solution.


raineorshine (approved):
Thanks, looks good!

> It also would work if the thought were sorted during `newThought` and had its proper rank applied at that time. If that would be a desirable UI change, I can switch to that solution.

It's intentional that we avoid sorting new empty thoughts until the user starts typing. This ensures that New Subthought initially insert the thought at the end of the list and New Subthought Above initially inserts the thought at the beginning of the list, matching the user's expectations of those commands.

ethan-james:
> It's intentional that we avoid sorting new empty thoughts until the user starts typing. This ensures that New Subthought initially insert the thought at the end of the list and New Subthought Above initially inserts the thought at the beginning of the list, matching the user's expectations of those commands.

Cool, I figured as much, although we did change that behavior for Created sort in #3927.

raineorshine:
Good observation, Created is an exception to that rule. Since the `created` timestamp doesn't change, I felt that the order of thoughts sorted by Created should not change either. Created sort is the only one that is insensible to typing. This invariant thus takes priority over the ordering semantics of New Subthought and New Subthought Above.

BayuAri (commented):
Issue is fixed.

PR #4482, opened 2026-06-30
Failing browserstack test before main merge

ethan-james:
References #4394 and #4291

#4407 is the PR to merge

This is the baseline for the failing test before `main` is merged into the branch and fixes it.


src/e2e/iOS/helpers/tap.ts
```diff
@@ -5,6 +5,8 @@ import type { Element } from 'webdriverio'
 interface Options {
   // Where in the horizontal line (inside) of the target node should be tapped
   horizontalTapLine?: 'left' | 'right'
+  // Pointer type to use for the tap action. Defaults to 'mouse'.
+  pointerType?: 'mouse' | 'touch'
```
raineorshine, line 9: I understand that `pointerType: 'mouse'` solves a problem, but I'm concerned about the consequences of doing this in a mobile test. We want the iOS Appium tests to behave exactly like a touch device. It seems like setting `pointerType: 'mouse'` on a simulated touch device would create an artificial environment that doesn't reflect a real situation that could be encountered. This would put into question any test that uses this tap helper, because it's not guaranteed that an actual touch device would behave the same way.

raineorshine (commented):
Hi, can you provide guidance on this PR? Is this meant to be merged or something you are using for troubleshooting? Thanks!

ethan-james:
> Hi, can you provide guidance on this PR? Is this meant to be merged or something you are using for troubleshooting? Thanks!

Eventually, I want to merge the test once @fbmcipher and I manage to get it working properly. It should have been in draft the whole time, sorry about that. I don't have a working Browserstack API key, so it's easier to create a PR and do the real testing with CI.

src/e2e/iOS/helpers/tap.ts
ethan-james: Yes, `pointerType: 'mouse'` was hard-coded into the helper prior to this. It would probably be worthwhile to revisit the other tests that use the helper and see if they work with `pointerType: 'touch'` instead.

raineorshine: Ah, I didn't realize we were already doing that in `main`. Okay, something to keep in mind. Seems like a 🚩 at least.

ethan-james:
> > Hi, can you provide guidance on this PR? Is this meant to be merged or something you are using for troubleshooting? Thanks!
> 
> Eventually, I want to merge the test once @fbmcipher and I manage to get it working properly. It should have been in draft the whole time, sorry about that. I don't have a working Browserstack API key, so it's easier to create a PR and do the real testing with CI.

Although, actually, #4407 will be merged, not this one.

PR #4498, opened 2026-07-02
Disable virtual keyboard while formatSelection is focusing an editable

### LINKED ISSUE #3999: [Mobile Safari] Thought vibrates and then jumps up  when user adds Text Formatting
## Step to Reproduce
1. Create Thought One on Mobile Safari
2. Tap any Text Formatting on the toolbar

## Current behavior
Thought vibrates and then jumps up 
Note: Not reproducible on iOS 18

https://github.com/user-attachments/assets/f161952f-ce0c-47ce-94e1-cc51efadbb2e

## Expected behavior
Nothing should happens.
Refer to what happens when cursor is on other thought on the list.

ethan-james:
Fixes #3999 

## Root Cause

Calling [selection.select](https://github.com/ethan-james/em/blob/a337788e0a3a3e4f94109f26c301b27c0a8df6ee/src/actions/formatSelection.ts#L59) focuses the editable, which starts to open the virtual keyboard on iOS. It's not possible to prevent the focus behavior once it's begun.

## Solution

It is possible to disable the keyboard with `inputmode="none"`.

## Complications

The toolbar still shifted slightly. Similar to #4435, this could be fixed by calling `editable.focus({ preventScroll: true })` before `selection.select`.

## Alternate Solution
**Confidence:** low

Since the keyboard is down, there is guaranteed to be no selection. As a result, it should be simpler than usual to apply formatting to the entire thought without focusing, selecting, and using `document.execCommand`. The low confidence is related to the case where part of the thought already has the desired formatting, and the new formatting path would be responsible for stripping that formatting from part of the thought and applying it to the entire thought. That might be a case that `DOMParser` already handlers, but I haven't tested it.


src/e2e/iOS/__tests__/format.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    // 3. Measure the scroll position of the viewport before formatting.
+    const scrollBefore = await browser.execute(() => window.scrollY)
+
+    // 4. Apply bold by tapping the Bold toolbar icon.
+    const boldButton = await browser.$('[data-testid="toolbar-icon"][aria-label="Bold"]').getElement()
+    await tap(boldButton, { y: 60 })
+
+    // Allow any focus-induced scroll to settle before measuring.
+    await browser.pause(1000)
```
raineorshine: Smooth scrolling is disabled in e2e tests, so any scroll should be completed in a single frame. Default autoscroll should be disabled anyway.

Adding sleeps to the tests slows down the test suite for everyone, so we make a good effort to ensure every delay is justified.

raineorshine (changes requested):
Thanks!

Does this have an associated issue number?

raineorshine:
> Since the keyboard is down, there is guaranteed to be no selection. As a result, it should be simpler than usual to apply formatting to the entire thought without focusing, selecting, and using `document.execCommand`. The low confidence is related to the case where part of the thought already has the desired formatting, and the new formatting path would be responsible for stripping that formatting from part of the thought and applying it to the entire thought. That might be a case that `DOMParser` already handlers, but I haven't tested it.

True, but at the cost of diverging implementations. The hope with `execCommand` is that it would handle the details for us, but I could imagine a larger migration to `DOMParser` if we need more control.

raineorshine:
I'm going to mark this back as Draft until the CI passes. Thanks!

ethan-james:
> I'm going to mark this back as Draft until the CI passes. Thanks!

OK, merging in `main` fixed CI, and I verified in #4499 that the Browserstack test fails on `main`.

src/e2e/iOS/__tests__/format.ts
```diff
@@ -0,0 +1,29 @@
...
+    // 2. Make it the cursor thought but not focused: dismiss the keyboard, which blurs the
+    // editable (isKeyboardOpen → false) while leaving the cursor set on the thought.
+    await hideKeyboardByTappingDone()
+
+    // 3. Measure the scroll position of the viewport before formatting.
+    const scrollBefore = await browser.execute(() => window.scrollY)
+
+    // 4. Apply bold by tapping the Bold toolbar icon.
+    const boldButton = await browser.$('[data-testid="toolbar-icon"][aria-label="Bold"]').getElement()
+    await tap(boldButton, { y: 60 })
```
raineorshine, line 23: @karunkop You may want to add this to #4446.

raineorshine (approved):

raineorshine:
Discovered a regression and reverted in #4606

PR #4502, opened 2026-07-03
Fix Pin All with existing children attributes

### LINKED ISSUE #4443: Pin All does does nothing on thought with existing =children meta attribute
## Steps to Reproduce

```
- A
  - B
    - =children
      - =bullet
        - None
    - C
      - D
    - E
      - F
```

Set the cursor on `A/B/C` and activate Pin All. 

## Current Behavior

`=children` is removed.  

```
- A
  - B
    - C
      - D
    - E
      - F
```

## Expected Behavior

`C` should be pinned open. 

```
- A
  - B
    - =children
      - =bullet
        - None
      - =pin
    - C
      - D
    - E
      - F
```

Ruby95515:
## Summary

Fixes #4443 .

This fixes the underlying generic attribute update behavior instead of special-casing `Pin All`.

Previously, nested attribute operations could treat hidden/meta children as if the parent were empty, causing unrelated meta attributes such as `=children/=bullet` to be removed when adding or deleting `=children/=pin`.

Changes:
- Preserve sibling meta attributes when setting terminal meta attributes.
- Preserve sibling meta attributes when toggling terminal meta attributes.
- Only delete attribute containers when they have no children at all, including meta children.
- Simplify `Pin All` so it directly sets/deletes `=children/=pin/true`.

## Testing

`node .yarn\releases\yarn-4.17.0.cjs lint:tsc` passed
`node .yarn\releases\yarn-4.17.0.cjs eslint src/commands/pinAll.ts src/commands/__tests__/pinAll.ts` passed
`node .yarn\releases\yarn-4.17.0.cjs test src/commands/__tests__/pinAll.ts` passed
`node .yarn\releases\yarn-4.17.0.cjs test` passed

### DIFF
--- a/src/commands/__tests__/pinAll.ts
+++ b/src/commands/__tests__/pinAll.ts
@@ -42,6 +42,40 @@ it('toggle on when there is no =children attribute', () => {
       - g`)
 })
 
+it('toggle on when there is an unrelated =children attribute', () => {
+  store.dispatch([
+    importText({
+      text: `
+        - A
+          - B
+            - =children
+              - =bullet
+                - None
+            - C
+              - D
+            - E
+              - F
+    `,
+    }),
+    setCursor(['A', 'B', 'C']),
+  ])
+
+  executeCommand(pinAllCommand, { store })
+
+  const exported = exportContext(store.getState(), [HOME_TOKEN], 'text/plain')
+  expect(exported).toEqual(`- __ROOT__
+  - A
+    - B
+      - =children
+        - =bullet
+          - None
+        - =pin
+      - C
+        - D
+      - E
+        - F`)
+})
+
 it('toggle on when =children/=pin is false', () => {
   // import thoughts
   store.dispatch([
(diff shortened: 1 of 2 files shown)


Ruby95515:
Hi @raineorshine , Should I fix the failed checks?

raineorshine:
The Puppeteer failure is preexisting. We're working on a fix. The other tests stopped short for some reason, so I triggered them again. I will follow up in the morning.

Ruby95515:
Got it.
Thanks.
I will wait for you.
Should I work on another issue?

raineorshine (changes requested):
https://github.com/cybersemics/em/issues/4443#issuecomment-4894776813

Ruby95515:
Based on the new expected behavior, I pushed the new commit to fix the issue.
Please check and let me know the result.
Thanks.

raineorshine (approved):
Thank you!

Eventually it would be good to generalize this functionality to toggle any deep attribute value, but this is a fine solution for the original issue.

PR #4504, opened 2026-07-03
Compare formatting tags independent of applied order

### LINKED ISSUE #3977: Sorting order is based on the first text formatting applied when the text has multiple text formatting
```
- A
- B
- C
- D
- E
```

## Step to Reproduce
1. Tap Sort - Alphabetically - Descending
2. Add Bold to E and C
3. Verify that E and C are moved to the top of the list as C then E
4. Add Italic to D
5. Verify that D is moved under E
6. Add Bold to D

## Current behavior
Nothing happens to _**D**_
The position is not updated.
The sorting order sees the D as Italic disregard it is being Bold.
Note: When the Italic is removed from D, the thought is moved between C and E accordingly.

Image

**Video:**

https://github.com/user-attachments/assets/b44c54f5-f613-4998-911a-66e6173f86e6

## Expected behavior
D should be moved to between C and E.

The  descending sorting order is as follow
1. Blank thought
2. Bold
3. Italic
4. Underline
5. Strikethrough
6. Metaprogramming
7. Symbol
8. Emoticon
9. Number
10. Alphabet 

For discussion:
When the app handles the multiple text formatting in correct order when the formatting comes in the right order, it starts to be a problem when the formatting is applied not in correct order.
For example, text is applied with text formatting with this order: Strikethrough, Underline, Italic and Bold as the last one; this text will be sorted as Strikethrough as the first applied and will be always loss against any text with Bold formatting as the first being applied.
I think, the text formatting should be sorted during the app giving rank to the text to avoid this issue
@raineorshine 
What do you think?

ethan-james:
Fixes #3977 

Tag matching was only performed on the outermost tag, so its priority was used to determine the sort ranking of the thought. Now, all tags that apply to the first readable character are parsed and the highest-priority tag is used for comparison. Additionally, the number of tags is counted and thoughts are sorted by highest combo formatting down to lowest combo formatting.

## Complications

[normalizeCharacters](https://github.com/ethan-james/em/blob/1b2eae4219895c0caa386160b939c9a745c817f5/src/util/compareThought.ts#L91) was comparing strings with tags included in them, which led to sorting based on invisible characters rather than the visible text of the thought. Now, `stripTags` is called during the `compareReadableText` step.

## Follow Up

1. It seems like [compareFormatting](https://github.com/ethan-james/em/blob/1b2eae4219895c0caa386160b939c9a745c817f5/src/util/compareThought.ts#L176) may have been subsumed by `compareFormattingTagPriority` and it is a potential candidate for removal.
2. As mentioned in https://github.com/cybersemics/em/issues/3977#issuecomment-4870458833, it may be preferable to treat 2+ tags as a single combo stack rather than sorting 4 tags above 3 tags above 2 tags above 1 tag above 0 tags.

### DIFF
--- a/src/e2e/puppeteer/__tests__/sort.ts
+++ b/src/e2e/puppeteer/__tests__/sort.ts
@@ -1,0 +1,47 @@
+import click from '../helpers/click'
+import clickThought from '../helpers/clickThought'
+import command from '../helpers/command'
+import exportThoughts from '../helpers/exportThoughts'
+import paste from '../helpers/paste'
+
+vi.setConfig({ testTimeout: 60000, hookTimeout: 60000 })
+
+/** Returns the visible order of the single-letter thoughts A–E from the exported outline. Strips markdown formatting (e.g. **E**, ***D***) and meta thoughts (=sort, Alphabetical, Desc). */
+const getLetterOrder = (exported: string): string[] =>
+  exported
+    .split('\n')
+    .map(line => line.replace(/[-*\s]/g, ''))
+    .filter(value => ['A', 'B', 'C', 'D', 'E'].includes(value))
+
+// https://github.com/cybersemics/em/issues/3977
+it('a thought formatted with multiple styles is sorted by its highest-priority format (bold) in reverse alphabetical order', async () => {
+  await paste(`
+    - A
+    - B
+    - C
+    - D
+    - E
+  `)
+
+  // set the cursor on a home child so that Sort targets the home context
+  await clickThought('A')
+
+  // Sort - Alphabetically - Descending (None → Alphabetical/Asc → Alphabetical/Desc)
+  await command('toggleSort')
+  await command('toggleSort')
+
+  // Add Bold to E and C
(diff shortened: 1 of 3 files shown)


raineorshine (approved):

raineorshine:
> 1. It seems like [compareFormatting](https://github.com/ethan-james/em/blob/1b2eae4219895c0caa386160b939c9a745c817f5/src/util/compareThought.ts#L176) may have been subsumed by `compareFormattingTagPriority` and it is a potential candidate for removal.

Yes, `compareFormatting` assumed the old binary ordering of any formatting < no formatting. If `compareFormattingTagPriority` handles all that and more, we can switch over (and probably rename to `compareFormatting`).

BayuAri (commented):
The order of multi formatting is based on 
- Bold 
- Italic
- Underline
- Strikethrough

Thought with more number of formatting is placed on the above of the less one. 

**Descending Arrow**
Image

**Ascending Arrow**
Image

PR #4507, opened 2026-07-04
fix(estimate): paginate Everhour project tasks in backfill

raineorshine:
## Problem

The backfill script logged `Found 250 tasks in Everhour project` even though the project (gh:143808059) has far more than 250 tasks. `EverhourClient.getProjectTasks` fetched `/projects/:id/tasks` with no pagination params, and Everhour's tasks endpoint returns at most one page (default/max page size 250). Every task beyond the first page was silently never processed.

## Fix

- **`everhour/client.ts`**: `getProjectTasks` now accepts `page` (1-based) and `limit` (page size, default 250) and appends them as `?limit=…&page=…` query params.
- **`backfill.ts`**: Replaced the fetch-all-then-loop with a page-at-a-time `while` loop. It fetches one page, filters to tasks missing estimates, processes them (respecting the existing `LIMIT` processing cap), then advances to the next page only when needed. Paging stops once the processing limit is reached or a page shorter than `PAGE_SIZE` signals the last page — so we never load every page at once.

Per the request, tasks are fetched and processed one page at a time rather than all up front.

## Verification

- `tsc --noEmit` passes for `scripts/estimate`.
- Existing estimate unit tests pass (14/14 via vitest).
- Prettier applied.

## Notes

Confirmed against the Everhour API docs: the tasks endpoint is an array response paginated via `limit` (max 250) + `page`; a page shorter than the requested `limit` indicates the final page.

### DIFF
--- a/scripts/estimate/.env.example
+++ b/scripts/estimate/.env.example
@@ -1,4 +1,5 @@
-# LIMIT=10 DRY_RUN=true yarn backfill
+# LIMIT=10 PAGE=1 DRY_RUN=true yarn backfill
+# PAGE sets the 1-based Everhour task page to start from (default 1).
 # API Token can be refreshed at https://app.everhour.com/#/account/profile
 EVERHOUR_API_KEY=
 EVERHOUR_PROJECT_ID=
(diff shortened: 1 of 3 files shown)


raineorshine:
@copilot+claude-opus-4.8 resolve the merge conflicts in this pull request

PR #4528, opened 2026-07-06
Fix uncategorize undo save error

### LINKED ISSUE #4425: Uncategorize duplicate of uncle breaks undo: Error: undefined is not an object (evaluating 'R.length')
## Steps to Reproduce

```
- a
- b
  - a
    - c
```

1. Set the cursor on `b/a`. 
2. Uncategorize (↓→)
3. Undo

## Current Behavior

`Error: undefined is not an object (evaluating 'R.length')`

## Expected Behavior

Should undo the uncategorize command without error.

Ruby95515:
## Summary

Fixes #4425.

This fixes a crash/save error when uncategorizing a duplicated uncle thought and then undoing the change.

## Problem

For a tree like:

- a
- b
  - a
    - c

Uncategorizing `b/a` and then undoing could produce:

> Cannot read properties of undefined (reading 'length')

The visible tree operation was mostly correct, but the undo persistence path could enqueue malformed lexeme updates.

## Changes

- Prevent uncategorize from merging a child back into the category being deleted when they have the same value.
- Fix undo/redo push queue reconstruction to persist the full lexeme from state instead of forwarding nested JSON patch leaf values.
- Add regression coverage for:
  - uncategorizing same-valued nested thoughts
  - undoing the duplicate-uncle uncategorize case
  - ensuring undo persistence sends complete lexeme updates
  - ensuring the initialized persistence path does not raise the save-error alert

## Testing

Passed:

- `yarn test src/actions/__tests__/uncategorize.ts src/commands/__tests__/uncategorize.ts src/commands/__tests__/undo-redo.ts`
- `yarn lint:tsc`
- `yarn eslint src/redux-enhancers/undoRedoEnhancer.ts src/actions/uncategorize.ts src/actions/__tests__/uncategorize.ts src/commands/__tests__/uncategorize.ts`

### DIFF
--- a/src/commands/__tests__/uncategorize.ts
+++ b/src/commands/__tests__/uncategorize.ts
@@ -1,9 +1,13 @@
 import { importTextActionCreator as importText } from '../../actions/importText'
+import { undoActionCreator as undo } from '../../actions/undo'
 import { executeCommandWithMulticursor } from '../../commands'
 import { HOME_TOKEN } from '../../constants'
+import db from '../../data-providers/yjs/thoughtspace'
+import { initialize } from '../../initialize'
 import exportContext from '../../selectors/exportContext'
 import store from '../../stores/app'
 import { addMulticursorAtFirstMatchActionCreator as addMulticursor } from '../../test-helpers/addMulticursorAtFirstMatch'
+import expectPathToEqual from '../../test-helpers/expectPathToEqual'
 import initStore from '../../test-helpers/initStore'
 import { setCursorFirstMatchActionCreator as setCursor } from '../../test-helpers/setCursorFirstMatch'
 import uncategorizeCommand from '../uncategorize'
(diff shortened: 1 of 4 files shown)


src/commands/__tests__/uncategorize.ts
```diff
@@ -1,16 +1,113 @@
...
+    expect(exportContext(state, [HOME_TOKEN], 'text/plain')).toEqual(`- ${HOME_TOKEN}
+  - a
+  - b
+    - a
+      - c`)
+    expectPathToEqual(state, state.cursor, ['b', 'a'])
+  })
+
+  it('pushes complete lexeme updates when undoing uncategorize of a duplicate uncle', async () => {
+    const updateThoughtsSpy = vi.spyOn(db, 'updateThoughts').mockResolvedValue(undefined)
```
raineorshine: The data provider is undergoing a major upgrade right now (to a custom sync engine), so unfortunately touching this is off limits.

If there is a way to test this without reaching into the data provider, that would be great. If not, we can skip test coverage on this one.

src/commands/__tests__/uncategorize.ts
```diff
@@ -1,16 +1,113 @@
...
+        setCursor(['b', 'a']),
+      ])
+
+      executeCommandWithMulticursor(uncategorizeCommand, { store })
+      store.dispatch(undo())
+
+      await vi.runOnlyPendingTimersAsync()
+      await Promise.resolve()
+
+      expect(store.getState().alert?.value).not.toContain('not able to save the last change')
```
raineorshine: This isn't a good success criteria. If the error message changes, this test will always pass. I would either mock the error reducer or find another way to check the last error state.

raineorshine (changes requested):
Fix looks good. Just some requests around test architecture.

raineorshine (approved):

BayuAri (approved):
Error is no longer prompted after Undo the Uncategorize with auto merged duplicate parent.

raineorshine:
Thank you!

